PR #184, opened 2018-08-17
Centralize business logic for starting background prefetch.

jrbriggs:
Move the check for unattended enlistment from `InProcessMount` to `BackgroundPrefetcher`.

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/BackgroundPrefetcher.cs
+++ b/GVFS/GVFS.Common/Prefetch/BackgroundPrefetcher.cs
@@ -29,14 +29,14 @@ namespace GVFS.Common.Prefetch
 
             this.prefetchJobThread = null;
 
-            if (gitObjects.IsUsingCacheServer())
+            if (gitObjects.IsUsingCacheServer() && !GVFSEnlistment.IsUnattended(tracer))
             {
                 this.prefetchJobTimer = new Timer((state) => this.LaunchPrefetchJobIfIdle(), null, this.timerPeriod, this.timerPeriod);
                 this.tracer.RelatedInfo(nameof(BackgroundPrefetcher) + ": starting background prefetch timer");
             }
             else
             {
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Prefetch/BackgroundPrefetcher.cs
```diff
@@ -29,14 +29,14 @@ public BackgroundPrefetcher(ITracer tracer, GVFSEnlistment enlistment, PhysicalF
...
-            if (gitObjects.IsUsingCacheServer())
+            if (gitObjects.IsUsingCacheServer() && !GVFSEnlistment.IsUnattended(tracer))
             {
                 this.prefetchJobTimer = new Timer((state) => this.LaunchPrefetchJobIfIdle(), null, this.timerPeriod, this.timerPeriod);
                 this.tracer.RelatedInfo(nameof(BackgroundPrefetcher) + ": starting background prefetch timer");
             }
             else
             {
-                this.tracer.RelatedInfo(nameof(BackgroundPrefetcher) + ": no configured cache server, not starting background prefetch timer");
+                this.tracer.RelatedInfo(nameof(BackgroundPrefetcher) + ": no configured cache server or enlistment is unattended, not starting background prefetch timer");
```
sanoursa: For diagnosing issues, it would be nice to keep these as two separate trace statements, so we know which reason caused it to be disabled.

jrbriggs: @sanoursa I considered that, but it added some complexity to this logging path for little value, because just a few lines prior to this statement in the mount process log file is "GVFS_UNATTENDED" if the enlistment is running in unattend mode, and at the head of the mount process log we log the remote url and objects endpoint url. I think that should give us the information we would need to diagnose.

Thoughts?

sanoursa: Got it, yea that makes sense

Issue #185, opened 2018-08-17
ForceCheckoutRequiresCheckout test fails


jamill:
**Test failure logs:**
```
2018-08-17T02:06:15.5470876Z 1) Failed : GVFS.FunctionalTests.Tests.FastFetchTests.ForceCheckoutRequiresCheckout()
2018-08-17T02:06:15.5471137Z   Expected: 0
2018-08-17T02:06:15.5471330Z   But was:  1
2018-08-17T02:06:15.5471883Z    at GVFS.Tests.Should.ValueShouldExtensions.ShouldEqual[T](T actualValue, T expectedValue, String message) in C:\agent\_work\4\s\GVFS\GVFS.Tests\Should\ValueShouldExtensions.cs:line 34
2018-08-17T02:06:15.5472355Z    at GVFS.FunctionalTests.Tests.FastFetchTests.RunFastFetch(String args) in C:\agent\_work\4\s\GVFS\GVFS.FunctionalTests\Tests\FastFetchTests.cs:line 496
2018-08-17T02:06:15.5472804Z    at GVFS.FunctionalTests.Tests.FastFetchTests.ForceCheckoutRequiresCheckout() in C:\agent\_work\4\s\GVFS\GVFS.FunctionalTests\Tests\FastFetchTests.cs:line 145
```

**Failed Builds:**
[18229.1](https://gvfs.visualstudio.com/ci/_build/results?buildId=552&view=logs)

jrbriggs:
@kyle-rader can you please take a look at this test flakiness?

sanoursa:
This is a reliable failure, not intermittent or flaky. The test case is calling a method that expects the fastfetch verb to succeed, even though the test is explicitly running a command that will fail. @kyle-rader please rework this test ASAP to make it pass.

@kewillford why did this make it through our build policy? Looking in the PR (https://github.com/Microsoft/VFSForGit/pull/159), Kyle ran the full tests and they show success in the PR, but when you click through to the report (https://gvfs.visualstudio.com/ci/_build/results?buildId=551&view=logs) it shows that the test actually failed there too.

PR #186, opened 2018-08-17
Enable runFastFetch to expect an error

kyle-rader:
This should fix the somehow missed test failure for FastFetch addressing #185

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
@@ -142,7 +142,10 @@ namespace GVFS.FunctionalTests.Tests
             this.RunFastFetch($"--checkout --folders \"/Scripts\" -b {Settings.Default.Commitish}");
 
             // Run a second time in the same repo on the same branch with more folders but expect an error.
-            string result = this.RunFastFetch($"--force-checkout --folders \"/GVFS;/Scripts\" -b {Settings.Default.Commitish}", expectError: true);
+            ProcessResult result = this.RunFastFetch($"--force-checkout --folders \"/GVFS;/Scripts\" -b {Settings.Default.Commitish}");
+
+            string[] expectedResults = new string[] { "Cannot use --force-checkout option without --checkout option." };
+            result.Output.ShouldContain(expectedResults);
         }
 
         [TestCase]
@@ -230,8 +233,8 @@ namespace GVFS.FunctionalTests.Tests
 
             // Reset the index and use fastfetch to update the index. Compare against 'git status' baseline.
             GitProcess.Invoke(this.fastFetchRepoRoot, $"-c index.version= {indexVersion} read-tree HEAD");
-            string fastfetchoutput = this.RunFastFetch("--checkout --Allow-index-metadata-update-from-working-tree");
-            Trace.WriteLine(fastfetchoutput); // Written to log file for manual investigation
+            ProcessResult fastFetchResult = this.RunFastFetch("--checkout --Allow-index-metadata-update-from-working-tree");
+            Trace.WriteLine(fastFetchResult.Output); // Written to log file for manual investigation
             string lsfilesAfterUpdate = GitProcess.Invoke(this.fastFetchRepoRoot, "ls-files --debug");
             lsfilesAfterUpdate.ShouldEqual(lsfilesAfterStatus, "git status and fastfetch didn't result in the same index");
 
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -491,9 +491,13 @@ private string RunFastFetch(string args)
...
             
             ProcessResult result = ProcessHelper.Run(processInfo);
-            result.Output.Contains("Error").ShouldEqual(false, result.Output);
-            result.Errors.ShouldBeEmpty(result.Errors);
-            result.ExitCode.ShouldEqual(0);
+            result.Output.Contains("Error").ShouldEqual(expectError, result.Output);
+            if (!expectError)
+            {
+                result.Errors.ShouldBeEmpty(result.Errors);
+                result.ExitCode.ShouldEqual(0);
```
wilbaker: If `expectError` is `true` can\should we verify that `ExitCode` is non-zero?

kyle-rader: Oh yep, thanks

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -491,9 +491,18 @@ private string RunFastFetch(string args)
             processInfo.RedirectStandardError = true;
             
             ProcessResult result = ProcessHelper.Run(processInfo);
-            result.Output.Contains("Error").ShouldEqual(false, result.Output);
-            result.Errors.ShouldBeEmpty(result.Errors);
-            result.ExitCode.ShouldEqual(0);
+            result.Output.Contains("Error").ShouldEqual(expectError, result.Output);
+            if (expectError)
+            {
+                Assert.IsTrue(result.Errors.Length > 0);
```
wilbaker: In our tests we try to use the `Should` syntax whenever possible.  Rather than doing:

`Assert.IsTrue(result.Errors.Length > 0);`

Switch to something like one of these:

`result.Errors.Length.ShouldBeNonEmpty("Expected an error message");`
`result.Errors.Length.ShouldBeAtLeast(1, "Expected an error message");`

kyle-rader: What would Should be non empty mean for an integer?

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -491,9 +491,18 @@ private string RunFastFetch(string args)
             processInfo.RedirectStandardError = true;
             
             ProcessResult result = ProcessHelper.Run(processInfo);
-            result.Output.Contains("Error").ShouldEqual(false, result.Output);
-            result.Errors.ShouldBeEmpty(result.Errors);
-            result.ExitCode.ShouldEqual(0);
+            result.Output.Contains("Error").ShouldEqual(expectError, result.Output);
+            if (expectError)
+            {
+                result.Errors.Length.ShouldBeAtLeast(1, "Expected at lest 1 error");
```
wilbaker: Typo: `lest` -> `least`

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
wilbaker: Doh, that option should have been `result.Errors.ShouldBeNonEmpty("Expected an error message");`

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -141,11 +141,8 @@ public void ForceCheckoutRequiresCheckout()
         {
             this.RunFastFetch($"--checkout --folders \"/Scripts\" -b {Settings.Default.Commitish}");
 
-            // Run a second time in the same repo on the same branch with more folders. 
-            string result = this.RunFastFetch($"--force-checkout --folders \"/GVFS;/Scripts\" -b {Settings.Default.Commitish}");
-
-            string[] expectedResults = new string[] { "Cannot use --force-checkout option without --checkout option." };
-            result.ShouldContain(expectedResults);
+            // Run a second time in the same repo on the same branch with more folders but expect an error.
```
wilbaker: @kyle-rader, could you update this comment to explain why we're expecting an error?

(I'm not sure why the check you removed was failing)

PR #189, opened 2018-08-17
Spike: capture git_trace during test runs

jamill:
NOT FOR MERGING

This is a test PR to run functional tests against and get more detailed logging on failed background git status commands.


jamill:
We have captured the test data from this PR - it is no longer needed.

Issue #152, opened 2018-08-13
Mac: Add missing file notifications


wilbaker:
Add all of the file notifications required for `git status` to report the correct\expected results.

wilbaker:
See #156 for "file modified" notification code changes

PR #191, opened 2018-08-20
Mac: New file created notification

wilbaker:
Part of the work required for #152.

Changes include:

- Adding "new file created" notification to kext and user-mode
- Adding new file notification handling to `MirrorProvider`
- Enabling additional functional tests
- Categorizing more functional tests (that were previously uncategorized)
- Removing the code that blocks modifications to the repo when GVFS is offline

Regarding the last point above, the code for blocking notifications was also preventing copying in new hooks and repairing broken repos (now that newly created files are flagged as being in the root).  The code for supporting this feature can be revisited once we have the rest of notifications and projection changes working on Mac.

It should also be noted that the "new file" notification will be raised for both newly created files, and any files that were created as part of `gvfs clone` (prior to the root of the repo getting marked as such).  This doesn't cause a problem for VFSForGit because it ignores new files created in the .git folder.

Functional test run with these changes:

```
Test Run Summary
  Overall result: Warning
  Test Count: 166, Passed: 164, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 2
    Skipped Tests - Ignored: 2, Explicit: 0, Other: 0
  Start time: 2018-08-20 16:50:23Z
    End time: 2018-08-20 16:54:19Z
    Duration: 236.021 seconds
```

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -35,9 +35,10 @@ static int HandleFileOpOperation(
 static int GetPid(vfs_context_t context);
 
 static uint32_t ReadVNodeFileFlags(vnode_t vn, vfs_context_t context);
-static bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
-static bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
-static bool ActionBitsNotSet(kauth_action_t action, kauth_action_t mask);
+static inline bool HasAncestorFlaggedAsInRoot(vnode_t vnode, vfs_context_t context);
+static inline bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
+static inline bool FileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t context);
+static inline bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
 
 static bool IsFileSystemCrawler(char* procname);
 
@@ -225,6 +226,7 @@ void KauthHandler_HandleKernelMessageResponse(uint64_t messageId, MessageType re
         case MessageType_KtoU_NotifyFileModified:
         case MessageType_KtoU_NotifyFilePreDelete:
         case MessageType_KtoU_NotifyDirectoryPreDelete:
+        case MessageType_KtoU_NotifyFileCreated:
             KextLog_Error("KauthHandler_HandleKernelMessageResponse: Unexpected responseType: %d", responseType);
             break;
     }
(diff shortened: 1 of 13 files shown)


ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -782,20 +735,49 @@ static uint32_t ReadVNodeFileFlags(vnode_t vn, vfs_context_t context)
...
+    // or the root of the file system
+    while (NULLVP != vnode && !vnode_isvroot(vnode))
+    {
+        uint32_t vnodeFileFlags = ReadVNodeFileFlags(vnode, context);
+        if (FileFlagsBitIsSet(vnodeFileFlags, FileFlags_IsInVirtualizationRoot))
+        {
+            inRoot = true;
+            break;
+        }
+        vnode_t parent = vnode_getparent(vnode);
```
sanoursa: If you need a 100% reliable answer, this does not seem to be a good way to walk up the vnode parent chain, because `vnode_getparent` can return null even if the current vnode does have a parent. I'm not sure what we need to do to make it more reliable though.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -382,28 +384,66 @@ static int HandleFileOpOperation(
...
+            goto CleanupAndReturn;
+        }
+
+        bool fileFlaggedInRoot = FileIsFlaggedAsInRoot(currentVnode, context);
+        if (fileFlaggedInRoot && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
+        {
+            goto CleanupAndReturn;
+        }
+		
+        if (!fileFlaggedInRoot && !HasAncestorFlaggedAsInRoot(currentVnode, context))
```
sanoursa: Do you actually need this call? The next call to `ShouldHandleFileOpEvent` already calls `VirtualizationRoots_FindForVnode` and will return a root if one exists. So it seems like `ShouldHandleFileOpEvent` already returns false if there is no ancestor that is a root.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -450,6 +450,20 @@ static void HandleKernelRequest(Message request, void* messageMemory)
                 PrjFS_NotificationType_PreDelete);
             break;
         }
+        
+        case MessageType_KtoU_NotifyFileCreated:
+        {
+            char fullPath[PrjFSMaxPath];
+            CombinePaths(s_virtualizationRootFullPath.c_str(), request.path, fullPath);
+            SetBitInFileFlags(fullPath, FileFlags_IsInVirtualizationRoot, true);
```
sanoursa: What if this fails?

wilbaker: Thanks for pointing this out, I had not considered that scenario.

I think VFSForGit should log if this happens, what would you think if the kext let the provider know in the NewFileCreated callback (e.g. add a `bool fileSuccessfullyFlagged`)?

As far as impact on VFSForGit, if we fail to set the bit:
- There could be duplicate NewFileCreated notifications
- The kext will not deliver the PreDelete notification (because the file is not flagged)

So far I've not been able to think of a scenario where git would behave incorrectly if this happened, the main issue would be missing (or duplicate callbacks).

Other than logging, do you have any ideas on what else we could\should do if this occurs?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
sanoursa: If I understand the design direction correctly (@pmj ?), we will be getting rid of our file flags at some point and our vnode cache will be the source of truth for whether something is a root or not. So I don't think we should spread the implementation details of what it means to be a root. At a minimum, this lock-free approach should be moved into `VirtualizationRoots.hpp/cpp`, but yea it could also be a premature optimization if you haven't actually measured a perf hit.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
sanoursa: This is a general implementation hole that @nickgra is looking into - we need to make sure that the lib's IO is all-or-nothing, and doesn't leave files in a halfway updated state. This one is a bit more problematic, since it's in response to a fileop event and therefore we don't get a chance to redo it. At the very least, we should leave a TODO here, and logging it is a good idea as well. Down the road, we need some strategy for recovering from lost fileop events (and that's a bigger scope than what @nickgra is working on right now, so let's address that one separately)

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
sanoursa: See the other comment related to this. For now, we should just avoid spreading the implementation details of how to find a root.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: Ok thanks, in that case I'll just take this out for now and rely on `ShouldHandleFileOpEvent`

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: See other comment, I will remove this function

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: > At the very least, we should leave a TODO here, and logging it is a good idea as well.

I'll leave a TODO for now.  Long term if we can have failures here (once we've stopped using flags) we'll either want a way to let providers know so that they can log the failure to a log file, or we'll want a PrjFSLib log file where we can log it.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -450,6 +450,22 @@ static void HandleKernelRequest(Message request, void* messageMemory)
                 PrjFS_NotificationType_PreDelete);
             break;
         }
+        
+        case MessageType_KtoU_NotifyFileCreated:
+        {
+            char fullPath[PrjFSMaxPath];
+            CombinePaths(s_virtualizationRootFullPath.c_str(), request.path, fullPath);
+			
+			// TODO(Mac): Handle SetBitInFileFlags failures
```
wilbaker: Remove tabs

PR #199, opened 2018-08-21
Update name of file that indicates installation version

jamill:
Update name of file that the GVFS installer creates when an on disk version 16 capable GVFS is first installed.

After discussions on how GVFS determines whether the GitStatusCache should be enabled to ensure that the machine has been rebooted since a GitStatusCache aware GVFS was installed, we decided on two changes:

1) The name of the file created by the GVFS installer should be changed to `OnDiskVersion16CapableInstallation.dat`

2) Move the logic for determining whether the GitStatusCache should enabled for a GVFS installation from the GVFS.Service to individual mount processes. This means we could get rid of the `EnableGitStatusCacheToken.dat` file that exists in the `GVFS.Service` common application directory.

This PR addresses issue 1.

I want to get agreement on the changes for the file written out by the installer. I can follow up with the changes for the second issue, either for the M139 release, or as a future change.

I am doing another round of testing on these changes, and can do a final verification after we agree on the name.

### DIFF
--- a/GVFS/GVFS.Service/GvfsService.cs
+++ b/GVFS/GVFS.Service/GvfsService.cs
@@ -287,27 +287,36 @@ namespace GVFS.Service
             try
             {
                 string statusCacheVersionTokenPath = Path.Combine(Paths.GetServiceDataRoot(GVFSConstants.Service.ServiceName), GVFSConstants.GitStatusCache.EnableGitStatusCacheTokenFile);
-
-                if (!File.Exists(statusCacheVersionTokenPath))
+                if (File.Exists(statusCacheVersionTokenPath))
                 {
-                    DateTime lastRebootTime = NativeMethods.GetLastRebootTime();
+                    this.tracer.RelatedInfo($"CheckEnableGitStatusCache: EnableGitStatusCacheToken file already exists at {statusCacheVersionTokenPath}.");
+                    return;
+                }
 
-                    // When a version of GVFS that supports the GitStatusCache is installed, it will create
-                    // the following file. By checking the time the file was created, we know when that
-                    // version of GVFS was installed.
-                    string fileToCheck = Path.Combine(Configuration.AssemblyPath, "GitStatusCacheAvailable");
-                    if (File.Exists(fileToCheck))
+                DateTime lastRebootTime = NativeMethods.GetLastRebootTime();
+
+                // GitStatusCache was included with GVFS on disk version 16. The 1st time GVFS that is at or above on disk version
+                // is installed, it will write out a file indicating that the installation is "OnDiskVersion16Capable".
+                // We can query the properties of this file to get the installation time, and compare this with the last reboot time for
+                // this machine.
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -287,28 +287,37 @@ private void CheckEnableGitStatusCacheTokenFile()
             try
             {
                 string statusCacheVersionTokenPath = Path.Combine(Paths.GetServiceDataRoot(GVFSConstants.Service.ServiceName), GVFSConstants.GitStatusCache.EnableGitStatusCacheTokenFile);
```
sanoursa: I think we should get rid of this file. Comparing the capability file and reboot time should be enough, shouldn't it?

sanoursa: I see that you discussed this in the PR comments and left it as a TODO. I do think we should drop it, since it's another point of failure and not a necessary one, it would seem.

jrbriggs: @sanoursa I think we should take this PR in isolation for M139 so we can get it out the door.

That would leave us in a state where we could take the second point up in master moving forward, if we felt we needed to revisit it.

sanoursa: That works for me. The file that is dropped by the installer is one that we will have to maintain forever. The status cache token file is not, so I agree we could clean it up later.

jamill:
Scenarios I tested:

- mounting a repository with out an `EnableGitStatusCacheToken.dat` file leaves git status cache disabled
- mount a repository with an `EnableGitStatusCacheToken.dat` file enables git status cache


- `OnDiskVersion16CapableInstallation.dat` is written by the installer
- `EnableGitStatusCacheToken.dat` is not written if `OnDiskVersion16CapableInstallation.dat`  is not present
```
[8/21/2018 1:40:30 PM] Error {"ErrorMessage":"Unable to determine GVFS installation time: C:\\Program Files\\GVFS\\OnDiskVersion16CapableInstallation.dat does not exist."}
```

- `EnableGitStatusCacheToken.dat` is not written if the creation time for `OnDiskVersion16CapableInstallation.dat`  is before last reboot
```
[8/21/2018 1:52:32 PM] Information {"Message":"CheckEnableGitStatusCache: Not writing EnableGitStatusCacheToken file - machine has not been rebooted since OnDiskVersion16Capable installation. GVFS installation time: 8/21/2018 1:48:55 PM, last reboot time: 8/21/2018 10:45:16 AM"}
```

- `EnableGitStatusCacheToken.dat` is written if the creation time for `OnDiskVersion16CapableInstallation.dat`  is after last reboot
```
[8/21/2018 1:56:29 PM] Information {"Message":"CheckEnableGitStatusCache: Writing out EnableGitStatusCacheToken file. GVFS installation time: 8/21/2018 1:48:55 PM, last Reboot time: 8/21/2018 1:56:06 PM."}
```

GVFS/GVFS.Service/GvfsService.cs
jamill: Thanks! This approach seems reasonable.

Issue #202, opened 2018-08-22
Mac: Folders should be expanded before being deleted


wilbaker:
The first step of a folder rename is a PreDelete notification, and when handling this notification we must expand the folder (and all of its children) to ensure all children are on disk prior to the rename proceeding.

PR #203, opened 2018-08-22
Hardlink notifications on Mac & Windows and rename notifications on Mac

### LINKED ISSUE #200: VFSForGit should listen to hardlink creation events and treat them the same as creating a new file
If hardlinks are created inside the repo, VFSForGit will not react to them as it's not listening to hardlink notifications.  It should listen to these notifications and react them as it would for new files.

wilbaker:
Part of the work for #152 and fixes #200.

*Summary of changes*

- Added hardlink notifications to VFSForGit for Mac and Windows
- Added new hardlink functional tests
- Added rename notifications on Mac
- Enabled additional functional tests on Mac

*Future work*

Unlike VFSForGit on Windows, on Mac users can rename\move folders have not been expanded.  To ensure that all placeholders are on disk prior to the rename the folder must be expanded in the PreDelete notification (#202)

Functional test run (on Mac) with these changes:

```
Test Run Summary
  Overall result: Warning
  Test Count: 206, Passed: 204, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 2
    Skipped Tests - Ignored: 2, Explicit: 0, Other: 0
  Start time: 2018-08-22 22:09:59Z
    End time: 2018-08-22 22:14:22Z
    Duration: 263.482 seconds
```

### DIFF
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/PowerShellRunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/PowerShellRunner.cs
@@ -1,4 +1,5 @@
 ﻿using GVFS.Tests.Should;
+using NUnit.Framework;
 using System.IO;
 
 namespace GVFS.FunctionalTests.FileSystemRunners
@@ -31,6 +32,11 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         {
             "PermissionDenied"
         };
+
+        public override bool SupportsHardlinkCreation
+        {
+            get { return false; }
+        }
 
         protected override string FileName
         {
@@ -103,6 +109,11 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         public override void CreateEmptyFile(string path)
         {
             this.RunProcess(string.Format("-Command \"&{{ New-Item -ItemType file {0}}}\"", path));
+        }
+
+        public override void CreateHardLink(string targetPath, string newLinkPath)
+        {
+            Assert.Fail($"{nameof(PowerShellRunner)} does not support {nameof(this.CreateHardLink)}");
         }
 
         public override void WriteAllText(string path, string contents)
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/PowerShellRunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/PowerShellRunner.cs
@@ -1,5 +1,4 @@
 ﻿using GVFS.Tests.Should;
-using NUnit.Framework;
 using System.IO;
 
 namespace GVFS.FunctionalTests.FileSystemRunners
@@ -32,11 +31,6 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         {
             "PermissionDenied"
         };
-
-        public override bool SupportsHardlinkCreation
-        {
-            get { return false; }
-        }
 
         protected override string FileName
         {
@@ -109,11 +103,6 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         public override void CreateEmptyFile(string path)
         {
             this.RunProcess(string.Format("-Command \"&{{ New-Item -ItemType file {0}}}\"", path));
-        }
-
-        public override void CreateHardLink(string targetPath, string newLinkPath)
-        {
-            Assert.Fail($"{nameof(PowerShellRunner)} does not support {nameof(this.CreateHardLink)}");
         }
 
         public override void WriteAllText(string path, string contents)
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
@@ -189,8 +189,6 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         [TestCase, Order(8)]
         public void RenamedFileAddedToModifiedPathsFile()
         {
-            GitMoveRenameTests.IgnoreSystemIOMoveOnMac(this.fileSystem);
-
             string fileToRenameEntry = "Test_EPF_MoveRenameFileTests/ChangeUnhydratedFileName/Program.cs";
             string fileToRenameTargetEntry = "Test_EPF_MoveRenameFileTests/ChangeUnhydratedFileName/Program2.cs";
             this.VerifyWorktreeBit(fileToRenameEntry, LsFilesStatus.SkipWorktree);
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
@@ -187,18 +187,17 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         }
 
         [TestCase, Order(8)]
-        [Category(Categories.Mac.M2TODO)]
-        public void RenamedFileAddedToSparseCheckoutAndSkipWorktreeBitCleared()
+        public void RenamedFileAddedToModifiedPathsFile()
         {
+            GitMoveRenameTests.IgnoreSystemIOMoveOnMac(this.fileSystem);
+
             string fileToRenameEntry = "Test_EPF_MoveRenameFileTests/ChangeUnhydratedFileName/Program.cs";
             string fileToRenameTargetEntry = "Test_EPF_MoveRenameFileTests/ChangeUnhydratedFileName/Program2.cs";
-            string fileToRenameRelativePath = "Test_EPF_MoveRenameFileTests\\ChangeUnhydratedFileName\\Program.cs";
-            string fileToRenameTargetRelativePath = "Test_EPF_MoveRenameFileTests\\ChangeUnhydratedFileName\\Program2.cs";
             this.VerifyWorktreeBit(fileToRenameEntry, LsFilesStatus.SkipWorktree);
 
             this.fileSystem.MoveFile(
-                this.Enlistment.GetVirtualPathTo(fileToRenameRelativePath), 
-                this.Enlistment.GetVirtualPathTo(fileToRenameTargetRelativePath));
+                this.Enlistment.GetVirtualPathTo(fileToRenameEntry), 
+                this.Enlistment.GetVirtualPathTo(fileToRenameTargetEntry));
             this.Enlistment.WaitForBackgroundOperations().ShouldEqual(true, "Background operations failed to complete.");
 
             GVFSHelpers.ModifiedPathsShouldContain(this.fileSystem, this.Enlistment.DotGVFSRoot, fileToRenameEntry);
@@ -209,19 +208,16 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         }
 
         [TestCase, Order(9)]
-        [Category(Categories.Mac.M2TODO)]
-        public void RenamedFileAndOverwrittenTargetAddedToSparseCheckoutAndSkipWorktreeBitCleared()
+        public void RenamedFileAndOverwrittenTargetAddedToModifiedPathsFile()
         {
             string fileToRenameEntry = "Test_EPF_MoveRenameFileTests_2/MoveUnhydratedFileToOverwriteUnhydratedFileAndWrite/RunUnitTests.bat";
             string fileToRenameTargetEntry = "Test_EPF_MoveRenameFileTests_2/MoveUnhydratedFileToOverwriteUnhydratedFileAndWrite/RunFunctionalTests.bat";
-            string fileToRenameRelativePath = "Test_EPF_MoveRenameFileTests_2\\MoveUnhydratedFileToOverwriteUnhydratedFileAndWrite\\RunUnitTests.bat";
-            string fileToRenameTargetRelativePath = "Test_EPF_MoveRenameFileTests_2\\MoveUnhydratedFileToOverwriteUnhydratedFileAndWrite\\RunFunctionalTests.bat";
             this.VerifyWorktreeBit(fileToRenameEntry, LsFilesStatus.SkipWorktree);
             this.VerifyWorktreeBit(fileToRenameTargetEntry, LsFilesStatus.SkipWorktree);
 
             this.fileSystem.ReplaceFile(
-                this.Enlistment.GetVirtualPathTo(fileToRenameRelativePath),
-                this.Enlistment.GetVirtualPathTo(fileToRenameTargetRelativePath));
+                this.Enlistment.GetVirtualPathTo(fileToRenameEntry),
+                this.Enlistment.GetVirtualPathTo(fileToRenameTargetEntry));
             this.Enlistment.WaitForBackgroundOperations().ShouldEqual(true, "Background operations failed to complete.");
 
             GVFSHelpers.ModifiedPathsShouldContain(this.fileSystem, this.Enlistment.DotGVFSRoot, fileToRenameEntry);
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
@@ -3,7 +3,6 @@ using GVFS.FunctionalTests.Should;
 using GVFS.FunctionalTests.Tools;
 using GVFS.Tests.Should;
 using NUnit.Framework;
-using System;
 using System.Collections.Generic;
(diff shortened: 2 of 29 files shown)


GVFS/GVFS.FunctionalTests/FileSystemRunners/PowerShellRunner.cs
```diff
@@ -105,6 +111,11 @@ public override void CreateEmptyFile(string path)
             this.RunProcess(string.Format("-Command \"&{{ New-Item -ItemType file {0}}}\"", path));
         }
 
+        public override void CreateHardLink(string targetPath, string newLinkPath)
+        {
+            Assert.Fail($"{nameof(PowerShellRunner)} does not support {nameof(this.CreateHardLink)}");
```
sanoursa: Nit: move this to the base class?

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1223,35 +1223,14 @@ private HResult NotifyPreDeleteHandler(string virtualPath, bool isDirectory)
...
-                EventMetadata metadata = this.CreateEventMetadata(virtualPath, e);
-                metadata.Add("destinationPath", destinationPath);
-                metadata.Add("isDirectory", isDirectory);
-                this.LogUnhandledExceptionAndExit(nameof(this.NotifyFileRenamedHandler), metadata);
-            }
+        private void NotifyHardlinkCreated(
+            string relativePath,
+            string destinationPath)
+        {
+            this.OnHardLinkCreated(relativePath, destinationPath);
```
sanoursa: Are these parameters ordered properly? The first parameter is ambiguous: `relativePath` could map to either `relativeTargetPath` or `relativeNewLinkPath`. But the second parameter sure looks like it maps better to `relativeTargetPath`, which makes me think these are swapped?

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -208,6 +208,62 @@ protected void OnWorkingDirectoryFileOrFolderDeleted(string relativePath, bool i
...
+            catch (Exception e)
+            {
+                EventMetadata metadata = this.CreateEventMetadata(relativeSourcePath, e);
+                metadata.Add("destinationPath", relativeDestinationPath);
+                metadata.Add("isDirectory", isDirectory);
+                this.LogUnhandledExceptionAndExit(nameof(this.OnFileRenamed), metadata);
+            }
+        }
+
+        protected void OnHardLinkCreated(string relativeTargetPath, string relativeNewLinkPath)
```
sanoursa: Why do we even need the `relativeTargetPath` parameter? It doesn't seem to be used at all.

sanoursa: And why would it always be relative? Couldn't you create a hardlink inside your repo to some path outside of it? It seems like the target would need to be treated as an absolute path.

MirrorProvider/MirrorProvider.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -164,6 +166,16 @@ private void OnNewFileCreated(string relativePath, bool isDirectory)
         {
             Console.WriteLine($"OnNewFileCreated (isDirectory: {isDirectory}): {relativePath}");
         }
+
+        private void OnFileRenamed(string relativeDestinationPath, bool isDirectory)
```
sanoursa: It seems really odd to get a rename event with only a single path in it. I understand the reasons why, but it's confusing from an API perspective. Should we just treat this as a new file event instead, and just not have a rename event?

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -316,6 +320,22 @@ private void OnFileModifiedOrDeleted(string relativePath, bool isDirectory, bool
             Console.WriteLine($"OnFileModifiedOrDeleted: `{relativePath}`, isDirectory: {isDirectory}, isModfied: {isFileDeleted}, isDeleted: {isFileDeleted}");
         }
 
+        private void OnFileRenamed(
+            string relativePath,
```
sanoursa: relativeSourcePath?

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -316,6 +320,22 @@ private void OnFileModifiedOrDeleted(string relativePath, bool isDirectory, bool
...
+            string relativeDestinationPath,
+            bool isDirectory,
+            ref NotificationType notificationMask)
+        {
+            Console.WriteLine($"OnFileRenamed (isDirectory: {isDirectory}), relativePath: {relativePath}, relativeDestinationPath: {relativeDestinationPath}");
+        }
+
+        private void OnHardlinkCreated(
+            string relativePath,
```
sanoursa: Same question as elsewhere - there's no guarantee this is relative, right?

sanoursa: Er, for one of these two parameters. I actually can't tell which one is the path of the link and which one is the path of the target :-)

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -35,9 +35,9 @@ static int HandleFileOpOperation(
 static int GetPid(vfs_context_t context);
 
 static uint32_t ReadVNodeFileFlags(vnode_t vn, vfs_context_t context);
-static inline bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
-static inline bool FileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t context);
-static inline bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
```
sanoursa: Just curious, why did you take the `inline`s back out?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -376,19 +379,66 @@ static int HandleFileOpOperation(
...
     {
-        vnode_t currentVnode = reinterpret_cast<vnode_t>(arg0);
-        // arg1 is the (const char *) path
-        int closeFlags = static_cast<int>(arg2);
+        // arg0 is the (const char *) fromPath (or the file being linked to)
+        const char* newPath = (const char*)arg1;
         
-        vtype vnodeType = vnode_vtype(currentVnode);
-        if (ShouldIgnoreVnodeType(vnodeType, currentVnode))
+        // TODO(Mac): Improve error handling
```
sanoursa: And more importantly, we just lost a notification here. We will need some pattern for dealing with that.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -376,19 +379,66 @@ static int HandleFileOpOperation(
...
+        }
+        else
+        {
+            messageType = MessageType_KtoU_NotifyFileHardLinkCreated;
+        }
+
+        int kauthResult;
+        int kauthError;
+        if (!TrySendRequestAndWaitForResponse(
+            root,
```
sanoursa: For consistency with your other code, indent one more level here

GVFS/GVFS.FunctionalTests/FileSystemRunners/PowerShellRunner.cs
wilbaker: Sure, I'll move a default `SupportsHardlinkCreation` to the base as well

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
wilbaker: These are ordered correctly.  

I agree the naming could be better (and I could change these names), but as a start I went with the approach of matching the parameter names declared in the ProjFS Managed API:

```
    /// <param name="relativePath">The path, relative to the virtualization root, of the file for
    ///     which the hard link was created.
    ///     <para>This parameter will be "" to indicate that the hard link name was created inside
    ///     the virtualization instance, i.e. a new hard link was created inside the virtualization
    ///     instance to a file that exists outside the virtualization instance.</para>
    /// </param>
    /// <param name="destinationPath">The path, relative to the virtualization root, of the new hard
    ///     link name.
    ///     <para>This parameter will be "" to indicate that the hard link name was created
    ///     outside the virtualization instance, i.e. a new hard link was created outside the
    ///     virtualization instance for a file that exists inside the virtualization instance.</para>
    /// </param>
```

Let me know if you think it's better to match the API names or use our own names.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
wilbaker: > Why do we even need the relativeTargetPath parameter? It doesn't seem to be used at all.

It's not *strictly* needed, but we do log it if there's an exception:

`metadata.Add(nameof(relativeTargetPath), relativeTargetPath);`

My thought was that if something goes wrong it's helpful to have as much information as we can.

> And why would it always be relative? Couldn't you create a hardlink inside your repo to some path outside of it? It seems like the target would need to be treated as an absolute path.

See related comment as well (for detailed descriptions of these paths), if either of the paths are outside of the repo they will be empty strings.

MirrorProvider/MirrorProvider.Mac/MacFileSystemVirtualizer.cs
wilbaker: For historical purposes, it should be noted that we *could* deliver both paths on Mac, but because VFSForGit does not strictly need the source path on Mac we opted to keep the kext simpler and only deliver the destination path for now.

>  but it's confusing from an API perspective. Should we just treat this as a new file event instead, and just not have a rename event?

I've found it helpful when debugging\developing to keep these as separate events (and Keeping them as separate events also makes us more consistent with ProjFS on Windows).

I lean toward keeping them separate and having VFSForGit decide to handle this notification in the same way as a new file notification.

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
wilbaker: See other comment as well.  Currently these names match the names used in the ProjFS API, let me know if you think it's better to keep them consistent or change them.

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
wilbaker: > Same question as elsewhere - there's no guarantee this is relative, right?

As discussed elsewhere, any paths outside of the root will be an empty string (by design)

> I actually can't tell which one is the path of the link and which one is the path of the target :-)

Same comment here too, do you think we should use our own names or stay consistent with ProjFS?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: We can add them back.  

I remembered while working on the code that the `inline` is only required on the definition and not the declaration (see: https://isocpp.org/wiki/faq/inline-functions#inline-nonmember-fns).

But it doesn't hurt anything to leave them (and makes it clearer that they are expected to be inline) so I can add them back in.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: Good point, I will add some more details to this comment

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: :+1:

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
```diff
@@ -38,10 +38,35 @@ public void CreateFileTest()
...
             this.Enlistment.WaitForBackgroundOperations().ShouldEqual(true, "Background operations failed to complete.");
             GVFSHelpers.ModifiedPathsShouldContain(this.fileSystem, this.Enlistment.DotGVFSRoot, emptyFileName);
-            this.Enlistment.GetVirtualPathTo(fileName).ShouldBeAFile(this.fileSystem);
+            this.Enlistment.GetVirtualPathTo(emptyFileName).ShouldBeAFile(this.fileSystem);
         }
 
         [TestCase, Order(2)]
+        public void CreateHardLinkTest()
+        {
+            if (!this.fileSystem.SupportsHardlinkCreation)
```
wilbaker: This test is not getting any coverage on Mac as it's not using BashRunner

GVFS/GVFS.FunctionalTests/Tests/GitCommands/AddStageTests.cs
```diff
@@ -32,12 +32,30 @@ public void StageBasicTest()
         }
 
         [TestCase, Order(3)]
+        [Category(Categories.Mac.M2)]
+        public void AddAndStageHardLinksTest()
+        {
+            if (!this.FileSystem.SupportsHardlinkCreation)
```
wilbaker: This is not adding any coverage is AddStageTests are only run with `SystemIORunner`

MirrorProvider/MirrorProvider.Mac/MacFileSystemVirtualizer.cs
wilbaker: I've added some additional comments in the notification callback

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
wilbaker: Test fixture updated to use all runners (whatever used to cause failures with `BashRunner` is no longer an issue)

GVFS/GVFS.FunctionalTests/Tests/GitCommands/AddStageTests.cs
```diff
@@ -32,12 +32,30 @@ public void StageBasicTest()
...
         [TestCase, Order(3)]
+        [Category(Categories.Mac.M2)]
+        public void AddAndStageHardLinksTest()
+        {
+            if (!this.FileSystem.SupportsHardlinkCreation)
+            {
+                return;
+            }
+
+            this.CreateHardLink("Readme.md", "ReadmeLink.md");
```
sanoursa: Nit: in the `IPlatformFileSystem` interface up above, I saw that the link name was first and the target name was second. I think it would help us keep these straight if we keep a consistent ordering through all of our own code.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -215,6 +215,24 @@ protected void EditFile(string filePath, string content)
...
+            string virtualTargetFile = Path.Combine(this.Enlistment.RepoRoot, targetPath);
+            string controlTargetFile = Path.Combine(this.ControlGitRepo.RootPath, targetPath);
+            string virtualNewLinkFile = Path.Combine(this.Enlistment.RepoRoot, newLinkPath);
+            string controlNewLinkFile = Path.Combine(this.ControlGitRepo.RootPath, newLinkPath);
+
+            // GitRepoTests are only run with SystemIORunner (which does not support hardlink
+            // creation) so use a BashRunner instead.
+            this.FileSystem.ShouldBeOfType<SystemIORunner>();
+            this.FileSystem.SupportsHardlinkCreation.ShouldBeFalse(
+                "FileSystem *does* support hard link creation, CreateHardLink no longer needs to create a BashRunner");
```
sanoursa: This message is confusing. You're saying that if in the future it ends up supporting hardlinks, we can remove this code?

wilbaker: Yep that's right, let me try re-working this message.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/AddStageTests.cs
wilbaker: :+1: I'll clean this up

Issue #117, opened 2018-06-30
Remove temp files from ModifiedPaths on deletion


sanoursa:
When a new file is created, it has to be added to the ModifiedPaths collection, even if it is currently .gitignore'd (because it could get unignored later and the file system still has to function correctly if that happens). However, if that file is then deleted before it is ever added to the index or committed, then it should be safe to remove it from the ModifiedPaths list.

We should determine the scenarios in which it is safe to remove entries from the ModifiedPaths list to avoid status runtimes from blowing out as a result of temp files written inside the src directory.

wilbaker:
Additionally, ideally recursive folder entries would be removed after the folder has been deleted (when safe).

For example, if a user creates a folder "foo" and deletes it, and then checks out a branch that has a folder named "foo" that folder should be projected by GVFS rather than having its contents written out by git.

PR #207, opened 2018-08-23
Remove temp paths from ModifiedPathsDatabase

kewillford:
Builds and other processes will write temp files or create temp directories in the source directory which then adds the file or folder path to the ModifiedPathsDatabase.  This causes performance issues because the ModifiedPathsDatabase is currently append only.

This PR is an initial change to address the issue of files or folder being created and deleted.  This change only removes from the ModifiedPathsDatabase when the file or folder was created and then delete and the index has not changed in between to solve the issue with builds and other processes creating and deleting a bunch of files and folders.

See #117

### DIFF
--- a/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
+++ b/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
@@ -138,7 +138,11 @@ namespace GVFS.Common
             metadata.Add("Area", "ModifiedPathsDatabase");
             metadata.Add(nameof(entry), entry);
             metadata.Add(nameof(isFolder), isFolder);
-            metadata.Add("Exception", e.ToString());
+            if (e != null)
+            {
+                metadata.Add("Exception", e.ToString());
+            }
+
             return metadata;
         }
 
--- a/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
+++ b/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
@@ -75,36 +75,97 @@ namespace GVFS.Common
                 }
                 catch (IOException e)
                 {
-                    if (this.Tracer != null)
-                    {
-                        EventMetadata metadata = new EventMetadata();
-                        metadata.Add("Area", "ModifiedPathsDatabase");
-                        metadata.Add(nameof(entry), entry);
-                        metadata.Add(nameof(isFolder), isFolder);
-                        metadata.Add("Exception", e.ToString());
-                        this.Tracer.RelatedWarning(metadata, $"IOException caught while processing {nameof(this.TryAdd)}");
-                    }
-
+                    this.TraceWarning(isFolder, entry, e, nameof(this.TryAdd));
                     return false;
                 }
                 catch (Exception e)
                 {
-                    if (this.Tracer != null)
-                    {
-                        EventMetadata metadata = new EventMetadata();
-                        metadata.Add("Area", "ModifiedPathsDatabase");
-                        metadata.Add(nameof(entry), entry);
-                        metadata.Add(nameof(isFolder), isFolder);
-                        metadata.Add("Exception", e.ToString());
-                        this.Tracer.RelatedError(metadata, $"Exception caught while processing {nameof(this.TryAdd)}");
-                    }
-
+                    this.TraceError(isFolder, entry, e, nameof(this.TryAdd));
                     isRetryable = false;
                     return false;
                 }
+
+                return true;
             }
 
             return true;
+        }
+
+        public bool TryRemove(string path, bool isFolder, out bool isRetryable)
+        {
+            isRetryable = true;
+            string entry = this.NormalizeEntryString(path, isFolder);
+            if (this.modifiedPaths.Contains(entry))
+            {
+                isRetryable = true;
+                try
+                {
+                    this.WriteRemoveEntry(entry, () => this.modifiedPaths.TryRemove(entry));
+                }
+                catch (IOException e)
+                {
+                    this.TraceWarning(isFolder, entry, e, nameof(this.TryRemove));
+                    return false;
+                }
+                catch (Exception e)
+                {
+                    this.TraceError(isFolder, entry, e, nameof(this.TryRemove));
+                    isRetryable = false;
+                    return false;
+                }
+
+                return true;
+            }
+
+            return true;
+        }
+
+        public void WriteAllEntriesAndFlush()
+        {
+            try
+            {
+                this.WriteAndReplaceDataFile(this.GenerateDataLines);
+            }
+            catch (Exception e)
+            {
+                throw new FileBasedCollectionException(e);
+            }
+        }
+
+        private static EventMetadata CreateEventMetadata(bool isFolder, string entry, Exception e)
+        {
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("Area", "ModifiedPathsDatabase");
+            metadata.Add(nameof(entry), entry);
+            metadata.Add(nameof(isFolder), isFolder);
+            metadata.Add("Exception", e.ToString());
+            return metadata;
+        }
+
+        private IEnumerable<string> GenerateDataLines()
+        {
+            foreach (string entry in this.modifiedPaths)
+            {
+                yield return this.FormatAddLine(entry);
+            }
+        }
+
+        private void TraceWarning(bool isFolder, string entry, Exception e, string method)
+        {
+            if (this.Tracer != null)
+            {
+                EventMetadata metadata = CreateEventMetadata(isFolder, entry, e);
+                this.Tracer.RelatedWarning(metadata, $"{e.GetType().Name} caught while processing {method}");
+            }
+        }
+
+        private void TraceError(bool isFolder, string entry, Exception e, string method)
+        {
+            if (this.Tracer != null)
+            {
+                EventMetadata metadata = CreateEventMetadata(isFolder, entry, e);
+                this.Tracer.RelatedError(metadata, $"{e.GetType().Name} caught while processing {method}");
+            }
         }
 
         private bool TryParseAddLine(string line, out string key, out string value, out string error)
--- a/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
+++ b/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
@@ -84,8 +84,6 @@ namespace GVFS.Common
                     isRetryable = false;
                     return false;
                 }
-
-                return true;
             }
 
             return true;
@@ -113,8 +111,6 @@ namespace GVFS.Common
                     isRetryable = false;
                     return false;
                 }
-
-                return true;
             }
 
             return true;
--- a/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
+++ b/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
@@ -324,24 +324,30 @@ namespace GVFS.FunctionalTests.Windows.Tests
             this.Enlistment.MountGVFS();
             this.Enlistment.UnmountGVFS();
 
-            string expectedModifiedPaths = @"A .gitattributes
-A developer/me/
-A developer/me/JLANGE9._prerazzle
-A developer/me/StateSwitch.Save
-A tools/x86/remote.exe
-A tools/x86/runelevated.exe
-A tools/amd64/remote.exe
-A tools/amd64/runelevated.exe
-A tools/perllib/MS/TraceLogging.dll
-A tools/managed/v2.0/midldd.CheckedInExe
-A tools/managed/v4.0/sdapi.dll
-A tools/managed/v2.0/midlpars.dll
-A tools/managed/v2.0/RPCDataSupport.dll
-A tools/managed/v2.0/MidlStaticAnalysis.dll
-A tools/perllib/MS/Somefile.txt
-";
+            string[] expectedModifiedPaths =
+                {
+                    "A .gitattributes",
+                    "A developer/me/",
+                    "A developer/me/JLANGE9._prerazzle",
+                    "A developer/me/StateSwitch.Save",
+                    "A tools/x86/remote.exe",
+                    "A tools/x86/runelevated.exe",
+                    "A tools/amd64/remote.exe",
+                    "A tools/amd64/runelevated.exe",
+                    "A tools/perllib/MS/TraceLogging.dll",
+                    "A tools/managed/v2.0/midldd.CheckedInExe",
+                    "A tools/managed/v4.0/sdapi.dll",
+                    "A tools/managed/v2.0/midlpars.dll",
+                    "A tools/managed/v2.0/RPCDataSupport.dll",
+                    "A tools/managed/v2.0/MidlStaticAnalysis.dll",
+                    "A tools/perllib/MS/Somefile.txt",
+                };
 
-            modifiedPathsDatabasePath.ShouldBeAFile(this.fileSystem).WithContents(expectedModifiedPaths);
+            modifiedPathsDatabasePath.ShouldBeAFile(this.fileSystem);
+            this.fileSystem.ReadAllText(modifiedPathsDatabasePath)
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
+                .OrderBy(x => x)
+                .ShouldMatchInOrder(expectedModifiedPaths.OrderBy(x => x));
 
             this.ValidatePersistedVersionMatchesCurrentVersion();
         }
--- a/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
+++ b/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
@@ -661,6 +661,9 @@ namespace GVFS.Virtualization
                     break;
 
                 case FileSystemTask.OperationType.OnFilePreDelete:
+                    // This code assumes that the current implementations of FileSystemVirtualizer will call either
+                    // the PreDelete or the Delete not both so if a new implementation starts calling both
+                    // this will need to be cleaned up to not duplicate the work that is being done.
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
                     if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
                     {
@@ -682,6 +685,9 @@ namespace GVFS.Virtualization
                     break;
 
                 case FileSystemTask.OperationType.OnFileDeleted:
+                    // This code assumes that the current implementations of FileSystemVirtualizer will call either
+                    // the PreDelete or the Delete not both so if a new implementation starts calling both
+                    // this will need to be cleaned up to not duplicate the work that is being done.
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
                     if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
                     {
@@ -796,6 +802,9 @@ namespace GVFS.Virtualization
                     break;
 
                 case FileSystemTask.OperationType.OnFolderPreDelete:
+                    // This code assumes that the current implementations of FileSystemVirtualizer will call either
+                    // the PreDelete or the Delete not both so if a new implementation starts calling both
+                    // this will need to be cleaned up to not duplicate the work that is being done.
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
                     if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
                     {
@@ -817,6 +826,9 @@ namespace GVFS.Virtualization
                     break;
 
                 case FileSystemTask.OperationType.OnFolderDeleted:
+                    // This code assumes that the current implementations of FileSystemVirtualizer will call either
+                    // the PreDelete or the Delete not both so if a new implementation starts calling both
+                    // this will need to be cleaned up to not duplicate the work that is being done.
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
                     if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
                     {
--- a/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
+++ b/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
@@ -438,6 +438,11 @@ namespace GVFS.Virtualization
             this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnFileDeleted(relativePath));
         }
 
+        public void OnFilePreDelete(string relativePath)
+        {
+            this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnFilePreDelete(relativePath));
+        }
+
         public void OnFolderCreated(string relativePath)
         {
             this.newlyCreatedFileAndFolderPaths.Add(relativePath);
@@ -452,6 +457,11 @@ namespace GVFS.Virtualization
         public void OnFolderDeleted(string relativePath)
         {
             this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnFolderDeleted(relativePath));
+        }
+
+        public void OnFolderPreDelete(string relativePath)
+        {
+            this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnFolderPreDelete(relativePath));
         }
 
         public void OnPlaceholderFileCreated(string relativePath, string sha, string triggeringProcessImageFileName)
@@ -650,6 +660,27 @@ namespace GVFS.Virtualization
 
                     break;
 
+                case FileSystemTask.OperationType.OnFilePreDelete:
+                    metadata.Add("virtualPath", gitUpdate.VirtualPath);
+                    if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
+                    {
+                        string fullPathToFolder = Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, gitUpdate.VirtualPath);
+                        if (!this.context.FileSystem.FileExists(fullPathToFolder))
+                        {
+                            result = this.TryRemoveModifiedPath(gitUpdate.VirtualPath, isFolder: false);
+                        }
+                        else
+                        {
+                            result = FileSystemTaskResult.Success;
+                        }
+                    }
+                    else
+                    {
+                        result = this.AddModifiedPathAndRemoveFromPlaceholderList(gitUpdate.VirtualPath);
+                    }
+
+                    break;
+
                 case FileSystemTask.OperationType.OnFileDeleted:
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
                     if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
@@ -764,6 +795,27 @@ namespace GVFS.Virtualization
 
                     break;
 
+                case FileSystemTask.OperationType.OnFolderPreDelete:
+                    metadata.Add("virtualPath", gitUpdate.VirtualPath);
+                    if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
+                    {
+                        string fullPathToFolder = Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, gitUpdate.VirtualPath);
+                        if (!this.context.FileSystem.DirectoryExists(fullPathToFolder))
+                        {
+                            result = this.TryRemoveModifiedPath(gitUpdate.VirtualPath, isFolder: true);
+                        }
+                        else
+                        {
+                            result = FileSystemTaskResult.Success;
+                        }
+                    }
+                    else
+                    {
+                        result = this.TryAddModifiedPath(gitUpdate.VirtualPath, isFolder: true);
+                    }
+
+                    break;
+
                 case FileSystemTask.OperationType.OnFolderDeleted:
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
                     if (this.newlyCreatedFileAndFolderPaths.Contains(gitUpdate.VirtualPath))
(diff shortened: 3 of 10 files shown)


GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -69,44 +69,86 @@ public bool TryAdd(string path, bool isFolder, out bool isRetryable)
...
+
+        private IEnumerable<string> GenerateDataLines()
+        {
+            foreach (string entry in this.modifiedPaths)
+            {
+                yield return this.FormatAddLine(entry);
+            }
+        }
+
+        private bool TryAddOrRemove(string entry, bool isFolder, out bool isRetryable, bool addEntry)
```
sanoursa: This seems like it should be two separate methods, since it does two non-overlapping actions based on the value of a bool parameter

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -33,6 +33,7 @@ public class FileSystemCallbacks : IDisposable, IHeartBeatMetadataProvider
         private GVFSContext context;
         private GVFSGitObjects gitObjects;
         private ModifiedPathsDatabase modifiedPaths;
+        private ConcurrentHashSet<string> newPaths;
```
sanoursa: This name isn't super clear. Are the "paths" new? Or are these the paths of newly created files?

I would rename this to `newlyCreatedFiles`, or even better though more verbose, `newlyCreatedFilesAndDirectories`

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -759,6 +781,17 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
...
         }
 
+        private FileSystemTaskResult TryRemoveModifiedPath(string virtualPath, bool isFolder)
+        {
+            if (!this.modifiedPaths.TryRemove(virtualPath, isFolder, out bool isRetryable))
+            {
+                return isRetryable ? FileSystemTaskResult.RetryableError : FileSystemTaskResult.FatalError;
+            }
+
+            this.InvalidateGitStatusCache();
```
sanoursa: Why did you add this line here? And how did you know you needed it?

In general, as we continue to make changes in this code, what are the principles we should keep in mind about when to invalidate the status cache? cc @jamill

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -793,7 +826,7 @@ private FileSystemTaskResult AddModifiedPathAndRemoveFromPlaceholderList(string
 
         private FileSystemTaskResult PostBackgroundOperation()
         {
-            this.modifiedPaths.ForceFlush();
+            this.modifiedPaths.WriteAllEntriesAndFlush();
```
sanoursa: Why did this need to change, and what is the difference from before?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -630,7 +636,15 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
 
                 case FileSystemTask.OperationType.OnFileDeleted:
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
-                    result = this.AddModifiedPathAndRemoveFromPlaceholderList(gitUpdate.VirtualPath);
+                    if (this.newPaths.Contains(gitUpdate.VirtualPath))
+                    {
+                        result = this.TryRemoveModifiedPath(gitUpdate.VirtualPath, isFolder: false);
```
sanoursa: Also remove it from `newPaths`?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -630,7 +636,15 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
 
                 case FileSystemTask.OperationType.OnFileDeleted:
                     metadata.Add("virtualPath", gitUpdate.VirtualPath);
-                    result = this.AddModifiedPathAndRemoveFromPlaceholderList(gitUpdate.VirtualPath);
+                    if (this.newPaths.Contains(gitUpdate.VirtualPath))
```
wilbaker: Two related comments:

1. Currently `OnFileDeleted` is scheduled from the PreDelete callback on Mac.  However, on Mac we can't remove files from ModifiedPaths in PreDelete because we don't actually know that the file has been deleted.

2. Related to 1 (and something @jrbriggs and I were just discussing), should we try to design this to work on mac and Windows from the start?  Or at this point should we just ensure what whatever we do for Windows does not break Mac?  

For this specific feature it might only take a few tweaks to get things working on both platforms (e.g. on Mac we could take an extra step of checking if the file actually was deleted before calling `TryRemoveModifiedPath`).  Thoughts?

CC: @sanoursa for his thoughts as well.

wilbaker: > However, on Mac we can't remove files from ModifiedPaths in PreDelete because we don't actually know that the file has been deleted.

The bare minimum change for this PR could be to wrap the new code here in an IsUnderConstruction check.

There's still the open question though of whether we design\implement for both platforms from the start.

wilbaker:
@sanoursa I filed #211 for the Mac functional test failures related to `mapping.dat` in the .gvfsCache folder

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
sanoursa: I think it would be nice to start thinking about how to make all new features cross-platform from the start. Unless we're up against a part of the Mac design that is still unknown or in heavy flux, I would take the time now to make this work for both.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford: Yeah I went this way because there were essentially ~15 lines of duplicate code with 2 separate methods but this does feel a little hacky and not as straight forward, do I'll switch it back.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: I'll go with `newlyCreatedFileAndFolderPaths` if that works?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: This is here because I was following the `TryAddModifiedPath` method and wanted to make sure the status cache would stay correct.  I will think this over more to determine if it is really needed and any input from @jamill would be great as well.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: This was changed so that the ModifiedPaths file will be clean after the background operations, meaning that there will only be the "A path" entries.  Otherwise the file will become littered with "A path", "D path" entries and would never get cleaned up.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
jamill: We need to invalidate the git status cache when we observe / record operations that possibly affect the results of a Git status call. In this case, we are processing an event that indicates a file / directory on disk has been mutated, so this is appropriate.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/PersistedModifiedPathsTests.cs
```diff
@@ -116,8 +155,39 @@ public void ExcludeSparseFileSavedAfterRemount(FileSystemRunner fileSystem)
...
+
+        private string CreateFile(FileSystemRunner fileSystem, string relativePath)
+        {
+            string tempFile = this.Enlistment.GetVirtualPathTo(relativePath);
+            fileSystem.WriteAllText(tempFile, $"Contents for the {relativePath} file");
+            tempFile.ShouldBeAFile(fileSystem);
+            return tempFile;
+        }
+
+        private void ValidateModifiedPathsDoNotContain(FileSystemRunner fileSystem, params string[] paths)
```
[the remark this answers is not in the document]
kewillford: I think I can use it but a little bit differently than intended because it is just searching the entire string from the modified paths so if I have a line like
Temp/dir1/file.txt
Temp/, dir1/, Temp/dir1/ would all be found even though the modified paths does not contain those paths directly which is what I am looking for.
Looking to use it like this which should match or not match a line in the modified paths.
```
GVFSHelpers.ModifiedPathsShouldNotContain(fileSystem, this.Enlistment.DotGVFSRoot, paths.Select(x => $"A {x}" + Environment.NewLine).ToArray());
GVFSHelpers.ModifiedPathsShouldNotContain(fileSystem, this.Enlistment.DotGVFSRoot, paths.Select(x => $"D {x}" + Environment.NewLine).ToArray());
```

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -364,6 +366,8 @@ public virtual void OnIndexFileChange()
                 this.GitIndexProjection.InvalidateProjection();
                 this.InvalidateGitStatusCache();
             }
+
+            this.newlyCreatedFileAndFolderPaths.Clear();
```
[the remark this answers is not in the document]
kewillford: Since this change is only meant to help when some process is creating and deleting temporary files I decided to clear every index change to be safe and because `git status` would be some user initiated action but if they ran this while some process was creating and deleting files, some paths might not be removed.  So hopefully most of the time there won't be a bunch of files created, user runs some git command, and then the files are deleted.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -75,38 +75,99 @@ public bool TryAdd(string path, bool isFolder, out bool isRetryable)
...
+            }
+        }
+
+        private static EventMetadata CreateEventMetadata(bool isFolder, string entry, Exception e)
+        {
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("Area", "ModifiedPathsDatabase");
+            metadata.Add(nameof(entry), entry);
+            metadata.Add(nameof(isFolder), isFolder);
+            metadata.Add("Exception", e.ToString());
```
wilbaker: What would you think of adding a `null` check before calling `e.ToString()`?  Most of our other `CreateEventMetadata` helper methods handle null.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -763,6 +785,26 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
             return result;
         }
 
+        private FileSystemTaskResult TryRemoveModifiedPath(string virtualPath, bool isFolder)
```
wilbaker: Should this method also be called for renames (with the old path when the old path is in `newlyCreatedFileAndFolderPaths`)?

If so, this might be a reason to plumb that through the Mac callbacks as well (as a separate follow up issue, CC: @sanoursa for his thoughts).

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -763,6 +785,26 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
...
+            }
+
+            if (!this.modifiedPaths.TryRemove(virtualPath, isFolder, out bool isRetryable))
+            {
+                return isRetryable ? FileSystemTaskResult.RetryableError : FileSystemTaskResult.FatalError;
+            }
+
+            this.newlyCreatedFileAndFolderPaths.TryRemove(virtualPath);
+
+            this.InvalidateGitStatusCache();
```
wilbaker: Why do we have to invalidate the status cache now (or, why didn't we have to before)?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: If we do this then shouldn't the new path be added to the `newlyCreatedFileAndFolderPaths` so that if a file is created, renamed and then deleted nothing will be left in the modified paths?

wilbaker: That's sounds like a good idea as well.  When I left my original comment I was thinking about the common tmp+rename pattern:

1. Create tmp file
2. Write to tmp file
3. Rename+overwrite to final path

But you're right that after step #3 that file could be deleted (and if it were new to start with it's something else we could remove from modified paths).

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: Anything that would affect the status cache should invalidate it.  In this case we are removing paths from the modified paths.  So this is more of being paranoid because with the way it is coded right now the `newlyCreatedFileAndFolderPaths` would get cleared with any status update and this code path not hit.

But this brings up a question for @jamill, if the background status runs will that clear the `newlyCreatedFileAndFolderPaths`, which right now if it changes the index it would, so this makes the argument to only clear when the projection changes stronger.  If that change is made then the invalidation of the status cache would be required because a user could create a file, run git status, delete file, run git status and since the modified paths changed with the delete file the status cache would need to be invalidated.  Right now when that delete happens we add to the modified paths which invalidates the status cache.

I will do some testing around this to make sure the modified paths are getting clean up even with the background status running.

wilbaker: > or, why didn't we have to before

I found that we're currently invalidating the status cache in TryAddModifiedPath, and @kewillford your explanation about why we need it for delete as well makes sense. Thanks!

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -678,14 +700,20 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
                                         foreach (DirectoryItemInfo itemInfo in this.context.FileSystem.ItemsInDirectory(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPath)))
                                         {
                                             string itemVirtualPath = Path.Combine(folderPath, itemInfo.Name);
-                                            if (itemInfo.IsDirectory)
+                                            string oldItemVirtualPath = gitUpdate.OldVirtualPath + itemVirtualPath.Substring(gitUpdate.VirtualPath.Length);
+                                            if (!this.newlyCreatedFileAndFolderPaths.Contains(itemVirtualPath))
                                             {
-                                                relativeFolderPaths.Enqueue(itemVirtualPath);
+                                                this.newlyCreatedFileAndFolderPaths.Add(itemVirtualPath);
```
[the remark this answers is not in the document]
kewillford: We are adding the top level folder `result = this.TryAddModifiedPath(gitUpdate.VirtualPath, isFolder: true);` so we don't need to add all the subfolder and files.  This is kind of left over from when we would add the top level folder to that sparse-checkout and then need to add files to the always_exclude.  Since this is just one list now we only need the top level folder and git will handle it correctly for both the sparse and exclude cases.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -669,10 +685,13 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
                         result = this.TryAddModifiedPath(gitUpdate.VirtualPath, isFolder: true);
                         if (result == FileSystemTaskResult.Success)
                         {
+                            this.newlyCreatedFileAndFolderPaths.Add(gitUpdate.VirtualPath);
+                            this.TryRemoveModifiedPath(gitUpdate.OldVirtualPath, isFolder: true);
```
wilbaker: Should we check the return value of `TryRemoveModifiedPath` here?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -768,6 +792,31 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
             return result;
         }
 
+        private FileSystemTaskResult TryRemoveModifiedPath(string virtualPath, bool isFolder)
+        {
+            if (!this.newlyCreatedFileAndFolderPaths.Contains(virtualPath))
+            {
+                return FileSystemTaskResult.Invalid;
+            }
```
wilbaker: For what it's worth, I think it was OK when you had the `newlyCreatedFileAndFolderPaths` check outside of `TryRemoveModifiedPath`.

When I suggested adding a helper I wasn't sure what you'd think of something like:

```
private FileSystemTaskResult TryUpdateModifiedPathsForRemoval(string virtualPath, bool isFolder)
{
    if (this.newlyCreatedFileAndFolderPaths.Contains(virtualPath))
    {
        return this.TryRemoveModifiedPath(virtualPath, isFolder);
    }

    return this.TryAddModifiedPath(virtualPath, isFolder);
}
```

And then calling that method in the three spots that needs that logic (up to you which you like better, I'm just throwing this out there 😃)

kewillford: I thought about it but couldn't come up with a good name for the method that wasn't `TryRemoveFromModifiedPathsIfNewlyCreatedOrAddIfNotNewlyCreated` :)

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: I was up in the air about that because on the one hand there won't be any issues if we fail to remove the path except the modified paths doesn't get as small.  On the other we could retry when needed and keep the modified paths small but I'm not sure we should kill the mount process if we get a `FatalError`.  What are you thoughts?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -768,6 +792,31 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
             return result;
         }
 
+        private FileSystemTaskResult TryRemoveModifiedPath(string virtualPath, bool isFolder)
+        {
+            if (!this.newlyCreatedFileAndFolderPaths.Contains(virtualPath))
+            {
+                return FileSystemTaskResult.Invalid;
```
wilbaker: Just noticed this now, but I believe that `FileSystemTaskResult.Invalid` is never supposed to be used in code (i.e. it's an invalid value for `FileSystemTaskResult`).

Given the choice I think it would be cleaner to move this check out of `TryRemoveModifiedPath` and that way the return value of `TryRemoveModifiedPath` reflects only whether we succeeded\failed when trying to update ModifiedPaths.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
wilbaker: I agree there's minimal harm to this failing, but I'm also thinking that if there are failures it could indicate some larger issue with the state of the file system (that I'm not sure we should ignore).

I lean toward checking the value (and killing the process if we see a `FatalError`).  @sanoursa, what are your thoughts?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -625,7 +632,11 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
                     result = FileSystemTaskResult.Success;
                     if (!string.IsNullOrEmpty(gitUpdate.OldVirtualPath) && !IsPathInsideDotGit(gitUpdate.OldVirtualPath))
                     {
-                        result = this.AddModifiedPathAndRemoveFromPlaceholderList(gitUpdate.OldVirtualPath);
+                        result = this.TryRemoveModifiedPath(gitUpdate.OldVirtualPath, isFolder: false);
+                        if (result == FileSystemTaskResult.Invalid)
```
sanoursa: Our general pattern is that enum values of `Invalid` are meant to be illegal, and you should throw if you ever see that value. Please create a new enum value if you intend to communicate some sort of error condition here. And actually from reading this code, I can't tell what that condition is, so a more descriptive name would be very helpful.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -768,6 +792,31 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
             return result;
         }
 
+        private FileSystemTaskResult TryRemoveModifiedPath(string virtualPath, bool isFolder)
```
sanoursa: I would rename this to make its intent more clear. It's not meant to remove a path from ModifiedPaths. It's meant to remove it if and only if it's a newly created item, i.e. a temp file. Maybe call it `TryRemoveTempFileFromModifiedPaths`?

sanoursa: Maybe not. Now that I read its implementation, I agree with William that it's just doing too much.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
sanoursa: I see that William commented on the `Invalid` value too, and I agree, from reading this code it seems like this check maybe just doesn't belong in here, at least as currently written.

I would probably change the usage of this method to something like:
```
    if (this.IsTempFile(virtualPath))
    {
        result = this.TryRemoveModifiedPath(virtualPath, isFolder);
    }
    else
    {
        result = this.TryAddModifiedPath(virtualPath, isFolder);
    }
```

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
sanoursa: Yea I don't think we should ever ignore a `FatalError`. That said, why is `TryRemoveModifiedPath` ever returning `FatalError`s if we don't intend to shut down as a result? It feels like that method shouldn't even be a `Try` method - if we focus it down to tmp file scenarios, it can just be a best effort and not even return a result.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
sanoursa: Nevermind on creating a new value, see other comments below.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: It is still trying to write to the modified paths database file and the pattern is to retry if there was and IOException and return Fatal if there is some other exception.  To keep with the pattern I think we should shut down because we don't know what the state of the modified paths database file is in if we get a fatal error writing to it.

GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
```diff
@@ -317,24 +317,30 @@ public void MountCreatesModifiedPathsDatabase()
...
+                    "A tools/managed/v4.0/sdapi.dll",
+                    "A tools/managed/v2.0/midlpars.dll",
+                    "A tools/managed/v2.0/RPCDataSupport.dll",
+                    "A tools/managed/v2.0/MidlStaticAnalysis.dll",
+                    "A tools/perllib/MS/Somefile.txt",
+                };
+
+            modifiedPathsDatabasePath.ShouldBeAFile(this.fileSystem);
+            this.fileSystem.ReadAllText(modifiedPathsDatabasePath)
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
```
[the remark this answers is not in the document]
kewillford: Since the other places in this file are using `char` I will leave this so it matches.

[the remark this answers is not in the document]
wilbaker: Ok sounds good :+1:

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -75,38 +75,103 @@ public bool TryAdd(string path, bool isFolder, out bool isRetryable)
...
-                        metadata.Add(nameof(isFolder), isFolder);
-                        metadata.Add("Exception", e.ToString());
-                        this.Tracer.RelatedError(metadata, $"Exception caught while processing {nameof(this.TryAdd)}");
-                    }
+                    this.TraceError(isFolder, entry, e, nameof(this.TryAdd));
+                    isRetryable = false;
+                    return false;
+                }
 
+                return true;
```
sanoursa: Remove this line since the very next line is also the same :-)

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -75,38 +75,103 @@ public bool TryAdd(string path, bool isFolder, out bool isRetryable)
...
+                    return false;
+                }
+                catch (Exception e)
+                {
+                    this.TraceError(isFolder, entry, e, nameof(this.TryRemove));
                     isRetryable = false;
                     return false;
                 }
+
+                return true;
```
sanoursa: Same here, remove redundant line

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -75,38 +75,103 @@ public bool TryAdd(string path, bool isFolder, out bool isRetryable)
...
+            }
+
+            return metadata;
+        }
+
+        private IEnumerable<string> GenerateDataLines()
+        {
+            foreach (string entry in this.modifiedPaths)
+            {
+                yield return this.FormatAddLine(entry);
```
sanoursa: What is the perf overhead of the `yield return` syntax if there are many modified paths? Would it be more efficient to switch to `Select` instead? Even better, could we just pass the list of paths to the underlying method and have it write them out in a tight loop? (Assuming the perf hit is meaningful here)

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -420,6 +425,7 @@ public void OnFileConvertedToFull(string relativePath)
 
         public virtual void OnFileRenamed(string oldRelativePath, string newRelativePath)
         {
+            this.newlyCreatedFileAndFolderPaths.Add(newRelativePath);
```
sanoursa: Do we actually need this? How often does someone *rename* into a temp path? If we don't expect benefit from this, I would remove it so that we're not adding unnecessary contention on the concurrent data structure.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -768,6 +807,26 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
             return result;
         }
 
+        private FileSystemTaskResult TryRemoveModifiedPath(string virtualPath, bool isFolder)
+        {
+            string fullPathToItem = Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, virtualPath);
+            if ((isFolder && this.context.FileSystem.DirectoryExists(fullPathToItem)) ||
```
sanoursa: Nit: I would remove the redundant parens. To me, they actually make it harder to read.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford: `Select` will in effect do the same thing as here.  See [Enumerable.SelectIterator](https://github.com/Microsoft/referencesource/blob/60a4f8b853f60a424e36c7bf60f9b5b5f1973ed1/System.Core/System/Linq/Enumerable.cs#L53).  So I don't see that being different in perf.  The `PlaceholderListDatabase` is doing more work and has been fast enough.  This is essentially a tight loop since the yield return will be the `in` for the loop in `TryWriteTempFile`

```
using (Stream tempFile = this.fileSystem.OpenFileStream(this.tempFilePath, FileMode.Create, FileAccess.Write, FileShare.None, callFlushFileBuffers: true))
using (StreamWriter writer = new StreamWriter(tempFile))
{
	foreach (string line in getDataLines())
	{
		writer.Write(line + NewLine);
	}

	tempFile.Flush();
}
```

I will do some measuring and see if I need to look into this.

sanoursa: What `yield return` does is generate a class that implements `IEnumerator` and maintains and updates several fields every time you call `MoveNext`. So it's definitely not free, compared to a `foreach` over an array or list, for example. It could be that this is not a significant hit in this usage, but it would be nice to confirm.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: I will remove this for now because we aren't measuring how often this is happening so we can't measure any benefit if we don't know what or if there are current difficulties that are from this.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford: I'm not seeing any significant performance difference and couldn't find anything online with performance numbers.  I will add some tracing around this so that we can at least see the numbers if it does cause an issue.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: Removing these caused a style cop error.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -637,9 +660,38 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
 
                     break;
 
+                case FileSystemTask.OperationType.OnFilePreDelete:
```
wilbaker: Maybe add a comment in the `PreDelete` and `Delete` cases and mention that currently the code assumes that `PreDelete` and `Delete` will never both be queued up?

If in the future they are both queued, we'll want some sort of `GVFSPlatform` check (e.g. `SupportsPostDeleteNotification`) to avoid doing this work twice.

PR #210, opened 2018-08-23
Don't perform upgrade when local ring is set to None

alameenshah:
- No upgrade check when local ring is set to None.
- Cleanup - Moved common upgrade tests from Upgrade Verb and Tool to base class.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -84,10 +84,20 @@ namespace GVFS.Common
         {
             List<Release> releases;
 
-            if (this.TryLoadRingConfig(out errorMessage) && 
-                this.TryFetchReleases(out releases, out errorMessage))
+            newVersion = null;
+            if (!this.TryLoadRingConfig(out errorMessage))
             {
-                newVersion = null;
+                return false;
+            }
+
+            if (this.Ring == RingType.None)
+            {
+                errorMessage = "Upgrade ring set to None. No upgrade check was performed.";
+                return false;
+            }
+
+            if (this.TryFetchReleases(out releases, out errorMessage))
+            {
                 foreach (Release nextRelease in releases)
                 {
                     Version releaseVersion;
@@ -105,7 +115,6 @@ namespace GVFS.Common
                 return true;
             }
 
-            newVersion = null;
             return false;
         }
 
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -37,6 +37,7 @@ namespace GVFS.Common
         {
             this.installedVersion = new Version(currentVersion);
             this.fileSystem = new PhysicalFileSystem();
+            this.Ring = RingType.Invalid;
 
             string upgradesDirectoryPath = GetUpgradesDirectoryPath();
             this.fileSystem.CreateDirectory(upgradesDirectoryPath);
@@ -63,6 +64,11 @@ namespace GVFS.Common
 
         public RingType Ring { get; protected set; }
 
+        public bool IsNoneRing()
+        {
+            return this.TryLoadRingConfig(out string _) && this.Ring == RingType.None;
+        }
+
         public bool IsGVFSUpgradeRunning()
         {
             return !this.globalUpgradeLock.IsFree();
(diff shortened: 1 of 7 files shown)


GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -84,10 +84,20 @@ public bool ReleaseUpgradeLock()
...
-                this.TryFetchReleases(out releases, out errorMessage))
+            newVersion = null;
+            if (!this.TryLoadRingConfig(out errorMessage))
+            {
+                return false;
+            }
+
+            if (this.Ring == RingType.None)
+            {
+                errorMessage = "Upgrade ring set to None. No upgrade check was performed.";
```
sanoursa: Why is this treated as an error? What does the user see as a result of this?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -62,41 +62,21 @@ public void UpgradeSequence()
         }
 
         [TestCase]
-        public void InvalidUpgradeRing()
+        public override void NoneLocalRing()
```
sanoursa: Why do you have an override that does nothing but call the base method? What is the goal here?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeTests.cs
```diff
@@ -35,6 +35,63 @@ public virtual void Setup()
...
 
+        public virtual void NoneLocalRing()
+        {
+            string errorString = "Upgrade ring set to None. No upgrade check was performed.";
+            this.ConfigureRunAndVerify(
+                configure: () =>
+                {
+                    this.Upgrader.LocalRingConfig = ProductUpgrader.RingType.None;
+                },
+                expectedReturn: ReturnCode.GenericError,
```
sanoursa: Yea I don't know if this should be an error. If we think of "None" as effectively being an empty upgrade ring, then you're just always up to date and that's a success for the other two rings. @jrbriggs do you have an opinion on that?

GVFS/GVFS.Common/ProductUpgrader.cs
alameenshah: This is what the user sees -

PS M:\Work\GVFS\VFSForGit> gvfs upgrade
Checking for GVFS upgrades...Failed
Upgrade ring set to None. No upgrade check was performed.
PS M:\Work\GVFS\VFSForGit>

sanoursa: I think what I would expect is either no spinner at all and a success message about None or just being up to date; or if you want to show the spinner, then it should end in Succeeded and a similar message.

There's really no need for a spinner at all if we already know we're going to noop, so I would personally favor the first approach. This could be rolled into the prereq checks, right?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
alameenshah: As discussed, review and changes to existing TestCases can be done in a separate PR.

GVFS/GVFS.Common/ProductUpgrader.cs
alameenshah: - Spinner removed.
- This is not treated as error anymore.
- New messaging (the message text can be reviewed and updated in a future PR)

PS M:\Work\GVFS\VFSForGit> gvfs upgrade
Upgrade ring set to None. No upgrade check was performed.
PS M:\Work\GVFS\VFSForGit>

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -76,6 +76,15 @@ private bool TryRunProductUpgrade()
             string error = null;
             Version newestVersion = null;
             bool isInstallable = false;
+
+            if (this.upgrader.IsNoneRing())
+            {
+                string message = "Upgrade ring set to None. No upgrade check was performed.";
+                this.tracer.RelatedInfo($"{nameof(this.TryRunProductUpgrade)}: {message}");
```
sanoursa: `ReportInfoToConsole` (next line) will already write the message to the log file. Do we need this one as well? (I see the same applies in a few other places in this class too)

alameenshah: right its there in other places too. Will clean this up in an upcoming PR.

PR #213, opened 2018-08-23
Add Mac specific FileBasedLock implementation

### LINKED ISSUE #211: Mac: FileBasedLock does not work as expected on Mac
`FileBasedLock` relies on Windows specific behavior, and because the locking is not working as expected the functional tests are occasionally failing.

Examples:

https://mseng.visualstudio.com/VSOnline/_build/results?buildId=7054343&view=logs

https://mseng.visualstudio.com/VSOnline/VSOnline%20Team/_build/results?buildId=7047738&_a=summary&view=logs

wilbaker:
Fixes #211 

*Issue*

The previous implementation of `FileBasedLock` is that it depended on Windows specific file sharing rules\behavior.

*Changes*

- Move `FileBasedLock` to the Windows platform project and rename it to `WindowsFileBasedLock`
- Create a new `MacFileBasedLock` that uses `flock` for locking on Mac
- Add a new abstract `FileBasedLock` that is implemented by the two platform specific classes mentioned above.
- Add `CreateFileBasedLock` to the `GVFSPlatform`

Note that with this change it's the `flock` that ensures\indicates exclusive access (unlike on Windows where the lock file being present indicates ownership of the lock)


GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+
+                this.Tracer.RelatedInfo($"Lock released: '{this.LockPath}'");
+
+                this.lockAcquired = false;
+            }
+
+            if (this.lockFileDescriptor != InvalidFileDescriptor)
+            {
+                if (Close(this.lockFileDescriptor) != 0)
```
wilbaker: @sanoursa note that with the current changes the lock file is not removed from the file system.  This allows us to avoid the race conditions described here:

https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition
https://stackoverflow.com/questions/33166460/how-to-delete-a-locked-flock-file-without-race-condition-before-or-after-rele

It appears that there *are* some options for removing it if we play some tricks with `stat`, but unless you see a major issue with this approach I'm thinking we stick with this simpler approach and avoid the complexity\races of attempting to delete this file.

Note that if we were not cross platform we could `flock` the file we're interested in itself (e.g. `mapping.dat`) but to keep the cross-platform code simple and consistent it's easiest if we use a separate file on Mac like we do on Windows.

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,119 @@
...
+        public override bool TryAcquireLock()
+        {
+            if (this.lockAcquired)
+            {
+                return true;
+            }
+
+            if (this.lockFileDescriptor == InvalidFileDescriptor)
+            {
+                this.lockFileDescriptor = Creat(this.LockPath, Convert.ToInt32("644", 8));
```
wilbaker: - Use `open` instead of `creat` to make intention clearer
- Don't truncate file on open as at that point we have not acquired the `flock` yet

sanoursa: Is 644 always the right mode? What if someone has left this file behind with a different mode on it?

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,119 @@
...
+        private static extern int Creat(string pathname, int mode);
+
+        [DllImport("libc", EntryPoint = "write", SetLastError = true)]
+        private static extern long Write(int fd, byte[] buf, ulong count);
+
+        [DllImport("libc", EntryPoint = "close", SetLastError = true)]
+        private static extern int Close(int fd);
+
+        [DllImport("libc", EntryPoint = "flock", SetLastError = true)]
+        private static extern int Flock(int fd, int operation);
```
sanoursa: Nit: I would capitalize this as `FLock`

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,119 @@
...
+
+                    return false;
+                }
+            }
+
+            if (Flock(this.lockFileDescriptor, LockEx | LockNb) != 0)
+            {
+                int errno = Marshal.GetLastWin32Error();
+
+                // Log error if not EWOULDBLOCK
```
sanoursa: Is this comment accurate? I don't see an `if` in here

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (bytesWritten == -1)
+            {
+                int errno = Marshal.GetLastWin32Error();
+
+                this.Tracer.RelatedWarning($"Failed to write signature for '{this.LockPath}': {errno}");
+            }
+
+            this.Tracer.RelatedInfo($"Lock acquired for for '{this.LockPath}'");
+
+            return true;
```
sanoursa: Should we be setting `this.lockAcquired = true` here?

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public override void Dispose()
+        {
+            if (this.lockAcquired)
+            {
+                if (Flock(this.lockFileDescriptor, LockUn) != 0)
+                {
+                    this.Tracer.RelatedWarning($"Failed to release lock for: '{this.LockPath}'");
```
sanoursa: Then maybe `IDisposable` isn't the proper interface for this class. Do we need a `TryUnlock` instead so we can retry?

sanoursa: And should we avoid clearing `lockAcquired` when this happens?

GVFS/GVFS.Common/FileBasedLock.cs
```diff
@@ -1,300 +1,30 @@
...
-    public class FileBasedLock : IDisposable
+    public abstract class FileBasedLock : IDisposable
     {
-        private const int HResultErrorSharingViolation = -2147024864; // -2147024864 = 0x80070020 = ERROR_SHARING_VIOLATION
-        private const int HResultErrorFileExists = -2147024816; // -2147024816 = 0x80070050 = ERROR_FILE_EXISTS
-        private const int DefaultStreamWriterBufferSize = 1024; // Copied from: http://referencesource.microsoft.com/#mscorlib/system/io/streamwriter.cs,5516ce201dc06b5f
-        private const long InvalidFileLength = -1;
-        private const string EtwArea = nameof(FileBasedLock);
-        private static readonly Encoding UTF8NoBOM = new UTF8Encoding(false, true); // Default encoding used by StreamWriter
+        protected readonly PhysicalFileSystem FileSystem;
```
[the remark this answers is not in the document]
wilbaker: Sure I'll make the change

GVFS/GVFS.Common/FileBasedLock.cs
```diff
@@ -1,300 +1,30 @@
...
-            {
-                this.Signature = signature;
-                this.Message = message;
-            }
-
-            public string Signature { get; }
-
-            public string Message { get; }
-        }
+        public abstract void Dispose();
```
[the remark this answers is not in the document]
wilbaker: Yeah, this might be leftover from when I was moving things around, let me try removing it

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: I think this code will be clearer with the switch to `open` (`creat` is the same as `open` with the flags we're interested in here).  

The `mode` parameter is only used when creating a new file, it's ignored if the file already exists:

> The mode argument specifies the file mode bits be applied when
              a new file is created.  This argument must be supplied when
              O_CREAT or O_TMPFILE is specified in flags; if neither O_CREAT
              nor O_TMPFILE is specified, then mode is ignored.  The
              effective mode is modified by the process's umask in the usual
              way: in the absence of a default ACL, the mode of the created
              file is (mode & ~umask).  Note that this mode applies only to
              future accesses of the newly created file;

http://man7.org/linux/man-pages/man2/open.2.html

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: This is still WIP, for now I'm logging everything just to make sure things are working as expected

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: Yes we should, good catch :)

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: :+1:

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: Reading through the comments for `flock` the lock will be released automatically when the file descriptor is closed, and so strictly speaking I don't think we need this call in the first place:

> Locks created by flock() are associated with an open file description
       (see open(2)).  This means that duplicate file descriptors (created
       by, for example, fork(2) or dup(2)) refer to the same lock, and this
       lock may be modified or released using any of these file descriptors.
       Furthermore, the lock is released either by an explicit LOCK_UN
       operation on any of these duplicate file descriptors, or when all
       such file descriptors have been closed.

http://man7.org/linux/man-pages/man2/flock.2.html

(Note also that all the error values described for `flock` only apply when acquiring the lock and not releasing it)

wilbaker: Note also that if `close` fails it should not be retried, as the `fd` has been successfully released (any errors are just for diagnostic purposes):

> A careful programmer will check the return value of close(), since it
       is quite possible that errors on a previous write(2) operation are
       reported only on the final close() that releases the open file
       description.  Failing to check the return value when closing a file
       may lead to silent loss of data.  This can especially be observed
       with NFS and with disk quota.
>
>Note, however, that a failure return should be used only for
       diagnostic purposes (i.e., a warning to the application that there
       may still be I/O pending or there may have been failed I/O) or
       remedial purposes (e.g., writing the file once more or creating a
       backup).
>
>  Retrying the close() after a failure return is the wrong thing to do,
       since this may cause a reused file descriptor from another thread to
       be closed.  This can occur because the Linux kernel always releases
       the file descriptor early in the close operation, freeing it for
       reuse; the steps that may return an error, such as flushing data to
       the filesystem or device, occur only later in the close operation.

http://man7.org/linux/man-pages/man2/close.2.html

GVFS/GVFS.Common/FileBasedLock.cs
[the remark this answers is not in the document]
wilbaker: It ends up the abstract `Dispose` is required, without it we get the error:

`FileBasedLock.cs(7,43): error CS0535: 'FileBasedLock' does not implement interface member 'IDisposable.Dispose()' [/Users/wilbaker/Repos/VFSForGit/src/GVFS/GVFS.Common/GVFS.Common.csproj]`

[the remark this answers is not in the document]
sanoursa: Yea you're right. I must be spending too much time in C++ these days :-)

GVFS/GVFS.UnitTests.Windows/Windows/WindowsFileBasedLockTests.cs
```diff
@@ -1,14 +1,15 @@
...
 
-namespace GVFS.UnitTests.Common
+namespace GVFS.UnitTests.Windows
 {
     [TestFixture]
-    public class FileBasedLockTests
+    public class WindowsFileBasedLockTests
     {
         [TestCase]
         public void CreateLockWhenDirectoryMissing()
```
wilbaker: Check if this scenario works on Mac

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,159 @@
...
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace GVFS.Platform.Mac
+{
+    public class MacFileBasedLock : FileBasedLock
+    {
+        // #define O_WRONLY    0x0001      /* open for writing only */
+        private const int OpenWriteOnly = 0x0001;
```
sanoursa: Nit: since you created a `NativeMethods` subclass (which I like), I would also move all of these constants inside there since that is the only context in which the constants are needed

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,159 @@
...
+            ITracer tracer,
+            string lockPath,
+            string signature)
+            : base(fileSystem, tracer, lockPath, signature) 
+        {
+            this.lockFileDescriptor = InvalidFileDescriptor;
+        }
+
+        [Flags]
+        private enum FLockOperations
```
sanoursa: Also a nit: you're currently using constants for some and an enum for others. I would keep them consistent, and the constants seem to make the usage less cluttered (no need for a cast)

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,159 @@
...
+            {
+                byte[] signatureBytes = Encoding.UTF8.GetBytes(this.Signature);
+                long bytesWritten = NativeMethods.Write(
+                    this.lockFileDescriptor,
+                    signatureBytes,
+                    Convert.ToUInt64(signatureBytes.Length));
+
+                if (bytesWritten == -1)
+                {
+                    int errno = Marshal.GetLastWin32Error();
```
sanoursa: Why are we still returning true if we hit this, or the other error right below it? Should we fail here?

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -0,0 +1,159 @@
...
+            return true;
+        }
+
+        public override void Dispose()
+        {
+            if (this.lockFileDescriptor != InvalidFileDescriptor)
+            {
+                if (NativeMethods.Close(this.lockFileDescriptor) != 0)
+                {
+                    int errno = Marshal.GetLastWin32Error();
```
sanoursa: I would suggest leaving a comment here that an error from close is advisory only

GVFS/GVFS.Platform.Windows/WindowsFileBasedLock.cs
```diff
@@ -0,0 +1,157 @@
...
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.ComponentModel;
+using System.IO;
+using System.Text;
+
+namespace GVFS.Platform.Windows
+{
+    public class WindowsFileBasedLock : FileBasedLock
```
sanoursa: I'm guessing there are no meaningful changes in this code compared to the old `FileBasedLock`?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -331,6 +331,7 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
         goto CleanupAndFail;
     }
     
+    // TODO(Mac): Only call chmod is fileMode is different than the default file mode
```
sanoursa: Good call

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: Thats a good idea, I'll move these

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: I'll make them all contants

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: > Why are we still returning true if we hit this

We're returning true because `FLock` successfully acquired the lock

> or the other error right below it? Should we fail here?

I don't think that we need to.  Currently the signature is only to help with diagnostics, it need not be written for the locking to work correctly.  (I also considered not writing the signature at all, it would speed things up and eliminate a class of possible failures, thoughts on that option?)

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
wilbaker: :+1:

GVFS/GVFS.Platform.Windows/WindowsFileBasedLock.cs
wilbaker: There were a lot of changes to this file to remove dead code and clean up the interface.

If you'd like to see the changes you can check out the first commit:

https://github.com/Microsoft/VFSForGit/pull/213/commits/2ab1b2b6d7275e0a7af42a0bf8e2bf158fb52d7f

I did not perform the file rename in that commit to make it easier to see the diffs in this file (see GVFS/GVFS.Common/FileBasedLock.cs in the above commit)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I see I made a typo in this comment, I'll fix that up

GVFS/GVFS.Platform.Windows/WindowsFileBasedLock.cs
wilbaker: (Note that these changes pre-dated the creation of the abstract FileBasedLock, and so the private members were not yet moved out)

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
sanoursa: Yea that's a good point. The reason we added the signature in the first place is that we were also creating the index.lock from the mount process. It was useful to know, for diagnostics, if an abandoned index.lock had been created by git or by us. Now that we no longer create the index.lock... is the signature adding any value at all? Even on Windows? We could just drop it if it's not serving a purpose any longer.

GVFS/GVFS.Platform.Windows/WindowsFileBasedLock.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        public override bool TryAcquireLock()
+        {
+            try
+            {
+                lock (this.deleteOnCloseStreamLock)
+                {
+                    if (this.deleteOnCloseStream != null)
+                    {
+                        throw new InvalidOperationException("Lock has already been acquired");
```
wilbaker: @sanoursa let me know what you think of this change (to throw an exception) in my latest commit.

I made this change (rather than returning `true`) to help enforce the rule that `TryAcquireLock` should not be called on a `FileBasedLock` after it has successfully acquired the lock.

sanoursa: Throwing `InvalidOperationException` makes perfect sense here. A re-entrant usage of this lock violates its contract and could result in other bugs if we don't make that obvious.

PR #214, opened 2018-08-24
Remove temporary directories and files after upgrade

alameenshah:
- Delete ProgramData/GVFS.Upgrade directory when GVFS is uninstalled.
- Delete downloaded upgrade installer files after upgrade is finished.
- Updated UT.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -224,12 +224,13 @@ namespace GVFS.Common
                 Exception exception;
                 if (!this.TryDeleteDownloadedAsset(asset, out exception))
                 {
-                    error += $"Could not delete {asset.LocalPath}. {exception.ToString()}.";
+                    error += $"Could not delete {asset.LocalPath}. {exception.ToString()}." + Environment.NewLine;
                 }
             }
 
             if (!string.IsNullOrEmpty(error))
             {
+                error.TrimEnd(Environment.NewLine.ToCharArray());
                 return false;
             }
 
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -216,6 +216,35 @@ namespace GVFS.Common
             return false;
         }
 
+        public bool TryCleanup(out string error)
+        {
+            bool isError = false;
+            string exceptionInfo = string.Empty;
+            foreach (Asset asset in this.newestRelease.Assets)
+            {
+                Exception exception;
+                if (!this.TryDeleteDownloadedAsset(asset, out exception))
+                {
+                    exceptionInfo += $"Could not delete {asset.LocalPath}. {exception.ToString()}.";
+                    isError = true;
+                }
+            }
+
+            if (isError)
+            {
+                error = exceptionInfo;
+                return false;
+            }
+
+            error = null;
+            return true;
+        }
+
+        protected virtual bool TryDeleteDownloadedAsset(Asset asset, out Exception exception)
+        {
+            return this.fileSystem.TryDeleteFile(asset.LocalPath, out exception);
+        }
+
         protected virtual bool TryLoadRingConfig(out string error)
         {
             string errorAdvisory = "Please run 'git config --system gvfs.upgrade-ring [\"Fast\"|\"Slow\"|\"None\"]' and retry.";
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -218,21 +218,18 @@ namespace GVFS.Common
 
         public bool TryCleanup(out string error)
         {
-            bool isError = false;
-            string exceptionInfo = string.Empty;
+            error = string.Empty;
             foreach (Asset asset in this.newestRelease.Assets)
             {
                 Exception exception;
                 if (!this.TryDeleteDownloadedAsset(asset, out exception))
                 {
-                    exceptionInfo += $"Could not delete {asset.LocalPath}. {exception.ToString()}.";
-                    isError = true;
+                    error += $"Could not delete {asset.LocalPath}. {exception.ToString()}.";
                 }
             }
 
-            if (isError)
+            if (!string.IsNullOrEmpty(error))
             {
-                error = exceptionInfo;
                 return false;
             }
 
--- a/GVFS/GVFS.Installer/Setup.iss
+++ b/GVFS/GVFS.Installer/Setup.iss
@@ -165,6 +165,7 @@ DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe";
 [UninstallDelete]
 ; Deletes the entire installation directory, including files and subdirectories
 Type: filesandordirs; Name: "{app}";
+Type: filesandordirs; Name: "{sd}\ProgramData\GVFS\GVFS.Upgrade";
 
 [Registry]
 Root: HKLM; Subkey: "{#EnvironmentKey}"; \
--- a/GVFS/GVFS.Installer/Setup.iss
+++ b/GVFS/GVFS.Installer/Setup.iss
@@ -165,7 +165,7 @@ DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe";
 [UninstallDelete]
 ; Deletes the entire installation directory, including files and subdirectories
 Type: filesandordirs; Name: "{app}";
-Type: filesandordirs; Name: "{sd}\ProgramData\GVFS\GVFS.Upgrade";
+Type: filesandordirs; Name: "{commonappdata}\GVFS\GVFS.Upgrade";
 
 [Registry]
 Root: HKLM; Subkey: "{#EnvironmentKey}"; \
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockProductUpgrader.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockProductUpgrader.cs
@@ -32,7 +32,9 @@ namespace GVFS.UnitTests.Windows.Mock.Upgrader
             GitDownload = 0x4,
             GVFSDownload = 0x8,
             GitInstall = 0x10,
-            GVFSInstall = 0x20
+            GVFSInstall = 0x20,
+            GVFSCleanup = 0x40,
+            GitCleanup = 0x80
         }
 
         public RingType LocalRingConfig { get; set; }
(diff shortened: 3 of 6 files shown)


GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -216,6 +216,35 @@ public virtual bool TryCreateToolsDirectory(out string upgraderToolPath, out str
...
+        public bool TryCleanup(out string error)
+        {
+            bool isError = false;
+            string exceptionInfo = string.Empty;
+            foreach (Asset asset in this.newestRelease.Assets)
+            {
+                Exception exception;
+                if (!this.TryDeleteDownloadedAsset(asset, out exception))
+                {
+                    exceptionInfo += $"Could not delete {asset.LocalPath}. {exception.ToString()}.";
```
sanoursa: Add a newline at the end?

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -216,6 +216,35 @@ public virtual bool TryCreateToolsDirectory(out string upgraderToolPath, out str
...
+                    isError = true;
+                }
+            }
+
+            if (isError)
```
sanoursa: Do you need a second variable for this or can you just check if `exceptionInfo` is non-empty?

GVFS/GVFS.Installer/Setup.iss
```diff
@@ -165,6 +165,7 @@ DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe";
 [UninstallDelete]
 ; Deletes the entire installation directory, including files and subdirectories
 Type: filesandordirs; Name: "{app}";
+Type: filesandordirs; Name: "{sd}\ProgramData\GVFS\GVFS.Upgrade";
```
sanoursa: I'm curious, what does `{sd}` mean?

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockProductUpgrader.cs
```diff
@@ -32,7 +32,9 @@ public enum ActionType
             GitDownload = 0x4,
             GVFSDownload = 0x8,
             GitInstall = 0x10,
-            GVFSInstall = 0x20
+            GVFSInstall = 0x20,
+            GVFSCleanup = 0x40,
+            GitCleanup = 0x80
```
sanoursa: Nit: a trailing comma will make future diffs cleaner (e.g. see the `GVFSInstall` line, that would have been white if it had a trailing comma in the original change)

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -198,28 +201,43 @@ private bool TryRunCleanUp(out string error)
...
 
-                    return true;
+                    string downloadsCleanupError;
+                    if (this.deleteDownloadedAssets && !this.upgrader.TryCleanup(out downloadsCleanupError))
+                    {
+                        EventMetadata metadata = new EventMetadata();
+                        metadata.Add("Upgrade Step", nameof(this.TryRunCleanUp));
+                        this.tracer.RelatedError(metadata, $"{nameof(this.upgrader.TryCleanup)} failed. {downloadsCleanupError}");
+                        errorMessage += downloadsCleanupError + Environment.NewLine;
+                        success = false;
```
sanoursa: Why do we need to fail the upgrade (in the UX) if this step fails? The upgrade has actually completed successfully right? And running upgrade again will just tell them they're caught up. If that's true, then this is not an error.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -198,28 +201,43 @@ private bool TryRunCleanUp(out string error)
...
-                        this.tracer.RelatedError(metadata, $"{nameof(this.preRunChecker.TryMountAllGVFSRepos)} failed. {errorMessage}");
-                        return false;
+                        this.tracer.RelatedError(metadata, $"{nameof(this.preRunChecker.TryMountAllGVFSRepos)} failed. {remountError}");
+                        errorMessage += remountError + Environment.NewLine;
+                        success = false;
                     }
 
-                    return true;
+                    string downloadsCleanupError;
+                    if (this.deleteDownloadedAssets && !this.upgrader.TryCleanup(out downloadsCleanupError))
```
sanoursa: Is there anything that will come back through and delete these files if this step fails? What is the actual goal here? Is it just a best-effort cleanup or do we care about making it reliable?

GVFS/GVFS.Common/ProductUpgrader.cs
wilbaker: Related: do you need exceptionInfo, or can use just use `error`?

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -198,28 +201,43 @@ private bool TryRunCleanUp(out string error)
...
 
-                    if (this.remount && !this.preRunChecker.TryMountAllGVFSRepos(out errorMessage))
+                    string remountError;
+                    if (this.remount && !this.preRunChecker.TryMountAllGVFSRepos(out remountError))
                     {
                         EventMetadata metadata = new EventMetadata();
                         metadata.Add("Upgrade Step", nameof(this.TryRunCleanUp));
-                        this.tracer.RelatedError(metadata, $"{nameof(this.preRunChecker.TryMountAllGVFSRepos)} failed. {errorMessage}");
-                        return false;
+                        this.tracer.RelatedError(metadata, $"{nameof(this.preRunChecker.TryMountAllGVFSRepos)} failed. {remountError}");
```
wilbaker: I recommend adding `remountError` to `metadata` rather than including it in the error message, it will make the error easier to query in Kusto.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -198,28 +201,43 @@ private bool TryRunCleanUp(out string error)
...
+                        success = false;
                     }
 
-                    return true;
+                    string downloadsCleanupError;
+                    if (this.deleteDownloadedAssets && !this.upgrader.TryCleanup(out downloadsCleanupError))
+                    {
+                        EventMetadata metadata = new EventMetadata();
+                        metadata.Add("Upgrade Step", nameof(this.TryRunCleanUp));
+                        this.tracer.RelatedError(metadata, $"{nameof(this.upgrader.TryCleanup)} failed. {downloadsCleanupError}");
```
wilbaker: Same comment here, add `downloadsCleanupError` to `metadata`

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -198,28 +201,43 @@ private bool TryRunCleanUp(out string error)
...
+                        errorMessage += downloadsCleanupError + Environment.NewLine;
+                        success = false;
+                    }
+
+                    return success;
                 },
                 "Finishing upgrade"))
             {
-                error = errorMessage;
+                error = errorMessage.TrimEnd(Environment.NewLine.ToCharArray());
```
wilbaker: What was the issue with leaving a trailing NewLine?

GVFS/GVFS.Installer/Setup.iss
jrbriggs: System drive.

I think we should use `{commonappdata}` instead based on [this SO post](https://stackoverflow.com/questions/39655574/inno-setup-constant-for-programdata-directory?rq=1).

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: - GVFS Un-installer removes downloaded assets that are left behind, but otherwise the assets will remain there when delete fails. 
- The goal is to try to throw away downloaded assets after use, so they don't pile up.
- Reliability - We don't want unwanted disk usage. Upgrader tool runs elevated, so I'm not expecting this delete to fail. The case where user terminates the tool before it gets to do the cleanup is not handled now.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: It leaves a blank line between the last error and "finished upgrade" message on the Console.

GVFS/GVFS.Common/ProductUpgrader.cs
alameenshah: UpgradeOrchestrator does that. It adds a newline when the message goes to Console, and Tracer adds the line when logging to file.

GVFS/GVFS.Common/ProductUpgrader.cs
alameenshah: I can use just the error.

GVFS/GVFS.Common/ProductUpgrader.cs
sanoursa: But this code is in a loop here. If you have multiple exceptions, you're going to end up with a run-on string and no breaks between them.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -216,6 +216,32 @@ public virtual bool TryCreateToolsDirectory(out string upgraderToolPath, out str
...
 
+        public bool TryCleanup(out string error)
+        {
+            error = string.Empty;
+            foreach (Asset asset in this.newestRelease.Assets)
+            {
+                Exception exception;
+                if (!this.TryDeleteDownloadedAsset(asset, out exception))
+                {
+                    error += $"Could not delete {asset.LocalPath}. {exception.ToString()}.";
```
derrickstolee: Could you add a newline here so these sentences are split better?

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -216,6 +216,32 @@ public virtual bool TryCreateToolsDirectory(out string upgraderToolPath, out str
...
+            if (!string.IsNullOrEmpty(error))
+            {
+                return false;
+            }
+
+            error = null;
+            return true;
+        }
+
+        protected virtual bool TryDeleteDownloadedAsset(Asset asset, out Exception exception)
```
derrickstolee: Are you expecting this method to be more complicated in the future? Why not just call `TryDeleteFile` directly?

derrickstolee: (nevermind, I see you are overriding it in `MockProductUpgrader`)

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
derrickstolee: Hm. Is that a problem? I would almost want to have some split between the error message list and the "finished upgrade" message.

PR #216, opened 2018-08-27
For Peer Review: Atomic Placeholder spec

nickgra:
Please leave comments on the .md file here. Once we've reach consensus on a design, I will abandon this PR and copy the text into a GitHub Issue so we can keep track of our design documents going forward.

### DIFF
--- a/Atomic.md
+++ b/Atomic.md
@@ -11,7 +11,7 @@ level, we go through the following workflow:
 
 -   (If creating a file) Update the mode of the file
 
-This could result in scenarios where a partial placeholder appears in
+This could result in scenarios where a incomplete placeholder appears in
 the virtualization root. The full set of failure modes is documented
 below:
 
@@ -22,14 +22,14 @@ WritePlaceholderDirectory:
     -   The enumeration fails as a directory wasn't created. This should
          be retriable as a second attempt should see the IsEmpty bit on
          the parent folder and call WritePlaceholderDirectory again. No
-         partial placeholders were left in the parent directory.
+         incomplete placeholders were left in the parent directory.
 
 -   We crash between mkdir() and InitializeEmptyPlaceholder
 
-    -   Placeholder folder appears to be not in virtualization root and
-         not empty.
+    -   Placeholder folder does not have flags for virtualization root and
+        empty.
 
-    -   This folder will never be enumerated as the kext will exit when
+    -   This folder will never be expanded as the kext will exit when
          we do not see the InVirtualizationRoot bit
 
 -   One of the SetBitInFileFlags calls fails in
@@ -37,12 +37,12 @@ WritePlaceholderDirectory:
 
     -   Folder is marked as in virtualization root but not empty.
 
-        -   This folder will never be enumerated as the kext will exit
+        -   This folder will never be expanded as the kext will exit
              when we do not see the IsEmpty bit
 
     -   Folder is marked as empty but not in virtualization root
 
-        -   This folder will never be enumerated as the kext will exit
+        -   This folder will never be expanded as the kext will exit
              when we do not see the InVirtualizationRoot bit (we'll
              never reach the IsEmpty test)
 
@@ -52,13 +52,13 @@ WritePlaceholderFile:
 
     -   This should be retriable as a second attempt should see the
          IsEmpty bit on the parent folder and call WritePlaceholderFile
-         again. No partial placeholders were left in the parent
+         again. No incomplete placeholders were left in the parent
          directory.
 
 -   Ftruncate fails to expand the file to the correct size (or is never
      reached because of a crash)
 
-    -   We have a partial placeholder file in the virtualization root
+    -   We have a incomplete placeholder file in the virtualization root
          that cannot be recovered. Future attempts of
          WritePlaceholderFile will fail because fopen(path, "wbx")
          won't let us overwrite the existing file.
@@ -66,11 +66,11 @@ WritePlaceholderFile:
 -   Memcpy'ing the providerID xattr fails (or is never reached because
      of a crash)
 
-    -   We have a partial placeholder file in the virtualization root
+    -   We have a incomplete placeholder file in the virtualization root
         that cannot be recovered. There are no flags set on this file.
 
     -   The MirrorProvider doesn't currently use the providerID for
-         anything (although it should fail if its empty)
+         anything (although it should fail if it's empty)
 
     -   VFSForGit will be unable to parse the placeholder version from
          the providerID, OnGetFileStream will fail so hydrating this
@@ -79,7 +79,7 @@ WritePlaceholderFile:
 -   Memcpy'ing the contentID xattr fails (or is never reached because of
      a crash)
 
-    -   We have a partial placeholder file in the virtualization root
+    -   We have a incomplete placeholder file in the virtualization root
         that cannot be recovered. There are no flags set on this file.
 
 
@@ -91,10 +91,10 @@ WritePlaceholderFile:
 
 -   We crash before reaching InitializeEmptyPlaceholder
 
-    -   Placeholder appears to be not in virtualization root and not
-         empty.
+    -   Placeholder does not have flags for virtualization root and
+        empty.
 
-    -   This file will never be enumerated as the kext will exit when we
+    -   This file will never be expanded as the kext will exit when we
          do not see the InVirtualizationRoot bit
 
 -   One of the SetBitInFileFlags calls fails in
@@ -132,7 +132,7 @@ provider's folder (.gvfs/.mirror), marks them up and then uses rename()
 to move them into the virtualization root.
 
 A sample of this change is implemented here (user mode component only):
-<https://github.com/nickgra/VFSForGit/tree/atomic
+https://github.com/nickgra/VFSForGit/tree/atomic
 
 Currently, the provider registers its virtualization root with the kext
 at clone time. We have code today that will log on files detected
@@ -167,12 +167,12 @@ WritePlaceholderDirectory:
 
     -   Folder is marked as in virtualization root but not empty.
 
-        -   A partial placeholder is in .temp that can be safely
+        -   A incomplete placeholder is in .temp that can be safely
             overwritten/deleted when the enumeration is retried
 
     -   Folder is marked as empty but not in virtualization root
 
-        -   A partial placeholder is in .temp that can be safely
+        -   A incomplete placeholder is in .temp that can be safely
             overwritten/deleted when the enumeration is retried
 
     -   The rename() to copy the completed placeholder from .temp into
@@ -198,13 +198,13 @@ WritePlaceholderFile:
 -   Memcpy'ing the providerID xattr fails (or is never reached because
      of a crash)
 
-    -   We have a partial placeholder file in .temp that will be
+    -   We have a incomplete placeholder file in .temp that will be
         overwritten by a retry.
 
 -   Memcpy'ing the contentID xattr fails (or is never reached because of
      a crash)
 
-    -   We have a partial placeholder file in .temp that will be
+    -   We have a incomplete placeholder file in .temp that will be
         overwritten by a retry.
 
 -   We crash before reaching InitializeEmptyPlaceholder
@@ -217,12 +217,12 @@ WritePlaceholderFile:
 
     -   Folder is marked as in virtualization root but not empty.
 
-        -   A partial placeholder is in .temp that can be safely
+        -   A incomplete placeholder is in .temp that can be safely
             overwritten/deleted when the enumeration is retried
 
     -   Folder is marked as empty but not in virtualization root
 
-        -   A partial placeholder is in .temp that can be safely
+        -   A incomplete placeholder is in .temp that can be safely
             overwritten/deleted when the enumeration is retried
 
 -   Chmod() fails or is never reached due to a crash
@@ -236,6 +236,12 @@ WritePlaceholderFile:
         -   We can retry from this error since nothing ever appeared in
             the virtualization root. The placeholder will be overwritten
             as part of the retry
+
+With this approach, each placeholder file that will be created will be moved in one at a time to their final location rather than moving all placeholders that may be present in a folder at once. 
+
+In a case where we crash (or the machine crashes, or the user kills the mount process) while expanding a placeholder directory, we will be able to recover by replacing all placeholders that are present in that folder and continuing the enumeration. We will not try to short circuit this process and restart the enumeration from where we crashed. Any user content that is present in this directory will not be overwritten by the directory enumeration.
+
+Designs where we moved an entire folder's worth of placeholders were considered, but were tabled in favor of this approach (making any expansion/hydration retriable).
 
 All of the potential holes identified above are filled by this design,
 so I believe this to be our best bet to create placeholders in an atomic
@@ -268,3 +274,5 @@ Miscellanea/Things to Validate
 
 -   All of the notification work will need to be validated under the new
     hydration model to ensure that behavior will not be changed.
+
+-  Do we need to ensure that that flags and xattrs are flushed to disk before renaming the placeholder?
--- a/Atomic.md
+++ b/Atomic.md
@@ -37,13 +37,13 @@ WritePlaceholderDirectory:
 
 WritePlaceholderFile:
 
--   Fopen() fails to create a file
+-   fopen() fails to create a file
 
     -   This is retriable today as a second attempt should see the
          IsEmpty bit on the parent directory and call WritePlaceholderFile
          again. 
 
--   Ftruncate fails to expand the file to the correct size (or is never
+-   ftruncate() fails to expand the file to the correct size (or is never
      reached because of a crash)
 
     -   We have an incomplete placeholder file in the virtualization root
@@ -90,7 +90,7 @@ WritePlaceholderFile:
     -   This file will never be expanded as the kext will exit
             when we do not see the IsEmpty bit
 
--   Chmod() fails or is never reached due to a crash
+-   chmod() fails or is never reached due to a crash
 
     -   The default mode should be 755 (unless a user has played with
          umask, but this hole is out of scope for this design) so
@@ -107,14 +107,13 @@ Proposed Solution:
 ------------------
 
 This solution creates placeholders in a .staging directory within the
-provider's directory (.gvfs/.mirror), marks them up and then uses rename()
+provider's directory (.gvfs or .mirror), marks them up and then uses rename()
 to move them into the virtualization root.
 
-A sample of this change is implemented here (user mode component only):
-https://github.com/nickgra/VFSForGit/tree/atomic
+In order to protect this directory from user tampering, we will have the kext block accesses to this directory that are not coming from the special identity that the mount process will use (this will be described in a separate design document).
 
 Currently, the provider registers its virtualization root with the kext
-at clone time. We have code today that will log on files detected
+at mount time. We have code today that will log on files detected
 outside of the virtualization root, which could be helpful for
 diagnosing customer issues.
 
@@ -163,12 +162,12 @@ WritePlaceholderDirectory:
 
 WritePlaceholderFile:
 
--   Fopen() fails to create a file
+-   fopen() fails to create a file
 
     -   We can simply retry and have this succeed. Nothing was created
         in .staging
 
--   Ftruncate fails to expand the file to the correct size (or is never
+-   ftruncate() fails to expand the file to the correct size (or is never
      reached because of a crash)
 
     -   We can simply retry and have this succeed. The file written in
@@ -199,7 +198,7 @@ WritePlaceholderFile:
         -   An incomplete placeholder is in .staging that can be safely
             overwritten/deleted when the enumeration is retried
 
--   Chmod() fails or is never reached due to a crash
+-   chmod() fails or is never reached due to a crash
 
     -   We can retry and have this succeed as the placeholder hasn't
         left .staging yet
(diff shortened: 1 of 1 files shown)


Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+placeholders. Today, when creating a placeholder directory, at a high
+level, we go through the following workflow:
+
+-   Create placeholder folder/file in the virtualization root
+
+-   'Mark up' the file with the proper flags/xattrs
+
+-   (If creating a file) Update the mode of the file
+
+This could result in scenarios where a partial placeholder appears in
```
wilbaker: What do you mean by "partial placeholder"?  If you mean a placeholder that is missing some/all of its metadata, consider using a term like "incomplete placeholder" to avoid confusion with the "partial files" term on Windows.

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+-   mkdir() fails to create the placeholder directory
+
+    -   The enumeration fails as a directory wasn't created. This should
+         be retriable as a second attempt should see the IsEmpty bit on
+         the parent folder and call WritePlaceholderDirectory again. No
+         partial placeholders were left in the parent directory.
+
+-   We crash between mkdir() and InitializeEmptyPlaceholder
+
+    -   Placeholder folder appears to be not in virtualization root and
```
wilbaker: > appears to be not in virtualization root

To confirm, in today's implementation the directory would not be flagged as inside the root?

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+         not empty.
+
+    -   This folder will never be enumerated as the kext will exit when
```
wilbaker: Nit: "never be enumerated" -> "never be expanded"

It can still be enumerated, it will just appear empty

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+
+-   One of the SetBitInFileFlags calls fails in
+     InitializeEmptyPlaceholder
+
+    -   Folder is marked as in virtualization root but not empty.
+
+        -   This folder will never be enumerated as the kext will exit
+             when we do not see the IsEmpty bit
+
+    -   Folder is marked as empty but not in virtualization root
```
wilbaker: Is it possible to have any combination of failing to set the flags today?  Or will VFSForGit bail out as soon as it fails to set a flag?

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+         won't let us overwrite the existing file.
+
+-   Memcpy'ing the providerID xattr fails (or is never reached because
+     of a crash)
+
+    -   We have a partial placeholder file in the virtualization root
+        that cannot be recovered. There are no flags set on this file.
+
+    -   The MirrorProvider doesn't currently use the providerID for
+         anything (although it should fail if its empty)
```
wilbaker: Nit: `it's`

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+    -   VFSForGit will be unable to get the SHA of the object that it
+         needs to download and hydrating this file will fail. Retrying
+         will not help.
+
+-   We crash before reaching InitializeEmptyPlaceholder
+
+    -   Placeholder appears to be not in virtualization root and not
+         empty.
+
+    -   This file will never be enumerated as the kext will exit when we
```
wilbaker: What do you mean by "This file will never be enumerated"?

It's on disk and so it will appear in enumerations.

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+
+-   One of the SetBitInFileFlags calls fails in
+     InitializeEmptyPlaceholder
+
+    -   Placeholder is marked as in virtualization root but not empty.
+
+        -   This file will never be hydrated as the kext will exit when
+             we do not see the IsEmpty bit
+
+    -   Placeholder is marked as empty but not in virtualization root
```
wilbaker: How can this happen?  If VFSForGit fails to mark the placeholder as inside the root does it continue onto the next (is empty) bit?

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+    -   The rename() to copy the completed placeholder from .temp into
+        the virtualization root fails
+
+        -   We can retry from this error since nothing ever appeared in
+            the virtualization root. The placeholder will be overwritten
+            as part of the retry
+
+All of the potential holes identified above are filled by this design,
+so I believe this to be our best bet to create placeholders in an atomic
+matter.
```
wilbaker: Based on my reading of the design, it sounds like placeholders are created in `.temp` and then moved into their final location one at a time.  What happens if VFSForGit is killed in the middle of this process?  Can it resume where it left off in the next enumeration?  How will this work?

Atomic.md
```diff
@@ -0,0 +1,270 @@
...
+
+Miscellanea/Things to Validate
```
wilbaker: Something else to validate:  Do we need to ensure that that flags and xattrs are flushed to disk before renaming the placeholder?

Atomic.md
nickgra: It will bail out today as soon as we fail to set the flag. We call SetBitInFileFlags in InitializeEmptyPlaceholder

```     
if (!InitializeEmptyPlaceholder(fullPath))
    {
        goto CleanupAndFail;
    }
```

nickgra: It will bail out today as soon as we fail to set the flag. We call SetBitInFileFlags in InitializeEmptyPlaceholder

```     
if (!InitializeEmptyPlaceholder(fullPath))
    {
        goto CleanupAndFail;
    }
```

nickgra: It will bail out today as soon as we fail to set the flag. We call SetBitInFileFlags in InitializeEmptyPlaceholder

```     
if (!InitializeEmptyPlaceholder(fullPath))
    {
        goto CleanupAndFail;
    }
```

nickgra: It will bail out today as soon as we fail to set the flag. We call SetBitInFileFlags in InitializeEmptyPlaceholder

```     
if (!InitializeEmptyPlaceholder(fullPath))
    {
        goto CleanupAndFail;
    }
```

Atomic.md
nickgra: This is referring to expansion, not enumeration. Thanks for correcting me on this earlier.

nickgra: This is referring to expansion, not enumeration. Thanks for correcting me on this earlier.

nickgra: This is referring to expansion, not enumeration. Thanks for correcting me on this earlier.

Atomic.md
nickgra: It should not be possible, but I wanted to call it out either way. Perhaps it's clutter and should be cut then?

Atomic.md
nickgra: That is the correct reading of the document, I'll explicitly call that out here. 

We will replace all placeholders that are present in this folder (the scenario I had in mind was VFSForGit crashes/is killed, on re-mounting, the customer uses this time to checkout a new branch before retrying their enumeration to ensure they get stamped with the right contentID) with new ones rather than trying to optimize. We will NOT touch files that don't belong to VFSForGit (if a user goes to that folder without the kext loaded and starts creating new files).

Atomic.md
nickgra: It will bail out today as soon as we fail to set the flag. We call SetBitInFileFlags in InitializeEmptyPlaceholder

```     
if (!InitializeEmptyPlaceholder(fullPath))
    {
        goto CleanupAndFail;
    }
```

Atomic.md
wilbaker: > It should not be possible, but I wanted to call it out either way. Perhaps it's clutter and should be cut then?

My vote is to cut it.  There's already a lot of failure modes to take in, and it might make it easier to reason about the best fix if we don't also consider the impossible scenarios.

(I think it *is* possible for a file to end up like this from a user mucking with it, but that's unrelated to the focus of your document)

Atomic.md
```diff
@@ -0,0 +1,278 @@
+Atomic placeholder operations in VFSForGit on MacOS
+===================================================
+
+A potential hole in our current design is in our creation of
+placeholders. Today, when creating a placeholder directory, at a high
```
wilbaker: Minor comment (unrelated to the proposed fix):

> when creating a placeholder directory

I wasn't sure when I read this if "creating a placeholder directory" meant expanding a directory, or writing a placeholder for a directory.

The bullet points are not directory specific, and so you might want to simplify this to "when creating a placeholder"

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+level, we go through the following workflow:
+
+-   Create placeholder folder/file in the virtualization root
```
wilbaker: > placeholder folder/file

Nit: Everywhere else you use the term "directory" rather than "folder".  Consider doing the same here for consistency.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+
+-   'Mark up' the file with the proper flags/xattrs
+
+-   (If creating a file) Update the mode of the file
+
+This could result in scenarios where a incomplete placeholder appears in
```
wilbaker: Nit: Should be "an incomplete"

Atomic.md
sanoursa: Correct, this is an issue about creating any placeholder. Not specific to placeholder directories.

Atomic.md
```diff
@@ -0,0 +1,278 @@
+Atomic placeholder operations in VFSForGit on MacOS
+===================================================
+
+A potential hole in our current design is in our creation of
```
sanoursa: It's a definite hole, not potential :-)

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+below:
+
+WritePlaceholderDirectory:
+
+-   mkdir() fails to create the placeholder directory
+
+    -   The enumeration fails as a directory wasn't created. This should
+         be retriable as a second attempt should see the IsEmpty bit on
+         the parent folder and call WritePlaceholderDirectory again. No
+         incomplete placeholders were left in the parent directory.
```
sanoursa: You're assuming here that mkdir failed on the creation of the first placeholder within the parent directory. If this is the second or subsequent placeholder, then the parent now has children in it, and the next attempt to enumerate the parent will fail.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+
+-   We crash between mkdir() and InitializeEmptyPlaceholder
+
+    -   Placeholder folder does not have flags for virtualization root and
```
sanoursa: Nit: folder -> directory for consistency

sanoursa: Here and anywhere else this word appears

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+    -   This folder will never be expanded as the kext will exit when
+         we do not see the InVirtualizationRoot bit
+
+-   One of the SetBitInFileFlags calls fails in
+     InitializeEmptyPlaceholder
+
+    -   Folder is marked as in virtualization root but not empty.
+
+        -   This folder will never be expanded as the kext will exit
+             when we do not see the IsEmpty bit
```
sanoursa: And to be clear, this will manifest as a directory that is allowed to be enumerated, but appears empty. In other words, it will look like files are missing from the repo.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+             when we do not see the InVirtualizationRoot bit (we'll
+             never reach the IsEmpty test)
+
+WritePlaceholderFile:
+
+-   Fopen() fails to create a file
+
+    -   This should be retriable as a second attempt should see the
+         IsEmpty bit on the parent folder and call WritePlaceholderFile
+         again. No incomplete placeholders were left in the parent
```
sanoursa: Again, only if this is the first placeholder being created within the parent directory.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+    -   VFSForGit will be unable to get the SHA of the object that it
+         needs to download and hydrating this file will fail. Retrying
+         will not help.
+
+-   We crash before reaching InitializeEmptyPlaceholder
+
+    -   Placeholder does not have flags for virtualization root and
+        empty.
+
+    -   This file will never be expanded as the kext will exit when we
```
sanoursa: And will manifest as a file that is the right size but is missing all of its contents

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+         chmodding this file.
+
+I believe that all of these problems can be addressed by moving to a
+model where we create placeholders in a temporary location, mark them up
+and then rename them into place.
+
+Proposed Solution:
+------------------
+
+This solution creates placeholders in a .temp directory within the
```
sanoursa: Please use a name other than `.temp`. We don't want to convey that it is safe to delete this directory.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+
+Currently, the provider registers its virtualization root with the kext
+at clone time. We have code today that will log on files detected
+outside of the virtualization root, which could be helpful for
+diagnosing customer issues.
+
+4974: 218.308) Note: No virtualization root found for file with set
+flag. (vnode path: \'/Users/nickgra/TestRoot/.temp/mov\')
+
+In future, the consequences could be worse than logging, so the idea is
```
sanoursa: I think even now, the consequences are worse than logging. The biggest issue is that we will, for a short period of time, have files that have the `IsInVirtualizationRoot` bit set on them, even though they are not actually inside a virtualization root. This can break various assumptions in our kext, and we will likely respond by failing IO to these files, without special care to allow the provider to do this in a well known location

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+    -   The rename() to copy the completed placeholder from .temp into
+        the virtualization root fails
+
+        -   We can retry from this error since nothing ever appeared in
+            the virtualization root. The placeholder will be overwritten
+            as part of the retry
+
+With this approach, each placeholder file that will be created will be moved in one at a time to their final location rather than moving all placeholders that may be present in a folder at once. 
+
+In a case where we crash (or the machine crashes, or the user kills the mount process) while expanding a placeholder directory, we will be able to recover by replacing all placeholders that are present in that folder and continuing the enumeration. We will not try to short circuit this process and restart the enumeration from where we crashed. Any user content that is present in this directory will not be overwritten by the directory enumeration.
```
sanoursa: > Any user content that is present in this directory will not be overwritten by the directory enumeration.

How are you ensuring that?

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+
+-   A user could write into one of our unhydrated placeholders (if our
+    kext isn't loaded, we can't block their access to the file). We
+    should validate that we never lose any changes that a user makes in
+    this way.
+
+-   We should clear the .temp directory on mounting to prevent abandon
+    placeholders from cluttering up disk space.
+
+-   All of the notification work will need to be validated under the new
```
sanoursa: I think you can just rely on functional tests to handle this. This is sort of assumed for every new feature - make sure you don't break any other feature. Not needed to call it out here, I think.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+    hydration model to ensure that behavior will not be changed.
+
+-  Do we need to ensure that that flags and xattrs are flushed to disk before renaming the placeholder?
```
sanoursa: That would be ideal, yes. But we also need to measure the impact on perf.

Atomic.md
```diff
@@ -0,0 +1,278 @@
...
+temporary path alongside the kext changes to actually ignore activity in
+registered temp directories in the kext.
+
+Let's enumerate all the failure cases again under this new design:
+
+WritePlaceholderDirectory:
+
+-   mkdir() fails to create the placeholder directory
+
+    -   We can simply retry and have this succeed. Nothing was created
```
sanoursa: Do we retry within the same enumeration attempt? Or fail this one and the retry happens the next time the user tries some IO?

Atomic.md
wilbaker: Yep I agree.  Even without the flush this proposal is a big improvement.

Atomic.md
nickgra: That makes sense, will do.

Atomic.md
nickgra: Ah, I wrote this section with the design that is proposed below in mind. I'll rework that.

Atomic.md
```diff
@@ -0,0 +1,253 @@
...
+         chmodding this file.
+
+I believe that all of these problems can be addressed by moving to a
+model where we create placeholders in a temporary location, mark them up
+and then rename them into place.
+
+Proposed Solution:
+------------------
+
+This solution creates placeholders in a .staging directory within the
```
sanoursa: I like the new `.staging` name

One implementation detail you'll very likely have to worry about is how to avoid inheriting owner/group/acl's from the `.staging` folder as you create your placeholder files. If a user decides to update the permissions/inheritance on this folder, we will need to detect and update that before we create placeholders inside it. We will also need some comprehensive tests covering these Mac-specific hydration concerns.

Atomic.md
```diff
@@ -0,0 +1,253 @@
...
+at clone time. We have code today that will log on files detected
+outside of the virtualization root, which could be helpful for
+diagnosing customer issues.
+
+4974: 218.308) Note: No virtualization root found for file with set
+flag. (vnode path: \'/Users/nickgra/TestRoot/.staging/mov\')
+
+In the kext, we have widely assumed that files with `IsInVirtualizationRoot` will live inside our virtualization root. We will  fail to respond correctly to these kind of requests unless changes are made to handle the temporary directory scenario. Functionally, we want changes in registered temporary directories to be ignored. Changes will be needed on the provider side to register the
+temporary path alongside the kext changes to actually ignore activity in
+registered temp directories in the kext.
```
sanoursa: Can we also use our kext to prevent the user from doing anything to our `.staging` folder? Ideally, limit _all_ IO to just the `gvfsmountuser` identity that we think we'll use to avoid the auth cache poison issues.

Atomic.md
```diff
@@ -0,0 +1,253 @@
...
+        left .staging yet
+
+    -   The rename() to copy the completed placeholder from .staging into
+        the virtualization root fails
+
+        -   We can retry from this error since nothing ever appeared in
+            the virtualization root. The placeholder will be overwritten
+            as part of the retry
+
+With this approach, each placeholder file that will be created will be moved in one at a time to their final location rather than moving all placeholders that may be present in a directory at once.
```
sanoursa: Will we create/move each placeholder, or first create all and then move all? Is there a meaningful difference between the two?

Atomic.md
```diff
@@ -0,0 +1,253 @@
...
+            the virtualization root. The placeholder will be overwritten
+            as part of the retry
+
+With this approach, each placeholder file that will be created will be moved in one at a time to their final location rather than moving all placeholders that may be present in a directory at once. 
+
+In a case where we crash (or the machine crashes, or the user kills the mount process) while expanding a placeholder directory, we will be able to recover by calling UpdatePlaceholderInformation on files that were enumerated by the previous attempt and creating the rest of the placeholders as expected. 
+
+We need to call UpdatePlaceholderInformation in order to prevent data loss in scenarios where users manage to create files in the partially enumerated directory (if the kext is unmounted, the user can write files into the directory that will look like full files to VFSForGit).
+UpdatePlaceholderInformation currently will add any file to ModifiedPaths (which will make 'git status' begin tracking it, if any changes are present) and will one day be smart enough to only update files that are still placeholders.
```
wilbaker: Small correction here:

`UpdatePlaceholderIfNeeded` will (in the future) fail if the file is not currently a placeholder.  

If this happens then VFSForGit's expansion code (that is calling `UpdatePlaceholderIfNeeded`) will need to update ModifiedPaths.

(It's currently the code for updating the projection that handles failures of `UpdatePlaceholderIfNeeded` and adds paths to ModifiedPaths)

Atomic.md
```diff
@@ -0,0 +1,253 @@
...
+
+-   Is it possible to hydrate a placeholder in a directory that isn't
+    fully enumerated? It shouldn't be, will revalidate experimentally.
+
+-   Placeholders that are being created in .staging should be given a
+    temporary name (we'll rename() into the actual name). If we give
+    them the exact name of the final file, there is a hole where some
+    issue occurs, they get abandoned in the .staging directory. Once that
+    happens, we'll encounter issues on a retry where we try creating a
+    file where one exists with the same name.
```
wilbaker: I'm having trouble understanding this hole.  

I could see wanting to use temp names in case multiple folders are being expanded in parallel, but why couldn't we just overwrite existing files in .staging (if needed)?

Atomic.md
```diff
@@ -0,0 +1,253 @@
...
+         we do not see the InVirtualizationRoot bit
+
+-   We crash between setting the bits for 'in virtualization root' and 'is empty'
+
+    -   This directory will never be expanded as the kext will exit
+            when we do not see the IsEmpty bit. This effectively puts a user into a state where it looks like they're missing files in their repo.
+
+WritePlaceholderFile:
+
+-   Fopen() fails to create a file
```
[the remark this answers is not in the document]
nickgra: A quirk of writing this in Word, which really wants to capitalize the beginning of sentences.

Atomic.md
nickgra: We will create and move each placeholder incrementally rather than move all placeholders at once. I added some more content to why we didn't choose rename_xp() based designs.

sanoursa: Yes I understand that we can't move them "all at once", meaning atomically. I was more wondering, should the flow be: first create all the placeholders, one at a time; then move all the placeholders, one at a time.

The one advantage I can think of is that we at least got as far as _creating_ all of the placeholders, before we attempt to move any of them into the actual virtualization root, so there's less chance of needing to clean anything up later.

Atomic.md
nickgra: That's a better justification for this hole, I will switch to that explanation.

nickgra:
Copied to #234

PR #218, opened 2018-08-28
Add EnlistmentId and MountId to ETW data for correlation

kewillford:
When matching up data with ETW event there is an EnlistmentId and mountId that it tied to data about the enlistment and the mounted process.  Other events are not tied to this data so when an exception happens getting the mount data that corresponds to it is not directly correlated.

This adds the EnlistmentId and MountId to the ETW data so that the data can be correlated.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
+++ b/GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
@@ -32,14 +32,18 @@ namespace GVFS.Platform.Windows
         private const long MeasureKeyword = 0x400000000000;
 
         private EventSource eventSource;
+        private string enlistmentId;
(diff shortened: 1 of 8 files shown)


GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
```diff
@@ -107,20 +111,37 @@ private EventSourceOptions CreateOptions(EventLevel level, Keywords keywords, Ev
         
         // Needed to pass relatedId without metadata
```
wilbaker: Should this comment be removed now?

kewillford: Removed.

Issue #219, opened 2018-08-29
Mac: Support projection changes that add new files


wilbaker:
Update VFSForGit for Mac to support projection changes where the new commit has new files (that VFSFotGit needs to create on disk as part of its projection change logic)

See also: #196

wilbaker:
Fixed by #239

PR #220, opened 2018-08-29
Mac: Support basic projection changes and re-categorize functional tests

### LINKED ISSUE #196: Mac: Support basic projection change scenarios
Make the changes required to support basic projection change scenarios.  Follow up items can be filed for scenarios not covered by this work (e.g. updating the projection after moving to a branch that has new files)

wilbaker:
**Summary of VFSForGit product changes**

- Added update\delete placeholder functionality to `ProjFS.Mac`
- Update VFSForGit (for Mac) to add files\folders to the placeholder list as they are created during enumeration
- Fixed bugs\issues in `MacFileSystemVirtualizer.UpdatePlaceholderIfNeeded`

With these changes *most* projection change scenarios are covered.  #219 has been filed for supporting projection changes where new files need to be created by VFSForGit (something not covered by the changes in this PR). 

**Summary of test changes**

- Eliminated `--mac-only` flag and replaced it with an OS check in the functional tests
- Switched from a allow-list of tests that should run on Mac to a deny-list of tests that should not run on Mac
- Fixed up Windows specific paths
- Enabled many more tests on Mac (283 now passing, up from 204)

```
 Test Run Summary
   Overall result: Warning
   Test Count: 286, Passed: 282, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 4
     Skipped Tests - Ignored: 4, Explicit: 0, Other: 0
     Start time: 2018-08-29 23:15:15Z
     End time: 2018-08-29 23:28:25Z
     Duration: 790.114 seconds
```

**Follow up work items**
#219 
#223 
#231 

Fixes #196

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Categories.cs
+++ b/GVFS/GVFS.FunctionalTests/Categories.cs
@@ -6,13 +6,17 @@
         public const string FastFetch = "FastFetch";
         public const string GitCommands = "GitCommands";
 
-        public const string Windows = "Windows";
+        public const string WindowsOnly = "WindowsOnly";
+        public const string MacOnly = "MacOnly";
 
-        public static class Mac
+        public static class MacTODO
         {
-            public const string M1 = "M1_CloneAndMount";
+            // The FailsOnBuildAgent category is for tests that pass on dev
+            // machines but not on the build agents
+            public const string FailsOnBuildAgent = "FailsOnBuildAgent";
+
+            public const string NeedsLockHolder = "NeedsDotCoreLockHolder"; 
             public const string M2 = "M2_StaticViewGitCommands";
-            public const string M2TODO = "M2_StaticViewGitCommandsStillTODO";
             public const string M3 = "M3_AllGitCommands";
             public const string M4 = "M4_All";
         }
--- a/GVFS/GVFS.FunctionalTests/Program.cs
+++ b/GVFS/GVFS.FunctionalTests/Program.cs
@@ -60,17 +60,21 @@ namespace GVFS.FunctionalTests
 
             if (runner.HasCustomArg("--windows-only"))
             {
-                includeCategories.Add(Categories.Windows);
+                includeCategories.Add(Categories.WindowsOnly);
             }
 
-            if (runner.HasCustomArg("--mac-only"))
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
-                includeCategories.Add(Categories.Mac.M1);
-                includeCategories.Add(Categories.Mac.M2);
-                excludeCategories.Add(Categories.Mac.M2TODO);
-                excludeCategories.Add(Categories.Mac.M3);
-                excludeCategories.Add(Categories.Mac.M4);
-                excludeCategories.Add(Categories.Windows);
+                excludeCategories.Add(Categories.MacTODO.NeedsLockHolder);
+                excludeCategories.Add(Categories.MacTODO.FailsOnBuildAgent);
+                excludeCategories.Add(Categories.MacTODO.M2);
+                excludeCategories.Add(Categories.MacTODO.M3);
+                excludeCategories.Add(Categories.MacTODO.M4);
+                excludeCategories.Add(Categories.WindowsOnly);
+            }
+            else
+            {
+                excludeCategories.Add(Categories.MacOnly);
             }
 
             GVFSTestConfig.RepoToClone =
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -14,6 +14,8 @@ namespace GVFS.Platform.Mac
 {
     public class MacFileSystemVirtualizer : FileSystemVirtualizer
     {
+        public static readonly byte[] PlaceholderVersionId = ToVersionIdByteArray(new byte[] { PlaceholderVersion });
+
         private VirtualizationInstance virtualizationInstance;
 
         public MacFileSystemVirtualizer(GVFSContext context, GVFSGitObjects gitObjects)
@@ -81,11 +83,18 @@ namespace GVFS.Platform.Mac
             out UpdateFailureReason failureReason)
         {
             UpdateFailureCause failureCause = UpdateFailureCause.NoFailure;
+
+            // TODO(Mac): Add functional tests that include:
+            //     - Mode + content changes between commits
+            //     - Mode only changes (without any change to content, see issue #223)
+            ushort fileMode = this.FileSystemCallbacks.GitIndexProjection.GetFilePathMode(relativePath);
+
             Result result = this.virtualizationInstance.UpdatePlaceholderIfNeeded(
                 relativePath,
-                GetPlaceholderVersionId(),
-                ConvertShaToContentId(shaContentId),
+                PlaceholderVersionId,
+                ToVersionIdByteArray(ConvertShaToContentId(shaContentId)),
                 (ulong)endOfFile,
+                fileMode,
                 (UpdateType)updateFlags,
                 out failureCause);
             failureReason = (UpdateFailureReason)failureCause;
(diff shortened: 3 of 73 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -352,6 +353,74 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
...
+        << "PrjFS_UpdatePlaceholderFileIfNeeded("
+        << relativePath << ", "
+        << (int)providerId[0] << ", "
+        << (int)contentId[0] << ", "
+        << fileSize << ", "
+        << std::oct << fileMode << std::dec << ", "
+        << std::hex << updateFlags << std::dec << ")" << std::endl;
+#endif
+    
+    PrjFS_Result result = PrjFS_DeleteFile(relativePath, updateFlags, failureCause);
```
sanoursa: Is this just a temporary approach? Will we be able to update the existing placeholder in place?

ProjFS.Mac/PrjFSLib/PrjFSLib.h
```diff
@@ -89,6 +89,7 @@ typedef enum
 {
     PrjFS_UpdateType_Invalid                        = 0x00000000,
     
+    PrjFS_UpdateType_AllowDirtyData                 = 0x00000002,
```
sanoursa: Why do we need this flag?

ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs
```diff
@@ -106,10 +113,28 @@ public virtual Result StopVirtualizationInstance()
             byte[] providerId,
             byte[] contentId,
             ulong fileSize,
+            ushort fileMode,
```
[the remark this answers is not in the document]
wilbaker: That's a really good point and I'm not sure, I can add a TODO and file an issue for this

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -352,6 +353,74 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
...
+    std::cout
+        << "PrjFS_UpdatePlaceholderFileIfNeeded("
+        << relativePath << ", "
+        << (int)providerId[0] << ", "
+        << (int)contentId[0] << ", "
+        << fileSize << ", "
+        << std::oct << fileMode << std::dec << ", "
+        << std::hex << updateFlags << std::dec << ")" << std::endl;
+#endif
+
```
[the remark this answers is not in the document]
wilbaker: Note to self: The "if needed" that we're missing is comparing the `contentId`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I was not thinking this (delete+add) would be a temporary approach:

- ProjFS on Windows internally does a delete+add when updating placeholders
- Git does a delete+add when updating the working tree as part of a checkout

We do need to compare content ID, but I was thinking we'd still with delete+add so that applications see the same events with VFSForGit as they do with Git.

ProjFS.Mac/PrjFSLib/PrjFSLib.h
wilbaker: Good point, we do not.

I added it here because it was already present in `PrjFSLib.Mac.UpdateType`, but I think rather than adding it here I can remove it from `UpdateType`

ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs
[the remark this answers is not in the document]
wilbaker: Filed #223 for this

wilbaker:
@sanoursa I believe I've addressed\replied to all of you feedback so far.  The latest changes are ready for another look.

GVFS/GVFS.FunctionalTests/Categories.cs
```diff
@@ -6,13 +6,17 @@ public static class Categories
...
+        public const string WindowsOnly = "WindowsOnly";
+        public const string MacOnly = "MacOnly";
 
-        public static class Mac
+        public static class MacTODO
         {
-            public const string M1 = "M1_CloneAndMount";
+            // The FailsOnBuildAgent category is for tests that pass on dev
+            // machines but not on the build agents
+            public const string FailsOnBuildAgent = "FailsOnBuildAgent";
```
sanoursa: I'm just curious, any theories as to why some of the tests are failing on the build agent?

GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -60,17 +60,20 @@ public static void Main(string[] args)
 
             if (runner.HasCustomArg("--windows-only"))
```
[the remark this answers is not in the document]
sanoursa: Outside the scope of this PR, but do we need to maintain the ability to run the core tests in the .NET Framework project, if we never run them from that project? Not saying we should necessarily, but it seems like we could simplify this so that the .NET Framework test project contains only the Windows-specific tests. Any reason not to?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -81,11 +81,18 @@ public override void Stop()
...
+            // TODO(Mac): Add functional tests that include:
+            //     - Mode + content changes between commits
+            //     - Mode only changes (without any change to content, see issue #223)
+            ushort fileMode = this.FileSystemCallbacks.GitIndexProjection.GetFilePathMode(relativePath);
+
             Result result = this.virtualizationInstance.UpdatePlaceholderIfNeeded(
                 relativePath,
-                GetPlaceholderVersionId(),
-                ConvertShaToContentId(shaContentId),
+                ToVersionIdByteArray(GetPlaceholderVersionId()),
```
sanoursa: This affects both Mac and Windows - can we just define a single readonly `byte[]` in the base class for the version id and just use that for all placeholders? This is essentially a constant, and right now we're paying the cost of allocating it for every single placeholder. And on Mac it's actually a double allocation.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -352,6 +353,77 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
...
+    
+    // TODO(Mac): Populate failure cause appropriately
+    *failureCause = PrjFS_UpdateFailureCause_Invalid;
+    
+    if (nullptr == relativePath)
+    {
+        return PrjFS_Result_EInvalidArgs;
+    }
+
+    // TODO(Mac): Unless allowed by updateFlags, ensure file is not full before proceeding
```
sanoursa: Not something we're planning on supporting, right?

GVFS/GVFS.FunctionalTests/Categories.cs
wilbaker: @nickgra had investigated this a bit.  The failures all have to do with non-ASCII file names, and I think Nick mentioned there must be a setting on the build agents that impacts how the file names are displayed.  

From what I remember it was something to do with continuation characters being added into the names, @nickgra, do you remember any of the details?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
wilbaker: Good idea, I'll make the change

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: Yep you're right, I'll update the comment

GVFS/GVFS.FunctionalTests/Program.cs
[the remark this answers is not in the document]
sanoursa: I can't remember I ran the functional tests from the IDE :-). If that's something you do, I'm totally fine with making it easier.

Running all of the Windows tests on .NET Framework may actually make more sense, since the Windows client is still using .NET Framework. The current setup is also working well enough, so this is a someday-maybe clean up as far as I'm concerned.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
wilbaker: I will define a readonly `byte[]` but it can't be shared as its length relies on the platform specific `VirtualizationInstance`.

GVFS/GVFS.FunctionalTests/Categories.cs
nickgra: You've nailed it, that's about as far as I got in the investigation. I have some ideas on how we can debug this further, I'll follow up offline.

PR #221, opened 2018-08-29
Allow Git downgrade if latest release contains an older version

alameenshah:


alameenshah:
Contains commits from open PRs. Difficult to review.

PR #224, opened 2018-08-29
Allow upgrade when release contains an older version of Git.

alameenshah:


alameenshah:
accidentally lost the branch in my fork. Created a new PR, which has updated branch info.

PR #228, opened 2018-08-30
Abort upgrade when GVFS.Service is not running

alameenshah:
- Adding precheck for GVFS.Service. If Service is not running, then upgrade will abort with error. 
- In case of fresh install, this check will not be performed.
- Added platform specific method to check Service status.
- Updated UT.

### DIFF
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -147,14 +147,11 @@ namespace GVFS.Upgrader
             return GVFSPlatform.Instance.KernelDriver.IsGVFSUpgradeSupported();
         }
 
-        protected virtual bool FreshInstallOrServiceRunning()
+        protected virtual bool IsServiceInstalledAndNotRunning()
         {
-            bool isInstalled;
-            bool isRunning;
+            GVFSPlatform.Instance.IsServiceInstalledAndRunning(GVFSConstants.Service.ServiceName, out bool isInstalled, out bool isRunning);
 
-            GVFSPlatform.Instance.IsServiceInstalledAndRunning(GVFSConstants.Service.ServiceName, out isInstalled, out isRunning);
-
-            return isInstalled ? isRunning : true;
+            return isInstalled && !isRunning;
         }
 
         protected virtual bool IsUnattended()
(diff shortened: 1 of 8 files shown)


GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -147,6 +147,16 @@ protected virtual bool IsProjFSInboxed()
...
         }
 
+        protected virtual bool FreshInstallOrServiceRunning()
+        {
+            bool isInstalled;
+            bool isRunning;
+
+            GVFSPlatform.Instance.IsServiceInstalledAndRunning(GVFSConstants.Service.ServiceName, out isInstalled, out isRunning);
+
+            return isInstalled ? isRunning : true;
```
derrickstolee: This is essentially "isInstalled implies isRunning" which can be given as `!isInstalled || isRunning` (which also seems to match your function name a bit better).

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -147,6 +147,16 @@ protected virtual bool IsProjFSInboxed()
             return GVFSPlatform.Instance.KernelDriver.IsGVFSUpgradeSupported();
         }
 
+        protected virtual bool FreshInstallOrServiceRunning()
+        {
+            bool isInstalled;
+            bool isRunning;
+
+            GVFSPlatform.Instance.IsServiceInstalledAndRunning(GVFSConstants.Service.ServiceName, out isInstalled, out isRunning);
```
derrickstolee: with C# 7, you should be able to define the bools inside the method:

`GVFSPlatform.Instance.IsServiceInstalledAndRunning(GVFSConstants.Service.ServiceName, out bool isInstalled, out bool isRunning)`

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -248,6 +258,12 @@ private bool IsGVFSUpgradeAllowed(out string error)
                 return false;
             }
 
+            if (!this.FreshInstallOrServiceRunning())
+            {
+                error = "GVFS Service is not running.\nStart \"GVFS.Service\" and run \"gvfs upgrade\" again.";
```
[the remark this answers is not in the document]
alameenshah: >>Will this error show up on a fresh install
No. For fresh install using Upgrader tool directly, this would throw no errors.

>> Should we rephrase this method as if (!this.ServiceRunning()) and say no if it isn't even installed?
Rephrased it to IsServiceInstalledAndNotRunning.

Issue #235, opened 2018-08-31
NamedPipeServerStream.EndWaitForConnection() throws System.Net.Sockets.SocketException, which terminates the whole process


turbonaitis:
On Mac, during GVFS clone command, a `NamedPipeServer` is started and then disposed. Then `OnNewConnection` method is invoked and it tries to invoke `pipe.EndWaitForConnection(ar);`, which  throws `System.Net.Sockets.SocketException`, which is currently unhandled and reaches generic `LogErrorAndExit` path, which terminates the whole process, before it has completed prefetch or started the mount process. It seems on Windows this might behave differently and throw `ObjectDisposedException`, which is handled appropriately.

sanoursa:
Fixed in #244

PR #236, opened 2018-08-31
Handling SocketException in NamedPipesServer

turbonaitis:
Solves #235 by catching the SocketException and handling it the same way, that ObjectDisposedException is currently handled.

### DIFF
--- a/GVFS/GVFS.Common/NamedPipes/NamedPipeServer.cs
+++ b/GVFS/GVFS.Common/NamedPipes/NamedPipeServer.cs
@@ -3,6 +3,7 @@ using System;
 using System.IO;
 using System.IO.Pipes;
 using System.Threading;
+using System.Net.Sockets;
 
 namespace GVFS.Common.NamedPipes
 {
@@ -122,6 +123,13 @@ namespace GVFS.Common.NamedPipes
                     metadata.Add("Exception", e.ToString());
                     metadata.Add(TracingConstants.MessageKey.WarningMessage, "OnNewConnection: Connection broken");
                     this.tracer.RelatedEvent(EventLevel.Warning, "OnNewConnectionn_EndWaitForConnection_IOException", metadata);
+                }
+                catch(SocketException)
+                {
+                    if (!this.isStopping)
+                    {
+                        throw;
+                    }
                 }
                 catch (ObjectDisposedException)
                 {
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Common/NamedPipes/NamedPipeServer.cs
```diff
@@ -123,6 +124,13 @@ private void OnNewConnection(IAsyncResult ar, bool createNewThreadIfSynchronous)
                     metadata.Add(TracingConstants.MessageKey.WarningMessage, "OnNewConnection: Connection broken");
                     this.tracer.RelatedEvent(EventLevel.Warning, "OnNewConnectionn_EndWaitForConnection_IOException", metadata);
                 }
+                catch(SocketException)
```
sanoursa: Can you provide a little more info on your repro steps? It doesn't seem to happen on my machine when I call `gvfs clone`, or on any of our build agents, since they can all clone successfully. Any idea what would be causing this exception to get thrown in your environment?

That being said, I think this change is safe, since if we're not stopping, all we do is rethrow the exception, so for the most part the behavior should be unchanged from before.

turbonaitis: I'm yet to test this on Windows, but to me it happens on Mac all the time, when I do GVFS clone. The named pipe server, created in `CloneVerb.TryClone`, has already been disposed, for some reason `NamedPipeServer.OnNewConnection` is invoked (didn't dig deep enough to understand why) and once `pipe.EndWaitForConnection(ar);` is called - SocketException is thrown. My guess is that it might be a bug/difference in the underlying dotnet implementation, where on Windows it appropriately throws ObjectDisposedException and on Mac it ends up trying to access a closed socket. As far as I can tell, I'm not doing anything special.

turbonaitis: Just confirmed, that this is not happening with a windows client (same server).

sanoursa:
Thanks again for this PR, and it actually does point to a more serious underlying issue with disposed tracers. See #244 for my attempt at fixing this.

turbonaitis:
Ok, great to see you've fixed the root cause of this! With your fixes in, I assume we can abandon this PR.

Issue #231, opened 2018-08-30
Mac: PrjFSLib should provide a way to delete directory placeholders


wilbaker:
`PrjFS_DeleteFile` will only work for files because it uses `unlink`.

When this change is made we'll need to revisit the projection change code that aggressively tries to delete empty directories.

wilbaker:
Fixed by #239

Issue #223, opened 2018-08-29
Mac: UpdatePlaceholderIfNeeded should be called when a file's mode changes


wilbaker:
Even if a file's content does not change, if its mode changes UpdatePlaceholderIfNeeded should be called.

wilbaker:
For the first part:

> `git status` should report files as modified when their mode changes

It appears that the the kext receives a `KAUTH_VNODE_WRITE_SECURITY` operation when `chmod` is used.

wilbaker:
Splitting this issue into two issues.

PR #239, opened 2018-09-04
Mac: Support projection changes that add new files and delete folder

wilbaker:
Changes for #231 and #223 

*Summary of Changes*
- Updated `PrjFS_DeleteFile` so that it can delete files and folders on Mac (#231)
- Updated VFSForGit to track expanded folders
- Updated the projection change code so that after updating file placeholders VFSForGit will re-expand folders that were previously expanded.  VFSForGit does this by checking the contents of each expanded folder in the the projection and ensuring that all of the that folder's children (files and folders) are on disk.

*Folder re-expansion logic*
1. GVFS records what folders have been expanded
2. When a projection change occurs VFSForGit looks up the expanded folders in the projection
3. If there are any children in the projection for the folder it means that their skip-worktree bit is on, and if the skip-worktree bit is on and the file\folder is not in the placeholder list VFSForGit needs to write it out.

### TODOs
- [X] Address new TODOs that have been added in the code
- [X] Enable additional functional tests

```
 Test Run Summary
   Overall result: Warning
   Test Count: 330, Passed: 326, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 4
     Skipped Tests - Ignored: 4, Explicit: 0, Other: 0
   Start time: 2018-09-06 19:21:38Z
     End time: 2018-09-06 19:37:06Z
     Duration: 928.064 seconds
```

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
@@ -3,6 +3,7 @@
     public interface IPlatformFileSystem
     {
         bool SupportsFileMode { get; }
+        bool EnumerationExpandsDirectories { get; }
         void FlushFileBuffers(string path);
         void MoveAndOverwriteFile(string sourceFileName, string destinationFilename);
         void CreateHardLink(string newLinkFileName, string existingFileName);
--- a/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
@@ -3,7 +3,6 @@
     public interface IPlatformFileSystem
     {
         bool SupportsFileMode { get; }
-        bool EnumerationExpandsDirectories { get; }
         void FlushFileBuffers(string path);
         void MoveAndOverwriteFile(string sourceFileName, string destinationFilename);
         void CreateHardLink(string newLinkFileName, string existingFileName);
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -7,6 +7,7 @@ namespace GVFS.Common
         public const int ShaStringLength = 40;
         public const int MaxPath = 260;
         public const string AllZeroSha = "0000000000000000000000000000000000000000";
+        public const string ExpandedFolderSha = "XF00000000000000000000000000000000000000";
 
         public const char GitPathSeparator = '/';
         public const string GitPathSeparatorString = "/";
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -7,7 +7,6 @@ namespace GVFS.Common
         public const int ShaStringLength = 40;
         public const int MaxPath = 260;
         public const string AllZeroSha = "0000000000000000000000000000000000000000";
-        public const string ExpandedFolderSha = "X_EXPANDED_FOLDER00000000000000000000000";
 
         public const char GitPathSeparator = '/';
         public const string GitPathSeparatorString = "/";
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -7,7 +7,7 @@ namespace GVFS.Common
         public const int ShaStringLength = 40;
         public const int MaxPath = 260;
         public const string AllZeroSha = "0000000000000000000000000000000000000000";
-        public const string ExpandedFolderSha = "XF00000000000000000000000000000000000000";
+        public const string ExpandedFolderSha = "X_EXPANDED_FOLDER00000000000000000000000";
 
         public const char GitPathSeparator = '/';
         public const string GitPathSeparatorString = "/";
--- a/GVFS/GVFS.Common/PlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/PlaceholderListDatabase.cs
@@ -8,7 +8,12 @@ namespace GVFS.Common
 {
     public class PlaceholderListDatabase : FileBasedCollection
     {
+        public const string PartialFolderValue = "@FPARTIAL_000000000000000000000000000000";
+        public const string ExpandedFolderValue = "@FEXPANDED_00000000000000000000000000000";
+
         private const char PathTerminator = '\0';
+        private const string FolderValuePrefix = "@F";
+        private const string ExpandedFolderPrefix = "@FE";
 
         // This list holds entries that would otherwise be lost because WriteAllEntriesAndFlush has not been called, but a file 
         // snapshot has been taken using GetAllEntries.
@@ -51,26 +56,15 @@ namespace GVFS.Common
             output = temp;
             return true;
         }
-        
-        public void AddAndFlush(string path, string sha)
+
+        public void AddAndFlushFile(string path, string sha)
         {
-            try
-            {
-                this.WriteAddEntry(
-                    path + PathTerminator + sha,
-                    () =>
-                    {
-                        this.EstimatedCount++;
-                        if (this.placeholderDataEntries != null)
-                        {
-                            this.placeholderDataEntries.Add(new PlaceholderDataEntry(path, sha));
-                        }
-                    });
-            }
-            catch (Exception e)
-            {
-                throw new FileBasedCollectionException(e);
-            }
+            this.AddAndFlush(path, sha);
+        }
+
+        public void AddAndFlushFolder(string path, bool isExpanded)
+        {
+            this.AddAndFlush(path, isExpanded ? ExpandedFolderValue : PartialFolderValue);
         }
 
         public void RemoveAndFlush(string path)
@@ -98,13 +92,13 @@ namespace GVFS.Common
         {
             try
             {
-                List<PlaceholderData> output = new List<PlaceholderData>(Math.Max(1, this.EstimatedCount));
+                List<PlaceholderData> placeholders = new List<PlaceholderData>(Math.Max(1, this.EstimatedCount));
 
                 string error;
                 if (!this.TryLoadFromDisk<string, string>(
                     this.TryParseAddLine,
                     this.TryParseRemoveLine,
-                    (key, value) => output.Add(new PlaceholderData(path: key, sha: value)),
+                    (key, value) => placeholders.Add(new PlaceholderData(path: key, fileShaOrFolderValue: value)),
                     out error,
                     () =>
                     {
@@ -119,7 +113,52 @@ namespace GVFS.Common
                     throw new InvalidDataException(error);
                 }
 
-                return output;
+                return placeholders;
+            }
+            catch (Exception e)
+            {
+                throw new FileBasedCollectionException(e);
+            }
+        }
+
+        public void GetAllEntries(out List<PlaceholderData> filePlaceholders, out List<PlaceholderData> folderPlaceholders)
+        {
+            try
+            {
+                List<PlaceholderData> filePlaceholdersFromDisk = new List<PlaceholderData>(Math.Max(1, this.EstimatedCount));
+                List<PlaceholderData> folderPlaceholdersFromDisk = new List<PlaceholderData>(Math.Max(1, (int)(this.EstimatedCount * .3)));
+
+                string error;
+                if (!this.TryLoadFromDisk<string, string>(
+                    this.TryParseAddLine,
+                    this.TryParseRemoveLine,
+                    (key, value) =>
+                    {
+                        if (value.StartsWith(FolderValuePrefix, StringComparison.Ordinal))
+                        {
+                            folderPlaceholdersFromDisk.Add(new PlaceholderData(path: key, fileShaOrFolderValue: value));
+                        }
+                        else
+                        {
+                            filePlaceholdersFromDisk.Add(new PlaceholderData(path: key, fileShaOrFolderValue: value));
+                        }
+                    },
+                    out error,
+                    () =>
+                    {
+                        if (this.placeholderDataEntries != null)
+                        {
+                            throw new InvalidOperationException("PlaceholderListDatabase should always flush queue placeholders using WriteAllEntriesAndFlush before calling GetAllEntries again.");
+                        }
+
+                        this.placeholderDataEntries = new List<PlaceholderDataEntry>();
+                    }))
+                {
+                    throw new InvalidDataException(error);
+                }
+
+                filePlaceholders = filePlaceholdersFromDisk;
+                folderPlaceholders = folderPlaceholdersFromDisk;
             }
             catch (Exception e)
             {
@@ -181,6 +220,27 @@ namespace GVFS.Common
             }
         }
 
+        private void AddAndFlush(string path, string sha)
+        {
+            try
+            {
+                this.WriteAddEntry(
+                    path + PathTerminator + sha,
+                    () =>
+                    {
+                        this.EstimatedCount++;
+                        if (this.placeholderDataEntries != null)
+                        {
+                            this.placeholderDataEntries.Add(new PlaceholderDataEntry(path, sha));
+                        }
+                    });
+            }
+            catch (Exception e)
+            {
+                throw new FileBasedCollectionException(e);
+            }
+        }
+
         private bool TryParseAddLine(string line, out string key, out string value, out string error)
         {
             // Expected: <Placeholder-Path>\0<40-Char-SHA1>
@@ -197,7 +257,7 @@ namespace GVFS.Common
             {
                 key = null;
                 value = null;
-                error = "Invalid SHA1 length: " + line;
+                error = $"Invalid SHA1 length {line.Length - idx - 1}: " + line;
                 return false;
             }
 
@@ -218,10 +278,10 @@ namespace GVFS.Common
 
         public class PlaceholderData
         {
-            public PlaceholderData(string path, string sha)
+            public PlaceholderData(string path, string fileShaOrFolderValue)
             {
                 this.Path = path;
-                this.Sha = sha;
+                this.Sha = fileShaOrFolderValue;
             }
 
             public string Path { get; }
@@ -230,8 +290,8 @@ namespace GVFS.Common
             public bool IsFolder
             {
                 get 
-                { 
-                    return this.Sha == GVFSConstants.AllZeroSha || this.IsExpandedFolder;
+                {
+                    return this.Sha.StartsWith(FolderValuePrefix, StringComparison.Ordinal);
                 }
             }
 
@@ -239,7 +299,7 @@ namespace GVFS.Common
             {
                 get
                 {
-                    return this.Sha == GVFSConstants.ExpandedFolderSha;
+                    return this.Sha.StartsWith(ExpandedFolderPrefix, StringComparison.Ordinal);
                 }
             }
         }
--- a/GVFS/GVFS.Common/PlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/PlaceholderListDatabase.cs
@@ -8,12 +8,13 @@ namespace GVFS.Common
 {
     public class PlaceholderListDatabase : FileBasedCollection
     {
-        public const string PartialFolderValue = "@FPARTIAL_000000000000000000000000000000";
-        public const string ExpandedFolderValue = "@FEXPANDED_00000000000000000000000000000";
+        // Special folder values must:
+        // - Be 40 characters long
+        // - Not be a valid SHA-1 value (to avoid collisions with files)
+        public const string PartialFolderValue = "                          PARTIAL FOLDER";
+        public const string ExpandedFolderValue = "                         EXPANDED FOLDER";
 
         private const char PathTerminator = '\0';
-        private const string FolderValuePrefix = "@F";
-        private const string ExpandedFolderPrefix = "@FE";
 
         // This list holds entries that would otherwise be lost because WriteAllEntriesAndFlush has not been called, but a file 
         // snapshot has been taken using GetAllEntries.
@@ -134,7 +135,7 @@ namespace GVFS.Common
                     this.TryParseRemoveLine,
                     (key, value) =>
                     {
-                        if (value.StartsWith(FolderValuePrefix, StringComparison.Ordinal))
+                        if (value == PartialFolderValue || value == ExpandedFolderValue)
                         {
                             folderPlaceholdersFromDisk.Add(new PlaceholderData(path: key, fileShaOrFolderValue: value));
                         }
@@ -291,7 +292,7 @@ namespace GVFS.Common
             {
                 get 
                 {
-                    return this.Sha.StartsWith(FolderValuePrefix, StringComparison.Ordinal);
+                    return this.Sha == PartialFolderValue || this.IsExpandedFolder;
                 }
             }
 
@@ -299,7 +300,7 @@ namespace GVFS.Common
             {
                 get
                 {
-                    return this.Sha.StartsWith(ExpandedFolderPrefix, StringComparison.Ordinal);
+                    return this.Sha == ExpandedFolderValue;
                 }
             }
         }
--- a/GVFS/GVFS.Common/PlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/PlaceholderListDatabase.cs
@@ -229,7 +229,18 @@ namespace GVFS.Common
 
             public bool IsFolder
             {
-                get { return this.Sha == GVFSConstants.AllZeroSha; }
+                get 
+                { 
+                    return this.Sha == GVFSConstants.AllZeroSha || this.IsExpandedFolder;
+                }
+            }
+
+            public bool IsExpandedFolder
+            {
+                get
+                {
+                    return this.Sha == GVFSConstants.ExpandedFolderSha;
+                }
             }
         }
 
--- a/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
+++ b/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
@@ -17,7 +17,7 @@ namespace GVFS.FunctionalTests.Windows.Tests
     [Category(Categories.WindowsOnly)]
     public class DiskLayoutUpgradeTests : TestsWithEnlistmentPerTestCase
     {
-        public const int CurrentDiskLayoutMajorVersion = 16;
+        public const int CurrentDiskLayoutMajorVersion = 17;
         public const int CurrentDiskLayoutMinorVersion = 0;
 
         public const string BlobSizesCacheName = "blobSizes";
@@ -153,34 +153,7 @@ namespace GVFS.FunctionalTests.Windows.Tests
         [TestCase]
         public void MountWritesFolderPlaceholdersToPlaceholderDatabase()
         {
-            // Create some placeholder data
-            this.fileSystem.ReadAllText(Path.Combine(this.Enlistment.RepoRoot, "Readme.md"));
-            this.fileSystem.ReadAllText(Path.Combine(this.Enlistment.RepoRoot, "Scripts\\RunUnitTests.bat"));
-            this.fileSystem.ReadAllText(Path.Combine(this.Enlistment.RepoRoot, "GVFS\\GVFS.Common\\Git\\GitRefs.cs"));
-
-            // Create a full folder
-            this.fileSystem.CreateDirectory(Path.Combine(this.Enlistment.RepoRoot, "GVFS\\FullFolder"));
-            this.fileSystem.WriteAllText(Path.Combine(this.Enlistment.RepoRoot, "GVFS\\FullFolder\\test.txt"), "Test contents");
-
-            // Create a tombstone
-            this.fileSystem.DeleteDirectory(Path.Combine(this.Enlistment.RepoRoot, "GVFS\\GVFS.Tests\\Properties"));
-
-            string junctionTarget = Path.Combine(this.Enlistment.EnlistmentRoot, "DirJunction");
-            string symlinkTarget = Path.Combine(this.Enlistment.EnlistmentRoot, "DirSymlink");
-            Directory.CreateDirectory(junctionTarget);
-            Directory.CreateDirectory(symlinkTarget);
-
-            string junctionLink = Path.Combine(this.Enlistment.RepoRoot, "DirJunction");
-            string symlink = Path.Combine(this.Enlistment.RepoRoot, "DirLink");
-            ProcessHelper.Run("CMD.exe", "/C mklink /J " + junctionLink + " " + junctionTarget);
-            ProcessHelper.Run("CMD.exe", "/C mklink /D " + symlink + " " + symlinkTarget);
-
-            string target = Path.Combine(this.Enlistment.EnlistmentRoot, "GVFS", "GVFS", "GVFS.UnitTests");
-            string link = Path.Combine(this.Enlistment.RepoRoot, "UnitTests");
-            ProcessHelper.Run("CMD.exe", "/C mklink /J " + link + " " + target);
-            target = Path.Combine(this.Enlistment.EnlistmentRoot, "GVFS", "GVFS", "GVFS.Installer");
-            link = Path.Combine(this.Enlistment.RepoRoot, "Installer");
-            ProcessHelper.Run("CMD.exe", "/C mklink /D " + link + " " + target);
+            this.PerformIOBeforePlaceholderDatabaseUpgradeTest();
 
             this.Enlistment.UnmountGVFS();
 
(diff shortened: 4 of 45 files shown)


GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
```diff
@@ -3,6 +3,7 @@
     public interface IPlatformFileSystem
     {
         bool SupportsFileMode { get; }
+        bool EnumerationExpandsDirectories { get; }
```
sanoursa: Should this live on `IKernelDriver`?

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -7,6 +7,7 @@ public static partial class GVFSConstants
         public const int ShaStringLength = 40;
         public const int MaxPath = 260;
         public const string AllZeroSha = "0000000000000000000000000000000000000000";
+        public const string ExpandedFolderSha = "X_EXPANDED_FOLDER00000000000000000000000";
```
sanoursa: This is no longer a sha. Does that matter? Could we just pick another fake, but actual, sha value?

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -7,6 +7,7 @@ public static partial class GVFSConstants
         public const int ShaStringLength = 40;
         public const int MaxPath = 260;
         public const string AllZeroSha = "0000000000000000000000000000000000000000";
```
sanoursa: Rename this to `EmptyFolderSha`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1094,14 +1099,25 @@ private void UpdatePlaceholders()
...
+                    IEnumerable<PlaceholderListDatabase.PlaceholderData> folderPlaceholderData = placeholderListCopy.Where(x => x.IsFolder);
+
+                    HashSet<string> folderPlaceholders = new HashSet<string>(folderPlaceholderData.Select(x => x.Path), StringComparer.OrdinalIgnoreCase);
+                    foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in folderPlaceholderData.OrderByDescending(x => x.Path))
+                    {
+                        if (GVFSPlatform.Instance.FileSystem.EnumerationExpandsDirectories && folderPlaceholder.IsExpandedFolder)
+                        {
+                            this.ReExpandFolder(blobSizesConnection, folderPlaceholder.Path, updatedPlaceholderList, folderPlaceholders, folderPlaceholdersToKeep);
+                        }
+
```
sanoursa: Are we missing an else here?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1094,14 +1099,25 @@ private void UpdatePlaceholders()
...
                 this.blobSizes.Flush();
 
-                // Waiting for the UpdateOrDeleteFilePlaceholder to fill the folderPlaceholdersToKeep before trying to remove folder placeholders
-                // so that we don't try to delete a folder placeholder that has file placeholders and just fails
-                foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in placeholderListCopy.Where(x => x.IsFolder).OrderByDescending(x => x.Path))
-                {
-                    this.TryRemoveFolderPlaceholder(folderPlaceholder, updatedPlaceholderList, folderPlaceholdersToKeep);
+                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.blobSizes.CreateConnection())
+                {
+                    // Waiting for the UpdateOrDeleteFilePlaceholder to fill the folderPlaceholdersToKeep before trying to remove folder placeholders
```
sanoursa: Where are we waiting? I don't see it in the code.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1257,17 +1273,107 @@ private string GetNewProjectedShaForPlaceholder(string path)
...
+                }
+                else
+                {
+                    if (!updatedPlaceholderList.ContainsKey(childRelativePath))
+                    {
+                        FileData childFileData = childEntry as FileData;
+                        string sha = childFileData.Sha.ToString();
+
+                        // TODO(Mac): Issue #245, handle failures of WritePlaceholder
+                        this.fileSystemVirtualizer.WritePlaceholder(childRelativePath, childFileData.Size, isDirectory: false, shaContentId: sha);
```
sanoursa: Break out the parameters onto separate lines to match the style above?

sanoursa: And actually the contents of this if block and the one above are virtually identical. Is there an easy way to avoid the duplication?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1257,17 +1273,107 @@ private string GetNewProjectedShaForPlaceholder(string path)
...
+                            childRelativePath,
+                            new PlaceholderListDatabase.PlaceholderData(childRelativePath, sha));
+
+                        folderPlaceholdersToKeep.Add(relativeFolderPath);
+                    }
+                }
+            }
+        }
+
         private void TryRemoveFolderPlaceholder(
```
sanoursa: Odd to have a Try method with void return, rather than a bool. Why does the Try mean in this context? Is this method just best-effort?

GVFS/GVFS.Common/GVFSConstants.cs
wilbaker: I think I picked the wrong spot to put this constant because this is really specific to the placeholder database (and as you mentioned it's not a sha).  I'm thinking we do something like:

- Leave `AllZeroSha` here (as-is)
- Add `EmptyFolderSha` to `PlaceholderListDatabase`
- Move `ExpandedFolderSha` to `PlaceholderListDatabase` and rename it something like `ExpandedFolderValue`

Optionally `AddAndFlush(string path, string sha)` could be renamed to something like `AddAndFlushFile(string path, string sha)` and we could add `AddAndFlushFolder(string path, bool isExpanded)`.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
sanoursa: I see now that you just moved this comment, so it's not new. But I'm having a hard time parsing what it means.

wilbaker: The wait happens in `ProcessListOnThreads`:

```
for (int i = 0; i < processThreads.Length; i++)
{
    processThreads[i].Join();
}
```
But I that's not clear from the comment.  I can change it to something like:

```
// Don't attempt to delete folder placeholders until UpdateOrDeleteFilePlaceholder 
// has finished populating folderPlaceholdersToKeep
```

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: There's no else because we still want to attempt to remove empty placeholders on Mac.

The call to `ReExpandFolder` happens first to (potentially) add more entries to `folderPlaceholdersToKeep`.

Let me know if you think a comment would be useful here.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: I'll work on cleaning this up

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: Yep it looks like best effort.  I think this can be renamed to something like `RemoveFolderPlaceholderIfEmpty`.

GVFS/GVFS.FunctionalTests/Categories.cs
```diff
@@ -14,8 +14,8 @@ public static class MacTODO
             // The FailsOnBuildAgent category is for tests that pass on dev
             // machines but not on the build agents
             public const string FailsOnBuildAgent = "FailsOnBuildAgent";
-
-            public const string NeedsLockHolder = "NeedsDotCoreLockHolder"; 
+            public const string NeedsLockHolder = "NeedsDotCoreLockHolder";
+            public const string NeedsCachePoisonFix = "NeedsCachePoisonFix";
```
sanoursa: Could you please add this category to `MultithreadedReadWriteTests.CanReadVirtualFileInParallel` and remove the `WindowsOnly` category from it?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
sanoursa: I think I would just remove the comment. My guess is that it was written before the ProcessListOnThreads was factored out, and we were actually waiting on threads to complete right before this comment. Now the comment is just confusing - of course we're waiting for it, this code won't run until that other method returns, and folderPlaceholdersToKeep is an output of that method.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
sanoursa: Yea a comment would be helpful, I think. And a functional test that fails if we add that else, because it's so very tempting to do so.

GVFS/GVFS.Common/GVFSConstants.cs
wilbaker: > That all makes sense, but what about the value itself?

The rules for the value:

- Must be 40 characters long (due to validation in `PlaceholderListDatabase.TryParseAddLine`)
- Must not be an actual valid SHA (otherwise we risk a file's SHA matching `ExpandedFolderValue` and VFSForGit incorrectly thinking the entry is for a folder)

We have a lot of freedom within those rules, is there a different string you'd prefer?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
```diff
@@ -237,7 +247,7 @@ public void CheckoutBranchAfterReadingAllFilesAndVerifyContentsCorrect()
                 .WithDeepStructure(this.FileSystem, this.ControlGitRepo.RootPath, compareContent: true);
 
             // Verify modified paths contents
-            GVFSHelpers.ModifiedPathsContentsShouldEqual(this.FileSystem, this.Enlistment.DotGVFSRoot, "A .gitattributes" + Environment.NewLine);
+            GVFSHelpers.ModifiedPathsContentsShouldEqual(this.FileSystem, this.Enlistment.DotGVFSRoot, "A .gitattributes\r\n");
```
wilbaker: Suggestion from Kevin: Use a named constant for "\r\n" here (and possibly in the the database code as well)

GVFS/GVFS.Common/GVFSConstants.cs
sanoursa: Isn't `0000000000000000000000000000000000000000` a valid sha? It seems like we'd have just as much chance of hitting `0000000000000000000000000000000000000001` as all-zeros, yea?

wilbaker: I believe that when @kewillford first added folders to the placeholder list he confirmed that the all-zero SHA will never be a valid SHA for a git blob.  @kewillford, can you confirm?

(And if it *could*, then I'd argue we're better off adding an upgrade step to convert those entries to a different invalid SHA rather than using 00...0001 for expanded folders)

sanoursa: Yep, why have a value here that even looks like a sha, if we want it to _not_ be a sha?

wilbaker: > There is a check in the git code when writing out the index that will error out if there is a null sha found in the index unless an environment variable is set. That said, not using something that looks like a sha at all for these seems like a better fit for our case especially now that there is more that one.

@kewillford @sanoursa, any objections to adding a major upgrade step to convert the all-zero SHAs to something that doesn't look like a sha?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -409,7 +409,7 @@ protected void RunFileDirectoryEnumerateTest(string command, string commandBranc
         protected void RunFileDirectoryReadTest(string command, string commandBranch = DirectoryWithFileAfterBranch)
         {
             this.SetupForFileDirectoryTest(commandBranch);
-            this.FileContentsShouldMatch("file.txt\\file.txt");
+            this.FileContentsShouldMatch("file.txt", "file.txt");
```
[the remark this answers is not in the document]
wilbaker: Yep it's happening inside `FileContentsShouldMatch`:

```
        protected void FileContentsShouldMatch(params string[] filePathPaths)
        {
            string filePath = Path.Combine(filePathPaths);
```

GVFS/GVFS.Tests/Should/EnumerableShouldExtensions.cs
```diff
@@ -122,6 +122,11 @@ public static IEnumerable<T> ShouldMatchInOrder<T>(this IEnumerable<T> group, IE
             return group;
         }
 
+        public static IEnumerable<T> ShouldMatchInOrder<T>(this IEnumerable<T> group, params T[] expectedValues)
+        {
+            return group.ShouldMatchInOrder(expectedValues, (t1, t2) => t1.Equals(t2));
```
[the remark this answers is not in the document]
wilbaker: Good idea, I had trouble doing that before, but I think I forgot to cast `expectedValues` to an `IEnumerable`

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1084,7 +1089,7 @@ private void UpdatePlaceholders()
             using (ITracer activity = this.context.Tracer.StartActivity("UpdatePlaceholders", EventLevel.Informational, metadata))
             {
                 ConcurrentHashSet<string> folderPlaceholdersToKeep = new ConcurrentHashSet<string>();
-                ConcurrentBag<PlaceholderListDatabase.PlaceholderData> updatedPlaceholderList = new ConcurrentBag<PlaceholderListDatabase.PlaceholderData>();
+                ConcurrentDictionary<string, PlaceholderListDatabase.PlaceholderData> updatedPlaceholderList = new ConcurrentDictionary<string, PlaceholderListDatabase.PlaceholderData>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
wilbaker: > Also how much overhead does adding the string key add? 

Let me measure that.  If it's significant we might need to further split this code out between Mac and Windows (because we do need the dictionary on Mac).

> Should this be a ConcurrentDictionary<string, string> and we build the PlaceholderListDatabase.PlaceholderData when calling WriteAllEntriesAndFlush?

That's a good idea, let me give that a try before making any perf measurements.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1094,14 +1099,25 @@ private void UpdatePlaceholders()
...
-                foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in placeholderListCopy.Where(x => x.IsFolder).OrderByDescending(x => x.Path))
-                {
-                    this.TryRemoveFolderPlaceholder(folderPlaceholder, updatedPlaceholderList, folderPlaceholdersToKeep);
+                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.blobSizes.CreateConnection())
+                {
+                    // Waiting for the UpdateOrDeleteFilePlaceholder to fill the folderPlaceholdersToKeep before trying to remove folder placeholders
+                    // so that we don't try to delete a folder placeholder that has file placeholders and just fails
+                    IEnumerable<PlaceholderListDatabase.PlaceholderData> folderPlaceholderData = placeholderListCopy.Where(x => x.IsFolder);
+
+                    HashSet<string> folderPlaceholders = new HashSet<string>(folderPlaceholderData.Select(x => x.Path), StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
wilbaker: Let me measure that as well.

One question I had looking at this code, do you remember why the folder placeholder list is ordered by descending before calling TryRemoveFolderPlaceholder? 

I assumed it was being done so that child folders would be removed before their parents, does that sound right to you?

>  is now being enumerated twice. Here and the line below

Another option I'll look into is removing this call (higher up in the method):

``` placeholderListCopy.Where(x => !x.IsFolder).ToList(),```

And replacing it with code that will walk `placeholderListCopy` to build both a file list and the `HashSet`.

File: GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs [hunk 14]
```diff
@@ -1289,14 +1395,17 @@ private string GetNewProjectedShaForPlaceholder(string path)
                     metadata.Add("result.RawResult", result.RawResult);
                     metadata.Add("UpdateFailureCause", failureReason.ToString());                    
                     this.context.Tracer.RelatedEvent(EventLevel.Informational, nameof(this.TryRemoveFolderPlaceholder) + "_DeleteFileFailure", metadata);
+
+                    updatedPlaceholderList.TryAdd(placeholder.Path, placeholder);
```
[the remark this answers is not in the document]
wilbaker: I'm glad you mentioned this, the idea was that VFSForGit should still track the placeholder if it fails to delete it, but I think I'll replace this line with a comment that references #245 and update #245 to include handling failures to delete folder placeholders.

I think that on Mac if we fail to delete the placeholder folder we should add its path to ModifiedPaths.dat and show the user a message (otherwise the folder will essentially be abandoned and never be re-expanded or deleted again).

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
kewillford: Yes, we have to make sure all the children are removed first so that we can remove the folder when it is empty since we are only removing individual folders and are depending on getting the directory not empty error if we can't delete because of existing files or folders.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
wilbaker: (I will dig more into our failure options when working on #245)

GVFS/GVFS.Common/GVFSConstants.cs
sanoursa: I think that's a good idea. And I would request that these special values that don't look like they are part of the path when manually looking at the PlaceholderList file. (i.e. put some delimiter char(s) at the beginning of the value)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
sanoursa: Do we have tests that fail if we drop that ordering? This is a very subtle detail that is easy to lose track of over time.

GVFS/GVFS.Common/GVFSConstants.cs
wilbaker: > I think that's a good idea. And I would request that these special values that don't look like they are part of the path when manually looking at the PlaceholderList file. (i.e. put some delimiter char(s) at the beginning of the value)

Sounds good, I'll start the values with `@`

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
wilbaker: > Do we have tests that fail if we drop that ordering? This is a very subtle detail that is easy to lose track of over time.

I'll try taking that out and see if any tests fail.  The tricky part is even if the ordering were to be removed, I'm not sure that we can come up with a test that would be guaranteed to catch it (as whatever random ordering VFSForGit would use in the future might just happen to work with the test).

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
wilbaker: > Should this be a ConcurrentDictionary<string, string> and we build the PlaceholderListDatabase.PlaceholderData when calling WriteAllEntriesAndFlush

I'm actually going to start with the perf overhead (starting that testing this morning) and see how things look before and after this change.  We're using the same string instances in both the `ConcurrentDictionary` and the `PlaceholderData` and so the memory overhead is just an extra reference.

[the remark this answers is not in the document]
wilbaker: > Should this be a ConcurrentDictionary<string, string> and we build the PlaceholderListDatabase.PlaceholderData when calling WriteAllEntriesAndFlush?

Also, to confirm, what was the main concern here? Perf or memory usage?

wilbaker:
@sanoursa @kewillford 

Here are the initial numbers comparing the total time of `UpdatePlaceholders`:

Results | Baseline | With Changes
------------ | ------------- | -------------
Average | 4056ms | 4735ms
Min | 3436ms | 2985ms
Max | 4619ms | 5767ms
StdDev | 354ms | 866ms
Count | 11 | 11

We are noticabley worse with the changes, and so I'll keep digging into where the time is going.

*Details*

- The time reported is the time of the full `UpdatePlaceholders` procedure
- File placeholders: 404,130
- Folder placeholders: 60,612
- Ran `git checkout` to move between two branches with the same commit (and so no placeholder would need to be updated on disk)

wilbaker:
@sanoursa @kewillford I just pushed a commit to get our times back down when running on Windows:

Results | Latest Commit
------------ | -------------
Average | 4019ms
Min | 2944ms
Max | 4935ms
StdDev | 638ms
Count | 11

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -218,18 +278,29 @@ private bool TryParseRemoveLine(string line, out string key, out string error)
...
 
             public string Path { get; }
             public string Sha { get; }
 
             public bool IsFolder
             {
-                get { return this.Sha == GVFSConstants.AllZeroSha; }
+                get 
+                {
+                    return this.Sha.StartsWith(FolderValuePrefix, StringComparison.Ordinal);
```
sanoursa: Why not just compare the full strings? Are the prefixes creating a significant perf win?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1078,30 +1087,85 @@ private void UpdatePlaceholders()
...
 
-                // Waiting for the UpdateOrDeleteFilePlaceholder to fill the folderPlaceholdersToKeep before trying to remove folder placeholders
-                // so that we don't try to delete a folder placeholder that has file placeholders and just fails
-                foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in placeholderListCopy.Where(x => x.IsFolder).OrderByDescending(x => x.Path))
+                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.blobSizes.CreateConnection())
+                {
+                    // A hash of the folder placeholders is only required if the platform expands directories
+                    HashSet<string> folderPlaceholders = 
+                        GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories ?
+                                    new HashSet<string>(placeholderFoldersListCopy.Select(x => x.Path), StringComparer.OrdinalIgnoreCase) :
```
sanoursa: Nit: fix the spacing

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1078,30 +1087,85 @@ private void UpdatePlaceholders()
...
+                int numThreads = Math.Max(8, Environment.ProcessorCount);
+                numThreads = Math.Min(numThreads, placeholderFilesListCopy.Count / minItemsPerThread);
+                numThreads = Math.Max(numThreads, 1);
+
                 ConcurrentHashSet<string> folderPlaceholdersToKeep = new ConcurrentHashSet<string>();
-                ConcurrentBag<PlaceholderListDatabase.PlaceholderData> updatedPlaceholderList = new ConcurrentBag<PlaceholderListDatabase.PlaceholderData>();
+
+                // On platforms that expand on enumeration we need to keep the updated placeholders in a collection
+                // that supports fast searching (for ReExpandFolder)
+                ConcurrentDictionary<string, PlaceholderListDatabase.PlaceholderData> updatedPlaceholderDictionary;
```
sanoursa: It's a little concerning that we're keeping two data structures for the same data. How do we guarantee that they stay in sync?

sanoursa: Or is the requirement that it be one or the other? In that case, I'd add a few strategic checks that enforce that and throw an exception if it's violated.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1078,30 +1087,85 @@ private void UpdatePlaceholders()
...
-                foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in placeholderListCopy.Where(x => x.IsFolder).OrderByDescending(x => x.Path))
+                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.blobSizes.CreateConnection())
+                {
+                    // A hash of the folder placeholders is only required if the platform expands directories
+                    HashSet<string> folderPlaceholders = 
+                        GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories ?
+                                    new HashSet<string>(placeholderFoldersListCopy.Select(x => x.Path), StringComparer.OrdinalIgnoreCase) : 
+                                    null;
+                    
+                    foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in placeholderFoldersListCopy.OrderByDescending(x => x.Path))
```
sanoursa: Is this `OrderByDescending` here to ensure that we walk through the tree from the bottom up? I would add some comments to clarify that.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1257,15 +1319,104 @@ private string GetNewProjectedShaForPlaceholder(string path)
...
+            HashSet<string> existingFolderPlaceholders)
+        {
+            FolderData folderData;
+            if (!this.TryGetOrAddFolderDataFromCache(relativeFolderPath, out folderData))
+            {
+                // Folder is no longer in the projection
+                return;
+            }
+
+            folderData.PopulateSizes(
```
sanoursa: Aren't we already gathering the necessary sizes elsewhere? How do we ensure that we're not doing lots of repeated round trips to the server here?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1257,15 +1319,104 @@ private string GetNewProjectedShaForPlaceholder(string path)
...
-                updatedPlaceholderList.Add(placeholder);
-                return;
+                addPlaceholderToUpdatedPlaceholders(placeholder);
+                return false;
+            }
+
+            if (GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories)
+            {
+                // If enumeration expands directories we should leave folder placeholders
+                // that are still in the projection on disk (even if they are empty).
```
sanoursa: Why do we leave them even if they are empty? How then do empty directories after a checkout get cleaned up?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -411,7 +411,7 @@ PrjFS_Result PrjFS_DeleteFile(
     
     char fullPath[PrjFSMaxPath];
     CombinePaths(s_virtualizationRootFullPath.c_str(), relativePath, fullPath);
-    if (0 != unlink(fullPath))
+    if (0 != remove(fullPath))
```
sanoursa: Why was this changed?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: > Why do we leave them even if they are empty?

I think should update this comment.  "Empty" here means that on the physical disk the folder is empty (e.g. because the folder has not been expanded yet).

> How then do empty directories after a checkout get cleaned up?

If they are empty from git's perspective they will *not* be in the projection, and so `TryGetOrAddFolderDataFromCache` will return false and VFSForGit will proceed with the delete of the folder.

(On Windows we can't make this "in projection" check because we also have to worry about tombstones that might be on disk)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: Filed #255 for this

GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
```diff
@@ -348,16 +378,16 @@ private string[] GetPlaceholderDatabaseLinesBeforeUpgrade(string placeholderData
...
-            lines.ShouldContain(x => x == "A GVFS\\GVFS.Common\0" + TestConstants.AllZeroSha);
-            lines.ShouldContain(x => x == "A GVFS\\GVFS.Common\\Git\0" + TestConstants.AllZeroSha);
-            lines.ShouldContain(x => x == "A GVFS\\GVFS.Tests\0" + TestConstants.AllZeroSha);
-            lines.ShouldContain(x => x == "A GVFS\\GVFS.Tests\\Properties\0" + TestConstants.AllZeroSha);
+            lines.ShouldContain(x => x == "A Scripts\0" + TestConstants.PartialFolderPlaceholderDatabaseValue);
+            lines.ShouldContain(x => x == "A GVFS\0" + TestConstants.PartialFolderPlaceholderDatabaseValue);
+            lines.ShouldContain(x => x == "A GVFS\\GVFS.Common\0" + TestConstants.PartialFolderPlaceholderDatabaseValue);
+            lines.ShouldContain(x => x == "A GVFS\\GVFS.Common\\Git\0" + TestConstants.PartialFolderPlaceholderDatabaseValue);
+            lines.ShouldContain(x => x == "A GVFS\\GVFS.Tests\0" + TestConstants.PartialFolderPlaceholderDatabaseValue);
+            lines.ShouldContain(x => x == "A GVFS\\GVFS.Tests\\Properties\0" + TestConstants.PartialFolderPlaceholderDatabaseValue);
```
kewillford: Is there some way to group some of these so we don't have to repeat 2 more times?

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -13,6 +13,8 @@ public abstract class FileSystemVirtualizer : IDisposable
     {
         public const byte PlaceholderVersion = 1;
 
+        protected static readonly byte[] FolderContentId = Encoding.Unicode.GetBytes(GVFSConstants.AllZeroSha);
```
kewillford: Why did this say as `AllZeroSha`?

wilbaker: This is what's actually stored on disk in the folder placeholders.  I wanted to avoid changing the on-disk format of the folder placeholders themselves.

GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/DiskLayoutUpgradeTests.cs
wilbaker: That's a good idea, let me see how I can clean these up.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1374,7 +1554,7 @@ private string GetNewProjectedShaForPlaceholder(string path)
             PlaceholderListDatabase.PlaceholderData placeholder,
```
kewillford: Not for this PR but the method name `ProcessGvUpdateDeletePlaceholderResult` needs to change at some point.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
wilbaker: > So we never check this value and don't expect to need to tell the difference using ContentId whether a folder is partial or expanded?

That's right.  We're currently only checking the contentID of files (when we get a hydration request).

On Mac the kext uses the`FileFlags_IsEmpty` flag to check if the directory has been expanded or not.

```if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))```

I'd rather not change this now unless we have to (to leave things cleaner for ourselves if we do need to change it some day).

kewillford: Sounds good.  Thanks!

PR #240, opened 2018-09-04
Allow upgrade when release contains an older version of Git.

alameenshah:


GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -45,15 +45,6 @@ public bool TryRunPreUpgradeChecks(GitVersion gitVersion, out string error)
...
-
-            this.tracer.RelatedInfo("Checking if Git upgrade can be run on this machine.");
-            if (gitVersion != null && !this.IsGitUpgradeAllowed(gitVersion, out error))
-            {
-                this.tracer.RelatedError($"{nameof(TryRunPreUpgradeChecks)}: {error}");
-                return false;
-            }
-
-            this.tracer.RelatedInfo("Okay to run Git upgrade.");
             this.tracer.RelatedInfo("Successfully finished pre upgrade checks.");
```
wilbaker: Should we keep both of these trace statements?

```
this.tracer.RelatedInfo("Okay to run GVFS upgrade.");
this.tracer.RelatedInfo("Successfully finished pre upgrade checks.");
```

Or can they be collapsed them into one?

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -350,24 +353,30 @@ private bool TryRunInstallerForAsset(string name, out int exitCode, out string e
...
-
             foreach (Asset asset in this.newestRelease.Assets)
             {
                 string extension = Path.GetExtension(asset.Name);
                 if (extension != null && extension == ".exe")
                 {
-                    if ((name == GitAssetNamePrefix && asset.Name.StartsWith(GitInstallerFileNamePrefix)) ||
-                        (name == GVFSAssetNamePrefix && asset.Name.StartsWith(GVFSInstallerFileNamePrefix)))
+                    path = asset.LocalPath;
+                    if (name == GitAssetNamePrefix && asset.Name.StartsWith(GitInstallerFileNamePrefix))
```
wilbaker: Not new to this PR, but you might want to use the version of `StartsWith` that takes a `StringComparison` so that we can ignore case when comparing (I doubt the case will change, but it might be a bit more robust, let me know what you think).

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -350,24 +353,30 @@ private bool TryRunInstallerForAsset(string name, out int exitCode, out string e
...
 
-        private bool TryGetLocalInstallerPath(string name, out string path)
+        private bool TryGetLocalInstallerPath(string name, out string path, out string args)
         {
-            path = null;
-
             foreach (Asset asset in this.newestRelease.Assets)
             {
                 string extension = Path.GetExtension(asset.Name);
                 if (extension != null && extension == ".exe")
```
wilbaker: There is no need to check `extension != null`

GVFS/GVFS.Common/ProductUpgrader.cs
wilbaker: (Same comment for the .exe check)

PR #248, opened 2018-09-06
Picking up changes from master.

alameenshah:
Merging with master


kewillford:
What is this commit `Merge branch 'features/auto_upgrade' into temp` and why was it added?  Is there a better description we can use for it?

PR #258, opened 2018-09-11
Remove Windows specific code from UpdatePlaceholderTests and enable tests on Mac

wilbaker:
- Removed use of Windows specific `CreateFile`
- Tests that depend on Windows file sharing behavior marked as `WindowsOnly`

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/UpdatePlaceholderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/UpdatePlaceholderTests.cs
@@ -8,12 +8,12 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 
 namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
 {
     [TestFixture]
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/UpdatePlaceholderTests.cs
```diff
@@ -34,18 +34,18 @@ public virtual void SetupForTest()
             this.GitStatusShouldBeClean(NewFilesAndChangesCommitId);
         }
 
+        // WindowsOnly because test requires Windows specific file sharing behavior
         [TestCase, Order(1)]
```
sanoursa: How important is the relative/interleaved ordering of all the tests in this class? The code would definitely be simpler if we pull all the WindowsOnly tests out into a separate class.

wilbaker: I don't think the ordering is that important, and I like the idea.  

I will see if I can pull all of the WindowsOnly tests out into a separate class that lives in the Windows specific functional test project (that also gives us the option of keeping the code that uses `CreateFile` as well).  

Thanks!

Issue #142, opened 2018-07-31
Harden parsing of messages from pipe


jamill:
This bug presents itself when you stage a region in VS Code.

Open a GVFS Repository with VS Code
Make some changes to a file
Select a region of the file in the editor, right click and “Stage selected range”

This is the command run:

```
/.git/hooks/pre-command.exe update-index --cacheinfo 100644 '7e012a87091a0f76de43fcbd83098b3611791643
' Scripts/ReinstallGVFS.bat --git-pid=20360
```

Notice the newline in the command. 

The InMount process does not like this, and crashes with something similar to (this doesn't necessarily correspond to the instance above):

```
The parsedCommand is an unexpected length, got: 123 from message: '7580|False|False|123|git update-index --cacheinfo 100644 f7ce833c943b1d4e33fce3c842236c9c044561c9'\r\n
```
If this is a valid format for a git command, then VFS for Git should be able to parse it.

wilbaker:
I believe the root cause of the issue in VFSForGit is that it uses the newline character to separate messages on the pipe (and uses `ReadLine` to read in new messages).

The hook process is not stripping the newline before passing the command to VFSForGit, resulting in it only reading in a part of the message.

benpeart:
I don't know of one specifically but there could be a situation where stripping the newline out will cause other/different problems. In general, we've had to use a nul to indicate EOL (and a double nul to indicate the end of the list) when communicating with git and other hooks (see the read object hook for an example).

PR #259, opened 2018-09-11
Tweak Named pipe protocol

jamill:
Tweak the protocol used for communicating over Named Pipes. Named Pipes is the mechanism used for interprocess communication between the git hooks, mount process, service, etc... This is to address problems of transmitting messages that include newline characters (i.e. #142).

Data is currently encoded / transmitted using a `TextWriter` derived class, which uses newline characters to delineate the end of a line of text. This approach will have trouble transmitting messages that include newline characters as part of the message. Any encoding scheme we choose will need to replicated for both a managed and native implementation.

To address this issue, this PR switches to writing bytes into the named pipe (instead of text into a `StreamWriter`), and switching the byte value used to delineate the end of a message to `0x3` (ASCII code for End of text). Text will be encoded via UTF-8, as that is the format currently used implicitly via the `StreamWriter`. This will result in the minimal change, and not require further reactions on the native pipe clients to handle a different encoding.


GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+                    bytes.AddRange(new ArraySegment<byte>(this.buffer, 0, bytesRead - 1));
+                    break;
+                }
+
+                bytes.AddRange(new ArraySegment<byte>(this.buffer, 0, bytesRead));
+                bytesRead = this.Read();
+                Debug.Assert(bytesRead != 0, "Not expecting 0 bytes at this time");
+            }
+
+            return Encoding.UTF8.GetString(bytes.ToArray());
```
sanoursa: It seems like the real value add of this overall change is using a different terminator character. What additional protection do we get from using UTF8 encoding? At first glance, it appears that this will just add perf overhead.

NamedPipeProtocol.md
```diff
@@ -0,0 +1,15 @@
+# The GVFS Named Pipe Protocol (v1)
```
sanoursa: I don't think we need to create a protocol doc for this. This feels like it could just be a class or method header comment up above.

This is not a standard protocol that we expect any other service or product to implement, and it doesn't even hold true across different versions of VFS for Git (meaning this protocol itself doesn't need to be versioned). This is just the internal details of how various processes in our codebase talk to each other.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: Good points - you are correct - the main value of this change is using a different terminator character. We don't need to use UTF8 (it doesn't add any additional protection) - we just need the sender / receiver to agree on the encoding. There are some reasons I chose UTF8 for this iteration:

1. GVFS currently sends bytes as UTF8 (implicitly via the `StreamReader` / `StreamWriter` classes). Using UTF8 again would be keeping with current behavior, so there shouldn't be any additional overhead.

2. The native code that reads / writes data from the named pipe treats the data as ASCII (which I understand is a subset of UTF8). If we switched away from UTF8, we would need to do convert the encoding in the native code.

3. The ModifiedPathsList hook also reads data from the named pipe and directly sends those bytes to git, which works with UTF8 (at least for our current scenarios), but would not work with UTF16 (which is the default representation of strings in C#).

I propose we continue using UTF8 (explicitly now) for this change? We can consider if a different encoding would make more sense...

NamedPipeProtocol.md
jamill: I agree - this wasn't meant to be versioned or for external components. My intention was to define the behavior in a single place for internal components. I will move this into a class comment - I think that would be a better fit.

jamill:
> I don't think this is really a new protocol per se, 

Right - this is just a tweak / refinement of the existing behavior. I can tweak the PR description to make the intention more clear. 

> Changing the encoding could also be a reasonable thing to do, as long as it's meaningfully improving reliability and not overly hurting perf.

I wasn't trying to change the encoding (as you point out - UTF8 encoding doesn't afford extra protection) - but am just being explicit about it now. GVFS currently uses UTF8 (implicitly via the `StreamWriter`), and the native readers currently expect ASCII / UTF8. I can also include this in the PR description to make this point more clear.

GVFS/GVFS.ReadObjectHook/main.cpp
```diff
@@ -18,9 +18,9 @@
 #define DLO_REQUEST_LENGTH (4 + SHA1_LENGTH + 1)
 
 // Expected response:
-// "S\n" -> Success
-// "F\n" -> Failure
-#define DLO_RESPONSE_LENGTH 2
+// "S\x3" -> Success
+// "F\x3" -> Failure
+#define DLO_RESPONSE_LENGTH 3
```
wilbaker: Why was this changed from 2 to 3?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/UpdateIndexTests.cs
```diff
@@ -63,6 +63,16 @@ public void UpdateIndexRemoveAddFileOpenForWrite()
             GitHelpers.InvokeGitAgainstGVFSRepo(this.Enlistment.RepoRoot, "update-index --add Test_ConflictTests/AddedFiles/AddedByBothDifferentContent.txt");
         }
 
+        [TestCase]
+        [Category(Categories.MacTODO.M4)]
```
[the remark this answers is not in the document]
jamill: I have not tried running this on the Mac - I will test and see what happens (or remove this, if possible)

GVFS/GVFS.ReadObjectHook/main.cpp
jamill: This should be a 2 - I will undo that

jamill:
This is ready for review now. Thanks for the initial feedback on the approach!

/cc @sanoursa @benpeart @wilbaker

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                if (this.buffer[bytesRead - 1] == TerminatorByte)
+                {
+                    bytes.AddRange(new ArraySegment<byte>(this.buffer, 0, bytesRead - 1));
+                    break;
+                }
+
+                bytes.AddRange(new ArraySegment<byte>(this.buffer, 0, bytesRead));
+                bytesRead = this.Read();
+                Debug.Assert(bytesRead != 0, "Not expecting 0 bytes at this time");
```
sanoursa: Please use exceptions rather than debug asserts

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            while (true)
+            {
+                if (this.buffer[bytesRead - 1] == TerminatorByte)
+                {
+                    bytes.AddRange(new ArraySegment<byte>(this.buffer, 0, bytesRead - 1));
+                    break;
+                }
+
+                bytes.AddRange(new ArraySegment<byte>(this.buffer, 0, bytesRead));
```
sanoursa: What are the perf implications, if any, of using an `ArraySegment` here?

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public string ReadMessage()
+        {
+            List<byte> bytes = new List<byte>();
+
+            int bytesRead = this.Read();
+            if (bytesRead == 0)
+            {
+                return null;
```
sanoursa: Is this the correct behavior? Or should we assert/throw like we do in the loop below? Either way, it seems like they should be consistent with each other.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+                Debug.Assert(bytesRead != 0, "Not expecting 0 bytes at this time");
+            }
+
+            return Encoding.UTF8.GetString(bytes.ToArray());
+        }
+
+        /// <summary>
+        /// Read the next chunk of bytes from the stream.
+        /// </summary>
+        /// <returns></returns>
```
sanoursa: Fill this in or remove the empty line

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <summary>
+    ///   The GVFS NamePipeProtocol:
+    ///       GVFS can send "messages" over the wire. A message is defined as a
+    ///       sequence of UTF8 encoded text whose end is delineated by a TerminatorByte.
+    /// </summary>
+    public class NamedPipeStreamReader
+    {
+        public const int BufferLength = 1024;
+
+        private const byte TerminatorByte = 0x3;
```
sanoursa: Have you considered a design where you include the message length as the first part of the message, rather than use a terminator byte? It seems like that could give us better perf when reading messages - especially since there's no need for a `List`, along with dynamic reallocations, and then converting it back to an array afterwards - but I'm not sure how big that perf impact is, necessarily.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public NamedPipeStreamReader(Stream stream)
+        {
+            this.stream = stream;
+            this.buffer = new byte[NamedPipeStreamReader.BufferLength];
+        }
+
+        public string ReadMessage()
+        {
+            List<byte> bytes = new List<byte>();
```
sanoursa: If you do stick with using a `List`, I would figure out how many bytes we're using for our current set of messages, and preallocate the list with at least that much room, to avoid having to do a reallocation while reading a message.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace GVFS.Common.NamedPipes
+{
+    /// <summary>
```
sanoursa: I would drop this comment, since it duplicates the info on `NamedPipeServer`. Let's document it in one place.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,67 @@
...
+    ///       sequence of UTF8 encoded text whose end is delineated by a TerminatorByte.
+    /// </summary>
+    public class NamedPipeStreamReader
+    {
+        public const int BufferLength = 1024;
+
+        private const byte TerminatorByte = 0x3;
+
+        private byte[] buffer;
+        private int bufferContentLength;
```
sanoursa: I don't see this variable being used anywhere. Drop it?

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: Yes - I did consider that approach (and in general, I like that approach) - however, it would require more substantive changes. This change focuses on correctness of the underlying protocol to address the associated issue, without regressing performance. This follows the same approach as used by the current `StreamReader` implementation. We have other  more impactful items to address for performance before we need to focus on the wire protocol here (at least, I am not aware of this popping up on any perf investigations).

I do like your suggestion to size the `List` based on the expected message size (or even a focused optimization to avoid allocating a `List` if the entire message is read in the initial read).

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: This one here is correct / expected behavior. This indicates the stream has been closed.

We don't expect the stream to close in the middle of reading a message (e.g. before reading the terminating byte). I will add a code comment here, and tweak the logic in the other location to throw an exception.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: ah - yes I meant to do this. Will update in next iteration.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: How do we want to handle "unexpected" messages  - messages that do not end with the expected terminating byte? 

We could throw an exception - but then I suspect we also would want a caller to catch this exception and trace what happened. Alternatively, we could trace this occurance and return null...

Preferences?

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: `ArraySegment` is a struct that contains a reference to the underlying array and an offset into it - It is not a full copy of the array.

`ArraySegment` also Implements `ICollection`, which means `AddRange` can copy the items using `CopyTo` and `Array.Copy`, instead of inserting items individually (If I am reading the underlying source correctly).

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
sanoursa: We already handle and log exceptions from all thread entry points, so throwing is equivalent to: Something went horribly wrong, log it and shut down.

If this situation ever happens, we've got a badly misbehaving client, or potentially mismatched versions installed on a box. In any case, it's extremely dangerous to keep going, so throwing seems like the best option.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
sanoursa: Why would that be a more substantive change? My naive reading of the code leads me to think that the changes would all be contained in NamedPipeStreamReader/Writer. Are there more complications?

jamill: There are also native clients that implement this protocol, and they would need to be updated as well.

sanoursa: Ah that makes sense, thanks. So I guess the question is: do we think that would be a better solution? And if so, is there still value in building an interim solution here? This is not a pressing need, so I would vote that we build whichever one we think is the better long term fix.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                bytesRead = this.Read();
+
+                if (bytesRead == 0)
+                {
+                    // We have read a partial message (the last byte received does not indicate that
+                    // this was the end of the message), but the stream has been closed. Throw an exception
+                    // and let upper layer deal with this condition.
+
+                    throw new IOException("Incomplete message read from stream. Stream closed before message terminating byte.");
```
sanoursa: I would throw a `BrokenPipeException` here

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary>
+        /// Read a message from the stream.
+        /// </summary>
+        /// <returns>The message read from the stream, or null if the end of the input stream has been reached. </returns>
+        public string ReadMessage()
+        {
+            int bytesRead = this.Read();
+            if (bytesRead == 0)
+            {
+                // The end of the input stream has been reach - return null to indicate this.
```
sanoursa: Typo nitpick: "reached"

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Text;
+
+namespace GVFS.Common.NamedPipes
+{
+    /// <summary>
+    /// Implements the NamedPipe protocol as described in NamedPipeServer.
+    /// </summary>
+    public class NamedPipeStreamReader
+    {
+        public const int BufferLength = 1024;
```
sanoursa: I think the only external code that uses this is unit tests, so please make it private. In my opinion it's not a good idea to share constants between production and test code, because they will just end up agreeing with each other no matter what.

NamedPipeProtocol.md
sanoursa: It looks like this file got added back in?

GVFS/GVFS.Common/NamedPipes/NamedPipeServer.cs
```diff
@@ -200,10 +201,11 @@ public string ReadRequest()
             {
                 try
                 {
-                    return this.reader.ReadLine();
+                    return this.reader.ReadMessage();
                 }
-                catch (IOException)
+                catch (IOException ex)
                 {
+                    this.tracer.RelatedError($"Error reading message from NamedPipe: {ex.Message}");
```
wilbaker: Minor suggestions:
- Include the full stack trace
- Create an EventMetadata and add a value for the stack trace

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -0,0 +1,86 @@
...
+                if (encounteredTerminatorByte)
+                {
+                    break;
+                }
+
+                bytesRead = this.Read();
+
+                if (bytesRead == 0)
+                {
+                    // We have read a partial message (the last byte received does not indicate that
```
wilbaker: Should we log the partial message?

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: I was trying to enforce wat that we test sending messages that are larger than the buffer size. If we completely hide this, then we could update the production code to use a buffer larger than what the test is using.

What if we instead expose an option for callers to control the buffer size. Then the tests could force a smaller buffer size and ensure it is testing the expected functionality...

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: I can do that - I was trying to decide between those two exceptions. Currently, `NamedPipeClient` and `NamedPipeServer` both handle `IOException` , which is why I chose this exception. If I switch to `BrokenPipeException`, I will also update the handling code in `NamedPipeServer` to handle `BrokenPipeExceptions` in the same manner as it currently does for `BrokenPipeExceptions`.

I can go with either approach - do you have a preference?

sanoursa: I'm guessing we were handling IOExceptions before because that's what the built in reader/writer were throwing. I would update this to throw/catch our specific exception type, and remove the handler for IOException unless there's still a valid reason for it.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
sanoursa: I wouldn't do either of those things. A unit test should be testing a _requirement_, not knowledge of an _implementation_. So the unit test should decide how big of a message is reasonable to send and test that limit, and the implementation is then responsible for handling that.

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
jamill: Yeah - the underlying Stream class can still throw `IOException`. We could have the `NamedPipeStreamReader` and `NamedPipeStreamWriter` class catch and wrap all underlying `IOException`s as `BrokenPipeException`s.

Would you be OK with keeping this as an IOException for now, which would allow us to keep all existing error handling in place? I think it would be easier if we extracted that change into its isolated change, so we can see the update to the exception type (and all handlers) by itself?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/UpdateIndexTests.cs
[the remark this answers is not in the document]
jamill: This test fails on the "cleanup" code (not related to the protocol changes). After "resetting" the repository, there are unexpected changes in the test repository. Because of this (and because the failure is not related to the protocol changes), I will leave this as a to for Mac

PR #261, opened 2018-09-12
Clean up code that was referring to sparse-check our always_exclude

kewillford:
Since VFSForGit no longer is using the sparse-checkout or always_exclude to track files and folders and instead is using the ModifiedPaths, this cleans up where we were incorrectly referencing them or the code was no longer used.

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/Git/GitIndexGenerator.cs
+++ b/GVFS/GVFS.Common/Prefetch/Git/GitIndexGenerator.cs
@@ -109,8 +109,7 @@ namespace GVFS.Common.Prefetch.Git
                     LsTreeEntry entry;
                     while (this.entryQueue.TryTake(out entry, Timeout.Infinite))
                     {
-                        bool skipWorkTree = false;
-                        this.WriteEntry(writer, version, entry.Sha, entry.Filename, skipWorkTree, ref lastStringLength);
+                        this.WriteEntry(writer, version, entry.Sha, entry.Filename, ref lastStringLength);
                     }
 
                     // Update entry count
@@ -140,7 +139,7 @@ namespace GVFS.Common.Prefetch.Git
             return filename.Substring(0, idx + 1);
         }
 
-        private void WriteEntry(BinaryWriter writer, uint version, string sha, string filename, bool skipWorktree, ref uint lastStringLength)
+        private void WriteEntry(BinaryWriter writer, uint version, string sha, string filename, ref uint lastStringLength)
         {
             long startPosition = writer.BaseStream.Position;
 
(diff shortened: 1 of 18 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -65,10 +65,7 @@ public void DehydrateShouldBackupFiles()
 
             // .git folder items
             string gitFolder = Path.Combine(backupFolderItems[0], ".git");
-            this.DirectoryShouldContain(gitFolder, "index", "info");
-
-            string gitInfoFolder = Path.Combine(gitFolder, "info");
-            this.DirectoryShouldContain(gitInfoFolder, "sparse-checkout");
```
wilbaker: Unrelated to your changes, does the sparse-checkout do anything when the ModifiedPaths.dat file is being used?

kewillford: The file is not doing anything when using the virtual projection feature instead but git is still using some of the sparse-checkout code internally.  I think for git it is either virtual projection or sparse-checkout since the virtual projection turns on sparse-checkout internally. @benpeart is that right?

benpeart: Correct.  The virtual projection uses the internal skip-worktree bit (and excludes) logic but just sets the bits using different logic (ie it doesn't use the sparse-checkout and always-excludes files, but the virtual projection file instead).

GVFS/GVFS.Common/Prefetch/Git/GitIndexGenerator.cs
```diff
@@ -109,10 +109,7 @@ private void WriteAllEntries(uint version, HashSet<string> sparseCheckoutEntries
                     LsTreeEntry entry;
                     while (this.entryQueue.TryTake(out entry, Timeout.Infinite))
                     {
-                        bool skipWorkTree = 
-                            sparseCheckoutEntries != null && 
-                            !sparseCheckoutEntries.Contains(entry.Filename) && 
-                            !sparseCheckoutEntries.Contains(this.GetDirectoryNameForGitPath(entry.Filename));
+                        bool skipWorkTree = false;
                         this.WriteEntry(writer, version, entry.Sha, entry.Filename, skipWorkTree, ref lastStringLength);
```
sanoursa: It looks like you can remove the `skipWorktree` parameter now? And all of the code inside `WriteEntry` that deals with it.

mwasplund:
Does this mean that this documentation is also out of date?
https://docs.microsoft.com/en-us/azure/devops/learn/git/gvfs-architecture
I was playing around and noticed that GVFS was not generating either the sparse-check or the always_exclude files as documented.

PR #265, opened 2018-09-12
Update readme to include additional build badges

kewillford:


Readme.md
```diff
@@ -2,15 +2,16 @@
...
 
 ## Mac
 |Branch|Unit Tests|Functional Tests|
 |:--:|:--:|:--:|
-|**master**|[![Build status](https://gvfs.visualstudio.com/ci/_apis/build/status/CI%20-%20Mac%20-%20master?branchName=master)](https://gvfs.visualstudio.com/ci/_build/latest?definitionId=15)||
+|**master**|[![Build status](https://dev.azure.com/gvfs/ci/_apis/build/status/CI%20-%20Mac?branchName=master)](https://dev.azure.com/gvfs/ci/_build/latest?definitionId=15)|[![Build status](https://dev.azure.com/mseng/VSOnline/_apis/build/status/GVFS/GitHub%20VFSForGit%20Mac%20Functional%20Tests?branchName=master)](https://dev.azure.com/mseng/VSOnline/_build/latest?definitionId=7350)|
+|**shipped**|[![Build status](https://dev.azure.com/gvfs/ci/_apis/build/status/CI%20-%20Mac?branchName=releases%2Fshipped)](https://dev.azure.com/gvfs/ci/_build/latest?definitionId=15)||
```
wilbaker: Should we have a badge for shipped functional tests as well?

Readme.md
```diff
@@ -2,15 +2,16 @@
 
 ## Windows
 
-|Branch|Unit Tests|Functional Tests|
-|:--:|:--:|:--:|
-|**master**|[![Build status](https://gvfs.visualstudio.com/ci/_apis/build/status/CI%20-%20Windows%20-%20master?branchName=master)](https://gvfs.visualstudio.com/ci/_build/latest?definitionId=7)|[![Build status](https://gvfs.visualstudio.com/ci/_apis/build/status/CI%20-%20Windows%20-%20Full%20Functional%20Tests?branchName=master)](https://gvfs.visualstudio.com/ci/_build/latest?definitionId=6)|
-
+|Branch|Unit Tests|Functional Tests|Large Repo Perf|
```
wilbaker: Is the Large Repo Build test also a CI test?

Readme.md
```diff
@@ -2,15 +2,16 @@
 
 ## Windows
 
-|Branch|Unit Tests|Functional Tests|
-|:--:|:--:|:--:|
-|**master**|[![Build status](https://gvfs.visualstudio.com/ci/_apis/build/status/CI%20-%20Windows%20-%20master?branchName=master)](https://gvfs.visualstudio.com/ci/_build/latest?definitionId=7)|[![Build status](https://gvfs.visualstudio.com/ci/_apis/build/status/CI%20-%20Windows%20-%20Full%20Functional%20Tests?branchName=master)](https://gvfs.visualstudio.com/ci/_build/latest?definitionId=6)|
-
+|Branch|Unit Tests|Functional Tests|Large Repo Perf|
+|:--:|:--:|:--:|:--:|
+|**master**|[![Build status](https://dev.azure.com/gvfs/ci/_apis/build/status/CI%20-%20Windows?branchName=master)](https://dev.azure.com/gvfs/ci/_build/latest?definitionId=7)|[![Build status](https://dev.azure.com/gvfs/ci/_apis/build/status/CI%20-%20Windows%20-%20Full%20Functional%20Tests?branchName=master)](https://dev.azure.com/gvfs/ci/_build/latest?definitionId=6)|[![Build status](https://dev.azure.com/mseng/VSOnline/_apis/build/status/GVFS/GitHub%20VFSForGit%20Large%20Repo%20Perf%20Tests?branchName=master)](https://dev.azure.com/mseng/VSOnline/_build/latest?definitionId=7179)|
```
sanoursa: Why are there two badges in some of the cells?

Readme.md
kewillford: No, but it is scheduled to run nightly so decided to add it.

Readme.md
kewillford: What cells?  I only see one badge in each cell.

Readme.md
wilbaker: Sorry, what I meant was could we also add Large Repo Build test? (I'm not sure if it's run automatically as well)

Readme.md
sanoursa: I was looking here: https://github.com/Microsoft/VFSForGit/pull/265/files?short_path=1e290ac#diff-1e290ac8433d555bce009b162cb869d0

kewillford: Maybe the diff is causing that?  If you look at the file here https://github.com/Microsoft/VFSForGit/blob/1741cacf77230fa77701f90743fbdf0a601ead2c/Readme.md there is only one per cell.

Readme.md
kewillford: Sorry, I could add that but right now it is only ran on the PRs so rarely ran on master or shipped.  I could set up a CI build so that we would have that as well.  I could also setup the Large Repo Perf build to run as CI instead of nightly but running nightly makes it easy to find the last run against master to compare performance.

Readme.md
sanoursa: Thanks, that looks cleaner. Why is it that all of the badges within one column all take you to the same build results? Can we set it up so that each one takes you to the actual latest build for that cell?

Readme.md
wilbaker: I think the other advantage to nightly is reducing the contention for the one machine that can run the tests.  I don't think you need to change the Build test settings as part of this PR, I was thinking if it was already running nightly though we might as well add a badge for it.

Readme.md
kewillford: I will have to ask around and see if there is a way to do that.  I already have them looking into an issue with the wrong badge showing for the releases/shipped branch of the Large Repo Perf build.

Readme.md
kewillford: I added a scheduled nightly build for the Large Repo Build and going to add those build badges.

Readme.md
kewillford: It should work by adding `&branchName=X` but is not.  They have someone looking into it.  I will update all of the URLs so that at least those should be correct when they get it fixed.

PR #266, opened 2018-09-12
Update PowerShellRunner to support hard link creation and enable more functional tests on Mac

### LINKED ISSUE #262: Add hardlink creation support to PowerShellRunner
@kewillford found that PowerShell does support creating hardlinks in Windows 10:

```
New-Item -ItemType HardLink -Name <new link name> -Value <existing file>
```

We should add an implementation of `CreateHardLink` to `PowerShellRunner` that uses `New-Item`

wilbaker:
Fixes #262 

- Added `CreateHardLink` implementation to `PowerShellRunner`
- Removed OS check from `ModifiedPathsTests` when building list of runners
- Enabled `CheckoutTests.CheckoutBranchThatHasFolderShouldGetDeleted` test on Mac
- Enabled `GitCommands.RebaseTests` on Mac

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
@@ -120,7 +120,7 @@ A Scripts/
             }
         }
 
-        [TestCaseSource(typeof(HardLinkRunners), HardLinkRunners.TestRunners)]
+        [TestCaseSource(typeof(FileSystemRunner), FileSystemRunner.TestRunners)]
         public void ModifiedPathsCorrectAfterHardLinking(FileSystemRunner fileSystem)
         {
             const string ExpectedModifiedFilesContentsAfterHardlinks =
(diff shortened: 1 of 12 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
```diff
@@ -172,15 +171,22 @@ public static object[] Runners
             {
                 get
                 {
-                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    List<object[]> hardLinkRunners = new List<object[]>();
+                    foreach (object[] runner in FileSystemRunner.Runners.ToList())
                     {
-                        return new[]
+                        FileSystemRunner fileSystem = runner.ToList().First() as FileSystemRunner;
```
sanoursa: Nit pick: you're using `as` but not checking for null. I would just do a normal cast here so we end up with a more appropriate exception if someone ever inserts in invalid type into the list.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
```diff
@@ -172,15 +171,22 @@ public static object[] Runners
...
+                            hardLinkRunners.Add(new object[] { fileSystem });
+                        }
                     }
 
+                    if (hardLinkRunners.Count > 0)
+                    {
+                        return hardLinkRunners.ToArray();
+                    }
+
+                    // Always return at least one runner that supports creating hard links
```
sanoursa: Isn't it a bug in our code if we get this far? I would just throw.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
wilbaker: > I would just do a normal cast here so we end up with a more appropriate exception if someone ever inserts in invalid type into the list.

Sounds good I'll fix this up

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
wilbaker: @sanoursa this is not a bug because when we run without `--full-suite` the only runner will be `SystemIORunner` which does not support hard link creation.

I can either update the comment to include those details, or I can mark the test as full suite only, let me know if you have a preference.

sanoursa: Got it, yea a comment would be helpful

wilbaker: Actually, I just noticed that we had this same issue for `RenameDirectory` and addressed it by having `SystemIORunner.RenameDirectory` PInvoke to the native API.  I'll see if I can do the same for hard link creation and then all of this "does runner support hard link" code can be removed.

PR #267, opened 2018-09-13
Mac: Prevent deletes of the index by non-git processes and enable related functional tests

### LINKED ISSUE #222: Mac: Prevent applications other than git from replacing the index file
VFSForGit on Mac should work like VFSForGit on Windows and only allow the index to be deleted while a git command is is progress

wilbaker:
*Summary of changes*
- Block deletes of the index when the `GVFSLock` is not held by git
- Updated `BashRunner` to expect different messages on Windows and Mac when deletes are blocked
- Enabled `BasicFileSystemTests. DeleteIndexFileFails` test on Mac
- Enabled all `GVFSLockTests` tests on Mac

Closes #222

### DIFF
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
@@ -114,7 +114,7 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         {
             // BashRunner does nothing special when a failure is expected, so just confirm source file is still present
             this.MoveFile(sourcePath, targetPath);
-            this.FileExists(sourcePath).ShouldBeTrue();
+            this.FileExists(sourcePath).ShouldBeTrue($"{sourcePath} does not exist when it should");
         }
 
         public override void MoveFile_FileShouldNotBeFound(string sourcePath, string targetPath)
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
@@ -29,14 +29,12 @@ namespace GVFS.FunctionalTests.FileSystemRunners
             "Function not implemented"
         };
 
-        private static string[] permissionDeniedWindowsMessage = new string[]
+        private static string[] windowsPermissionDeniedMessage = new string[]
         {
             "Permission denied"
         };
 
-        // TODO(Mac): Update this message when the kext returns a more accurate
-        // error code
-        private static string[] permissionDeniedMacMessage = new string[]
+        private static string[] macPermissionDeniedMessage = new string[]
         {
             "Resource temporarily unavailable"
         };
@@ -116,7 +114,7 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         {
             // BashRunner does nothing special when a failure is expected, so just confirm source file is still present
             this.MoveFile(sourcePath, targetPath);
-            this.FileExists(sourcePath).ShouldEqual(true);
+            this.FileExists(sourcePath).ShouldBeTrue();
         }
 
         public override void MoveFile_FileShouldNotBeFound(string sourcePath, string targetPath)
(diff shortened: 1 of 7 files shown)


GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
```diff
@@ -28,11 +29,18 @@ public class BashRunner : ShellRunner
...
+        private static string[] permissionDeniedWindowsMessage = new string[]
         {
             "Permission denied"
         };
 
+        // TODO(Mac): Update this message when the kext returns a more accurate
+        // error code
+        private static string[] permissionDeniedMacMessage = new string[]
+        {
+            "Resource temporarily unavailable"
```
sanoursa: We haven't really settled on what error codes we're going to return, so this is subject to change. That said, does the specific error message matter? Shouldn't we just check that we tried to delete and the delete failed?

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
```diff
@@ -246,7 +254,14 @@ public override void DeleteFile_FileShouldNotBeFound(string path)
 
         public override void DeleteFile_AccessShouldBeDenied(string path)
         {
-            this.DeleteFile(path).ShouldContain(permissionDeniedMessage);
```
sanoursa: I think I would rather change this to `this.DeleteFile(path); path.ShouldBeAFile();`. Regardless of what the error is, what we care about is that the file is still present. And that check is also platform-agnostic.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSLockTests.cs
```diff
@@ -102,15 +101,28 @@ private void OverwritingIndexShouldFail(string testFilePath)
...
             byte[] newIndexContents = File.ReadAllBytes(indexPath);
 
             indexContents.SequenceEqual(newIndexContents).ShouldBeTrue("Index contenst should not have changed");
 
             this.fileSystem.DeleteFile(testFilePath);
         }
+
+        private bool RenameAndOverwrite(string oldPath, string newPath)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
sanoursa: Every time I see a new platform check, I wonder if we're creating a maintenance headache for ourselves down the road. What are your thoughts on this? Is it a big enough issue to worry about yet? Is there a cleaner pattern that we could use to at least encapsulate the platform checks into one place? (I'm happy to keep deferring this if that's what makes sense, just wanted to bring it up for discussion)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSLockTests.cs
```diff
@@ -91,6 +87,9 @@ public void LockPreventsRenameOfIndexLockOnTopOfIndex()
             string newFileName,
             uint flags);
 
+        [DllImport("libc", EntryPoint = "rename", SetLastError = true)]
+        private static extern int Rename(string oldPath, string newPath);
```
sanoursa: Consider renaming this method to `MacRename`, and similarly `MoveFileEx`->`WindowsMoveFileEx`, or some other similar pattern to make the platform-specific usage more clear. Or see my other comment, at some point we may want to think about a refactor to encapsulate platform-specific calls like we did for the product code.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -32,6 +34,14 @@ public MacFileSystemVirtualizer(GVFSContext context, GVFSGitObjects gitObjects)
             this.virtualizationInstance = virtualizationInstance ?? new VirtualizationInstance();
         }
 
+        protected override string EtwArea
```
sanoursa: Nit pick: this can be rewritten as `protected override string EtwArea => ClassName;` and avoid taking up so much screen space for boilerplate.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -330,6 +340,21 @@ private Result OnPreDelete(string relativePath, bool isDirectory)
...
+                    {
+                        string lockedGitCommand = this.Context.Repository.GVFSLock.GetLockedGitCommand();
+                        if (string.IsNullOrEmpty(lockedGitCommand))
+                        {
+                            EventMetadata metadata = new EventMetadata();
+                            metadata.Add("Area", this.EtwArea);
+                            metadata.Add(TracingConstants.MessageKey.WarningMessage, "Blocked index delete outside the lock");
+                            this.Context.Tracer.RelatedEvent(EventLevel.Warning, $"{nameof(OnPreDelete)}_BlockedIndexDelete", metadata);
+
+                            // TODO(Mac): Is this the correct Result to return?
```
sanoursa: Looks reasonable to me

PR #275, opened 2018-09-18
Upgrade verb for manually installing new product release

alameenshah:
1. GVFS.Service checks for upgrade availability periodically. When a new release is available it will be downloaded in the background.
2. GVFS Git Hooks will remind user to install the upgrade.
3. GVFS Upgrade [--confirm] verb will enable user to manually trigger installation of the upgrade release.
4. Use gvfs config command to specify an upgrade ring. 
     Example: `gvfs config upgrade.ring "Fast"`. 
5. New GVFS name format for GVFS installer VFSGit.<version>.<platform_specific_installer_extension>
4. Unit test cases.
5. Functional test for upgrade reminder messaging.

#### Testing note
For testing I have created multiple releases of this branch in my clone - https://github.com/alameenshah/VFSForGit/releases

[[Old PR that was accidentally closed](https://github.com/Microsoft/VFSForGit/pull/254)]


GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+            {
+                this.output.WriteLine(GVFSConstants.UpgradeVerbMessages.NoneRingConsoleAlert);
+                this.output.WriteLine(GVFSConstants.UpgradeVerbMessages.SetUpgradeRingCommand);
+            }
+            else
+            {
+                try
+                {
+                    Version newVersion = null;
+                    if (!this.TryRunUpgradeInstall(out newVersion, out error))
```
sanoursa: "UpgradeInstall" is redundant

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+                this.output.WriteLine(error);
+            }
+
+            if (this.input == Console.In)
+            {
+                this.output.WriteLine("Press Enter to exit.");
+                this.input.ReadLine();
+            }
+            
+            if (this.shouldExitOnError)
```
sanoursa: Can you drop this bool if you just set `Environment.ExitCode` instead of calling `Environment.Exit`?

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+                    }
+
+                    if (!this.TryDownloadUpgrade(newGVFSVersion, out errorMessage))
+                    {
+                        return false;
+                    }
+
+                    return true;
+                },
+                "Downloading"))
```
sanoursa: This is going to say "downloading" even if we never got to downloading because of the pre-req checks.

Does the call to `TryRunPreUpgradeChecks` need to happen so late in the process? Can't it be the first thing you check, before even asking GitHub for new releases? If we can move `TryRunPreUpgradeChecks` to be the first check, and do it before this spinner, that will result in a cleaner experience.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+                () =>
+                {
+                    if (!this.TryInstallGVFSUpgrade(newGVFSVersion, out errorMessage))
+                    {
+                        return false;
+                    }
+
+                    return true;
+                },
+                $"Installing VFSForGit version: {newGVFSVersion}"))
```
sanoursa: "VFSForGit" -> "VFS for Git"

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+            }
+
+            this.LogVersionInfo(newGVFSVersion, newGitVersion, "Newly Installed Version");
+
+            newVersion = newGVFSVersion;
+            consoleError = null;
+            return true;
+        }
+        
+        private bool TryRemountRepositories(out string consoleError)
```
sanoursa: Rename to `TryMountRepositories`

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+            string errorMessage = string.Empty;
+            if (this.remount && !this.LaunchInsideSpinner(
+                () =>
+                {
+                    string remountError;
+                    if (!this.preRunChecker.TryMountAllGVFSRepos(out remountError))
+                    {
+                        EventMetadata metadata = new EventMetadata();
+                        metadata.Add("Upgrade Step", nameof(this.TryRemountRepositories));
+                        metadata.Add("Remount Error", remountError);
```
sanoursa: Change to "mount" everywhere

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -113,6 +113,14 @@ protected override void Execute(GVFSEnlistment enlistment)
                             this.ServiceName,
                             copySubFolders: true);
 
+                        // upgrader
+                        this.CopyAllFiles(
```
sanoursa: Do we need to copy all of these files? How big can they get? Would a directory listing be good enough, and avoid adding unnecessary size and bulk to diagnose?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,325 @@
...
+                {
+                    if (this.TryCopyUpgradeTool(out upgraderPath, out errorMessage) &&
+                        this.TryLaunchUpgradeTool(upgraderPath, out errorMessage))
+                    {
+                        return true;
+                    }
+
+                    return false;
+                },
+                "Launching upgrade tool",
```
sanoursa: Since we're releasing the console back to the user after this point, it may be worth including a message like 
```
Installer launched in a new window. Do not run any git or gvfs commands until the installer has completed.
```

cc @jrbriggs  thoughts?

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+                    {
+                        this.ExitCode = ReturnCode.GenericError;
+                    }
+                }
+                finally
+                {
+                    string remountError = null;
+                    if (!this.TryRemountRepositories(out remountError))
+                    {
+                        remountError = Environment.NewLine + "WARNING: " + remountError;
```
sanoursa: I forced a downgrade failure here, and the result is very confusing:
```
Downloading...Succeeded
Unmounting repositories...Succeeded
Installing Git version: 2.17.1.gvfs.2.5...Succeeded
Installing VFSForGit version: 1.0.18234.1...Succeeded
Mounting repositories...Failed

WARNING: GVFS error.
Mounting repo at c:\GVFSRepos\GVFS
Error: Changes to GVFS disk layout do not allow mounting after downgrade. Try mounting again using a more recent version of GVFS.
Enlistment disk layout version: 17
GVFS disk layout version: 16

The following repos failed to mount:
c:\GVFSRepos\GVFS

Press Enter to exit.
```

Problems:
1) "WARNING: GVFS Error". Is it a warning or an error?
2) On the default window size that the upgrade tool opened up, the text is wrapping, making it even harder to read:
image
3) I see duplicate messages about the path `C:\GVFSRepos\GVFS` and it's hard to know how to make sense of it
4) There is nothing here to indicate to the user that the installation actually succeeded, even though one of the repos failed to mount. Their natural response will be to just run the installer again, which is not going to fix anything.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+        public bool TryMountAllGVFSRepos(out string consoleError)
+        {
+            return this.TryRunGVFSWithArgs("service --mount-all", out consoleError);
+        }
+
+        public bool TryUnmountAllGVFSRepos(out string consoleError)
+        {
+            consoleError = null;
+
+            this.tracer.RelatedInfo("Unmounting any mounted GVFS repositories.");
```
sanoursa: Instead of this pattern of tracing "Starting foo" and and the end "Finished foo", call `ITracer.StartActivity` and put it in a using statement. It'll automatically log the start and stop messages, but also record the duration of the activity. Please apply it to any place you have this pattern.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,325 @@
...
+
+                        if (!this.TryCheckUpgradeInstallable(out errorMessage))
+                        {
+                            upgradeIsInstallable = false;
+                        }
+                    }
+
+                    return upgradeCheckSuccess;
+                },
+                 "Checking for GVFS upgrades",
```
sanoursa: Elsewhere you've already started calling it "VFS for Git". Please make it consistent.

GVFS/GVFS.Hooks/Program.cs
```diff
@@ -83,6 +83,12 @@ public static void Main(string[] args)
 
         private static void RunPreCommands(string[] args)
         {
+            if (ProductUpgrader.IsLocalUpgradeAvailable())
```
sanoursa: What is the cost of this check? How much time are we adding to every git command now?

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -33,6 +33,7 @@ public static class GitConfig
             public const string HooksPrefix = GitConfig.GVFSPrefix + "clone.default-";
             public const string GVFSTelemetryId = GitConfig.GVFSPrefix + "telemetry-id";
             public const string HooksExtension = ".hooks";
+            public const string UpgradeRing = GVFSPrefix + "upgradering";
```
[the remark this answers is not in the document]
jrbriggs: Let's stop the inconsistency with Git rather than introduce more. But I also don't want to be inconsistent with what we have. What about just `ring`?

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -215,5 +219,13 @@ public static class Unmount
                 public const string SkipLock = "skip-wait-for-lock";
             }
         }
+
+        public static class UpgradeVerbMessages
+        {
+            public const string NoneRingConsoleAlert = "Upgrade ring set to None. No upgrade check was performed.";
+            public const string SetUpgradeRingCommand = "To set or change upgrade ring, run `git config --global gvfs.upgradering [\"Fast\"|\"Slow\"|\"None\"] from an elevated command prompt.";
```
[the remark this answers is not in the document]
jrbriggs: It should not be necessary, with the change from system config to global config. We should remove that portion of the message.

GVFS/GVFS.Common/GVFSConstants.cs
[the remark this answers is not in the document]
sanoursa: Honestly I'm much more interested in having a name that's both descriptive and easy to read. Is the lack or presence of `-` a major issue? I vote `upgrade-ring` but also you can override that to whatever you think makes sense :-)

GVFS/GVFS.Hooks/Program.cs
```diff
@@ -83,6 +83,12 @@ public static void Main(string[] args)
 
         private static void RunPreCommands(string[] args)
         {
+            if (ProductUpgrader.IsLocalUpgradeAvailable())
+            {
+                Console.WriteLine(GVFSConstants.UpgradeVerbMessages.UpgradeAvailable);
```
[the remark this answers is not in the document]
jrbriggs: Yes -- this is the area I'm most concerned about.

We need to check where we can put this output that will be minimally disruptive. I agree it should be one line at the end separated by a blank line. We also need to test common tools like VS and SourceTree to ensure we're not breaking parsers. We also might want to simplify the message to "A new version of GVFS is available. Run `gvfs upgrade` to install" and let that guide them into starting the flow and the details of needing to elevate.

It should also not fire every time. It probably should be no more than x times per hour or some similar metric so as not to unfairly punish those who use the command line more frequently.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,389 @@
...
+            
+            if (this.ExitCode == ReturnCode.GenericError)
+            {
+                error = Environment.NewLine + "ERROR: " + error;
+                this.output.WriteLine(error);
+            }
+
+            if (this.input == Console.In)
+            {
+                this.output.WriteLine("Press Enter to exit.");
```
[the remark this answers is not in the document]
jrbriggs: I like being more affirmative when it succeeeds.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: This is a warning, agree with comments 2, 3 & 4. These errors are as shown in the existing manual installer (attached screenshot).

How about we display the mount error in a single line and log the detailed failure reason in the upgrade log file. Like this -

```
Downloading...Succeeded
Unmounting repositories...Succeeded
Installing Git version: 2.17.1.gvfs.2.5...Succeeded
Installing VFSForGit version: 1.0.18234.1...Succeeded
Mounting repositories...Failed

WARNING: Mounting of one or more repos failed. Refer <upgrade.log path> for more info.

Upgrade completed successfully!
Press Enter to exit.```

<img width="416" alt="screen shot 2018-09-18 at 5 18 08 pm" src="https://user-images.githubusercontent.com/378580/45717363-c1aa7700-bb67-11e8-9661-3b31d5ebe489.png">

@sanoursa @jrbriggs - Please share your thoughts.

File: GVFS/GVFS/CommandLine/UpgradeVerb.cs [hunk 17]
```diff
@@ -0,0 +1,325 @@
...
+            {
+                this.ReportInfoToConsole($"Great news, you're all caught up on upgrades in the {this.upgrader.Ring} ring!");
+                return true;
+            }
+
+            this.ReportInfoToConsole("New GVFS version available: {0}", newestVersion.ToString());
+                        
+            if (!this.Confirmed && isInstallable)
+            {
+                this.ReportInfoToConsole("Run `gvfs upgrade --confirm` to install it");
```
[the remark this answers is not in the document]
alameenshah: We are also reminding the user about the need to upgrade in GVFS Git hooks.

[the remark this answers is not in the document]
sanoursa: I'm not sure I follow. The problem here isn't about reminding them. The user already ran the upgrade command. The problem is that it looks like the command completed, even though the user still needs to take another action. We need to make that action more obvious.

GVFS/GVFS.Hooks/Program.cs
alameenshah: Tested on GVFS repo using “Measure-Command { foreach ($i in 0..200) {git status}}”

Trials | No reminder hook | Reminder hook enabled & upgrade NOT available | Reminder Hook enabled & upgrade available
------ | ------------ | ------------- | -----------
1 | 52647.2346 | 54122.9319 | 67319.0552
2 | 66861.3138 | 64679.6913 | 63246.1476
3 | 66897.6017 | 68131.4394 | 68101.7228
4 | 62314.6312 | 67122.4883 | 68498.0093
5 | 68277.4996 | 67653.2103 | 68913.034
Average | 63399.65618 | 64341.95224 | 67215.59378
**Per call** | **317ms** | **321ms** | **336ms**

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: There are too many occurrences of GVFS, so renamed "VFS for Git" to "GVFS". Will replace "GVFS" references with "VFS for Git" in a future PR.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: - Modified GVFS Service verb to log failed repos list to stderr. Upgrade tool will display stderr from GVFS CLI. 
- This new Service verb option is enabled using the command line switch "log-mount-failure-in-stderr"

screen shot 2018-09-20 at 4 54 52 pm

alameenshah: ### Testing suggestion 
The new command line option obviously is not available in the current release version of GVFS. After downloading and installing upgrade release, when Upgrader tool invokes newly installed GVFS CLI, the CLI wouldn't know the new command line option and reply with error (Screenshot attached.) One workaround is to use a custom GVFS build that has  TryInstallGVFSUpgrade() invocation in UpgradeOrchestrator::TryRunUpgradeInstall() method commented out.

screen shot 2018-09-20 at 4 49 50 pm

GVFS/GVFS/CommandLine/UpgradeVerb.cs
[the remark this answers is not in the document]
alameenshah: Updated.

```
PS M:\Work\gvfs\VFSForGit_Ameen> gvfs upgrade
Upgrade ring is not set. No upgrade check was performed.

ERROR: Unable to determine upgrade ring.
Run `git config --global gvfs.upgradering ["Fast"|"Slow"|"None"]` and run `gvfs upgrade [--confirm]` again.
```

GVFS/GVFS/CommandLine/UpgradeVerb.cs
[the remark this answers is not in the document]
alameenshah: Inserted a blank line before final instruction

```
PS M:\Work\TestRepos\GVFS_1\src> gvfs upgrade
New GVFS version available: 1.0.18234.1

Run `gvfs upgrade --confirm` to install it
```

GVFS/GVFS.Hooks/Program.cs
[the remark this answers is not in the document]
alameenshah: - Updated to display reminder notification 10% of the time.
- Moved the reminder from pre-command hook to post-command hook.
- Added a blank line before inserting the reminder.
- Single line message.
- Removed the text to run `gvfs upgrade`

New output.
```
PS M:\Work\TestRepos\GVFS_1\src> git status
On branch master
Your branch is up-to-date with 'origin/master'.

A new version of GVFS is available.
```

GVFS/GVFS/CommandLine/ServiceVerb.cs
```diff
@@ -35,6 +35,13 @@ public class ServiceVerb : GVFSVerb
             HelpText = "Prints a list of all mounted repos")]
         public bool List { get; set; }
 
+        [Option(
+            "log-mount-failure-in-stderr",
+            Default = false,
+            Required = false,
+            HelpText = "This parameter is reserved for internal use.")]
+        public bool RedirectMountFailuresToStderr { get; set; }
```
sanoursa: I don't think we need a new flag for this

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -92,6 +92,32 @@ public static Result GetFromSystemConfig(string gitBinPath, string settingName)
...
+                if (!GitVersion.TryParseGitVersionCommandResult(version, out gitVersion))
+                {
+                    error = "Unable to determine installed git version. " + version;
+                    return false;
+                }
+
+                error = null;
+                return true;
+            }
+            catch (NotSupportedException exception)
```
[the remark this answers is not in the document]
alameenshah: This is thrown only during unit testing by MockPlatform. MockPlatform does not implement IGitInstallation yet. Will make a note and take this up in a future PR.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        private ITracer tracer;
+                
+        public InstallerPreRunChecker(ITracer tracer)
+        {
+            this.tracer = tracer;
+            this.CommandToRerun = string.Empty;
+        }
+
+        public string CommandToRerun { get; set; }
```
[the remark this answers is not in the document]
alameenshah: MockInstallerPreRunChecker needs access to this. Have made it a protected property with a private get method.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+            this.tracer.RelatedInfo("Checking if GVFS upgrade can be run on this machine.");
+
+            if (this.IsUnattended())
+            {
+                consoleError = "`gvfs upgrade` is not supported in unattended mode";
+                this.tracer.RelatedError($"{nameof(TryRunPreUpgradeChecks)}: {consoleError}");
+                return false;
+            }
+
+            if (this.IsDevelopmentVersion())
```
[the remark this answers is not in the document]
alameenshah: Removed it :)

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+            {
+                if (process.Id == currentProcessId || !BlockingProcessSet.Contains(process.ProcessName))
+                {
+                    continue;
+                }
+
+                matchingNames.Add(process.ProcessName);
+            }
+
+            processes = matchingNames.ToList();
```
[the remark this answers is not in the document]
alameenshah: Yes, updated to return HashSet.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+
+namespace GVFS.Upgrader
+{
+    public class InstallerPreRunChecker
```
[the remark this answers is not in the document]
alameenshah: Agree. This class is doing more than pre-run operations. Have a follow-up task created for moving mount/un-mount methods out - https://github.com/Microsoft/VFSForGit/issues/293

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+            processes = matchingNames.ToList();
+            return processes.Count > 0;
+        }
+
+        protected virtual bool TryRunGVFSWithArgs(string args, out string consoleError)
+        {
+            consoleError = null;
+
+            string gvfsPath = Path.Combine(
+                ProcessHelper.WhereDirectory(GVFSPlatform.Instance.Constants.GVFSExecutableName),
```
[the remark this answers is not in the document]
alameenshah: right. Added the null check.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+            ProcessResult processResult = ProcessHelper.Run(gvfsPath, args);
+            if (processResult.ExitCode == 0)
+            {
+                return true;
+            }
+            else
+            {
+                string output = string.IsNullOrEmpty(processResult.Output) ? string.Empty : processResult.Output;
+                string errorString = string.IsNullOrEmpty(processResult.Errors) ? "GVFS error" : processResult.Errors;
+                consoleError = string.Format("{0}. {1}", errorString, output);
```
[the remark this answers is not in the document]
alameenshah: This has been updated to show only stderr. Attaching latest screenshot.

screen shot 2018-09-20 at 4 54 52 pm

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            return Paths.GetServiceDataRoot(RootDirectory);
+        }
+        
+        public static string GetLogDirectoryPath()
+        {
+            return Path.Combine(Paths.GetServiceDataRoot(RootDirectory), LogDirectory);
+        }
+
+        private static bool TryCreateDirectory(string path, out Exception exception)
```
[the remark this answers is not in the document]
alameenshah: Moved

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -0,0 +1,477 @@
...
+    public partial class ProductUpgrader
+    {
+        private const string GitHubReleaseURL = @"https://api.github.com/repos/microsoft/vfsforgit/releases";
+        private const string JSONMediaType = @"application/vnd.github.v3+json";
+        private const string UserAgent = @"GVFS_Auto_Upgrader";
+        private const string CommonInstallerArgs = "/VERYSILENT /CLOSEAPPLICATIONS /SUPPRESSMSGBOXES /NORESTART";
+        private const string GVFSInstallerArgs = CommonInstallerArgs + " /MOUNTREPOS=false";
+        private const string GitInstallerArgs = CommonInstallerArgs + " /ALLOWDOWNGRADE=1";
+        private const string GitAssetNamePrefix = "Git";
+        private const string GVFSAssetNamePrefix = "GVFS";
```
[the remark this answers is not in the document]
alameenshah: Yes, changing name will break this logic. 
Future proofing - [GitHub Release API](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) only provides "Content-Type", "name" and "label" for us to specify an Asset. A numeric identifier or ability to specify a unique asset identifier would have helped. I can post a feature request in the GitHub page.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public ProductUpgradeTimer(JsonTracer tracer)
+        {
+            this.tracer = tracer;
+        }
+
+        public void Start()
+        {
+            Random random = new Random();
+            TimeSpan startTime = TimeSpan.FromMinutes(random.Next(0, 60));
```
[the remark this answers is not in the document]
alameenshah: Removed the randomization.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockInstallerPreRunChecker.cs
```diff
@@ -0,0 +1,129 @@
...
+            BlockingProcessesRunning = 0x4,
+            UnattendedMode = 0x8,
+            IsDevelopmentVersion = 0x10,
+            IsGitUpgradeAllowed = 0x20,
+            UnMountRepos = 0x40,
+            RemountRepos = 0x80,
+            IsServiceInstalledAndNotRunning = 0x100
+        }
+        
+        public List<string> GVFSArgs { get; private set; } = new List<string>();
```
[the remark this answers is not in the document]
alameenshah: removed it.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: Yes, that is possible. Removed the bool.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: Renamed it back to "GVFS".

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
alameenshah: Max size I've seen of a gvfs_*.log file is 4KB. Git and GVFS installer logs reach 2MB and 30KB respectively. Also it gets generated only when a user runs upgrade.

GVFS/GVFS/CommandLine/ServiceVerb.cs
alameenshah: Removed

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockTextWriter.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace GVFS.UnitTests.Windows.Mock.Upgrader
+{
+    public class MockTextWriter : TextWriter
```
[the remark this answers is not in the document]
alameenshah: Upgrade verb output is not being verified in Functional tests.

GVFS/GVFS.Common/Git/GitProcess.cs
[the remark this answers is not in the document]
sanoursa: @alameenshah I don't think we can leave this as a TODO. This can hide serious bugs in our production code.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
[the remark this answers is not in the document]
sanoursa: The warning output looks good, but I have a couple of questions:

* What happened to the line about installing GVFS?
* Should we update the final line to something like "Even though some repos failed to mount, the upgrade completed successfully"? But more importantly, is there anything a user can do to recover here or is this just going to generate a support email?

GVFS/GVFS.Common/ProductUpgrader.cs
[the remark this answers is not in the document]
sanoursa: We have the ability to future-proof this ourselves, without requiring a change from GitHub. E.g. we could rename our installer now, if we want to.

If we don't, then whenever we change the installer name in the future, we will either need to publish duplicate installers under both names, or else users on the pre-rename installed version will not be able to upgrade to the post-rename release.

cc @jrbriggs what are your thoughts?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -92,6 +92,32 @@ public static Result GetFromSystemConfig(string gitBinPath, string settingName)
             return new GitProcess(gitBinPath, workingDirectoryRoot: null, gvfsHooksRoot: null).InvokeGitOutsideEnlistment("config --system " + settingName);
         }
 
+        public static bool TryGetVersion(out GitVersion gitVersion, out string error)
```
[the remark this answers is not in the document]
sanoursa: Yes I understand that it's being used, but it seems like a very simple cleanup to fix it now, and more work and technical debt to leave it for later.

I'm not going to block this PR on this item, but in general we should clean as we go and not just push things off into issues that may never get fixed in the future.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
sanoursa: @alameenshah fine to do this in a future PR, but I think a directory listing would give us all the info we need, and we already have code to do that. There's no need to copy the installers themselves into the zip file.

alameenshah: >>There's no need to copy the installers themselves into the zip file.

Oh. We Don't copy the installers at all. Only the log files.

sanoursa: Ah yea I missed that it was just the logs, I thought it was the installers themselves.

In that case, I think we _also_ need a directory listing of the downloaded installers, since that is what drives the hooks to show an upgrade reminder to the user. When someone complains about that, we need to have a way to diagnose what's happening. So please add that as well.

GVFS/GVFS.Common/Git/GitProcess.cs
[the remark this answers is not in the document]
alameenshah: Made the change.

Version() returns ```2.18.0.gvfs.1.26.gf61ade4```
TryGetVersion() returns ```2.18.0.gvfs.1.26```
Unit tests pass. Will keep an eye on Functional tests for any side effects.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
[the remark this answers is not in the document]
alameenshah: >> What happened to the line about installing GVFS?

It will be there in the actual release. This output is from my custom build. The latest release on GitHub/Release does not log mount failure list to stderr - you know. I commented out the code that installs GVFS upgrade, built and installed it with all the new mount-stderr related changes. Then ran gvfs upgrade. Now upgrade will download both Git and GVFS installers from Github/release, but will only install Git. Since the PC has the latest GVFS with mount-stderr changes, it shows the new stderr logging on the Console. 

>>Should we update the final line to something like "Even though some repos failed to mount, the upgrade completed successfully"?

I think that would break the existing pattern.
```
Progress spinners

ERROR/WARNING: Failure reason
Possible remedy in case of failure

[Upgrade completed successfully!] /*in case of success:*/
Press Enter to exit.
```

>>But more importantly, is there anything a user can do to recover here or is this just going to generate a support email?

It depends on the specific reason for the mount failure, right? We can't give a generic recovery step to cover all failure reasons. User can view the mount logs and take any recovery steps suggested there or email support. I'm not sure if there can be an alternate option.

GVFS/GVFS.Common/Git/GitProcess.cs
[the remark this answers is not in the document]
alameenshah: Added new MockGitInstallation class for use in UT. Removed the NotSupportedException handler.

GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -335,5 +347,20 @@ private void LogExceptionAndExit(Exception e, string method)
...
+        private void EnableAccessToAuthenticatedUsers(string rootDirectory)
+        {
+            // GVFS Config is written to a temporary file and then renamed to its final destination.
+            // For this rename operation to succeed, user needs to have delete permission on the 
+            // destination file, in case it is pre-existing. If the pre-existing file was created 
+            // by a different user, then the delete will fail.
+            // Reference: https://stackoverflow.com/questions/22107812/privileges-owner-issue-when-writing-in-c-programdata
+            // This work around allows safe write to succeed in C:\ProgramData directory.
+
+            DirectorySecurity security = Directory.GetAccessControl(Path.GetDirectoryName(this.serviceDataLocation));
```
sanoursa: What path does this resolve to?

alameenshah: "C:\ProgramData\GVFS"

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -0,0 +1,425 @@
...
+            }
+
+            if (this.ExitCode == ReturnCode.GenericError)
+            {
+                error = Environment.NewLine + "ERROR: " + error;
+                this.output.WriteLine(error);
+            }
+            else
+            {
+                this.output.WriteLine(Environment.NewLine + "Upgrade completed successfully!");
```
sanoursa: I still find this output confusing, and I think users will be similarly wondering if this is an error or a success:

```
Mounting repositories...Failed

WARNING: GVFS service --mount-all failed.

Upgrade completed successfully!
Press Enter to exit.
```

I think that final success message needs to acknowledge that mount failed, but also that upgrading the product succeeded. Otherwise people will just think this is a bug and will be emailing us about it.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -0,0 +1,489 @@
...
+                    Enum.IsDefined(typeof(RingType), ringType) &&
+                    ringType != RingType.Invalid)
+                {
+                    this.Ring = ringType;
+                    error = null;
+                    return true;
+                }
+                else
+                {
+                    error = "Invalid upgrade ring `" + ringConfig + "` specified in gvfs config.";
```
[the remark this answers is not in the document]
alameenshah: This is expected behavior today - When no ring is specified, or the ring that is specified is mis-spelled, it is treated as an error. We show the error and advice user on how to set the right ring config. This behavior is different from "None" ring, which is an acceptable ring setting.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,379 @@
...
+                    string message = string.Join(
+                        Environment.NewLine + Environment.NewLine, 
+                        upgradeAvailableMessage,
+                        GVFSConstants.UpgradeVerbMessages.UnmountRepoWarning,
+                        GVFSConstants.UpgradeVerbMessages.UpgradeInstallAdvice);
+                    this.ReportInfoToConsole(message);
+                }
+                else
+                {
+                    this.ReportInfoToConsole($"{Environment.NewLine}{upgradeAvailableMessage}");
```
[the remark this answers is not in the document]
alameenshah: Added.
Example - 
screen shot 2018-10-03 at 6 06 11 pm

GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -0,0 +1,62 @@
+﻿using CommandLine;
+using GVFS.Common;
+using System;
+
+namespace GVFS.CommandLine
+{
+    [Verb(ConfigVerbName, HelpText = "Get and set GVFS options.")]
+    public class ConfigVerb : GVFSVerb.ForNoEnlistment
```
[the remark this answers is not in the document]
alameenshah: Will make a note to support all CRUD and List operations.

GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -0,0 +1,62 @@
...
+                    Console.WriteLine(value);
+                }
+                else
+                {
+                    this.ReportErrorAndExit(error);
+                }
+            }
+            else
+            {
+                if (!this.localConfig.TrySetConfig(this.Key, this.Value, out error, tracer: null))
```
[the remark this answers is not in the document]
alameenshah: This is similar in behavior to ```git config```.

screen shot 2018-10-03 at 6 10 50 pm

UpgradeVerb owns the upgrade.ring config and it is up to UpgradeVerb to decide what is the correct value for the settings that it owns. This separation helps keep ConfigVerb focused on its value addition - it manages CRUD operations of Key Value pairs. Leaving the specialty knowledge of acceptable config settings to the consumers of ConfigVerb.

But am open to suggestions - please let me know if you think it is better for ConfigVerb to validate the settings.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,379 @@
...
+                return false;
+            }
+
+            if (newestVersion == null)
+            {
+                this.ReportInfoToConsole($"Great news, you're all caught up on upgrades in the {this.upgrader.Ring} ring!");
+                return true;
+            }
+            
+            string upgradeAvailableMessage = $"New GVFS version available: {newestVersion.ToString()}";
```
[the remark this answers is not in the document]
alameenshah: Added ring info.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -0,0 +1,489 @@
...
+                if (installerExitCode != 0 && string.IsNullOrEmpty(error))
+                {
+                    error = name + " installer failed. Error log: " + logFilePath;
+                }
+
+                installerIsRun = true;
+            }
+            else
+            {
+                error = "Could not find downloaded installer for " + name;
```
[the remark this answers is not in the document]
alameenshah: >>It should have known that a file was missing after the download step. It should not have upgraded Git when it was guaranteed that upgrading GVFS would fail.

Agree. Added early exit, when either Git or GVFS asset installers are missing.

>>It failed to upgrade GVFS, and yet it still remounted the repos. We've now put the user into a state where the Git and GVFS versions do not match.

Updated to not auto re-mount when GVFS installation failed.

>> There was a long delay between "Mounting repositories...Succeeded" and "Error: Could not find". At least 10 seconds, if not longer, with nothing moving on the screen. Any idea what happened?

I couldn't repro this, but there is a step to delete the downloaded assets after Mounting Finished. Am wondering if something (like AV - am guessing) slowed the file deletion on your PC.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: screen shot 2018-10-03 at 3 42 06 pm

GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -0,0 +1,62 @@
...
+    public class ConfigVerb : GVFSVerb.ForNoEnlistment
+    {
+        private const string ConfigVerbName = "config";
+        private LocalGVFSConfig localConfig;
+
+        [Value(
+                0,
+                Required = true,
+                Default = "",
+                MetaName = "GVFS config setting name",
```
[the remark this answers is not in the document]
alameenshah: Drop or replace it with VFSForGit?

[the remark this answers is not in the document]
sanoursa: I would drop it. It's very redundant in context.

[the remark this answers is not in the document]
jrbriggs: +1. I think these should be "Setting name" and "Setting value" since we're already in the context of "gvfs config"

jrbriggs:
When I run from non-elevated I get:
```

New GVFS version 1.0.18234.1 available in ring Fast

The installer needs to be run from an elevated command prompt.
To install, run `gvfs upgrade --confirm` from an elevated command prompt.
```

When I run from elevated I get:
```
New GVFS version 1.0.18234.1 available in ring Fast

When ready, run `gvfs upgrade --confirm` from an elevated command prompt.
```

My concern here is a user that goes to a non-elevated prompt and types `gvfs upgrade` will be instructed to run `gvfs upgrade --confirm` from an elevated prompt and miss the details about needing to be at a stopping point. I think should show that at both levels. Thoughts @alameenshah @sanoursa ?

GVFS/GVFS/CommandLine/ConfigVerb.cs
sanoursa: Turns out that this is case sensitive, and I think that'll be surprising for users. Check out the contents of my config file:

```
C:\Repos\VFSForGit\src>type "C:\ProgramData\GVFS\gvfs.config"
A {"Key":"upgrade.ring","Value":"slow"}
A {"Key":"Upgrade.Ring","Value":"fast"}
```

GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            this.localConfig = new LocalGVFSConfig();
+
+            string error = null;
+            bool isRead = string.IsNullOrEmpty(this.Value);
+
+            if (isRead)
+            {
+                string value = null;
+                if (this.localConfig.TryGetConfig(this.Key, out value, out error, tracer: null))
```
sanoursa: Same here. Even though I have two competing variants of the setting already configured, this returned nothing:

```
C:\Repos\VFSForGit\src>gvfs config upgrade.Ring

```

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,384 @@
...
+                this.ReportInfoToConsole($"Cannot upgrade GVFS on this machine.");
+                this.Output.WriteLine(errorOutputFormat, cannotInstallReason);
+                this.tracer.RelatedError($"{nameof(this.TryRunProductUpgrade)}: Upgrade is not installable. {cannotInstallReason}");
+                return false;
+            }
+
+            if (!this.TryLoadUpgradeRing(out ring, out error))
+            {
+                this.tracer.RelatedError($"{nameof(this.TryRunProductUpgrade)}: Could not load upgrade ring. {error}");
+                this.ReportInfoToConsole(GVFSConstants.UpgradeVerbMessages.InvalidRingConsoleAlert);
```
sanoursa: The opposite of what I had reported before. I now configured an invalid ring, and this is telling me no ring is set.

```
C:\Repos\VFSForGit\src>gvfs config upgrade.ring Foo

C:\Repos\VFSForGit\src>gvfs upgrade
Upgrade ring is not set. No upgrade check was performed.

ERROR: Invalid upgrade ring `Foo` specified in gvfs config.
To set or change upgrade ring, run `gvfs config upgrade.ring ["Fast"|"Slow"|"None"]` from a command prompt.
```

I think we need to have different behavior for no config vs invalid config. No config behaves like None, it's not an error, though we should tell them how to configure it. An invalid ring is an error, and it shouldn't say that it's not set.

GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -0,0 +1,62 @@
...
+    [Verb(ConfigVerbName, HelpText = "Get and set GVFS options.")]
+    public class ConfigVerb : GVFSVerb.ForNoEnlistment
+    {
+        private const string ConfigVerbName = "config";
+        private LocalGVFSConfig localConfig;
+
+        [Value(
+                0,
+                Required = true,
+                Default = "",
```
sanoursa: Drop these `Default = ""` lines. It's causing the help output to display a confusing `(Default: )` 

```
  GVFS config setting name (pos. 0)     Required. (Default: ) Name of GVFS config setting that is to be set or read

  GVFS config setting value (pos. 1)    (Default: ) Value of GVFS config setting to be set
```

start
```diff
@@ -0,0 +1 @@
+prjflt
```
sanoursa: What is this file for?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,384 @@
...
+
+                checkSucceeded = this.upgrader.TryGetNewerVersion(out version, out consoleError);
+                if (!checkSucceeded)
+                {
+                    return false;
+                }
+
+                latestVersion = version;
+
+                activity.RelatedInfo("Successfully checked server for GVFS upgrades.");
```
sanoursa: This log statement should include the version that was discovered.

In general, I'm not sure that the logs we're generating right now are all that useful for debugging user errors. Have you thought through what information needs to be included in them so we can diagnose user issues that get reported later?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,384 @@
...
+
+        private bool TryInitializeUpgrader()
+        {
+            OperatingSystem os_info = Environment.OSVersion;
+
+            if (os_info.Platform == PlatformID.Win32NT)
+            {
+                if (this.upgrader == null)
+                {
+                    JsonTracer jsonTracer = new JsonTracer(GVFSConstants.GVFSEtwProviderName, "UpgradeVerb");
```
sanoursa: Here's an example of a log file from a successful upgrade run:
```
[10/4/2018 8:31:05 AM] TryCheckUpgradeInstallable (Start)
[10/4/2018 8:31:05 AM] TryRunPreUpgradeChecks (Start)
[10/4/2018 8:31:06 AM] Information {"Message":"Successfully finished pre upgrade checks. Okay to run `gvfs upgrade`."}
[10/4/2018 8:31:06 AM] TryRunPreUpgradeChecks (Stop) {"DurationMs":132}
[10/4/2018 8:31:06 AM] Information {"Message":"Upgrade is installable."}
[10/4/2018 8:31:06 AM] TryCheckUpgradeInstallable (Stop) {"DurationMs":138}
[10/4/2018 8:31:06 AM] TryCheckUpgradeAvailable (Start)
[10/4/2018 8:31:06 AM] Information {"Message":"Successfully checked server for GVFS upgrades."}
[10/4/2018 8:31:06 AM] TryCheckUpgradeAvailable (Stop) {"DurationMs":269}
[10/4/2018 8:31:06 AM] TryCopyUpgradeTool (Start)
[10/4/2018 8:31:06 AM] Information {"Message":"Successfully Copied upgrade tool to C:\\ProgramData\\GVFS\\GVFS.Upgrade\\Tools\\GVFS.Upgrader.exe"}
[10/4/2018 8:31:06 AM] TryCopyUpgradeTool (Stop) {"DurationMs":77}
[10/4/2018 8:31:06 AM] TryLaunchUpgradeTool (Start)
[10/4/2018 8:31:06 AM] Information {"Message":"Successfully launched upgrade tool."}
[10/4/2018 8:31:06 AM] TryLaunchUpgradeTool (Stop) {"DurationMs":358}
```

I can't tell from this log:
* what version we started from
* what version we went to
* if the installers had any issues or if everything installed correctly


And here's an example of a log file from a failed install:
```
[10/4/2018 8:33:21 AM] TryCheckUpgradeInstallable (Start)
[10/4/2018 8:33:21 AM] TryRunPreUpgradeChecks (Start)
[10/4/2018 8:33:23 AM] Error {"ErrorMessage":"TryRunPreUpgradeChecks: `gvfs upgrade` is not supported because you have previously installed an out of band ProjFS driver.\r\nCheck your team's documentation for how to upgrade."}
[10/4/2018 8:33:23 AM] TryRunPreUpgradeChecks (Stop) {"DurationMs":1329}
[10/4/2018 8:33:23 AM] TryCheckUpgradeInstallable (Stop) {"DurationMs":1335}
[10/4/2018 8:33:23 AM] Error {"ErrorMessage":"TryLoadUpgradeRing failed. Invalid upgrade ring `Foo` specified in gvfs config.\r\nTo set or change upgrade ring, run `gvfs config upgrade.ring [\"Fast\"|\"Slow\"|\"None\"]` from a command prompt."}
[10/4/2018 8:33:23 AM] Error {"ErrorMessage":"TryRunProductUpgrade: Could not load upgrade ring. Invalid upgrade ring `Foo` specified in gvfs config.\r\nTo set or change upgrade ring, run `gvfs config upgrade.ring [\"Fast\"|\"Slow\"|\"None\"]` from a command prompt."}
```

The problem with this is that this is the exact same info that was shown to the user. So the log files aren't actually giving us any extra diagnostic info right now.

sanoursa: I can see that the other log file from the upgrade tool has much more useful info in it:
```
[10/4/2018 8:31:07 AM] TryCheckIfUpgradeAvailable (Start)
[10/4/2018 8:31:07 AM] Information {"Message":"Successfully checked for new release. 1.0.18234.4"}
[10/4/2018 8:31:07 AM] TryCheckIfUpgradeAvailable (Stop) {"DurationMs":389}
[10/4/2018 8:31:07 AM] TryGetNewGitVersion (Start)
[10/4/2018 8:31:07 AM] Information {"Message":"Successfully read Git version 2.18.0.gvfs.1.26"}
[10/4/2018 8:31:07 AM] TryGetNewGitVersion (Stop) {"DurationMs":2}
[10/4/2018 8:31:07 AM] Installed Version {"installedGVFSVersion":"0.2.173.2","installedGitVersion":"2.18.0.gvfs.1.26"}
[10/4/2018 8:31:07 AM] Available Version {"gvfsVersion":"1.0.18234.4","gitVersion":"2.18.0.gvfs.1.26"}
[10/4/2018 8:31:07 AM] TryRunPreUpgradeChecks (Start)
[10/4/2018 8:31:07 AM] Information {"Message":"Successfully finished pre upgrade checks. Okay to run `gvfs upgrade`."}
[10/4/2018 8:31:07 AM] TryRunPreUpgradeChecks (Stop) {"DurationMs":5}
[10/4/2018 8:31:07 AM] TryDownloadUpgrade(1.0.18234.4) (Start)
[10/4/2018 8:31:13 AM] Information {"Message":"Successfully downloaded version: 1.0.18234.4"}
[10/4/2018 8:31:13 AM] TryDownloadUpgrade(1.0.18234.4) (Stop) {"DurationMs":5945}
[10/4/2018 8:31:13 AM] Information {"Message":"Unmounting any mounted GVFS repositories."}
[10/4/2018 8:31:13 AM] TryUnmountAllGVFSRepos (Start)
[10/4/2018 8:31:14 AM] Information {"Message":"Checking if GVFS or dependent processes are running."}
[10/4/2018 8:31:15 AM] Information {"Message":"Successfully unmounted repositories."}
[10/4/2018 8:31:15 AM] TryUnmountAllGVFSRepos (Stop) {"DurationMs":1344}
[10/4/2018 8:31:15 AM] TryInstallGitUpgrade(2.18.0.gvfs.1.26) (Start)
[10/4/2018 8:31:47 AM] Information {"Message":"Successfully installed Git version: 2.18.0.gvfs.1.26"}
[10/4/2018 8:31:47 AM] TryInstallGitUpgrade(2.18.0.gvfs.1.26) (Stop) {"DurationMs":32863}
[10/4/2018 8:31:47 AM] TryInstallGVFSUpgrade(1.0.18234.4) (Start)
[10/4/2018 8:32:10 AM] Information {"Message":"Successfully installed GVFS version: 1.0.18234.4"}
[10/4/2018 8:32:10 AM] TryInstallGVFSUpgrade(1.0.18234.4) (Stop) {"DurationMs":22052}
[10/4/2018 8:32:10 AM] Newly Installed Version {"gvfsVersion":"1.0.18234.4","gitVersion":"2.18.0.gvfs.1.26"}
```

So unless we think we can add some useful diagnostic info to the first log file, I would just drop it.

(And the upgrade tool's log file still needs to add what version we're starting from)

GVFS/GVFS/CommandLine/ConfigVerb.cs
jrbriggs: I just hit this myself. Let's make all keys case insensitive.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
jrbriggs: In the same vein, gvfs diagnose does not capture the upgrade logs. Should it @sanoursa ?

sanoursa: Yes it captured the logs for me. It's in the `GVFS.Upgrade` folder in the top level of the zip file.

jrbriggs: @sanoursa pointed out i upgraded to a public version that doesn't support this -- please disregard.

alameenshah: These are the errors that can get logged into this file
1. FileBasedCollectionException info when trying to load ring config.
2. Network exception while trying to fetch Github/release JSON.
3. Parse errors while parsing Release JSON.
4. File system exceptions while trying to copy upgrader tool to temp directory in ProgramData.
5. Process exceptions while trying to launch Upgrader tool.

These errors don’t happen normally, thats the reason the log file looks un-eventful and more like a debug trace log. Its still good to keep this log file, b’coz if for some reason Upgrader tool didn’t get launched, or upgrade check failed, this is the file where we would look, for the detailed exception info.

Example: I simulated a file copy error and this is what the error log looked like
```
[10/4/2018 4:32:33 PM] TryCheckUpgradeInstallable (Start)
[10/4/2018 4:32:33 PM] TryRunPreUpgradeChecks (Start)
[10/4/2018 4:32:35 PM] Information {"Message":"Successfully finished pre upgrade checks. Okay to run `gvfs upgrade`."}
[10/4/2018 4:32:35 PM] TryRunPreUpgradeChecks (Stop) {"DurationMs":1820}
[10/4/2018 4:32:35 PM] Information {"Message":"Upgrade is installable."}
[10/4/2018 4:32:35 PM] TryCheckUpgradeInstallable (Stop) {"DurationMs":1825}
[10/4/2018 4:32:35 PM] TryCheckUpgradeAvailable (Start)
[10/4/2018 4:32:35 PM] Information {"Message":"Successfully checked server for GVFS upgrades."}
[10/4/2018 4:32:35 PM] TryCheckUpgradeAvailable (Stop) {"DurationMs":450}
[10/4/2018 4:32:35 PM] TryCopyUpgradeTool (Start)
[10/4/2018 4:32:35 PM] Error {"Method":"TrySetupToolsDirectory","Exception":"System.IO.FileNotFoundException: Could not find file 'M:\\Work\\GVFS\\BuildOutput\\GVFS.Windows\\bin\\x64\\Debug\\GVFS.Upgrader.exe'.\r\nFile name: 'M:\\Work\\GVFS\\BuildOutput\\GVFS.Windows\\bin\\x64\\Debug\\GVFS.Upgrader.exe'\r\n   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)\r\n   at System.IO.File.InternalCopy(String sourceFileName, String destFileName, Boolean overwrite, Boolean checkHost)\r\n   at GVFS.Common.ProductUpgrader.TrySetupToolsDirectory(String& upgraderToolPath, String& error)","ErrorMessage":"Error copying M:\\Work\\GVFS\\BuildOutput\\GVFS.Windows\\bin\\x64\\Debug\\GVFS.Upgrader.exe to C:\\ProgramData\\GVFS\\GVFS.Upgrade\\Tools\\GVFS.Upgrader.exe."}
[10/4/2018 4:32:35 PM] TryCopyUpgradeTool (Stop) {"DurationMs":5}
[10/4/2018 4:32:35 PM] Error {"ErrorMessage":"TryRunProductUpgrade: Could not launch upgrade tool. File copy error - Could not find file 'M:\\Work\\GVFS\\BuildOutput\\GVFS.Windows\\bin\\x64\\Debug\\GVFS.Upgrader.exe'."}
```

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: Added version info.

Also, please see my comment below. This file will contain detailed exception info, which would be missing in the Console logs.

start
alameenshah: That got added by accident, my bad. Removed it.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: Added NoConfig Ring. None and NoConfig will not be treated as errors anymore. No change in behavior for Invalid. Attaching the latest screenshot.

screen shot 2018-10-04 at 3 37 40 pm

GVFS/GVFS/CommandLine/ConfigVerb.cs
alameenshah: good catch. Thank you!

alameenshah:
#### TODO
1. Reliability issue - replace GVFS CLI invocations for repo mount/unmount with in-proc alternative using GVFSVerb. (We discussed this – GVFS and Git installer exes will continue to be launched as separate processes.)
2. Refactoring –
a.       Move out repo mount/unmount methods out of InstallerPreRunChecker to a new class in ProductUpgrader.csproj.
b.       Currently ProductUpgrader is using HttpClient to fetch Github/releases info and WebClient for asset downloads. Investigate if it is better to replace the usage of two separate APIs with one.
c.       Evaluate using Newtonsoft.json for parsing Github/release response. Right now I’m using System.Runtime.Serialization.Json.
3. Support CRUD in gvfs config
4. <Will update soon>

GVFS/GVFS.Common/ProductUpgrader.cs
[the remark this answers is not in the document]
alameenshah: >> Otherwise everyone's first interaction with this command will be an error.

Makes sense. Removed ERROR. Have attached screenshot with the new behavior in a comment below.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -0,0 +1,379 @@
...
+                {
+                    if (this.TryCopyUpgradeTool(out upgraderPath, out errorMessage) &&
+                        this.TryLaunchUpgradeTool(upgraderPath, out errorMessage))
+                    {
+                        return true;
+                    }
+
+                    return false;
+                },
+                "Launching upgrade tool",
```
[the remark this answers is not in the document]
alameenshah: It is fast on my PC too. Also there is a customer request to remove it. Removed.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -0,0 +1,212 @@
...
+                return false;
+            }
+
+            this.tracer.RelatedInfo("Successfully finished pre upgrade checks. Okay to run GVFS upgrade.");
+
+            consoleError = null;
+            return true;
+        }
+        
+        // TODO: Move repo mount calls to GVFS.Upgrader project.
```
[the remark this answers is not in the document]
alameenshah: https://github.com/Microsoft/VFSForGit/issues/293

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: Updated the messaging.

screen shot 2018-10-04 at 11 43 38 am

PR #280, opened 2018-09-18
Catch and log exceptions thrown by TryPrepareFolderForCallbacks

### LINKED ISSUE #278: TryPrepareFolderForCallbacks can throw, and the exception should be caught and logged by the clone verb
Example:

```
Cannot clone @ D:\repo: System.IO.FileNotFoundException: Could not load file or assembly 'ProjectedFSLib.Managed.dll' or one of its dependencies. The specified module could not be found.
File name: 'ProjectedFSLib.Managed.dll'
   at GVFS.Platform.Windows.ProjFSFilter.TryPrepareFolderForCallbacks(String folderPath, String& error)
   at GVFS.CommandLine.CloneVerb.CreateClone(ITracer tracer, GVFSEnlistment enlistment, GitObjectsHttpRequestor objectRequestor, GitRefs refs, String branch) in E:\A\_work\174\s\GVFS\GVFS\CommandLine\CloneVerb.cs:line 503
   at GVFS.CommandLine.CloneVerb.TryClone(JsonTracer tracer, GVFSEnlistment enlistment, CacheServerInfo cacheServer, RetryConfig retryConfig, GVFSConfig gvfsConfig, String resolvedLocalCacheRoot) in E:\A\_work\174\s\GVFS\GVFS\CommandLine\CloneVerb.cs:line 382
   at GVFS.CommandLine.CloneVerb.<>c__DisplayClass35_2.<Execute>b__2() in E:\A\_work\174\s\GVFS\GVFS\CommandLine\CloneVerb.cs:line 193
   at GVFS.Common.ConsoleHelper.<>c__DisplayClass1_0.<ShowStatusWhileRunning>b__0()
   at GVFS.Common.ConsoleHelper.ShowStatusWhileRunning(Func`1 action, String message, TextWriter output, Boolean showSpinner, String gvfsLogEnlistmentRoot, Int32 initialDelayMs)
   at GVFS.CommandLine.GVFSVerb.ShowStatusWhileRunning(Func`1 action, String message, String gvfsLogEnlistmentRoot) in E:\A\_work\174\s\GVFS\GVFS\CommandLine\GVFSVerb.cs:line 170
   at GVFS.CommandLine.CloneVerb.Execute() in E:\A\_work\174\s\GVFS\GVFS\CommandLine\CloneVerb.cs:line 200
```

A better message should also be shown to the user.

wilbaker:
Fixes #278 

`TryPrepareFolderForCallbacks ` can throw a `FileNotFoundException` if ProjFS is not properly installed\enabled.  Prior to these changes, when an exception occurred the exception would not be logged and the full call stack would be displayed to the user on the console.

With these changes any exceptions thrown by `TryPrepareFolderForCallbacks` will be caught and recorded to the log file.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
+++ b/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
@@ -30,7 +30,6 @@ namespace GVFS.Platform.Windows
         private const string FilterLoggerSessionName = "Microsoft-Windows-ProjFS-Filter-Log";
 
         private const string ProjFSNativeLibFileName = "ProjectedFSLib.dll";
-        private const string ProjFSManagedLibFileName = "ProjectedFSLib.Managed.dll";
 
         private const uint OkResult = 0;
         private const uint NameCollisionErrorResult = 0x801F0012;
@@ -321,40 +320,14 @@ namespace GVFS.Platform.Windows
             return sb.ToString();
         }
 
-        public bool TryPrepareFolderForCallbacks(string folderPath, out string error, out Exception exception)
+        public bool TryPrepareFolderForCallbacks(string folderPath, out string error)
         {
-            exception = null;
             error = string.Empty;
             Guid virtualizationInstanceGuid = Guid.NewGuid();
-
-            try
+            HResult result = VirtualizationInstance.ConvertDirectoryToVirtualizationRoot(virtualizationInstanceGuid, folderPath);
+            if (result != HResult.Ok)
             {
-                HResult result = VirtualizationInstance.ConvertDirectoryToVirtualizationRoot(virtualizationInstanceGuid, folderPath);
-                if (result != HResult.Ok)
-                {
-                    error = "Failed to prepare \"" + folderPath + "\" for callbacks, error: " + result.ToString("F");
-                    return false;
-                }
-            }
-            catch (FileNotFoundException e)
-            {
-                exception = e;
-
-                if (e.FileName.Equals(ProjFSManagedLibFileName, StringComparison.OrdinalIgnoreCase))
-                {
-                    error = $"Failed to load {ProjFSManagedLibFileName}. Ensure that ProjFS is installed and enabled";
-                }
-                else
-                {
-                    error = $"FileNotFoundException while trying to prepare \"{folderPath}\" for callbacks: {e.Message}";
-                }
-
-                return false;
-            }
-            catch (Exception e)
-            {
-                exception = e;
-                error = $"Exception while trying to prepare \"{folderPath}\" for callbacks: {e.Message}";
+                error = "Failed to prepare \"" + folderPath + "\" for callbacks, error: " + result.ToString("F");
                 return false;
             }
 
--- a/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
+++ b/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
@@ -30,6 +30,7 @@ namespace GVFS.Platform.Windows
         private const string FilterLoggerSessionName = "Microsoft-Windows-ProjFS-Filter-Log";
 
         private const string ProjFSNativeLibFileName = "ProjectedFSLib.dll";
+        private const string ProjFSManagedLibFileName = "ProjectedFSLib.Managed.dll";
 
         private const uint OkResult = 0;
         private const uint NameCollisionErrorResult = 0x801F0012;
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -320,14 +321,40 @@ public string FlushDriverLogs()
...
-        public bool TryPrepareFolderForCallbacks(string folderPath, out string error)
+        public bool TryPrepareFolderForCallbacks(string folderPath, out string error, out Exception exception)
         {
+            exception = null;
             error = string.Empty;
             Guid virtualizationInstanceGuid = Guid.NewGuid();
-            HResult result = VirtualizationInstance.ConvertDirectoryToVirtualizationRoot(virtualizationInstanceGuid, folderPath);
-            if (result != HResult.Ok)
+
+            try
```
wilbaker: Manual testing is showing that this might need to get moved one level up

GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -609,14 +609,29 @@ private void CheckNotInsideExistingRepo(string normalizedEnlistmentRootPath)
             }
 
             // Prepare the working directory folder for GVFS last to ensure that gvfs mount will fail if gvfs clone has failed
-            string prepFileSystemError;
-            if (!GVFSPlatform.Instance.KernelDriver.TryPrepareFolderForCallbacks(enlistment.WorkingDirectoryRoot, out prepFileSystemError))
+            Result prepForCallbacksResult = new Result(true);
+            try
+            {
+                string prepFileSystemError;
+                if (!GVFSPlatform.Instance.KernelDriver.TryPrepareFolderForCallbacks(enlistment.WorkingDirectoryRoot, out prepFileSystemError))
```
sanoursa: I think a better fix would be to have `TryPrepareFolderForCallbacks` handle its own exceptions - that's the contract of a TryFoo method after all, to convert exceptions to a false return value. And it would also enable us to leave the calling code here and in `DehydrateVerb` unchanged.

wilbaker: @sanoursa I actually started with that approach (see the first commit here: https://github.com/Microsoft/VFSForGit/pull/280/commits/2c2d116711e613752d556378ecd3f9b3d5ca03e7), but I found that when testing the changes that the exception is happening when attempting to call `TryPrepareFolderForCallbacks` rather than when it attempts to call `ConvertDirectoryToVirtualizationRoot`.

My guess was that .NET must be validating that all dependencies for the method are available before the method is called, but I am not 100% sure.

Thinking about this more though I *could* do something like making `TryPrepareFolderForCallbacks` a wrapper for a helper method and catch the exceptions in `TryPrepareFolderForCallbacks`.  Thoughts on that approach?

sanoursa: Yep the runtime validates all dependencies for a method at the time the method is called. Having `TryPrepareFolderForCallbacks` call another method to do the actual work is the best way to handle that type of error.

wilbaker:
@sanoursa I made another round of changes based on your feedback.  Please let me know what you think.

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -320,18 +321,34 @@ public string FlushDriverLogs()
             return sb.ToString();
         }
 
-        public bool TryPrepareFolderForCallbacks(string folderPath, out string error)
+        public bool TryPrepareFolderForCallbacks(string folderPath, out string error, out Exception exception)
```
sanoursa: Minor: both of the out params are only used for tracing errors. Should this method just take an ITracer as an input param and writes out its own errors?

wilbaker: > both of the out params are only used for tracing errors

@sanoursa, the `exception` parameter is only used for tracing, but the `error` out param is used for both tracing and for letter the user know (on the command line) why their clone as failed.

I can update `TryPrepareFolderForCallbacks` to take an `ITracer` (and remove the out `exception`), but we'd still need an out `error` parameter.  

Let me know if you think it's worth making that change.

wilbaker: Discussed with @sanoursa offline and I will merge this in as-is.

PR #291, opened 2018-09-21
Allow functional tests to be started from VS for Mac

wilbaker:
The issue was that the functional tests need to look in the Publish directory for `gvfs`.

To fix this, the functional tests will check if they were started from the debugger on Mac and adjust the path to `gvfs` as needed.

With these changes I am able to run the tests by right-clicking on the`GVFS.FunctionalTests` project and selecting "Start Debugging Item".

Note that it's still required to manually load the ProjFS kext before starting the tests in VS.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
+++ b/GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
@@ -1,4 +1,5 @@
 ﻿using System.IO;
+using System.Runtime.InteropServices;
 
 namespace GVFS.FunctionalTests
 {
@@ -16,14 +17,23 @@ namespace GVFS.FunctionalTests
 
         public static bool ReplaceInboxProjFS { get; set; }
 
+        public static bool StartedFromDebugger { get; set; }
+
         public static string PathToGVFS
         {
             get
             {
-                return
-                    TestGVFSOnPath ? 
-                    Properties.Settings.Default.PathToGVFS : 
-                    Path.Combine(Properties.Settings.Default.CurrentDirectory, Properties.Settings.Default.PathToGVFS);
+                if (TestGVFSOnPath)
+                {
+                    return Properties.Settings.Default.PathToGVFS;
+                }
+
+                if (StartedFromDebugger && RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    return Path.Combine(Properties.Settings.Default.CurrentDirectory, "..", "..", "..", "..", "..", "..", "Publish", Properties.Settings.Default.PathToGVFS);
+                }
+
+                return Path.Combine(Properties.Settings.Default.CurrentDirectory, Properties.Settings.Default.PathToGVFS);
             }
         }
     }
--- a/GVFS/GVFS.FunctionalTests/Program.cs
+++ b/GVFS/GVFS.FunctionalTests/Program.cs
@@ -13,6 +13,8 @@ namespace GVFS.FunctionalTests
     {
         public static void Main(string[] args)
         {
+            GVFSTestConfig.StartedFromDebugger = Debugger.IsAttached;
+            
             Properties.Settings.Default.Initialize();
             NUnitRunner runner = new NUnitRunner(args);
 


GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
```diff
@@ -16,14 +17,23 @@ public static class GVFSTestConfig
...
-                return
-                    TestGVFSOnPath ? 
-                    Properties.Settings.Default.PathToGVFS : 
-                    Path.Combine(Properties.Settings.Default.CurrentDirectory, Properties.Settings.Default.PathToGVFS);
+                if (TestGVFSOnPath)
+                {
+                    return Properties.Settings.Default.PathToGVFS;
+                }
+
+                if (StartedFromDebugger && RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
sanoursa: Couldn't you just check `Debugger.IsAttached` here? Why do we need a new property?

GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
```diff
@@ -16,14 +17,23 @@ public static class GVFSTestConfig
...
+                {
+                    return Path.Combine(Properties.Settings.Default.CurrentDirectory, "..", "..", "..", "..", "..", "..", "Publish", Properties.Settings.Default.PathToGVFS);
```
sanoursa: What is your goal with this change? Based on the code here, it looks like you still have to remember to publish before you run your tests. So for example, if you make a change and then just build and run, you will run the tests on the _last version you published_, not the version you just built. That seems likely to cause some a lot of time and sanity :-)

Is there any way we can remove the requirement to publish before you can run the tests? That would be ideal because we can run the tests from the debugger, and also publishing takes a non-trivial amount of time and it would be nice to get rid of that while testing changes.

GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
wilbaker: > Couldn't you just check Debugger.IsAttached here?

If we did that then the tests would start looking in the wrong path if you attached the debugger to tests started from the command line.

> Why do we need a new property?

We need the property to remember that we were started from the debugger.

GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
wilbaker: > So for example, if you make a change and then just build and run, you will run the tests on the last version you published, not the version you just built. That seems likely to cause someone a lot of time and sanity :-)

Ahh good point, I did not realize that building in VS for Mac would not perform the publish step.

>Is there any way we can remove the requirement to publish before you can run the tests? That would be ideal because we can run the tests from the debugger, and also publishing takes a non-trivial amount of time and it would be nice to get rid of that while testing changes.

I think we might be able to change this so that when started from the debugger we call `dotnet` rather than `gvfs`.  Let me give that a try.

GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
sanoursa: That's a good point, but I think the best way to guarantee a stable path would be to compute it once and store the path itself.

wilbaker: That's a good idea, I'll switch this to `{ get; set; }` and we can set it when the program starts.

GVFS/GVFS.FunctionalTests/GVFSTestConfig.cs
wilbaker: > I think we might be able to change this so that when started from the debugger we call dotnet rather than gvfs. Let me give that a try.

@sanoursa after spending a bunch of time digging into it, the approach I suggested above appears to be a real rat's nest.

I think it will be easier\clean to add a post-build step to GVFS.FunctionalTests on Mac that does a `dotnet publish` I'll see if I can get this approach to work.

wilbaker:
Closing this PR for now, will re-open once a new set of changes are ready.

PR #295, opened 2018-09-24
Add GCM_URL_OVERRIDE support for "git credential fill"

jeschu1:
GCM_URL_OVERRIDE is required for new DevOps urls to authenticate in Git Credential Manager, when command line arguments are not available.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -113,6 +113,7 @@ namespace GVFS.Common.Git
             {
                 Result gitCredentialOutput = this.InvokeGitAgainstDotGitFolder(
                     "credential fill",
+                    repoUrl,
                     stdin => stdin.Write("url=" + repoUrl + "\n\n"),
                     parseStdOutLine: null);
 
@@ -258,7 +259,7 @@ namespace GVFS.Common.Git
 
         public Result DiffTree(string sourceTreeish, string targetTreeish, Action<string> onResult)
         {
-            return this.InvokeGitAgainstDotGitFolder("diff-tree -r -t " + sourceTreeish + " " + targetTreeish, null, onResult);
+            return this.InvokeGitAgainstDotGitFolder("diff-tree -r -t " + sourceTreeish + " " + targetTreeish, null, null, onResult);
         }
 
         public Result CreateBranchWithUpstream(string branchToCreate, string upstreamBranch)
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -414,6 +417,12 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
 
             processInfo.EnvironmentVariables["GIT_TERMINAL_PROMPT"] = "0";
             processInfo.EnvironmentVariables["GCM_VALIDATE"] = "0";
+
+            if (!string.IsNullOrEmpty(repoUrl))
+            {
+                processInfo.EnvironmentVariables["GCM_URL_OVERRIDE"] = repoUrl;
```
sanoursa: As we discussed last week, I don't think this approach makes sense, because it means we now only support the GCM for cloning from Azure Repos, and all other git credential helpers are broken.

I thought what we had agreed to was set the `credential.useHttpPath` config setting for now, and then looking into adding a new config setting that improves the behavior?

jeschu1: @sanoursa credential.useHttpPath isn't an option because gcm would use that setting for both github/bitbucket.  Further to that useHttpPath does have a purpose as outlined (here) [https://github.com/Microsoft/Git-Credential-Manager-for-Windows/issues/749].

I believe we said a more preferable solution may be to add another environment variable to git.  That would be a longer way out and hasn't been prioritized.

sanoursa: It sounds like we agree on a long term fix, just not on which thing we want to break in the short term :-)

My take on it is: using the GCM-specific environment variable means we are breaking all non-GCM cred helpers. Setting the useHttpPath just means that for services other than Azure DevOps, folks get a slightly less optimal experience, but it still works. I would go for the latter as the short term solution.

sanoursa: In fact, this current solution is probably leaving our own Mac client broken as well. Does the GCM for Mac understand this new environment variable? How does it behave with the git config setting?

jeschu1: @sanoursa We had a discussion here and the new plan is to use useHttpPath as you suggest.  We would save credentials using the full path as the flag documents.  This may mean a few extra auth entries for GVFS repos, but since we're not expecting a user to have dozens of different GVFS repos on their machines it's not a real concern.  Looking at the "git credential fill" flow for useHttpPath it looks like there is some more work to do in GCM to support this.  I'm going to jump to that and then back on a new GVFS PR.

To answer your second question about GCM for Mac, I would expect *all* VSTS credential managers to be broken with the new urls :-( .  Previously you could determine tenant and save key by using hostname alone.  Now you need to know to use hostname + path[0].

@jrbriggs @jamill

PR #296, opened 2018-09-24
Update multi-pack-index to use upstream version

### LINKED ISSUE #294: Replace custom MIDX with upstream Multi-Pack-Index
We implemented the multi-pack-index feature in a rush to support our many-packs scenario in VFS for Git, but then upstreamed the commit-graph first. We have recently achieved feature-parity upstream, including all multi-pack-index features we care about like 64-bit offsets and 'git multi-pack-index verify'.

There were a few things that changed during the upstream review process:

- The `git-midx` builtin is now `git-multi-pack-index`
- The parameters are now `git  multi-pack-index write --object-dir=<dir>`
- The file storage no longer has a `midx-head` file pointing to a `midx-<hash>.midx` file. Instead, we have a `multi-pack-index` file in the pack dir.
- We no longer have a `--read` option, since the equivalent was moved to a test-tool helper. Instead, we can use the `verify` option in our tests.

So, we will want to do the following:

1. Update all references from `git midx --write --update-head --pack-dir=<dir>` to `git multi-pack-index write --object-dir=<dir>`
2. Remove code that deletes stale `.midx` files on prefetch.
3. Add minor-version upgrade code to delete `midx-head` and `*.midx` files.
4. Compute multi-pack-index on mount, so we minimize time where users are without a midx after upgrade. This action will be a quick no-op if their midx is up-to-date in later mounts.

derrickstolee:
Resolves #294.

Update GitForWindows to use upstream multi-pack-index code instead of our private version. See [the PR in Microsoft/git](https://github.com/Microsoft/git/pull/24) for more info on the Git change (and don't merge this until that is final).

This does the following:

- Update the command-line interactions that update the multi-pack-index.
- On mount, launch the post-fetch background job to ensure a multi-pack-index is calculated. (Very low cost if it already exists.)
- Since we have the multi-pack-index in more cases than before, we discover that we need to delete the multi-pack-index when deleting bad pack-files during prefetch.
- Since we trigger the post-fetch job on every mount, we now need to purposefully care about platform before shutting down the background thread. More work is required and tracked by #409.

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181012.4</GitPackageVersion>
+    <GitPackageVersion>2.20181018.1</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -155,33 +155,13 @@ namespace GVFS.Common.Git
             using (ITracer activity = tracer.StartActivity(nameof(this.TryWriteMultiPackIndex), EventLevel.Informational, Keywords.Telemetry, metadata: null))
             {
                 GitProcess process = new GitProcess(enlistment);
-                GitProcess.Result result = process.WriteMultiPackIndex(enlistment.GitPackRoot);
+                GitProcess.Result result = process.WriteMultiPackIndex(enlistment.GitObjectsRoot);
 
-                if (!result.HasErrors)
-                {
-                    string midxHash = result.Output.Trim();
-                    activity.RelatedInfo("Updated midx-head to hash {0}", midxHash);
-
-                    string expectedMidxHead = Path.Combine(enlistment.GitPackRoot, "midx-" + midxHash + ".midx");
-
-                    List<string> midxFiles = new List<string>();
-
-                    midxFiles.AddRange(fileSystem.GetFiles(enlistment.GitPackRoot, "midx-*.midx"));
-                    midxFiles.AddRange(fileSystem.GetFiles(enlistment.GitPackRoot, "tmp_midx_*"));
-
-                    foreach (string midxFile in midxFiles)
-                    {
-                        if (!midxFile.Equals(expectedMidxHead, StringComparison.OrdinalIgnoreCase) && !fileSystem.TryDeleteFile(midxFile))
-                        {
-                            activity.RelatedWarning("Failed to delete MIDX file {0}", midxFile);
-                        }
-                    }
-                }
-                else
+                if (result.HasErrors)
                 {
                     EventMetadata errorMetadata = new EventMetadata();
                     errorMetadata.Add("Operation", nameof(this.TryWriteMultiPackIndex));
-                    errorMetadata.Add("packDir", enlistment.GitPackRoot);
+                    errorMetadata.Add("objectDir", enlistment.GitObjectsRoot);
                     errorMetadata.Add("Errors", result.Errors);
                     errorMetadata.Add("Output", result.Output.Length > 1024 ? result.Output.Substring(1024) : result.Output);
                     activity.RelatedError(errorMetadata, result.Errors, Keywords.Telemetry);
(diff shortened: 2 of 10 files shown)


GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -155,33 +155,13 @@ public bool TryWriteMultiPackIndex(ITracer tracer, GVFSEnlistment enlistment, Ph
             using (ITracer activity = tracer.StartActivity(nameof(this.TryWriteMultiPackIndex), EventLevel.Informational, Keywords.Telemetry, metadata: null))
             {
                 GitProcess process = new GitProcess(enlistment);
-                GitProcess.Result result = process.WriteMultiPackIndex(enlistment.GitPackRoot);
+                GitProcess.Result result = process.WriteMultiPackIndex(enlistment.GitObjectsRoot);
 
-                if (!result.HasErrors)
```
kewillford: Are we going to leave any midx files in the pack directory and the user will have to clean up if they so desire?  Also I'm assuming git is what will keep the midx in the objects directory cleaned up?

derrickstolee: The new mechanism is to have a constant-name file ($objdir/pack/multi-pack-index) that is swapped out like the .git/index file (and the $objdir/info/commit-graph file).

We will probably want to delete the old MIDX files at some point. I wonder if that clean up should be an upgrade step, or if we should leave them in-place for a version so users can downgrade and still use their old midxes if they want.

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -546,6 +546,9 @@ private void MountAndStartWorkingDirectoryCallbacks(CacheServerInfo cache)
 
             this.heartbeat = new HeartbeatThread(this.tracer, this.fileSystemCallbacks);
             this.heartbeat.Start();
+
+            // Launch a background job to compute the multi-pack-index. Will do nothing if up-to-date.
+            this.fileSystemCallbacks.LaunchPostFetchJob(packIndexes: new List<string>());
```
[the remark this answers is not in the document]
derrickstolee: But now, users are expecting the performance benefits of a multi-pack-index, so this gives them the minimum amount of time without having one.

GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20180814.4</GitPackageVersion>
+    <GitPackageVersion>2.20180924.2-pr</GitPackageVersion>
```
kewillford: Update to release version

GVFS/GVFS.Common/Git/GitObjects.cs
derrickstolee: The old code uses a `midx-head` file containg a hash that points to a `midx-<hash>.midx` file. The new code only knows about the `multi-pack-index` file.

So, when a user downgrades, they will still have their old `midx-head` as a valid pointer, and the next prefetch would extend it.

We can do a cleanup step after this multi-pack-index version is released.

wilbaker: Ok thanks!

wilbaker:
@derrickstolee the latest commit looks good to me

derrickstolee:
@wilbaker I also figured out the Mac problem. I now launch the post-fetch job on mount to ensure we build a multi-pack-index right away. This triggered the ThreadAbort to happen (because the thread is non-null), but for now I'm adding a change to the platform to say `SupportsThreadAbort` so we can avoid this.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -198,7 +198,11 @@ public void Stop()
             lock (this.postFetchJobLock)
             {
                 // TODO(Mac): System.PlatformNotSupportedException: Thread abort is not supported on this platform
-                this.postFetchJobThread?.Abort();
+
+                if (!GVFSPlatform.Instance.IsUnderConstruction)
```
sanoursa: Can we just clean this up with a cross-plat solution now? Would it be more complex than setting a flag so the thread can terminate itself?

derrickstolee: The thread spends most of its time waiting for Git to return with a result. We can't watch a flag at the same time without creating _yet another thread_ to watch the flag and then somehow kill the other thread. Did you have a real solution in mind?

sanoursa: In that case the solution seems to be to terminate the Git process. Every Windows-specific behavior we write is technical debt that we have to come back to, so I'd prefer to see a complete fix unless there's a strong argument against it.

sanoursa: In fact even on Windows that would be the preferred behavior. It seems that today we are killing our own thread but orphaning the Git process for however long it continues to run.

derrickstolee: I added #409 to track your concerns here.

PR #298, opened 2018-09-24
Mac: Support projecting symbolic links

### LINKED ISSUE #285: Mac: Add support for projecting symlinks
Right now VFS for Git does not understand that some entries in the git index represent symlinks, so they are not projected correctly. And on Mac, the issue is made even worse because symlinks have a mode of 0, which makes the file totally inaccessible to any user except for root.

wilbaker:
Resolves #285

The changes in this PR include support for the following (on Mac):

- Creating symbolic links (where appropriate) when expanding folders
- Projection changes scenarios where links change to files (and vice versa)
- Modification of tracked symbolic links

The changes in this PR do *not* include support for:

- Detecting the creation of new symbolic links within the repo
- Upgrading to a version of Git that reports untracked symbolic links (whose paths are included in ModifiedPaths.dat)

(Support for these scenarios will be added in a follow up PR for issue #320)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -181,8 +185,35 @@ private FileSystemTaskResult ParseIndex(ITracer tracer, Stream indexStream, Func
...
+                            case FileType.GitLink:
+                                if (mode != 0)
+                                {
+                                    throw new InvalidDataException($"Invalid file mode {Convert.ToString(mode, 8)} found for link file({type:X}) in index");
+                                }
+
+                                break;
+
+                            default:
+                                throw new InvalidDataException($"Invalid object type {type:X} found in index");
```
wilbaker: "object type" -> "file type"

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -165,6 +232,67 @@ private static byte[] ToVersionIdByteArray(byte[] version)
...
         }
 
+        private bool TryGetSymLinkObjectContents(string sha, out string contents)
+        {
+            contents = null;
+            StringBuilder objectContents = new StringBuilder();
+
+            try
+            {
+                if (!this.GitObjects.TryCopyBlobContentStream(
```
wilbaker: @sanoursa let me know if you have any suggestions for a cleaner way to do this

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -165,6 +232,67 @@ private static byte[] ToVersionIdByteArray(byte[] version)
...
+                                ++bufferIndex;
+                            }
+
+                            while (bufferIndex < buffer.Length)
+                            {
+                                buffer[bufferIndex] = 0;
+                                ++bufferIndex;
+                            }
+
+                            objectContents.Append(System.Text.Encoding.ASCII.GetString(buffer));
```
wilbaker: This needs to be UTF8

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -165,6 +232,67 @@ private static byte[] ToVersionIdByteArray(byte[] version)
...
+            try
+            {
+                if (!this.GitObjects.TryCopyBlobContentStream(
+                    sha,
+                    CancellationToken.None,
+                    GVFSGitObjects.RequestSource.SymLinkCreation,
+                    (stream, blobLength) =>
+                    {
+                        // TODO(Mac): Find a better solution than reading from the stream one byte at at time
+                        byte[] buffer = new byte[4096];
```
wilbaker: We can't use chunks as we could cross a UTF8 character barrier

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -80,15 +81,44 @@ public override void Stop()
...
+                    PlaceholderVersionId,
+                    ToVersionIdByteArray(FileSystemVirtualizer.ConvertShaToContentId(sha)),
+                    (ulong)endOfFile,
+                    (ushort)(fileTypeAndMode & GitIndexProjection.FileModeMask));
+
+                return new FileSystemResult(ResultToFSResult(result), unchecked((int)result));
+            }
+            else if (fileType == GitIndexProjection.FileType.SymLink)
+            {
+                string symLinkContents;
```
sanoursa: What is `symLinkContents`? At first I thought it is trying to get the contents of the target file... but now I think it's maybe just the path of the target file? It would be nice to clarify that.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -165,6 +232,67 @@ private static byte[] ToVersionIdByteArray(byte[] version)
...
+
+            try
+            {
+                if (!this.GitObjects.TryCopyBlobContentStream(
+                    sha,
+                    CancellationToken.None,
+                    GVFSGitObjects.RequestSource.SymLinkCreation,
+                    (stream, blobLength) =>
+                    {
+                        // TODO(Mac): Find a better solution than reading from the stream one byte at at time
```
sanoursa: Can you remind me - why do we need to read one byte at a time?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
sanoursa: You probably don't need to worry about chunking anyway, if the contents are here are actually a path. Just make sure you have one buffer that's large enough for a path.

GVFS/GVFS.Virtualization/Background/FileSystemTask.cs
```diff
@@ -29,7 +29,8 @@ public enum OperationType
             OnFolderFirstWrite,
             OnIndexWriteWithoutProjectionChange,
             OnPlaceholderCreationsBlockedForGit,
-            OnFileHardLinkCreated
+            OnFileHardLinkCreated,
+            OnFileSymLinkCreated
```
sanoursa: Nit: trailing comma

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -16,6 +16,10 @@ internal partial class GitIndexParser
             private const ushort ExtendedBit = 0x4000;
             private const ushort SkipWorktreeBit = 0x4000;
 
+            private static readonly ushort FileMode755 = Convert.ToUInt16("755", 8);
+            private static readonly ushort FileMode664 = Convert.ToUInt16("644", 8);
```
sanoursa: The value here should be 664, not 644

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -54,7 +58,7 @@ public partial class GitIndexProjection : IDisposable, IProfilerOnlyIndexProject
 
         // nonDefaultFileModes is only populated when the platform supports file mode
         // On platforms that support file modes, file paths that are not in nonDefaultFileModes have mode 644
-        private Dictionary<string, ushort> nonDefaultFileModes = new Dictionary<string, ushort>(StringComparer.OrdinalIgnoreCase);
+        private Dictionary<string, ushort> nonDefaultFileTypesAndModes = new Dictionary<string, ushort>(StringComparer.OrdinalIgnoreCase);
```
sanoursa: Since this is a sparse data structure anyway, could we just store the result as a class with pre-parsed fields? I saw that we're already masking out the sub-fields when parsing anyway, then storing the combined ushort, then the Mac virtualizer has to mask them out again.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -355,6 +355,38 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
     return PrjFS_Result_EIOError;
 }
 
+PrjFS_Result PrjFS_WriteSymLink(
```
sanoursa: I'm wondering if we actually need symlink methods in our API? I don't think it hurts anything, and we can always think about cleaning this up later, but just to make sure I'm understanding it properly - there's no special logic in here related to our placeholder formats or anything, right? The provider could just as easily create a symlink, or delete a placeholder to replace with a symlink, right?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: >could we just store the result as a class with pre-parsed fields? 

Yep we could do that.  

Do you have a strong preference on storing the data as separate fields vs a single `ushort` with properties that mask out the values for the caller?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: >there's no special logic in here related to our placeholder formats or anything, right? The provider could just as easily create a symlink, or delete a placeholder to replace with a symlink, right?

I had just written up a response that you're correct and that I only did this for API consistency but I then realized that we do need some special logic here.

After the symlink is created it should be flagged as being in the root.  I will make that change (and it's one more reason for keeping these symlink APIs in our library).

sanoursa: Ah yea that's a good point. In fact, the provider MUST NOT create any sort of file on its own, if it expects to ever get back notifications on those files.

I can see two ways of dealing with this, though all of this is out of scope for this PR:
1) panic if the provider creates a new file on its own
2) rather than early-existing our kext on all IO from the provider, handle some of the events that are safe. E.g. new file events need to get sent up to the lib so the files can get stamped with the right flags, but the lib could be smart about not forwarding that on to the provider if the provider itself was the cause of the event. And now that I've typed that out, it feels like a can of worms and option 1 may be the safer one :-)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SymbolicLinkTests.cs
```diff
@@ -0,0 +1,163 @@
...
+using GVFS.FunctionalTests.Tools;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    // MacOnly until issue #297 (add SymLink support for Windows) is complete
+    [Category(Categories.MacOnly)]
+    [TestFixture]
+    public class SymbolicLinkTests : TestsWithEnlistmentPerFixture
```
kewillford: Do we need any tests the have symlinks to directories or directories to symlinks?  Also do we need any tests for syslinks and modified paths so files that are written to?

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+namespace GVFS.Virtualization.Projection
+{
+    public struct FileTypeAndMode
```
kewillford: Do we need this whole new class instead of just using the ushort and having some method to check the ushort for values?

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
```diff
@@ -86,15 +94,22 @@ public static void DeleteDirectoryWithUnlimitedRetries(string path)
             }
         }
 
-        public override bool FileExists(string path)
+        public bool IsSymbolicLink(string path)
```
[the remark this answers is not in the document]
wilbaker: > This would make it so the test didn't have to use the BashRunner directly.

My thought had been that symbolic link tests would just always use BashRunner (much like how many of the tests explicitly use `SystemIORunner`). 

BashRunner natively supports checking if a path is a symlink (and it's cross platform) which made it a good candidate for the symlink tests.

I'm not sure that running these tests with additional runners gets us much more coverage.  

What do you think about this approach?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SymbolicLinkTests.cs
wilbaker: > Do we need any tests the have symlinks to directories or directories to symlinks?

I think adding symlinks to directories is a good idea.  What did you have in mind for "directories to symlinks"?

> Also do we need any tests for syslinks and modified paths so files that are written to?

That's a good idea as well. Symlinks are added to ModifiedPaths.dat as soon as they're written by VFS for Git, and so I should be able to modify these tests to confirm that paths are being added there.

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
wilbaker: > Do we need this whole new class instead of just using the ushort and having some method to check the ushort for values?

There was a suggestion in a previous round of PR feedback that we use a more explicit type for type+mode.  

@sanoursa, after seeing how the code looks with a wrapper `struct` do you have a preference on sticking with the struct or just providing helper methods for the `ushort`?

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
```diff
@@ -53,6 +53,14 @@ public BashRunner()
             }
         }
 
+        private enum FileType
```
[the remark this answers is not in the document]
sanoursa: I concur that all directories are also files. We're not totally consistent everywhere, but if we want to make a change, I would want to go towards more consistency on directories-as-files.

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
```diff
@@ -0,0 +1,78 @@
...
+        public enum FileType : ushort
+        {
+            Invalid = 0,
+
+            Regular = 0x8000,
+            SymLink = 0xA000,
+            GitLink = 0xE000,
+        }
+
+        public FileType Type => (FileType)(this.fileTypeAndMode & FileTypeMask);
```
sanoursa: To me, part of the value of creating a new type is to not pay the cost of masking every time we access a property. I would either store it as a ushort and add helper methods for masking (still keeping those details private to the parsing code), or else make this a class with separate fields. I also don't see much value in making this a struct.

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
sanoursa: I think it depends on the cost of masking out the bits of the ushort. Right now we're doing it twice - once in the index parsing code, and then again in the provider. If that cost doesn't matter, than storing a ushort with helper methods could suffice, though I think a new class makes the code cleaner.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -181,8 +181,36 @@ private FileSystemTaskResult ParseIndex(ITracer tracer, Stream indexStream, Func
...
-                        ushort mode = this.ReadUInt16();
-                        this.resuableParsedIndexEntry.FileMode = (ushort)(mode & 0x1FF);
+                        ushort indexFormatTypeAndMode = this.ReadUInt16();
+
+                        FileTypeAndMode typeAndMode = new FileTypeAndMode(indexFormatTypeAndMode);
+
+                        switch (typeAndMode.Type)
+                        {
+                            case FileTypeAndMode.FileType.Regular:
+                                if (typeAndMode.Mode != FileTypeAndMode.FileMode755 &&
```
sanoursa: Here we're actually paying the masking cost 3 times.

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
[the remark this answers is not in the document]
kewillford: I agree that we should try to be consistent whatever we do.  Since this is private would it make any sense to use `FileAttributes` enum instead.  Doesn't look like there is something specific for symlinks other than ReparsePoint so maybe that doesn't quite fit.  Just throwing it out there.

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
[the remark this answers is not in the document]
kewillford: That should be fine for now but if users can create symlinks in other ways like System.IO or cmd we should probably be testing those to make sure git and VFSforGit are handling them correctly.  Maybe all the runners are hitting the same code at some point so that would make it easier to say they are testing the same thing?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SymbolicLinkTests.cs
kewillford: Sorry what I meant on this was during a checkout when an item in the index can change from one type to another, so for example you are on a branch where `foo` is a directory and checkout another branch and now `foo` is a symlink.

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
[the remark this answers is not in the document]
wilbaker: > That should be fine for now but if users can create symlinks in other ways like System.IO or cmd we should probably be testing those to make sure git and VFSforGit are handling them correctly.

Yep I agree.  When we add code to support the detection of user-created symlinks we'll want to make sure that all of our runners can create them .

Currently the tests are just validating that VFS for Git is writing symlinks when appropriate, and so I think it's less critical for these tests to check with every runner.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SymbolicLinkTests.cs
wilbaker: Ahh that makes sense thanks! I will add those tests as well.

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
wilbaker: > If that cost doesn't matter, than storing a ushort with helper methods could suffice, though I think a new class makes the code cleaner.

The cost of masking is very low, and it's also true that we won't add much memory overhead to store the values separately.

I can switch to applying the mask once and having our own values for type (rather than re-using the values from git).  If in the future we need the memory back we can switch to using a `ushort`.

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
wilbaker: > I also don't see much value in making this a struct.

The value I saw in using a struct if we avoid having to store reference + data in memory.

What do you see as the downside of using a struct?

GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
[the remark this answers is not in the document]
wilbaker: > Since this is private would it make any sense to use FileAttributes enum instead. Doesn't look like there is something specific for symlinks other than ReparsePoint so maybe that doesn't quite fit. Just throwing it out there.

I think in this case we're probably best with our own enum as this is really just to control how we're checking for a few specific types of files.  Perhaps a better name for the enum would be a better option.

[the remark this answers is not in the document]
kewillford: Yeah this is good for now and if we need something different we can revisit it.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SymbolicLinkTests.cs
wilbaker: @kewillford I've added the tests and validate you suggested, and fixed a bug in `ModifiedPathsShouldContain` where `ModifiedPathsShouldContain` would succeed if a line existed that ended with the path being validated (e.g. ModifiedPathsShouldContain("foo") would pass if the file had the line `bar/foo`)

kewillford: Looks good thanks!

GVFS/GVFS.UnitTests/Platform.Mac/MacFileSystemVirtualizerTests.cs
```diff
@@ -108,7 +107,7 @@ public void UpdatePlaceholderIfNeeded()
                 mockVirtualization.UpdatePlaceholderIfNeededFailureCause = UpdateFailureCause.NoFailure;
                 virtualizer
                     .UpdatePlaceholderIfNeeded(
-                        "test.txt",
+                        filePath,
```
kewillford: This changed from `"test.txt"` to `"test/test.txt"`  Was that intended?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -80,15 +83,43 @@ public override void Stop()
...
         {
             // TODO(Mac): Add functional tests that validate file mode is set correctly
-            ushort fileMode = this.FileSystemCallbacks.GitIndexProjection.GetFilePathMode(relativePath);
-            Result result = this.virtualizationInstance.WritePlaceholderFile(
-                relativePath,
-                PlaceholderVersionId,
-                ToVersionIdByteArray(FileSystemVirtualizer.ConvertShaToContentId(sha)),
-                (ulong)endOfFile,
-                fileMode);
+            FileTypeAndMode fileTypeAndMode = this.FileSystemCallbacks.GitIndexProjection.GetFileTypeAndMode(relativePath);
```
sanoursa: As far as API goes, is there any reason FileType and Mode should be tied together? Or are we still leaking private details of the index format?

I think this `FileTypeAndMode` class is good for storing the values internally, but seeing this usage, I'm wondering if we should just have separate `GetFileType` and `GetMode` methods. What do you think?

Overall, I'd say this is pretty minor now, since at least the calling code isn't dealing with masks.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -165,6 +232,86 @@ private static byte[] ToVersionIdByteArray(byte[] version)
...
+                            else
+                            {
+                                buffer[bufferIndex - 1] = 0;
+
+                                EventMetadata metadata = this.CreateEventMetadata();
+                                metadata.Add(nameof(sha), sha);
+                                metadata.Add("bufferContents", Encoding.UTF8.GetString(buffer));
+                                this.Context.Tracer.RelatedError(metadata, $"{nameof(this.TryGetSymLinkTarget)}: SymLink target exceeds buffer size");
+
+                                throw new GetSymLinkTargetException("SymLink target exceeds buffer size");;
```
sanoursa: Would an `InvalidDataException` suffice here?

GVFS/GVFS.UnitTests/Platform.Mac/MacFileSystemVirtualizerTests.cs
```diff
@@ -189,7 +279,7 @@ public void OnEnumerateDirectoryReturnsSuccessWhenResultsNotInMemory()
                 backgroundFileSystemTaskRunner: backgroundTaskRunner,
                 fileSystemVirtualizer: virtualizer))
             {
-                gitIndexProjection.MockFileModes.TryAdd("test" + Path.DirectorySeparatorChar + "test.txt", FileMode644);
+                gitIndexProjection.MockFileTypesAndModes.TryAdd("test" + Path.DirectorySeparatorChar + "test.txt", Regular644FileTypeAndMode);
```
kewillford: This path is used a bunch but in some places it is using Path.Combine and others `"test" + Path.DirectorySeparatorChar + "test.txt"`.  Do you know why?  Is this an inconsistency that we should clean up?  Should there be a const for this path?

GVFS/GVFS.UnitTests/Platform.Mac/MacFileSystemVirtualizerTests.cs
wilbaker: Yep this was to match the path we're using above when adding the test file to `gitIndexProjection.MockFileTypesAndModes`.

GVFS/GVFS.Virtualization/Projection/FileTypeAndMode.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            switch (typeAndModeInIndexFormat & FileTypeMask)
+            {
+                case RegularFileIndexEntry:
+                    this.Type = FileType.Regular;
+                    break;
+                case SymLinkFileIndexEntry:
+                    this.Type = FileType.SymLink;
+                    break;
+                case GitLinkFileIndexEntry:
```
sanoursa: We don't actually support gitlinks right? If we ever see a repo with this data, we may want to fail hard as early as possible, rather than wait and see...

wilbaker: I thought about making a change so that we'd throw when parsing the index if we encounter a git link (i.e. submodule).  

However, doing so would allow someone to make their repo unmountable:

1. Clone repo
2. Add git submodule
3. The next time VFS for Git parses the index it would throw an exception that results in GVFS.Mount.exe shutting down
4. Repo cannot be mounted or repaired

This also seemed outside the scope of adding symlink support, and so if we do want to add this validation (and prevent a repo from becoming unmountable) it's probably a big enough task to put in its own PR.

GVFS/GVFS.UnitTests/Platform.Mac/MacFileSystemVirtualizerTests.cs
wilbaker: Actually, I need to dig more into this and see how the mock paths are getting set up.

Thanks for pointing this out!

PR #308, opened 2018-09-26
Compress the entries in the modified paths database when mounting

kewillford:
This will search the modified paths for a parent folder and if one
exists it will remove the entry from the modified paths list


GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -52,6 +52,34 @@ public static bool TryLoadOrCreate(ITracer tracer, string dataDirectory, Physica
             return true;
         }
 
+        public void Compress(ITracer tracer)
```
wilbaker: How long does this take with a large number of entries?

sanoursa: I think this isn't so much "Compress" as it is "RemoveDuplicateEntries", right?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -89,6 +89,9 @@ public FileSystemCallbacks(GVFSContext context, GVFSGitObjects gitObjects, RepoM
                 throw new InvalidRepoException(error);
             }
 
+            this.modifiedPaths.Compress(this.context.Tracer);
```
sanoursa: This is a lot of work to be doing in a constructor, and it can have surprising consequences. Is there anywhere that we can do this work after this point?

GVFS/GVFS.UnitTests/Common/ModifiedPathsDatabaseTests.cs
```diff
@@ -74,6 +86,18 @@ public void DirectorySeparatorAddedForFolder()
             TestAddingPath(pathToAdd: Path.Combine("dir", "subdir"), pathInList: Path.Combine("dir", "subdir") + Path.DirectorySeparatorChar, isFolder: true);
         }
 
+        [TestCase]
+        public void CompressEntries()
+        {
+            ModifiedPathsDatabase mpd = CreateModifiedPathsDatabase(EntriesToCompress);
```
sanoursa: `mpd` sounds like a police department. Please avoid abbreviations.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -52,6 +52,34 @@ public static bool TryLoadOrCreate(ITracer tracer, string dataDirectory, Physica
...
+            int startingCount = this.modifiedPaths.Count;
+            using (ITracer activity = tracer.StartActivity("Compress ModifiedPaths", EventLevel.Informational))
+            {
+                foreach (var item in this.modifiedPaths)
+                {
+                    int pathSeparatorIndex = item.IndexOf('/');
+                    while (pathSeparatorIndex >= 0 && pathSeparatorIndex < item.Length - 1)
+                    {
+                        string folder = item.Substring(0, pathSeparatorIndex + 1);
+                        if (this.modifiedPaths.Contains(folder))
```
sanoursa: Won't this end up deleting every folder from the collection? Every folder entry ends with a `/`, and so the `folder` variable will end up consuming the entire string, and then find a match, right?

sanoursa: Hm, perhaps what's preventing that is the `item.Length - 1` up on line 63. That's super subtle, and easy to get wrong. Since we're doing a bunch of string parsing/creating here anyway, it seems far safer and not a big perf hit to just call `string.Split` on `/`, and then ensure that we only consider the path parts up to, but not including, the leaf part.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -52,6 +52,34 @@ public static bool TryLoadOrCreate(ITracer tracer, string dataDirectory, Physica
...
+        {
+            int startingCount = this.modifiedPaths.Count;
+            using (ITracer activity = tracer.StartActivity("Compress ModifiedPaths", EventLevel.Informational))
+            {
+                foreach (var item in this.modifiedPaths)
+                {
+                    int pathSeparatorIndex = item.IndexOf('/');
+                    while (pathSeparatorIndex >= 0 && pathSeparatorIndex < item.Length - 1)
+                    {
+                        string folder = item.Substring(0, pathSeparatorIndex + 1);
```
sanoursa: This is another subtle detail. The `+ 1` here is super important, to ensure that there is a trailing `/`, and avoid matching on a file with the same name rather than a folder with that name.

I would make this much more explicit by removing the `+ 1`, and adding a `+ "/"` at the end. But this issue will be moot if you use `string.Split`, since the `/`'s won't be part of the strings, and you'll have to forcibly reinsert them anyway.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford:
```
[9/25/2018 6:16:15 PM] Compress ModifiedPaths (Stop) {"startingCount":1839665,"EndCount":9528,"DurationMs":1451}
```

GVFS/GVFS.UnitTests/Common/ModifiedPathsDatabaseTests.cs
```diff
@@ -17,6 +18,17 @@ public class ModifiedPathsDatabaseTests
         private const string ExistingEntries = @"A file.txt
 A dir/file2.txt
 A dir1/dir2/file3.txt
+";
+        private const string EntriesToCompress = @"A file.txt
```
sanoursa: I think we need to test more data shapes than this. Adds and deletes, for one thing, but also things like both a file and a folder with the same name, and perhaps others as well.

This feature feels like it carries some risk, so we definitely need to make sure we're getting enough test coverage on it. Thoughts on adding some functional tests as well?

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford: Maybe "RemoveCoveredEntries" or "RemoveFolderEntries" because these really aren't duplicates.
`dir/file.txt` is not a duplicate of `dir/` but will be removed because of it.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford: I will test using split and see what the performance and memory use of that is.

GVFS/GVFS.UnitTests/Common/ModifiedPathsDatabaseTests.cs
kewillford: Sure we already have the `ModifiedPathsTests` that we can add to.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: We can move it to the `TryStart` method.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
sanoursa: "Duplicate" in the sense that it adds no new behavior. "Compress" in the sense that we're removing redundant string and making the file smaller. Yea this can go either way, but some comments explaining its purpose would probably be helpful in this case since it's not something that's easy to capture in just a name.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
kewillford: Memory usage was about the same and performance was
```
Current: {"startingCount":1839665,"EndCount":9528,"DurationMs":1409}
Using Split: {"startingCount":1839665,"EndCount":9528,"DurationMs":1990}
```

Since there isn't a huge difference in performance and using Split is more readable, I'll switch to use split.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/ModifiedPathsTests.cs
```diff
@@ -138,9 +133,16 @@ public void ModifiedPathsSavedAfterRemount(FileSystemRunner fileSystem)
...
-            string folderToDelete = this.Enlistment.GetVirtualPathTo(FolderToDelete);
-            fileSystem.DeleteDirectory(folderToDelete);
-            folderToDelete.ShouldNotExistOnDisk(fileSystem);
+            string folderToDeleteFullPath = this.Enlistment.GetVirtualPathTo(FolderToDelete);
+            fileSystem.WriteAllText(Path.Combine(folderToDeleteFullPath, "NewFile.txt"), "Contents for new file");
+            string newFileToDelete = Path.Combine(folderToDeleteFullPath, "NewFileToDelete.txt");
+            fileSystem.WriteAllText(newFileToDelete, "Contents for new file");
+            fileSystem.DeleteFile(newFileToDelete);
+            fileSystem.WriteAllText(Path.Combine(folderToDeleteFullPath, "CreateCommonVersionHeader.bat"), "Changing the file contents");
+            fileSystem.DeleteFile(Path.Combine(folderToDeleteFullPath, "RunUnitTests.bat"));
```
wilbaker: What was the reason for adding this additional I/O to this test?

kewillford: I wanted to make sure there were entries in the modified paths under a folder that would be should be removed on the remount.

kewillford:
@sanoursa do you have any additional feedback or requested changes for this PR?

PR #313, opened 2018-09-28
CloneVerb: Force download of commit and trees on failed initial checkout

### LINKED ISSUE #306: `gvfs clone` may fail due to prefetch packfile "gaps"
The prefetch packfiles use timestamps to signal "this is what the cache server knew about at this time". When a client prefetches, they contact _a_ cache server with their latest timestamp and the server responds with a list of packfiles. As we call different cache servers, we can get into a state where there are "gaps": some commits and trees never exist in a prefetch pack.

In a normal enlistment, these gaps can be refilled on-demand. However, at clone time we may be directed to an existing shared object cache, see that the cache has the commit and root tree that we expect, but be missing a tree that is reachable from that root tree. As we attempt to run a `git checkout`, we will fail to find that object and we don't have the read-object hook available.

This is a very odd data shape that has always been possible, but it is more likely to be hit now that we have background prefetch (we are more likely to satisfy `CommitAndRootTreeExists()` for the latest ref).

One fix is to run `git rev-list --objects <root-tree-hash>` instead of just checking if the objects exist. If that command succeeds, then we can continue without downloading the initial packfile. This will take less time than the `git checkout` step, so may be reasonable.

Another fix is to always download that initial packfile. This is likely not the best solution because we rarely _need_ that data a second time, and we would be slowing down clones that are usually "trivial".

derrickstolee:
On Clone, we need to create the `.git/index` file during a checkout, which requires having all trees reachable from the root tree of the tip commit. To ensure we have all of the objects, we call `TryDownloadCommit()`. However, this method doesn't actually try downloading a pack of trees if we already have the commit and root tree.

Due to the shared object cache, we may have these objects for a few reasons:

1. A recent fetch included a ref update that points to that commit, and we downloaded the commit and tree as loose objects.
2. A background prefetch downloaded a prefetch pack containing that commit and tree, but some of the reachable trees are missing due to clock skew between cache servers.

This fixes the issue by forcibly downloading the packfile if the checkout fails and trying again.

Resolves #306.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/SharedCacheTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/SharedCacheTests.cs
@@ -96,7 +96,7 @@ namespace GVFS.FunctionalTests.Tests.MultiEnlistmentTests
             enlistment1.Status().ShouldContain("Mount status: Ready");
             enlistment2.Status().ShouldContain("Mount status: Ready");
         }
-        
+
         [TestCase]
         public void ParallelReadsInASharedCache()
         {
@@ -265,7 +265,30 @@ namespace GVFS.FunctionalTests.Tests.MultiEnlistmentTests
             newObjectsRoot.ShouldContain(newCacheKey);
             newObjectsRoot.ShouldBeADirectory(this.fileSystem);
 
-            this.AlternatesFileShouldHaveGitObjectsRoot(enlistment);            
+            this.AlternatesFileShouldHaveGitObjectsRoot(enlistment);
+        }
+        
+        [TestCase]
+        public void SecondCloneSucceedsWithMissingTrees()
+        {
+            string newCachePath = Path.Combine(this.localCacheParentPath, ".customGvfsCache2");
+            GVFSFunctionalTestEnlistment enlistment1 = this.CreateNewEnlistment(localCacheRoot: newCachePath);
+            File.ReadAllText(Path.Combine(enlistment1.RepoRoot, WellKnownFile));
+
+            this.AlternatesFileShouldHaveGitObjectsRoot(enlistment1);
+
+            string packDir = Path.Combine(enlistment1.GetObjectRoot(this.fileSystem), "pack");
+            string[] packDirFiles = Directory.EnumerateFiles(packDir, "*", SearchOption.AllDirectories).ToArray();
+            for (int i = 0; i < packDirFiles.Length; i++)
+            {
+                this.fileSystem.DeleteFile(Path.Combine(packDir, packDirFiles[i]));
+            }
+
+            // Enumerate the root directory, populating the tip commit and root tree
+            this.fileSystem.EnumerateDirectory(enlistment1.RepoRoot);
+
+            GVFSFunctionalTestEnlistment enlistment2 = this.CreateNewEnlistment(localCacheRoot: newCachePath);
+            this.fileSystem.EnumerateDirectory(enlistment2.RepoRoot);
         }
 
         // Override OnTearDownEnlistmentsDeleted rathern than using [TearDown] as the enlistments need to be unmounted before
--- a/GVFS/GVFS/CommandLine/CloneVerb.cs
+++ b/GVFS/GVFS/CommandLine/CloneVerb.cs
@@ -572,7 +572,7 @@ namespace GVFS.CommandLine
                 // It is possible to have the above TryDownloadCommit() fail because we
                 // already have the commit and root tree we intend to check out, but
                 // don't have a tree further down the working directory. If we fail
-                // checkout here, it may be due to not having these trees and the
+                // checkout here, its' because we don't have these trees and the
                 // read-object hook is not available yet. Force downloading the commit
                 // again and retry the checkout.
 
@@ -583,7 +583,7 @@ namespace GVFS.CommandLine
                     gitObjects,
                     gitRepo,
                     out errorMessage,
-                    ignoreIfRootExists: false))
+                    checkLocalObjectCache: false))
                 {
                     return new Result(errorMessage);
                 }
--- a/GVFS/GVFS/CommandLine/CloneVerb.cs
+++ b/GVFS/GVFS/CommandLine/CloneVerb.cs
@@ -567,6 +567,30 @@ namespace GVFS.CommandLine
             }
 
             GitProcess.Result forceCheckoutResult = git.ForceCheckout(branch);
+            if (forceCheckoutResult.HasErrors && forceCheckoutResult.Errors.IndexOf("unable to read tree") > 0)
+            {
+                // It is possible to have the above TryDownloadCommit() fail because we
+                // already have the commit and root tree we intend to check out, but
+                // don't have a tree further down the working directory. If we fail
+                // checkout here, it may be due to not having these trees and the
+                // read-object hook is not available yet. Force downloading the commit
+                // again and retry the checkout.
+
+                if (!this.TryDownloadCommit(
+                    refs.GetTipCommitId(branch),
+                    enlistment,
+                    objectRequestor,
+                    gitObjects,
+                    gitRepo,
+                    out errorMessage,
+                    ignoreIfRootExists: false))
+                {
+                    return new Result(errorMessage);
+                }
+
+                forceCheckoutResult = git.ForceCheckout(branch);
+            }
+
             if (forceCheckoutResult.HasErrors)
             {
                 string[] errorLines = forceCheckoutResult.Errors.Split('\n');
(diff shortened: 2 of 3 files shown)


GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -405,11 +405,23 @@ protected void ValidatePathParameter(string path)
...
-                    return false;
-                }
+                return this.ForceTryDownloadCommit(commitId, objectRequestor, gitObjects, out error);
+            }
+
+            error = null;
+            return true;
+        }
+
+        protected bool ForceTryDownloadCommit(
```
sanoursa: I don't see anything being "forced" here. This is really referring to the _other_ method, which first checks the cache. I would suggest just adding a `skipLocalCacheCheck` param to the existing method instead.

GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -567,6 +567,27 @@ private void CheckNotInsideExistingRepo(string normalizedEnlistmentRootPath)
...
+            if (forceCheckoutResult.HasErrors)
+            {
+                // It is possible to have the above TryDownloadCommit() fail because we
+                // already have the commit and root tree we intend to check out, but
+                // don't have a tree further down the working directory. If we fail
+                // checkout here, it may be due to not having these trees and the
+                // read-object hook is not available yet. Force downloading the commit
+                // again and retry the checkout.
+
+                if (!this.ForceTryDownloadCommit(
```
sanoursa: I don't think we should just re-download on all errors. We know exactly when this specific error happens because the error message says something about missing trees. I would check for that and only redownload/re-checkout in that case.

sanoursa: How much work would it be to write a functional test for this scenario?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -405,11 +405,23 @@ protected void ValidatePathParameter(string path)
...
+            string commitId,
+            GitObjectsHttpRequestor objectRequestor,
+            GVFSGitObjects gitObjects,
+            out string error)
+        {
+            if (!gitObjects.TryDownloadCommit(commitId))
```
wilbaker: Does the objects endpoint always respond with all of the trees reachable by `commitId`?

wilbaker:
>Due to the shared object cache, we may have these objects for a few reasons:
>
> 1. A recent fetch included a ref update that points to that commit, and we downloaded the commit and tree as loose objects.

I'm trying to make sure I understand this scenario.  If this were to happen, is the problem that not *all* reachable trees were downloaded?  Or is the issue that they were downloaded as loose objects?

GVFS/GVFS/CommandLine/CloneVerb.cs
derrickstolee: @sanoursa @jrbriggs How much are we dependent on the installed language (for Git)? If we search for a specific string, then we will be broken if someone installs using a different locale. Or, is there a way to force the language in a specific call?

derrickstolee: I've thought about how to set this up, and it is really difficult to do. I don't think the severity of this issue is worth engineering time on the test.

GVFS/GVFS/CommandLine/GVFSVerb.cs
derrickstolee: All trees reachable from its root tree, yes.

derrickstolee:
> 
> 
> > Due to the shared object cache, we may have these objects for a few reasons:
> > 
> > 1. A recent fetch included a ref update that points to that commit, and we downloaded the commit and tree as loose objects.
> 
> I'm trying to make sure I understand this scenario. If this were to happen, is the problem that not _all_ reachable trees were downloaded? Or is the issue that they were downloaded as loose objects?

The issue is that we don't have all reachable trees. When we ask for objects one-at-a-time, we don't close by reachability. But on clone time, we ask for and expect a reachable closure of trees. Our condition of checking just the commit and root tree is insufficient for this case.

GVFS/GVFS/CommandLine/GVFSVerb.cs
derrickstolee: :+1:

GVFS/GVFS/CommandLine/CloneVerb.cs
sanoursa: What is the difficulty? We already have some tests that can work with multiple repos, so at first glance, it seems like you could just clone once at one branch, touch some other branch's root tree object, then clone again at that second branch.

sanoursa: That's a great question, but we're already dependent on some error strings. E.g. see the very next section in this method, we're checking the checkout errors for certain errors that we ignore and consider non-errors.

derrickstolee: The problem is that you need to somehow guarantee a missing tree. This means we need to disable the background prefetch for the first enlistment in order to ensure we don't get the objects.

sanoursa: Would turning on unattended mode for this test achieve that?

sanoursa: Or alternatively, after you unmount the first repo, delete any packfiles from its cache.

sanoursa: Actually, this should be a non-issue. Background prefetch is already disabled if there is no cache server configured, and our functional tests do not use a cache server.

derrickstolee: I can add an extra check here, if we are already tied to the message. We will need to consider multi-language support at some time, and working with a non-English Git will be important before we localize VFS for Git.

derrickstolee: It took some trial-and-error, but deleting the packfiles, downloading the commit and root tree as loose, then re-cloning worked. I ran the full functional test suite to verify and inspected the logs to ensure it did everything as expected.

GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/SharedCacheTests.cs
```diff
@@ -60,6 +60,28 @@ public void SecondCloneDoesNotDownloadAdditionalObjects()
...
+            GVFSFunctionalTestEnlistment enlistment1 = this.CloneAndMountEnlistment();
+            File.ReadAllText(Path.Combine(enlistment1.RepoRoot, WellKnownFile));
+
+            this.AlternatesFileShouldHaveGitObjectsRoot(enlistment1);
+
+            string packDir = Path.Combine(enlistment1.GetObjectRoot(this.fileSystem), "pack");
+            string[] packDirFiles = Directory.EnumerateFiles(packDir, "*", SearchOption.AllDirectories).ToArray();
+            for (int i = 0; i < packDirFiles.Length; i++)
+            {
+                this.fileSystem.DeleteFile(Path.Combine(packDir, packDirFiles[i]));
```
sanoursa: Technically, this is changing the behavior of all the tests that follow. Can we do a prefetch at the end of this test to get back to the original state?

sanoursa: Your other option would be to use a new cache location for this test case only, and clone with `--no-prefetch`

GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -567,6 +567,30 @@ private void CheckNotInsideExistingRepo(string normalizedEnlistmentRootPath)
             }
 
             GitProcess.Result forceCheckoutResult = git.ForceCheckout(branch);
+            if (forceCheckoutResult.HasErrors && forceCheckoutResult.Errors.IndexOf("unable to read tree") > 0)
+            {
+                // It is possible to have the above TryDownloadCommit() fail because we
+                // already have the commit and root tree we intend to check out, but
+                // don't have a tree further down the working directory. If we fail
+                // checkout here, it may be due to not having these trees and the
```
sanoursa: It's no longer "may be" since you're now checking the specific error

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -401,9 +401,10 @@ protected void ValidatePathParameter(string path)
             GitObjectsHttpRequestor objectRequestor,
             GVFSGitObjects gitObjects,
             GitRepo repo,
-            out string error)
+            out string error,
+            bool ignoreIfRootExists = true)
```
sanoursa: This name is confusing, and results in a double-negative check down below with `!ignore`. Rename it to `checkLocalObjectCache`?

PR #314, opened 2018-09-28
Check modified paths for parent directories before adding

kewillford:


GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -183,6 +185,22 @@ private bool TryParseRemoveLine(string line, out string key, out string error)
             return true;
         }
 
+        private bool ContainsParentDirectory(string modifiedPath)
+        {
+            string[] pathParts = modifiedPath.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            StringBuilder parentFolder = new StringBuilder();
```
sanoursa: I think this `StringBuilder` is actually causing us to do extra work here, since we have to convert every path to a string anyway on line 195. I would just use normal string concatenation in this code.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -183,6 +185,22 @@ private bool TryParseRemoveLine(string line, out string key, out string error)
             return true;
         }
 
+        private bool ContainsParentDirectory(string modifiedPath)
+        {
+            string[] pathParts = modifiedPath.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            StringBuilder parentFolder = new StringBuilder();
+            foreach (string pathPart in pathParts.Take(pathParts.Length - 1))
+            {
+                parentFolder.Append(pathPart + "/");
```
sanoursa: Use the constant for git path separator?

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -183,6 +185,22 @@ private bool TryParseRemoveLine(string line, out string key, out string error)
             return true;
         }
 
+        private bool ContainsParentDirectory(string modifiedPath)
+        {
+            string[] pathParts = modifiedPath.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
```
sanoursa: Same here, use the existing constant

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -183,6 +185,22 @@ private bool TryParseRemoveLine(string line, out string key, out string error)
             return true;
         }
 
+        private bool ContainsParentDirectory(string modifiedPath)
+        {
+            string[] pathParts = modifiedPath.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            StringBuilder parentFolder = new StringBuilder();
+            foreach (string pathPart in pathParts.Take(pathParts.Length - 1))
```
sanoursa: Not sure how much of a perf impact this has, but `Take` is causing us to construct a new object, whereas we could just use a for loop here and the code would look very similar. I doubt this matters at all when checking a single entry, but it could add up for someone who has 100's of thousands of entries.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
```diff
@@ -77,7 +77,6 @@ public void CreateFileInFolderTest()
             this.Enlistment.WaitForBackgroundOperations().ShouldEqual(true, "Background operations failed to complete.");
 
             GVFSHelpers.ModifiedPathsShouldContain(this.fileSystem, this.Enlistment.DotGVFSRoot, folderName + "/");
-            GVFSHelpers.ModifiedPathsShouldContain(this.fileSystem, this.Enlistment.DotGVFSRoot, folderName + "/" + fileName);
```
sanoursa: Add the reverse check, that it should not contain that path?

PR #317, opened 2018-09-28
Design review: always-hydrate feature

sanoursa:
The goal of this PR is to review the always-hydrate design doc. Once this PR is wrapped up, I'll copy the contents of the md file to an issue and discard this PR.

Feel free to peruse or ignore the code changes. None of it has been fleshed out or tested. I wrote just enough code to validate that I can inject the logic into the correct places to satisfy the design.

### DIFF
--- a/GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
@@ -18,7 +18,8 @@ namespace GVFS.Virtualization.FileSystem
         protected static readonly GitCommandLineParser.Verbs CanCreatePlaceholderVerbs =
             GitCommandLineParser.Verbs.AddOrStage | GitCommandLineParser.Verbs.Move | GitCommandLineParser.Verbs.Status;
 
-        private BlockingCollection<FileOrNetworkRequest> fileAndNetworkRequests;
+        private BlockingCollection<FileOrNetworkRequest> highPriorityQueue;
+        private BlockingCollection<FileOrNetworkRequest> lowPriorityQueue;
         private Thread[] fileAndNetworkWorkerThreads;
         private int numWorkerThreads;
 
@@ -36,7 +37,9 @@ namespace GVFS.Virtualization.FileSystem
 
             this.Context = context;
             this.GitObjects = gvfsGitObjects;
-            this.fileAndNetworkRequests = new BlockingCollection<FileOrNetworkRequest>();
+
+            this.highPriorityQueue = new BlockingCollection<FileOrNetworkRequest>();
+            this.lowPriorityQueue = new BlockingCollection<FileOrNetworkRequest>();
 
             this.numWorkerThreads = numWorkerThreads;
         }
@@ -82,7 +85,9 @@ namespace GVFS.Virtualization.FileSystem
 
         public void PrepareToStop()
         {
-            this.fileAndNetworkRequests.CompleteAdding();
+            this.highPriorityQueue.CompleteAdding();
+            this.lowPriorityQueue.CompleteAdding();
+
             foreach (Thread t in this.fileAndNetworkWorkerThreads)
             {
                 t.Join();
@@ -112,10 +117,16 @@ namespace GVFS.Virtualization.FileSystem
 
         public void Dispose()
         {
-            if (this.fileAndNetworkRequests != null)
+            if (this.highPriorityQueue != null)
             {
-                this.fileAndNetworkRequests.Dispose();
-                this.fileAndNetworkRequests = null;
+                this.highPriorityQueue.Dispose();
+                this.highPriorityQueue = null;
+            }
+
+            if (this.lowPriorityQueue != null)
+            {
+                this.lowPriorityQueue.Dispose();
+                this.lowPriorityQueue = null;
             }
         }
 
@@ -307,13 +318,31 @@ namespace GVFS.Virtualization.FileSystem
             return metadata;
         }
 
-        protected bool TryScheduleFileOrNetworkRequest(FileOrNetworkRequest request, out Exception exception)
+        protected bool TryScheduleHighPriorityRequest(FileOrNetworkRequest request, out Exception exception)
         {
             exception = null;
 
             try
             {
-                this.fileAndNetworkRequests.Add(request);
+                this.highPriorityQueue.Add(request);
+                return true;
+            }
+            catch (InvalidOperationException e)
+            {
+                // Attempted to call Add after CompleteAdding has been called
+                exception = e;
+            }
+
+            return false;
+        }
+
+        protected bool TryScheduleLowPriorityRequest(FileOrNetworkRequest request, out Exception exception)
+        {
+            exception = null;
+
+            try
+            {
+                this.lowPriorityQueue.Add(request);
                 return true;
             }
             catch (InvalidOperationException e)
@@ -347,10 +376,8 @@ namespace GVFS.Virtualization.FileSystem
         {
             try
             {
-                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.FileSystemCallbacks.BlobSizes.CreateConnection())
-                {
-                    FileOrNetworkRequest request;
-                    while (this.fileAndNetworkRequests.TryTake(out request, Timeout.Infinite))
+                Action<BlobSizes.BlobSizesConnection, FileOrNetworkRequest> processRequest =
+                    (blobSizesConnection, request) =>
                     {
                         try
                         {
@@ -370,6 +397,28 @@ namespace GVFS.Virtualization.FileSystem
                         {
                             EventMetadata metadata = this.CreateEventMetadata(relativePath: null, exception: e);
                             this.LogUnhandledExceptionAndExit($"{nameof(this.ExecuteFileOrNetworkRequest)}_Cleanup", metadata);
+                        }
+                    };
+
+                BlockingCollection<FileOrNetworkRequest>[] anyQueue =
+                    new[]
+                    {
+                        this.highPriorityQueue,
+                        this.lowPriorityQueue,
+                    };
+
+                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.FileSystemCallbacks.BlobSizes.CreateConnection())
+                {
+                    FileOrNetworkRequest request;
+                    while (BlockingCollection<FileOrNetworkRequest>.TryTakeFromAny(anyQueue, out request, Timeout.Infinite) != -1)
+                    {
+                        processRequest(blobSizesConnection, request);
+
+                        // Make sure to drain all high priority requests
+                        // before checking to see if there are any low priority ones
+                        while (this.highPriorityQueue.TryTake(out request, TimeSpan.Zero))
+                        {
+                            processRequest(blobSizesConnection, request);
                         }
                     }
                 }
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1194,7 +1194,7 @@ namespace GVFS.Virtualization.Projection
             }
 
             // SNTODO: now that we've updated all placeholder files, go through the always-hydrate list 
-            // and schedule a low priority request to hydrate all matching paths
+            // and schedule a low priority download+hydrate request for every matching path
         }
 
         private void ProcessListOnThreads<T>(
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1192,6 +1192,9 @@ namespace GVFS.Virtualization.Projection
                 TimeSpan duration = activity.Stop(null);
                 this.context.Repository.GVFSLock.Stats.RecordUpdatePlaceholders((long)duration.TotalMilliseconds);
             }
+
+            // SNTODO: now that we've updated all placeholder files, go through the always-hydrate list 
+            // and schedule a low priority request to hydrate all matching paths
         }
 
         private void ProcessListOnThreads<T>(
--- a/GVFS/GVFS/CommandLine/PrefetchVerb.cs
+++ b/GVFS/GVFS/CommandLine/PrefetchVerb.cs
@@ -54,6 +54,13 @@ namespace GVFS.CommandLine
         public bool HydrateFiles { get; set; }
 
         [Option(
+            "always-hydrate",
+            Required = false,
+            Default = false,
+            HelpText = "Specify this flag to indicate that these files should always be hydrated, even after a projection change")]
+        public bool AlwaysHydrate { get; set; }
+
+        [Option(
             'c',
             "commits",
             Required = false,
@@ -312,6 +319,8 @@ namespace GVFS.CommandLine
                     : "Fetching blobs ";
                 this.ShowStatusWhileRunning(doPrefetch, message + this.GetCacheServerDisplay(cacheServer, enlistment.RepoUrl));
             }
+
+            // SNTODO: if AlwaysHydrate is true, also add the Files and Folders params to the always-exclude file
 
             if (blobPrefetcher.HasFailures)
             {
--- a/always-hydrate.md
+++ b/always-hydrate.md
@@ -26,13 +26,13 @@ the UX to the existing `gvfs prefetch` verb.
 
 Side note on the existing prefetch verb: it's really two different commands in one, and it's
 unfortunate that they are combined. The command `gvfs prefetch --commits` is entirely different in
-its goals, usage, and implementation than either of `gvfs prefetch --files` or `gvfs prefetch -
-folders` (and the latter two can be combined with each other, but not with `--commits`). Ideally we
+its goals, usage, and implementation than either of `gvfs prefetch --files` or `gvfs prefetch 
+--folders` (and the latter two can be combined with each other, but not with `--commits`). Ideally we
 would have had `gvfs prefetch-commits` and `gvfs prefetch-files` as two separate commands, which
 would have then led to a cleaner design for this new addition. I'm open to other suggestions here
 if anyone sees a better pattern for extending the current verbs.
 
-###Proposed UX for always-hydrate commands:
+### Proposed UX for always-hydrate commands:
 Adding new entries: 
 
 Add an `--always-hydrate` flag to the existing `--files` and `--folders` variants of `gvfs prefetch`.
@@ -50,8 +50,8 @@ Syntax:
       [--[always-]hydrate]
 ```
 
-The only change from today's syntax is the optional `[always-]` on the hydrate flag. `--always
-hydrate` implies `--hydrate`.
+The only change from today's syntax is the optional `[always-]` on the hydrate flag. 
+`--always-hydrate` implies `--hydrate`.
 
 Passing in the `--always-hydrate` flag has all the same behaviors that `prefetch --hydrate` has today
 and also adds the specified files and folders to the `always-hydrate` file, so that those files will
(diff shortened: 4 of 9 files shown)


always-hydrate.md
```diff
@@ -0,0 +1,130 @@
...
+
+```
+    gvfs prefetch
+      [<Repo Root Path>]
+      [--files <semicolon-delimited list of file paths, prefix * wildcards allowed>] 
+      [--folders <semicolon-delimited list of folder paths, no wildcards>]
+      --remove-always-hydrate
+```
+
+Any entries that match the specified `--files` and `--folders` arguments will be removed from the
```
sanoursa: Note: `--remove-always-hydrate` is incompatible with `--[always-]hydrate` or `--list-always-hydrate`

always-hydrate.md
```diff
@@ -0,0 +1,130 @@
...
+always-hydrate file. To remove all entries, the user can specify `gvfs prefetch --remove-always
```
sanoursa: Missed a `-` here in the `--remove-always-hydrate` flag

always-hydrate.md
```diff
@@ -0,0 +1,130 @@
...
+
+Side note on the existing prefetch verb: it's really two different commands in one, and it's
+unfortunate that they are combined. The command `gvfs prefetch --commits` is entirely different in
+its goals, usage, and implementation than either of `gvfs prefetch --files` or `gvfs prefetch -
+folders` (and the latter two can be combined with each other, but not with `--commits`). Ideally we
+would have had `gvfs prefetch-commits` and `gvfs prefetch-files` as two separate commands, which
+would have then led to a cleaner design for this new addition. I'm open to other suggestions here
+if anyone sees a better pattern for extending the current verbs.
+
+###Proposed UX for always-hydrate commands:
```
kyle-rader: title `###` spacing

always-hydrate.md
```diff
@@ -0,0 +1,130 @@
...
+    gvfs prefetch 
+      [<Repo Root Path>]
+      [--files <semicolon-delimited list of file paths, prefix * wildcards allowed>] 
+      [--folders <semicolon-delimited list of folder paths, no wildcards>]
+      [--folders-list <path of file containing newline-delimited paths of folders, no wildcards>] 
+      [--[always-]hydrate]
+```
+
+The only change from today's syntax is the optional `[always-]` on the hydrate flag. `--always
+hydrate` implies `--hydrate`.
```
kyle-rader: `--always-hyrdrate` space vs `-`

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -313,6 +320,8 @@ private void PrefetchBlobs(ITracer tracer, GVFSEnlistment enlistment, GitObjects
                 this.ShowStatusWhileRunning(doPrefetch, message + this.GetCacheServerDisplay(cacheServer, enlistment.RepoUrl));
             }
 
+            // SNTODO: if AlwaysHydrate is true, also add the Files and Folders params to the always-exclude file
```
sanoursa: to the _always-hydrate_ file. Gotta break that habit...

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+command, and we will need to remember if it was a file or a folder. Current plan is to go with a
+`FileBasedDictionary` unless we find the need to store more fields.
+
+### Hydrating files after a projection change
+Whenever the projection changes, the `GitIndexProjection` class currently kicks in and updates all
+placeholder files on disk with their new blob ids (for any files that changed). This is all done
+synchronously and blocks the git command that caused the projection change.
+
+In addition, this code will now also schedule background downloads of all paths that match the
+`always-hydrate` collection. We will just pass the list of file and folder parameters to the existing
```
wilbaker: > of all paths that match the `always-hydrate` collection

As mentioned in the code changes, limit this list to the placeholder that needed to be updated?

wilbaker: > We will just pass the list of file and folder parameters

typo: parameters -> placeholders?

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+      [--folders-list <path of file containing newline-delimited paths of folders, no wildcards>] 
+      [--[always-]hydrate]
+```
+
+The only change from today's syntax is the optional `[always-]` on the hydrate flag. 
+`--always-hydrate` implies `--hydrate`.
+
+Passing in the `--always-hydrate` flag has all the same behaviors that `prefetch --hydrate` has today
+and also adds the specified files and folders to the `always-hydrate` file, so that those files will
+continue to be hydrated in the future.
```
wilbaker: Possible UX feature:

`--always-hydrate` could use `git ls-files` to check how many files (approximately) will by hydrated (before staring any of the prefetch work).

We could also consider warning the user if the number is too large (to prevent them from unintentionally hydrating their entire repo).

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+command should be saved and reapplied after every projection change
+
+Syntax:
+
+```
+    gvfs prefetch 
+      [<Repo Root Path>]
+      [--files <semicolon-delimited list of file paths, prefix * wildcards allowed>] 
+      [--folders <semicolon-delimited list of folder paths, no wildcards>]
+      [--folders-list <path of file containing newline-delimited paths of folders, no wildcards>]
```
wilbaker: Will we save the location of`--folders-list` and re-read it each time?  Or will it be imported at the time `gvfs prefetch` is run (and not read again unless prefetch is called with `--folders-list` again)?

always-hydrate.md
sanoursa: What is too large? We actually currently allow a user to prefetch and hydrate all with `--files * --hydrate`, and I don't see a strong argument against `--files * --always-hydrate`. In fact, some of the folks in Windows and Office who routinely build the entire repo will probably want exactly that behavior.

The one protection that we currently have in place, and was technically a breaking change when it went in because people kept doing it on accident, is that running `gvfs prefetch` would download all blobs and `gvfs prefetch --hydrate` would hydrate all files. Now they fail (not warn!) and force you to specify `--files *` to indicate that you really mean it. We'll definitely extend that same protection to `--always-hydrate`. Beyond that, it's harder for me to see what would be a reasonable limit.

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+The only change from today's syntax is the optional `[always-]` on the hydrate flag. 
+`--always-hydrate` implies `--hydrate`.
+
+Passing in the `--always-hydrate` flag has all the same behaviors that `prefetch --hydrate` has today
+and also adds the specified files and folders to the `always-hydrate` file, so that those files will
+continue to be hydrated in the future.
+
+Managing existing entries:
+
+In addition to adding entries to the `always-hydrate` file, users will also want to list out the
```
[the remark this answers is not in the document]
wilbaker: > I think that's exactly what we're doing :-)

I can see it both ways :)

My thinking had been along these lines:

- `prefetch` -> synchronous VFS for Git command that will download objects and\or hydrate files.

- `always-hydrate` -> VFS for Git feature that enables more aggressive downloading of objects and hydrating of files.

But maybe that's over thinking it.  Users need to run `gvfs mount` to enable auto-mount, and that's similar to the experience here.

always-hydrate.md
wilbaker: That makes sense, thanks!

always-hydrate.md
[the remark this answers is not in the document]
sanoursa: To answer your question though - no I don't think `gvfs config` would be an appropriate place to add this because that verb is design to capture machine-wide config, not repo-specific. E.g. what upgrade ring you want and whether to automount.

I do think that the `--always-hydrate` flag makes good sense on the prefetch verb. It's actually doing the prefetching and hydrating, but _also_ recording that we want to keep prefetching and hydrating these paths later. 

The other two flags for listing and removing entries definitely feel strange here, and maybe that's what you meant. Those two variants don't do any prefetching, rather they just read or remove from the `always-hydrate` file. I'm open to ideas on where they should go, but overall my feeling was that it makes more sense for them to be on the same verb as the `--always-hydrate` flag, and that one makes the most sense on the existing `prefetch` verb.

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+its goals, usage, and implementation than either of `gvfs prefetch --files` or `gvfs prefetch 
+--folders` (and the latter two can be combined with each other, but not with `--commits`). Ideally we
+would have had `gvfs prefetch-commits` and `gvfs prefetch-files` as two separate commands, which
+would have then led to a cleaner design for this new addition. I'm open to other suggestions here
+if anyone sees a better pattern for extending the current verbs.
+
+### Proposed UX for always-hydrate commands:
+Adding new entries: 
+
+Add an `--always-hydrate` flag to the existing `--files` and `--folders` variants of `gvfs prefetch`.
```
sanoursa: Clarify: this doesn't just save the config, it also executes the same prefetch and hydrate here and now, as it would today.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1192,6 +1192,9 @@ private void UpdatePlaceholders()
                 TimeSpan duration = activity.Stop(null);
                 this.context.Repository.GVFSLock.Stats.RecordUpdatePlaceholders((long)duration.TotalMilliseconds);
             }
+
+            // SNTODO: now that we've updated all placeholder files, go through the always-hydrate list 
+            // and schedule a low priority download+hydrate request for every matching path
```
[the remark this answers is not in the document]
wilbaker: After discussing offline, I missed that we might also need to hydrate files that are still virtual (e.g. moving to a branch that has new files).

always-hydrate.md
sanoursa: I was thinking import, not re-read.

always-hydrate.md
sanoursa: And yea as we discussed, this could end up hydrating files that don't currently exist as placeholders, so we do need to start from the index and apply our filters on that. Luckily we already have it all parsed and in memory, so we should be able to just use that.

always-hydrate.md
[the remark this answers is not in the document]
kewillford: I think it would be helpful to users if we follow what git has done with the excludes and have a machine-wide and repo specific always-hydrate.  Most users I think would want to set the machine-wide list and then just have it work, whether I clone, prefetch, checkout, etc.  If I am creating a new clone, having to find the right repo, get the list of files or folders, copy it, clone, prefetch with the list seems like a lot of steps.

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+Today, a user can partially communicate this desire with the `gvfs prefetch --files` or `gvfs
+prefetch --folders` commands to bulk download the files they care about. And they can further specify
+the `--hydrate` flag to either command to bulk hydrate those files into the working directory.
+
+However, this is not a great experience because it's just a one-time prefetch and hydrate. What that
+means is that the next time the projection changes, e.g. after a `git checkout`, anywhere from some
+to all of those files can become dehydrated, and the user will again pay the cost of file-by-file
+downloads while trying to build.
+
+With this feature, we want the developer to make one gesture and communicate that a set of files
```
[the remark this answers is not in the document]
nickgra: I can see this being helpful, but this would conflict with the way Office plans to consume this feature and would be problematic for build lab scenarios. It would need to have an opt-out flag at least (although I think I would prefer opt-in).

always-hydrate.md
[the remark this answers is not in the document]
sanoursa: I don't agree that most users will want that functionality. And I don't think we envision that users will have to do much of this manually. The Office team is building up their own tool that analyzes their build dependencies, figures out what components the dev will need, and prefetch those files. We're giving them the ability to additionally indicate that they don't want to just prefetch once, but also keep this files/folders hydrated in the future.

That being said, the ability to configure some machine-wide defaults would be an easy extension of this design. I think we could easily add it in later on, but it's not something that I think we should include in an initial release. Also keep in mind that we have no other such pattern of machine-wide and repo-local config settings for VFS for Git, and doing so adds lots of complexities that I'd rather avoid if customers aren't actually asking for it.

always-hydrate.md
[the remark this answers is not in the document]
sanoursa: See the other thread on this topic, I don't think most users, at least not in Office, will want the behavior you're describing.

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+```
+    gvfs prefetch 
+      [<Repo Root Path>]
+      [--files <semicolon-delimited list of file paths, prefix * wildcards allowed>] 
+      [--folders <semicolon-delimited list of folder paths, no wildcards>]
+      [--folders-list <path of file containing newline-delimited paths of folders, no wildcards>] 
+      [--[always-]hydrate]
+```
+
+The only change from today's syntax is the optional `[always-]` on the hydrate flag.
```
jrbriggs: This implies that we support a mixed set of folders and files in the `always-hydrate` set. Is that the correct level of granularity? Should we start with only folders, since it seems quite fiddly ask a user to enumerate interesting files for us? Is there a scenario I'm not considering?

sanoursa: The scenario where it's useful is `--files *.csproj`, or any other build definition files

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+      [<Repo Root Path>]
+      [--files <semicolon-delimited list of file paths, prefix * wildcards allowed>] 
+      [--folders <semicolon-delimited list of folder paths, no wildcards>]
+      --remove-always-hydrate
+```
+
+The `--remove-always-hydrate` flag is incompatible with any combination of `--[always-]hydrate` or
+`--list-always-hydrate`.
+
+Any entries that match the specified `--files` and `--folders` arguments will be removed from the
```
jrbriggs: This doesn't feel right to me yet. Imagine I want three folders always-hydrated:

```
gvfs prefetch --folders a;b;c --always-hydrate
```

But then realize I wanted d, not c.
```
gvfs prefetch --folders c --remove-always-hydrate
gvfs prefetch --folders d --always-hydrate
```
It's weird to go through a "prefetch" verb to remove something from the set of items to prefetch. Should we consider allowing the user to directly edit the `always-hydrate` file like a `.gitignore` (but with a vastly simpler syntax)?

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+      --list-always-hydrate
+```
+
+The `--list-always-hydrate` flag is incompatible with any of the other flags. It will read out the
+`always-hydrate` file and display the entries to the user.
+
+```
+    gvfs prefetch
+      [<Repo Root Path>]
+      [--files <semicolon-delimited list of file paths, prefix * wildcards allowed>]
```
jrbriggs: Should we fix this while we're in here, since semicolons are perfectly valid in paths?

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+--remove-always-hydrate --files *`.
+
+## Implementation
+### Storing always-hydrate entries
+We will create a new file in the `.gvfs` folder called `always-hydrate`. This file is technically
+human readable and editable, but like all files in the `.gvfs` folder, carries no contract around
+maintaining the same format moving forward. All user interactions with this file should go through
+the `gvfs prefetch` command.
+
+Therefore the specific format of the file is not that interesting. For the sake of easy
```
jrbriggs: I think we should consider guaranteeing a very simple versioned file format. The obvious con is not being able to break it. But in the pro column, we wouldn't need to provide gestures to interact with it beyond a text editor. This helps clear up the awkward `gvfs prefetch --folders x --remove-always-hydrate`

Then tie everything to the file. To turn it off, delete the file. To remove something, edit the file. You can add something by following the pattern. Onboarding still makes sense with a `gvfs prefetch --folders x --always-hydrate` but the flag is an alternate way to say "create this file for me."

always-hydrate.md
```diff
@@ -0,0 +1,133 @@
...
+
+In addition, this code will now also schedule background downloads of all paths that match the
+`always-hydrate` collection. We will just pass the list of file and folder parameters to the existing
+`BlobPrefetcher` that the `PrefetchVerb` uses to download and hydrate files today. 
+
+That class will determine which blobs need to be downloaded and which files need to be hydrated.
+However, the existing functionality of that class tries to spin up several threads and do bulk
+discovery, download, and hydration of files. Since we want these to be low-priority downloads (see 
+next section), we will instead just take the resulting list of blobs and paths, and schedule 
+individual requests for each one. (Note: some experimentation will be needed here. It may be more
```
jrbriggs: Let's think about this differently -- it will definitely be batched, but we can start with a batch size of 1. This makes it easier to experiment with later.

always-hydrate.md
sanoursa: Totally agree that the current syntax works for adding, but is very awkward for listing/deleting. Let's first find agreement on whether the always-hydrate file format should be open or closed, and then come back to this.

always-hydrate.md
sanoursa: That was my initial thought as well. However, I really hesitate to go with this approach. Right now, the entire .gvfs folder is technically opaque to users, and this gives us a lot of freedom as we keep adding more features, making things faster, etc. If we open up this one file, not only do we have to guarantee this file's format moving forward, we also risk people not realizing that other files are off limits, and then we can get locked into current formats. Even worse, some of these files are human readable/editable, but will break your repo's functionality if you modify them incorrectly (e.g. the list of modified paths). So I really don't want to open up that rabbit hole for this feature, which is why I settled on saying that the CLI is the only supported way to interact with this feature.

I suppose we could store this file in a location other than .gvfs, but I don't really know where that location would be. In the clone root? In the .git folder? Somewhere else?

The other alternative is to have a new `gvfs always-hydrate` verb. That has its own set of issues, but it may result in a cleaner syntax.

Thoughts?

always-hydrate.md
sanoursa: It would be a breaking change, so I think we should tackle it separately as it's not really related to any of this functionality

always-hydrate.md
sanoursa: Agreed

always-hydrate.md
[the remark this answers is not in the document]
kewillford: What if we use the gvfs section of the git config to store this information?  That would give us the machine-wide vs repo setting for free and we wouldn't have another file to manage.  Listing, adding, deleting would all be taken care of as well.  I don't know how many files and folders we plan on saving here so that could be a limitation?

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -307,13 +318,31 @@ protected void OnHardLinkCreated(string relativeExistingFilePath, string relativ
             return metadata;
         }
 
-        protected bool TryScheduleFileOrNetworkRequest(FileOrNetworkRequest request, out Exception exception)
+        protected bool TryScheduleHighPriorityRequest(FileOrNetworkRequest request, out Exception exception)
```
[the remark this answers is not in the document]
sanoursa: This is all throw-away code, just focusing on design and UX right now

always-hydrate.md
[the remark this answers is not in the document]
sanoursa: You might have a folder named Foo in two completely unrelated repos, and you wouldn't want your global config affecting them both. So we'd have to tie the config to a URL at least, though even that is unreliable.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -36,7 +37,9 @@ protected FileSystemVirtualizer(GVFSContext context, GVFSGitObjects gvfsGitObjec
 
             this.Context = context;
             this.GitObjects = gvfsGitObjects;
-            this.fileAndNetworkRequests = new BlockingCollection<FileOrNetworkRequest>();
+
+            this.highPriorityQueue = new BlockingCollection<FileOrNetworkRequest>();
+            this.lowPriorityQueue = new BlockingCollection<FileOrNetworkRequest>();
```
alameenshah: nit - About the queue names. lowPriorityQueue and highPriorityQueue sounds generic, it hints at the existence of low and high priority operations outside the context of always-hydrate. Naming it more specific, like alwaysHydrateQueue might be more exact.

sanoursa: I disagree, this is actually unrelated to the always-hydrated feature per se. We have other reasons we may want to use the low pri queue, e.g. predictive prefetching. The point here is that this class only knows that some download requests are more important than other download requests.

(That being said, this is just a review of the doc. All of the code changes are throw away, so we'll come back to a real code review later on)

sanoursa:
After further discussion with the Office team, we're actually going to park this whole feature request for now as it may not even be needed.

What we will look at instead is making a noop `gvfs prefetch --files --folders --hydrate` much faster than it is today

PR #322, opened 2018-10-02
Mac kext: Virtualisation root refactoring & array resizing

pmj:
The way the `VirtualizationRoot` struct has been used in the ProjFS kext has been rather messy so far - sometimes by index, sometimes by pointer, often non-threadsafe (although the risks have been fairly small).

The ultimate goals of this PR are:
 * Lift the limit on number of virtualisation roots.
 * Fix the thread safety issues.

The reason I made the virtualisation root array a fixed size initially was because allowing resizes means the pointers will become invalid unless you introduce some fairly strict locking requirements. Locking the root array for near enough the entire duration of a vnode/fileop callback could cause some significant contention though and would furthermore raise issues with performing vnode I/O and acquiring other locks while holding the vroot lock.

This PR approaches the issue a different way: consistently only expose vroot indices in the API, never pointers to the struct itself, and provide accessor functions which ensure thread safety as needed. As long as indices stay linked to root identities, it's safe to drop the lock while holding a root index.

To get there, I made a bunch of other smaller preparatory and in some cases only tangentially related changes; **all the changes are in logically distinct git commits, so when reviewing you may wish to review one commit at a time** - you'll have an easier ride that way.

Quick commit summary:

1. <s>*Don't assert/KP on attribute reading failure.* The assertion on getting vnode attributes successfully failed a few times during development & testing. It's not entirely clear why yet, but the assert doesn't give much clue as to the reason, so I've changed this to a soft failure with some extra logging. If you encounter instances of this error while testing etc., please do send me --reports.</s>
2. *Handle named streams.* Relatedly, I've logged I/O errors on a bunch of other operations, and quite a number of them originated in vnodes corresponding to resource forks. (You'll recognise these by their path, which looks something like `/path/to/a/regular/file/..namedfork/rsrc`.) These aren't standalone files, but alternate data streams/forks on a regular file. For the most part, they are uninteresting to VFS for Git, but they behave oddly when querying things like xattrs, and we really should be applying the logic according to their underlying file. So this changes the vnode op callback handler to operate on that underlying file when encountering a named stream.
3. *Replaces some C-style casts with C++ style.* We've been fairly consistent with using C++ style casts (`static_cast`, `const_cast`, `reinterpret_cast`) in the kext, but I stumbled across some C-style casts that crept in. This changes them to C++ style.
4. *Enables warnings for shadowed variables.* Activates another compiler warning - shadowing variables is rarely what we want, and more frequently is a bug. (I made just such a mistake while refactoring the `VirtualizationRoot` API and spent half an hour chasing it, if you must know.)
5. *Adds typed array memory allocation function*. This adds a helper function that wraps our memory allocator with a typed array-based one, which make for slightly more readable and less error prone code. (Particularly with regard to overflow on size calculations.)
6. *Use only `VirtualizationRoot` indices, not direct pointers.* This rejigs the `VirtualizationRoot.hpp` API to never leak pointers, and indeed moves the struct definition to the `.cpp` file. It also adds accessor functions for situations where previously pointers were dereferenced, and wraps them with locks so access is threadsafe.
7. *Dynamically alloc virtualisation root array & resize when full.* This finally gets rid of the nasty assertion failure when our virtualisation root array is full and grows it instead!

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -42,11 +42,9 @@ static inline bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
 
 static bool IsFileSystemCrawler(char* procname);
 
-static const char* GetRelativePath(const char* path, const char* root);
-
 static void Sleep(int seconds, void* channel);
 static bool TrySendRequestAndWaitForResponse(
-    const VirtualizationRoot* root,
+    VirtualizationRootHandle root,
     MessageType messageType,
     const vnode_t vnode,
     const FsidInode& vnodeFsidInode,
@@ -65,7 +63,7 @@ static bool ShouldHandleVnodeOpEvent(
     kauth_action_t action,
 
     // Out params:
-    VirtualizationRoot** root,
+    VirtualizationRootHandle* root,
     vtype* vnodeType,
     uint32_t* vnodeFileFlags,
     FsidInode* vnodeFsidInode,
@@ -80,7 +78,7 @@ static bool ShouldHandleFileOpEvent(
     kauth_action_t action,
 
     // Out params:
-    VirtualizationRoot** root,
+    VirtualizationRootHandle* root,
     FsidInode* vnodeFsidInode,
     int* pid);
 
@@ -277,7 +275,7 @@ static int HandleVnodeOperation(
     char vnodePathBuffer[PrjFSMaxPath];
     int vnodePathLength = PrjFSMaxPath;
 
-    VirtualizationRoot* root = nullptr;
+    VirtualizationRootHandle root = RootHandle_None;
     vtype vnodeType;
     uint32_t currentVnodeFileFlags;
     FsidInode vnodeFsidInode;
@@ -435,7 +433,7 @@ static int HandleFileOpOperation(
             goto CleanupAndReturn;
         }
         
-        VirtualizationRoot* root = nullptr;
+        VirtualizationRootHandle root = RootHandle_None;
         FsidInode vnodeFsidInode;
         int pid;
         if (!ShouldHandleFileOpEvent(
@@ -495,7 +493,7 @@ static int HandleFileOpOperation(
             goto CleanupAndReturn;
         }
             
-        VirtualizationRoot* root = nullptr;
+        VirtualizationRootHandle root = RootHandle_None;
         FsidInode vnodeFsidInode;
         int pid;
         if (!ShouldHandleFileOpEvent(
@@ -571,7 +569,7 @@ static bool ShouldHandleVnodeOpEvent(
     kauth_action_t action,
 
     // Out params:
-    VirtualizationRoot** root,
+    VirtualizationRootHandle* root,
     vtype* vnodeType,
     uint32_t* vnodeFileFlags,
     FsidInode* vnodeFsidInode,
@@ -579,8 +577,8 @@ static bool ShouldHandleVnodeOpEvent(
     char procname[MAXCOMLEN + 1],
     int* kauthResult)
 {
-    *root = nullptr;
     *kauthResult = KAUTH_RESULT_DEFER;
+    *root = RootHandle_None;
     
     if (!VirtualizationRoot_VnodeIsOnAllowedFilesystem(vnode))
     {
@@ -626,19 +624,22 @@ static bool ShouldHandleVnodeOpEvent(
     }
     
     *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context);
-    *root = VirtualizationRoots_FindForVnode(vnode, *vnodeFsidInode);
+    *root = VirtualizationRoot_FindForVnode(vnode, *vnodeFsidInode);
     
-    if (nullptr == *root)
+    if (RootHandle_ProviderTemporaryDirectory == *root)
+    {
+        *kauthResult = KAUTH_RESULT_DEFER;
+        return false;
+    }
+    else if (RootHandle_None == *root)
     {
         KextLog_FileNote(vnode, "No virtualization root found for file with set flag.");
         
         *kauthResult = KAUTH_RESULT_DEFER;
         return false;
     }
-    else if (nullptr == (*root)->providerUserClient)
+    else if (!VirtualizationRoot_IsOnline(*root))
     {
-        // There is no registered provider for this root
-        
         // TODO(Mac): Protect files in the worktree from modification (and prevent
         // the creation of new files) when the provider is offline
         
@@ -647,7 +648,7 @@ static bool ShouldHandleVnodeOpEvent(
     }
     
     // If the calling process is the provider, we must exit right away to avoid deadlocks
-    if (*pid == (*root)->providerPid)
+    if (VirtualizationRoot_PIDMatchesProvider(*root, *pid))
     {
         *kauthResult = KAUTH_RESULT_DEFER;
         return false;
@@ -663,10 +664,12 @@ static bool ShouldHandleFileOpEvent(
     kauth_action_t action,
 
     // Out params:
-    VirtualizationRoot** root,
+    VirtualizationRootHandle* root,
     FsidInode* vnodeFsidInode,
     int* pid)
 {
+    *root = RootHandle_None;
+
     vtype vnodeType = vnode_vtype(vnode);
     if (ShouldIgnoreVnodeType(vnodeType, vnode))
     {
@@ -674,21 +677,17 @@ static bool ShouldHandleFileOpEvent(
     }
     
     *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context);
-    *root = VirtualizationRoots_FindForVnode(vnode, *vnodeFsidInode);
-    if (nullptr == *root)
+    *root = VirtualizationRoot_FindForVnode(vnode, *vnodeFsidInode);
+    if (!VirtualizationRoot_IsValidRootHandle(*root))
     {
-        return false;
-    }
-    else if (nullptr == (*root)->providerUserClient)
-    {
-        // There is no registered provider for this root
+        // This VNode is not part of a root
         return false;
     }
     
-    // If the calling process is the provider, we must exit right away to avoid deadlocks
     *pid = GetPid(context);
-    if (*pid == (*root)->providerPid)
+    if (VirtualizationRoot_PIDMatchesProvider(*root, *pid))
     {
+        // If the calling process is the provider, we must exit right away to avoid deadlocks
         return false;
     }
     
@@ -696,7 +695,7 @@ static bool ShouldHandleFileOpEvent(
 }
 
 static bool TrySendRequestAndWaitForResponse(
-    const VirtualizationRoot* root,
+    VirtualizationRootHandle root,
     MessageType messageType,
     const vnode_t vnode,
     const FsidInode& vnodeFsidInode,
@@ -719,7 +718,7 @@ static bool TrySendRequestAndWaitForResponse(
         return false;
     }
     
-    const char* relativePath = GetRelativePath(vnodePath, root->path);
+    const char* relativePath = VirtualizationRoot_GetRootRelativePath(root, vnodePath);
     
     int nextMessageId = OSIncrementAtomic(&s_nextMessageId);
     
@@ -748,7 +747,7 @@ static bool TrySendRequestAndWaitForResponse(
     
     // TODO(Mac): Should we pass in the root directly, rather than root->index?
     //            The index seems more like a private implementation detail.
-    if (!isShuttingDown && 0 != ActiveProvider_SendMessage(root->index, messageSpec))
+    if (!isShuttingDown && 0 != ActiveProvider_SendMessage(root, messageSpec))
     {
         // TODO: appropriately handle unresponsive providers
         
@@ -883,19 +882,6 @@ static bool IsFileSystemCrawler(char* procname)
     }
     
     return false;
-}
-
-static const char* GetRelativePath(const char* path, const char* root)
-{
-    assert(strlen(path) >= strlen(root));
-    
-    const char* relativePath = path + strlen(root);
-    if (relativePath[0] == '/')
-    {
-        relativePath++;
-    }
-    
-    return relativePath;
 }
 
 static bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode)
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -425,7 +425,7 @@ static int HandleFileOpOperation(
         KAUTH_FILEOP_LINK == action)
     {
         // arg0 is the (const char *) fromPath (or the file being linked to)
-        const char* newPath = (const char*)arg1;
+        const char* newPath = reinterpret_cast<const char*>(arg1);
         
         // TODO(Mac): We need to handle failures to lookup the vnode.  If we fail to lookup the vnode
         // it's possible that we'll miss notifications
@@ -481,7 +481,7 @@ static int HandleFileOpOperation(
     else if (KAUTH_FILEOP_CLOSE == action)
     {
         vnode_t currentVnode = reinterpret_cast<vnode_t>(arg0);
-        const char* path = (const char*)arg1;
+        const char* path = reinterpret_cast<const char*>(arg1);
         int closeFlags = static_cast<int>(arg2);
         
         if (vnode_isdir(currentVnode))
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -260,6 +260,18 @@ static int HandleVnodeOperation(
     // arg2 is the (vnode_t) parent vnode
     int* kauthError =       reinterpret_cast<int*>(arg3);
     int kauthResult = KAUTH_RESULT_DEFER;
+    bool putVnodeWhenDone = false;
+
+    // A lot of our file checks such as attribute tests behave oddly if the vnode
+    // refers to a named fork/stream; apply the logic as if the vnode operation was
+    // occurring on the file itself. (/path/to/file/..namedfork/rsrc)
+    if (vnode_isnamedstream(currentVnode))
+    {
+        vnode_t mainFileFork = vnode_getparent(currentVnode);
+        assert(NULLVP != mainFileFork);
+        currentVnode = mainFileFork;
+        putVnodeWhenDone = true;
+    }
 
     const char* vnodePath = nullptr;
     char vnodePathBuffer[PrjFSMaxPath];
@@ -384,6 +396,11 @@ static int HandleVnodeOperation(
     }
     
 CleanupAndReturn:
+    if (putVnodeWhenDone)
+    {
+        vnode_put(currentVnode);
+    }
+    
     atomic_fetch_sub(&s_numActiveKauthEvents, 1);
     return kauthResult;
 }
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
@@ -1,5 +1,6 @@
 #include <kern/debug.h>
 #include <kern/assert.h>
+#include <stdatomic.h>
 
 #include "PrjFSCommon.h"
 #include "PrjFSXattrs.h"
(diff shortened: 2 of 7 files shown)


pmj:
I've been undecided whether to submit 1-5 and 6-7 as 2 separate PRs, so if review finds that 1-5 are uncontroversial while 6&7 need more work, I'd like to merge that first lot on its own rather than having to keep rebasing on master.

**Not** addressed by this PR, to be handled in a future change:
 * GC of dead roots. (e.g. if the user unmounts a root and deletes it with `rm -r`) We should recycle the array slot in this case, which is currently not done.
 * Linear searches on the root array. Searching by vnode or inode should be done via a hash table or similarly efficient data structure.

nickgra:
> Don't assert/KP on attribute reading failure. The assertion on getting vnode attributes successfully failed a few times during development & testing. It's not entirely clear why yet, but the assert doesn't give much clue as to the reason, so I've changed this to a soft failure with some extra logging. If you encounter instances of this error while testing etc., please do send me reports.

I've been encountering this all the time locally, I suspect the problem is that we're trying to operate on a vnode type that doesn't support us reading attributes. 
Since we're handling every fileop close event, we catch some events like this.  Pulling ShouldIgnoreVnodeType out from ShouldHandleFileOpEvent before FileIsFlaggedAsInRoot (the offending callsite according to the panic report) fixes the panics for me, but seems to have a larger performance hit than I would like. 

```
@@ -462,10 +462,16 @@ static int HandleFileOpOperation(
         if (vnode_isdir(currentVnode))
         {
             goto CleanupAndReturn;
         }
 
+        vtype vnodeType = vnode_vtype(currentVnode);
+        if (ShouldIgnoreVnodeType(vnodeType, currentVnode))
+        {
+        }
+        
         bool fileFlaggedInRoot = FileIsFlaggedAsInRoot(currentVnode, context);
         if (fileFlaggedInRoot && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
         {
             goto CleanupAndReturn;
         }
```
I haven't done any formal perf testing, but your solution seems like it'll be much cheaper if this actually fixes the panic. I'm happy to take your fix if you think it's safe with this context :)

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -610,16 +608,23 @@ static bool ShouldHandleVnodeOpEvent(
         }
     }
     
-    *root = VirtualizationRoots_FindForVnode(vnode);
+    *rootIndex = VirtualizationRoot_FindForVnode(vnode);
+    
+    // This VNode is in the temp directory, do not act on it.
```
nickgra: Is this leftover from when you were looking into the atomic placeholder/temp directory work?

pmj: Yeah, I was unsure whether to remove it - it's also in your atomic enumeration branch. I definitely wanted to pull in the enum with the "special" root indices and finally get rid of the -1s everywhere. It seemed silly to remove the temp directory enum value, and we're presumably going to need this condition here, even if it's not going to kick in in this version of the code. I can remove it though.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.hpp
```diff
@@ -3,31 +3,22 @@
...
+    RootIndex_Indeterminate              = -2,
+    // Vnode is not in a virtualization root, but below a provider's registered temp directory
+    RootIndex_ProviderTemporaryDirectory = -3,
 };
 
 kern_return_t VirtualizationRoots_Init(void);
 kern_return_t VirtualizationRoots_Cleanup(void);
 
-VirtualizationRoot* VirtualizationRoots_FindForVnode(vnode_t vnode);
+int16_t VirtualizationRoot_FindForVnode(vnode_t vnode);
```
sanoursa: Instead of passing around `int16_t`'s, does it make sense to typedef it to a `VirtualizationRootHandle`?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -256,12 +254,26 @@ static int HandleVnodeOperation(
...
     int* kauthError =       reinterpret_cast<int*>(arg3);
     int kauthResult = KAUTH_RESULT_DEFER;
+    bool putVnodeWhenDone = false;
+
+    // A lot of our file checks such as attribute tests behave oddly if the vnode
+    // refers to a named fork/stream; apply the logic as if the vnode operation was
+    // occurring on the file itself. (/path/to/file/..namedfork/rsrc)
+    if (vnode_isnamedstream(currentVnode))
+    {
+        vnode_t mainFileFork = vnode_getparent(currentVnode);
```
sanoursa: What should we do if this call fails and returns `NULLVP`? We now know that `currentVnode` is referring to a named stream and that we failed to resolve to the actual file, so should we give up?

sanoursa: Related: is there any way to guarantee that we can find the parent? The `vnode_getparent` function is racy since it only resolves vnodes that are still in the name cache. Is there an alternative?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -591,16 +608,23 @@ static bool ShouldHandleVnodeOpEvent(
         }
     }
     
-    *root = VirtualizationRoots_FindForVnode(vnode);
+    *rootIndex = VirtualizationRoot_FindForVnode(vnode);
+    
+    // This VNode is in the temp directory, do not act on it.
```
sanoursa: Nit: I don't think this comment says anything that the following code doesn't already tell us. I would expand the comment to explain why (not what), or just drop it.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -591,16 +608,23 @@ static bool ShouldHandleVnodeOpEvent(
...
+    if (RootIndex_ProviderTemporaryDirectory == *rootIndex)
+    {
+        *kauthResult = KAUTH_RESULT_DEFER;
+        return false;
+    }
     
-    if (nullptr == *root)
+    if (*rootIndex == RootIndex_None)
```
sanoursa: Nit: for consistency, put the value first and the variable second

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -628,38 +652,37 @@ static bool ShouldHandleFileOpEvent(
...
 
-    *root = VirtualizationRoots_FindForVnode(vnode);
-    if (nullptr == *root)
-    {
-        return false;
-    }
-    else if (nullptr == (*root)->providerUserClient)
+    *rootIndex = VirtualizationRoot_FindForVnode(vnode);
+
+    if (RootIndex_None >= *rootIndex)
```
sanoursa: This check seems pretty magical, and it really assumes knowledge of the fact that these are indexes (which I think we should keep private). Make this a helper function instead?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -12,18 +12,70 @@
...
+    if (rootIndex < 0)
+    {
+        return false;
+    }
+    
+    bool result;
+    RWLock_AcquireShared(s_rwLock);
+    {
+        result =
+        (rootIndex < s_maxVirtualizationRoots)
```
sanoursa: Indent

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -12,18 +12,70 @@
...
+    return result;
+}
+
+bool VirtualizationRoot_PIDMatchesProvider(int16_t rootIndex, pid_t pid)
+{
+    bool result;
+    RWLock_AcquireShared(s_rwLock);
+    {
+        result =
+        (rootIndex >= 0 && rootIndex < s_maxVirtualizationRoots)
```
sanoursa: Indent

pmj:
> Pulling ShouldIgnoreVnodeType out from ShouldHandleFileOpEvent before FileIsFlaggedAsInRoot (the offending callsite according to the panic report) fixes the panics for me, but seems to have a larger performance hit than I would like.

Hmm. `vnode_vtype` is just a pointer dereferencing/struct member read, and the switch block in `ShouldIgnoreVnodeType()` ought to compile down to something minimal, so I'd be extremely surprised if there was any significant overhead introduced by that extra check. I feel like we're getting sidetracked by the vnode I/O failure issue - I'll open a github issue to track the various different instances of the problem, because there do seem to be a few pieces to it.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: Sounds sensible, I've created a `VirtualizationRoot_IsValidRootHandle()` helper function instead.

pmj:
I've just pushed an update that fixes the formatting issues, and which changes the terminology to "handles" outside of VirtualizationRoots.cpp, including a typedef as suggested. I've kept the index terminology inside the .cpp for now as they are actually known to be indices there.

pmj:
vnode I/O errors now tracked as [issue #328](https://github.com/Microsoft/VFSForGit/issues/328).

wilbaker:
@pmj I'm reviewing these changes now, please hold off on merging until I've posted comments.  Thanks!

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -56,18 +119,17 @@ kern_return_t VirtualizationRoots_Cleanup()
...
-VirtualizationRoot* VirtualizationRoots_FindForVnode(vnode_t vnode)
+VirtualizationRootHandle VirtualizationRoot_FindForVnode(vnode_t vnode)
 {
-    VirtualizationRoot* root = nullptr;
+    VirtualizationRootHandle rootIndex = RootHandle_None;
     
     vnode_get(vnode);
     // Search up the tree until we hit a known virtualization root or THE root of the file system
-    while (nullptr == root && NULLVP != vnode && !vnode_isvroot(vnode))
+    while (rootIndex == RootHandle_None && NULLVP != vnode && !vnode_isvroot(vnode))
```
wilbaker: @sanoursa, we have a mix right now with our checks, sometimes it's `VALUE == var` and othertimes it's `var == VALUE`.  Did we decide on preferring one over the other, and what is the rule for `!=`?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -56,18 +119,17 @@ kern_return_t VirtualizationRoots_Cleanup()
...
     {
-        int16_t rootIndex = VirtualizationRoots_LookupVnode(vnode, nullptr);
-        if (rootIndex >= 0)
+        rootIndex = VirtualizationRoots_FindRootAtVnode(vnode, nullptr);
+        if (rootIndex != RootHandle_None)
```
wilbaker: Would it be safer to check `!VirtualizationRoot_IsValidRootHandle(rootIndex)`?

(Or is part of the contract with `VirtualizationRoots_FindRootAtVnode` that it will only return `RootHandle_None` or a valid index?)

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -112,9 +174,9 @@ int16_t VirtualizationRoots_LookupVnode(vnode_t vnode, vfs_context_t context)
             RWLock_AcquireExclusive(s_rwLock);
             {
                 // Vnode may already have been inserted as a root in the interim
-                rootIndex = FindRootForVnode_Locked(vnode, vid, fsidInode);
+                rootIndex = FindRootAtVnode_Locked(vnode, vid, fsidInode);
                 
-                if (rootIndex < 0)
+                if (rootIndex == RootHandle_None)
```
wilbaker: Similar comment\question on checking None vs < 0 here.  

When should we be checking against None and when should be be checking `VirtualizationRoot_IsValidRootHandle`?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -131,27 +193,58 @@ int16_t VirtualizationRoots_LookupVnode(vnode_t vnode, vfs_context_t context)
...
+    if (RootHandle_None == rootIndex)
+    {
+        // No space, resize array
+        uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
+        if (newLength <= s_maxVirtualizationRoots)
+        {
+            return RootHandle_None;
+        }
+        
+        VirtualizationRoot* grownArray = Memory_AllocArray<VirtualizationRoot>(newLength);
```
wilbaker: Check if `Memory_AllocArray` returned `nullptr`?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
wilbaker: (No need to clean this up in this PR, I'm just wondering what our style guideline is for this)

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
sanoursa: I'm happy to leave the bigger investigation for a separate change, but I think we should either assert or return a failure here if this call fails. Otherwise we know we will misbehave downstream of this code.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -803,6 +821,12 @@ static uint32_t ReadVNodeFileFlags(vnode_t vn, vfs_context_t context)
     struct vnode_attr attributes = {};
     errno_t err = GetVNodeAttributes(vn, context, &attributes);
     // TODO: May fail on some file system types? Perhaps we should early-out depending on mount point anyway.
+    if (0 != err)
+    {
+        KextLog_FileError(vn, "ReadVNodeFileFlags failed with error %d (0x%x)", err, err);
```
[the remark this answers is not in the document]
sanoursa: Yea I don't think we can just log and ignore. We need to either assert, or else change this function signature so we can return an error and fail the IO.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -12,17 +12,73 @@
...
+}
 
-static VirtualizationRoot s_virtualizationRoots[MaxVirtualizationRoots] = {};
+bool VirtualizationRoot_PIDMatchesProvider(VirtualizationRootHandle rootIndex, pid_t pid)
+{
+    bool result;
+    RWLock_AcquireShared(s_rwLock);
+    {
+        result =
+            (rootIndex >= 0 && rootIndex < s_maxVirtualizationRoots)
```
sanoursa: Style nit: I see 3 different parentheses styles in this one condition, and none of them are required :-). I personally favor: only use parens when they change the order of operations, otherwise they just add visual and mental clutter.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -37,9 +93,16 @@ kern_return_t VirtualizationRoots_Init()
         return KERN_FAILURE;
     }
     
-    for (uint32_t i = 0; i < MaxVirtualizationRoots; ++i)
+    s_maxVirtualizationRoots = 128;
+    s_virtualizationRoots = Memory_AllocArray<VirtualizationRoot>(s_maxVirtualizationRoots);
+    if (nullptr == s_virtualizationRoots)
+    {
+        return KERN_RESOURCE_SHORTAGE;
```
sanoursa: Everywhere else we return KERN_FAILURE if we fail to allocate memory. Should we update those to return this error code instead?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
sanoursa: I definitely prefer yoda-speak in C/C++ code to eliminate any risk of accidental assignments, but we have not enforced it consistently.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: OK, an assertion sounds the most sensible approach here.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.hpp
```diff
@@ -3,42 +3,39 @@
...
-errno_t ActiveProvider_SendMessage(int32_t rootIndex, const Message message);
+errno_t ActiveProvider_SendMessage(VirtualizationRootHandle rootHandle, const Message message);
 bool VirtualizationRoot_VnodeIsOnAllowedFilesystem(vnode_t vnode);
+bool VirtualizationRoot_IsOnline(VirtualizationRootHandle rootHandle);
+bool VirtualizationRoot_PIDMatchesProvider(VirtualizationRootHandle rootHandle, pid_t pid);
+bool VirtualizationRoot_IsValidRootHandle(VirtualizationRootHandle rootHandle);
+const char* VirtualizationRoot_GetRootRelativePath(VirtualizationRootHandle rootHandle, const char* path);
 
-int16_t VirtualizationRoots_LookupVnode(vnode_t vnode, vfs_context_t context);
+int16_t VirtualizationRoots_FindRootAtVnode(vnode_t vnode, vfs_context_t context);
```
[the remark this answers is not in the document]
pmj: Thinking about this some more, in addition to making it a static function, I'm renaming it to `FindOrDetectRootAtVnode` to distinguish it from `FindRootAtVnode_Locked`. The latter just checks against known roots, while this function subsequently checks for xattrs to see if it's a root that's previously not been detected yet.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -37,9 +93,16 @@ kern_return_t VirtualizationRoots_Init()
         return KERN_FAILURE;
     }
     
-    for (uint32_t i = 0; i < MaxVirtualizationRoots; ++i)
+    s_maxVirtualizationRoots = 128;
+    s_virtualizationRoots = Memory_AllocArray<VirtualizationRoot>(s_maxVirtualizationRoots);
```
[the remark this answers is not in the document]
pmj: There can't be any other threads accessing the data structures at this point yet, so a full lock is technically not needed, although you're right that other threads will need a memory barrier to guarantee propagation. How do you feel about an explicit `atomic_thread_fence(memory_order_seq_cst);` at the end of this init function?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: `VirtualizationRoots_FindRootAtVnode()` (via `FindRootAtVnode_Locked()`) [will eventually be able to return `RootHandle_ProviderTemporaryDirectory`](https://github.com/Microsoft/VFSForGit/pull/225/files#diff-86d4504941139fd962dc83a58c348753R205), in which case we've actually found what we're looking for and want to stop the search and return that special handle from this function. This is evidently not obvious, so I'll put in a comment to that effect.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Yeah, same answer as above - a result of `RootHandle_ProviderTemporaryDirectory` means we don't need to drop to file I/O level as we've found what we're looking for.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Given that the return value is only checked as a boolean, it doesn't really matter much I suppose. If I was designing it from scratch, I'd probably either pass the full return value back out from the kext start function, or simply return booleans from the various init functions. I don't have sufficiently strong feelings either way and can also change this to `KERN_FAILURE` if preferred.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
pmj: OK, I'm going to cull that patch from this PR. It's tangential to the core functionality, and too important to piggy-back on it.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Good catch! Added the check.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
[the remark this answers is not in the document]
wilbaker: >How do you feel about an explicit atomic_thread_fence(memory_order_seq_cst); at the end of this init function?

That sounds good to me, thanks!

PR #323, opened 2018-10-02
WIP: Windows: Use a faster comparison when enumeration filter strings don't have wildcards

wilbaker:
Update `ActiveEnumeration` to use `string.Equals` rather than `Utils.IsFileNameMatch` when the enumeration filter string does not have wildcards (first commit), and to exit early when the filter does not have wildcards (second commit).

Times with scenario provided by @kyle-rader 

  | Baseline Times (sec) | First commit (sec) | Second commit (sec)
-- | -- | -- | --
  | 48.23 | 9.83 | 8.6
  | 41.16 | 9.37 | 8.75
  | 44.49 | 10.76 | 9.06
  | 41.18 | 10.35 | 8.99
Average | 43.77 | 10.08 | 8.85

### DIFF
--- a/GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
+++ b/GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
@@ -6,10 +6,11 @@ namespace GVFS.Platform.Windows
 {
     public class ActiveEnumeration
     {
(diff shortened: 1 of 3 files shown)


wilbaker:
Closing PR until the second commit is fixed up

PR #324, opened 2018-10-02
Windows: Use a faster comparison when enumeration filter strings don't have wildcards

wilbaker:
Update `ActiveEnumeration` to use `string.Equals` rather than `Utils.IsFileNameMatch` when the enumeration filter string does not have wildcards.

Times with scenario provided by @kyle-rader 

  | Baseline Times (sec) | PR Changes (sec)
-- | -- | --
  | 48.23 | 9.83
  | 41.16 | 9.37 
  | 44.49 | 10.76 
  | 41.18 | 10.35 
Average | 43.77 | 10.08

### DIFF
--- a/GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
+++ b/GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
@@ -6,10 +6,11 @@ namespace GVFS.Platform.Windows
 {
     public class ActiveEnumeration
     {
-        private static FileNamePatternMatcher doesPatternMatch = null;
+        private static FileNamePatternMatcher wildcardPatternMatcher = null;
 
         // Use our own enumerator to avoid having to dispose anything
         private ProjectedFileInfoEnumerator fileInfoEnumerator;
+        private FileNamePatternMatcher patternMatcher;
 
         private string filterString = null;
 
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
```diff
@@ -6,10 +6,11 @@ namespace GVFS.Platform.Windows
 {
     public class ActiveEnumeration
     {
-        private static FileNamePatternMatcher doesPatternMatch = null;
+        private static FileNamePatternMatcher wildcardPatternMatcher = null;
 
         // Use our own enumerator to avoid having to dispose anything
         private ProjectedFileInfoEnumerator fileInfoEnumerator;
+        private FileNamePatternMatcher patternMatcher;
```
wilbaker: Rename these to better names:

patternMatcher -> doesPatternMatch
wildcardPatternMatcher -> doesWildcardPatternMatch

GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
```diff
@@ -6,10 +6,11 @@ namespace GVFS.Platform.Windows
 {
     public class ActiveEnumeration
```
CAllredMSFT: @wilbaker is this class reset if the enumeration is restarted with a new filter?

wilbaker: @CAllredMSFT Yep when the enumeration is reset the callback calls `RestartEnumeration`

```
/// <summary>
/// Resets the enumerator and advances it to the first ProjectedFileInfo in the enumeration
/// </summary>
/// <param name="filter">Filter string to save.  Can be null.</param>
public void RestartEnumeration(string filter)
{
    this.ResetEnumerator();
    this.IsCurrentValid = this.fileInfoEnumerator.MoveNext();
    this.SaveFilter(filter);
}
```

And `SaveFilter` determines which pattern matcher we use.

You can see the call to `RestartEnumeration` [here](https://github.com/Microsoft/VFSForGit/blob/355ead878bac3fade5f213aadf239332ff396224/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs#L498)

PR #326, opened 2018-10-03
GitProcess: allow GIT_TRACE to point to full path

### LINKED ISSUE #276: Don't remove GIT_TRACE environment variable if it's set to a file path
Currently VFSForGit removes all GIT_TRACE environment variables when calling git.exe (to avoid breaking the parsing of git output).

https://github.com/Microsoft/VFSForGit/blob/06793f215847ad3023977380d0623cbf59b6cde1/GVFS/GVFS.Common/Git/GitProcess.cs#L409

It would be helpful for diagnosing issues if GIT_TRACE was *not* removed if it's set to a file path.

derrickstolee:
The GIT_TRACE variable can be helpful for diagnosing problems during Git commands. However, there are many places in VFS for Git that require parsing the Git output. Using `GIT_TRACE=1` sends the output to stdout, which messes up this parsing. In the past, we have blocked that variable for this reason.

It would still be helpful to allow logging the trace output. Git allows a fully-qualified path as the value of `GIT_TRACE` and will append to that file instead. This is how we would like users to use `GIT_TRACE` when investigating issues. Unblock that type of setting.

Resolves #276

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -1,6 +1,5 @@
 using GVFS.Common.FileSystem;
 using GVFS.Common.Tracing;
-using Microsoft.Win32;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
@@ -175,7 +174,7 @@ namespace GVFS.Common.Git
             Result result = this.InvokeGitAgainstDotGitFolder("rev-parse --show-toplevel");
             return !result.HasErrors;
         }
-                
+
         public Result RevParse(string gitRef)
         {
             return this.InvokeGitAgainstDotGitFolder("rev-parse " + gitRef);
@@ -441,11 +440,24 @@ namespace GVFS.Common.Git
 
             // Removing trace variables that might change git output and break parsing
             // List of environment variables: https://git-scm.com/book/gr/v2/Git-Internals-Environment-Variables
-            foreach (string key in processInfo.EnvironmentVariables.Keys.Cast<string>()
-                .Where(x => x.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase))
-                .ToList())
+            foreach (string key in processInfo.EnvironmentVariables.Keys.Cast<string>().ToList())
             {
-                processInfo.EnvironmentVariables.Remove(key);
+                // If GIT_TRACE is set to a fully-rooted path, then Git sends the trace
+                // output to that path instead of stdout (GIT_TRACE=1) or stderr (GIT_TRACE=2).
+                if (key.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase))
+                {
+                    try
+                    {
+                        if (!Path.IsPathRooted(processInfo.EnvironmentVariables[key]))
+                        {
+                            processInfo.EnvironmentVariables.Remove(key);
+                        }
+                    }
+                    catch (ArgumentException)
+                    {
+                        processInfo.EnvironmentVariables.Remove(key);
+                    }
+                }
             }
 
             processInfo.EnvironmentVariables["GIT_TERMINAL_PROMPT"] = "0";
@@ -637,7 +649,7 @@ namespace GVFS.Common.Git
                 parseStdOutLine: parseStdOutLine,
                 timeoutMs: -1);
         }
-        
+
         public class Result
         {
             public const int SuccessCode = 0;
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/StatusVerbTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/StatusVerbTests.cs
@@ -1,0 +1,28 @@
+﻿using GVFS.FunctionalTests.FileSystemRunners;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixture]
+    public class StatusVerbTests : TestsWithEnlistmentPerFixture
+    {
+        [TestCase]
+        public void GitTrace()
+        {
+            Dictionary<string, string> environmentVariables = new Dictionary<string, string>();
+
+            this.Enlistment.Status(trace: "1");
+            this.Enlistment.Status(trace: "2");
+
+            string logPath = Path.Combine(this.Enlistment.RepoRoot, "log-file.txt");
+            this.Enlistment.Status(trace: logPath);
+
+            FileSystemRunner fileSystem = new SystemIORunner();
+            fileSystem.FileExists(logPath).ShouldBeTrue();
+            string.IsNullOrWhiteSpace(fileSystem.ReadAllText(logPath)).ShouldBeFalse();
+        }
+    }
+}
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
@@ -211,9 +211,9 @@ namespace GVFS.FunctionalTests.Tools
             return this.gvfsProcess.Diagnose();
         }
 
-        public string Status()
+        public string Status(string trace = null)
         {
-            return this.gvfsProcess.Status();
+            return this.gvfsProcess.Status(trace);
         }
 
         public bool WaitForBackgroundOperations(int maxWaitMilliseconds = DefaultMaxWaitMSForStatusCheck)
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
@@ -8,14 +8,14 @@ namespace GVFS.FunctionalTests.Tools
         private readonly string pathToGVFS;
         private readonly string enlistmentRoot;
         private readonly string localCacheRoot;
-        
+
         public GVFSProcess(string pathToGVFS, string enlistmentRoot, string localCacheRoot)
         {
             this.pathToGVFS = pathToGVFS;
             this.enlistmentRoot = enlistmentRoot;
             this.localCacheRoot = localCacheRoot;
         }
-        
+
         public void Clone(string repositorySource, string branchToCheckout, bool skipPrefetch)
         {
             string args = string.Format(
@@ -60,9 +60,9 @@ namespace GVFS.FunctionalTests.Tools
             return this.CallGVFS("diagnose \"" + this.enlistmentRoot + "\"");
         }
 
-        public string Status()
+        public string Status(string trace = null)
         {
-            return this.CallGVFS("status " + this.enlistmentRoot);
+            return this.CallGVFS("status " + this.enlistmentRoot, trace: trace);
         }
 
         public string CacheServer(string args)
@@ -90,7 +90,7 @@ namespace GVFS.FunctionalTests.Tools
             return this.CallGVFS("service " + argument, failOnError: true);
         }
 
-        private string CallGVFS(string args, bool failOnError = false)
+        private string CallGVFS(string args, bool failOnError = false, string trace = null)
         {
             ProcessStartInfo processInfo = null;
             processInfo = new ProcessStartInfo(this.pathToGVFS);
@@ -99,6 +99,11 @@ namespace GVFS.FunctionalTests.Tools
             processInfo.WindowStyle = ProcessWindowStyle.Hidden;
             processInfo.UseShellExecute = false;
             processInfo.RedirectStandardOutput = true;
+
+            if (trace != null)
+            {
+                processInfo.EnvironmentVariables["GIT_TRACE"] = trace;
+            }
 
             using (Process process = Process.Start(processInfo))
             {


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -405,11 +405,15 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
...
-            foreach (string key in processInfo.EnvironmentVariables.Keys.Cast<string>()
-                .Where(x => x.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase))
-                .ToList())
+            foreach (string key in processInfo.EnvironmentVariables.Keys.Cast<string>())
             {
-                processInfo.EnvironmentVariables.Remove(key);
+                // If GIT_TRACE is set to a fully-rooted path, then Git sends the trace
+                // output to that path instead of stdout (GIT_TRACE=1) or stderr (GIT_TRACE=2).
+                if (key.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase) &&
+                    !(key.Equals("GIT_TRACE") && Path.IsPathRooted(processInfo.EnvironmentVariables[key])))
```
jeschu1: Any reason why we wouldn't want to leave it in fancy linq?
Aka
            foreach (string key in processInfo.EnvironmentVariables.Keys.Cast<string>()
                .Where(x => x.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase) 
                    && !Path.IsPathRooted(processInfo.EnvironmentVariables[x]))
                .ToList())
            {
                processInfo.EnvironmentVariables.Remove(key);
            }

Also IsPathRooted can throw an exception if value contains invalid chars,  We should guard against that I think https://docs.microsoft.com/en-us/dotnet/api/system.io.path.ispathrooted?redirectedfrom=MSDN&view=netframework-4.7.2.

derrickstolee: Good find on that exception! I'll need to check for invalid chars first, I guess.

The extra complication we are doing here is something that is too large to fit nicely into a `.Where()` clause. It's also kind of pointless that we `.ToList()` something that is looped in a `foreach`.

jeschu1: Thanks for going after this bug, it was actually a change I requested :-)

Why do you need the !(key.Equals(GIT_TRACE) bit?  

Agree we can drop the ToList.  It reads nicely to me in the Linq, but I'm happy to follow your lead here!

derrickstolee: There are other `GIT_TRACE_` variables, like `GIT_TRACE_PERF`, but I suppose we could stop interfering with those, too.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -405,11 +404,21 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
...
-                processInfo.EnvironmentVariables.Remove(key);
+                // If GIT_TRACE is set to a fully-rooted path, then Git sends the trace
+                // output to that path instead of stdout (GIT_TRACE=1) or stderr (GIT_TRACE=2).
+                if (key.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase))
+                {
+                    string value = processInfo.EnvironmentVariables[key];
+
+                    if (!key.Equals("GIT_TRACE")
+                        || value.IndexOfAny(Path.GetInvalidPathChars()) >= 0
+                        || !Path.IsPathRooted(processInfo.EnvironmentVariables[key]))
```
kewillford: Don't we want all the `GIT_TRACE` permutations to be able to use a fully-rooted path?  Would it make the code easier to understand by checking all the conditions for value like 1, 2, true?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -45,6 +46,48 @@ public void StatusTest()
...
         }
 
+        [TestCase]
+        public void StatusWithGitTraceTest()
+        {
+            Dictionary<string, string> environmentVariables = new Dictionary<string, string>();
+
+            string tracePath = Path.Combine(this.Enlistment.EnlistmentRoot, "trace-log.txt");
+            environmentVariables["GIT_TRACE"] = tracePath;
+            this.ValidateGitCommand(environmentVariables, "status");
```
kewillford: I wonder if this should just be a flag or file name and we set different paths for the vfs repo and control repo so that we can compare the contents of the files against each other?

derrickstolee: Trace logs have two problems when comparing them directly:

1. Every line starts with a timestamp. We'd need to ignore these.

2. If we verify equality, then we are actually saying "these two commands did the exact same things underneath" which is not always appropriate.

The current setup ignores this by only verifying the path exists and an event that happens exactly once in each execution happens twice in the combined log.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -405,11 +404,20 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
...
+            foreach (string key in processInfo.EnvironmentVariables.Keys.Cast<string>())
             {
-                processInfo.EnvironmentVariables.Remove(key);
+                // If GIT_TRACE is set to a fully-rooted path, then Git sends the trace
+                // output to that path instead of stdout (GIT_TRACE=1) or stderr (GIT_TRACE=2).
+                if (key.StartsWith("GIT_TRACE", StringComparison.OrdinalIgnoreCase))
+                {
+                    string value = processInfo.EnvironmentVariables[key];
+
+                    if (value.IndexOfAny(Path.GetInvalidPathChars()) >= 0
```
wilbaker: I am guessing most of the time the environment variable will not have invalid path characters.  Would it be better from a perf standpoint to just call `IsPathRooted` and catch `ArgumentException` (in case there is an invalid char)?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -45,6 +46,48 @@ public void StatusTest()
...
+                }
+            }
+
+            // The GIT_TRACE output is appended to the file, so we should have
+            // one copy for each run: normal and virtualized.
+            count.ShouldEqual(2);
+        }
+
+        [TestCase]
+        public void StatusWithGitTraceInvalidCharactersTest()
```
wilbaker: Do we already have a test where `status` is run with a valid value for "GIT_TRACE" (that is not a file)? If not, can\should we add one?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
kewillford: I still think keeping 2 separate trace files for the control and vfs repo helps to keep one from affecting the other in unforeseen ways.  Also keeping them separate makes it easier to check the sequence of 
1. File was written
2. Expect it to contain X

Right now the control/vfs repo could have created the file and written the 2 lines that you are counting and the other repo might have done nothing but the test would pass.

GVFS/GVFS.Common/Git/GitProcess.cs
derrickstolee: It is almost always better to check a condition than to catch an exception.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
derrickstolee: Hm. Perhaps I could do it with `GIT_TRACE=2` to pipe to stderr, since if we test with `GIT_TRACE=1` the output will be different.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -3,8 +3,7 @@
 using GVFS.FunctionalTests.Tools;
 using GVFS.Tests.Should;
 using NUnit.Framework;
-using NUnit.Framework.Interfaces;
-using System;
+using System.Collections.Generic;
```
derrickstolee: No need for this change anymore.

GVFS/GVFS.Common/Git/GitProcess.cs
sanoursa: Not when the condition has a perf cost and will almost never yield a hit, whereas the exception handler is free to set up. I would favor the exception handler here as well.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -45,6 +46,75 @@ public void StatusTest()
...
+            environmentVariables["GIT_TRACE"] = tracePath;
+
+            string command = "status";
+            string controlRepoRoot = this.ControlGitRepo.RootPath;
+            string gvfsRepoRoot = this.Enlistment.RepoRoot;
+
+            ProcessResult expectedResult = GitProcess.InvokeProcess(controlRepoRoot, command, environmentVariables);
+            ProcessResult actualResult = GitHelpers.InvokeGitAgainstGVFSRepo(gvfsRepoRoot, command, environmentVariables);
+
+            GitHelpers.ValidateGitOutput(command, expectedResult, actualResult);
```
kewillford: Here you can just call `ValidateGitCommand` and remove 5 lines of code.  Makes me wonder if we can somehow get that method or add a method to handle `GIT_TRACE` environment variables and simplify these tests but that might be messier and if these are the only tests that will use it, doesn't make as much sense.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -45,6 +46,75 @@ public void StatusTest()
...
+
+            string.IsNullOrWhiteSpace(expectedResult.Errors).ShouldBeFalse();
+            string.IsNullOrWhiteSpace(actualResult.Errors).ShouldBeTrue();
```
kewillford: Why did one write to Errors but the other didn't?  Shouldn't the `ValidateGitOutput` be checking the Errors?

derrickstolee: `ValidateGitOutput` does check the errors, which is why this test was failing.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
derrickstolee: In this one case, I can inject the environment into `ValidateGitCommand` and rely on its comparisons. That does save code overall, and makes it easier to add tests like it in the future.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
derrickstolee: These tests are testing the wrong thing :(. The environment stuff we are checking is only for Git commands invoked from GVFS, while these tests are calling Git directly (not going through the code paths I'm checking). That's why they are failing.

derrickstolee:
@kewillford Could you take a look at the new test structure? It took me a while to figure out the right thing to do, and it's very different than what you approved.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/StatusVerbTests.cs
```diff
@@ -0,0 +1,34 @@
...
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixture]
+    public class StatusVerbTests : TestsWithEnlistmentPerFixture
+    {
+        private enum ExpectedReturnCode
```
kewillford: I don't see where this is used?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/StatusVerbTests.cs
```diff
@@ -0,0 +1,34 @@
...
+            ParsingError = 1,
+        }
+
+        [TestCase]
+        public void GitTrace()
+        {
+            Dictionary<string, string> environmentVariables = new Dictionary<string, string>();
+
+            this.Enlistment.Status(trace: "1");
+            this.Enlistment.Status(trace: "2");
```
kewillford: Do we need to verify anything for these?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/StatusVerbTests.cs
derrickstolee: Oops. It was needed for a helper function I had copied in an earlier version. Will delete.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/StatusVerbTests.cs
derrickstolee: So `gvfs status` could fail to parse the `git status` it calls if the `GIT_TRACE` values are not removed.

kewillford: Looks like we are sending a `gvfs status` command.  Will that cause a `git status` to run as well?  I wasn't able to find that in the quick search I did.  Could you point me at that code?

derrickstolee: You're right. I thought it did, but it doesn't. I must call _something_, since `logPath` is filled with contents.

derrickstolee: Ok, I took a look at my logs, and we actually run `git config` to get the remote url:

```
07:52:55.681330 exec-cmd.c:236          trace: resolved executable dir: C:/Program Files/Git/mingw64/bin
07:52:55.684326 run-command.c:638       trace: run_command: C:/_git/ForTests2/src/.git/hooks/pre-command.exe config --local remote.origin.url --git-pid=16732
07:52:55.808825 git.c:477               trace: built-in: git config --local remote.origin.url
07:52:55.809326 run-command.c:638       trace: run_command: C:/_git/ForTests2/src/.git/hooks/post-command.exe config --local remote.origin.url --git-pid=16732 --exit_code=0
```

This is run by `GitProcess.GetOriginUrl()`.

derrickstolee: The output is interpreted [here in `Enlistment.cs`](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Common/Enlistment.cs#L36-L47) and would fail if we didn't redirect `GIT_TRACE=1` (the output would not be a URL) or `GIT_TRACE=2` (`originResult.HasErrors` would be true).

Issue #289, opened 2018-09-21
Diagnose logs incorrect cache server url


jrbriggs:
**Steps to reproduce:**
1. Have a cache server.
2. Run gvfs diagnose.
3. Look in diagnostics.log

**Expected Behavior:**
1. Should have correct cache server url.

Instead shows invalid://repoUrl.

derrickstolee:
This also affects jobs like `gvfs repair` and can even lead to the URL being passed to a credential manager.

jeschu1:
There are 2 different bugs here

1.  Under certain conditions we don't correctly init the enlistment on repair, causing us to use an invalid url.  This is do to a miss on refactoring.
2. If we don't have a cache server we are showing an invalid url instead of nothing.  This is due to us using the default url since it wasn't overwritten on the config lookup.

I will addresss each in separate PRs.

PR #327, opened 2018-10-03
Use correct repoUrl for credential

jeschu1:
This change allows us to reinitialize the GitProcess with valid repository information and pass it correctly to the credential call.

The bug was caused by the refactor [here](https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/commit/cec8bb00b1a4954329fd5a6e5dd6243f47fe381e?refName=refs%2Fheads%2Fmaster&_a=compare&path=%2FGVFS%2FGVFS.CLI%2FRepairJobs%2FGitConfigRepairJob.cs).  
- Previously, we used the repoUrl from the enlistment on GitProcess creation.
- Now, we use the url passed in on  TryGetCredentials.

A new unit test was added that exposed the failure.

#289

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
@@ -14,6 +14,14 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
     public class RepairTests : TestsWithEnlistmentPerTestCase
     {
         [TestCase]
+        public void NoFixesNeeded()
+        {
+            this.Enlistment.UnmountGVFS();
+
+            this.Enlistment.Repair();
+        }
+
+        [TestCase]
         public void FixesCorruptHeadSha()
         {
             this.Enlistment.UnmountGVFS();
--- a/GVFS/GVFS/RepairJobs/GitConfigRepairJob.cs
+++ b/GVFS/GVFS/RepairJobs/GitConfigRepairJob.cs
@@ -45,6 +45,7 @@ namespace GVFS.RepairJobs
 
             // At this point, we've confirmed that the repo url can be gotten, so we have to 
             // reinitialize the GitProcess with a valid repo url for 'git credential fill'
+            string repoUrl = null;
             try
             {
                 GVFSEnlistment enlistment = GVFSEnlistment.CreateFromDirectory(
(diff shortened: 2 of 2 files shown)


GVFS/GVFS/RepairJobs/RepairJob.cs
```diff
@@ -37,7 +37,7 @@ public enum FixResult
 
         protected ITracer Tracer { get; }
         protected TextWriter Output { get; }
-        protected GVFSEnlistment Enlistment { get; }
+        protected GVFSEnlistment Enlistment { get; set; }
```
sanoursa: Yea I definitely think we should keep the property readonly. In general, readonly state is much easier to reason about than mutable state.

GVFS/GVFS/RepairJobs/GitConfigRepairJob.cs
```diff
@@ -47,11 +47,11 @@ public override IssueType HasIssue(List<string> messages)
             // reinitialize the GitProcess with a valid repo url for 'git credential fill'
             try
             {
-                GVFSEnlistment enlistment = GVFSEnlistment.CreateFromDirectory(
+                this.Enlistment = GVFSEnlistment.CreateFromDirectory(
```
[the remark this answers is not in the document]
sanoursa: See my other comment. Given that you had to add a new set accessor, I would go with your other idea for the fix.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
```diff
@@ -13,6 +13,14 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
     [Category(Categories.MacTODO.M4)]
     public class RepairTests : TestsWithEnlistmentPerTestCase
     {
+        [TestCase]
+        public void NoFixesNeeded()
+        {
+            this.Enlistment.UnmountGVFS();
+
+            this.Enlistment.Repair();
```
kewillford: Do you know if the repair without the `--confirm` would have caught this as well?  I was thinking that for each of our tests in this class would should run the repair without `--confirm` before running the confirm version to catch other possible code paths.

jeschu1: @kewillford , yes without --confirm we still get the error.  It's up to us if we want to do another run without the flag.  My only concern would be the length of time to run the tests, but if it's a test worth doing we should do it!

kewillford: I was thinking we could do it in the test cases we already have not adding new test cases.
Run repair without `--confirm` then run with it in the same test case.  Hopefully this wouldn't add to much time.

PR #329, opened 2018-10-03
Show "None" message in diagnose instead of invalid url

jeschu1:
When a cache server is not set up we shouldn't be showing the invalid default url.  

Instead let's show a helpful message to the user.

#289

### DIFF
--- a/GVFS/GVFS.Common/GVFSEnlistment.cs
+++ b/GVFS/GVFS.Common/GVFSEnlistment.cs
@@ -13,7 +13,6 @@ namespace GVFS.Common
         public const string BlobSizesCacheName = "blobSizes";
 
         private const string GitObjectCacheName = "gitObjects";
-        private const string InvalidRepoUrl = "invalid://repoUrl";
 
         private string gitVersion;
         private string gvfsVersion;
@@ -90,7 +89,7 @@ namespace GVFS.Common
                     return null;
                 }
 
-                return new GVFSEnlistment(enlistmentRoot, InvalidRepoUrl, gitBinRoot, gvfsHooksRoot);
+                return new GVFSEnlistment(enlistmentRoot, string.Empty, gitBinRoot, gvfsHooksRoot);
             }
 
             return null;
--- a/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
+++ b/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
@@ -53,7 +53,7 @@ namespace GVFS.CommandLine
                 this.WriteMessage(GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath());
                 this.WriteMessage(string.Empty);
                 this.WriteMessage("Enlistment root: " + enlistment.EnlistmentRoot);
-                this.WriteMessage("Cache Server: " + CacheServerResolver.GetUrlFromConfig(enlistment));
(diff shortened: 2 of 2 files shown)


GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -11,9 +11,9 @@ namespace GVFS.Common
     public partial class GVFSEnlistment : Enlistment
     {
         public const string BlobSizesCacheName = "blobSizes";
+        public const string InvalidRepoUrl = "invalid://repoUrl";
```
sanoursa: I don't think we should make this constant public. The point of it was that it should never leak out of this class.

Right now, I think you're fixing the symptom by checking for this constant before displaying it. But how did the invalid URL leak out of here in the first place? What other code paths might be affected? Rather than special case all uses, I'd like to understand the root cause better.

sanoursa: I don't think we should make this constant public. The point of it was that it should never leak out of this class.

Right now, I think you're fixing the symptom by checking for this constant before displaying it. But how did the invalid URL leak out of here in the first place? What other code paths might be affected? Rather than special case all uses, I'd like to understand the root cause better.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -53,7 +53,9 @@ protected override void Execute(GVFSEnlistment enlistment)
                 this.WriteMessage(GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath());
                 this.WriteMessage(string.Empty);
                 this.WriteMessage("Enlistment root: " + enlistment.EnlistmentRoot);
-                this.WriteMessage("Cache Server: " + CacheServerResolver.GetUrlFromConfig(enlistment));
+                string cacheServer = CacheServerResolver.GetUrlFromConfig(enlistment);
```
sanoursa: Note that `GetUrlFromConfig` should have returned the origin URL if there was no cache server configured. So this is not just a case of "no cache server". Did we actually somehow create a bogus origin remote? Or is there some other bug that is causing that bad URL to leak out of this method?

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -90,7 +89,7 @@ public static GVFSEnlistment CreateWithoutRepoUrlFromDirectory(string directory,
                     return null;
                 }
 
-                return new GVFSEnlistment(enlistmentRoot, InvalidRepoUrl, gitBinRoot, gvfsHooksRoot);
+                return new GVFSEnlistment(enlistmentRoot, string.Empty, gitBinRoot, gvfsHooksRoot);
```
jeschu1: @sanoursa we discussed sending null in here, however in the base if you send in null it will attempt to look up details for the repository [here](https://github.com/Microsoft/VFSForGit/blob/4a489012e5bef89adf2afd366b37928c5be2eec7/GVFS/GVFS.Common/Enlistment.cs#L30).  We don't want that in this case.  Previously there was a bug fix to diagnose explicitly saying don't fail if you can't get origin info.

We could look at other options like sending in a flag rather than the string.Empty.  Let me know what you think.

sanoursa: I think this is fine. Any place where we try to use this as a URL, we will still fail. And your fix in DiagnoseVerb now ensures that we show a good display string.

PR #334, opened 2018-10-03
Mac: Prevent double hydration and directory expansion

### LINKED ISSUE #315: Mac: Double hydration race condition leads to sporadic failures of MultithreadedReadWriteTests.CanReadVirtualFileInParallel
Example failure:

```
2018-09-28T16:48:12.0939760Z 1) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.MultithreadedReadWriteTests.CanReadVirtualFileInParallel()
2018-09-28T16:48:12.0940220Z   At least one of the reads failed
2018-09-28T16:48:12.0940410Z   Expected: null
2018-09-28T16:48:12.0941870Z   But was:  <System.IO.IOException: The process cannot access the file '/GVFS.FT/test/c6474bc299d348b68d54/src/GVFS/GVFS.FunctionalTests/Tests/LongRunningEnlistment/GitMoveRenameTests.cs' because it is being used by another process.
2018-09-28T16:48:12.0942170Z    at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
2018-09-28T16:48:12.0942320Z    at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
2018-09-28T16:48:12.0942520Z    at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
2018-09-28T16:48:12.0942680Z    at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize)
2018-09-28T16:48:12.0942810Z    at System.IO.File.InternalReadAllText(String path, Encoding encoding)
2018-09-28T16:48:12.0942930Z    at System.IO.File.ReadAllText(String path)
2018-09-28T16:48:12.0943080Z    at GVFS.FunctionalTests.FileSystemRunners.SystemIORunner.ReadAllText(String path) in /Users/vsts/agent/2.140.2/work/1/s/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs:line 70
2018-09-28T16:48:12.0943320Z    at GVFS.FunctionalTests.Tests.EnlistmentPerFixture.MultithreadedReadWriteTests.<>c__DisplayClass0_0.<CanReadVirtualFileInParallel>b__0() in /Users/vsts/agent/2.140.2/work/1/s/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MultithreadedReadWriteTests.cs:line 36>
2018-09-28T16:48:12.0943580Z    at GVFS.Tests.Should.ValueShouldExtensions.ShouldBeNull[T](T obj, String message) in /Users/vsts/agent/2.140.2/work/1/s/GVFS/GVFS.Tests/Should/ValueShouldExtensions.cs:line 76
2018-09-28T16:48:12.0943820Z    at GVFS.FunctionalTests.Tests.EnlistmentPerFixture.MultithreadedReadWriteTests.CanReadVirtualFileInParallel() in /Users/vsts/agent/2.140.2/work/1/s/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MultithreadedReadWriteTests.cs:line 52

```

https://dev.azure.com/mseng/VSOnline/_build/results?buildId=7373385&view=logs

wilbaker:
Fixes #315 

There were a few issues in the ProjFS for Mac user-mode library:

- All messages from the kernel would be coalesced into a single callback to the provider, meaning it was possible that notifications could get lost.

- The path comparisons when deciding if messages should be collapsed were case sensitive.

- When messages were collapsed, the library was leaking the memory allocated for any of the collapsed messages.

- The hydrate\enumerate code was not checking if the file\directory had already been expanded.  As a result, if two requests came in sequentially, it's possible that with the right timing they would both run and problems would occur when trying to expand the file\directory a second time.

This PR addresses these issues with the following changes:

- Messages are no longer collapsed

- There is a new per-path collection of mutexes to ensure that two threads don't expand the same path at the same time.

- The hydration\enumeration handles check if the path is expanded prior to asking the provider to expand it.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -9,8 +9,9 @@
 #include <unistd.h>
 #include <dirent.h>
 #include <queue>
-#include <unordered_map>
+#include <memory>
 #include <set>
+#include <map>
 #include <IOKit/IOKitLib.h>
 #include <IOKit/IODataQueueClient.h>
 #include <mach/mach_port.h>
@@ -35,14 +36,16 @@ using std::hex;
 using std::is_pod;
 using std::lock_guard;
 using std::make_pair;
+using std::make_shared;
+using std::map;
 using std::move;
 using std::mutex;
 using std::oct;
 using std::pair;
 using std::queue;
 using std::set;
+using std::shared_ptr;
 using std::string;
-using std::unordered_map;
 
 typedef lock_guard<mutex> mutex_lock;
 
@@ -69,7 +72,7 @@ static void CombinePaths(const char* root, const char* relative, char (&combined
 static errno_t SendKernelMessageResponse(uint64_t messageId, MessageType responseType);
 static errno_t RegisterVirtualizationRootPath(const char* path);
 
-static void HandleKernelRequest(Message requestSpec, void* messageMemory);
+static void HandleKernelRequest(void* messageMemory, uint32_t messageSize);
 static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* path);
 static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHeader* request, const char* path);
 static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const char* path);
@@ -94,10 +97,27 @@ static PrjFS_Callbacks s_callbacks;
 static dispatch_queue_t s_messageQueueDispatchQueue;
 static dispatch_queue_t s_kernelRequestHandlingConcurrentQueue;
 
-// Map of relative path -> set of pending message IDs for that path, plus mutex to protect it.
-static unordered_map<string, set<uint64_t>> s_PendingRequestMessageIDs;
-static mutex s_PendingRequestMessageMutex;
+struct CaseInsensitiveStringCompare
+{
+    bool operator() (const std::string& lhs, const std::string& rhs) const
+    {
+        return strcasecmp(lhs.c_str(), rhs.c_str()) < 0;
+    }
+};
 
+struct MutexAndUseCount
+{
+    shared_ptr<mutex> mutex;
+    int useCount;
+};
+
+// Map of relative path -> MutexAndUseCount for that path, plus mutex to protect the map itself.
+typedef map<string, MutexAndUseCount, CaseInsensitiveStringCompare> PathToMutexMap;
+PathToMutexMap s_inProgressExpansions;
+mutex s_inProgressExpansionsMutex;
+
+static shared_ptr<mutex> CheckoutPathMutex(const string& fullPath);
+static void ReturnPathMutex(const string& fullPath, const shared_ptr<mutex>& mutex);
 
 // The full API is defined in the header, but only the minimal set of functions needed
 // for the initial MirrorProvider implementation are listed here. Calling any other function
@@ -192,37 +212,11 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
                 cerr << "Unexpected result dequeueing message - result 0x" << hex << result << " dequeued " << dequeuedSize << "/" << messageSize << " bytes\n";
                 abort();
             }
-            
-            Message message = ParseMessageMemory(messageMemory, messageSize);
-            
-            // At the moment, we expect all messages to include a path
-            assert(message.path != nullptr);
-
-            // Ensure we don't run more than one request handler at once for the same file
-            {
-                mutex_lock lock(s_PendingRequestMessageMutex);
-                typedef unordered_map<string, set<uint64_t>>::iterator PendingMessageIterator;
-                    PendingMessageIterator file_messages_found = s_PendingRequestMessageIDs.find(message.path);
-                if (file_messages_found == s_PendingRequestMessageIDs.end())
-                {
-                    // Not handling this file/dir yet
-                    pair<PendingMessageIterator, bool> inserted =
-                        s_PendingRequestMessageIDs.insert(make_pair(string(message.path), set<uint64_t>{ message.messageHeader->messageId }));
-                    assert(inserted.second);
-                }
-                else
-                {
-                    // Already a handler running for this path, don't handle it again.
-                    file_messages_found->second.insert(message.messageHeader->messageId);
-                    continue;
-                }
-            }
-            
 
             dispatch_async(
                 s_kernelRequestHandlingConcurrentQueue,
                 ^{
-                    HandleKernelRequest(message, messageMemory);
+                    HandleKernelRequest(messageMemory, messageSize);
                 });
         }
     });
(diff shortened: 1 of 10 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -94,10 +97,27 @@ static PrjFS_Callbacks s_callbacks;
...
+};
+
+struct MutexAndUseCount
+{
+    shared_ptr<mutex> mutex;
+    int useCount;
+};
 
+// Map of relative path -> MutexAndUseCount for that path, plus mutex to protect the map itself.
+typedef map<string, MutexAndUseCount, CaseInsensitiveStringCompare> PathToMutexMap;
```
wilbaker: @pmj do you know if we need to worry about being case insensitive?  Or will the kext always use the same\actual case when sending paths to user-mode?

wilbaker: CC @sanoursa  as well

pmj: I'm not sure; the path is ultimately coming from `vn_getpath()` and I believe it gets its information from the name cache. The question is therefore if the name cache is populated with canonical names or the component names that came in from user space.

One correct way to check whether 2 paths refer to the same file would be to `stat()` them both and check if fsid & inodes match. The kext should already have this information, so perhaps we should be sending that tuple along with each kernel -> user message and using that as our map key?

Given that macOS does support case-sensitive file systems, case-insensitive string comparison is not really ideal here. I mean, we can of course declare that we won't support VFS for Git on case-sensitive file systems, but I'd still be worried about edge cases.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1031,3 +1056,34 @@ static const char* NotificationTypeToString(PrjFS_NotificationType notificationT
...
+        return newEntry.first->second.mutex;
+    }
+    else
+    {
+        iter->second.useCount++;
+        return iter->second.mutex;
+    }
+}
+
+static void ReturnPathMutex(const string& fullPath, const shared_ptr<mutex>& mutex)
```
pmj: It's a fairly minor point, but given that `CheckoutPathMutex`/`ReturnPathMutex` are always paired in a function, I'm wondering if `CheckoutPathMutex` should return an iterator/mutex `pair`, and `ReturnPathMutex` would accept that iterator to avoid looking up the fullPath afresh.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Incidentally, the inode approach also tackles any hardlink issues.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: That's a good idea thanks!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: > One correct way to check whether 2 paths refer to the same file would be to stat() them both and check if fsid & inodes match. The kext should already have this information, so perhaps we should be sending that tuple along with each kernel -> user message and using that as our map key?

That's a good idea, I'll see about plumbing that through.

wilbaker: > check if fsid & inodes match

@pmj Totally unrelated to this PR, your comment reminded me of something I'd been thinking about related to how the kext could handle a failures of `vn_getpath` (or perhaps it wouldn't even call it at all).

Do you have any thoughts an approach where the kext passes fsid & inode and then the user-mode library calls `fsgetpath`?  I wasn't sure if that would potentially be more reliable than `vn_getpath`.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -94,10 +97,27 @@ static PrjFS_Callbacks s_callbacks;
...
+
+struct MutexAndUseCount
+{
+    shared_ptr<mutex> mutex;
+    int useCount;
+};
 
+// Map of relative path -> MutexAndUseCount for that path, plus mutex to protect the map itself.
+typedef map<string, MutexAndUseCount, CaseInsensitiveStringCompare> PathToMutexMap;
+PathToMutexMap s_inProgressExpansions;
```
sanoursa: Rename to just `s_fileLocks`? Today, this is protecting expansions, but I think it will need to expand to protect all placeholder transitions, e.g. the placeholder->full and hydrated->empty transitions as well.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -94,10 +97,27 @@ static PrjFS_Callbacks s_callbacks;
...
+mutex s_inProgressExpansionsMutex;
+
+static shared_ptr<mutex> CheckoutPathMutex(const string& fullPath);
```
sanoursa: Style nit: all the function prototypes are declared up above in one section

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1031,3 +1056,34 @@ static const char* NotificationTypeToString(PrjFS_NotificationType notificationT
...
+    if (iter == s_inProgressExpansions.end())
+    {
+        pair<PathToMutexMap::iterator, bool> newEntry = s_inProgressExpansions.insert(
+            PathToMutexMap::value_type(fullPath, { make_shared<mutex>(), 1 }));
+        assert(newEntry.second);
+        return newEntry.first->second.mutex;
+    }
+    else
+    {
+        iter->second.useCount++;
```
sanoursa: In theory, this could happen outside of the global lock (and use interlocked operations instead). How much do you think that would reduce contention during a parallel build that is doing lots of hydrations? Is it worth the trouble?

wilbaker: > In theory, this could happen outside of the global lock (and use interlocked operations instead).

I think there is a scenario where that approach would not be thread safe:

Thread 1: Calls `CheckoutPathMutex` and a new mutex is added to the global collection
Thread 2: Calls `CheckoutPathMutex` and release global lock before incrementing `useCount`
Thread 1: Calls `ReturnPathMutex`, useCount drops to 0 and so the mutex is removed from the global collection
Thread 3: Calls `CheckoutPathMutex`, a new mutex is created for the same path

Thread 2 and 3 would now be holding two different mutexes for the same path.

I think we can leave this as-is for now, and if we identify that it's causing perf problems we can look into improvements\alternatives.

wilbaker:
> LGTM, I agree with Saeed that we should get this in before considering moving to fsid & inode matching rather than blocking on it.

@sanoursa @nickgra, I just pushed a new commit that updates the map key to be fsid+inode, could you take another look?

PR #339, opened 2018-10-04
Pass all required components to vswhere in BuildGVFSForWindows.bat

jrbriggs:
When installing new versions of Visual Studio for testing, I often do not install all the workloads. This PR makes vswhere smarter about choosing the latest version of VS that can actually build VFS for Git, rather than the latest version of VS period.

### DIFF
--- a/Scripts/BuildGVFSForWindows.bat
+++ b/Scripts/BuildGVFSForWindows.bat
@@ -19,14 +19,20 @@ SET vswhere=%~dp0..\..\packages\vswhere.%vswherever%\tools\vswhere.exe
 
 :: Use vswhere to find the latest VS installation (including prerelease installations) with the msbuild component.
 :: See https://github.com/Microsoft/vswhere/wiki/Find-MSBuild
-for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -products * -requires Microsoft.Component.MSBuild -property installationPath`) do (
+for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -products * -requires Microsoft.Component.MSBuild Microsoft.VisualStudio.Workload.ManagedDesktop Microsoft.VisualStudio.Workload.NativeDesktop Microsoft.VisualStudio.Workload.NetCoreTools Microsoft.Component.NetFX.Core.Runtime Microsoft.VisualStudio.Component.Windows10SDK.10240 -property installationPath`) do (
   set VsInstallDir=%%i
+)
+
+IF NOT DEFINED VsInstallDir (
+  echo ERROR: Could not locate a Visual Studio installation with required components.
+  echo Refer to Readme.md for a list of the required Visual Studio components.
+  exit /b 10
 )
 
 SET msbuild="%VsInstallDir%\MSBuild\15.0\Bin\amd64\msbuild.exe"
 IF NOT EXIST %msbuild% (
-	echo Error: Could not find msbuild
-	exit /b 1
+  echo ERROR: Could not find msbuild
+  exit /b 1
 )
 
 %msbuild% %~dp0\..\GVFS.sln /p:GVFSVersion=%GVFSVersion% /p:Configuration=%SolutionConfiguration% /p:Platform=x64 || exit /b 1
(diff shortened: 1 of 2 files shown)


Scripts/BuildGVFSForWindows.bat
```diff
@@ -19,14 +19,20 @@ SET vswhere=%~dp0..\..\packages\vswhere.%vswherever%\tools\vswhere.exe
 
 :: Use vswhere to find the latest VS installation (including prerelease installations) with the msbuild component.
 :: See https://github.com/Microsoft/vswhere/wiki/Find-MSBuild
-for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -products * -requires Microsoft.Component.MSBuild -property installationPath`) do (
+for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -products * -requires Microsoft.Component.MSBuild Microsoft.VisualStudio.Workload.NativeDesktop Microsoft.VisualStudio.Workload.NetCoreTools Microsoft.Component.NetFX.Core.Runtime Microsoft.VisualStudio.Component.VC.CLI.Support Microsoft.VisualStudio.Component.Windows10SDK.10240 -property installationPath`) do (
```
sanoursa: Fails to build on my machine, even though I can build master.

```
C:\Repos\VFSForGit\src>Scripts\BuildGVFSForWindows.bat
Feeds used:
  C:\Users\sanoursa\.nuget\packages\
  https://www.myget.org/F/gvfsbuilddependencies/api/v3/index.json
  https://api.nuget.org/v3/index.json

Package "vswhere.2.5.2" is already installed.
ERROR: Could not locate a Visual Studio installation with required components.
Refer to Readme.md for a list of the required Visual Studio components.
```

One of these listed components must not actually be required. Does this list match what's in the readme?

jrbriggs: In reviewing the list, I discovered I had actually missed a "required" component.

Here's the mapping (from https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=vs-2017#net-desktop-development)
* Install Visual Studio 2017 Community Edition or higher (https://www.visualstudio.com/downloads/). 
  * Include the following workloads:
    * .NET desktop development => `Microsoft.VisualStudio.Workload.ManagedDesktop `
    * Desktop development with C++ => `Microsoft.VisualStudio.Workload.NativeDesktop `
    * .NET Core cross-platform development => `Microsoft.VisualStudio.Workload.NetCoreTools `
  * Include the following additional components:
    * .NET Core runtime => `Microsoft.Component.NetFX.Core.Runtime`
    * C++/CLI support => `Microsoft.VisualStudio.Component.VC.CLI.Support`
    * Windows 10 SDK (10.0.10240.0) => `Microsoft.VisualStudio.Component.Windows10SDK.10240`

Would you check your installation to see which of these shouldn't actually be required? I'm guessing it's .NET Core since you're probably getting that from the next step in the readme that says to install the .NET Core 2.1 SDK.

sanoursa: The component I'm missing from your list is `Microsoft.VisualStudio.Component.VC.CLI.Support`, which is weird because I can build the C++/CLI project. 

Dropping that item makes vswhere happy:
```
C:\Repos\VFSForGit\src>..\packages\vswhere.2.5.2\tools\vswhere.exe -all -prerelease -latest -products *  -property installationPath -requires Microsoft.Component.MSBuild Microsoft.VisualStudio.Workload.NativeDesktop Microsoft.VisualStudio.Workload.NetCoreTools Microsoft.Component.NetFX.Core.Runtime Microsoft.VisualStudio.Component.Windows10SDK.10240
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise
```

It's also been a long time since I updated VS because I've been almost exclusively on my Mac. Updating everything now to see if that resolves it, stand by...

sanoursa: @wilbaker  pointed out that we no longer build any C++/CLI code, we now consume that from the ProjFS nuget package. I don't remember when I disabled the C++/CLI component, but apparently I did, and I can still build. So I think we can just drop that as a requirement (here and in the readme)

jrbriggs: Thanks for helping track that down. I've removed it from the readme and the required components.

PR #343, opened 2018-10-04
Simplify the logic for `NamedPipeStreamReader` and fix handling of multiple sequential messages

### LINKED ISSUE #333: NamedPipeStreamReader might read in two messages as one
Example functional test failure:

```
1) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerTestCase.CaseOnlyFolderRenameTests.CaseRenameFoldersAndRemountAndRenameAgain()
Unmounting...Failed. Run 'gvfs log /GVFS.FT/test/3449aa1662704686a6a4' for more info.
Unrecognized response to unmount: AckComplete
  Expected: 0
  But was:  3
   at GVFS.Tests.Should.ValueShouldExtensions.ShouldEqual[T](T actualValue, T expectedValue, String message) in /Users/wilbaker/Repos/VFSForGit/src/GVFS/GVFS.Tests/Should/ValueShouldExtensions.cs:line 34
   at GVFS.FunctionalTests.Tools.GVFSProcess.CallGVFS(String args, Boolean failOnError) in /Users/wilbaker/Repos/VFSForGit/src/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs:line 109
   at GVFS.FunctionalTests.Tools.GVFSProcess.Unmount() in /Users/wilbaker/Repos/VFSForGit/src/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs:line 76
   at GVFS.FunctionalTests.Tools.GVFSFunctionalTestEnlistment.UnmountGVFS() in /Users/wilbaker/Repos/VFSForGit/src/GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs:line 227
   at GVFS.FunctionalTests.Tests.EnlistmentPerTestCase.CaseOnlyFolderRenameTests.CaseRenameFoldersAndRemountAndRenameAgain() in /Users/wilbaker/Repos/VFSForGit/src/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/CaseOnlyFolderRenameTests.cs:line 56
```

I suspect this is due to the change to reading messages with `NamedPipeStreamReader.ReadMessage` instead of `ReadLine`.  If there are two messages in the stream and `ReadMessage` reads in both at once it looks like it will combine them into a single message.

jamill:
Update the `NamedPipeStreamReader` to read individual bytes from the underlying stream. This avoids complexity associated with reading a larger chunk of bytes and handling the different cases associated with that approach. This change also fixes handling of multiple sequential messages (#333)

### DIFF
--- a/GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
+++ b/GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
@@ -11,23 +11,16 @@ namespace GVFS.Common.NamedPipes
     /// </summary>
     public class NamedPipeStreamReader
     {
-        private const int DefaultBufferSize = 1024;
+        private const int InitialListSize = 1024;
         private const byte TerminatorByte = 0x3;
+        private readonly byte[] buffer;
 
-        private int bufferSize;
-        private byte[] buffer;
         private Stream stream;
 
-        public NamedPipeStreamReader(Stream stream, int bufferSize)
+        public NamedPipeStreamReader(Stream stream)
         {
             this.stream = stream;
-            this.bufferSize = bufferSize;
-            this.buffer = new byte[this.bufferSize];
-        }
-
-        public NamedPipeStreamReader(Stream stream)
-            : this(stream, DefaultBufferSize)
-        {
+            this.buffer = new byte[1];
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -11,23 +11,16 @@ namespace GVFS.Common.NamedPipes
     /// </summary>
     public class NamedPipeStreamReader
     {
-        private const int DefaultBufferSize = 1024;
+        private const int InitialListSize = 1024;
         private const byte TerminatorByte = 0x3;
-
-        private int bufferSize;
         private byte[] buffer;
```
sanoursa: Mark this readonly now that we expect it to be a fixed 1-byte array forever

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -36,38 +29,23 @@ public NamedPipeStreamReader(Stream stream)
         /// <returns>The message read from the stream, or null if the end of the input stream has been reached. </returns>
         public string ReadMessage()
         {
-            int bytesRead = this.Read();
-            if (bytesRead == 0)
+            byte readByte;
```
sanoursa: Nit: this sounds like the name of a delegate. Rename to `currentByte`?

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -76,17 +54,30 @@ public string ReadMessage()
...
 
         /// <summary>
-        /// Read the next chunk of bytes from the stream.
+        /// Read a byte from the stream.
         /// </summary>
-        /// <returns>The number of bytes read.</returns>
-        private int Read()
+        /// <param name="readByte">The byte read from the stream</param>
+        /// <returns>True if byte read, false if end of stream has been reached</returns>
+        private bool ReadByte(out byte readByte)
```
sanoursa: This method follows the Try pattern. Rename it to `TryReadByte`

GVFS/GVFS.Common/NamedPipes/NamedPipeStreamReader.cs
```diff
@@ -76,17 +54,30 @@ public string ReadMessage()
...
         {
-            return this.stream.Read(this.buffer, 0, this.buffer.Length);
+            int numBytesRead = this.stream.Read(this.buffer, 0, 1);
+
+            if (numBytesRead > 0)
```
sanoursa: I think you can drop this condition and always execute line 73. The caller won't use the value if we return false, so there's no point in doing the check twice.

jeschu1:
I'm just curious why we don't specify a size at the start of the message.  That should make it simple to read and performant.

jamill:
> I'm just curious why we don't specify a size at the start of the message.

I think that would be another fine option. I didn't go with it because it would require more changes to different components - we have native and managed components that communicate over the named pipe. 

The original problem was that the existing design did not handle messages with new line characters - performance was not a problem. The thought was that it would be better to just switch out the byte that represented the end of a message, rather than deeper changes to the protocol.

sanoursa:
@jamill thanks for the perf numbers. In almost all cases, it actually appears that the simple solution is faster than the complex one. Definitely looks good to go.

Issue #243, opened 2018-09-05
Mac: ReadFilesJob references kernel32


turbonaitis:
ReadFilesJob references kernel32.dll in attempt to do native file open/read, which results in an exception being thrown during `gvfs prefetch --files *.txt --hydrate <enlistment>`

sanoursa:
This is still a general issue because we have not yet scrubbed all uses of kernel32. Good to keep track of this instance though.

turbonaitis:
I guess these particular uses should go to `IPlatformFileSystem` ? I may create a PR for this next week, to unblock one of the scenarios we're trying to go through.

sanoursa:
Yes, exactly. If I recall correctly, the only reason we're calling kernel32 functions here is that the .NET APIs always read a full page even if you request a single byte, so calling `ReadFile` directly was a bit more efficient. So as a short-term hack to test your scenario, you could just drop the kernel32 code and use `FileStream`, taking care to open for read access only (opening for write will immediately convert the file from placeholder to full, even if you never write to it). If you want to go for the proper fix, then yes I would add a method called `ReadOneByte` to `IPlatformFileSystem`.

PR #345, opened 2018-10-04
Mac: Unblock 'gvfs prefetch --hydrate'

nickgra:
Unblock 'gvfs prefetch --hydrate' by going for a cheap, inefficient .NET implementation that will force a hydration on each prefetched file. This should be replaced with an efficient native version down the line. I turned the tests for this behavior on while I was here.

This 'fixes' #243 by implementing a quick .NET solution, but we'll keep the issue open to track the real fix.

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
+++ b/GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
@@ -60,11 +60,8 @@ namespace GVFS.Common.Prefetch.Jobs
                         {
                             using (FileStream filestream = new FileStream(path, FileMode.Open, FileAccess.Read))
                             {
-                                int thisByte = filestream.ReadByte();
-                                if (thisByte != -1)
-                                {
-                                    succeeded = true;
-                                }
+                                filestream.ReadByte();
+                                succeeded = true;
                             }
                         }
 
--- a/GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
+++ b/GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
@@ -44,11 +44,27 @@ namespace GVFS.Common.Prefetch.Jobs
                     foreach (string path in this.blobIdToPaths[blobId])
                     {
                         bool succeeded = false;
-                        using (SafeFileHandle handle = NativeFileReader.Open(path))
+
+                        // TODO(Mac): Replace this with a native method as opposed to a slow .NET method
+                        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                         {
-                            if (!handle.IsInvalid)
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
```diff
@@ -44,11 +44,24 @@ protected override void DoWork()
...
+                            }
+                        }
+                        else
                         {
-                            if (!handle.IsInvalid)
+                            using (FileStream filestream = new FileStream(path, FileMode.Open, FileAccess.Read))
                             {
-                                succeeded = NativeFileReader.ReadOneByte(handle, buffer);
+                                filestream.ReadByte();
+                                succeeded = true;
```
alameenshah: you are assuming ReadByte always succeeds.
Something like - ```succeeded = -1 != filestream.ReadByte();``` would be more correct.

sanoursa: -1 is not an error here, it just indicates end of file. I believe you would get that if this is a 0-byte file, and it would still be a success case.

GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
```diff
@@ -44,11 +44,24 @@ protected override void DoWork()
...
+                                if (!handle.IsInvalid)
+                                {
+                                    succeeded = NativeFileReader.ReadOneByte(handle, buffer);
+                                }
+                            }
+                        }
+                        else
                         {
-                            if (!handle.IsInvalid)
+                            using (FileStream filestream = new FileStream(path, FileMode.Open, FileAccess.Read))
```
sanoursa: That said, this code can throw IOExceptions, and we need to catch those and report them as errors on the individual files, not fail the whole prefetch operation.

GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
```diff
@@ -44,11 +44,24 @@ protected override void DoWork()
                     foreach (string path in this.blobIdToPaths[blobId])
                     {
                         bool succeeded = false;
-                        using (SafeFileHandle handle = NativeFileReader.Open(path))
+
+                        // TODO(Mac): Replace this with a native method as opposed to a slow .NET method
```
sanoursa: Nit: this comment is out of place. Either move it into the Mac else branch, or even just drop it. The `IsUnderConstruction` flag already serves as a TODO reminder.

wilbaker:
@nickgra, with these changes can these functional tests be enabled on Mac?

- `PrefetchByFileExtensionWithHydrate`
- `PrefetchByFilesWithHydrateWhoseObjectsAreAlreadyDownloaded`

nickgra:
Abandoning this version in favor of a native implementation.

PR #346, opened 2018-10-05
Update Readme.md (Typo)

priyankamadhwal:


sanoursa:
Thank you for the suggestion but I don't think that was a typo :-)

PR #349, opened 2018-10-05
UpgradeVerb cleanups for release

alameenshah:
- Removed wildcard matching while searching for downloaded installers
- Updated FileBasedDictionary to support custom Key Comparators to enable case-insensitive lookups
- Fixed partial failure in build caused by UninstallGVFS.bat script.
- Updated TryGetVersion in GitProcess to accept git path as input argument.

### DIFF
--- a/GVFS/GVFS.Common/FileBasedDictionary.cs
+++ b/GVFS/GVFS.Common/FileBasedDictionary.cs
@@ -15,10 +15,10 @@ namespace GVFS.Common
             ITracer tracer,
             PhysicalFileSystem fileSystem, 
             string dataFilePath, 
-            IEqualityComparer<TKey> keyComparer = null) 
+            IEqualityComparer<TKey> keyComparer) 
             : base(tracer, fileSystem, dataFilePath, collectionAppendsDirectlyToFile: false)
         {
-            this.data = new ConcurrentDictionary<TKey, TValue>(keyComparer ?? EqualityComparer<TKey>.Default);
+            this.data = new ConcurrentDictionary<TKey, TValue>(keyComparer);
         }
 
         public static bool TryCreate(
@@ -29,7 +29,12 @@ namespace GVFS.Common
             out string error,
             IEqualityComparer<TKey> keyComparer = null)
         {
-            output = new FileBasedDictionary<TKey, TValue>(tracer, fileSystem, dictionaryPath, keyComparer);
+            output = new FileBasedDictionary<TKey, TValue>(
+                tracer, 
+                fileSystem, 
+                dictionaryPath, 
+                keyComparer ?? EqualityComparer<TKey>.Default);
+
             if (!output.TryLoadFromDisk<TKey, TValue>(
                 output.TryParseAddLine,
                 output.TryParseRemoveLine,
--- a/GVFS/GVFS.Common/FileBasedDictionary.cs
+++ b/GVFS/GVFS.Common/FileBasedDictionary.cs
@@ -11,14 +11,28 @@ namespace GVFS.Common
     {
         private ConcurrentDictionary<TKey, TValue> data = new ConcurrentDictionary<TKey, TValue>();
 
-        private FileBasedDictionary(ITracer tracer, PhysicalFileSystem fileSystem, string dataFilePath) 
+        private FileBasedDictionary(
+            ITracer tracer,
+            PhysicalFileSystem fileSystem, 
+            string dataFilePath, 
+            IEqualityComparer<TKey> keyComparer = null) 
             : base(tracer, fileSystem, dataFilePath, collectionAppendsDirectlyToFile: false)
         {
+            if (keyComparer != null)
+            {
+                this.data = new ConcurrentDictionary<TKey, TValue>(keyComparer);
+            }
         }
 
-        public static bool TryCreate(ITracer tracer, string dictionaryPath, PhysicalFileSystem fileSystem, out FileBasedDictionary<TKey, TValue> output, out string error)
+        public static bool TryCreate(
+            ITracer tracer, 
+            string dictionaryPath, 
+            PhysicalFileSystem fileSystem,
+            out FileBasedDictionary<TKey, TValue> output, 
+            out string error,
+            IEqualityComparer<TKey> keyComparer = null)
         {
-            output = new FileBasedDictionary<TKey, TValue>(tracer, fileSystem, dictionaryPath);
+            output = new FileBasedDictionary<TKey, TValue>(tracer, fileSystem, dictionaryPath, keyComparer);
             if (!output.TryLoadFromDisk<TKey, TValue>(
                 output.TryParseAddLine,
                 output.TryParseRemoveLine,
--- a/GVFS/GVFS.Common/FileBasedDictionary.cs
+++ b/GVFS/GVFS.Common/FileBasedDictionary.cs
@@ -9,7 +9,7 @@ namespace GVFS.Common
 {
     public class FileBasedDictionary<TKey, TValue> : FileBasedCollection
     {
-        private ConcurrentDictionary<TKey, TValue> data = new ConcurrentDictionary<TKey, TValue>();
+        private ConcurrentDictionary<TKey, TValue> data;
 
         private FileBasedDictionary(
             ITracer tracer,
@@ -18,10 +18,7 @@ namespace GVFS.Common
             IEqualityComparer<TKey> keyComparer = null) 
             : base(tracer, fileSystem, dataFilePath, collectionAppendsDirectlyToFile: false)
         {
-            if (keyComparer != null)
-            {
-                this.data = new ConcurrentDictionary<TKey, TValue>(keyComparer);
-            }
+            this.data = new ConcurrentDictionary<TKey, TValue>(keyComparer ?? EqualityComparer<TKey>.Default);
         }
 
         public static bool TryCreate(
--- a/GVFS/GVFS.Common/LocalGVFSConfig.cs
+++ b/GVFS/GVFS.Common/LocalGVFSConfig.cs
@@ -1,5 +1,6 @@
 ﻿using GVFS.Common.FileSystem;
 using GVFS.Common.Tracing;
+using System;
 using System.IO;
 
 namespace GVFS.Common
@@ -35,7 +36,7 @@ namespace GVFS.Common
                         
             try
             {
-                this.allSettings.TryGetValue(this.KeyFromConfigName(name), out value);
+                this.allSettings.TryGetValue(name, out value);
                 error = null;
                 return true;
             }
@@ -67,7 +68,7 @@ namespace GVFS.Common
 
             try
             {
-                this.allSettings.SetValueAndFlush(this.KeyFromConfigName(name), value);
+                this.allSettings.SetValueAndFlush(name, value);
                 error = null;
                 return true;
             }
@@ -85,11 +86,6 @@ namespace GVFS.Common
             }
         }
 
-        private string KeyFromConfigName(string configName)
-        {
-            return configName.ToUpperInvariant();
-        }
-
         private bool TryLoadSettings(ITracer tracer, out string error)
         {
             if (this.allSettings == null)
@@ -100,7 +96,8 @@ namespace GVFS.Common
                     dictionaryPath: this.configFile,
                     fileSystem: this.fileSystem,
                     output: out config,
-                    error: out error))
+                    error: out error,
+                    keyComparer: StringComparer.OrdinalIgnoreCase))
                 {
                     this.allSettings = config;
                     return true;
--- a/GVFS/GVFS.Common/LocalGVFSConfig.cs
+++ b/GVFS/GVFS.Common/LocalGVFSConfig.cs
@@ -21,21 +21,21 @@ namespace GVFS.Common
         }
 
         public bool TryGetConfig(
-            string key, 
+            string name, 
             out string value, 
             out string error,
             ITracer tracer)
         {
             if (!this.TryLoadSettings(tracer, out error))
             {
-                error = $"Error getting config value {key}. {error}";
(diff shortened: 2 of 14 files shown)


GVFS/GVFS.Common/LocalGVFSConfig.cs
```diff
@@ -76,15 +76,20 @@ public LocalGVFSConfig()
...
-                error = string.Format(ErrorFormat, key, value, this.configFile, exception.Message);
+                error = string.Format(ErrorFormat, name, value, this.configFile, exception.Message);
                 value = null;
                 return false;
             }
         }
 
+        private string KeyFromConfigName(string configName)
+        {
+            return configName.ToUpperInvariant();
```
sanoursa: I don't think we should change what the user is telling us to store, if we can avoid it. What are your thoughts on allowing for case-insensitive lookups in the dictionary?

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -19,12 +18,25 @@ public static bool IsLocalUpgradeAvailable()
...
-                    PotentialInstallerName, 
-                    SearchOption.TopDirectoryOnly);
-                return installers.Length > 0;
+                // This method is used Only by Git hooks. Git hooks does not have access
+                // to GVFSPlatform to read platform specific file extensions. That is the
+                // reason possible installer file extensions are defined here.
+                HashSet<string> extensions = new HashSet<string>() { "EXE", "DMG", "RPM", "DEB" };
+                HashSet<string> installerNames = new HashSet<string>()
+                {
+                    GVFSInstallerFileNamePrefix.ToUpperInvariant(),
```
sanoursa: You don't need to change any of these strings. Just use case-insensitive lookups in the HashSet.

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -19,12 +18,25 @@ public static bool IsLocalUpgradeAvailable()
...
             if (Directory.Exists(downloadDirectory))
             {
-                const string PotentialInstallerName = "*VFS*.*";
-                string[] installers = Directory.GetFiles(
-                    downloadDirectory,
-                    PotentialInstallerName, 
-                    SearchOption.TopDirectoryOnly);
-                return installers.Length > 0;
+                // This method is used Only by Git hooks. Git hooks does not have access
+                // to GVFSPlatform to read platform specific file extensions. That is the
```
sanoursa: This comment is not true, the hooks access parts of GVFSPlatform already today.

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -19,12 +18,25 @@ public static bool IsLocalUpgradeAvailable()
...
+                // reason possible installer file extensions are defined here.
+                HashSet<string> extensions = new HashSet<string>() { "EXE", "DMG", "RPM", "DEB" };
```
sanoursa: I don't think we should allow all possible extensions here, we should make this specific to each platform

GVFS/GVFS.Common/LocalGVFSConfig.cs
alameenshah: Have updated the dictionary to accept a case insensitive comparer. Removed all ToUpper() calls. Is this similar to what you had in mind?

GVFS/GVFS.Common/FileBasedDictionary.cs
```diff
@@ -11,14 +11,28 @@ public class FileBasedDictionary<TKey, TValue> : FileBasedCollection
...
+        private FileBasedDictionary(
+            ITracer tracer,
+            PhysicalFileSystem fileSystem, 
+            string dataFilePath, 
+            IEqualityComparer<TKey> keyComparer = null) 
             : base(tracer, fileSystem, dataFilePath, collectionAppendsDirectlyToFile: false)
         {
+            if (keyComparer != null)
+            {
+                this.data = new ConcurrentDictionary<TKey, TValue>(keyComparer);
```
wilbaker: With this change `this.data` will be initialized twice.  Should we remove the creation of the `ConcurrentDictionary` here:

```private ConcurrentDictionary<TKey, TValue> data = new ConcurrentDictionary<TKey, TValue>();```

And always create it once in the `FileBasedDictionary` constructor?

File: GVFS/GVFS.Common/ProductUpgrader.Shared.cs (hunk 3 above)
Comment (alameenshah) [reply GVFS/GVFS.Common/ProductUpgrader.Shared.cs]: Ya... removed the comment.

File: GVFS/GVFS.Common/FileBasedDictionary.cs [hunk 6]
```diff
@@ -9,16 +9,27 @@ namespace GVFS.Common
...
     {
-        private ConcurrentDictionary<TKey, TValue> data = new ConcurrentDictionary<TKey, TValue>();
+        private ConcurrentDictionary<TKey, TValue> data;
 
-        private FileBasedDictionary(ITracer tracer, PhysicalFileSystem fileSystem, string dataFilePath) 
+        private FileBasedDictionary(
+            ITracer tracer,
+            PhysicalFileSystem fileSystem, 
+            string dataFilePath, 
+            IEqualityComparer<TKey> keyComparer = null)
```
sanoursa: This is a private constructor, so do we actually need a default value for this parameter?

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -14,17 +14,27 @@ public partial class ProductUpgrader
...
             string downloadDirectory = GetAssetDownloadsPath();
             if (Directory.Exists(downloadDirectory))
             {
-                const string PotentialInstallerName = "*VFS*.*";
-                string[] installers = Directory.GetFiles(
-                    downloadDirectory,
-                    PotentialInstallerName, 
-                    SearchOption.TopDirectoryOnly);
-                return installers.Length > 0;
+                HashSet<string> extensions = new HashSet<string>(installerExtensions, StringComparer.OrdinalIgnoreCase);
```
sanoursa: Both of the platform implementations currently have a single extension, so do we need to construct a new HashSet for this every time we run the hooks?

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -14,17 +14,27 @@ public partial class ProductUpgrader
...
+                HashSet<string> installerNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
sanoursa: Same here, why do we need a hashset for this? With the current logic, this will return true if there is a git installer but not a GVFS installer. Is that desired?

If we're ok with partial results, then I would make this a single check for the GVFS installer name.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -50,9 +50,10 @@ protected override void Execute(GVFSEnlistment enlistment)
                 this.WriteMessage(string.Empty);
                 this.WriteMessage("gvfs version " + ProcessHelper.GetCurrentProcessVersion());
 
-                GitVersion gitVersion;
-                string error;
-                if (GitProcess.TryGetVersion(out gitVersion, out error))
+                GitVersion gitVersion = null;
+                string error = null;
+                string gitPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+                if (!string.IsNullOrEmpty(gitPath) && GitProcess.TryGetVersion(gitPath, out gitVersion, out error))
```
sanoursa: Use `out string _` like you did in GVFSVerb?

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -50,9 +50,10 @@ protected override void Execute(GVFSEnlistment enlistment)
                 this.WriteMessage(string.Empty);
                 this.WriteMessage("gvfs version " + ProcessHelper.GetCurrentProcessVersion());
 
-                GitVersion gitVersion;
-                string error;
-                if (GitProcess.TryGetVersion(out gitVersion, out error))
+                GitVersion gitVersion = null;
+                string error = null;
+                string gitPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
```
sanoursa: This method is being called again a few lines later. Since you're already storing it in a variable, use the variable on line 65 too

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -619,8 +619,9 @@ private void CheckFileSystemSupportsRequiredFeatures(ITracer tracer, Enlistment
 
         private void CheckGitVersion(ITracer tracer, GVFSEnlistment enlistment, out string version)
         {
-            GitVersion gitVersion;
-            if (!GitProcess.TryGetVersion(out gitVersion, out string _))
+            string gitPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
```
sanoursa: The `enlistment` already has this path, there's no need to call this method again

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -414,9 +414,8 @@ private void LogInstalledVersionInfo()
 
             GitVersion installedGitVersion = null;
             string error = null;
-            if (GitProcess.TryGetVersion(
-                out installedGitVersion,
-                out error))
+            string gitPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
```
sanoursa: Does the caller already have this path?

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
alameenshah: Current logic- Check for the files "SetupGVFS.<version>.exe" or "VFSForGit.<version>.exe". Return true if one of these GVFS installers is found. Both these files are GVFS installers (in the new and old name format), we don't check for Git installer.

We need a Collection class (HashSet or List) only because of the need to support both current and new GVFS installer names in future.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
alameenshah: we are writing the error here to console, in case TryGetVersion fails.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
alameenshah: updated line 65 to read the git path from enlistment.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: no caller does not.

GVFS/GVFS.Common/FileBasedDictionary.cs
alameenshah: Right. we don't need it. Updated.

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
alameenshah: Right. Have updated both platforms to return ```string``` rather than an ```array``` for ```InstallerExtension```. Replaced HashSet with string.

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
```diff
@@ -4,6 +4,11 @@ namespace GVFS.Hooks.HooksPlatform
 {
     public static class GVFSHooksPlatform
     {
+        public static string InstallerExtension()
```
sanoursa: Nit: either rename to `GetInstallerExtension` or change to a property with this same name

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
```diff
@@ -5,6 +5,8 @@ namespace GVFS.Platform.Mac
 {
     public partial class MacPlatform
     {
+        public static readonly string InstallerExtension = "dmg";
```
sanoursa: Nit: Replace `static readonly` with `const`

GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs
```diff
@@ -8,6 +8,8 @@ namespace GVFS.Platform.Windows
 {
     public partial class WindowsPlatform
     {
+        public static readonly string InstallerExtension = "exe";
```
sanoursa: const

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -50,9 +50,9 @@ protected override void Execute(GVFSEnlistment enlistment)
                 this.WriteMessage(string.Empty);
                 this.WriteMessage("gvfs version " + ProcessHelper.GetCurrentProcessVersion());
 
-                GitVersion gitVersion;
-                string error;
-                if (GitProcess.TryGetVersion(out gitVersion, out error))
+                GitVersion gitVersion = null;
+                string error = null;
+                if (!string.IsNullOrEmpty(enlistment.GitBinPath) && GitProcess.TryGetVersion(enlistment.GitBinPath, out gitVersion, out error))
```
sanoursa: Just to double check - is enlistment.GitBinPath always guaranteed to be initialized properly for `DiagnoseVerb`?

alameenshah: Yes - GVFSVerb.CreateEnlistment sets it. When Git is not installed, gvfs diagnose will early exit.

PR #350, opened 2018-10-05
Windows: Add delay and retry to attributes check in ExpandedFileAttributesAreUpdated

### LINKED ISSUE #348: Add retry logic to BasicFileSystemTests.ExpandedFileAttributesAreUpdated test
Sometimes `ExpandedFileAttributesAreUpdated` fails because the following attributes are also set:

```
#define FILE_ATTRIBUTE_SPARSE_FILE          0x00000200  // winnt
#define FILE_ATTRIBUTE_REPARSE_POINT        0x00000400  // winnt 
#define FILE_ATTRIBUTE_RECALL_ON_OPEN       0x00040000  // winnt 
```

Per the ProjFS team there is currently a race where these still might be set on the file after it's expanded to full.  Rather than failing immediately, the test should have retry\sleep logic when checking the file attributes.

wilbaker:
Fixes #348.

Added sleep\retry logic to `ExpandedFileAttributesAreUpdated` because the ProjFS related attributes are not removed synchronously when a file is converted to full.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
@@ -171,7 +171,9 @@ namespace GVFS.FunctionalTests.Tests.LongRunningEnlistment
                 
                 ++retryCount;
                 Thread.Sleep(500);
-                attributes = new FileInfo(virtualFile).Attributes & ~FileAttributes.Archive;
+
+                info.Refresh();
+                attributes = info.Attributes & ~FileAttributes.Archive;
             }
                    
             attributes.ShouldEqual(FileAttributes.Hidden, $"Attributes do not match, expected: {FileAttributes.Hidden} actual: {attributes}");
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
```diff
@@ -8,12 +8,17 @@
...
+using System.Threading;
 
 namespace GVFS.FunctionalTests.Tests.LongRunningEnlistment
 {
     [TestFixture]
     public class BasicFileSystemTests : TestsWithEnlistmentPerFixture
     {
+        private const int FileAttributeSparseFile = 0x00000200;
+        private const int FileAttributeReparsePoint = 0x00000400;
+        private const int FileAttributeRecallOnOpen = 0x00040000;
```
jrbriggs: As per Abhijeet's email, should this be `FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS` (`0x00400000`) ?

wilbaker: Yep thanks!

PR #353, opened 2018-10-08
Make GVFS.Upgrader rely on GVFS.cs.props to fix incremental build issue

jrbriggs:
Also eliminates the need to manually specify the OutputPath in the Upgrader csproj -- it will inherit this from GVFS.cs.props.


alameenshah:
Did build and do some basic testing with this change. Looking Good.

PR #357, opened 2018-10-08
Mac: New folder notifications

### LINKED ISSUE #197: Mac: Notify VFS for Git when new directories are created
#152 does not include a notification when new directories are created.

Make sure new directory scenarios work as expected.

### LINKED ISSUE #264: Mac: Folders created by git are not expanded after git command completes
This issue can be reproduced by running the `CheckoutTests.DeleteFileThenCheckout` functional test.

The test fails because when `TestValidationAndCleanup` runs the repo is in this state:

- `GitCommandsTests/DeleteFileTests/1/#test` is in ModifiedPaths.Dat
- `GitCommandsTests/DeleteFileTests` is not on disk

Then `TestValidationAndCleanup` checks out the control commit (which does have `GitCommandsTests/DeleteFileTests`, and it has several child folders).

When the checkout occurs git creates `DeleteFileTests`, `DeleteFileTests/1`, and `DeleteFileTests/1/#test`.  When git creates the folders (unlike in VFSForGit on Windows) the folders are not marked as empty (or recorded as needing to be re-expanded).

As a result, when the command completes `DeleteFileTests` (and `DeleteFileTests/1`) are never expanded again and all of their children (other than `#test`) are missing.

wilbaker:
Resolves #197
Resolves #264 

The ProjFS kext is unable to raise notifications when new folders are created, however, notifications are raised when new files are created within new folders.  With the changes in this PR the user-mode library will check for untracked folders whenever files are create (or renamed\hard-linked) and raise a notification to VFS for Git when previously unknown folders are found.

`git status` ignores new folders until they have a new file within them, and so these changes take advantage of that behavior.

For issue #264 VFS for Git has been updated to add folders created by git as "expanded folders" so that VFS for Git will re-expand them during projection changes.

The PR also includes some cleanup of the Mac functional test categories (that are not yet passing).

The following scenario is not covered by this PR:

- User creating a new folder and checking out a branch that has a file or folder with the same name (#360)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -608,16 +610,40 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
+                
+                char parentFullPath[PrjFSMaxPath];
+                CombinePaths(s_virtualizationRootFullPath.c_str(), parentPath.c_str(), parentFullPath);
+                
+                if (!IsBitSetInFileFlags(parentFullPath, FileFlags_IsInVirtualizationRoot))
+                {
+                    // TODO(Mac): Handle SetBitInFileFlags failures
+                    SetBitInFileFlags(parentFullPath, FileFlags_IsInVirtualizationRoot, true);
+                    
+                    HandleFileNotification(
```
wilbaker: We need to walk up the tree and notify VFS for Git of all new folders it has not seen yet.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -608,16 +614,54 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
+                {
+                    newFolderPaths.emplace(make_pair(parentPath, parentFullPath));
+                    lastDirSeparator = parentPath.find_last_of('/');
+                }
+            }
+            
+            while (!newFolderPaths.empty())
+            {
+                const pair<string /*relative path*/, string /*full path*/>& parentFolderPath = newFolderPaths.top();
+                newFolderPaths.pop();
```
wilbaker: Don't pop until the bottom of the code block when we're done using `parentFolderPath`

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -525,17 +528,26 @@ private void OnNewFileCreated(string relativePath, bool isDirectory)
...
-                        GitCommandLineParser gitCommand = new GitCommandLineParser(this.Context.Repository.GVFSLock.GetLockedGitCommand());
-                        if (gitCommand.IsValidGitCommand)
-                        {
-                            // TODO(Mac): Ensure that when git creates a folder all files\folders within that folder are written to disk
-                            EventMetadata metadata = this.CreateEventMetadata(relativePath);
-                            metadata.Add("isDirectory", isDirectory);
-                            this.Context.Tracer.RelatedWarning(metadata, $"{nameof(this.OnNewFileCreated)}: Git created a folder, currently an unsupported scenario on Mac");
-                        }
-                        else
+                        if (!relativePath.Equals(GVFSConstants.DotGit.Root, StringComparison.OrdinalIgnoreCase))
```
sanoursa: Isn't `relativePath` the name of the new file that was created? How could it be equal to `.git`?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -525,17 +528,26 @@ private void OnNewFileCreated(string relativePath, bool isDirectory)
...
+                            GitCommandLineParser gitCommand = new GitCommandLineParser(lockedGitCommand);
+                            if (gitCommand.IsValidGitCommand)
+                            {
+                                EventMetadata metadata = this.CreateEventMetadata(relativePath);
+                                metadata.Add(nameof(lockedGitCommand), lockedGitCommand);
+                                metadata.Add(TracingConstants.MessageKey.InfoMessage, "Git command created new folder");
+                                this.Context.Tracer.RelatedEvent(EventLevel.Informational, $"{nameof(this.OnNewFileCreated)}_GitCreatedFolder", metadata);
+
+                                // Record this folder as expanded so that GitIndexProjection will re-expand the folder
+                                // when the projection change completes.  This is the closest we can get to converting this
```
sanoursa: Why is this the closest we can get? Why can't we just mark the folder as empty again?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -608,16 +614,55 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
         case MessageType_KtoU_NotifyDirectoryRenamed:
         case MessageType_KtoU_NotifyFileHardLinkCreated:
         {
+            // Walk up the directory tree and notify the provider about any directories
```
sanoursa: I think we should keep the `HandleKernelRequest` function as pure routing code, and move this logic down into the other function that handles this type of message.

sanoursa: Or a separate function for all of this logic here, since it's actually generating more callbacks to the provider

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -838,6 +883,7 @@ static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const
 static PrjFS_Result HandleFileNotification(
     const MessageHeader* request,
     const char* path,
+    const char* fullPath_opt,
```
sanoursa: What is `_opt`? Optional?

sanoursa: And does it need to be optional?

wilbaker: > What is _opt? Optional?

Yep, and you're right we can just make this required.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
sanoursa: I don't think we need a new API. `PrjFS_ConvertDirectoryToVirtualizationRoot` could enumerate the root directory and mark everything that's already present as being in the root. Given that we're becoming more and more reliant on that flag being accurate, this seems like a good change to make.

Also this code as written is somewhat confusing, because we're inside an if block that just checked `if (!Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath))` - if we're not in `.git`, then how could it be `.git`? I get why, the `IsPathInsideDotGit` looks for `.git\`, not `.git`, but it ends up being hard to follow.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
sanoursa: I wasn't asking for an alternate approach, just pointing out that the comment is making a false claim :-). Marking the folder as expanded seems like a fine approach.

And yea, with the code as it is _today_, marking the folder empty will likely fail, because we will fail when creating a duplicate placeholder. But that will change soon with @nickgra 's PR for atomic placeholder creation.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
wilbaker: Both good pieces of feedback, if the first suggestion works out I think we can stick with `IsPathInsideDotGit` being limited to checking `.git\` for now

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
wilbaker: > I wasn't asking for an alternate approach, just pointing out that the comment is making a false claim :-).

Ok thanks :) I'll update this comment.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -866,6 +902,47 @@ static PrjFS_Result HandleFileNotification(
...
+    // not flagged as being in the root
+    stack<pair<string /*relative path*/, string /*full path*/>> newFolderPaths;
+    string parentPath(path);
+    size_t lastDirSeparator = parentPath.find_last_of('/');
+    while (lastDirSeparator != string::npos && lastDirSeparator > 0)
+    {
+        parentPath = parentPath.substr(0, lastDirSeparator);
+        char parentFullPath[PrjFSMaxPath];
+        CombinePaths(s_virtualizationRootFullPath.c_str(), parentPath.c_str(), parentFullPath);
+        if (IsBitSetInFileFlags(parentFullPath, FileFlags_IsInVirtualizationRoot))
```
sanoursa: Does it matter that we're setting the bit before we notified the provider? What happens if we lose power between here and there? Would it be more robust if we notify the provider first?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -835,9 +850,33 @@ static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const
...
+    bool isDirectory,
+    PrjFS_NotificationType notificationType)
+{
+    // Whenever a new file shows up in the root, we need to check if its ancestor
+    // directories are flagged as in root.  If they are not, flag them as in root and
+    // notify the provider
+    FindNewFoldersInRootAndNotifyProvider(request, path);
+    
+    // TODO(Mac): Handle SetBitInFileFlags failures
+    SetBitInFileFlags(fullPath, FileFlags_IsInVirtualizationRoot, true);
```
sanoursa: Same here, though this one is not new in this PR. Should we notify first, and set the bit second?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -525,17 +528,22 @@ private void OnNewFileCreated(string relativePath, bool isDirectory)
...
                             EventMetadata metadata = this.CreateEventMetadata(relativePath);
-                            metadata.Add("isDirectory", isDirectory);
-                            this.Context.Tracer.RelatedWarning(metadata, $"{nameof(this.OnNewFileCreated)}: Git created a folder, currently an unsupported scenario on Mac");
-                        }
-                        else
+                            metadata.Add(nameof(lockedGitCommand), lockedGitCommand);
+                            metadata.Add(TracingConstants.MessageKey.InfoMessage, "Git command created new folder");
+                            this.Context.Tracer.RelatedEvent(EventLevel.Informational, $"{nameof(this.OnNewFileCreated)}_GitCreatedFolder", metadata);
+
+                            // Record this folder as expanded so that GitIndexProjection will re-expand the folder
```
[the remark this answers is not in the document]
wilbaker: > Some comments about the why would be useful here.

Sure will do

> E.g. we're doing this because a new file was created by a git command, and it's inside a folder that may need to project other items after the git command is completed. (Did I get that right?)

You did get that right.  The issue is that when git creates a folder it will only create files in that folder that are in ModifiedPaths.dat.  That's an issue because there could be other files in the folder (that are not in ModifiedPaths.dat) that VFS needs to create placeholders for.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -866,6 +902,47 @@ static PrjFS_Result HandleFileNotification(
...
+    string parentPath(path);
+    size_t lastDirSeparator = parentPath.find_last_of('/');
+    while (lastDirSeparator != string::npos && lastDirSeparator > 0)
+    {
+        parentPath = parentPath.substr(0, lastDirSeparator);
+        char parentFullPath[PrjFSMaxPath];
+        CombinePaths(s_virtualizationRootFullPath.c_str(), parentPath.c_str(), parentFullPath);
+        if (IsBitSetInFileFlags(parentFullPath, FileFlags_IsInVirtualizationRoot))
+        {
+            break;
```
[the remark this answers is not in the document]
wilbaker: Filed #390 for this

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: > make sure we cover this scenario in the follow up robustness issue.

I filed #391 for this

PR #359, opened 2018-10-09
RFC: Enhance automount to wait for volume to be available

mjcheetham:
_**Not ready for merge. Just a draft/RFC to get some initial feedback on the approach.**_

_(there are still some TODOs in the code, and a lack of tests! These will follow later :smile:)_

Enhance the auto-mounting logic to watch for new volumes to be attached (such as external drives), and to wait for BitLocker drives to be unlocked before mounting.

This changes the notion of 'auto mount' from a single action that runs on logon to an on-going process of mounting repositories as they become 'available'. At first logon we still attempt to mount all registered repositories as before.

I've used the term 'available'/'unavailable' in this change as it captures both the missing external drive case, as well as the encypted volume cases (such as BitLocker on Windows and any equivalents on macOS). I've used the term 'volume' rather than 'drive' as you can also mount drives as NTFS folder paths on Windows, and in macOS where volumes can be mounted anywhere under "/".

Added ability to list the state of all known repositories using the cmd: `gvfs service --list-all`, to compliment the existing `gvfs service --list-mounted` option which lists only the mounted repositories.

Open questions:
1. We can now detect when a volume is removed. Currently if you remove a volume the mount process continues to run and is reported as 'mounted' (this is the existing behaviour).
Should we try and kill the mount processes for repositories on volumes that have been removed/locked? Should the mount process itself be responsible for killing itself or should the service send a "kill" message?

    1.1. If we kill the process on `VolumeUnavilable`, and then you quickly lock and then unlock a volume (or plug/unplug an external drive) then do we want the removal to be considered a 'transient'; should we wait a fixed time (10 seconds) before killing the mount process?

    1.2. We can't send the 'unmount' message because this does two things we don't want: a) unregister the repository, b) read/write from disk! it's too late to unmount.. we can only 'die quickly'.

2. Currently if a repository does not exist and we're trying to mount it (at logon) we remove it from the repo-registry file (assuming it's been deleted). This might not be the case if a volume is locked, or has not been mounted yet. In this PR I've removed the 'remove repo on missing path' code, but this might lead to lots of redundant entries over time with no clean-up.
We could consider a heuristic that if the _volume_ of the repository does not exist then do NOT remove the registration, but if the volume _does_ exist and the enlistment does _not_ then it has likely been deleted and should be unregistered.

3. The only way to watch for BitLocker lock/unlock events is using the `ManagementEventWatcher`, which polls WMI. What polling interval should be used (I picked 10 seconds arbitrarily)? Too short and you end up using CPU. Too long and we increase the gap between the volume being unlocked and the call to `gvfs mount`.

4. [Future]: Users might want to select some repositories to not auto-mount. Currently they can do this by expliclty running `gvfs unmount` which sets "IsActive" to false in the registration - we don't ever try and auto-mount it.

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
@@ -8,5 +8,42 @@
         void CreateHardLink(string newLinkFileName, string existingFileName);
         bool TryGetNormalizedPath(string path, out string normalizedPath, out string errorMessage);
         void ChangeMode(string path, int mode);
+
+        /// <summary>
+        /// Check if a path exists as a subpath of the specified directory.
+        /// </summary>
+        /// <param name="directoryPath">Directory path.</param>
+        /// <param name="path">Path to query.</param>
+        /// <returns>True if <see cref="path"/> exists as a subpath of <see cref="directoryPath"/>, false otherwise.</returns>
+        bool IsPathUnderDirectory(string directoryPath, string path);
+
+        /// <summary>
+        /// Get the path to the volume root that the given path.
+        /// </summary>
+        /// <param name="path">File path to find the volume for.</param>
+        /// <returns>Path to the root of the volume.</returns>
+        string GetVolumeRoot(string path);
+
+        /// <summary>
+        /// Check if the volume for the given path is available and ready for use.
+        /// </summary>
+        /// <param name="path">Path to any directory or file on a volume.</param>
+        /// <remarks>
+        /// A volume might be unavailable for multiple reasons, including:
+        /// <para/>
+        ///  - the volume resides on a removable device which is not present
+        /// <para/>
+        ///  - the volume is not mounted in the operating system
+        /// <para/>
+        ///  - the volume is encrypted or locked
+        /// </remarks>
+        /// <returns>True if the volume is available, false otherwise.</returns>
+        bool IsVolumeAvailable(string path);
+
+        /// <summary>
+        /// Create an <see cref="IVolumeStateWatcher"/> which monitors for changes to the state of volumes on the system.
+        /// </summary>
+        /// <returns>Volume watcher</returns>
+        IVolumeStateWatcher CreateVolumeStateWatcher();
     }
 }
--- a/GVFS/GVFS.Common/FileSystem/IVolumeStateWatcher.cs
+++ b/GVFS/GVFS.Common/FileSystem/IVolumeStateWatcher.cs
@@ -1,0 +1,25 @@
+﻿using System;
+
+namespace GVFS.Common.FileSystem
+{
+    /// <summary>
+    /// Monitors the system for changes to the state of any disk volume, notifying subscribers when a change occurs.
+    /// </summary>
+    public interface IVolumeStateWatcher : IDisposable
+    {
+        /// <summary>
+        /// Raised when the state of a volume has changed, such as becoming available.
+        /// </summary>
+        event EventHandler<VolumeStateChangedEventArgs> VolumeStateChanged;
+
+        /// <summary>
+        /// Start watching for changes to the states of volumes on the system.
+        /// </summary>
+        void Start();
+
+        /// <summary>
+        /// Stop watching for changes to the states of volumes on the system.
+        /// </summary>
+        void Stop();
+    }
+}
--- a/GVFS/GVFS.Common/NativeMethods.cs
+++ b/GVFS/GVFS.Common/NativeMethods.cs
@@ -164,6 +164,13 @@ namespace GVFS.Common
             return DateTime.Now - uptime;
         }
 
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        public static extern bool GetVolumePathName(
+            string volumeName,
+            StringBuilder volumePathName,
+            uint bufferLength);
+
         [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
         private static extern bool MoveFileEx(
             string existingFileName,
--- a/GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
+++ b/GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using System.Linq;
+using System.Management;
+
+namespace GVFS.Platform.Windows
+{
+    internal static class BitLockerHelpers
+    {
+        public const string VolumeEncryptionNamespace = @"root\cimv2\security\MicrosoftVolumeEncryption";
+
+        public static bool TryGetVolumeLockStatus(string driveName, out bool isLocked)
+        {
+            string queryString = $"SELECT * FROM Win32_EncryptableVolume WHERE DriveLetter = '{driveName}'";
+
+            using (var searcher = new ManagementObjectSearcher(VolumeEncryptionNamespace, queryString))
+            {
+                ManagementObjectCollection results = searcher.Get();
+                ManagementObject mo = results.OfType<ManagementObject>().FirstOrDefault();
+                if (mo != null)
+                {
+                    // Check the protection status
+                    // ProtectionStatus == 0 means the drive is not protected by BitLocker and therefore cannot be locked
+                    if (mo.Properties["ProtectionStatus"].Value is uint protectedInt && protectedInt == 0)
+                    {
+                        isLocked = false;
+                        return true;
+                    }
+
+                    // Check the lock status
+                    // Lock status is not a property and must be retrieved by the GetLockStatus method
+                    var args = new object[1];
+                    if (TryInvokeMethod(mo, "GetLockStatus", args) && args[0] is uint lockedInt)
+                    {
+                        // Unlocked
+                        if (lockedInt == 0)
+                        {
+                            isLocked = false;
+                            return true;
+                        }
+
+                        // Locked
+                        if (lockedInt == 1)
+                        {
+                            isLocked = true;
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            isLocked = false;
+            return false;
+        }
+
+        private static bool TryInvokeMethod(ManagementObject mo, string methodName, object[] args)
+        {
+            try
+            {
+                return mo.InvokeMethod(methodName, args) is uint result && result == 0;
+            }
+            catch (Exception)
+            {
+                return false;
+            }
+        }
+    }
+}
--- a/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
+++ b/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
@@ -53,7 +53,7 @@ namespace GVFS.Platform.Windows
             try
             {
                 StringBuilder volumePathName = new StringBuilder(GVFSConstants.MaxPath);
-                if (!NativeMethods.GetVolumePathName(enlistmentRoot, volumePathName, GVFSConstants.MaxPath))
+                if (!Common.NativeMethods.GetVolumePathName(enlistmentRoot, volumePathName, GVFSConstants.MaxPath))
                 {
                     errorMessage = "Could not get volume path name";
                     tracer.RelatedError($"{nameof(TryAttach)}:{errorMessage}");
@@ -626,13 +626,6 @@ namespace GVFS.Platform.Windows
                 string instanceName,
                 uint createdInstanceNameLength = 0,
                 string createdInstanceName = null);
-
-            [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
-            [return: MarshalAs(UnmanagedType.Bool)]
-            public static extern bool GetVolumePathName(
-                string volumeName,
-                StringBuilder volumePathName,
-                uint bufferLength);
         }
     }
 }
--- a/GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
@@ -1,4 +1,6 @@
-﻿using GVFS.Common;
+﻿using System;
+using System.Text;
+using GVFS.Common;
 using GVFS.Common.FileSystem;
 
 namespace GVFS.Platform.Windows
@@ -27,6 +29,39 @@ namespace GVFS.Platform.Windows
 
         public void ChangeMode(string path, int mode)
         {
+        }
+
+        public bool IsPathUnderDirectory(string directoryPath, string path)
+        {
+            // TODO: Normalize paths
+            // We can't use the existing TryGetNormalizedPathImplementation method
+            // because it relies on actual calls to the disk to check if directories exist.
+            // This may be called with paths or volumes which do not actually exist.
+            return path.StartsWith(directoryPath, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public string GetVolumeRoot(string path)
+        {
+            var volumePathName = new StringBuilder(GVFSConstants.MaxPath);
+            if (NativeMethods.GetVolumePathName(path, volumePathName, GVFSConstants.MaxPath))
+            {
+                return volumePathName.ToString();
+            }
+
+            return null;
+        }
+
+        public bool IsVolumeAvailable(string path)
+        {
+            // No paths 'exist' on locked BitLocker volumes so it is sufficent to just
+            // check if the directory/file exists using the framework APIs.
+            return System.IO.Directory.Exists(path) || System.IO.File.Exists(path);
+        }
+
+        public IVolumeStateWatcher CreateVolumeStateWatcher()
+        {
+            // TODO: Extract the polling interval to a configuration value?
+            return new WindowsVolumeStateWatcher(TimeSpan.FromSeconds(15));
         }
 
         public bool TryGetNormalizedPath(string path, out string normalizedPath, out string errorMessage)
--- a/GVFS/GVFS.Platform.Windows/WindowsVolumeStateWatcher.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsVolumeStateWatcher.cs
@@ -1,0 +1,144 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Management;
+using GVFS.Common.FileSystem;
+
+namespace GVFS.Platform.Windows
+{
+    public class WindowsVolumeStateWatcher : IVolumeStateWatcher
+    {
+        private readonly ManagementEventWatcher volumeWatcher;
+        private readonly ManagementEventWatcher cryptoVolumeWatcher;
+        private readonly IDictionary<string, bool> cryptoVolumeLockStatuses =
+            new ConcurrentDictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
+
+        public WindowsVolumeStateWatcher(TimeSpan pollingInterval)
+        {
+            int intervalSeconds = (int)pollingInterval.TotalSeconds;
+
+            // Watch for mount and unmount of volumes
+            string volumeQuery = $"SELECT * FROM Win32_VolumeChangeEvent WITHIN {intervalSeconds}";
+            this.volumeWatcher = new ManagementEventWatcher(volumeQuery);
+            this.volumeWatcher.EventArrived += this.OnVolumeEvent;
+
+            // Watch for changes to BitLocker-protected volume states (unlock, lock, etc)
+            string cryptoVolumeQuery = $"SELECT * FROM __InstanceModificationEvent WITHIN {intervalSeconds} WHERE TargetInstance ISA 'Win32_EncryptableVolume'";
+            this.cryptoVolumeWatcher = new ManagementEventWatcher(BitLockerHelpers.VolumeEncryptionNamespace, cryptoVolumeQuery);
+            this.cryptoVolumeWatcher.EventArrived += this.OnCryptoVolumeEvent;
+        }
+
+        public event EventHandler<VolumeStateChangedEventArgs> VolumeStateChanged;
+
+        public void Start()
+        {
+            this.volumeWatcher.Start();
+            this.cryptoVolumeWatcher.Start();
+        }
+
+        public void Stop()
+        {
+            this.cryptoVolumeWatcher.Stop();
+            this.volumeWatcher.Stop();
+        }
+
+        public void Dispose()
+        {
+            this.Stop();
+
+            this.volumeWatcher.EventArrived -= this.OnVolumeEvent;
+            this.volumeWatcher.Dispose();
+
+            this.cryptoVolumeWatcher.EventArrived -= this.OnCryptoVolumeEvent;
+            this.cryptoVolumeWatcher.Dispose();
+        }
+
+        protected void RaiseVolumeStateChanged(string volumePath, VolumeStateChangeType changeType)
+        {
+            this.VolumeStateChanged?.Invoke(this, new VolumeStateChangedEventArgs(volumePath, changeType));
+        }
+
+        private void OnVolumeEvent(object sender, EventArrivedEventArgs e)
+        {
+            ManagementBaseObject mbo = e.NewEvent;
+
+            var driveName = (string)mbo["DriveName"];
+            var eventType = (ushort)mbo["EventType"];
+
+            string volumePath = $@"{driveName}\";
+
+            if (eventType == 2)
+            {
+                // Device Arrival
+                // Check if the volume is not left locked by BitLocker on mount
+                if (BitLockerHelpers.TryGetVolumeLockStatus(driveName, out bool isLocked) && !isLocked)
+                {
+                    this.RaiseVolumeStateChanged(volumePath, VolumeStateChangeType.VolumeAvailable);
+                }
+            }
+            else if (eventType == 3)
+            {
+                // Device Removal
+                this.RaiseVolumeStateChanged(volumePath, VolumeStateChangeType.VolumeUnavailable);
+            }
+        }
+
+        private void OnCryptoVolumeEvent(object sender, EventArrivedEventArgs e)
+        {
+            ManagementBaseObject mbo = e.NewEvent;
+            var targetMbo = (ManagementBaseObject)mbo["TargetInstance"];
+
+            var driveName = (string)targetMbo?["DriveLetter"];
+            if (driveName == null)
+            {
+                return;
+            }
+
+            string volumePath = $@"{driveName}\";
+
+            // Get the new lock status of the volume
+            // Note: we can only invoke the required "GetLockStatus" method on instances of type
+            // ManagementObject, but management events only return instances of ManagementBaseObject.
+            if (!BitLockerHelpers.TryGetVolumeLockStatus(driveName, out bool newLockStatus))
+            {
+                return;
+            }
+
+            // Get previous lock status
+            // We only want to raise the 'VolumeStateChanged' event if the lock status has changed, but this event could
+            // be raised for any modification to the BitLocker volume's configuration such as auto unlock on/off.
+            // We track the previously seen lock statuses for all volumes so that we can detect changes to the locked status
+            // and only raise the 'VolumeStateChanged' event in those cases.
+            if (this.cryptoVolumeLockStatuses.TryGetValue(driveName, out bool prevLockStatus))
+            {
+                if (prevLockStatus && !newLockStatus)
+                {
+                    // Locked -> Unlocked
+                    this.RaiseVolumeStateChanged(volumePath, VolumeStateChangeType.VolumeAvailable);
+                }
+                else if (!prevLockStatus && newLockStatus)
+                {
+                    // Unlocked -> Locked
+                    this.RaiseVolumeStateChanged(volumePath, VolumeStateChangeType.VolumeUnavailable);
+                }
+            }
+            else
+            {
+                // We've never seen this volume before (must have just been mounted).. report the current locked state as an event
+                if (!newLockStatus)
+                {
+                    // Unlocked
+                    this.RaiseVolumeStateChanged(volumePath, VolumeStateChangeType.VolumeAvailable);
+                }
+                else
+                {
+                    // Locked
+                    this.RaiseVolumeStateChanged(volumePath, VolumeStateChangeType.VolumeUnavailable);
+                }
+            }
+
+            // Update new lock status
+            this.cryptoVolumeLockStatuses[driveName] = newLockStatus;
+        }
+    }
+}
--- a/GVFS/GVFS.Service/GVFSMountProcessManager.cs
+++ b/GVFS/GVFS.Service/GVFSMountProcessManager.cs
@@ -6,13 +6,13 @@ using System;
 
 namespace GVFS.Service
 {
-    public class GVFSMountProcess : IDisposable
+    public class GVFSMountProcessManager : IDisposable
     {
         private const string ParamPrefix = "--";
 
         private readonly ITracer tracer;
 
-        public GVFSMountProcess(ITracer tracer, int sessionId)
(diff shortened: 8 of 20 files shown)


GVFS/GVFS.Platform.Windows/WindowsVolumeStateWatcher.cs
```diff
@@ -0,0 +1,144 @@
...
+namespace GVFS.Platform.Windows
+{
+    public class WindowsVolumeStateWatcher : IVolumeStateWatcher
+    {
+        private readonly ManagementEventWatcher volumeWatcher;
+        private readonly ManagementEventWatcher cryptoVolumeWatcher;
+        private readonly IDictionary<string, bool> cryptoVolumeLockStatuses =
+            new ConcurrentDictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
+
+        public WindowsVolumeStateWatcher(TimeSpan pollingInterval)
```
mjcheetham: This should probably take an `ITracer`.

GVFS/GVFS.Service/GVFSMountProcessManager.cs
```diff
@@ -6,34 +6,34 @@
 
 namespace GVFS.Service
 {
-    public class GVFSMountProcess : IDisposable
```
mjcheetham: This name confused me to no-end! 🤦‍♂️  I thought this represented a mount process instance itself, but it's really a sort of factory for processes right? I originally added a `StopMount(string repoRoot)` method to this along with the rename which would be invoked on `VolumeUnavailable` but removed that logic in favour of open question 1.

GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -115,18 +119,30 @@ protected override void OnSessionChange(SessionChangeDescription changeDescripti
...
                         using (ITracer activity = this.tracer.StartActivity("LogonAutomount", EventLevel.Informational))
                         {
-                            this.repoRegistry.AutoMountRepos(changeDescription.SessionId);
-                            this.repoRegistry.TraceStatus();
+                            var mounter = new RepoAutoMounter(this.tracer, this.repoRegistry, sessionId);
+                            this.repoMounters.Add(sessionId, mounter);
+
+                            mounter.Start();
+
+                            // TODO: this.repoRegistry.TraceStatus();
```
mjcheetham: Tracing of current state the repositories after logon should probably still happen, but do we want to trace the full status after each 'auto mount' or just at logon?

GVFS/GVFS.Service/Handlers/GetActiveRepoListHandler.cs
```diff
@@ -42,14 +43,7 @@ public void Run()
                 {
                     if (!this.IsValidRepo(repoRoot))
                     {
-                        if (!this.registry.TryRemoveRepo(repoRoot, out errorMessage))
```
mjcheetham: This is the line which is 'cleaning up' missing repositories _on read_. I don't think we should be clearing away entries on read? A 'clean up registrations' message would be more explicit, and could also do the check (`volume exists && repo not exists`) listed in open question 2.

GVFS/GVFS.Service/RepoAutoMounter.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
+using GVFS.Common;
+using GVFS.Common.FileSystem;
+using GVFS.Common.NamedPipes;
+using GVFS.Common.Tracing;
+using GVFS.Service.Handlers;
+
+namespace GVFS.Service
+{
+    public class RepoAutoMounter : IDisposable
```
mjcheetham: Currently there is no Mac daemon equivalent to the Windows service, but when there is in the future, this class is platform independent.

GVFS/GVFS.Service/RepoAutoMounter.cs
```diff
@@ -0,0 +1,150 @@
...
+            this.tracer.RelatedEvent(EventLevel.Informational, "VolumeStateChange", metadata);
+
+            switch (e.ChangeType)
+            {
+                case VolumeStateChangeType.VolumeAvailable:
+                    this.MountAll(rootPath: e.VolumePath);
+                    break;
+                case VolumeStateChangeType.VolumeUnavailable:
+                    // There is no need to do anything here to stop any potentially orphaned mount processes
+                    // since they will self-terminate if the volume is removed.
```
mjcheetham: This comment lies. Mount processes do not self-terminate.. but should they? Open question 1.

GVFS/GVFS.Service/RepoRegistry.cs
```diff
@@ -232,31 +202,38 @@ public void AutoMountRepos(int sessionId)
...
+                                string enlistmentPath = registration.EnlistmentRoot;
+
+                                // Try and normalize the enlistment path if the volume is available, otherwise just take the
+                                // path verbatim.
+                                string volumePath = GVFSPlatform.Instance.FileSystem.GetVolumeRoot(registration.EnlistmentRoot);
+                                if (GVFSPlatform.Instance.FileSystem.IsVolumeAvailable(volumePath))
                                 {
-                                    if (!normalizedEnlistmentRootPath.Equals(registration.EnlistmentRoot, StringComparison.OrdinalIgnoreCase))
+                                    string normalizedPath;
+                                    if (GVFSPlatform.Instance.FileSystem.TryGetNormalizedPath(registration.EnlistmentRoot, out normalizedPath, out errorMessage))
```
mjcheetham: `TryGetNormalizedPath` will fail if the path doesn't actually exist. For repositories on volumes which are still locked by BitLocker this means we can read the registry entries! Not sure why we'd ever get non-normalised paths in the registry, but changed this to only bother trying if the volume is actually available.

GVFS/GVFS.Service/RepoAutoMounter.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            using (var activity = this.tracer.StartActivity("AutoMount", EventLevel.Informational, metadata))
+            {
+                string volumeRoot = GVFSPlatform.Instance.FileSystem.GetVolumeRoot(enlistmentRoot);
+                if (GVFSPlatform.Instance.FileSystem.IsVolumeAvailable(volumeRoot))
+                {
+                    // TODO #1043088: We need to respect the elevation level of the original mount
+                    if (this.mountProcessManager.StartMount(enlistmentRoot))
+                    {
+                        this.SendNotification("GVFS AutoMount", "The following GVFS repo is now mounted:\n{0}", enlistmentRoot);
```
mjcheetham: Aside: these toast notifications (existing code) from "GVFS.Service.UI" never actually seemed to show for me, even on released versions of VFS for Git.

GVFS/GVFS.Service/RepoAutoMounter.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        public RepoAutoMounter(ITracer tracer, RepoRegistry repoRegistry, int sessionId)
+        {
+            this.tracer = tracer ?? throw new ArgumentNullException(nameof(tracer));
+            this.repoRegistry = repoRegistry ?? throw new ArgumentNullException(nameof(repoRegistry));
+            this.sessionId = sessionId;
+
+            // Create a mount process factory for this session/user
+            this.mountProcessManager = new GVFSMountProcessManager(this.tracer, sessionId);
+            this.userSid = this.mountProcessManager.CurrentUser.Identity.User?.Value;
```
wilbaker: Have you ever seen `this.mountProcessManager.CurrentUser.Identity.User` be null?

GVFS/GVFS.Service/GVFSMountProcessManager.cs
wilbaker: > I thought this represented a mount process instance itself, but it's really a sort of factory for processes right?

I'm not sure that I'd say it's a factory (as it doesn't produce any objects) but you have the right idea.  This class is a helper for calling the `gvfs` CLI.

Part of the naming confusion might be because prior to calling `gvfs mount` this class would start `GVFS.Mount.exe` directly.

GVFS/GVFS.Service/Handlers/GetActiveRepoListHandler.cs
wilbaker: It would still be nice to clean up old entries without requiring user interaction, and so your suggestion here:

> and could also do the check (volume exists && repo not exists) listed in open question 2.

Sounds like a good option.

wilbaker:
> We can now detect when a volume is removed. Currently if you remove a volume the mount process continues to run and is reported as 'mounted' (this is the existing behaviour).
Should we try and kill the mount processes for repositories on volumes that have been removed/locked? Should the mount process itself be responsible for killing itself or should the service send a "kill" message?

I think this could probably be its own issue\PR.  @sanoursa, what do you think?

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
```diff
@@ -8,5 +8,42 @@ public interface IPlatformFileSystem
         void CreateHardLink(string newLinkFileName, string existingFileName);
         bool TryGetNormalizedPath(string path, out string normalizedPath, out string errorMessage);
         void ChangeMode(string path, int mode);
+
+        /// <summary>
```
sanoursa: Style: As you can see, we don't tend to have many xml docs on our methods. I know this is a subject that we could debate and have well-informed opinions on both sides, but in this codebase our approach has been to favor descriptive method and param names over xml docs. We still do add the comments where they describe something that's non-obvious, but in this example here, the comments are only duplicating the info that is already present in the method signature.

Basically, I would drop all of the comments in this file.

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
```diff
@@ -8,5 +8,42 @@ public interface IPlatformFileSystem
...
         bool TryGetNormalizedPath(string path, out string normalizedPath, out string errorMessage);
         void ChangeMode(string path, int mode);
+
+        /// <summary>
+        /// Check if a path exists as a subpath of the specified directory.
+        /// </summary>
+        /// <param name="directoryPath">Directory path.</param>
+        /// <param name="path">Path to query.</param>
+        /// <returns>True if <see cref="path"/> exists as a subpath of <see cref="directoryPath"/>, false otherwise.</returns>
+        bool IsPathUnderDirectory(string directoryPath, string path);
```
sanoursa: Why does this method exist here in this interface? Isn't this a platform-agnostic question?

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
```diff
@@ -0,0 +1,67 @@
...
+        public const string VolumeEncryptionNamespace = @"root\cimv2\security\MicrosoftVolumeEncryption";
+
+        public static bool TryGetVolumeLockStatus(string driveName, out bool isLocked)
+        {
+            string queryString = $"SELECT * FROM Win32_EncryptableVolume WHERE DriveLetter = '{driveName}'";
+
+            using (var searcher = new ManagementObjectSearcher(VolumeEncryptionNamespace, queryString))
+            {
+                ManagementObjectCollection results = searcher.Get();
+                ManagementObject mo = results.OfType<ManagementObject>().FirstOrDefault();
```
sanoursa: Please avoid abbreviated variable names

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -29,6 +31,39 @@ public void ChangeMode(string path, int mode)
         {
         }
 
+        public bool IsPathUnderDirectory(string directoryPath, string path)
+        {
+            // TODO: Normalize paths
+            // We can't use the existing TryGetNormalizedPathImplementation method
+            // because it relies on actual calls to the disk to check if directories exist.
+            // This may be called with paths or volumes which do not actually exist.
+            return path.StartsWith(directoryPath, StringComparison.OrdinalIgnoreCase);
```
sanoursa: Referring back to my question about why `IsPathUnderDirectory` was added to the interface - the implementations of this method do not seem to be doing anything platform-specific, so they do not seem to belong here.

The way to think of the `GVFSPlatform` and the interfaces that it uses are that they represent the capabilities of the OS that we rely on. Resolving a path to a volume root is a good example, because what we require is that we can resolve to a root, and each platform will do that differently. Normalizing a path is also a good example. But given those two primitives, there is no need for a platform-specific method to do just a string comparison.

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -29,6 +31,39 @@ public void ChangeMode(string path, int mode)
...
+            }
+
+            return null;
+        }
+
+        public bool IsVolumeAvailable(string path)
+        {
+            // No paths 'exist' on locked BitLocker volumes so it is sufficent to just
+            // check if the directory/file exists using the framework APIs.
+            return System.IO.Directory.Exists(path) || System.IO.File.Exists(path);
```
sanoursa: This also seems like it's not particular to any platform

GVFS/GVFS.Platform.Windows/WindowsVolumeStateWatcher.cs
```diff
@@ -0,0 +1,144 @@
...
+            this.volumeWatcher = new ManagementEventWatcher(volumeQuery);
+            this.volumeWatcher.EventArrived += this.OnVolumeEvent;
+
+            // Watch for changes to BitLocker-protected volume states (unlock, lock, etc)
+            string cryptoVolumeQuery = $"SELECT * FROM __InstanceModificationEvent WITHIN {intervalSeconds} WHERE TargetInstance ISA 'Win32_EncryptableVolume'";
+            this.cryptoVolumeWatcher = new ManagementEventWatcher(BitLockerHelpers.VolumeEncryptionNamespace, cryptoVolumeQuery);
+            this.cryptoVolumeWatcher.EventArrived += this.OnCryptoVolumeEvent;
+        }
+
+        public event EventHandler<VolumeStateChangedEventArgs> VolumeStateChanged;
```
sanoursa: What benefit are we getting here from using a C# event vs just a callback delegate? I don't think we want/need to support multiple listeners, right? Events tend to make diagnostics a little harder, so I would constrain this down if we don't need the flexibility.

GVFS/GVFS.Service/GVFSMountProcessManager.cs
sanoursa: I would leave the name unchanged. I can understand the confusion, but it's analogous to the .NET `Process` class, or our own `GitProcess` class. The class is not the process - it's code for interacting with the process.

GVFS/GVFS/CommandLine/ServiceVerb.cs
```diff
@@ -33,7 +32,14 @@ public class ServiceVerb : GVFSVerb.ForNoEnlistment
...
             HelpText = "Prints a list of all mounted repos")]
-        public bool List { get; set; }
+        public bool ListMounted { get; set; }
+
+        [Option(
+            "list-all",
+            Default = false,
+            Required = false,
+            HelpText = "Prints a list of all known repos and their statuses")]
+        public bool ListAll { get; set; }
```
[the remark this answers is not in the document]
sanoursa: This feels more like a debugging feature than an end user feature anyway, so I would take it out. For debugging, we can just inspect the file directly.

GVFS/GVFS.Service/RepoAutoMounter.cs
mjcheetham: I have not seen it to be null, but some helpful(?) tooling indicated this property _could possibly_ be null.

https://docs.microsoft.com/en-us/dotnet/api/system.security.principal.windowsidentity.user?view=netframework-4.7.2#remarks

It appears the `WindowsIdentity.User` property is `null` only if the identity is the 'anonymous' user. Sessions could be created for anonymous users (file share connections open to 'everyone' perhaps?).. not sure if the fact we only act for `SessionChangeReason.SessionLogon` in `OnSessionChanged` means we don't see anonymous user sessions.

I can remove the `?.` if you'd like.

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -53,7 +53,7 @@ public static bool TryAttach(ITracer tracer, string enlistmentRoot, out string e
             try
             {
                 StringBuilder volumePathName = new StringBuilder(GVFSConstants.MaxPath);
-                if (!NativeMethods.GetVolumePathName(enlistmentRoot, volumePathName, GVFSConstants.MaxPath))
+                if (!Common.NativeMethods.GetVolumePathName(enlistmentRoot, volumePathName, GVFSConstants.MaxPath))
```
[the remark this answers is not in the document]
mjcheetham: There are several classes named `NativeMethods` thoughout the codebase.
This callsite used to use the inner-class `private static class NativeMethods` which had the `GetVolumePath` definition.

I needed to use this P/Invoke call outside of this `ProjFSFilter` class so I moved it to the `GVFS.Common.NativeMethods` class instead. Using just `NativeMethods.blah` here refers to the inner-class (which still has the `FilterAttach` P/Invoke method) whereas I want to reference `GVFS.Common.NativeMethods` instead now.

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Linq;
+using System.Management;
+
+namespace GVFS.Platform.Windows
+{
+    internal static class BitLockerHelpers
+    {
+        public const string VolumeEncryptionNamespace = @"root\cimv2\security\MicrosoftVolumeEncryption";
```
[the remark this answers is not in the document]
mjcheetham: https://docs.microsoft.com/en-us/windows/desktop/secprov/win32-encryptablevolume#requirements

image

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace GVFS.Platform.Windows
+{
+    internal static class BitLockerHelpers
+    {
+        public const string VolumeEncryptionNamespace = @"root\cimv2\security\MicrosoftVolumeEncryption";
+
+        public static bool TryGetVolumeLockStatus(string driveName, out bool isLocked)
+        {
+            string queryString = $"SELECT * FROM Win32_EncryptableVolume WHERE DriveLetter = '{driveName}'";
```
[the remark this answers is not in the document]
mjcheetham: So far I have found reading the _current_ BitLocker state it to be very reliable.

Watching for events (in `class WindowsVolumeStateWatcher`) has also been reliable.
Events for `Win32_VolumeChangeEvent` monitor for general drive added/removed; this was pretty instant.
Events for `__InstanceModificationEvent` to `Win32_EncryptableVolume` monitor for BitLocker volume status changes; these are a little slower due to the filtering required in the query over all `__InstanceModificationEvent`s.. we don't need to respond immediately within microseconds to an unlock however do we?

AFAIK this is the only exposed API we have to query BitLocker information and to watch/wait for changes to lock/unlock state, without having the BitLocker filter driver expose information and events like PrjFlt does.

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Linq;
+using System.Management;
+
+namespace GVFS.Platform.Windows
+{
+    internal static class BitLockerHelpers
```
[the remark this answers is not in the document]
mjcheetham: I can fold this into the `WindowsVolumeStateWatcher` class.

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
mjcheetham: I see. I was under the impression that it was policy for all `public` types and methods in product assemblies to be fully documented. Is VFS for Git a supported product? Should all VFS for Git assemblies not be considered public (even if they physically are) and carry no compatibility guarentees? Should these types be made internal instead?

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
mjcheetham: Perhaps, although case sensitivity when comparing path segments is file system (and transitively OS) dependent?

GVFS/GVFS.Common/FileSystem/IVolumeStateWatcher.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace GVFS.Common.FileSystem
+{
+    /// <summary>
+    /// Monitors the system for changes to the state of any disk volume, notifying subscribers when a change occurs.
+    /// </summary>
+    public interface IVolumeStateWatcher : IDisposable
```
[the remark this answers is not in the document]
mjcheetham: Would you imagine this to be a 'fire once/self unregister' callback? Or return some cookie/handle as part of the registration process to be used when unregistering otherwise.

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
mjcheetham: Sure. I usually don't use abbreviated variable names, but based on [this](https://github.com/mjcheetham/VFSForGit/blob/723f21ff814a0003feae614a82763d8c2800951a/GVFS/GVFS.Platform.Windows/WindowsPhysicalDiskInfo.cs#L100) prior art I thought it was acceptable for `ManagementBaseObject` and `ManagementObject` to be `mbo`/`mo`.

Will change to `win32CryptoVolume` or similar. Should the other instances of `mbo` short var names (such as in WindowPhysicalDiskInfo.cs) be expanded too?

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
```diff
@@ -0,0 +1,67 @@
...
+            string queryString = $"SELECT * FROM Win32_EncryptableVolume WHERE DriveLetter = '{driveName}'";
+
+            using (var searcher = new ManagementObjectSearcher(VolumeEncryptionNamespace, queryString))
+            {
+                ManagementObjectCollection results = searcher.Get();
+                ManagementObject mo = results.OfType<ManagementObject>().FirstOrDefault();
+                if (mo != null)
+                {
+                    // Check the protection status
+                    // ProtectionStatus == 0 means the drive is not protected by BitLocker and therefore cannot be locked
```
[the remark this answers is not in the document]
mjcheetham: I think the part of the comment that a drive being unprotected means it cannot be locked is still useful for those reading in the future not knowing the difference between 'protection' and 'lock'.
I'll extract the integers as private constants with commented links to the [relevant](https://docs.microsoft.com/en-us/windows/desktop/secprov/win32-encryptablevolume#properties) [source docs](https://docs.microsoft.com/en-us/windows/desktop/secprov/getlockstatus-win32-encryptablevolume).

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
mjcheetham: Is case sensitivity of the underlying file system not a platform dependent property? HFS+ on Mac (and I guess NTFS on Windows) can support case sensitive modes or directories.. or is the stake in the ground that VFS for Git will never support case sensitive file systems (Linux/ext*)?

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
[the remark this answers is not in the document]
wilbaker: Thanks, i missed that `NativeMethods` was still a private class here.  The P/Invokes are kind of a mess right now, long term I don't think we want any in `GVFS.Common` (as they're all platform specific) but no need to clean that up now.

GVFS/GVFS.Common/NativeMethods.cs
```diff
@@ -164,6 +164,13 @@ public static DateTime GetLastRebootTime()
             return DateTime.Now - uptime;
         }
 
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        public static extern bool GetVolumePathName(
+            string volumeName,
+            StringBuilder volumePathName,
+            uint bufferLength);
+
```
alameenshah: nit - couple things.
1. Is volumePathName the output here? And the return value indicates success/failure? If yes, then may be this could be renamed to TryGetVolumePathName()? 
2. Usage of both “Path” and “Name” in method & argument names look confusing to me. If it means the full path to a mounted volume, then may be “Path” is enough? Like TryGetVolumePath(.., out string path, …)?

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
```diff
@@ -0,0 +1,67 @@
...
+            return false;
+        }
+
+        private static bool TryInvokeMethod(ManagementObject mo, string methodName, object[] args)
+        {
+            try
+            {
+                return mo.InvokeMethod(methodName, args) is uint result && result == 0;
+            }
+            catch (Exception)
```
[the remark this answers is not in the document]
sanoursa: I was really asking what it means to fail gracefully :-). Catching all exceptions (even with logging) is more akin to succeeding blindly. We tend to favor a fail-fast approach, but it's a little different here since this code is not handling any repo data and is less likely to get itself into a bad state.

GVFS/GVFS.Service/RepoAutoMounter.cs
sanoursa: They do self-terminate the next time they try to do some IO and it fails.

GVFS/GVFS.Service/RepoAutoMounter.cs
sanoursa: Yea they have been very problematic and inconsistent for us. I suspect something changed in the platform and we're no longer using their APIs correctly.

GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -26,6 +28,7 @@ public class GVFSService : ServiceBase
         private string serviceName;
         private string serviceDataLocation;
         private RepoRegistry repoRegistry;
+        private IDictionary<int, RepoAutoMounter> repoMounters;
```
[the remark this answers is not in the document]
sanoursa: It's not even clear that we need a dictionary here. From the current usage, a List would be fine, but it seems that we could even get away with nothing at all.

GVFS/GVFS.Service/RepoAutoMounter.cs
sanoursa: What would this do if User is null? What does it mean for us to store a userSid of 0? How will that behave when we try to mount later?

The issue isn't so much the presence of the `?` as it is that this will likely lead to cascading failures down the line. If you're going to handle null users here, then you also need to do something other than carry on as if it succeeded.

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
sanoursa: I see, that is a valid concern. However, note that we are currently very unprepared to deal with case sensitive file systems. Note that even putting it here in this interface doesn't solve that problem, since it's not a per-OS question but rather a per-file-system-and-machine-configuration question. We will have to deal with this as a broader issue since we are now porting to Linux as well, but it is outside the scope of this PR.

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
sanoursa: I'm fairly certain I asked for no abbreviations on that other code as well :-). I do like your suggested rename, as it describe the thing the object represents, rather than its type in the API.

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
sanoursa: Unfortunately it is not a platform-dependent property, where platform means OS. On the same OS, a file system can be case sensitive or insensitive. We currently have no pattern to deal with this, but will need to add one soon.

GVFS/GVFS.Common/FileSystem/IVolumeStateWatcher.cs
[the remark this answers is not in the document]
sanoursa: Fire once makes the most sense to me, but I've realized that we just have different assumptions about what this feature is trying to do. See more on my upcoming note.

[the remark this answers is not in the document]
sanoursa: Today's behavior is: one and only one best effort attempt to automount after initial login. I don't think we should expand the scope of that feature in _this_ PR, but should instead focus on extending that same behavior to BitLockered drives. 

Expanding the intent of the feature could also be a fine thing to do, but it carries a lot more risk (we have nothing today that will automount without either an explicit user gesture or a fatal error, and nothing that will automount at any time other than initial login) and so it could lead to surprising interactions for users and issues that are difficult to diagnose. So going after those other scenarios could be nice, but will require a much more thorough design to ensure we don't create any unexpected behaviors.

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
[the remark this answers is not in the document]
sanoursa: That will just bloat the responsibilities of that other class :-)

GVFS/GVFS.Platform.Windows/BitLockerHelpers.cs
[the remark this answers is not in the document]
sanoursa: The problem we had with the WMI queries in the past is that they worked 99% of the time but would occasionally not deliver an event. I strongly suspect you'll have the same problem here. 

If we continue to treat automount as a best effort one-time behavior after initial login, I think that lack of reliability is tolerable. If we're trying to continuously monitor volumes and react, this becomes a lot more problematic.

GVFS/GVFS.Common/NativeMethods.cs
sanoursa: This is not a new method, it's just mapping an existing Win32 API into C# code. So we need to stick with the names/signatures of that existing method.

GVFS/GVFS.Service/RepoAutoMounter.cs
mjcheetham: If user is null, then we'd cascade the nulls and have a `(string)null` SID. The call later when mounting to get the list of registered, active repositories for that user's SID would return no repositories (it filters the collection by `string.Compare(repo.OwnerSID, ownerSID, ..)`.

I guess this is a silent event that we might want to know about (that we have a null/anonymous user). I'll remove the `?.` to revert to the existing behaviour?

GVFS/GVFS.Common/NativeMethods.cs
alameenshah: oh...ok

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -29,6 +31,39 @@ public void ChangeMode(string path, int mode)
...
+        }
+
+        public IVolumeStateWatcher CreateVolumeStateWatcher()
+        {
+            // TODO: Extract the polling interval to a configuration value?
+            return new WindowsVolumeStateWatcher(TimeSpan.FromSeconds(15));
```
alameenshah: 1. If we are going to be polling anyways, wouldn't it be simpler to just iterate though the yet-to-be-mounted repository list and check if the Enlistment root path became available and readable in the Timer callback? Thats all that we need to know, right? Or does the volumeWatcher and cryptoVolumeWatcher offer anything more?

2. After all the un-available repos get mounted eventually, do we shutdown this timer?

jrbriggs:
I'm going to close this PR for now. We're going to pick up the work in a subsequent PR, but the approach will change enough that I want to start fresh there.

PR #362, opened 2018-10-09
Add check-ignore as command that allows placeholders to be created

### LINKED ISSUE #361: Add check-ignore to the list of commands allowed to create placeholders
VSCode is running `git check-ignore -z --stdin` while enumerating directories and checking for files which means placeholders need to be created while VSCode is passing files to the check-ignore command.

kewillford:
This fixes #361

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CreatePlaceholderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CreatePlaceholderTests.cs
@@ -1,0 +1,73 @@
+﻿using GVFS.FunctionalTests.Should;
+using GVFS.FunctionalTests.Tools;
+using NUnit.Framework;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace GVFS.FunctionalTests.Tests.GitCommands
+{
+    [TestFixture]
+    [Category(Categories.GitCommands)]
+    public class CreatePlaceholderTests : GitRepoTests
+    {
+        private static readonly string FileToRead = Path.Combine("GVFS", "GVFS", "Program.cs");
+
+        public CreatePlaceholderTests() : base(enlistmentPerTest: true)
+        {
+        }
+
+        [TestCase("check-attr")]
+        [TestCase("check-ignore")]
+        [TestCase("check-mailmap")]
+        [TestCase("diff-tree")]
+        [TestCase("fetch-pack")]
+        [TestCase("hash-object")]
+        [TestCase("index-pack")]
+        [TestCase("name-rev")]
+        [TestCase("notes")]
+        [TestCase("send-pack")]
+        [TestCase("rev-list")]
+        [TestCase("update-ref")]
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -16,7 +16,10 @@ public abstract class FileSystemVirtualizer : IDisposable
         protected static readonly byte[] FolderContentId = Encoding.Unicode.GetBytes(GVFSConstants.AllZeroSha);
 
         protected static readonly GitCommandLineParser.Verbs CanCreatePlaceholderVerbs =
-            GitCommandLineParser.Verbs.AddOrStage | GitCommandLineParser.Verbs.Move | GitCommandLineParser.Verbs.Status;
+            GitCommandLineParser.Verbs.AddOrStage |
+            GitCommandLineParser.Verbs.Move |
+            GitCommandLineParser.Verbs.Status |
+            GitCommandLineParser.Verbs.CheckIgnore;
```
sanoursa: Do we always want to allow `check-ignore`, or only when it has a `--stdin` parameter? What about other verbs that take a `--stdin` parameter, could they be in a similar boat?

sanoursa: Can we add a functional test that fails without this change?

kewillford: Commands that could be waiting on stdin
```
check-attr
check-ignore
check-mailmap
checkout-index
commit-graph
diff-tree
fetch-pack
hash-object
index-pack
name-rev
notes
reset  -  Experimental
send-pack
rev-list
update-index
update-ref
```

The fact that these wait on stdin isn't necessarily an issue, but placeholders will not be able to be created if these are running and waiting on input.

kewillford: Running some tests and the failures are for:
```
check-ignore
check-mailmap
checkout-index
reset
update-index
```

Most of the others are in the list to not get the lock which is probably where `check-ignore` and `check-mailmap` belong.  I'll update the PR.  The others will have to stay for now since they are changing the index.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/CreatePlaceholderTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        [TestCase("diff-tree")]
+        [TestCase("fetch-pack")]
+        [TestCase("hash-object")]
+        [TestCase("index-pack")]
+        [TestCase("name-rev")]
+        [TestCase("notes")]
+        [TestCase("send-pack")]
+        [TestCase("rev-list")]
+        [TestCase("update-ref")]
+        public void AllowsPlaceholderCreation(string commandToRun)
```
sanoursa: Nit: `AllowsPlaceholderCreationWhileGitCommandIsRunning`?

PR #369, opened 2018-10-10
Fix missing untracked files when created in subfolder 

### LINKED ISSUE #358: `git status` does not report untracked files that are more than one folder deep
This is a new break that is likely related to the changes to cleanup ModifiedPaths.dat.

If a new file is created several folders deep, git status will not report it.

*Last released version*

```
M:\>gvfs clone https://gvfs.visualstudio.com/ci/_git/ForTests
M:\>cd ForTests\src
M:\ForTests\src>mkdir foo
M:\ForTests\src>mkdir foo\bar
M:\ForTests\src>echo test > foo\bar\test
M:\ForTests\src>git status
On branch master
Your branch is up-to-date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        foo/

nothing added to commit but untracked files present (use "git add" to track)

```

*Latest master: Untracked file is not reported*

```
M:\>gvfs clone https://gvfs.visualstudio.com/ci/_git/ForTests ForTestsMaster
M:\>cd ForTestsMaster
M:\ForTestsMaster>cd src
M:\ForTestsMaster\src>mkdir foo
M:\ForTestsMaster\src>mkdir foo\bar
M:\ForTestsMaster\src>echo test > foo\bar\test
M:\ForTestsMaster\src>git status
On branch master
Your branch is up-to-date with 'origin/master'.

```

kewillford:
Add to the test to create multiple level of folders and files in those folder to make sure they show up as untracked files.

Update the git for windows version that has the fix.

Fixes #358

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20180814.4</GitPackageVersion>
+    <GitPackageVersion>2.20181011.3</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
@@ -16,6 +16,8 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         private const string EncodingFilename = "ريلٌأكتوبرûمارسأغسطسºٰٰۂْٗ۵ريلٌأك.txt";
         private const string ContentWhenEditingFile = "// Adding a comment to the file";
         private const string UnknownTestName = "Unknown";
+        private const string TopLevelFolderToCreate = "level1";
+        private const string SubFolderToCreate = "level2";
 
         private static readonly string EditFilePath = Path.Combine("GVFS", "GVFS.Common", "GVFSContext.cs");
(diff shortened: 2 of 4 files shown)


GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20180814.4</GitPackageVersion>
+    <GitPackageVersion>2.20181009.1-pr</GitPackageVersion>
```
wilbaker: What was the fix in Git?

kewillford: See https://github.com/Microsoft/git/pull/33

A folder entry in the modified paths was not being used as recursive in the case of untracked files.  So if there was the entry foo/ in the modified paths, git was not going into any of the folders of foo to look for untracked files.

This had to do with the fact that while using the exclude we were "unexcluding" folders so if we had the file foo/bar/test.txt, we would need the !foo/, !foo/bar/ in order for git to traverse the directories and find the untracked file.  We still do that but because we no longer add to the modified paths if there is a parent directory entry, git needs to treat the folder entry as recursive.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -1088,6 +1088,9 @@ private void CommitChangesSwitchBranchSwitchBack(Action fileSystemAction, [Calle
         private void CreateFile()
         {
             this.CreateFile("Some content here", Path.GetRandomFileName() + "tempFile.txt");
+            this.CreateFolder("foo");
```
sanoursa: I'd suggest better names (Level1/Level2?) and constants for these folder names

PR #372, opened 2018-10-10
Mac: Turn on 'gvfs prefetch --hydrate' 

### LINKED ISSUE #243: Mac: ReadFilesJob references kernel32

nickgra:
Fixes #243 

Implement a native macOS solution for 'gvfs prefetch --hydrate' and enable associated functional tests.

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacFileSystem.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystem.cs
@@ -38,10 +38,44 @@ namespace GVFS.Platform.Mac
             return MacFileSystem.TryGetNormalizedPathImplementation(path, out normalizedPath, out errorMessage);
         }
 
+        public bool HydrateFile(string fileName, byte[] buffer)
+        {
+            return NativeFileReader.TryReadFirstByteOfFile(fileName, buffer);
+        }
+
         [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
         private static extern int Chmod(string pathname, int mode);
 
         [DllImport("libc", EntryPoint = "rename", SetLastError = true)]
         private static extern int Rename(string oldPath, string newPath);
+
+        private class NativeFileReader
+        {
+            private const int ReadOnly = 0x0000;
+
+            public static bool TryReadFirstByteOfFile(string fileName, byte[] buffer)
+            {
+                int fileDescriptor = Open(fileName, ReadOnly);
+                return TryReadOneByte(fileDescriptor, buffer);
+            }
+
+            private static bool TryReadOneByte(int fileDescriptor, byte[] buffer)
+            {
+                int numBytes = Read(fileDescriptor, buffer, 1);
+
+                if (numBytes == -1)
+                {
+                    return false;
+                }
+
+                return true;
+            }
+
+            [DllImport("libc", EntryPoint = "open", SetLastError = true)]
+            private static extern int Open(string path, int flag);
+
+            [DllImport("libc", EntryPoint = "read", SetLastError = true)]
+            private static extern int Read(int fd, [Out] byte[] buf, int count);
+        }
     }
 }
--- a/GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
@@ -1,5 +1,8 @@
 ﻿using GVFS.Common;
 using GVFS.Common.FileSystem;
+using Microsoft.Win32.SafeHandles;
+using System.IO;
+using System.Runtime.InteropServices;
 
 namespace GVFS.Platform.Windows
 {
@@ -32,6 +35,59 @@ namespace GVFS.Platform.Windows
         public bool TryGetNormalizedPath(string path, out string normalizedPath, out string errorMessage)
         {
             return WindowsFileSystem.TryGetNormalizedPathImplementation(path, out normalizedPath, out errorMessage);
+        }
+
+        public bool HydrateFile(string fileName, byte[] buffer)
+        {
+            return NativeFileReader.TryReadFirstByteOfFile(fileName, buffer);
+        }
+
+        private class NativeFileReader
+        {
+            private const uint GenericRead = 0x80000000;
+            private const uint OpenExisting = 3;
+
+            public static bool TryReadFirstByteOfFile(string fileName, byte[] buffer)
+            {
+                using (SafeFileHandle handle = Open(fileName))
+                {
+                    if (!handle.IsInvalid)
+                    {
+                        return ReadOneByte(handle, buffer);
+                    }
+                }
+
+                return false;
+            }
+
+            private static SafeFileHandle Open(string fileName)
+            {
+                return CreateFile(fileName, GenericRead, (uint)(FileShare.ReadWrite | FileShare.Delete), 0, OpenExisting, 0, 0);
+            }
+
+            private static bool ReadOneByte(SafeFileHandle handle, byte[] buffer)
+            {
+                int bytesRead = 0;
+                return ReadFile(handle, buffer, 1, ref bytesRead, 0);
+            }
+
+            [DllImport("kernel32", SetLastError = true, ThrowOnUnmappableChar = true, CharSet = CharSet.Unicode)]
+            private static extern SafeFileHandle CreateFile(
+                string fileName,
+                uint desiredAccess,
+                uint shareMode,
+                uint securityAttributes,
+                uint creationDisposition,
+                uint flagsAndAttributes,
+                int hemplateFile);
+
+            [DllImport("kernel32", SetLastError = true)]
+            private static extern bool ReadFile(
+                SafeFileHandle file,
+                [Out] byte[] buffer,
+                int numberOfBytesToRead,
+                ref int numberOfBytesRead,
+                int overlapped);
         }
     }
 }
--- a/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
@@ -8,5 +8,6 @@
         void CreateHardLink(string newLinkFileName, string existingFileName);
         bool TryGetNormalizedPath(string path, out string normalizedPath, out string errorMessage);
         void ChangeMode(string path, int mode);
+        bool HydrateFile(string fileName, byte[] buffer);
     }
 }
--- a/GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
+++ b/GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
@@ -1,9 +1,6 @@
 ﻿using GVFS.Common.Tracing;
-using Microsoft.Win32.SafeHandles;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
-using System.Runtime.InteropServices;
 using System.Threading;
 
 namespace GVFS.Common.Prefetch.Jobs
@@ -43,15 +40,7 @@ namespace GVFS.Common.Prefetch.Jobs
                 {
                     foreach (string path in this.blobIdToPaths[blobId])
                     {
-                        bool succeeded = false;
-                        using (SafeFileHandle handle = NativeFileReader.Open(path))
-                        {
-                            if (!handle.IsInvalid)
-                            {
-                                succeeded = NativeFileReader.ReadOneByte(handle, buffer);
-                            }
-                        }
-
+                        bool succeeded = GVFSPlatform.Instance.FileSystem.HydrateFile(path, buffer);
                         if (succeeded)
                         {
                             Interlocked.Increment(ref this.readFileCount);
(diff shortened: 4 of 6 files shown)


GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -64,6 +64,8 @@ public static void Register(GVFSPlatform platform)
             ITracer tracer,
             string lockPath);
 
+        public abstract bool OpenAndReadFile(string fileName);
```
wilbaker: Naming suggestion: `TryReadFirstByte` or `TryReadFirstByteOfFile`.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -285,5 +287,61 @@ private static object GetValueFromRegistry(RegistryHive registryHive, string key
...
+                uint shareMode,
+                uint securityAttributes,
+                uint creationDisposition,
+                uint flagsAndAttributes,
+                int hemplateFile);
+
+            [DllImport("kernel32", SetLastError = true)]
+            private static extern unsafe bool ReadFile(
+                SafeFileHandle file,
+                void* buffer,
```
wilbaker: Can we use `[Out] byte []` here as well and avoid having to add `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>`?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -140,5 +141,41 @@ public override bool IsGitStatusCacheSupported()
...
+
+                if (success == -1)
+                {
+                    return false;
+                }
+
+                return true;
+            }
+
+            private static unsafe int ReadOneByte(int fileDescriptor)
```
wilbaker: Maybe make this `TryReadOneByte` and have it return a bool?

It looked a bit odd to see:

```int success = ReadOneByte(fileDescriptor);```

When `ReadOneByte` is returning the number of bytes read.

File: GVFS/GVFS.Platform.Mac/MacPlatform.cs [hunk 4]
```diff
@@ -140,5 +141,41 @@ public override bool IsGitStatusCacheSupported()
...
+            {
+                byte[] buffer = new byte[1];
+                return Read(fileDescriptor, buffer, 1);
+            }
+
+            [DllImport("libc", EntryPoint = "open", SetLastError = true)]
+            private static extern int Open(string path, int flag);
```
wilbaker: Not something for this PR, but at some point we might want to move all the native interop calls to a common class in `MacPlatform`

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -285,5 +292,57 @@ private static object GetValueFromRegistry(RegistryHive registryHive, string key
...
+
+                return false;
+            }
+
+            private static SafeFileHandle Open(string fileName)
+            {
+                return CreateFile(fileName, GenericRead, (uint)(FileShare.ReadWrite | FileShare.Delete), 0, OpenExisting, 0, 0);
+            }
+
+            private static unsafe bool ReadOneByte(SafeFileHandle handle)
```
wilbaker: Name here too: `TryReadOneByte`

GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs
```diff
@@ -114,6 +114,11 @@ public override bool IsGitStatusCacheSupported()
         public override FileBasedLock CreateFileBasedLock(PhysicalFileSystem fileSystem, ITracer tracer, string lockPath)
         {
             return new MockFileBasedLock(fileSystem, tracer, lockPath);
-        }
+        }
+
+        public override bool OpenAndReadFile(string fileName)
+        {
+            throw new NotImplementedException();
```
sanoursa: NotSupportedException

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: I would move this to `IPlatformFileSystem`, and rename to `ReadOneByte`

GVFS/GVFS.Platform.Mac/MacPlatform.cs
sanoursa: Does this actually need to be unsafe?

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: Or `TryReadOneByte` as William suggested.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
nickgra: I agree, I was thinking the same thing while working on this. Want me to open a work item for it?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -6,7 +6,8 @@
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Pipes;
-
+using System.Runtime.InteropServices;
```
sanoursa: Drop this?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -4,12 +4,14 @@
 using GVFS.Common.Tracing;
 using GVFS.Platform.Windows.DiskLayoutUpgrades;
 using Microsoft.Win32;
+using Microsoft.Win32.SafeHandles;
```
sanoursa: Drop these changes too

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -36,6 +36,42 @@ public void ChangeMode(string path, int mode)
...
+        }
+
+        public bool TryReadFirstByteOfFile(string fileName)
+        {
+            return NativeFileReader.TryReadFirstByteOfFile(fileName);
+        }
+
+        private class NativeFileReader
+        {
+            // #define O_RDONLY     0x0000
```
sanoursa: I would drop this comment

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -63,7 +63,7 @@ public static void Register(GVFSPlatform platform)
             PhysicalFileSystem fileSystem,
             ITracer tracer,
             string lockPath);
-
+
```
sanoursa: Drop this

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -6,7 +6,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Pipes;
-
+
```
sanoursa: Drop this change

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -275,7 +275,7 @@ public override bool IsGitStatusCacheSupported()
         public override bool TryGetGVFSEnlistmentRoot(string directory, out string enlistmentRoot, out string errorMessage)
         {
             return WindowsPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
-        }
+        }
```
sanoursa: This and the next file are both space-only changes, drop them

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -36,6 +36,41 @@ public void ChangeMode(string path, int mode)
...
+
+            public static bool TryReadFirstByteOfFile(string fileName)
+            {
+                int fileDescriptor = Open(fileName, ReadOnly);
+                return TryReadOneByte(fileDescriptor);
+            }
+
+            private static bool TryReadOneByte(int fileDescriptor)
+            {
+                byte[] buffer = new byte[1];
```
sanoursa: We'll actually need to fix this pointer before passing it over to the native code. Otherwise the risk is that the GC can move this memory while the native code is running, and it will write to a now invalid pointer

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
sanoursa: FYI we need to use fixed not just for the type conversion, but also to ensure the GC doesn't move the memory. So we're stuck with the unsafe keywords here.

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -32,6 +35,63 @@ public void ChangeMode(string path, int mode)
...
+                uint creationDisposition,
+                uint flagsAndAttributes,
+                int hemplateFile);
+
+            [DllImport("kernel32", SetLastError = true)]
+            private static extern unsafe bool ReadFile(
+                SafeFileHandle file,
+                [Out] byte* buffer,
+                int numberOfBytesToRead,
+                [Out] int numberOfBytesRead,
```
sanoursa: How did this translate to an int? Shouldn't it be a pointer?

sanoursa: I think what's happening is that the type is not translated correctly here, but because the value you're passing in happens to be initialized to 0, the native code is probably seeing this as a null pointer and ignoring it. But anything other than a 0 will probably cause a crash.

sanoursa: And I can see now that this is the pre-existing Windows code, and I may have been the person who wrote this incorrect code :-)

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -36,12 +36,50 @@ public void ChangeMode(string path, int mode)
...
+                int fileDescriptor = Open(fileName, ReadOnly);
+                return TryReadOneByte(fileDescriptor);
+            }
+
+            private static unsafe bool TryReadOneByte(int fileDescriptor)
+            {
+                byte[] buffer = new byte[1];
+                fixed (byte* p = buffer)
+                {
+                    int resultCode = Read(fileDescriptor, p, 1);
```
sanoursa: The return is actually `numBytes`, not `resultCode`, right? It just so happens that -1 is a special value.

GVFS/GVFS.Platform.Windows/GVFS.Platform.Windows.csproj
```diff
@@ -12,6 +12,7 @@
     <AssemblyName>GVFS.Platform.Windows</AssemblyName>
     <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
wilbaker: Is this change still needed?

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -32,6 +35,59 @@ public void ChangeMode(string path, int mode)
...
+
+            private static SafeFileHandle Open(string fileName)
+            {
+                return CreateFile(fileName, GenericRead, (uint)(FileShare.ReadWrite | FileShare.Delete), 0, OpenExisting, 0, 0);
+            }
+
+            private static unsafe bool ReadOneByte(SafeFileHandle handle, byte[] buffer)
+            {
+                int bytesRead = 0;
+                return ReadFile(handle, buffer, 1, &bytesRead, 0);
```
wilbaker: Why are you passing in `&bytesRead`?  If you just pass in `bytesRead` there should be no need to mark this `unsafe`.

wilbaker: Oh sorry, I missed that this was for `int* numberOfBytesRead`

wilbaker: If you change the `DllImport` from:

`[Out] int* numberOfBytesRead,`

to 

`ref int numberOfBytesRead,`

You can drop the `unsafe`

wilbaker: See example here (the second parameter)

https://docs.microsoft.com/en-us/dotnet/framework/interop/marshaling-different-types-of-arrays

*C API*
```
int TestRefArrayOfInts(int** ppArray, int* pSize);  
```

*DllImport in C#*
```
   [ DllImport( "..\\LIB\\PinvokeLib.dll" )]
   public static extern int TestRefArrayOfInts( ref IntPtr array, 
      ref int size );
```

nickgra: Thanks!

PR #375, opened 2018-10-11
Mac: Update build scripts to handle new Profiling configuration

nickgra:
Update build scripts to be able to handle the 'Profiling(Release)' configuration. Only pass that configuration value into xcodebuild as .NET requires 'Release.Mac' to be the configuration.

### DIFF
--- a/ProjFS.Mac/Scripts/Build.sh
+++ b/ProjFS.Mac/Scripts/Build.sh
@@ -14,5 +14,10 @@ PROJFS=$SRCDIR/ProjFS.Mac
 
 xcodebuild -configuration $CONFIGURATION -workspace $PROJFS/PrjFS.xcworkspace build -scheme PrjFS -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
 
+# If we're building the Profiling(Release) configuration, remove Profiling() for building .NET code
+if [ "$CONFIGURATION" == "Profiling(Release)" ]; then
+  CONFIGURATION=Release
+fi
+
 dotnet restore $PROJFS/PrjFSLib.Mac.Managed/PrjFSLib.Mac.Managed.csproj /p:Configuration=$CONFIGURATION /p:Platform=x64 --packages $PACKAGES || exit 1
 dotnet build $PROJFS/PrjFSLib.Mac.Managed/PrjFSLib.Mac.Managed.csproj /p:Configuration=$CONFIGURATION /p:Platform=x64 || exit 1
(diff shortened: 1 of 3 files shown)


Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -38,6 +38,11 @@ GITPATH="$(find $PACKAGES/gitformac.gvfs.installer/$GITVERSION -type f -name *.d
 # Now that we have a path containing the version number, generate GVFSConstants.GitVersion.cs
 $SCRIPTDIR/GenerateGitVersionConstants.sh "$GITPATH" $BUILDDIR || exit 1
 
+# If we're building the Profiling(Release) configuration, remove Profiling() for building .NET code
+if [ "$CONFIGURATION" == "Profiling(Release)" ]; then
+  CONFIGURATION=Release
```
sanoursa: Shouldn't this be `DOTNETCONFIGURATION`?

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -14,5 +14,10 @@ PROJFS=$SRCDIR/ProjFS.Mac
 
 xcodebuild -configuration $CONFIGURATION -workspace $PROJFS/PrjFS.xcworkspace build -scheme PrjFS -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
 
+# If we're building the Profiling(Release) configuration, remove Profiling() for building .NET code
+if [ "$CONFIGURATION" == "Profiling(Release)" ]; then
```
sanoursa: Use `DOTNETCONFIGURATION` for consistency?

Do we need a change in the MirrorProvider scripts too?

Scripts/Mac/BuildGVFSForMac.sh
sanoursa: Oh I see, `DOTNETCONFIGURATION` hasn't been set yet. I would probably clean this up so we're deciding how to set `DOTNETCONFIGURATION` rather than changing `CONFIGURATION`

ProjFS.Mac/Scripts/Build.sh
nickgra: We do need a change in the MirrorProvider, adding that. On making that change, I discovered that the MirrorProvider's script does something different (doesn't bother making a DOTNETCONFIGURATION variable so I'll make that change in the BuildGVFSForMac script too.

Scripts/Mac/BuildGVFSForMac.sh
nickgra: We have to set DOTNETCONFIGURATION for the Debug/default case too so I'd rather have two separate conversions. One to reset 'Profiling(Release)' to Release and then another to unconditionally add '.Mac' to the configuration, regardless of if it's Debug/Release/malformed input.

PR #384, opened 2018-10-15
Include UTC offset in log timestamp.

mitesch:
Change logging to include the date in consistent year-month-day format and include the machine's time zone offset.
This simplifies correlating user logs with server logs.

Previously
`[10/11/2018 3:16:00 PM] DownloadObjects`

Now
`[2018-10-11 15:16:00 -04:00] DownloadObjects`


jrbriggs:
@mitesch I like the second proposal that makes explicit the timezone but still leaves logs in local time. It's a good balance of server/client debugability.

mitesch:
Updated the description.

PR #394, opened 2018-10-16
Mac: Update and delete placeholder functions should check if file is still a placeholder

### LINKED ISSUE #332: Mac: Update/Delete placeholder functions should check the file's state before proceeding
Currently `PrjFS_UpdatePlaceholderFileIfNeeded` and `PrjFS_DeleteFile` are not checking if a file is full before proceeding, and they should.

wilbaker:
Resolves #332

- Update `PrjFS_DeleteFile` so that it can only delete files and directories.  Additionally, `PrjFS_DeleteFile` will now only allow partial files to be deleted.
- Update `HandleFileNotification` to remove the `PrjFSFileXAttrName` attribute when a partial file is modified
- Enable functional tests that depended on this behavior

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -91,6 +91,7 @@ static bool InitializeEmptyPlaceholder(const char* fullPath);
 template<typename TPlaceholder> static bool InitializeEmptyPlaceholder(const char* fullPath, TPlaceholder* data, const char* xattrName);
 static bool AddXAttr(const char* fullPath, const char* name, const void* value, size_t size);
 static bool GetXAttr(const char* fullPath, const char* name, size_t size, _Out_ void* value);
+static bool RemoveXAttr(const char* fullPath, const char* name);
 
 static inline PrjFS_NotificationType KUMessageTypeToNotificationType(MessageType kuNotificationType);
 
@@ -524,7 +525,6 @@ PrjFS_Result PrjFS_DeleteFile(
         << hex << updateFlags << dec << ")" << endl;
 #endif
     
-    // TODO(Mac): Populate failure cause appropriately
     *failureCause = PrjFS_UpdateFailureCause_Invalid;
     
     if (nullptr == relativePath)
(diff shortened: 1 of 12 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -533,10 +533,31 @@ PrjFS_Result PrjFS_DeleteFile(
...
+    }
+    
+    if (S_ISREG(path_stat.st_mode))
+    {
+        // TODO(Mac): Determine if we need a similar check for directories as well
+        PrjFSFileXAttrData xattrData = {};
+        if (!GetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData))
+        {
+            *failureCause = PrjFS_UpdateFailureCause_FullFile;
+            return PrjFS_Result_EVirtualizationInvalidOperation;
```
sanoursa: I don't think this is an invalid operation. This can easily happen because of a race that causes a placeholder file to get converted to full right as the provider is trying to delete it.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -533,10 +533,31 @@ PrjFS_Result PrjFS_DeleteFile(
...
+        // Anything else should be treated as a full
+        *failureCause = PrjFS_UpdateFailureCause_FullFile;
+        return PrjFS_Result_EVirtualizationInvalidOperation;
+    }
+    
+    if (S_ISREG(path_stat.st_mode))
+    {
+        // TODO(Mac): Determine if we need a similar check for directories as well
+        PrjFSFileXAttrData xattrData = {};
+        if (!GetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData))
```
sanoursa: Speaking of races, I think we need to grab the file lock here so that we're not deleting at the same time as transitioning to full.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -533,10 +533,31 @@ PrjFS_Result PrjFS_DeleteFile(
...
     
     char fullPath[PrjFSMaxPath];
     CombinePaths(s_virtualizationRootFullPath.c_str(), relativePath, fullPath);
+    
+    struct stat path_stat;
+    stat(fullPath, &path_stat);
+    if (!(S_ISREG(path_stat.st_mode) || S_ISDIR(path_stat.st_mode)))
+    {
+        // Only files and directories can be deleted with PrjFS_DeleteFile
+        // Anything else should be treated as a full
```
sanoursa: Nit: "as a full file"

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -938,6 +959,13 @@ static PrjFS_Result HandleFileNotification(
         isDirectory,
         notificationType,
         nullptr /* destinationRelativePath */);
+    
+    if (partialFile && PrjFS_NotificationType_FileModified == notificationType)
```
sanoursa: "&& result == PrjFS_Result_Success"?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
sanoursa: I personally think that's a mistake. An "invalid operation" should indicate that the caller of the API did something wrong, i.e. they have a bug. This is a situation that is totally unavoidable from the provider's perspective, so it's a runtime error, not a bug.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: As discussed offline, I'll see if we can\should make `HandleFileNotification` have a `void` return type.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: As discussed offline, because the file modified notification is not a vnode callback, we can't synchronize converting a file to full and updating a placeholder.   However, the same race exists with git (without VFS4G), and so for now we can stick with the current changes and not add in locking.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: @sanoursa I checked and the `OnNotifyOperation` operation is also used for `PreDelete` (which might have a non-success result).

I think we could change this behavior, but the scope is outside of this PR.  For now I'll add a comment, let me know if you'd like me to file a follow up issue.

PR #399, opened 2018-10-17
Mac: Close file descriptors when 'prefetch --hydrate'-ing

nickgra:
When attempting to 'prefetch --hydrate' a very large repo, VFSForGit will die and spew errors that point at the process running out of file descriptors. This only happens on very large repos, on a repo the size of VFSForGit, we don't hit this limit.

We need to close the file descriptors when we are opening it to read a single byte to force a hydration.

Additionally, fix a potential kernel panic where a build agent might still have the kext loaded when trying to install Git as part of PrepFunctionalTests.sh.


GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -56,7 +56,9 @@ private class NativeFileReader
             public static bool TryReadFirstByteOfFile(string fileName, byte[] buffer)
             {
                 int fileDescriptor = Open(fileName, ReadOnly);
-                return TryReadOneByte(fileDescriptor, buffer);
+                bool readStatus = TryReadOneByte(fileDescriptor, buffer);
+                Close(fileDescriptor);
```
wilbaker: What would you think of putting the `Close` inside of a `finally`?

`TryReadOneByte` shouldn't ever throw, but using a `try`\ `finally` gets us the same behavior as a `using` statement.

PR #401, opened 2018-10-17
Add WaitForBackgroundOperations call with checking modified paths

kewillford:
This looks like more changes than it really is because I changed the `ModifiedPaths..` methods to take a `GVFSFunctionalTestEnlistment ` and removed the `string dotGVFSRoot` parameter since all of them were passing `this.Enlistment.DotGVFSRoot`

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
@@ -104,7 +104,7 @@ namespace GVFS.FunctionalTests.Tools
 
         public static void ModifiedPathsContentsShouldEqual(GVFSFunctionalTestEnlistment enlistment, FileSystemRunner fileSystem, string contents)
         {
-            enlistment.WaitForBackgroundOperations().ShouldBeTrue("Background operations failed to complete.");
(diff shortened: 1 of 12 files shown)


GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
```diff
@@ -102,16 +102,18 @@ public static string ReadAllTextFromWriteLockedFile(string filename)
             }
         }
 
-        public static void ModifiedPathsContentsShouldEqual(FileSystemRunner fileSystem, string dotGVFSRoot, string contents)
+        public static void ModifiedPathsContentsShouldEqual(GVFSFunctionalTestEnlistment enlistment, FileSystemRunner fileSystem, string contents)
         {
-            string modifiedPathsDatabase = Path.Combine(dotGVFSRoot, TestConstants.Databases.ModifiedPaths);
+            enlistment.WaitForBackgroundOperations().ShouldBeTrue("Background operations failed to complete.");
```
sanoursa: It looks like every single call to `WaitForBackgroundOperations` throughout our test code always has to append the `ShouldBeTrue("same error message")`. I would move that check inside the `WaitForBackgroundOperations` method and get rid of all of that duplication.

PR #406, opened 2018-10-18
[Engineering] Remove most relative paths from Windows scripts

jrbriggs:
Our scripts on Windows have many relative paths. Remove those and call a common script to set environment variables that we can rely on during script execution.


Scripts/Macros.cmd.txt
```diff
@@ -0,0 +1,7 @@
+b=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat $*
+br=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat Release $*
+ut=%VFS_SCRIPTSDIR%\RunUnitTests.bat $*
+utr=%VFS_SCRIPTSDIR%\RunUnitTests.bat Release $*
+ft=%VFS_SCRIPTSDIR%\RunFunctionalTests.bat $*
+ftr=%VFS_SCRIPTSDIR%\RunFunctionalTests.bat Release $*
+scorch=%VFS_SCRIPTSDIR%\NukeBuildOutputs.bat $*
```
sanoursa: This script is really intended for build machines, not dev machines. It's not _too_ far off, except that I don't want it killing my GVFS.Service Windows service just because I'm deleting my build outputs.

Scripts/BuildGVFSForWindows.bat
```diff
@@ -1,12 +1,14 @@
 @ECHO OFF
 SETLOCAL
 
+CALL %~dp0\EnsureVfsDevShell.bat || EXIT /b 10
```
sanoursa: Nit: Vfs -> VFS

Scripts/ReinstallGVFS.bat
```diff
@@ -1,6 +1,9 @@
+@ECHO OFF
```
sanoursa: This is another script that is intended for build machines only.

Maybe outside the scope of this PR, but I think the following would be a very nice additional cleanup:
* Move all scripts in the `Scripts` folder into `Scripts/Windows`
* Further separate the scripts into those intended to be run on dev machines and those that are for build machines only. At a first glance, all of the scripts you have aliased (except for scorch) could go in `Scripts/Windows/Dev` (or just leave them in `Scripts/Windows` to save some letters) and many of the rest could be moved to `Scripts/Windows/BuildAgents`

Thoughts?

Scripts/Macros.cmd.txt
```diff
@@ -0,0 +1,7 @@
+b=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat $*
```
sanoursa: Do you plan to do something similar for Mac development?

init.cmd
```diff
@@ -0,0 +1,55 @@
+@ECHO OFF
```
sanoursa: Sorry, I'm gonna take back my Approve vote for a minute while we discuss this :-)

I don't think Windows-specific scripts should go in the root of the repo, at least not with such a generic name. What are your thoughts for how this will extend to Mac and Linux?

Scripts/Macros.cmd.txt
```diff
@@ -0,0 +1,7 @@
+b=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat $*
+br=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat Release $*
+ut=%VFS_SCRIPTSDIR%\RunUnitTests.bat $*
+utr=%VFS_SCRIPTSDIR%\RunUnitTests.bat Release $*
+ft=%VFS_SCRIPTSDIR%\RunFunctionalTests.bat $*
```
sanoursa: Nit: explicitly pass in Debug here, so that I can pass in extra params and not have to remember to specify that param. Same for all of the macros, be explicit about the Debug param.

Scripts/Macros.cmd.txt
```diff
@@ -0,0 +1,7 @@
+b=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat $*
+br=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat Release $*
+ut=%VFS_SCRIPTSDIR%\RunUnitTests.bat $*
+utr=%VFS_SCRIPTSDIR%\RunUnitTests.bat Release $*
+ft=%VFS_SCRIPTSDIR%\RunFunctionalTests.bat $*
+ftr=%VFS_SCRIPTSDIR%\RunFunctionalTests.bat Release $*
```
sanoursa: Not sure how useful this one is

Scripts/Macros.cmd.txt
```diff
@@ -0,0 +1,7 @@
+b=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat $*
+br=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat Release $*
```
sanoursa: Other scripts that I run _all_ the time are `..\BuildOutput\GVFS.Build\InstallG4W.bat` and `..\BuildOutput\GVFS.Build\InstallGVFS.bat`

Scripts/Macros.cmd.txt
```diff
@@ -0,0 +1,7 @@
+b=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat $*
+br=%VFS_SCRIPTSDIR%\BuildGVFSForWindows.bat Release $*
+ut=%VFS_SCRIPTSDIR%\RunUnitTests.bat $*
+utr=%VFS_SCRIPTSDIR%\RunUnitTests.bat Release $*
```
sanoursa: Not sure if this one is that useful either

Maybe this isn't everyone's experience, but these are the things I do locally (and I rely on the build agents for everything else)
* Build, 99% of the time it's debug only
* Install what I just built
* Run a specific functional test or test class

Unit tests are my default project in VS, so I don't need a macro for that. Full functional tests take way too long so I rely on the build agents and only run failed ones locally. And there's really no point in running the release versions locally.

Thoughts?

Scripts/Macros.cmd.txt
jrbriggs: Yes -- that way a developer on either platform has access to the same shortcuts, even if the implementation is different.

Scripts/ReinstallGVFS.bat
jrbriggs: I like that idea. Let me see what I can do...

Scripts/Macros.cmd.txt
jrbriggs: Great point, thanks.

Scripts/Macros.cmd.txt
jrbriggs: A subsequent PR will condense those to an InstallProduct script (since we require both to be in sync) for the yaml build support. But your use case probably wants them separated still, so you don't need to pay for Git installation when you're iterating on VFS for Git?

Scripts/Macros.cmd.txt
jrbriggs: That's a fair point. The reason I was offering these as shortcuts was to make very clear how to get started as a dev. I might cut the release variants but I like the idea of here's the basics: build, unit test, functional test. Install probably makes the cut for "how to get productive"

Scripts/Macros.cmd.txt
sanoursa: Yep for my local testing, I only want to pay the cost of installing G4W once, and then I can build/install GVFS over and over again while tweaking things

Scripts/Macros.cmd.txt
sanoursa: Well for a "getting started" scenario, the functional test one is tricky. You have to first build, then install both G4W and GVFS, and _then_ you can run the functional tests. Personally I would tailor this more to our own productivity, and refer new devs to a wiki page.

sanoursa: Like it would be awesome to have a macro to run just one functional test, so I don't have to go find the syntax for that every single time :-). e.g. `ft Name.Of.Class.Or.Test`

init.cmd
nickgra: I agree that we need to have a cross platform solution here, but this is a tricky problem due to what ships inbox on Windows and Unix systems. There's no scripting language that's common between them.

Batch files don't run on MacOS. Powershell ships with Windows and is available on MacOS, but requires an installation.
MacOS ships with bash, which is sort of an option on Windows if we want to take a dependency on a Git bash window (WSL can't meet our needs, we need a real Windows environment for everything other than bash). It also ships with Perl/Python, which requires an installation.

We might just have to have two scripts that do identical things that are implemented in different languages. Looking at projects similar to ours like the dotnet family of repos, they ship .cmd/.sh copies of each of their scripts. 

https://github.com/dotnet/corefx

Scripts/Macros.cmd.txt
nickgra: I agree with Saeed's comments completely. We've already got some starter documentation on getting started as a new developer (which could be improved), I'm more interested in optimizing for our productivity.

init.cmd
jrbriggs: Yep, @nickgra, this is the direction I'm thinking we head. I don't want to hide important scripts down two levels of the tree. I'm happy to consider other names but I don't have a better suggestion than `init`ialize the environment.

sanoursa: What's wrong with a `Scripts\Windows\init.cmd` and `Scripts/Mac/init.sh` approach? Again, these scripts are not (in my opinion) designed for discoverability by a new person, so I would follow our existing organizational pattern if possible.

jrbriggs: What's wrong with having them at the root? Why make me type "Scripts\Windows\init" instead of init? I think we should have these at the root for clarity, and we can squirrel the rest of implementation details in the Scripts\[Platform] pattern. 

I think init is special because it provides the facade over `Scripts\Windows\BuildGVFSForWindows` and `Scripts\Mac\BuildGVFSForMac`. A user, on either platform, types "init" and then "b". At least that part of the docs can be platform-agnostic, and it allows for us to make more of engineering system platform-agnostic in the future.

jrbriggs:
> Did you look at or consider replacing the scripts with a .net core app? This might allow us to share more. We would still need to have separate scripts if we wanted to have the shortcuts for building, testing, etc from cmd or bash.

Yes @kewillford I think this is something to investigate, be it dotnet or xplat scripting environment acquired during init. I'm not a fan of having so much of the engineering system logic as platform-specific scripts. We should look at converging as much as possible.

init.cmd
sanoursa: I see, yea so just have an init.cmd and init.sh in the root? That can work, but two questions with that:
* Why does Windows get two scripts? Do we need both .cmd and .ps1?
* How do we differentiate Mac and Linux, since both support .sh?

jrbriggs: Unfortunately running the .bat from powershell doesn't persist the environment as the child process terminates, which is why we extract the environment into a temp file and read it into the powershell environment. I think we should provide one, but we could consider making it less in your face and moving it to Scripts\Windows

WRT sh support, my hope is we don't need to differentiate Mac and Linux at this layer. It's really just crafting the environment and providing some convenience shortcuts. If we needed to fork those shortcuts, I'd guess we can just interrogate `uname` and set the environment pointing at the platform we think we're running on. However, I haven't tested this yet.

Scripts/ReinstallGVFS.bat
jrbriggs: If we go moving all of this around now, before we are on yaml pipelines, we will have to modify the build definitions. Let's hold off on this cleanup until our pipelines are versioned with the code.

Scripts/Macros.cmd.txt
jrbriggs: Sounds good, I'll trim this list down.

Scripts/ReinstallGVFS.bat
sanoursa: Yep doing it once we move to yaml is a good idea.

init.cmd
sanoursa: We will definitely need to fork some of the shortcuts. e.g. the Mac build scripts invoke xcodebuild, and Linux most definitely will not, so we will end up with Linux-specific build scripts.

On the filp side, some of the commands are already cross-plat-ready, e.g. invoking the unit tests just needs to execute a .NET Core application (though the scripts are currently doing slightly different things so they would need some work to be totally unified)

Scripts/ReinstallGVFS.bat
jrbriggs: Deferring for now.

Scripts/SetupDevService.bat
```diff
@@ -1,4 +1,6 @@
 @ECHO OFF
+CALL %~dp0\EnsureVFSDevShell.bat || EXIT /b 10
+
 IF "%1"=="" (SET "Configuration=Debug") ELSE (SET "Configuration=%1")
 
-sc create %Configuration%.GVFS.Service binPath=%~dp0\..\..\BuildOutput\GVFS.Service\bin\x64\%Configuration%\GVFS.Service.exe
\ No newline at end of file
+sc create %Configuration%.GVFS.Service binPath=%VFS_OUTPUTDIR%\GVFS.Service\bin\x64\%Configuration%\GVFS.Service.exe
```
wilbaker: I didn't realize we had this script.  @sanoursa do you know if anyone\anything uses this script?  Or should we just remove it?

init.cmd
```diff
@@ -0,0 +1,56 @@
...
+TITLE VFS ^for Git Developer Shell ^(%VFS_SRCDIR%^)
+doskey /MACROFILE=.\Scripts\Macros.cmd.txt
+
+ECHO ==============================================
+ECHO * Welcome to the VFS for Git developer shell *
+ECHO *                                            *
+ECHO * Build: b                                   *
+ECHO * Install Git: ig                            *
+ECHO * Install VFS for Git: iv                    *
+ECHO * Run functional test: test [pattern]        *
```
wilbaker: Update `test` to `ftest`, and consider changing `[pattern]` to `[test name(s)]`

jrbriggs: Whoops, good catch.

Scripts/SetupDevService.bat
sanoursa: I have never used them, and I can't find any other scripts that call them. I think they must have been added to help with testing changes to the built GVFS.Service without installing it, but I don't think they add any value anymore. I'd be good with deleting all of the `*DevService.bat` scripts.

init.cmd
```diff
@@ -0,0 +1,56 @@
+@ECHO OFF
+
+REM Don't run inside of devenv shells
+IF NOT "%VSCMD_VER%"=="" (
```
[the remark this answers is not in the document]
jrbriggs: Longer term, I would like to isolate the build from the version of VS that's on the machine. This isn't necessary in this PR though, so I can remove it.

init.cmd
```diff
@@ -0,0 +1,56 @@
...
+
+REM Don't run inside of devenv shells
+IF NOT "%VSCMD_VER%"=="" (
+  ECHO ERROR: Do not run from a Developer Command prompt.
+  ECHO Use a plain shell and try again.
+  EXIT /b 10
+)
+
+REM Don't run twice
+IF NOT "%VFS_DEVSHELL%"=="" (
```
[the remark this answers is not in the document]
jrbriggs: You would only need to rerun if envvars changed, or to pick up a new macro. Let's try it like this, since it means we don't need to worry with environment upgrades. If it proves clunky we'll revisit.

init.cmd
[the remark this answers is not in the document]
sanoursa: Sounds good. Isolating the build makes sense, but the devenv shell gives me other things, like nuget.exe. We are now downloading that ourselves, so we could also add that to the PATH and maybe that eliminates my need for the devenv shell (I'm sure I've opened it for other reasons too, but nuget has been the big one)

init.cmd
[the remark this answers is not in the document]
sanoursa: OK but when that happens I'll probably just comment out this line rather than restart my shell :-)

init.cmd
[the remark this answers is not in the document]
jrbriggs: Got it. I've both removed the restriction on devenv shells and put nuget on the path.

Scripts/SetupDevService.bat
jrbriggs: I will do that in the subsequent PR that also refactors the scripts dir.

jrbriggs:
@sanoursa Do you want further changes or is this ready to go in?

init.ps1
```diff
@@ -0,0 +1,18 @@
+# Delegate the work to init.cmd and save off the environment it creates to a temp file.
```
sanoursa: I thought you had mentioned moving this file into `Scripts\Windows`? How important is it to even have it?

init.cmd
```diff
@@ -0,0 +1,51 @@
+@ECHO OFF
```
sanoursa: If it's important to you to leave this script in the root, how do you feel about renaming it to `InitWindowsEnvironment.cmd`? The usage will be the exact same as what you have now: `ini<tab><enter>` but its purpose is clearer that way. Though personally I still think it would be nicest to put it in `Scripts\Windows` and follow the same pattern as all of our scripts.

jrbriggs: I'm rethinking these scripts as I investigate the macOS counterparts. Just running init.sh doesn't leave the environment variables exported. A user would need to source the script.

We can either be ok with that or just make these scripts pre-reqs to all other scripts and remove the convenience macros. Opinions?

sanoursa: Yea I've always thought of the scripts _as_ the convenience macros. I'd be just fine with dropping those, and having each script include the base script to inherit the environment variables.

Readme.md
```diff
@@ -44,7 +44,8 @@ If you'd like to build your own VFS for Git Windows installer:
 * Install the .NET Core 2.1 SDK (https://www.microsoft.com/net/download/dotnet-core/2.1)
 * Create a folder to clone into, e.g. `C:\Repos\VFSForGit`
 * Clone this repo into the `src` subfolder, e.g. `C:\Repos\VFSForGit\src`
-* Run `\src\Scripts\BuildGVFSForWindows.bat`
+* Run `src\init.cmd` (Command Prompt) or `src\init.ps1` (PowerShell).
```
kewillford: This needs to be updated now.

jrbriggs: Good point, thanks.

Issue #403, opened 2018-10-17
Untracked files are not reported after 'reset --mixed' if those files existed as placeholders prior to the reset command running


wilbaker:
To reproduce:

1. Ensure placeholders are on disk for all files that did not exist in HEAD~1
2. `git reset --mixed HEAD~1`
3. `git status` does not report the untracked files (that did not exist in HEAD~1)

This is an issue on Mac and Windows.

wilbaker:
@jrbriggs @sanoursa Although this issue is not Mac specific I've added it to the "Porting to Mac" project because it's resulting in several functional test failures.

wilbaker:
Fixed by #408

PR #408, opened 2018-10-19
Add untracked files from reset mixed to ModifiedPaths.dat

wilbaker:
Fixeds #403 

**Issue**

(This was an issue on both Windows and Mac)

`git reset --mixed` can result in untracked files when users `reset` to a commit that does not have all of the tracked files present in the working directory.

`git status` was not reporting placeholder files in this state because:

- They were not in ModifiedPaths.dat
- git does not touch them during the reset (it only removes the path(s) from the index), and so there was no I/O during the the git command that would result in them being added to ModifiedPaths.dat

**Fix**

VFS for Git already had a background task to scan the index and check for files that had their skip-worktree bit cleared by `git reset`.  That task has been expanded to also check for placeholders that are not longer in the projection.  If there are placeholders missing from the projection, they are added to the modified paths, removed from the placeholder database, and `git status` properly reports them as untracked.


GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -167,6 +167,58 @@ public void GetAllEntries(out List<PlaceholderData> filePlaceholders, out List<P
...
+                    this.TryParseRemoveLine,
+                    (key, value) =>
+                    {
+                        if (value == PartialFolderValue || value == ExpandedFolderValue)
+                        {
+                            folderPlaceholdersFromDisk.Add(new PlaceholderData(path: key, fileShaOrFolderValue: value));
+                        }
+                        else
+                        {
+                        filePlaceholdersFromDiskByPath[key] = new PlaceholderData(path: key, fileShaOrFolderValue: value);
```
wilbaker: Fix spacing

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -529,12 +532,25 @@ public FileSystemTaskResult AddMissingModifiedFiles()
             {
                 if (this.modifiedFilesInvalid)
                 {
-                    FileSystemTaskResult result = this.indexParser.AddMissingModifiedFiles(this.context.Tracer, this.indexFileStream);
+                    Dictionary<string, PlaceholderListDatabase.PlaceholderData> placeholderDataByPath;
+                    List<PlaceholderListDatabase.PlaceholderData> placeholdersToKeep;
+                    this.placeholderList.GetAllEntries(out placeholderDataByPath, folderPlaceholders: out placeholdersToKeep);
```
sanoursa: I'm not quite following the changes here in this chunk of code. 

First question is, what is `GetAllEntries` returning? Are `placeholderDataByPath` and `placeholdersToKeep` duplicate data, just stored in different data structures? Are they mutually exclusive (and what's the division)? Are they separate but possibly overlapping sets?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -529,12 +532,25 @@ public FileSystemTaskResult AddMissingModifiedFiles()
...
+
+                    FileSystemTaskResult result = this.indexParser.AddMissingModifiedFiles(this.context.Tracer, this.indexFileStream, placeholderDataByPath, placeholdersToKeep);
                     if (result == FileSystemTaskResult.Success)
                     {
                         this.modifiedFilesInvalid = false;
                     }
 
+                    // Any paths that are not in the the index must be added to ModifiedPaths.dat so that git
+                    // reports them as untracked
+                    foreach (string path in placeholderDataByPath.Keys)
```
sanoursa: I would have thought `placeholdersToKeep` are the ones we need to add. Again, not really following the purpose of the two data structures.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -529,12 +532,25 @@ public FileSystemTaskResult AddMissingModifiedFiles()
...
+                    {
+                        this.AddModifiedPath(path);
+                    }
+
+                    this.placeholderList.WriteAllEntriesAndFlush(placeholdersToKeep);
```
sanoursa: Is this a replace? Or an addition to the existing list? And why do we need to add to the list of placeholders now? I thought we just needed to update the ModifiedPaths?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: I think some better naming would help.

The flow in here is:

- `placeholderDataByPath` -> Populated with file placeholders by `GetAllEntries`.  This should be renamed to something like `placeholderFilesByPath`
- `placeholdersToKeep` -> Populated by folder placeholders by `GetAllEntries`.

As `AddMissingModifiedFiles` is walking the index, it will move entries from `placeholderDataByPath` to `placeholdersToKeep` if the files still exist in the index and have their skip-worktree bit turned on.

Then, when `AddMissingModifiedFiles` completes any items still left in `placeholderDataByPath` were either not in the index or they did not have their skip-worktree bit turned on.  Those entries should be added to ModifiedPaths.dat and removed from the placeholder list.

At the end of the method `WriteAllEntriesAndFlush` is called with `placeholdersToKeep` because it now contains:
- All folder placeholders
- All file placeholders that are still in the index with their skip-worktree bit turned on

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: `placeholdersToKeep` are the entries that we should keep in the placeholder list.  At this point in the method any entries in `placeholderDataByPath` were not found in the index and need to be added to ModifedPaths.dat

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: There were two reasons for this:

- The design of `placeholderList` requires that each call to `GetAllEntries` be paired with a corresponding call to `WriteAllEntriesAndFlush`.
- As we add the entries to ModifiedPaths we must remove them from the placeholder list.  If we did not, we could run into a situation like this:

1. Add files to ModifiedPaths.dat here (and leave them in placeholder list)
2. Checkout a branch that does not have the untracked files
3. The usual UpdatePlaceholders logic runs, and it does not find these placeholders in the projection and decides they should be deleted
4. The delete succeeds because the files are still placeholders (i.e. they are not full).

However, all that said, I think we can probably remove the first limitation mentioned above (that each Get requires a Write).  This was originally added so that the collapsing of the placeholder list that occurs during a projection change does not drop and placeholders created during that time.

However, we don't need to collapse the list here (we could just write deletes for the files that should be removed), and so it would be a nice speed up to eliminate the restriction.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -529,12 +532,25 @@ public FileSystemTaskResult AddMissingModifiedFiles()
...
                     if (result == FileSystemTaskResult.Success)
                     {
                         this.modifiedFilesInvalid = false;
                     }
 
+                    // Any paths that are not in the the index must be added to ModifiedPaths.dat so that git
+                    // reports them as untracked
+                    foreach (string path in placeholderDataByPath.Keys)
+                    {
+                        this.AddModifiedPath(path);
```
wilbaker: `this.AddModifiedPath(path)` can return an error and we need to handle it (also rename `AddModifiedPath` to `TryAddModifiedPath`)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
sanoursa: I don't know if this answered my question, but I think I can infer it - you're saying that we're not _adding_ to the placeholder list here, we're actually removing from it? That was not obvious from the code, but I'm also not totally sold on it if that is true. Still processing...

sanoursa: And what does it mean to have placeholder files on disk but not in out PlaceholderList database?

wilbaker: > I don't know if this answered my question, but I think I can infer it - you're saying that we're not adding to the placeholder list here, we're actually removing from it? 

Yep that's right, and I think it will be more obvious with the change discussed above (to not call `WriteAllEntriesAndFlush` and to instead call `RemoveAndFlush`).

We remove from the placeholder list by doing the following (and a higher level):

1. Get a list of all placeholders
2. Remove some from that (in memory) list
3. Replace what's on disk with the (updated) list

The "remove" happens by moving items from `placeholderDataByPath` to `placeholdersToKeep` as we decide we want to keep them.  `placeholdersToKeep` starts with all of the folder placeholders as we know from the start we want to keep those.

> And what does it mean to have placeholder files on disk but not in out PlaceholderList database?

It means that we'll still hydrate them (if needed), but we won't update them when the projection changes (because git will).  If we wanted a more detailed name, we could call the list `PlaceholdersOnDiskThatGitIsNotTracking`.  In other words, the database is the intersection of placeholders on disk and index entries that have their 

We've actually supported this state for a long time.  When a placeholder file is renamed:
- It remains a placeholder
- It is added to ModifedPaths.dat
- It is removed from the placeholder list

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -105,18 +113,30 @@ private FileSystemTaskResult AddToProjection(GitIndexEntry data)
...
                     // A git command (e.g. 'git reset --mixed') may have cleared a file's skip worktree bit without
                     // triggering an update to the projection.  Ensure this file is in GVFS's modified files database
-                    data.ParsePath();
                     return this.projection.AddModifiedPath(data.GetFullPath());
                 }
                 else
                 {
                     data.ClearLastParent();
+
+                    string placeholderPath = data.GetFullPath();
```
wilbaker: Make sure that on both platforms we end up with the right path separators (and add a functional test that will fail on Windows without this fix).

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
sanoursa: Thanks that all makes sense and I think with your proposed renames, and updating the pattern to remove from, rather than replace, the placeholder list, this logic will become a lot clearer

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -167,6 +167,58 @@ public void GetAllEntries(out List<PlaceholderData> filePlaceholders, out List<P
             }
         }
 
+        public void GetAllEntries(
+            out Dictionary<string, PlaceholderListDatabase.PlaceholderData> placeholderDataByPath,
+            out List<PlaceholderData> folderPlaceholders)
+        {
+            try
+            {
+                Dictionary<string, PlaceholderListDatabase.PlaceholderData> filePlaceholdersFromDiskByPath =
```
kewillford: What is the memory usage with this going to be if a user has the entire repo as placeholders?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
kewillford: This could be moved outside the if/else since both code paths call `GetFullPath`.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: Most of the `Try` methods return a `bool` to check where this is returning a `FileSystemTaskResult` so I'm not sure it fits how we are using TryX methods.

wilbaker: That's true.  After taking a look at the other methods in this class it looks like we're not using `Try` for those that can fail and return `FileSystemTaskResult`.  I'll just update the code to check the return value.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -105,21 +134,50 @@ private FileSystemTaskResult AddToProjection(GitIndexEntry data)
...
                 {
                     // A git command (e.g. 'git reset --mixed') may have cleared a file's skip worktree bit without
-                    // triggering an update to the projection.  Ensure this file is in GVFS's modified files database
-                    data.ParsePath();
-                    return this.projection.AddModifiedPath(data.GetFullPath());
+                    // triggering an update to the projection. If git cleared the skip-worktree bit then git will
+                    // be responsible for updating the file and we need to:
+                    //    - Ensure this file is in GVFS's modified files database
+                    //    - Remove this path from the placeholders list (if present)
+                    result = this.projection.AddModifiedPath(placeholderRelativePath);
```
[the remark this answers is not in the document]
kewillford: Ah yes missed that.  I just saw that this changed from using `GetFullPath` which used the git path separator to `GetRelativePath` which is using the OS path separator.  Do you think there will be any performance impact on Windows since it will need to replace separators now?  I don't think so since this seems to be a limited code path.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -105,21 +134,50 @@ private FileSystemTaskResult AddToProjection(GitIndexEntry data)
...
+            /// Adjusts the modifed paths and placeholders list for an index entry.
+            /// </summary>
+            /// <param name="gitIndexEntry">Index entry</param>
+            /// <param name="filePlaceholders">
+            /// Dictionary of file placeholders.  AdjustModifedPathsAndPlaceholdersForIndexEntry will
+            /// remove enties from filePlaceholders as they are found in the index.  After
+            /// AdjustModifedPathsAndPlaceholdersForIndexEntry is called for all entries in the index 
+            /// filePlaceholders will contain only those placeholders that are not in the index.
+            /// </param>
+            private FileSystemTaskResult AdjustModifedPathsAndPlaceholdersForIndexEntry(
```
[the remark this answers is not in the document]
kewillford: Yeah that's as good as anything I would be able to come up with.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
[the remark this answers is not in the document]
wilbaker: > Do you think there will be any performance impact on Windows since it will need to replace separators now? I don't think so since this seems to be a limited code path.

You're right that this is a limited code path.  

We do have the option of getting both path formats, but then it's a trade off between the replacement and building a new path.

Your comment did remind me that I wanted to add some tracing to record how long this scan\update took.  I'll add that in.

If we find that in practice this new code is taking to long we can look for further optimizations, let me know if you think it's work doing more perf analysis now.

wilbaker:
>We need more tests that run the prefetch so we make sure we are covering more of that scenario but that is outside the scope of this PR.

@kewillford yep I agree we'll want more tests like this.  I didn't want to add the complexity of it to this PR, but I was thinking rather than prefetching we could also just make sure to open handles to all files in the repo (but not read them) to ensure the placeholders are on disk.  As we add this to more tests the prefetching times might start to add up.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -167,6 +167,37 @@ public void GetAllEntries(out List<PlaceholderData> filePlaceholders, out List<P
...
+                string error;
+                if (!this.TryLoadFromDisk<string, string>(
+                    this.TryParseAddLine,
+                    this.TryParseRemoveLine,
+                    (key, value) =>
+                    {
+                        if (value != PartialFolderValue && value != ExpandedFolderValue)
+                        {
+                            filePlaceholdersFromDiskByPath[key] = new PlaceholderData(path: key, fileShaOrFolderValue: value);
+                        }
```
jeschu1: @wilbaker to make sure my understanding is correct his returns all the files (instead of both files and directories).

Did we decide against just doing something like:
   var AllEntries = GetAllEntries();
   var results = AllEntries.where(x => x.value != PartialFolderValue && x.value != ExpandedFolderValue);
Due to performance reasons?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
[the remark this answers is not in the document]
wilbaker: Thanks for the suggestions.  I went with `AddEntryToModifiedPathsAndRemoveFromPlaceholdersIfNeeded` because it left a little more wiggle room.  

`MoveEntryFromPlaceholderToModifiedPathsIfInIndex` captured one of the things done by this method, but I thought it gave the impression that the placeholder list was the only thing driving if an entry is added to ModifiedPaths (which is not the case).

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -360,8 +360,12 @@ public virtual void OnIndexFileChange()
             }
             else if (this.GitCommandLeavesProjectionUnchanged(gitCommand))
             {
-                this.GitIndexProjection.InvalidateModifiedFiles();
-                this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnIndexWriteWithoutProjectionChange());
+                if (gitCommand.IsResetSoftOrMixed())
```
wilbaker: This should be `IsResetMixed`

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -18,11 +21,23 @@ internal class GitIndexEntry
             private const int MaxParts = MaxPathBufferSize / 2;
             private const byte PathSeparatorCode = 0x2F;
 
+            private static readonly string PathSeparatorString = Path.DirectorySeparatorChar.ToString();
+
             private int previousFinalSeparatorIndex = int.MaxValue;
 
-            public GitIndexEntry()
+            private LazyUTF8String[] lazyPathParts;
+            private string[] utf16PathParts;
```
wilbaker: Add a comment for these

wilbaker:
@sanoursa @kewillford @jeschu1 

These changes are ready for another look.  

I updated the code so that:
- VFS4G will only run through the modified path validation for reset mixed
- I've added some extra logging to this path
- The BG task thread will no longer use LazyPaths when parsing the index, which eliminates the race condition that I saw in an earlier test run

*Example run with no placeholders (on our large repo)*
```
[2018-10-24 09:48:30 -07:00] AddMissingModifiedFilesAndRemoveThemFromPlaceholderList (Start)
[2018-10-24 09:48:30 -07:00] AddMissingModifiedFilesAndRemoveThemFromPlaceholderList_FilePlaceholderCount {"FilePlaceholderCount":0}
[2018-10-24 09:48:31 -07:00] Information {"Message":"386880/3125678 index entries parsed."}
[2018-10-24 09:48:31 -07:00] Information {"Message":"872565/3125678 index entries parsed."}
[2018-10-24 09:48:32 -07:00] Information {"Message":"1418704/3125678 index entries parsed."}
[2018-10-24 09:48:32 -07:00] Information {"Message":"1896620/3125678 index entries parsed."}
[2018-10-24 09:48:33 -07:00] Information {"Message":"2389030/3125678 index entries parsed."}
[2018-10-24 09:48:33 -07:00] Information {"Message":"2905393/3125678 index entries parsed."}
[2018-10-24 09:48:33 -07:00] Information {"Message":"Finished parsing 3125678 index entries."}
[2018-10-24 09:48:33 -07:00] AddMissingModifiedFilesAndRemoveThemFromPlaceholderList (Stop) {"DurationMs":3268}
```

*Example run with ~144k placeholders*
```
[2018-10-24 09:41:35 -07:00] AddMissingModifiedFilesAndRemoveThemFromPlaceholderList (Start)
[2018-10-24 09:41:36 -07:00] AddMissingModifiedFilesAndRemoveThemFromPlaceholderList_FilePlaceholderCount {"FilePlaceholderCount":143589}
[2018-10-24 09:41:36 -07:00] Information {"Message":"409300/3125678 index entries parsed."}
[2018-10-24 09:41:37 -07:00] Information {"Message":"830295/3125678 index entries parsed."}
[2018-10-24 09:41:37 -07:00] Information {"Message":"1274830/3125678 index entries parsed."}
[2018-10-24 09:41:38 -07:00] Information {"Message":"1748263/3125678 index entries parsed."}
[2018-10-24 09:41:38 -07:00] Information {"Message":"2196634/3125678 index entries parsed."}
[2018-10-24 09:41:39 -07:00] Information {"Message":"2640680/3125678 index entries parsed."}
[2018-10-24 09:41:39 -07:00] Information {"Message":"3093452/3125678 index entries parsed."}
[2018-10-24 09:41:39 -07:00] Information {"Message":"Finished parsing 3125678 index entries."}
[2018-10-24 09:41:39 -07:00] AddMissingModifiedFilesAndRemoveThemFromPlaceholderList (Stop) {"DurationMs":3868}
```

wilbaker:
> These changes are ready for another look.

It looks like the most recent changes have broken the `UpdateIndexWithCacheInfo()` test, I'm taking a look now.  I suspect we might need to run the background validation task for UpdateIndex as well.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/UpdateIndexTests.cs
```diff
@@ -53,7 +52,11 @@ public void UpdateIndexRemoveAddFileOpenForWrite()
...
 
             // Open Test_ConflictTests/AddedFiles/AddedByBothDifferentContent.txt for write so that it's added to the modified paths database
-            using (FileStream stream = File.Open(Path.Combine(this.Enlistment.RepoRoot, @"Test_ConflictTests\AddedFiles\AddedByBothDifferentContent.txt"), FileMode.Open, FileAccess.Write))
+            using (
+                FileStream stream = File.Open(
+                    Path.Combine(this.Enlistment.RepoRoot, "Test_ConflictTests", "AddedFiles", "AddedByBothDifferentContent.txt"), 
+                    FileMode.Open, 
+                    FileAccess.Write))
             {
                 // TODO 940287: Remove this File.Open once update-index --add\--remove are working as expected
```
derrickstolee: While we are here, is this TODO still accurate?

wilbaker: Let me give it a try with the `File.Open` removed.  Based on the description in the issue I'd guess it's still needed, but it's worth giving it a shot with it removed:

> On a regular git enlistment on NTFS git update-index --remove <file path> will only remove the file from the path if that file is not on disk.

> However, git.exe does not check if the file is on disk if the skip-worktree bit is set.  As a result, on GVFS, git update-index --remove <projected file path> will always remove the file from the path, and this differs from the behavior of a vanilla enlistment on NTFS.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
sanoursa: @wilbaker do we do anything to enforce that second point? Otherwise having both a `GetAllEntries` and `GetAllFileEntries` is going to lead to a bug down the road when someone calls the wrong one.

But also, I don't quite understand the issue. What would happen if we call `GetAllEntries` without calling `WriteAllEntriesAndFlush`? I don't see `GetAllEntries` changing any state, so I would not have known not to call it in this situation.

GVFS/GVFS.Virtualization/Background/FileSystemTask.cs
```diff
@@ -27,7 +27,7 @@ public enum OperationType
             OnFolderRenamed,
             OnFolderDeleted,
             OnFolderFirstWrite,
-            OnIndexWriteWithoutProjectionChange,
+            OnIndexWriteRequiringModifiedPathsValidation,
```
sanoursa: Did the meaning of this event change? Or is this just a rename to make it more descriptive? (I'm wondering if we need an upgrade step for this)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -105,11 +139,22 @@ public unsafe void ParsePath()
...
 
                     byte* forLoopPtr = pathPtr + forLoopStartIndex;
+                    byte* bufferPtr;
+                    int bufferLength;
                     for (int i = forLoopStartIndex; i < this.PathLength + 1; i++)
                     {
                         if (*forLoopPtr == PathSeparatorCode)
                         {
-                            this.PathParts[partIndex] = LazyUTF8String.FromByteArray(pathPtr + currentPartStartIndex, i - currentPartStartIndex);
+                            bufferPtr = pathPtr + currentPartStartIndex;
```
sanoursa: This is a very hot code path, and writing these temp values to variables could cause a measurable perf hit if the compiler thinks it has to write it out from a register into memory. Please measure this to make sure we're not hurting perf here. It's possible that moving the variable scope into this inner block will help the compiler figure it out, but in the end you may be better off duplicating the math and not writing to a variable at all.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -122,7 +167,16 @@ public unsafe void ParsePath()
                     }
 
                     // We unrolled the final part calculation to after the loop, to avoid having to do a 0-byte check inside the for loop
```
sanoursa: This comment is an indication of just how hot the loop above is. A single if statement had a measurable perf impact, so writing to memory will undoubtedly slow it down more than we'd like.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -18,11 +21,26 @@ internal class GitIndexEntry
             private const int MaxParts = MaxPathBufferSize / 2;
             private const byte PathSeparatorCode = 0x2F;
 
+            private static readonly string PathSeparatorString = Path.DirectorySeparatorChar.ToString();
```
sanoursa: It looks like you want the platform-dependent path separator here - should this live on `GVFSPlatform` instead? I think that'll make the intent more clear.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -135,14 +189,29 @@ public void ClearLastParent()
...
 
-            public LazyUTF8String GetChildName()
+            public LazyUTF8String GetLazyChildName()
             {
-                return this.PathParts[this.NumParts - 1];
+                return this.GetLazyPathPart(this.NumParts - 1);
             }
 
-            public string GetFullPath()
+            public string GetGitPath()
```
sanoursa: `GetGitFullPath`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -135,14 +189,29 @@ public void ClearLastParent()
...
             {
-                return string.Join(GVFSConstants.GitPathSeparatorString, this.PathParts.Take(this.NumParts).Select(x => x.GetString()));
+                return this.GetPath(GVFSConstants.GitPathSeparatorString);
+            }
+
+            public string GetRelativePath()
```
sanoursa: `GetPlatformRelativePath`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -1,6 +1,9 @@
 ﻿using GVFS.Common;
+using System;
+using System.IO;
 using System.Linq;
-
+using System.Text;
+
 namespace GVFS.Virtualization.Projection
 {
     public partial class GitIndexProjection
```
sanoursa: The changes in this file feel very complicated. Can you describe the problem that was being solved here?

GVFS/GVFS.Common/PlaceholderListDatabase.cs
wilbaker: > I don't see GetAllEntries changing any state, so I would not have known not to call it in this situation.

The state change and enforcement in `GetAllEntries` is here:

```
() =>
{
    if (this.placeholderDataEntries != null)
    {
        throw new InvalidOperationException("PlaceholderListDatabase should always flush queue placeholders using WriteAllEntriesAndFlush before calling GetAllEntries again.");
    }

    this.placeholderDataEntries = new List<PlaceholderDataEntry>();
}))
```
> do we do anything to enforce that second point?

If you were to call `GetAllEntries` without calling `WriteAllEntriesAndFlush`, whenever the next call to `GetAllEntries` happens (e.g. during a projection change) it will throw.

> But also, I don't quite understand the issue.

At a high level, the issue is that the `GetAllEntries` API was specifically tailored for the "update projection" scenario.  In that scenario `GetAllEntries` and `WriteAllEntriesAndFlush` are always paired.

In this new scenario we just want to delete some entries from the placeholder list (and not write them all back) and so we can't use the `GetAllEntries` getter.

What would you think if for now I renamed `GetAllEntries` to something like `GetAllEntriesAndPrepForWriteAllEntries`?  It's a bit of a mouthful but makes it clear this is not a generic getter.

sanoursa: Thanks, I see it now, though I still can't claim I understand what's happening :-). (Also I just found the comments on the `placeholderDataEntries` variable, but even that doesn't explain it - _why_ are these entries going to be lost?)

Given that the method throws if you call it incorrectly, that's at least some protection. I'm not sure `PrepForWriteAllEntries` explains what's happening, and I don't yet have a better suggestion. I'll keep trying to digest what's happening though.

wilbaker: > but even that doesn't explain it - why are these entries going to be lost?

I had to refresh myself on all this code when I made this change, and I was able to track the "lost entry" scenario down:

1. VFS4G parses the updated index (as part of a projection change)
2. VFS4G starts the work to update placeholders
3. VFS4G calls `GetAllEntries`
4. VFS4G starts updating placeholders
5. Some application reads a pure-virtual file (creating a new placeholder) while VFS4G is updating existing placeholders. (**That new placeholder is added to `placeholderDataEntries`**)
6. VFS4G completes updating the placeholders and calls `WriteAllEntriesAndFlush`. (**Note:** this list does *not* include the placeholders created in step 5, as the were not included in `GetAllEntries`).
7. `WriteAllEntriesAndFlush` writes *both* the entires in `placeholderDataEntries` and those that were passed in as the parameter.

Without `placeholderDataEntries` VFS4G would lose track of the placeholder creations that crossed with updating placeholders.

sanoursa: OK thanks, that makes sense. I would consider adding that explanation in the code, and possibly rename the variable to something like `placeholdersCreatedDuringProjectionUpdate`. As for the method, how about `GetAllEntriesForProjectionUpdate`?

I know this was a detour from the point of this PR, but it was useful for understanding what's happening. Thanks for digging up those details.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
sanoursa: Yea this code is intentionally not thread-safe, because the index parsing always happens on one thread and we didn't want to take that perf hit. Let me try to better understand why there's a race from the background thread and see if I can suggest anything to simplify this down.

wilbaker: As discussed offline, I will try removing all the new checks from `ParsePath` so that we don't impact the hot-path.  Stay tuned for that commit and let me know what you think once it's up.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
jeschu1: Thanks for the long explanation @wilbaker!  I have a better understanding of what's going on now.

kewillford: @wilbaker would it be possible and safer to have a method that takes a delegate like
`Func<List<PlaceholderListDatabase.PlaceholderData>, IEnumerable<PlaceholderData> updatedPlaceholders>`
Then we could hide the fact that these need to be tied together by having the method be something like
```
void UpdatePlaceholders(Func<List<PlaceholderListDatabase.PlaceholderData>, IEnumerable<PlaceholderData> updatedPlaceholders> updateWork)
{
    List<PlaceholderListDatabase.PlaceholderData> placeholders = this.GetAllEntries();
    this.WriteAllEntriesAndFlush(updateWork(placeholders));
}
```
And make the 2 methods private.

GVFS/GVFS.Virtualization/Background/FileSystemTask.cs
kewillford: Are these serialized as `int` values or `string`?  I always forget.  If it is as strings then we will hit the default case in the switch.
```
default:
                     throw new InvalidOperationException("Invalid background operation");	                     throw new InvalidOperationException("Invalid background operation");
```

GVFS/GVFS.Common/PlaceholderListDatabase.cs
wilbaker: >would it be possible and safer to have a method that takes a delegate like
Func<List<PlaceholderListDatabase.PlaceholderData>, IEnumerable<PlaceholderData> updatedPlaceholders>
Then we could hide the fact that these need to be tied together by having the method be something like

@kewillford I think that would be possible and safer, but I think the changes in this PR have become complicated enough that I'd rather not also include refactoring of code (beyond renaming) unrelated to the changes required to fix the issue this PR is addressing.
 
I've filed #426 to make the cleanup you're suggesting.

GVFS/GVFS.Virtualization/Background/FileSystemTask.cs
wilbaker: Fortunately these are serialized as ints:

*Example*

```
E:\os6\.gvfs\databases>type BackgroundGitOperations.dat
A 3 6 ReleaseConfig.dc
A 4 1 foofoo
```

GVFS/GVFS.FunctionalTests/Tests/GitCommands/UpdateIndexTests.cs
wilbaker: Thanks for pointing this out, I was able to remove the `File.Open` as it's no longer needed.  I'm also looking into enabling more of these tests.

GVFS/GVFS.UnitTests/Virtualization/Projection/GitIndexEntryTests.cs
```diff
@@ -2,15 +2,36 @@
...
+
         private const int DefaultIndexEntryCount = 10;
+        private bool shouldTestLazyPaths;
+
+        public GitIndexEntryTests(bool shouldTestLazyPaths)
+        {
+            this.shouldTestLazyPaths = shouldTestLazyPaths;
+        }
+
+        public static object[] ShouldTestLazyPaths
```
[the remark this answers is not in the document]
nickgra: Thanks for educating me about NUnit. I've got a PR out to fix the code I wrote.

wilbaker:
@derrickstolee @jeschu1 @kewillford @sanoursa @nickgra these change are ready for another look.

In the most recent commit I cleaned up naming\comments, updating the update-index functional tests, and make sure that the projection parsing hot path is not impacted by these changes.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -122,7 +146,89 @@ public unsafe void ParsePath()
...
+            }
+
+            /// <summary>
+            /// Parses the path without using LazyUTF8Strings. It should only be called when running a background task.
+            /// </summary>
+            /// <remarks>
+            /// Code in this method has been fine-tuned for performance. None of it is shared with 
+            /// BuildingProjection_ParsePath to avoid overhead.
+            /// </remarks>
+            public unsafe void BackgroundTask_ParsePath()
```
wilbaker: As discussed offline with @sanoursa , this method can use a `StringBuilder` instead of `string[]`, and we don't need to be concerned with parts (like we do when parsing for projection updates)

sanoursa: Related: it would be nice to get the profiling project set up to measure this method too, and get some numbers on it

wilbaker: @sanoursa that's a good idea.  I'll work on that first to get some comparison numbers between the two options.

GVFS/GVFS.UnitTests/Virtualization/Projection/GitIndexEntryTests.cs
```diff
@@ -2,15 +2,36 @@
...
+
         private const int DefaultIndexEntryCount = 10;
+        private bool buildingNewProjection;
+
+        public GitIndexEntryTests(bool shouldTestLazyPaths)
+        {
+            this.buildingNewProjection = shouldTestLazyPaths;
+        }
+
+        public static object[] BuildingNewProjection
```
wilbaker: Merge with master and pick up the common class @nickgra added for this.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
wilbaker: Actually, it looks like we already have this:

```
        /// <summary>
        /// Force the index file to be parsed to add missing paths to the modified paths database.
        /// This method should only be used to measure index parsing performance.
        /// </summary>
        void IProfilerOnlyIndexProjection.ForceAddMissingModifiedPaths(ITracer tracer)
        {
            using (FileStream indexStream = new FileStream(this.indexPath, FileMode.Open, FileAccess.ReadWrite, FileShare.Read, IndexFileStreamBufferSize))
            {
                // Not checking the FileSystemTaskResult here because this is only for profiling
                this.indexParser.AddMissingModifiedFilesAndRemoveThemFromPlaceholderList(tracer, indexStream);
            }
        }
```

I'll make sure this still works and get us some data from test runs.

wilbaker: Perf numbers:

*Commit from 10/25*
Average Time 20 runs - ValidateModifiedPaths 3337.791105 ms

*StringBuilder changes (coming in 10/26) commit*
Average Time 20 runs - ValidateModifiedPaths 1012.179935 ms

wilbaker:
@derrickstolee @jeschu1 @kewillford @sanoursa @nickgra

Please let me know what you think of the latest comment.  I've made some perf improvements to the BG task for validating ModifiedPaths, and added some additional unit tests for non-ASCII file names.

Once I have sign off I'll merge with master and switch to using the `true`/`false` test helper that @nickgra recently added.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -89,7 +102,14 @@ public void RemoveAndFlush(string path)
             }
         }
 
-        public List<PlaceholderData> GetAllEntries()
+        /// <summary>
+        /// Gets all entries and intializes placeholderChangesWhileRebuildingList in preparation fr a call to WriteAllEntriesAndFlush.
```
sanoursa: Spelling: `intializes` -> `initializes` and `fr` -> `for`

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -122,7 +142,14 @@ public List<PlaceholderData> GetAllEntries()
             }
         }
 
-        public void GetAllEntries(out List<PlaceholderData> filePlaceholders, out List<PlaceholderData> folderPlaceholders)
+        /// <summary>
+        /// Gets all entries and intializes placeholderChangesWhileRebuildingList in preparation fr a call to WriteAllEntriesAndFlush.
```
sanoursa: Same comments here

GVFS/GVFS.Common/PlaceholderListDatabase.cs
sanoursa: What you're describing here is internal implementation details of this class, not really relevant to the caller. What's relevant to the caller is: if you call this method and don't call `WriteAllEntriesAndFlush`, you can lose data.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -80,69 +97,136 @@ public unsafe void ParsePath()
...
+                    {
+                        if (*bufferPtrForLoop <= 127)
+                        {
+                            if (*bufferPtrForLoop == PathSeparatorCode)
+                            {
+                                this.backgroundTaskRelativePathBuilder.Append(GVFSPlatform.GVFSPlatformConstants.PathSeparator);
+                            }
+                            else
+                            {
+                                this.backgroundTaskRelativePathBuilder.Append(Convert.ToChar(*bufferPtrForLoop));
```
sanoursa: `Convert.ToChar(byte)` is just doing a cast internally. Do you know if this call is getting inlined or not? It may be a worth a perf test with just a cast in here instead of this method call, to see if it makes a difference.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexParser.cs
```diff
@@ -105,21 +153,48 @@ private FileSystemTaskResult AddToProjection(GitIndexEntry data)
...
+            /// remove enties from filePlaceholders as they are found in the index.  After
+            /// AddEntryToModifiedPathsAndRemoveFromPlaceholdersIfNeeded is called for all entries in the index 
+            /// filePlaceholders will contain only those placeholders that are not in the index.
+            /// </param>
+            private FileSystemTaskResult AddEntryToModifiedPathsAndRemoveFromPlaceholdersIfNeeded(
+                GitIndexEntry gitIndexEntry,
+                Dictionary<string, PlaceholderListDatabase.PlaceholderData> filePlaceholders)
             {
-                if (!data.SkipWorktree)
+                gitIndexEntry.BackgroundTask_ParsePath();
```
sanoursa: At this point, we know that we're only going to use background-task-specific parts of the `GitIndexEntry` class, so we could cast it to a subclass, e.g. `GitIndexEntry.ForBackgroundTask`. That would allow the implementation of that class to group together those related methods and member variables, and not have to use name prefixes to keep them organized.

Though this approach could potentially have its own drawbacks too so take it with a grain of salt. Just an idea that could help clean up the `GitIndexEntry` and make the intent a bit more clear.

wilbaker: > That would allow the implementation of that class to group together those related methods and member variables, and not have to use name prefixes to keep them organized.

After discussing this option with @sanoursa offline we decided that because this option wouldn't come for free at runtime (downcasts must be checked) we're better off leaving this as-is for now and we can revisit the design if we end up needing to expand it further.

PR #411, opened 2018-10-19
Managed interface for Linux projected filesystem

chrisd8088:
Add an initial Linux implementation of a VirtualizationInstance with the same types, callbacks, and functions as the Mac interface.

The underlying dynamic library used by the interoperability layer is not included here due to anticipated licensing requirements, but would be provided separately.

Note that it may be valuable to consider a future refactoring of identical classes between the Linux and Mac implementations (such as PrjFSLib.Linux.Result and PrjFSLib.Mac.Result).

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/NotificationMapping.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/NotificationMapping.cs
@@ -1,0 +1,8 @@
+﻿namespace PrjFSLib.Linux
+{
+    public class NotificationMapping
+    {
+        public NotificationType NotificationMask { get; set; }
+        public string NotificationRelativeRoot { get; set; }
+    }
+}
--- a/ProjFS.Linux/Scripts/Build.sh
+++ b/ProjFS.Linux/Scripts/Build.sh
@@ -17,10 +17,5 @@ PROJFS=$SRCDIR/ProjFS.Linux
 #              into $ROOTDIR/BuildOutput/ProjFS.Linux/Native with
 #              given $CONFIGURATION
 
-# If we're building the Profiling(Release) configuration, remove Profiling() for building .NET code
-if [ "$CONFIGURATION" == "Profiling(Release)" ]; then
-  CONFIGURATION=Release
-fi
-
 dotnet restore $PROJFS/PrjFSLib.Linux.Managed/PrjFSLib.Linux.Managed.csproj /p:Configuration=$CONFIGURATION /p:Platform=x64 --packages $PACKAGES || exit 1
 dotnet build $PROJFS/PrjFSLib.Linux.Managed/PrjFSLib.Linux.Managed.csproj /p:Configuration=$CONFIGURATION /p:Platform=x64 || exit 1
--- a/ProjFS.Linux/Scripts/Build.sh
+++ b/ProjFS.Linux/Scripts/Build.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+CONFIGURATION=$1
+if [ -z $CONFIGURATION ]; then
+  CONFIGURATION=Debug
+fi
+
+SCRIPTDIR=$(dirname ${BASH_SOURCE[0]})
+SRCDIR=$SCRIPTDIR/../..
+ROOTDIR=$SRCDIR/..
+PACKAGES=$ROOTDIR/packages
+
+PROJFS=$SRCDIR/ProjFS.Linux
+
+# TODO(Linux): use pkg-config to look for installed libprojfs;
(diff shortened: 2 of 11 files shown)


ProjFS.Linux/Scripts/Build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+PROJFS=$SRCDIR/ProjFS.Linux
+
+# TODO(Linux): use pkg-config to look for installed libprojfs;
+#              if not found, retrieve latest stable from repo, build
+#              into $ROOTDIR/BuildOutput/ProjFS.Linux/Native with
+#              given $CONFIGURATION
+
+# If we're building the Profiling(Release) configuration, remove Profiling() for building .NET code
+if [ "$CONFIGURATION" == "Profiling(Release)" ]; then
```
sanoursa: I think this is pretty Xcode-specific, you can probably drop it

ProjFS.Linux/PrjFSLib.Linux.Managed/NotificationMapping.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace PrjFSLib.Linux
+{
+    public class NotificationMapping
```
sanoursa: I would drop this file since it's not being used yet. It's not even clear that it'll be needed in the end (we still haven't decided to add it on Mac, for example)

ProjFS.Linux/Scripts/Build.sh
chrisd8088: Sure, will do.

ProjFS.Linux/PrjFSLib.Linux.Managed/NotificationMapping.cs
chrisd8088: OK ... NotificationMapping.cs is over in the Mac implementation; should I take it out of the Linux one or leave it to stay in sync with the Mac?

sanoursa: Ah interesting. I'm actually inclined to take it out of the Mac code, since all we've done there (prematurely it seems) is define the native/managed structs but no functions are using them.

@wilbaker thoughts?

@chrisd8088 no need to block this PR on the decision though, so either way.

chrisd8088: OK, I might leave it for now then to stay in sync.

chrisd8088:
Thanks for the review, @sanoursa -- and before I merge this, can I just check that since I'm treating the features/linuxprototype branch as "master", that means that unless someone else commits to that branch, I don't need to merge from the real master before committing.  Would you agree?

At certain points, though, I guess I'll submit a PR to merge all recent "true master" changes back into the feature branch.  Any advice on that would be appreciated!  Thanks very much, Chris.

sanoursa:
@chrisd8088 Yep exactly right

chrisd8088:
Great, thanks @sanoursa!

It looks like I don't write access to this repo, so can I ask you to do the merge for me?

sanoursa:
@chrisd8088 @wilbaker I'm not sure I agree on refactoring across Mac/Linux :-). While all three ProjFS APIs (Windows, Mac, Linux) look similar and are designed to allow for ease of cross-platform development, they are actually different products. They will even likely end up living in their own separate repos once their code has stabilized. If we were to find that Mac and Linux are so close that they can share a very high percentage of their managed libraries, then that would be worth exploring. But if it's just sharing a couple enums/structs, that does not seem worth it to tightly couple the two platforms together and make future development of each platform more difficult.

So my vote is to hold off on any such refactoring until we can decide if it is actually a net benefit.

wilbaker:
> they are actually different products
> So my vote is to hold off on any such refactoring until we can decide if it is actually a net benefit.

@sanoursa, I agree with all the points you've brought up and that we should hold off on any refactoring of the ProjFS APIs until the end.

Once the work starts in VFS for Git itself starts I think it would make sense to share the POSIX compliant helpers we have (e.g. `FileBasedLock`) in a common base class. Unlike ProjFS on Mac\Linux, VFS for Git is a single product and so I think we can be more proactive about sharing code in its codebase.

sanoursa:
@wilbaker Agreed, we definitely need to factor out the posix-compliant code in the VFS for Git codebase so that can be shared across Mac and Linux. That's something we can do from the get-go as we extend to Linux. My previous comment was exclusively about the ProjFS APIs on each platform.

chrisd8088:
And I would just add that I know the Linux ProjFS API will be changing over the next while as I implement the underlying library.

PR #412, opened 2018-10-20
Mac: Ignore symlinks when 'prefetch --hydrate'-ing

### LINKED ISSUE #405: Mac: 'gvfs prefetch --hydrate' needs to handle symlinks correctly
When calling 'prefetch --hydrate', if the list of files contains symlinks, the command will fail to properly hydrate these files. 

Logs will show errors like so:
```
[2018-10-17 11:42:53 -07:00] Error {"ErrorMessage":"Failed to read /foo/bar/Some.framework/Versions/Current"}
```

nickgra:
Fixes #405.

`ParseFromLsTreeLine` assumes all tree entries with type `blob` are files, so symlinks get added to `FileAddOperations`, which will trigger a `read()` on MacOS. `read()` will fail on directories, regardless of if they're symlinks or not.

Instead of handling symlinks that point to directories differently, ignore all symlinks when calling 'prefetch --hydrate'. We are already prefetched the blobs that they are pointing to so not hydrating the symlink itself seems reasonable. 

The FastFetch tool needs different behavior than 'prefetch --hydrate' when it comes to the DiffHelper as it will need to consider the symlinks when determining if everything has been downloaded.


GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -312,7 +315,12 @@ private void EnqueueOperationsFromDiffTreeLine(ITracer activity, string repoRoot
         }
 
         private bool ShouldIncludeResult(DiffTreeResult blobAdd)
-        {
+        {
+            if (blobAdd.TargetIsSymlink && !this.includeSymlinks)
```
nickgra: I'm not happy with this as it breaks the pattern of the rest of this function, I'm open to suggestions on how revisit this.

GVFS/FastFetch/CheckoutJob.cs
```diff
@@ -39,7 +39,7 @@ public CheckoutJob(int maxParallel, IEnumerable<string> folderList, string targe
         {
             this.tracer = tracer.StartActivity(AreaPath, EventLevel.Informational, Keywords.Telemetry, metadata: null);
             this.enlistment = enlistment;
-            this.diff = new DiffHelper(tracer, enlistment, new string[0], folderList);
+            this.diff = new DiffHelper(tracer, enlistment, new string[0], folderList, true);
```
sanoursa: Use a named parameter for the literal

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -28,6 +28,7 @@ public enum Operations
         public Operations Operation { get; set; }
         public bool SourceIsDirectory { get; set; }
         public bool TargetIsDirectory { get; set; }
+        public bool TargetIsSymlink { get; set; }
```
sanoursa: Nit: elsewhere we've been using "SymLink"

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -229,7 +229,7 @@ public virtual void Prefetch(string branchOrCommit, bool isBranch)
                 }
             }
 
-            DiffHelper blobEnumerator = new DiffHelper(this.Tracer, this.Enlistment, this.FileList, this.FolderList);
+            DiffHelper blobEnumerator = new DiffHelper(this.Tracer, this.Enlistment, this.FileList, this.FolderList, false);
```
sanoursa: Use a named parameter

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
sanoursa: How is it breaking the pattern?

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -24,18 +24,21 @@ public class DiffHelper
         private Enlistment enlistment;
         private GitProcess git;
 
-        public DiffHelper(ITracer tracer, Enlistment enlistment, IEnumerable<string> fileList, IEnumerable<string> folderList)
-            : this(tracer, enlistment, new GitProcess(enlistment), fileList, folderList)
+        private bool includeSymlinks;
```
wilbaker: Same comment that Saeed left in a different file, if we've been using SymLink use that casing here as well

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -44,7 +44,7 @@ public class DiffHelperTests
         public void CanParseDiffForwards()
         {
             MockTracer tracer = new MockTracer();
-            DiffHelper diffForwards = new DiffHelper(tracer, new MockGVFSEnlistment(), new List<string>(), new List<string>());
+            DiffHelper diffForwards = new DiffHelper(tracer, new MockGVFSEnlistment(), new List<string>(), new List<string>(), false);
```
wilbaker: Are there any tests we can add (either functional or unit) to test out the behavior of `includeSymlinks` and the changes to `DiffTreeResult`?

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -66,11 +67,12 @@ public void CanParseDiffForwards()
         }
 
         // Parses Diff B => A
+        // Note that backward.txt does not contain a symlink entry so that we can get coverage on the includeSymLinks false case
```
wilbaker: It might be worth testing `backward.txt` with `includeSymLinks` true and false.

Could you parameterize `DiffHelperTests` so and make `includeSymLinks` a member (that is set by the parameter)?

wilbaker: (You can take a look at `ActiveEnumerationTests` for an example of how to do this)

nickgra: That's a nice pattern, thanks for the pointer.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -13,6 +13,8 @@ public class DiffHelper
     {
         private const string AreaPath = nameof(DiffHelper);
 
+        private static bool shouldIncludeSymLinks;
```
wilbaker: Why was this changed to static?

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -11,9 +11,27 @@
...
+
+        private static readonly bool[] SymLinkSupport = new bool[]
+        {
+            true,
+            false,
+        };
+
+        public DiffHelperTests(bool symLinkSupport)
+        {
+            DiffHelper.SetIncludeSymLinks(symLinkSupport);
```
wilbaker: I can't see how the test is validating any difference in behavior based on `symLinkSupport`.  Can you add a test that validates the changes to `ShouldIncludeResult`? (e.g verifying that `FileAddOperations` does\doesn't contain the symlinks)

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -47,9 +66,9 @@ public void CanParseDiffForwards()
             DiffHelper diffForwards = new DiffHelper(tracer, new MockGVFSEnlistment(), new List<string>(), new List<string>());
             diffForwards.ParseDiffFile(GetDataPath("forward.txt"), "xx:\\fakeRepo");
 
-            // File added, file edited, file renamed, folder => file, edit-rename file
+            // File added, file edited, file renamed, folder => file, edit-rename file, SymLink added
             // Children of: Add folder, Renamed folder, edited folder, file => folder
-            diffForwards.RequiredBlobs.Count.ShouldEqual(9);
+            diffForwards.RequiredBlobs.Count.ShouldEqual(10);
```
wilbaker: Related to above comment on test coverage, will `diffForwards.FileAddOperations` be different based on `SymLinkSupport`?

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -44,12 +65,17 @@ public class DiffHelperTests
...
-            // File added, file edited, file renamed, folder => file, edit-rename file
+            // File added, file edited, file renamed, folder => file, edit-rename file, SymLink added
             // Children of: Add folder, Renamed folder, edited folder, file => folder
-            diffForwards.RequiredBlobs.Count.ShouldEqual(9);
+            diffForwards.RequiredBlobs.Count.ShouldEqual(10);
+
+            if (diffForwards.shouldIncludeSymLinks)
+            {
+                diffForwards.FileAddOperations.ContainsKey("symLinkToBeCreated.txt");
+            }
```
wilbaker: I don't think this check is validating anything as it is right now (there is nothing checking the return value of `ContainsKey`.

Maybe replace this if block with:

```
diffForwards.FileAddOperations.ContainsKey("symLinkToBeCreated.txt").ShouldEqual(diffForwards.shouldIncludeSymLinks);
```

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -44,12 +65,17 @@ public class DiffHelperTests
         public void CanParseDiffForwards()
         {
             MockTracer tracer = new MockTracer();
-            DiffHelper diffForwards = new DiffHelper(tracer, new MockGVFSEnlistment(), new List<string>(), new List<string>());
+            DiffHelper diffForwards = new DiffHelper(tracer, new MockGVFSEnlistment(), new List<string>(), new List<string>(), includeSymLinks: includeSymLinks);
```
wilbaker: Here (and elsewhere in this file) use `this.includeSymLinks`

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -82,6 +83,7 @@ public static DiffTreeResult ParseFromDiffTreeLine(string line, string repoRoot)
             DiffTreeResult result = new DiffTreeResult();
             result.SourceIsDirectory = ValidTreeModes.Contains(parts[0]);
             result.TargetIsDirectory = ValidTreeModes.Contains(parts[1]);
+            result.TargetIsSymLink = parts[1] == "120000";
```
wilbaker: Minor, there is no need to check `parts[1]` again if `result.TargetIsDirectory` is `true`

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -11,45 +11,69 @@
...
+            DiffHelper diffForwards = new DiffHelper(tracer, new MockGVFSEnlistment(), new List<string>(), new List<string>(), includeSymLinks: this.includeSymLinks);
             diffForwards.ParseDiffFile(GetDataPath("forward.txt"), "xx:\\fakeRepo");
 
-            // File added, file edited, file renamed, folder => file, edit-rename file
+            // File added, file edited, file renamed, folder => file, edit-rename file, SymLink added (if applicable)
             // Children of: Add folder, Renamed folder, edited folder, file => folder
-            diffForwards.RequiredBlobs.Count.ShouldEqual(9);
+            diffForwards.RequiredBlobs.Count.ShouldEqual(diffForwards.shouldIncludeSymLinks ? 10: 9);
+
+            // SymLink should only be present if DiffHelper is configured for SymLinks
```
wilbaker: You probably don't need this comment, the check is pretty self explanatory

PR #414, opened 2018-10-22
Implement MirrorProvider support for Linux

chrisd8088:
Add initial MirrorProvider support for Linux based on the Mac implementation, including updates to the VisualStudio solution file to add Linux build targets.

Note that some small changes may be required in the future to accommodate the Linux ProjFS API as it evolves; in particular, a call to StopVirtualizationInstance may be needed in the generic MirrorProvider.  These will be submitted as separate PRs; for now, we just try to sync with the Mac implementation as a starting point.

### DIFF
--- a/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
+++ b/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
@@ -1,0 +1,247 @@
+﻿using PrjFSLib.Linux;
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace MirrorProvider.Linux
+{
+    public class LinuxFileSystemVirtualizer : FileSystemVirtualizer
+    {
+        private VirtualizationInstance virtualizationInstance = new VirtualizationInstance();
+
+        public override bool TryConvertVirtualizationRoot(string directory, out string error)
+        {
+            Result result = VirtualizationInstance.ConvertDirectoryToVirtualizationRoot(directory);
+
+            error = result.ToString();
+            return result == Result.Success;
+        }
+
+        public override bool TryStartVirtualizationInstance(Enlistment enlistment, out string error)
+        {
+            this.virtualizationInstance.OnEnumerateDirectory = this.OnEnumerateDirectory;
+            this.virtualizationInstance.OnGetFileStream = this.OnGetFileStream;
+            this.virtualizationInstance.OnFileModified = this.OnFileModified;
+            this.virtualizationInstance.OnPreDelete = this.OnPreDelete;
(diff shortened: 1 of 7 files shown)


MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            Console.WriteLine($"OnFileRenamed (isDirectory: {isDirectory}) destination: {relativeDestinationPath}");
+        }
+
+        private void OnHardLinkCreated(string relativeNewLinkPath)
+        {
+            Console.WriteLine($"OnHardLinkCreated: {relativeNewLinkPath}");
+        }
+
+        private bool TryGetSymLinkTarget(string relativePath, out string symLinkTarget)
```
wilbaker: I noticed that this appears to be identical to `MacFileSystemVirtualizer.TryGetSymLinkTarget`.

Can `MacFileSystemVirtualizer.TryGetSymLinkTarget` be moved to a common spot rather than duplicating it here?

MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -0,0 +1,247 @@
...
+                // The target needs to be adjusted to point inside the src root
+                symLinkTarget = Path.Combine(
+                    this.Enlistment.SrcRoot.TrimEnd(Path.DirectorySeparatorChar),
+                    symLinkTarget.Substring(this.Enlistment.MirrorRoot.Length).TrimStart(Path.DirectorySeparatorChar));
+            }
+
+            return true;
+        }
+
+        private static byte[] ToVersionIdByteArray(byte version)
```
wilbaker: Same comment for these two methods, does it make sense to move them to a common spot? @sanoursa, and preferences here?

sanoursa: My own preference is to keep the MirrorProvider simple. Rather than refactoring for maximum code reuse, this simpler layout makes it easier to understand how any given platform works.

Creating a common helper class could be fine. Refactoring with base classes and other complications like we're planning to do in VFS for Git seems overkill for the MirrorProvider.

wilbaker: That sounds good to me thanks

chrisd8088: Thanks, @wilbaker and @sanoursa ... and I should note again that I expect I'll have adjustments as we actually start to wire up methods.

chrisd8088:
Thanks for the merge, @sanoursa!  Now I'll be quiet(er) for a bit while I work on the libprojfs side of things ...

PR #418, opened 2018-10-23
Mac: Do not panic when file attributes cannot be read during vnode callback

wilbaker:
WIP for now, after getting in some test runs we can decide how to proceed.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -71,7 +71,8 @@ static bool ShouldHandleVnodeOpEvent(
     FsidInode* vnodeFsidInode,
     int* pid,
     char procname[MAXCOMLEN + 1],
-    int* kauthResult);
+    int* kauthResult,
+    int* kauthError);
 
 static bool ShouldHandleFileOpEvent(
     // In params:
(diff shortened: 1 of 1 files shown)


ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -879,8 +879,17 @@ static uint32_t ReadVNodeFileFlags(vnode_t vn, vfs_context_t context)
...
-    assert(0 == err);
+    if (0 != err)
+    {
+        // TODO(Mac): May fail on some file system types? Perhaps we should early-out depending on mount point anyway.
+        // We should also consider:
+        //   - Logging this error
+        //   - Falling back on vnode lookup (or custom cache) to determine if file is in the root
+        //   - Assuming files are empty if we can't read the flags
+        
+        return 0;
```
sanoursa: I don't think we should return 0 here. Instead, I think we should change the signature of this function to use the Try pattern, and when this fails, explicitly fail the initiating event. As written here, we just silently move on as if the vnode is not ours even if it is, and that can cause other downstream errors, whereas failing the IO will cause a more immediate alarm.

wilbaker: > As written here, we just silently move on as if the vnode is not ours even if it is, and that can cause other downstream errors, whereas failing the IO will cause a more immediate alarm.

What are your thoughts on the best way to handles files here in the FileOp path?  We can't fail the I/O, do you think for now we should continue to panic?

sanoursa: Do we know for sure if this panic is even getting hit by fileop events? As a first pass, it might be interesting to do an experiment where we fail vnode events and panic on fileop and see what happens.

PR #420, opened 2018-10-24
PostFetchJob: Kill running Git process when unmounting

### LINKED ISSUE #409: Issues closing post-fetch job on unmount
We have a variety of things to figure out related to closing the post-fetch job thread in the event of an unmount:

1. We can't abort the thread on Mac. We need a different model to kill the background activity.

2. The ThreadAbort action may not kill the Git process that is running. This could be an existing issue on Windows already.

Things to consider as we investigate these issues: 

* What happens to the post-fetch lock on the shared object cache on thread abort (I'm pretty sure we clean it up on a ThreadAbortException)?

* How can we test that all Git processes are closed as the Mount process closes?

derrickstolee:
The post-fetch job runs up to two Git commands: 'git multi-pack-index write' and 'git commit-graph write'. These can each take up to 30s depending on the amount of work to do. When we unmount, we want to ensure these prcoesses terminate.

Before, we were aborting the post-fetch job thread. This is not cross-platform compliant.

The new approach is to store a `GitProcess` object and pass a Kill signal to that process during unmount. Watch that call for some possible exceptions. Also, add an additional `processLock` to `GitProcess` that holds around our state changes for the process (starting or killing the process) so we don't race by killing a process before it starts and have it start anyway.

As part of revealing this `GitProcess` to the necessary layer, I pulled the implementation of `TryWriteMultiPackIndex` out of the `GitObjects` class.

The test suite was previously exercising this code path during the functional tests, as the functional tests frequently ran faster than the background Git processes. That should give some good data on how this works.

Resolves #409

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -17,12 +17,24 @@ namespace GVFS.Common.Git
         private static readonly Encoding UTF8NoBOM = new UTF8Encoding(false);
         private static bool failedToSetEncoding = false;
 
+        /// <summary>
+        /// Lock taken for duration of running executingProcess.
+        /// </summary>
         private object executionLock = new object();
+
+        /// <summary>
+        /// Lock taken when changing the running state of executingProcess.
+        ///
+        /// Can be taken within executionLock.
+        /// </summary>
+        private object processLock = new object();
 
         private string gitBinPath;
         private string workingDirectoryRoot;
         private string dotGitRoot;
         private string gvfsHooksRoot;
+        private Process executingProcess;
+        private bool stopping;
 
         static GitProcess()
         {
@@ -106,6 +118,35 @@ namespace GVFS.Common.Git
 
             error = null;
             return true;
+        }
+
+        public bool TryKillRunningProcess()
+        {
+            this.stopping = true;
+
+            try
+            {
+                lock (this.processLock)
+                {
+                    Process process = this.executingProcess;
+
+                    if (process != null)
+                    {
+                        process.Kill();
+                        return true;
+                    }
+                }
+            }
+            catch (Win32Exception)
+            {
+                // Thrown when process is already terminating
+            }
+            catch (InvalidOperationException)
+            {
+                // Process already terminated
+            }
+
+            return false;
         }
 
         public virtual void RevokeCredential(string repoUrl)
@@ -496,19 +537,19 @@ namespace GVFS.Common.Git
                 // From https://msdn.microsoft.com/en-us/library/system.diagnostics.process.standardoutput.aspx
                 // To avoid deadlocks, use asynchronous read operations on at least one of the streams.
                 // Do not perform a synchronous read to the end of both redirected streams.
-                using (Process executingProcess = this.GetGitProcess(command, workingDirectory, dotGitDirectory, useReadObjectHook, redirectStandardError: true))
+                using (this.executingProcess = this.GetGitProcess(command, workingDirectory, dotGitDirectory, useReadObjectHook, redirectStandardError: true))
                 {
                     StringBuilder output = new StringBuilder();
                     StringBuilder errors = new StringBuilder();
 
-                    executingProcess.ErrorDataReceived += (sender, args) =>
+                    this.executingProcess.ErrorDataReceived += (sender, args) =>
                     {
                         if (args.Data != null)
                         {
                             errors.Append(args.Data + "\n");
                         }
                     };
-                    executingProcess.OutputDataReceived += (sender, args) =>
+                    this.executingProcess.OutputDataReceived += (sender, args) =>
                     {
                         if (args.Data != null)
                         {
@@ -525,29 +566,39 @@ namespace GVFS.Common.Git
 
                     lock (this.executionLock)
                     {
-                        executingProcess.Start();
-
-                        if (writeStdIn != null)
+                        lock (this.processLock)
                         {
-                            writeStdIn(executingProcess.StandardInput);
+                            if (this.stopping)
+                            {
+                                return new Result(string.Empty, nameof(GitProcess) + " is stopping", Result.GenericFailureCode);
+                            }
+
+                            this.executingProcess.Start();
                         }
 
-                        executingProcess.BeginOutputReadLine();
-                        executingProcess.BeginErrorReadLine();
+                        writeStdIn?.Invoke(this.executingProcess.StandardInput);
 
-                        if (!executingProcess.WaitForExit(timeoutMs))
+                        this.executingProcess.BeginOutputReadLine();
+                        this.executingProcess.BeginErrorReadLine();
+
+                        if (!this.executingProcess.WaitForExit(timeoutMs))
                         {
-                            executingProcess.Kill();
+                            this.executingProcess.Kill();
+
                             return new Result(output.ToString(), "Operation timed out: " + errors.ToString(), Result.GenericFailureCode);
                         }
                     }
 
-                    return new Result(output.ToString(), errors.ToString(), executingProcess.ExitCode);
+                    return new Result(output.ToString(), errors.ToString(), this.executingProcess.ExitCode);
                 }
             }
             catch (Win32Exception e)
             {
                 return new Result(string.Empty, e.Message, Result.GenericFailureCode);
+            }
+            finally
+            {
+                this.executingProcess = null;
             }
         }
 
--- a/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
+++ b/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
@@ -39,6 +39,7 @@ namespace GVFS.Virtualization
         private FileSystemVirtualizer fileSystemVirtualizer;
         private FileProperties logsHeadFileProperties;
         private Thread postFetchJobThread;
+        private GitProcess postFetchGitProcess;
         private object postFetchJobLock;
         private bool stopping;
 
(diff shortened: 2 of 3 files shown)


GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -39,6 +39,7 @@ public class FileSystemCallbacks : IDisposable, IHeartBeatMetadataProvider
         private FileSystemVirtualizer fileSystemVirtualizer;
         private FileProperties logsHeadFileProperties;
         private Thread postFetchJobThread;
+        private GitProcess backgroundGitProcess;
```
sanoursa: The word "background" is heavily overloaded in this codebase, so let's find something different. How about `postFetchGitProcess`?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -584,8 +624,19 @@ private void PostFetchJob(List<string> packIndexes)
...
+                                message: PostFetchTelemetryKey + ": Not launching 'git commit-graph' because the mount is stopping",
+                                keywords: Keywords.Telemetry);
+                            return;
+                        }
+
+                        GitProcess.Result result = this.backgroundGitProcess.WriteCommitGraph(this.context.Enlistment.GitObjectsRoot, packIndexes);
+                        this.backgroundGitProcess = null;
 
                         if (result.HasErrors)
                         {
```
sanoursa: I can't put this comment on the correct line but - do we still need the `ThreadAbortException` catch block below?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -568,12 +590,30 @@ private void PostFetchJob(List<string> packIndexes)
...
+                        {
+                            this.context.Tracer.RelatedWarning(
+                                metadata: null,
+                                message: PostFetchTelemetryKey + ": Not launching 'git multi-pack-index' because the mount is stopping",
+                                keywords: Keywords.Telemetry);
+                            return;
+                        }
+
+                        GitProcess.Result result = this.backgroundGitProcess.WriteMultiPackIndex(this.context.Enlistment.GitObjectsRoot);
+                        this.backgroundGitProcess = null;
```
sanoursa: Why are we setting to null here and then constructing a new `GitProcess` again down at line 627? Can we just use the same instance?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -584,8 +624,19 @@ private void PostFetchJob(List<string> packIndexes)
...
+                                metadata: null,
+                                message: PostFetchTelemetryKey + ": Not launching 'git commit-graph' because the mount is stopping",
+                                keywords: Keywords.Telemetry);
+                            return;
+                        }
+
+                        GitProcess.Result result = this.backgroundGitProcess.WriteCommitGraph(this.context.Enlistment.GitObjectsRoot, packIndexes);
+                        this.backgroundGitProcess = null;
 
                         if (result.HasErrors)
```
sanoursa: HasErrors will now be true if we are stopping. It would be good to check that flag when writing the warning message, so we know that this is an expected error from `commit-graph`

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -568,12 +590,30 @@ private void PostFetchJob(List<string> packIndexes)
...
+                        }
+
+                        GitProcess.Result result = this.backgroundGitProcess.WriteMultiPackIndex(this.context.Enlistment.GitObjectsRoot);
+                        this.backgroundGitProcess = null;
+
+                        if (result.HasErrors)
+                        {
+                            this.context.Tracer.RelatedWarning(
+                                metadata: null,
+                                message: PostFetchTelemetryKey + ": Failed to generate multi-pack-index for new packfiles:" + result.Errors,
```
sanoursa: Same here, record if we are stopping, so that we can be sure this was expected.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -197,11 +198,32 @@ public void Stop()
...
+                        }
+                        else
+                        {
+                            this.context.Tracer.RelatedEvent(
+                                EventLevel.Informational,
+                                PostFetchTelemetryKey + ": failed to kill background Git process during " + nameof(this.Stop),
+                                metadata: null);
+                        }
+                    }
+                    catch (NullReferenceException)
```
sanoursa: This catch statement is too broad because of the amount of code inside the try block. E.g. we've had bugs before where a tracer object was null, and this would hide that bug.

I think you can pretty easily avoid having to catch `NullReferenceException` here using an interlocked compare/exchange.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -108,6 +109,23 @@ public static bool TryGetVersion(string gitBinPath, out GitVersion gitVersion, o
...
+        {
+            try
+            {
+                if (this.executingProcess != null)
+                {
+                    this.executingProcess.Kill();
+                    return true;
+                }
+            }
+            catch (NullReferenceException)
```
sanoursa: Same here, you should be able to drop this if you use an interlocked compare/exchange to set the `executingProcess` to null.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -108,6 +109,23 @@ public static bool TryGetVersion(string gitBinPath, out GitVersion gitVersion, o
             return true;
         }
 
+        public bool KillRunningProcess()
```
sanoursa: Naming nit: rename to `TryKillRunningProcess`

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -108,6 +109,23 @@ public static bool TryGetVersion(string gitBinPath, out GitVersion gitVersion, o
             return true;
         }
 
+        public bool KillRunningProcess()
+        {
+            try
+            {
+                if (this.executingProcess != null)
+                {
+                    this.executingProcess.Kill();
```
sanoursa: What will `Kill` do if the process has already been killed? Will it throw?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
derrickstolee: in this case, we can simply create a local variable and use plain assignment, then refer only to the local copy. A similar pattern can occur in `TryKillRunningProcess()`.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
derrickstolee: We can. The `GitProcess` will handle setting the `Process` it creates to `null` as necessary.

GVFS/GVFS.Common/Git/GitProcess.cs
derrickstolee: It will! I'll add the necessary `catch` statements as specified in [the Process.Kill() docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.kill?view=netframework-4.7.2).

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
derrickstolee: Oops. I misread this the first time, and you are correct, we should still log this but track the stopping variable.

derrickstolee:
@sanoursa Thanks for the review! I think I've responded to everything.

Here are the things that you probably should take second looks at:

1. `Interlocked.CompareExchange()` didn't seem like the right thing for this application, but I focused on using local variables to avoid nullrefs.

2. I added an extra object lock and `stopping` bool in `GitProcess` to handle changing the inner `Process` state (can't `Start()` or `Kill()` without holding that lock). This prevents a `Kill()` from happening before a `Start()`.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -568,12 +581,34 @@ private void PostFetchJob(List<string> packIndexes)
...
+                        {
+                            string message = PostFetchTelemetryKey
+                                + ": Failed to generate multi-pack-index for new packfiles"
+                                + (this.stopping ? " (expected because the process is stopping):" : ":")
+                                + result.Errors;
+
+                            this.context.Tracer.RelatedWarning(
+                                metadata: null,
+                                message: message,
+                                keywords: Keywords.Telemetry);
```
sanoursa: Do we need this to go to telemetry if it's because of a shutdown? It may make more sense to restrict that to actual midx failures.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -584,24 +619,33 @@ private void PostFetchJob(List<string> packIndexes)
...
+                        }
 
-                        if (result.HasErrors)
+                        GitProcess.Result result = this.postFetchGitProcess.WriteCommitGraph(this.context.Enlistment.GitObjectsRoot, packIndexes);
+
+                        if (!this.stopping && result.HasErrors)
                         {
+                            string message = PostFetchTelemetryKey
+                                + ": Failed to generate commit-graph for new packfiles"
+                                + (this.stopping ? " (expected because the process is stopping):" : ":")
```
sanoursa: We already know `this.stopping` is false here (see line 633), so in this case we do not record a log event or telemetry on shutdown. Whatever the right pattern is, let's make sure the two code paths are consistent with each other.

derrickstolee: Oops. I had intended to remove the case from the `if` statement, but if we prefer to not write anything when `stopping`, then I'll leave it in the `if` and remove the tracing.

PR #422, opened 2018-10-24
PostFetchJob: Delete stale .lock files

derrickstolee:
Before writing commit-graph or multi-pack-index, delete the .lock
file that prevents Git from writing to those files. This will
unblock users who have a stale .lock file from a previous early
termination (BSOD, process kill due to unmount, etc.). If the
Git commands are terminating due to legitimate problems, then their
logs will be filled with the correct error instead of a message
about a .lock file preventing the write.

Both of these commands are maintenance commands that users should
not be writing on their own, and we do them under our own file-
based lock.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbWithoutSharedCacheTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbWithoutSharedCacheTests.cs
@@ -277,6 +277,33 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             otherFilePath.ShouldBeAFile(this.fileSystem).WithContents(otherFileContents);
         }
 
+        [TestCase, Order(9)]
+        public void PrefetchCleansUpOphanedLockFiles()
+        {
+            // Multi-pack-index write happens even if the prefetch downloads nothing, while
+            // the commit-graph write happens only when the prefetch downloads at least one pack
+
+            string graphPath = Path.Combine(this.Enlistment.GetObjectRoot(this.fileSystem), "info", "commit-graph");
+            string graphLockPath = graphPath + ".lock";
+            string midxPath = Path.Combine(this.PackRoot, "multi-pack-index");
+            string midxLockPath = midxPath + ".lock";
+
+            this.fileSystem.CreateEmptyFile(graphLockPath);
+
+            // Force deleting the prefetch packs to make the prefetch non-trivial.
+            this.fileSystem.DeleteDirectory(this.PackRoot);
+            this.fileSystem.CreateDirectory(this.PackRoot);
+            this.fileSystem.CreateEmptyFile(midxLockPath);
+
+            this.Enlistment.Prefetch("--commits");
+            this.PostFetchJobShouldComplete();
+
+            this.fileSystem.FileExists(graphLockPath).ShouldBeFalse();
+            this.fileSystem.FileExists(midxLockPath).ShouldBeFalse();
+            this.fileSystem.FileExists(graphPath).ShouldBeTrue();
+            this.fileSystem.FileExists(midxPath).ShouldBeTrue();
+        }
+
         private void PackShouldHaveIdxFile(string pathPath)
         {
             string idxPath = Path.ChangeExtension(pathPath, ".idx");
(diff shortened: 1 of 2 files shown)


derrickstolee:
TODO: add a functional test that verifies we can clear up these stale files.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbWithoutSharedCacheTests.cs
```diff
@@ -277,6 +277,33 @@ public void PrefetchCleansUpStaleTempPrefetchPacks()
             otherFilePath.ShouldBeAFile(this.fileSystem).WithContents(otherFileContents);
         }
 
+        [TestCase, Order(9)]
+        public void PrefetchCleansUpMultiPackIndexLockFile()
```
jrbriggs: naming nit: we're testing both midx and graph lock file cleanup. Perhaps `PrefetchCleansUpOrphanedLockFiles()` or similar?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -585,6 +585,9 @@ private void PostFetchJob(List<string> packIndexes)
 
                     using (ITracer activity = this.context.Tracer.StartActivity("TryWriteMultiPackIndex", EventLevel.Informational, Keywords.Telemetry, metadata: null))
                     {
+                        string midxLockFile = Path.Combine(this.context.Enlistment.GitPackRoot, "multi-pack-index.lock");
```
jrbriggs: Should we have these lock files as named constants, like the post fetch lock file?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -585,6 +587,9 @@ private void PostFetchJob(List<string> packIndexes)
 
                     using (ITracer activity = this.context.Tracer.StartActivity("TryWriteMultiPackIndex", EventLevel.Informational, Keywords.Telemetry, metadata: null))
                     {
+                        string midxLockFile = Path.Combine(this.context.Enlistment.GitPackRoot, MultiPackIndexLock);
+                        this.context.FileSystem.TryDeleteFile(midxLockFile);
```
wilbaker: @derrickstolee how does it behave on Mac\Linux if our TryDeleteFile crosses with an actual read\in-progress MIDX or commit graph write?

derrickstolee: @wilbaker If another Git process is holding the lock with a write handle, then the delete should fail, and then the following Git command that _we_ run will also fail because the lock is held. I'm not sure of the specifics of what will happen with our implementation of TryDeleteFile(), except that I would not expect it to throw (the catch mechanisms in its implementations seem sufficient for this case).

All of this is to say: I don't know for sure, but based on what I can see it looks like it will be fine. At the very least, the only way to hit that case is if a user (or another, non-VFS for Git process) runs one of these commit-graph or multi-pack-index commands.

wilbaker: > If another Git process is holding the lock with a write handle, then the delete should fail, and then the following Git command that we run will also fail because the lock is held.

Have you tested this on Linux or Mac?  I would actually expect the delete to succeed on Mac even if there is an open file descriptor with write access.

`File.Delete` calls `unlink` on Mac (see [FileSystem.Unix.cs](https://github.com/dotnet/corefx/blob/25d197892c37067f45cf97619bb0dfb28413a682/src/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs#L158)).

`unlink` succeeds even if another process has an open file descriptor for the file:

> The unlink() function removes the link named by path from its directory and decrements the link count of the file which was referenced by the link.  If that decrement reduces the link count of the file to zero, and no process has the file open, then all resources associated with the file are reclaimed.  If one or more process have the file open when the last link is removed, the link is removed, but the removal of the file is delayed until all references to it have been closed.

[From the `unlink` manpage]

I have not tested this specific scenario myself, but the behavior above is why the `FileBasedLock` had to be re-written to use `flock` on Mac instead of exclusive handles.

derrickstolee: Ah, then we will stomp on that process (which will fail) and our process will succeed.

Again: we should be the only ones writing these files.

wilbaker: > Again: we should be the only ones writing these files.

Ahh, yep, sorry I missed that we only perform this while holding `postFetchFileLock`.

I agree we'll be good, thanks!

Issue #340, opened 2018-10-04
Mac: Kernel panic when touching FUSE mount


hemancuso:
Using the latest code from Github and building using XCode 10, changing no settings other than signing certificate, I can consistently panic my system [10.13] by loading PrjFSKext and then browsing to a FUSE volume (Google Drive File Stream). I have not been able to get a symbolic backtrace unfortunately. It does panic instantly so it should be easy to replicate.

Thu Oct  4 14:12:11 2018

*** Panic Report ***
panic(cpu 6 caller 0xffffff800c188c2d): Kernel trap at 0xffffff800c2cab24, type 14=page fault, registers:
CR0: 0x000000008001003b, CR2: 0x0000000000000008, CR3: 0x000000041d314015, CR4: 0x00000000003627e0
RAX: 0x0000000000000000, RBX: 0x0000000000000000, RCX: 0x0000000003000000, RDX: 0xffffff803c67f5a0
RSP: 0xffffff91fdf435b0, RBP: 0xffffff91fdf435b0, RSI: 0x0000000000000002, RDI: 0x0000000000000000
R8:  0x0000000000000073, R9:  0x0000000000000003, R10: 0xffffff8034721a40, R11: 0x0000000000000000
R12: 0xffffff803c812c18, R13: 0x0000000000000002, R14: 0xffffff91fdf43728, R15: 0xffffff803273a000
RFL: 0x0000000000010246, RIP: 0xffffff800c2cab24, CS:  0x0000000000000008, SS:  0x0000000000000000
Fault CR2: 0x0000000000000008, Error code: 0x0000000000000000, Fault CPU: 0x6, PL: 0, VF: 0

Backtrace (CPU 6), Frame : Return Address
0xffffff91fdf43060 : 0xffffff800c06e166 
0xffffff91fdf430b0 : 0xffffff800c196714 
0xffffff91fdf430f0 : 0xffffff800c188a00 
0xffffff91fdf43170 : 0xffffff800c020180 
0xffffff91fdf43190 : 0xffffff800c06dbdc 
0xffffff91fdf432c0 : 0xffffff800c06d99c 
0xffffff91fdf43320 : 0xffffff800c188c2d 
0xffffff91fdf434a0 : 0xffffff800c020180 
0xffffff91fdf434c0 : 0xffffff800c2cab24 
0xffffff91fdf435b0 : 0xffffff7f90612d80 
0xffffff91fdf43670 : 0xffffff7f906069b6 
[...]

PR #424, opened 2018-10-25
Mac kext: Fix incorrect use of nullptr as VFS context

pmj:
Bug #340 highlighted that it's not valid to pass a `nullptr` to file I/O kernel APIs which expect a `vfs_context_t`. This change annotates all VFS context function parameters in the kext to require non-null pointers (`_Nonnull`)  and subsequently fixes the instance where a `nullptr` was used in place of a valid context pointer.


pmj:
Obviously, the `_Nonnull` annotations aren't actually required to fix this issue. They're a clang extension to C & C++ (they originate in the Objective-C to Swift bridging layer) and so are definitely nonstandard, but I have my doubts we'll ever compile the kext with anything but Apple's compilers.

Do we want to introduce these annotations into the kext codebase? To some extent, this will catch errors at compile time vs requiring runtime `assert()`s, but they do make the code a bit more noisy.

wilbaker:
> Do we want to introduce these annotations into the kext codebase? To some extent, this will catch errors at compile time vs requiring runtime assert()s, but they do make the code a bit more noisy. Additionally, the compiler will complain if header files are only partially annotated - as soon as you annotate one parameter or return value, it will point out all the non-annotated ones in the same file. (And we have warnings-as-errors enabled, so that will stop the build.)

I personally would rather keep the annotations (unless we hit some scenario down the road where 3rd-party headers force us to stop).  I agree that's unlikely we'll move to a compiler that does not support them.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -257,7 +257,7 @@ static int HandleVnodeOperation(
 
     ProfileSample functionSample(Probe_VnodeOp);
     
-    vfs_context_t context = reinterpret_cast<vfs_context_t>(arg0);
+    vfs_context_t _Nonnull context = reinterpret_cast<vfs_context_t>(arg0);
```
wilbaker: I am a bit surprised the compiler does not complain that `arg0` is not flagged as `_Nonnull`.  Do you know if it's because we're casting?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -575,7 +575,7 @@ static int HandleFileOpOperation(
 
 static bool ShouldHandleVnodeOpEvent(
     // In params:
-    vfs_context_t context,
+    vfs_context_t __nonnull context,
```
sanoursa: Why do we have `__nonnull` (two underscores, lower case) here and `_Nonnull` (one underscore, upper case N) everywhere else? Do they mean anything different?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -144,7 +144,7 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(vnode_t vnode, const Fs
     {
         ProfileSample iterationSample(Probe_VirtualizationRoot_FindIteration);
 
-        rootHandle = FindOrDetectRootAtVnode(vnode, nullptr /* vfs context */, vnodeFsidInode);
+        rootHandle = FindOrDetectRootAtVnode(vnode, vfsContext, vnodeFsidInode);
```
sanoursa: Is this the actual bug fix? I can't tell why this matters though, because even though we're passing the vfsContext in here, it's not being used anywhere inside `FindOrDetectRootAtVnode`. Can we actually remove this parameter?

sanoursa: Note: `FindOrDetectRootAtVnode` passes the context over to `Vnode_ReadXattr`, but that function ignores it completely. So it looks like we can remove it from both.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -132,7 +132,7 @@ kern_return_t VirtualizationRoots_Cleanup()
     return KERN_FAILURE;
 }
 
-VirtualizationRootHandle VirtualizationRoot_FindForVnode(vnode_t vnode, const FsidInode& vnodeFsidInode)
+VirtualizationRootHandle VirtualizationRoot_FindForVnode(vnode_t _Nonnull vnode, const FsidInode& vnodeFsidInode, vfs_context_t _Nonnull vfsContext)
```
sanoursa: And it looks like you can drop context from here too

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Huh! You appear to be correct, I somehow missed that. I guess all the NULL vfs_context cases must already have been dealt with previously. Well, that pretty much defeats the point of this PR…

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: Ah, I missed that one. I originally started putting in the `__nonnull`/`__nullable` in, then remembered that that's the old syntax and [you're supposed to use `_Nonnull` etc. now](https://clang.llvm.org/docs/AttributeReference.html#nullability-attributes) Looks like I missed one in the find-all.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
sanoursa: I'm in favor of compile-time enforcement wherever we can get it

pmj: Alright, no true bugfix then, but I've kept the annotations and removed the obsolete parameter instead of passing the context through.

PR #425, opened 2018-10-25
GitProcess: return true when process is null

derrickstolee:
In #420, we changed how we stopped the background process. However, there was a gap: if the `Process` was `null`, then we returned `false`, causing extra error messages. This fixes that problem and avoids the noise in the logs on unmount.


kewillford:
Do we need any tests for this?

PR #427, opened 2018-10-25
Modify PrepFunctionalTests.sh to work out of the output directories

jrbriggs:
Simplify PrepFunctionalTests by offloading determining exactly which version to install to the build script. PrepFunctionalTests will now assume the one-and-only dmg it finds in $VFS_PUBLISHDIR/Git is the version that needs to be installed to run the tests for $VFS_PUBLISHDIR

This refactoring will help support creating a minimal "functional test enlistment" for Mac.

### DIFF
--- a/Scripts/Mac/BuildGVFSForMac.sh
+++ b/Scripts/Mac/BuildGVFSForMac.sh
@@ -45,13 +45,8 @@ echo 'Copying native binaries to Publish directory...'
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.ReadObjectHook $VFS_PUBLISHDIR || exit 1
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.VirtualFileSystemHook $VFS_PUBLISHDIR || exit 1
 
-echo 'Copying Git installer to Publish directory...'
-GITPUBLISH=$VFS_PUBLISHDIR/Git
-if [[ -d $GITPUBLISH ]] ; then
-  rm -rf $GITPUBLISH
-fi
-mkdir $GITPUBLISH
-cp $GITPATH $GITPUBLISH
+echo 'Copying Git installer to the output directory...'
+$VFS_SCRIPTDIR/PublishGit.sh $GITPATH || exit 1
 
 echo 'Running VFS for Git unit tests...'
 $VFS_PUBLISHDIR/GVFS.UnitTests || exit 1
--- a/Scripts/Mac/BuildGVFSForMac.sh
+++ b/Scripts/Mac/BuildGVFSForMac.sh
@@ -45,4 +45,13 @@ echo 'Copying native binaries to Publish directory...'
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.ReadObjectHook $VFS_PUBLISHDIR || exit 1
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.VirtualFileSystemHook $VFS_PUBLISHDIR || exit 1
 
+echo 'Copying Git installer to Publish directory...'
+GITPUBLISH=$VFS_PUBLISHDIR/Git
+if [[ -d $GITPUBLISH ]] ; then
+  rm -rf $GITPUBLISH
+fi
+mkdir $GITPUBLISH
+cp $GITPATH $GITPUBLISH
+
+echo 'Running VFS for Git unit tests...'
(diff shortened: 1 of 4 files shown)


Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -45,4 +45,13 @@ echo 'Copying native binaries to Publish directory...'
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.ReadObjectHook $VFS_PUBLISHDIR || exit 1
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.VirtualFileSystemHook $VFS_PUBLISHDIR || exit 1
 
+echo 'Copying Git installer to Publish directory...'
+GITPUBLISH=$VFS_PUBLISHDIR/Git
+if [[ -d $GITPUBLISH ]] ; then
+  rm -rf $GITPUBLISH
+fi
+mkdir $GITPUBLISH
+cp $GITPATH $GITPUBLISH
```
sanoursa: Why is it necessary to actually copy? Couldn't we just update the init script to set the `GITPATH` variable appropriately?

Two concerns with this:
* build+publish+unit tests is already getting to be too long. Adding a copy of a large package is just going to add to that. Maybe do the copy only if it's not already there?
* The `VFS_PUBLISHDIR` currently contains just the outputs of `dotnet publish GVFS.sln`. With this change, that publish folder will now contain different contents if I build+publish from VS or build+publish with this script. So if we do need to copy, I'd say copy the Git package to a parallel folder, not inside the same publish folder.

Thoughts?

jrbriggs:
>I'm not really clear on what a "minimal functional test enlistment" is either. Can you elaborate on that?

If you look at https://gvfs.visualstudio.com/ci/_build/results?buildId=2699&view=logs notice how the functional test job doesn't build anymore. It instead sucks down the binaries it needs to run the tests on Windows. I'm starting to restructure the mac code to enable the same thing.

Once we have that working, we can teach the test runner to run one slice and let azure pipelines distribute the job across multiple agents, so we can parallelize the functional test runs.

sanoursa:
@jrbriggs  Yep I've seen that but I'm still a bit fuzzy on it. Maybe I'm just hung up on the word "enlistment". Would it be accurate to say you're just publishing the build artifacts from one job, and then consuming them from another?

jrbriggs:
@sanoursa yes, that's exactly it. It's an "enlistment" insofar as the layout is structured, but I'm happy to drop that verbage.

Scripts/Mac/BuildGVFSForMac.sh
jrbriggs: > Couldn't we just update the init script to set the GITPATH variable appropriately?

I'm not a fan of that, since it means we need to sweep up the packages (which I view as build _input_) and pass it between jobs. I think that if we need something in a package available to end users, the build should promote it to an output folder.

> build+publish+unit tests is already getting to be too long. Adding a copy of a large package is just going to add to that. Maybe do the copy only if it's not already there?

Makes sense to me. I'll make that change.

> So if we do need to copy, I'd say copy the Git package to a parallel folder, not inside the same publish folder.

Also makes sense. Thanks!

sanoursa: To add one more element here - today I can do a clean build+publish from VS, then run the prep script, and then run the functional tests.

With this change, we would now require that the build happen via this build script, at least once. It would be nice to keep the ability to build from VS if possible.

jrbriggs: @sanoursa and I agreed offline that, unfortunately, clean build+publish from VS already doesn't work, and the build script is already a prereq. I'm adding to the problem here. We'll move more of this into the build phase itself in subsequent refactorings, but it's deferred while we work on moving to yaml pipelines.

nickgra: Sounds good to me. Note that I plan to spend some time next week digging into getting our build working well in VSMac (or at least understanding why things don't work well today).

Scripts/Mac/PublishGit.sh
```diff
@@ -0,0 +1,15 @@
...
+
+GITPATH=$1
+INSTALLER=$(basename $GITPATH)
+
+GITPUBLISH=$VFS_OUTPUTDIR/Git
+if [[ ! -d $GITPUBLISH ]] ; then
+  mkdir $GITPUBLISH
+fi
+
+find $GITPUBLISH -type f ! -name $INSTALLER -delete
```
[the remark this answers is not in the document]
jrbriggs: Yes, that's the intent. Since we're copying a versioned installer into a non-versioned directory, I want to ensure that we don't have multiple. The correctness of the installation during PrepFunctionalTests.sh depends on their being exactly the one dmg we want to install in $VFS_OUTPUTDIR/Git

PR #161, opened 2018-08-14
Mac kext: Workaround for Kextlog/IOSharedDataQueue stall issue

pmj:
This is a refresh of my earlier PR - no code changes other than that I've reapplied the patches on latest master.

**Background:**
I did some deep digging and it turned out the stall problems we were having with the logging system were a regression in IOSharedDataQueue itself. When a message is `enqueue()`d on the ring buffer, the writer thread must decide whether the reader process needs to be woken up with a Mach message. The decision is based on whether the ring buffer was empty (head==tail) before enqueueing or was emptied during the enqueue operation as both of those situations imply that the reader thread has run out of messages to dequeue, and thus will likely have gone to sleep.

The bug relates to updates to the `head` and `tail` fields of the `IODataQueueMemory` struct, i.e. the core mechanism of the ring buffer. There seem to be 2 regressions since OS X 10.11.6:

1. In macOS 10.12, the kernel side `IOSharedDataQueue` class was changed to use C11 atomics instead of `OSAtomic` functions from libkern. The C11 atomics chosen mostly use `__ATOMIC_RELAXED` semantics with weak memory ordering guarantees. This causes rare occurrences of the bug.
2. Something seems to have changed on the user space side between 10.12.6 and 10.13.6. I can't see any differences until 10.13.3 (the last version with source) but they would at minimum have been compiled with different compilers. It's possible the `IODataQueueDequeue()` function itself has been changed between 10.13.3 and 10.13.6. In any case, this change makes the stall bug to become extremely frequent.

The changes cause the reader and writer thread to end up with inconsistent views of the shared `IODataQueueMemory` data structure. For example:

1. Reader thread sees `head = 100`, `tail = 200`. This means there are messages to be dequeued.
2. Writer thread wants to enqueue a new message, notes the initial state of the ring buffer is `head = 100`, `tail = 200`.
3. Reader dequeues a message, updates `head` to 200.
4. Writer has queued a new message, updates `tail` to 300.
5. Reader checks for more messages. `head` and `tail` are found to be equal at `200`, the `tail` write of 300 has not propagated to this thread yet.
6. Writer checks whether the ring buffer was emptied during the enqueue operation. `head` is still `100`, as the write has not yet propagated.

This can be avoided by enforcing strict ordering on the reads and writes, which used to be the case in OS X 10.11.6. I guess those OSAtomic functions are being deprecated so the implementation was switched to fancy new C11 atomics.

[...]

PR #429, opened 2018-10-25
Mac PrjFSLib: Use IOSharedDataQueue workaround library if available

pmj:
This allows the Mac PrjFSLib native library and prjfs-log tool to use fixed versions of the `IODataQueueDequeue()` and `IODataQueuePeek()` functions which suffer from severe regressions in macOS 10.13.6 and 10.14. (Darwin 17.7.0 and 18.0)

Assuming you have a [dynamic library containing alternate versions of the aforementioned functions](https://github.com/pmj/SharedDataQueue) and place it in the library search path as "libSharedDataQueue.dylib", this change will cause PrjFSLib and prjfs-log load that library and use the versions of the aforementioned functions it provides instead of the OS's IOKit.framework's.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
@@ -3,6 +3,20 @@
 #include <IOKit/IOKitLib.h>
 #include <iostream>
 #include <mach/mach_port.h>
+#include <sys/utsname.h>
+#include <dlfcn.h>
+
+struct DarwinVersion
+{
+    unsigned long major, minor, revision;
+};
+
+typedef decltype(IODataQueueDequeue)* ioDataQueueDequeueFunctionPtr;
+static ioDataQueueDequeueFunctionPtr ioDataQueueDequeueFunction = nullptr;
+typedef decltype(IODataQueuePeek)* ioDataQueuePeekFunctionPtr;
+static ioDataQueuePeekFunctionPtr ioDataQueuePeekFunction = nullptr;
+
+static void InitDataQueueFunctions();
 
 
 io_connect_t PrjFSService_ConnectToDriver(enum PrjFSServiceUserClientType clientType)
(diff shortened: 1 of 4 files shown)


pmj:
Previous discussion is at #161.

ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
```diff
@@ -122,3 +136,98 @@ bool PrjFSService_DataQueueInit(
...
+    }
+    return ioDataQueuePeekFunction(dataQueue);
+}
+
+
+static bool GetDarwinVersion(DarwinVersion& outVersion)
+{
+    utsname unameInfo = {};
+    if (0 != uname(&unameInfo))
+    {
```
pmj: I'm unfortunately not aware of a nicer way to get the macOS version with just the C system libraries. (as opposed to Objective-C/Cocoa) There's also a "kern.osrelease" sysctl, but that returns the exact same version string as `uname()` that still needs to be parsed.

nickgra: Did you consider `sw_vers -productVersion`? That should give you the macOS version number, not Darwin. You'd still have to do some parsing though.

nickgra:
Thanks @pmj for putting this together. We're having some packaging issues at the moment, but once that's done, I'll get your repo packaged up and set up a PR to consume it. Do you want this PR to go in first?

ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
pmj: That would require shelling out or reading the file `/System/Library/CoreServices/SystemVersion.plist` and subsequently parsing the string anyway as you say. I think a simple syscall is preferable in that regard. The darwin version tends to change on very minor revisions (security updates, supplementary updates, etc.) too, so in theory that would allow us to turn off the workaround if Apple changed their mind and supplied a fix with a supplementary update to 10.13.6.

pmj:
I wouldn't mind getting this into the tree now, as it's one fewer local patch I need to keep maintaining - and the workaround is pretty much essential for preventing the log from seizing up during development/testing. Until the dylib is packaged, we can just build it manually and place it in the same directory as prjfs-log.

PR #435, opened 2018-10-26
[Engineering] Move build dependencies feed.

jrbriggs:
Upgrade the feeds to myget enterprise.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -5,11 +5,11 @@
   </config>
   <packageSources>
     <clear />
-    <add key="GVFSBuildDependencies" value="https://www.myget.org/F/gvfsbuilddependencies/api/v3/index.json" />
+    <add key="VFSForGitBuildDependencies" value="https://vfsforgit.myget.org/F/build-dependencies/api/v3/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
   <activePackageSource>
-    <add key="GVFSBuildDependencies" value="https://www.myget.org/F/gvfsbuilddependencies/api/v3/index.json" />
+    <add key="All" value="(Aggregate source)" />
   </activePackageSource>
   <disabledPackageSources>
     <add key="Microsoft and .NET" value="true" />


nuget.config
```diff
@@ -5,11 +5,12 @@
   </config>
   <packageSources>
     <clear />
-    <add key="GVFSBuildDependencies" value="https://www.myget.org/F/gvfsbuilddependencies/api/v3/index.json" />
+    <add key="VFSForGitBuildDependencies" value="https://vfsforgit.myget.org/F/build-dependencies/api/v3/index.json" />
+    <add key="ArchivedGVFSBuildDependencies" value="https://vfsforgit.myget.org/F/gvfsbuilddependencies/api/v3/index.json" />
```
sanoursa: Why do we need this key? The new feed has all of our latest packages, right? And the older commits still reference the original feed and it's not going away?

nuget.config
```diff
@@ -5,11 +5,12 @@
   </config>
   <packageSources>
     <clear />
-    <add key="GVFSBuildDependencies" value="https://www.myget.org/F/gvfsbuilddependencies/api/v3/index.json" />
+    <add key="VFSForGitBuildDependencies" value="https://vfsforgit.myget.org/F/build-dependencies/api/v3/index.json" />
```
kewillford: I'm not seeing this feed in with my user.  Have we pushed any packages to this new feed?  Do we need to update the builds to push to it?

nuget.config
jrbriggs: `vfsforgit.myget.org/F/build-dependencies` is currently empty.

`vfsforgit.myget.org/F/gvfsbuilddependencies` is what was migrated from `www.myget.org/F/gvfsbuilddependencies`.

I guess another option would be to rename `gvfsbuilddependencies` but I couldn't figure out how to do that in a couple minutes of poking around the myget ui. Given that, I wanted to take this opportunity to move away from the gvfs name.

nuget.config
jrbriggs: Yes, we'll need to update the relevant builds to push to the new feed.

nuget.config
sanoursa: I think we should just migrate the latest version of each package over to the new feed, and drop this archived feed from our code altogether. There's not much value in preserving two feeds here.

jrbriggs: Makes sense. I'll make those changes.

kewillford:
The API key for myget is also used to push packages for the ProjFS team.  I'm assuming with this new feed there will be a new API key to use and we will need to let them know.

nuget.config
jrbriggs: I've moved the packages that I think are needed to build this commit. CI should be able to fact check me :)

nuget.config
jrbriggs: @kewillford do you see the feed when you log into vfsforgit.myget.org?

jrbriggs: I've mirrored the permissions from the old feed. Hopefully that fixes the issue.

PR #437, opened 2018-10-29
Telemetry: Record when a mount succesfully starts

jeschu1:
Add a telemetry marker so we can tell if a mount successfully came up

### DIFF
--- a/GVFS/GVFS.Mount/InProcessMount.cs
+++ b/GVFS/GVFS.Mount/InProcessMount.cs
@@ -134,14 +134,15 @@ namespace GVFS.Mount
                 Console.Title = "GVFS " + ProcessHelper.GetCurrentProcessVersion() + " - " + this.enlistment.EnlistmentRoot;
 
                 this.tracer.RelatedEvent(
-                    EventLevel.Critical,
+                    EventLevel.Informational,
                     "Mount",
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -141,7 +141,8 @@ public void Mount(EventLevel verbosity, Keywords keywords)
                         // Use TracingConstants.MessageKey.InfoMessage rather than TracingConstants.MessageKey.CriticalMessage
                         // as this message should not appear as an error
                         { TracingConstants.MessageKey.InfoMessage, "Virtual repo is ready" },
-                    });
+                    },
+                    Keywords.Telemetry);
```
wilbaker: Should this still be `EventLevel.Critical`, or can we make this Info level? 

CC: @sanoursa. Saeed, do you remember if we perviously got push back for Critical + Telemetry events?

wilbaker: I discussed this with @sanoursa offline and he agreed this should no longer be `Critical`.

jeschu1: Thanks @wilbaker , I updated as suggested

PR #438, opened 2018-10-29
Basic Mac installer

alameenshah:
### How to build installer
1. Run Mac Build ```Scripts/Mac/BuildGVFSForMac.sh```

### How to install
1. ```BuildInstaller.sh``` would create a Pkg (installer package) file in the ```BuildOutput/GVFS.Installer.Mac``` directory. 

screen shot 2018-11-07 at 12 26 32 pm

2. Open Terminal and run the following command to install. ```sudo installer -pkg <BuildOutput_Path>/GVFS.Installer.Mac/bin/x64/Release/netcoreapp2.1/osx-x64/VFSForGit.0.2.173.2.pkg -target /```
3. Double clicking on the installer package ```.pkg``` file would launch the GUI installer wizard.

### Running GVFS
Installer copies all GVFS assemblies (both .Net Core and Native) to the directory ```/usr/local/vfsforgit```. It also installs ```/usr/local/bin/gvfs``` which is a link to ```/usr/local/vfsforgit/gvfs```. ```gvfs``` command should be visible in the Terminal. The ```ProjFS.kext``` has to be manually loaded before cloning a gvfs repository.

### TODO
Add License, Readme text (in an upcoming PR.)

### Test package
[Download Mac Installer](
https://microsoft-my.sharepoint.com/:u:/p/alsha/EYZyNqhMWotHkBGLYAynAFIB1r8llhpg2mlqpS7R4etU0A?e=l81AIn) - The installer is not signed, please follow [these instructions](https://support.apple.com/kb/PH25088?locale=en_US) to install.

### High level description of the BuildInstaller.sh script
buildinstaller

ProductBuild and Code-signing of the installer package is not part of this PR.


jrbriggs:
Could you also reword BuildInstaller.sh to use InitializeEnvironment.sh for most of those shared directories?

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function SetPermissions()
+{
+	chmodCommand="chmod -R 755 \"${INSTALLROOT}\""
+	echo $chmodCommand
+	eval $chmodCommand || exit 1
+}
+
+function CreateInstallerRoot()
```
nickgra: We need to run this code as part of the installation of the package, not as part of the script that creates the package. You can pass pre/postinstall scripts to pkgbuild and they'll be added to the package and ran at the correct times.

jrbriggs:
The dmg should also go to a subdirectory in BuildOutput, I think.

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,121 @@
...
+function CheckBuildIsAvailable()
+{
+	if [ ! -d $BUILDOUTPUT ] || [ ! -d $PUBLISHDIR ]; then
+  		echo "Could not find VFSForGit Build to package."
+  		exit 1
+	fi
+}
+
+function SetPermissions()
+{
```
nickgra: Same thing as CreateInstallerRoot. 

Additionally, I think if a script in pkgbuild creates the directory, the permissions shouldn't need to be changed (not sure if this prototype requires elevation. but if the install directory is owned by root, we might need more changes here)

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function CreateInstaller()
+{
+	pkgBuildCommand="/usr/bin/pkgbuild --identifier com.microsoft.vfsforgit.pkg --root \"${INSTALLROOT}\" \"${ROOTDIR}\"/$INSTALLERPACKAGENAME.pkg"
+	echo $pkgBuildCommand
+	eval $pkgBuildCommand || exit 1
+}
+
+function CreateDiskImage
```
nickgra: I'm against the pattern of creating a .dmg file to house a .pkg. It's a relic of the old days when .pkg files used to not be flat files (they used to be packages/bundles, see https://en.wikipedia.org/wiki/Installer_(macOS)#Installer_package). These days, all we get is a very minor compression in return for increasing the amount of clicks needed for a user to install our software. @pmj might also have opinions on MacOS installers that are different than my experiences, so I'm curious to know where he stands on this.

https://apple.stackexchange.com/questions/30695/why-are-so-many-mac-downloads-a-dmg-just-that-contains-a-single-pkg

FWIW: I think this should also be changed in our Git installer, but that's another story :)

nickgra:
```
The ProjFS.kext has to be manually loaded before cloning a gvfs repository.
```
Do we want to implement a very bare bones LaunchAgent that will automatically load the kext now? It can be as a simple as the one I had in my previous prototype.
I think that's an important convenience that we'll want before we stand up a full GVFS.Service on MacOS.

sanoursa:
Re: creating a barebones LaunchAgent - that sounds like a fine idea, but outside the scope of this PR. Let's use this PR to just figure out how to construct the installer.

alameenshah:
> Could you also reword BuildInstaller.sh to use InitializeEnvironment.sh for most of those shared directories?

Updated to read from InitializeEnvironment.

alameenshah:
> The dmg should also go to a subdirectory in BuildOutput, I think.

Moved the location of DMG to BuildOutput directory.

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,97 @@
...
+
+CreateInstallerRoot
+
+CopyBinariesToInstall
+
+SetPermissions
+
+CreateInstaller
+
+CreateDiskImage
```
sanoursa: Nit: end with a newline, it makes it easier to cat files on the command line

Scripts/Mac/BuildInstaller.sh
nickgra: Ameen and I talked offline, this pattern works fine, it's just different than what I expected :)

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,88 @@
...
+	currentDirectory=`pwd`
+	cd "${INSTALLROOT}/usr/local/bin"
+	linkCommand="ln -s ../vfsforgit/gvfs gvfs"
+	eval $linkCommand
+	cd $currentDirectory
+}
+
+function CreateInstaller()
+{
+	pkgBuildCommand="/usr/bin/pkgbuild --identifier com.microsoft.vfsforgit.pkg --root \"${INSTALLROOT}\" \"${ROOTDIR}\"/$INSTALLERPACKAGENAME.pkg"
```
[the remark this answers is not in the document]
nickgra: It might matter for codesigning, but we can cross that bridge when we get there. 
Should it be io.gvfs or io.vfsforgit?

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,88 @@
...
+{
+	if [ ! -d $VFS_OUTPUTDIR ] || [ ! -d $VFS_PUBLISHDIR ]; then
+  		echo "Could not find VFSForGit Build to package."
+  		exit 1
+	fi
+}
+
+function CreateInstallerRoot()
+{
+	mkdirVfsForGit="mkdir -p \"${INSTALLROOT}/$VFSFORGITDESTINATION\""
```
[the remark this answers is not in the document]
alameenshah: Adding spaces breaks the bash script. [Found some good description](https://unix.stackexchange.com/questions/258727/spaces-in-variable-assignments-in-shell-scripts) of what space means in bash assignments.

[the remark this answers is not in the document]
alameenshah: Yes, it is intentional. I moved this now to BuildOutput folder. The script lays out the files to be installed in a hierarchy. The hierarchy is an exact replica of how & where we want our payload to appear in the target Mac after install. ```productBuild``` tool archives this entire hierarchy (along with permissions and ownership if specified) using ```pax``` tool (I think) and creates the installer package.

pmj:
>> The ProjFS.kext has to be manually loaded before cloning a gvfs repository.
> Do we want to implement a very bare bones LaunchAgent that will automatically load the kext now?
If we're just going to load the kext on boot, we can forego the launch agent (did you mean launch daemon?) and simply install the kext into `/Library/Extensions`, where it will be auto-loaded on boot anyway.

Scripts/Mac/BuildInstaller.sh
pmj: Unless there's a good reason to do otherwise, my vote is for a standalone .pkg as well.

pmj:
As a general comment, you might want to build the final installer .pkg with `productbuild` even at this stage. It's pretty much inevitable that we'll end up doing that anyway, what with adding license agreements, readmes, or other installer steps of interest. [This StackOverflow answer](https://stackoverflow.com/a/11487658/48660) might help with that.

alameenshah:
> As a general comment, you might want to build the final installer .pkg with `productbuild` even at this stage. It's pretty much inevitable that we'll end up doing that anyway, what with adding license agreements, readmes, or other installer steps of interest. [This StackOverflow answer](https://stackoverflow.com/a/11487658/48660) might help with that.

Agree. It is there in the TODO to add the license and release notes. Will add it soon.

alameenshah:
> @alameenshah:
> 
> > > The ProjFS.kext has to be manually loaded before cloning a gvfs repository.
> 
> @nickgra:
> 
> > Do we want to implement a very bare bones LaunchAgent that will automatically load the kext now?
> 
> If we're just going to load the kext on boot, we can forego the launch agent (did you mean launch daemon?) and simply install the kext into `/Library/Extensions`, where it will be auto-loaded on boot anyway.

Wouldn't it need a restart to load the Kext then? We can do a one time load in the installer ```postflight``` and subsequent loads can happen on restart.

Scripts/Mac/BuildInstaller.sh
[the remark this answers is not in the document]
alameenshah: Have removed "microsoft", new text - "com.vfsforgit.pkg"

pmj:
> Wouldn't it need a restart to load the Kext then? We can do a one time load in the installer `postflight` and subsequent loads can happen on restart.

It should in theory load automatically after the kext cache has been updated (done automatically by `installer`/`Installer.app` if installing a kext to /Library/Extensions), due to the way the IOKit personality is set up in the Info.plist. (matching IOResources) I say in theory, because in practice, [user-approved kext loading](https://developer.apple.com/library/archive/technotes/tn2459/_index.html) tends to muddy the waters somewhat.

alameenshah:
> > Wouldn't it need a restart to load the Kext then? We can do a one time load in the installer `postflight` and subsequent loads can happen on restart.
> 
> It should in theory load automatically after the kext cache has been updated (done automatically by `installer`/`Installer.app` if installing a kext to /Library/Extensions), due to the way the IOKit personality is set up in the Info.plist. (matching IOResources) I say in theory, because in practice, [user-approved kext loading](https://developer.apple.com/library/archive/technotes/tn2459/_index.html) tends to muddy the waters somewhat.

Tried it on my Mac (MacBook Pro - 10.13.4), it didn't auto-load. [user-approved kext loading](https://developer.apple.com/library/archive/technotes/tn2459/_index.html) - this would continue to be an issue even if we were to load it using launchd-daemon/installer-postflight.

pmj:
> Tried it on my Mac (MacBook Pro - 10.13.4), it didn't auto-load.

Just to make sure I wasn't going mad, I just copied an IOResources-matching kext on one of my test machines on 10.14 into `/Library/Extensions`, and it worked first time.

So I suppose the question is, *how* did you try it? A kext cache rebuild is kicked off whenever the mtime on `/Library/Extensions` changes. So if you do anything other than an atomic move/rename of the kext package after preparing it elsewhere first, there can be a race condition between the scan for the list of loadable kexts and your kext actually becoming loadable. Of course, a `sudo touch /Library/Extensions` will force a rescan, which is useful if you need to perform more than one action, and at that point the kext should become marked as valid. Except of course if the approved kext loading mechanism has added your invalid-because-partially-installed kext to the naughty list, in which case only deleting the kext completely and reinstalling will fix it. Hence my comment about that.

Installer.app will, by the way, do the right thing in terms of all of the above. (Atomic install with correct permissions before moving into place.)

Note that this *won't* autoload an *updated* version of a kext if another version is already loaded. Here you're right, you either need (a) a reboot, or (b) logic for safely unloading the old and loading the new, specific to the software being installed.

alameenshah:
> > Tried it on my Mac (MacBook Pro - 10.13.4), it didn't auto-load.
> 
> Just to make sure I wasn't going mad, I just copied an IOResources-matching kext on one of my test machines on 10.14 into `/Library/Extensions`, and autoloading worked first time. Approval popup showed up, clicked "Allow", and the service turned up in the IORegistry. (See `ioreg` or IORegistryExplorer.app)
> 
> So I suppose the question is, _how_ did you try it? A kext cache rebuild is kicked off whenever the mtime on `/Library/Extensions` changes. So if you do anything other than an atomic move/rename of the kext package after preparing it elsewhere first, there can be a race condition between the scan for the list of loadable kexts and your kext actually becoming loadable. Of course, a `sudo touch /Library/Extensions` will force a rescan, which is useful if you need to perform more than one action, and at that point the kext should become marked as valid. Except of course if the approved kext loading mechanism has added your invalid-because-partially-installed kext to the naughty list, in which case only deleting the kext completely and reinstalling will fix it. Hence my comment about that.
> 
> Installer.app will, by the way, do the right thing in terms of all of the above. (Atomic install with correct permissions before moving into place.)
> 
> Note that this _won't_ autoload an _updated_ version of a kext if another version is already loaded. Here you're right, you either need (a) a reboot, or (b) logic for safely unloading the old and loading the new, specific to the software being installed.

I updated the installer to install the kext in /Library/Extensions. Sharing here - [Download](https://microsoft-my.sharepoint.com/:u:/p/alsha/EXMRw4GzLsZMjod9Pc2EBcoBrncj6ZewH-hSlc2BXV3x1w?e=7c83YP).

Scripts/Mac/BuildInstaller.sh
alameenshah: Removed DMG.

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -50,3 +50,7 @@ $VFS_SCRIPTDIR/PublishGit.sh $GITPATH || exit 1
 
 echo 'Running VFS for Git unit tests...'
 $VFS_PUBLISHDIR/GVFS.UnitTests || exit 1
+
+echo 'Creating installer package...'
```
nickgra: I think this should remain a separate step for now. For your developer inner loop scenarios, we don't need to build the installer each time.

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
nickgra: Nit: For developer convenience, can we rename this away from starting with Build? I really like tab-completing B<tab>uildGVFSForMac.sh automatically.
Maybe CreateInstaller.sh?

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,89 @@
...
+if [ -z $CONFIGURATION ]; then
+  CONFIGURATION=Debug
+fi
+
+PACKAGEVERSION=$2
+if [ -z $PACKAGEVERSION ]; then
+  PACKAGEVERSION="0.0.0.1"
+fi
+
+INSTALLROOT=$VFS_OUTPUTDIR/"InstallerRoot"
```
sanoursa: There's no need for "Root" in the name of the folder on disk. Please follow the pattern used by the other projects, which is that the `BuildOutput` folder's children are named for the project that built them, and also have their own subfolders for platform and configuration. On Windows, the installer ends up in `$VFS_OUTPUTDIR/GVFS.Installer/bin/x64/$CONFIGURATION`. Using that pattern will also enable us to integrate this installer into our `GVFS.sln` with a csproj that calls out to this script. That will enable the IDE build to create the installer, and it will also give us incremental builds "for free" if we set up the csproj correctly.

sanoursa: Variable naming nit: This is not the install root, it's the build output root for this installer project. This name confused me at first because I thought it represented where the bits will be installed, but it's just the staging area for creating the installer.

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,89 @@
...
+VFSFORGITDESTINATION="usr/local/vfsforgit"
+KEXTDESTINATION="$VFSFORGITDESTINATION"
```
sanoursa: I don't think the kext files should go in the same folder as the rest of the VFS for Git files. At the very least, I would create a subdirectory for them.

sanoursa: Though again, look at the pattern on Windows. We lay down the driver and its lib in a subfolder, though we will also need to copy the lib into a location that allows the GVFS.Mount process to find it.

Scripts/Mac/BuildGVFSForMac.sh
sanoursa: Ideally, this step should be wrapped in a csproj file that is then included in GVFS.sln, similar to what we did on Windows (I would rename the existing project to `GVFS.Installer.Windows` and create a new one called `GVFS.Installer.Mac`). Our "source of truth" for the build should be the sln file, not these build scripts, though that's not 100% satisfied at the moment. This is something @jrbriggs  is working on cleaning up.

Scripts/Mac/BuildInstaller.sh
sanoursa: Agreed on avoiding that collision, though this script should really live in `src/GVFS/GVFS.Installer.Mac`, not here in the Scripts folder. This is a source file that produces a build output, not a helper script.

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
```
sanoursa: Now that I'm thinking about how this script actually integrates into the build, you should not be calling this script. Instead, you should take the necessary parameters as arguments to the script, so that the csproj can pass them in.

Scripts/Mac/BuildInstaller.sh
```diff
@@ -0,0 +1,89 @@
...
+
+INSTALLROOT=$VFS_OUTPUTDIR/"InstallerRoot"
+VFSFORGITDESTINATION="usr/local/vfsforgit"
+KEXTDESTINATION="$VFSFORGITDESTINATION"
+INSTALLERPACKAGENAME="VFSForGit.$PACKAGEVERSION"
+INSTALLERPACKAGEID="com.vfsforgit.pkg"
+
+function CheckBuildIsAvailable()
+{
+    if [ ! -d $VFS_OUTPUTDIR ] || [ ! -d $VFS_PUBLISHDIR ]; then
```
sanoursa: Do we need to be dependent on publishing? Can we set it up to trigger the publish if it hasn't happened yet? What if the build happened, the publish hasn't, and this script is now packaging up stale binaries?

Scripts/Mac/BuildInstaller.sh
alameenshah: Renamed the script to CreateMacInstaller.sh and moved it to inside of GVFS.Installer directory.

screen shot 2018-11-07 at 12 35 03 pm

Scripts/Mac/BuildInstaller.sh
alameenshah: Have created a new installer csproj for Mac, which invokes the CreateMacInstaller.sh script during build. Please take a look.

alameenshah: Renamed to Staging.

Scripts/Mac/BuildInstaller.sh
alameenshah: Updated the script to take arguments from GVFS.Installer.Mac.csproj.

Scripts/Mac/BuildInstaller.sh
alameenshah: This script is now being called from GVFS.Installer.Mac.csproj. The installer creation is hooked on the "Publish" target. Please take a look at the new code and let me know your thoughts.

Scripts/Mac/BuildInstaller.sh
alameenshah: Moved the Kext to /Library/Extensions. Testing /usr/local/lib directory for the libs.

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sanoursa: In Visual Studio for Mac, this project is showing files that are for Windows only (packages.config and Setup.iss).

I think the real issue is that you should create two separate folders for the Mac and Windows installers here, rather than have two csproj files in the same folder. The projects that we have that put two csproj's into the same folder do so because those csproj's build the same source files, just targeted to different runtimes, e.g. GVFS.Mac.csproj and GVFS.Windows.csproj. However, in cases where they build different entirely different source files, we create separate folders for them as well, e.g. GVFS.Platform.Mac and GVFS.Platform.Windows. The installer is much more like the latter example.

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <RootNamespace>GVFS.Installer.Mac</RootNamespace>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <Platforms>x64</Platforms>
+    <RunTimeIdentifiers>osx-x64</RunTimeIdentifiers>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+  </PropertyGroup>
+    
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <Version>$(GVFSVersion)</Version>
+    <NoWarn>2008</NoWarn>
```
sanoursa: What does this line mean? I don't see it in our other projects.

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+  
+  <Target Name="DeleteInstaller" AfterTargets="Clean">
+    <Message Text="Deleting $(TargetDir).." Importance="high"></Message>
+    <RemoveDir Directories="$(TargetDir).." />
+  </Target>
+  
+  <Target Name="AfterCopyFiles" AfterTargets="CopyFilesToOutputDirectory">
+        <Exec Command="/bin/rm -Rf $(TargetDir)*" />
```
sanoursa: What is the purpose of this line? Why are we deleting files here?

And nit: fix the spacing

GVFS/GVFS.Installer/CreateMacInstaller.sh
```diff
@@ -0,0 +1,100 @@
...
+
+BUILDOUTPUTDIR=$3
+if [ -z $BUILDOUTPUTDIR ]; then
+  echo "Build output directory not specified"
+  exit 1
+fi
+
+STAGINGDIR=$BUILDOUTPUTDIR"Staging"
+VFSFORGITDESTINATION="usr/local/vfsforgit"
+KEXTDESTINATION="/Library/Extensions"
```
sanoursa: Why is the kext going into this folder? Won't this cause the kext to auto-load on reboot? We're not ready for that yet.

GVFS/GVFS.Installer/CreateMacInstaller.sh
```diff
@@ -0,0 +1,100 @@
...
+INSTALLERPACKAGENAME="VFSForGit.$PACKAGEVERSION"
+INSTALLERPACKAGEID="com.vfsforgit.pkg"
```
sanoursa: @ethomson can you remind me which domain(s) we registered with the new name?

GVFS/GVFS.Installer/CreateMacInstaller.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
+
+CONFIGURATION=$1
+if [ -z $CONFIGURATION ]; then
```
sanoursa: Why are we allowing optional parameters to this script? The csproj always passes in all the params, and making them optional here just seems to increase the risk of a bug when someone edits the csproj. I'd prefer to blow up if the params aren't exactly right.

sanoursa: And to that point, have you confirmed that an error from this script will fail the csproj build?

GVFS/GVFS.Installer/CreateMacInstaller.sh
ethomson: It is indeed vfsforgit.com.  Thanks for the reminder, I should have some time this sprint to get the site moved over.

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
alameenshah: This is to get the following warning ignored
```
  GVFS.Mac -> /Users/Ameen/Work/GVFS/BuildOutput/GVFS.Mac/bin/x64/Debug/netcoreapp2.1/osx-x64/gvfs.dll
CSC : warning CS2008: No source files specified. [/Users/Ameen/Work/GVFS/VFSForGit/GVFS/GVFS.Installer.Mac/GVFS.Installer.Mac.csproj]
  GVFS.Installer.Mac -> /Users/Ameen/Work/GVFS/BuildOutput/GVFS.Installer.Mac/bin/x64/Debug/netcoreapp2.1/osx-x64/GVFS.Installer.Mac.dll
  GVFS.Tests -> /Users/Ameen/Work/GVFS/BuildOutput/GVFS.Tests/bin/x64/Debug/netcoreapp2.1/GVFS.Tests.dll
  GVFS.FunctionalTests -> /Users/Ameen/Work/GVFS/BuildOutput/GVFS.FunctionalTests/bin/x64/Debug/netcoreapp2.1/osx-x64/GVFS.FunctionalTests.dll
  GVFS.UnitTests -> /Users/Ameen/Work/GVFS/BuildOutput/GVFS.UnitTests/bin/x64/Debug/netcoreapp2.1/osx-x64/GVFS.UnitTests.dll

Build succeeded.

CSC : warning CS2008: No source files specified. [/Users/Ameen/Work/GVFS/VFSForGit/GVFS/GVFS.Installer.Mac/GVFS.Installer.Mac.csproj]
    1 Warning(s)
    0 Error(s)
```

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
alameenshah: ```GVFS.Installer.Mac``` project build produces output assemblies (even though no source code file is included in the project.) I could not reliably override the ```csproj``` ```Compile``` or ```Build``` targets to prevent this generation of un-wanted assembly. The work around is to delete the generated assemblies after the are copied to ```TargetDir```. This is not optimal. Will continue to look for options to prevent the generation of unwanted assembly output.

GVFS/GVFS.Installer/CreateMacInstaller.sh
alameenshah: removed optional.

alameenshah: ```BuildGVFSForMac.sh``` logs red error text in the Console and exits after Publish.

```
  GVFS.Installer.Mac -> /Users/Ameen/Work/GVFS/BuildOutput/GVFS.Installer.Mac/bin/x64/Debug/netcoreapp2.1/osx-x64/GVFS.Installer.Mac.dll
  GVFS.Installer.Mac -> /Users/Ameen/Work/GVFS/Publish/
  /Users/Ameen/Work/GVFS/VFSForGit/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh Debug 0.2.173.2 /Users/Ameen/Work/GVFS/VFSForGit/../BuildOutput/GVFS.Installer.Mac/bin/x64/Debug/netcoreapp2.1/osx-x64/
  Build output directory not specified
/Users/Ameen/Work/GVFS/VFSForGit/GVFS/GVFS.Installer.Mac/GVFS.Installer.Mac.csproj(34,5): error MSB3073: The command "/Users/Ameen/Work/GVFS/VFSForGit/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh Debug 0.2.173.2 /Users/Ameen/Work/GVFS/VFSForGit/../BuildOutput/GVFS.Installer.Mac/bin/x64/Debug/netcoreapp2.1/osx-x64/" exited with code 1.
```

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
alameenshah: Created separate directories for Mac & Windows installer projects.

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
sanoursa: This will also prevent the creation of unnecessary build outputs, so it seems like the right thing to do. Take a look at the prebuild csproj as an example.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -0,0 +1,102 @@
...
+
+function CreateInstaller()
+{
+    pkgBuildCommand="/usr/bin/pkgbuild --identifier $INSTALLERPACKAGEID --root \"${STAGINGDIR}\" \"${BUILDOUTPUTDIR}\"$INSTALLERPACKAGENAME.pkg"
+    eval $pkgBuildCommand || exit 1
+}
+
+function DeleteStaging()
+{
+    rmCommand="/bin/rm -Rf \"${STAGINGDIR}\""
```
sanoursa: Why are we deleting the staging folder? This seems like it would be useful to keep to help with diagnosing installer creation issues?

jrbriggs: @sanoursa I disagree here -- If dotnet publish is laying down those files, I'd rather not be in the business of mucking around with it, even if it does have some extra cruft. If anything, we should file an issue for dotnet to minimize the set of dependencies laid down by publish.

/cc @alameenshah

sanoursa: Yep that's fair. It's definitely simpler and more robust if we take what publish gives us.

alameenshah: keeping the Staging folder.

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
alameenshah: removed the spacing

GVFS/GVFS.Installer/GVFS.Installer.Mac.csproj
alameenshah: Agree. Was able to [figure out](https://stackoverflow.com/questions/47179705/net-core-override-default-build-targets) why my build target overrides were not working.

Now overriding Build, Rebuild and Publish.

GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181018.1</GitPackageVersion>
+    <GitPackageVersion>2.20181107.1</GitPackageVersion>
```
alameenshah: This change is incorrectly shown by GitHub, after I picked changes on master from upstream to my fork and rebased mac_installer_prototype branch on to master.

Use this [URL](https://github.com/Microsoft/VFSForGit/compare/master...alameenshah:mac_installer_prototype) to view the correct diff.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -100,6 +100,7 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
                 { "index.version", "4" },
                 { "merge.stat", "false" },
                 { "merge.renames", "false" },
+                { "pack.useBitmaps", "false" },
```
alameenshah: This change is incorrectly shown by GitHub, after I picked changes in upstream:master to my fork and rebased mac_installer_prototype branch on to master.

Use this URL to view the correct diff.

sanoursa: It looks like you did a rebase, and then did a merge before pushing, so you have some duplicate commits in this PR and that is messing up the diff. You can fix this by dropping the latest merge commit, and force pushing the results of your rebased topic branch.

alameenshah: I had a messed up git commit graph. Also there was a Github issue with identifying the right merge-bases which @derrickstolee investigated (turns out Github support is already aware of the issue).

PR #439, opened 2018-10-29
Mac: Consume libSharedDataQueue.dylib

nickgra:
#429 updated our code to be able to consume both the system provided IOSharedDataQueue and libSharedDataQueue.dylib (which is needed to work around issues described in #161 that effect MacOS <10.14.1). 

This PR adds a script to consume that dylib via a NuGet package and puts it into DYLD_LIBRARY_PATH to be found by dlopen.

For now, we'll call this script as part of PrepFunctionalTests.sh to ensure everyone is on a healthy version of the library. When 10.14.1 is officially released and all of our machines are updated, we can consider removing this call.

### DIFF
--- a/Scripts/Mac/InstallSharedDataQueueDylib.sh
+++ b/Scripts/Mac/InstallSharedDataQueueDylib.sh
@@ -1,0 +1,9 @@
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
+
+BUILDDIR=$VFS_OUTPUTDIR/GVFS.Build
+cp $VFS_SRCDIR/nuget.config $BUILDDIR
+dotnet new classlib -n GVFS.Restore -o $BUILDDIR --force
+dotnet add $BUILDDIR/GVFS.Restore.csproj package --package-directory $VFS_PACKAGESDIR SharedDataQueueDylib --version '1.0.0'
+
+# DYLD_LIBRARY_PATH contains /usr/local/lib by default, so we'll copy this library there.
+cp $VFS_PACKAGESDIR/shareddataqueuedylib/1.0.0/libSharedDataQueue.dylib /usr/local/lib
--- a/Scripts/Mac/InstallSharedDataQueueDylib.sh
+++ b/Scripts/Mac/InstallSharedDataQueueDylib.sh
@@ -1,9 +1,0 @@
-. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
-
-BUILDDIR=$VFS_OUTPUTDIR/GVFS.Build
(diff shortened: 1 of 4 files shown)


Scripts/Mac/InstallSharedDataQueueDylib.sh
```diff
@@ -0,0 +1,9 @@
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
+
+BUILDDIR=$VFS_OUTPUTDIR/GVFS.Build
+cp $VFS_SRCDIR/nuget.config $BUILDDIR
+dotnet new classlib -n GVFS.Restore -o $BUILDDIR --force
+dotnet add $BUILDDIR/GVFS.Restore.csproj package --package-directory $VFS_PACKAGESDIR SharedDataQueueDylib --version '1.0.0'
+
+# DYLD_LIBRARY_PATH contains /usr/local/lib by default, so we'll copy this library there.
+cp $VFS_PACKAGESDIR/shareddataqueuedylib/1.0.0/libSharedDataQueue.dylib /usr/local/lib
```
sanoursa: Nit: end with a newline

Scripts/Mac/InstallSharedDataQueueDylib.sh
```diff
@@ -0,0 +1,9 @@
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
+
+BUILDDIR=$VFS_OUTPUTDIR/GVFS.Build
+cp $VFS_SRCDIR/nuget.config $BUILDDIR
+dotnet new classlib -n GVFS.Restore -o $BUILDDIR --force
+dotnet add $BUILDDIR/GVFS.Restore.csproj package --package-directory $VFS_PACKAGESDIR SharedDataQueueDylib --version '1.0.0'
```
sanoursa: Just some naming nits:
* We're already using `GVFS.Restore.csproj` elsewhere, use a different name here to avoid confusion. And actually, it would be nice to rename both to be more descriptive to what they're actually restoring.
* "SharedDataQueueDyLib" is not very descriptive. I would give it a name that indicates that it's a temporary package with a bug fix.

Scripts/Mac/PrepFunctionalTests.sh
```diff
@@ -37,6 +37,8 @@ fi
 
 git-credential-manager install
 
+$VFS_SCRIPTDIR/InstallSharedDataQueueDylib.sh || exit 1
```
sanoursa: Same here, rename the script to indicate it's installing a bug fix. This usage is currently unclear about _why_ it's happening.

PR #441, opened 2018-10-30
FastFetch: properly handle items with special tokens in path

jamill:
This is a fix for an internally reported issue where FastFetch
created a directory for a file that contained the string " tree ".

FastFetch does not correctly handle paths with certain text. For
example, if a file includes the string " tree ", then FastFetch will
consider this a tree entry, and will create a directory for it.

This change it to only look at the type field from ls-tree when
determining the type of entry.

### DIFF
--- a/GVFS/GVFS.Common/Git/DiffTreeResult.cs
+++ b/GVFS/GVFS.Common/Git/DiffTreeResult.cs
@@ -7,9 +7,10 @@ namespace GVFS.Common.Git
 {
     public class DiffTreeResult
     {
-        public const string TreeMarker = " tree ";
-        public const string BlobMarker = " blob ";
-        public const string CommitMarker = " commit ";
+        public const string TreeMarker = "tree ";
+        public const string BlobMarker = "blob ";
+
+        public const int TypeMarkerStartIndex = 7;
 
         private static readonly HashSet<string> ValidTreeModes = new HashSet<string>() { "040000" };
 
@@ -137,8 +138,7 @@ namespace GVFS.Common.Git
              */
 
             // Everything from ls-tree is an add.
-            int treeIndex = line.IndexOf(TreeMarker);
-            if (treeIndex >= 0)
+            if (IsLsTreeLineOfType(line, TreeMarker))
             {
                 DiffTreeResult treeAdd = new DiffTreeResult();
                 treeAdd.TargetIsDirectory = true;
@@ -149,12 +149,11 @@ namespace GVFS.Common.Git
             }
             else
             {
-                int blobIndex = line.IndexOf(BlobMarker);
-                if (blobIndex >= 0)
+                if (IsLsTreeLineOfType(line, BlobMarker))
                 {
                     DiffTreeResult blobAdd = new DiffTreeResult();
                     blobAdd.TargetIsSymLink = line.StartsWith("120000");
-                    blobAdd.TargetSha = line.Substring(blobIndex + BlobMarker.Length, GVFSConstants.ShaStringLength);
+                    blobAdd.TargetSha = line.Substring(TypeMarkerStartIndex + BlobMarker.Length, GVFSConstants.ShaStringLength);
                     blobAdd.TargetPath = ConvertPathToAbsoluteUtf8Path(repoRoot, line.Substring(line.LastIndexOf("\t") + 1));
                     blobAdd.Operation = DiffTreeResult.Operations.Add;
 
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -167,6 +171,23 @@ public static DiffTreeResult ParseFromLsTreeLine(string line, string repoRoot)
...
+        /// <param name="typeMarker">The type being checked for (e.g. "blob", "tree")</param>
+        /// <returns>True if type of entry matches type specified.</returns>
+        public static bool IsLsTreeLineOfType(string line, string typeMarker)
+        {
+            if (line.Length <= TypeMarkerStartIndex + typeMarker.Length)
+            {
+                return false;
+            }
+
+            return line.IndexOf(typeMarker, TypeMarkerStartIndex, typeMarker.Length, StringComparison.OrdinalIgnoreCase) >= 0;
```
sanoursa: This is the crux of the fix (only look for "tree" at the expected location, not in the name field), but it's not very clear that that's what's happening here. I would change the condition to `== TypeMarkerStartIndex`, not `>=0`, to make that more explicit.

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -167,6 +171,23 @@ public static DiffTreeResult ParseFromLsTreeLine(string line, string repoRoot)
             }
         }
 
+        /// <summary>
```
sanoursa: I don't think this comment block is adding any value. It's just duplicating exactly what the names of the method and params already tell me.

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -7,9 +7,15 @@ namespace GVFS.Common.Git
 {
     public class DiffTreeResult
     {
-        public const string TreeMarker = " tree ";
-        public const string BlobMarker = " blob ";
-        public const string CommitMarker = " commit ";
+        public const string TreeMarker = "tree ";
+        public const string BlobMarker = "blob ";
+
+        /// <summary>
```
sanoursa: A one line comment that shows the format of the line would be more useful here, I think. 5 lines of comments to just say "the type is at position 7" doesn't tell me any more information than the constant declaration itself.

wilbaker:
@jamill I took a look over the changes and have no comments to add (beyond what @sanoursa has already mentioned).

jamill:
Thanks @sanoursa and @wilbaker. I made the suggested, except for the functional tests.

Regarding the functional tests, the way I would see them working is that I would add an example of the problematic path to the repositories that are fetched, and then confirm that they are fetched appropriately. Does that seem reasonable? Would this cause a period of time where the test code and test repository were out of sync, leading to test failures?

wilbaker:
> Would this cause a period of time where the test code and test repository were out of sync, leading to test failures?

This shouldn't cause any problems.  I think you'd just need to create a new `FunctionalTests` with the problematic path in the test repo, and then update the functional test(s) to use that new branch.

jrbriggs:
@sanoursa @jamill I question the value of a functional test duplicating what is covered by the unit tests that are added in this PR. What additional validation would it provide?

/cc @wilbaker

sanoursa:
@jrbriggs the unit tests are great but they're not validating that the bug is fixed when called on real data by the full code paths of `fastfetch` and `gvfs prefetch` (and the two could be manifesting the bug in different ways, we don't actually know right now). I think whenever possible, a bug fix should be accompanied by a _functional_ test that repros the issue.

sanoursa:
@jrbriggs I disagree but I'll leave it up to you. Can we at least test `gvfs prefetch` with the data shape in question and make sure there are no other surprises there?

jrbriggs:
Agreed @sanoursa -- @jamill please make sure we test the new parser manually on prefetch to make sure there aren't any unexpected issues.

wilbaker:
> they're not validating that the bug is fixed when called on real data by the full code paths of `fastfetch` and `gvfs prefetch`

I'll add that historically we've caught breaks with functional tests that unit tests would have missed because the unit tests make assumptions about how the software we're interacting with behaves.

For example, in the future the output of `ls-tree` could change (e.g. maybe we start passing in a new argument that impacts the output format) and we forget to update this parsing code.  

The unit tests would continue to pass whereas the functional tests would catch the issue.

jamill:
Thanks everyone for the feedback. I will verify with prefetch before merging this. I did verify with fast fetch (and found some other behavior that looks incorrect, for which I will file issues for as needed).

PR #445, opened 2018-10-30
Mac: Run GitRepoTests with and without validating the working directory during setup

### LINKED ISSUE #250: Mac: GitRepoTests should be run without expanding all folders in the repo first
Currently `GitRepoTests.SetupForTest` calls `this.Enlistment.RepoRoot.ShouldBeADirectory(this.FileSystem).WithDeepStructure(this.FileSystem, this.ControlGitRepo.RootPath);` which will result in all placeholders in the repo getting created before the test runs.

This gets us *some* coverage, but it would be good to run these tests without `WithDeepStructure` as well to ensure that git commands work properly when placeholders are not already on disk.

TBD: If we should always run the tests both ways, or only without calling `WithDeepStructure`

wilbaker:
Parameterize `GitRepoTests` so that the Mac functional tests are run with and without validation during test setup (so that the working tree is and is not expanded during setup).

Resolves #250

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Program.cs
+++ b/GVFS/GVFS.FunctionalTests/Program.cs
@@ -1,5 +1,6 @@
 ﻿using GVFS.FunctionalTests.Tests;
 using GVFS.FunctionalTests.Tools;
+using GVFS.FunctionalTests.Tests.GitCommands;
 using GVFS.Tests;
 using System;
 using System.Collections.Generic;
@@ -43,6 +44,13 @@ namespace GVFS.FunctionalTests
             {
                 Console.WriteLine("Running the full suite of tests");
 
+                GVFSTestConfig.GitCommandTestWorkTreeValidation =
+                new object[]
+                {
+                    new object[] { GitRepoTests.ValidateWorkingTreeOptions.ValidateWorkingTree },
+                    new object[] { GitRepoTests.ValidateWorkingTreeOptions.DoNotValidateWorkingTree }
+                };
+
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
                     GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.AllWindowsRunners;
@@ -51,11 +59,17 @@ namespace GVFS.FunctionalTests
                 {
                     GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.AllMacRunners;
                 }
+
             }
             else
             {
                 excludeCategories.Add(Categories.FullSuiteOnly);
                 GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.DefaultRunners;
+                GVFSTestConfig.GitCommandTestWorkTreeValidation = 
+                new object[] 
+                { 
+                    new object[] { GitRepoTests.ValidateWorkingTreeOptions.ValidateWorkingTree } 
+                };
             }
 
             if (runner.HasCustomArg("--windows-only"))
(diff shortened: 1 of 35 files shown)


GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -43,6 +44,13 @@ public static void Main(string[] args)
             {
                 Console.WriteLine("Running the full suite of tests");
 
+                GVFSTestConfig.GitCommandTestWorkTreeValidation =
+                new object[]
```
sanoursa: Nit: indent

GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -56,6 +64,11 @@ public static void Main(string[] args)
             {
                 excludeCategories.Add(Categories.FullSuiteOnly);
                 GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.DefaultRunners;
+                GVFSTestConfig.GitCommandTestWorkTreeValidation = 
+                new object[]
```
sanoursa: Indent

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -13,6 +13,8 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
     [TestFixture]
     public abstract class GitRepoTests
     {
+        public const string ValidateWorkingTree = "WorkTreeValidation";
```
sanoursa: Is this redirection buying us anything? Can we just use `nameof(WorkTreeValidation)` in all of the callers?

And actually, why not just reference `TestFixtureSource(typeof(GVFSTestConfig), nameof( GVFSTestConfig.GitCommandTestWorkTreeValidation)` directly?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -23,13 +25,27 @@ public abstract class GitRepoTests
...
 
-        public GitRepoTests(bool enlistmentPerTest)
+        public GitRepoTests(bool enlistmentPerTest, ValidateWorkingTreeOptions validateWorkingTree)
         {
             this.enlistmentPerTest = enlistmentPerTest;
+            this.validateWorkingTree = validateWorkingTree == ValidateWorkingTreeOptions.ValidateWorkingTree;
             this.FileSystem = new SystemIORunner();
         }
 
+        public enum ValidateWorkingTreeOptions
```
sanoursa: What is the enum buying us? Can we simplify the pattern even further with the existing `AllBools` data source?

sanoursa: By that argument, we should drop the `AllBools` data source and use enums everywhere instead :-). I'm not opposed to doing it either way, but I'd like to simplify our usage of this NUnit API if we can.

wilbaker: I think I might be putting too much value into the pretty printing.  

My guess is that when tests fail we'll be most interested in the specific check that failed (and not care as much whether or not we validated the working tree during setup).

Now I'm thinking I'll take this out and stick with `AllBools` for simplicity.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/HashObjectTests.cs
```diff
@@ -28,7 +29,7 @@ public void CanReadFileAfterHashObject()
             // TODO 1087312: Fix 'git hash-oject' so that it works for files that aren't on disk yet
             GitHelpers.InvokeGitAgainstGVFSRepo(
                 this.Enlistment.RepoRoot,
-                "hash-object " + fileName);
+                "hash-object " + fileName.Replace("\\", "/"));
```
kewillford: Do we have `const` values or a helper for this replace?

PR #447, opened 2018-10-30
Add ability to specify an iKey for the ETW listener

kewillford:


wilbaker:
What is an iKey used for?

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
```diff
@@ -85,62 +83,60 @@ public override void Dispose()
...
-            if (jsonPayload != null)
+            if (string.IsNullOrEmpty(jsonPayload))
             {
-                JsonPayload payload = new JsonPayload(jsonPayload, this.enlistmentId, this.mountId);
+                jsonPayload = "{}";
+            }
+
+            if (string.IsNullOrEmpty(this.ikey))
+            {
+                var payload = new
```
wilbaker: Nit: No `var`

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
```diff
@@ -85,62 +83,60 @@ public override void Dispose()
...
+                    Json = jsonPayload,
+                    EnlistmentId = this.enlistmentId,
+                    MountId = this.mountId,
+                };
                 this.eventSource.Write(eventName, ref options, ref activityId, ref parentActivityId, ref payload);
             }
             else
             {
-                Payload payload = new Payload(this.enlistmentId, this.mountId);
+                var payload = new
```
wilbaker: `var` is here too

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
```diff
@@ -85,62 +83,60 @@ public override void Dispose()
...
+                {
+                    PartA_iKey = this.ikey,
+                    Json = jsonPayload,
+                    EnlistmentId = this.enlistmentId,
+                    MountId = this.mountId,
+                };
                 this.eventSource.Write(eventName, ref options, ref activityId, ref parentActivityId, ref payload);
```
wilbaker: The call to `this.eventSource.Write` is the same in both the `if` and the `else` block, can it be moved out?

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
kewillford: This is an anonymous type now and has to use `var`.

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
kewillford: Because it is using an anonymous type the `Write` is not using the same type for the last parameter.

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
sanoursa: Do you have to use an anonymous type here? Is there a perf hit from using a dynamic type here? This code gets executed in perf critical code paths, so let's make sure we're not causing any issues with that.

wilbaker: This might be a bit overkill, but is there a reason it has to be an anonymous type (and did the `Payload` types have to be removed)?

kewillford: >Do you have to use an anonymous type here?

No we could continue to use the explicit types as before
>Is there a perf hit from using a dynamic type here?

An anonymous type is not the same thing as a dynamic type. It is exactly the same as an explicit type.  There is just compiler magic going on creating the type declaration for you.  So there would be no performance difference with the explicit types.

> is there a reason it has to be an anonymous type (and did the Payload types have to be removed)?

There is no reason it _has_ to be an anonymous type but the `EventData` types that were created were only and only ever would be used here and nowhere else.  So it seems like more maintenance to have to update both the class and the use of the class instead of updating one place.  I'm not taking a hardline stance here so if you still want the explicit types that is fine with me.

sanoursa: To me it seems like a small amount of maintenance work to keep the explicit types, and that helps (at least in my brain) with the ability to read the code down the line. I always like to optimize for ease of readability, so that's my vote.

PR #451, opened 2018-11-01
Mac: Don't assert/panic when we fail to read attributes during a FileOp operation

nickgra:
Historically, many of us have seen a panic on reading the attempting to read the attributes on some path outside of our virtualization root (Most reports of this have been when no virtualization root is attached). For now, let's avoid the panic here and log the path where the failure occurred.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -500,8 +500,14 @@ static int HandleFileOpOperation(
             goto CleanupAndReturn;
         }
 
+        // We used to assert if we couldn't read the attributes of a file here.
+        // In practice, we always found that these panics were on accesses outside of a
+        // virtualization root, so we'll log as opposed to panicing.
         bool fileFlaggedInRoot;
-        assert(TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot));
+        if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
+        {
+            KextLog_Info("Failed to read attributes when handling FileOp operation. Path is %s", path);
+        }
         
         if (fileFlaggedInRoot && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
         {


ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -500,8 +500,14 @@ static int HandleFileOpOperation(
             goto CleanupAndReturn;
         }
 
+        // We used to assert if we couldn't read the attributes of a file here.
```
sanoursa: I don't think it's very useful to write a comment that says what the code "used to" do. These sorts of comments just end up being confusing later on.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -500,8 +500,14 @@ static int HandleFileOpOperation(
...
+        // In practice, we always found that these panics were on accesses outside of a
+        // virtualization root, so we'll log as opposed to panicing.
         bool fileFlaggedInRoot;
-        assert(TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot));
+        if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
+        {
+            KextLog_Info("Failed to read attributes when handling FileOp operation. Path is %s", path);
+        }
```
sanoursa: `goto CleanupAndReturn` here. We don't want to do anything more with this event.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -500,8 +500,14 @@ static int HandleFileOpOperation(
...
         }
 
+        // We used to assert if we couldn't read the attributes of a file here.
+        // In practice, we always found that these panics were on accesses outside of a
+        // virtualization root, so we'll log as opposed to panicing.
         bool fileFlaggedInRoot;
-        assert(TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot));
+        if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
+        {
+            KextLog_Info("Failed to read attributes when handling FileOp operation. Path is %s", path);
```
wilbaker: Can you add a `// Mac(TODO)` comment here that this `KextLog_Info` should be added to the kext log file?

wilbaker: (I suspect that once we have a log file implementation we won't want to include *all* events in it, but this seems like one we will want)

sanoursa: I would argue that we _do_ want to capture all kext log statements into the log file. Why else would we record them?

nickgra: I agree with Saeed, if you look at our existing use of KextLog, we log all those messages today and I can't imagine when we wouldn't want all of these messages to go to the logger.
If it ends up being problematic down the line, we could always revisit.

wilbaker: > Why else would we record them?

If we wanted to take the same approach as ProjFS on Windows a subset of our logging calls would always be written to file and the rest would be disabled unless a user enables verbose logging (to track down specific issues they might have encountered).

> we log all those messages today and I can't imagine when we wouldn't want all of these messages to go to the logger.

Perhaps we start with only logging error and\or critical events?  

I think there's going to be a balance between log file size, and how far back we can go in log file history (assuming we do circular logging).

wilbaker: (All this logging discussion is outside the scope of this PR, so I don't think there's any need to block on it)

PR #461, opened 2018-11-01
Post index changed hook

### LINKED ISSUE #413: Mac: Rebasing with conflicts creates temporary file (does not happen without VFS4G)
*To reproduce*
- Run `RebaseConflictTests.RebaseConflict` (or do the steps manually)

*Output without VFS for Git*
> CONFLICT (modify/delete): Test_ConflictTests/ModifiedFiles/ChangeInTargetDeleteInSource.txt deleted in HEAD and modified in Changes to the target branch for testing various git commands. Version Changes to the target branch for testing various git commands of Test_ConflictTests/ModifiedFiles/ChangeInTargetDeleteInSource.txt **left in tree**.

*Output with VFS for Git*
> CONFLICT (modify/delete): Test_ConflictTests/ModifiedFiles/ChangeInTargetDeleteInSource.txt deleted in HEAD and modified in Changes to the target branch for testing various git commands. Version Changes to the target branch for testing various git commands of Test_ConflictTests/ModifiedFiles/ChangeInTargetDeleteInSource.txt **left in tree at Test_ConflictTests/ModifiedFiles/ChangeInTargetDeleteInSource.txt~Changes to the target branch for testing various git commands.**

After taking a look at the git code it looks like these messages are coming from `handle_change_delete`:

```
/*
    * Despite the four nearly duplicate messages and argument
    * lists below and the ugliness of the nested if-statements,
    * having complete messages makes the job easier for
    * translators.
    *
    * The slight variance among the cases is due to the fact
    * that:
    * 1) directory/file conflicts (in effect if
    * !alt_path) could cause us to need to write the
    * file to a different path.
    * 2) renames (in effect if !old_path) could mean that
    * there are two names for the path that the user
    * may know the file by.
    */
if (!alt_path) {
    if (!old_path) {
        output(o, 1, _("CONFLICT (%s/delete): %s deleted in %s "
                "and %s in %s. Version %s of %s left in tree."),
                change, path, delete_branch, change_past,
                change_branch, change_branch, path);
    } else {
        output(o, 1, _("CONFLICT (%s/delete): %s deleted in %s "
                "and %s to %s in %s. Version %s of %s left in tree."),
                change, old_path, delete_branch, change_past, path,
                change_branch, change_branch, path);
    }
} else {
    if (!old_path) {
        output(o, 1, _("CONFLICT (%s/delete): %s deleted in %s "
                "and %s in %s. Version %s of %s left in tree at %s."),
                change, path, delete_branch, change_past,
                change_branch, change_branch, path, alt_path);
    } else {
        output(o, 1, _("CONFLICT (%s/delete): %s deleted in %s "
                "and %s to %s in %s. Version %s of %s left in tree at %s."),
                change, old_path, delete_branch, change_past, path,
                change_branch, change_branch, path, alt_path);
    }
}
```

And so perhaps the issue is with `would_lose_untracked` (which is part of deciding which path git takes in the code above)?

```
     * This may look like it can be simplified to:
     * return !was_tracked(o, path) && file_exists(path)
     * but it can't. This function needs to know whether path was in
     * the working tree due to EITHER having been tracked in the index
     * before the merge OR having been put into the working copy and
     * index by unpack_trees(). Due to that either-or requirement, we
     * check the current index instead of the original one.
```

kewillford:
These are the initial changes for allowing the post-indexchanged hook.  This works by have a hook in git that will be called whenever git has finished renaming the index.lock to index.  
There are 2 arguments passed to this hook:
1. Indicator if the index change was caused by git updating the working directory
2. Indicator if the index change was caused by a `reset --mixed`
Neither of these indicators are used in the current code changes.  It continues to rely on the lock data to determine what command is running and the code path to take.

This is not necessarily where this should end up, it was just the fastest change to test and see what issues would be caused.  

The main change to get this to work was in GitIndexProjection adding a method (`AllowIndexToBeParsed`) to allow the background index parsing and placeholder updates to run without actually releasing the lock.  The lock release continues to happen in the post-command.

Fixes #413 

Todo:
- [x] Make sure the xcode project changes are right
- [x] See what it would take to use the indicators from the hook instead of the lock data
- [x] Find a way to report the updated placeholder errors.  Running on my machine the `WindowsUpdatePlaceholderTests` were the only ones that were failing.


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -229,6 +229,18 @@ public virtual void Shutdown()
             this.indexParsingThread.Wait();
         }
 
+        public void AllowIndexToBeParsed()
```
kewillford: Have `TryReleaseExternalLock` call this method.

GVFS.sln
```diff
@@ -55,9 +55,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GVFS.Service", "GVFS\GVFS.S
 		{2D23AB54-541F-4ABC-8DCA-08C199E97ABB} = {2D23AB54-541F-4ABC-8DCA-08C199E97ABB}
 		{5A6656D5-81C7-472C-9DC8-32D071CB2258} = {5A6656D5-81C7-472C-9DC8-32D071CB2258}
 		{BDA91EE5-C684-4FC5-A90A-B7D677421917} = {BDA91EE5-C684-4FC5-A90A-B7D677421917}
+		{24D161E9-D1F0-4299-BBD3-5D940BEDD535} = {24D161E9-D1F0-4299-BBD3-5D940BEDD535}
 	EndProjectSection
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "GVFS.ReadObjectHook", "GVFS\GVFS.ReadObjectHook\GVFS.ReadObjectHook.Windows.vcxproj", "{5A6656D5-81C7-472C-9DC8-32D071CB2258}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "GVFS.ReadObjectHook.Windows", "GVFS\GVFS.ReadObjectHook\GVFS.ReadObjectHook.Windows.vcxproj", "{5A6656D5-81C7-472C-9DC8-32D071CB2258}"
```
kewillford: Not sure why VS changed this.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -229,6 +229,18 @@ public virtual void Shutdown()
             this.indexParsingThread.Wait();
         }
 
+        public void AllowIndexToBeParsed()
+        {
+            // Inform the index parsing thread that git has indicated the index has changed,
+            // which means the projection needs to be updated to be correct before git proceeds
+            this.externalLockReleaseRequested.Set();
+
+            this.ClearNegativePathCacheIfPollutedByGit();
```
sanoursa: Do we actually want to call this here? Or wait until the git command has completed?

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -339,6 +343,28 @@ private void HandleReleaseLockRequest(string messageBody, NamedPipeServer.Connec
...
+            }
+            else if (this.currentState != MountState.Ready)
+            {
+                response = new NamedPipeMessages.PostIndexChanged.Response(NamedPipeMessages.MountNotReadyResult);
+            }
+            else
+            {
+                this.fileSystemCallbacks.OnIndexFileChanged(request.WorkingDirectoryUpdated, request.OnlyIndexUpdated);
+                this.fileSystemCallbacks.UpdateIndexProjection();
+                response = new NamedPipeMessages.PostIndexChanged.Response(NamedPipeMessages.PostIndexChanged.SuccessResult);
```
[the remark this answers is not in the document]
kewillford: Yes but the failures to update placeholders are tracked and then output at the end of the command and what would be the response if it did fail?  Throw and exception and crash?  git wouldn't care because it has already renamed the index.

GVFS/GVFS.PostIndexChangedHook/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+        die(PostIndexChangedErrorReturnCode::ErrorPostIndexChangedProtocol, "Invalid value passed for second argument");
+    }
+
+    DisableCRLFTranslationOnStdPipes();
+
+    PATH_STRING pipeName(GetGVFSPipeName(argv[0]));
+    PIPE_HANDLE pipeHandle = CreatePipeToGVFS(pipeName);
+
+    // Construct index changed request message
+    // Format:  "PICN|<working directory updated flag><only index updated flag>"
```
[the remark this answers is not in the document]
kewillford: I think the second flag name needs to be changed.  It used to be the `reset --mixed` flag to indicate that there was a change to the index and it needed to parse the index to update the modified paths instead of the placeholders, but `update-index` was also a command that needs to pass this flag.  The `00` is for commands like `add`, `commit`, etc that update the index and don't need to parse the index but VFSForGit still needs the notification to clear the `newlyCreatedFileAndFolderPaths` list.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -358,15 +363,14 @@ public IEnumerable<string> GetAllModifiedPaths()
             return this.modifiedPaths.GetAllModifiedPaths();
         }
 
-        public virtual void OnIndexFileChange()
+        public virtual void CheckLockWhenIndexChanged()
```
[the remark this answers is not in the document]
kewillford: The only purpose of this method now is to detect when the index is changed/renamed without the lock so the projection can be updated.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -358,15 +363,14 @@ public IEnumerable<string> GetAllModifiedPaths()
...
         {
             string lockedGitCommand = this.context.Repository.GVFSLock.GetLockedGitCommand();
             GitCommandLineParser gitCommand = new GitCommandLineParser(lockedGitCommand);
             if (!gitCommand.IsValidGitCommand)
             {
                 // Something wrote to the index without holding the GVFS lock, so we invalidate the projection
-                this.GitIndexProjection.InvalidateProjection();
-                this.InvalidateGitStatusCache();
+                this.OnIndexFileChanged(workingDirectoryUpdated: true, onlyIndexUpdated: false);
```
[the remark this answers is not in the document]
kewillford: We don't.  We only know that something renamed the index without having the lock so we invalidate the projection which is the `workingDirectoryUpdated` flag.  These flags correspond with what comes from git but maybe here they should be renamed to something more appropriate for the context like `invalidateProjection` and `invalidateModifiedPaths`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
kewillford: This is only not async because the code has a `Wait`.  It is allowing the index parsing thread to run and parse the index.  What would be a better name? `AllowIndexParsingThreadToRun`,  `ReleaseIndexParsingThreadWait`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -229,6 +229,18 @@ public virtual void Shutdown()
             this.indexParsingThread.Wait();
         }
 
+        public void AllowIndexToBeParsed()
+        {
+            // Inform the index parsing thread that git has indicated the index has changed,
+            // which means the projection needs to be updated to be correct before git proceeds
+            this.externalLockReleaseRequested.Set();
```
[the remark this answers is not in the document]
kewillford: Yes the way it is being used now is not just an external lock release request so I will think of a better name for it.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: Hmm. I'm not sure.  I was keeping the code sequence together but this is to clear the negative path cache that the driver keeps.  Seems like we still want this to happen whenever we update the projection.  @wilbaker what are your thoughts?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1107,8 +1119,6 @@ private void ClearUpdatePlaceholderErrors()
 
         private void UpdatePlaceholders()
         {
-            this.ClearUpdatePlaceholderErrors();
```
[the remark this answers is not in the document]
kewillford: Because we are updating placeholders multiple times within one command we need to keep this error list around until the command ends so that all the update placeholder errors will be output for the user.  If we clear here, one index changed hook notification would clear it and it would not be output.  We could output with the hook being called but then our output could be intermixed with git's output which would be messy.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
kewillford: Looking over the code does `parseIndexRequested` make this more understandable?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: I am only just starting to look at the changes, but my initial thought is that it would be OK to wait for the git command to complete (but functionally it doesn't hurt to call it whenever we update the projection).

The reason that we need this call is because we're blocking git from being able to create placeholders (and we don't want the blocked paths to still be in the cache after the git command completes).  If this is a projection change in the middle of a git command, VFS4G will continue to block git and so we probably don't need to clear the cache (assuming the GVFSLock is released when rebase stops due to conflicts).

GVFS/GVFS.Common/FileSystem/HooksInstaller.cs
```diff
@@ -13,6 +13,12 @@ namespace GVFS.Common.FileSystem
     public static class HooksInstaller
     {
         private static readonly string ExecutingDirectory;
+        private static readonly HookData[] Hooks = new[]
```
wilbaker: Super minor: Maybe rename to `NativeHookData[] NativeHooks` (since the pre\post command hooks are not part of this collection)?

GVFS/GVFS.Common/NamedPipes/NamedPipeMessages.cs
```diff
@@ -137,6 +138,46 @@ public Message CreateMessage()
...
+                        throw new InvalidOperationException($"Invalid PostIndexChanged message. Expected at least 2 characters, got: {message.Body.Length} from message: '{message.Body}'");
+                    }
+
+                    this.WorkingDirectoryUpdated = message.Body[0] == '1';
+                    this.OnlyIndexUpdated = message.Body[1] == '1';
+                }
+
+                public bool WorkingDirectoryUpdated { get; }
+
+                public bool OnlyIndexUpdated { get; }
```
wilbaker: Related to the comment in `GVFS/GVFS.PostIndexChangedHook/main.cpp` it's not clear to me what `OnlyIndexUpdated` means.  Consider adding a comment here (and\or updating the name)

GVFS/GVFS.Mount/InProcessMount.cs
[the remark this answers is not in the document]
wilbaker: > git wouldn't care because it has already renamed the index.

If the git command running updates the index multiple times, should the command stop\fail in the middle if VFS4G fails to update its projection?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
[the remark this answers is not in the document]
wilbaker: > maybe here they should be renamed to something more appropriate for the context like invalidateProjection and invalidateModifiedPaths

I think renaming them like you suggested would make the code easier to follow

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -375,24 +379,24 @@ public virtual void OnIndexFileChange()
...
-                {
-                    this.GitIndexProjection.InvalidateModifiedFiles();
-                    this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnIndexWriteRequiringModifiedPathsValidation());
-                }
+        }
 
-                this.InvalidateGitStatusCache();
-            }
-            else
+        public void OnIndexFileChanged(bool workingDirectoryUpdated, bool onlyIndexUpdated)
```
wilbaker: Maybe rename this to `OnIndexFileChangedHook` to make it clearer that this method is driven by the git hook and not by file notifications.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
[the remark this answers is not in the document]
wilbaker: Maybe `WakeIndexParsingThreadAndWait`?  I like your suggestion of `ReleaseIndexParsingThreadWait` as well.

GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181107.1</GitPackageVersion>
+    <GitPackageVersion>2.20181106.5-pr</GitPackageVersion>
```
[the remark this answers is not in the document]
kewillford: I talked this over with Ben and want to change the indicator to be updatedSkipWorktreeBits since it is not just the `reset --mixed` that we care about it is when the index changes and the skip-worktree bits could have been changed.  I also need to update the description since the implementation has changed

GVFS/GVFS.Common/NamedPipes/NamedPipeMessages.cs
kewillford: I'm going to change this to SkipWorktreeBitsUpdated.  Hopefully that makes more sense.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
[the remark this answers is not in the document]
kewillford: I would really like the caller to be able to know which `OnIndexChanged` to call since there would be 2 if the name is kept the same.  Maybe they should be `OnIndexChangedByGit` and `OnIndexChangeDetected` or `OnIndexChangedPostCommand`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: I will test turning off the placeholder creation and see what the failures are.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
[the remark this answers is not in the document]
wilbaker: I like your idea of making the names `OnIndexChange[OriginOfChangeNotification]`.

GVFS/GVFS.Common/NamedPipes/NamedPipeMessages.cs
```diff
@@ -137,6 +138,46 @@ public Message CreateMessage()
...
+                        throw new InvalidOperationException($"Invalid PostIndexChanged message. Expected 2 characters, got: {message.Body.Length} from message: '{message.Body}'");
+                    }
+
+                    this.WorkingDirectoryUpdated = message.Body[0] == '1';
+                    this.SkipWorktreeBitsUpdated = message.Body[1] == '1';
+                }
+
+                public bool WorkingDirectoryUpdated { get; }
+
+                public bool SkipWorktreeBitsUpdated { get; }
```
sanoursa: This sounds like a request - "Skip the worktree bits update"

I would reword both of these: `UpdatedWorkingDirectory` and `UpdatedSkipWorktreeBits`

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -229,6 +229,18 @@ public virtual void Shutdown()
             this.indexParsingThread.Wait();
         }
 
+        public void ReleaseIndexParsingRequestWait()
```
sanoursa: This name is still very confusing. My problem with it is that it assumes the caller knows or cares about the internal details of how this is going to work - but the caller should just care about what they want done, not how. I would suggest something like `ParseIndex` or `UpdateProjection` or something to indicate the action it will be taking, rather than the mechanism.

GVFS/GVFS.Mount/InProcessMount.cs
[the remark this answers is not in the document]
kewillford: Thinking about this and I think it would be a big change to how things have been working. Right now we only inform the user at the end of the git command of the placeholders that VFS4G was unable to update/delete.  Halting the git code in the middle could cause unknown issues especially ones like rebase that are spawning other commands like `git am`.  The outer command would need to detect the failure and do the right thing.  Git without VFS4G in many cases is a best effort anyways like when trying to delete files.  Sometimes it will inform the user and retry if the user indicates they want to but others it will just continue on.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: I am going to leave stopping blocking placeholder creation during a git command for a future PR.  I was running into some test failures and I'm not sure if it is something that needs to change on the git side or VFSForGit side.

kewillford:
@wilbaker @sanoursa please take another look.  I believe I have addressed the initial feedback.  Thanks!

GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181107.1</GitPackageVersion>
+    <GitPackageVersion>2.20181114.1-pr</GitPackageVersion>
```
wilbaker: Update to non-PR build before merging.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -229,6 +229,18 @@ public virtual void Shutdown()
             this.indexParsingThread.Wait();
         }
 
+        public void UpdateProjection()
+        {
+            // Inform the index parsing thread that the index has changed,
+            // which means the projection needs to be updated to be correct before git proceeds
+            this.parseIndexRequested.Set();
```
wilbaker: I've read through this a few times now, and I'm not sure that we need `parseIndexRequested`.  Previously we used `externalLockReleaseRequested` to force the parsing thread to wait for the git command to finish before it runs, but it seems that now we always want it to run right away.

Did you have any scenarios in mind that where `parseIndexRequested` is useful?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -991,7 +998,7 @@ private void ParseIndexThreadMain()
 
                     while (this.context.Repository.GVFSLock.GetExternalHolder() != null)
```
wilbaker: Another comment related to `parseIndexRequested`.  Do we still need to check `GetExternalHolder`?  Or could this block:

```
while (this.context.Repository.GVFSLock.GetExternalHolder() != null)
{
   if (this.parseIndexRequested.Wait(ParseIndexRequestTimeoutMs))
    {
        break;
    }

    if (this.isStopping)
    {
        return;
    }
}
```

Be replaced with:

```
    if (this.isStopping)
    {
        return;
    }
```

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: I don't _think_ we need this anymore.  The code that runs in between waking up the thread and releasing it here is the status cache invalidation and newly created list clearing which I don't think would affect the projection getting updated. I will remove this and run the tests to make sure.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -572,6 +560,23 @@ public void LaunchPostFetchJob(List<string> packIndexes)
...
         }
 
+        private void InvalidateState(bool invalidateProjection, bool invalidateModifiedPaths)
+        {
+            if (invalidateProjection)
+            {
+                this.GitIndexProjection.InvalidateProjection();
+            }
+
+            if (invalidateModifiedPaths)
```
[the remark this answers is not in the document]
wilbaker: I guess there's actually a bigger question, can `invalidateModifiedPaths` ever be true during a git command that will update the projection?

I don't think that any of this is necessarily an issue, the questions are for me to better understand what's happening :smile:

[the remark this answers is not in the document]
kewillford: Right now that should never happen since we are passing the UpdatedSkipworktreeBits only when in the reset mixed and update-index paths for git, so something in git would need to change for the UpdatedWorkingDirectory flag to get passed as well because it would need to run through the unpack_trees code which is updating the working directory with index entries and both those commands should never have to do that.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: > The code that runs in between waking up the thread and releasing it here is the status cache invalidation and newly created list clearing which I don't think would affect the projection getting updated. I will remove this and run the tests to make sure.

I have not been able to come up with any scenarios where that would be an issue.  Because we're in a git command we would have already finished processing anything that uses the list before the command starts (I think).  

I was thinking if this causes an issue we could move the clearing of the list up (so that it happens before starting the index parse), but I'm not sure if there are any scenarios where that would be an issue.  I think it comes down to whether we need to guarantee that the list has been cleared when we release the index-changed hook.

GVFS/GVFS.Common/FileSystem/HooksInstaller.cs
```diff
@@ -48,20 +54,15 @@ public static bool InstallHooks(GVFSContext context, out string error)
...
-                if (!TryHooksInstallationAction(() => CopyHook(context, installedReadObjectHookPath, targetReadObjectHookPath), out error))
-                {
-                    error = "Failed to copy " + installedReadObjectHookPath + "\n" + error;
-                    return false;
-                }
-
-                string installedVirtualFileSystemHookPath = Path.Combine(ExecutingDirectory, GVFSPlatform.Instance.Constants.GVFSVirtualFileSystemHookExecutableName);
-                string targetVirtualFileSystemHookPath = Path.Combine(context.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Hooks.VirtualFileSystemPath + GVFSPlatform.Instance.Constants.ExecutableExtension);
-                if (!TryHooksInstallationAction(() => CopyHook(context, installedVirtualFileSystemHookPath, targetVirtualFileSystemHookPath), out error))
+                foreach (HookData hook in NativeHooks)
```
derrickstolee: why are the precommand and postcommand hooks separate?

kewillford: Pre and Post command hooks work a little differently than the others because they are **not** direct hooks that git calls.  They are using the GitHooksLoader to read a .hooks file that lists the programs to run for during pre and post command of which the pre and post commands hooks are listed.

See [TryInstallGitCommandHooks](https://github.com/Microsoft/VFSForGit/blob/3337d8643ac81f73e2a4a2e34edab986fdd3243f/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs#L222) and [CreateHookCommandConfig](https://github.com/Microsoft/VFSForGit/blob/3337d8643ac81f73e2a4a2e34edab986fdd3243f/GVFS/GVFS.Platform.Windows/WindowsGitHooksInstaller.cs#L18)

Issue #396, opened 2018-10-16
Mac: Create log file for kext logging


wilbaker:
Kext logging of warnings\error should be recorded to a log file that `gvfs diagnose` can collect.

See #395 for related user-mode issue

pmj:
I'm not sufficiently familiar with the `gvfs diagnose` end, but if you need anything done on the kext/native userlib end for this, let me know.

Do we want to include kext profiling support into this as well in some form? (Certain builds only of course.)

pmj:
My take-away from our out-of-band discussion earlier:

 * As a first deliverable for this, extend/replace `prjfs-log` to be/with a proper daemon that is installed by the installer and runs all the time by default.
 * Add log level filtering in the kext to minimise overhead.
 * By default, log only error-level errors.
 * Allow end-users to temporarily enable more verbose logging.

Things that were discussed elsewhere:

 * Start by logging to macOS's syslog database, and see how we go with that. (+ write some scripts for gathering the logs up from syslog)

jrbriggs:
@pmj That all sounds good to me. Is there any value to log warning-level errors as well by default? Where will `prjfs-log` log to by default?

/cc @wilbaker

pmj:
I've been reading up on the details of macOS's Unified Logging system a bit before I start implementing the log daemon based on it. It seems pretty reasonable to me in general; interestingly, it has an "info" log level that by default doesn't hit disk unless sometime later an error is also logged. So rather than logging everything all the time, it effectively retroactively enables verbose logging if an error occurs. Is that something we'd be interested in using?

The only downside I can see to the whole thing is that there's no programmatic API for querying the log database, only the `log` CLI tool. This tool at least is fairly powerful and also supports a metadata-preserving (i.e. searchable, filterable, etc.) export/archival file format that e.g. Console.app supports reading; so that's something that users could send back to us.

Documentation I've been using:

* [WWDC presentation on (then) new logging system](https://developer.apple.com/videos/play/wwdc2016/721/)
* [Reference documentation](https://developer.apple.com/documentation/os/logging?language=occ)

Given the extremely low barrier to entry (I expect we'll simply be calling [os_log_with_type()](https://developer.apple.com/documentation/os/os_log_with_type?language=objc) instead of printf()ing to stdout in our code.) I suggest the following:

 * Create a variant of prjfs-log that works as a daemon
 * This daemon writes to the unified logging system.
[...]

Issue #395, opened 2018-10-16
Mac: Add logging support to the ProjFS user-mode library


wilbaker:
There should be some way for the ProjFS user-mode library to write to a log file (either directly or via VFS4G).

wilbaker:
Error logging has been added to ProjFS, but we should add info/warning logging as well

PR #465, opened 2018-11-02
Mac: Support 'gvfs diagnose' verb

### LINKED ISSUE #459: Mac: Support 'gvfs diagnose' verb
Support `gvfs diagnose` verb.

Creating and collecting ProjFS logs will be handled in #395  and #396.

wilbaker:
Resolves #459 

Minimum changes to get `gvfs diagnose` working on the Mac.  

Collecting logs from ProjFS will be covered in follow-up issues (#395 and #396).

### DIFF
--- a/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
+++ b/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
@@ -16,8 +16,6 @@ namespace GVFS.CommandLine
     public class DiagnoseVerb : GVFSVerb.ForExistingEnlistment
     {
         private const string DiagnoseVerbName = "diagnose";
-        private const string WindowsExecutableExtension = ".exe";
-        private const string MacNamedPipeSocketName = "GVFS_NetCorePipe";
 
         private TextWriter diagnosticLogFileWriter;
 
@@ -398,9 +396,9 @@ namespace GVFS.CommandLine
                 string fileName = Path.GetFileName(filePath);
                 try
                 {
-                    string fileExtension = Path.GetExtension(fileName);
-                    if (!string.Equals(fileExtension, WindowsExecutableExtension, StringComparison.OrdinalIgnoreCase) &&
-                        !string.Equals(fileName, MacNamedPipeSocketName, StringComparison.OrdinalIgnoreCase))
+                    string sourceFilePath = Path.Combine(sourcePath, fileName);
+                    if (!GVFSPlatform.Instance.FileSystem.IsSocket(sourceFilePath) &&
+                        !GVFSPlatform.Instance.FileSystem.IsExecutable(sourceFilePath))
                     {
                         File.Copy(
                             Path.Combine(sourcePath, fileName),
--- a/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
+++ b/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
@@ -90,8 +90,8 @@ namespace GVFS.CommandLine
 
                         if (!GVFSPlatform.Instance.IsUnderConstruction)
                         {
-                            // filter
-                            this.FlushFilterLogBuffers();
+                            // driver
+                            this.FlushKernelDriverLogs();
                             string kernelLogsFolderPath = GVFSPlatform.Instance.KernelDriver.LogsFolderPath;
 
                             // This copy sometimes fails because the OS has an exclusive lock on the etl files. The error is not actionable
@@ -473,7 +473,7 @@ namespace GVFS.CommandLine
             }
         }
 
-        private void FlushFilterLogBuffers()
+        private void FlushKernelDriverLogs()
         {
             string errors = GVFSPlatform.Instance.KernelDriver.FlushLogs();
             this.diagnosticLogFileWriter.WriteLine(errors);
--- a/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
+++ b/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
@@ -4,7 +4,6 @@ using GVFS.Common.FileSystem;
 using GVFS.Common.Git;
 using GVFS.Common.Http;
 using GVFS.Common.Tracing;
-using Microsoft.Win32;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -17,7 +16,8 @@ namespace GVFS.CommandLine
     public class DiagnoseVerb : GVFSVerb.ForExistingEnlistment
     {
         private const string DiagnoseVerbName = "diagnose";
-        private const string System32LogFilesRoot = @"%SystemRoot%\System32\LogFiles";
+        private const string WindowsExecutableExtension = ".exe";
+        private const string MacNamedPipeSocketName = "GVFS_NetCorePipe";
 
         private TextWriter diagnosticLogFileWriter;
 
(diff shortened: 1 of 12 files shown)


GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -390,7 +396,8 @@ private void RecursiveFileCopyImpl(string sourcePath, string targetPath, bool co
                 try
                 {
                     string fileExtension = Path.GetExtension(fileName);
-                    if (!string.Equals(fileExtension, ".exe", StringComparison.OrdinalIgnoreCase))
+                    if (!string.Equals(fileExtension, WindowsExecutableExtension, StringComparison.OrdinalIgnoreCase) &&
+                        !string.Equals(fileName, MacNamedPipeSocketName, StringComparison.OrdinalIgnoreCase))
```
kewillford: Maybe the platform should have a method like `bool ShouldCopyForDiagnose(string filename)`?

sanoursa: Or `IsExecutable`

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -115,27 +118,30 @@ protected override void Execute(GVFSEnlistment enlistment)
...
-                            ProductUpgrader.GetUpgradesDirectoryPath(), 
-                            Path.Combine(archiveFolderPath, ProductUpgrader.UpgradeDirectoryName),
-                            ProductUpgrader.DownloadDirectory, 
-                            "downloaded-assets.txt");
+                        this.CopyAllFiles(enlistment.DotGVFSRoot, Path.Combine(archiveFolderPath, GVFSConstants.DotGVFS.Root), GVFSConstants.DotGVFS.CorruptObjectsName, copySubFolders: false);
+
+                        if (!GVFSPlatform.Instance.IsUnderConstruction)
+                        {
+                            // service
+                            this.CopyAllFiles(
```
sanoursa: This could be wrapped in `GVFSPlatform.Instance.SupportsGVFSService` instead of `IsUnderConstruction`

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -115,27 +118,30 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                                Paths.GetServiceDataRoot(string.Empty),
+                                archiveFolderPath,
+                                this.ServiceName,
+                                copySubFolders: true);
+
+                            // upgrader
+                            this.CopyAllFiles(
```
sanoursa: Maybe add a `GVFSPlatform.SupportsGVFSUpgrade`? And actually the `UpgradeVerb` code would benefit from that too, since it's still incorrectly checking the OS instead of a platform capability. cc @alameenshah in case he already has a fix in progress for that.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -88,15 +88,18 @@ protected override void Execute(GVFSEnlistment enlistment)
...
-                        // filter
-                        this.FlushFilterLogBuffers();
-                        string system32LogFilesPath = Environment.ExpandEnvironmentVariables(System32LogFilesRoot);
-
-                        // This copy sometimes fails because the OS has an exclusive lock on the etl files. The error is not actionable
-                        // for the user so we don't write the error message to stdout, just to our own log file.
-                        this.CopyAllFiles(system32LogFilesPath, archiveFolderPath, GVFSPlatform.Instance.KernelDriver.DriverLogFolderName, copySubFolders: false, hideErrorsFromStdout: true);
+                        this.CopyAllFiles(enlistment.EnlistmentRoot, archiveFolderPath, GVFSConstants.DotGVFS.Root, copySubFolders: false);
+
+                        if (!GVFSPlatform.Instance.IsUnderConstruction)
```
sanoursa: I'm not sure we need an under construction flag here - we could just check if the log folder exists.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
wilbaker: > Maybe the platform should have a method like bool ShouldCopyForDiagnose(string filename)

I think this would be cleaner, but I'm not sure if it violates the spirit of the platforms.  @sanoursa, any preferences\objections?

I actually had started down the road of adding `IsExecutable` and `IsSocket` to the platform (which seemed like a better fit), but it ended up being quite complicated on Mac (and didn't seem like something we absolutely needed for `gvfs diagnose`).  The trouble I ran into was that the output `struct` of `stat(...)` can vary based on `ifdefs`, and it wasn't clear how we could determine which format would be returned when PInvoking to libc.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
wilbaker: > I'm not sure we need an under construction flag here - we could just check if the log folder exists.

We don't know the log path to return on Mac (yet), but I could switch from using a `LogsFolderPath` property to a `TryGetLogsFolderPath` method and eliminate the under construction flag (I'm not sure that's actually buying us much though).  

Thoughts?

sanoursa: Or just return null for now? And this will get cleaned up in the very next PR right? So you could also just leave it as is.

wilbaker: > Or just return null for now? And this will get cleaned up in the very next PR right? So you could also just leave it as is.

I'll go with whatever is cleanest.  I was hoping to avoid code that looks permanent (e.g. checking if the path is `null`) that was added *only* to avoid calling `IsUnderConstruction` (I'd rather be more explicit about why we have a check and use `IsUnderConstruction`).

As you mentioned, this will get cleaned up as soon as we have kext logging (which should be soon) and so I don't think it's a big deal to leave in the `IsUnderConstruction` (assuming that ends up being the cleanest option).

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
sanoursa: Yep as we discussed offline, the `GVFSPlatform` is intended to represent behaviors of the platform, not GVFS-specific business logic. `IsExecutable` and `IsSocket` are a better fit than `ShouldCopyForDiagnose`.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
wilbaker: I'll make this change

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
wilbaker: I'll go ahead and add `IsExecutable` and `IsSocket`

wilbaker: > but it ended up being quite complicated on Mac 

I figured out the issue I had here as well, I will bring back my earlier changes for PInvoking to `stat`

wilbaker:
@kewillford @sanoursa my latest push address all of the outstanding feedback in this PR, please let me know what you think.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -467,7 +475,7 @@ private ReturnCode RunAndRecordGVFSVerb<TVerb>(string archiveFolderPath, string
 
         private void FlushFilterLogBuffers()
```
kewillford: Rename to `FlushKernelDriverLogs` or just move this code inline since it is only called from one place?

PR #466, opened 2018-11-02
Create a script to create a build drop to enable running functional tests on boxes that didn't build the product.

jrbriggs:
This will drop the minimal layout to successfully run the functional tests for a build that can be transferred to a box that did not build the product.

### DIFF
--- a/Scripts/Mac/CI/CreateBuildDrop.sh
+++ b/Scripts/Mac/CI/CreateBuildDrop.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+. "$(dirname ${BASH_SOURCE[0]})/../InitializeEnvironment.sh"
+
+CONFIGURATION=$1
+BUILDDROP_ROOT=$2
+if [ -z $BUILDDROP_ROOT ] || [ -z $CONFIGURATION ]; then
+  echo 'ERROR: Usage: CreateBuildDrop.sh [configuration] [build drop root directory]'
+  exit 1
+fi
+
+# Set up some paths
+BUILDDROP_BUILDOUTPUT=$BUILDDROP_ROOT/BuildOutput
+BUILDDROP_SRC=$BUILDDROP_ROOT/src
+BUILDDROP_PROJFS=$BUILDDROP_SRC/ProjFS.Mac
+BUILDDROP_KEXT=$BUILDDROP_BUILDOUTPUT/ProjFS.Mac/Native/Build/Products/$CONFIGURATION
+
+# Set up the build drop directory structure
+rm -rf $BUILDDROP_ROOT
+mkdir -p $BUILDDROP_BUILDOUTPUT
+mkdir -p $BUILDDROP_SRC
+mkdir -p $BUILDDROP_PROJFS
+mkdir -p $BUILDDROP_KEXT
+
+# Copy to the build drop, retaining directory structure.
+rsync -avm $VFS_OUTPUTDIR/Git $BUILDDROP_BUILDOUTPUT
+rsync -avm $VFS_PUBLISHDIR $BUILDDROP_ROOT
+rsync -avm $VFS_SCRIPTDIR $BUILDDROP_SRC/Scripts
+rsync -avm $VFS_SRCDIR/ProjFS.Mac/Scripts $BUILDDROP_PROJFS
+rsync -avm $VFS_OUTPUTDIR/ProjFS.Mac/Native/Build/Products/$CONFIGURATION/PrjFSKext.kext $BUILDDROP_KEXT


Scripts/Mac/LayoutFunctionalTests.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
+
+CONFIGURATION=$1
+LAYOUT_ROOT=$2
```
sanoursa: I think the meaning of "layout" will become fuzzy very quickly. This script is just creating a build drop that can be published from one build and then consumed from another, right?

How about the name `CreateBuildDrop.sh` for the script name, and `BuildDrop` anywhere you use `Layout` inside the script?

I also didn't find an equivalent script for Windows, how is it being created there?

Scripts/Mac/LayoutFunctionalTests.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
sanoursa: We had discussed reorganizing the Windows scripts into dev-oriented and CI-oriented scripts. This one very clearly feels like a script intended for use on the CI/PR builds, and never on a dev box, so we could we stick it in a different location and not mix it in with the dev scripts?

jrbriggs: Makes sense, I'll move it.

Scripts/Mac/LayoutFunctionalTests.sh
jrbriggs: `BuildDrop` makes sense. I'll retool this.

I'm going to create a similar script for Windows. I kept it out of this PR in the interest of focus, so I don't have to hop between two boxes. It will the new PR I post. :)

Scripts/Mac/CI/CreateBuildDrop.sh
```diff
@@ -0,0 +1,30 @@
...
+BUILDDROP_ROOT=$2
+if [ -z $BUILDDROP_ROOT ] || [ -z $CONFIGURATION ]; then
+  echo 'ERROR: Usage: CreateBuildDrop.sh [configuration] [build drop root directory]'
+  exit 1
+fi
+
+# Set up some paths
+BUILDDROP_BUILDOUTPUT=$BUILDDROP_ROOT/BuildOutput
+BUILDDROP_SRC=$BUILDDROP_ROOT/src
+BUILDDROP_PROJFS=$BUILDDROP_SRC/ProfFS.Mac
```
kewillford: ProjFS.Mac?

Scripts/Mac/CI/CreateBuildDrop.sh
```diff
@@ -0,0 +1,30 @@
...
+BUILDDROP_KEXT=$BUILDDROP_BUILDOUTPUT/ProjFS.Mac/Native/Build/Products/$CONFIGURATION
+
+# Set up the build drop directory structure
+rm -rf $BUILDDROP_ROOT
+mkdir -p $BUILDDROP_BUILDOUTPUT
+mkdir -p $BUILDDROP_SRC
+mkdir -p $BUILDDROP_PROJFS
+mkdir -p $BUILDDROP_KEXT
```
kewillford: There are cases in the build scripts that use `sudo`.  How do we know when we will need to use that vs not?

Scripts/Mac/CI/CreateBuildDrop.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Copy to the build drop, retaining directory structure.
+rsync -avm $VFS_OUTPUTDIR/Git $BUILDDROP_BUILDOUTPUT
+rsync -avm $VFS_PUBLISHDIR $BUILDDROP_ROOT
+rsync -avm $VFS_SCRIPTDIR $BUILDDROP_SRC/Scripts
+rsync -avm $VFS_SRCDIR/ProjFS.Mac/Scripts $BUILDDROP_PROJFS
```
kewillford: Should this keep the `Scripts` folder?

Scripts/Mac/CI/CreateBuildDrop.sh
jrbriggs: (facepalm)

Scripts/Mac/CI/CreateBuildDrop.sh
jrbriggs: It keeps the script folders so that we can keep the logic on how we run functional tests or install/start the kext encoded in the script.

Scripts/Mac/CI/CreateBuildDrop.sh
jrbriggs: I'm making the assumption in this script that the CI agent will have access to wherever it is asked to publish the drop. Also, this caught my clumsiness a few times where I fat fingered /[something] instead of ~/[something].

PR #469, opened 2018-11-05
WriteFile: Fix Logging

### LINKED ISSUE #443: Failures of WriteFile due to FileClosed are incorrectly being logged as an error
We should be breaking out of the switch statement here:

https://github.com/Microsoft/VFSForGit/blob/8dba3b4aae0076ba71289f5c59793db0e3518838/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs#L1041

But instead VFS4G is logging:

`WriteFile failed, error: 80070006(FileClosed)`

jeschu1:
Don't throw on FileClose since it is an expected condition 

resolves #443

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
@@ -1038,8 +1038,8 @@ namespace GVFS.Platform.Windows
                                 {
                                     switch (writeResult)
                                     {
-                                        case (HResult)HResultExtensions.HResultFromNtStatus.FileClosed:
-                                            // StatusFileClosed is expected, and occurs when an application closes a file handle before OnGetFileStream
+                                        case HResult.FileClosed:
+                                            // HResult.FileClosed is expected, and occurs when an application closes a file handle before OnGetFileStream
                                             // is complete
                                             break;
 
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
@@ -822,7 +822,6 @@ namespace GVFS.UnitTests.Windows.Virtualization
                     mockGVFSGitObjects.FileLength = fileLength;
 
                     MockTracer mockTracker = this.Repo.Context.Tracer as MockTracer;
-                    mockTracker.WaitRelatedEventName = "GetFileStreamHandlerAsyncHandler_OperationCancelled";
 
                     mockVirtualization.WriteFileReturnResult = HResult.InternalError;
                     mockVirtualization.OnGetFileStream(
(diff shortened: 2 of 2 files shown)


GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1034,15 +1034,14 @@ private HResult QueryFileNameHandler(string virtualPath)
                                 HResult writeResult = this.virtualizationInstance.WriteFile(streamGuid, targetBuffer, (ulong)writeOffset, bytesToCopy);
                                 remainingData -= bytesToCopy;
 
-                                if (writeResult != HResult.Ok)
+                                if (writeResult != HResult.Ok &&
+
+                                    // StatusFileClosed is expected, and occurs when an application closes a file handle before OnGetFileStream
```
wilbaker: Minor, can you update this comment to say `FileClosed` instead of `StatusFileClosed` (the old text is referring to an enum that we no longer have in our code).

jeschu1: Done!

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1034,15 +1034,14 @@ private HResult QueryFileNameHandler(string virtualPath)
                                 HResult writeResult = this.virtualizationInstance.WriteFile(streamGuid, targetBuffer, (ulong)writeOffset, bytesToCopy);
                                 remainingData -= bytesToCopy;
 
-                                if (writeResult != HResult.Ok)
+                                if (writeResult != HResult.Ok &&
+
+                                    // FileClosed is expected, and occurs when an application closes a file handle before OnGetFileStream
+                                    // is complete
+                                    writeResult != (HResult)HResultExtensions.HResultFromNtStatus.FileClosed)
```
kewillford: @wilbaker what will happen when the `TryCompleteCommand` runs after we received this HResult?  Looks like it will get called with `HResult.Ok`?  Is that right?

wilbaker: @kewillford that's a good point, we should pass back the return value of `WriteFile` to ProjFS when it's not `Ok`.

@jeschu1, I don't think we want to stop throwing a `GetFileStreamException` here.  Based on the issue's description I suspect the bug is in our switch statement below.

The switch statement currently has:

`case (HResult)HResultExtensions.HResultFromNtStatus.FileClosed:`

Which is supposed to prevent us from logging errors when `WriteFile` returns `FileClosed`.  

I took another look at the message reported in the bug (`WriteFile failed, error: 80070006(FileClosed)`) and it looks like the issue is that our switch statement is not handling `HResult.FileClosed`.

In other words, if we update the switch statement to:

```
case (HResult)HResultExtensions.HResultFromNtStatus.FileClosed:
case HResult.FileClosed:
   // FileClosed is expected, and occurs when an application closes a file handle before OnGetFileStream
   // is complete
   break;
```

And I'm not sure that we actually need to handle `(HResult)HResultExtensions.HResultFromNtStatus.FileClosed` at all anymore.  

@CAllredMSFT, do you know if `WriteFile` will ever return `HRESULT_FROM_NT(STATUS_FILE_CLOSED)`?

CAllredMSFT: @wilbaker `0x80070006` is `HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE)`, which is what `STATUS_FILE_CLOSED` becomes.

I've had a look at some code to confirm, and it is possible for `PrjWriteFileData()` to return `HRESULT_FROM_WIN32(ERROR_INVALID_HANDLE)`.  Like the comment says, if the app closes the file handle before one of your `WriteFile` calls completes, then this can happen.  I'd expect it to be uncommon though, since I think it would require unusual app behavior.

> we should pass back the return value of `WriteFile` to ProjFS when it's not `Ok`

That's correct.

CAllredMSFT: Aside from STATUS_CANNOT_DELETE for [denying delete](https://docs.microsoft.com/en-us/windows/desktop/projfs/file-system-operation-notifications#receiving-notifications) in a pre-delete notification (for now anyway), I don't think you need any NTSTATUS codes.  I suspect they're all left over from a long time ago when our API used to use a mix of NTSTATUS and HRESULT codes.

wilbaker: @CAllredMSFT the one I wasn't sure about was `IoReparseTagNotHandled`.  

What will ProjFS return if we call `UpdatePlaceholderIfNeeded` on a file that has a non-ProjFS reparse point?

CAllredMSFT: None of our APIs return NTSTATUS codes or even `HRESULT_FROM_NT(status)` codes.  All of the NTSTATUS codes that the filter returns through the API go through the Win32-standard `RtlNtStatusToDosError()` transformation before being converted to an HRESULT.

So if that API did trip over STATUS_IO_REPARSE_TAG_NOT_HANDLED it would turn into `HRESULT_FROM_WIN32(ERROR_CANT_ACCESS_FILE)`.

wilbaker: @CAllredMSFT thanks again for the details!

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1038,7 +1038,7 @@ private HResult QueryFileNameHandler(string virtualPath)
                                 {
                                     switch (writeResult)
                                     {
-                                        case (HResult)HResultExtensions.HResultFromNtStatus.FileClosed:
+                                        case HResult.FileClosed:
                                             // StatusFileClosed is expected, and occurs when an application closes a file handle before OnGetFileStream
```
wilbaker: Minor: Can you update this comment (`StatusFileClosed` -> `HResult.FileClosed`)

PR #472, opened 2018-11-05
Add functional tests for LsTree parsing issue

jamill:
Functional tests for issue uncovered with Fast Fetch.

The original issue was fixed with #441. This adds functional test coverage and allows discussion specifically on the functional tests.


GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -19,6 +19,8 @@ namespace GVFS.FunctionalTests.Tests
     [Category(Categories.MacTODO.M4)]
     public class FastFetchTests
     {
+        private const string LsTreeTypeInPathBranchName = "FunctionalTests/10181105_LsTreeTypeInPath";
```
nickgra: Is **1018** intentional? I see you've pushed a branch that starts with 2018 too. The convention is to start with the year, so 1018 seems wrong.
Apologies if this was discussed elsewhere where I wasn't looking.

jamill: You don't think this branch could have been created 1,000 years ago? This has been updated to reflect the current year - thanks for catching this!

nickgra:
Is this ready to be merged in? :)

Issue #234, opened 2018-08-30
Atomic placeholder operations in VFSForGit on MacOS


nickgra:
A hole in our current design is in our creation of
placeholders. Today, when creating a placeholder, at a high
level, we go through the following workflow:

-   Create placeholder directory/file in the virtualization root

-   'Mark up' the file with the proper flags/xattrs

-   (If creating a file) Update the mode of the file

This could result in scenarios where an incomplete placeholder appears in
the virtualization root. The full set of failure modes is documented
below:

WritePlaceholderDirectory:

-   mkdir() fails to create the placeholder directory

    -   The enumeration fails as a directory wasn't created. Today, we are       retriable here as a second attempt should see the IsEmpty bit on
         the parent directory and call WritePlaceholderDirectory again. However, the retry will fail as we will be unable to overwrite the existing children (if any exist regardless of if they're files or directories) in this directory.

-   We crash between mkdir() and InitializeEmptyPlaceholder

    -   Placeholder directory does not have flags for virtualization root and
        empty.

    -   This directory will never be expanded as the kext will exit when
         we do not see the InVirtualizationRoot bit

-   We crash between setting the bits for 'in virtualization root' and 'is empty'

    -   This directory will never be expanded as the kext will exit
            when we do not see the IsEmpty bit. This effectively puts a user into a state where it looks like they're missing files in their repo.

WritePlaceholderFile:

-   fopen() fails to create a file

    -   This is retriable today as a second attempt should see the
         IsEmpty bit on the parent directory and call WritePlaceholderFile
         again. 

-   ftruncate() fails to expand the file to the correct size (or is never
     reached because of a crash)

    -   We have an incomplete placeholder file in the virtualization root
         that cannot be recovered. Future attempts of
         WritePlaceholderFile will fail because fopen(path, "wbx")
         won't let us overwrite the existing file.

-   Memcpy'ing the providerID xattr fails (or is never reached because
     of a crash)

    -   We have an incomplete placeholder file in the virtualization root
        that cannot be recovered. There are no flags set on this file.

    -   The MirrorProvider doesn't currently use the providerID for
         anything (although it should fail if it's empty)

    -   VFSForGit will be unable to parse the placeholder version from
         the providerID, OnGetFileStream will fail so hydrating this
         file will fail. Retrying will not help.

-   Memcpy'ing the contentID xattr fails (or is never reached because of
     a crash)

    -   We have an incomplete placeholder file in the virtualization root
[...]

PR #473, opened 2018-11-06
Atomic Placeholder Creation for MacOS

nickgra:
This PR (previous discussion at #225) implements the design described at #234 to increase the reliability of VFSForGit by making the process in which we create placeholder files/directories 'atomic' (we create them outside of the virtualization root in a known temporary directory and move them into place once they're 'complete').

- [x] Functionality in the provider to create placeholders in the temp directory
- [x] Provider can now register a temp directory with the kext
- [x] Kext will no longer consider changes in a registered temp directory as one that should be considered
- [x] Rename .temp and put it under the provider folder
- [x] Use random names when creating placeholders in the temp directory
- [ ] Clear temp folder on provider mount
- [ ] Investigate if flushing is needed after setting file flags (and perf impact)
- [ ] Validate functionality on the large VFSForGit repo
- [ ] Performance analysis on the large VFSForGit repo
- [ ] Call UpdatePlaceholderIfNeeded when in a retry scenario where we could potentially be overwriting user data
- [ ] Use our kext to prevent the user from doing anything to our .staging folder

### DIFF
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -95,6 +95,7 @@ namespace GVFS.Common
         public static class DotGVFS
         {
             public const string Root = ".gvfs";
+            public const string TempRoot = ".staging";
             public const string CorruptObjectsName = "CorruptObjects";
 
             public static readonly string LogPath = Path.Combine(DotGVFS.Root, "logs");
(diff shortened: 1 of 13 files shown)


nickgra:
The kext has moved on so far from when I started working on this that I'm going to retire this PR as it's only useful as a reference for when we pick this idea back up. 

(I think the rebase would be messy)

Issue #471, opened 2018-11-05
TryGetFromGitConfig should not treat warning strings as errors


wilbaker:
There are a few places in the code that read from git config files like this:

```
GitProcess.Result result = git.GetFromConfig(configName);
if (result.HasErrors)
{
    if (result.Errors.Any())
    {
        error = "Error while reading '" + configName + "' from config: " + result.Errors;
        return false;
    }

    // Git returns non-zero for non-existent settings and errors.
    return true;
}
```

This check can cause VFS4G failures when trying to read non-existent config entries.

*Example*

> Failed to determine GVFS timeout and max retries: Error while reading 'gvfs.max-retries' from config: warning: could not open '\\.\pipe\git' for tracing: No such file or directory

**TBD**

- Will git always prepend warnings with "warning"?
- We need to ensure that if there are multiple lines written to `stderr`(some warnings and some errors, that VFS4G *does* fail)

jrbriggs:
Closed with #478

PR #478, opened 2018-11-06
Ignore non-zero exit codes when calling `git config` and all lines on stderr are prefixed 'warning'

jrbriggs:
This addresses #471

### DIFF
--- a/GVFS/FastFetch/CheckoutPrefetcher.cs
+++ b/GVFS/FastFetch/CheckoutPrefetcher.cs
@@ -111,7 +111,7 @@ namespace FastFetch
                         string remoteBranch = refs.GetBranchRefPairs().Single().Key;
                         GitProcess git = new GitProcess(this.Enlistment);
                         GitProcess.Result result = git.SetUpstream(branchOrCommit, remoteBranch);
-                        if (result.HasErrors)
+                        if (result.ExitCodeIsFailure)
                         {
                             activity.RelatedError("Could not set upstream for {0} to {1}: {2}", branchOrCommit, remoteBranch, result.Errors);
                             this.HasFailures = true;
@@ -204,7 +204,7 @@ namespace FastFetch
             uint valueCoreGvfs;
 
             // No errors getting the configuration and it is either "true" or numeric with the right bit set.
-            return !configCoreGvfs.HasErrors &&
+            return !configCoreGvfs.ExitCodeIsFailure &&
                 !string.IsNullOrEmpty(configCoreGvfs.Output) &&
                 (configCoreGvfs.Output.Equals("true", StringComparison.OrdinalIgnoreCase) ||
                 (uint.TryParse(configCoreGvfs.Output, out valueCoreGvfs) &&
--- a/GVFS/FastFetch/CheckoutPrefetcher.cs
+++ b/GVFS/FastFetch/CheckoutPrefetcher.cs
@@ -204,7 +204,7 @@ namespace FastFetch
             uint valueCoreGvfs;
 
             // No errors getting the configuration and it is either "true" or numeric with the right bit set.
-            return !configCoreGvfs.ExitCodeIsFailure &&
+            return configCoreGvfs.ExitCodeIsSuccess &&
                 !string.IsNullOrEmpty(configCoreGvfs.Output) &&
                 (configCoreGvfs.Output.Equals("true", StringComparison.OrdinalIgnoreCase) ||
                 (uint.TryParse(configCoreGvfs.Output, out valueCoreGvfs) &&
--- a/GVFS/FastFetch/CheckoutPrefetcher.cs
+++ b/GVFS/FastFetch/CheckoutPrefetcher.cs
@@ -200,14 +200,20 @@ namespace FastFetch
             const uint CoreGvfsUnsignedIndexFlag = 1;
 
             GitProcess git = new GitProcess(this.Enlistment);
-            GitProcess.Result configCoreGvfs = git.GetFromConfig("core.gvfs");
+            GitProcess.ConfigResult configCoreGvfs = git.GetFromConfig("core.gvfs");
+            string coreGvfs;
+            string error;
+            if (!configCoreGvfs.TryParseAsString(out coreGvfs, out error))
+            {
+                return false;
+            }
+
             uint valueCoreGvfs;
 
             // No errors getting the configuration and it is either "true" or numeric with the right bit set.
-            return configCoreGvfs.ExitCodeIsSuccess &&
-                !string.IsNullOrEmpty(configCoreGvfs.Output) &&
-                (configCoreGvfs.Output.Equals("true", StringComparison.OrdinalIgnoreCase) ||
-                (uint.TryParse(configCoreGvfs.Output, out valueCoreGvfs) &&
+            return !string.IsNullOrEmpty(coreGvfs) &&
+                (coreGvfs.Equals("true", StringComparison.OrdinalIgnoreCase) ||
+                (uint.TryParse(coreGvfs, out valueCoreGvfs) &&
                 ((valueCoreGvfs & CoreGvfsUnsignedIndexFlag) == CoreGvfsUnsignedIndexFlag)));
         }
     }
--- a/GVFS/GVFS.Common/Enlistment.cs
+++ b/GVFS/GVFS.Common/Enlistment.cs
@@ -35,7 +35,7 @@ namespace GVFS.Common
             else
             {
                 GitProcess.Result originResult = gitProcess.GetOriginUrl();
-                if (originResult.HasErrors)
+                if (originResult.ExitCodeIsFailure)
                 {
                     if (originResult.Errors.Length == 0)
                     {
--- a/GVFS/GVFS.Common/Enlistment.cs
+++ b/GVFS/GVFS.Common/Enlistment.cs
@@ -34,18 +34,18 @@ namespace GVFS.Common
             }
             else
             {
-                GitProcess.Result originResult = gitProcess.GetOriginUrl();
-                if (originResult.ExitCodeIsFailure)
+                GitProcess.ConfigResult originResult = gitProcess.GetOriginUrl();
+                if (!originResult.TryParseAsString(out string originUrl, out string error))
                 {
-                    if (originResult.Errors.Length == 0)
-                    {
-                        throw new InvalidRepoException("Could not get origin url. remote 'origin' is not configured for this repo.'");
-                    }
-
-                    throw new InvalidRepoException("Could not get origin url. git error: " + originResult.Errors);
+                    throw new InvalidRepoException("Could not get origin url. git error: " + error);
                 }
 
-                this.RepoUrl = originResult.Output.Trim();
+                if (originUrl == null)
+                {
+                    throw new InvalidRepoException("Could not get origin url. remote 'origin' is not configured for this repo.'");
+                }
+
+                this.RepoUrl = originUrl.Trim();
             }
             
             this.Authentication = authentication ?? new GitAuthentication(gitProcess, this.RepoUrl);
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -719,6 +719,18 @@ namespace GVFS.Common.Git
             {
                 get { return !this.ExitCodeIsSuccess; }
             }
+
+            public bool StderrContainsErrors()
+            {
+                if (!string.IsNullOrWhiteSpace(this.Errors))
+                {
+                    return !this.Errors
+                        .Split(new char[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
+                        .All(line => line.TrimStart().StartsWith("warning:", StringComparison.OrdinalIgnoreCase));
+                }
+
+                return false;
+            }
         }
     }
 }
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -731,8 +731,8 @@ namespace GVFS.Common.Git
 
         public class ConfigResult
         {
-            private Result result;
-            private string configName;
+            private readonly Result result;
+            private readonly string configName;
 
             public ConfigResult(Result result, string configName)
             {
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -276,7 +276,7 @@ namespace GVFS.Common.Git
         /// otherwise it will run it from outside the enlistment.
         /// </param>
         /// <returns>The value found for the setting.</returns>
-        public virtual Result GetFromConfig(string settingName, bool forceOutsideEnlistment = false, PhysicalFileSystem fileSystem = null)
+        public virtual ConfigResult GetFromConfig(string settingName, bool forceOutsideEnlistment = false, PhysicalFileSystem fileSystem = null)
         {
             string command = string.Format("config {0}", settingName);
             fileSystem = fileSystem ?? new PhysicalFileSystem();
@@ -284,13 +284,13 @@ namespace GVFS.Common.Git
             // This method is called at clone time, so the physical repo may not exist yet.
             return
                 fileSystem.DirectoryExists(this.workingDirectoryRoot) && !forceOutsideEnlistment
-                    ? this.InvokeGitAgainstDotGitFolder(command)
-                    : this.InvokeGitOutsideEnlistment(command);
+                    ? new ConfigResult(this.InvokeGitAgainstDotGitFolder(command), settingName)
+                    : new ConfigResult(this.InvokeGitOutsideEnlistment(command), settingName);
         }
 
-        public Result GetFromLocalConfig(string settingName)
+        public ConfigResult GetFromLocalConfig(string settingName)
         {
-            return this.InvokeGitAgainstDotGitFolder("config --local " + settingName);
+            return new ConfigResult(this.InvokeGitAgainstDotGitFolder("config --local " + settingName), settingName);
         }
 
         /// <summary>
@@ -308,12 +308,8 @@ namespace GVFS.Common.Git
             value = null;
             try
             {
-                Result result = this.GetFromConfig(settingName, forceOutsideEnlistment, fileSystem);
-                if (result.ExitCodeIsSuccess)
-                {
-                    value = result.Output;
-                    return true;
-                }
+                ConfigResult result = this.GetFromConfig(settingName, forceOutsideEnlistment, fileSystem);
+                return result.TryParseAsString(out value, out string _);
             }
             catch
             {
@@ -322,9 +318,9 @@ namespace GVFS.Common.Git
             return false;
         }
 
-        public Result GetOriginUrl()
+        public ConfigResult GetOriginUrl()
         {
-            return this.InvokeGitAgainstDotGitFolder("config --local remote.origin.url");
+            return new ConfigResult(this.InvokeGitAgainstDotGitFolder("config --local remote.origin.url"), "remote.origin.url");
         }
 
         public Result DiffTree(string sourceTreeish, string targetTreeish, Action<string> onResult)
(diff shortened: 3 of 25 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -714,6 +714,24 @@ public bool HasErrors
             {
                 get { return this.ReturnCode != SuccessCode; }
```
kewillford: What do you think about changing this logic to be `==` and the name of this property to be `WasSuccessful` since how it is now is not really and indication that there are errors, it is checking the return code for success?  Then your new logic could be called `HasErrors` and would be more accurate to me.  `HasErrors` sounds like it should be some indication of what is in the `Errors` property which it doesn't now.  I realize this is a lot more invasive of a change but would make more sense at least to me.

GVFS/GVFS.Common/RetryConfig.cs
```diff
@@ -134,8 +134,9 @@ private static bool TryGetFromGitConfig(GitProcess git, string configName, int d
             value = defaultValue;
             error = string.Empty;
 
+            System.Diagnostics.Debugger.Launch();
```
kewillford: Don't think we want to keep this in production code. :)

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
```diff
@@ -14,5 +14,80 @@ public void TryKillRunningProcess_NeverRan()
             GitProcess process = new GitProcess(new MockGVFSEnlistment());
             process.TryKillRunningProcess().ShouldBeTrue();
         }
+
+        [TestCase]
+        public void ResultHasNonWarningErrors_SingleLine_AllWarnings()
```
kewillford: I realize we didn't have any before but should we have at least a couple of success tests?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -714,6 +714,24 @@ public bool HasErrors
...
+            public bool HasNonWarningErrors
+            {
+                get
+                {
+                    if (this.HasErrors && this.Errors.Any())
+                    {
+                        return !this.Errors
+                            .Trim()
+                            .Split('\n')
+                            .All(line => line.TrimStart().StartsWith("warning:"));
```
kewillford: Does the `StartsWith` need to be case insensitive?

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
```diff
@@ -14,5 +14,80 @@ public void TryKillRunningProcess_NeverRan()
...
+            result.HasNonWarningErrors.ShouldBeFalse();
+        }
+
+        [TestCase]
+        public void ResultHasNonWarningErrors_Multiline_AllWarnings()
+        {
+            GitProcess.Result result = new GitProcess.Result(
+                string.Empty,
+                @"warning: this line should not be considered an error
+warning: neither should this.",
```
kewillford: Do we need tests with spaces before "warning:"  or casing differences? I saw a couple of messages in the git source that use "Warning:" and "WARNING:"

GVFS/GVFS.Common/RetryConfig.cs
jrbriggs: I'm sure it'll be fine :)

Thanks for catching that...

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
jrbriggs: Likely so, I'll make that change.

GVFS/GVFS.Common/Git/GitProcess.cs
jrbriggs: Yeah, good point.

GVFS/GVFS.Common/Git/GitProcess.cs
jrbriggs: I considered such a direction, but the scope was larger than I wanted to pick off for this particular bug. It's something we should consider for the backlog, but I'm wary of churning so much of the codebase when the bugfix can be much more targeted. Thoughts @kewillford @sanoursa @wilbaker ?

GVFS/GVFS.Common/GitStatusCacheConfig.cs
```diff
@@ -86,7 +86,7 @@ private static bool TryGetFromGitConfig(GitProcess git, string configName, int d
             error = string.Empty;
 
             GitProcess.Result result = git.GetFromConfig(configName);
-            if (result.HasErrors)
+            if (result.HasNonWarningErrors)
             {
                 if (result.Errors.Any())
```
wilbaker: (This comment applies to change in `RetryConfig.cs` as well)

Rather than the current change, should this be updated to:

```
if (result.HasErrors)
{
    if (result.HasNonWarningErrors)
    {
        error = "Error while reading '" + configName + "' from config: " + result.Errors;
```

The current changes confused me a bit because based on this comment:

```
// Git returns non-zero for non-existent settings and errors.
```

This `if` block was intended to handle all non-zero return values from git.  With the new changes, this code block is only handling "fatal errors" and other errors (e.g. non-existing setting) are skipping the block and continuing on to the code that's meant to parse the output when git was able to successfully read the config value.

GVFS/GVFS.Common/Git/GitProcess.cs
wilbaker: I agree with Kevin that the current naming can make it a bit tricky to follow the current code (see comment in `GitStatusCacheConfig.cs` for an example), but I also agree that we want to limit the scope of this fix.

To throw an option 3 out there, what if for this PR we renamed `HasNonWarningErrors` to something like `HasErrorMessages` and have it simply check `Errors` for lines that don't start with "warning:".


We could then replace calls to `result.Errors.Any()` with `result.HasErrorMessages`:

*Example*

```
if (result.HasErrors)
{
    if (result.HasErrorMessages)
    {
        error = "Error while reading '" + configName + "' from config: " + result.Errors;
```
> should consider for the backlog

I think the cleanup is definitely worth it, whether we do it now or add it to the backlog.

jrbriggs:
> Independent of these code changes - should we fix git to avoid printing those warnings? Are they actionable for users or are they just going to break any tools that try to parse git output?

Yes @sanoursa. These are the same warnings that would be printed if a user was directing `GIT_TRACE` to a bad location. We should be more resilient to this output.

GVFS/GVFS.Common/GitStatusCacheConfig.cs
jrbriggs: > this code block is only handling "fatal errors" and other errors (e.g. non-existent setting) are skipping the block and continuing on to the code that's meant to parse the output when git was able to successfully read the config value

Yes and no. The very next statement after this if block tests for empty output, meaning use the default. Semantically I don't think anything changed but I agree it could be cleaner. I'm looking at other options.

wilbaker: > Yes and no. The very next statement after this if block tests for empty output, meaning use the default. Semantically I don't think anything changed but I agree it could be cleaner. I'm looking at other options.

Yep that's a good point.  Maybe just remove (or move?) the `// Git returns non-zero for non-existent settings and errors.` comment?  I'm not sure that it's in the best place now.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -714,6 +714,24 @@ public bool HasErrors
             {
                 get { return this.ReturnCode != SuccessCode; }
             }
+
+            public bool HasNonWarningErrors
```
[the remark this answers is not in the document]
jrbriggs: I'm going to rename `HasErrors` to be more specifically about testing the exit code. I think that's part of the confusion here.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -714,6 +714,24 @@ public bool HasErrors
             {
                 get { return this.ReturnCode != SuccessCode; }
             }
+
+            public bool HasNonWarningErrors
+            {
+                get
+                {
+                    if (this.HasErrors && this.Errors.Any())
```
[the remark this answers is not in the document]
jrbriggs: Yep. I'll make this change.

GVFS/GVFS.Common/Git/GitProcess.cs
jrbriggs: This is simplified in the new patch series

GVFS/FastFetch/CheckoutPrefetcher.cs
```diff
@@ -204,7 +204,7 @@ private bool GetIsIndexSigningOff()
             uint valueCoreGvfs;
 
             // No errors getting the configuration and it is either "true" or numeric with the right bit set.
-            return !configCoreGvfs.HasErrors &&
+            return !configCoreGvfs.HasNonWarningErrors &&
```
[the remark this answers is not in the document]
jrbriggs: Good catch, fixed.

GVFS/GVFS.Common/GitStatusCacheConfig.cs
jrbriggs: I simplified this codepath.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -605,7 +605,7 @@ private void PostFetchJob(List<string> packIndexes)
 
                         GitProcess.Result result = this.postFetchGitProcess.WriteMultiPackIndex(this.context.Enlistment.GitObjectsRoot);
 
-                        if (!this.stopping && result.HasErrors)
+                        if (!this.stopping && result.ExitCodeIsFailure)
```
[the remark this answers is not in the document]
jrbriggs: @jeschu1 this change will not address that, as my refactoring here was just a mechanical rename to make it more clear what, exactly, we're testing.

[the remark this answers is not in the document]
jeschu1: @jrbriggs ah you are correct here.  I've filed a separate bug for the empty errors [here](https://github.com/Microsoft/VFSForGit/issues/491).

GVFS/GVFS.Common/GitStatusCacheConfig.cs
```diff
@@ -86,16 +86,10 @@ private static bool TryGetFromGitConfig(GitProcess git, string configName, int d
             error = string.Empty;
 
             GitProcess.Result result = git.GetFromConfig(configName);
```
kewillford: Do you know if there are any unit tests or functional tests covering this code path?  We should be able to use the `MockGitProcess` if not.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -699,20 +699,38 @@ public class Result
...
+                get { return !this.ExitCodeIsSuccess; }
+            }
+
+            public bool StderrContainsErrors()
+            {
+                if (!string.IsNullOrWhiteSpace(this.Errors))
+                {
+                    return !this.Errors
+                         .Trim()
+                         .Split('\n')
```
sanoursa: Should we specify `StringSplitOptions.RemoveEmptyEntries` here? Otherwise a blank line between warning messages would cause us to return false here, right?

GVFS/GVFS.Common/Enlistment.cs
```diff
@@ -35,7 +35,7 @@ public abstract class Enlistment
             else
             {
                 GitProcess.Result originResult = gitProcess.GetOriginUrl();
-                if (originResult.HasErrors)
+                if (originResult.ExitCodeIsFailure)
                 {
                     if (originResult.Errors.Length == 0)
```
wilbaker: `if (!originResult.StderrContainsErrors())` instead of `if (originResult.Errors.Length == 0)`?

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
```diff
@@ -14,5 +14,104 @@ public void TryKillRunningProcess_NeverRan()
...
+            result.StderrContainsErrors().ShouldBeFalse();
+        }
+
+        [TestCase]
+        public void ResultHasNonWarningErrors_Multiline_AllWarnings()
+        {
+            GitProcess.Result result = new GitProcess.Result(
+                string.Empty,
+                @"warning: this line should not be considered an error
+WARNING: neither should this.",
```
wilbaker: Related to Saeed's comment, add a test with blank lines between warning lines?

GVFS/GVFS.Common/Http/CacheServerResolver.cs
```diff
@@ -136,7 +136,7 @@ private static string GetValueFromConfig(GitProcess git, string configName, bool
                 ? git.GetFromLocalConfig(configName)
                 : git.GetFromConfig(configName);
 
-            if (!result.HasErrors)
+            if (result.ExitCodeIsSuccess)
             {
                 return result.Output.TrimEnd('\n');
             }
```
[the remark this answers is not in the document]
jrbriggs: Yep, good catch

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
jrbriggs: Yep, thanks

GVFS/GVFS/RepairJobs/GitConfigRepairJob.cs
```diff
@@ -23,7 +23,7 @@ public override IssueType HasIssue(List<string> messages)
         {
             GitProcess git = new GitProcess(this.Enlistment);
             GitProcess.Result result = git.GetOriginUrl();
-            if (result.HasErrors)
+            if (result.ExitCodeIsFailure)
             {
                 if (result.Errors.Length == 0)
```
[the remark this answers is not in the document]
wilbaker: I am marking this as approved so that I don't prevent the changes from going in when they're ready, but I will still take a look at the new commit when it's ready.

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
```diff
@@ -103,12 +103,12 @@ public override void Dispose()
         private static string GetConfigValue(string gitBinRoot, string configKey)
         {
             GitProcess.Result result = GitProcess.GetFromSystemConfig(gitBinRoot, configKey);
-            if (result.HasErrors || string.IsNullOrEmpty(result.Output.TrimEnd('\r', '\n')))
+            if (result.ExitCodeIsFailure || string.IsNullOrEmpty(result.Output.TrimEnd('\r', '\n')))
```
jrbriggs: Looks like I missed some config interactions. Taking another pass through.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -699,20 +695,99 @@ public class Result
...
+                    error = "Error while reading '" + this.configName + "' from config: " + this.result.Errors;
+                    return false;
+                }
 
-            public bool HasErrors
+                if (this.result.ExitCodeIsSuccess)
+                {
+                    value = this.result.Output;
+                }
+
```
wilbaker: Should `TryParseAsString` have a `defaultValue` parameter like `TryParseAsInt`?

GVFS/GVFS.Mount/InProcessMountVerb.cs
```diff
@@ -129,16 +129,16 @@ private JsonTracer CreateTracer(GVFSEnlistment enlistment, EventLevel verbosity,
             string mountId = null;
 
             GitProcess git = new GitProcess(enlistment);
-            GitProcess.Result configResult = git.GetFromLocalConfig(GVFSConstants.GitConfig.EnlistmentId);
-            if (!configResult.HasErrors)
+            GitProcess.ConfigResult configResult = git.GetFromLocalConfig(GVFSConstants.GitConfig.EnlistmentId);
+            if (configResult.TryParseAsString(out enlistmentId, out string _))
             {
-                enlistmentId = configResult.Output.Trim();
+                enlistmentId = enlistmentId.Trim();
```
wilbaker: Check if `enlistmentId` is `null` (or allow `TryParseAsString` to take a default value and pass in `string.Empty`)

GVFS/GVFS.Platform.Windows/WindowsGitHooksInstaller.cs
```diff
@@ -43,7 +43,7 @@ private static string MergeHooks(GVFSContext context, string configSettingName,
             string filename;
             string[] defaultHooksLines = { };
 
-            if (configProcess.TryGetFromConfig(configSettingName, forceOutsideEnlistment: true, value: out filename))
+            if (configProcess.TryGetFromConfig(configSettingName, forceOutsideEnlistment: true, value: out filename) && filename != null)
             {
                 filename = filename.Trim(' ', '\n');
                 defaultHooksLines = File.ReadAllLines(filename);
```
wilbaker: I wonder if we just had a bug here before if the config value was not set.  It looks like `File.ReadAllLines` would have  been called with an empty string.

Just mentioned this to confirm that this change in behavior is expected.

GVFS/GVFS.Mount/InProcessMountVerb.cs
```diff
@@ -129,16 +129,16 @@ private JsonTracer CreateTracer(GVFSEnlistment enlistment, EventLevel verbosity,
...
-                enlistmentId = configResult.Output.Trim();
+                enlistmentId = enlistmentId.Trim();
             }
 
             configResult = git.GetFromLocalConfig(GVFSConstants.GitConfig.MountId);
-            if (!configResult.HasErrors)
+            if (configResult.TryParseAsString(out mountId, out string _))
             {
-                mountId = configResult.Output.Trim();
+                mountId = mountId.Trim();
```
wilbaker: Same thing here, `mountId` could be `null`.

GVFS/GVFS.Common/Http/CacheServerResolver.cs
```diff
@@ -126,26 +126,22 @@ public bool TrySaveUrlToLocalConfig(CacheServerInfo cache, out string error)
...
-                return result.Output.TrimEnd('\n');
-            }
-            else if (result.Errors.Any())
-            {
-                throw new InvalidRepoException("Error while reading '" + configName + "' from config: " + result.Errors);
+                throw new InvalidRepoException(error);
             }
 
-            return null;
+            return value;
```
wilbaker: We used to call `.TrimEnd('\n')` before returning the result, and it looks like now it's not happening (as best I can tell `TryParseAsString` does not do this).

GVFS/GVFS.Common/GitStatusCacheConfig.cs
```diff
@@ -82,42 +82,8 @@ private static bool TryLoadBackOffTime(GitProcess git, out int backoffTimeSecond
...
-
-            if (value < minValue)
-            {
-                error = string.Format("Invalid value {0} for setting {1}, value must be greater than or equal to {2}", value, configName, minValue);
-                return false;
-            }
-
-            return true;
+            GitProcess.ConfigResult result = git.GetFromConfig(configName);
+            return result.TryParseAsInt(defaultValue, minValue, out value, out error);
```
wilbaker: Previously we'd call `TrimEnd('\n')` on the output here before trying to parse it as an int.  Does `TryParseAsInt` need to do the same?

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
```diff
@@ -14,5 +14,207 @@ public void TryKillRunningProcess_NeverRan()
...
+
+            result.TryParseAsInt(1, -1, out int value, out string error).ShouldBeTrue();
+            value.ShouldEqual(1);
+        }
+
+        [TestCase]
+        public void ConfigResult_TryParseAsInt_ParsesWhenNoError()
+        {
+            GitProcess.ConfigResult result = new GitProcess.ConfigResult(
+                new GitProcess.Result("32", string.Empty, 0),
```
wilbaker: Related to comments elsewhere, consider adding `ConfigResult` tests where the output has a trailing newline.

GVFS/GVFS.Common/GitStatusCacheConfig.cs
jrbriggs: I validated my assumption that int.TryParse() would do that trimming for us. I think it's cleaner to leave that to the framework.

GVFS/GVFS.Mount/InProcessMountVerb.cs
jrbriggs: Fixed by allowing a default parameter, defaulting to string.Empty.

GVFS/GVFS.Mount/InProcessMountVerb.cs
jrbriggs: Now defaults to string.Empty

GVFS/GVFS.Common/Git/GitProcess.cs
jrbriggs: Yes, good thought. Making that change helped clarify other bits that didn't feel so great. :)

GVFS/GVFS.Platform.Windows/WindowsGitHooksInstaller.cs
jrbriggs: I don't think the behavior is changed here. Previously, TryGetFromConfig would fail unless exitcode was success, meaning that a config key was present and the value was read. 

The condition changed to guard against null now that TryGetFromConfig can return null. We should figure out if that's the desired API behavior. This is the only caller of TryGetFromConfig, fwiw.

wilbaker: Ahh you're right, I must have gotten the `TryGetFromConfig` implementation mixed up with one of the other helpers.

GVFS/GVFS.Platform.Windows/ETWTelemetryEventListener.cs
jrbriggs: Cleaned this up with the latest push.

Issue #488, opened 2018-11-08
Maintain shared object cache in the background


derrickstolee:
The shared cache is growing to be large. This includes:

* Object Cache:
  * Number of loose objects.
  * Number of pack-files.
  * Size of pack-files.
* Object Size Database

This document proposes a background maintenance task to help reduce the
size of this cache and improve Git performance as a side-effect.

Object Cache Maintenance
------------------------

There are some things we can do to clean up the object cache, with varying
levels of difficulty and benefit. Here are the four items I believe will
have a strong impact on our data size and performance:

I. Delete loose objects that exist in pack-files.
II. Place loose objects into pack-files.
III. Repack some prefetch packs into larger packs.
IV. Delete packs with no objects referenced by the multi-pack-index.

These are numbered by order of implementation, but they work together in an
interesting way. If we place loose objects into pack-files (I), then the next
time we delete loose objects (II) we will delete those objects that we just put
into a pack-file. If we repack some prefetch packs into larger packs (III), the
multi-pack-index will use the "newer" pack for all objects in those packs, and
we can delete them (IV).

In order to perform this maintenance without disrupting users, we must be
careful in what order we perform these steps. The order must be as given
below:

A. Loose Object Cleanup Step
  1. Delete loose objects that exist in pack-files.
  2. Place loose objects into pack-files.
B. Prefetch Pack Cleanup Step
  1. Delete packs with no objects referenced by the multi-pack-index.
  2. Repack some prefetch packs into larger packs.

Care must be taken to ensure that Git processes that initialized their object
store before the maintenance still succeed. Thus, we should never delete
anything that those processes could be reading from. This is why we delete loose
objects only if they are referenced by pack-files. This is why we delete
prefetch packs only if they have no objects referenced by the multi-pack-index.
Since a second run of these jobs could delete objects that were repacked in the
previous run, we need to space our job runs such that no Git processes span both
runs.

The loose object maintenance can happen at a different cadence than the prefetch
pack maintenance, or they can happen at the same time.

Loose Object Cleanup Step
---------------------------

The loose object step can save users who have many loose object downloads.
These loose objects are populated by several user activities: running
`git fetch` will get loose commits and trees for the ref tips that are not
covered by prefetch packs, hydrating a file will trigger a loose blob, and many
other. Older enlistments have many loose commits and trees due to the behavior
of `git fetch` before the `--no-show-forced-updates` option.

### Delete loose objects that appear in a pack-file

[...]

Issue #481, opened 2018-11-07
BackgroundPrefetch brings down service if "Device is not ready"


jeschu1:
**I'm seeing several issues that follow the same pattern.**

1. Mount has been up and running for  while without very recent activity
2. BackgroundPrefetch is unable to run and sends a warning with  "The device is not ready.\r\n",
3. Telemetry for the BackgroundPrefetch attempts to write and throws an error with  "System.IO.IOException: The device is not ready.\r\n\r\n"

**Some examples are:**
timestamp	                            MountId
2018-11-06 21:14:48.8584682    b747d97273b9494b8e39c10f469bf732
2018-11-06 19:01:34.6583537	   8234a2ea76774fae81bf594b96f824e7
2018-11-06 18:54:41.1197685	   c2e88b5eb82648ed94a4dc2354204ff1
2018-11-06 18:26:07.3041028	   9b50c5145c184f55aee77e9c5405cdd9
2018-11-06 17:40:46.6437567	   040e03593168474b8d3a3cf0b4a38000
2018-11-06 17:40:24.7076743   dbfc3836d2d54f09a2a29134a7fec929

**Questions.**
1. Should a failed BackgroundPrefetch take down the mount?
2. If we decide it does take the mount should this be a warning not an error?  Since this wasn't user initiated it could be that the drive has been disconnected, locked etc...  
3. As a general rule should we count "Device is not Ready" against us for mount failures?
@sanoursa @wilbaker

PR #492, opened 2018-11-08
[RFC] Maintenance Job Design Document

derrickstolee:
This PR is for discussion on the RFC originally created as #488.

### DIFF
--- a/MaintenanceJobDesignDoc.md
+++ b/MaintenanceJobDesignDoc.md
@@ -1,0 +1,140 @@
+Maintenance Job Design Document
+===============================
+
+The (shared) object cache is growing to be large. This includes:
+
+* Number of loose objects.
+* Number of pack-files.
+* Size of pack-files.
+
+There are some things we can do to improve this situation, with varying levels
+of difficulty and benefit. Here are the four items I believe will have a strong
+impact on our data size and performance:
+
+I. Delete loose objects that exist in pack-files.
+II. Place loose objects into pack-files.
+III. Repack some prefetch packs into larger packs.
+IV. Delete packs with no objects referenced by the multi-pack-index.
+
+These are numbered by order of implementation, but they work together in an
+interesting way. If we place loose objects into pack-files (I), then the next
+time we delete loose objects (II) we will delete those objects that we just put
+into a pack-file. If we repack some prefetch packs into larger packs (III), the
+multi-pack-index will use the "newer" pack for all objects in those packs, and
+we can delete them (IV).
+
+In order to perform this maintenance without disrupting users, we can actually
+do these interactions in two independent jobs, with their sub-actions in a
+particular order:
+
+A. Loose Object Maintenance
+  1. Delete loose objects that exist in pack-files.
+  2. Place loose objects into pack-files.
+B. Prefetch Pack Maintenance
+  1. Delete packs with no objects referenced by the multi-pack-index.
+  2. Repack some prefetch packs into larger packs.
+
+Care must be taken to ensure that Git processes that initialized their object
+store before the maintenance   still succeed. Thus, we should never delete
+anything that those processes could be reading from. This is why we delete loose
+objects only if they are referenced by pack-files. This is why we delete
+prefetch packs only if they have no objects referenced by the multi-pack-index.
+Since a second run of these jobs could delete objects that were repacked in the
+previous run, we need to space our job runs such that no Git processes span both
+runs.
+
+Loose Object Job
+--------------
+
+The loose object job can save users who have many loose object downloads. These
+loose objects are populated by several user activities: running `git fetch` will
+get loose commits and trees for the ref tips that are not covered by prefetch
+packs, hydrating a file will trigger a loose blob, and many other. Older
+enlistments have many loose commits and trees due to the behavior of `git fetch`
+before the `--no-show-forced-updates` option.
+
+### Delete loose objects that appear in a pack-file
+
+By running `git prune-packed -q` with the environment variable
+`GIT_OBJECT_DIRECTORY` set to the object cache, Git will delete the loose
+objects that exist in a pack-file. This will use the multi-pack-index to test
+containment, so is only bound by the file-deletion operations.
+
+### Pack loose objects that do not appear in a pack-file
+
+We can use `git pack-objects -q <object-dir>/pack/loose-<timestamp>` to create a
+pack-file filled with a list of loose objects. We can pass the list of loose
+objects to this command through stdin. We can choose to batch this operation by
+limiting the number of objects we pack at a time, if we want. We may also want
+to skip this step if the number of loose objects is low (&lt; 1,000? &lt;
+10,000?)
+
+By listing a timestamp in the pack-file name, we can store the "most-recent" 
+time this job was run.
+
+*Note:* While we are thinking about packing loose objects, we can consider
+_deleting_ some loose blobs. It is easier to delete loose blobs than packed
+blobs. If we want to delete "stale" blobs, then we should add a step to this
+job.
+
+Prefetch Pack Job
+---------------
+
+We store a list of prefetch packs as `<object-dir>/pack/prefetch-<timestamp>-<sha>.[pack|idx]`.
+These are downloaded at a regular cadence by the post-fetch job, and could have
+a new pack-file for every hour. In addition to these prefetch packs, some daily
+prefetch packs will appear and occasionally a very large prefetch pack covering
+everything more than 30 days old. This large prefetch pack appears on initial
+clone and sometimes again if a user does not prefetch for 30 days (this occurs 
+less often now that background prefetch exists).
+
+### Delete pack-files with no objects referenced by the multi-pack-index
+
+The multi-pack-index stores a single copy of every packed object. When an object
+appears in multiple pack-files covered by the multi-pack-index, we select the
+pack-file with the most-recent _modified time_ as reported by the filesystem.
+For example, if someone downloads a second large prefetch pack when prefetching
+for the first time in 31 days, then that large prefetch pack is referenced and
+all older prefetch packs are "stale" and not referenced by the multi-pack-index.
+
+We can extend the `git multi-pack-index` builtin to delete pack-files that have
+no objects referenced by the multi-pack-index. This should happen at the same
+time as we update the multi-pack-index since we don't want to have this list of
+pack-files in the multi-pack-index if they don't exist. We also don't want to
+remove them from the multi-pack-index while the pack-files still exist, because
+then a Git process will start referencing those pack-files in the `packed_git`
+linked list.
+
+### Repack the prefetch pack-files
+
+We can reduce the number of prefetch packs and improve file locality by
+combining prefetch packs into new "repacked" pack-files. Use `git show-index
+<object-dir>/pack/prefetch-<timestamp[i]>-<hash[i]>.idx` to list the objects
+in the prefetch packs and pipe that output to `git pack-objects -q
+<object-dir>/pack/repacked-<timestamp>`.
+
+We should refrain from packing the most-recent prefetch packs, as we don't want
+to get in the way of the prefetch machinery for determining the most-recent
+timestamp. We could, however, start repacking the prefetch packs from
+oldest-to-newest, stopping in the most-recent week, or when the prefetch packs
+are large enough to satisfy a batch. Some stopping conditions could include
+"pack-file size is above 1GB".
+
+Running the Jobs
+----------------
+
+We already have a post-fetch job that is automatically queued in the background.
+We should rename this job to be a "maintenance" job that could trigger several
+actions. The maintenance job should run every 15 minutes (as the post-fetch job
+currently does) but then it should check each "sub-job" for a timing condition
+to run again.
+
+For example:
+
+* The post-fetch job doesn't trigger if the most-recent prefetch pack timestamp
+  is within the last 70 minutes.
+* The loose object job doesn't trigger if the most-recent "loose" pack timestamp
+  is within the last 24 hours.
+* The prefetch repack job doesn't trigger if the most-recent "repacked" pack
+  timestamp is within the last 7 days or the most-recent "repacked" pack is
+  modified within the last 24 hours.
--- a/MaintenanceJobDesignDoc.md
+++ b/MaintenanceJobDesignDoc.md
@@ -1,15 +1,25 @@
 Maintenance Job Design Document
 ===============================
 
-The (shared) object cache is growing to be large. This includes:
+The shared cache is growing to be large. This includes:
 
-* Number of loose objects.
-* Number of pack-files.
-* Size of pack-files.
+* Object Cache:
+  * Number of loose objects.
+  * Number of pack-files.
+  * Size of pack-files.
+* Object Size Database
 
-There are some things we can do to improve this situation, with varying levels
-of difficulty and benefit. Here are the four items I believe will have a strong
-impact on our data size and performance:
+This document proposes a background maintenance task to help reduce the
+size of this cache and improve Git performance as a side-effect. This
+proposal focuses on concrete actions on the object cache, but the concepts
+of scheduling background tasks can extend to the object size database.
+
+Object Cache Maintenance
+------------------------
+
+There are some things we can do to clean up the object cache, with varying
+levels of difficulty and benefit. Here are the four items I believe will
+have a strong impact on our data size and performance:
 
 I. Delete loose objects that exist in pack-files.
 II. Place loose objects into pack-files.
@@ -23,19 +33,19 @@ into a pack-file. If we repack some prefetch packs into larger packs (III), the
 multi-pack-index will use the "newer" pack for all objects in those packs, and
 we can delete them (IV).
 
-In order to perform this maintenance without disrupting users, we can actually
-do these interactions in two independent jobs, with their sub-actions in a
-particular order:
+In order to perform this maintenance without disrupting users, we must be
+careful in what order we perform these actions. The order must be as given
+below:
 
-A. Loose Object Maintenance
+A. Loose Object Cleanup Action
   1. Delete loose objects that exist in pack-files.
   2. Place loose objects into pack-files.
-B. Prefetch Pack Maintenance
+B. Prefetch Pack Cleanup Action
   1. Delete packs with no objects referenced by the multi-pack-index.
   2. Repack some prefetch packs into larger packs.
 
 Care must be taken to ensure that Git processes that initialized their object
-store before the maintenance   still succeed. Thus, we should never delete
+store before the maintenance still succeed. Thus, we should never delete
 anything that those processes could be reading from. This is why we delete loose
 objects only if they are referenced by pack-files. This is why we delete
 prefetch packs only if they have no objects referenced by the multi-pack-index.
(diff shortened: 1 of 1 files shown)


MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+multi-pack-index will use the "newer" pack for all objects in those packs, and
+we can delete them (IV).
+
+In order to perform this maintenance without disrupting users, we can actually
+do these interactions in two independent jobs, with their sub-actions in a
+particular order:
+
+A. Loose Object Maintenance
+  1. Delete loose objects that exist in pack-files.
+  2. Place loose objects into pack-files.
```
kewillford: Maybe this has changed and could be different with the read-object hook but I remember in the git code that it would
1. check pack files
2. check loose object
3. check pack files again

In the comments it said something about objects that may have just been repacked.  Will this keep the loose objects around until we run again and step 1 will delete the loose objects that were moved into pack-files?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+  2. Repack some prefetch packs into larger packs.
+
+Care must be taken to ensure that Git processes that initialized their object
+store before the maintenance   still succeed. Thus, we should never delete
+anything that those processes could be reading from. This is why we delete loose
+objects only if they are referenced by pack-files. This is why we delete
+prefetch packs only if they have no objects referenced by the multi-pack-index.
+Since a second run of these jobs could delete objects that were repacked in the
+previous run, we need to space our job runs such that no Git processes span both
+runs.
```
kewillford: Maybe we should have some tracking of the git processes that are running before we start and mark things for deletion and have something that only checks what can be deleted and makes sure the git processes are gone so we can know for sure that something can be deleted?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+packs, hydrating a file will trigger a loose blob, and many other. Older
+enlistments have many loose commits and trees due to the behavior of `git fetch`
+before the `--no-show-forced-updates` option.
+
+### Delete loose objects that appear in a pack-file
+
+By running `git prune-packed -q` with the environment variable
+`GIT_OBJECT_DIRECTORY` set to the object cache, Git will delete the loose
+objects that exist in a pack-file. This will use the multi-pack-index to test
+containment, so is only bound by the file-deletion operations.
```
kewillford: If the file is locked and the delete fails does git just skip it an move on, trying again the next time?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Pack loose objects that do not appear in a pack-file
```
kewillford: Will this only be for loose object in the cache or loose object in the repo as well?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+then a Git process will start referencing those pack-files in the `packed_git`
+linked list.
+
+### Repack the prefetch pack-files
+
+We can reduce the number of prefetch packs and improve file locality by
+combining prefetch packs into new "repacked" pack-files. Use `git show-index
+<object-dir>/pack/prefetch-<timestamp[i]>-<hash[i]>.idx` to list the objects
+in the prefetch packs and pipe that output to `git pack-objects -q
+<object-dir>/pack/repacked-<timestamp>`.
```
kewillford: Any idea how long this would run or how much memory it would use?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+
+For example:
+
+* The post-fetch job doesn't trigger if the most-recent prefetch pack timestamp
+  is within the last 70 minutes.
+* The loose object job doesn't trigger if the most-recent "loose" pack timestamp
+  is within the last 24 hours.
+* The prefetch repack job doesn't trigger if the most-recent "repacked" pack
+  timestamp is within the last 7 days or the most-recent "repacked" pack is
+  modified within the last 24 hours.
```
kewillford: How did you come up with these numbers for running?  Is there data to back up this being the most efficient or is it based on the prefetch packs?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+timestamp. We could, however, start repacking the prefetch packs from
+oldest-to-newest, stopping in the most-recent week, or when the prefetch packs
+are large enough to satisfy a batch. Some stopping conditions could include
+"pack-file size is above 1GB".
+
+Running the Jobs
+----------------
+
+We already have a post-fetch job that is automatically queued in the background.
+We should rename this job to be a "maintenance" job that could trigger several
```
sanoursa: "Maintenance job" sounds too generic. E.g. we also have plans to do maintenance of placeholder files. How about "Cache Maintenance" instead of just "Maintenance"?

MaintenanceJobDesignDoc.md
derrickstolee: The comments are regarding missing objects. So Git will follow this mechanism:

1. Check pack-files.
2. check loose object.
3. If first time at this step, check read-object hook. If success, goto 1.
4. If first time at this step, reprepare_packed_git and goto 1. 
5. Fail to find object.

The reprepare_packed_git is intended for Git processes that loaded the packed_git list before a repack finished, but tried to access an object that was previously loose after the repack finishes.

You are correct: we intend to leave loose objects around until the _next_ run of the loose object job. That next run should be far out enough that we are sure no Git processes live across that boundary.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
+Maintenance Job Design Document
+===============================
+
+The (shared) object cache is growing to be large. This includes:
+
+* Number of loose objects.
+* Number of pack-files.
+* Size of pack-files.
```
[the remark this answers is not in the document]
derrickstolee: But a similar mechanism could be added that cleans up that data store. I'll keep in mind that this maintenance doesn't need to be strictly about the git objects.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+IV. Delete packs with no objects referenced by the multi-pack-index.
+
+These are numbered by order of implementation, but they work together in an
+interesting way. If we place loose objects into pack-files (I), then the next
+time we delete loose objects (II) we will delete those objects that we just put
+into a pack-file. If we repack some prefetch packs into larger packs (III), the
+multi-pack-index will use the "newer" pack for all objects in those packs, and
+we can delete them (IV).
+
+In order to perform this maintenance without disrupting users, we can actually
```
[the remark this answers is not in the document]
derrickstolee: I split the loose objects and prefetch pack maintenance into separate jobs just because the timing involved seems like it should be different. Running A.1, A.2, B.1, B.2 (using numbers from below) in order will not disrupt users. However, if we did A.2 _then_ A.1, we may disrupt users, because they may try to access a loose object as we delete it.

MaintenanceJobDesignDoc.md
derrickstolee: Such a thing could be possible. I'll look into it. Seems like overkill for a job that will run at most once a day, but we can see how disruptive it would be to add a check after the job runs.

MaintenanceJobDesignDoc.md
derrickstolee: It uses `unlink_or_warn(path)`, so I believe it is robust against someone having a handle on the file.

MaintenanceJobDesignDoc.md
derrickstolee: Currently, I'm focused on the object cache, as that is where the majority of objects are. We could expand this to include objects in the .git folder of each enlistment (objects created by the user).

sanoursa: I would make it an explicit design goal that we only clean up objects in our own cache. I think we should never mess with user-created objects.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+no objects referenced by the multi-pack-index. This should happen at the same
+time as we update the multi-pack-index since we don't want to have this list of
+pack-files in the multi-pack-index if they don't exist. We also don't want to
+remove them from the multi-pack-index while the pack-files still exist, because
+then a Git process will start referencing those pack-files in the `packed_git`
+linked list.
+
+### Repack the prefetch pack-files
+
+We can reduce the number of prefetch packs and improve file locality by
```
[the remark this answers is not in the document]
derrickstolee: This is mostly a file-locality concern (fewer pack-file handles the better), but also an issue with duplicate objects (due to clock skew between cache servers, daily packs overriding hourly packs).

The multi-pack-index solves the "where is my object" question, while this will help object parsing.

MaintenanceJobDesignDoc.md
derrickstolee: This is something to investigate during implementation, but I think if we plan to batch the repack (only repack prefetch packs that are smaller than 1GB until the batch is larger than 1GB) then the computation time would be under a minute. Definitely something to try.

I do want to make an extra-careful statement that we are _not_ repacking the whole pack directory into a single pack. We are just taking the many small packs into slightly fewer, slightly bigger packs.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+combining prefetch packs into new "repacked" pack-files. Use `git show-index
+<object-dir>/pack/prefetch-<timestamp[i]>-<hash[i]>.idx` to list the objects
+in the prefetch packs and pipe that output to `git pack-objects -q
+<object-dir>/pack/repacked-<timestamp>`.
+
+We should refrain from packing the most-recent prefetch packs, as we don't want
```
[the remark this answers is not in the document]
derrickstolee: True. That would suffice for the needs of the most-recent timestamp.

We should also ensure that we have enough of a buffer so that the next time we run a repack, we are more likely to fill our batch size.

MaintenanceJobDesignDoc.md
derrickstolee: The loose timing comes from "24 hours should be enough for no Git processes to survive, and not too obtrusive to users". This could be flexible, depending on how often we think these loose objects can be filled.

The prefetch repack timing is based on looking at the shared object cache and seeing the size of the hourly and daily prefetch packs for the Windows repo.

* Looks like recent activity has been around 80MB a day. This amounts to less than 500 MB a week, so maybe it isn't enough to fill a 1GB batch. This does raise a good point: we should leave a timestamp file for "most-recent check for a batch" to avoid repeated checks for "do we have a large enough batch?".

* The 24 limit on the modified time is to allow us to catch up for really old repos with a long history of prefetch packs, but also because we want to delete pack-files on the next run.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+particular order:
+
+A. Loose Object Maintenance
+  1. Delete loose objects that exist in pack-files.
+  2. Place loose objects into pack-files.
+B. Prefetch Pack Maintenance
+  1. Delete packs with no objects referenced by the multi-pack-index.
+  2. Repack some prefetch packs into larger packs.
+
+Care must be taken to ensure that Git processes that initialized their object
```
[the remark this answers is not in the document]
derrickstolee: I imagine functional test scenarios will involve testing the verb on a mount that has background maintenance disabled (say, by running in unattended mode or by modifying the mount verb with an option). We can do things like check which loose objects are not in packs and ensure they are not deleted.

I don't think we can cover "what happens when a Git process is running at the same time" but must instead focus on which conditions allow us to delete a file and verify those are followed.

MaintenanceJobDesignDoc.md
derrickstolee: While I agree that this design should focus on the cache, I think that eventually we will need to help users clean up their .git folder. While they have many fewer local loose objects, as we have users keep enlistments for longer and longer, we will see performance degrade due to that, too. Normally, `git gc --auto` would kick in and do it for them, but we've disabled this command.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,140 @@
...
+### Delete loose objects that appear in a pack-file
+
+By running `git prune-packed -q` with the environment variable
+`GIT_OBJECT_DIRECTORY` set to the object cache, Git will delete the loose
+objects that exist in a pack-file. This will use the multi-pack-index to test
+containment, so is only bound by the file-deletion operations.
+
+### Pack loose objects that do not appear in a pack-file
+
+We can use `git pack-objects -q <object-dir>/pack/loose-<timestamp>` to create a
```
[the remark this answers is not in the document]
derrickstolee: Good point. Maybe `fromloose-<timestamp>`? `readobject-` (since objects were downloaded by read-object hook)?

MaintenanceJobDesignDoc.md
sanoursa: Yes we should make it _possible_ for a user to clean up their git objects, but I don't think it's something we should do as a proactive cleanup. We own the cache folder, and we should clean that up, and we should treat the repo's .git folder as the user's own data and not mess with it without an explicit gesture. Two ways to handle that particular issue down the road would be to re-enable `git gc` to work in VFS for Git repos, or we can add a new verb of our own to do repo-local cleanup. Either way, outside the scope of a cache maintenance feature.

derrickstolee: You're right. It shouldn't be in the background. Something to consider as an extension to the manual way to trigger maintenance. Perhaps `gvfs maintenance --local` or something similar. I'll drop it for now.

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,247 @@
...
+The `Stop()` method is important to test in the following states:
+
+* `Execute()` has not run.
+* `Execute()` has completed.
+* `Execute()` is running. This may require injecting "pauses" into the action
+  to check different stages during the execution. We could insert callbacks in
+  the action's constructor (that are `null` outside of tests) that allow us to
+  call `Stop()` at specific times in the `Execute()` method.
+
+How does the post-fetch job fit in?
```
derrickstolee: I had another idea here after I woke up this morning:

The post-fetch job could be the first action to leave a data file, and that file can store the last timestamp of a prefetch pack that was used in the commit-graph. We can then scan the pack-files to see which ones should be used. This can allow us to drop the message-passing algorithm, because we will do it in the background within 15 minutes. That changes the behavior slightly (we don't compute the multi-pack-index and commit-graph immediately after a `gvfs prefetch --commits`, but that is less and less likely to be a non-trivial action).

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,247 @@
...
+
+The multi-pack-index stores a single copy of every packed object. When an object
+appears in multiple pack-files covered by the multi-pack-index, we select the
+pack-file with the most-recent _modified time_ as reported by the filesystem.
+For example, if someone downloads a second large prefetch pack when prefetching
+for the first time in 31 days, then that large prefetch pack is referenced and
+all older prefetch packs are "stale" and not referenced by the multi-pack-index.
+
+We can extend the `git multi-pack-index` builtin to delete pack-files that have
+no objects referenced by the multi-pack-index. This should happen at the same
```
wilbaker: If there is a pack file with only a few objects referenced by the multi-pack-index would it be possible (and if so, should we) give it priority when deciding what to re-pack?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,247 @@
...
+  any Git processes that are being run.
+
+The actions are then run in sequence by an `ActionRunner` that triggers a run
+even on some interval, runs the actions that are ready, and tracks the
+currently-running action (so the mount process can tell the `ActionRunner` to
+`Stop()` and it will stop the running action). The `ActionRunner` can keep an
+ordered list of actions, and during each run of the background task it can
+check `IsReady(DateTime.UtcNow)` and run `Execute()` as necessary.
+
+Here are some initial ideas for the frequency of each action:
```
wilbaker: Something that might be worth considering (though I don't know how difficult it would be) is checking the state of the machine as well.  If we were able to give preference to times when the machine is locked (or the monitor has gone to sleep) we might be able to reduce this chances of these BG tasks taking resources from whatever the user happens to be doing on the machine.

wilbaker: Related: Currently can multiple prefetch actions run at the same time (if repos are using different .gvfsCache folders, or different subfolders of the same .gvfsCache folder)?  

If so, I wonder if we should have any cross-repo coordination to avoid having more than one of these running at once (we could also wait and see if anyone actually hits and perf\memory issues from this scenario).

MaintenanceJobDesignDoc.md
derrickstolee: That's an interesting idea! If we go this route, we want to specifically target the objects that are uniquely in those pack-files. This may be a better way to go (instead of repacking entire pack-files).

Here's how I see the approach you suggest working:

1. We create a `git multi-pack-index repack` verb (or a special mode for `git repack` that is multi-pack-index aware).
2. This Git process inspects the counts of packs marked by the multi-pack-index, and finds certain conditions for a pack being "stale" (say, modified time is old enough or there are at most X referenced objects).
3. The git process calls `git pack-objects`, passing the list of _objects_ (not the list of pack-files) and creates a pack.
4. The git process then updates the multi-pack-index with pack-level metadata saying the new pack-files are "repacked" pack-files and the old pack-files are "expired".
5. We keep the expired pack-files in the multi-pack-index until we repack again.

This is something I can really dig into and see what is possible. I'll want to make an RFC to the mailing list, as this seems like a big reason to have the multi-pack-index. In fact, [making `repack` use the `multi-pack-index` in a smart way is part of the design document's "Future Work" section](https://github.com/git/git/blob/master/Documentation/technical/multi-pack-index.txt#L86-L93).

(The good news is that I can let that idea simmer for a while as we work on the refactor and loose-object improvements.)

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,247 @@
...
+  as a timestamp in a list of pack-files or as the content of a file in
+  the shared object directory.)
+* Most logic happens during an `Execute()` method.
+* If the mount process needs to shut down while the action is active, we need
+  a `Stop()` method for the action to interrupt itself. This includes halting
+  any Git processes that are being run.
+
+The actions are then run in sequence by an `ActionRunner` that triggers a run
+even on some interval, runs the actions that are ready, and tracks the
+currently-running action (so the mount process can tell the `ActionRunner` to
```
jeschu1: Please make sure the new ActionRunner checks if the device is ready before running to address #481

MaintenanceJobDesignDoc.md
jeschu1: @derrickstolee would be good to know the result of your "Is the machine awake" investigation.  As you may be aware we are currently getting a lot of "Device is not ready" exceptions.  Do we expect success when the machine is locked?

MaintenanceJobDesignDoc.md
```diff
@@ -0,0 +1,247 @@
...
+<object-dir>/pack/repacked-<timestamp>`.
+
+We should refrain from packing the most-recent prefetch packs, as we don't want
+to get in the way of the prefetch machinery for determining the most-recent
+timestamp. We could, however, start repacking the prefetch packs from
+oldest-to-newest, stopping in the most-recent week, or when the prefetch packs
+are large enough to satisfy a batch. Some stopping conditions could include
+"pack-file size is above 1GB".
+
+Scheduling Actions
```
jeschu1: @derrickstolee will the first time this job is run on an existing enlistment be expensive?  

From the cadence suggestions below I'd expect subsequent runs to be cheaper (since the packs were already updated X time ago).

MaintenanceJobDesignDoc.md
wilbaker: >where would we put it? Do we have some application-level storage?

Currently our machine-wide data lives in C:\ProgramData\GVFS.  The one tricky part about that folder though is you *might* need elevated permissions to write to it.  I'm not sure if that can be adjusted on a per-file basis.

> For now, I think we should keep this global lock in the back of our minds, and instead focus on making the jobs run infrequently and in reasonably-sized batches.

Sounds good

derrickstolee: Good point! If every "idle" time actually locks the drive, then we will be sunk.

MaintenanceJobDesignDoc.md
derrickstolee: The hope is to batch the actions (say: only repack 1GB of data at a time) so even right away the cost is low. But we will want to detect that we are in the initial "catchup" phase and have a more frequent cadence for these batch-limited runs (daily instead of weekly?).

MaintenanceJobDesignDoc.md
derrickstolee: Thanks for that. I just pushed to #493 a commit that includes a check for the device being ready. It's a bit crude, but perhaps it is enough? [Please let me know what you think](https://github.com/Microsoft/VFSForGit/pull/493/commits/8226cb77bfe99ba206b4b6a3a59a3be6e11d064c).

derrickstolee:
Closing PR and remaining discussion is here:

* #493 is for the refactor of existing logic.
* #488 is for the new logic to add after the refactor.

PR #493, opened 2018-11-08
Refactor post-fetch job and background prefetch into GitMaintenanceSteps

derrickstolee:
The post-fetch job and background prefetch features used different mechanisms to run in the background. Let's unify these as much as possible, and create a framework for more background actions to happen with minimal interruption.

Most of the logic happens in `GitMaintenanceStep`, which does the following:

1. Takes a new "object cache lock" to prevent other cleanup steps from running. This can be ignored for user-initiated prefetches (by setting `RequireCacheLock` to `false`).
2. The `RunGitCommand` method safely runs an action on a `GitProcess`, allowing that process to be halted by the `Stop()` method.

The actions in these cleanup steps are best-effort: if we can't get the object cache lock, then we don't run the step.

Reoccurring cleanup steps are initiated by the `GitMaintenanceScheduler`. This is currently only the background prefetch step.

The `GitMaintenanceQueue` receives cleanup steps and runs them in a background thread until the queue is empty (or receives a `Stop()` request). It is responsible for checking if the filesystem is ready for the cleanup actions.


GVFS/GVFS.UnitTests/Maintenance/PackfileHelper.cs
```diff
@@ -0,0 +1,152 @@
...
+using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.UnitTests.Maintenance
+{
+    public class PackfileHelper
```
sanoursa: Did you intend to duplicate this class? I don't see it being used anywhere.

GVFS/GVFS.Common/Maintenance/MaintenanceJobManager.cs
```diff
@@ -0,0 +1,341 @@
...
+using GVFS.Common.Prefetch;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+
+namespace GVFS.Common.Maintenance
+{
+    public class MaintenanceJobManager : IDisposable
```
sanoursa: As I noted in your design doc, I think this name is too generic. I would rename it to reference "CacheMaintenance" so we can later on extend to repo maintenance as well.

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -549,15 +550,16 @@ private void MountAndStartWorkingDirectoryCallbacks(CacheServerInfo cache)
             this.heartbeat.Start();
 
             // Launch a background job to compute the multi-pack-index. Will do nothing if up-to-date.
-            this.fileSystemCallbacks.LaunchPostFetchJob(packIndexes: new List<string>());
+            this.maintenanceJobManager.LaunchPostFetchJob(packIndexes: new List<string>());
```
sanoursa: I think the callers of this new class don't really care about "jobs". We aren't even really doing anything that resembles a job, we're just doing some of the work asynchronously.

E.g. this call here could be simplified to `this.localCache.UpdateMidx`. The caller really doesn't care if that is sync or async, just that it's telling the local cache that there are potentially new packfiles available.

GVFS/GVFS.Common/Maintenance/MaintenanceJobManager.cs
sanoursa: The word "Manager" is usually a code smell that a class is doing too much, and in fact this class is doing more than just maintenance since it also generates new midx files post-fetch.

And as I noted lower down, we don't really need a concept of a "job" here. Jobs tend to imply that you create some background process that you have a handle to, so that you can check its status, cancel it, check if it is completed, etc. We don't have, or need, any of that here. It's just that some of our work happens to be asynchronous, so let's drop this job concept and simplify the mental model.

All of that being said, I think this is a useful class, and to me it might make sense to keep it as is and just call it the `LocalGVFSCache`. It has one public method for asking it to generate a new midx, and it also runs a background thread to clean itself up as needed. 

I would also move it out of the Maintenance namespace... unless you split it into two classes, one for the active operations on the cache and another for maintenance. I haven't thought too deeply yet about that approach, but that may be a cleaner separation of concerns.

GVFS/GVFS.UnitTests/Maintenance/PackfileHelper.cs
derrickstolee: I did not! I created it in the wrong place and cross-project move instead was a copy. Thanks for catching.

GVFS/GVFS.Common/Maintenance/MaintenanceJobManager.cs
derrickstolee: The main reason this class has the post-fetch operations and the background prefetch operations is the "stopping" logic, and how we cancel things during unmount. However, it is worth thinking of a better way to remove the operation logic of what a post-fetch is doing or a background prefetch is doing from these classes.

Each needs to have a `Stop()` method (which is a better name than the `TerminateJobs` that I have below).

derrickstolee: Also: some jobs need a shared file-based lock, but that could be in a separate constants class.

GVFS/GVFS.Common/Maintenance/MaintenanceJobManager.cs
```diff
@@ -0,0 +1,341 @@
...
+        public void WaitForMaintenanceJobToFinish()
+        {
+            this.maintenanceJobThread?.Join();
+        }
+
+        private void RunMaintenanceJobs()
+        {
+            try
+            {
+                this.BackgroundPrefetch();
```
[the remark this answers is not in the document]
sanoursa: I think this class already has too many responsibilities. It should not grow to handle even more.

It sounds like your primary goal is to have a shared implementation of "I need to run a thing in the background, and be able to shut it down in a reliable way". But that is a separate concern from the business logic of when/how to run cache maintenance, when/how to create a new midx, and when/how to clean up placeholder files. We should not mix all of those business logic concerns into the same class just because they need to share the same mechanism for doing background work.

GVFS/GVFS.Common/Maintenance/MaintenanceJobManager.cs
derrickstolee: I've got some ideas to help separate these concerns, improve testability, and make manual runs easier. I'm going to add them to the design document before I come back to this PR and re-do the refactor.

derrickstolee:
I updated this PR in the most-recent push to match the object model described in #492. Along the way, I found our special casing around unattended mode and no-cache-server cases. I plan to update the design document to match the changes I put in here.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
```diff
@@ -0,0 +1,184 @@
...
+            {
+                tracer.RelatedEvent(
+                    EventLevel.Informational,
+                    "Failed to kill background Git process during " + nameof(TryStopGitProcess),
+                    metadata: null);
+                return false;
+            }
+        }
+
+        public abstract bool IsReady(DateTime now);
```
derrickstolee: Move up above the implemented methods.

GVFS/GVFS.Common/Actions/WriteMultiPackIndexAction.cs
```diff
@@ -0,0 +1,46 @@
...
+            DateTime multiPackIndexTime = this.Context.FileSystem
+                                                     .GetFileProperties(midxPath)
+                                                     .LastWriteTimeUTC;
+
+            return this.HasNewerPrefetchPack(multiPackIndexTime);
+        }
+
+        protected override void RunGitAction()
+        {
+            string midxLock = Path.Combine(this.Context.Enlistment.GitPackRoot, MultiPackIndexLock);
```
derrickstolee: Add a similar lock-deletion to `WriteCommitGraphAction`. Add unit tests for this.

GVFS/GVFS.Common/Actions/WriteCommitGraphAction.cs
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }
+
+            if (packs.Count > MaxNumPacks)
+            {
+                packs.Sort();
+                packs.RemoveRange(0, packs.Count - MaxNumPacks);
+            }
+
+            this.RunGitCommand(process => process.WriteCommitGraph(this.Context.Enlistment.GitObjectsRoot, packs));
```
derrickstolee: It's possible that `packs.Count == 0`, in which case don't run this command. (For instance, when running `gvfs prefetch --commits` in unattended mode and not downloading any packs.)

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
```diff
@@ -0,0 +1,184 @@
...
+                        this.Context.Tracer.RelatedEvent(
+                            EventLevel.Informational,
+                            this.TelemetryKey + ": failed to kill background Git process during " + nameof(this.Stop),
+                            metadata: null);
+                    }
+                }
+            }
+        }
+
+        protected abstract void RunGitAction();
```
derrickstolee: Comment this method with the preconditions it guarantees (specifically, we hold the object-cache lock before calling this method).

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
```diff
@@ -0,0 +1,184 @@
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common.Actions
+{
+    public abstract class AbstractObjectCacheAction : IAction
```
sanoursa: Having read through the code, I'm not sure that we should have one class that manages both "background tasks that generate new content (prefetch, midx, etc)" and "background tasks that perform cache cleanup". I can see why you're trying to share their implementations and that makes sense - but my worry is that their goals and tradeoffs will be different and that mismatch will create bugs.

The fact that it's so hard to even name these classes is a good clue that they are doing too much. The best I can come up with is "BackgroundTask", "CleanupJob", etc, all of which are overly broad.

An alternative approach could be: create a "BestEffortBackgroundQueue" (as opposed to the required one that we have today) that allows any code to queue up optional work, and maintains a single thread (or pool of threads) to process that queue. Then you can have classes like "BackgroundPrefetch", "MidxGenerator", "LocalCacheCleanup", etc that can all have their own business logic and custom implementation tradeoffs, but share just the mechanism of how to run a thing in the background, and how to cancel it when unmounting.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
```diff
@@ -0,0 +1,184 @@
...
+                            this.TelemetryKey + ": failed to kill background Git process during " + nameof(this.Stop),
+                            metadata: null);
+                    }
+                }
+            }
+        }
+
+        protected abstract void RunGitAction();
+
+        protected bool HasNewerPrefetchPack(DateTime time)
```
derrickstolee: This can be pulled out as a static method in a `PackfileHelper` class.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
```diff
@@ -0,0 +1,184 @@
...
+            try
+            {
+                using (FileBasedLock cacheLock = GVFSPlatform.Instance.CreateFileBasedLock(
+                    this.Context.FileSystem,
+                    this.Context.Tracer,
+                    Path.Combine(this.Context.Enlistment.GitObjectsRoot, ObjectCacheLock)))
+                {
+                    if (!cacheLock.TryAcquireLock())
+                    {
+                        this.Context.Tracer.RelatedInfo(this.TelemetryKey + ": Skipping post-fetch work since another process holds the lock");
```
derrickstolee: Drop "post-fetch" from this string.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
derrickstolee: (I'm using "actions" for this discussion as I don't have a better word right now.)

You're right that this name is too generic, especially because it is doing a lot of logic around starting a `GitProcess` that it can terminate during `Stop()`. That should be the core of this class. As such, I should [move the pack-file modified time logic](https://github.com/Microsoft/VFSForGit/pull/493/files/da9e1b29d0fc0077420284304bf0c69f91b5f649#r233604138) into another class. The other logic around the `ObjectCacheLock` is important for protecting actions in different mount processes from fighting.

In hindsight, the `PrefetchAction` also runs `git` processes, and could benefit from this machinery (it doesn't extend this class). It would start multiple processes during the course of the action, but it could use `RunGitCommand` to do that in a safe way. This means we can remove `IAction` and have every action extend this class and rename it something like `GitObjectCacheMaintenanceStep`. We would want to split out an interface later if we decide to use this mechanism for things like the shared object size database, but YAGNI says we shouldn't do it now.

> ...that allows any code to queue up optional work...

The current post-fetch job works this way, so would be easy to replace the old `LaunchPostFetchJobIfIdle()` with a `QueuePostFetchJobs()` method. We could then separate the regularly-scheduled maintenance into `Timer`s that have different periods and all they do is queue a job. This sounds a bit like a background job agent.

We could then split this into:

* `BackgroundQueue` takes `QueueJob` requests and adds them to a `ConcurrentList`. It also holds a `Thread` object, that pulls jobs off of this list and runs them until the list is empty. After adding the job to the list, check if the thread is active. If the thread is active, then it will pick up the job when it can. If the thread is inactive, launch a new thread to empty the queue.
* `MaintenanceScheduler` creates a list of `Timer` objects that fire at different periods to queue jobs. The background prefetch job would run every 70 minutes. The loose object cleanup job could run less frequently.
* The `InProcessMount` class can listen for post-fetch job requests and queue those jobs according to the message contents.

I'm still not thrilled that the background prefetch sends a message over the named pipe to queue more background activity, but at least we consolidate the thread where we do the work.

The final thought I'm worried about is the "best effort" nature of this work. This is really for the actions to decide (hence why I dropped it from the class name), but we should also be careful if we actually want these actions to all take the same lock. For instance, can we write a commit-graph while we clean up loose objects? Probably. We don't want to do write two commit-graphs at the same time, and we don't want to write a multi-pack-index at the same time we may delete a pack-file. This means we should be more granular about the `ObjectCacheLock` that I'm currently managing in the base class.

kewillford: Wondering if it would simplify things if we had the service coordinate some of this since there is only one service that is running?  Mount processes could send requests to the service to run something but it would be the service's job to keep track of what is running and when?

derrickstolee: @kewillford The service is a nice singleton to help coordinate, but is it better than using a file-based lock? The stumbling block I am trying to overcome is what happens when a mount process suddenly dies while holding a "service lock"? The file-based lock opens up because the handle is dropped, right? Or maybe I'm misunderstanding what you are proposing.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
```diff
@@ -0,0 +1,184 @@
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common.Actions
```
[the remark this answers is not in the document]
derrickstolee: @sanoursa: We've been going in circles with names. My interpretation of your concerns are that we are too generic. I propose we err on the side of being too specific to what we are currently implementing. We are only cleaning up git data, so how about a base class of `GitCleanupStep`? The queue could then be `GitCleanupQueue`.

If we are instead going to keep this more generic, then the queue could be `BackgroundQueue` and the items going in the queue could be `Worker`s. This is after `Job`, `Action`, and `Task` are more likely to collide with existing C# ideas.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
sanoursa: I've thought about that as well but this would introduce a new failure mode for a mount process. Currently, the GVFS.Service is entirely optional for a mount process to function correctly. Moving some of the mount process's logic over to the service means that now we absolutely require the service to be running, or else you lose things like midx generation.

Also that move wouldn't change any of the design discussions we're having in this PR. It just changes which process that code runs in.

sanoursa: @derrickstolee 
> The final thought I'm worried about is the "best effort" nature of this work. This is really for the actions to decide (hence why I dropped it from the class name)

I disagree on that point. We have an existing background queue, and its special quality is that it is for absolutely required work, and that's why the queue is backed by disk. It is resilient to crash, power loss, etc, because that work cannot be lost.

The nature of this new queue is that it is best effort, meaning that it does not need to be backed by disk. A caller of this queue cannot decide that its action is required - if that is true, it is using the wrong queue.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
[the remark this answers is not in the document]
sanoursa: Yes the names are too generic, but that's just a symptom of the other discussion down below that the responsibilities of the classes are not yet well defined. Your new names are good - except your existing classes are not "cleanup steps", they generate new content. That's what I'm most worried about here - we're mixing steps that generate new content that is important for good perf with steps that cleanup a cache that would be nice to do when the machine is idle. They do not share the same goals, just a similar queueing mechanism.

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
derrickstolee: (I still need to do this in the new version.)

GVFS/GVFS.Common/Actions/AbstractObjectCacheAction.cs
derrickstolee: Thanks, both.

> The nature of this new queue is that it is best effort, meaning that it does not need to be backed by disk. A caller of this queue cannot decide that its action is required - if that is true, it is using the wrong queue.

Hopefully the new, very-specific name for the queue helps clarify the purpose of the queue, and the fact that it is "best-effort" is understood for the steps that are run by the queue.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,155 @@
...
+            {
+                return false;
+            }
+        }
+
+        private void RunQueue()
+        {
+            while (true)
+            {
+                lock (this.threadLock)
```
jeschu1: Can the lock just be around the:
this.thread = null;
portion?

There is also a very small race condition here where the drive becomes unavailable after the EnlistmentRootReady check, but the risk there is pretty small and we really don't have another option.

derrickstolee: But back to your actual question, I do think the lock could just be around the assignment.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,155 @@
...
+                    }
+                    catch (ThreadStateException e)
+                    {
+                        this.LogError(nameof(GitCleanupQueue), nameof(this.Enqueue), e);
+                    }
+                    catch (OutOfMemoryException e)
+                    {
+                        this.LogError(nameof(GitCleanupQueue), nameof(this.Enqueue), e);
+                    }
+                }
```
jeschu1: @derrickstolee I think we can greatly simplify this class by using a BlockingQueue.  I put a PR into your fork showing what i mean [here](https://github.com/derrickstolee/VFSForGit/pull/2/files).

derrickstolee: Thanks for this! It is a much cleaner approach that I would never create on my own.

derrickstolee:
@wilbaker when you get a chance (I know you're a bit busy), this PR is ready to go. @jeschu1 simplified the queue significantly, and I think the result is quite nice.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,103 @@
+﻿using GVFS.Common.Tracing;
+using System;
+using System.Collections.Concurrent;
+using System.IO;
+using System.Threading;
+
+namespace GVFS.Common.Cleanup
```
kewillford: I'm not sold on the word `Cleanup`.  Is Prefetch cleaning anything up or generation of the midx or commit graph?  What about "Optimize" or "Optimization"?

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,103 @@
...
+        public void Stop()
+        {
+            this.cancellationToken.Cancel();
+            this.currentStep?.Stop();
+        }
+
+        /// <summary>
+        /// This method is public for test purposes only.
+        /// </summary>
+        public bool EnlistmentRootReady()
```
kewillford: Can this be added to the `MockFileSystem` instead of the production code?

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,103 @@
...
+                if (this.EnlistmentRootReady())
+                {
+                    try
+                    {
+                        this.currentStep.Execute();
+                    }
+                    catch (Exception e)
+                    {
+                        this.LogErrorAndExit(
+                            telemetryKey: nameof(GitCleanupQueue),
```
kewillford: `telemetryKey` seems like the wrong word here.  It is more of an message prefix.  We use "Area" in other places in the code so maybe that is the right metadata field to use?

GVFS/GVFS.Common/Cleanup/GitCleanupScheduler.cs
```diff
@@ -0,0 +1,42 @@
...
+            this.prefetchStepTimer?.Dispose();
+            this.prefetchStepTimer = null;
+        }
+
+        public void ScheduleRecurringCleanupSteps()
+        {
+            if (!this.context.Unattended && this.gitObjects.IsUsingCacheServer())
+            {
+                this.prefetchStepTimer = new Timer(
+                (state) => this.queue.Enqueue(new PrefetchStep(this.context, this.gitObjects, requireCacheLock: true)),
```
kewillford: I realize we only have one of these right now but since the method name has `Steps` should we have a list of timers and recurring steps so it is easier to add to this later on.  I'm okay with waiting until we do have more than one to make this change though.  

The reason I kind of thought about this was because I was wondering why it was newing up the PrefetchStep every time the timer goes off.  Could this be a member variable?

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
```diff
@@ -0,0 +1,198 @@
...
+        private readonly object gitProcessLock = new object();
+
+        public GitCleanupStep(GVFSContext context, GitObjects gitObjects, bool requireCacheLock = false)
+        {
+            this.Context = context;
+            this.GitObjects = gitObjects;
+            this.RequireCacheLock = requireCacheLock;
+        }
+
+        public abstract string TelemetryKey { get; }
```
kewillford: Again I don't think `TelemetryKey` is the right name for this.

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
```diff
@@ -0,0 +1,198 @@
...
+                if (this.Stopping)
+                {
+                    this.Context.Tracer.RelatedWarning(
+                        metadata: null,
+                        message: this.TelemetryKey + ": Not launching Git process because the mount is stopping",
+                        keywords: Keywords.Telemetry);
+                    return null;
+                }
+
+                GitProcess.Result result = work.Invoke(this.GitProcess);
```
kewillford: Does this need the `gitProcessLock`?  If not I'm a little confused why we need the lock at all.

GVFS/GVFS.Common/Cleanup/PostFetchCleanupStep.cs
```diff
@@ -0,0 +1,47 @@
+﻿using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GVFS.Common.Cleanup
+{
+    public class PostFetchCleanupStep : GitCleanupStep
```
kewillford: Wondering if there should be a `WriteMultiPackIndexStep` and `WriteCommitGraphStep`  Are these two tied to each other?  Do they have to be run in order?  What if there are more post fetch steps?  Will they all be added to this class?

GVFS/GVFS.Common/Cleanup/PostFetchCleanupStep.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            using (ITracer activity = this.Context.Tracer.StartActivity("TryWriteMultiPackIndex", EventLevel.Informational, Keywords.Telemetry, metadata: null))
+            {
+                string multiPackIndexLockPath = Path.Combine(this.Context.Enlistment.GitPackRoot, MultiPackIndexLock);
+                this.Context.FileSystem.TryDeleteFile(multiPackIndexLockPath);
+
+                this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+            }
+
+            if (this.packIndexes == null || this.packIndexes.Count == 0)
```
kewillford: Why do we skip the commit graph but always write the multi pack index?

GVFS/GVFS.Common/Cleanup/GitCleanupScheduler.cs
```diff
@@ -0,0 +1,42 @@
...
+                state: null,
+                dueTime: this.prefetchPeriod,
+                period: this.prefetchPeriod);
```
kewillford: If there are network issues or for some reason the step is taking longer than the 15 minutes could we end up with a bunch of steps enqueued?  Should it start timing after the step is completed?

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -1,105 +1,147 @@
...
-                TrySchedulePostFetchJob(tracer, enlistment.NamedPipeName, packIndexes);
+                this.Context.Tracer.RelatedError(error);
+                return;
             }
 
-            return true;
+            DateTime lastDateTime = EpochConverter.FromUnixEpochSeconds(last);
+            DateTime now = DateTime.UtcNow;
+
+            if (now <= lastDateTime + this.timeBetweenPrefetches)
```
kewillford: Ah okay here is where it is making sure the prefetch step isn't running too often.  This almost seems like the schedulers job.  It should know when the step ran and finished so should it be checking time between and just not schedule the work?

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -147,27 +191,34 @@ public static class CommitPrefetcher
...
                         error = $"Unable to delete {packPath}";
                         return false;
                     }
                 }
             }
 
+            if (this.Stopping)
+            {
+                error = nameof(this.Stopping);
+                return false;
```
kewillford: Is stopping really an error condition?

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -204,7 +255,7 @@ private static bool TrySchedulePostFetchJob(ITracer tracer, string namedPipeName
             return false;
         }
 
-        private static long? GetTimestamp(string packName)
```
kewillford: Is there a reason for this change?

GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -47,9 +47,10 @@ public virtual bool TryDownloadCommit(string commitSha)
             const bool PreferLooseObjects = false;
             IEnumerable<string> objectIds = new[] { commitSha };
 
+            GitProcess gitProcess = new GitProcess(this.Enlistment);
```
kewillford: Would this be better as a member variable instead of passing it around everywhere?

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -29,7 +29,8 @@ public class InProcessMount
         private FileSystemCallbacks fileSystemCallbacks;
         private GVFSEnlistment enlistment;
         private ITracer tracer;
-        private BackgroundPrefetcher prefetcher;
+        private GitCleanupQueue cleanupQueue;
+        private GitCleanupScheduler cleanupScheduler;
```
kewillford: I'm wondering if we can encapsulate the queue in the scheduler so it will never be exposed and have a method on the scheduler to add a one time step instead of recurring?

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -1,105 +1,147 @@
...
-            for (int i = 0; i < orderedPacks.Count; ++i)
+            for (int i = 0; !this.Stopping && i < orderedPacks.Count; ++i)
             {
                 long timestamp = orderedPacks[i].Timestamp;
                 string packPath = orderedPacks[i].Path;
                 string idxPath = Path.ChangeExtension(packPath, ".idx");
-                if (!fileSystem.FileExists(idxPath))
+                if (!this.Context.FileSystem.FileExists(idxPath))
                 {
                     EventMetadata metadata = new EventMetadata();
```
wilbaker: Here (and elsewhere we're creating a new `EventMetadata` directly) we could switch to `CreateEventMetadata()`

GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -91,8 +92,23 @@ public virtual void DeleteStaleTempPrefetchPackAndIdxs()
...
         }
+        
+        public virtual void DeleteTemporaryFiles()
+        {
+            string[] temporaryFiles = this.fileSystem.GetFiles(this.Enlistment.GitPackRoot, "tmp_*");
+            foreach (string temporaryFilePath in temporaryFiles)
+            {
+                EventMetadata metadata = CreateEventMetadata();
+                metadata.Add(nameof(temporaryFilePath), temporaryFilePath);
+                metadata.Add(TracingConstants.MessageKey.InfoMessage, @"Deleting temporary file");
```
wilbaker: Nit: remove the `@` here: `@"Deleting temporary file"`

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -1,105 +1,147 @@
...
+            this.RunGitCommand(process =>
+            {
+                this.TryPrefetchCommitsAndTrees(out error, process);
+                return null;
+            });
+
+            if (!string.IsNullOrEmpty(error))
+            {
+                this.Context.Tracer.RelatedWarning(
+                    metadata: null,
```
wilbaker: Minor: Pass in `CreateEventMetadata()` here as well

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
derrickstolee: The scheduler is for running `Execute()` on some schedule.

The steps are responsible for checking if we should do anything during that `Execute()` step. For instance, the prefetch step can check the timestamps on the prefetch packs. The later steps can't use this, and they'll need their own mechanism.

Specifically for the larger maintenance steps in the future: we will need some custom data file in the object cache to describe how long it has been since it ran (say, about once a week for repack), but need to check shortly after mount because of users who restart their machines daily.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public void Enqueue(GitCleanupStep step)
+        {
+            this.queue.Add(step);
+        }
+
+        public void Stop()
+        {
+            this.cancellationToken.Cancel();
+            this.currentStep?.Stop();
```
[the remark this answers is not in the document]
jeschu1: Looking at the api/source we can probably replace the cancellationtoken in the class with a call to CompleteAdding().  That would make the class even smaller / sleeker!  We'd have to check what happens if you try to add in that scenario (does it throw an exception) and make sure we handle that case correctly.

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
derrickstolee: No, but returning `false` results in logging _something_, so it may as well be a message that we are stopping the step early.

I could rename `error` to `message`.

Otherwise, we would need to return an `enum` of return codes, but that seems overkill for something that won't happen often.

GVFS/GVFS.Common/Cleanup/PostFetchCleanupStep.cs
derrickstolee: The `multi-pack-index` call will quickly no-op if there are no new pack-files, but there could also be new pack-files (from a `gvfs prefecth --files` call) so there may be something to do.

The commit-graph command literally cannot do anything without a list of pack-files, as it uses that to determine commits to add to the file.

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -1,105 +1,147 @@
...
+            {
+                gitProcess = new GitProcess(this.Context.Enlistment);
+            }
+
             List<string> packIndexes;
             using (FileBasedLock prefetchLock = GVFSPlatform.Instance.CreateFileBasedLock(
-                fileSystem,
-                tracer,
-                Path.Combine(enlistment.GitPackRoot, PrefetchCommitsAndTreesLock)))
+                this.Context.FileSystem,
```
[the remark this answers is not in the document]
derrickstolee: We want foreground and background prefetches to not stomp on each other, but we want foreground prefetches to not be blocked on the background maintenance lock. I can add a comment.

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
derrickstolee: Nope! Will fix.

GVFS/GVFS.Common/Git/GitObjects.cs
derrickstolee: We need to pass it around to allow us to stop the background prefetch during an unmount (the `PrefetchStep` gets a `GitProcess` from its base class, and the base class handles shutting down the process on `Stop()`). Other commands that use those methods need to create their own process to pass to those methods.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
jeschu1: We just need to update the comment.  It's needed for production code to verify the device is ready

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
derrickstolee: I'll use `Maintenance` in the next version.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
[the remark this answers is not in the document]
derrickstolee: If we drop the cancellation token, then we just need to replace it with a `Stopping` boolean again. I'll add the `CompleteAdding()` to avoid adding some more, but I think we should keep the cancellation token.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
derrickstolee: This is necessary logic to check if the drive is locked due to BitLocker or an unplugged USB drive. We need it in production and only make it public so I could write unit tests.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        private void RunQueue()
+        {
+            while (true)
+            {
+                try
+                {
+                    this.queue.TryTake(out this.currentStep, Timeout.Infinite, this.cancellationToken.Token);
```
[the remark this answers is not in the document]
derrickstolee: If we don't use the cancellation token, then we will still successfully drain the queue. This would require our own `Stopping` boolean to prevent starting new steps.

So, do we want to replace the cancellation token with a boolean `Stopping` check?

GVFS/GVFS.Common/Cleanup/GitCleanupScheduler.cs
derrickstolee: When we have a list of three or four recurring steps, where do you want those specific steps to be added to the schedule? Do you want the `InProcessMount` to be responsible for that, or a third class that initializes the scheduler?

To me, pulling out the choices of _what_ is scheduled away from the logic of _how_ it recurs is unnecessary complication.

GVFS/GVFS.Common/Cleanup/GitCleanupScheduler.cs
derrickstolee: I think the steps should be responsible for checking if they "need to run" as the first thing inside of `Execute()`. This should be done under the lock. If we have five prefetch steps queued, then at least four of them will return quickly because the prefetch timestamp is too recent. Later steps will have similar ways to record when the last step of that type was run.

Note that it's not enough to limit the jobs in a single process, as multiple enlistments will be running steps at different times. If we have five enlistments trying to run background prefetch, the first to take the lock should win, and the rest give up (because of the lock or because of the recent timestamp).

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
```diff
@@ -0,0 +1,198 @@
+﻿using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common.Cleanup
+{
+    public abstract class GitCleanupStep
+    {
+        public const string ObjectCacheLock = "git-cleanup-step.lock";
```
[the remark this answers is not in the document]
derrickstolee: Good catch that `requireCacheLock` should default to `true`.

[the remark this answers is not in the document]
derrickstolee: I'm going to make the parameter explicit.

GVFS/GVFS.Common/Cleanup/GitCleanupScheduler.cs
jeschu1: @wilbaker , we could simply check the queue for the same task before adding it right?  Or alternatively draining any dup cleanupEvents off the queue when we take one off.  

This isn't a bad idea if we think we're going to suffer from backlogs.  With todays design would we just have 2 cleanup jobs (potentially the same) running at the same time in that scenario.

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                return false;
+            }
+
+            if (process.TryKillRunningProcess())
+            {
+                tracer.RelatedEvent(
+                    EventLevel.Informational,
+                    "Killed background Git process during " + nameof(TryStopGitProcess),
+                    metadata: null);
```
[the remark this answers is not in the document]
derrickstolee: (actually, this method is never called and should be deleted)

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
```diff
@@ -0,0 +1,198 @@
...
+                if (this.RequireCacheLock)
+                {
+                    using (FileBasedLock cacheLock = GVFSPlatform.Instance.CreateFileBasedLock(
+                        this.Context.FileSystem,
+                        this.Context.Tracer,
+                        Path.Combine(this.Context.Enlistment.GitObjectsRoot, ObjectCacheLock)))
+                    {
+                        if (!cacheLock.TryAcquireLock())
+                        {
+                            this.Context.Tracer.RelatedInfo(this.TelemetryKey + ": Skipping work since another process holds the lock");
```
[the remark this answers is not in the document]
derrickstolee: Yes, a prefetch step could block a post-fetch step. The basic philosophy of this background queue is "best effort" and that these steps never _need_ to run. It's just nice when they do.

If we want to work harder to ensure things run later if the lock is blocking, then we can extend the design later (say, by building a way to reschedule). I do think that is out of scope for this PR as it would be a significant change of behavior.

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
```diff
@@ -0,0 +1,198 @@
...
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Implement this method to perform a cleanup step. If the object-cache lock is required
+        /// (as specified by <see cref="RequireCacheLock"/>), then this step is not run unless we
+        /// hold the lock.
+        /// </summary>
```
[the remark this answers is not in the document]
derrickstolee: By "is always called" I think you mean the problem where the requireCacheLock was accidentally false.

I'll reword that first sentence.

GVFS/GVFS.Common/Cleanup/GitCleanupStep.cs
derrickstolee: The process lock is for changing the state of the process (initializing it, calling `TryKillRunningProcess()` during stop). It prevents us from creating a new process in `CreateProcessAndRun()` if we are stopping.

Locking the process during the execution of a git command would stop us from being able to halt the process.

GVFS/GVFS.Mount/InProcessMount.cs
jeschu1: same here like a cleanupScheduler. ScheduleOneTimeCleanupStep

GVFS/GVFS.Common/Cleanup/PostFetchCleanupStep.cs
derrickstolee: 1. They are currently run together, and this PR is intended to be a non-functional change when possible.
2. The event "I just ran prefetch" would trigger both of these events.
3. The commit-graph specifically needs the list of packfiles to run.
4. The commit-graph write will use the updated multi-pack-index file, but the multi-pack-index write does not use the commit-graph file.

So, in the future it _could_ make sense to queue these as separate cleanup jobs, but they would both be triggered by the same event. I'll leave that to a later change if we are interested in that.

jeschu1: For us newbies it would be great if we put some comments at the top of the class saying what it Cleans up.  I.E. Those two steps above with a short reasoning.

GVFS/GVFS.Common/Cleanup/PrefetchStep.cs
```diff
@@ -1,105 +1,147 @@
...
+                .Where(pack => this.GetTimestamp(pack).HasValue)
+                .Select(pack => new PrefetchPackInfo(this.GetTimestamp(pack).Value, pack))
                 .OrderBy(packInfo => packInfo.Timestamp)
                 .ToList();
 
             maxGoodTimestamp = -1;
 
             int firstBadPack = -1;
-            for (int i = 0; i < orderedPacks.Count; ++i)
+            for (int i = 0; !this.Stopping && i < orderedPacks.Count; ++i)
```
jeschu1: We check for this.Stopping a lot in the step.  Wondering if it's worth just letting the kill process do it's job so it isn't a frequent check in the step.

GVFS/GVFS.Common/Git/GitObjects.cs
jeschu1: I added a GitProcess to GVFSEnlistment in a previous PR.  Wondering if this should have been done at Enlistment itself.  Then this class could just take advantage of it.

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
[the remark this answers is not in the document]
wilbaker: > This would require our own Stopping boolean to prevent starting new steps.

Could we check `this.queue.IsAddingComplete` instead of our own Stopping boolean?

GVFS/GVFS.Common/Cleanup/GitCleanupQueue.cs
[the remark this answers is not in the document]
wilbaker: > I'll add the CompleteAdding() to avoid adding some more, but I think we should keep the cancellation token.

What scenario do we still need the cancellation token for? We can check `this.queue.IsAddingComplete` to see if we've called `CompleteAdding()` (and stop running steps if `IsAddingComplete` is true).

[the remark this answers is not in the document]
jeschu1: @derrickstolee pointed this out to me.

There could already be Steps already waiting in the queue.  By using the cancellation token we won't try to read them.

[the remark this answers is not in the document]
wilbaker: > There could already be Steps already waiting in the queue. By using the cancellation token we won't try to read them.

I think if we we switch to an if rather than a try\catch we should be able to avoid that scenario:

Something like this:

```
while (true)
{
    if (!this.queue.TryTake(out this.currentStep, Timeout.Infinite) || this.queue.IsAddingComplete)
    {
        // Stop has been requested
       return;
```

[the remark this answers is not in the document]
jeschu1: @wilbaker CompleteAdding() prevents new items from being added.  I would expect it to break out of the TryTake so that would wait there indefinitely right?

I like the approach of cancelling the TryTake which causes the background thread to end gracefully.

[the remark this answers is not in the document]
derrickstolee: That'll do! thanks.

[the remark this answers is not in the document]
derrickstolee: Good point, @jeschu1. Need to think about that. I know that the unit tests would wait until that `TryTake` completes if we didn't call `Stop()` directly. They seem to stop fine now, but I'll do some digging into the `TryTake()`.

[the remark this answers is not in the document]
derrickstolee: The `CompleteAdding` will [cause the `TryTake` to return `false`](https://stackoverflow.com/a/21195735).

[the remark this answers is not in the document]
wilbaker: > @wilbaker CompleteAdding() prevents new items from being added. I wouldn't expect it to break out of the TryTake so that would wait there indefinitely right?

`TryTake` will break out and return false if `CompleteAdding()` is called while it's waiting.

Internally the BlockingCollection uses a CancellationToken to break out of TryTakes that are waiting (without a user provided cancellation token).

*From TryTake*
https://referencesource.microsoft.com/#System/sys/system/collections/concurrent/BlockingCollection.cs,712

```
// set the combined token source to the combinedToken paramater if it is not null (came from GetConsumingEnumerable)
CancellationTokenSource linkedTokenSource = combinedTokenSource;
try
{
    waitForSemaphoreWasSuccessful = m_occupiedNodes.Wait(0);
    if (waitForSemaphoreWasSuccessful == false && millisecondsTimeout != 0)
    {
        // create the linked token if it is not created yet
        if (combinedTokenSource == null)
            linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken,
                                                                                m_ConsumersCancellationTokenSource.Token);
        waitForSemaphoreWasSuccessful = m_occupiedNodes.Wait(millisecondsTimeout, linkedTokenSource.Token);
    }
}
//The collection became completed while waiting on the semaphore.
catch (OperationCanceledException)
{
    if (cancellationToken.IsCancellationRequested)
        throw new OperationCanceledException(SR.GetString(SR.Common_OperationCanceled), cancellationToken);

    return false;
}
```

*From CompleteAdding:*
https://referencesource.microsoft.com/#System/sys/system/collections/concurrent/BlockingCollection.cs,1516
```
if (Interlocked.CompareExchange(ref m_currentAdders, observedAdders | COMPLETE_ADDING_ON_MASK, observedAdders) == observedAdders)
{
    spinner.Reset();
    while (m_currentAdders != COMPLETE_ADDING_ON_MASK) spinner.SpinOnce();

    if (Count == 0)
    {
        CancelWaitingConsumers();
    }
```

We have used this approach elsewhere VFS4G and haven't seen shutdown issues (e.g. `GitIndexGenerator.WriteAllEntries` and `FileSystemVirtualizer.ExecuteFileOrNetworkRequest`)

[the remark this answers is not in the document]
jeschu1: Thanks @wilbaker for the education there, yes that will definitely work nicely and eliminate the variable, nice! :-)

GVFS/GVFS.Common/Cleanup/GitCleanupScheduler.cs
derrickstolee: I started playing with this, but we actually want to be very careful with the schedule here.

* We don't want any background steps when unattended.

* We don't want to run prefetch automatically if we don't have a cache server. We'll still want to run the other steps, though.

* The other steps will want a shorter `dueTime` than their `period`, as we want them to run at most once per day, but want the steps to run when people restart their machine nightly.

I can make this class slightly more generic by creating a `List<Timer>` of timers, and add the timers to the list as necessary. Those timers can then be disposed from the list instead of one at a time.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -0,0 +1,172 @@
...
+                else
+                {
+                    this.CreateProcessAndRun();
+                }
+            }
+            catch (IOException e)
+            {
+                this.Context.Tracer.RelatedWarning(
+                    metadata: this.CreateEventMetadata(e),
+                    message: this.Area + ": IOException while running action: " + e.Message,
```
kewillford: Seems inconsistent how `this.Area` is being used.  In some places below it is only in the message and others it is in the metadata field `"Area"`.  Here it is in both. I think we would just need it in the metadata field and leave it out of the message.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -0,0 +1,172 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Implement this method perform the mainteance actions. If the object-cache lock is required
+        /// (as specified by <see cref="RequireObjectCacheLock"/>), then this step is not run unless we
+        /// hold the lock.
+        /// </summary>
+        protected abstract void RunGitAction();
```
kewillford: Should this be named PerformMaintenance? Currently it makes it sound like it will run one git action but in the post fetch it is running multiple.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -0,0 +1,172 @@
...
+                        message: this.Area + ": Git process failed with errors:" + result.Errors,
+                        keywords: Keywords.Telemetry);
+                    return result;
+                }
+
+                return result;
+            }
+        }
+
+        protected void LogWarning(string telemetryKey, string methodName, Exception exception)
```
kewillford: Is this being used anywhere?  I did a search in the web and couldn't find any.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        [TestCase]
+        public void GitMaintenanceStepRunsGitAction()
+        {
+            ITracer tracer = new MockTracer();
+            GVFSEnlistment enlistment = new MockGVFSEnlistment();
+            PhysicalFileSystem fileSystem = new MockFileSystem(new MockDirectory(enlistment.EnlistmentRoot, null, null));
+
+            GVFSContext context = new GVFSContext(tracer, fileSystem, null, enlistment);
+            GitObjects gitObjects = new MockPhysicalGitObjects(tracer, fileSystem, enlistment, null);
```
kewillford: Move these 5 lines of code to a `GVFSContext InitContext(out GitObjects gitObjects)` method to use in all the tests?

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
```diff
@@ -0,0 +1,200 @@
...
+                enlistment.EnlistmentRoot,
+                enlistment.GitObjectsRoot
+            });
+
+            GVFSContext context = new GVFSContext(tracer, fileSystem, null, enlistment);
+            GitObjects gitObjects = new MockPhysicalGitObjects(tracer, null, null, null);
+
+            GitMaintenanceQueue queue = new GitMaintenanceQueue(context);
+
+            queue.Stop();
```
kewillford: Is there anything we need to assert here?  Would failure be an exception?

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            // We need to have the EnlistmentRoot and GitObjectsRoot available for jobs to run
+            ReadyFileSystem fileSystem = new ReadyFileSystem(new string[]
+            {
+                enlistment.EnlistmentRoot,
+                enlistment.GitObjectsRoot
+            });
+
+            GVFSContext context = new GVFSContext(tracer, fileSystem, null, enlistment);
+            GitObjects gitObjects = new MockPhysicalGitObjects(tracer, null, null, null);
```
kewillford: All these tests have the same setup so can we have a method for that?

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            // then checks if Stop() was called.
+            WatchForStopStep watchForStop = new WatchForStopStep(queue, context, gitObjects);
+
+            queue.Enqueue(watchForStop);
+            Assert.IsTrue(watchForStop.EventTriggered.WaitOne(this.maxWaitTime));
+            watchForStop.SawStopping.ShouldBeTrue();
+
+            // Ensure we don't start a job after the Stop() call
+            TestGitMaintenanceStep watchForStart = new TestGitMaintenanceStep(context, gitObjects);
+            queue.Enqueue(watchForStart);
```
kewillford: Should this throw an exception or have a return value that we can check if it is something that isn't allowed?

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            TestGitMaintenanceStep step1 = new TestGitMaintenanceStep(context, gitObjects);
+            TestGitMaintenanceStep step2 = new TestGitMaintenanceStep(context, gitObjects);
+
+            GitMaintenanceQueue queue = new GitMaintenanceQueue(context);
+
+            queue.Enqueue(step1);
+            queue.Enqueue(step2);
+
+            Assert.IsTrue(step1.EventTriggered.WaitOne(this.maxWaitTime)
+                && step2.EventTriggered.WaitOne(this.maxWaitTime));
```
kewillford: Wondering if the EventTriggered should be a counter so we make sure it is only ran once?

GVFS/GVFS.Common/Maintenance/GitMaintenanceQueue.cs
```diff
@@ -0,0 +1,109 @@
...
+                if (!this.queue.TryTake(out this.currentStep, Timeout.Infinite)
+                    || this.queue.IsAddingCompleted)
+                {
+                    // A stop was requested
+                    this.queue.Dispose();
+                    this.queue = null;
+                    return;
+                }
+
+                if (this.EnlistmentRootReady())
```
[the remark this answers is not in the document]
derrickstolee: We shouldn't do any retry logic. We will keep scheduling things and will catch up any missed work by running future work.

GVFS/GVFS.Common/Maintenance/GitMaintenanceQueue.cs
```diff
@@ -0,0 +1,109 @@
...
+        public void Stop()
+        {
+            this.queue?.CompleteAdding();
+            this.currentStep?.Stop();
+        }
+
+        /// <summary>
+        /// This method is public for test purposes only.
+        /// </summary>
+        public bool EnlistmentRootReady()
```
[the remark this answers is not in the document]
derrickstolee: @mjcheetham Looks like similar things going on in #359. That PR doesn't look like it is ready, and we need to stop running background tasks when the drive becomes unavailable. We can make a note to replace this method with the deeper approach taken in #359 when it's available.

GVFS/GVFS.Common/Maintenance/GitMaintenanceQueue.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        private void LogError(string area, string methodName, Exception exception)
+        {
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("Method", methodName);
+            metadata.Add("ExceptionMessage", exception.Message);
+            metadata.Add("StackTrace", exception.StackTrace);
+            this.context.Tracer.RelatedError(
+                metadata: metadata,
+                message: area + ": Unexpected Exception while running maintenance steps (fatal): " + exception.Message,
```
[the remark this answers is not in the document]
derrickstolee: We have a `string area` param that I just forgot to add to the metadata. Will fix.

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                TimeSpan prefetchPeriod = TimeSpan.FromMinutes(15);
+                this.stepTimers.Add(new Timer(
+                    (state) => this.queue.Enqueue(new PrefetchStep(this.context, this.gitObjects, requireCacheLock: true)),
+                    state: null,
+                    dueTime: prefetchPeriod,
+                    period: prefetchPeriod));
+            }
+
+            // TODO: Schedule more mainenance steps here.
```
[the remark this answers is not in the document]
derrickstolee: The TODO is for  @jeschu1 to fill in as part of #488. I can remove the comment.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
derrickstolee: I audited instances of `CreateEventMetadata()` and removed the duplicate areas.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
derrickstolee: I'll do both, as we need to wait for the calls before we `Stop()` the queue.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
derrickstolee: Pretty much checking for failure here. Ensuring nothing is `null` or empty during the `Stop()` call.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceQueueTests.cs
derrickstolee: As long as we keep in mind that product code doesn't need to check for the return value. As far as the process is concerned, this is a noisy channel and steps can be dropped with no warning (because the lock is taken, because the drive is unavailable, because someone already called `Stop()` on the queue, etc.). Since we've encapsulated the queue in the scheduler, this surface is very small.

GVFS/GVFS.Common/Maintenance/GitMaintenanceQueue.cs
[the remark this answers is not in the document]
kewillford: So this will cause anything in the queue not to run and _hope_ that it will get scheduled again?  Maybe we need some comments around `EnqueueOneTimeStep` that there are no guarantees that it will run? Or maybe around the scheduler that it is a best effort?

derrickstolee:
@kewillford I've had this branch installed for a while, and I've been checking the mount logs to see the background behavior. I'll merge this and am available for a quick response to any issues you have.

PR #496, opened 2018-11-08
Stand up FastFetch on MacOS

nickgra:
This PR ports FastFetch to .NET Core and in doing so, stands up our first .NET Core project on Windows. 

I've attempted to make the commits individually reviewable as this review was getting large.

1.  Splits out the platform specific native code into Windows and implements the counterpart in the Mac platform. Converts FastFetch to a .NET Core SDK style project and makes it build on MacOS.
2. Fixes a bug that I introduced when porting our codebase to .NET Core. In RestorePacakges.bat, we only restored two of our .NET Standard projects. This results in these projects being restored into the user's home folder/.nuget.  Since FastFetch was .NET Framework up until now, we never had issues with LibGit2Sharp not being present for Windows builds to consume. Once it was removed, we no longer had a place to consume the native .dll from. Now we properly restore that entire .sln and point it at the VFSForGit packages directory, not the default home folder location.
3. Remove build dependency in .NET Framework projects on FastFetch (which breaks the build)
4. Changes to make FastFetch 'dotnet publish'-able on Windows as part of our build script. Attempting to publish the entire .sln results in a lot of issues with the vcxproj and .NET Framework projects in our .sln so we'll settle for just FastFetch now. We can't run stylecop in the .NET Core environment and we need to ensure all dependencies of FastFetch are targeting the same RunTime.
5. Changes to make the functional tests run on Windows (and pass) and some cleanup of Windows specific idioms.
6. Changes needed to track the mode of a file that is being hydrated through DiffTree so that we can create files with the proper modes when FastFetch-ing.
7-10. Address PR feedback. Most notably, instead of dangling new functions off of the GVFSPlatform, put all the platform specific code inside FastFetch itself and refactor more of FastFetch's code to be self contained (or to live inside FastFetch itself).

Things out of the scope of this PR that follow up issues are opened for:
- We have a self contained version of FastFetch now that we can have the build definition publish as an artifact. How should we publish FastFetch going forward?
- FastFetch.dll/.exe is no longer signed by GVFS.SignFiles. Can we sign .NET Core executables in the same way we sign Framework? Can we retarget SignFiles to be Standard if so? (A Framework project can't take a dependency on a Core/Standard project)


GVFS/FastFetch/FastFetch.csproj
```diff
@@ -1,104 +1,46 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\GVFS.Build\GVFS.cs.props" />
+  <Import Project="..\LibGit2Sharp.NativeBinaries.props" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <AssemblyName>fastfetch</AssemblyName>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <Platforms>x64</Platforms>
+    <RunTimeIdentifiers>osx-x64;win-x64</RunTimeIdentifiers>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
```
nickgra: After digging into how publishing works, I believe TargetLatestRuntimePatch is the source of our problems between building on the CLI and building in the IDE on MacOS. I made the changes here needed to get FastFetch buildable/publishable between the CLI and IDE on Windows, but I think more standardization here will fix this pain point for us on MacOS. I'll open an issue to track this down and get it fixed.

See https://docs.microsoft.com/en-us/dotnet/core/deploying/runtime-patch-selection for more info.

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -43,33 +46,71 @@ public bool HydrateFile(string fileName, byte[] buffer)
...
         [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
         private static extern int Chmod(string pathname, int mode);
 
         [DllImport("libc", EntryPoint = "rename", SetLastError = true)]
         private static extern int Rename(string oldPath, string newPath);
 
+        [DllImport("libc", EntryPoint = "write", SetLastError = true)]
+        private static unsafe extern IntPtr Write(int fileDescriptor, void* buf, IntPtr count);
+
         private class NativeFileReader
```
nickgra: I debated restructuring this class here since it's no longer just being used for the FileReader code in Prefetch, but I think it might make sense as a follow up PR so this one doesn't get too large.

Scripts/RestorePackages.bat
```diff
@@ -11,7 +11,8 @@ IF NOT EXIST %nuget% (
 
 %nuget% restore %VFS_SRCDIR%\GVFS.sln || exit /b 1
 
-dotnet restore %VFS_SRCDIR%\GVFS\GVFS.Common\GVFS.Common.csproj || exit /b 1
-dotnet restore %VFS_SRCDIR%\GVFS\GVFS.Virtualization\GVFS.Virtualization.csproj || exit /b 1
+SET VCTargetsPath=C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140
```
nickgra: This is needed for the .vcxproj's in our solution. They import .props files in this directory which is only set by the Framework version of msbuild. These projects won't actually restore anything under .NET Core, but the whole process will fail if they can't resolve their Imports.

Scripts/BuildGVFSForWindows.bat
```diff
@@ -39,4 +39,6 @@ IF NOT EXIST %msbuild% (
 
 %msbuild% %VFS_SRCDIR%\GVFS.sln /p:GVFSVersion=%GVFSVersion% /p:Configuration=%SolutionConfiguration% /p:Platform=x64 || exit /b 1
 
+SET SolutionDir=%VFS_SRCDIR%\
```
nickgra: This seems odd, but the .NET Core version of MSBuild is failing to resolve BuildOutputDir locations correctly without this change:
```
CSC : error CS2001: Source file 'C:\Users\nickgra\VFSForGit..\BuildOutput\CommonAssemblyVersion.cs' could not be found. [C:\Users\nickgra\VFSForGit\GVFS\GVFS.GVFlt\GVFS.GVFlt.csproj]
```

I think this change will be safe to make in the top level Initialize script, but I'll dig in deeper here.

nickgra:
Removing the WIP tag as this is ready for an initial review (more changes are needed for functional tests, but I hope they're relatively minor versus the changes here.)

Scripts/RestorePackages.bat
nickgra: I'll rework this (and other similar patterns) to pass this variable directly into msbuild/dotnet rather than setting environment variables per an offline discussion about this pattern.

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -153,6 +157,7 @@ public static DiffTreeResult ParseFromLsTreeLine(string line, string repoRoot)
                 {
                     DiffTreeResult blobAdd = new DiffTreeResult();
                     blobAdd.TargetIsSymLink = line.StartsWith("120000");
+                    blobAdd.TargetMode = line.Substring(0, 6);
```
wilbaker: Minor, do you think these two lines would be easier to follow if the `StartsWith` was removed?

```
blobAdd.TargetMode = line.Substring(0, 6);
blobAdd.TargetIsSymLink = blobAdd.TargetMode == "120000";
```

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -84,6 +86,8 @@ public static DiffTreeResult ParseFromDiffTreeLine(string line, string repoRoot)
             DiffTreeResult result = new DiffTreeResult();
             result.SourceIsDirectory = ValidTreeModes.Contains(parts[0]);
             result.TargetIsDirectory = ValidTreeModes.Contains(parts[1]);
+            result.SourceMode = parts[0];
+            result.TargetMode = parts[1];
 
             if (!result.TargetIsDirectory)
             {
```
wilbaker: This line:

`result.TargetIsSymLink = parts[1] == "120000";`

Can now be changed to:

`result.TargetIsSymLink = result.TargetMode == "120000";`

Which (I think) is more readable.

GVFS/GVFS.Common/Prefetch/Jobs/ReadFilesJob.cs
```diff
@@ -38,8 +39,9 @@ protected override void DoWork()
                 string blobId;
                 while (this.availableBlobs.TryTake(out blobId, Timeout.Infinite))
                 {
-                    foreach (string path in this.blobIdToPaths[blobId])
-                    {
+                    foreach (Tuple<string, string> modeAndPath in this.blobIdToPaths[blobId])
+                    {
+                        string path = modeAndPath.Item2;
```
wilbaker: With your own class type you could drop the `path` temporary variable and just pass in `modeAndPath.Path` to `HydrateFile`.

Scripts/RunFunctionalTests.bat
```diff
@@ -4,7 +4,7 @@ CALL %~dp0\InitializeEnvironment.bat || EXIT /b 10
 IF "%1"=="" (SET "Configuration=Debug") ELSE (SET "Configuration=%1")
 
 SETLOCAL
-SET PATH=C:\Program Files\GVFS;C:\Program Files\Git\cmd;%PATH%
+SET PATH=C:\Program Files\GVFS;%VFS_PUBLISHDIR%\FastFetch;C:\Program Files\Git\cmd;%PATH%
```
wilbaker: How was this working before?

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -26,7 +26,7 @@ public class FastFetchTests
         public void InitControlRepo()
         {
             Directory.CreateDirectory(this.fastFetchControlRoot);
-            GitProcess.Invoke("C:\\", "clone -b " + Settings.Default.Commitish + " " + GVFSTestConfig.RepoToClone + " " + this.fastFetchControlRoot);
+            GitProcess.Invoke(this.fastFetchControlRoot, "clone -b " + Settings.Default.Commitish + " " + GVFSTestConfig.RepoToClone + " " + this.fastFetchControlRoot);
```
wilbaker: Will this call fail if `this.fastFetchControlRoot` doesn't exist?  The first parameter is `executionWorkingDirectory` and so I'm wondering if we should use a different path (e.g. the parent of `this.fastFetchControlRoot`).

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -45,14 +45,28 @@ public void InitRepo()
...
         {
-            CmdRunner.DeleteDirectoryWithUnlimitedRetries(this.fastFetchControlRoot);
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                CmdRunner.DeleteDirectoryWithUnlimitedRetries(this.fastFetchRepoRoot);
+            }
+            else
+            {
+                BashRunner.DeleteDirectoryWithUnlimitedRetries(this.fastFetchRepoRoot);
+            }
```
wilbaker: Put this duplicated code in a helper method? Ideally it would be shared with `SharedCacheTests` and `ControlGitRepo` as well (they also have the same code).

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -185,9 +200,10 @@ public void CanFetchAndCheckoutBranchIntoEmptyGitRepo()
 
             this.fastFetchRepoRoot.ShouldBeADirectory(FileSystemRunner.DefaultRunner)
                 .WithDeepStructure(FileSystemRunner.DefaultRunner, this.fastFetchControlRoot);
-        }
-
-        [TestCase]
+        }
+
+        [TestCase]
+        [Category(Categories.MacTODO.M4)]
```
wilbaker: Why are some of the tests failing on Mac?

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -476,16 +493,24 @@ private ProcessResult RunFastFetch(string args)
...
 
-            string fastfetch = Path.Combine(Settings.Default.CurrentDirectory, "fastfetch.exe");
-            if (!File.Exists(fastfetch))
-            {
-                fastfetch = "fastfetch.exe";
+            string fastfetch;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                fastfetch = Path.Combine(Settings.Default.CurrentDirectory, "netcoreapp2.1", "fastfetch.dll");
+                if (!File.Exists(fastfetch))
```
wilbaker: Do you know if this check is still needed? I can't remember why we'd fall back on "fastfetch.exe", @sanoursa, do you have any ideas?

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -54,6 +59,86 @@ public bool IsSocket(string fileName)
...
+            [MarshalAs(UnmanagedType.LPTStr)] string filename,
+            [MarshalAs(UnmanagedType.U4)] FileAccess access,
+            [MarshalAs(UnmanagedType.U4)] FileShare share,
+            IntPtr securityAttributes, // optional SECURITY_ATTRIBUTES struct or IntPtr.Zero
+            [MarshalAs(UnmanagedType.U4)] FileMode creationDisposition,
+            [MarshalAs(UnmanagedType.U4)] FileAttributes flagsAndAttributes,
+            IntPtr templateFile);
+
+        [DllImport("kernel32.dll")]
+        [return: MarshalAs(UnmanagedType.Bool)]
```
wilbaker: I can't see a reason for including `[return: MarshalAs(UnmanagedType.Bool)]`

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -1,14 +1,19 @@
 ﻿using GVFS.Common;
 using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
 using Microsoft.Win32.SafeHandles;
 using System;
+using System.ComponentModel;
 using System.IO;
 using System.Runtime.InteropServices;
+using static GVFS.Common.Git.LibGit2Repo;
```
wilbaker: What was the reason for using `static` here?  Was it messy to use `LibGit2Repo` in the code below?

GVFS/FastFetch/FastFetch.csproj
```diff
@@ -1,104 +1,46 @@
...
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <Version>$(GVFSVersion)</Version>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <Version>$(GVFSVersion)</Version>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
nickgra: It looks like we've only set DebugType in .NET Framework projects and the MirrorProvider csprojs. A quick search of how DebugType interacts with .NET Core shows that SDK style projects imply the new style portable PDBs, which we'll want for debugging our code on MacOS. I want to do more research into this before I remove it from the MirrorProvider .csprojs, but I've been able to debug VFS for Git's .NET Core on MacOS with the new portable PDBs, so I think it's safe to remove.

https://stackoverflow.com/questions/46902377/what-is-the-difference-between-using-debugtypefull-debugtype-and-debugtype

GVFS/GVFS.Build/GVFS.cs.props
```diff
@@ -1,16 +1,21 @@
...
+
   <PropertyGroup>
     <OutputPath>$(BuildOutputDir)\$(MSBuildProjectName)\bin\$(Platform)\$(Configuration)\</OutputPath>
-    <IntermediateOutputPath>$(BuildOutputDir)\$(MSBuildProjectName)\obj\$(Platform)\$(Configuration)\</IntermediateOutputPath>
+    <IntermediateOutputPath>$(BuildOutputDir)\$(MSBuildProjectName)\obj\$(Platform)\$(Configuration)\</IntermediateOutputPath>
     <IntDir>$(BuildOutputDir)\$(MSBuildProjectName)\intermediate\$(Platform)\$(Configuration)\</IntDir>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(MSBuildRuntimeType)' == 'Core'">
+    <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'true'">true</IsOSX>
```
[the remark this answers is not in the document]
nickgra: Moved. What do you think is so dangerous about this? I can see a future version of .NET Core defining properties with these names.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -40,7 +42,7 @@ public DiffHelper(ITracer tracer, Enlistment enlistment, GitProcess git, IEnumer
 
             this.DirectoryOperations = new ConcurrentQueue<DiffTreeResult>();
             this.FileDeleteOperations = new ConcurrentQueue<string>();
-            this.FileAddOperations = new ConcurrentDictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
+            this.FileAddOperations = new ConcurrentDictionary<string, HashSet<Tuple<string, string>>>(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
nickgra: Thanks, moving to my own class made this a lot cleaner and readable!

GVFS/GVFS.Virtualization/GVFS.Virtualization.csproj
```diff
@@ -14,8 +15,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Data.Sqlite" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <PackageReference Include="StyleCop.Error.MSBuild" Version="1.0.0" Condition="'$(OS)' == 'Windows_NT'" />
-    <PackageReference Include="StyleCop.MSBuild" Version="5.0.0" Condition="'$(OS)' == 'Windows_NT'" />
+    <PackageReference Include="StyleCop.Error.MSBuild" Version="1.0.0" Condition="('$(OS)' == 'Windows_NT') AND ('$(MSBuildRuntimeType)' != 'Core')" />
```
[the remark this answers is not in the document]
nickgra: We need to disable Stylecop when `dotnet publish`-ing as it uses System.Windows.Forms (for some reason). This is why it's never worked for us in the command line build but works fine in Visual Studio Mac (VSMac runs in Mono and Mono provides a copy of Windows.Forms)

It's worth calling out that this change will disable Stylecop running in VSMac. I've never had good luck with it running in that configuration. Whenever I have the bandwidth, I'd like to move us onto the modern version of Stylecop (https://github.com/DotNetAnalyzers/StyleCopAnalyzers) which should work correctly everywhere without hacks like this.


```
error MSB4062: The "StyleCopTask" task could not be loaded from the assembly C:\Users\nickgra\.nuget\packages\stylecop.msbuild\5.0.0\build\..\tools\StyleCop.dll. Could not load file or assembly 'System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [C:\Users\nickgra\VFSForGit\GVFS\GVFS.Virtualization\GVFS.Virtualization.csproj]
```

Scripts/RunFunctionalTests.bat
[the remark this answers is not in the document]
nickgra: This is unneeded. It looks like today we don't have a story for testing FastFetch on the PATH (unless we manually put it there) so this was an experiment that can be reverted.

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
nickgra: The two tests that are failing are due to the way we create the index in FastFetch. I'll address these in a follow up PR to make this one smaller. 

In the tests, we compare the index created by FastFetch with one that 'git status' creates. We compare them by 'git ls-files --debug' which outputs in this format. 

Settings.StyleCop
  ctime: 1540245129:777055100
  mtime: 1540245129:777055100
  **dev: 0        ino: 0**
  uid: 0        gid: 0
  size: 10738   flags: 0
nuget.config
  ctime: 1540245129:777055100
  mtime: 1541723805:620569300
  **dev: 0        ino: 0**
  uid: 0        gid: 0
  size: 617     flags: 0

The differences between the two indexes ends up being that dev/ino are 0 in FastFetch's index while they have real random values in the git created index.

I believe the fix is to just ignore dev/ino when comparing the ls-files output but I want to dig a little deeper before sweeping this under the rug.

More context:
These are unix internals that track the /dev device that contains the file and a unique inode number for the file. These don't mean anything on Windows, so Git for Windows writes these out as 0s (experimentally, still need to dig through the source to prove this). As far as I can tell, Git doesn't use these values for anything.

https://github.com/git/git/blob/master/Documentation/technical/index-format.txt
```
        // We can't accurated fill times and length in realtime, so **we block write the zeroes and probably save time.**
        private static readonly byte[] EntryHeader = new byte[] 
        {
            0, 0, 0, 0,
            0, 0, 0, 0, // ctime
            0, 0, 0, 0,
            0, 0, 0, 0, // mtime
            0, 0, 0, 0, // stat(2) dev
            0, 0, 0, 0, // stat(2) ino
            0, 0, 0x81, 0xA4, // filemode (0x81A4 in little endian)
            0, 0, 0, 0, // stat(2) uid
            0, 0, 0, 0, // stat(2) gid
            0, 0, 0, 0  // file length
        };
```

GVFS/GVFS.Platform.Windows/GVFS.Platform.Windows.csproj
```diff
@@ -131,9 +132,9 @@
     <Error Condition="!Exists('..\ProjectedFSLib.NativeBinaries.props')" Text="$([System.String]::Format('$(ErrorText)', '..\ProjectedFSLib.NativeBinaries.props'))" />
     <Error Condition="!Exists('..\..\..\packages\StyleCop.MSBuild.5.0.0\build\StyleCop.MSBuild.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\..\packages\StyleCop.MSBuild.5.0.0\build\StyleCop.MSBuild.targets'))" />
   </Target>
-  <Import Project="..\..\..\packages\StyleCop.Error.MSBuild.1.0.0\build\StyleCop.Error.MSBuild.Targets" Condition="Exists('..\..\..\packages\StyleCop.Error.MSBuild.1.0.0\build\StyleCop.Error.MSBuild.Targets')" />
+  <Import Project="..\..\..\packages\StyleCop.Error.MSBuild.1.0.0\build\StyleCop.Error.MSBuild.Targets" Condition="Exists('..\..\..\packages\StyleCop.Error.MSBuild.1.0.0\build\StyleCop.Error.MSBuild.Targets') AND ('$(MSBuildRuntimeType)' != 'Core')" />
```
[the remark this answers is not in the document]
nickgra: See the other comment on StyleCop and why this is required now that we're `dotnet publishing` on Windows.

GVFS/GVFS.SignFiles/GVFS.SignFiles.csproj
```diff
@@ -36,7 +36,6 @@
   <ItemGroup>
     <FilesToSign Include="
       $(BuildOutputDir)\GitHooksLoader\bin\$(Platform)\$(Configuration)\GitHooksLoader.exe;
-      $(BuildOutputDir)\FastFetch\bin\$(Platform)\$(Configuration)\FastFetch.exe;
```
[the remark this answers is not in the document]
nickgra: We don't have a FastFetch.exe at this point as it's created by `dotnet publish`. We could do something like what Ameen's installer PR does to run a publish as part of a build, but I'm not sure how the signing infrastructure works and if it would be able to handle a .NET Core binary. Additionally, now that FastFetch is .NET Core, the SignFiles .NET Framework project can't take a build dependency on it anymore. I would hope we could make SignFiles .NET Standard to work around this but I'll wait for @jrbriggs to get back.

This is something else I want to follow up with.

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
nickgra: Check out mseng PR#269997 for the PR where this change was introduced. From the comments there, it looks like this was to work around issues in the Windows build lab. Since you left the comments on this code there, I'll defer to your good judgment on if this is now dead code. :)

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -387,12 +389,13 @@ private void EnqueueFileAddOperation(ITracer activity, DiffTreeResult operation)
                 activity.RelatedEvent(EventLevel.Warning, "CaseConflict", metadata);
             }
 
+            // TODO(Nick): Write EqualityComparison for HashSet<Tuple<string,string>>
```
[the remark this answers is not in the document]
nickgra: After looking at this code again after reworking it to use a class, I don't think I need to mess with `Equals` and `GetHashCode`. The AddOrUpdate code will prevent duplicates from entering FileAddOperations and we can still use the default StringComparison for the SHA1 keys of this dictionary. 
Is there something I'm missing here?

Scripts/BuildGVFSForWindows.bat
nickgra: Your explanation is correct. I fixed this and chose to pass SolutionDir into `dotnet` since that is technically the missing variable (once it's defined, BuildOutputDir and friends are all defined correctly per our props files).
Publishing the solution at this time was getting complicated due to some of our custom MSBuild logic in GVFS.Upgrader. I'd prefer to address that as part of a future PR. I know saying that is what caused our bug with `dotnet restore`, but I'd like to rationalize all of our .NET Core stuff sooner rather than later.

GVFS/GVFS.Common/Git/DiffTreeResult.cs
nickgra: This code is now a little bit stranger because of the ushort conversions, but if you have a good suggestion on how to make this cleaner, I'd love to take it.

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -45,14 +45,28 @@ public void InitRepo()
...
-        {
-            CmdRunner.DeleteDirectoryWithUnlimitedRetries(this.fastFetchRepoRoot);
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                CmdRunner.DeleteDirectoryWithUnlimitedRetries(this.fastFetchRepoRoot);
+            }
+            else
+            {
+                BashRunner.DeleteDirectoryWithUnlimitedRetries(this.fastFetchRepoRoot);
```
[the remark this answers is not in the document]
nickgra: This code was commented out by @wilbaker with the following comment in PR:
```
When this code was uncommented I saw the functional tests would get stuck in DeleteDirectoryWithRetry, I have not yet debugged things further.
```

This was in standing up the hooks for VFSForGit on MacOS, which have been stable for a while now so I'm going to try moving all uses of this pattern to a common function. We'll see if it causes hangs on the functional tests now.

GVFS/GVFS.Common/Prefetch/Git/PathWithMode.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public class PathWithMode
+    {
+        public PathWithMode(ushort mode, string path)
+        {
+            this.Mode = mode;
+            this.Path = path;
+        }
+
+        public ushort Mode { get; set; }
```
sanoursa: Remove these setters, this class should be immutable (and I don't see anything other the constructor setting these properties currently anyway)

GVFS/FastFetch/CheckoutJob.cs
```diff
@@ -262,23 +262,23 @@ private void HandleAllFileAddOperations()
                     
                     Interlocked.Increment(ref this.shasReceived);
 
-                    HashSet<string> paths;
-                    if (this.diff.FileAddOperations.TryRemove(availableBlob, out paths))
+                    HashSet<PathWithMode> modesAndPaths;
```
sanoursa: Nit: why `modesAndPaths`? The path is the key so I would make that first.

GVFS/GVFS.FunctionalTests/Settings.cs
```diff
@@ -47,6 +47,8 @@ public static void Initialize()
                     string root = "/GVFS.FT";
                     EnlistmentRoot = Path.Combine(root, "test");
                     ControlGitRepoRoot = Path.Combine(root, "control");
+                    FastFetchRoot = Path.Combine(root, "FFtest");
```
sanoursa: Please don't abbreviate the folder names

sanoursa: And I would try to keep a consistent naming pattern as we do for the Windows tests (see line 40)

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
sanoursa: The intended behavior of the original code was to run the command outside of any repo, which is why it was using the `C:\` drive root. I don't think we should pass in the control root. Any path outside of it (the parent directory?) would do fine.

GVFS/FastFetch/CheckoutJob.cs
sanoursa: Or just leave the name `paths`

GVFS/FastFetch/FastFetch.csproj
```diff
@@ -1,104 +1,52 @@
...
+
+  <PropertyGroup Condition="'$(MSBuildRuntimeType)' == 'Core'">
+    <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'true'">true</IsOSX>
+    <IsLinux Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">true</IsLinux>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
+    <ProjectReference Include="..\GVFS.Virtualization\GVFS.Virtualization.csproj" />
```
sanoursa: What are we using from `GVFS.Virtualization.csproj`?

sanoursa: From the looks of it, nothing, except that PlatformLoader uses a type from that namespace - which FastFetch never uses.

I think we should avoid adding this reference, FastFetch should have no dependency on virtualization code. All it would take is to split up the GVFSPlatformLoader class into two separate (partial?) classes, to avoid that unwanted dependency.

GVFS/GVFS.Common/Prefetch/Git/PathWithMode.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace GVFS.Common.Prefetch.Git
+{
+    public class PathWithMode
+    {
+        public PathWithMode(ushort mode, string path)
```
sanoursa: Super nitpicky - make path first

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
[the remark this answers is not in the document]
wilbaker: Previously this data structure:

```
this.FileAddOperations = new ConcurrentDictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
```

Would prevent the same path from being added to the `HashSet` more than once.

With the new changes:

```
this.FileAddOperations = new ConcurrentDictionary<string, HashSet<PathWithMode>>(StringComparer.OrdinalIgnoreCase);
```

The same path can be added to the `HashSet` multiple times.

Based on the description in the PR that included the change, it sounds like something we don't want to lose:

> Use a hashset for destination filepaths. (ls-tree was producing the same filename twice for many files. This change is the biggest win, although 3 provides similar value.)

GVFS.sln
```diff
@@ -29,7 +29,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GVFS.Common", "GVFS\GVFS.Co
 		{A4984251-840E-4622-AD0C-66DFCE2B2574} = {A4984251-840E-4622-AD0C-66DFCE2B2574}
```
sanoursa: This isn't showing up in any of your diffs, so I'm leaving it here.

In a few places, the fastfetch code does "git to Windows" and "Windows to git" conversions. It appears that those code paths are doing the right thing since they're mostly actually doing "git to .NET" and ".NET to git", and the .NET code is platform-aware. But please review all of those instances, confirm that they are correct, and rename them as appropriate.

The ones I have found are:
* `Index.ToWindowsTime` and `ToUnixTime`
* `FromGitRelativePathToWindowsFullPath` and `FromWindowsFullPathToGitRelativePath`, both called from `Index.cs`

That was just from a quick search for the word "Windows" in the fastfetch code. A more thorough search would be helpful as well.

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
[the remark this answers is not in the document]
nickgra: This still causes hangs so the copy in GVFSFunctionalTestEnlistment will stay commented out. I've opened an issue to track this: #554

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
wilbaker: Could you file a follow up issue for this?  It would be good to confirm if it's OK that FastFetch leaves these fields as 0 on Mac.

CC: @kewillford who might be more familiar with how git uses these fields

GVFS/GVFS.Common/Git/DiffTreeResult.cs
wilbaker: Maybe something like:

```
private const ushort SymLinkFileIndexEntry = 0xA000;

result.TargetIsSymLink = result.TargetMode == SymLinkFileIndexEntry;
```

Note that SymLinkFileIndexEntry is already in `GitIndexProjection.FileTypeAndMode`, and so you could move it from there to a common location .

wilbaker: You could do the same check below at line 159 as well:

`blobAdd.TargetIsSymLink = line.StartsWith("120000");`

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
kewillford: The index is a cache of what was on disk that last time status ran.  In most cases a status call will just update these values the next time it runs and stats the file.  There are places that the fields are used like [match_stat_data](https://github.com/Microsoft/git/blob/43ac10287170d00704cf69aca4702c28ffd2a03d/read-cache.c#L150) but that is mainly to know if the index entry needs to be refreshed.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -369,9 +369,9 @@ private void EnqueueFileAddOperation(ITracer activity, DiffTreeResult operation)
             // Each filepath should be case-insensitive unique. If there are duplicates, only the last parsed one should remain.
             if (!this.filesAdded.Add(operation.TargetPath))
             {
-                foreach (KeyValuePair<string, HashSet<string>> kvp in this.FileAddOperations)
+                foreach (KeyValuePair<string, HashSet<PathWithMode>> kvp in this.FileAddOperations)
                 {
-                    if (kvp.Value.Remove(operation.TargetPath))
+                    if (kvp.Value.Any(m => m.Path == operation.TargetPath))
```
wilbaker: This line used to remove the entry from this.FileAddOperations, and now it's just checking for existence (and `Any` will be linear time rather than constant).

Can you switch back to `Remove` and create a temporary `PathWithMode` with a dummy mode?  With the change to use the path for GetHashCode\Equals you should still be able to call `Remove`.

GVFS/GVFS.Common/Prefetch/Git/PathWithMode.cs
```diff
@@ -0,0 +1,32 @@
...
+            this.Mode = mode;
+        }
+
+        public ushort Mode { get; }
+        public string Path { get; }
+
+        public override bool Equals(object obj)
+        {
+            PathWithMode x = obj as PathWithMode;
+            if (x.Path.Equals(this.Path, StringComparison.OrdinalIgnoreCase))
```
wilbaker: Check for `null` first.

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
```diff
@@ -145,15 +145,15 @@ public void DeleteEnlistment()
         {
             TestResultsHelper.OutputGVFSLogs(this);
 
+            // TODO(Mac): Figure out why the call to DieleteDirectoryWithRetry is not returning.
+            // Once this is resolved, we can replace this duplicated code with RepositoryHelpers.DeleteTestDirectory
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                // Use cmd.exe to delete the enlistment as it properly handles tombstones and reparse points
```
wilbaker: Please leave in this comment

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
```diff
@@ -145,15 +145,15 @@ public void DeleteEnlistment()
         {
             TestResultsHelper.OutputGVFSLogs(this);
 
+            // TODO(Mac): Figure out why the call to DieleteDirectoryWithRetry is not returning.
```
wilbaker: Typo: `DieleteDirectoryWithRetry`

GVFS/GVFS.FunctionalTests/Tools/RepositoryHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace GVFS.FunctionalTests.Tools
+{
+    public static class RepositoryHelpers
+    {
+        public static void DeleteTestDirectory(string repoPath)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                CmdRunner.DeleteDirectoryWithUnlimitedRetries(repoPath);
```
wilbaker: Add `// Use cmd.exe to delete the enlistment as it properly handles tombstones and reparse points` comment here as well

GVFS/GVFS.Common/Prefetch/Git/PrefetchLibGit2Repo.cs
```diff
@@ -42,38 +36,9 @@ public virtual bool TryCopyBlobToFile(string sha, IEnumerable<string> destinatio
...
-                                    }
-                                }
-                                catch (Exception e)
-                                {
-                                    this.Tracer.RelatedError("Exception writing {0}: {1}", destination, e);
-                                    throw;
-                                }
+                            foreach (PathWithMode destination in destinations)
+                            {
+                                GVFSPlatform.Instance.FileSystem.WriteFile(this.Tracer, originalData, originalSize, destination.Path, destination.Mode);
```
wilbaker: How ugly would it be to leave all of the `unsafe` code in this class (for both platforms)?

That would avoid having to spread `unsafe` to so many places.

@sanoursa, what are your thoughts on that approach?

Scripts/RunFunctionalTests.bat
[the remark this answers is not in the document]
wilbaker: > This is unneeded. It looks like today we don't have a story for testing FastFetch on the PATH (unless we manually put it there) so this was an experiment that can be reverted.

This comment was marked as resolved but I still see the code change.  Is the plan to revert the change?

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -476,16 +480,21 @@ private ProcessResult RunFastFetch(string args)
...
 
-            string fastfetch = Path.Combine(Settings.Default.CurrentDirectory, "fastfetch.exe");
-            if (!File.Exists(fastfetch))
-            {
-                fastfetch = "fastfetch.exe";
+            string fastfetch;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                fastfetch = Path.Combine(Settings.Default.CurrentDirectory, "netcoreapp2.1", "fastfetch.dll");
+                File.Exists(fastfetch).ShouldBeTrue();
```
wilbaker: Please move this line:

`File.Exists(fastfetch).ShouldBeTrue();`

After the else block.  It should be true on both platforms.

GVFS/GVFS.FunctionalTests/Settings.cs
```diff
@@ -36,7 +37,8 @@ public static void Initialize()
                     PathToGit = @"C:\Program Files\Git\cmd\git.exe";
                     PathToBash = @"C:\Program Files\Git\bin\bash.exe";
 
-                    ControlGitRepoRoot = @"C:\Repos\GVFSFunctionalTests\ControlRepo";
+                    ControlGitRepoRoot = @"C:\Repos\GVFSFunctionalTests\ControlRepo";
+                    FastFetchBaseRoot = @"C:\Repos\GVFSFunctionalTests\FastFetch";
                     FastFetchRoot = @"C:\Repos\GVFSFunctionalTests\FastFetch\Test";
                     FastFetchControl = @"C:\Repos\GVFSFunctionalTests\FastFetch\Control";
```
wilbaker: Use `Path.Combine` here as well for `FastFetchRoot` and `FastFetchControl`

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
nickgra: Thanks guys! I went ahead and made a change here to remove the lines starting with dev: and uid: when doing the index comparison in the tests. This behaves as expected.

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace GVFS.Common.FileSystem
+﻿using GVFS.Common.Tracing;
```
wilbaker: `using GVFS.Common.Tracing;` can be removed

GVFS/LibGit2Sharp.NativeBinaries.props
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <ItemGroup>
-        <None Condition="Exists('..\..\..\packages\LibGit2Sharp.NativeBinaries.1.0.165\runtimes\win7-x64\native\git2-1196807.dll')" Include="..\..\..\packages\LibGit2Sharp.NativeBinaries.1.0.165\runtimes\win7-x64\native\git2-1196807.dll">
+        <None Condition="Exists('..\..\..\packages\libgit2sharp.nativebinaries\1.0.165\runtimes\win7-x64\native\git2-1196807.dll')" Include="..\..\..\packages\libgit2sharp.nativebinaries\1.0.165\runtimes\win7-x64\native\git2-1196807.dll">
```
wilbaker: Why was this change made?

GVFS/GVFS.SignFiles/GVFS.SignFiles.csproj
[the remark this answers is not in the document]
wilbaker: If we don't have a work item already to sign FastFetch.exe, can you create one?

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using GVFS.Common;
 using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
 using System;
+using System.ComponentModel;
 using System.IO;
```
wilbaker: No need to add `using GVFS.Common.Tracing;` or `using System.ComponentModel;`

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
```diff
@@ -145,15 +145,16 @@ public void DeleteEnlistment()
         {
             TestResultsHelper.OutputGVFSLogs(this);
 
+            // TODO(Mac): Figure out why the call to DeleteDirectoryWithRetry is not returning.
```
wilbaker: Can you file an issue for this?

GVFS/GVFS.Common/Prefetch/Git/PathWithMode.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public override bool Equals(object obj)
+        {
+            if (obj == null)
+            {
+                return false;
+            }
+
+            PathWithMode x = obj as PathWithMode;
+            if (x.Path.Equals(this.Path, StringComparison.OrdinalIgnoreCase))
```
wilbaker: Check for `null` (after `as`) to avoid exception when accessing `Path`

GVFS/GVFS.PlatformLoader/PlatformLoader.Mac.cs
```diff
@@ -1,17 +1,10 @@
...
-using GVFS.Common.Git;
 using GVFS.Platform.Mac;
-using GVFS.Virtualization.FileSystem;
 
 namespace GVFS.PlatformLoader
 {
-    public static class GVFSPlatformLoader
+    public static partial class GVFSPlatformLoader
     {
-        public static FileSystemVirtualizer CreateFileSystemVirtualizer(GVFSContext context, GVFSGitObjects gitObjects)
```
wilbaker: Are the `PlatformLoader` changes still needed (now that FastFetch has the native calls for both platforms in its own project)?

GVFS/FastFetch/FastFetchLibGit2Repo.cs
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    switch (Native.Object.GetType(objHandle))
+                    {
+                        case Native.ObjectTypes.Blob:
+                            byte* originalData = Native.Blob.GetRawContent(objHandle);
+                            long originalSize = Native.Blob.GetRawSize(objHandle);
+
+                            foreach (PathWithMode destination in destinations)
+                            {
+                                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
wilbaker: `IsOSPlatform` performs a string comparison:

*IsOSPlatform(Mac)*
https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Runtime.InteropServices.RuntimeInformation/src/System/Runtime/InteropServices/RuntimeInformation/RuntimeInformation.Unix.cs#L22

*IsOSPlatform(Windows)*
https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Runtime.InteropServices.RuntimeInformation/src/System/Runtime/InteropServices/RuntimeInformation/RuntimeInformation.Windows.cs#L19

*OSPlatform.Equals*
https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Runtime.InteropServices.RuntimeInformation/src/System/Runtime/InteropServices/RuntimeInformation/OSPlatform.cs#L37

And so we should switch to setting either a `bool` or an `Action<...>` in the constructor to avoid calling `IsOSPlatform` for every blob.

GVFS/FastFetch/NativeMacMethods.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public const int ReadOnly = 0x0000;
+        public const int WriteOnly = 0x0001;
+
+        public const int Create = 0x0200;
+        public const int Truncate = 0x0400;
+
+        public static unsafe void WriteFile(ITracer tracer, byte* originalData, long originalSize, string destination, ushort mode)
+        {
+            int fileDescriptor = -1;
```
wilbaker: Minor: Add a named constant for -1 (e.g. `InvalidFileDescriptor`)

GVFS/FastFetch/NativeMacMethods.cs
```diff
@@ -0,0 +1,51 @@
...
+                throw;
+            }
+            finally
+            {
+                Close(fileDescriptor);
+            }
+        }
+
+        [DllImport("libc", EntryPoint = "open", SetLastError = true)]
+        public static extern int Open(string path, int flag, int creationMode);
```
wilbaker: I think that `creationMode` should be a `ushort`.

From the `open` manpage:

> The oflag argument may indicate that the file is to be created if it does not exist (by specifying the O_CREAT flag).  In this case, open() and openat() require an additional argument mode_t mode; the file is created with mode mode as described in chmod(2) and modified by the process' umask value (see umask(2)).

*_mode_t.h*
`typedef	__darwin_mode_t		mode_t;`

*_types.h*
`typedef __uint16_t	__darwin_mode_t;	/* [???] Some file attributes */`

GVFS/GVFS.Common/FileSystem/IPlatformFileSystem.cs
wilbaker: `using GVFS.Common.Tracing;` can be removed

GVFS/LibGit2Sharp.NativeBinaries.props
wilbaker: Why was this change made?

GVFS/GVFS.SignFiles/GVFS.SignFiles.csproj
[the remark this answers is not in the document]
wilbaker: If we don't have a work item already to sign FastFetch.exe, can you create one?

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using GVFS.Common;
 using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
 using System;
+using System.ComponentModel;
```
wilbaker: No need to add `using GVFS.Common.Tracing;` or `using System.ComponentModel;`

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
```diff
@@ -145,15 +145,16 @@ public void DeleteEnlistment()
         {
             TestResultsHelper.OutputGVFSLogs(this);
 
+            // TODO(Mac): Figure out why the call to DeleteDirectoryWithRetry is not returning.
+            // Once this is resolved, we can replace this duplicated code with RepositoryHelpers.DeleteTestDirectory
```
wilbaker: Can you file an issue for this?

GVFS/GVFS.Common/Prefetch/Git/PathWithMode.cs
wilbaker: Check for `null` (after `as`) to avoid exception when accessing `Path`

GVFS/GVFS.PlatformLoader/PlatformLoader.Mac.cs
```diff
@@ -1,17 +1,10 @@
 using GVFS.Common;
-using GVFS.Common.Git;
 using GVFS.Platform.Mac;
-using GVFS.Virtualization.FileSystem;
 
 namespace GVFS.PlatformLoader
 {
-    public static class GVFSPlatformLoader
+    public static partial class GVFSPlatformLoader
```
wilbaker: Are the `PlatformLoader` changes still needed (now that FastFetch has the native calls for both platforms in its own project)?

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -476,16 +503,21 @@ private ProcessResult RunFastFetch(string args)
...
-                fastfetch = "fastfetch.exe";
+                fastfetch = Path.Combine(Settings.Default.CurrentDirectory, "fastfetch.dll");
             }
 
+            File.Exists(fastfetch).ShouldBeTrue();
             Console.WriteLine($"Using {fastfetch}");
 
-            ProcessStartInfo processInfo = new ProcessStartInfo(fastfetch);
-            processInfo.Arguments = args;
+            ProcessStartInfo processInfo = new ProcessStartInfo("dotnet");
```
wilbaker: Why are the tests running against "fastfetch.dll"? Can we point them to the published FastFetch?

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -214,34 +216,59 @@ public void CanFetchAndCheckoutAfterDeletingIndex()
...
+        //  nuget.config
+        //  ctime: 1540245129:777055100
+        //  mtime: 1541723805:620569300
+        //  dev: 0 ino: 0
+        //  uid: 0 gid: 0
+        //  size: 617 flags: 0
+        //
+        // The index that FastFetch crafts fills the dev/ino/uid/gid fields of the index with 0 for speed
+        // Git for Windows writes these fields out as 0, but once a Unix version of Git writes out the index,
+        // they get filled with actual values. This breaks our Index comparison logic, so we'll remove these
```
wilbaker: > but once a Unix version of Git writes out the index, they get filled with actual values

Does FastFetch need to use actual values (on Mac)?  What is the impact if it does not?

GVFS/FastFetch/FastFetchLibGit2Repo.cs
nickgra: Great catch, thanks! I moved this into the constructor.

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
nickgra: #554

GVFS/GVFS.PlatformLoader/PlatformLoader.Mac.cs
nickgra: It's not needed as part of this PR after pulling code directly into FastFetch. However, I think this change is still virtuous, although it's rather small. Would you rather me split this off into a separate PR?

GVFS/GVFS.SignFiles/GVFS.SignFiles.csproj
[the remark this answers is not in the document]
nickgra: We have #574 for figuring out the future of distributing FastFetch. I assume that would cover the work to sign it. Would you prefer that I open a separate issue?

GVFS/LibGit2Sharp.NativeBinaries.props
nickgra: This is in the second commit's description in the PR:
> Fixes a bug that I introduced when porting our codebase to .NET Core. In RestorePacakges.bat, we only restored two of our .NET Standard projects. This results in these projects being restored into the user's home folder/.nuget. Since FastFetch was .NET Framework up until now, we never had issues with LibGit2Sharp not being present for Windows builds to consume. Once it was removed, we no longer had a place to consume the native .dll from. Now we properly restore that entire .sln and point it at the VFSForGit packages directory, not the default home folder location.

GVFS/GVFS.PlatformLoader/PlatformLoader.Mac.cs
wilbaker: > I think this change is still virtuous

What's the benefit of this change?

> Would you rather me split this off into a separate PR?

I'd rather this change be its own small PR.  This PR is large enough that it will be helpful to restrict the changes to only those needed to enable FastFetch.

GVFS/GVFS.SignFiles/GVFS.SignFiles.csproj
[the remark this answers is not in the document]
wilbaker: > We have #574 for figuring out the future of distributing FastFetch. I assume that would cover the work to sign it. Would you prefer that I open a separate issue?

Could you file a separate issue?  The signing issue is separate from how the binaries are distributed.

/ cc: @jrbriggs

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
nickgra: On MacOS, we run the FunctionalTests out of the Publish folder, so we can easily find the published FastFetch executable. On Windows, we're currently running the FunctionalTests out of the BuildOutput folder. Instead of reworking how we run tests on Windows at this time, I went for the cheap fix of running FastFetch via `dotnet`. Whenever we move the rest of the project to .NET Core (or extract FastFetch into its own codebase), this can be cleaned up.

GVFS/GVFS.Build/GVFS.cs.props
```diff
@@ -1,11 +1,11 @@
...
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
+
   <Import Project="GVFS.props" />
-
+
   <PropertyGroup>
     <OutputPath>$(BuildOutputDir)\$(MSBuildProjectName)\bin\$(Platform)\$(Configuration)\</OutputPath>
-    <IntermediateOutputPath>$(BuildOutputDir)\$(MSBuildProjectName)\obj\$(Platform)\$(Configuration)\</IntermediateOutputPath>
+    <IntermediateOutputPath>$(BuildOutputDir)\$(MSBuildProjectName)\obj\$(Platform)\$(Configuration)\</IntermediateOutputPath>
```
wilbaker: Did something change on this line?

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -0,0 +1,146 @@
...
+                    IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
+                    if (result.ToInt32() == -1)
+                    {
+                        throw new Win32Exception(Marshal.GetLastWin32Error());
+                    }
+                }
+            }
+            catch (Win32Exception e)
+            {
+                tracer.RelatedError("Error writing file {0}. Win32Exception: {1}", destination, e);
```
wilbaker: - Put `destination` and `exception` in an `EventMetadata`
- Call `e.ToString()` so that we have the exception call stack

GVFS/FastFetch/NativeWindowsMethods.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.IO.MemoryMappedFiles;
+using System.Runtime.InteropServices;
+
+namespace FastFetch
+{
+    public class NativeWindowsMethods
+    {
+        private const int AccessDeniedWin32Error = 5;
+
+        public static unsafe void WriteFile(ITracer tracer, byte* originalData, long originalSize, string destination, ushort mode /* ignored on windows */)
```
wilbaker: Since there's no inheritance relationship now, can the `mode` parameter be dropped from this method?

GVFS/FastFetch/NativeMethods.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace FastFetch
+{
+    internal static class NativeMethods
+    {
+        public static bool isUnixOS = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) || RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
+
+        public static unsafe void WriteFile(ITracer tracer, byte* originalData, long originalSize, string destination, ushort mode)
+        {
+            if (!isUnixOS)
```
wilbaker: Super minor style comment (and this might just be me), but I think this code (and below in `StatAndUpdateIndexForFile`) would be a bit more readable if you preferred affirmative checks (i.e. `if (isUnixOS)`) rather than negative checks (`if (!isUnixOS)`).

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -0,0 +1,146 @@
...
+                throw;
+            }
+            finally
+            {
+                Close(fileDescriptor);
```
wilbaker: Is is OK to call `close` on an invalid file descriptor?

GVFS/FastFetch/NativeMethods.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                NativeWindowsMethods.WriteFile(tracer, originalData, originalSize, destination, mode);
+            }
+            else
+            {
+                NativeUnixMethods.WriteFile(tracer, originalData, originalSize, destination, mode);
+            }
+        }
+
+        public static bool StatAndUpdateIndexForFile(string path, MemoryMappedViewAccessor indexView, long offset)
```
wilbaker: Naming suggestion: `StatAndUpdateIndexForFile` -> `TryStatFileAndUpdateIndex` (since we're not stating the index).

(This is assuming the return value indicates success)

GVFS/FastFetch/NativeWindowsMethods.cs
```diff
@@ -0,0 +1,121 @@
...
+                        }
+
+                        size -= written;
+                        data = data + written;
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                tracer.RelatedError("Exception writing {0}: {1}", destination, e);
```
wilbaker: Same comment here about using metadata and calling `ToString` (I realize the code might have been like this before but we might as well clean it up).

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+        public static bool StatAndUpdateIndexForFile(string path, MemoryMappedViewAccessor indexView, long offset)
+        {
+            NativeStat.StatBuffer st = StatFile(path);
```
wilbaker: Should we be catching exceptions thrown by `StatFile` and returning false? 

Related: I noticed the Windows implementation returns false if the file does not exist and we're not doing that here.

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -0,0 +1,146 @@
...
+            // #define S_IXOTH     0000001     /* [XSI] X for other */
+            private static readonly ushort IXOTH = Convert.ToUInt16("1", 8);
+
+            public static bool IsSock(ushort mode)
+            {
+                // #define  S_ISSOCK(m) (((m) & S_IFMT) == S_IFSOCK)    /* socket */
+                return (mode & IFMT) == IFSOCK;
+            }
+
+            public static bool IsExecutable(ushort mode)
```
wilbaker: If `FastFetch` is not using any of the readonly ushorts (e.g, `IXGRP`) or helper methods (`IsSock`, `IsExecutable`) they can be removed.

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
wilbaker: Could you file a follow up issue to run the tests against the exe?  In general we've tried to run the functional tests as black-box tests against the final binaries.

GVFS/FastFetch/Index.cs
```diff
@@ -187,15 +188,14 @@ private bool UpdateFileInformationFromWorkingTree(MemoryMappedViewAccessor index
...
                             {
-                                IndexEntry indexEntry = new IndexEntry(indexView, offset);
-                                indexEntry.Mtime = file.LastWriteTimeUtc;
-                                indexEntry.Ctime = file.CreationTimeUtc;
-                                indexEntry.Size = (uint)file.Length;
-                                Interlocked.Increment(ref updatedEntries);
+                                if (NativeMethods.StatAndUpdateIndexForFile(gitPath, indexView, offset))
+                                {
+                                    Interlocked.Increment(ref updatedEntries);
+                                }
```
wilbaker: It looks like the old code did not have a failure mode once making it past `if (this.indexEntryOffsets.TryGetValue(gitPath, out offset))`.

If we get a failure now, should we log an error here?

GVFS/GVFS.SignFiles/GVFS.SignFiles.csproj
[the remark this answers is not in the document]
nickgra: #623

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
nickgra: #624

GVFS/FastFetch/Index.cs
nickgra: I'll log an error here. Note that before we incremented updatedEntries without checking for errors here, which seemed like a bug/mistake from looking at the internal repo's history.

GVFS/FastFetch/NativeUnixMethods.cs
nickgra: Close will return -1 and set errno = EBADF. We don't check the return code of close here, so this is fine as is.

GVFS/FastFetch/NativeUnixMethods.cs
nickgra: Updated it to catch the exception and take a tracer. We'll get an exception from stat if the file doesn't exist so we should be covered for that case. Alternatively, we can test to see if the file exists here, but this is a very hot path so I'm a little concerned about perf.

GVFS/FastFetch/Index.cs
```diff
@@ -213,14 +213,18 @@ private bool UpdateFileInformationFromDiskForFiles(MemoryMappedViewAccessor inde
...
                         if (this.indexEntryOffsets.TryGetValue(gitPath, out offset))
                         {
-                            if (this.TryUpdateEntryFromDisk(indexView, localPath, offset))
+                            if (NativeMethods.TryStatFileAndUpdateIndex(this.tracer, gitPath, indexView, offset))
                             {
                                 Interlocked.Increment(ref updatedEntriesFromDisk);
                             }
+                            else
+                            {
+                                this.tracer.RelatedError("Failed to update file information from disk for file {0}", gitPath);
```
wilbaker: Not a big deal because these errors are only coming from FastFetch, but I've found it helpful to include the method name that's generating the error.

Example:

```
this.tracer.RelatedError($"{nameof(this.UpdateFileInformationFromDiskForFiles)}: Failed ...
```

GVFS/FastFetch/NativeMethods.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                NativeWindowsMethods.WriteFile(tracer, originalData, originalSize, destination);
+            }
+        }
+
+        public static bool TryStatFileAndUpdateIndex(ITracer tracer, string path, MemoryMappedViewAccessor indexView, long offset)
+        {
+            if (isUnixOS)
+            {
+                return NativeUnixMethods.StatAndUpdateIndexForFile(tracer, path, indexView, offset);
```
wilbaker: Rename the platform specific helpers to `TryStatFileAndUpdateIndex` as well

GVFS/FastFetch/NativeWindowsMethods.cs
```diff
@@ -0,0 +1,127 @@
...
+                // Skip these.
+            }
+            catch (System.UnauthorizedAccessException)
+            {
+                // Skip these.
+            }
+
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("path", path);
+            tracer.RelatedError(metadata, "Error stat-ing file.");
```
wilbaker: This means we'll log an error when the file does not exist, and I'm not sure that's what we want.

Based on how this method is called in `UpdateFileInformationForAllEntries` it appears that it's expected that it will sometimes return false.

I'm thinking the logging you added in `UpdateFileInformationFromDiskForFiles` should be sufficient, what do you think? (if you do decide to log something here, it should probably be Info rather than Error.

nickgra: I agree that the logging in `UpdateFileInformationFromDiskForFiles` should be sufficient here. I'll remove this.

PR #497, opened 2018-11-09
Adjust line endings

dscho:
In a single-platform project, one might get away with not caring about line endings. But as soon as multiple platforms are involved, it is important to clear up the picture.

I suspect that this PR would have prevented the problems described in https://github.com/Microsoft/git/issues/63


dscho:
An offline chat suggests that this PR is actually not the solution for https://github.com/Microsoft/git/issues/63. Therefore I'll close this.

Issue #284, opened 2018-09-20
Mac: unmount should unregister virtualization root in the kext


sanoursa:
Unmount should send a message to the kext to unregister the virtualization root and do any necessary cleanup as a result. For example, when the final virtualization root on a mount point is unregistered, we could re-enable the auth cache.

Similar logic should kick in when the kext detects that the provider has gone down without a clean handshake.

pmj:
I've got a prototype implementation for this now, although it currently uses linear search in an array, and the nuts & bolts of the data structure are mixed in with the business logic, so I'm going to split this up before posting the PR.

PR #500, opened 2018-11-12
Mac: Reset auth cache on provider disconnecting, improve VirtualizationRoot array correctness

pmj:
This fixes issue #284.

The actual fix is implemented in the final commit (5177369) and is quite simple: when a provider connects, check if the (VFS) mount point inside which its virtualisation root lives is already in use by another provider. If not, save the mount point's kauth TTL in an array of mount points and disable the kauth cache on the mount point. Either way, increment a counter.

When a provider disconnects, it looks up the saved mount point, decrementing the use counter. When the counter hits 0, the mount point's kauth cache TTL is reset to the saved, original value, and removes the mount point from the array.

The other 2 commits contain the messy logic of the resizeable associative array. My original approach to this fix had the array logic intermingled with the auth cache logic, and it wasn't especially nice, so I moved the array memory management, key search, insertion and removal to a reusable templated class instead. Some nice side effects of this, aside from keeping the business logic clean:

 * It's reusable within the kext and also unit testable, so I've also supplied 2 minimalistic test cases using the XCTest framework Apple provides with Xcode.
 * We can easily swap it out with a different data structure later (For example a hash table if the vnode cache goes ahead as I'm envisaging it.)

I'm not deeply invested in this approach, and it's certainly different from how we've previously done this sort of thing. For example, for the array of `VirtualizationRoot` structs, we directly manage the memory in `FindUnusedIndexOrGrow_Locked()`, although the way it's done there could technically be unsafe as we call `MemoryAlloc` without first dropping the lock. Anyway, as the kext increases in complexity, I thought it might be worth abstracting some if this sort of thing, so critique away! I can publish the non-abstracted version too if you like, but I'd need to update it with some bug fixes this version introduces, and I won't bother with that if nobody's interested anyway.

**Addendum:** I haven't added the tests to the overall VFSForGit test runner yet, as that'll require a bit of digging for me to figure out how to do it, and if we throw out the PR based on its general approach, I'll have wasted my time. So I'll do that if and when this gets positive feedback.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
+++ b/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
@@ -1,11 +1,158 @@
+#include "FileSystemMountPointsPrivate.hpp"
 #include "FileSystemMountPoints.hpp"
 #include "kernel-header-wrappers/mount.h"
+#include "Locks.hpp"
+#include "KextLog.hpp"
+#include <kern/assert.h>
+
+#ifdef KEXT_UNIT_TESTING
+#include "FileSystemMountPointsTestable.hpp"
+#endif
+
+static Mutex s_mountPointsMutex;
+KEXT_STATIC MountPoint s_usedMountPoints[MaxUsedMountPoints];
+static bool s_reuseMountPointSlots;
+
+bool MountPoint_Init()
+{
+    s_reuseMountPointSlots = true;
+    s_mountPointsMutex = Mutex_Alloc();
+    if (!Mutex_IsValid(s_mountPointsMutex))
+    {
+        return false;
+    }
+    
+    return true;
+}
+
+void MountPoint_Cleanup()
+{
+    if (Mutex_IsValid(s_mountPointsMutex))
+    {
+        Mutex_FreeMemory(&s_mountPointsMutex);
+    }
+    
+    for (size_t i = 0; i < MaxUsedMountPoints; ++i)
+    {
+        // by the time this function is called, any provider user clients
+        // should have been terminated & deregistered
+        assert(s_usedMountPoints[i].authCacheDisableCount == 0);
+        assertf(
+            !s_reuseMountPointSlots || s_usedMountPoints[i].mountPoint == nullptr,
+            "If reusing slots is still allowed, they should all have been reset to the empty state by now; slot reuse allowed: %s, mount point registered in slot %lu: %p",
+            s_reuseMountPointSlots ? "yes" : "no",
+            i,
+            KextLog_Unslide(s_usedMountPoints[i].mountPoint));
+    }
+}
+
+KEXT_STATIC ssize_t FindMountPoint_Locked(mount_t mountPoint)
+{
+    for (size_t i = 0; i < MaxUsedMountPoints; ++i)
+    {
+        if (s_usedMountPoints[i].mountPoint == mountPoint)
+        {
+            return i;
+        }
+    }
+    
+    return -1;
+}
+
+KEXT_STATIC_INLINE ssize_t FindEmptyMountPointSlot_Locked()
+{
+    return FindMountPoint_Locked(nullptr);
+}
+
+KEXT_STATIC ssize_t TryFindOrInsertMountPoint_Locked(mount_t mountPoint)
+{
+    ssize_t mountPointIndex = FindMountPoint_Locked(mountPoint);
+    if (mountPointIndex < 0)
+    {
+        mountPointIndex = FindEmptyMountPointSlot_Locked();
+        if (mountPointIndex >= 0)
+        {
+            s_usedMountPoints[mountPointIndex].mountPoint = mountPoint;
+            assert(s_usedMountPoints[mountPointIndex].authCacheDisableCount == 0);
+        }
+    }
+    
+    return mountPointIndex;
+}
 
 void MountPoint_DisableAuthCache(mount_t mountPoint)
 {
-    vfs_setauthcache_ttl(mountPoint, 0);
+    Mutex_Acquire(s_mountPointsMutex);
+    {
+        ssize_t mountPointIndex = TryFindOrInsertMountPoint_Locked(mountPoint);
+        if (mountPointIndex < 0)
+        {
+            // No space in array, just disable auth cache and don't save previous state
+            KextLog_Info("MountPoint_DisableAuthCache: No space to save auth cache state for mount point '%s', restoring will not be possible.", vfs_statfs(mountPoint)->f_mntonname);
+            vfs_setauthcache_ttl(mountPoint, 0);
+            // No longer allow reuse of slots once all are used; allowing it would
+            // open the possibility of miscounting "restore" operations, thus
+            // restoring the auth cache TTL despite mount points being in use.
+            s_reuseMountPointSlots = false;
+        }
+        else
+        {
+            MountPoint* usedMountPoint = &s_usedMountPoints[mountPointIndex];
+            if (usedMountPoint->authCacheDisableCount == 0)
+            {
+                // New entry, save state and disable cache
+                usedMountPoint->savedAuthCacheTTL = vfs_authcache_ttl(mountPoint);
+                usedMountPoint->authCacheDisableCount = 1;
+                vfs_setauthcache_ttl(mountPoint, 0);
+            }
+            else
+            {
+                ++usedMountPoint->authCacheDisableCount;
+            }
+        }
+        
+    }
+    Mutex_Release(s_mountPointsMutex);
 }
 
 void MountPoint_RestoreAuthCache(mount_t mountPoint)
 {
+    Mutex_Acquire(s_mountPointsMutex);
+    {
+        ssize_t mountPointIndex = FindMountPoint_Locked(mountPoint);
+        if (mountPointIndex < 0)
+        {
+            KextLog_Info("MountPoint_RestoreAuthCache: previous auth cache state for mount point '%s' not found, can't restore.", vfs_statfs(mountPoint)->f_mntonname);
+        }
+        else
+        {
+            MountPoint* usedMountPoint = &s_usedMountPoints[mountPointIndex];
+            usedMountPoint->authCacheDisableCount--;
+            if (0 == usedMountPoint->authCacheDisableCount)
+            {
+                int savedTTL = usedMountPoint->savedAuthCacheTTL;
+                // vfs_clearauthcache_ttl(mount) has a slightly different effect in the kernel than
+                // vfs_setauthcache_ttl(mount, CACHED_RIGHT_INFINITE_TTL) so make sure to call the
+                // appropriate one when resetting to the original state.
+                if (savedTTL == CACHED_RIGHT_INFINITE_TTL)
+                {
+                    KextLog_Info("DecrementAuthCacheDisableCount: resetting mount point '%s' to default auth cache behaviour",
+                        vfs_statfs(mountPoint)->f_mntonname);
+                    vfs_clearauthcache_ttl(mountPoint);
+                }
+                else
+                {
+                    KextLog_Info("DecrementAuthCacheDisableCount: resetting mount point '%s' TTL to %d",
+                        vfs_statfs(mountPoint)->f_mntonname, savedTTL);
+                    vfs_setauthcache_ttl(mountPoint, usedMountPoint->savedAuthCacheTTL);
+                }
+                
+                if (s_reuseMountPointSlots)
+                {
+                    usedMountPoint->mountPoint = nullptr;
+                }
+            }
+        }
+    }
+    Mutex_Release(s_mountPointsMutex);
 }
--- a/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
+++ b/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
@@ -1,0 +1,11 @@
+#include "FileSystemMountPoints.hpp"
+#include "kernel-header-wrappers/mount.h"
+
+void MountPoint_DisableAuthCache(mount_t mountPoint)
+{
+    vfs_setauthcache_ttl(mountPoint, 0);
+}
+
+void MountPoint_RestoreAuthCache(mount_t mountPoint)
+{
+}
--- a/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.hpp
+++ b/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.hpp
@@ -2,5 +2,8 @@
 
 struct mount;
 
+bool MountPoint_Init();
+void MountPoint_Cleanup();
+
 void MountPoint_DisableAuthCache(mount* mountPoint);
 void MountPoint_RestoreAuthCache(mount* mountPoint);
--- a/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.hpp
+++ b/ProjFS.Mac/PrjFSKext/FileSystemMountPoints.hpp
@@ -1,0 +1,6 @@
+#pragma once
+
+struct mount;
+
+void MountPoint_DisableAuthCache(mount* mountPoint);
+void MountPoint_RestoreAuthCache(mount* mountPoint);
--- a/ProjFS.Mac/PrjFSKext/FileSystemMountPointsPrivate.hpp
+++ b/ProjFS.Mac/PrjFSKext/FileSystemMountPointsPrivate.hpp
@@ -1,0 +1,12 @@
+#pragma once
+
+#include <sys/kernel_types.h>
+
+struct MountPoint
+{
+    mount_t  mountPoint;
+    uint32_t authCacheDisableCount;
+    int      savedAuthCacheTTL;
+};
+
+static constexpr size_t MaxUsedMountPoints = 4;
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -14,6 +14,7 @@
 #include "kernel-header-wrappers/mount.h"
 #include "VnodeUtilities.hpp"
 #include "PerformanceTracing.hpp"
+#include "FileSystemMountPoints.hpp"
 
 #ifdef KEXT_UNIT_TESTING
 #include "VirtualizationRootsTestable.hpp"
@@ -386,6 +387,13 @@ KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProvid
     {
         assert(rootIndex < s_maxVirtualizationRoots);
         assert(!s_virtualizationRoots[rootIndex].inUse);
+        
+        // Retain a strong reference to the vnode if we have an active provider on it
+        if (userClient != nullptr && 0 != vnode_get(vnode))
+        {
+            return RootHandle_None;
+        }
+        
         VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
         
         root->providerUserClient = userClient;
@@ -475,7 +483,10 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
                             strlcpy(root.path, virtualizationRootCanonicalPath, sizeof(root.path));
                             KextLog_File(virtualizationRootVNode, "VirtualizationRoot_RegisterProviderForPath: registered provider (PID %d, IOUC %p) for virtualization root %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u.",
                                 clientPID, KextLog_Unslide(userClient), rootIndex, root.path, root.rootFsid.val[0], root.rootFsid.val[1], root.rootInode, KextLog_Unslide(virtualizationRootVNode), rootVid);
-                            virtualizationRootVNode = NULLVP; // transfer ownership
+
+                            // Acquire strong reference while provider is connected
+                            err = vnode_get(virtualizationRootVNode);
+                            assert(err == 0); // we already hold 1 strong reference from vnode_lookup so this should always succeed
                         }
                     }
                     else
@@ -484,14 +495,11 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
                         if (rootIndex >= 0)
                         {
                             assert(rootIndex < s_maxVirtualizationRoots);
-                        
-                            virtualizationRootVNode = NULLVP; // prevent vnode_put later; active provider should hold vnode reference
-                        
+                            
                             KextLog("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootCanonicalPath);
                         }
                         else
                         {
-                            // TODO: scan the array for roots on mounts which have disappeared, or grow the array
                             KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root");
                             err = ENOMEM;
                         }
@@ -507,17 +515,16 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
         }
     }
     
+    if (VirtualizationRoot_IsValidRootHandle(rootIndex))
+    {
+        MountPoint_DisableAuthCache(vnode_mount(virtualizationRootVNode));
+    }
+
     if (NULLVP != virtualizationRootVNode)
     {
         vnode_put(virtualizationRootVNode);
     }
-    
-    if (rootIndex >= 0)
-    {
-        VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
-        vfs_setauthcache_ttl(vnode_mount(root->rootVNode), 0);
-    }
-    
+
     vfs_context_rele(vfsContext);
     
     return VirtualizationRootResult { err, rootIndex };
@@ -535,20 +542,22 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex, PrjFSProvider
         assertf(userClient == root->providerUserClient, "ActiveProvider_Disconnect: disconnecting provider IOUC %p for root index %d (%s), but expecting IOUC %p",
             KextLog_Unslide(userClient), rootIndex, root->path, KextLog_Unslide(root->providerUserClient));
         
-        assert(NULLVP != root->rootVNode);
-        
         KextLog_File(root->rootVNode, "ActiveProvider_Disconnect: disconnecting provider (PID %d, IOUC %p) for virtualization root %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u.",
             root->providerPid, KextLog_Unslide(root->providerUserClient), rootIndex, root->path, root->rootFsid.val[0], root->rootFsid.val[1], root->rootInode, KextLog_Unslide(root->rootVNode), root->rootVNodeVid);
 
         
-        vnode_put(root->rootVNode);
+        vnode_t rootVNode = root->rootVNode;
+        assert(NULLVP != rootVNode);
+
         root->providerPid = 0;
-        
         root->providerUserClient = nullptr;
 
         RWLock_DropExclusiveToShared(s_virtualizationRootsLock);
 
         ProviderMessaging_AbortOutstandingEventsForProvider(rootIndex);
+        MountPoint_RestoreAuthCache(vnode_mount(rootVNode));
+
+        vnode_put(rootVNode);
     }
     RWLock_ReleaseShared(s_virtualizationRootsLock);
 }
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -30,13 +30,17 @@ KEXT_STATIC VirtualizationRoot* s_virtualizationRoots = nullptr;
 static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t vid, FsidInode fileId);
 
 static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHandle, vnode_t vnode, uint32_t vid, FsidInode fileId);
+static void GrowVirtualizationRootArrayWithMemory_Locked(VirtualizationRoot* newMemory, uint16_t newLength);
+static bool FsidsAreEqual(fsid_t a, fsid_t b);
+static const char* GetRelativePath(const char* fullPath, const char* root);
 
 // Looks up the vnode and fsid/inode pair among the known roots, and if not found,
 // detects if there is a hitherto-unknown root at vnode by checking attributes.
 static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t vnode, const FsidInode& vnodeFsidInode);
 
 static VirtualizationRootHandle FindUnusedIndex_Locked();
-KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path);
+
+KEXT_STATIC VirtualizationRootHandle FindOrInsertVirtualizationRoot_LockedMayUnlock(vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path);
 
 ActiveProviderProperties VirtualizationRoot_GetActiveProvider(VirtualizationRootHandle rootHandle)
 {
@@ -81,7 +85,8 @@ kern_return_t VirtualizationRoots_Init()
         return KERN_FAILURE;
     }
     
-    s_maxVirtualizationRoots = 128;
+    // Start with a small size so the resizing logic is regularly tested
+    s_maxVirtualizationRoots = 4;
     s_virtualizationRoots = Memory_AllocArray<VirtualizationRoot>(s_maxVirtualizationRoots);
     if (nullptr == s_virtualizationRoots)
     {
@@ -225,18 +230,10 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode, 
  
             RWLock_AcquireExclusive(s_virtualizationRootsLock);
             {
-                // Vnode may already have been inserted as a root in the interim
-                rootIndex = FindRootAtVnode_Locked(vnode, vid, vnodeFsidInode);
+                rootIndex = FindOrInsertVirtualizationRoot_LockedMayUnlock(vnode, vid, vnodeFsidInode, path);
                 
-                if (RootHandle_None == rootIndex)
-                {
-                    // Insert new offline root
-                    rootIndex = InsertVirtualizationRoot_Locked(nullptr, 0, vnode, vid, vnodeFsidInode, path);
-                    
-                    // TODO: error handling
-                    assert(rootIndex >= 0);
-                }
-
+                // TODO: error handling
+                assert(rootIndex >= 0);
             }
             RWLock_ReleaseExclusive(s_virtualizationRootsLock);
         }
@@ -270,42 +267,19 @@ static VirtualizationRootHandle FindUnusedIndex_Locked()
     return RootHandle_None;
 }
 
-static VirtualizationRootHandle FindUnusedIndexOrGrow_Locked()
+static void GrowVirtualizationRootArrayWithMemory_Locked(VirtualizationRoot* newMemory, uint16_t newLength)
 {
-    VirtualizationRootHandle rootIndex = FindUnusedIndex_Locked();
-    
-    if (RootHandle_None == rootIndex)
+    uint32_t oldSizeBytes = sizeof(s_virtualizationRoots[0]) * s_maxVirtualizationRoots;
+    memcpy(newMemory, s_virtualizationRoots, oldSizeBytes);
+    Memory_FreeArray(s_virtualizationRoots, s_maxVirtualizationRoots);
+    s_virtualizationRoots = newMemory;
+
+    for (uint16_t i = s_maxVirtualizationRoots; i < newLength; ++i)
     {
-        // No space, resize array
-        uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
-        if (newLength <= s_maxVirtualizationRoots)
-        {
-            assertf(newLength > 0, "s_maxVirtualizationRoot was likely not initialized");
-            // Already at max size, nothing to do.
-            return RootHandle_None;
-        }
-
-        VirtualizationRoot* grownArray = Memory_AllocArray<VirtualizationRoot>(newLength);
-        if (nullptr == grownArray)
-        {
-            return RootHandle_None;
-        }
-        
-        uint32_t oldSizeBytes = sizeof(s_virtualizationRoots[0]) * s_maxVirtualizationRoots;
-        memcpy(grownArray, s_virtualizationRoots, oldSizeBytes);
-        Memory_Free(s_virtualizationRoots, oldSizeBytes);
-        s_virtualizationRoots = grownArray;
-
-        for (uint16_t i = s_maxVirtualizationRoots; i < newLength; ++i)
-        {
-            s_virtualizationRoots[i] = VirtualizationRoot{ };
-        }
-        
-        rootIndex = s_maxVirtualizationRoots;
-        s_maxVirtualizationRoots = newLength;
+        s_virtualizationRoots[i] = VirtualizationRoot{ };
     }
-    
-    return rootIndex;
+
+    s_maxVirtualizationRoots = newLength;
 }
 
 static bool FsidsAreEqual(fsid_t a, fsid_t b)
@@ -378,42 +352,75 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
     rootEntry.rootVNodeVid = vid;
 }
 
-// Returns negative value if it failed, or inserted index on success
-KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path)
+KEXT_STATIC VirtualizationRootHandle FindOrInsertVirtualizationRoot_LockedMayUnlock(vnode_t virtualizationRootVNode, uint32_t rootVid, FsidInode persistentIds, const char* path)
 {
-    VirtualizationRootHandle rootIndex = FindUnusedIndexOrGrow_Locked();
-    
-    if (RootHandle_None != rootIndex)
+    VirtualizationRootHandle rootIndex;
+    do
     {
-        assert(rootIndex < s_maxVirtualizationRoots);
-        assert(!s_virtualizationRoots[rootIndex].inUse);
-        
-        // Retain a strong reference to the vnode if we have an active provider on it
-        if (userClient != nullptr && 0 != vnode_get(vnode))
+        rootIndex = FindRootAtVnode_Locked(virtualizationRootVNode, rootVid, persistentIds);
+        if (rootIndex >= 0)
         {
-            return RootHandle_None;
+            return rootIndex;
         }
         
-        VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
-        
-        root->providerUserClient = userClient;
-        root->providerPid = clientPID;
-        root->inUse = true;
-
-        root->rootVNode = vnode;
-        root->rootVNodeVid = vid;
-        KextLog_File(vnode, "InsertVirtualizationRoot_Locked: virtualization root inserted at index %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u, user client PID %d, IOUC %p.",
-            rootIndex, path, persistentIds.fsid.val[0], persistentIds.fsid.val[1], persistentIds.inode, KextLog_Unslide(vnode), vid, clientPID, KextLog_Unslide(userClient));
-        
-        root->rootFsid = persistentIds.fsid;
-        root->rootInode = persistentIds.inode;
-
-        if (path != nullptr)
+        rootIndex = FindUnusedIndex_Locked();
+        if (rootIndex < 0)
         {
-            strlcpy(root->path, path, sizeof(root->path));
+            // No space, resize array
+            uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
+            if (newLength <= s_maxVirtualizationRoots)
+            {
+                KextLog_Error("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array for root at '%s' failed: out of valid indices, array has reached %u items\n", path, s_maxVirtualizationRoots);
+                return RootHandle_None;
+            }
+        
+            VirtualizationRoot* grownArray = Memory_AllocArrayNoBlock<VirtualizationRoot>(newLength);
+            if (grownArray != nullptr)
+            {
+                KextLog_Info("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array for root at '%s' with non-blocking alloc succeeded\n", path);
+                rootIndex = s_maxVirtualizationRoots;
+            }
+            else
+            {
+                // Must drop lock to safely allocate memory with blocking alloc
+                RWLock_ReleaseExclusive(s_virtualizationRootsLock);
+                grownArray = Memory_AllocArray<VirtualizationRoot>(newLength);
+                KextLog_Info("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array to %u items for root at '%s' with non-blocking alloc failed, blocking alloc returned %p\n", newLength, path, grownArray);
+                RWLock_AcquireExclusive(s_virtualizationRootsLock);
+                
+                uint16_t newLengthAgain = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
+                if (newLengthAgain != newLength || newLength == s_maxVirtualizationRoots)
+                {
+                    KextLog_Info("FindOrInsertVirtualizationRoot_LockedMayUnlock: Beaten to the resize (newLength = %u, newLengthAgain = %u) by another thread, starting over.", newLength, newLengthAgain);
+                    // another thread already resized, start over.
+                    Memory_FreeArray(grownArray, newLength);
+                    continue;
+                }
+            }
+            
+            GrowVirtualizationRootArrayWithMemory_Locked(grownArray, newLength);
         }
-    }
+    } while (rootIndex < 0);
+
+    assert(rootIndex < s_maxVirtualizationRoots);
+    assert(!s_virtualizationRoots[rootIndex].inUse);
+
+    VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
     
+    root->inUse = true;
+
+    root->rootVNode = virtualizationRootVNode;
+    root->rootVNodeVid = rootVid;
+    root->rootFsid = persistentIds.fsid;
+    root->rootInode = persistentIds.inode;
+    if (path != nullptr)
+    {
+        strlcpy(root->path, path, sizeof(root->path));
+    }
+
+    KextLog_File(virtualizationRootVNode, "InsertVirtualizationRoot_Locked: virtualization root inserted at index %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u.",
+            rootIndex, path, persistentIds.fsid.val[0], persistentIds.fsid.val[1], persistentIds.inode, KextLog_Unslide(virtualizationRootVNode), rootVid);
+
     return rootIndex;
 }
 
@@ -462,13 +469,16 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
                 
                 RWLock_AcquireExclusive(s_virtualizationRootsLock);
                 {
-                    rootIndex = FindRootAtVnode_Locked(virtualizationRootVNode, rootVid, vnodeIds);
+                    rootIndex = FindOrInsertVirtualizationRoot_LockedMayUnlock(virtualizationRootVNode, rootVid, vnodeIds, virtualizationRootCanonicalPath);
+                    
                     if (rootIndex >= 0)
                     {
                         RefreshRootVnodeIfNecessary_Locked(rootIndex, virtualizationRootVNode, rootVid, vnodeIds);
                         
-                        // Reattaching to existing root
-                        if (nullptr != s_virtualizationRoots[rootIndex].providerUserClient)
+                        VirtualizationRoot& root = s_virtualizationRoots[rootIndex];
+                        assert(root.rootVNode == virtualizationRootVNode);
+
+                        if (nullptr != root.providerUserClient)
                         {
                             // Only one provider per root
                             err = EBUSY;
@@ -476,8 +486,6 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
                         }
                         else
                         {
-                            VirtualizationRoot& root = s_virtualizationRoots[rootIndex];
-                            assert(root.rootVNode == virtualizationRootVNode);
                             root.providerUserClient = userClient;
                             root.providerPid = clientPID;
                             strlcpy(root.path, virtualizationRootCanonicalPath, sizeof(root.path));
@@ -491,18 +499,8 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
                     }
                     else
                     {
-                        rootIndex = InsertVirtualizationRoot_Locked(userClient, clientPID, virtualizationRootVNode, rootVid, vnodeIds, virtualizationRootCanonicalPath);
-                        if (rootIndex >= 0)
-                        {
-                            assert(rootIndex < s_maxVirtualizationRoots);
-                            
-                            KextLog("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootCanonicalPath);
-                        }
-                        else
-                        {
-                            KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root");
-                            err = ENOMEM;
-                        }
+                        KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root '%s' for provider with PID %u", virtualizationRootPath, clientPID);
+                        err = ENOMEM;
                     }
                 }
                 RWLock_ReleaseExclusive(s_virtualizationRootsLock);
--- a/ProjFS.Mac/PrjFSKextTests/FileSystemMountPointTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/FileSystemMountPointTests.mm
@@ -1,0 +1,212 @@
+#include "../PrjFSKext/FileSystemMountPoints.hpp"
+#include "../PrjFSKext/FileSystemMountPointsTestable.hpp"
+#include "MockVnodeAndMount.hpp"
+#include "KextMockUtilities.hpp"
+#import <XCTest/XCTest.h>
+#include <array>
+
+using std::shared_ptr;
+using std::array;
+using std::generate;
+
+
+@interface FileSystemMountPointTests : XCTestCase
+
+@end
+
+@implementation FileSystemMountPointTests
+{
+    array<shared_ptr<mount>, MaxUsedMountPoints> dummyMounts;
+}
+
+- (void) fillAllMountPointSlots
+{
+    for (unsigned i = 0; i < MaxUsedMountPoints; ++i)
+    {
+        s_usedMountPoints[i].authCacheDisableCount = 1;
+        s_usedMountPoints[i].mountPoint = self->dummyMounts[i].get();
+    }
+}
+
+- (void) fillAllMountPointSlotsExcept:(unsigned)leaveEmptyIndex
+{
+    for (unsigned i = 0; i < MaxUsedMountPoints; ++i)
+    {
+        if (i != leaveEmptyIndex)
+        {
+            s_usedMountPoints[i].authCacheDisableCount = 1;
+            s_usedMountPoints[i].mountPoint = self->dummyMounts[i].get();
+        }
+    }
+}
+
+
+- (void)setUp
+{
+    srand(0);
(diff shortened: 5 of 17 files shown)


pmj:
I'm not quite clear on what's causing the Windows functional test failure, the exception stack doesn't look like anything my changes could have caused:

```
Unhandled Exception: System.InvalidOperationException: Stack empty.
   at System.Collections.Generic.Stack`1.ThrowForEmptyStack()
   at System.Collections.Generic.Stack`1.Pop()
   at NUnit.Framework.Internal.Execution.WorkItemQueue.Restore() in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Execution\WorkItemQueue.cs:line 369
   at NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.RestoreQueues() in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Execution\ParallelWorkItemDispatcher.cs:line 292
   at NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.OnEndOfShift(WorkShift endingShift) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Execution\ParallelWorkItemDispatcher.cs:line 328
   at NUnit.Framework.Internal.Execution.WorkShift.<StartWorkers>b__26_1(TestWorker s, WorkItem ea) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Execution\WorkShift.cs:line 181
   at NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc() in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Execution\TestWorker.cs:line 143
   at System.Threading.Thread.ThreadMain_ThreadStart()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()

Process completed with exit code -532462766 and had 1 error(s) written to the error stream.```

**Comment by sanoursa** [comment]
That test failure one is an intermittent failure caused by a bug in the NUnit framework. Just rerun the tests and they ought to pass.

File: ProjFS.Mac/PrjFSKext/AssociativeArrayTests/AssociativeArrayTests.mm [hunk 1]
```diff
@@ -0,0 +1,135 @@
+#import <XCTest/XCTest.h>
```
sanoursa: This file looks like it's Objective C perhaps? Can you give us some context on how this is all working?

pmj: Yeah, the `XCTest` framework is designed for Objective-C and Swift. You can mix Objective-C and C++ in the combined offshoot that is Objective-C++ - this is what the `.mm` suffix indicates, `.m` would be Objective-C. (c.f. `.c` for C, and `.cc` for C++)

All methods whose name begins with "test" in any subclasses of XCTestCase are run as tests, success and failure determined by the outcomes of any `XCTestAssert` macros. While entirely OK for testing C++ code via the detour to Objective-C++, there certainly remains a certain syntactic disconnect, and there are some oddities lurking in the assert macros. (such as use of `self`, which makes placing asserts in C++ functions a little trickier, for example in the [test `Memory_Free()`](https://github.com/Microsoft/VFSForGit/pull/500/files#diff-3971b7af86d2ce078fb9151e50d21928R48).)

If the Scheme is configured correctly, you can hit the "Test" button (click and hold "Run" in Xcode, then select "Test" from the dropdown) and the tests will run straight in Xcode, failing tests are shown in the warning/error UI, and success/failure is indicated on the source code line where each test is defined. `xcodebuild test` runs the tests from the command line.

As we've avoided bringing in any external dependencies so far, I thought I'd prototype it with XCTest and find out what you all think of the overall approach. (And whether we even want to start unit testing bits of the kext code outside the kext itself.) There's certainly no shortage of alternatives. I can see that [Visual Studio ships with 4 different unit test frameworks](https://docs.microsoft.com/en-us/visualstudio/test/writing-unit-tests-for-c-cpp?view=vs-2017) so that's probably a good starting point:

 * I can't seem to find a way to obtain the "Microsoft Unit Testing Framework for C++" independently from Visual Studio, so I'm guessing that might be a nonstarter on macOS.
 * CTest only really makes sense when using CMake for the build; we're not currently using that on VFS For Git on macOS, so it's probably not a good choice unless we also change our build system.
 * That leaves Google Test and Boost.Test; both work on macOS. The lowest-friction way to get them running is probably via homebrew or MacPorts, but that's probably not so great for the test farm. Google Test is very easy to fetch & build on a clean macOS system with only Xcode installed.

Anyway, this probably is all irrelevant if we don't go with the templated data structure approach, unit testing the intertwined data structure/business logic is probably going to be more trouble than it's worth.

nickgra: My opinion here is to keep using XCTest since we will only ever be using this code on MacOS. If it needed to be cross-plat, Google Test sounds like a good choice but I'd rather us use the native solution that our environment (Xcode) provides rather than add a new large dependency. 

We could easily call ```xcodebuild test``` or ```xctest -Test``` from our Build and UT script.

I'm a little biased though as I've used XCTest before and think it works well for unit testing for native MacOS code. I could be persuaded though :)

ProjFS.Mac/PrjFSKext/PrjFSKext/Memory.hpp
```diff
@@ -5,8 +5,18 @@ kern_return_t Memory_Init();
 kern_return_t Memory_Cleanup();
 
 void* Memory_Alloc(uint32_t size);
+// Variant which is safe to call while lock is held, but with low chance of success.
```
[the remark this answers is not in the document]
pmj: > Which lock?

Any lock that may impede progress of a thread that may be preventing the allocation to complete.

> Under what circumstances would allocating memory block?

If the allocator must itself acquire a lock in order to fulfil the request. So there may be a lock free fast path if there is a free block of memory in the suitably sized zone, but if it triggers an allocation of fresh pages in the zone, that probably requires a lock. That in itself wouldn’t necessarily be bad, but in low memory situations, this can lead to a deadlock as hinted at above. Essentially, if there are no free pages available, some memory may need to be pages out to disk. This causes I/O, and our callbacks are technically in the path of such I/O. In theory, the auth checks should already have passed by the time a memory mapped file gets flushed to disk, but I wouldn’t be surprised if there were specific circumstances where our callback is called as a result of a pageout. If that callback then attempts to acquire the lock we originally held while trying to allocate memory, you’ve got yourself a deadlock.

I’ve not seen this sequence of events happen in kauth vnode listeners, but I’d still be wary. I’ve definitely encountered this type of deadlock in block device driver kexts and MAC framework callbacks (these tend to be a little higher traffic than kauth) that weren’t careful about allocations. I think we’re a little more at risk of this than normal because we have this situation where the kext blocks progress until it hears back from user space.

To prevent this kind of situation entirely, you’d probably need to ensure you never have to allow on a critical path at all, but allocating without holding a lock should be safe enough - I’ve never seen a report where the problem still occurred in that case.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+{
+    bool   memoryWasUsed;
+    uint32_t releasedMemorySize;
+    void*  releasedMemory;
+};
+
+template <typename KEY, typename VALUE, bool EQUAL_FN(const KEY&, const VALUE&)>
+class AssociativeArray
+{
+    AssociativeArray(const AssociativeArray&) = delete;
```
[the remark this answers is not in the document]
wilbaker: > No, that’s only for classes derived from OSObject or one of its subclasses

Ok thanks! I missed that its use is specific to those cases.

ProjFS.Mac/PrjFSKext/PrjFSKext/Memory.hpp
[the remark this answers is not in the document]
wilbaker: @pmj thanks for all the details and helping me get up to speed on this.  A few more questions regarding the memory allocation:

> Any lock that may impede progress of a thread that may be preventing the allocation to complete.

When you say "any lock" do you mean any lock that is owned\managed by the allocator?  Or is it any lock used in kernel space (e.g. would our `s_virtualizationRootsLock` count because it uses kernel locking primitives)?

> This causes I/O, and our callbacks are technically in the path of such I/O. In theory, the auth checks should already have passed by the time a memory mapped file gets flushed to disk, but I wouldn’t be surprised if there were specific circumstances where our callback is called as a result of a pageout. If that callback then attempts to acquire the lock we originally held while trying to allocate memory, you’ve got yourself a deadlock.

Thanks I think I've got it.  Am I correct in summarizing the scenario as the following sequence of events?

1. The allocator acquired a lock
2. The allocator decided to page out to disk
3. Our kext was called (due to paging out to disk)
4. Our kext attempts to allocate memory
5. Deadlock

> To prevent this kind of situation entirely, you’d probably need to ensure you never have to allow on a critical path at all, but allocating without holding a lock should be safe enough - I’ve never seen a report where the problem still occurred in that case.

When you say "where the problem still occurred in that case", do you mean the case of allocating without holding a lock? (i.e. calling `OSMalloc_noblock`?) . Or was this referring to a different case?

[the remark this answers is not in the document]
pmj: > When you say "any lock" do you mean any lock that is owned\managed by the allocator? Or is it any lock used in kernel space (e.g. would our `s_virtualizationRootsLock` count because it uses kernel locking primitives)?

Yes, I do mean any lock, and `s_virtualizationRootsLock` counts.

> 1. The allocator acquired a lock
> 2. The allocator decided to page out to disk
> 3. Our kext was called (due to paging out to disk)
> 4. Our kext attempts to allocate memory
> 5. Deadlock

No, that's not quite it; the allocator should be (and as far as I'm aware, *is*) written to not to be vulnerable to quite such a severe weakness. I'm referring to:

1. Our kext acquires `s_virtualizationRootsLock` and finds it needs to resize the mount points array.
2. We allocate a new array while `s_virtualizationRootsLock` is exclusively locked.
3. The allocator doesn't have the memory available without blocking and needs to kick off some paging operations.
4. The same thread or another thread pages out some memory, causing a file write.
5. The file write causes a kauth callback into our kext.
6. Our kext needs to look up the file, so tries to acquire `s_virtualizationRootsLock`, which is already held by our original callback call.
7. Deadlock.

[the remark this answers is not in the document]
pmj: > When you say "where the problem still occurred in that case", do you mean the case of allocating without holding a lock? (i.e. calling `OSMalloc_noblock`?) . Or was this referring to a different case?

Not holding any lock while making a blocking allocation, or using `OSMalloc_noblock` to allocate while holding a lock. Note that the latter frequently fails, so you need a backup plan. This generally means you need to drop your lock, use blocking alloc, re-lock and retry your transaction.

ProjFS.Mac/PrjFSKext/PrjFSKext/KextCppUtilities.hpp
```diff
@@ -0,0 +1,25 @@
...
+    template <typename T> struct remove_reference<T&>
+    {
+        typedef T type;
+    };
+    template <typename T> struct remove_reference<T&&>
+    {
+        typedef T type;
+    };
+    
+    template <typename T> typename remove_reference<T>::type&& move(T&& o) __attribute__((visibility("hidden")));
```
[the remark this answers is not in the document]
pmj: It prevents symbols from being exported by the linker. Templates tend to generate a lot of symbols, junking up the symbol table. A better solution would probably be to set the kext linker to not export anything by default, and only export symbols explicitly named in a list. (As we don't have any kexts depending on ours, we can limit this to the kext start & stop functions, and any `OSObject` based classes which the IOKit reflection and vtable patching system needs to some extent.)

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -34,12 +35,22 @@ struct VirtualizationRoot
     char                        path[PrjFSMaxPath];
 };
 
+struct UsedMountPoint
```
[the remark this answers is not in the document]
pmj: Yeah, I left this one generic because I anticipate ending up using this data structure for other purposes. For example, instead of `VirtualizationRoot_VnodeIsOnAllowedFilesystem()`, we could early-out of kauth callbacks based on looking up `vnode_mount(currentVnode)` in this data structure. It would need to be something more efficient than an array in that case though.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+        }
+        lastItem->~VALUE();
+
+        --this->numberUsed;
+    }
+    
+    // Find item matching key; if not found, insert newItem, growing the array if necessary
+    // (during which the lock may be dropped and then re-acquired (exclusively))
+    // Returns the found or inserted value, or nullptr if resizing failed.
+    VALUE* FindOrInsertLockedAllowResize(KEY key, VALUE&& newItem, RWLock& lock)
```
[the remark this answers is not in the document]
pmj: This is a direct consequence of avoiding memory allocations while holding a lock. If this were not the kernel, we'd do something like:

 * Lock `s_virtualizationRootsLock` with exclusive access.
 * Call a `FindOrInsert()` function on the data structure.
 * Item not found, so insert.
 * No space left, so resize.
 * **Allocate some memory** to hold the resized array.
 * Move items across.
 * Perform the insertion.
 * Return from `FindOrInsert()`
 * Unlock

Note how this allocates memory while a lock is held. So in a kext, I'd implement it as this instead:

 * Lock `s_virtualizationRootsLock` with exclusive access.
 * Call a `FindOrInsert()` function on the data structure.
 * Item not found, so insert.
 * No space left, so attempt to resize.
 * **Try allocating some memory with non-blocking allocation** to hold the resized array<br>(*Assume for illustrative purposes the allocation fails; this doesn't always happen of course, but I'm showing the worst case.*)
 * Record how much memory we need for resizing.
 * Drop the lock.
 * **Allocate memory required for resizing using *blocking* alloc.**
 * Re-acquire the lock.
 * Search for the item again. (NB: Another thread may have inserted it since.)
 * Item not found, so try to insert. (NB: Another thread may have resized the data structure or removed items since we last checked for space.)
 * No space left, try to resize using the memory we allocated while not holding the lock.<br>(*This could fail if another thread already resized and then filled all the new space, so our allocated memory is no longer large enough; so you actually have to attempt this in a loop.*)
 * We have the right amount of memory, go ahead with resizing.
 * Move items across.
 * Perform the insertion.
 * Return from `FindOrInsert()`
 * Unlock

This gets extremely complicated if we only allow dropping the lock in the calling code; the calling code will then have to deal with the memory allocation while dropping the lock, explicitly perform the resize, then try inserting again, etc.

Sometimes you *have* to make the calling code do the leg work - this is the case if the insertion and another operation must be atomic, so the lock can't simply be dropped at will. Where that's not a problem though, passing the lock that needs dropping during allocations into the data structure mutation function is by far the most straightforward.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+    
+    uint32_t MemorySizeToGrow() const
+    {
+        return this->NextGrowCapacity() * sizeof(this->data[0]);
+    }
+    
+    static uint32_t MemorySizeForCapacity(uint32_t capacity)
+    {
+        return capacity * sizeof(VALUE);
```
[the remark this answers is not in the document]
pmj: `this` is not available in `MemorySizeForCapacity` as it's a `static` function, but I can change them both to `VALUE`.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+        for (size_t i = 0; i < this->numberUsed; ++i, memoryPosition += sizeof(VALUE))
+        {
+            VALUE* moved = new(memoryPosition) VALUE(KextCpp::move(this->data[i]));
+            (void)moved;
+            this->data[i].~VALUE();
+        }
+        
+        ResizeResult result = { true, MemorySizeForCapacity(this->capacity), this->data };
+        this->data = static_cast<VALUE*>(memory);
+        this->capacity = memorySize / sizeof(VALUE);
```
jeschu1: should we be able to just do
this->capacity = newCapacity;

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -535,3 +555,77 @@ const char* VirtualizationRoot_GetRootRelativePath(VirtualizationRootHandle root
...
+                    vfs_statfs(mountPoint)->f_mntonname, usedMountPoint->savedAuthCacheTTL);
+            }
+       
+            usedMountPoint->authCacheDisableCount++;
+        }
+    }
+    RWLock_ReleaseExclusive(s_virtualizationRootsLock);
+}
+
+static void DecrementAuthCacheDisableCount_Locked(mount_t mountPoint)
```
jeschu1: Is someone force kills a mount (i.e. no unmount), is there a way to detect and call this?

ProjFS.Mac/PrjFSKext/AssociativeArrayTests/AssociativeArrayTests.mm
```diff
@@ -0,0 +1,135 @@
...
+{
+    uint32_t size;
+    AllocationList::iterator listEntry;
+    alignas(16) uint8_t allocatedBytes[];
+};
+static_assert(offsetof(MemoryHeader, allocatedBytes) == sizeof(MemoryHeader), "Bytes should be at end of struct");
+
+static AllocationList s_allocationList;
+
+void* Memory_Alloc(uint32_t size)
```
wilbaker: What keeps the tests from finding the implementation of `Memory_Alloc` defined in Memory.cpp?  Is that source file not compiled as part of building `AssociativeArrayTests`?

ProjFS.Mac/PrjFSKext/AssociativeArrayTests/AssociativeArrayTests.mm
```diff
@@ -0,0 +1,135 @@
...
+    uint32_t value;
+};
+
+static bool TestArrayElementKeyEquals(const uint32_t& key, const TestArrayElement& item)
+{
+    return item.key == key;
+}
+
+struct MemoryHeader;
+typedef std::list<MemoryHeader*> AllocationList;
```
jeschu1: @pmj I really like the groundwork you've set for unit tests here!

Looking at AllocationList, it seems like we could verify how many memory blocks where allocated and in what sizes.  I.e. really have fun testing out any growing scenarios.  Is that correct?

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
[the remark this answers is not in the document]
wilbaker: > Where that's not a problem though, passing the lock that needs dropping during allocations into the data structure mutation function is by far the most straightforward.
> This should hopefully also answer your other questions about seemingly odd API design.

All that makes sense, and thanks again for the details.  While reviewing the changes with @jeschu1 this morning we were wondering about the exact scenario you call out here:

> Sometimes you have to make the calling code do the leg work - this is the case if the insertion and another operation must be atomic, so the lock can't simply be dropped at will. 

But it sounds like that's a complexity you (as a the author of the calling code) must be aware of and make sure that if you're using this function it is safe to have the lock dropped and reacquired.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
pmj: Yes, good point. Saves an integer divide. :-)

ProjFS.Mac/PrjFSKext/PrjFSKext/Memory.hpp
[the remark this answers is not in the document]
pmj: > I'm glad you called this out is it's something I'll need to account for in the vnode cache implementation.

Yep, definitely. Our virtualisation root array is currently also vulnerable to this. Like this one, it'll be resized extremely rarely, so the chances of the problem occurring are vanishingly small, but I'm planning to fix it. I was waiting with that until we have some generic data structures that solve the problem centrally (like this one) though.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+                return value;
+            }
+        
+            uint32_t memoryNeeded = this->MemorySizeToGrow();
+            void* reservedMemory = Memory_AllocNoBlock(memoryNeeded);
+            if (nullptr == reservedMemory)
+            {
+                RWLock_ReleaseExclusive(lock);
+                {
+                    reservedMemory = Memory_Alloc(memoryNeeded);
```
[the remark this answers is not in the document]
pmj: This should be clarified now, if not, let me know.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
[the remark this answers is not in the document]
pmj: > But it sounds like that's a complexity you (as a the author of the calling code) must be aware of and make sure that if you're using this function it is safe to have the lock dropped and reacquired.

Yes, I was on the fence on whether to add a rather clumsy `MayDropLock` suffix to the function name to be super explicit about it. Any better suggestions?

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+    {
+        return this->NextGrowCapacity() * sizeof(this->data[0]);
+    }
+    
+    static uint32_t MemorySizeForCapacity(uint32_t capacity)
+    {
+        return capacity * sizeof(VALUE);
+    }
+    
+    ResizeResult GrowUsingMemory(void* memory, uint32_t memorySize)
```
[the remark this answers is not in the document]
pmj: Does the deadlock avoidance/lock-dropping explanation clarify this aspect of the API design? Any alternative suggestions bearing in mind that constraint? (I deliberately decomposed it so that you can either call the kitchen sink find-insert-resize-possibly-dropping-lock function, or you can call these individual functions in situations where momentarily dropping the lock is not so simple.)

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -124,6 +138,8 @@ kern_return_t VirtualizationRoots_Init()
 
 kern_return_t VirtualizationRoots_Cleanup()
 {
+    s_usedMountPoints.ClearAndFreeMemory();
```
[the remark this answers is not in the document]
pmj: By this point, the vnode callbacks have been shut down, so no other threads will be accessing the data structure.

ProjFS.Mac/PrjFSKext/AssociativeArrayTests/AssociativeArrayTests.mm
pmj: Yes, definitely. This was very much a minimal approach illustrating what's possible, as there was a good chance it needed rewriting or would be thrown out.

ProjFS.Mac/PrjFSKext/AssociativeArrayTests/AssociativeArrayTests.mm
pmj: Yeah, that file isn't linked; neither is most of the kext code - most of the `#includes` will fail, so you need complicated ifdefs.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
```diff
@@ -0,0 +1,203 @@
...
+
+struct ResizeResult
+{
+    bool   memoryWasUsed;
+    uint32_t releasedMemorySize;
+    void*  releasedMemory;
+};
+
+template <typename KEY, typename VALUE, bool EQUAL_FN(const KEY&, const VALUE&)>
+class AssociativeArray
```
wilbaker: @pmj thanks again for the details and for bearing with me as I continue to learn the details of our kext and kext programming in general.

Now that I have a better understanding of the fix for the underlying issue (5177369) and the associative array implementation, I wanted to get your thoughts on a few approaches for simplifying the changes:

_Make the array a fixed size_ 
 
The number of `mount_t` mount points will always be <= the number of virtualization roots on disk, and we don't anticipate users having a lot (e.g. 30+) of virtualization roots.  If we picked a reasonable fixed size (128?) we could avoid the complexity (and failure modes) of resizing the array.

If there were more `mount_t`s than the kext could store it could always fall back to the current behavior of not resetting the TTL and log a warning.

**NOTE:** This suggestion assumes that virtualization roots created in the same partition would share a common `mount_t`, and so our functional tests creating a bunch of test roots in the same folder (`/GVFS.FT`) would not result in a bunch of different `mount_t` entries in the array. 

_Make the array specific to `UsedMountPoint`_

It looks like there's a fair bit of complexity in the implementation of `AssociativeArray` that comes from making it generic.  Did you have other uses in mind for `AssociativeArray`?  If we don't have any other uses for it in the near-term it could be renamed to `MountPointArray` and simplified so that it only works with `UsedMountPoint`s.

(In another comment I saw you mentioned wanting to improve our handling of the virtualization root array, but it wasn't clear to me if you had `AssociativeArray` in mind for that task.)

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: That's already covered; in this situation, the [`IOUserClient::clientDied()` function is called, which calls `IOUserClient::clientClose()`](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/iokit/Kernel/IOUserClient.cpp#L1555). Our provider user client class [overrides that function](https://github.com/Microsoft/VFSForGit/blob/0459b281d2ec259f0749157e2924bb0ec0580f37/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp#L95), which eventually calls down to the auth cache decrement via `ActiveProvider_Disconnect()`.

ProjFS.Mac/PrjFSKext/AssociativeArrayTests/AssociativeArrayTests.mm
pmj: When I get a chance, I'll see how easy it is to wrap the most awkward of Objective-C bits in something more palatable to C++ developers. The reliance on `self` is my main grievance with XCTest, which is what necessitates the contortions for carrying around the test object:

```
static AssociativeArrayTests* s_test;
…
void Memory_Free(void* memory, uint32_t size)
{
    AssociativeArrayTests* self = s_test;
```

There's got to be a better way to do that in C++, but there are more pressing concerns at the moment. :-)

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
pmj: I'll address the suggestions in reverse order as the reasoning in my response makes more sense that way:

> _Make the array a fixed size_
> 
> The number of `mount_t` mount points will always be <= the number of virtualization roots on disk,

It's actually the number of virtualisation roots with active providers, which is <= the number on disk; and the number of mount points will usually be much smaller still.

> and we don't anticipate users having a lot (e.g. 30+) of virtualization roots. If we picked a reasonable fixed size (128?) we could avoid the complexity (and failure modes) of resizing the array.

It's definitely an option, although I feel the complexity is now fairly well contained. We initially had a fixed size array for the virtualisation roots which caused us a lot of problems in the functional tests; of course, for that data structure, the failure to grow is much more serious. As you say, we can just not reset the TTL, which is in fact what this PR does if for some reason the resize fails altogether.

My second argument against this is also my response to the other suggestion:

pmj: > _Make the array specific to `UsedMountPoint`_
> 
> It looks like there's a fair bit of complexity in the implementation of `AssociativeArray` that comes from making it generic.

Looking over the code, I actually can't find that many instances of code that's not necessary for `UsedMountPoint` - that struct doesn't have constructors & destructors, so invoking them could be removed, which simplifies the resizing code ever so slightly. I'm counting about 9 LOC from about 200, so 5%. I'm not sure that's worth the saving? Granted, we can also remove the move semantics helper code (KextCppUtilities.hpp) but I can remove that anyway if preferred, while still keeping the data structure generic.

> Did you have other uses in mind for `AssociativeArray`? If we don't have any other uses for it in the near-term it could be renamed to `MountPointArray` and simplified so that it only works with `UsedMountPoint`s.
> 
> (In another comment I saw you mentioned wanting to improve our handling of the virtualization root array, but it wasn't clear to me if you had `AssociativeArray` in mind for that task.)

Right, that one isn't really an associative array - the virtualisation root handles are just indices, and while we've got some linear searches that really should be replaced with something more efficient (`FindRootAtVnode_Locked()` & `FindUnusedIndex_Locked()`) those are not the primary lookup means; the way to speed them up would be via an ancillary data structure. (e.g. vnode -> root handle hash table for active providers, fsid/inode -> root handle for offline roots)

However, taking on your earlier idea of passing a lambda to the `Find` function, we could isolate the resizing and linear search logic in a simple generic `Vector` class; the associative array could be based on top of this, while the virtualisation root array uses the vector directly. Given that we definitely want the virtualisation root array to be resizable, this would concentrate the resizing logic for both arrays into one generic implementation, with the instantiation for the roots array getting more testing, but as it's the same code, we can be more confident about resizes of the mount point array too.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
jeschu1: Thanks for that explanation makes sense!

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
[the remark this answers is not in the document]
wilbaker: @pmj, after revisiting this code (and reading the update on your other comment), I'm now wondering if we need the complexity of dropping the lock when increasing the size of `s_usedMountPoints`.

Could we instead add a new `s_usedMountPointsLock` and use that when modifying `s_usedMountPoints`?

Because `s_usedMountPoints` is not accessed during any of the kext callbacks, I'm wondering if using a second lock would allow us to avoid deadlocks without the complexity of dropping\re-acquiring the lock.  

We would not need to worry about the allocator paging out memory because if it does (and our kext is called) none of those code paths would acquire `s_usedMountPointsLock`.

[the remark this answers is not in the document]
pmj: Yes, this is a good point, we can completely isolate that aspect, as we're not attempting to make the auth cache disabling atomic with regard to provider registration. In any case, we could always increment *before* registering the provider, and decrement in case it fails. And yes, it *should* be safe to alloc memory while holding that lock, although this sort of contradicts the philosophy you floated yesterday, that you were thinking we shouldn't have generic data structures as it would make memory allocations too easy for new devs. In the same vein, I would say that allocating memory while holding this newly introduced lock would possibly make it too easy for new code to be introduced where the lock is acquired in a context where those allocations *could* cause a deadlock.

So I think adding and using the new lock is definitely a good idea simply because we *can* do it, but if we're going with dynamic allocations, I'd still vote against allocating while the lock is held.

ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp
pmj: > All of this might be moot, however, if we can avoid the complexity of having to drop\reacquire the lock completely (see other comment just posted to ProjFS.Mac/PrjFSKext/PrjFSKext/AssociativeArray.hpp).

The new lock won't help with reallocating the virtualisation root array though.

pmj:
Given the relative lack of urgency on this issue and the apparently sizeable amount of work involved in getting this PR to a point where we'll be merging it, I'm putting the effort on hold until I've run out of higher priority tasks. For future reference, I think I have the following still to do:

 * Move the auth cache disabling code to a separate code file, using a new mutex to protect the data structure, as there's no need to be further junking up VirtualizationRoots.cpp.
 * There's not a clear consensus on the final implementation other than that the current PR isn't ideal; we appear to have the following options:
   1. Fixed size array of mount points. If a user has providers on more than this fixed number, those mount points that don't fit will simply not have their auth cache re-enabled on provider disconnect.<br>👍: substantially less code<br>👎: inelegant, error handling code paths will rarely be exercised, doesn't solve `s_virtualizationRoots` realloc
   2. Non-generic resizable data structure (with or without lock held while allocating)<br>👍: code potentially easier to read than generic option, no rare edge cases<br>👎: if we need something similar in future, we'll need to copy-paste or refactor into a generic version after all; potentially harder to isolate for unit testing; doesn't solve `s_virtualizationRoots` realloc
   3. Generic resizable vector, to be used directly for `s_virtualizationRoots` and via an associative array adapter for the auth cache mount points.<br>👍: widely used generic code gets more active coverage, easy to unit test; `s_virtualizationRoots` realloc solved.<br>👎: harder to read, possibly over-engineered
   4. If vnode cache is implemented as a generic hash table, we can re-use it for the auth cache mount points.<br>👍: widely used generic code gets more active coverage, easy to unit test, more elegant than linear search<br>👎: possibly over-engineered, doesn't solve `s_virtualizationRoots` realloc
 * The unit tests using XCTest are awkward for our purposes due to various Objective-C quirks, so it would be nicer to have some kind of C++ adapter for these.

pmj:
This latest push is a fresh attempt at this task. For now, this implements the mount point reset with a fixed-size array, in a separate code file with an independent mutex. This corresponds to option (i) in my previous comment. I've kept the array deliberately small (4 volumes) so that the edge case can easily be tested. We can add a functional test that creates virtualization roots in a bunch of disk images for this, assuming we don't want to move to a more general solution for the resizeable variant.

I also noticed a potential race condition with regard to obtaining the mount point from the vnode, so there's a fix for that now too.

Additionally, I've changed the `s_virtualizationRoots` array resizing so that the lock is dropped during the blocking memory allocation. As expected, this necessitated a change to the structure of the resizing code. (Incidentally, I made a bunch of mistakes in my first few attempts at this, including one I didn't catch by reviewing my code, but by testing; it's easy to get wrong, which is another reason why I still prefer the generic approach.)

I've done a force push with fresh commits, the old ones are preserved at https://github.com/pmj/VFSForGit/tree/reset-mount-point-auth-cache-old for reference. The commits are once again self-contained, so for review I once again recommend tackling them individually.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
```diff
@@ -0,0 +1,134 @@
...
+            KextLog_Info("MountPoint_RestoreAuthCache: previous auth cache state for mount point '%s' not found, can't restore.", vfs_statfs(mountPoint)->f_mntonname);
+        }
+        else
+        {
+            MountPoint* usedMountPoint = &s_usedMountPoints[mountPointIndex];
+            usedMountPoint->authCacheDisableCount--;
+            if (0 == usedMountPoint->authCacheDisableCount)
+            {
+                int savedTTL = usedMountPoint->savedAuthCacheTTL;
+                if (savedTTL == CACHED_RIGHT_INFINITE_TTL)
```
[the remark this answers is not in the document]
pmj: Calling [`vfs_clearauthcache_ttl(mountPoint)`](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/kpi_vfs.c#L560) vs [`vfs_setauthcache_ttl(mountPoint, CACHED_RIGHT_INFINITE_TTL)`](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/kpi_vfs.c#L551) has subtly different behaviour in the kernel with regard to the `mnt_kern_flag` bitfield on mountpoints.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+void MountPoint_DisableAuthCache(mount_t mountPoint)
+{
+    Mutex_Acquire(s_mountPointsMutex);
+    {
+        ssize_t mountPointIndex = FindMountPoint(mountPoint);
+        if (mountPointIndex < 0)
+        {
+            mountPointIndex = FindMountPoint(nullptr);
```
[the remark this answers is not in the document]
pmj: 1. Check if there's already a slot for this mount point.
2. If not, find an empty slot.
3. If that succeeds, fill the slot and use it, otherwise follow the full array logic.

ProjFS.Mac/PrjFSKext/PrjFSKext.exp
```diff
@@ -0,0 +1,18 @@
+# Symbols that should be exported (public) by the linker. * is wildcard.
```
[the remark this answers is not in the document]
pmj: I guess now that we haven't got any templated code left in this PR, it's a weird PR to include it in. Still not a bad idea for general symbol hygiene. I had it in this PR because the original code contained templates, and template instances generate tons of symbols which junk up the symbol tables.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -105,7 +105,7 @@ kern_return_t VirtualizationRoots_Init()
         return KERN_FAILURE;
     }
     
-    s_maxVirtualizationRoots = 128;
+    s_maxVirtualizationRoots = 4;
```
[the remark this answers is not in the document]
pmj: Making sure the various different code paths are regularly exercised now that complexity has increased a bit.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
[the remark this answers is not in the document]
nickgra: The `nullptr` mount point was the non-intuitive part to me at my first pass through this. Maybe a comment here would make this more intuitive?

ProjFS.Mac/PrjFSKext/PrjFSKext.exp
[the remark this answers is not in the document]
nickgra: Got it, thanks! I think I would prefer to remove the exported symbols for now (at least as part of this PR) until we have templates that will bloat our symbols. I don't feel super strongly about this however.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
```diff
@@ -1,11 +1,139 @@
...
+            return i;
+        }
+    }
+    
+    return -1;
+}
+
+static inline ssize_t FindEmptyMountPointSlot()
+{
+    return FindMountPoint(nullptr);
```
pmj: I've now turned this into a hopefully self-documenting helper/wrapper function.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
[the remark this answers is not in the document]
pmj: Comment now added

ProjFS.Mac/PrjFSKextTests/FileSystemMountPointTests.mm
```diff
@@ -0,0 +1,106 @@
...
+{
+    memset(s_usedMountPoints, 0, sizeof(s_usedMountPoints));
+}
+
+- (void)tearDown
+{
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+}
+
+- (void)testFindEmptyMountPointSlot_AllFull
```
pmj: I've added a few unit tests for the new code with this latest push. Coverage is currently not good, but I wanted to get something off the ground at least.

* For testing `MountPoint_Init()`/`MountPoint_Cleanup()` we need to implement testing versions of the `Mutex_` functions which I haven't got around to yet.
 * `MountPoint_DisableAuthCache` and `MountPoint_RestoreAuthCache` call KPIs, so I'll wait until we know what we're doing regarding mocking KPIs.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
```diff
@@ -0,0 +1,131 @@
...
+        {
+            mountPointIndex = FindEmptyMountPointSlot();
+            if (mountPointIndex >= 0)
+            {
+                s_usedMountPoints[mountPointIndex].mountPoint = mountPoint;
+                assert(s_usedMountPoints[mountPointIndex].authCacheDisableCount == 0);
+            }
+        }
+        
+        if (mountPointIndex < 0)
```
jeschu1: This would read easier for me if was added as an else for the above
`if (mountPointIndex < 0)`
followed by a return

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
```diff
@@ -0,0 +1,131 @@
...
+    for (size_t i = 0; i < MaxUsedMountPoints; ++i)
+    {
+        // by the time this function is called, any provider user clients
+        // should have been terminated & deregistered
+        assert(s_usedMountPoints[i].mountPoint == nullptr);
+        assert(s_usedMountPoints[i].authCacheDisableCount == 0);
+    }
+}
+
+KEXT_STATIC ssize_t FindMountPoint(mount_t mountPoint)
```
jeschu1: Obviously we can only call this under lock.  Is there any way to enforce that?

ProjFS.Mac/PrjFSKext/FileSystemMountPointsPrivate.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
jeschu1: Can you elaborate on the file name convention?  Previously we've used "Testable" to facilitate testing.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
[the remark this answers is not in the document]
jeschu1: To clarify we wouldn't want to leave this at 4 full stop right?  Was it just for a functional test run?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -330,27 +302,69 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
...
+            VirtualizationRoot* grownArray = Memory_AllocArrayNoBlock<VirtualizationRoot>(newLength);
+            if (grownArray != nullptr)
+            {
+                KextLog_Info("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array for root at '%s' with non-blocking alloc succeeded\n", path);
+                rootIndex = s_maxVirtualizationRoots;
+            }
+            else
+            {
+                // Must drop lock to safely allocate memory with blocking alloc
+                RWLock_ReleaseExclusive(s_virtualizationRootsLock);
```
jeschu1: I believe I understand why we need to release the lock, but it seems risky.  Could it cause us to get bad data?

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
pmj: Not conventionally, but I've clearly forgotten to apply the naming convention of suffixing with `Locked` which I myself introduced to other parts of the project. Oops.

Re "not conventionally": [Rust's mutex implementation](https://doc.rust-lang.org/1.0.0/book/concurrency.html#safe-shared-mutable-state) is interesting in this regard - the mutex is a generic, templated on the type of the object it is protecting. (or an aggregate of the multiple objects it is protecting) Acquiring the mutex returns a reference to the protected object for the duration of the held lock. Rust's type system (a limited version of linear typing) is powerful enough to prevent you from holding onto that reference outside of the locked block.

You could replicate that same programming pattern in C++ by turning our `Mutex` struct into a template and making lock acquisition scoped:

```
struct UsedMountPoints
{
    MountPoint mountPoints[MaxUsedMountPoints];
};

static Mutex<UsedMountPoints> s_mountPointsMutex;

// …

ssize_t FindMountPoint(MutexGuard<UsedMountPoints>& lockedMountPoints, mount_t mountPoint)
{
  for (size_t i = 0; i < MaxUsedMountPoints; ++i)
  {
    if (lockedMountPoints->usedMountPoints[i].mountPoint == mountPoint)
    {
      return i;
    }
  }

  return -1;
}

void MountPoint_DisableAuthCache(mount_t mountPoint)
{
  {
    MutexGuard<UsedMountPoints> lockedMountPoints(s_mountPointsMutex);
    ssize_t mountPointIndex = FindMountPoint(lockedMountPoints, mountPoint);
  }
} 
```

`FindMountPoint` can now only be called by a lock holder as it expects a `MutexGuard` reference.

In the expression `lockedMountPoints->usedMountPoints[i]` you can see that I'm implementing `operator->` on the `MutexGuard` to return `UsedMountPoints*` which can safely be mutated because the guard is holding the lock.

Or you could perhaps use some slightly different notation.

You could still break out of this by explicitly calling `operator->` and saving that pointer beyond where the guard goes out of scope, but this should in theory be more a much more obvious transgression than accidentally accessing an informally mutex-protected variable outside of a lock.

Whether or not we want to apply this sort of pattern to the kext code is another question of course!

ProjFS.Mac/PrjFSKext/FileSystemMountPointsPrivate.hpp
pmj: The problem I had here is that I needed to access the contents of `struct MountPoint` in the tests in order to prepare the test case. `MountPoint` should however be opaque to the rest of the kext - this is why it was originally defined in `FileSystemMountPoints.cpp`, not the header file which is included by `VirtualizationRoots.cpp` as well.

We explicitly did *not* want the kext code to depend on anything defined in the "…Testable.hpp" headers, i.e. the tail wagging the dog. So if not the main `FileSystemMountPoints.hpp` header, and not `FileSystemMountPointsTestable.hpp`, then there must be some new header that declares and defines things that are private to the implementation but also accessible to tests. Hence `FileSystemMountPointsPrivate.hpp`.

Right solution? Discuss!

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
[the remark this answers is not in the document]
pmj: No, I'm fully intending for this to stay in so that any future breakage is found before it ships to users who really do have more than 128 repos.

The resizing is so rare and has such a minuscule performance impact as to effectively be irrelevant, so I think picking a "big enough for everyone*" value here is explicitly harmful, and we *should* choose an artificially small size.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: I've satisfied myself that this code is not accessing any shared state protected by the lock outside of that lock being held. Beyond that, I've checked that the callers of `FindOrInsertVirtualizationRoot_LockedMayUnlock()` don't make any assumptions about the state before the call being consistent with state after the call - i.e. unlocking temporarily would not cause any inconsistencies. I gave the function the awkward name suffix so that developers (including my forgetful self) are made painfully aware that such assumptions must not be made in future changes.

Have I got it right? I don't know, please do try to poke holes in my code!

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
pmj: Do you mean:

        if (mountPointIndex < 0)
        {
            mountPointIndex = FindEmptyMountPointSlot();
            if (mountPointIndex >= 0)
            {
                s_usedMountPoints[mountPointIndex].mountPoint = mountPoint;
                assert(s_usedMountPoints[mountPointIndex].authCacheDisableCount == 0);
            }
            else
            {
               // No space in array, just disable auth cache and don't save previous state
               KextLog_Info("MountPoint_DisableAuthCache: No space to save auth cache state for mount point '%s', restoring will not be possible.", vfs_statfs(mountPoint)->f_mntonname);
               vfs_setauthcache_ttl(mountPoint, 0);
               Mutex_Release(s_mountPointsMutex); // <--- I don't really like this here
               return;
            }
        }

        MountPoint* usedMountPoint = &s_usedMountPoints[mountPointIndex];
        …

? I'm not a fan of breaking up the mutex release like that. Another option would be a `goto` (again, not a fan), or wrapping the whole function apart from the mutex lock/unlock in `do … while(false);` so that we're allowed to put a `break;` statement after the `vfs_setauthcache_ttl(mountPoint, 0);` call above.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Incidentally, not having to keep writing/reviewing complicated code like this every time we need to resize an array is why I do still like the generic approach.

jeschu1: This all sounds good @pmj thank you!  I should be back full time on this tomorrow and will go thru the comments more.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
pmj: I've just pushed a new version of the code that moves this chunk of functionality into a new `TryFindOrInsertMountPoint_Locked()` helper function. This also allows us to easily unit test this code, so there are now a few tests too.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
pmj: The latest push suffixes the function names with `_Locked` as appropriate.

ProjFS.Mac/PrjFSKext/FileSystemMountPointsPrivate.hpp
wilbaker: > So if not the main `FileSystemMountPoints.hpp` header, and not `FileSystemMountPointsTestable.hpp`, then there must be some new header that declares and defines things that are private to the implementation but also accessible to tests. Hence `FileSystemMountPointsPrivate.hpp`.

> Right solution yes/no? Discuss!

When we ran into this issue writing unit tests on Windows that needed access to private ProjFS structs the approach we took was to duplicate the struct somewhere that the test could find it.

@pmj @jeschu1 what do you think of taking the same approach here?  It obviously comes at the cost of maintaining the same struct in two spots, but it comes at the benefit of ensuring the `MountPoint` internals don't leak out to the rest of the kext (and reduces how much we have to contort the code to make it testable).

I also expect that if the `MountPoint` struct is getting updated the unit tests should/will need to get updated anyway, so (in my opinion) it doesn't seem like too heavy of a burden to have a second declaration of the struct for the unit tests.

If we *really* wanted to get fancy we could have a pre-process step that auto-generates a header with `MountPoint` for the unit tests to consume (based on metadata we include in `FileSystemMountPoints.cpp`), but unless maintaining two copies of the struct becomes too much of a hassle that feels like overkill to me.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -339,31 +311,72 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
...
-
-        root->rootVNode = vnode;
-        root->rootVNodeVid = vid;
-        KextLog_File(vnode, "InsertVirtualizationRoot_Locked: virtualization root inserted at index %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u, user client PID %d, IOUC %p.",
-            rootIndex, path, persistentIds.fsid.val[0], persistentIds.fsid.val[1], persistentIds.inode, KextLog_Unslide(vnode), vid, clientPID, KextLog_Unslide(userClient));
+        rootIndex = FindUnusedIndex_Locked();
+        if (rootIndex < 0)
+        {
+            // No space, resize array
+            uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
```
wilbaker: I don't think we'll ever get this high, but for my own understanding do you remember why `INT16_MAX + 1` selected?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: >Have I got it right? I don't know, please do try to poke holes in my code! The main sensitive edge case is where 1 thread decides the data structure needs resizing, and drops the lock to alloc some memory. Meanwhile, another thread jumps on the lock during this time, _also_ decides the array needs to grow, so _also_ drops the lock and allocs some memory. Now we have 2 chunks of memory for growing the array. The condition a few lines down is what's supposed to catch and untangle that situation.

There's only one other scenario that's coming to mind, and it seems extremely unlikely.  But @pmj I want to throw it out there for you to confirm that it's unlikely:

1. The no-block alloc fails
2. We release `s_virtualizationRootsLock` and call the blocking alloc
3. The allocator kicks off paging operations
4. The paging operations result in a write that calls into our kext
5. In handling the write our kext goes down a path that requires a memory allocation
6. The allocator is called ...

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -339,31 +311,72 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
...
+            {
+                KextLog_Error("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array for root at '%s' failed: out of valid indices, array has reached %u items\n", path, s_maxVirtualizationRoots);
+                return RootHandle_None;
+            }
         
-        root->rootFsid = persistentIds.fsid;
-        root->rootInode = persistentIds.inode;
-        strlcpy(root->path, path, sizeof(root->path));
-    }
+            VirtualizationRoot* grownArray = Memory_AllocArrayNoBlock<VirtualizationRoot>(newLength);
```
wilbaker: Minor: The switch to to use `Memory_AllocArrayNoBlock` when resizing the VirtualizationRoot array is a bit unrelated to the current PR title.  Could you update the title?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -412,43 +425,35 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
...
-                            virtualizationRootVNode = NULLVP; // prevent vnode_put later; active provider should hold vnode reference
-                        
-                            KextLog("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootCanonicalPath);
-                        }
-                        else
-                        {
-                            // TODO: scan the array for roots on mounts which have disappeared, or grow the array
-                            KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root");
-                        }
+                        KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root");
```
wilbaker: Unrelated to your changes for this PR, could you update this log message to include `virtualizationRootPath`?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: `VirtualizationRootHandle` is typedef'd to an `int16_t`,  so we can have roots `0`…`INT16_MAX`.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: > * In handling the write our kext goes down a path that requires a memory allocation

This implies that there's an open descriptor to a file in a virtualisation root that was previously unknown to the kext. The only way this can happen is if the kext has only just been loaded.

> * The allocator is called ...

This won't actually lead to deadlock; the system will acknowledge the need to free up more physical memory, and do some more paging, etc. Eventually, something will need to make progress, because not every single open file can be on a previously-undiscovered virtualisation root.

ProjFS.Mac/PrjFSKext/FileSystemMountPointsPrivate.hpp
pmj: > When we ran into this issue writing unit functional tests on Windows that needed access to private ProjFS structs the approach we took was to duplicate the struct somewhere that the test could find it.
> 
> @pmj @jeschu1 what do you think of taking the same approach here? It obviously comes at the cost of maintaining the same struct in two spots, but it comes at the benefit of ensuring the `MountPoint` internals don't leak out to the rest of the kext (and reduces how much we have to contort the code to make it testable).

I'm not a fan of duplicating struct definitions at all. It's fragile and leads to weird issues that can lead to time wasted head-scratching even if the root cause is discovered before any code gets committed/pushed.

If you're that worried that someone might accidentally `#include` a "forbidden" `…Private.hpp` file, we could add a build step that uses `grep` or whatever to search for them and fails the build.

> One other point, I don't think the function declarations (e.g. `KEXT_STATIC ssize_t FindMountPoint_Locked(mount_t mountPoint);`) should be in a "Private" file, those should be able to stay in a "Testable" file.

Yes, we can move those to stay 100% consistent.

wilbaker: Closing the loop on this, as we discussed offline we can stick with the "Private" file approach to avoid duplication.  Thanks!

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+struct mount;
```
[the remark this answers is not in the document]
pmj: I'm not sure that's really necessary. What else would a forward declaration be for?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -339,31 +311,72 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
     return RootHandle_None;
 }
 
-// Returns negative value if it failed, or inserted index on success
```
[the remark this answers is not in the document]
pmj: I guess I didn't think it was really necessary anymore as all of the code consistently uses index handles and the "special" handle values for various error or other special cases. The comment is probably from before we had the `VirtualizationRootHandle` typedef.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -261,40 +254,19 @@ static VirtualizationRootHandle FindUnusedIndex_Locked()
     return RootHandle_None;
 }
 
-static VirtualizationRootHandle FindUnusedIndexOrGrow_Locked()
+static void GrowVirtualizationRootArrayWithMemory_Locked(VirtualizationRoot* newMemory, uint16_t newLength)
```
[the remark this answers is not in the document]
pmj: There were a few more prototypes for `static` functions missing in this file, I've gone through all of them and added them too.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Done.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.cpp
```diff
@@ -0,0 +1,137 @@
...
+                    vfs_clearauthcache_ttl(mountPoint);
+                }
+                else
+                {
+                    KextLog_Info("DecrementAuthCacheDisableCount: resetting mount point '%s' TTL to %d",
+                        vfs_statfs(mountPoint)->f_mntonname, savedTTL);
+                    vfs_setauthcache_ttl(mountPoint, usedMountPoint->savedAuthCacheTTL);
+                }
+                
+                usedMountPoint->mountPoint = nullptr;
```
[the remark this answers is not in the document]
pmj: Ouch, this is a great catch, thanks for that! I guess this is another edge case introduced by the "simplification" of a static sized array. I'm still not sure that is the right decision, but I'll go with it for this PR, it's an improvement over not resetting the cache at all.

I've worked around the reuse issue, and added a comprehensive test case that detects the problem if you allow slot reuse after filling the array. That test also contains a bunch of commented-out `MockCalls::DidCallFunction(` asserts *with* argument value checks, a feature which is not available on the master branch as of this writing. We can uncomment those lines once #952 is merged.

pmj:
The master branch has moved on quite a bit since the last round of reviews on this PR, so I've gone through and rebased the changes, incorporated the previously unaddressed feedback, and then updated the test code to make use of some of the new facilities we've gained in the interim.

Functionally, the only change from last time is the workaround for the [reuse issue that @wilbaker found](https://github.com/Microsoft/VFSForGit/pull/500#discussion_r260494625) but as I said, the contextual code has changed somewhat so it would probably not be a bad idea to review the latest iteration in some depth in case I broke anything in my (pretty thorough, but you never know) rebase.

We could now also unit test the `VirtualizationRoot` array resizing code's lock-dropping behaviour - at the time I started this PR it wouldn't even compile in the test target - but that would require a few extra mock facilities, such as tracking locks, forcing non-blocking allocs to fail, etc. I might add those to this PR if I get time after I finish the other high-priority tasks on my list. If we want to wait with merging until we have those tests, please review commits 1-3 and 5 in the series, and if those are good to go, I'd like to at least get them merged.

ProjFS.Mac/PrjFSKext/FileSystemMountPoints.hpp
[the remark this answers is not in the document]
wilbaker: > I'm not sure that's really necessary. What else would a forward declaration be for?

Yeah I agree.  It's been a long time since I left this comment, and I'm not sure why I thought it was worth adding a comment at that time.

wilbaker:

Issue #499, opened 2018-11-10
Mac: Investigate hangs during shutdown when virtualization root is attached


nickgra:
_Sometimes_ when our kext is loaded and a provider is running, even if we're not actively triggering I/O in the virtualization root, if you shutdown or reboot the machine the shutdown will 'hang' on a black screen. It will sit like this until physically turned off and CPU load will be at 100% (I've seen this in VMs on my machine and @wilbaker has had his laptop get extremely hot while in this state). My theory is that something about the state our kext is in makes it impossible to be unloaded (or we're in a deadlock). 

Partial repro steps (doesn't repro 100% of the time, but will frequently):
1. Load kext and mount a provider
2. Run some I/O through the provider (or get lucky with some background I/O like a crawler interacting with our kext).
3. 'sudo reboot'
4. The reboot won't finish and the machine will hang on a black screen 

Running applications/daemons will be terminated (sshd, agent.listener, screensharingd) so we can't remotely connect to investigate. However, the machine can still be pinged by IP address so some services are still running on the machine.

pmj:
I have a strong suspicion this is because we're not flushing outstanding messages (`s_outstandingMessages`) when a provider disconnects. When you disconnect a provider manually, your system is usually in a steady state, so chances are there aren't any outstanding messages. On reboot, a lot is going on in the system, and the provider will be violently disconnected, so it can easily happen that there are outstanding messages waiting for the provider that has just died.

The solution is of course to flush any such messages in `ActiveProvider_Disconnect()`. Preferably, the list of outstanding messages should be per virtualisation root, which would make this a lot easier. Additionally, it'll make the search handling a reply message easier too. (and prevent providers from replying to other providers' messages)

I'm doing some work in this area anyway (#284) so I'd be happy to sort this out.

pmj:
To answer your question about debugging this, connecting with a kernel debugger should work, yes. You'll want to enable break-to-KDP with an NMI (`debug` bit `0x4` in `boot-args`, and by my personal preference, I also enable bit `0x8000` so I [don't have to press that ridiculous key combo](https://asciiwwdc.com/2013/sessions/707#t=2743.876)). If it's late enough in the reboot process that networking has been torn down, you probably will struggle with ethernet kdp, but it should still work with firewire, although that's only really an option on a real Mac.

For hangs like this, a core dump should of course also work instead of a kdp session.

pmj:
[...]

PR #501, opened 2018-11-12
Mac kext: Abort events awaiting response when their provider disconnects

pmj:
This change ensures that any events awaiting a response from a provider get flushed out if their corresponding provider process disconnects/dies before responding. I think this should fix issue #499, but even if #499 has another cause, this is worth doing.

I was originally going to fix this by creating an outstanding message list for each provider rather than a single global one, but that would have taken some significantly larger code changes, and from an efficiency point of view, there's not much in it as the list tends to be fairly short.

The first commit in the series fixes a potential `nullptr` dereference KP, but more importantly syncs the event sleep on the outstanding message with the mutex. The second introduces a wrapper function for dropping a held exclusive lock to a shared lock without giving up the entire lock in between. The third commit then implements the fix by calling a new function which walks the list of messages and marks those matching the disconnecting provider as aborted, waking the corresponding threads.

I didn't run into any more deadlock problems when suddenly killing a provider in the middle of a build process. (And some debug output, which I inserted locally but have not included in these commits, indicated that pending events were indeed resumed/aborted.)

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -91,6 +91,7 @@ typedef struct OutstandingMessage
     MessageHeader request;
     MessageType response;
     bool    receivedResponse;
+    VirtualizationRootHandle rootHandle;
     
     LIST_ENTRY(OutstandingMessage) _list_privates;
     
@@ -195,7 +196,7 @@ kern_return_t KauthHandler_Cleanup()
     return result;
 }
 
-void KauthHandler_HandleKernelMessageResponse(uint64_t messageId, MessageType responseType)
+void KauthHandler_HandleKernelMessageResponse(VirtualizationRootHandle providerVirtualizationRootHandle, uint64_t messageId, MessageType responseType)
 {
     switch (responseType)
     {
@@ -207,7 +208,7 @@ void KauthHandler_HandleKernelMessageResponse(uint64_t messageId, MessageType re
                 OutstandingMessage* outstandingMessage;
                 LIST_FOREACH(outstandingMessage, &s_outstandingMessages, _list_privates)
                 {
-                    if (outstandingMessage->request.messageId == messageId)
+                    if (outstandingMessage->request.messageId == messageId && outstandingMessage->rootHandle == providerVirtualizationRootHandle)
                     {
                         // Save the response for the blocked thread.
                         outstandingMessage->response = responseType;
@@ -237,11 +238,31 @@ void KauthHandler_HandleKernelMessageResponse(uint64_t messageId, MessageType re
         case MessageType_KtoU_NotifyFileRenamed:
         case MessageType_KtoU_NotifyDirectoryRenamed:
         case MessageType_KtoU_NotifyFileHardLinkCreated:
+        case MessageType_Response_Aborted:
             KextLog_Error("KauthHandler_HandleKernelMessageResponse: Unexpected responseType: %d", responseType);
             break;
     }
     
     return;
+}
+
+void KauthHandler_AbortOutstandingEventsForProvider(VirtualizationRootHandle providerVirtualizationRootHandle)
+{
+    // Mark all outstanding messages for this root as aborted and wake up the waiting threads
+    Mutex_Acquire(s_outstandingMessagesMutex);
+    {
+        OutstandingMessage* outstandingMessage;
+        LIST_FOREACH(outstandingMessage, &s_outstandingMessages, _list_privates)
+        {
+            if (outstandingMessage->rootHandle == providerVirtualizationRootHandle)
+            {
+                outstandingMessage->receivedResponse = true;
+                outstandingMessage->response = MessageType_Response_Aborted;
+                wakeup(outstandingMessage);
+            }
+        }
+    }
+    Mutex_Release(s_outstandingMessagesMutex);
 }
 
 // Private functions
@@ -753,8 +774,11 @@ static bool TrySendRequestAndWaitForResponse(
 {
     bool result = false;
     
-    OutstandingMessage message;
-    message.receivedResponse = false;
+    OutstandingMessage message =
+    {
+        .receivedResponse = false,
+        .rootHandle = root,
+    };
     
     if (nullptr == vnodePath)
     {
@@ -826,7 +850,7 @@ static bool TrySendRequestAndWaitForResponse(
             result = true;
             goto CleanupUnlockAndReturn;
         }
-        else
+        else // failure reported by provider or aborted due disconnected provider
         {
             // Default error code is EACCES. See errno.h for more codes.
             *kauthError = EAGAIN;
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -791,45 +791,51 @@ static bool TrySendRequestAndWaitForResponse(
     }
     Mutex_Release(s_outstandingMessagesMutex);
     
-    // TODO(Mac): Should we pass in the root directly, rather than root->index?
-    //            The index seems more like a private implementation detail.
-    if (!isShuttingDown && 0 != ActiveProvider_SendMessage(root, messageSpec))
+    if (isShuttingDown)
     {
-        // TODO: appropriately handle unresponsive providers
-        
-        *kauthResult = KAUTH_RESULT_DEFER;
-        goto CleanupAndReturn;
+        // Message never inserted into list, don't try to remove it on cleanup.
+        return false;
     }
     
-    while (!message.receivedResponse &&
-           !s_isShuttingDown)
-    {
-        Sleep(5, &message);
-    }
-    
-    if (s_isShuttingDown)
-    {
-        *kauthResult = KAUTH_RESULT_DENY;
-        goto CleanupAndReturn;
-    }
-
-    if (MessageType_Response_Success == message.response)
-    {
-        *kauthResult = KAUTH_RESULT_DEFER;
-        result = true;
-        goto CleanupAndReturn;
-    }
-    else
-    {
-        // Default error code is EACCES. See errno.h for more codes.
-        *kauthError = EAGAIN;
-        *kauthResult = KAUTH_RESULT_DENY;
-        goto CleanupAndReturn;
-    }
-    
-CleanupAndReturn:
+    errno_t sendError = ActiveProvider_SendMessage(root, messageSpec);
+   
     Mutex_Acquire(s_outstandingMessagesMutex);
     {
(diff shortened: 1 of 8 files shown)


ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -791,45 +815,51 @@ static bool TrySendRequestAndWaitForResponse(
...
-        *kauthResult = KAUTH_RESULT_DENY;
-        goto CleanupAndReturn;
-    }
+        if (MessageType_Response_Success == message.response)
+        {
+            *kauthResult = KAUTH_RESULT_DEFER;
+            result = true;
+            goto CleanupUnlockAndReturn;
+        }
+        else // failure reported by provider or aborted due disconnected provider
```
sanoursa: I don't think this comment adds any real value, I would drop it

sanoursa: And more importantly, it duplicates the information that is already encoded in the enum value definitions. We shouldn't have to remember to update this comment every time we add a new enum value

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -791,45 +815,51 @@ static bool TrySendRequestAndWaitForResponse(
...
     
-    if (s_isShuttingDown)
-    {
-        *kauthResult = KAUTH_RESULT_DENY;
-        goto CleanupAndReturn;
-    }
+        while (!message.receivedResponse &&
+           !s_isShuttingDown)
+        {
+            Mutex_Sleep(s_outstandingMessagesMutex, &message, 5 /* seconds */);
```
sanoursa: I think the existing call to `msleep` inside our `Sleep` wrapper function already has the ability to sleep a mutex, we're just not passing the mutex pointer through to it. I'd like to avoid having two different sleep implementations if we can avoid it - any way to combine these?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -423,7 +424,11 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex)
         
         root->providerUserClient = nullptr;
     }
-    RWLock_ReleaseExclusive(s_rwLock);
+    RWLock_DropExclusiveToShared(s_rwLock);
```
sanoursa: Pattern note: I would vote to delete the closing brace at 426 and opening brace at 428. We're not releasing the lock, so I would not close out the code block

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -29,6 +29,7 @@ typedef enum
     // Responses
     MessageType_Response_Success,
     MessageType_Response_Fail,
+    MessageType_Response_Aborted,
```
sanoursa: This is listed in the "responses" section here, and yet it is explicitly disallowed as a response in the message handling code. How important is it to have a new enum value in the first place? I don't think we're doing anything different as a result of this response - do you have future plans for it?

pmj:
> These code changes look fine, but I'm not seeing how it fixes the issue. We already had code that, at kext unload time, wakes up all threads waiting on outstanding messages. Did you find a code path there where a thread would not wake up properly?

Well, that existing code only runs on kext unload, which doesn’t apply here. Kexts are unloaded when:

* Specifically requested to with kextunload or the corresponding OSKext API
* For IOKit kexts, after a timeout from the point where no instances of OSObject subclasses implemented by the kext exist. (The idea being that a hot-unplug event terminated all driver instances, so the whole kext can be unloaded.)

Only the first applies to us, as we keep the PrjFSService object running indefinitely, listening for providers and kauth events. Note that neither case is triggered by system shutdown or reboot. In that situation, all that happens is that all user space processes are sent SIGTERM, and if that doesn’t help, SIGKILL signals. So for us, the user clients will disconnect; previously, this left kauth events dangling, which this fixes.

ProjFS.Mac/PrjFSKext/public/Message.h
pmj: Yeah, I was on the fence for this, but ultimately, I think we should be handling this case identically to there not being an active provider for that root in the first place, which I’m not sure is always going to be the same as when an active, responding provider is explicitly making a “failed” response?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: OK, I’ll just try to emphasise the change in lock state by surrounding it with blank lines then.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: I don't feel that "sleep for N seconds" and "atomically drop mutex, await event with timeout, reacquire mutex" are conceptually close enough to warrant sharing an API, and that the former purpose (by passing a `nullptr` for the mutex argument) is particularly obvious, but we can certainly go that way.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -791,45 +815,51 @@ static bool TrySendRequestAndWaitForResponse(
     }
     Mutex_Release(s_outstandingMessagesMutex);
     
-    // TODO(Mac): Should we pass in the root directly, rather than root->index?
-    //            The index seems more like a private implementation detail.
-    if (!isShuttingDown && 0 != ActiveProvider_SendMessage(root, messageSpec))
+    if (isShuttingDown)
```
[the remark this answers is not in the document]
pmj: Surely, we need to do that while the lock is held, and we can't return without dropping the lock. I mean obviously we can release the lock in the `if` clause as well as the fall-through, but that in turn would violate our coding conventions?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -791,45 +815,51 @@ static bool TrySendRequestAndWaitForResponse(
...
+            // Default error code is EACCES. See errno.h for more codes.
+            *kauthError = EAGAIN;
+            *kauthResult = KAUTH_RESULT_DENY;
+            goto CleanupUnlockAndReturn;
+        }
     
-CleanupAndReturn:
-    Mutex_Acquire(s_outstandingMessagesMutex);
-    {
+CleanupUnlockAndReturn:
```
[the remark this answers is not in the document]
pmj: Similarly to the `isShuttingDown` comment: we need to make sure the lock is being handled correctly, so I don't follow what you're suggesting here.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
sanoursa: I don't think you need to hold the lock to read the boolean. Either you'll catch that we're shutting down at the top and early exit, or else you'll see it in the loop below while waiting for a response.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
sanoursa: Looking at this more closely - in this case you don't even need any of the goto's or the label. If/else's will do the trick just as cleanly (or more so, actually)

ProjFS.Mac/PrjFSKext/public/Message.h
sanoursa: Yea I can see that. I think I would remove the `_Response_` portion of its name, since it's not a response.

ProjFS.Mac/PrjFSKext/PrjFSKext/Locks.cpp
```diff
@@ -2,6 +2,7 @@
 #include <kern/locks.h>
 #include <kern/thread.h>
 #include <kern/assert.h>
+#include <kern/clock.h>
```
sanoursa: No need for this header any more, I think?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
pmj: I suppose as we re-check the flag before sleeping, there is no actual damage done by inserting into the list of outstanding messages after the previous occupants of this list have already been woken up, as the list item will immediately be removed again, same as if the shutdown came during the period where we dropped the lock for sending the message. And as we perform this second check during a held lock, and the flag never returns to false, it should also be safe from a memory coherence point of view. I don't particularly like accessing variables inconsistently inside vs outside of the mutexes that protect them, but it does make the code slightly shorter in this case.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
pmj: Yup, I've now reworked this part of the code more thoroughly and got rid of the gotos.

ProjFS.Mac/PrjFSKext/public/Message.h
pmj: Using the same logic, the `receivedResponse` and `response` field names in the struct no longer seemed entirely appropriate, so I've [renamed the "response" portion to "result"](https://github.com/Microsoft/VFSForGit/pull/501/commits/81ed1352b63a2b32e4d17a730a7911d947e7b115) there. Sound sensible?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -422,8 +423,12 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex)
         root->providerPid = 0;
         
         root->providerUserClient = nullptr;
+
+        RWLock_DropExclusiveToShared(s_virtualizationRootsLock);
```
wilbaker: What's the reason for dropping this to shared before calling `KauthHandler_AbortOutstandingEventsForProvider`?

pmj: Because we can, essentially. The `s_virtualizationRootsLock` is pretty busy, but most threads only need shared access to it. So this just helps moving things along while we clear out the list of pending messages for the disconnecting provider. We do still need to keep the shared lock as we want to avoid a situation where a new provider for the same root turns up while we're still going through the message list.

PR #507, opened 2018-11-13
Add lower dir and handle to Linux VFS API

chrisd8088:
The VFSforGit API, as defined for Windows and MacOS, does not support the use of an opaque handle as required to allow the Linux libprojfs library to track its event loop thread (and thus allow it to start and, later, stop the thread).

Further, we need to pass the lower-level storageRoot directory path to libprojfs as it will be a stackable Linux filesystem which passes most or all operations through to the lower-level filesystem.

The GVFS.Virtualization.FileSystem.FileSystemVirtualizer class has a Stop() method which returns void, so we follow that lead for MirrorProvider as well.

And as we don't expect to return a value from the corresponding function in the Linux library either, we can also remove the return value from StopVirtualizationInstance() in the Linux PrjFSLib.

### DIFF
--- a/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
+++ b/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
@@ -9,6 +9,7 @@ namespace MirrorProvider.Linux
     public class LinuxFileSystemVirtualizer : FileSystemVirtualizer
     {
         private VirtualizationInstance virtualizationInstance = new VirtualizationInstance();
+        private IntPtr mountHandle = IntPtr.Zero;
 
         public override bool TryConvertVirtualizationRoot(string directory, out string error)
         {
@@ -20,6 +21,10 @@ namespace MirrorProvider.Linux
 
         public override bool TryStartVirtualizationInstance(Enlistment enlistment, out string error)
         {
+            string storageRoot = Path.Combine(enlistment.DotMirrorRoot, "src");
+
+            Directory.CreateDirectory(storageRoot);
+
             this.virtualizationInstance.OnEnumerateDirectory = this.OnEnumerateDirectory;
             this.virtualizationInstance.OnGetFileStream = this.OnGetFileStream;
             this.virtualizationInstance.OnFileModified = this.OnFileModified;
@@ -29,8 +34,10 @@ namespace MirrorProvider.Linux
             this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
 
             Result result = this.virtualizationInstance.StartVirtualizationInstance(
+                storageRoot,
                 enlistment.SrcRoot,
-                poolThreadCount: (uint)Environment.ProcessorCount * 2);
+                poolThreadCount: (uint)Environment.ProcessorCount * 2,
+                mountHandle: out this.mountHandle);
 
             if (result == Result.Success)
             {
@@ -41,6 +48,12 @@ namespace MirrorProvider.Linux
                 error = result.ToString();
                 return false;
             }
+        }
+
+        public override void Stop()
+        {
+            this.virtualizationInstance.StopVirtualizationInstance(
+                this.mountHandle);
         }
 
         private Result OnEnumerateDirectory(
--- a/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
+++ b/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
@@ -9,7 +9,6 @@ namespace MirrorProvider.Linux
     public class LinuxFileSystemVirtualizer : FileSystemVirtualizer
     {
         private VirtualizationInstance virtualizationInstance = new VirtualizationInstance();
-        private IntPtr mountHandle = IntPtr.Zero;
 
         public override bool TryConvertVirtualizationRoot(string directory, out string error)
         {
@@ -36,8 +35,7 @@ namespace MirrorProvider.Linux
             Result result = this.virtualizationInstance.StartVirtualizationInstance(
                 storageRoot,
                 enlistment.SrcRoot,
-                poolThreadCount: (uint)Environment.ProcessorCount * 2,
-                mountHandle: out this.mountHandle);
+                poolThreadCount: (uint)Environment.ProcessorCount * 2);
 
             if (result == Result.Success)
             {
@@ -52,8 +50,7 @@ namespace MirrorProvider.Linux
 
         public override void Stop()
         {
-            this.virtualizationInstance.StopVirtualizationInstance(
-                this.mountHandle);
+            this.virtualizationInstance.StopVirtualizationInstance();
         }
 
         private Result OnEnumerateDirectory(
--- a/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
+++ b/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
@@ -20,7 +20,7 @@ namespace MirrorProvider.Linux
 
         public override bool TryStartVirtualizationInstance(Enlistment enlistment, out string error)
         {
-            string storageRoot = Path.Combine(enlistment.DotMirrorRoot, "src");
+            string storageRoot = Path.Combine(enlistment.DotMirrorRoot, "lower");
 
             Directory.CreateDirectory(storageRoot);
 
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
@@ -16,7 +16,7 @@ namespace PrjFSLib.Linux.Interop
             string virtualizationRootFullPath,
             Callbacks callbacks,
             uint poolThreadCount,
-            ref IntPtr mountHandlePtr);
+            ref IntPtr mountHandle);
 
         [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_StopVirtualizationInstance")]
         public static extern void StopVirtualizationInstance(
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
@@ -12,9 +12,15 @@ namespace PrjFSLib.Linux.Interop
         // TODO(Linux): revise library functions for Linux
         [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_StartVirtualizationInstance")]
         public static extern Result StartVirtualizationInstance(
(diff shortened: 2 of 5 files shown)


ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
```diff
@@ -12,9 +12,15 @@ internal static class PrjFSLib
         // TODO(Linux): revise library functions for Linux
         [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_StartVirtualizationInstance")]
         public static extern Result StartVirtualizationInstance(
+            string storageRootFullPath,
             string virtualizationRootFullPath,
             Callbacks callbacks,
-            uint poolThreadCount);
+            uint poolThreadCount,
+            ref IntPtr mountHandlePtr);
```
sanoursa: Nit: rename to `mountHandle`

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -36,15 +38,23 @@ public static Result ConvertDirectoryToVirtualizationRoot(string fullPath)
                 OnNotifyOperation = this.preventGCOnNotifyOperationDelegate = new NotifyOperationCallback(this.OnNotifyOperation),
             };
 
-            return Interop.PrjFSLib.StartVirtualizationInstance(
+            IntPtr newMountHandle = IntPtr.Zero;
```
sanoursa: I don't think you need a new variable here. Just use `mountHandle`?

MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -20,6 +21,10 @@ public override bool TryConvertVirtualizationRoot(string directory, out string e
 
         public override bool TryStartVirtualizationInstance(Enlistment enlistment, out string error)
         {
+            string storageRoot = Path.Combine(enlistment.DotMirrorRoot, "src");
```
sanoursa: I think it'll be confusing to have this folder called `src`. How about `.storage`? Or some other appropriate "dot" name.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -26,8 +26,10 @@ public static Result ConvertDirectoryToVirtualizationRoot(string fullPath)
         }
 
         public virtual Result StartVirtualizationInstance(
+            string storageRootFullPath,
             string virtualizationRootFullPath,
-            uint poolThreadCount)
+            uint poolThreadCount,
+            out IntPtr mountHandle)
```
sanoursa: It doesn't seem like there's a need to expose the `mountHandle` in this managed API. The `VirtualizationInstance` could just track it as a private member variable.

That said, I'm not really clear on why you need this handle in the first place. Can you give us more detail on how it's being used in the C lib? Can it just be contained to that code and not exposed out of the API at all?

MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: OK ... we choose .gvfs/src (and by extension .mirror/src) in the meeting in Redmond, but I'm agnostic.  Happy to change to .storage if you prefer.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: I can't see an obvious way to keep this internal.  In the MacOS implementation, you can start a background event loop to listen for kernel events via Grand Central Dispatch, and that just gets shut down for you later by the OS.  But in the Linux world, we need a way to keep a handle to a struct which records the event loop thread ID, so we can stop it later.

ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
chrisd8088: I used `mountHandlePtr` because this is a `**thing` into which the start function writes the `*thing`, whereas in the stop function it just takes the `*thing`.  So I'm inclined to stick with this unless it's highly non-idiomatic in C#; let me know what you think!

MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
sanoursa: We definitely agreed to put it inside the `.gvfs` folder, but we hadn't settled on a name that I can recall. I'm open to suggestions, but calling it `src` seems like it will lead to confusion.

ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
sanoursa: It's definitely non-idiomatic, but I also think we can probably drop this :-). So let's discuss that point further.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
sanoursa: To address your second point first - how will you handle a process crash or kill? You can't guarantee that your stop function will be called, so do we need some other way to clean up in that situation?

As for the handle, I agree that you need one. My assertion is that you can probably keep it contained within the C lib, and not bubble it up out of your API. The API's already exposes a "handle" - the virtualization root path. You can map that to whatever internal handle you like, and not leak implementation details out of the API. Is there anything that would prevent that approach?

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: Yup, that works -- thanks, it's cleaner now.  Still learning the C# stuff!  :-)

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: On a crash or kill, there's not much that can be done; we can't trap `SIGKILL`, obviously.  The user would have to `umount` manually to get back to a clean state by hand.  However, on startup we can try to detect this (an existing mount) and either warn or attempt to clean up automatically, then remount.

On `SIGTERM` things are a little different; it's not ideal to have a library set up its own handler (as signal handlers are global across the whole process) so typically the application does that.  But, I could put a `SIGTERM` handler into the "translation" layer (see below).

Re the handle, I see where you're coming from now.  It's non-idiomatic in standard C library terms, but we could have the VFSforGit "translation" layer keep a global path-to-struct map and lookup the root path on the stop function.  So if you prefer, I can move that into "translation" C code (all the VFSforGit API function names and wrappers) which the library only builds if you choose the build-for-VFSforGit configuration option.  Otherwise you'd get only the more conventional C approach.  Since it sounds like that would likely be your strong preference, let me take a stab at that.

sanoursa: I'd rather you not have a "build-for-VFSForGit" flag :-). That path-to-handle translation could also be done in the C# interop layer, just to avoid exposing it to the C# provider. And actually, you wouldn't need to do any real mapping - just have `VirtualizationInstance` store the handle, and not bubble it up any higher than that. From the C# provider's perspective, the `VirtualizationInstance` object _is_ the handle.

chrisd8088: Yes, I like that better, thank you -- and it avoids the problem that if we try to retain state, in the form of paths to mount points, in the library, then it becomes a problem if the user renames some directory above the mount point (e.g., from /repos/this-gvfs-repo to /my-repos/this-gvfs-repo).  I can definitely just keep the handle in `PrjFS.Linux.VirtualizationInstace`.  Thanks!!

MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: Done!

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -26,25 +28,35 @@ public static Result ConvertDirectoryToVirtualizationRoot(string fullPath)
         }
 
         public virtual Result StartVirtualizationInstance(
+            string storageRootFullPath,
             string virtualizationRootFullPath,
             uint poolThreadCount)
         {
+            if (mountHandle != IntPtr.Zero)
```
sanoursa: Style nit: prefix with `this.`

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -26,25 +28,35 @@ public static Result ConvertDirectoryToVirtualizationRoot(string fullPath)
...
+                ref this.mountHandle);
+
+            return result;
         }
 
-        public virtual Result StopVirtualizationInstance()
+        public virtual void StopVirtualizationInstance()
         {
-            throw new NotImplementedException();
+            Interop.PrjFSLib.StopVirtualizationInstance(this.mountHandle);
```
sanoursa: Minor: clear out `this.mountHandle`

ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/PrjFSLib.cs
chrisd8088: OK, done, thanks for the advice!

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: Fixed.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: Good catch!  The latest commit 81f4923 should do this now.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: I think this is fully implemented now, as described.  Thanks for the critique, that was helpful!

chrisd8088:
I think this is up-to-date with all the changes proposed and discussed, @sanoursa; thanks for the keen-eyed review!

Issue #495, opened 2018-11-08
GVFS.Service should not attempt to download updates when running in unattended mode


wilbaker:
Currently GVFS.Service logs the the following message as an Error when running in unattended mode:

> TryRunPreUpgradeChecks: `gvfs upgrade` is not supported in unattended mode

Even if the upgrade ring is set to None.

When running in unattended mode rather than logging an error GVFS.Service should simply not attempt to download an update (and if it's not already, GVFS.Service should log on startup if it's running in unattended mode).

PR #508, opened 2018-11-13
Service should no check for upgrades when unattended

alameenshah:
GVFS.Service - dont schedule upgrade check timer when running in un-attended mode.

#### Issues
Fixes https://github.com/Microsoft/VFSForGit/issues/495

#### Testing
Restart GVFS.Service after updating the System Environment Variable```GVFS_UNATTENDED``` for the changes to take effect.

### DIFF
--- a/GVFS/GVFS.Service/ProductUpgradeTimer.cs
+++ b/GVFS/GVFS.Service/ProductUpgradeTimer.cs
@@ -7,7 +7,7 @@ using System.Threading;
 
 namespace GVFS.Service
 {
-    public class ProductUpgradeTimer
+    public class ProductUpgradeTimer : IDisposable
     {
         private static readonly TimeSpan TimeInterval = TimeSpan.FromDays(1);
         private JsonTracer tracer;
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -20,15 +20,21 @@ public ProductUpgradeTimer(JsonTracer tracer)
...
+                TimeSpan startTime = TimeSpan.Zero;
 
-            this.tracer.RelatedInfo("Starting auto upgrade checks.");
-            this.timer = new Timer(
-                this.TimerCallback,
-                state: null,
-                dueTime: startTime,
-                period: TimeInterval);
+                this.tracer.RelatedInfo("Starting auto upgrade checks.");
+                this.timer = new Timer(
```
jrbriggs: Unrelated to this change, I noticed that ProductUpgradeTimer doesn't implement `IDisposable`. Should it?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -20,15 +20,21 @@ public ProductUpgradeTimer(JsonTracer tracer)
 
         public void Start()
         {
-            Random random = new Random();
-            TimeSpan startTime = TimeSpan.Zero;
+            if (!GVFSEnlistment.IsUnattended(this.tracer))
```
jeschu1: @alameenshah rookie question here, what does it mean for an enlistment to unattended?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: Makes sense. Added Disposable.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: It means VFSForGit is running un-attended by any user, may be inside a lab. This state can be enabled by setting the environment variable ```GVFS_UNATTENDED```.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -20,21 +20,36 @@ public ProductUpgradeTimer(JsonTracer tracer)
...
```
sanoursa: I think you may need to check this inside the `TimerCallback`, because the environment variable can change. I'm not sure though if an already-running process will pick that up, so it may be that someone who turns on unattended mode for the first time also needs to restart the service. Can you please test that out?

jrbriggs: `Environment.GetEnvironment` will only pick up those values that were available at process startup, unless they're set inside the process with `SetEnvironmentVariable`, which we don't do here. Given that, I think this is fine as-is, and consistent with how we handle the mount process deciding whether to start the background prefetch timer.

It begs the larger question of whether this configuration _should_ be environmental or managed more directly through the application, but I don't think that's in scope for this bugfix.

Issue #398, opened 2018-10-16
Harden InProcessMount enlistment/mount id handling


jrbriggs:
Rather than silently ignoring missing config values, we should:
- pull enlistment id from the repometadata and rewrite the config setting
- generate a fresh mount id and rewrite the config setting.

kewillford:
@jeschu1 I know that you did some work around this after my initial implementation.  Is this still outstanding work that we need to do to make it better?  Is there other work that we still need to do to make these ids more reliable?

jeschu1:
@kewillford the process we have seems pretty reliable.  We may want to add EnlistmentId/MountId on Dehydrate, but that seems like a separate issue.

jrbriggs:
Sounds like this issue can be resolved. Please reactivate if you disagree @kewillford @jeschu1

PR #510, opened 2018-11-14
Add EnlistmentId to RepairVerb

jeschu1:
If we can read the enlismentId from the config file we will add it to telemetry.

### DIFF
--- a/GVFS/GVFS.Common/GVFSEnlistment.cs
+++ b/GVFS/GVFS.Common/GVFSEnlistment.cs
@@ -16,6 +16,7 @@ namespace GVFS.Common
         private string gitVersion;
         private string gvfsVersion;
         private string gvfsHooksVersion;
+        private GitProcess gitProcess;
 
         // New enlistment
         public GVFSEnlistment(string enlistmentRoot, string repoUrl, string gitBinPath, string gvfsHooksRoot, GitAuthentication authentication)
@@ -218,6 +219,16 @@ namespace GVFS.Common
             return true;
         }
 
+        public string GetMountId()
+        {
+            return this.GetId(GVFSConstants.GitConfig.MountId);
+        }
+
+        public string GetEnlistmentId()
+        {
+            return this.GetId(GVFSConstants.GitConfig.EnlistmentId);
+        }
+
         private void SetOnce<T>(T value, ref T valueToSet)
         {
             if (valueToSet != null)
@@ -237,6 +248,29 @@ namespace GVFS.Common
         {
             DirectoryInfo dir = Directory.CreateDirectory(path);
             dir.Attributes = FileAttributes.Hidden;
+        }
+
+        private string GetId(string key)
+        {
+            GitProcess.ConfigResult configResult = this.GetGitProcess().GetFromLocalConfig(key);
+            string value;
+            string error;
+            if (!configResult.TryParseAsString(out value, out error, defaultValue: string.Empty))
+            {
+                return value.Trim();
+            }
+
+            return string.Empty;
+        }
+
+        private GitProcess GetGitProcess()
+        {
+            if (this.gitProcess == null)
+            {
+                this.gitProcess = new GitProcess(this);
+            }
+
+            return this.gitProcess;
         }
     }
 }
--- a/GVFS/GVFS/CommandLine/RepairVerb.cs
+++ b/GVFS/GVFS/CommandLine/RepairVerb.cs
@@ -95,7 +95,7 @@ To actually execute any necessary repair(s), run 'gvfs repair --confirm'
 
             this.Output.WriteLine();
 
-            using (JsonTracer tracer = new JsonTracer(GVFSConstants.GVFSEtwProviderName, "RepairVerb"))
+            using (JsonTracer tracer = new JsonTracer(GVFSConstants.GVFSEtwProviderName, "RepairVerb", enlistment.GetEnlistmentId(), mountId: null))
             {
                 tracer.AddLogFileEventListener(
                     GVFSEnlistment.GetNewGVFSLogFileName(enlistment.GVFSLogsRoot, GVFSConstants.LogFileTypes.Repair),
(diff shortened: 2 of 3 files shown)


GVFS/GVFS/CommandLine/RepairVerb.cs
```diff
@@ -95,7 +97,15 @@ public override void Execute()
 
             this.Output.WriteLine();
 
-            using (JsonTracer tracer = new JsonTracer(GVFSConstants.GVFSEtwProviderName, "RepairVerb"))
+            string enlistmentId = null;
```
derrickstolee: This is good to do, but we should stop repeating ourselves! I can see this becoming an important thing to do in more places.

I would add an `EnlistmentId` property to `GVFSEnlistment` that initializes on first call using this logic. Then replace the logic in `InProcessMount.CreateTracer()` with `enlistment.EnlistmentId`.

```
        private bool initializedIds;
        private string enlistmentId;
        private string mountId;

...

        public string EnlistmentId
        {
            get
            {
                this.InitializeIds();
                return this.enlistmentId;
            }
        }

        public string MountId
        {
            get
            {
                this.InitializeIds();
                return this.mountId;
            }
        }

...

        private void InitializeIds()
        {
            if (this.initializedIds)
            {
                return;
            }

            GitProcess git = new GitProcess(this);


            GitProcess.Result configResult = git.GetFromLocalConfig(
                                                GVFSConstants.GitConfig.EnlistmentId);
            if (!configResult.HasErrors)
            {
                this.enlistmentId = configResult.Output.Trim();
            }

            configResult = git.GetFromLocalConfig(GVFSConstants.GitConfig.MountId);
            if (!configResult.HasErrors)
            {
                this.mountId = configResult.Output.Trim();
            }

            this.initializedIds = true;
        }
```

kewillford: Yeah I agree we need this logic in one place.  We also should also use the RepoMetadata to get the `EnlistmentId` since this is where it is stored to begin with.  We could even try both places.  We should probably also detect when there is not an entry and create a new id if there is not one.  Not sure if this would only be if the call succeeded but there wasn't a value or for an error condition as well.  If there was an error and we change the EnlistmentId reporting on the enlistment would not be right but we at least have an EnlistmentId.  There is issue #398 that is for this work.

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -79,6 +82,24 @@ public string GVFSHooksVersion
             get { return this.gvfsHooksVersion; }
         }
 
+        public string EnlistmentId
+        {
+            get
+            {
+                this.InitializeIds();
```
sanoursa: I don't think it's a good idea to have a property do things as expensive as shelling out to a process.

Is there a good reason to lazy-initialize these values? Can we just initialize them up front and get rid of this extra complexity?

jeschu1: @kewillford please do correct me if I'm wrong

GVFSVerb.ForExistingEnlistment will create a GVFSEnlistment, but I"m not seeing any of the verbs that extend it use enlistmentId.  Ex Dehydrate/Diagnose/Status/Prefetch.  I do think it would be great if we have the value there though!

MountId gets a little bit more tricky.  We don't need it for Repair for instance, but probably not a killer if we look up another config setting.  However, MountId can change from it's previous config value while we run the mount process.  I think making this one into a function instead of a property and looking it up on-demand is probably safest.  It should at most be called once and that would match the current functionality.

kewillford: Currently only the CloneVerb and MountVerb are calling `LogEnlistmentInfoAndSetConfigValues` that initialized the EnlistmentId and MountId and sets them in the config.  Later code was added in `CreateTracer` of the InProcessMountVerb to lookup up the ids so that the tracer that gets created could always have those for any ETW event.  Having them as a part of the Enlistment makes sense and should then be available when needed but in some cases the Tracer is created before any Enlistment so we should check those cases and see if there is something we need to do there.

kewillford: In most cases we only have one Enlistment object that gets created and what if at some point we want `gvfs status` to output the EnlistmentId and MountId so we can ask a user for them and run queries against them?

jeschu1: @kewillford that's for the clarification here!  I've updated it to be a lookup on demand which is hopefully pretty straightforward.  Since it's on Enlistment it's a pretty central place to get at EnlistmentId/MountId.  

My understanding (and @derrickstolee  feel free to correct me) is that we want status to be as fast as possible so a config lookup wouldn't be in the cards.

derrickstolee: I think `gvfs status` is ok to slow down a bit _if we actually show the information_. Kevin knows this area the best, so I defer to his opinions.

kewillford: I think we can defer initializing in the constructor until we need them initialized in more cases.

jeschu1: I really do appreciate the conversation @kewillford / @derrickstolee !  At a minimum now we do have an easy way to grab enlistment/mountIds.  

I will look at adding them to more verbs when we have the need.

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -218,6 +221,16 @@ public bool TryCreateEnlistmentFolders()
             return true;
         }
 
+        public string MountId()
```
kewillford: These methods should probably start with `Get`

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -238,5 +251,26 @@ private void CreateHiddenDirectory(string path)
...
+            GitProcess.Result configResult = this.GetGitProcess().GetFromLocalConfig(key);
+            if (!configResult.HasErrors)
+            {
+                return configResult.Output.Trim();
+            }
+
+            return string.Empty;
+        }
+
+        private GitProcess GetGitProcess()
```
kewillford: This opens an interesting question about how we are using the `GitProcess` in other places since in most of those places we pass the `Enlistment` it might make sense to just grab this one that is in the enlistment object.  But that would be for a future PR.

PR #511, opened 2018-11-14
Upgrade reminder causing functional test failure

### LINKED ISSUE #428: Functional tests can fail because of unexpected `A new version of GVFS is available. Run `gvfs upgrade` to start the upgrade.`
I just tried running the functional tests on my Windows machine and they failed because of this unexpected text printed by the git command:

>A new version of GVFS is available. Run `gvfs upgrade` to start the upgrade.

alameenshah:
Previously downloaded upgrade installers were triggering reminder messaging from git hook. This in-turn cause FunctionalTests to fail. This PR deletes the downloaded installers before starting FunctionalTests. GVFS.Service is also updated to Not download upgrades when development version is installed. User can still run ```gvfs upgrade --confirm``` to manually upgrade gvfs.

Fixes #428

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
@@ -60,20 +60,30 @@ namespace GVFS.FunctionalTests.Tools
             string gvfsRepoRoot,
             string command,
             Dictionary<string, string> environmentVariables = null,
-            bool cleanErrors = true)
+            bool removeWaitingMessages = true,
+            bool removeUpgradeMessages = true)
         {
             ProcessResult result = GitProcess.InvokeProcess(gvfsRepoRoot, command, environmentVariables);
-
             string errors = result.Errors;
-            if (cleanErrors)
-            {
-                string[] lines = errors.Split(new string[] { "\r\n" }, StringSplitOptions.None);
-                errors = string.Join("\r\n", lines.Where(line => !line.StartsWith("Waiting for ")));
 
-                if (errors.Length > 0 && string.IsNullOrWhiteSpace(errors))
+            if (!string.IsNullOrEmpty(errors) && (removeWaitingMessages || removeUpgradeMessages))
+            {
+                IEnumerable<string> errorLines = errors.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
+                IEnumerable<string> filteredErrorLines = errorLines.Where(line =>
                 {
-                    errors = string.Empty;
-                }
+                    if (string.IsNullOrWhiteSpace(line) ||
+                        (removeUpgradeMessages && line.StartsWith("A new version of GVFS is available.")) ||
+                        (removeWaitingMessages && line.StartsWith("Waiting for ")))
+                    {
+                        return false;
+                    }
+                    else
+                    {
+                        return true;
+                    }
+                });
+
+                errors = filteredErrorLines.Any() ? string.Join(Environment.NewLine, filteredErrorLines) : string.Empty;
             }
 
             return new ProcessResult(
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -122,6 +122,17 @@ private static void RunBeforeAnyTests()
...
+
+                string upgradeDownloadPath = Path.Combine(
+                    Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData),
+                    "GVFS",
+                    "GVFS.Upgrade",
+                    "Downloads");
+
+                if (Directory.Exists(upgradeDownloadPath))
+                {
+                    Directory.Delete(upgradeDownloadPath, recursive: true);
```
sanoursa: This can throw exceptions, and its recursive implementation is quite unreliable. We have our own method for doing a reliable recursive directory delete.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -23,12 +23,19 @@ public void Start()
             Random random = new Random();
             TimeSpan startTime = TimeSpan.Zero;
 
-            this.tracer.RelatedInfo("Starting auto upgrade checks.");
-            this.timer = new Timer(
-                this.TimerCallback,
-                state: null,
-                dueTime: startTime,
-                period: TimeInterval);
+            if (!ProcessHelper.IsDevelopmentVersion())
```
sanoursa: As we've discussed in past PRs, I don't think you should ever disable a feature based on it being a development version. It is incredibly risky to ship code to customers that is different than the code we run ourselves.

And in this case, this check will not prevent the failures anyway, at least not on dev boxes. If I am dogfooding the product and have a non-dev version installed and running, it'll happily download the next upgrade. Then when I build and run the tests on my dev version, it will find those downloads. Yes, you're currently deleting them on test startup, but that is racy as the download could happen right after it.

GVFS/GVFS.Hooks/Program.cs
```diff
@@ -109,7 +109,7 @@ private static void RemindUpgradeAvailable()
             // sure that the reminder is displayed only 10% of the times a git 
             // command is run, check that the random number is between 0 and 10, 
             // which will have a probability of 10/100 == 10%.
-            int reminderFrequency = 10;
+            int reminderFrequency = 100;
```
alameenshah: This is only to product reminder messaging during all the git command tests. Will remove before final commit.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: Alternate approach is to look for upgrade messaging in the git output and ignore it in the Functional Tests. Have incorporated this, please take a look.

GVFS/GVFS.FunctionalTests/Program.cs
alameenshah: This not needed anymore, removed. Updated Functional Tests to ignore reminder messaging.

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
```diff
@@ -60,11 +60,21 @@ public static void CheckGitCommandAgainstGVFSRepo(string virtualRepoRoot, string
...
 
             string errors = result.Errors;
+            if (trimUpgradeReminder)
+            {
+                string upgradeReminder = Environment.NewLine + "A new version of GVFS is available. Run `gvfs upgrade` to start the upgrade." + Environment.NewLine;
+                if (errors.EndsWith(upgradeReminder))
+                {
+                    errors = errors.Substring(0, errors.LastIndexOf(upgradeReminder));
+                }
+            }
```
kewillford: Could this be incorporated into the code below so you would have something like
```
IEnumerable<string> lines = errors.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
if (trimUpgradeReminder)
{
    lines = lines.Where(line => !line.StartsWith("A new version..."));
}

if (cleanErrors)
{
    lines = lines.Where(line => !line.StartsWith("Waiting for "));
}

errors = string.Join(Environment.NewLine, lines);
```

You could also check both flags and only do the `Split` and `Join` if at least one was set to avoid to perf hit when both are false.

Issue #487, opened 2018-11-08
GVFS should use http client SSL authentication settings from git config


turbonaitis:
Git config has many http configuration items in it, some of them authentication related (https://git-scm.com/docs/git-config#git-config-httpemptyAuth). The relevant ones to our use case are these:
- `http.sslCert` - specifies client SSL certificate
- `http.sslKey` - specifies the private key for the client SSL certificate
- `http.sslCertPasswordProtected`

turbonaitis:
Thinking about implementation - currently the simplified authentication flow is this:
* GVFS tries to access server anonymously. If that succeeds, we continue without user credentials
* If anonymous request fails with `Unauthorized`, we try to get credentials from `git credential` and use them 

Now, with these SSL settings:
1. Do we want to still try anonymous request, before loading the configured certificate? I'd say yes, if we can get away with anonymous access, it'd be great to avoid the cost of getting the certificate etc. 
2. Do we try to make a request with just the configured SSL certificate, or do we still set authorization header from `git credential`? My suggestion is to try to use username/password from `git credential` immediately along with the certificate, otherwise we'll end up with the question which combinations of client cert + username/passwrod we use and in what order.

turbonaitis:
After looking into how git uses those settings, I actually think, that we should always send the ssl client certificates, if configured, as this is how git behaves. It also simplifies the code quite a bit.

jrbriggs:
Closing this issue: the PR was merged.

PR #512, opened 2018-11-15
Load SSL certificate according to git config

turbonaitis:
This PR addresses #487.

It covers the scenarios, when sslCert configured in .gitconfig points either to a certificate with private key in a keychain, or to a file, containing both public and private keys.

One thing missing is sslKey support - if our .gitconfig specifies sslCert, containing just the public part, and sslKey, containing the private key, the current PR doesn't load it. Main reason being, is that dotnet currently does not support easy handling of PEM files. I'll create a separate PR addressing that, as I believe it will lead to much more discussins.


GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -31,22 +41,51 @@ static HttpRequestor()
...
-            this.client.Timeout = retryConfig.Timeout;
             this.RetryConfig = retryConfig;
-            this.authentication = authentication;
+
+            this.authentication = enlistment.Authentication;
 
             this.Tracer = tracer;
 
+            var httpClientHandler = new HttpClientHandler() { UseDefaultCredentials = true };
+            httpClientHandler.ClientCertificateOptions = ClientCertificateOption.Manual;
```
jrbriggs: I don't think we should change anything about the mainline flow. Can we only set this when there is a cert provided?

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -31,22 +41,51 @@ static HttpRequestor()
             availableConnections = new SemaphoreSlim(ServicePointManager.DefaultConnectionLimit);
         }
 
-        public HttpRequestor(ITracer tracer, RetryConfig retryConfig, GitAuthentication authentication)
+        protected HttpRequestor(ITracer tracer, RetryConfig retryConfig, Enlistment enlistment)
```
jrbriggs: Should the ssl settings be a property of the authentication object?

GVFS/GVFS.Common/Git/GitSslSettings.cs
```diff
@@ -0,0 +1,40 @@
...
+                }
+                else
+                {
+                    this.SslCertPasswordProtected = false;
+                }
+            }
+            else
+            {
+                this.SslCertificate = default(string);
+                this.SslCertPasswordProtected = default(bool);
```
jrbriggs: Let's be explicit about `null` and `false` as the default values.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -158,6 +158,58 @@ public virtual void RevokeCredential(string repoUrl)
...
+            string certificatePath,
+            out string password,
+            out string errorMessage)
+        {
+            password = null;
+            errorMessage = null;
+
+            using (ITracer activity = tracer.StartActivity("TryGetCertificatePassword", EventLevel.Informational))
+            {
+                Result gitCredentialOutput = this.InvokeGitAgainstDotGitFolder(
```
jrbriggs: Have you tested this on Windows?

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -249,10 +303,42 @@ private static string GetSingleHeaderOrEmpty(HttpHeaders headers, string headerN
...
+            }
+
+            return null;
+        }
+
+        private X509Certificate2 LoadCertificate(string certId, string certificatePassword)
+        {
+            if (File.Exists(certId))
+            {
+                return new X509Certificate2(certId, certificatePassword);
```
kewillford: Do we need some exception handling in this method since this can throw an exception?

GVFS/GVFS.Common/Git/GitProcess.cs
turbonaitis: No, sadly I do not have access to a windows machine. Do you think something in particular may behave differently here?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -253,6 +305,19 @@ public Result SetInFileConfig(string configFile, string settingName, string valu
                  value));
         }
 
+        public bool TryGetConfigUrlMatch(string section, string repositoryUrl, out Dictionary<string, GitConfigSetting> configSettings)
+        {
+            Result result = this.InvokeGitAgainstDotGitFolder($"config --get-urlmatch {section} {repositoryUrl}");
+            if (result.HasErrors)
```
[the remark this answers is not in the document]
turbonaitis: Ideally, we would, but I was following the pattern, established by other TryGetConfig methods, which don't take. Also, neither Enlistment constructor (current call site), nor GitAuthentication constructor (potential new call site) have ITracer instance to pass here.

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -21,8 +23,16 @@ public abstract class HttpRequestor : IDisposable
 
         private readonly ProductInfoHeaderValue userAgentHeader;
 
+        private readonly GitAuthentication authentication;
+
+        private readonly Lazy<X509Store> store = new Lazy<X509Store>(() =>
+        {
+            var s = new X509Store();
```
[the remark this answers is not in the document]
turbonaitis: Out of interest, is there a reason for this? Also, do you guys have some coding guidelines, which I could follow in the future?

GVFS/GVFS.Common/Git/GitConfigHelper.cs
```diff
@@ -79,13 +79,14 @@ public static bool TrySanitizeConfigFileLine(string fileLine, out string sanitiz
...
         /// <param name="input">The lines of text with the settings to parse.</param>
+        /// <param name="delimiter">The delimiter char, separating key from value</param>
         /// <returns>A dictionary of settings, keyed off the setting name representing the settings parsed from input.</returns>
-        public static Dictionary<string, GitConfigSetting> ParseKeyValues(IEnumerable<string> input)
+        public static Dictionary<string, GitConfigSetting> ParseKeyValues(IEnumerable<string> input, char delimiter = '=')
         {
             Dictionary<string, GitConfigSetting> configSettings = new Dictionary<string, GitConfigSetting>(StringComparer.OrdinalIgnoreCase);
             foreach (string line in input)
             {
-                string[] fields = line.Trim().Split(new[] { '=' }, 2, StringSplitOptions.None);
```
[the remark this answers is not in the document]
turbonaitis: The reason I removed trim is to cater for cases, where the separator is a space and the value is empty. We are trimming `key` and `value` below.

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Good point, added.

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Moved the ClienCertificateOption assignment further down, to happen just before setting the certificate. The httpClientHandler isn't really new - it was just being created inline, in the HttpClient constructor.

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Moved it there.

GVFS/GVFS.Common/Enlistment.cs
```diff
@@ -46,9 +46,18 @@ public abstract class Enlistment
                 }
 
                 this.RepoUrl = originResult.Output.Trim();
+            }
+
+            this.Authentication = authentication ?? new GitAuthentication(gitProcess, this.RepoUrl);
+
+            if (gitProcess.TryGetConfigUrlMatch("http", this.RepoUrl, out var configSettings))
```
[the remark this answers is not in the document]
turbonaitis: Moved `SslSettings` property and its assignment to `GitAuthentication` class.

GVFS/GVFS.Common/Git/GitSslSettings.cs
turbonaitis: Great suggestion, changes made.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -158,6 +158,58 @@ public virtual void RevokeCredential(string repoUrl)
                 null);
         }
 
+        /// <summary>
+        /// Input for certificate crednetials looks like
```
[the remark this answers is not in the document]
turbonaitis: Fixed.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -158,6 +158,58 @@ public virtual void RevokeCredential(string repoUrl)
...
+            using (ITracer activity = tracer.StartActivity("TryGetCertificatePassword", EventLevel.Informational))
+            {
+                Result gitCredentialOutput = this.InvokeGitAgainstDotGitFolder(
+                    "credential fill",
+                    stdin => stdin.Write("protocol=cert\npath=" + certificatePath + "\nusername=\n\n"),
+                    parseStdOutLine: null);
+
+                if (gitCredentialOutput.HasErrors)
+                {
+                    EventMetadata errorData = new EventMetadata();
```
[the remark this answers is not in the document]
turbonaitis: Added.

GVFS/GVFS.Common/Http/HttpRequestor.cs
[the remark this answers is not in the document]
kewillford: >is there a reason for this?

We mainly do this for code readability when doing code reviews so you don't have to go searching for the type that is declared.  For example if you have the following:
```
var a = SomeMethod();
<a bunch of code>
var b = a;
<a bunch of code>
var c = b;
```
Doing a code review, you would have to go searching to know what the type of a, b, or c is.

There are also subtle bugs that can slip through.  

I know that in most cases the compiler will catch things and in your case here the type is right there for the reviewer to see, but to be more explicit we have opted to not use it.
>do you guys have some coding guidelines, which I could follow in the future?

We have mostly relied on stylecop to catch any style type issues but that is a good idea of something to add.  I added #537 to track this.

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -31,22 +42,51 @@ static HttpRequestor()
...
-            this.client.Timeout = retryConfig.Timeout;
             this.RetryConfig = retryConfig;
-            this.authentication = authentication;
+
+            this.authentication = enlistment.Authentication;
 
             this.Tracer = tracer;
 
+            var httpClientHandler = new HttpClientHandler() { UseDefaultCredentials = true };
+#if DEBUG
```
sanoursa: I don't think the other reviews have called this out yet. Please avoid precompiler constants - you'll notice that we don't have a single one in any of our C# files.

Two other techniques that you can use: 
* git config setting that enables/disables self-signed certs
* Call `ProcessHelper.IsDevelopmentVersion`. This is dangerous though, because we don't want to ship different code than what we test, so I would personally favor a bool config setting that just says yes/no to the question of self-signed certs

I would also encapsulate all of the logic related to self-signed certs into one place. Right now I see a sprinkling of these `#if DEBUG` checks, and that is going to be difficult to maintain.

turbonaitis: Git actually already has http.sslVerify setting, which I'll reuse for this.

GVFS/GVFS.Common/Git/GitProcess.cs
jrbriggs: Prompting for a password from potentially a background process is just a flow we don't have good automation around, which is why I was asking.

turbonaitis:
@jrbriggs I think I've addressed all of the comments above. Are there any other outstanding actions, that I should take, before we can merge this PR?

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -249,10 +306,63 @@ private static string GetSingleHeaderOrEmpty(HttpHeaders headers, string headerN
...
         }
+
+        private string LoadCertificatePassword(string certId, GitProcess git)
+        {
+            if (git.TryGetCertificatePassword(this.Tracer, certId, out var password, out var error))
+            {
+                return password;
+            }
+
+            return null;
```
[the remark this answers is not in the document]
turbonaitis: We will try to load the certificate without the password. If that fails, we'll return null for certificate, which means, that we'll try to access our HTTP endpoint without it, which will likely result in an authentication error. 
If there was an error, while retrieving password, it will be logged. 

Another way would be to stop the flow immediately, if either password or certificate can not be retrieved, but I think allowing the customer to proceed till he actually fails a request due to the missing certificate is less disruptive. 

One thing we should probably do, is to log a warning, stating that password was configured, but not retrieved (and same for certificate).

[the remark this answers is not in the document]
turbonaitis: Added the warning messages

jrbriggs:
There's a couple of stylecop warnings to fix.

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -21,8 +24,16 @@ public abstract class HttpRequestor : IDisposable
...
         private readonly ProductInfoHeaderValue userAgentHeader;
 
+        private readonly GitAuthentication authentication;
+
+        private readonly Lazy<X509Store> store = new Lazy<X509Store>(() =>
+        {
+            X509Store s = new X509Store();
+            s.Open(OpenFlags.OpenExistingOnly | OpenFlags.ReadOnly);
+            return s;
+        });
```
kewillford: Can we move most of this code closer to where it belongs?  For example this `store` should be in the `GitSslSettings` class along with the `LoadCertificatePassword` and `LoadCertificate`.  Probably want to rename the class to `GitSsl` since it would be more than just the settings.

Then I would have a method on the `GitAuthentication` class that is 
```
void SetupSslIfNeeded(ITracer tracer, HttpClientHandler httpClientHandler, GitProcess gitProcess)
```
Which does all the work in the constructor of this class so it would only have 
```
HttpClientHandler httpClientHandler = new HttpClientHandler() { UseDefaultCredentials = true };
enlistment.Authentication.SetupSslIfNeeded(tracer, httpClientHandler, enlistment.CreateGitProcess());
```

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -31,22 +42,63 @@ static HttpRequestor()
...
-            this.client = new HttpClient(new HttpClientHandler() { UseDefaultCredentials = true });
-            this.client.Timeout = retryConfig.Timeout;
             this.RetryConfig = retryConfig;
-            this.authentication = authentication;
+
+            this.authentication = enlistment.Authentication;
 
             this.Tracer = tracer;
 
+            var httpClientHandler = new HttpClientHandler() { UseDefaultCredentials = true };
```
kewillford: Please remove all uses of `var` since we don't use that in this code base.  Thanks!

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -249,10 +316,69 @@ private static string GetSingleHeaderOrEmpty(HttpHeaders headers, string headerN
...
             if (headers.TryGetValues(headerName, out values))
             {
-                return values.First();                
+                return values.First();
             }
 
             return string.Empty;
         }
+
+        private string LoadCertificatePassword(string certId, GitProcess git)
```
kewillford: This should be `GetCertificatePassword` I think because we aren't really loading it anywhere.  Same with `LoadCertificate`.

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -249,10 +316,69 @@ private static string GetSingleHeaderOrEmpty(HttpHeaders headers, string headerN
...
+        private X509Certificate2 LoadCertificate(string certId, string certificatePassword, bool onlyLoadValidCertificateFromStore)
+        {
+            EventMetadata metadata = new EventMetadata
+            {
+                { "certId", certId },
+                { "isPasswordSpecified", string.IsNullOrEmpty(certificatePassword) },
+                { "shouldVerify", onlyLoadValidCertificateFromStore }
+            };
+
+            if (File.Exists(certId))
```
kewillford: What is the purpose of this check?  Null or empty was already checked before this method was called and the file could be deleted between this call and the constructor for `X509Certificate2` and you are already handling the exception for that but it will be a different code path.  

If the file doesn't exist you run `Find`. If it was deleted between this call and the `X509Certificate2` constructor which means it doesn't exist it will return null.  Not sure how much this matters, just pointing out the difference in code paths.

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -249,10 +316,69 @@ private static string GetSingleHeaderOrEmpty(HttpHeaders headers, string headerN
...
+                    metadata.Add("Exception", cryptEx);
+                    this.Tracer.RelatedError(metadata, "Error, while loading certificate from disk");
+                    return null;
+                }
+            }
+
+            try
+            {
+                var findResults = this.store.Value.Certificates.Find(X509FindType.FindBySubjectName, certId, onlyLoadValidCertificateFromStore);
+                if (findResults?.Count > 0)
```
kewillford: If there is more than 1, do we care?  How do we know we are picking the right one?

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Great suggestions, will do!

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Sure, went and rooted out every one of them in the files, that I've touched.

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Renamed to `GetCertificactePassword`, but I'm not sure I agree with suggestion to rename `LoadCertificate`, as we are loading it - from file or from certificate store. Of course, this is your code base, so if you feel strongly about this, I can rename it too.

kewillford: I prefer to use `Get` because we are getting it either from a file or from the certificate store.  The difference to me is `Get` implies a return value.  `Load` to me doesn't necessarily imply a return value. It could be loading it into a class variable or something else.

GVFS/GVFS.Common/Http/HttpRequestor.cs
kewillford: There is another option of:
3. Try each one until one works or fail if none work, so add a foreach loop.

Reading up on this and don't we want to use [`FindBySubjectDistinguishedName`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509findtype?view=netframework-4.7.2)?  The documentation says that `FindBySubjectName`
>will find all certificates with the subject name containing that string, regardless of other subject values. Searching by distinguished name is a more precise search.

To me that mean that if I pass "test", I will get the certificates for "mytest", "test", "testing", etc.  

So option 3 would be the only one to work in that case.  Do we have the full name of the cert so we can use `FindBySubjectDistinguishedName`?

GVFS/GVFS.Common/Http/HttpRequestor.cs
kewillford: There might be a way to parse the string and determine if the format is a path or name but that seems overly complex so this check will be fine.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Security.Cryptography.X509Certificates;
+using GVFS.Common.Tracing;
+
+namespace GVFS.Common.Git
+{
+    public class GitSsl : IDisposable
+    {
+        public readonly string SslCertificate;
+        public readonly bool SslCertPasswordProtected;
+        public readonly bool SslVerify;
```
kewillford: Some naming suggestions:  Drop `Ssl`
`CertificatePathOrName`
`IsCertPasswordProtected`
`ShouldVerify`

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: Makes sense, renamed LoadCertificate to GetCertificate.

GVFS/GVFS.Common/Http/HttpRequestor.cs
turbonaitis: It's too complex, and you can't determine, if a string "certificate" refers to a file in your current working directory, or to a subject name in a certificate store.

GVFS/GVFS.Common/Http/HttpRequestor.cs
kewillford: I like your plan of logging a warning and narrowing down the certs.  Like you said hopefully there will be enough information in the logs to find out what is going on.  Thanks

turbonaitis: Pushed changes, implementing the above.

turbonaitis:
Rebased on latest master to prevent merge-conflicts.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            return null;
+        }
+
+        public X509Certificate2 GetCertificate(ITracer tracer, string certificatePassword, bool onlyLoadValidCertificateFromStore)
+        {
+            EventMetadata metadata = new EventMetadata
+            {
+                { "CertificatePathOrSubjectCommonName", this.CertificatePathOrSubjectCommonName },
```
kewillford: Use `nameof(this.CertificatePathOrSubjectCommonName)` to avoid the string getting out of sync with the property name.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+                { "isPasswordSpecified", string.IsNullOrEmpty(certificatePassword) },
+                { "shouldVerify", onlyLoadValidCertificateFromStore }
+            };
+
+            var result =
+                this.GetCertificateFromFile(tracer, metadata, certificatePassword, onlyLoadValidCertificateFromStore) ??
+                this.GetCertificateFromStore(tracer, metadata, onlyLoadValidCertificateFromStore);
```
kewillford: We should be able to remove the `onlyLoadValidCertificateFromStore` variable and use the `ShouldVerify` property right?

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+            X509Store store = null;
+            try
+            {
+                store = new X509Store();
+                store.Open(OpenFlags.OpenExistingOnly | OpenFlags.ReadOnly);
+
+                X509Certificate2Collection findResults = store.Certificates.Find(X509FindType.FindBySubjectName, this.CertificatePathOrSubjectCommonName, onlyLoadValidCertificateFromStore);
+                if (findResults?.Count > 0)
+                {
+                    LogCertificateCounts(tracer, metadata, findResults.OfType<X509Certificate2>(), "Found {0} certificates by provided name. Matching DNs: {1}");
```
kewillford: I don't think the string with format should be so far away from where it is used in case there is something that needs to change and one needs to pass more that 2 items for formatting.  Can we create the string here instead?
```
IEnumerable<X509Certificate2> certificates = findResults.OfType<X509Certificate2>();
string matchingNames = string.Join(Environment.NewLine, certificates.Select(x => x.Subject));
string logMessage = $"Found {certificates.Count()} certificates by provided name. Matching DNs: {matchingNames}";
```

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                    X509Certificate2[] certsWithMatchingCns = findResults
+                        .OfType<X509Certificate2>()
+                        .Where(x => x.HasPrivateKey && Regex.IsMatch(x.Subject, string.Format("(^|,\\s?)CN={0}(,|$)", this.CertificatePathOrSubjectCommonName))) // We only want certificates, that have private keys, as we need them. We also want a complete CN match
```
kewillford: Not a fan of regex.  They are hard to understand and maintain.  Can we just use
`x.Subject.IndexOf($"CN={this.CertificatePathOrSubjectCommonName}") >= 0` ?

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+                        .OrderByDescending(x => x.Verify()) // Ordering by validity in a descending order will bring valid certificates to the beginning
+                        .ThenBy(x => x.NotBefore) // We take the one, that was issued earliest, first
+                        .ThenByDescending(x => x.NotAfter) // We then take the one, that is valid for the longest period
+                        .ToArray();
```
kewillford: You can remove the call to `ToArray` and eliminate enumerating the collection multiple times.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            return null;
+        }
+
+        private X509Certificate2 GetCertificateFromStore(ITracer tracer, EventMetadata metadata, bool onlyLoadValidCertificateFromStore)
+        {
+            X509Store store = null;
+            try
+            {
+                store = new X509Store();
```
kewillford: Use a using for the store?
```
using (X509Store store = new X509Store())
{
```
and you can remove the `finally` at the end

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                    X509Certificate2[] certsWithMatchingCns = findResults
+                        .OfType<X509Certificate2>()
+                        .Where(x => x.HasPrivateKey && Regex.IsMatch(x.Subject, string.Format("(^|,\\s?)CN={0}(,|$)", this.CertificatePathOrSubjectCommonName))) // We only want certificates, that have private keys, as we need them. We also want a complete CN match
+                        .OrderByDescending(x => x.Verify()) // Ordering by validity in a descending order will bring valid certificates to the beginning
+                        .ThenBy(x => x.NotBefore) // We take the one, that was issued earliest, first
+                        .ThenByDescending(x => x.NotAfter) // We then take the one, that is valid for the longest period
+                        .ToArray();
+
+                    LogCertificateCounts(tracer, metadata, certsWithMatchingCns, "Found {0} certificates with a private key and an exact CN match. DNs (sorted by priority, will take first): {1}");
```
kewillford: Same as the first comment for `LogCertificateCounts`.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                    return certsWithMatchingCns.FirstOrDefault();
+                }
+            }
+            catch (CryptographicException cryptEx)
+            {
+                metadata.Add("Exception", cryptEx);
```
kewillford: Call `ToString()` on `cryptEx`.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: It depends - do we want to keep the logs stable, regardless of potential renames of variables? If so, having a hardcoded string (which diverge from the variable name in the future) would give us that.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: They are hard to understand, if you don't know them :)  I agree though, that they're not really readable, but they allow us to achieve precise string matching.
Your suggested code would also match "CN={subjectName}AndSomethingElse" as well as "DCN={subjectName}" - so not an exact match.

As always, this is your code and your standards, so if you're fine with imprecise matching, I can implement your suggested changes.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: But `ToArray` is giving us precisely that - preventing from enumerating the whole collection multiple times. If we remove the `ToArray` call, we'll evaluate the whole expression three times:
1. When getting the number of results for logging
2. When selecting the subject names for logging
3. On the `return` statement`, when selecting the `FirstOrDefault`

When we do `ToArray`, we only operate on the resulting array in all the three cases, without any ordering, or filtering.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+                    loggingFunction = tracer.RelatedInfo;
+                    break;
+                default:
+                    loggingFunction = tracer.RelatedWarning;
+                    break;
+            }
+
+            loggingFunction(
+                metadata,
+                logMessage);
```
kewillford: We can get rid of the `loggingFunction` now and just call the `tracer.Related` method in each of the case statements.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+                if (configSettings.TryGetValue(GitConfigSetting.HttpSslVerify, out GitConfigSetting sslVerify))
+                {
+                    this.ShouldVerify = sslVerify.Values.Select(bool.Parse).Single();
+                }
+            }
+        }
+
+        public bool ShouldVerify { get; }
+
+        public string GetCertificatePassword(ITracer tracer, GitProcess git)
```
kewillford: Looks like this can be private now?

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+            {
+                certificatePassword = this.GetCertificatePassword(tracer, gitProcess);
+
+                if (string.IsNullOrEmpty(certificatePassword))
+                {
+                    tracer.RelatedWarning(
+                        new EventMetadata
+                        {
+                                { "SslCertificate", this.certificatePathOrSubjectCommonName }
+                        },
```
kewillford: Can this use the `metadata` parameter that is passed in?

GVFS/GVFS.Common/Git/GitSsl.cs
kewillford: Yes that is true but if it gets renamed to something extremely different like `CertificateExpirationDate` the name would not match the data so there is a balance there.  I try to use `nameof` by default if the name would just be the name of the property then if the property name ever changes you can decide if it better to keep the old name or allow it to change.  The code does use both hard coded strings and `nameof` so there aren't any hard and fast rules here so I'll leave it up to you.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+                    if (findResults?.Count > 0)
+                    {
+                        LogWithAppropriateLevel(
+                            tracer,
+                            metadata,
+                            findResults.OfType<X509Certificate2>(),
+                            string.Format(
+                                "Found {0} certificates by provided name. Matching DNs: {1}",
+                                findResults.Count,
+                                GetSubjectNameLineForLogging(findResults.OfType<X509Certificate2>())));
```
kewillford: Will you break this single line up and use some temp variables for better readability?  Like
```
string subjects = GetSubjectNameLineForLogging(findResults.OfType<X509Certificate2>());
string logMessage = $"Found {findResults.Count} certificates by provided name. Matching DNs: {subjects}";
LogWithAppropriateLevel(tracer, metadata, findResults.OfType<X509Certificate2>(), logMessage);
```

GVFS/GVFS.Common/Git/GitSsl.cs
kewillford: We can keep it.  It is pretty simple and should not be a performance concern in this code path.

GVFS/GVFS.Common/Git/GitSsl.cs
kewillford: Thanks I see that now.

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+            this.ShouldVerify = true;
+        }
+
+        public GitSsl(IDictionary<string, GitConfigSetting> configSettings) : this()
+        {
+            if (configSettings != null)
+            {
+                if (configSettings.TryGetValue(GitConfigSetting.HttpSslCert, out GitConfigSetting sslCerts))
+                {
+                    this.certificatePathOrSubjectCommonName = sslCerts.Values.Single();
```
wilbaker: `Single` can throw.  Do we have a guarantee at this point that there is exactly one item in `sslCerts.Values`?  If not, do we need to handle the exception (or check that there's exactly one item first)?

(Same comment applies elsewhere in this method where we're using `Single`)

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+                }
+
+                if (configSettings.TryGetValue(GitConfigSetting.HttpSslCertPasswordProtected, out GitConfigSetting isSslCertPasswordProtected))
+                {
+                    this.isCertificatePasswordProtected = isSslCertPasswordProtected.Values.Select(bool.Parse).Single();
```
wilbaker: Similar question for `bool.Parse`, what will happen if it throws?

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+                { "ShouldVerify", this.ShouldVerify }
+            };
+
+            X509Certificate2 result =
+                this.GetCertificateFromFile(tracer, metadata, gitProcess) ??
+                this.GetCertificateFromStore(tracer, metadata);
+
+            if (result == null)
+            {
+                tracer.RelatedError("Certificate {0} not found", this.certificatePathOrSubjectCommonName);
```
wilbaker: Pass in `metadata` here?

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -152,6 +166,25 @@ public bool TryInitializeAndRequireAuth(ITracer tracer, out string errorMessage)
             return false;
         }
 
+        public void SetupSslIfNeeded(ITracer tracer, HttpClientHandler httpClientHandler, GitProcess gitProcess)
```
wilbaker: Naming suggestion: 

When I first read `SetupSslIfNeeded` I thought this method was doing something to setup `this.GitSsl`.  Assuming my understanding is correct it actually looks like the method configures ssl on the `HttpClientHandler`.

If that's the case consider renaming this to something like `ConfigureHttpClientHandlerSSLIfNeeded`.  If my understanding is not correct, please point me in the right direction 😄

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -173,6 +175,7 @@ private MockGitProcess GetGitProcess()
             MockGitProcess gitProcess = new MockGitProcess();
             gitProcess.SetExpectedCommandResult("config gvfs.FunctionalTests.UserName", () => new GitProcess.Result(string.Empty, string.Empty, GitProcess.Result.GenericFailureCode));
             gitProcess.SetExpectedCommandResult("config gvfs.FunctionalTests.Password", () => new GitProcess.Result(string.Empty, string.Empty, GitProcess.Result.GenericFailureCode));
+            gitProcess.SetExpectedCommandResult("config --get-urlmatch http mock://repoUrl", () => new GitProcess.Result($"http.sslCert {CertificatePath}\nhttp.sslCertPasswordProtected true\n\n", string.Empty, GitProcess.Result.SuccessCode));
```
wilbaker: It looks like with this change we won't get any coverage for the scenario where `http.sslCert` is not in the configuration.

Could you parameterize these tests so that we can get coverage for both scenarios? (see `DiffHelperTests` for an example of where we've parameterized unit tests in the past).

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+                        {
+                                { "SslCertificate", this.certificatePathOrSubjectCommonName }
+                        },
+                        "Git config indicates, that certificate is password protected, but retrieved password was null or empty!");
+                }
+
+                metadata.Add("isPasswordSpecified", string.IsNullOrEmpty(certificatePassword));
+            }
+
+            if (File.Exists(this.certificatePathOrSubjectCommonName))
```
wilbaker: For unit testing `PhysicalFileSystem.FileExists` should be used instead of `File.Exists`

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text.RegularExpressions;
+using GVFS.Common.Tracing;
+
+namespace GVFS.Common.Git
+{
+    public class GitSsl
```
wilbaker: Add unit tests for this class?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -164,6 +164,61 @@ public virtual void RevokeCredential(string repoUrl)
...
+                        "Git could not get credentials: " + gitCredentialOutput.Errors,
+                        Keywords.Network | Keywords.Telemetry);
+                    errorMessage = gitCredentialOutput.Errors;
+
+                    return false;
+                }
+
+                password = ParseValue(gitCredentialOutput.Output, "password=");
+
+                bool success = password != null;
```
wilbaker: Are empty passwords allowed?

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -25,6 +28,15 @@ public GitAuthentication(GitProcess git, string repoUrl)
...
             this.git = git;
             this.repoUrl = repoUrl;
+
+            if (git.TryGetConfigUrlMatch("http", this.repoUrl, out Dictionary<string, GitConfigSetting> configSettings))
+            {
+                this.GitSsl = new GitSsl(configSettings);
+            }
+            else
+            {
+                this.GitSsl = new GitSsl();
```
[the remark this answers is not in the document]
turbonaitis: It was used more widely, so having default values was good. We can leave it as null and use null propagation now.

GVFS/GVFS.Common/Git/GitAuthentication.cs
turbonaitis: Sure, will make the name more explicit. Just FYI, this exact name was suggested by one of the other reviewers, so I hope he doesn't come back asking me to revert it.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: If we have this git config
```
git config --get-urlmatch http https://localhost:8081 

http.sslcert testCert2 test
http.sslverify fake news
```

Then when you invoke git, this is what you get
```
git ls-remote https://localhost:8081
fatal: bad numeric config value 'fake news' for 'http.sslverify': invalid unit
```

So we should fail, but the question is, if we should catch, and log a friendlier message.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: Currently, if you have multiple git settings, like
```
http.sslcert = cert1
http.sslcert = cert2
```

Git uses the last one. With that in mind, I'll replace Single() with Last(), although with current git behavior, it should only output one line per config key anyway.

GVFS/GVFS.Common/Git/GitProcess.cs
turbonaitis: This is the same behavior as here:
https://github.com/Microsoft/VFSForGit/blob/7390eba79729141c9042b93750a108cdf4dafc2a/GVFS/GVFS.Common/Git/GitProcess.cs#L200

The idea is, that empty password will be an empty string, not a null.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: Unless you insist - I'd prefer not to. Adding these temp variables will just pollute the code IMO.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: I'd love to, but it's a bit challenging, due to the fact, that certificate store has no interface, that we could mock and I don't think we want to write certificates to the actual store during unit testing, as this might not work on CI servers. Also, the certificates are being validated using native libraries (that is - it's outside our control), so it would be very hard to setup a valid certificate. If you have ideas how to approach this - they would be appreciated.

GVFS/GVFS.Common/Git/GitSsl.cs
wilbaker: > the question is, if we should catch, and log a friendlier message

Correct me if I'm wrong but it looks like:

- `GitSsl` is only created in the `GitAuthentication` constructor
- The `GitAuthentication` constructor is only called from the `Enlistment` constructor.

I've taken a look at some of the paths that hit the `Enlistment` constructor, and it looks like they *only* expect it to throw `InvalidRepoException` (and they expect the exception to contain a helpful error message).

*Example from `Enlistment` constructor*

```
GitProcess.ConfigResult originResult = gitProcess.GetOriginUrl();
if (!originResult.TryParseAsString(out string originUrl, out string error))
{
    throw new InvalidRepoException("Could not get origin url. git error: " + error);
}

if (originUrl == null)
{
    throw new InvalidRepoException("Could not get origin url. remote 'origin' is not configured for this repo.'");
}
```

I think we should maintain that contract (only throwing helpful `InvalidRepoException` messages from the `Enlistment` constructor), and so I think we have a few options:

- Move the code that can throw out of the constructor and into a `TryInitialize` type method
- Update `GitSsl` to catch the exception that can get thrown in its constructor and throw `InvalidRepoException` exceptions instead (with a helpful error message)

We could also catch everything in the `Enlistment` constructor and convert it to a `InvalidRepoException`, but my concern is that message will not have as much context for the user (and these are user facing messages):

*Example of code that handles `InvalidRepoException`*

```
catch (InvalidRepoException e)
{
    this.ReportErrorAndExit(
        "Error: '{0}' is not a valid GVFS enlistment. {1}",
        enlistmentRootPath,
        e.Message);
}
```

GVFS/GVFS.Common/Git/GitSsl.cs
wilbaker: Those are great points.  @kewillford, @jrbriggs, do you have any thoughts on additional tests that could be added to validate `GitSsl`?

My main concern is that we might inadvertently break this feature\functionality if we don't have sufficient coverage of it in our unit and functional tests.

turbonaitis: Absolutely, covering this with tests is something, that I think all of us want. One way to achieve this is to add additional layers of abstraction, like:
```
interface ICertificateLoader
{
   X509Certificate2 LoadCertificateBySubjectName(string subjectName);
}

interface ICertificateVerifier
{
   bool VerifyCertificate(X509Certificate2 certificate);
}
```
Then use them in our GitSsl class and mock those in tests. That's something, that would work, but I'd like to hear your opinion on the solution first, as I didn't see similar code in GVFS code base.

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: Makes sense, will make sure we throw exceptions with messages, that are helpful to the user.

turbonaitis: Looked at uses - it seems CloneVerb only handles generic exceptions, not `InvalidRepoException`:
https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS/CommandLine/CloneVerb.cs#L244-L261

Added handling for this exception and started throwing it from GitSsl constructor, when bool parsing fails.

GVFS/GVFS.Common/Git/GitSsl.cs
kewillford: @wilbaker What are you thoughts?

wilbaker: I'm also not a fan of the multi-line indent, but I would recommend the following:

- Push the formatting work down to the helper method
- Put all of the interesting data into an `EventMetadata`
- Remove `GetSubjectNameLineForLogging `

*Example calling code*

```
LogWithAppropriateLevel(
    tracer, 
    metadata, 
    findResults.OfType<X509Certificate2>(), 
    "Found certificates by provided name");
```

And then move the certificate related formatting into `LogWithAppropriateLevel`

```
private static void LogWithAppropriateLevel(ITracer tracer, EventMetadata metadata, IEnumerable<X509Certificate2> certificates, string logMessage)
{
    int numberOfCertificates = certificates.Count();

    EventMetadata certificateMetadata = new EventMetadata(metadata);
    certificateMetadata.Add("CertificateCount", numberOfCertificates);
    certificateMetadata.Add("CertificateDNs", string.Join(Environment.NewLine, certificates.Select(x => x.Subject)));

    switch (numberOfCertificates)
    {
        case 0:
            tracer.RelatedError(certificateMetadata, logMessage);
            break;
        case 1:
            tracer.RelatedInfo(certificateMetadata, logMessage);
            break;
        default:
            tracer.RelatedWarning(certificateMetadata, logMessage);
            break;
    }
}
```

@kewillford what do you think?

GVFS/GVFS.Common/Git/GitSsl.cs
turbonaitis: So, finally got time to get back to this and pushed unit tests for GitSsl. Would you have a look @wilbaker ?

GVFS/GVFS.Common/GVFS.Common.csproj
```diff
@@ -28,4 +28,10 @@
     </PackageReference>
     <PackageReference Include="System.Security.Cryptography.Algorithms" Version="4.3.0" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="X509Certificates\" />
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="System.Security.Cryptography.X509Certificates" />
+  </ItemGroup>
```
wilbaker: I don't think any of these lines are needed.  I found that with them the build on Windows complains:

```
2019-01-25T18:13:45.7237754Z   C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64\Microsoft.Common.CurrentVersion.targets(2106,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "System.Security.Cryptography.X509Certificates". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\agent\_work\11\s\GVFS\GVFS.Common\GVFS.Common.csproj]
2019-01-25T18:13:45.7238315Z   C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin\amd64\Microsoft.Common.CurrentVersion.targets(2106,5): warning MSB3243: No way to resolve conflict between "System.Security.Cryptography.X509Certificates, Version=4.1.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "System.Security.Cryptography.X509Certificates". Choosing "System.Security.Cryptography.X509Certificates, Version=4.1.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" arbitrarily. [C:\agent\_work\11\s\GVFS\GVFS.Common\GVFS.Common.csproj]
```

And with them removed the build is able to locate the System.Security.Cryptography.X509Certificates assembly.

GVFS/GVFS.Common/X509Certificates/SystemCertificateStore.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public void Dispose()
+        {
+            this.store.Dispose();
+        }
+
+        public X509Certificate2Collection Find(X509FindType findType, string searchString, bool validOnly)
+        {
+            if (!this.store.IsOpen)
```
wilbaker: This line is causing a compilation failure on Windows (where GVFS.Common builds with .NET Framework):

```
5>------ Build started: Project: GVFS.Common, Configuration: Debug x64 ------
5>X509Certificates\SystemCertificateStore.cs(20,29,20,35): error CS1061: 'X509Store' does not contain a definition for 'IsOpen' and no extension method 'IsOpen' accepting a first argument of type 'X509Store' could be found (are you missing a using directive or an assembly reference?)
5>Done building project "GVFS.Common.csproj" -- FAILED.
```

And based on the context Window I saw in the IDE, it looks like `IsOpen` is not available in the .NET Standard API.

image

Is there a different API you can use to perform this check?

(If you don't have a Windows machine available you can use the required PR builds to check if the build on Windows is successful)

turbonaitis: @wilbaker after I fixed these two, there's another issue, that `GVFS.UnitTests.Windows` uses .NET 4.6.1, which does not have `CertificateRequest` class, which I'm using in my unit tests to generate self signed certificates.  Would it be possible to target 4.7.2 in the unit tests project?

wilbaker: > Would it be possible to target 4.7.2 in the unit tests project?

My preference is to leave the unit tests targeting 4.6.1 as that's what the VFS4G executables on Windows are using (and we're not mocking out everything in the .NET Framework).

Rather than retargeting `GVFS.UnitTests.Windows` what would you think of only running these tests as part of `GVFS.UnitTest`?

The cleanest way I've found for doing that is adding the following to `GitSslTests.cs`:

```
#if NETCOREAPP2_1 // .NET Framework 4.6.1 does not support System.Security.Cryptography.X509Certificates.CertificateRequest

// All of the contents of GitSslTests.cs is here and unchanged

#endif
```

I still need to take a look at the new tests you've added, and will post feedback after taking a look.  Thanks!

CC: @kewillford and @nickgra for their thoughts as well.

nickgra: After some digging through history, I agree that `#if NETCOREAPP2_1` is likely the best solution here even though it's a code smell. 

In the past when we started migrating to .NET Core, we had some tests that broke due to changes in how .NET hashes strings. While the problem was actually an issue with the way the test was written, it concerns me that we could write a UT that would behave differently than the actual product code would if they're targeting different platform versions.

In the medium to long term, I plan to move the entire project (Windows and Mac) to .NET Core 3.0 when it comes out (We need C++/CLI support on Windows, which is going to be supported in 3.0), so I can clean this up when that time comes. I think that's a better investment than us moving forward to 4.7.2 right now.

turbonaitis: Sounds great, will do this!

GVFS/GVFS.UnitTests/Common/Git/GitSslTests.cs
```diff
@@ -0,0 +1,266 @@
...
+using System.Security.Cryptography.X509Certificates;
+
+namespace GVFS.UnitTests.Common.Git
+{
+    [TestFixture]
+    public class GitSslTests
+    {
+        public static object[] BoolGitSettings = new[]
+        {
+            GitConfigSetting.HttpSslCertPasswordProtected,
```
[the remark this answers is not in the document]
turbonaitis: Thanks for the tip, didn't know of this interesting NUnit behavior!

GVFS/GVFS.Common/Git/GitProcess.cs
turbonaitis: IMO, this doesn't differ from `TryGetCredentials`, a method you already have. It invokes the same git credential helper, with the same potential of being invoked from a background process. So the way I see it, I'm not introducing anything new here. And yes, when credentials change, while VFS4G is running, and when getting new credentials requires user interaction in terminal (be it for certificate or for http), bad things will happen, but that's something, that is already in the code.

turbonaitis: @wilbaker do you agree with this? Or do you think that my changes introduce some additional risks?

GVFS/GVFS.UnitTests/Mock/ReusableMemoryStream.cs
```diff
@@ -16,6 +16,12 @@ public ReusableMemoryStream(string initialContents)
             this.length = this.contents.Length;
         }
 
+        public ReusableMemoryStream(byte[] initialContnets)
```
kewillford: Spelling

GVFS/GVFS.Common/Git/GitSsl.cs
```diff
@@ -0,0 +1,237 @@
...
+                    this.certificatePathOrSubjectCommonName = sslCerts.Values.Last();
+                }
+
+                this.isCertificatePasswordProtected = SetBoolSettingOrThrow(configSettings, GitConfigSetting.HttpSslCertPasswordProtected, this.isCertificatePasswordProtected);
+
+                this.ShouldVerify = SetBoolSettingOrThrow(configSettings, GitConfigSetting.HttpSslVerify, this.ShouldVerify);
+            }
+        }
+
+        private GitSsl(Func<SystemCertificateStore> certificateStoreFactory, CertificateVerifier certificateVerifier, PhysicalFileSystem fileSystem)
```
kewillford: Parameter is `certificateStoreFactory` and `createCertificateStore` above.  They should probably match.

jrbriggs:
Thanks for the contribution, @turbonaitis !

PR #527, opened 2018-11-16
[RFC] Vnode Cache Design Document

wilbaker:
This PR is for discussion of the Vnode Cache Design Document

### DIFF
--- a/VnodeCacheDesign.md
+++ b/VnodeCacheDesign.md
@@ -1,0 +1,55 @@
+# Vnode Cache Design
+
+## Goals
+
+- Reduce the time required to determine if a file is inside of a virtualization root
+- Reduce the time required to find a file's virtualization root
+- **TBD:** Eliminate the use of file flags to determine if a file is inside of a root
+
+## Design 
+
+### Cached Data
+
+PrjFSKext will cache the following information for each vnode:
+
+- The `vnode_t` itself
+- The vnode's `uint32_t` vid (generation number)
+- `VirtualizationRootHandle` for the vnode (or `RootHandle_None` if the vnode is not in a root).  **NOTE** This assumes that `VirtualizationRootHandle`  are not recycled as VFS4G instances are unmounted\remounted.
+
+The data for each vnode will be stored in a hash table that assumes there will be `desiredvnodes` total vnodes.
+
+### Using the Cache
+
+When PrjFSKext needs to determine the virtualization root of a file (or if a file is inside of a virtualization root at all), it will look up the vnode in its cache.
+
+- If the vnode exists in the cache and its `vid` has not changed, PrjFSKext will use the `VirtualizationRootHandle` in the cache.
+
+- If the vnode exists in the cache and its `vid` has changed, PrjFSKext will lookup its `VirtualizationRootHandle` (using the existing code in PrjFSKext) and update the cache entry.
+
+- If the vnode does not exist in the cache,  PrjFSKext will lookup its `VirtualizationRootHandle` (using the existing code in PrjFSKext) and insert the vnode into the cache.
+
+### Special Case: Renames
+
+*Directory renames* 
+
+Directory renames are a special case because all of the directory's children may need to have their `VirtualizationRootHandle` updated. Rather than recording\tracking all of the parent\child relationships among the vnodes PrjFSKext will simply invalidate the entire cache when it detects that a directory has been renamed in\out of a virtualization root.  If PrjFSKext is unable to determine if the rename was within the same root (e.g. the directory's vnode was recycled) then PrjFSKext will invalidate the cache.
+
+*File renames*
+
+When a file is renamed PrjFSKext will update its `VirtualizationRootHandle` and\or `vid` as needed.
+
+## Future Enhancements
+
+- The vnode cache could track which vnodes have received the `KAUTH_FILEOP_WILL_RENAME` file operation (on Mojave) so that when handling `KAUTH_VNODE_DELETE` the kext can know if the delete vnode operation is for a rename.
+
+## Issues not addressed by the cache
+
+- Improving the speed and reliability of finding a vnode's path
+- Addressing failures to get a vnode's parent (when the vnode is not in the cache)
+- Finding the roots for hard links that span repos (and general handling of hard links that cross repo boundaries)
+
+## Open Questions
+
+- During a vnode operation, when a vnode does not exist in the cache (or has been recycled since being added to the cache) should PrjFSKext check `FileFlags_IsInVirtualizationRoot` before it walks up the tree looking for a root?
+- Will PrjFSKext invalidate the cache too frequently if it does so on every directory rename?
+- What is the timeline for allowing breaking changes to existing repos (as we refine the approach that we're taking)?


VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Using the Cache
+
+When PrjFSKext needs to determine the virtualization root of a file (or if a file is inside of a virtualization root at all), it will look up the vnode in its cache.
+
+- If the vnode exists in the cache and its `vid` has not changed, PrjFSKext will use the `VirtualizationRootHandle` in the cache.
+
+- If the vnode exists in the cache and its `vid` has changed, PrjFSKext will lookup its `VirtualizationRootHandle` (using the existing code in PrjFSKext) and update the cache entry.
+
+- If the vnode does not exist in the cache,  PrjFSKext will lookup its `VirtualizationRootHandle` (using the existing code in PrjFSKext) and insert the vnode into the cache.
```
sanoursa: "does not exist" and "exists and its vid has changed" are equivalent, correct? I think this bullet and the one above are indistinguishable from each other.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
+# Vnode Cache Design
+
+## Goals
+
+- Reduce the time required to determine if a file is inside of a virtualization root
+- Reduce the time required to find a file's virtualization root
+- **TBD:** Eliminate the use of file flags to determine if a file is inside of a root
```
sanoursa: I think getting rid of the `IsInVirtualizationRoot` flag should be a primary goal of this design. That flag exists only as a way to short-circuit our logic for paths outside of any virtualization root, but it is flaky for several reasons. The main benefit of this cache is that we can answer the question cheaply and reliably.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+- The vnode cache could track which vnodes have received the `KAUTH_FILEOP_WILL_RENAME` file operation (on Mojave) so that when handling `KAUTH_VNODE_DELETE` the kext can know if the delete vnode operation is for a rename.
+
+## Issues not addressed by the cache
+
+- Improving the speed and reliability of finding a vnode's path
+- Addressing failures to get a vnode's parent (when the vnode is not in the cache)
+- Finding the roots for hard links that span repos (and general handling of hard links that cross repo boundaries)
+
+## Open Questions
```
sanoursa: My biggest open question is: what is the design of the hash table? As we've discussed, we don't need a full-fledged hash table since we want to store every vnode, but we do need to at least implement a hash function and index collisions. What does that look like?

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+- During a vnode operation, when a vnode does not exist in the cache (or has been recycled since being added to the cache) should PrjFSKext check `FileFlags_IsInVirtualizationRoot` before it walks up the tree looking for a root?
```
sanoursa: Again, "does not exist" and "has been recycled" are the same thing. I don't think we need to keep repeating that, unless you think there is some subtle difference between them?

sanoursa: To your question: I think we should eliminate that flag entirely.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+- Will PrjFSKext invalidate the cache too frequently if it does so on every directory rename?
```
sanoursa: Related question: how much of a perf hit do we see after clearing the cache?

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Design 
+
+### Cached Data
+
+PrjFSKext will cache the following information for each vnode:
+
+- The `vnode_t` itself
+- The vnode's `uint32_t` vid (generation number)
+- `VirtualizationRootHandle` for the vnode (or `RootHandle_None` if the vnode is not in a root).  **NOTE** This assumes that `VirtualizationRootHandle`  are not recycled as VFS4G instances are unmounted\remounted.
```
sanoursa: I think that is a safe assumption. A "handle" here is really just an index into the array. We should make sure that we only ever expand that array, and never move entries around. That is its current behavior, but some comments / unit tests would help reinforce that.

VnodeCacheDesign.md
sanoursa: Related: what thread-safety guarantees do we need?

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+When a file is renamed PrjFSKext will update its `VirtualizationRootHandle` and\or `vid` as needed.
+
+## Future Enhancements
+
+- The vnode cache could track which vnodes have received the `KAUTH_FILEOP_WILL_RENAME` file operation (on Mojave) so that when handling `KAUTH_VNODE_DELETE` the kext can know if the delete vnode operation is for a rename.
+
+## Issues not addressed by the cache
+
+- Improving the speed and reliability of finding a vnode's path
```
sanoursa: Will that be in a separate design?

VnodeCacheDesign.md
wilbaker: Great questions:

> we do need to at least implement a hash function 

Based on what I've read for pointers it's common to simply use the value of the pointer, or the value right-shifted by 2 to 3 bits.  I can test out some different options and see how they perform. 

> and index collisions

Index collisions will be handled by adding a new entry to the list in each bucket.  We can compare `vnode_t` values to know when we've found a match (when walking the list).

> what thread-safety guarantees do we need

Another great question, I'll add more specifics in this area to the doc.

VnodeCacheDesign.md
wilbaker: As mentioned earlier the difference is whether or not PrjFSKext will need to insert a new entry into the hash table.  It's probably not worth calling out that difference here.

VnodeCacheDesign.md
wilbaker: That's a great question, I think we'll need to build it and see :)

I will add your question to the list.

VnodeCacheDesign.md
wilbaker: > I think getting rid of the IsInVirtualizationRoot flag should be a primary goal of this design. That flag exists only as a way to short-circuit our logic for paths outside of any virtualization root, but it is flaky for several reasons. The main benefit of this cache is that we can answer the question cheaply and reliably.

Agree on all points.

I think the open question is how frequently we'll miss in our cache.  If the miss % is too high then we'll end up slowing everything down walking up the vnode hierarchy.

VnodeCacheDesign.md
sanoursa: > Index collisions will be handled by adding a new entry to the list in each bucket. 

That assumes that you're implementing buckets. Certainly a valid hash table design choice, but seems to be more complexity than we'd need here, and also potentially wasteful on memory.

We should at least consider other options (linear probing? secondary hash?). Perhaps we'll find that the pointer value hashes cluster more than we'd like and we do want the buckets; in that case, we may not want to preallocate the `desiredvnodes` entries up front, and then have the complexity of needing to reallocate in some scenarios.

All of that is to say - I think we need more design detail on the hash table implementation :-)

jrbriggs: >  Certainly a valid hash table design choice, but seems to be more complexity than we'd need here, and also potentially wasteful on memory.

Linear probing has some big advantages: no memory allocations and locality of reference in a collision scenario. In this application it's probably a great choice.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Cached Data
+
+PrjFSKext will cache the following information for each vnode:
+
+- The `vnode_t` itself
+- The vnode's `uint32_t` vid (generation number)
+- `VirtualizationRootHandle` for the vnode (or `RootHandle_None` if the vnode is not in a root).  **NOTE** This assumes that `VirtualizationRootHandle`  are not recycled as VFS4G instances are unmounted\remounted.
+
+The data for each vnode will be stored in a hash table that assumes there will be `desiredvnodes` total vnodes.
```
jrbriggs: It wasn't clear to me what the key is and what the value is in this cache. I'm inferring that this is a cache from vnode to vid and handle, but let's clarify that explicitly. Especially if I'm wrong :)

VnodeCacheDesign.md
jrbriggs: This distinction was helpful to me. Let's keep it, but port some of your last comment into the doc itself.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Future Enhancements
+
+- The vnode cache could track which vnodes have received the `KAUTH_FILEOP_WILL_RENAME` file operation (on Mojave) so that when handling `KAUTH_VNODE_DELETE` the kext can know if the delete vnode operation is for a rename.
+
+## Issues not addressed by the cache
+
+- Improving the speed and reliability of finding a vnode's path
+- Addressing failures to get a vnode's parent (when the vnode is not in the cache)
+- Finding the roots for hard links that span repos (and general handling of hard links that cross repo boundaries)
```
jrbriggs: Should we disallow a hardlink across virtualization roots? That seems unnecessary and fraught with complexity.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Open Questions
+
+- During a vnode operation, when a vnode does not exist in the cache (or has been recycled since being added to the cache) should PrjFSKext check `FileFlags_IsInVirtualizationRoot` before it walks up the tree looking for a root?
+- Will PrjFSKext invalidate the cache too frequently if it does so on every directory rename?
+- What is the timeline for allowing breaking changes to existing repos (as we refine the approach that we're taking)?
```
jrbriggs: I'm unclear how a cache in the kext is a breaking change unless it's persisted somewhere, which I didn't see called out. Can you elaborate?

VnodeCacheDesign.md
pmj: I'd definitely avoid the linked list approach here, we want to avoid memory allocations while holding the hash table lock, which makes the insertion code very complex. Linked lists are also awful for locality of reference/cache behaviour. So I'd almost certainly go with open addressing if we're talking hash tables. Linear probing ought to be good enough if the hash function is decent.

Speaking of hash functions: don't forget they're not arbitrary pointers, as vnodes use the zone allocator: If I'm reading the output of `zprint` correctly, `vnode`s are 248 bytes on macOS 10.13.6, and allocated in 8KiB slabs. (33 vnodes per slab) So each vnode pointer will have one of only 33 values in the low 13 bits (uniformly distributed), a whole bunch more non-uniform entropy in bits 13 and up, at least to a point; there will probably be very little entropy in the upper ~32 bits. Anyway, this sort of pattern is poison for the identity hash function if using power-of-2 capacities, but it really shouldn't be hard to come up with a hash function that works well. A good start would be to collect all vnode pointer values we see during a series of intensive test runs (with reboots) and dump them to a file we can then use as input for offline tests of hash functions & tables.

This is assuming we go with a hash table. Given the bit pattern, we could also go with a radix/prefix tree. Those are especially good if memory use is dominated by the key; in our case, each table entry is fairly large compared to the key, so the memory saving from not storing the vnode pointer itself in each table entry is negligible. On the other hand, we don't have to contend with load factors, collision resolution, etc.

Given how critical this data structure is to performance, it might be worth trying a few different options.

VnodeCacheDesign.md
pmj: We'll probably want to recycle indices if a root is deleted/destroyed. However, in that case we should be able to ensure the cache does not refer to that index anymore.

VnodeCacheDesign.md
pmj: Thinking about it some more, I guess the absence of the file flag will be a breaking change, but initially, I assume we'll still be writing it, it'll just be ignored.

sanoursa: Yep it's due to the fact that the cache allows us to change what we're persisting on placeholder files. We're also considering a separate change to store each placeholder file's canonical relative path in its xattr (i.e. the path that we created it with) and avoiding any reliance on resolving paths in the kext.

Breaking changes are a non issue at the moment since we have no active users. We'll definitely need to decide when to set that expectation, but I don't think it needs to be a day 1 requirement necessarily. For reference, we had real users on Windows for at least 6 months before we started guaranteeing that a repo could be used after upgrade. In the beginning, it was more important, and valuable to our users, to be able to settle on the right design without being constrained by prior data shapes. I think we'll settle much more quickly for the Mac client though since the VFSForGit data shapes are already stable, so we just need to let the placeholder file formats stabilize.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+The data for each vnode will be stored in a hash table that assumes there will be `desiredvnodes` total vnodes.
+
+### Using the Cache
+
+When PrjFSKext needs to determine the virtualization root of a file (or if a file is inside of a virtualization root at all), it will look up the vnode in its cache.
+
+- If the vnode exists in the cache and its `vid` has not changed, PrjFSKext will use the `VirtualizationRootHandle` in the cache.
+
+- If the vnode exists in the cache and its `vid` has changed, PrjFSKext will lookup its `VirtualizationRootHandle` (using the existing code in PrjFSKext) and update the cache entry.
```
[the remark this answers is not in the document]
wilbaker: @pmj that's a good point, I'll update the document.

There is a corner case here that I don't have a good answer for.  What should the kext do if `vnode_getparent` fails (returns `NULL`) while walking up the tree.

Should we simply consider the vnode as not being in any of our roots?

I don't have a good understanding of under what conditions `vnode_getparent` might return `NULL`, any ideas?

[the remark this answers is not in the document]
pmj: I've not seen `vnode_getparent()` return `NULL` myself yet, and I'm not entirely clear on the circumstances, although based on my reading of the source code I'm certainly willing to believe it happens. I've added logging for this case in the #555 branch, so I'll be keeping an eye out for this case.

VnodeCacheDesign.md
wilbaker: > Additionally, are we interested in removing the need for the 'empty' flag too, or are we considering that to be out of scope for now?

@pmj I think it would be great to eliminate the "empty" flag too, but I think that's out of scope for this first iteration of the vnode cache.

Out of curiosity, did you have any idea(s) in mind for how the vnode cache could address this?  

Even if it's out of scope right in the initial implementation I'd like to avoid going down a path that makes it harder to add later.

VnodeCacheDesign.md
wilbaker: I will update the document to clarify this point, thanks!

VnodeCacheDesign.md
wilbaker: > I'm not sure we can prevent them with kauth vnode listeners?

This was my thought as well, I don't believe that this is something that we can prevent.  

For what it's worth, hard links can cause problems for VFS4G on Windows as well:
- VFS4G will not receive notifications when a file is modified via a hard link that's outside the repo
- If a file inside the repo has a hard link that's also within in the same repo, VFS4G will only receive file modified notifications for the path that was used to modify the file (and not for both).

I'd expect the behavior on Mac to be *slightly* different, as we don't guarantee that the path used to make the modification will be delivered to VFS4G (when a file has hard links).

@jrbriggs perhaps as a starting point our message can be that `git status` results may be inconsistent when repos contain hard links.  Thoughts?

VnodeCacheDesign.md
wilbaker: Yep @sanoursa is correct, and I will remove this question from the document (strictly speaking it's not related to the vnode cache).

The change to remove the flag is not, in and of itself, a breaking change.  However, if we make that change and then discover unforeseen circumstances that require bringing back the flag we'd have a breaking change.

I raised this question in the document because it felt like we're starting down a path that could lead to a breaking change, and I wanted to make sure everyone was on the same page as to how we'd handle that.  I agree with @sanoursa's point that we should make sure we've settled on the right design before locking down the format.

VnodeCacheDesign.md
wilbaker: > If the position in the prefix tree already encodes the vnode pointer, we can more or less halve the memory cost.

What pieces of data were you thinking would be recorded to encode portions of the pointer address while moving down the tree?  In other words, how would each node record what chunks of the pointer's address are encoded by each of its children (or how would the children themselves record this information)?

The options that I've come up with have all been pretty inefficient.

VnodeCacheDesign.md
pmj: We can persist the empty state in the xattr, and cache that information in the vnode cache as a simple flag when we end up reading the xattr. We can invalidate the cache when we receive an xattr write check in the vnode listener, and also after receiving a success response for a hydrate file provider message.

VnodeCacheDesign.md
pmj: This is a very quick & dirty sketch of the sort of thing I was thinking of: https://gist.github.com/pmj/a25abf621b830ba9d7f7dee6d6959d15

There are no doubt some optimisations to be made, and the insertion logic is, uh, *left as an exercise to the reader*, but I **think** it should be possible to turn it into something practical. Whether it's faster than a hash table remains to be seen.

VnodeCacheDesign.md
wilbaker: @pmj thanks for the suggestion!  I'll include the details you've suggested in the doc

pmj: FWIW, I believe @sanoursa did not like this approach when I first suggested it back in ~June, but then he was pretty lukewarm on the vnode cache idea in general. 😃 If I remember correctly, the core concern was that we'd be tracking state in more than one place and the possibility that these different versions of the same information might go out of sync.

VnodeCacheDesign.md
wilbaker: > This is a very quick & dirty sketch of the sort of thing I was thinking of: 

Thanks!

> Whether the approach is faster than a hash table remains to be seen.

Agreed, it's looking like the big takeaway is we'll need to implement and perf test a few options to see what's best.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,55 @@
...
+
+- If the vnode does not exist in the cache,  PrjFSKext will lookup its `VirtualizationRootHandle` (using the existing code in PrjFSKext) and insert the vnode into the cache.
+
+### Special Case: Renames
+
+*Directory renames* 
+
+Directory renames are a special case because all of the directory's children may need to have their `VirtualizationRootHandle` updated. Rather than recording\tracking all of the parent\child relationships among the vnodes PrjFSKext will simply invalidate the entire cache when it detects that a directory has been renamed in\out of a virtualization root.  If PrjFSKext is unable to determine if the rename was within the same root (e.g. the directory's vnode was recycled) then PrjFSKext will invalidate the cache.
+
```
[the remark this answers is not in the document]
wilbaker: > Only invalidate cache entries which match the root handle from which we're renaming. (When moving a directory from inside root 7 to not-in-root, I don't think the cached information suddenly becomes incorrect for anything other than vnodes below root 7?)

Good point, it should be sufficient to only invalidate the entries with the root we're moving from.

> We can combine the 2 optimisations by storing the last generation before which the cache was reset for a particular root handle value. On lookup, we check whether the stored generation is equal or newer than the last generation when the cached root handle's zone was invalidated.

The generation id is a great idea, I will update the document to include it.  I think a per-root generation ID is a nice optimization as well.

wilbaker:
@jrbriggs @pmj @sanoursa thanks for the great feedback so far.  

I just pushed a commit with changes your feedback, please let me know if you have any additional thoughts\questions on the design.

VnodeCacheDesign.md
```diff
@@ -2,37 +2,83 @@
...
+The data for each vnode will be stored in a fixed-size hash table with the `vnode_t` pointer serving as the key.
+
+#### Collision Handling
+
+Index collisions will be handled using open addressing.  The exact technique (e.g. linear probing, double hashing, etc.) will be determined once we have sample sets of `vnode_t` pointers and can evaluate the quality of our hash function.
+
+#### Size
+
+The hash table will be a fixed size of length `desiredvnodes`.
```
pmj: `desiredvnodes` is a target value; if I understand the code correctly, there can actually be a few more than that, as the recycling vnodes aren't counted towards the total. So it's not quite a fixed size, but "not much larger than `desiredvnodes`".

VnodeCacheDesign.md
```diff
@@ -2,37 +2,83 @@
...
 ### Special Case: Renames
 
 *Directory renames* 
 
-Directory renames are a special case because all of the directory's children may need to have their `VirtualizationRootHandle` updated. Rather than recording\tracking all of the parent\child relationships among the vnodes PrjFSKext will simply invalidate the entire cache when it detects that a directory has been renamed in\out of a virtualization root.  If PrjFSKext is unable to determine if the rename was within the same root (e.g. the directory's vnode was recycled) then PrjFSKext will invalidate the cache.
+Directory renames are a special case because all of the directory's children may need to have their `VirtualizationRootHandle` updated. Rather than recording\tracking all of the parent\child relationships among the vnodes PrjFSKext will (potentially) invalidate a portion of the vnode cache:
+
+- If the rename is within the same virtualization root: All entries in the cache are still valid.
+
+- If the rename source and target are outside of any virutalization roots: All entries in the cache are still valid.
```
pmj: very minor: typo in "virtualization"

VnodeCacheDesign.md
```diff
@@ -0,0 +1,101 @@
...
+
+- If the rename moves a directory out of a virtualization root: Increase that root's `vrgid` (thereby invalidating all entries in the cache for that root)
+
+- If the rename moves a directory from outside of any virtualization root to inside a virtualization root: Increase the global "no root" `vrgid` (thereby invalidating all entries in the cache that are not inside of a virtualization root)
+
+- If PrjFSKext is unable to determine if the rename was within the same root: Increase the `vrgid` for that vnode's virtualization root (or the "no root" `vrgid` if it was not inside a root)
+
+*File renames*
+
+When a file is renamed PrjFSKext will update its `VirtualizationRootHandle` and\or `vid` as needed.
```
pmj: I can't think of any connection between renames and `vid` change - what did you have in mind here?

VnodeCacheDesign.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Design 
+
+### Cached Data
+
+PrjFSKext will cache the following information for each `vnode_t` pointer:
+
+- The `vnode_t` itself
+- The vnode's `uint32_t` vid (vnode generation number)
+- A `uint32_t` virtualization\cache generation number (hereby referred to as `vrgid`).  If a cache entry's `vrgid` does not match its roots it means that all entries for that root have been invalidated and the `VirtualizationRootHandle` in the cache entry is no longer valid.  There will also be a global `vrgid`  for all vnodes that are not inside of any virtualization roots.
```
pmj: I would perhaps not nail down the data type for this just yet; it looks like a uint32_t will push us over another 8-byte boundary, so instead of 4 bytes, it'll cost us 10 bytes per entry. A `uint16_t` might be good enough for this purpose, and would fit in the padding left by the rest of the fields.

VnodeCacheDesign.md
wilbaker: > Presumably we'll use RootHandle_Indeterminate as the "unknown" (invalidated cache) state.

When portions of the cache are invalidated we won't need to update the `VirtualizationRootHandle` member of any of the entries in the cache.  We'll know they are invalidated because their `vrgid` won't match the current `vrgid` for their `VirtualizationRootHandle`.

VnodeCacheDesign.md
wilbaker: > there can actually be a few more than that, as the recycling vnodes aren't counted towards the total

Good point.  The "fixed size" comment was because we're looking to avoid the cost of growing the cache if it there's not enough space.  It sounds like we might instead want to pick a size some percentage larger than `desiredvnodes` (e.g. `desiredvnodes * 1.25`).

VnodeCacheDesign.md
wilbaker: @pmj I wasn't sure if renames would change a `vnode`'s `vid.  Based on your question though it makes me think that's not the case, and I'll remove the mention of `vid`.

VnodeCacheDesign.md
```diff
@@ -0,0 +1,101 @@
...
+The data for each vnode will be stored in a fixed-size hash table with the `vnode_t` pointer serving as the key.
+
+#### Collision Handling
+
+Index collisions will be handled using open addressing.  The exact technique (e.g. linear probing, double hashing, etc.) will be determined once we have sample sets of `vnode_t` pointers and can evaluate the quality of our hash function.
+
+#### Size
+
+The hash table will be a fixed size that is some percent larger than the value of `desiredvnodes` (exact percentage is still TBD).
+
```
jeschu1: The Collision Handing, Table Size, and Hash Key are pretty key pieces.  Are there similar drivers out there we can follow for guidance?

wilbaker: Not that I'm aware of.  @nickgra, @sanoursa, and I recently met with two other teams building similar kexts and neither are caching all of the vnodes.

VnodeCacheDesign.md
pmj: Yeah, we'll need to try and see how it behaves in practice; I think we should at least support resizing in case it becomes necessary (e.g. as a result of some other kext changing `desiredvnodes`), even if it introduces a brief performance stumble.

wilbaker: @pmj good point, I'll update the doc

PR #530, opened 2018-11-16
Fix 'Profiling(Release)' build configuration after GVFS.Installer.Mac inclusion

nickgra:
This configuration fails today as it can't find the native code in the right location. While we investigate the perf overhead of the Profiling configuration (this whole configuration could be rolled into the Release configuration if the overhead isn't large), unblock the build of this configuration by also building 'Release' flavored native code (The profiling test infrastructure will pick up the build output directly, not from the installer).


nickgra:
Abandoning in favor of handling this in the build definition

PR #531, opened 2018-11-19
[RFC] Add new repair job for Git refs/heads using the reflog

mjcheetham:
Some users experence torn writes by Git after a crash to their branch refs (.git/refs/heads/**/*) leaving them unable to interact with their branches.

We're already fixing the user's HEAD ref if that becomes corrupt by reading the last location from the relevant reflog. This change generalises the fix to enable fixing of any ref using the corresponding reflog.

**Changes:**
- Generalise the work done by `GitHeadRepairJob` into an abstract job `GitRefsRepairJob`.
- Refactor the exising `GitHeadRepairJob` job to extend the abstract job.
- Introduce a new `GitHeadsRepairJob` that attempts to fix refs/heads/**/*.

**Questions:**
1. The only refs we'll now attempt to fix are HEAD, and refs/heads/**/*, not remotes (or tags).
Do we want to try fixing remote refs? Can't these be re-fetched from the remote? (what about tags?)

**TODO:**
- [x] ~Add functional test of the new repair job for corrupt refs/heads/**/*~ Done.

### DIFF
--- a/GVFS/GVFS.Common/Paths.Shared.cs
+++ b/GVFS/GVFS.Common/Paths.Shared.cs
@@ -68,5 +68,24 @@ namespace GVFS.Common
         {
             return path.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
         }
+
+        public static string MakeRelative(string rootPath, string path)
+        {
+            // Ensure trailing slash on the rootPath
+            if (rootPath.Length > 0 && rootPath.Last() != Path.DirectorySeparatorChar)
+            {
+                rootPath += Path.DirectorySeparatorChar;
+            }
+
+            Uri rootUri = new Uri(rootPath, UriKind.Absolute);
+            Uri pathUri = new Uri(path, UriKind.Absolute);
+
+            Uri relativeUri = rootUri.MakeRelativeUri(pathUri);
+
+            string relativePath = relativeUri.ToString();
+
+            // Convert to native path separators
+            return relativePath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
+        }
     }
 }
--- a/GVFS/GVFS.Common/Paths.Shared.cs
+++ b/GVFS/GVFS.Common/Paths.Shared.cs
@@ -69,23 +69,54 @@ namespace GVFS.Common
             return path.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
         }
 
-        public static string MakeRelative(string rootPath, string path)
+        public static string GetRelativePath(string relativeTo, string path)
         {
-            // Ensure trailing slash on the rootPath
-            if (rootPath.Length > 0 && rootPath.Last() != Path.DirectorySeparatorChar)
+            if (!Path.IsPathRooted(relativeTo))
             {
-                rootPath += Path.DirectorySeparatorChar;
+                throw new ArgumentException("Path must be absolute.", nameof(relativeTo));
             }
 
-            Uri rootUri = new Uri(rootPath, UriKind.Absolute);
+            if (!Path.IsPathRooted(path))
+            {
+                throw new ArgumentException("Path must be absolute.", nameof(path));
+            }
+
+            // Normalize paths
+            relativeTo = Path.GetFullPath(relativeTo);
+            path = Path.GetFullPath(path);
+
+            // Handle calculation of relative paths to self
+            if (StringComparer.OrdinalIgnoreCase.Equals(relativeTo, path))
+            {
+                return ".";
+            }
+
+            // For UNIX style paths we must prepend the "file://" scheme explicitly to create a System.Uri
+            const char UnixPathRoot = '/';
+            if (relativeTo.Length > 0 && relativeTo[0] == UnixPathRoot)
+            {
+                relativeTo = $"{Uri.UriSchemeFile}://{relativeTo}";
+            }
+
+            if (path.Length > 0 && path[0] == UnixPathRoot)
+            {
+                path = $"{Uri.UriSchemeFile}://{path}";
+            }
+
+            Uri relativeToUri = new Uri(relativeTo, UriKind.Absolute);
             Uri pathUri = new Uri(path, UriKind.Absolute);
 
-            Uri relativeUri = rootUri.MakeRelativeUri(pathUri);
+            Uri relativeUri = relativeToUri.MakeRelativeUri(pathUri);
 
             string relativePath = relativeUri.ToString();
 
+            // Decode/un-escape characters (e.g, "%20" back to " ")
+            relativePath = Uri.UnescapeDataString(relativePath);
+
             // Convert to native path separators
-            return relativePath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
+            relativePath = relativePath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
+
+            return relativePath;
         }
     }
 }
--- a/GVFS/GVFS.Common/Paths.Shared.cs
+++ b/GVFS/GVFS.Common/Paths.Shared.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
@@ -69,54 +70,27 @@ namespace GVFS.Common
             return path.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
         }
 
-        public static string GetRelativePath(string relativeTo, string path)
+        public static IEnumerable<string> GetFilesRecursive(string rootDirectory)
         {
-            if (!Path.IsPathRooted(relativeTo))
+            return GetFilesRecursive(new DirectoryInfo(rootDirectory), string.Empty);
+        }
+
+        private static IEnumerable<string> GetFilesRecursive(DirectoryInfo directory, string prefix)
+        {
+            foreach (FileInfo file in directory.GetFiles())
             {
-                throw new ArgumentException("Path must be absolute.", nameof(relativeTo));
+                yield return Path.Combine(prefix, file.Name);
             }
 
-            if (!Path.IsPathRooted(path))
+            foreach (DirectoryInfo childDirectory in directory.GetDirectories())
             {
-                throw new ArgumentException("Path must be absolute.", nameof(path));
+                string childPrefix = Path.Combine(prefix, childDirectory.Name);
+
+                foreach (string childFileName in GetFilesRecursive(childDirectory, childPrefix))
+                {
+                    yield return childFileName;
+                }
             }
-
-            // Normalize paths
-            relativeTo = Path.GetFullPath(relativeTo);
-            path = Path.GetFullPath(path);
-
-            // Handle calculation of relative paths to self
-            if (StringComparer.OrdinalIgnoreCase.Equals(relativeTo, path))
-            {
-                return ".";
-            }
-
-            // For UNIX style paths we must prepend the "file://" scheme explicitly to create a System.Uri
-            const char UnixPathRoot = '/';
-            if (relativeTo.Length > 0 && relativeTo[0] == UnixPathRoot)
-            {
-                relativeTo = $"{Uri.UriSchemeFile}://{relativeTo}";
-            }
-
-            if (path.Length > 0 && path[0] == UnixPathRoot)
-            {
-                path = $"{Uri.UriSchemeFile}://{path}";
-            }
-
-            Uri relativeToUri = new Uri(relativeTo, UriKind.Absolute);
-            Uri pathUri = new Uri(path, UriKind.Absolute);
-
-            Uri relativeUri = relativeToUri.MakeRelativeUri(pathUri);
-
-            string relativePath = relativeUri.ToString();
-
-            // Decode/un-escape characters (e.g, "%20" back to " ")
-            relativePath = Uri.UnescapeDataString(relativePath);
-
-            // Convert to native path separators
-            relativePath = relativePath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
-
-            return relativePath;
         }
     }
 }
--- a/GVFS/GVFS.Common/Paths.Shared.cs
+++ b/GVFS/GVFS.Common/Paths.Shared.cs
@@ -69,28 +69,5 @@ namespace GVFS.Common
         {
             return path.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
         }
-
-        public static IEnumerable<string> GetFilesRecursive(string rootDirectory)
-        {
-            return GetFilesRecursive(new DirectoryInfo(rootDirectory), string.Empty);
-        }
-
-        private static IEnumerable<string> GetFilesRecursive(DirectoryInfo directory, string prefix)
-        {
-            foreach (FileInfo file in directory.GetFiles())
-            {
-                yield return Path.Combine(prefix, file.Name);
-            }
-
-            foreach (DirectoryInfo childDirectory in directory.GetDirectories())
-            {
-                string childPrefix = Path.Combine(prefix, childDirectory.Name);
-
-                foreach (string childFileName in GetFilesRecursive(childDirectory, childPrefix))
-                {
-                    yield return childFileName;
-                }
-            }
-        }
     }
 }
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
@@ -50,18 +50,19 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
         }
 
         [TestCase]
-        public void FixesCorruptHeadsSha()
+        public void FixesCorruptRefsHeadsSha()
         {
             const string InvalidRefContents = "0000";
 
             string branch1Name = "testBranch1";
             string branch2Name = "test/branch/number2";
-            string branchStartSha = this.Enlistment.GetRevParse("HEAD");
+            string branchStartSha = this.GetRevParse("HEAD");
 
-            this.Enlistment.CreateBranch(branch1Name, branchStartSha);
-            this.Enlistment.CreateBranch(branch2Name, branchStartSha);
+            this.CreateBranch(branch1Name, branchStartSha);
+            this.CreateBranch(branch2Name, branchStartSha);
             this.Enlistment.UnmountGVFS();
 
+            // Write invalid data in the reference
             string branch1RefFilePath = Path.Combine(this.Enlistment.RepoRoot, ".git", "refs", "heads", branch1Name);
             string branch2RefFilePath = Path.Combine(this.Enlistment.RepoRoot, ".git", "refs", "heads", branch2Name);
             File.WriteAllText(branch1RefFilePath, InvalidRefContents);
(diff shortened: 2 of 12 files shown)


GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
```diff
@@ -0,0 +1,163 @@
...
+        protected GitRefsRepairJob(ITracer tracer, TextWriter output, GVFSEnlistment enlistment) 
+            : base(tracer, output, enlistment)
+        {
+        }
+
+        public override IssueType HasIssue(List<string> messages)
+        {
+            int numBadRefs = 0;
+
+            foreach (string @ref in this.GetRefs())
```
kewillford: What is the `@` for?

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            return numBadRefs > 0 ? IssueType.Fixable : IssueType.None;
+        }
+
+        /// <summary>
+        /// Fixes the refs using the reflog to find the last SHA.
+        /// </summary>
+        public override FixResult TryFixIssues(List<string> messages)
+        {
+            IEnumerable<string> badRefs = this.GetRefs().Where(x => !this.TryParseRef(x));
```
kewillford: Why did you use a `foreach` above and then use Linq here?  Can we be use the same in both?

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
```diff
@@ -0,0 +1,163 @@
...
+            {
+                refContents = File.ReadAllText(refPath).Trim();
+            }
+            catch (IOException ex)
+            {
+                messages?.Add($"IOException while reading '{refPath}': {ex.Message}");
+                return false;
+            }
+
+            if (this.IsValidRefContents(fullSymbolicRef, refContents))
```
kewillford: How are we resolving symbolic refs that have a symbolic ref?  For example files in `refs/heads/`  can have a ref in them and don't have to be a sha by running `git symbolic-ref refs/heads/branch2 refs/heads/branch1`

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
```diff
@@ -0,0 +1,163 @@
...
+                refLog = null;
+                error = $"IOException while reading reflog '{fullSymbolicRef}': " + ex.Message;
+                return false;
+            }
+
+            error = null;
+            return true;
+        }
+
+        private bool TryParseRef(string fullSymbolicRef, List<string> messages = null)
```
kewillford: Rename `messages` to `errorMessages` and maybe it should just be and `out string errorMessage` since it doesn't look like we are adding multiple messages?

kewillford: Looks like this could be applied to multiple places in these changes.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
```diff
@@ -49,6 +49,42 @@ public void FixesCorruptHeadSymRef()
             this.RepairWithConfirmShouldFix();
         }
 
+        [TestCase]
+        public void FixesCorruptHeadsSha()
```
kewillford: Not sure if we have any but will you add/make sure there are some tests where the ref is good and doesn't need fixed and where the reflog file for the ref is missing.

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
mjcheetham: `ref` is a C# keyword. Prefixing with @ escapes the keyword. Could alternatively pick a different name for this variable? `reference` perhaps?

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
mjcheetham: The `abstract class RepairJob` defines the `HasIssue` and `TryFixIssue` methods and gives you a `List<string>` for all messages to be output as part of the job execution (not just errors).

Keeping the list seems simpler as the caller can just pass in its reference to `messages` to all the called methods. It also avoids the caller(s) having to do a null check for the returned string, then an add to it's list if non-null.

Also I was trying to keep the original implementation of `GitHeadRepairJob` as much as possible since it was pre-existing. 😄

GVFS/GVFS.Common/Paths.Shared.cs
```diff
@@ -68,5 +68,24 @@ public static string ConvertPathToGitFormat(string path)
...
+            // Ensure trailing slash on the rootPath
+            if (rootPath.Length > 0 && rootPath.Last() != Path.DirectorySeparatorChar)
+            {
+                rootPath += Path.DirectorySeparatorChar;
+            }
+
+            Uri rootUri = new Uri(rootPath, UriKind.Absolute);
+            Uri pathUri = new Uri(path, UriKind.Absolute);
+
+            Uri relativeUri = rootUri.MakeRelativeUri(pathUri);
```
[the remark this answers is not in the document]
mjcheetham: There is no 'make a relative path' in the BCL, which is a shame. The closest you can get without having to implement your own is using `System.Uri`. I was surprised the VFS4G codebase doesn't already have such a utility yet (unless I'm missing something?).

You're right about needing to unescape spaces from Uri.ToString however.. I'll look at replacing this and doing the relative path calc manually (with tests!).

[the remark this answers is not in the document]
mjcheetham: \* Grumble \* 😢 

Covering all the edge cases here is actually trickier that first imagined..

It also turns out there _is_ a [BCL API for getting a relative path](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=netcore-2.1#applies-to).. but only in .NET Core 2.0 onwards, not in .NET Standard or Framework. 😭 (this project is multi-targeted)

I've updated the current code (using `System.Uri`) to unescape spaces correctly and handle both Windows- and UNIX-style paths. Added tests too.

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
mjcheetham: I pushed the symbolic ref check down into the base.

GVFS/GVFS/RepairJobs/GitHeadsRepairJob.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using GVFS.Common;
+using GVFS.Common.Tracing;
+
+namespace GVFS.RepairJobs
+{
+    public class GitHeadsRepairJob : GitRefsRepairJob
```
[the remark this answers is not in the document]
mjcheetham: Renamed.

GVFS/GVFS/RepairJobs/GitHeadRepairJob.cs
```diff
@@ -31,114 +30,49 @@ public override IssueType HasIssue(List<string> messages)
...
-                refContents = File.ReadAllText(refPath).Trim();
-            }
-            catch (IOException ex)
-            {
-                messages.Add("IOException while reading .git\\HEAD: " + ex.Message);
-                return false;
-            }
+        protected override bool IsValidRefContents(string fullSymbolicRef, string refContents)
+        {
+            Debug.Assert(fullSymbolicRef == GVFSConstants.DotGit.HeadName, "Only expecting to be called with the HEAD ref");
```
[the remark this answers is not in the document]
mjcheetham: Removed.

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
```diff
@@ -180,6 +180,20 @@ public void CloneAndMount(bool skipPrefetch)
...
+            result.ExitCode.ShouldEqual(0, $"Failed to create branch {branchName} at {sha}");
+        }
+
+        public string GetRevParse(string @ref)
+        {
+            ProcessResult result = GitProcess.InvokeProcess(this.RepoRoot, $"rev-parse {@ref}");
+            result.ExitCode.ShouldEqual(0, $"Failed to get commit for ref '{@ref}'");
+            string refSha = result.Output.Trim();
+            return refSha;
+        }
```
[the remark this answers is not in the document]
mjcheetham: Moved to the test class.

GVFS/GVFS/RepairJobs/GitHeadRepairJob.cs
```diff
@@ -182,5 +103,17 @@ private bool CanBeRepaired(List<string> messages)
 
             return true;
         }
+
+        private string GetHeadRefSha()
+        {
+            string headRefFilePath = Path.Combine(this.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Head);
+
+            var contents = File.ReadAllText(headRefFilePath);
```
sanoursa: Don't use `var`. Please scrub all uses of it in this PR.

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
sanoursa: I would go with `gitRef`, to avoid any chance of ambiguity

GVFS/GVFS.Common/Paths.Shared.cs
```diff
@@ -68,5 +68,55 @@ public static string ConvertPathToGitFormat(string path)
...
+            {
+                throw new ArgumentException("Path must be absolute.", nameof(relativeTo));
+            }
+
+            if (!Path.IsPathRooted(path))
+            {
+                throw new ArgumentException("Path must be absolute.", nameof(path));
+            }
+
+            // Normalize paths
```
sanoursa: What do you mean by "normalize"? You already checked that they are full paths up above, so isn't `GetFullPath` going to be a noop? Also we have a method `IPlatformFileSystem.TryGetNormalizedPath`, which on Windows ends up calling `GetFinalPathNameByHandle` to normalize the path. 

I'm not really sure I understand the goal of this `GetRelativePath` method though. In general, paths are much more complicated than they appear and converting them in this way seems likely to be very error-prone.

GVFS/GVFS/RepairJobs/GitRefsHeadsRepairJob.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            get { return @".git\refs\heads\**\*"; }
+        }
+
+        protected override IEnumerable<string> GetRefs()
+        {
+            string refsHeadsPath = Path.Combine(this.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Refs.Heads.RootFolder);
+            string dotGitPath = Path.Combine(this.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Root) + Path.DirectorySeparatorChar;
+
+            return Directory.EnumerateFiles(refsHeadsPath, "*", SearchOption.AllDirectories)
```
sanoursa: I think a better way to write this code would be to use `DirectoryInfo` to enumerate the children, so that you can keep everything in relative-path-space, and not have to convert from relative to full back to relative paths. As you've seen, paths are hard, and I can't say with any confidence that the current `GetRelativePath` implementation is correct, or even what it might be missing. We should avoid manipulating them to whatever extent possible.

GVFS/GVFS/RepairJobs/GitHeadRepairJob.cs
```diff
@@ -182,5 +103,17 @@ private bool CanBeRepaired(List<string> messages)
...
         }
+
+        private string GetHeadRefSha()
+        {
+            string headRefFilePath = Path.Combine(this.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Head);
+
+            var contents = File.ReadAllText(headRefFilePath);
+            var sha = contents.Trim();
+
+            Debug.Assert(SHA1Util.IsValidShaFormat(sha), "Fix to HEAD ref should have written a valid SHA");
```
sanoursa: Please don't use `Debug.Assert`. If it's a bug, it's a bug even in production, so just throw an exception.

GVFS/GVFS/RepairJobs/GitHeadRepairJob.cs
```diff
@@ -31,114 +30,36 @@ public override IssueType HasIssue(List<string> messages)
...
-            }
-
-            messages.Add("Invalid contents found in '" + GVFSConstants.DotGit.Head + "': " + refContents);
-            return false;
+        protected override IEnumerable<string> GetRefs()
+        {
+            return new[] { GVFSConstants.DotGit.HeadName };
         }
 
         private bool CanBeRepaired(List<string> messages)
```
sanoursa: Why didn't `CanBeRepaired` make it into the base class?

GVFS/GVFS/RepairJobs/GitHeadRepairJob.cs
sanoursa: Also this condition seems invalid. The HEAD file does not have to contain a sha, it can contain the name of another ref.

I realize that you're currently calling this method after writing a sha, but only the calling code knows that, this method does not.

GVFS/GVFS.FunctionalTests/Tools/GVFSFunctionalTestEnlistment.cs
```diff
@@ -19,7 +19,7 @@ public class GVFSFunctionalTestEnlistment
         private static readonly string ZeroBackgroundOperations = "Background operations: 0" + Environment.NewLine;
 
         private GVFSProcess gvfsProcess;
-        
+
```
sanoursa: Drop these noop changes

GVFS/GVFS.UnitTests/Common/PathsTests.cs
```diff
@@ -31,5 +36,56 @@ public void CanConvertOSPathToGitFormat()
...
+        [TestCase(@"Z:\a\b\", @"C:\a\b\c\file.txt", @"C:\a\b\c\file.txt")]
+        [TestCase(@"C:\a\spacey dir\", @"C:\a\spacey dir\c\spacey file.txt", @"c\spacey file.txt")]
+        public void GetRelativePath_Windows(string relativeTo, string path, string expected)
+        {
+            var actual = Paths.GetRelativePath(relativeTo, path);
+
+            // If we're running on non-Windows the relative path is returned with platform-
+            // native path separators, even if a Windows-style path was passed in.
+            // Convert it back to the Windows-style separator "\".
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
sanoursa: I don't think we should make tests behave differently based on what platform they're running on. What is the actual goal here?

GVFS/GVFS/RepairJobs/GitRefsHeadsRepairJob.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class GitRefsHeadsRepairJob : GitRefsRepairJob
+    {
+        public GitRefsHeadsRepairJob(ITracer tracer, TextWriter output, GVFSEnlistment enlistment)
+            : base(tracer, output, enlistment)
+        {
+        }
+
+        public override string Name
+        {
+            get { return @".git\refs\heads\**\*"; }
```
sanoursa: This isn't what your code is actually doing. Your code is checking for anything in `.git\refs\heads\*`

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                messages.Add($"Not all references could be fixed. Failed to fix {numFailures} references.");
+                return FixResult.Failure;
+            }
+
+            return FixResult.Success;
+        }
+
+        /// <summary>
+        /// Get the list of full symbolic references (eg. HEAD, refs/remotes/origin/HEAD, refs/heads/master)
```
sanoursa: The example is somewhat confusing since none of the code inspects the `refs\remotes` folder.

GVFS/GVFS.UnitTests/Common/PathsTests.cs
```diff
@@ -31,5 +40,44 @@ public void CanConvertOSPathToGitFormat()
...
+        }
+
+        private static void CreateEmptyFiles(string rootDirectory, IEnumerable<string> files)
+        {
+            foreach (string file in files)
+            {
+                string fullFilePath = Path.Combine(rootDirectory, file);
+
+                string parentDirectory = Path.GetDirectoryName(fullFilePath);
+                Directory.CreateDirectory(parentDirectory);
```
sanoursa: Unit tests must not do actual IO. Either mock out the file system, or drop this test and rely on functional tests.

GVFS/GVFS.UnitTests/Common/PathsTests.cs
```diff
@@ -8,6 +12,11 @@ namespace GVFS.UnitTests.Common
     [TestFixture]
     public class PathsTests
     {
+        // Cannot use the System.IO.Path.(Alt)DirectorySeparatorChar field because
+        // this is different depending on the platform the tests are being run on.
+        private const char WindowsPathSeparator = '\\';
```
sanoursa: I don't think these are being used anymore?

GVFS/GVFS/RepairJobs/GitRefsHeadsRepairJob.cs
sanoursa: I guess it depends on what wildcard syntax you're using. I read this as `.git\refs\heads\<some subfolder here>\*`. Either way, this is just a display string and intended to inform users, so lets remove the ambiguity. You can just say `Local branches` here and be totally clear about what you mean.

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
sanoursa: Even if it was pre-existing, I would drop the examples. It's potentially quite misleading to a future reader who may think that it would be desirable to repair remote refs, when it definitely would not be.

GVFS/GVFS.Common/Paths.Shared.cs
```diff
@@ -68,5 +69,28 @@ public static string ConvertPathToGitFormat(string path)
         {
             return path.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
         }
+
+        public static IEnumerable<string> GetFilesRecursive(string rootDirectory)
```
sanoursa: I would make this a private method within the repair class. While its behavior is general, we don't need it anywhere else and it's unlikely that we will. And as a private method in that class, we don't have to be super picky about its perf characteristics (e.g. how much overhead is the `yield return` syntax adding?), but I wouldn't this code getting called from a perf critical code path without more analysis.

sanoursa: I would rename this to `GetRecursiveRelativeFilePaths` - the fact that they're relative is quite important here, and it's returning paths, not files.

GVFS/GVFS.UnitTests/Common/PathsTests.cs
sanoursa: See my other note, I would make the new relative path method private anyway. Functional test coverage should be adequate here.

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
```diff
@@ -0,0 +1,174 @@
...
+            error = null;
+            return true;
+        }
+
+        private bool TryParseRef(string fullSymbolicRef, List<string> messages = null)
+        {
+            string refPath = Path.Combine(this.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Root, fullSymbolicRef);
+            if (!File.Exists(refPath))
+            {
+                messages?.Add($"Could not find ref file for '{fullSymbolicRef}'");
```
sanoursa: I would indicate what path you were looking for here.

I just tried testing this code and my first attempt hit a failure to repair. My repro was to simply open up `.git/refs/heads/master` and delete its contents. The output tells me that it can't find the ref, but I'm not sure why not because it's there.

```
.git\refs\heads\**\*
    Could not find ref file for 'master'
    Repair failed. Run 'gvfs log C:\GVFSRepos\GVFS' for more info.
    Could not find reflog for ref 'master'
    Not all references could be fixed. Failed to fix 1 references.
```

sanoursa: I believe it's looking for a file named `.git/master`, rather than `.git/refs/heads/master`

GVFS/GVFS/RepairJobs/GitHeadRepairJob.cs
sanoursa: What if HEAD is pointing to the ref that you're currently fixing? In effect, you'd be changing HEAD during one of those conflict states.

GVFS/GVFS/RepairJobs/GitRefsRepairJob.cs
mjcheetham: Ah yes. I've got my base path/ref name prefixes mixed up here. Will address in next push.

GVFS/GVFS.UnitTests/Common/PathsTests.cs
mjcheetham: I'll drop it and rely on functional tests (given it's being pushed into the repair job as private).

jrbriggs:
Closing this for now, since it's not actively under development

PR #532, opened 2018-11-19
Allow multiple internal mount parameters

### LINKED ISSUE #479: Log if mount is being manually run by user or kicked off by the service
We are seeing several errors come thru that say 'disk I/O error'

The two flavors are:
Failed to create src folder callback listener Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 10: 'disk I/O error'.
System.IO.IOException: The device is not ready.
The first is almost always on startup.  The second is sometimes on startup.  

Reaching out to a few users they just run mount or open/close razzle.  We'd like to get a better idea of this failure.  One hypothesis is that the service is kicking off mount when it can't run, and adding this data will help field that theory.  In general it's also helpful to know if users are having to manually mount versus the service.

Since the service kicks off "mount", this makes it tricky to determine how the start occurred.
Options discussed:
1. Add an Env variable in the service, that MountVerb will look for and log to determine start
2. Add an new optional parameter on Mount
@sanoursa, let me know your preference here on approach

jeschu1:
We are seeing several errors come thru that say 'disk I/O error'

**The two flavors are:**
1. Failed to create src folder callback listener Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 10: 'disk I/O error'.
2. System.IO.IOException: The device is not ready.
The first is almost always on startup. The second is sometimes on startup.

This change logs whether the mount is started by the service.  This will help us determine how often mount is failing when launched via the service.

**There were 3 evaluated approaches**
1. Walk the process tree
This was decided against since it's not cross-plat and walking process trees is non-trivial
2. Use an env var
This was decided against since we start processes using native code and adding an env variable there adds new complexity.  It's also less than ideal since we don't pass any other arguments this way.
3. Use the existing internal-only property on MountVerb
I extended this option to allow you send a series of var=val;var=val; type arguments.  I originally considered using json but passing args via json has it's own set of complexities (double escaping) and can become quite verbose.  Since this is internal I'm voting to keep it simple.

The third approach here is what I went with.  It seems the most standard as well as giving an easy way to add internal parameters in the future.

resolves #479

### DIFF
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -208,9 +208,11 @@ namespace GVFS.Common
 
         public static class VerbParameters
         {
+                public const string InternalUseOnly = "internal_use_only";
+
             public static class Mount
             {
-                public const string ServiceName = "internal_use_only_service_name";
+                public const string StartedByService = "StartedByService";
                 public const string Verbosity = "verbosity";
                 public const string Keywords = "keywords";
                 public const string DebugWindow = "debug-window";
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
@@ -94,7 +94,7 @@ namespace GVFS.FunctionalTests.Tools
         {
             ProcessStartInfo processInfo = null;
             processInfo = new ProcessStartInfo(this.pathToGVFS);
-            processInfo.Arguments = args + " --internal_use_only_service_name " + GVFSServiceProcess.TestServiceName;
+            processInfo.Arguments = args + " " + TestConstants.InternalUseOnlyFlag + " " + GVFSHelpers.GetInternalParameter();
 
             processInfo.WindowStyle = ProcessWindowStyle.Hidden;
             processInfo.UseShellExecute = false;
--- a/GVFS/GVFS.Mount/InProcessMountVerb.cs
+++ b/GVFS/GVFS.Mount/InProcessMountVerb.cs
@@ -47,6 +47,14 @@ namespace GVFS.Mount
             HelpText = "Show the debug window.  By default, all output is written to a log file and no debug window is shown.")]
         public bool ShowDebugWindow { get; set; }
 
+        [Option(
+            's',
+            GVFSConstants.VerbParameters.Mount.StartedByService,
+            Default = "false",
+            Required = false,
+            HelpText = "Service initiated mount.")]
+        public string StartedByService { get; set; }
+
         [Value(
                 0,
                 Required = true,
@@ -80,6 +88,7 @@ namespace GVFS.Mount
                 {
                     { "IsElevated", GVFSPlatform.Instance.IsElevated() },
                     { nameof(this.EnlistmentRootPathParameter), this.EnlistmentRootPathParameter },
+                    { nameof(this.StartedByService), this.StartedByService },
                 });
 
             AppDomain.CurrentDomain.UnhandledException += (object sender, UnhandledExceptionEventArgs e) =>
--- a/GVFS/GVFS.Service/GVFSMountProcess.cs
+++ b/GVFS/GVFS.Service/GVFSMountProcess.cs
@@ -58,7 +58,10 @@ namespace GVFS.Service
 
         private bool CallGVFSMount(string repoRoot)
         {
-            return this.CurrentUser.RunAs(Configuration.Instance.GVFSLocation, "mount " + repoRoot);
+            InternalVerbParameters mountInternal = new InternalVerbParameters(serviceName: null, startedByService: true);
+            return this.CurrentUser.RunAs(
+                Configuration.Instance.GVFSLocation,
+                $"mount {repoRoot} --{GVFSConstants.VerbParameters.InternalUseOnly} {mountInternal.ToJson()}");
         }
     }
 }
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -5,6 +5,7 @@ using GVFS.Common.Git;
 using GVFS.Common.Http;
 using GVFS.Common.NamedPipes;
 using GVFS.Common.Tracing;
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -23,9 +24,9 @@ namespace GVFS.CommandLine
         {
             this.Output = Console.Out;
             this.ReturnCode = ReturnCode.Success;
-            this.ServiceName = GVFSConstants.Service.ServiceName;
             this.validateOriginURL = validateOrigin;
-
+            this.ServiceName = GVFSConstants.Service.ServiceName;
+            this.StartedByService = false;
             this.Unattended = GVFSEnlistment.IsUnattended(tracer: null);
 
             this.InitializeDefaultParameterValues();
@@ -34,11 +35,36 @@ namespace GVFS.CommandLine
         public abstract string EnlistmentRootPathParameter { get; set; }
 
         [Option(
-            GVFSConstants.VerbParameters.Mount.ServiceName,
-            Default = GVFSConstants.Service.ServiceName,
+            GVFSConstants.VerbParameters.InternalUseOnly,
             Required = false,
             HelpText = "This parameter is reserved for internal use.")]
+        public string InternalParameters
+        {
+            set
+            {
+                if (!string.IsNullOrEmpty(value))
+                {
+                    try
+                    {
+                        InternalVerbParameters mountInternal = InternalVerbParameters.FromJson(value);
+                        if (!string.IsNullOrEmpty(mountInternal.ServiceName))
+                        {
+                            this.ServiceName = mountInternal.ServiceName;
+                        }
+
+                        this.StartedByService = mountInternal.StartedByService;
+                    }
+                    catch (JsonReaderException e)
+                    {
+                        this.ReportErrorAndExit("Failed to parse InternalParameters: {0}.\n {1}", value, e);
+                    }                    
+                }
+            }
+        }
+
         public string ServiceName { get; set; }
+
+        public bool StartedByService { get; set; }
 
         public bool Unattended { get; private set; }
 
--- a/GVFS/GVFS.Common/InternalVerbParameters.cs
+++ b/GVFS/GVFS.Common/InternalVerbParameters.cs
@@ -1,0 +1,26 @@
+﻿using Newtonsoft.Json;
+
+namespace GVFS.Common
+{
+    public class InternalVerbParameters
+    {
+        public InternalVerbParameters(string serviceName, bool startedByService)
+        {
+            this.ServiceName = serviceName;
+            this.StartedByService = startedByService;
+        }
+
+        public string ServiceName { get; private set; }
+        public bool StartedByService { get; private set; }
+
+        public static InternalVerbParameters FromJson(string json)
+        {
+            return JsonConvert.DeserializeObject<InternalVerbParameters>(json);
+        }
+
+        public string ToJson()
+        {
+            return JsonConvert.SerializeObject(this);
+        }
+    }
+}
--- a/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/JunctionAndSubstTests.cs
+++ b/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/JunctionAndSubstTests.cs
@@ -214,11 +214,11 @@ namespace GVFS.FunctionalTests.Windows.Tests
             string mountCommand;
             if (enlistmentPath != null)
             {
-                mountCommand = $"mount \"{enlistmentPath}\" --internal_use_only_service_name {GVFSServiceProcess.TestServiceName}";
+                mountCommand = $"mount \"{enlistmentPath}\" {TestConstants.InternalUseOnlyFlag} {GVFSHelpers.GetInternalParameter()}";
             }
             else
             {
-                mountCommand = $"mount --internal_use_only_service_name {GVFSServiceProcess.TestServiceName}";
+                mountCommand = $"mount {TestConstants.InternalUseOnlyFlag} {GVFSHelpers.GetInternalParameter()}";
             }
 
             ProcessStartInfo startInfo = new ProcessStartInfo(GVFSTestConfig.PathToGVFS);
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
@@ -192,7 +192,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             string enlistmentRoot = this.Enlistment.EnlistmentRoot;
 
             ProcessStartInfo processInfo = new ProcessStartInfo(GVFSTestConfig.PathToGVFS);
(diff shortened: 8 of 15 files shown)


jrbriggs:
@jeschu1 can you summarize the motivation for this change in the description?

jeschu1:
I put the PR up to double check the tests.  Once they pass I'll update the context.  Please don't dive into this PR yet friends!

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -210,7 +210,9 @@ public static class VerbParameters
         {
             public static class Mount
             {
-                public const string ServiceName = "internal_use_only_service_name";
+                public const string InternalParameters = "internal_use_only_parameters";
```
jeschu1: I'm not sure if changing internal_use_only_service_name to internal_use_only_parameters is a legit thing to do.  If anyone was using the parameter (which they shouldn't it's internal :-) ) they'd be broken.  I do like the new name as a better description though.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,12 +34,41 @@ public GVFSVerb(bool validateOrigin = true)
...
 
         [Option(
-            GVFSConstants.VerbParameters.Mount.ServiceName,
-            Default = GVFSConstants.Service.ServiceName,
+            GVFSConstants.VerbParameters.Mount.InternalParameters,
             Required = false,
-            HelpText = "This parameter is reserved for internal use.")]
+            HelpText = "This parameter is reserved for internal use." +
+                " Accepts input in the form key1=value;key2=value;key3=value etc..")]
+        public string InternalParameters
```
jeschu1: If we give this approach the thumbs up, I'll add unit tests for the below

GVFS/GVFS.Mount/InProcessMountVerb.cs
```diff
@@ -47,6 +47,14 @@ public InProcessMountVerb()
...
         public bool ShowDebugWindow { get; set; }
 
+        [Option(
+            's',
+            GVFSConstants.VerbParameters.Mount.StartedByService,
+            Default = "false",
+            Required = false,
+            HelpText = "Service initiated mount.")]
+        public string StartedByService { get; set; }
+
```
jeschu1: Arguably this isn't consistent with MountVerb which allows you to pass a series of internal parameters.  On the InProcessMountVerb itself, you only need 1 parameter so I'd rather not make it take a series until/unless the need arises.

sanoursa: The `MountVerb` gets called in-proc by other verbs. No such requirement here.

GVFS/GVFS.Common/GVFSConstants.cs
sanoursa: Not your doing, but this parameter doesn't belong in `VerbParameters.Mount`, since it's not specific to just `MountVerb` (I believe it was when it was first added, and later got moved to all verbs).

I would create a class named `VerbParameters.Internal` with a member named `Name` and other members for each of the parameters that it can contain.

sanoursa: Minor, but I would name it just "internal_use_only" and drop the "_parameters". Or even just "internal_use". Make the CLI, especially when captured in Task Manager, etc, slightly less verbose.

sanoursa: Or even better, define a class with properties, and use Json to serialize/deserialize its members. There's no need for string-typed values here since we don't need any cross-version compatibility.

GVFS/GVFS.Mount/InProcessMountVerb.cs
```diff
@@ -47,6 +47,14 @@ public InProcessMountVerb()
             HelpText = "Show the debug window.  By default, all output is written to a log file and no debug window is shown.")]
         public bool ShowDebugWindow { get; set; }
 
+        [Option(
+            's',
```
sanoursa: No need for a short name, since no user will be running this command

sanoursa: (I know that is inconsistent with the other params here, but pretty much all of them are holdovers from the early days and should be removed anyways)

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,12 +34,41 @@ public GVFSVerb(bool validateOrigin = true)
         public abstract string EnlistmentRootPathParameter { get; set; }
 
         [Option(
-            GVFSConstants.VerbParameters.Mount.ServiceName,
-            Default = GVFSConstants.Service.ServiceName,
+            GVFSConstants.VerbParameters.Mount.InternalParameters,
             Required = false,
-            HelpText = "This parameter is reserved for internal use.")]
+            HelpText = "This parameter is reserved for internal use." +
+                " Accepts input in the form key1=value;key2=value;key3=value etc..")]
```
sanoursa: I would drop this. Users don't need to know or care what the format is.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,12 +34,41 @@ public GVFSVerb(bool validateOrigin = true)
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
         public string ServiceName { get; set; }
 
+        public string StartedByService { get; set; }
```
sanoursa: This should be a bool, not a string. But this is moot if you parse the json into a strongly-typed `InternalParameters` instance.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,12 +34,41 @@ public GVFSVerb(bool validateOrigin = true)
...
+            {
+                if (!string.IsNullOrEmpty(value))
+                {
+                    string[] values = value.Split(';');
+                    foreach (string s in values)
+                    {
+                        string[] pair = s.Split('=');
+                        if (pair.Length == 2)
+                        {
+                            if (string.Equals(pair[0], GVFSConstants.VerbParameters.Mount.StartedByService))
```
mjcheetham: Do we care about case-sensitivity? I'd use `StringComparer.OrdinalIgnoreCase.Equals` here.

GVFS/GVFS.Service/GVFSMountProcess.cs
```diff
@@ -58,7 +58,10 @@ public void Dispose()
 
         private bool CallGVFSMount(string repoRoot)
         {
-            return this.CurrentUser.RunAs(Configuration.Instance.GVFSLocation, "mount " + repoRoot);
+            return this.CurrentUser.RunAs(
+                Configuration.Instance.GVFSLocation,
```
mjcheetham: This expression is now quite long with lots of string appending. Makes it hard to see if the spaces are in the correct place to separate arguments/values.

Consider making this an `$"interpolated {string}"` (or use a `StringBuilder`) and pull the final formatted string out of the `RunAs` call.

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
```diff
@@ -94,7 +94,7 @@ private string CallGVFS(string args, bool failOnError = false, string trace = nu
         {
             ProcessStartInfo processInfo = null;
             processInfo = new ProcessStartInfo(this.pathToGVFS);
-            processInfo.Arguments = args + " --internal_use_only_service_name " + GVFSServiceProcess.TestServiceName;
+            processInfo.Arguments = args + " --internal_use_only_parameters " + GVFSHelpers.GetTestServiceParamater();
```
mjcheetham: Should we (or have we already) pull out the name of the parameter into a constant somewhere?
```csharp
const string InternalUseParametersName = "internal_use_only_parameters";
```

GVFS/GVFS.Common/GVFSConstants.cs
jeschu1: I updated it to be a serialized Json class.  It's a bit more painful (supplying json across a command line is no fun!), but that will probably encourage folks not to use it which is what we want, while forcing type.

GVFS/GVFS/CommandLine/GVFSVerb.cs
jeschu1: I updated it to be a serialized class.  No ambiguity now and type safety.

GVFS/GVFS.Service/GVFSMountProcess.cs
```diff
@@ -58,7 +58,10 @@ public void Dispose()
 
         private bool CallGVFSMount(string repoRoot)
         {
-            return this.CurrentUser.RunAs(Configuration.Instance.GVFSLocation, "mount " + repoRoot);
+            MountInternal mountInternal = new MountInternal() { StartedByService = true };
+            return this.CurrentUser.RunAs(
+                Configuration.Instance.GVFSLocation,
+                $"mount {repoRoot} --internal_use_only {mountInternal.ToJson()}");
```
wilbaker: Use `GVFSConstants.VerbParameters.Mount.InternalUseOnly` here rather than `internal_use_only`

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,12 +34,30 @@ public GVFSVerb(bool validateOrigin = true)
...
+                    }
+
+                    this.StartedByService = mountInternal.StartedByService;
+                }
+            }
+        }
+
         public string ServiceName { get; set; }
 
+        public bool StartedByService { get; set; }
```
wilbaker: Minor: Can `ServiceName` and `StartedByService` be changed to `protected`?

(It actually looks like `Unattended`, `ServicePipeName`, and `Output` can probably be `protected` as well)

GVFS/GVFS.Common/MountInternal.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GVFS.Common
+{
+    public class MountInternal
```
sanoursa: This currently sounds like a method name rather than a class name, and also these parameters apply to all verbs, not just mount. Rename to `InternalVerbParameters`?

GVFS/GVFS.Common/MountInternal.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Text;
```
sanoursa: Remove unnecessary using statements

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,12 +34,30 @@ public GVFSVerb(bool validateOrigin = true)
...
+                    {
+                        this.ServiceName = mountInternal.ServiceName;
+                    }
+
+                    this.StartedByService = mountInternal.StartedByService;
+                }
+            }
+        }
+
         public string ServiceName { get; set; }
```
sanoursa: Make these setters private

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -210,7 +210,8 @@ public static class VerbParameters
         {
             public static class Mount
             {
-                public const string ServiceName = "internal_use_only_service_name";
+                public const string InternalUseOnly = "internal_use_only";
+                public const string StartedByService = "StartedByService";
```
sanoursa: I don't think this is being used any more?

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -210,7 +210,8 @@ public static class VerbParameters
         {
             public static class Mount
             {
-                public const string ServiceName = "internal_use_only_service_name";
+                public const string InternalUseOnly = "internal_use_only";
```
sanoursa: Move this up to the parent class, since it's not specific to Mount anymore

GVFS/GVFS/CommandLine/GVFSVerb.cs
sanoursa: They can be, but then so could everything else in these classes. This is the top of the stack so there isn't any other code calling these methods/properties other than the verbs themselves.

GVFS/GVFS.Common/MountInternal.cs
sanoursa: Also does it belong here in `GVFS.Common`? I think it could probably move into the CLI project instead.

GVFS/GVFS.Common/MountInternal.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace GVFS.Common
+{
+    public class MountInternal
+    {
+        public string ServiceName { get; set; }
+        public bool StartedByService { get; set; }
+
+        public static MountInternal FromJson(string json)
+        {
+            return JsonConvert.DeserializeObject<MountInternal>(json);
```
sanoursa: Can this throw if the json is not in the right format?

jeschu1: @sanoursa yes it can throw.  Since it's internal I'm thinking I'd prefer the mount not to start if we get bad input.

sanoursa: Yes, agreed that it should take down the process. Please just confirm that it is getting handled and logged as it tears down the process.

GVFS/GVFS.Common/MountInternal.cs
jeschu1: @sanoursa CLI project?  Sorry i'm not seeing it.  It's used by GVFS.Service/GVFS.Windows(Mac) which is why I was thinking Common might be a good home.

GVFS/GVFS.Common/MountInternal.cs
jeschu1: Done, but it's a little spooky you could spot that from just a PR :-)

GVFS/GVFS.Common/MountInternal.cs
sanoursa: `GVFS.Windows` and `GVFS.Mac` are what I meant by the CLI project(s), since they produce the CLI executable.

This class exists for interacting with the CLI, which is why I suggested putting it there. `GVFS.Service` should be able to take a dependency on `GVFS.Windows`.

derrickstolee: You're right, this is a class specifically for `MountVerb`.

It should also be used by whatever Mac equivalent of the `GVFS.Service` project will be. I imagine that the Mac service can take a dependence on `GVFS.Mac` and get the same functionality. Is that right?

sanoursa: It's not specifically for `MountVerb`, it gets used by all CLI verbs.

And yes the Mac equivalent of the service should be able to depend on the CLI project as well.

GVFS/GVFS.Common/GVFSConstants.cs
jeschu1: @sanoursa  Where would you recommend adding the StartedByService Constant?
It's used in both InProcessMountVerb (GVFS.Win) and MountVerb (GVFS.Mount.Win)

GVFS/GVFS.Common/MountInternal.cs
jeschu1: @sanoursa I tried to add a dependency on GVFS.Windows from GVFS.Service.

However since GVFS.Service has a dependency on GVFS.Service.UI which has a shared file ( PlatformLoader.Windows.cs), I'm getting build errors saying it doesn't know which GVFSPlatformLoader to use.  Not sure how to resolve this one.

jeschu1: I tried removing the shared file, to no avail.  As you've previously mentioned GVFS.Service.UI needs some attention in terms of refactoring / getting it to work again.  I'm going to leave this final change out of scope for the PR.  But hopefully we prioritize the notifications soon and then we can clean this up.

PR #533, opened 2018-11-19
Downloaded Installers of previous versions should not trigger upgrade reminders

### LINKED ISSUE #502: VFS for Git is claiming there are updates it isn't getting
A new version of GVFS is available. Run `gvfs upgrade` to start the upgrade.
 
E:\osgtools\src>gvfs --version
GVFS 1.0.18297.1
 
E:\osgtools\src>gvfs upgrade
Great news, you're all caught up on upgrades in the Fast ring!
 
E:\osgtools\src>gvfs --version
GVFS 1.0.18297.1

alameenshah:
1. GVFS.Service does its periodic upgrade check. When it finds that the current version is up-to-date, it deletes the previously downloaded releases. In between the user may see reminder messaging while running git commands.
2. User manually runs `gvfs upgrade`. When gvfs cli finds that the current version is up-to-date it deletes the un-needed release downloads. User would no longer see reminders.

Fixes #502

### DIFF
--- a/GVFS/GVFS.Service/ProductUpgradeTimer.cs
+++ b/GVFS/GVFS.Service/ProductUpgradeTimer.cs
@@ -79,6 +79,9 @@ namespace GVFS.Service
                 if (newerVersion == null)
                 {
                     // Already up-to-date
+                    // Make sure there a no asset installers remaining in the Downloads directory. This can happen if user
+                    // upgraded by manually downloading and running asset installers.
+                    productUpgrader.DeletePreviousDownloads();
                     errorMessage = null;
                     return true;
                 }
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -79,6 +79,9 @@ private bool TryDownloadUpgrade(out string errorMessage)
                 if (newerVersion == null)
                 {
                     // Already up-to-date
+                    // Make sure there a no asset installers remaining in the Downloads directory. This can happen if user
+                    // upgraded by manually downloading and running asset installers.
+                    productUpgrader.DeletePreviousDownloads();
```
sanoursa: What happens if this fails? What happens if it throws?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -128,6 +128,10 @@ private bool TryRunProductUpgrade()
             if (newestVersion == null)
             {
                 this.ReportInfoToConsole($"Great news, you're all caught up on upgrades in the {this.upgrader.Ring} ring!");
+
+                // Make sure there a no asset installers remaining in the Downloads directory. This can happen if user
+                // upgraded by manually downloading and running asset installers.
+                this.upgrader.DeletePreviousDownloads();
```
sanoursa: Same questions, what happens if this throws?

sanoursa: Can you add functional test coverage for this scenario?

alameenshah: Sure.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: Added exception tracing in ```DeletePreviousDownloads()``` method. This is not treated as a critical error, upgrade would log the error and proceed.

PR #534, opened 2018-11-19
gvfs config - error out while reading missing key

### LINKED ISSUE #402: `gvfs config [nonexistant key]` should exit with code 1
To be consistent with Git, we should not print a blank line and we should exit with a non-zero return value.

alameenshah:
gvfs config now treats an attempt to read a unknown key as error and exits with GenericError code. No error message is printed.

Fixes #402

Comparing ```gvfs config``` behavior vs ```git config```.

screen shot 2018-11-19 at 5 25 40 pm

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/ConfigVerbTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/ConfigVerbTests.cs
@@ -17,6 +17,11 @@ namespace GVFS.FunctionalTests.Tests.MultiEnlistmentTests
         private const string FloatSettingKey = "functionalTest_Float";
         private const string RegularStringSettingKey = "functionalTest_RegularString";
         private const string SpacedStringSettingKey = "functionalTest_SpacedString";
+        private const string SpacesOnlyStringSettingKey = "functionalTest_SpacesOnlyString";
+        private const string EmptyStringSettingKey = "functionalTest_EmptyString";
+        private const string NonExistentSettingKey = "functionalTest_NonExistentSetting";
+
+        private const int GenericErrorExitCode = 3;
 
         private readonly Dictionary<string, string> initialSettings = new Dictionary<string, string>()
         {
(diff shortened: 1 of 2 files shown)


GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -93,11 +93,12 @@ public override void Execute()
                 else
                 {
                     string valueRead = null;
-                    if (!this.localConfig.TryGetConfig(this.Key, out valueRead, out error))
+                    if (!this.localConfig.TryGetConfig(this.Key, out valueRead, out error) ||
+                        string.IsNullOrEmpty(valueRead))
```
kewillford: So we are now saying a empty value is error?

alameenshah: Right. Here is a comparison with ```git config```.
```
PS M:\Work\GVFS\VFSForGit_Ameen> git config un.defined.key "  "
PS M:\Work\GVFS\VFSForGit_Ameen> git config un.defined.key

PS M:\Work\GVFS\VFSForGit_Ameen> echo $?
True
PS M:\Work\GVFS\VFSForGit_Ameen> gvfs config un.defined.key "  "
PS M:\Work\GVFS\VFSForGit_Ameen> gvfs config un.defined.key

PS M:\Work\GVFS\VFSForGit_Ameen> echo $?
True
PS M:\Work\GVFS\VFSForGit_Ameen>
```

kewillford: Looks like we should be able to write some tests around this right?

alameenshah: Added Functional tests.

GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/ConfigVerbTests.cs
```diff
@@ -72,6 +75,26 @@ public void ReadSingleSetting()
         }
 
         [TestCase, Order(5)]
+        public void AddSpaceValueSetting()
```
kewillford: Add one more test for reading a non existing setting?

PR #535, opened 2018-11-20
Taking DEVELOPMENT_TEAM from environment variable, if set, allowing to easier build native mac binaries

turbonaitis:
Not sure, if DEVELOPMENT_TEAM environment variable is supposed to be used by xcodebuild, but it wasn't on my machine with this xcodebuild version
```
Xcode 10.1
Build version 10B61
```
Modified build scripts to pass this environment variable to xcodebuild, if it is set, so that people, who do not have the developer certificate, that is defined in the project files, can build easier.

### DIFF
--- a/ProjFS.Mac/Scripts/Build.sh
+++ b/ProjFS.Mac/Scripts/Build.sh
@@ -11,8 +11,14 @@ ROOTDIR=$SRCDIR/..
 PACKAGES=$ROOTDIR/packages
 
 PROJFS=$SRCDIR/ProjFS.Mac
-
-xcodebuild -configuration $CONFIGURATION -workspace $PROJFS/PrjFS.xcworkspace build -scheme PrjFS -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
+if [ -z "$DEVELOPMENT_TEAM" ]
+then 
+    echo "DEVELOPMENT_TEAM environment variable not set, using DEVELOPMENT_TEAM from project..."
+    xcodebuild -configuration $CONFIGURATION -workspace $PROJFS/PrjFS.xcworkspace build -scheme PrjFS -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
+else
+    echo "DEVELOPMENT_TEAM environment variable set to $DEVELOPMENT_TEAM, using it for build..."
+    xcodebuild -configuration $CONFIGURATION -workspace $PROJFS/PrjFS.xcworkspace build -scheme PrjFS -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native DEVELOPMENT_TEAM=$DEVELOPMENT_TEAM || exit 1
+fi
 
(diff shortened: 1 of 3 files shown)


ProjFS.Mac/Scripts/Build.sh
```diff
@@ -11,8 +11,14 @@ ROOTDIR=$SRCDIR/..
 PACKAGES=$ROOTDIR/packages
 
 PROJFS=$SRCDIR/ProjFS.Mac
-
-xcodebuild -configuration $CONFIGURATION -workspace $PROJFS/PrjFS.xcworkspace build -scheme PrjFS -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
+if [ -z "$DEVELOPMENT_TEAM" ]
```
sanoursa: I like making this more flexible, but please avoid using environment variables as implicit inputs to a build. A parameter to the script is much easier to track and reason about.

turbonaitis: I like the explicitness of a parameter as well, but passing development team (which is a random string) on every build is not really practical. It's like a PATH - you set once, where binaries are, and then it works.

pmj: How about making it an explicit "hardcoded" variable in the build script, so we can either go in and edit it in our local branch, or pass it on the command line, whichever is more convenient?

Another option is to query what IDs are available on the system and pick the only one available if there is only one. You can get a list using:

    security find-identity -v -p codesigning

I'm no shell scripting magician, but I've previously massaged this down to just the team IDs using the following: (although with intermediate variables in practice)

    security find-identity -v -p codesigning | grep "Developer ID Application" | sed -E 's/.*\"(.*)\"$/\1/' | sort | uniq | sed -n -E 's/.*\(([A-Z0-9]+)\)$/\1/p'

If there's only one choice, it's pretty obvious which to pick, if there's more than 1, we could require a choice via a command line argument?

turbonaitis: Hardcoding it to a script brings us back to square one - we need to make changes to scripts (currently - to project files), which we can't commit/push to upstream.

I like detecting the available certificate/team automatically, but this only works well, when there's only one of them (I have two, one for iPhone, one for Mac). We could filter out iPhone ones, but if someone has more than one valid cert (i.e. one personal, one company), what then? Do we ask them to choose every time? If no, where do we record that choice? An environment variable? :) If we ask to chose by passing in a command line argument, what do we ask them to pass? A number, indicating which cert they want to chose (1st, 2nd etc)? Name? Cert ID? 

The reason my PR suggests using an environment variable, is that I think, that this is something, that a developer sets once, and then forgets about it. I don't see a scenario, where a developer would want to use different signing certificates often (while if they do - they're free to do this by `DEVELOPMENT_TEAM=xxx build.sh`. 

If you think, that this is something, that would vary on someones machine quite often, I'd be happy to hear that scenario.

turbonaitis:
So, do we think this is a useful addition to the scripts? If so, how can we move forward with this PR? Otherwise, let's just close this diff to reduce clutter.

jrbriggs:
Closing this for now to reduce the number of active PRs.

PR #536, opened 2018-11-20
Add DiskInfo to HeartBeat

jeschu1:
This logs disk info on heat beat:

\"PhysicalDiskInfo\":{\"DriveLetter\":\"C\",\"VolumeDriveType\":\"Fixed\",\"VolumeFileSystem\":\"NTFS\",\"VolumeFileSystemLabel\":\"OSDisk\",\"VolumeSize\":\"510928089088\",\"VolumeSizeRemaining\":\"103830548480\",\"DiskNumber\":\"0\",\"DiskModel\":\"MTFDDAK512MBF1A\",\"DiskIsSystem\":\"True\",\"DiskIsBoot\":\"True\",\"DiskSerialNumber\":\"XXXXXX\",\"PhysicalMediaType\":\"SSD\",\"PhysicalBusType\":\"SATA\",\"PhysicalSpindleSpeed\":\"0\"}
This should come in very handy when investigating why users run out of disk space.

@sanoursa  I'm using the existing [GetPhysicalDiskInfo method](https://github.com/Microsoft/VFSForGit/blob/785ccb4a67281d0eb10f19cb53e6431b62b88555/GVFS/GVFS.Platform.Windows/WindowsPhysicalDiskInfo.cs#L57), which  brings back more info that we need.  Let me know if you think it's worth seperating queryVolumeString into it's own method and just calling that here.   I'll update the failing unit test once we make a decision there.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsPhysicalDiskInfo.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPhysicalDiskInfo.cs
@@ -54,7 +54,7 @@ namespace GVFS.Platform.Windows
         /// the given pathname.  For example, whether the drive is an SSD or HDD.
         /// </summary>
         /// <returns>A dictionary of platform-specific keywords and values.</returns>
-        public static Dictionary<string, string> GetPhysicalDiskInfo(string path)
+        public static Dictionary<string, string> GetPhysicalDiskInfo(string path, bool sizeStatsOnly)
         {
             // Use the WMI APIs to get details about the physical disk associated with the given path.
             // Some of these fields are avilable using normal classes, such as System.IO.DriveInfo:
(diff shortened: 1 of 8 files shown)


sanoursa:
Two questions:
* Is there a perf hit from querying all the extra stats that we don't care about?
* Are we bloating the logs / telemetry with redundant info, since most of those stats will be redundant on every heartbeat message?

Keep in mind that "queryVolumeString" is totally Windows-specific. What you might consider doing is add a flag to `GetPhysicalDiskInfo` so that you can ask it for all drive stats, or just size-related stats, and let each platform make its own decision about what that means.

jeschu1:
@sanoursa that makes sense

To answer both questions.
Yes there is a perf hitting in getting all the disk info (an extra query).  Yes it also bloats the logs.

I'll update GetPhysicalDiskInfo to have a sizeStatsOnly flag.  That should solve both problems nicely.  Thanks!

jeschu1:
@sanoursa any chance you may be able to give this a final review.  If your too busy let me know and I'll ask someone else.  Thank You!

GVFS/GVFS.Platform.Windows/WindowsPhysicalDiskInfo.cs
```diff
@@ -107,6 +107,11 @@ public class WindowsPhysicalDiskInfo
                     result.Add("VolumeSizeRemaining", FetchValue(mbo, "SizeRemaining"));
                 }
 
+                if (sizeStatsOnly)
```
sanoursa: Not all your doing, but this method is quite long, does two separate things, and now with an exit in the middle, can be pretty hard to follow. I would split it into two new methods, so that this top level method can call the first, then early exit if needed, and then call the second.

GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
```diff
@@ -141,12 +141,13 @@ public void GetMetadataForHeartBeatDoesSetsEventLevelWToInformationalWhenPlaceho
...
 
                 // "ModifiedPathsCount" should be 1 because ".gitattributes" is always present
-                metadata.Count.ShouldEqual(5);
+                metadata.Count.ShouldEqual(6);
                 metadata.ShouldContain("ProcessName1", "GVFS.UnitTests.exe");
                 metadata.ShouldContain("ProcessCount1", 1);
                 metadata.ShouldContain("ModifiedPathsCount", 1);
                 metadata.ShouldContain("PlaceholderCount", 1);
                 metadata.ShouldContain(nameof(RepoMetadata.Instance.EnlistmentId), RepoMetadata.Instance.EnlistmentId);
+                metadata.ContainsKey("PhysicalDiskInfo");
```
wilbaker: Validate the result of `ContainsKey` (e.g. `.ShouldBeTrue()`)?

GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
```diff
@@ -156,14 +157,15 @@ public void GetMetadataForHeartBeatDoesSetsEventLevelWToInformationalWhenPlaceho
...
-                metadata.Count.ShouldEqual(5);
+                metadata.Count.ShouldEqual(6);
 
                 // Only processes that have created placeholders since the last heartbeat should be named
                 metadata.ShouldContain("ProcessName1", "GVFS.UnitTests.exe2");
                 metadata.ShouldContain("ProcessCount1", 2);
                 metadata.ShouldContain("ModifiedPathsCount", 1);
                 metadata.ShouldContain("PlaceholderCount", 3);
                 metadata.ShouldContain(nameof(RepoMetadata.Instance.EnlistmentId), RepoMetadata.Instance.EnlistmentId);
+                metadata.ContainsKey("PhysicalDiskInfo");
```
wilbaker: Same thing here, add a `ShouldContainKey` or check the return value.

PR #540, opened 2018-11-20
Start the move to yaml pipelines.

jrbriggs:
It's better to have the pipeline config in code, so let's move to yaml.


.azure-pipelines/templates/windows-functional-test.yml
```diff
@@ -0,0 +1,34 @@
...
+      artifactName: FunctionalTests_$(platformFriendlyName)_$(configuration)
+      downloadPath: $(Build.BinariesDirectory)
+
+  - script: $(Build.BinariesDirectory)/FunctionalTests_$(platformFriendlyName)_$(configuration)/src/Scripts/ReinstallGVFS.bat $(configuration)
+    displayName: Run VFSForGit and G4W installers
+
+  - script: git config --global credential.interactive never
+    displayName: Disable interactive auth
+
+  - script: cmdkey /generic:LegacyGeneric:target=git:https://gvfs.visualstudio.com /user:"Agent PAT" /password:$(VFSForGitOrgCodePat)
```
jrbriggs: I can remove this now

.azure-pipelines/templates/windows-functional-test.yml
```diff
@@ -0,0 +1,34 @@
...
+    displayName: Save agent password is credential store
+
+  - script: cmdkey /generic:LegacyGeneric:target=git:https://gvfs.visualstudio.com/ci/_git/ForTests /user:"Agent PAT" /password:$(VFSForGitOrgCodePat)
```
jrbriggs: Same here.

.azure-pipelines/continuous-integration.yml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  - job: Build_Windows
+    variables:
+      platformFriendlyName: Windows
+    strategy:
+      maxParallel: 2
+      matrix:
+        #debug:
+        #  configuration: Debug
```
kewillford: Should we be building both Debug and Release?  Or should this be a config variable so we can chose for the build?  Having this commented out doesn't seem right if we have to uncomment and PR to run a Debug build.

.azure-pipelines/continuous-integration.yml
```diff
@@ -0,0 +1,53 @@
...
+    variables:
+      platformFriendlyName: Windows
+      configuration: Release
+    pool:
+      name: FunctionalTests
+    dependsOn: Build_Windows
+    condition: succeeded()
+    steps:
+    - checkout: none # We'll get the test enlistment from Build_Windows job
+    - template: templates/windows-functional-test.yml
```
kewillford: When will the functional tests for Mac run?

.azure-pipelines/pull-request.yml
```diff
@@ -0,0 +1,22 @@
...
+    pool:
+      vmImage: vs2017-win2016
+    steps:
+    - template: templates/windows-build-and-unit-test.yml
+
+  - job: macOS
+    pool:
+      vmImage: macOS 10.13
+    steps:
+    - template: templates/macos-build-and-unit-test.yml
```
kewillford: Will functional tests not run on pull request anymore?

.azure-pipelines/templates/windows-build-and-unit-test.yml
```diff
@@ -0,0 +1,40 @@
+steps:
+
+  - script: $(Build.Repository.LocalPath)\\Scripts\\NukeBuildOutputs.bat
```
kewillford: Shouldn't this be running on a clean VM and not need this step?

.azure-pipelines/templates/windows-functional-test.yml
```diff
@@ -0,0 +1,34 @@
+steps:
+
+  - task: DownloadBuildArtifacts@0
+    displayName: Download functional test enlistment
+    inputs:
+      buildType: current
+      downloadType: specific
+      artifactName: FunctionalTests_$(platformFriendlyName)_$(configuration)
```
kewillford: I think you mentioned this and looking at this, it looks like it is tied to the current build's artifacts so we don't need the build number in the `artifactName` right?

kewillford: Thinking about this a little bit though and what if I want to download artifacts to my machine?  I guess there will only be one and I will need to remember, rename, or check the date to know what build it was?  Just wondering if it would hurt anything to put the build number in the artifact name.

.azure-pipelines/templates/windows-functional-test.yml
kewillford: Is this the right URL?  Will this work with https://dev.azure.com/gvfs?

.azure-pipelines/templates/windows-functional-test.yml
```diff
@@ -0,0 +1,34 @@
+steps:
+
+  - task: DownloadBuildArtifacts@0
+    displayName: Download functional test enlistment
```
kewillford: Should this be Download and run functional tests?

Scripts/CI/CreateBuildDrop.bat
```diff
@@ -0,0 +1,33 @@
...
+SET VFS_STAGEDIR=%2
+
+REM Prepare the staging directories for functional tests.
+IF EXIST %VFS_STAGEDIR% (
+  rmdir /s /q %VFS_STAGEDIR%
+)
+mkdir %VFS_STAGEDIR%\src\Scripts
+mkdir %VFS_STAGEDIR%\BuildOutput\GVFS.Build
+mkdir %VFS_STAGEDIR%\BuildOutput\GVFS.FunctionalTests\bin\x64\%Configuration%\netcoreapp2.1
+mkdir %VFS_STAGEDIR%\BuildOutput\GVFS.FunctionalTests.Windows\bin\x64\%Configuration%\
```
kewillford: Wondering if we really need this deep of a folder structure?  Aren't we already naming the stage directory for the configuration and if bin and x64 are hard coded, do we need those?  Could we use `BuildOutput\GVFS.FunctionalTests.NetCore` and `BuildOutput\GVFS.FunctionalTests.Windows` and remove all the intermediary folders?

sanoursa: I think the idea is that it should match the same layout as if you had run a build on that machine? That way the test scripts/tools don't have to change their understanding of where the binaries live.

jrbriggs: @sanoursa is exactly right, @kewillford. This means the scripts run identically whether you built this from source on your box or if you downloaded the drop.

.azure-pipelines/templates/windows-functional-test.yml
jrbriggs: @kewillford I decided to not include the build number because we don't do that for other artifacts like SetupGVFS.exe. Since you already need a build number to get an artifact, I opted to avoid the duplication of that information.

jrbriggs:
@sanoursa said:
> Do you have any pipelines that are using these new yaml files? Everything seems fine, but I have no way to test or validate any of it

Yep: https://dev.azure.com/gvfs/ci/_build/results?buildId=4060&view=results

jrbriggs:
Ignore the failed check. I can't authorize access to the secret file without having the yaml out of my fork and into Microsoft/VFSForGit. Once I do, the checks should come online too.

GVFS/GVFS.Build/GenerateUnifiedInstallScript.cs
```diff
@@ -0,0 +1,37 @@
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.IO;
+using System.Text;
+
+namespace GVFS.PreBuild
+{
+    public class GenerateUnifiedInstallScript : Task
```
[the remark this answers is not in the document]
jrbriggs: The current install scripts have absolute paths _as built_ on the build agent. They're not portable. This generates a script with relative paths to the installers, so it can be invoked from the portable build drop.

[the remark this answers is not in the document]
jrbriggs: However it's not exactly consistent with how we're doing this on macOS. I'll look at reconciling those behaviors in the morning.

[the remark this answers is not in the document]
jrbriggs: After reviewing options, I slightly prefer how Windows is accomplishing this. I'll look at addressing the macOS behavior in a subsequent PR.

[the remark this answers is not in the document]
jrbriggs: However, I did teach the existing `GenerateInstallScripts` task how to also generate the full product installer, and had them reference the installers in the output directory, so it's cleaner now.

.azure-pipelines/pull-request.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  - job: Windows
+    pool:
+      vmImage: vs2017-win2016
+    steps:
+    - template: templates/windows-build-and-unit-test.yml
+
+  - job: macOS
+    pool:
+      vmImage: macOS 10.13
```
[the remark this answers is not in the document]
jrbriggs: Fair point. However, for now I'm going to leave this broken out so I have a bit more flexibility as I iterate.

.azure-pipelines/continuous-integration.yml
```diff
@@ -0,0 +1,53 @@
...
+variables:
+  majorAndMinorVersion: '0.5'
+  revision: '$(Build.BuildNumber)'
+
+jobs:
+  - job: Build_Windows
+    variables:
+      platformFriendlyName: Windows
+    strategy:
+      maxParallel: 2
```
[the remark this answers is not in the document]
jrbriggs: Thanks, removed.

jrbriggs:
@kewillford to set up the pipeline to post status, i first have to get the yaml file into the branch, as best as i can tell, because i need to authorize access to the secure file we have in the pipeline. Getting it all configured to post statuses will be my next PR. Thanks for the review!

PR #541, opened 2018-11-20
DiagnoseVerb: Option to specify copy-to location for zip

jeschu1:
It seems we often ask users to run diagnose and they copy to a shared drive for us to investigate.

This change allows them to specify the copy-to location as a parameter
i.e. "diagnose -p \\myshare\gvfsdiagnose\jeschu1"
I'm hoping to make this my holiday present to users!

Output looks as below
"Copying zip file.../
C:\Repos\testme21\.gvfs\diagnostics\gvfs_20181120_160608.zip has been copied to \\myshare\gvfsdiagnose\jeschu1\gvfs_20181120_160608.zip.
Copying zip file...Succeeded"

### DIFF
--- a/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
+++ b/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
@@ -23,6 +23,14 @@ namespace GVFS.CommandLine
         {
         }
 
+        [Option(
+            'p',
+            "path",
+            Default = null,
+            Required = false,
+            HelpText = "Path to copy created zip file to")]
+        public string PathToCopyTo { get; set; }
+
         protected override string VerbName
         {
             get { return DiagnoseVerbName; }
@@ -37,7 +45,8 @@ namespace GVFS.CommandLine
                 Directory.CreateDirectory(diagnosticsRoot);
             }
 
-            string archiveFolderPath = Path.Combine(diagnosticsRoot, "gvfs_" + DateTime.Now.ToString("yyyyMMdd_HHmmss"));
+            string diagnoseDirectory = "gvfs_" + DateTime.Now.ToString("yyyyMMdd_HHmmss");
+            string archiveFolderPath = Path.Combine(diagnosticsRoot, diagnoseDirectory);
             Directory.CreateDirectory(archiveFolderPath);
 
             using (FileStream diagnosticLogFile = new FileStream(Path.Combine(archiveFolderPath, "diagnostics.log"), FileMode.CreateNew))
@@ -143,7 +152,7 @@ namespace GVFS.CommandLine
                                 ProductUpgrader.DownloadDirectory,
                                 "downloaded-assets.txt");
                         }
-                     
+
                         return true;
                     },
                     "Copying logs");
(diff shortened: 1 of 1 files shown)


kewillford:
Can we make this even easier for users and set a default share location `\\share\gvfsdiagnose\` and create the username folder if it doesn't exist and copy.  Then the user would only need to run `gvfs diagnose --copy-to-share` or something similar?

sanoursa:
I appreciate the sentiment of what you're trying to do, but I vote against this feature. I would argue that having to discover/remember the syntax for telling the diagnose verb to copy the zip file is more work than just using the copy command. So I don't see any clear win for users here, but I do see us adding a feature that we now have to support on three platforms, add functional tests for, etc.

kewillford:
If you look at a majority of the support emails and they start by saying "Run 'gvfs diagnose' and copy the resulting zip file to share xyz". and while it is only a copy command that needs to be ran, there are a couple of things that slow the process down.
1. It is another step that the user needs to take and some support people go to the share and make sure there is a folder for the user or create one if there isn't already one there.
2. When someone does know about the diagnose command, they sometimes copy the files to a share where we don't have access and there has to be some back and forth before we can look at the files to help with the issue.

I agree that this isn't without risk though in code maintenance and users having issues with the copy not working anyways.  We need to determine if the benefit outweighs the risk/maintenance burden which is hard to argue when it is only a copy command the user needs to run.

sanoursa:
@kewillford this PR doesn't fix those issues though. It would be one thing if we could hardcode a location and always copy it, but we can't (even if we wanted to, good luck getting it to work on macOS). So the user still has to 1) discover the diagnose command, or more likely we have to tell them about it 2) discover the new flag, but actually we have to tell them about it and 3) make sure they discover the right path to copy to and have permissions to it, but in reality we have to tell them that as well.

In practice, the user is going to copy/paste the command we tell them. Whether the network share is an additional argument to the diagnose verb or a second command is hardly relevant if all the user is doing is copy/paste.

On the flip side, baking the copy into the verb is extra maintenance, and it also breaks the philosophy that each tool should do the thing it's designed for.

And lastly, this change only benefits users inside Microsoft who happen to be on corpnet and have access to our network share. It doesn't benefit any external users, and it's not even going to benefit any of our internal macOS users.

The more I think about it, the more I think we should not do this :-)

nickgra:
I agree with Saeed. On MacOS, we have to go through the process of mounting the share and authenticating with it, which isn't as straightforward as navigating to a \\<share> path on Windows. We've had enough trouble with authentication for purely internal tools that I'd prefer not to add that level of cruft to our codebase. 

I do see the problem of users copying the diagnose zips onto the wrong share often and it frustrates me too. 

Perhaps we could try pointing users to a script that lives on a share (so we control it) that would do the following for them when ran from a VFSForGit repo:
1. Run diagnose for the repo
2. Parse the output .zip location from the output
3. Create a folder on our share for this user if it doesn't already exist
4. Copy the zip to the share 

One step for our end users:
`Can you please run \\our\share\Diagnose.cmd from your problematic repo and reply when it's done so I can take a look?`

For MacOS we could do something similar. Have the user mount the share from Finder themselves (sidestepping the auth problem), then run a shell script from /Volumes/OurShare to copy the diagnose zip over for them. Not as seamless, but it's one fewer manual step.

mjcheetham:
How about this compromise?

1. Add an extra, optional `--output | -o` parameter where the ZIP will be placed/copied at the end of the diagnose execution.

    It does nothing special regarding mounting file systems, no `mkdir -p`, and no hard coding paths etc. It just attempts to write to the location and fails gracefully if it does not succeed; fall back to the current default location and write a simple warning message:
    > Failed to write to specified output location Y. Writing to Y instead.

    Plenty of standard tools include an option to direct the location of their file-based output.

2. Instruct internal users to run `\\our\share\diagnose.cmd` or `\\our\share\diagnose.sh` scripts, as suggested by @nickgra, which will force users to have access to (and have mounted) that location. The script will then do a `mkdir -p`, `gvfs diagnose -o \\our\share\%USERNAME%` and whatever we want to do.

    The addition of the `-o` parameter means we don't have to make the script as complicated/have to parse the default tmp output location of the ZIP; we specified it in the script.

The output parameter just doing a simple `try { copy(a, b); } catch { print("warning"); }` shouldn't be that difficult to implement or support.

mjcheetham:
Just to add that on my last expert rotation there were a few users who had errors that I reached out to; they ended up they had out of disk space problems on their drive. When they ran `gvfs diagnose` not all files were copied to the ZIP due to lack of space. Being able to specify another path to collect all the nessesary logs and files (say another volume with more space) would be potentially useful here.

jeschu1:
Thanks for the feedback everyone!  Let's have the discussion offline when everyone is back and then we can update PR accordingly.

jamill:
I like the general idea of making it more convenient / easier to collect and submit diagnostic logs. If we did want to make it easier for specific groups of users (e.g. internal users) to copy logs to a well known location, I like @mjcheetham and @nickgra general suggestions.

The groups distributing VFSForGit could produce a script to make gathering / submitting logs easier (or we could produce a script for these groups to include). This script could be on a network share, or could be included with their deployment vehicle for distributing VFSForGit. It seems they would be in the best position to automate this process for their users.

If there are changes we could make in the diagnose verb (maybe allowing the caller to control the output location, or at least making it easy for callers to get the specific diagnose artifacts created) to make it easier to consume in a helper script, that seems like it could provide value (i.e. making it easier for anyone to consume this functionality) without mixing concerns or making overly specific to internal usage.

wilbaker:
I also like the idea of directing users to an organization specific script that takes care of copying the zip to the correct location.

Regarding the disk usage scenario mentioned by @mjcheetham, another option could be updating `gvfs diagnose` to return a new exit code when it detects that the disk is full.  The script could check for that exit code and ask the user to free up space and run the script again.  I'm not sure that it's worth the added complexity of collecting diagnose logs on a separate volume (and potentially discovering what volume to use).

jrbriggs:
Closing this for now to reduce the number of active PRs.

Issue #544, opened 2018-11-25
Mac kext: Renaming/moving of virtualisation roots is not handled correctly


pmj:
While reviewing the kext code for potential security issues in preparation for #528, I noticed that we store the `vnode` pointer and the path string for each virtualisation root, and that the path is used for relative path calculations.

This means that if the virtualisation root directory is renamed/moved, the path will no longer match the vnode, and the relative path calculation will be incorrect. I'm about to push a change that detects the issue at the string handling level, but we should decide how we want to avoid it in the first place. (For example, we could disallow renaming virtualisation roots with active providers by catching the attempt in the vnode listener and denying it.)

pmj:
After thinking about this some more, I've now realised that blocking renames of the virtualisation root isn't going to fly, as renaming parent directories of the root will still cause the same problem, and are going to be a lot harder to match up to the root - and this behaviour also has the potential to annoy users or break other software.

So my current plan is to detect relative path calculation failures and re-evaluate the path for that root when they happen. Thoughts?

wilbaker:
> So my current plan is to detect relative path calculation failures and re-evaluate the path for that root when they happen. Thoughts?

I think that sounds like the right long term approach.  How big of a change do you think that will be?

I ask because root renames (or renames of their parents) are not currently a supported scenario, and user-mode changes will be required to fully support them.  Right now at a minimum we should make sure that our kext doesn't trigger panics or negatively impact I/O outside of our roots.

If the cost of the minimal changes and the proper fix you're suggesting are about the same I think it makes sense to go with the proper fix.  If the proper fix will be more expensive then I think it can be deferred until later.

cc: @jrbriggs and @sanoursa for their thoughts as well.

pmj:
The minimal changes to detect the problem and defend against it by erroring out are already implemented in #545 and appear to work in my tests. They're essentially a requirement for the proposed more proactive fix.

I think on the kext side this more advanced fix is pretty straightforward actually: if `VirtualizationRoot_GetRootRelativePath()` returns a `nullptr` we need to call a new function which

1. Acquires the lock for the root
2. Grabs a reference to the root vnode
3. Drops the lock
3. Gets the new path for the root vnode
4. Re-acquires the root lock with exclusive access
5. Stores the new root vnode path in the virtualisation root data structure
6. Drops the lock again.

Once that's complete, we can re-run `VirtualizationRoot_GetRootRelativePath()`

There are 2 more potential issues here:

[...]

Issue #528, opened 2018-11-16
Mac: sign PrjFSKext


sanoursa:
Sign the kext so that users don't have to disable SIP

PR #545, opened 2018-11-25
Mac kext review fixes

pmj:
While reviewing the kext code for potential security issues in preparation for #528, I found a few places in the code which aren't ideal, although they probably aren't exploitable.

1. The memory getter & notification port setter in the `IOUserClient` subclasses were not protected by the mutex. It's probably not a serious problem, but doesn't do any harm to fix this.
2. `GetRelativePath()` can be called with 2 unrelated paths. See issue #544. The immediate upshot is that you can easily KP the system by posing as a provider, creating a new root, and renaming it. We'll need to decide what to do about renaming virtualisation roots, but I've changed `GetRelativePath()` to behave more reasonably by returning `nullptr` if the arguments don't make sense, which appears to be safe at the sole call site.
3. I noticed some logging that could be improved by using the path-based logging macro.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
@@ -118,13 +118,20 @@ IOReturn PrjFSProviderUserClient::clientMemoryForType(
     {
     case ProviderMemoryType_MessageQueue:
         {
-            IOMemoryDescriptor* queueMemory = this->dataQueueMemory;
-            if (queueMemory != nullptr)
+            IOMemoryDescriptor* queueMemory;
+            
+            Mutex_Acquire(this->dataQueueWriterMutex);
             {
-                queueMemory->retain(); // Matched internally in IOUserClient
-                *memory = queueMemory;
-                return kIOReturnSuccess;
+                queueMemory = this->dataQueueMemory;
+                if (queueMemory != nullptr)
+                {
+                    queueMemory->retain(); // Matched internally in IOUserClient
+                }
             }
+            Mutex_Release(this->dataQueueWriterMutex);
+            
+            *memory = queueMemory;
+            return nullptr == queueMemory ? kIOReturnError : kIOReturnSuccess;
         }
         break;
     }
@@ -139,13 +146,18 @@ IOReturn PrjFSProviderUserClient::registerNotificationPort(
 {
     if (type == ProviderPortType_MessageQueue)
     {
-        assert(nullptr != this->dataQueue);
         if(port == MACH_PORT_NULL)
         {
             return kIOReturnError;
         }
+
+        Mutex_Acquire(this->dataQueueWriterMutex);
+        {
+            assert(nullptr != this->dataQueue);
+            this->dataQueue->setNotificationPort(port);
+        }
+        Mutex_Release(this->dataQueueWriterMutex);
         
-        this->dataQueue->setNotificationPort(port);
         return kIOReturnSuccess;
     }
     else
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
@@ -352,50 +352,55 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
         }
         else
         {
-            FsidInode vnodeIds = Vnode_GetFsidAndInode(virtualizationRootVNode, vfsContext);
-            uint32_t rootVid = vnode_vid(virtualizationRootVNode);
+            char virtualizationRootCanonicalPath[PrjFSMaxPath] = "";
+            int virtualizationRootCanonicalPathLength = sizeof(virtualizationRootCanonicalPath);
+            err = vn_getpath(virtualizationRootVNode, virtualizationRootCanonicalPath, &virtualizationRootCanonicalPathLength);
             
-            RWLock_AcquireExclusive(s_virtualizationRootsLock);
+            if (0 == err)
             {
-                rootIndex = FindRootAtVnode_Locked(virtualizationRootVNode, rootVid, vnodeIds);
-                if (rootIndex >= 0)
+                FsidInode vnodeIds = Vnode_GetFsidAndInode(virtualizationRootVNode, vfsContext);
+                uint32_t rootVid = vnode_vid(virtualizationRootVNode);
+                
+                RWLock_AcquireExclusive(s_virtualizationRootsLock);
                 {
-                    // Reattaching to existing root
-                    if (nullptr != s_virtualizationRoots[rootIndex].providerUserClient)
-                    {
-                        // Only one provider per root
-                        err = EBUSY;
-                        rootIndex = RootHandle_None;
-                    }
-                    else
-                    {
-                        VirtualizationRoot& root = s_virtualizationRoots[rootIndex];
-                        root.providerUserClient = userClient;
-                        root.providerPid = clientPID;
-                        virtualizationRootVNode = NULLVP; // transfer ownership
-                    }
-                }
-                else
-                {
-                    rootIndex = InsertVirtualizationRoot_Locked(userClient, clientPID, virtualizationRootVNode, rootVid, vnodeIds, virtualizationRootPath);
+                    rootIndex = FindRootAtVnode_Locked(virtualizationRootVNode, rootVid, vnodeIds);
                     if (rootIndex >= 0)
                     {
-                        assert(rootIndex < s_maxVirtualizationRoots);
-                        VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
-                    
-                        strlcpy(root->path, virtualizationRootPath, sizeof(root->path));
-                        virtualizationRootVNode = NULLVP; // prevent vnode_put later; active provider should hold vnode reference
-                    
-                        KextLog_Note("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootPath);
+                        // Reattaching to existing root
+                        if (nullptr != s_virtualizationRoots[rootIndex].providerUserClient)
+                        {
+                            // Only one provider per root
+                            err = EBUSY;
+                            rootIndex = RootHandle_None;
+                        }
+                        else
+                        {
+                            VirtualizationRoot& root = s_virtualizationRoots[rootIndex];
+                            root.providerUserClient = userClient;
+                            root.providerPid = clientPID;
+                            virtualizationRootVNode = NULLVP; // transfer ownership
+                        }
                     }
                     else
                     {
-                        // TODO: scan the array for roots on mounts which have disappeared, or grow the array
-                        KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root");
+                        rootIndex = InsertVirtualizationRoot_Locked(userClient, clientPID, virtualizationRootVNode, rootVid, vnodeIds, virtualizationRootCanonicalPath);
+                        if (rootIndex >= 0)
+                        {
+                            assert(rootIndex < s_maxVirtualizationRoots);
+                        
+                            virtualizationRootVNode = NULLVP; // prevent vnode_put later; active provider should hold vnode reference
+                        
+                            KextLog_Note("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootPath);
+                        }
+                        else
+                        {
+                            // TODO: scan the array for roots on mounts which have disappeared, or grow the array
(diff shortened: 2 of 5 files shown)


pmj:
Some notes:

* I'm not entirely done with the review, this is just what I found specifically looking for user space API issues and string handling problems.
* `GetRelativePath()` seems like the kind of function for which we could easily create unit tests for edge cases. Before I can do this, we need to resolve the questions I've previously raised in #500.
* I still need to do some manual testing on the code for this PR.

pmj:
I've done some testing and the `GetRelativePath()` changes do appear to successfully defend against the 3 edge cases:

 * Path shorter than root path (buffer overflow)
 * Root path not a prefix of path
 * Root path not a parent of path

So I guess this PR is now ready for review.

wilbaker:
> GetRelativePath() seems like the kind of function for which we could easily create unit tests for edge cases.  Before I can do this, we need to resolve the questions I've previously raised about unit testing native Mac code in #500.

Some unit tests here would be great, let me catch up on the discussion in #500.  If we don't already have a work item for adding unit tests for the kext code I can add one.

ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
```diff
@@ -118,13 +118,20 @@ IOReturn PrjFSProviderUserClient::clientMemoryForType(
...
+            
+            Mutex_Acquire(this->dataQueueWriterMutex);
             {
-                queueMemory->retain(); // Matched internally in IOUserClient
-                *memory = queueMemory;
-                return kIOReturnSuccess;
+                queueMemory = this->dataQueueMemory;
+                if (queueMemory != nullptr)
+                {
+                    queueMemory->retain(); // Matched internally in IOUserClient
```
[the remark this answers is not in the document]
pmj: After calling `clientMemoryForType()`, a mapping to user space is made with the returned memory descriptor, after which the latter is released:

https://github.com/apple/darwin-xnu/blob/0a798f6738bc1db01281fc08ae024145e84df927/iokit/Kernel/IOUserClient.cpp#L1645

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -486,13 +486,24 @@ bool VirtualizationRoot_VnodeIsOnAllowedFilesystem(vnode_t vnode)
 
 static const char* GetRelativePath(const char* path, const char* root)
 {
-    assert(strlen(path) >= strlen(root));
+    size_t rootLength = strlen(root);
+    size_t pathLength = strlen(path);
+    if (pathLength < rootLength || 0 != memcmp(path, root, rootLength))
```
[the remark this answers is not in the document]
pmj: I've now tested that change thoroughly and it's working as expected. I've added one more commit that exports the canonical path as a property on the user client's registry entry, so you can see which user client is the provider for which virtualisation root in IORegistryExplorer/`ioreg`.

pmj:
> Some unit tests here would be great, let me catch up on the discussion in #500. If we don't already have a work item for adding unit tests for the kext code I can add one.

As far as I'm concerned, this PR is ready to go once commits 4 & 5 have been reviewed, do you want to keep it open until we've got unit tests?

wilbaker:
> As far as I'm concerned, this PR is ready to go once commits 4 & 5 have been reviewed, do you want to keep it open until we've got unit tests?

Thanks I'll take a look, I think we can add unit tests in a follow up PR that adds unit test infrastructure for the kext.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -352,50 +352,60 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
         }
         else
         {
-            FsidInode vnodeIds = Vnode_GetFsidAndInode(virtualizationRootVNode, vfsContext);
-            uint32_t rootVid = vnode_vid(virtualizationRootVNode);
+            char virtualizationRootCanonicalPath[PrjFSMaxPath] = "";
+            int virtualizationRootCanonicalPathLength = sizeof(virtualizationRootCanonicalPath);
+            err = vn_getpath(virtualizationRootVNode, virtualizationRootCanonicalPath, &virtualizationRootCanonicalPathLength);
```
wilbaker: Super minor:  Update `// Return values:` comment to include that this function can return error value(s) returned by `vn_getpath`.

Also, do you know if our call to `vnode_lookup` will help prevent `vn_getpath` failures?  (I wasn't sure if there was any relationship between the name cache and `vnode_lookup`).

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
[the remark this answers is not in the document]
wilbaker: > I've just pushed a fourth commit to this PR which evaluates the root path using vn_getpath() during provider registration. I still need to do some manual testing on this change though, for which it's now a bit late here, so I'll do that tomorrow.

What is the advantage of this approach over using a case-insensitive string comparison (in place of memcpy)?  Is it a perf concern?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -352,50 +352,60 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
...
-                        virtualizationRootVNode = NULLVP; // prevent vnode_put later; active provider should hold vnode reference
                     
-                        KextLog_Note("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootPath);
-                    }
-                    else
+                    if (0 == err)
                     {
-                        // TODO: scan the array for roots on mounts which have disappeared, or grow the array
-                        KextLog_Error("VirtualizationRoot_RegisterProviderForPath: failed to insert new root");
+                        userClient->setProperty(PrjFSProviderPathKey, virtualizationRootCanonicalPath);
```
wilbaker: Minor, could this code (for `setProperty`) be moved outside of holding `s_virtualizationRootsLock`?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -352,50 +352,60 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
...
-                        root.providerPid = clientPID;
-                        virtualizationRootVNode = NULLVP; // transfer ownership
+                        rootIndex = InsertVirtualizationRoot_Locked(userClient, clientPID, virtualizationRootVNode, rootVid, vnodeIds, virtualizationRootCanonicalPath);
+                        if (rootIndex >= 0)
+                        {
+                            assert(rootIndex < s_maxVirtualizationRoots);
+                        
+                            virtualizationRootVNode = NULLVP; // prevent vnode_put later; active provider should hold vnode reference
+                        
+                            KextLog_Note("VirtualizationRoot_RegisterProviderForPath: new root not found in offline roots, inserted as new root with index %d. path '%s'", rootIndex, virtualizationRootPath);
```
wilbaker: In the log message use `virtualizationRootCanonicalPath` rather than `virtualizationRootPath` (or include both if you think we'd potentially care about them being different).

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
wilbaker: What is the purpose of setting this properly on `userClient`, is it used anywhere?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -486,13 +496,24 @@ bool VirtualizationRoot_VnodeIsOnAllowedFilesystem(vnode_t vnode)
...
+        return nullptr;
+    }
     
-    const char* relativePath = path + strlen(root);
+    const char* relativePath = path + rootLength;
     if (relativePath[0] == '/')
     {
         relativePath++;
     }
+    else if (rootLength > 0 && root[rootLength - 1] != '/' && pathLength > rootLength)
```
wilbaker: Can this check:

`else if (rootLength > 0 && root[rootLength - 1] != '/' && pathLength > rootLength)`

Be simplified to:

`else if (rootLength != pathLength)`

My thinking was that the root path can only be a parent of `path` at this point  if either:

- `relativePath[0] == '/'` is true
- `rootLength == pathLength` (i.e. `relativePath[0]` is `'\0'`)

Can you think of a scenario I'm missing?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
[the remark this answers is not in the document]
pmj: A case insensitive string comparison is neither correct here, nor efficient.

Correctness:
 * `strcasecmp()` only takes into account ASCII characters' case, and isn't locale-specific. It's also not what case-insensitive HFS+ and APFS use.
 * Not all components of the path necessarily reside on case-insensitive file systems.
 * With all paths under consideration now being obtained via `vn_getpath()`, there's no reason there would be a case discrepancy unless a rename operation had taken place.

Efficiency:
 * Case-insensitive comparison is more computationally intensive than simple equality.
 * `memcmp()` can in theory be more efficient than the `strcmp()` family of functions as the latter must not dereference the incoming pointers past a `\0` byte, meaning the only allowed way to implement is is to work byte-by-byte. As we don't care about ordering, I suppose `bcmp()` could be even more efficient, although in practice the kernel versions of none of these functions are optimised for speed.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: It might reduce the likelihood, but as we aren't locking the name cache, nothing can be guaranteed anyway.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: It's a diagnostic aid - you can look in the IO Registry using `ioreg` or IORegistryExplorer and check the path associated with each provider. (This is explained in the commit message IIRC.) I suppose you could argue we should only set it in debug builds, but I don't see any harm, and the updates are infrequent, so not a performance concern.

pmj: Yes, that's safe as long as we don't start updating it elsewhere. My understanding is that we're not going to bother with the more comprehensive root rename/move support from the [refresh-root-path branch](https://github.com/pmj/VFSForGit/tree/refresh-root-path) at least for now, where the thread safety of that becomes slightly more relevant.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: In principle, if root is `"/"`, and path is `"/something/else"`, `relativePath[0]` will not be `'/'` and the lengths won't match. Although of course at the moment that would mean a virtualisation root at /, which isn't terribly practical. I don't *think* `vn_getpath()` should be trailing slashes for any other kind of path, so assuming we'll never call `GetRelativePath()` with `root = "/"` we can make that change. I'm not sure it's worth doing that for performance's sake, considering `relativePath[0] == '/'` will be the hottest code path by far.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Fixed, thanks!

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
wilbaker: > In principle, if root is "/", and path is "/something/else", relativePath[0] will not be '/' and the lengths won't match.

Ahh, I'd been assuming that `root` would never end with `"/"` (even in the case of the root directory).

pmj: I've definitely seen `vn_getpath` return `"/"` for the root vnode. The empty path would be relative, not absolute, so I guess that's why.

wilbaker: > I've definitely seen vn_getpath return "/" for the root vnode. The empty path would be relative, not absolute, so I guess that's why.

Ok cool, thanks again for the details!

PR #547, opened 2018-11-26
Upgrade extensibility design document

jamill:
Document to describe proposed changes to enable upgrade extensibility.

The main purpose of this PR is to enable discussion, not to actually merge this document (we could, if people think it would add value).

This document will be updated as the feature is implemented.

### DIFF
--- a/Documents/UpgradeExtensibility.md
+++ b/Documents/UpgradeExtensibility.md
@@ -10,8 +10,8 @@ components besides just VFSForGit and Git.
 Motivation
 ----------
 
-VFSForGit can exist as a component of a larger engineering
-system. Some environments could include components that are coupled
+VFSForGit can be deployed as a component in a larger engineering
+system. Some engineering systems could include components that are coupled
 with specific versions of VFSForGit and should be updated with with
 VFSForGit. There could be other interactions specific to the
 environment that need to be accounted for when updating VFSForGit. As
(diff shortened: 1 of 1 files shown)


Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+the following directories:
+
+  G4W :: Contains Git for windows installer
+  
+  GVFS :: Contains the VFSForGitInstaller
+  
+  PreSetup :: (Optional) Contains the installer to run before setup
+  
+  PostSetup :: (Optional) Contains the installer to run after setup
+
```
jeschu1: We're basically creating a vehicle for installing anything on a developer box.  How will we ensure this isn't abused?

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+  - Installer logic delegated to ProductUpgrader
+- Remounting repositories
+- Deleting downloaded files
+  - cleanup delegated to Upgrader
+
+#### Upgrade with Alternative feed
+
+When upgrading from NuGet package feed, the NuGet package will be
+downloaded and the individual well know directories extracted. The
+installers from these directories will then be run.
```
jeschu1: Nugent packages generally aren't downloaded and then searched for particular files to run.  I do ensure the file names don't changes and this is extensible going forward.

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+  - Delegate the logic of querying the feed to Upgrader
+- Running prechecks
+- Downloading new version
+  - Download logic delegated to ProductUpgrader
+- Unmounting all repositories
+- Running installer
+  - Installer logic delegated to ProductUpgrader
+- Remounting repositories
+- Deleting downloaded files
+  - cleanup delegated to Upgrader
```
jeschu1: Will this be user initiatied or an autoupgrade?  What will the user experience be like during this process?

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+this feed. It must be a NuGet feed.
+
+### Upgrade Package Contents
+
+If using the alternative feed, the upgrade package must include
+installers for Git and VFSForGit. It can optionally include installers
+to run before / after installing Git and GVFS. The upgrade logic will
+look in the following locations for the pre / post setup installers:
+
+    PreSetup\install.exe :: Run before the Git and VFSForGit
```
nickgra: Do these have to be `install.exe`? I can imagine engineering system teams preferring something lighter weight like a .cmd/.sh script that could run some installers (to gather things like telemetry), configure some global git config values, set up mergetool, etc

nickgra: If we do stick with `.exe`, we'll need a different extension for MacOS.

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+- Running installer
+  - Installer logic delegated to ProductUpgrader
+- Remounting repositories
+- Deleting downloaded files
+  - cleanup delegated to Upgrader
+
+#### Upgrade with Alternative feed
+
+When upgrading from NuGet package feed, the NuGet package will be
+downloaded and the individual well know directories extracted. The
```
nickgra: typo: well know**n**

nickgra: Downloaded via Nuget.exe? If so, will we bundle Mono to run NuGet.exe for cross-plat?

One other option to consider is that .NET Core can restore packages for a .csproj. If the end user has the 'dotnet' CLI, we could create a temporary .csproj (and NuGet.config) that has the PackageReference and then 'dotnet restore' the .csproj's packages into a well known location to handle the downloading/unpacking.

This is how we download our version of VFSForGit-aware Git for our build process (see DownloadGVFSGit.sh).

Documents/UpgradeExtensibility.md
jamill: A script would make some operations easier - I have been thinking about what the right balance is here.

You are right, if we do stick with running an executable, it will have to be tweaked for the different platforms. I think we adjust the extension of the installer that we expect based on the platform being run on, and this would have to follow the same pattern. I will update the document to reflect this.

Documents/UpgradeExtensibility.md
jamill: The plan is to use the NuGet API (provided via the `NuGet.Protocol` NuGet package), rather than bundle NuGet.Exe (or use dotnet CLI).

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+
+    PostSetup\install.exe :: Run after the Git and VFSForGit
+    			     installers are run.
+
+This means the top level structure of the NuGet package will contain
+the following directories:
+
+  G4W :: Contains Git for windows installer
+  
+  GVFS :: Contains the VFSForGitInstaller
```
wilbaker: Nit: `GVFS` should be something like `VFS4G` or `VFSForGit`

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+VFSForGit will support an alternative "Feed" from which updates can be
+retrieved. When an alternative feed is configured, VFSForGit will
+query this feed for updates instead of the normal feed. This
+alternative feed can contain customized components in addition to the
+VFSForGit and Git installers.
+
+### Feed
+
+The alternative feed will be a NuGet feed. The feed URL is specified
+in the gvfs configuration file (gvfs.config) with the following entry:
```
wilbaker: Does modifying the `gvfs.config` file (either manually or with the verb) require elevation? (it's probably good if it does, but I wasn't sure if it did)

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+
+    upgrade.feedurl
+
+This is similar to how upgrade rings are specified. The upgrade.feed
+will take precedence over the upgrade ring configuration. When this
+value is configured, VFSForGit will query and download updates from
+this feed. It must be a NuGet feed.
```
wilbaker: - How will organizations configure their own upgrade rings?

- Should users see a warning\error if they attempt to change their upgrade ring when they have an alternate feed configured?

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+packages from these feeds. In this model, we would have a
+"GitHubReleasesUpgrader" and a "NuGetPackageUpgrader"
+
+The UpgradeOrchestrator will delegate the actual upgrade logic to
+individual ProductUpgraders. The existing upgrade logic for upgrading
+from the public GitHub releases API will be one ProductUpgrader
+implementation. We will implement another ProductUpgrader for the
+upgrades from NuGet feeds.
+
+- Checking for a new version
```
wilbaker: To confirm, are these the steps that the `UpgradeOrchestrator` will run?

Documents/UpgradeExtensibility.md
wilbaker: What version check(s) are performed after downloading installers from the alternate feed?

I'm thinking about a scenario where a user is on their org's fast feed and they switch to their org's slow feed.  If the slow feed has an older version of VFS4G we should not downgrade, and so maybe we shouldn't run any of the other installers either?

Related, what if an org wants to update their custom installers without changing VFS4G or git versions, will that be allowed?

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+installers from these directories will then be run.
+
+### Checking for new upgrades
+
+ProductUpgrader currently contains the logic for determining when an
+upgrade is available. It currently does this by checking if an
+installer has been downloaded.
+
+Now that we have to handle different feed sources, the installers will
+need to follow a convention for us to read the version from, or we
```
wilbaker: >follow a convention for us to read the version from

Which version(s)?

Documents/UpgradeExtensibility.md
```diff
@@ -0,0 +1,157 @@
...
+might want to put the version infromation in a well known file. The
+advantage of storing the version information in a file is that we can
+update this file by querying the feed endpoints, and not necessarily
+waiting to download the actual installer.
+
+### TBD:
```
wilbaker: I think it's worth calling out any scenarios that you don't think we should support (or if we should support them, include the details of how they will work).  

Some scenarios I'm not sure about:

- Switching to\from alternate feeds to public feeds (if a user switches to a public feed, will we run an uninstaller for the custom components)?
- Switching to\from feeds in different orgs
- Switching to\from different feeds in the same org (I assume we do want to support this)

Documents/UpgradeExtensibility.md
alameenshah: > Does modifying the `gvfs.config` file (either manually or with the verb) require elevation? (it's probably good if it does, but I wasn't sure if it did)

No elevation is needed while using the ```config``` verb.

Documents/UpgradeExtensibility.md
alameenshah: We also need to document how VFSForGit would authenticate with the NuGet feed. And how this would affect GVFS.Service's ability to do silent upgrade checks in the background.

alameenshah:
High level control flow

extensible upgrader

jrbriggs:
@jamill can we close this out if the design has been promoted to an issue?

jrbriggs:
Closing this to reduce the number of active PRs.

PR #552, opened 2018-11-27
Early exit a noop blob prefetch

sanoursa:
I'm still looking into ways to just make the index parsing faster and not have to create this extra file, but in the meantime, here's a WIP PR to show what this approach will look like.

In this PR, we save a successful blob prefetch's parameters in a `FileBasedDictionary`. This allows a subsequent prefetch with the same exact parameters (including commit id) to early exit because it's nearly guaranteed to be a noop. This approach does have the potential to create false positives, e.g. the user could have deleted their cache between the last prefetch and the current one, but that will result in a perf hit in those situations, not a correctness issue.

It'll be nicer to avoid this extra fragility if we can, so I'm still investigating alternatives. Right now, the vast majority of the cost of a noop prefetch is in the `DiffHelper.PerformDiff` method, which is just calling `git ls-files -r -t` and then parsing the string output. That is about a constant 20s overhead on the Windows repo. We should be able to get this down to 3-5s, though that will be at a high complexity cost and probably still not fast enough for running in every build, so it may be unworkable in the end.

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
+++ b/GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
@@ -4,6 +4,7 @@ using GVFS.Common.Http;
 using GVFS.Common.Prefetch.Git;
 using GVFS.Common.Prefetch.Pipeline;
 using GVFS.Common.Tracing;
+using Newtonsoft.Json;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
@@ -32,10 +33,13 @@ namespace GVFS.Common.Prefetch
         private const string AreaPath = nameof(BlobPrefetcher);
         private static string pathSeparatorString = Path.DirectorySeparatorChar.ToString();
 
+        private FileBasedDictionary<string, string> lastPrefetchArgs;
+
         public BlobPrefetcher(
             ITracer tracer,
             Enlistment enlistment,
             GitObjectsHttpRequestor objectRequestor,
+            FileBasedDictionary<string, string> lastPrefetchArgs,
             int chunkSize,
             int searchThreadCount,
             int downloadThreadCount,
@@ -52,6 +56,8 @@ namespace GVFS.Common.Prefetch
             this.GitObjects = new PrefetchGitObjects(tracer, enlistment, this.ObjectRequestor);
             this.FileList = new List<string>();
             this.FolderList = new List<string>();
+
+            this.lastPrefetchArgs = lastPrefetchArgs;
 
             // We never want to update config settings for a GVFSEnlistment
             this.SkipConfigUpdate = enlistment is GVFSEnlistment;
@@ -211,10 +217,8 @@ namespace GVFS.Common.Prefetch
 
             this.DownloadMissingCommit(commitToFetch, this.GitObjects);
 
-            // Configure pipeline
-            // LsTreeHelper output => FindMissingBlobs => BatchDownload => IndexPack
+            // For FastFetch only, examine the shallow file to determine the previous commit that had been fetched
             string shallowFile = Path.Combine(this.Enlistment.WorkingDirectoryRoot, GVFSConstants.DotGit.Shallow);
-
             string previousCommit = null;
 
             // Use the shallow file to find a recent commit to diff against to try and reduce the number of SHAs to check.
@@ -229,6 +233,17 @@ namespace GVFS.Common.Prefetch
                 }
             }
 
+            // TODO: make this a separate public method so that the caller can exit even earlier. Right now we're still wasting time
+            //       on auth and config even when we know there will be nothing to do
+            // TODO: always log the details from the last prefetch, to enable debugging of unexpected behaviors
+            if (this.IsNoopPrefetch(commitToFetch, hydrateFilesAfterDownload))
+            {
+                // TODO: prevent the CLI output from saying 0 matched/already cached
+
+                this.Tracer.RelatedInfo("This prefetch matches the last successful one, so there's nothing new to download");
+                return;
+            }
+            
             BlockingCollection<string> availableBlobs = new BlockingCollection<string>();
 
             ////
@@ -340,6 +355,15 @@ namespace GVFS.Common.Prefetch
                 {
                     this.HasFailures |= !this.UpdateRefSpec(this.Tracer, this.Enlistment, branchOrCommit, refs);
                 }
+            }
+
+            if (!this.HasFailures)
+            {
+                this.SavePrefetchArgs(commitToFetch, hydrateFilesAfterDownload);
+            }
+            else
+            {
+                this.SavePrefetchArgs(null, false);
             }
         }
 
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -229,6 +233,17 @@ public virtual void Prefetch(string branchOrCommit, bool isBranch)
                 }
             }
 
+            // TODO: make this a separate public method so that the caller can exit even earlier. Right now we're still wasting time
+            //       on auth and config even when we know there will be nothing to do
```
kewillford: Would this be easier if the data was abstracted into a `PrefetchRunData` class that wraps the `FileBasedDictionary` so it could be created and used outside this class?

sanoursa: I don't think a new wrapper class will change anything. The caller is already the one instantiating the class, so that's not the issue. Just need to put this code in a public method, is all.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -229,6 +233,17 @@ public virtual void Prefetch(string branchOrCommit, bool isBranch)
                 }
             }
 
+            // TODO: make this a separate public method so that the caller can exit even earlier. Right now we're still wasting time
+            //       on auth and config even when we know there will be nothing to do
+            // TODO: always log the details from the last prefetch, to enable debugging of unexpected behaviors
+            if (this.IsNoopPrefetch(commitToFetch, hydrateFilesAfterDownload))
+            {
+                // TODO: prevent the CLI output from saying 0 matched/already cached
```
sanoursa: Note: this is the cause of the one test failure:

```
2018-11-27T18:52:57.9017551Z Stats:
2018-11-27T18:52:57.9017646Z   Matched blobs:    0
2018-11-27T18:52:57.9017724Z   Already cached:   0
2018-11-27T18:52:57.9017808Z   Downloaded:       0
```

sanoursa: (Which means the noop behavior is working. Just need to update the output, and also fix up the tests to expect or prevent the noop)

sanoursa:
This PR is ready for review. No amount of optimization in index parsing is going to get us down to a sub-second noop prefetch, so this seems like a decent approach in the end.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -129,6 +149,44 @@ public static bool TryLoadFileList(Enlistment enlistment, string filesInput, Lis
...
+            ITracer tracer,
+            FileBasedDictionary<string, string> lastPrefetchArgs,
+            string commitId,
+            List<string> files,
+            List<string> folders,
+            bool hydrateFilesAfterDownload)
+        {
+            if (lastPrefetchArgs != null &&
+                lastPrefetchArgs.TryGetValue(PrefetchArgs.CommitId, out string lastCommitId) &&
+                lastPrefetchArgs.TryGetValue(PrefetchArgs.Files, out string filesString) &&
```
wilbaker: For consistency could you prepend "last" on the other out parameters as well? (e.g. lastFilesString, lastFoldersString, etc.)

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -129,6 +149,44 @@ public static bool TryLoadFileList(Enlistment enlistment, string filesInput, Lis
...
+
+                tracer.RelatedEvent(
+                    EventLevel.Informational,
+                    "BlobPrefetcher.IsNoopPrefetch", 
+                    new EventMetadata
+                    {
+                        { PrefetchArgs.CommitId, lastCommitId },
+                        { PrefetchArgs.Files, filesString },
+                        { PrefetchArgs.Folders, foldersString },
+                        { PrefetchArgs.Hydrate, hydrateString },
```
wilbaker: Also trace the new values?

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -341,6 +397,15 @@ public virtual void Prefetch(string branchOrCommit, bool isBranch)
...
                 }
             }
+
+            if (!this.HasFailures)
+            {
+                this.SavePrefetchArgs(commitToFetch, hydrateFilesAfterDownload);
+            }
+            else
+            {
+                this.SavePrefetchArgs(null, false);
```
wilbaker: Is there a downside to leaving the old prefetch args (from the last successful prefetch) on disk if this prefetch fails?

That would make the failure case consistent with the scenario where FastFetch crashes or is killed.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -226,32 +264,72 @@ private void PrefetchCommits(ITracer tracer, GVFSEnlistment enlistment, GitObjec
...
-                enlistment,
-                blobRequestor,
-                ChunkSize,
-                SearchThreadCount,
-                DownloadThreadCount,
-                IndexThreadCount);
+                Path.Combine(enlistment.DotGVFSRoot, "LastBlobPrefetch.dat"),
+                new PhysicalFileSystem(),
+                out lastPrefetchArgs,
+                out error);
```
wilbaker: Can we log something if `FileBasedDictionary<string, string>.TryCreate` fails?

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
wilbaker: Maybe log all of them when `"Result"` is false?  My main concern was being able to track down why prefetch might have decided that the operation was not a no-op.

In addition to the "expanded paths out of a file" scenario, we're also not logging the new commit ID anywhere.

sanoursa: I suspect more about false positives rather than false negatives. I just went ahead and logged it in all cases.

e.g. for the command
```
Saeeds-MacBook-Pro:GVFSTest sanoursa$ ~/Repos/VFSForGit/Publish/gvfs prefetch --folders 'GVFS;ProjFS.Mac;MirrorProvider' --hydrate
All requested files are already available. Nothing new to prefetch.
```

Got the log:
```
[2018-11-28 14:59:28 -08:00] Prefetch (Start) {"Version":"0.2.173.2","EnlistmentRoot":"/Users/sanoursa/GVFSTest","Remote":"https://dev.azure.com/gvfs/ci/_git/ForTests","ObjectsEndpoint":"https://dev.azure.com/gvfs/ci/_git/ForTests"}
[2018-11-28 14:59:28 -08:00] PerformPrefetch {"Commits":false,"Files":"","Folders":"GVFS;ProjFS.Mac;MirrorProvider","FoldersListFile":"","HydrateFiles":true}
[2018-11-28 14:59:29 -08:00] BlobPrefetcher.IsNoopPrefetch {"LastCommitId":"2797fbb8358bb2e0c12d6f3b42a60b43f7655edf","LastFiles":"[]","LastFolders":"[\"/Users/sanoursa/GVFSTest/src/GVFS/\",\"/Users/sanoursa/GVFSTest/src/ProjFS.Mac/\",\"/Users/sanoursa/GVFSTest/src/MirrorProvider/\"]","LastHydrate":"True","NewCommitId":"2797fbb8358bb2e0c12d6f3b42a60b43f7655edf","NewFiles":"[]","NewFolders":"[\"/Users/sanoursa/GVFSTest/src/GVFS/\",\"/Users/sanoursa/GVFSTest/src/ProjFS.Mac/\",\"/Users/sanoursa/GVFSTest/src/MirrorProvider/\"]","NewHydrate":"True","Result":true}
[2018-11-28 14:59:29 -08:00] Prefetch (Stop) {"DurationMs":615}
```

wilbaker: > I suspect we'll care more about false positives rather than false negatives. I just went ahead and logged it in all cases.

That's a good point, I agree false positives would be worse than false negatives.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -226,32 +264,75 @@ private void PrefetchCommits(ITracer tracer, GVFSEnlistment enlistment, GitObjec
...
-            if (!BlobPrefetcher.TryLoadFolderList(enlistment, this.Folders, this.FoldersListFile, blobPrefetcher.FolderList, out error))
+            
+            if (!FileBasedDictionary<string, string>.TryCreate(
+                    tracer,
+                    Path.Combine(enlistment.DotGVFSRoot, "LastBlobPrefetch.dat"),
+                    new PhysicalFileSystem(),
+                    out lastPrefetchArgs,
+                    out error))
+            {
+                tracer.RelatedWarning("Unable to warn last prefetch args: " + error);
```
wilbaker: Is there a typo here?

`Unable to warn`

Should this say "Unable to load" (or something similar)?

sanoursa: Typing faster than I can think apparently :-). I'll fix this up.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -226,32 +264,75 @@ private void PrefetchCommits(ITracer tracer, GVFSEnlistment enlistment, GitObjec
...
-                SearchThreadCount,
-                DownloadThreadCount,
-                IndexThreadCount);
-
             string error;
-            if (!BlobPrefetcher.TryLoadFolderList(enlistment, this.Folders, this.FoldersListFile, blobPrefetcher.FolderList, out error))
+            
+            if (!FileBasedDictionary<string, string>.TryCreate(
+                    tracer,
+                    Path.Combine(enlistment.DotGVFSRoot, "LastBlobPrefetch.dat"),
```
nickgra: Should this go into the .gvfs/databases folder or the .gvfs root?

Scripts/InitializeEnvironment.bat
```diff
@@ -3,6 +3,7 @@
 REM Set environment variables for interesting paths that scripts might need access to.
 PUSHD %~dp0
 SET VFS_SCRIPTSDIR=%CD%
+POPD
```
nickgra: This is just cleanup and not related to your change right? (I didn't know Windows had pushd/popd so I learned something new!)

GVFS/GVFS/CommandLine/PrefetchVerb.cs
sanoursa: This file has the same format as the other files in the `databases` folder, but those are required files for correct functionality, and this is just a convenient way to store the last prefetch args, not required for anything to work right. So no, I don't think it should go in the same folder.

We don't have any other files that look like this one, so I didn't see a logical place to put it other than the `.gvfs` root. The nice thing is that if we ever decide to move it, it's a non-breaking change. Just delete the old one and let the next prefetch recreate it wherever needed.

Scripts/InitializeEnvironment.bat
sanoursa: Yes I'm just fixing a bug in this script. I hit this because when I tried running the prefetch functional tests, it kept changing my current directory.

Issue #242, opened 2018-09-04
Mac: The read-object hook launched by `git push` hangs if any objects are downloaded


wilbaker:
On the Mac if `git push` needs to download any objects, the read-object hook will get stuck on the call to `fread` in `packet_bin_read`.

So far I've only found this issue with `git push`.

*`git cat-file` example (works)*

```
>git cat-file -t 1f337200b18a31b39d858e8117ab1a03402b8e1b
Starting read-object hooj 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Inside read-object loop 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
DownloadSHA: 1f337200b18a31b39d858e8117ab1a03402b8e1b 
Downloaded: 1f337200b18a31b39d858e8117ab1a03402b8e1b S 
Inside read-object loop 
Waiting on fread ... commit
fread read 0 
>
```

**Note** In the tracing above here "commit" on this line:

`Waiting on fread ... commit`

Is the output of the `git cat-file`command.

*`git push` example with the same commit object being downloaded (hangs)*

```
>git push origin users/wilbaker/read-object-test
Starting read-object hooj 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Inside read-object loop 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
Waiting on fread ... fread read 4 
DownloadSHA: 1f337200b18a31b39d858e8117ab1a03402b8e1b 
Downloaded: 1f337200b18a31b39d858e8117ab1a03402b8e1b S 
Inside read-object loop 
Waiting on fread ... 
```

At this point if I attach the debugger to read-object the call to `fread` will complete and read-objet will exit.

You can find the commit with extra read-object tracing here:

https://github.com/wilbaker/VFSForGit/commit/e5bb1b19482503e4ba1aad1963de3abe001c8838

wilbaker:
I noticed when the hang happens these processes are running:

```
~/Repos/VFSForGit/src>ps -ef | grep git
  502  3647   410   0  3:18PM ttys002    0:00.00 grep git
[...]

PR #553, opened 2018-11-27
Update git version to 2.19.1.gvfs.1.7.g43ac102

kewillford:
This change updates git to not spawn the read-object process in send-pack after the pack-objects process has been spawned but instead read-object will be spawned by the pack-objects process.  This was needed because the stdin/stdout/stderr handles were changed when the read-object process was spawned after the pack-objects process started but send-pack still needed to use the handles of pack-objects.


kewillford:
Fixes #242

PR #567, opened 2018-11-30
Mac: Kill any git/gvfs processes when cleaning up after Functional Tests

nickgra:
In some cases, a developer might make a change that causes the functional tests to hang or take an extremely long time. The job will time out after 120 minutes and we don't currently clean up any processes that could still be running. We should pkill any git/gvfs processes still running while we're cleaning up.

This is similar to what we're doing in UnInstallGVFS.bat on Windows.

### DIFF
--- a/Scripts/Mac/CleanupFunctionalTests.sh
+++ b/Scripts/Mac/CleanupFunctionalTests.sh
@@ -1,5 +1,10 @@
 . "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
 
+pkill -9 -l GVFS.FunctionalTests
+pkill -9 -l git
+pkill -9 -l gvfs
+pkill -9 -l prjfs-log
+
 $VFS_SRCDIR/ProjFS.Mac/Scripts/UnloadPrjFSKext.sh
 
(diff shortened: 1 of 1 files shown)


Scripts/Mac/CleanupFunctionalTests.sh
```diff
@@ -1,5 +1,8 @@
 . "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
 
+pkill -l git
+pkill -l gvfs
```
sanoursa: Will this also take down `gvfs.mount`?

pmj: We should probably also kill the prjfs-log tool unless that's already handled elsewhere.

nickgra: This will take down gvfs.mount as well as it matches substrings per the man page for pkill. I've validated this locally and confirmed that the PIDs that were killed match both 'gvfs' and 'gvfs.mount' processes that were running.

Thanks for catching `prjfs-log`!

jeschu1:
Thanks for looking into this @nickgra!

PR #568, opened 2018-11-30
Fallback to slow method of getting process

kewillford:


GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs
```diff
@@ -47,9 +48,23 @@ public static bool IsProcessActiveImplementation(int processId)
...
-                return false;
+                    return false;
+                }
+                else
+                {
+                    // The process.IsInvalid may be true when the mount process doesn't have access to call
+                    // OpenProcess for the specified processId. Fallback to slow way of finding process.
+                    try
+                    {
+                        Process.GetProcessById(processId);
```
sanoursa: What are the perf implications of this change?

sanoursa: I can see that some ProjFS callbacks end up in this code path, so we do need to be careful here. How much will this impact git command perf if a user mounts non-elevated and then runs git commands elevated?

kewillford: From an old thread by @benpeart a checkout that was taking 28 seconds 2 seconds was spent in `GetProcessById`.  This code has to be able to answer the question if the process is active reliably.  Right now the answer correct if we return `true` but it really means `maybe` when we return `false` which is the more important answer to get correct since it is when we release the lock because of what it thinks is a process that doesn't exist.  I will see if there are other changes we can make so that this code path is not hit as often

wilbaker:
If you end up going with the changes in this PR can you update the description to include the problem that the changes are fixing?

jeschu1:
@kewillford I updated your description to reference the bug we have.

GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs
jeschu1: To verify I understand correctly, the issue will happen when:
1. GVFS Mount was run with user permissions
2. User runs a git command elevated and the lock is is in use when we try to acquire it
IsProcessActive simply returns false since it can't open the Process and we incorrectly give out the lock.

Now we'll fallback onto the old slower code, and may be delayed depending on how long it takes to call GetProcesses.  However, we won't erroneously hand out locks.  

From a telemetry standpoint is it worth noting the perf hit when we go down the slow path?

wilbaker: [Updating thread with our conversation offline]

> From an old thread by @benpeart a checkout that was taking 28 seconds 2 seconds was spent in GetProcessById.

In that thread it looked like this was happening in the post-command hook, and I'm not sure that we actually care if the process is still running then.  Could we update the post-command hook to stop checking if the process is active (it can still compare PIDs if there is an external lock holder set)?

kewillford: I am working on some changes that will only check the process for active in the case where it will be releasing the lock since in the other code paths there are comments that indicate they don't care if it has terminated.

wilbaker: > if GetProcesses takes seconds to run there could still be a delay in starting a process right? Is that something worth auditing?

I think it's worth confirming that we're only hitting this code path where we absolutely need to.  

Under normal circumstances the pre-command hook shouldn't need to hit this code path because we only need to go down this code path when GVFS.Mount.exe thinks that there is an external process holding the lock (and usually the lock is cleaned up correctly when a git command completes).

kewillford: Just to clarify the steps I was seeing with this issue
1. GVFS.Mount is running as a non-admin user and not elevated
2. git is ran with admin user elevated
3. git command causes background operations to be added to the queue
4. 3 causes the GVFS.Mount BG operations to check if it can take the lock for GVFS
5. Because the GVFS.Mount process doesn't get the process back from OpenProcess, the lock is released and the BG operations take the lock for GVFS.
5. git tries to rename the index.lock and fails since it doesn't have the GVFSLock
6. the post command tries to release the lock and gets an error because it's PID doesn't own the lock anymore.

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Windows.cs
```diff
@@ -16,7 +16,7 @@ public static bool IsElevated()
 
         public static bool IsProcessActive(int processId)
         {
-            return WindowsPlatform.IsProcessActiveImplementation(processId);
+            return WindowsPlatform.IsProcessActiveImplementation(processId, tryGetProcessById: false);
```
wilbaker: Why pass in `false` here?

Are the hooks started with the same elevation as their parent git.exe process?

wilbaker: If you're able to confirm that could you add a comment describing why we're passing in `false`?

GVFS/GVFS.UnitTests/Common/GVFSLockTests.cs
```diff
@@ -0,0 +1,206 @@
...
+using GVFS.UnitTests.Virtual;
+using NUnit.Framework;
+using System;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class GVFSLockTests : TestsWithCommonRepo
+    {
+        private static readonly NamedPipeMessages.LockData DefaultLockData = new NamedPipeMessages.LockData(1234, false, false, "git command");
```
wilbaker: Minor: Named parameters might make this easier to follow

GVFS/GVFS.UnitTests/Common/GVFSLockTests.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        [TestCase]
+        public void TryAcquireLockForExternalRequestor()
```
wilbaker: The name of this test doesn't appear to match what it's doing (unless I'm missing something).

Should it be something like `ReleaseLockHeldByExternalProcess_WhenProcessDead`?

GVFS/GVFS.UnitTests/Common/GVFSLockTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            NamedPipeMessages.LockData newLockData = new NamedPipeMessages.LockData(4321, false, false, "git new");
+            NamedPipeMessages.LockData existingExternalHolder;
+            gvfsLock.TryAcquireLockForExternalRequestor(newLockData, out existingExternalHolder).ShouldBeFalse();
+            this.ValidateLockHeld(gvfsLock, DefaultLockData);
+            this.ValidateExistingExternalHolder(DefaultLockData, existingExternalHolder);
+            mockPlatform.ActiveProcesses.Remove(DefaultLockData.PID);
+        }
+
+        [TestCase]
+        public void AbleToAcquireLockAfterTerminatedProcess()
```
wilbaker: Minor naming suggestion:  Something like `TryAcquireLockForExternalRequestor_WhenExternalHolderTerminated` would better match the other `TryAcquireLockForExternalRequestor` tests.

GVFS/GVFS.UnitTests/Common/GVFSLockTests.cs
kewillford: This is the happy path for acquiring and releasing the lock and make sure it is free at the end.  Will rename to `TryAcquireAndReleaseLockForExternalRequestor`.

wilbaker: Thanks! I misread what the test was doing in my last review.

PR #569, opened 2018-12-02
Simplified delegate invocation

Thomas-S-B:
Simplified delegate invocation for leaner code.


derrickstolee:
Hi, @Thomas-S-B! Thanks for the contribution.

I'm pretty sure we have a lot of those in the codebase, so it would be helpful if you could find and replace them all. Otherwise, we've been cleaning them up when we change nearby code.

Thomas-S-B:
Hi Derrick,
ok, i'm replacing all in another PR.

PR #572, opened 2018-12-03
Move to Roslyn-based StyleCopAnalyzer

jamill:
We currently consume StyleCop via the [StyleCop.MSBuild](https://github.com/StyleCop/StyleCop) NuGet package. The project indicates that there is unlikely to be major new development on this, and that the Roslyn-based StyleCopAnalyzers project is recommended for developers who use VS 2015 or later.

There are some existing pain points with the StyleCop.MSBuild component, including:

 - Not being as reliable as desired on non-windows platforms (e.g. on the Mac)
 - Not supporting some newer language features (e.g. declaring out parameters in line).

The Roslyn based StyleCopAnalyzer:

 - Is supported on .NET Framework, Core, and Standard projects.
 - It will also show violations as you type, at design time, and at compile time.
 - Supports actions to automatically fix up violations.

## Transition

The StyleCop.MSBuild component used a Settings.StyleCop configuration file that is not used by StyleCopAnalyzer. The severity of certain rules can be customized with rulesets. After setting up the StyleCopAnalyzer with default rules, I went through the reported violations in the projects and resolved them as follows:

  - For rules that the project did not follow previously, I added an exception to `GVFS.ruleset`
  - For violations that it looks like should have been flagged previously, I fixed up the code in place
  - For violations in projects that might not rules might not checked, I resolved violations in place.

Now projects use our custom ruleset instead of the Microsoft Recommended Managed Rules. The Microsoft Recommended Managed rules included rules that controlled static analysis. However, this analysis was not run by default previously, and is not supported for .NET Core or .NET standard projects. As we are moving to .NET Core / .NET Standard, and we didn't utilize these rules previously, I left them out of our custom ruleset.

## Open Issues
- None that I know of - please let me know if you encounter any


GVFS/GVFS.Common/GVFS.Common.csproj
```diff
@@ -7,6 +7,8 @@
     <TargetFrameworks>netcoreapp2.1;netstandard2.0</TargetFrameworks>
     <Platforms>x64</Platforms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <CodeAnalysisRuleSet>..\..\GVFS.ruleset</CodeAnalysisRuleSet>
+    <StyleCopTreatErrorsAsWarnings>true</StyleCopTreatErrorsAsWarnings>
```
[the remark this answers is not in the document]
jamill: This (`StyleCopeTreatErrorAsWarnings`) was left over from me trying out different settings. This should be removed. The style cop analyzer will generate warnings, which are then treated as errors.

GVFS/GVFS.Installer.Windows/packages.config
```diff
@@ -1,5 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="MicroBuild.Core" version="0.2.0" developmentDependency="true" />
-  <package id="Tools.InnoSetup" version="5.5.9" targetFramework="net461" />
-</packages>
```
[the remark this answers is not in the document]
jamill: That was not intended (I must have removed the file as I was playing around with different versions)

GVFS.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27428.2015
+VisualStudioVersion = 15.0.28307.136
```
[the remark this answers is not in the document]
jamill: No - I reverted this change

GVFS/FastFetch/FastFetch.csproj
```diff
@@ -12,6 +12,7 @@
     <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <CodeAnalysisRuleSet>..\..\GVFS.ruleset</CodeAnalysisRuleSet>
```
jrbriggs: Let's move this into GVFS.props

GVFS.ruleset
```diff
@@ -0,0 +1,102 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Rules for GVFS" Description="Code analysis rules for GVFS projects" ToolsVersion="15.0">
```
nickgra: Are these rules a conversion from the old Settings.Stylecop list or a list of default/recommended warnings? If it's a list of defaults (which I suspect due to the number of rules here vs what we have today), can we start by turning on only the rules we had on in legacy Stylecop and make the infrastructure changes first then as a follow up, have a discussion on what new rules we should turn on and make the codebase clean for those rules?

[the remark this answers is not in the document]
nickgra: I believe the idea here is that these are all warnings that will be elevated to Errors by a global 'Warnings as Errors' property. I feel strongly that we don't want to add thousands of informational warnings that won't break the build.

GVFS/GVFS.Common/GVFS.Common.csproj
```diff
@@ -22,8 +23,10 @@
   <ItemGroup>
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="1.0.165" />
     <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
-    <PackageReference Include="StyleCop.Error.MSBuild" Version="1.0.0" Condition="'$(OS)' == 'Windows_NT'" />
-    <PackageReference Include="StyleCop.MSBuild" Version="5.0.0" Condition="'$(OS)' == 'Windows_NT'" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.0.2">
+      <PrivateAssets>all</PrivateAssets>
```
nickgra: Why are these lines needed only for .NET Core/SDK style projects but not in Framework projects?

GVFS.ruleset
jamill: The `CA*` rules are from the "Managed Binary Analysis" rules that are in the `Microsoft Managed Minimum / Recommended Rules" ruleset. I am still looking at how these rules interact with our build - the MSDN documentation seems to recommend that some of these should be included in custom rulesets.

The stylecop rules have a default severity. For most rules the default severity is "warning". The ones I have overriden to  "None" are those that we are not clean on currently. For some projects, I did clean up some of the existing violations, especially for rules that I thought we generally enforced already.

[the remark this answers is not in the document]
jamill: The rulesets are documented at: [rule-set-reference](https://docs.microsoft.com/en-us/visualstudio/code-quality/rule-set-reference?view=vs-2017)

The `CA*` rules that are included here came from the Microsoft Managed [Minumum ](https://docs.microsoft.com/en-us/visualstudio/code-quality/managed-minimum-rules-rule-set-for-managed-code?view=vs-2017)/ [Recommended ](https://docs.microsoft.com/en-us/visualstudio/code-quality/managed-recommended-rules-rule-set-for-managed-code?view=vs-2017) ruleset. I am still looking at how these rules interact with the build (and how custom rulesets interact with baseline configuration)

GVFS.ruleset
```diff
@@ -0,0 +1,102 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Rules for GVFS" Description="Code analysis rules for GVFS projects" ToolsVersion="15.0">
+  <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
```
sanoursa: Is there a rule for "no var"? :-)

GVFS/GVFS.Common/GVFS.Common.csproj
jamill: .NET Core consume NuGet package via the `PackageReference`, and the `PrivateAssets` and `IncludeAssets` control how dependencies are handled (see [here](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets)). .NET Framework projects can use `PackageReference`, but defaults to using packages.config.

GVFS.ruleset
```diff
@@ -0,0 +1,102 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
jamill: I moved this to be next to GVFS.props. I had it at the top of the repository originally as I considered it a replacement for `Settings.StyleCop`, which was at the root.

GVFS.ruleset
jamill: I didn't see a StyleCop rule for this :)

However, I think var usage *is* actually controllable via [.editorconfig](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017#c-code-style-settings)...

GVFS/GVFS.Common/GVFS.Common.csproj
nickgra: Reading that page closely, it looks like all we should need is `PrivateAssets = all` since PrivateAssets implies that the items will be consumed like IncludeAssets. 
```
Tag | Description | Default Value
IncludeAssets | These assets will be consumed | all
ExcludeAssets | These assets will not be consumed | none
PrivateAssets | These assets will be consumed but won't flow to the parent project | contentfiles;analyzers;build
```

This page makes it sound like just adding `PrivateAssets="All"` to our PackageReference is good enough. Does that not work for us? https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/DotNetCli.md

nickgra:
```
The Microsoft Recommended Managed rules included rules that controlled static analysis. However, this analysis was not run by default previously, and is not supported for .NET Framework or .NET standard projects. As we are moving to .NET Core / .NET Standard, and we didn't utilize these rules previously, I left them out of our custom ruleset.
```
Do you have a pointer to the docs that says this isn't supported on .NET Standard? I would be surprised if it worked on .NET Core but not on Standard. Perhaps the issue is that it can't run on .NET Standard projects that are being consumed by Framework projects?

In the long term, I think we'll want to move everything to .NET Core 3 (if we're able to consume C++/CLI there), should we file a work item to track turning on the Microsoft Recommended Managed Rules when we're fully on .NET Core?

GVFS/GVFS.Common/GVFS.Common.csproj
jamill: Your reading seems reasonable. The existing configuration was written out by the NuGet package manager. I can remove the `IncludeAssets` line and verify it still works.

jamill:
@derrickstolee - Thanks for trying this out.

What I think is happening is that analysis of the XML documentation is not enabled for builds, but when you have the file open in Visual Studio, the (design-time?) analyzer does do analysis of the XML documentation. Can you confirm that you only see this error when the document is open in Visual Studio (you might have to close the file and do a build)?

We could fix this by:

1) Enabling the checking of XML documentation during compile time, which would entail making sure the projects generate XML documentation files

2) Suppressing the existing documentation violations for now, and then deciding later on which ones we want to enable and whether we want to tweak the build to also generate XML documentation files.

jamill:
I removed all the XML documentation rules for now. I will go through and try to make sure I didn't miss any. If you encounter others I have missed, please let me know.

jamill:
@nickgra 

> Do you have a pointer to the docs that says this isn't supported on .NET Standard? I would be surprised if it worked on .NET Core but not on Standard.

I mistyped that. I have updated the text in the PR description, it should state that the static analysis is not supported on .NET Core and .NET Standard. It is only supported on .NET Framework.

> should we file a work item to track turning on the Microsoft Recommended Managed Rules when we're fully on .NET Core?

Possibly. I figure we can follow up with other analysers. There exists a NuGet package with FXCop analysers that we could look at, and they should apply accross all of our projects. We don't have to wait until we are on .NET Core. Maybe when someone starts that work they can open an issue to discuss it.

nickgra:
I'll do some manual testing on Windows and MacOS this afternoon, but this looks great. Thanks for putting this together Jameson!

GVFS/GVFS.Common/GVFS.Common.csproj
nickgra: Super nitpick: You can do this all in one line, which I think is cleaner:
```
    <PackageReference Include="StyleCop.Analyzers" Version="1.0.2" PrivateAssets="all"> 
```

jamill:
Thank you everyone for taking the time to test this and provide feedback!

@kewillford - I like your suggestion. I don't look at this change as the final goal, but a step in keeping the repository in good shape. I will follow up with further issues to continue to improve this.

If anyone hits issues after updating, please let me know. @nickgra had indicated that he had experienced some oddities and had to delete his .vs folder. At one point while switching between this branch and branches without this change I had to delete my BuildOutput folder as these Analyzers were not removed from the solution.

PR #555, opened 2018-11-28
DONTMERGE/WIP: Debug kext io errors

pmj:
I'm using this branch to test the kext logging feature added to the functional test pipeline. There are various non-production-ready commits for logging I/O errors and failed calls to `vnode_getparent()` in some detail. Related to #328 and #338 

I'm not currently expecting to merge this branch, nor does it require review at this point.

PR #576, opened 2018-12-04
Mac kext: Avoid working on file stream vnodes in fileop handler

pmj:
* Named file streams are not files themselves do not have file attributes, so attempting to do so on that type of vnode will fail.
 * In the vnode listener, we've already been checking for this edge case, but the fileop handler thus far has not checked for file forks, and we were getting errors in the KAUTH_FILEOP_CLOSE events.
 * This change moves the named stream check logic into a helper function which we can use from both vnode and fileop handlers. As this affects whether vnodes need to be released, I've unified the logic for this across the entire fileop handler.
 * Named forks can’t be renamed or hard linked, so we don’t expect to get those events for such vnodes, and we don’t need to defend against that case there.

This is one of the outcomes of my ongoing reliability/error logging efforts. (#555)

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
@@ -277,6 +277,30 @@ void KauthHandler_AbortOutstandingEventsForProvider(VirtualizationRootHandle pro
     Mutex_Release(s_outstandingMessagesMutex);
 }
 
+static void UseMainForkIfNamedStream(
+    // In+out params:
+    vnode_t& vnode,
+    // Out params:
+    bool&    putVnodeWhenDone)
+{
+    // A lot of our file checks such as attribute tests behave oddly if the vnode
+    // refers to a named fork/stream; apply the logic as if the vnode operation was
+    // occurring on the file itself. (/path/to/file/..namedfork/rsrc)
+    if (vnode_isnamedstream(vnode))
+    {
+        vnode_t mainFileFork = vnode_getparent(vnode);
+        assert(NULLVP != mainFileFork);
+        
+        if (putVnodeWhenDone)
+        {
+            vnode_put(vnode);
+        }
+        
+        vnode = mainFileFork;
+        putVnodeWhenDone = true;
+    }
+}
+
 // Private functions
 static int HandleVnodeOperation(
     kauth_cred_t    credential,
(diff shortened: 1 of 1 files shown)


ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -277,6 +277,30 @@ void KauthHandler_AbortOutstandingEventsForProvider(VirtualizationRootHandle pro
     Mutex_Release(s_outstandingMessagesMutex);
 }
 
+static void UseMainForkIfNamedStream(
+    // In+out params:
+    vnode_t& vnode,
+    // Out params:
+    bool&    putVnodeWhenDone)
```
wilbaker: Minor, if we're treating `putVnodeWhenDone` as an out parameter we should always assign something to it before leaving the function.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -277,6 +277,30 @@ void KauthHandler_AbortOutstandingEventsForProvider(VirtualizationRootHandle pro
...
+    // occurring on the file itself. (/path/to/file/..namedfork/rsrc)
+    if (vnode_isnamedstream(vnode))
+    {
+        vnode_t mainFileFork = vnode_getparent(vnode);
+        assert(NULLVP != mainFileFork);
+        
+        if (putVnodeWhenDone)
+        {
+            vnode_put(vnode);
+        }
```
wilbaker: What's the purpose of having this block of code?

```
        if (putVnodeWhenDone)
        {
            vnode_put(vnode);
        }
```

The callers of `UseMainForkIfNamedStream` are taking care of calling `vnode_put`

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: Ah yes, that should also be under in+out in the current iteration of the code.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: The intention is that since we're replacing the existing (incoming) vnode under consideration, we should also release it if needed. This would in principle be needed on a code path using `vnode_lookup()`. I subsequently decided that we probably won't see any named forks on our only code path where we use that though. So yeah, it's a toss-up over whether to leave the functionality in the helper function in case we need it in future or take it back out for now.

pmj: I've removed this functionality for now as it's not being used.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: Now that the code for releasing the incoming vnode is removed, this can be a pure output, so I've set it to false in the else clause.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: Thanks for the details here, all that makes sense.

pmj:
Yeah, that comment was definitely out of date - thanks for catching it! I've kept this as 2 separate commits so it'll be easy to pick up the original logic again should the need arise.

PR #578, opened 2018-12-05
Fix bad test by fixing maintenance step lock string

derrickstolee:
The test `PrefetchVerbWithoutSharedCacheTests.PrefetchCommitsToEmptyCache()` only runs with the full test suite, not the PR test suite. This was not run during #493, and hence broke the CI build immediately after.

The fix is a simple constant string change.


derrickstolee:
[Successfull Full Functional Test Build](https://gvfs.visualstudio.com/ci/_build/results?buildId=4452)

Issue #537, opened 2018-11-20
Add coding style guide for contributors to follow.


kewillford:
There are some coding styles that are not covered by stylecop rules like the use of `var` that would be good to document for contributors to follow.

PR #581, opened 2018-12-05
Fix config lookup

jeschu1:
This is a bad bug caused by yours truly.  I'll look at ways to unit/functional test, but in the interim let's fix it.

### DIFF
--- a/GVFS/GVFS.Common/GVFSEnlistment.cs
+++ b/GVFS/GVFS.Common/GVFSEnlistment.cs
@@ -16,7 +16,6 @@ namespace GVFS.Common
         private string gitVersion;
         private string gvfsVersion;
         private string gvfsHooksVersion;
-        private GitProcess gitProcess;
 
         // New enlistment
         public GVFSEnlistment(string enlistmentRoot, string repoUrl, string gitBinPath, string gvfsHooksRoot, GitAuthentication authentication)
@@ -252,25 +251,11 @@ namespace GVFS.Common
 
         private string GetId(string key)
         {
-            GitProcess.ConfigResult configResult = this.GetGitProcess().GetFromLocalConfig(key);
+            GitProcess.ConfigResult configResult = this.CreateGitProcess().GetFromLocalConfig(key);
             string value;
             string error;
-            if (!configResult.TryParseAsString(out value, out error, defaultValue: string.Empty))
-            {
-                return value.Trim();
-            }
-
-            return string.Empty;
-        }
-
-        private GitProcess GetGitProcess()
-        {
-            if (this.gitProcess == null)
-            {
-                this.gitProcess = new GitProcess(this);
-            }
-
-            return this.gitProcess;
+            configResult.TryParseAsString(out value, out error, defaultValue: string.Empty);
+            return value.Trim();
         }
     }
 }
--- a/GVFS/GVFS.UnitTests/Common/GVFSEnlistmentTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/GVFSEnlistmentTests.cs
@@ -1,0 +1,52 @@
+﻿using GVFS.Common;
+using GVFS.Common.Git;
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Mock.Common;
+using GVFS.UnitTests.Mock.Git;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class GVFSEnlistmentTests
+    {
+        private const string MountId = "85576f54f9ab4388bcdc19b4f6c17696";
+        private const string EnlistmentId = "520dcf634ce34065a06abaa4010a256f";
+
+        [TestCase]
+        public void CanGetMountId()
+        {
+            TestGVFSEnlistment enlistment = new TestGVFSEnlistment();
+            enlistment.GetMountId().ShouldEqual(MountId);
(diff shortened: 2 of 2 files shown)


sanoursa:
What are the symptoms of this bug?

jeschu1:
>           What are the symptoms of this bug?

No mountId / enlistmentId in telemetry :-(. Very bad!

sanoursa:
Got it. I would vote to get the test fixes into this PR. There's no pending release, so this is not an emergency fix.

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -255,7 +255,7 @@ private string GetId(string key)
             GitProcess.ConfigResult configResult = this.GetGitProcess().GetFromLocalConfig(key);
             string value;
             string error;
-            if (!configResult.TryParseAsString(out value, out error, defaultValue: string.Empty))
+            if (configResult.TryParseAsString(out value, out error, defaultValue: string.Empty))
```
derrickstolee: Because the `configResult` relies on a Git process, this is hard to unit test. Is it possible to mock out `GetGitProcess()` to inject a custom `GitProcess` class that returns certain values for `GetFromLocalConfig()`?

sanoursa: Yes that is the purpose of `Enlistment.CreateGitProcess`. It is already overridden by `MockGVFSEnlistment` to return a `MockGitProcess`.

I'm not sure when `GVFSEnlistment.GetGitProcess` was added, but it looks like we can probably drop it as a duplicate.

derrickstolee: Good point. I didn't see those mocks initially, so I recreated these things to [get a proof of concept](https://github.com/jeschu1/VFSForGit/pull/1). @jeschu1 you could take those ideas _and do them correctly by using the existing mocks_.

jeschu1: Thanks @sanoursa and @derrickstolee !

This was a good opportunity to clean the class up (dropped GetGitProcess) and add tests.  The MockGVFSEnlistment works nicely here.

Let me know what you think.

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -252,25 +251,10 @@ private void CreateHiddenDirectory(string path)
...
-        private GitProcess GetGitProcess()
-        {
-            if (this.gitProcess == null)
-            {
-                this.gitProcess = new GitProcess(this);
-            }
-
-            return this.gitProcess;
+            GitProcess.ConfigResult configResult = this.CreateGitProcess().GetFromLocalConfig(key);
+            string value, error;
```
sanoursa: Style: we tend to avoid combined declarations.

GVFS/GVFS.UnitTests/Common/GVFSEnlistmentTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using GVFS.UnitTests.Mock.Common;
+using GVFS.UnitTests.Mock.Git;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class GVFSEnlistmentTests
+    {
+        public const string MountId = "85576f54f9ab4388bcdc19b4f6c17696";
```
sanoursa: Why are these public?

GVFS/GVFS.UnitTests/Common/GVFSEnlistmentTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        public const string EnlistmentId = "520dcf634ce34065a06abaa4010a256f";
+
+        [TestCase]
+        public void CanGetMountId()
+        {
+            MockGVFSEnlistment enlistment = this.CreateEnlistment();
```
sanoursa: It's a little bit odd to use a mock object as the object under test. The more pure answer would be to update `GVFSEnlistment` to take a factory method as an argument, but that is overkill here, and would unnecessarily complicate the product code.

You might consider creating a private class here named `TestGVFSEnlistment` that derives from `GVFSEnlistment` and use that instead, so that future enhancements to the common mock object don't interfere with these unit tests.

derrickstolee:
@jeschu1 I requeued your unit test build because the failure was due to a race condition I wrote. The fix is in #586.

GVFS/GVFS.UnitTests/Common/GVFSEnlistmentTests.cs
derrickstolee: We are very inconsistent here. We have `public` constants in `MoveRenameFileTests` but `private` constants in `MoveRenameFileTests_2` and `protected` constants in `GitRepoTests`.

It does seem like we use `private` more often. This could be something to add to the contributor style guide (#537) and to consolidate into a cleanup PR during an expert week. (not in scope for this PR, except to change this variable.)

GVFS/GVFS.UnitTests/Common/GVFSEnlistmentTests.cs
jeschu1: @sanoursa you're correct here.  I created my own TestGVFSEnlistment, so we're protected against future changes to the MockGVFSEnlistment.

PR #585, opened 2018-12-05
Clean up some Mac TODOs

sanoursa:
* Fixed a disabled unit test
* Opened two issues for TODOs that could impact early adopters
* Removed some under construction checks that are no longer needed because the code works properly. Most important are the git version checks.
* Created specific flags for the remaining features that are under construction, and got rid of the generic `IsUnderConstruction` flag. This flag will become very ambiguous as soon as we start lighting up Linux features, so it's best to go with feature-specific flags rather than one generic one.
* Cleaned up some other TODO comments


GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -75,7 +75,8 @@ public static DiffTreeResult ParseFromDiffTreeLine(string line, string repoRoot)
             // Take the mode, sha, operation part and split on a space then add the paths that were split on a tab to the end
             parts = parts[0].Split(' ').Concat(parts.Skip(1)).ToArray();
 
-            if (parts.Length != 6)
+            if (parts.Length != 6 ||
+                parts[5].Contains('\t'))
```
sanoursa: Previously we were relying on a later call to `Path.Combine` to throw if there is a `\t` in the path, but that is brittle and not cross-platform. This check makes the condition explicit.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -18,7 +18,6 @@ public class ProjFSKext : IKernelDriver
 
         public bool IsGVFSUpgradeSupported()
         {
-            // TODO(Mac)
```
sanoursa: This is a rather large work item, covering much more than just this one function, so this TODO isn't adding much value. And there is an issue tracking upgrade on Mac.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1685,7 +1685,6 @@ private void AddParentFoldersToListToKeep(string parentKey, ConcurrentHashSet<st
             PlaceholderListDatabase.PlaceholderData placeholder, 
             out string gitPath)
         {
-            // TODO(Mac) This can be optimized if needed
```
sanoursa: We could optimize this a bit by not calling `Replace` on Mac, but since this only gets called in rare error scenarios, seems unnecessary to worry about is perf.

GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -82,8 +82,6 @@ public override void Execute()
         {
             int exitCode = 0;
 
-            // TODO(Mac): limit the length of the enlistment root based on the length constraints imposed by named pipes
```
sanoursa: This comment was written back when we let .NET create the socket file in its default location, which was rather deep inside some folder structure, and therefore limited the length of the repo root.

We now place the socket file inside the `.gvfs` folder, so there is no longer any such restriction.

sanoursa:
@wilbaker actually marking this as WIP, so feel free to wait on review. I see a couple of things marked `IsUnderConstruction` that might need to block early adoption (specifically it seems that we're currently not enforcing the proper git version on Mac, confirming...)

GVFS/GVFS/CommandLine/ConfigVerb.cs
```diff
@@ -46,7 +46,7 @@ protected override string VerbName
 
         public override void Execute()
         {
-            if (GVFSPlatform.Instance.IsUnderConstruction)
+            if (GVFSPlatform.Instance.SupportsGVFSConfig)
```
sanoursa: This is backwards

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -88,7 +88,7 @@ protected override void Execute(GVFSEnlistment enlistment)
                         // .gvfs
                         this.CopyAllFiles(enlistment.EnlistmentRoot, archiveFolderPath, GVFSConstants.DotGVFS.Root, copySubFolders: false);
 
-                        if (!GVFSPlatform.Instance.IsUnderConstruction)
+                        if (!GVFSPlatform.Instance.SupportsKernelLogs)
```
sanoursa: Also backwards

sanoursa:
This is now ready for review

GVFS/GVFS.Common/Git/DiffTreeResult.cs
jeschu1: Just adding there is already a unit test for this case (ParseFromDiffTreeLine_TwoPathLine) and the associated comment is updated to reflect the change.

GVFS/GVFS/CommandLine/MountVerb.cs
```diff
@@ -276,10 +276,6 @@ private bool TryMount(GVFSEnlistment enlistment, string mountExecutableLocation,
             }
 
             const string ParamPrefix = "--";
-            if (GVFSPlatform.Instance.IsUnderConstruction)
-            {
-                mountExecutableLocation = Path.Combine(ProcessHelper.GetCurrentProcessLocation(), "gvfs.mount");
```
jeschu1: I think this cleanup makes sense.  Previously it would have came in as "GVFS.Mount" but due to the change in MountExecutableName it'll correctly be "gvfs.mount".

jeschu1:
Changes mostly make sense to me!  I'll wait for @wilbaker / @kewillford to chime in as they have more exerience in the area

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -21,9 +21,11 @@ public GVFSPlatform(string executableExtension, string installerExtension)
         public abstract IGitInstallation GitInstallation { get; }
         public abstract IDiskLayoutUpgradeData DiskLayoutUpgrade { get; }
         public abstract IPlatformFileSystem FileSystem { get; }
-        public virtual bool IsUnderConstruction { get; } = false;
         public virtual bool SupportsGVFSService { get; } = true;
         public virtual bool SupportsGVFSUpgrade { get; } = true;
+        public virtual bool SupportsGVFSConfig { get; } = true;
+        public virtual bool SupportsKernelLogs { get; } = true;
+        public virtual bool UsesGitHooksLoader { get; } = false;
```
kewillford: Do you see these eventually being removed at some point when all the platforms we want to support have them?

GVFS/GVFS.Common/Git/DiffTreeResult.cs
kewillford: Are there other invalid path characters that we should be checking for here or do we still rely on `Path.Combine` to catch those?

sanoursa: @jeschu1 yep this was the code change that enabled that unit test to be re-enabled.

@kewillford I don't know, this is all that was required to make the test pass. In reality, the test is not 100% sound, it's just trying to confirm that we're not calling `git diff-tree` with the wrong parameters, but it's doing it in a really round-about way.

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: Yep all of these should be removed as the various platforms catch up in features.

I considered prefixing each of these properties with `UnderConstruction` to make that more clear. Thoughts?

GVFS/GVFS/CommandLine/MountVerb.cs
```diff
@@ -294,12 +290,6 @@ private bool TryMount(GVFSEnlistment enlistment, string mountExecutableLocation,
                     this.StartedByService.ToString()
                 });
 
-            if (GVFSPlatform.Instance.IsUnderConstruction)
```
sanoursa: Not really sure what we've done that fixed this. I remember adding this sleep, and back then connecting to the pipe without sleeping first would fail most of the time. Now it is succeeding reliably. My only guess is that since we've changed the location of the socket file, it has also changed the timing somewhat.

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: Take a look at my latest commit. I like that pattern because it makes it very obvious, both in the platform classes and at every usage site, that these are temporary flags. They are intended to be removed, and not intended to make features optional, for example.

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -120,13 +120,31 @@ public string GVFSPostIndexChangedHookExecutableName
 
             public string MountExecutableName
             {
-                get { return "GVFS.Mount" + this.ExecutableExtension; }
+                get { return "gvfs.mount" + this.ExecutableExtension; }
```
wilbaker: Why was this changed, and is the casing consistent between Mac and Windows?

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -120,13 +120,31 @@ public string GVFSPostIndexChangedHookExecutableName
...
+                bool supportsKernelLogs = true,
+                bool usesGitHooksLoader = false)
+            {
+            }
+
+            public bool SupportsGVFSService { get; }
+            public bool SupportsGVFSUpgrade { get; }
+            public bool SupportsGVFSConfig { get; }
+            public bool SupportsKernelLogs { get; }
+            public bool UsesGitHooksLoader { get; }
```
wilbaker: Is `UsesGitHooksLoader` here because it's on the way out in Windows?

kewillford: I like that this is a lot more explicit.  I only thing that is weird is the `UsesGitHooksLoader` makes it look like that is something that is under construction when it is something that is being removed so maybe all the flags should have `true` defaults and `UsesGitHooksLoader` should be something like `NotUsingGitHooksLoader`?

sanoursa: How about `RequiresDeprecatedGitHooksLoader`?

GVFS/GVFS/CommandLine/MountVerb.cs
wilbaker: Is it lower case now on Windows?

GVFS/GVFS.Common/GVFSPlatform.cs
kewillford: Sounds good.  Thanks!

GVFS/GVFS/CommandLine/MountVerb.cs
wilbaker: It might have also been fixed when we moved to a more recent version of .NET Core

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: The case doesn't matter on Windows, so I just changed this to match the casing on Mac.

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: Yep. See the discussion with Kevin, renaming this to `RequiresDeprecatedGitHooksLoader` to make that more clear.

GVFS/GVFS.Common/GVFSPlatform.cs
wilbaker: Out of curiosity, how did the casing matter on Mac?  I thought it was also case insensitive.

GVFS/GVFS/CommandLine/MountVerb.cs
sanoursa: Windows is case insensitive. We could certainly change the csproj's to all generate the same casing, but it is not required for this change.

GVFS/GVFS.Common/GVFSPlatform.cs
wilbaker: Never mind, I see the other thread on this topic.

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: Hm, good question actually. I keep getting tripped up that bash is case sensitive, but you're right that in theory this shouldn't have mattered on Mac either. The reason I changed it is that the "under construction" code I ripped out of CloneVerb was just replacing "GVFS.Mount" with "gvfs.mount". Let me give it a try and see.

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -120,13 +120,31 @@ public string GVFSPostIndexChangedHookExecutableName
...
+        public class UnderConstructionFlags
+        {
+            public UnderConstructionFlags(
+                bool supportsGVFSService = true,
+                bool supportsGVFSUpgrade = true,
+                bool supportsGVFSConfig = true,
+                bool supportsKernelLogs = true,
+                bool usesGitHooksLoader = false)
+            {
+            }
```
sanoursa: Kind of a big oversight :-)... set the properties to the passed in params

GVFS/GVFS.Common/GVFSPlatform.cs
sanoursa: Yep upper case worked fine. As you pointed out offline, that under construction flag was originally added back when we had to run `dotnet gvfs.mount.dll`, and it hasn't been needed since we started publishing an executable.

Issue #580, opened 2018-12-05
GVFS prefetch should support --file-list


turbonaitis:
1. In a scenario, where my build tool (like Buck, or Bazel), can give me a list of input files, needed to build a target, it would be great to be able to provided this list of files via an input file, instead of having to pass all of them in as a command-line argument
2. It would be even better, if --file-list would accept that list via stdin, if no input file is specified, so one could do
```
getListOfInputs | gvfs prefetch --file-list
```

PR #588, opened 2018-12-06
Allow --files-list option for prefetch

turbonaitis:
This PR addresses #580 and adds --files-list option to prefetch verb. It also adds a magic file name, to allow passing in both --files-list and --folders-list via stdin.

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
+++ b/GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
@@ -16,6 +16,7 @@ namespace GVFS.Common.Prefetch
 {
     public class BlobPrefetcher
     {
+        public const string StdInFileName = "stdin";
         protected const string RefsHeadsGitPath = "refs/heads/";
 
         protected readonly Enlistment Enlistment;
@@ -85,64 +86,73 @@ namespace GVFS.Common.Prefetch
 
         public static bool TryLoadFolderList(Enlistment enlistment, string foldersInput, string folderListFile, List<string> folderListOutput, out string error)
         {
-            folderListOutput.AddRange(
-                foldersInput.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
-                .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: true)));
+            return TryLoadFileOrFolderList(enlistment, foldersInput, folderListFile, true, folderListOutput, s => s.Contains("*") ? "Wildcards are not supported for folders. Invalid entry: " + s : null, out error);
+        }
 
-            if (!string.IsNullOrWhiteSpace(folderListFile))
+        public static bool TryLoadFileList(Enlistment enlistment, string filesInput, string filesListFile, List<string> fileListOutput, out string error)
             {
-                if (File.Exists(folderListFile))
+            return TryLoadFileOrFolderList(enlistment, filesInput, filesListFile, false, fileListOutput, s =>
                 {
-                    IEnumerable<string> allLines = File.ReadAllLines(folderListFile)
-                        .Select(line => line.Trim())
-                        .Where(line => !string.IsNullOrEmpty(line))
-                        .Where(line => !line.StartsWith(GVFSConstants.GitCommentSign.ToString()))
-                        .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: true));
+              if (s.IndexOf('*', 1) != -1)
+                {
+                return "Only prefix wildcards are supported. Invalid entry: " + s;
+                }
 
-                    folderListOutput.AddRange(allLines);
+              if (s.EndsWith(GVFSConstants.GitPathSeparatorString) ||
+                  s.EndsWith(pathSeparatorString))
+            {
+                return "Folders are not allowed in the file list. Invalid entry: " + s;
+            }
+
+              return null;
+            }
+            , out error);
+        }
+
+        private static bool TryLoadFileOrFolderList(Enlistment enlistment, string valueString, string listFileName, bool isFolder, List<string> output, Func<string, string> elementValidationFunction, out string error)
+        {
+            output.AddRange(
+                valueString.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: isFolder)));
+
+            if (!string.IsNullOrWhiteSpace(listFileName))
+            {
+                IEnumerable<string> linesFromFile = null;
+                if (File.Exists(listFileName))
+                {
+                    linesFromFile = File.ReadAllLines(listFileName);
+                }
+                else if(listFileName == StdInFileName)
+            {
+                    linesFromFile = Console.In
+                        .ReadToEnd()
+                        .Split('\n', StringSplitOptions.RemoveEmptyEntries)
+                        .Select(x => x.Trim())
+                        .Where(x => !string.IsNullOrWhiteSpace(x));
                 }
                 else
                 {
-                    error = string.Format("Could not find '{0}' for folder list.", folderListFile);
+                    error = string.Format("Could not find '{0}' list file.", listFileName);
                     return false;
                 }
+
+                IEnumerable<string> allLines = linesFromFile
+                        .Select(line => line.Trim())
+                        .Where(line => !string.IsNullOrEmpty(line))
+                        .Where(line => !line.StartsWith(GVFSConstants.GitCommentSign.ToString()))
+                        .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: isFolder));
+
+                    output.AddRange(allLines);
             }
 
-            folderListOutput.RemoveAll(string.IsNullOrWhiteSpace);
+            output.RemoveAll(string.IsNullOrWhiteSpace);
 
-            foreach (string folder in folderListOutput)
-            {
-                if (folder.Contains("*"))
+            string[] errorArray = output.Select(elementValidationFunction).Where(s => !string.IsNullOrWhiteSpace(s)).ToArray();
+
+            if(errorArray != null && errorArray.Length > 0)
                 {
-                    error = "Wildcards are not supported for folders. Invalid entry: " + folder;
+                error = string.Join('\n', errorArray);
                     return false;
-                }
-            }
-
-            error = null;
-            return true;
-        }
-
-        public static bool TryLoadFileList(Enlistment enlistment, string filesInput, List<string> fileListOutput, out string error)
-        {
-            fileListOutput.AddRange(
-                filesInput.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
-                .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: false)));
-
-            foreach (string file in fileListOutput)
-            {
-                if (file.IndexOf('*', 1) != -1)
-                {
-                    error = "Only prefix wildcards are supported. Invalid entry: " + file;
-                    return false;
-                }
-
-                if (file.EndsWith(GVFSConstants.GitPathSeparatorString) ||
-                    file.EndsWith(pathSeparatorString))
-                {
-                    error = "Folders are not allowed in the file list. Invalid entry: " + file;
-                    return false;
-                }
             }
 
             error = null;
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
```diff
@@ -145,6 +145,41 @@ public void PrefetchCleansUpPackDir()
...
+        {
+            var input = string.Join(
+                "\n",
+                new[]
+                {
+                    Path.Combine("GVFS", "GVFS", "packages.config"),
+                    Path.Combine("GVFS", "GVFS.FunctionalTests", "App.config")
+                });
+
+            this.ExpectBlobCount(this.Enlistment.Prefetch("--files-list stdin", standardInput: input), 2);
```
derrickstolee: If someone had a file named "stdin", this would break the existing behavior. Yes, unlikely, but easily fixed by creating new arguments, say `--stdin-files` and `--stdin-folders`.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -45,9 +45,16 @@ public class PrefetchVerb : GVFSVerb.ForExistingEnlistment
             "folders-list",
             Required = false,
             Default = "",
-            HelpText = "A file containing line-delimited list of folders to fetch. Wildcards are not supported.")]
+            HelpText = "A file containing line-delimited list of folders to fetch. Wildcards are not supported. Specify \"stdin\" for stdin")]
         public string FoldersListFile { get; set; }
 
+        [Option(
```
derrickstolee: As mentioned above, this would be better as `--stdin-files`. Unless you are _also_ needing a `--files-list <input-file>`, don't add this parameter. It is a natural parallel to `--folders-list <input-file>`, but unless there is a need let's keep it out.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
```diff
@@ -145,6 +145,41 @@ public void PrefetchCleansUpPackDir()
...
+            {
+                File.Delete("fileList");
+            }
+        }
+
+        [TestCase, Order(13)]
+        public void PrefetchFilesFromFileListStdIn()
+        {
+            var input = string.Join(
+                "\n",
```
derrickstolee: Use `Environment.NewLine`.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
derrickstolee: (Sorry, I see that you are adding `--files-list`, so there is no existing behavior to break. However, doing the same pattern in `--folders-list` would break existing behavior, and it would be best to stay consistent.)

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +479,59 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
             }
         }
 
+        private static bool TryLoadFileOrFolderList(Enlistment enlistment, string valueString, string listFileName, bool isFolder, List<string> output, Func<string, string> elementValidationFunction, out string error)
```
derrickstolee: nit: the tabbing is wrong in several places in this method.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +479,59 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
...
+        {
+            output.AddRange(
+                valueString.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: isFolder)));
+
+            if (!string.IsNullOrWhiteSpace(listFileName))
+            {
+                IEnumerable<string> linesFromFile = null;
+                if (File.Exists(listFileName))
```
derrickstolee: I see here that if the `stdin` file exists, then we will never read from `stdin`, if you needed an extra reason to create `--stdin-files` instead.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +479,59 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
...
+                        .Where(line => !string.IsNullOrEmpty(line))
+                        .Where(line => !line.StartsWith(GVFSConstants.GitCommentSign.ToString()))
+                        .Select(path => BlobPrefetcher.ToAbsolutePath(enlistment, path, isFolder: isFolder));
+
+                    output.AddRange(allLines);
+            }
+
+            output.RemoveAll(string.IsNullOrWhiteSpace);
+
+            string[] errorArray = output.Select(elementValidationFunction).Where(s => !string.IsNullOrWhiteSpace(s)).ToArray();
```
derrickstolee: While you are here, could you split these LINQ calls across multiple lines?

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +479,59 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
...
+                IEnumerable<string> linesFromFile = null;
+                if (File.Exists(listFileName))
+                {
+                    linesFromFile = File.ReadAllLines(listFileName);
+                }
+                else if (listFileName == StdInFileName)
+            {
+                    linesFromFile = Console.In
+                        .ReadToEnd()
+                        .Split(new char[] { '\n' }, StringSplitOptions.RemoveEmptyEntries)
```
derrickstolee: Use `Environment.NewLine`.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +479,59 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
...
+            {
+                IEnumerable<string> linesFromFile = null;
+                if (File.Exists(listFileName))
+                {
+                    linesFromFile = File.ReadAllLines(listFileName);
+                }
+                else if (listFileName == StdInFileName)
+            {
+                    linesFromFile = Console.In
+                        .ReadToEnd()
```
derrickstolee: `ReadToEnd().Split()` seems fraught with performance problems. Please replace with a `Console.ReadLine()` loop. Saves you the trouble with `Environment.NewLine`, too.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +479,59 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
...
+                    output.AddRange(allLines);
+            }
+
+            output.RemoveAll(string.IsNullOrWhiteSpace);
+
+            string[] errorArray = output.Select(elementValidationFunction).Where(s => !string.IsNullOrWhiteSpace(s)).ToArray();
+
+            if (errorArray != null && errorArray.Length > 0)
+                {
+                error = string.Join("\n", errorArray);
```
derrickstolee: `Environment.NewLine`

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
derrickstolee: Visual Studio format should fix that for you.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
turbonaitis: I'll keep the --files-list, to be consistent with --folders-list, but as per your suggestion, will add another pair, --stdin-files and --stdin-folders

turbonaitis:
As per @derrickstolee suggestions, added separate flags to trigger file or folder list loading from stdin

derrickstolee:
@turbonaitis Please leave FastFetch alone. That's kind of a separate project and should not be taken into account just for consistency-sake.

turbonaitis:
@derrickstolee removed. Out of interest, why do you want to prevent new features from being added to FastFetch?

jrbriggs:
@turbonaitis The fact that FastFetch lives in the VFS for Git codebase is a bit of a historical oddity. The projects try to do very different things. VFS for Git wants to virtualize an repository, only hydrating files on-demand. FastFetch wants to hydrate an entire enlistment as fast as it possibly can. The only reason it's here is because it depends on the /gvfs/objects endpoint.

In the future, I see FastFetch lifting out of the VFS for Git repository.

/cc @derrickstolee

turbonaitis:
> @turbonaitis The fact that FastFetch lives in the VFS for Git codebase is a bit of a historical oddity. The projects try to do very different things. VFS for Git wants to virtualize an repository, only hydrating files on-demand. FastFetch wants to hydrate an entire enlistment as fast as it possibly can. The only reason it's here is because it depends on the /gvfs/objects endpoint.
> 
> In the future, I see FastFetch lifting out of the VFS for Git repository.
> 
> /cc @derrickstolee

I understand, that it's a separate tool with a different purpose, but I fail to understand, how adding features to it is a bad thing. In any case, it's your decision - I've updated my PR to not include changes to FastFetch.

turbonaitis:
@derrickstolee are there any more outstanding changes for this PR?

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -83,70 +84,36 @@ public class BlobPrefetcher
 
         public List<string> FolderList { get; }
 
-        public static bool TryLoadFolderList(Enlistment enlistment, string foldersInput, string folderListFile, List<string> folderListOutput, out string error)
+        public static bool TryLoadFolderList(Enlistment enlistment, string foldersInput, string folderListFile, bool readListFromStdIn, List<string> folderListOutput, out string error)
```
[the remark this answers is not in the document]
turbonaitis: it can't be the last argument in the list, as the last argument is `out string error`, so I've moved it to be a penultimate argument. Also, that means, that `error:` needs to be passed by name as well. Both pushed.

wilbaker:
> @wilbaker do you mind taking a look? Most of the low-hanging comments should be covered.

No problem, taking a look now

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -45,9 +45,30 @@ public class PrefetchVerb : GVFSVerb.ForExistingEnlistment
...
+            Required = false,
+            Default = false,
+            HelpText = "Specify this flag to load folder list from stdin. Same format as when loading from file")]
+        public bool FoldersFromStdIn { get; set; }
+
+        [Option(
+            "files-list",
+            Required = false,
+            Default = "",
+            HelpText = "A file containing line-delimited list of files to fetch. Wildcards are supported. Specify \"stdin\" for stdin")]
```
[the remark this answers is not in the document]
turbonaitis: That's a left-over from previous implementation, where you would specify stdin magic file name for stdin. Removed.

GVFS/GVFS.Common/Prefetch/BlobPrefetcher.cs
```diff
@@ -513,9 +480,77 @@ protected void DownloadMissingCommit(string commitSha, GitObjects gitObjects)
...
+            }
+        }
+
+        private static bool TryLoadFileOrFolderList(Enlistment enlistment, string valueString, string listFileName, bool readListFromStdIn, bool isFolder, List<string> output, Func<string, string> elementValidationFunction, out string error)
+        {
+            output.AddRange(
+                GetFilesFromParameter(valueString)
+                .Union(GetFilesFromFile(listFileName, out string fileReadError))
+                .Union(GetFilesFromStdin(readListFromStdIn))
+                .Where(path => !path.StartsWith(GVFSConstants.GitCommentSign.ToString()))
```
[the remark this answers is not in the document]
turbonaitis: My idea was to keep the stdin format exactly the same as input from file. This makes the code simpler (less branching) and also makes `gvfs prefetch --folders-list list.txt` equivalent to `cat list.txt | gvfs prefetch --stdin-folders-list`. Do you think we should have different logic for files vs stdin?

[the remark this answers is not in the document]
wilbaker: > Do you think we should have different logic for files vs stdin?

Thanks for the details here, I think it makes sense to keep them the same.

GVFS/FastFetch/FastFetchVerb.cs
```diff
@@ -171,8 +171,8 @@ private int ExecuteWithExitCode()
             {
                 Console.WriteLine("Cannot use --force-checkout option without --checkout option.");
                 return ExitFailure;
-            }
-            
+            }
+
```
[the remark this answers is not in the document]
wilbaker: @turbonaitis This might actually stop being an issue, see #625 (opened after I created this comment)

turbonaitis:
Thanks for your comments!

Issue #457, opened 2018-11-01
Mac: Detect mode changes to placeholder files


wilbaker:
Related: #223 

`git status` should report placeholder files as modified when their mode changes

wilbaker:
As discussed offline with @sanoursa, this issue is lower priority than #223.  

A file's mode changing is probably more likely than a user changing the mode of a placeholder file (mode changes to full files are reported properly by `git status`)

PR #590, opened 2018-12-06
Allow core.filemode to be configured by platform

### LINKED ISSUE #302: Mac: "core.filemode" should be true
VFS for Git is always setting "core.filemode" to false, even on Mac:

https://github.com/Microsoft/VFSForGit/blob/a074575fe90d925d28d33ea0ea7cb5dbf0c8bb7f/GVFS/GVFS/CommandLine/GVFSVerb.cs#L89

Based on the documentation for "core.filemode" it should be true:

> Tells Git if the executable bit of files in the working tree is to be honored.
>
>Some filesystems lose the executable bit when a file that is marked as executable is checked out, or checks out a non-executable file with executable bit on. git-clone[1] or git-init[1] probe the filesystem to see if it handles the executable bit correctly and this variable is automatically set as necessary.
>
>A repository, however, may be on a filesystem that handles the filemode correctly, and this variable is set to true when created, but later may be made accessible from another environment that loses the filemode (e.g. exporting ext4 via CIFS mount, visiting a Cygwin created repository with Git for Windows or Eclipse). In such a case it may be necessary to set this variable to false. See git-update-index[1].
>
>The default is true (when core.filemode is not specified in the config file).

https://git-scm.com/docs/git-config

See also related issue #223

kewillford:
Fixes #302

### DIFF
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
@@ -189,11 +189,26 @@ namespace GVFS.FunctionalTests.FileSystemRunners
             this.ShouldFail<IOException>(() => { this.ReadAllText(path); });
         }
 
+        public override void ChangeMode(string path, int mode)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                throw new NotSupportedException();
+            }
+            else
+            {
+                Chmod(path, mode).ShouldEqual(0, $"Failed to chmod: {Marshal.GetLastWin32Error()}");
+            }
+        }
+
         [DllImport("kernel32", SetLastError = true)]
         private static extern bool MoveFileEx(string existingFileName, string newFileName, int flags);
 
         [DllImport("libc", EntryPoint = "link", SetLastError = true)]
         private static extern int MacCreateHardLink(string oldPath, string newPath);
+
+        [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
+        private static extern int Chmod(string pathname, int mode);
 
         [DllImport("kernel32.dll", EntryPoint = "CreateHardLink", SetLastError = true, CharSet = CharSet.Unicode)]
         private static extern bool WindowsCreateHardLink(
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
@@ -189,7 +189,7 @@ namespace GVFS.FunctionalTests.FileSystemRunners
             this.ShouldFail<IOException>(() => { this.ReadAllText(path); });
         }
 
-        public override void ChangeMode(string path, int mode)
+        public override void ChangeMode(string path, ushort mode)
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -208,7 +208,7 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         private static extern int MacCreateHardLink(string oldPath, string newPath);
 
         [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
-        private static extern int Chmod(string pathname, int mode);
+        private static extern int Chmod(string pathname, ushort mode);
 
         [DllImport("kernel32.dll", EntryPoint = "CreateHardLink", SetLastError = true, CharSet = CharSet.Unicode)]
         private static extern bool WindowsCreateHardLink(
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
@@ -558,6 +558,18 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
             this.CommitChangesSwitchBranchSwitchBack(fileSystemAction: this.MoveFolder);
         }
 
+        // MacOnly because Windows does not support file mode
+        [TestCase]
+        [Category(Categories.MacOnly)]
+        public void UpdateFileModeOnly()
+        {
+            const string TestFileName = "test-file-mode";
+            this.CreateFile("#!/bin/bash\n", TestFileName);
+            this.ChangeMode(TestFileName, Convert.ToInt32("755", 8));
+            this.ValidateGitCommand($"add {TestFileName}");
+            this.ValidateGitCommand($"ls-files --stage {TestFileName}");
+        }
+
         [TestCase]
         public void AddFileCommitThenDeleteAndCommit()
         {
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
@@ -565,7 +565,7 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         {
             const string TestFileName = "test-file-mode";
             this.CreateFile("#!/bin/bash\n", TestFileName);
-            this.ChangeMode(TestFileName, Convert.ToInt32("755", 8));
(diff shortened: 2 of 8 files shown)


GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -113,7 +113,7 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
                 { "core.safecrlf", "false" },
                 { "core.untrackedCache", "false" },
                 { "core.repositoryformatversion", "0" },
-                { "core.filemode", "false" },
+                { "core.filemode", GVFSPlatform.Instance.GitCoreFileMode ? "true" : "false" },
```
wilbaker: I am not sure how git actually uses this value.  Is it possible to write a functional test that would fail if this value were not set correctly?

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -24,6 +24,7 @@ public GVFSPlatform(string executableExtension, string installerExtension)
         public virtual bool IsUnderConstruction { get; } = false;
         public virtual bool SupportsGVFSService { get; } = true;
         public virtual bool SupportsGVFSUpgrade { get; } = true;
+        public virtual bool GitCoreFileMode { get; } = true;
```
sanoursa: I would drop the default value here. This is something that we should decide explicitly on each platform.

sanoursa: Also, I would move this to `IFileSystem`, as it's more a property of how the files are stored than just a general property of the platform.

kewillford: Should we just use the property that is already there `SupportsFileMode`?  Seems like that would mean the same thing?
CC: @wilbaker

sanoursa: Yep that's what it was for :-). So definitely drop this one.

GVFS/GVFS/CommandLine/GVFSVerb.cs
wilbaker: > I thought there was still some outstanding things there?

The two outstanding issues I'm aware of are:

1. VFS4G will not update a placeholder if *only* its mode changes during a projection change (#223)
2. If *only* the mode of a file is changed VFS4G is not notified and the file is not added to ModifiedPaths.dat (#457)

So if `core.filemode` controls whether git checks a file's mode at all, I'd think a test like this would catch it:

1. Modify file
2. git checkout -- <file>
3. Confirm git status is clean
4. chmod file
5. Confirm git status reports file

GVFS/GVFS.Common/GVFSPlatform.cs
wilbaker: Yeah that sounds right to me too

GVFS/GVFS/CommandLine/GVFSVerb.cs
jamill: We might want to check that the file mode of a file on disk is set correctly after a checkout. I think I noticed some issues with fast fetch where the execute bit was not set correctly after checking out files.

jamill: Do we need to explicitly set this here? I think this will normally be set appropriately by the platform specific git that is running, so on Windows, this is set to false, and on Mac, this is set to True. Does GVFS need to know about and explicitly set this?

wilbaker: > Does GVFS need to know about and explicitly set this?

My recollection is that we set this explicitly in case there is a global or system git config that sets `core.filemode` to something other than the platform default.

kewillford: Looking at history and this was added when the repair was added for the .git/config so when it is corrupted we can fix it by putting the right values in.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -557,6 +557,18 @@ public void MoveFolderCommitChangesSwitchBranchSwitchBackTest()
         {
             this.CommitChangesSwitchBranchSwitchBack(fileSystemAction: this.MoveFolder);
         }
+
+        // MacOnly because renames of partial folders are blocked on Windows
```
kewillford: Update comment

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
```diff
@@ -189,11 +189,26 @@ public override void ReadAllText_FileShouldNotBeFound(string path)
...
+        }
+
         [DllImport("kernel32", SetLastError = true)]
         private static extern bool MoveFileEx(string existingFileName, string newFileName, int flags);
 
         [DllImport("libc", EntryPoint = "link", SetLastError = true)]
         private static extern int MacCreateHardLink(string oldPath, string newPath);
+
+        [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
+        private static extern int Chmod(string pathname, int mode);
```
wilbaker: `mode` should be a `ushort` (@nickgra is fixing this in the product code in #496)

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -557,6 +557,18 @@ public void MoveFolderCommitChangesSwitchBranchSwitchBackTest()
...
+        // MacOnly because Windows does not support file mode
+        [TestCase]
+        [Category(Categories.MacOnly)]
+        public void UpdateFileModeOnly()
+        {
+            const string TestFileName = "test-file-mode";
+            this.CreateFile("#!/bin/bash\n", TestFileName);
+            this.ChangeMode(TestFileName, Convert.ToInt32("755", 8));
+            this.ValidateGitCommand($"add {TestFileName}");
+            this.ValidateGitCommand($"ls-files --stage {TestFileName}");
```
wilbaker: Did you confirm this test fails without your change?

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
kewillford: Will update.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
kewillford: Yep.

GVFS/GVFS/CommandLine/GVFSVerb.cs
wilbaker: > Looking at history and this was added when the repair was added for the .git/config so when it is corrupted we can fix it by putting the right values in.

Yep, for future reference, this was the conversation on this PR that led to this change (note the initial changes in the PR were to call `git init` when repairing the config):

>Saeed 7/10/2017
Why are we calling 'git init'? Why not just create a new empty config file?
It looks like that's what you're doing down below. I don't understand what the 'git init' is for.


> Forest 7/10/2017
git init populates four entries that GVFS does not. I don't know much about them, but I figured this is how clone initializes the config, so it was best to keep it in line with that.

>    [core]
repositoryformatversion = 0
filemode = false
bare = false
logallrefupdates = true


>Saeed 7/11/2017
It was just surprising to see "git init" called on an existing repo. What other side effects does it have besides recreating the config file?
If it's just a matter of creating those 4 config entries, I'd be fine with adding them to the list of entries in GVFSVerb, since they all seem like sensible values, and we're currently relying on them anyway so might as well make it explicit.


>Forest 7/11/2017
From the git man:
Running git init in an existing repository is safe. It will not overwrite things that are already there. The primary reason for rerunning git init is to pick up newly added templates (or to move the repository to another place if --separate-git-dir is given).
> So really, the primary side effect would be "accidentally fixing" any missing folders in .git (refs/heads, objects, etc). but I'm not opposed to just adding those entries to simplify/clarify things here.

kewillford:
@jamill Since Nick is also making some changes for this same method in a different place, I will leave this to match what he has for now and we can change both places in a later PR.

PR #594, opened 2018-12-07
Fix issue building VFS for Git inside of Visual Studio

jamill:
I am open to feedback / alternatives if there is a better way to fix this.

When building inside of Visual Studio, the application manifest for
GVFS.Service.exe.manifest is not created correctly in
GVFS.PreBuild.csproj. It appears the cause of this issue is that you
cannot use Item metadata from an ItemDefinitionGroup:

  Item metadata from an ItemGroup is not useful in an
  ItemDefinitionGroup metadata declaration because ItemDefinitionGroup
  elements are processed before ItemGroup elements.

https://docs.microsoft.com/en-us/visualstudio/msbuild/item-definitions?view=vs-2017

### DIFF
--- a/GVFS/GVFS.Build/GVFS.PreBuild.csproj
+++ b/GVFS/GVFS.Build/GVFS.PreBuild.csproj
@@ -162,7 +162,6 @@
   <ItemDefinitionGroup>
     <ApplicationNeedsManifest>
       <Visible>false</Visible>
-      <OutputFile>$(BuildOutputDir)\%(Identity).exe.manifest</OutputFile>
     </ApplicationNeedsManifest>
   </ItemDefinitionGroup>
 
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Build/GVFS.PreBuild.csproj
```diff
@@ -162,7 +162,6 @@
   <ItemDefinitionGroup>
     <ApplicationNeedsManifest>
       <Visible>false</Visible>
-      <OutputFile>$(BuildOutputDir)\%(Identity).exe.manifest</OutputFile>
     </ApplicationNeedsManifest>
```
jamill: I am not sure we need this `ItemDefinitionGroup` at all with this change (or maybe it is used elsewhere...)...

jrbriggs: Might want to test if removing it causes the .exe.manifest to show as a child of project in the VS solution hierarchy. If it doesn't then we can remove it.

jamill: ah - that it does (well, it causes a `GVFS.service` item to appear). Will keep it!

jamill:
> jrbriggs changed the title from Fix issue building GVFS inside of Visual Studio to Fix issue building VFS for Git inside of Visual Studio 2

In my defense, I am fixing the GVFS.Prebuild project... :)

PR #596, opened 2018-12-07
[WIP] MaintenanceVerb + LooseObjectsStep

jeschu1:
Lot of code here so plenty of room for feedback :-).
Still to do are unit tests & functional tests and the PR will not be completed without them.  

**LooseObjectsStep**
- We call prune-packed to remove loose objects that are already in pack-files
- This is scheduled to run every 12 hours, but only runs if it's been 7 days since last run or ignoreTimeRestriction=true
- There will be  part 2 of this Step that packs loose objects into a packfile.  That will come in a later iteration

**MaintenaceVerb**
-  This allows running a maintenance job (currently there is just 1 but this will be expanded)
-  Allows forcing of Step (ignoreTimeRestriction=true)
- Will also be used by functional tests to test running Steps

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -32,6 +32,7 @@ namespace GVFS.Common.Git
         private string gitBinPath;
         private string workingDirectoryRoot;
         private string dotGitRoot;
+        private string gitObjectsRoot;
         private string gvfsHooksRoot;
         private Process executingProcess;
         private bool stopping;
@@ -62,11 +63,11 @@ namespace GVFS.Common.Git
         }
 
         public GitProcess(Enlistment enlistment)
-            : this(enlistment.GitBinPath, enlistment.WorkingDirectoryRoot, enlistment.GVFSHooksRoot)
+            : this(enlistment.GitBinPath, enlistment.WorkingDirectoryRoot, enlistment.GVFSHooksRoot, enlistment.GitObjectsRoot)
         {
         }
 
-        public GitProcess(string gitBinPath, string workingDirectoryRoot, string gvfsHooksRoot)
+        public GitProcess(string gitBinPath, string workingDirectoryRoot, string gvfsHooksRoot, string gitObjectsRoot = null)
         {
             if (string.IsNullOrWhiteSpace(gitBinPath))
             {
@@ -76,6 +77,7 @@ namespace GVFS.Common.Git
             this.gitBinPath = gitBinPath;
             this.workingDirectoryRoot = workingDirectoryRoot;
             this.gvfsHooksRoot = gvfsHooksRoot;
+            this.gitObjectsRoot = gitObjectsRoot;
 
             if (this.workingDirectoryRoot != null)
             {
@@ -461,6 +463,11 @@ namespace GVFS.Common.Git
             return this.InvokeGitAgainstDotGitFolder("read-tree " + treeIsh);
         }
 
+        public Result PrunePacked()
+        {
+            return this.InvokeGitAgainstDotGitFolder("prune-packed -q");
+        }
+
         public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, bool redirectStandardError)
         {
             ProcessStartInfo processInfo = new ProcessStartInfo(this.gitBinPath);
@@ -504,6 +511,11 @@ namespace GVFS.Common.Git
                     ";",
                     this.gitBinPath,
                     this.gvfsHooksRoot ?? string.Empty);
+
+            if (this.gitObjectsRoot != null)
+            {
+                processInfo.EnvironmentVariables["GIT_OBJECT_DIRECTORY"] = this.gitObjectsRoot;
+            }
 
             if (!useReadObjectHook)
             {
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
@@ -54,6 +54,13 @@ namespace GVFS.Common.Maintenance
                     state: null,
                     dueTime: prefetchPeriod,
                     period: prefetchPeriod));
+
+                TimeSpan looseObjectsPeriod = TimeSpan.FromHours(12);
+                this.stepTimers.Add(new Timer(
+                    (state) => this.queue.TryEnqueue(new LooseObjectsStep(this.context, requireCacheLock: true)),
+                    state: null,
+                    dueTime: looseObjectsPeriod,
+                    period: looseObjectsPeriod));
             }
         }
     }
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -25,7 +25,7 @@ namespace GVFS.Common.Maintenance
         protected bool Stopping { get; private set; }
         protected bool RequireObjectCacheLock { get; }
 
-        public void Execute()
+        public bool Execute()
         {
             try
             {
@@ -39,15 +39,15 @@ namespace GVFS.Common.Maintenance
                         if (!cacheLock.TryAcquireLock())
                         {
                             this.Context.Tracer.RelatedInfo(this.Area + ": Skipping work since another process holds the lock");
-                            return;
+                            return false;
                         }
 
-                        this.CreateProcessAndRun();
+                        return this.CreateProcessAndRun();
                     }
                 }
                 else
                 {
-                    this.CreateProcessAndRun();
+                    return this.CreateProcessAndRun();
                 }
             }
             catch (IOException e)
@@ -65,6 +65,8 @@ namespace GVFS.Common.Maintenance
                     keywords: Keywords.Telemetry);
                 Environment.Exit((int)ReturnCode.GenericError);
             }
+
+            return true;
         }
 
         public void Stop()
@@ -100,7 +102,7 @@ namespace GVFS.Common.Maintenance
         /// (as specified by <see cref="RequireObjectCacheLock"/>), then this step is not run unless we
         /// hold the lock.
         /// </summary>
-        protected abstract void PerformMaintenance();
+        protected abstract bool PerformMaintenance();
 
         protected GitProcess.Result RunGitCommand(Func<GitProcess, GitProcess.Result> work)
         {
@@ -143,19 +145,19 @@ namespace GVFS.Common.Maintenance
             return metadata;
         }
 
-        private void CreateProcessAndRun()
+        private bool CreateProcessAndRun()
         {
             lock (this.gitProcessLock)
             {
                 if (this.Stopping)
                 {
-                    return;
+                    return false;
                 }
 
                 this.GitProcess = new GitProcess(this.Context.Enlistment);
             }
 
-            this.PerformMaintenance();
+            return this.PerformMaintenance();
         }
     }
 }
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -1,0 +1,123 @@
+﻿using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    public class LooseObjectsStep : GitMaintenanceStep
+    {
+        private const string LooseObjects = "loose-objects.history";
+        private readonly string looseObjectsHistoryPath = null;
+        private readonly bool ignoreTimeRestriction = false;
+
+        public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool ignoreTimeRestriction)
+            : this(context, requireCacheLock)
+        {
+            this.ignoreTimeRestriction = ignoreTimeRestriction;
+        }
+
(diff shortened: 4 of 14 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -62,11 +63,11 @@ static GitProcess()
         }
 
         public GitProcess(Enlistment enlistment)
-            : this(enlistment.GitBinPath, enlistment.WorkingDirectoryRoot, enlistment.GVFSHooksRoot)
+            : this(enlistment.GitBinPath, enlistment.WorkingDirectoryRoot, enlistment.GVFSHooksRoot, enlistment.GitObjectsRoot)
         {
         }
 
-        public GitProcess(string gitBinPath, string workingDirectoryRoot, string gvfsHooksRoot)
+        public GitProcess(string gitBinPath, string workingDirectoryRoot, string gvfsHooksRoot, string gitObjectsRoot = null)
```
jeschu1: I don't love having an optional param here.  Perhaps I should force the callers to send in null.  Let me know if there is a more preferred way to do this.

Since we use gitObjectRoot in the Environment vars, i think it should be passed in to the constructor (similar ot workingDirectoryRoot)

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -25,7 +25,7 @@ public GitMaintenanceStep(GVFSContext context, GitObjects gitObjects, bool requi
         protected bool Stopping { get; private set; }
         protected bool RequireObjectCacheLock { get; }
 
-        public void Execute()
+        public bool Execute()
```
jeschu1: This returns a bool, so the new maintenance job knows weather it was successful

GVFS/GVFS/GVFS.Windows.csproj
```diff
@@ -79,6 +79,7 @@
     </Compile>
     <Compile Include="CommandLine\CacheServerVerb.cs" />
     <Compile Include="CommandLine\CloneVerb.cs" />
+    <Compile Include="CommandLine\MaintenanceVerb.cs" />
```
jeschu1: This needs to be ordered

GVFS/GVFS/CommandLine/MaintenanceVerb.cs
```diff
@@ -0,0 +1,104 @@
...
+using GVFS.Common;
+using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Maintenance;
+using GVFS.Common.Tracing;
+
+namespace GVFS.CommandLine
+{
+    [Verb(MaintenanceVerbName, HelpText = "Runs a maintance job.")]
+    public class MaintenanceVerb : GVFSVerb.ForExistingEnlistment
```
wilbaker: What's the reason for adding a maintenance verb?  

`LooseObjectsStep` is scheduled to run automatically, when would (or how would) a user know that they should run the maintenance verb?

derrickstolee: The reason is for:

1. Functional tests. We'll run the step explicitly instead of relying on background timing.
2. Support. When we are supporting someone with a bad data shape, we can ask them to run the verb.

It's also possible that we will want to add notifications like "You just hydrated your repo, consider dehydrating and cleaning up loose objects by running..."

jeschu1: @wilbaker how would a user know to run dehydrate?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,123 @@
...
+
+                    // Verify no git processes are running
+                    IEnumerable<int> processIds = this.RunningGitProcessIds();
+                    if (processIds.Count() > 0)
+                    {                            
+                        activity.RelatedWarning("Skipping LooseObjectsStep due to git pids {0}", string.Join(",", processIds));
+                        return false;
+                    }
+
+                    // Perform prune-packed
```
wilbaker: To confirm, is the goal to prune loose objects that were created locally? (i.e. not those that were downloaded to the shared cache?)

jeschu1: It should use the shared cache (i.e.) D:\.gvfsCache\a34d3fab679e49d191e525a2f1ba1\gitObjects

GVFS/GVFS.Common/Git/GitProcess.cs
derrickstolee: hm. strange we can't do it during `PrunePacked()`. It certainly seems we could insert the variable there and then remove the variable after calling the invoke method. Maybe instead we should extend one of the `InvokeGitAgainstDotGitFolder()` methods?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: our goal is to not touch their `src/.git` folder (at this time). Instead, we will delete loose objects in their shared object cache _that already exist in packfiles in the shared object cache_.

GVFS/GVFS.Common/Git/GitProcess.cs
jeschu1: @derrickstolee that's basically what I was debating. :-)

Given we already have variables on the constructor that are used in the environment (and the value here doesn't change) having it here seems like a good option.  But I don't love it being an optional param, I could go and update the callers to pass in null.

Alternatively we could set it on PrunePacked, but that doesn't follow how the class handles env variables currently.  

@wilbaker  would love your input here too1

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: So one test that is planned is to run the step and verify that only known duplicate objects are deleted and other loose objects remain.

derrickstolee: It can be helpful to point to #488 for the 4-step plan for shared object cache maintenance.

GVFS/GVFS/CommandLine/MaintenanceVerb.cs
kewillford: Will there be any issues if a user discovers or we tell them about this verb and they decide to schedule it to run on their own at a low interval say every minute?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
wilbaker: Do we need to point `prune-packed` at the shared cache?  It looks like it's running against `src/.git`:

```
        public Result PrunePacked()
        {
            return this.InvokeGitAgainstDotGitFolder("prune-packed -q");
        }
```

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -505,6 +512,11 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
                     this.gitBinPath,
                     this.gvfsHooksRoot ?? string.Empty);
 
+            if (this.gitObjectsRoot != null)
+            {
+                processInfo.EnvironmentVariables["GIT_OBJECT_DIRECTORY"] = this.gitObjectsRoot;
```
kewillford: If this is only for the `PrunePacked` method can that method take the `gitObjectsRoot` variable and use it?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: @wilbaker this is a weird one because it uses an environment variable to find the scared cache.  
https://git-scm.com/docs/git-prune

It also seems inconsistent with the way the object dir is sent to git.  Maybe it does make sense for this to take in a parameter, and then pass that down to InvokeGitAgainstDogGitFolder to set in the environment.

GVFS/GVFS.Common/Git/GitProcess.cs
jeschu1: @kewillford yup

And i honestly don't love it.  

Based on the feedback so far i'm thinking maybe we should extend this function to allow passing in environment variables?  That way prune packed would take in the object dir and pass it down.

GVFS/GVFS/CommandLine/MaintenanceVerb.cs
jeschu1: @kewillford no more so if the user sets up 'git prune-packed' to run every minute ;-)

Their mount wouldn't fall over, but I wouldn't recommend doing that.  

We have telemetry on MaintenanceVerb so we could detect if someone does that.

Have users done similar things in the past or suggestions to steer away from that?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
wilbaker: Thanks for the details here.  Does `prune-packed` care about the .git folder at all?

I'm wondering if we should be calling `InvokeGitAgainstDotGitFolder` at all because it's a bit weird that we're passing in `--git-dir=` if git just ends up ignoring it.

Should we just call `InvokeGitImpl` directly in `PrunePacked()`?

GVFS/GVFS/CommandLine/MaintenanceVerb.cs
wilbaker: > Will there be any issues if a user discovers or we tell them about this verb and they decide to schedule it to run on their own at a low interval say every minute?

I think my concern is related to Kevin's comment.  Are the benefits of the new verb worth expanding out CLI's API surface and supporting it going forward?

For the specific scenarios mentioned by Stolee:

> Functional tests. We'll run the step explicitly instead of relying on background timing.

Could the functional tests just call `prune-packed` rather than relying on the verb?

> Support. When we are supporting someone with a bad data shape, we can ask them to run the verb.

Could we also just as the user to run `prune-packed` manually?  I think one difference between this support task and dehydrating is that there aren't manual steps for dehydrating as it requires calls to the ProjFS API. 

> @wilbaker how would a user know to run dehydrate?

When the dehydrate feature was added it was communicated to the organization using VFS4G via email and documented in their internal wiki.

You bring up a good point though, for new users to VFS4G dehydrate is not easily found. However, I'm still not sure we need to add a new verb if we can provide a git command that can be run manually to achieve the same result.  There is no manual alternative to `gvfs dehydrate`.

wilbaker: > Functional tests. We'll run the step explicitly instead of relying on background timing.

Sorry I think I misinterpreted this.  I was thinking we'd want to call it to clean stuff up, and you might have meant to test out the functionality itself.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: It's worth testing to see if `prune-packed` doesn't need a `.git` dir, but it may complain "I'm not in a git repo"

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
```diff
@@ -54,6 +54,13 @@ private void ScheduleRecurringSteps()
                     state: null,
                     dueTime: prefetchPeriod,
                     period: prefetchPeriod));
+
+                TimeSpan looseObjectsPeriod = TimeSpan.FromHours(12);
+                this.stepTimers.Add(new Timer(
+                    (state) => this.queue.TryEnqueue(new LooseObjectsStep(this.context, requireCacheLock: true)),
+                    state: null,
+                    dueTime: looseObjectsPeriod,
```
wilbaker: Should the `dueTime` be sooner than 12 hours?  

If someone has their machine configured to take Windows updates at 6 or 7pm every night, GVFS.Mount.exe might never hit the `dueTime` (assuming they long in after 7am in the morning).

derrickstolee: Yes, these steps with a long period should have a short dueTime.

jeschu1: @derrickstolee 6 hours?

jeschu1:
@kewillford @derrickstolee @wilbaker thanks for all the feedback!  I'll put up a V2 soon which nixes the MaintenceVerb and makes use of internal-only as well as other changes

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
derrickstolee: I was thinking on the order of 15 seconds.

jeschu1: That would be a lot of churn in the background.  Opening the files reading the date, etc...  we could go to 1 hour if we want to be agressive

derrickstolee: The `dueTime` is just for the first trigger. The `period` is used for the frequency.

derrickstolee: For these bigger maintenance steps, we should have _different_ values for `dueTime` and `period` (which is not necessary for background prefetch, as 15 minutes is good enough for both).

jeschu1: We can play with the values.  I remember you said there is a race condition here, so maybe 5 minutes for due time on this one and we can do the check hourly.  That should be pretty cheap and def allow the maintence jobs to run.

PR #597, opened 2018-12-07
stop blocking unsupported git commands

benpeart:
Now that git has learned to block unsupported commands on VFS4G repos, we can remove that logic from the pre-command hook.

The corresponding changes to git can be found in https://github.com/Microsoft/git/pull/85/commits/cfe88be4e34695a4b6c1f3e76d5b8fc4349ba031

The version of git specified includes two fixes:

1. [block unsupported git commands](https://github.com/Microsoft/git/pull/91)
2. [reset.quiet and post-indextchanged hook](https://github.com/Microsoft/git/pull/93)

Signed-off-by: Ben Peart <benpeart@microsoft.com>

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitBlockCommandsTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitBlockCommandsTests.cs
@@ -1,0 +1,106 @@
+﻿using GVFS.FunctionalTests.FileSystemRunners;
+using GVFS.FunctionalTests.Should;
+using GVFS.FunctionalTests.Tools;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixtureSource(typeof(FileSystemRunner), nameof(FileSystemRunner.Runners))]
+    [Category(Categories.GitCommands)]
+    public class GitBlockCommandsTests : TestsWithEnlistmentPerFixture
+    {
+        private FileSystemRunner fileSystem;
+        public GitBlockCommandsTests(FileSystemRunner fileSystem)
+        {
+            this.fileSystem = fileSystem;
+        }
+
+        [TestCase, Order(1)]
+        public void GitFsck()
+        {
+            ProcessResult result = GitHelpers.InvokeGitAgainstGVFSRepo(
+                this.Enlistment.RepoRoot,
+                "fsck");
+            result.ExitCode.ShouldNotEqual(0, result.Errors);
+        }
+
+        [TestCase, Order(2)]
+        public void GitGc()
+        {
+            ProcessResult result = GitHelpers.InvokeGitAgainstGVFSRepo(
+                this.Enlistment.RepoRoot,
+                "gc");
+            result.ExitCode.ShouldNotEqual(0, result.Errors);
+            result = GitHelpers.InvokeGitAgainstGVFSRepo(
+                this.Enlistment.RepoRoot,
+                "gc --auto");
+            result.ExitCode.ShouldEqual(0, result.Errors);
+        }
+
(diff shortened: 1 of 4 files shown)


jamill:
Could you also include a bit more detail on how someone could find out more about the corresponding Git changes? I think a link to the git commit with the changes (or at least the initial git commit with these changes) would help someone who wants to learn more about this.

Also, as Kevin mentioned, do we need to update the associated version of Git to one that includes these changes? or do we already have them? I don't think it would be much work to include functional tests here as well.

derrickstolee:
Please update GVFS.props to 2.20181213.2.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitBlockCommandsTests.cs
```diff
@@ -0,0 +1,106 @@
...
+    [Category(Categories.GitCommands)]
+    public class GitBlockCommandsTests : TestsWithEnlistmentPerFixture
+    {
+        private FileSystemRunner fileSystem;
+        public GitBlockCommandsTests(FileSystemRunner fileSystem)
+        {
+            this.fileSystem = fileSystem;
+        }
+
+        [TestCase, Order(1)]
```
kewillford: Does order really matter in these tests?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -1036,10 +1036,10 @@ public void RenameOnlyFileInFolder()
         public void UpdateIndexCannotModifySkipWorktreeBit()
```
kewillford: This seems like a duplicate test to the ones you added?  If so I would just delete it.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitBlockCommandsTests.cs
```diff
@@ -0,0 +1,106 @@
...
+        public void GitFsck()
+        {
+            ProcessResult result = GitHelpers.InvokeGitAgainstGVFSRepo(
+                this.Enlistment.RepoRoot,
+                "fsck");
+            result.ExitCode.ShouldNotEqual(0, result.Errors);
```
kewillford: I would maybe add a function that is:
```
private void CommandBlocked(string command)
{
	ProcessResult result = GitHelpers.InvokeGitAgainstGVFSRepo(
		this.Enlistment.RepoRoot,
		command);
	result.ExitCode.ShouldNotEqual(0, $"Command {command} not blocked when it should be.  Errors: {result.Errors}");
}
```
And have a similar method for `CommandNotBlocked`
Then you could have one test case that is:
```
CommandBlocked("fsck");
CommandBlocked("gc");
CommandNotBlocked("gc --auto");
CommandBlocked("prune");
etc...
```

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitBlockCommandsTests.cs
```diff
@@ -0,0 +1,106 @@
...
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixtureSource(typeof(FileSystemRunner), nameof(FileSystemRunner.Runners))]
```
wilbaker: There is no need to parameterize these tests on `FileSystemRunner`, as they never use `this.fileSystem`.

`[TestFixtureSource(typeof(FileSystemRunner), nameof(FileSystemRunner.Runners))]` can be replaced with `[TestFixture]` and the `FileSystemRunner fileSystem` parameter can be removed from the constructor.

wilbaker: @benpeart I see I was too late with this comment, I will create a PR with this change.

PR #600, opened 2018-12-10
Update Git to v2.20.0.vfs.1.1

derrickstolee:
Includes v2.20.0.windows.1.

It also disables `reset.quiet` (correctness issue) and `rebase.useBuiltin` (performance issue).

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181129.1</GitPackageVersion>
+    <GitPackageVersion>2.20181211.5</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
--- a/GVFS/GVFS.FunctionalTests/Tools/ControlGitRepo.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/ControlGitRepo.cs
@@ -53,6 +53,7 @@ namespace GVFS.FunctionalTests.Tools
             GitProcess.Invoke(this.RootPath, "config merge.renames false");
             GitProcess.Invoke(this.RootPath, "config advice.statusUoption false");
             GitProcess.Invoke(this.RootPath, "config core.abbrev 40");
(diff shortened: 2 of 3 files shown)


derrickstolee:
@kewillford The functional tests are breaking on things like 'reset' and 'status', and the biggest change in `gvfs-2.19.1` since the last time we had functional tests succeeding with a `2.20` release candidate is the post-index-changed hook. Could you take a look at these build failures?

kewillford:
@derrickstolee @benpeart  was there some change to the output of `reset --mixed` to not display the files that are updated/deleted.  Maybe we aren't setting a flag in the control repo that we are using in VFSForGit?  I will dig in and see what I can find.

derrickstolee:
@kewillford @benpeart @jrbriggs It appears that the "reset.quiet" setting doesn't work well in our setting. I tried setting it to true in our control repo, but that wasn't enough. Disabling it in our processes made the functional tests pass.

My guess is that the index-changed hook merged cleanly with the reset quiet code in 2.20.0, but that merge introduced a different behavior than the hook expects. For now, I'll clean up this PR to only update the Git version and remove the "reset.quiet" setting and we can work to fix the integration in a follow-up PR.

benpeart:
> I'm tempted to pull the change out of the v2.20.0 PR and work to reapply the change directly to that branch. We can then find the best way to halt auto-GC in a VFS for Git repo.

Yes, please pull this patch out of 2.20.0 - I'll go back and figure out best to handle the automatic 'git gc --auto' commands (which will fail silently later as we turn off auto GC in VFSForGIT repos).

My bad - I hadn't merged the PR to switch VFSForGIT to use the new version of git yet and didn't expect it to get pulled into 2.20.0 before I'd had a chance to run the functional tests.

benpeart:
> Do you know what is running the `gc --auto` or why it is running? We set [`{ "gc.auto", "0" },`](https://github.com/Microsoft/VFSForGit/blob/8c814fa190dca98bd58bcec0d2d79d2cea8742c3/GVFS/GVFS/CommandLine/GVFSVerb.cs#L125) in the config for the repo which I though should be disabling it.

Several git commands spawn "gc --auto" themselves (am, commit, fetch, merge, rebase, etc).  I've updated the patch in git to check for --auto and will only die if the config setting 'gc.auto' isn't "0" (ie turned off).

derrickstolee:
The latest push disables the builtin rebase. This is to check the performance results. It _may_ cause a failure in the functional tests due to different warning/error messages.

PR #601, opened 2018-12-10
Fix download cleanup issue caught by Functional test

alameenshah:
When pre-checks fail, service was not cleaning up the upgrade downloads directory. This issue was caught by FT in CI build nodes, which don't have ProjFS inboxed causing pre-check to fail.

Updated service to do the cleanup even when pre-check or download tasks fail.

### DIFF
--- a/GVFS/GVFS.Service/ProductUpgradeTimer.cs
+++ b/GVFS/GVFS.Service/ProductUpgradeTimer.cs
@@ -57,17 +57,24 @@ namespace GVFS.Service
             string errorMessage = null;
 
             InstallerPreRunChecker prerunChecker = new InstallerPreRunChecker(this.tracer, string.Empty);
-            if (prerunChecker.TryRunPreUpgradeChecks(out string _) && !this.TryDownloadUpgrade(out errorMessage))
+            ProductUpgrader productUpgrader = new ProductUpgrader(ProcessHelper.GetCurrentProcessVersion(), this.tracer);
+            if (prerunChecker.TryRunPreUpgradeChecks(out string _) && this.TryDownloadUpgrade(productUpgrader, out errorMessage))
+            {
+                return;
+            }
+
+            productUpgrader.CleanupDownloadDirectory();
+
+            if (errorMessage != null)
             {
                 this.tracer.RelatedError(errorMessage);
             }
         }
 
-        private bool TryDownloadUpgrade(out string errorMessage)
+        private bool TryDownloadUpgrade(ProductUpgrader productUpgrader, out string errorMessage)
         {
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -98,5 +105,11 @@ private bool TryDownloadUpgrade(out string errorMessage)
                 }
             }
         }
+
+        private void EmptyDownloadDirectory()
```
jamill: nit: I might suggest calling this `CleanupDownloadDirectory(...)`. `Empty` to me is a bit more ambiguous (is it used as a verb to "empty" a directory, ir is it checking if the directory is empty... ect...). 

As this is a private method and the calling code already has a product upgrade instance, I think we should re-use that instance.

alameenshah: renamed to CleanupDownloadDirectory. Removed the private helper method.

alameenshah:
>>The other thought I had is when do we want to clean up the download directory? Is it only if there are no newer updates? Do we do it in other places (after we have downloaded and installed an update)? What about with a failed download - do we want to clean up or is there value in keeping around the download artifacts for investigation?

Scenario 1
1. Install VFSForGit Version 1.0.18289.1.
2. Set the upgrade.ring to “Slow/Fast”.
3. Restart GVFS.Service. Now VFSForGit version 1.0.18297.1 upgrade release would get downloaded.
4. Manually download and install VFSForGit version 1.0.18297.1 from Github/Release.
5. Run git commands on an enlistment. The upgrade reminder message would show, even though you are running the latest version.
 
Scenario 2 (this is not reproducible today)
1. Install latest Slow release version of VFSForGit. Make sure there is a newer Fast release version that is available out there in Github/release.
2. After installing Slow release, set local upgrade.ring to “Fast”.
3. Restart GVFS.Service. This will download the latest Fast upgrade installer.
4. Switch back to “Slow” again. Now the PC will have the latest Fast release downloaded. But is up-to-date with respect to the Slow ring that it is configured to use.
 
A few options
1. GVFS.Service does its periodic upgrade check. When it finds that the current version is up-to-date, it can delete the previously downloaded releases. In between the user may see reminder messaging while running git commands.
2. User manually runs `gvfs upgrade`. When gvfs cli finds that the current version is up-to-date it deletes the un-needed release downloads. User would no longer see reminders.
3. A more complete solution would be for git post command hook to determine the current vfsforgit version and current upgrade ring. Then use that info to decide whether the downloaded release represents a valid upgrade before showing the reminder. This approach can negatively impact the Git hook perf a little.

PR #604, opened 2018-12-11
Pass ITracer to background process launching to log errors

### LINKED ISSUE #583: Mac: log failure to launch background process
E.g. by pass an ITracer instance to `GVFSPlatform.Instance.StartBackgroundProcess`

kewillford:
Resolves #583

### DIFF
--- a/GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
+++ b/GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using GVFS.Common.Tracing;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -12,7 +13,7 @@ namespace GVFS.Platform.Mac
         private const int StdOutFileNo = 1; // STDOUT_FILENO -> standard output file descriptor
         private const int StdErrFileNo = 2; // STDERR_FILENO -> standard error file descriptor
 
-        public static void StartBackgroundProcess(string programName, string[] args)
+        public static void StartBackgroundProcess(ITracer tracer, string programName, string[] args)
         {
             ProcessStartInfo startInfo = new ProcessStartInfo(programName);
             string[] envp = NetCoreMethods.CreateEnvp(startInfo);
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
```diff
@@ -89,6 +94,15 @@ private static string[] GenerateArgv(string fileName, string[] args)
             return argvList.ToArray();
         }
 
+        private static void LogErrorAndExit(ITracer tracer, string message)
+        {
+            int lastWin32Error = Marshal.GetLastWin32Error();
```
jamill: `Marshal.GetLastWin32Error()` is now quite a bit farther away from where the error is generated. We have to be careful to not run any action that might clear the last error before running this method. I think this would be harder to inadvertently break if the GetLastWin32Error() was closer to the native call... Maybe at least add a comment to call out this requirement in code to remind future people to keep this mind? (assuming my understanding of GetLastWin32Error is correct)

wilbaker: I've got the same concern, what if we made the native error code a parameter to `LogErrorAndExit` and continued to call `GetLastWin32Error` before entering this method?

GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
```diff
@@ -49,36 +50,40 @@ public static void StartBackgroundProcess(string programName, string[] args)
...
             string programName, 
             byte** argvPtr = null,
             byte** envpPtr = null)
         {
-            // TODO(Mac): Issue #583, log errors here to a log file
-
             int fdin = Open("/dev/null", (int)NetCoreMethods.OpenFlags.O_RDONLY);
+            if (fdin == -1)
+            {
+                LogErrorAndExit(tracer, "Unable to open file descriptor for stdin");
```
wilbaker: Have you tried writing to `tracer` in this method and checking the behavior?  

At this point we're in the forked child process and I'm not sure what will happen if you attempt to use the `tracer` from the parent process.

GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
```diff
@@ -37,7 +38,7 @@ public static void StartBackgroundProcess(string programName, string[] args)
 
                     if (processId == 0)
```
wilbaker: Should we also log above (when `processId == -1`) before throwing a `Win32Exception` (similar to what's being done in WindowsPlatform.cs)?

GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
kewillford: Works fine.  All handles are inherited in a child process so there doesn't seem to be any issues.

GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
kewillford: Makes sense.

GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
kewillford: Will do.

GVFS/GVFS.Platform.Mac/ProcessLauncher.cs
wilbaker: > Works fine. All handles are inherited in a child process so there doesn't seem to be any issues.

Nice 😃  I am surprised that managed references are inherited, but I confirmed it works on my machine as well 👍

PR #605, opened 2018-12-11
Loose Object Cleanup Step (Round 2)

jeschu1:
**LooseObjectsStep**
- We call prune-packed to remove loose objects that are already in pack-files
- This is scheduled to run every 6 hours, but only runs if it's been 7 days since last run or ignoreTimeRestriction=true
- There will be part 2 of this Step that packs loose objects into a packfile. That will come in a later iteration

**Testing**
- Unit tests have been added for coverage
- Dehydrate will call the step via the internal only flag.  
- Functional Tests are still WIP

### DIFF
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -44,7 +44,7 @@ namespace GVFS.Common.Git
 
         public static bool IsLooseObjectsDirectory(string value)
         {
-            return value.Length == 2 && !value.Any(c => !Uri.IsHexDigit(c));
+            return value.Length == 2 && value.All(c => Uri.IsHexDigit(c));
         }
 
         public virtual bool TryDownloadCommit(string commitSha)
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -42,6 +42,11 @@ namespace GVFS.Common.Git
             Error
         }
 
+        public static bool IsLooseObjectsDirectory(string value)
+        {
+            return value.Length == 2 && !value.Any(c => !Uri.IsHexDigit(c));
+        }
+
         public virtual bool TryDownloadCommit(string commitSha)
         {
             const bool PreferLooseObjects = false;
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -461,7 +461,16 @@ namespace GVFS.Common.Git
             return this.InvokeGitAgainstDotGitFolder("read-tree " + treeIsh);
         }
 
-        public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, bool redirectStandardError)
+        public Result PrunePacked(string gitObjectDirectory)
+        {
+            return this.InvokeGitAgainstDotGitFolder(
+                "prune-packed -q", 
+                writeStdIn: null,
+                parseStdOutLine: null,
+                gitObjectsDirectory: gitObjectDirectory);
+        }
+
+        public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, bool redirectStandardError, string gitObjectsDirectory)
         {
             ProcessStartInfo processInfo = new ProcessStartInfo(this.gitBinPath);
             processInfo.WorkingDirectory = workingDirectory;
@@ -505,6 +514,11 @@ namespace GVFS.Common.Git
                     this.gitBinPath,
                     this.gvfsHooksRoot ?? string.Empty);
 
+            if (gitObjectsDirectory != null)
+            {
+                processInfo.EnvironmentVariables["GIT_OBJECT_DIRECTORY"] = gitObjectsDirectory;
+            }
+
             if (!useReadObjectHook)
             {
                 command = "-c " + GitConfigSetting.CoreVirtualizeObjectsName + "=false " + command;
@@ -529,7 +543,8 @@ namespace GVFS.Common.Git
             bool useReadObjectHook,
             Action<StreamWriter> writeStdIn,
             Action<string> parseStdOutLine,
-            int timeoutMs)
+            int timeoutMs,
+            string gitObjectsDirectory = null)
         {
             if (failedToSetEncoding && writeStdIn != null)
             {
@@ -541,7 +556,7 @@ namespace GVFS.Common.Git
                 // From https://msdn.microsoft.com/en-us/library/system.diagnostics.process.standardoutput.aspx
                 // To avoid deadlocks, use asynchronous read operations on at least one of the streams.
                 // Do not perform a synchronous read to the end of both redirected streams.
-                using (this.executingProcess = this.GetGitProcess(command, workingDirectory, dotGitDirectory, useReadObjectHook, redirectStandardError: true))
+                using (this.executingProcess = this.GetGitProcess(command, workingDirectory, dotGitDirectory, useReadObjectHook, redirectStandardError: true, gitObjectsDirectory: gitObjectsDirectory))
                 {
                     StringBuilder output = new StringBuilder();
                     StringBuilder errors = new StringBuilder();
@@ -682,7 +697,8 @@ namespace GVFS.Common.Git
         private Result InvokeGitAgainstDotGitFolder(
             string command,
             Action<StreamWriter> writeStdIn,
-            Action<string> parseStdOutLine)
+            Action<string> parseStdOutLine,
+            string gitObjectsDirectory = null)
         {
             // This git command should not need/use the working directory of the repo.
             // Run git.exe in Environment.SystemDirectory to ensure the git.exe process
@@ -694,7 +710,8 @@ namespace GVFS.Common.Git
                 useReadObjectHook: false,
                 writeStdIn: writeStdIn,
                 parseStdOutLine: parseStdOutLine,
-                timeoutMs: -1);
+                timeoutMs: -1,
+                gitObjectsDirectory: gitObjectsDirectory);
         }
 
         public class Result
--- a/GVFS/GVFS.Common/InternalVerbParameters.cs
+++ b/GVFS/GVFS.Common/InternalVerbParameters.cs
@@ -4,14 +4,16 @@ namespace GVFS.Common
 {
     public class InternalVerbParameters
     {
-        public InternalVerbParameters(string serviceName, bool startedByService)
+        public InternalVerbParameters(string serviceName, bool startedByService, string maintenanceJob)
         {
             this.ServiceName = serviceName;
             this.StartedByService = startedByService;
+            this.MaintenanceJob = maintenanceJob;
         }
 
         public string ServiceName { get; private set; }
         public bool StartedByService { get; private set; }
+        public string MaintenanceJob { get; private set; }
 
         public static InternalVerbParameters FromJson(string json)
         {
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
@@ -7,6 +7,10 @@ namespace GVFS.Common.Maintenance
 {
     public class GitMaintenanceScheduler : IDisposable
     {
+        private readonly TimeSpan looseObjectsDueTime = TimeSpan.FromMinutes(5);
+        private readonly TimeSpan looseObjectsPeriod = TimeSpan.FromHours(6);
+        private readonly TimeSpan prefetchPeriod = TimeSpan.FromMinutes(15);
+
         private List<Timer> stepTimers;
         private GVFSContext context;
         private GitObjects gitObjects;
@@ -52,8 +56,14 @@ namespace GVFS.Common.Maintenance
                 this.stepTimers.Add(new Timer(
                     (state) => this.queue.TryEnqueue(new PrefetchStep(this.context, this.gitObjects, requireCacheLock: true)),
                     state: null,
-                    dueTime: prefetchPeriod,
-                    period: prefetchPeriod));
+                    dueTime: this.prefetchPeriod,
+                    period: this.prefetchPeriod));
+
+                this.stepTimers.Add(new Timer(
+                    (state) => this.queue.TryEnqueue(new LooseObjectsStep(this.context, requireCacheLock: true)),
+                    state: null,
+                    dueTime: this.looseObjectsDueTime,
+                    period: this.looseObjectsPeriod));
             }
         }
     }
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -1,7 +1,10 @@
 ﻿using GVFS.Common.Git;
 using GVFS.Common.Tracing;
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
 
 namespace GVFS.Common.Maintenance
 {
@@ -18,7 +21,8 @@ namespace GVFS.Common.Maintenance
         }
 
         public abstract string Area { get; }
-
+        protected virtual TimeSpan TimeBetweenRuns { get; }
+        protected virtual string LastRunTimeFilePath { get; set; }
         protected GVFSContext Context { get; }
         protected GitObjects GitObjects { get; }
         protected GitProcess GitProcess { get; private set; }
@@ -143,6 +147,45 @@ namespace GVFS.Common.Maintenance
             return metadata;
         }
 
+        protected bool EnoughTimeBetweenRuns()
+        {
+            if (!this.Context.FileSystem.FileExists(this.LastRunTimeFilePath))
+            {
+                return true;
+            }
+
+            string lastRunTime = this.Context.FileSystem.ReadAllText(this.LastRunTimeFilePath);
+            if (!long.TryParse(lastRunTime, out long result))
+            {
+                this.Context.Tracer.RelatedError("Failed to parse long: {0}", lastRunTime);
+                return true;
+            }
+
+            if (DateTime.UtcNow.Subtract(EpochConverter.FromUnixEpochSeconds(result)) >= this.TimeBetweenRuns)
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        protected void SaveLastRunTimeToFile()
+        {
+            if (!this.Context.FileSystem.TryWriteTempFileAndRename(
+                this.LastRunTimeFilePath,
+                EpochConverter.ToUnixEpochSeconds(DateTime.UtcNow).ToString(),
+                out Exception handledException))
+            {
+                this.Context.Tracer.RelatedError(this.CreateEventMetadata(handledException), "Failed to record run time");
+            }
+        }
+
+        protected IEnumerable<int> RunningGitProcessIds()
+        {
+            Process[] allProcesses = Process.GetProcesses();
+            return allProcesses.Where(x => x.ProcessName.Equals("git", StringComparison.OrdinalIgnoreCase)).Select(x => x.Id);
+        }
+
         private void CreateProcessAndRun()
         {
             lock (this.gitProcessLock)
@@ -152,7 +195,7 @@ namespace GVFS.Common.Maintenance
                     return;
                 }
 
-                this.GitProcess = new GitProcess(this.Context.Enlistment);
+                this.GitProcess = this.Context.Enlistment.CreateGitProcess();
             }
 
             this.PerformMaintenance();
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -60,7 +60,7 @@ namespace GVFS.Common.Maintenance
                         IEnumerable<int> processIds = this.RunningGitProcessIds();
                         if (processIds.Any())
                         {
-                            activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to git pids {0}", string.Join(",", processIds), Keywords.Telemetry);
+                            activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
                             return;
                         }
                     }
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -1,0 +1,87 @@
+﻿using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    // Remove loose objects that appear in packfiles
+    public class LooseObjectsStep : GitMaintenanceStep
+    {
+        public const string LooseObjectsLastRunFileName = "loose-objects.time";
+        private readonly bool forceRun;
+
+        public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRun = false)
+            : base(context, gitObjects: null, requireObjectCacheLock: requireCacheLock)
+        {
+            this.forceRun = forceRun;
+        }
+
+        public override string Area => nameof(LooseObjectsStep);
+
+        protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsLastRunFileName);
+        protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(7);
+
+        public int CountLooseObjects()
+        {
+            int count = 0;
+            
+            foreach (string directoryPath in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
+            {
+                string directoryName = directoryPath.TrimEnd(Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar).Last();
+
+                if (GitObjects.IsLooseObjectsDirectory(directoryName))
+                {                    
+                    count += this.Context.FileSystem.GetFiles(Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath), "*").Count();
+                }
+            }
+
+            return count;
+        }
+
+        protected override void PerformMaintenance()
+        {
+            using (ITracer activity = this.Context.Tracer.StartActivity(this.Area, EventLevel.Informational, Keywords.Telemetry, metadata: null))
+            {
+                try
+                {
+                    // forceRun is only currently true for functional tests
+                    if (!this.forceRun)
+                    {
+                        if (!this.EnoughTimeBetweenRuns())
+                        {
+                            activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to not enough time between runs");
+                            return;
+                        }
+
+                        IEnumerable<int> processIds = this.RunningGitProcessIds();
+                        if (processIds.Any())
+                        {
+                            activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to git pids {0}", string.Join(",", processIds), Keywords.Telemetry);
+                            return;
+                        }
+                    }
+
+                    int beforeCount = this.CountLooseObjects();
+                    GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot));
+                    int afterCount = this.CountLooseObjects();
+
+                    EventMetadata metadata = new EventMetadata();
+                    metadata.Add("StartingCount", beforeCount);
+                    metadata.Add("EndingCount", afterCount);
+                    metadata.Add("RemovedCount", beforeCount - afterCount);
+                    activity.RelatedEvent(EventLevel.Informational, this.Area, metadata, Keywords.Telemetry);
+
+                    this.SaveLastRunTimeToFile();
+                }
+                catch (Exception e)
+                {
+                    activity.RelatedWarning(this.CreateEventMetadata(e), "Failed to run LooseObjectsStep", Keywords.Telemetry);
+                }
+            }
+        }
+    }
+}
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -70,6 +70,7 @@ namespace GVFS.Common.Maintenance
                     int afterCount = this.CountLooseObjects();
 
                     EventMetadata metadata = new EventMetadata();
+                    metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
                     metadata.Add("StartingCount", beforeCount);
                     metadata.Add("EndingCount", afterCount);
                     metadata.Add("RemovedCount", beforeCount - afterCount);
--- a/GVFS/GVFS.Common/SHA1Util.cs
+++ b/GVFS/GVFS.Common/SHA1Util.cs
@@ -9,7 +9,7 @@ namespace GVFS.Common
     {
         public static bool IsValidShaFormat(string sha)
         {
-            return sha.Length == 40 && !sha.Any(c => !Uri.IsHexDigit(c));
+            return sha.Length == 40 && sha.All(c => Uri.IsHexDigit(c));
         }
 
         public static string SHA1HashStringForUTF8String(string s)
--- a/GVFS/GVFS.Common/SHA1Util.cs
+++ b/GVFS/GVFS.Common/SHA1Util.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 
@@ -8,7 +9,7 @@ namespace GVFS.Common
     {
         public static bool IsValidShaFormat(string sha)
         {
-            return sha.Length == 40 && !sha.Any(c => !(c >= '0' && c <= '9') && !(c >= 'a' && c <= 'f') && !(c >= 'A' && c <= 'F'));
+            return sha.Length == 40 && !sha.Any(c => !Uri.IsHexDigit(c));
         }
 
         public static string SHA1HashStringForUTF8String(string s)
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
@@ -1,0 +1,115 @@
+﻿using GVFS.FunctionalTests.FileSystemRunners;
+using GVFS.FunctionalTests.Tools;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixture]
+    public class LooseObjectStepTests : TestsWithEnlistmentPerFixture
+    {
+        private const string TempPackFolder = "tempPacks";
+        private FileSystemRunner fileSystem;
+
+        // Set forcePerRepoObjectCache to true to avoid any of the tests inadvertently corrupting
+        // the cache 
+        public LooseObjectStepTests()
+            : base(forcePerRepoObjectCache: true)
+        {
+            this.fileSystem = new SystemIORunner();
+        }
+
+        private string GitObjectRoot => this.Enlistment.GetObjectRoot(this.fileSystem);
+        private string PackRoot => this.Enlistment.GetPackRoot(this.fileSystem);
+        private string TempPackRoot=> Path.Combine(this.PackRoot, TempPackFolder);
+
+        [TestCase]
+        public void RemoveLooseObjectsInPackFiles()
+        {
+            // Delete any starting loose objects and verify
+            this.DeleteFiles(this.GetLooseObjectFiles());
+            Assert.AreEqual(0, this.GetLooseObjectFiles().Count);
+
+            // Move packfiles to temp
+            this.MovePackFilesToTemp();
+
+            // Expand 1 pack file and Copy it back to packs
+            this.ExpandOneTempPackAndMoveBack();
+
+            // Verify we have some LooseObjects
+            // These objects will also appear in the pack file that was moved back
+            Assert.AreNotEqual(0, this.GetLooseObjectFiles().Count);
+
+            // Run Cleanup
+            this.Enlistment.LooseObjectStep();
+
+            // Verify loose objects appearing in the pack file are removed
+            Assert.AreEqual(0, this.GetLooseObjectFiles().Count);
+        }
+
+        private List<string> GetLooseObjectFiles()
+        {
+            List<string> looseObjectFiles = new List<string>();
+            foreach (string directory in Directory.GetDirectories(this.GitObjectRoot))
+            {
+                // Check if the directory is 2 letter HEX
+                if (Regex.IsMatch(directory, @"[/\\][0-9a-fA-F]{2}$"))
+                {
+                    string[] files = Directory.GetFiles(directory);
+                    looseObjectFiles.AddRange(files);
+                }
+            }
+
+            return looseObjectFiles;
+        }
+
+        private void DeleteFiles(List<string> filePaths)
+        {
+            foreach (string filePath in filePaths)
+            {
+                File.Delete(filePath);
+            }
+        }
+
+        private void MovePackFilesToTemp()
+        {
+            string[] files = Directory.GetFiles(this.PackRoot);
+            foreach (string file in files)
+            {
+                string path2 = Path.Combine(this.TempPackRoot, Path.GetFileName(file));
+                File.Move(file, path2);
+            }
+        }
+
+        private void ExpandOneTempPackAndMoveBack()
+        {
+            // Find all pack files
+            string[] packFiles = Directory.GetFiles(this.TempPackRoot, "pack-*.pack");
+            Assert.Greater(packFiles.Length, 0);
+
+            // Pick the first one found
+            string packFile = packFiles[0];
+
+            // Send the contents of the packfile to unpack-objects to example the loose objects
+            // Note this won't work if the object exists in a pack file which is why we had to move them
+            using (FileStream packFileStream = File.OpenRead(packFile))
+            {
+                string output = GitProcess.InvokeProcess(
+                    this.Enlistment.RepoRoot, 
+                    "unpack-objects", 
+                    new Dictionary<string, string>() { { "GIT_OBJECT_DIRECTORY", this.GitObjectRoot } }, 
+                    inputStream: packFileStream).Output;
+            }
+
+            // Copy the pack file back to packs
+            string packFileName = Path.GetFileName(packFile);
+            File.Copy(packFile, Path.Combine(this.PackRoot, packFileName));
+
+            // Replace the '.pack' with '.idx' to copy the index file
+            string packFileIndexName = packFileName.Replace(".pack", ".idx");
+            File.Copy(Path.Combine(this.TempPackRoot, packFileIndexName), Path.Combine(this.PackRoot, packFileIndexName));
+        }
+    }
+}
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
@@ -133,9 +133,9 @@ namespace GVFS.FunctionalTests.Tools
             }
         }
 
-        public static string GetInternalParameter()
+        public static string GetInternalParameter(string maintenanceJob = "null")
         {
-            return $"\"{{\\\"ServiceName\\\":\\\"{GVFSServiceProcess.TestServiceName}\\\",\\\"StartedByService\\\":false}}\"";
+            return $"\"{{\\\"ServiceName\\\":\\\"{GVFSServiceProcess.TestServiceName}\\\",\\\"StartedByService\\\":false,\\\"MaintenanceJob\\\":{maintenanceJob}}}\"";
         }
 
         private static string GetModifiedPathsContents(GVFSFunctionalTestEnlistment enlistment, FileSystemRunner fileSystem)
--- a/GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
@@ -8,12 +8,17 @@ namespace GVFS.UnitTests.Common
     public class GitObjectsTests
     {
         [TestCase]
-        public void IsLooseObjectsDirectory()
+        public void IsLooseObjectsDirectory_ValidDirectories()
         {
             GitObjects.IsLooseObjectsDirectory("BB").ShouldBeTrue();
             GitObjects.IsLooseObjectsDirectory("bb").ShouldBeTrue();
             GitObjects.IsLooseObjectsDirectory("A7").ShouldBeTrue();
             GitObjects.IsLooseObjectsDirectory("55").ShouldBeTrue();
+        }
+
+        [TestCase]
+        public void IsLooseObjectsDirectory_InvalidDirectories()
+        {
             GitObjects.IsLooseObjectsDirectory("K7").ShouldBeFalse();
             GitObjects.IsLooseObjectsDirectory("A-").ShouldBeFalse();
             GitObjects.IsLooseObjectsDirectory("?B").ShouldBeFalse();
--- a/GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
@@ -10,15 +10,15 @@ namespace GVFS.UnitTests.Common
         [TestCase]
         public void IsLooseObjectsDirectory()
         {
-            GitObjects.IsLooseObjectsDirectory("BB").ShouldEqual(true);
-            GitObjects.IsLooseObjectsDirectory("bb").ShouldEqual(true);
-            GitObjects.IsLooseObjectsDirectory("A7").ShouldEqual(true);
-            GitObjects.IsLooseObjectsDirectory("55").ShouldEqual(true);
-            GitObjects.IsLooseObjectsDirectory("K7").ShouldEqual(false);
-            GitObjects.IsLooseObjectsDirectory("A-").ShouldEqual(false);
-            GitObjects.IsLooseObjectsDirectory("?B").ShouldEqual(false);
-            GitObjects.IsLooseObjectsDirectory("BBB").ShouldEqual(false);
-            GitObjects.IsLooseObjectsDirectory("B-B").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("BB").ShouldBeTrue();
+            GitObjects.IsLooseObjectsDirectory("bb").ShouldBeTrue();
+            GitObjects.IsLooseObjectsDirectory("A7").ShouldBeTrue();
+            GitObjects.IsLooseObjectsDirectory("55").ShouldBeTrue();
+            GitObjects.IsLooseObjectsDirectory("K7").ShouldBeFalse();
+            GitObjects.IsLooseObjectsDirectory("A-").ShouldBeFalse();
+            GitObjects.IsLooseObjectsDirectory("?B").ShouldBeFalse();
+            GitObjects.IsLooseObjectsDirectory("BBB").ShouldBeFalse();
+            GitObjects.IsLooseObjectsDirectory("B-B").ShouldBeFalse();
         }
     }
 }
--- a/GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
@@ -1,0 +1,24 @@
+﻿using GVFS.Common.Git;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class GitObjectsTests
+    {
+        [TestCase]
+        public void IsLooseObjectsDirectory()
+        {
+            GitObjects.IsLooseObjectsDirectory("BB").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("bb").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("A7").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("55").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("K7").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("A-").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("?B").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("BBB").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("B-B").ShouldEqual(false);
+        }
+    }
+}
--- a/GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
+++ b/GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
@@ -111,10 +111,10 @@ namespace GVFS.UnitTests.Maintenance
                 new List<MockFile>()
                 {
                      new MockFile(Path.Combine(enlistment.GitObjectsRoot, "F1", "test1"), string.Empty),
-                     new MockFile(Path.Combine(enlistment.GitObjectsRoot, "F2", "test2"), string.Empty)
+                     new MockFile(Path.Combine(enlistment.GitObjectsRoot, "F1", "test2"), string.Empty)
                 });
 
-            // Create NonHex Folder 2 with 2 Files
+            // Create NonHex Folder with 4 Files
             MockDirectory nonhex = new MockDirectory(
                 Path.Combine(enlistment.GitObjectsRoot, "ZZ"), 
                 null, 
--- a/GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
+++ b/GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
@@ -1,0 +1,141 @@
+﻿using GVFS.Common;
+using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Maintenance;
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Mock.Common;
+using GVFS.UnitTests.Mock.FileSystem;
+using GVFS.UnitTests.Mock.Git;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GVFS.UnitTests.Maintenance
+{
+    [TestFixture]
+    public class LooseObjectStepTests
+    {
+        private const string PrunePackedCommand = "prune-packed -q";
+        private MockTracer tracer;
+        private MockGitProcess gitProcess;
+        private GVFSContext context;
+
+        [TestCase]
+        public void LooseObjectsIgnoreTimeRestriction()
+        {
+            this.TestSetup(DateTime.UtcNow);
+
+            LooseObjectsStep step = new LooseObjectsStep(this.context, requireCacheLock: false, forceRun: true);
+            step.Execute();
+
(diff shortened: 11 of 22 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -461,7 +461,12 @@ public Result ReadTree(string treeIsh)
             return this.InvokeGitAgainstDotGitFolder("read-tree " + treeIsh);
         }
 
-        public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, bool redirectStandardError)
+        public Result PrunePacked(string gitObjectDirectory)
+        {
+            return this.InvokeGitAgainstDotGitFolder("prune-packed -q", gitObjectDirectory);
+        }
+
+        public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, bool redirectStandardError, string gitObjectsDirectory)
```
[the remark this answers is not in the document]
derrickstolee: Perhaps we are getting so large that we need a DTO, say `GitProcessContext`, that contains all of these values. A helper function to generate a default from an enlistment would make calling these methods easier.

Of course, we have a lot of methods in the codebase that take several parameters and could use a similar pattern. So you are not outside of normal to have so many parameters here.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -674,15 +685,16 @@ private Result InvokeGitOutsideEnlistment(string command)
         /// Invokes git.exe against an enlistment's .git folder.
         /// This method should be used only with git-commands that ignore the working directory
         /// </summary>
-        private Result InvokeGitAgainstDotGitFolder(string command)
+        private Result InvokeGitAgainstDotGitFolder(string command, string gitObjectsDirectory = null)
```
derrickstolee: It looks like you aren't using `gitObjectsDirectory` here?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    public class LooseObjectsStep : GitMaintenanceStep
+    {
+        public const string LooseObjects = "loose-objects.history";
```
derrickstolee: `LooseObjects` is too vague. Maybe `LooseObjectsFileName`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            if (!this.Context.FileSystem.FileExists(this.looseObjectsHistoryPath))
+            {
+                return true;
+            }
+
+            string lastRunTime = this.Context.FileSystem.ReadAllText(this.looseObjectsHistoryPath);
+            DateTime dateTime = Convert.ToDateTime(lastRunTime);
+            double daysDiff = (DateTime.Now - dateTime).TotalDays;
+            if (daysDiff >= 7)
```
derrickstolee: This seems like it would be better to use a `readonly TimeSpan` up with the constants, and then compare the values:

```
private readonly TimeSpan timeBetweenRuns = TimeSpan.FromDays(7);
```
...
```
            string lastRunTime = this.Context.FileSystem.ReadAllText(this.looseObjectsHistoryPath);
            DateTime lastRunDate = Convert.ToDateTime(lastRunTime);
            TimeSpan timeSince = DateTime.Now - lastRunDate;
            if (timeSince.CompareTo(this.TimeBetweenRuns) > 0)
            {
                return true;
            }
```

Also, this logic is not specific to this step, but will be something we want to replicate in other steps. Specifically:

1. We should remove the dependence in the `PrefetchStep` to use the filenames for timestamps and instead use this file mechanism. (This may require changing the self-healing logic present currently, so is a bigger job than it sounds.) The prefetch step will want to actually acquire this timestamp as a Unix Epoch value.
2. The repack step will need _exactly_ this logic, but with a different filename.

Related to all of this: we have `EpochConverter` to assist with using Unix-style epoch values instead of `DateTime` conversions (which may have localization issues?).

And finally, my super nit-pick: "history" doesn't seem like the right word, as we don't keep a log of multiple runs. Perhaps `.time` as the extension?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+        private readonly bool ignoreTimeRestriction = false;
+
+        public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool ignoreTimeRestriction)
+            : this(context, requireCacheLock)
+        {
+            this.ignoreTimeRestriction = ignoreTimeRestriction;
+        }
+
+        public LooseObjectsStep(GVFSContext context, bool requireCacheLock)
+            : base(context, gitObjects: null, requireObjectCacheLock: requireCacheLock)
```
derrickstolee: `null` gitObjects? That seems like it will break.

Also: why two constructiors? Shouldn't we just make `ignoreTimeRestriction` default to `false`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    public class LooseObjectsStep : GitMaintenanceStep
+    {
+        public const string LooseObjects = "loose-objects.history";
+        private readonly string looseObjectsHistoryPath = null;
+        private readonly bool ignoreTimeRestriction = false;
```
derrickstolee: nit: don't set these `readonly` values here, so the compiler will check that they are initialized in the constructor.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+        public override string Area => "LooseObjectsStep";
+
+        public int CountLooseObjects()
+        {
+            int count = 0;
+            int pathLength = this.Context.Enlistment.GitObjectsRoot.Length;
+            
+            foreach (string directory in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
+            {
+                // Check if the directory is 2 letter HEX
```
derrickstolee: Instead of a regex, could we just iterate through the 256 options and check existence before enumerating?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+                try
+                {
+                    if (!this.ignoreTimeRestriction && !this.EnoughTimeBetweenRuns())
+                    {
+                        activity.RelatedWarning("Skipping step due to not enough time between runs");
+                        return;
+                    }
+
+                    // Verify no git processes are running
+                    IEnumerable<int> processIds = this.RunningGitProcessIds();
```
derrickstolee: Should this method return a `bool` instead?

Also, better to use `.Any()` instead of `.Count() > 0` on an `IEnumerable`.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+                    int beforeCount = this.CountLooseObjects();
+                    GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot));
+                    int afterCount = this.CountLooseObjects();
+
+                    // Record Telemetry
+                    EventMetadata metadata = new EventMetadata();
+                    metadata.Add("StartingCount", beforeCount);
+                    metadata.Add("EndingCount", afterCount);
+                    metadata.Add("RemovedCount", afterCount - beforeCount);
+                    activity.RelatedEvent(EventLevel.Informational, "LooseObjects", metadata, Keywords.Telemetry);
```
derrickstolee: Use `this.Area` instead of `"LooseObjects"`.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+            double daysDiff = (DateTime.Now - dateTime).TotalDays;
+            if (daysDiff >= 7)
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        private IEnumerable<int> RunningGitProcessIds()
```
derrickstolee: Move to subclass so we can use this in another step?

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            this.TestSetup(DateTime.Now.AddDays(-7).ToString(), out MockTracer tracer, out MockGitProcess gitProcess, out GVFSContext context);
+
+            LooseObjectsStep step = new LooseObjectsStep(context, requireCacheLock: false, ignoreTimeRestriction: false);
+            int count = step.CountLooseObjects();
+
+            count.ShouldEqual(3);
+        }
+
+        private void TestSetup(string lastRunTime, out MockTracer tracer, out MockGitProcess gitProcess, out GVFSContext context)
```
derrickstolee: Instead of a lot of out-params, could we set private member variables? See `GitMaintenanceStepTests` for an example.

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,135 @@
...
+            this.TestSetup(DateTime.Now.ToString(), out MockTracer tracer, out MockGitProcess gitProcess, out GVFSContext context);
+
+            LooseObjectsStep step = new LooseObjectsStep(context, requireCacheLock: false, ignoreTimeRestriction: true);
+            step.Execute();
+
+            tracer.StartActivityTracer.RelatedErrorEvents.Count.ShouldEqual(0);
+            tracer.StartActivityTracer.RelatedWarningEvents.Count.ShouldEqual(0);
+            List<string> commands = gitProcess.CommandsRun;
+            commands.Count.ShouldEqual(1);
+            commands[0].ShouldEqual("prune-packed -q");
```
derrickstolee: It's great that we are checking the command we run. Is there any way to check the environment variables? Or, should we leave that to the functional tests?

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
```diff
@@ -26,8 +28,18 @@ public void SetExpectedCommandResult(string command, Func<Result> result, bool m
...
+            string command,
+            string workingDirectory,
+            string dotGitDirectory,
+            bool useReadObjectHook,
+            Action<StreamWriter> writeStdIn,
+            Action<string> parseStdOutLine,
+            int timeoutMs,
+            string gitObjectsDirectory = null)
         {
+            this.CommandsRun.Add(command);
```
derrickstolee: Ah, here is where we _could_ add information about the whole parameter set. Probably overkill, so ignore if you want.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -58,6 +59,19 @@ protected override void Execute(GVFSEnlistment enlistment)
                         { nameof(this.EnlistmentRootPathParameter), this.EnlistmentRootPathParameter },
                     });
 
+                if (this.MaintenanceJob != null
+                    && this.MaintenanceJob.Equals("LooseObjects"))
+                {
+                    this.WriteMessage(tracer, "Running LooseObject Step");
```
derrickstolee: We have some inconsistencies with the "LooseObject[s]" string. Perhaps it should be a fixed constant?

Alternatively, we could rewrite this portion of the method as:

```
if (this.MaintenanceJob != null) {
    this.InitializeLocalCacheAndObjectsPaths(tracer, enlistment, retryConfig: null, serverGVFSConfig: null, cacheServer: null);
    PhysicalFileSystem fileSystem = new PhysicalFileSystem();
    GitRepo gitRepo = new GitRepo(
                            tracer,
                            enlistment,
                            fileSystem);
    GVFSContext context = new GVFSContext(tracer, fileSystem, gitRepo, enlistment);
    GitMaintenanceStep[] steps = new [] {
        new LooseObjectsStep(context, requireCacheLock: true);
    };
    for (int i = 0; i < steps.Length; i++) {
        if (steps[i].Area.Equals(this.MaintenanceJob)) {
            this.WriteMessage(tracer, $"Running {this.MaintenanceJob} step");
            steps[i].Execute();
        }
    }
}
```
It makes the future extension of adding the packfile step very simple.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            Process[] allProcesses = Process.GetProcesses();
+            return allProcesses.Where(x => x.ProcessName == "git").Select(x => x.Id);
+        }
+
+        private void RecordLooseObjectHistory()
+        {
+            this.Context.FileSystem.WriteAllText(this.looseObjectsHistoryPath, DateTime.Now.ToString());
```
wilbaker: This could result in a torn write if there's a power-loss or BSOD.  

Can we use the approach of writing to a temporary file and then renaming it to `this.looseObjectsHistoryPath`? (see `FileSystem.MoveAndOverwriteFile`)

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+                    this.RecordLooseObjectHistory();
+                }
+                catch (Exception e)
+                {
+                    activity.RelatedError("Failed to run LooseObjectsStep: {0}", e.ToString());
+                }
+            }
+        }
+
+        // Job should be run 7 days apart
```
wilbaker: Rather than a comment use a named constant for 7 (e.g. `JobIntervalDays`)

GVFS/GVFS.Common/Git/GitProcess.cs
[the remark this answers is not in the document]
wilbaker: > Passing the extra variable seems pretty heavy, so I'd be happy to go back to the previous pattern if folks prefer.

I find this approach to be a easier to follow than the previous approach.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -58,6 +59,19 @@ protected override void Execute(GVFSEnlistment enlistment)
                         { nameof(this.EnlistmentRootPathParameter), this.EnlistmentRootPathParameter },
                     });
 
+                if (this.MaintenanceJob != null
```
wilbaker: Maybe add a comment here that this code is only intended to be run by the functional tests?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+            this.ignoreTimeRestriction = ignoreTimeRestriction;
+        }
+
+        public LooseObjectsStep(GVFSContext context, bool requireCacheLock)
+            : base(context, gitObjects: null, requireObjectCacheLock: requireCacheLock)
+        {
+            this.looseObjectsHistoryPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjects);         
+        }
+
+        public override string Area => "LooseObjectsStep";
```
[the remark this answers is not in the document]
derrickstolee: StyleCop does complain. It thinks you are talking about the constructor and insists `nameof(this.LooseObjectsStep)`. Not sure if that is desirable, and the other steps use a string literal.

GVFS/GVFS.Common/Git/GitProcess.cs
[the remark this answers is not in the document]
jeschu1: Thanks for the feedback @derrickstolee  and @wilbaker I'll leave it as is!

GVFS/GVFS.Common/Git/GitProcess.cs
jeschu1: Yup I found that out also through a new functional test :-).  Yay functional tests!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: We don't use the gitObjects so I set them to null.  If we have a need for them in the future i can take them in.  Right now I'd rather make it clear this isn't being used.

I'm updating to one constructor.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: Thanks all!  I like @wilbaker idea of making a function in SHA1Util that can also be used DiagnoseVerb to validate the directory name and I've updated accordingly. 

Long term I think a shared LogLooseObjectCount function makes a ton of sense.  It would require some retweaking in DiagnoseVerb, so I'm saving that for a later iteration.  I think it's definitely worth doing.

I'll let you know when the changes are pushed up.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: Moved, I wasn't sure if future steps would need this.

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
jeschu1: @derrickstolee will the test be run concurrently?  If so private variables aren't a good idea.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: Wow. That's totally my fault. Only the `PrefetchStep` needs the `GitObjects` member, and that is causing a lot of unneeded code passing one to the constructor of every `GitMaintenanceStep`. I'm sure there is a historical reason why that was done in the base class, but it isn't needed anymore.

While you are here, could you move the `GitObjects` out of `GitMaintenanceStep` and only into `PrefetchStep`? This will change constructors in `PostFetchStep` and the unit tests. Sorry for the pain here.

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
jeschu1: Unfortunately the env variables are harder to test.  The good news is that the functional tests definitely cover this case.  I found out the hard way :-).

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
jeschu1: @derrickstolee not sure what you mean, can you elaborate?

derrickstolee: I mean you could add every input to this function as data for checking in the unit test (it could check the `gitObjectsDirectory` parameter, for instance). Again, overkill.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
jeschu1: We definitely want the flexibility but I find the above hard to read.  I'm updating with an alternative using switch statements.

It means the MaintenanceJob my not equal the Area for the step.  I really think that's the way to go given the tradeoffs.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,120 @@
...
+                return true;
+            }
+
+            return false;
+        }
+
+        private IEnumerable<int> RunningGitProcessIds()
+        {
+            Process[] allProcesses = Process.GetProcesses();
+            return allProcesses.Where(x => x.ProcessName == "git").Select(x => x.Id);
```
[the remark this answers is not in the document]
jeschu1: Good question!  

Since it's the name 'git' works.  InstallerPreRunChecker also has similar logic. 

Is git-checkout.exe a child process?  The installer also doesn't check for it.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: @derrickstolee 

I've updated the timestamp reading and writing to a file to be in MaintenanceStep.  This gives steps the opportunity to use it.  I wouldn't however move PrefetchStep over to the new infrastructure in this design since it's a bit involved and this PR is already quite large.  I'd rather create an issue and seperate PR for it.  Sound fair?

I'm not sure why we'd need to use EpochConverter.  Since it's reading/writing from the same machine shouldn't any localization be accounted for?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: :-)  I'm happy to do the refactor here but I'd prefer to do it in a separate PR.  This one is already so big and I don't want to overload my very kind reviewers (yourself included!).

I've added an issue to track #617

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
jeschu1: Ah I see!  This is also where a GitContext would be nice as we could just have a list of them and compare.  I think we've definitely got better coverage now than we've had in the past, so as you suggest we'll punt that for now.  Good idea though!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
jeschu1: It works now!!!! I put it in a separate commit that i'll squash so everyone can see it.

Thanks @jamill !

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -18,7 +21,8 @@ public GitMaintenanceStep(GVFSContext context, GitObjects gitObjects, bool requi
         }
 
         public abstract string Area { get; }
-
+        protected virtual TimeSpan TimeBetweenRuns { get; }
+        protected virtual string TimeFilePath { get; set; }
```
derrickstolee: I'm guessing that you left this as `virtual` so the superclass can change the `get` method. Why is `set` still here? I have two guesses (and will see it when I get down to its use later):

1. You don't need it, and can delete it.
2. You need it, but you could change the protection to `{ get; protected set; }`.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public class LooseObjectsStep : GitMaintenanceStep
+    {
+        public const string LooseObjectsFileName = "loose-objects.time";
+        private readonly bool forceRun;
+
+        public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRun = false)
+            : base(context, gitObjects: null, requireObjectCacheLock: requireCacheLock)
+        {
+            this.TimeFilePath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsFileName);
```
derrickstolee: Since you are using the `virtual` pattern instead of the `readonly` pattern, you could put this down after `TimeBetweenRuns` as

```
protected override string TimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsFileName);
```

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
derrickstolee: (I now see it is already `protected` so the second option is silly.)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,139 @@
+﻿using GVFS.FunctionalTests.FileSystemRunners;
+using GVFS.FunctionalTests.Tools;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    // TODO(Mac): Before these tests can be enabled PostFetchJobShouldComplete needs
```
derrickstolee: Is this comment (and the `MacTODO.M4` category) needed?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        // Set forcePerRepoObjectCache to true to avoid any of the tests inadvertently corrupting
+        // the cache 
+        public LooseObjectStepTests()
+            : base(forcePerRepoObjectCache: true)
+        {
+            this.fileSystem = new SystemIORunner();
+        }
+
+        private string GitObjectRoot
```
derrickstolee: these `get`s could be done with the `=>` syntax for brevity.

GVFS/GVFS.UnitTests/Common/SHA1UtilTests.cs
```diff
@@ -74,5 +74,16 @@ public void IsValidFullSHATrueForLength40HexStrings()
             SHA1Util.IsValidShaFormat("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa").ShouldEqual(true);
             SHA1Util.IsValidShaFormat("ffffffffffffffffffffffffffffffffffffffff").ShouldEqual(true);
         }
+
+        [TestCase]
+        public void IsValidHex()
+        {
+            SHA1Util.IsValidHex("1234567890123456789012345678901234567890", 40).ShouldEqual(true);
+            SHA1Util.IsValidHex("abcdef7890123456789012345678901234567890", 20).ShouldEqual(false);
+            SHA1Util.IsValidHex("ab", 2).ShouldEqual(true);
```
derrickstolee: What happens if our string has a different length than the second parameter? This currently doesn't make it clear what should happen.

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
derrickstolee: I don't think rests in the same class are run concurrently, but that may be something about MSTest and not NUnit.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: Times are hard. Localization is hard. I never claim to know handling them will work.

For instance: what happens if someone updates their timezone? How will that interact with your DateTime serializations? No, it's not the end of the world if that happens, but an epoch is simple and related to our existing timestamp patterns.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,89 @@
...
+                    {
+                        if (!this.EnoughTimeBetweenRuns())
+                        {
+                            activity.RelatedWarning("Skipping step due to not enough time between runs");
+                            return;
+                        }
+
+                        // Verify no git processes are running
+                        IEnumerable<int> processIds = this.RunningGitProcessIds();
+                        if (processIds.Any())
```
kewillford: Would we still want to check for git process even in a forced run?  What could happen if there is a git process running that would affect the prune-packed?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        private string TempPackRoot
+        {
+            get
+            {
+                return Path.Combine(this.PackRoot, TempPackFolder);
+            }
+        }
+
+        [TestCase, Order(1)]
```
kewillford: Don't need `Order` do we?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            // Verify we have some LooseObjects
+            // These objects will also appear in the pack file that was moved back
+            Assert.AreNotEqual(0, this.GetLooseObjectFiles().Count);
+
+            // Run Cleanup
+            this.Enlistment.LooseObjectStep();
+
+            // Verify loose objects appearing in the pack file are removed
+            Assert.AreEqual(0, this.GetLooseObjectFiles().Count);
```
kewillford: Can we run an add or commit and create some objects that would be left so we can verify that some objects don't get removed or would that be a separate test?

GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
```diff
@@ -133,9 +133,9 @@ public static void ModifiedPathsShouldNotContain(GVFSFunctionalTestEnlistment en
             }
         }
 
-        public static string GetInternalParameter()
+        public static string GetInternalParameter(string maintenanceJob = "null")
```
kewillford: Could you explain the `"null"` string to me?  Will the object be `null`?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
kewillford: I was thinking about this and if the `MaintenanceJob` parameter should be something like a comma separated list of the class names for the steps to run so that we can test an individual step?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: @derrickstolee good questions!

These should really be abstract and extended by each class same as Area.  But I don't want to force each class to implement it until we do the refactoring in #617.  I left a comment on the issue ticket to address these.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
jeschu1: Good point, these comments are copy/paste issues and I've updated so we should run on Mac

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
jeschu1: Can you give the full example.  Sorry a bit confused here.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -143,6 +147,37 @@ protected EventMetadata CreateEventMetadata(Exception e = null)
...
+
+        protected void RecordRunTime()
+        {
+            if (!this.Context.FileSystem.TryWriteTempFileAndRename(this.TimeFilePath, DateTime.Now.ToString(), out Exception handledException))
+            {
+                this.Context.Tracer.RelatedError("Failed record run time: {0}", handledException?.ToString());
+            }
+        }
+
+        protected IEnumerable<int> RunningGitProcessIds()
```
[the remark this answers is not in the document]
jeschu1: We print the results for debugging purposes:
                            activity.RelatedWarning("Skipping LooseObjectsStep due to git pids {0}", string.Join(",", processIds));

We could add debugging on this method for the processIds and then another warning line in the LooseObjectStep (to confirm that's why we're exiting).

I perfer it as is honestly.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,89 @@
...
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    public class LooseObjectsStep : GitMaintenanceStep
```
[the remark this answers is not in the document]
jeschu1: Hrmm maybe some comments would be most helpful here.  I've added some. 

The class does all maintenance associated with LooseObjects so I think the name is appropriate.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: @derrickstolee and I talked about this offline and the forceRun is only set for functional tests.  We want to force certain scenarios here so yes we'd like it to run even if a git command is going.  We will add more testing around particular git commands.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
jeschu1: Seperate test.  And there are definitely more functional tests coming after this PR :-).  Trying to keep the scope narrow here as the PR is already huge.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -505,6 +510,11 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
                     this.gitBinPath,
                     this.gvfsHooksRoot ?? string.Empty);
 
+            if (gitObjectsDirectory != null)
+            {
+                processInfo.EnvironmentVariables["GIT_OBJECT_DIRECTORY"] = gitObjectsDirectory;
```
jamill: Maybe more of a question for @derrickstolee - how is it decided whether parameters are specified via an environment variable vs command line argument?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: @derrickstolee I think we can fix the timezone issue and keep the data readable by using Utc and a date format.  Let know what you think about the latest changes.

GVFS/GVFS.Common/Git/GitProcess.cs
jamill: If we are going to start adding optional parameters, can we cut out this middle function and just add optional parameters to the called function?

GVFS/GVFS.Common/InternalVerbParameters.cs
```diff
@@ -4,14 +4,16 @@ namespace GVFS.Common
 {
     public class InternalVerbParameters
     {
-        public InternalVerbParameters(string serviceName, bool startedByService)
+        public InternalVerbParameters(string serviceName, bool startedByService, string maintenanceJob)
```
jamill: It was unclear to me from just reading the code what a `maintenanceJob` string is....

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -143,6 +147,37 @@ protected EventMetadata CreateEventMetadata(Exception e = null)
...
         }
 
+        protected bool EnoughTimeBetweenRuns()
+        {
+            if (!this.Context.FileSystem.FileExists(this.TimeFilePath))
+            {
+                return true;
+            }
+
+            string lastRunTime = this.Context.FileSystem.ReadAllText(this.TimeFilePath);
```
jamill: How are IO errors handled?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -143,6 +147,37 @@ protected EventMetadata CreateEventMetadata(Exception e = null)
...
+            if (!this.Context.FileSystem.TryWriteTempFileAndRename(this.TimeFilePath, DateTime.Now.ToString(), out Exception handledException))
+            {
+                this.Context.Tracer.RelatedError("Failed record run time: {0}", handledException?.ToString());
+            }
+        }
+
+        protected IEnumerable<int> RunningGitProcessIds()
+        {
+            Process[] allProcesses = Process.GetProcesses();
+            return allProcesses.Where(x => x.ProcessName == "git").Select(x => x.Id);
```
jamill: Can we explicitly include the string comparison type? `OrdinalIgnoreCase`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
jamill: `LooseObjectsCleanupStep`?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: I just pushed some changes that address this.  Great point!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
derrickstolee:
```
private string GitObjectRoot => this.Enlistment.GetObjectRoot(this.fileSystem);
```

GVFS/GVFS.Common/Git/GitProcess.cs
jeschu1: I thought this exact same, but then looking at all the options we default decided against it:

                command,
                workingDirectory: Environment.SystemDirectory,
                dotGitDirectory: this.dotGitRoot,
                useReadObjectHook: false,
                writeStdIn: writeStdIn,
                parseStdOutLine: parseStdOutLine,
                timeoutMs: -1,
                gitObjectsDirectory: gitObjectsDirectory);

That's a lot to call directly and hope defaults don't change.  Maybe it's fine?

GVFS/GVFS.Common/Git/GitProcess.cs
derrickstolee: It seems to be a bit random, but also depends on whether it is a porcelain or plumbing command.

* [`git prune-packed`](https://git-scm.com/docs/git-prune-packed) doesn't have a command-line option.
* [`git unpack-objects`](https://git-scm.com/docs/git-unpack-objects) doesn't have a command-line option.
* [`git commit-graph`](https://git-scm.com/docs/git-commit-graph) has `--object-dir`.
* `git multi-pack-index` has `--object-dir`.

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
```diff
@@ -12,6 +12,9 @@ public class GitMaintenanceScheduler : IDisposable
         private GitObjects gitObjects;
         private GitMaintenanceQueue queue;
 
+        private TimeSpan looseObjectsDueTime = TimeSpan.FromMinutes(5);
```
derrickstolee: nit: `readonly`.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -143,6 +149,47 @@ protected EventMetadata CreateEventMetadata(Exception e = null)
...
 
+        protected bool EnoughTimeBetweenRuns()
+        {
+            if (!this.Context.FileSystem.FileExists(this.LastRunTimeFilePath))
+            {
+                return true;
+            }
+
+            string lastRunTime = this.Context.FileSystem.ReadAllText(this.LastRunTimeFilePath);
+            if (!DateTime.TryParseExact(
```
derrickstolee: This would be so much simpler if you used `EpochConverter` to store a unix-style timestamp.

jeschu1: I updated it to use EpochConverter is a seperate commit.

I honestly still prefer the formatter way, but this will work to.  Just add the unit tests as promised ;-)

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
jeschu1: I added the comments, there will be more comments as we expand what the Step does.

Part2 of this will take loose objects and put them in packfiles.

Let me know if anything seems confusing!

GVFS/GVFS.Common/InternalVerbParameters.cs
jeschu1: I think once you know what a mintenance job is it makes sense.  Let me know if comments would be helpful or what other suggestions you have for clarity.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
kewillford: I guess my question would be is this same class going to be doing the packing of loose objects as well or will that be a separate maintenance step?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
derrickstolee: #620

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
derrickstolee: It needs to be the same step, because if they are scheduled independently we could have loose objects be packed (the next part of this step) and then deleted soon after (this part of the step). By doing both (i) deleting and _then_ (ii) packing, we can guarantee that we are not interrupting Git commands that don't see the new pack-file and try to read from a loose object on disk.

A similar ordering is required for the packfile cleanup. See #488 for more discussion on this.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: I should add context here as to why I didn't love using the EpochConverter
- I like having readable datetimes in files.  This easily lets me verify when the last run was.  By saving a long we have to use a converter.
- EpochConverter has two functions ToUnixEpochSeconds/FromUnixEpochSeconds, however only one of them is in use (FromUnixEpochSeconds in Prefetch).  @derrickstolee has since adding unit tests so we know both functions work.
- Since we're redoing the way we look up last Prefetch in the future, I saw that as opportunity to nix the EpochConverter and switch everything to readable formated dates
- In high performance situations saving a long rather than a datetime makes a ton of sense.  It'll be faster for comparisons (+/- of long is faster than dates).  However, we're not doing that here.
- Prefetch files use long for datetime so using that method has consistency with the CacheServer
- The file timestamp should also be a clue to the last runtime, however we'd still need to convert a long to be sure

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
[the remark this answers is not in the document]
derrickstolee: I got this warning as I was running it on my machine:

```
[2018-12-14 10:15:46 -05:00] LooseObjectsStep (Start)
[2018-12-14 10:15:46 -05:00] Warning {"WarningMessage":"Skipping LooseObjectsStep due to git pids 1808"}
[2018-12-14 10:15:46 -05:00] LooseObjectsStep (Stop) {"DurationMs":18}
```

This seems helpful, especially if we start seeing users who never get the work done because they always have processes running. We can monitor via Kusto for user who always hit this warning instead of an actual job run.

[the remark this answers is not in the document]
jeschu1: This brings up a great point.  I'm currently not logging to telemetry.  We should, do that but carefully!

@derrickstolee I'm thinking if the time span has elapsed and we bail for active pids, that would be the only time we want a warning in telemetry.  Agree?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: I should certainly add a comment that forceRun is only done in testing for clarity.  I'll do that!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,90 @@
...
+                    // Perform prune-packed
+                    int beforeCount = this.CountLooseObjects();
+                    GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot));
+                    int afterCount = this.CountLooseObjects();
+
+                    // Record Telemetry
+                    EventMetadata metadata = new EventMetadata();
+                    metadata.Add("StartingCount", beforeCount);
+                    metadata.Add("EndingCount", afterCount);
+                    metadata.Add("RemovedCount", afterCount - beforeCount);
```
derrickstolee: This subtraction is backwards. Here are my logs:

```
[2018-12-14 10:21:44 -05:00] RunGitCommand (Start) {"Area":"LooseObjectsStep"}
[2018-12-14 10:21:46 -05:00] RunGitCommand (Stop) {"DurationMs":2230}
[2018-12-14 10:21:47 -05:00] LooseObjectsStep {"StartingCount":605193,"EndingCount":604491,"RemovedCount":-702}
[2018-12-14 10:21:47 -05:00] LooseObjectsStep (Stop) {"DurationMs":7314}
```

derrickstolee: Also, keep in mind that I over-hydrated my loose objects, so this "702" number is expected to be small. The next part will do the hard work of packing those loose objects.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,87 @@
...
+                    metadata.Add("StartingCount", beforeCount);
+                    metadata.Add("EndingCount", afterCount);
+                    metadata.Add("RemovedCount", beforeCount - afterCount);
+                    activity.RelatedEvent(EventLevel.Informational, this.Area, metadata, Keywords.Telemetry);
+
+                    this.SaveLastRunTimeToFile();
+                }
+                catch (Exception e)
+                {
+                    activity.RelatedError(this.CreateEventMetadata(e), "Failed to run LooseObjectsStep");
```
wilbaker: Should this be a warning rather than an error?

We've typically reserved errors for scenarios where:

- The mount process is dying
- Processes interacting with VFS4G will see an I\O error

We've taken this approach so that if a customer sees a failure (typically build failure) they can scan the VFS4G logs to see if VFS4G is the likely the culprit of the failure (if there are errors, it's our fault).

Rather than using `RelatedError` would it be sufficient to use `RelatedWarning` + `Keywords.Telemetry`?

GVFS/GVFS.Common/SHA1Util.cs
```diff
@@ -8,7 +8,12 @@ public static class SHA1Util
     {
         public static bool IsValidShaFormat(string sha)
         {
-            return sha.Length == 40 && !sha.Any(c => !(c >= '0' && c <= '9') && !(c >= 'a' && c <= 'f') && !(c >= 'A' && c <= 'F'));
+            return IsValidHex(sha, 40);
+        }
+
+        public static bool IsValidHex(string value, int length)
```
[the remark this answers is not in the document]
jeschu1: Created IsLooseObjectsDirectory in GitObjects.  

Since we call it in more than 1 spot let's make it obvious what it's checking and how.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -143,6 +147,45 @@ protected EventMetadata CreateEventMetadata(Exception e = null)
...
+                out Exception handledException))
+            {
+                this.Context.Tracer.RelatedError("Failed to record run time: {0}", handledException?.ToString());
+            }
+        }
+
+        protected IEnumerable<int> RunningGitProcessIds()
+        {
+            Process[] allProcesses = Process.GetProcesses();
+            return allProcesses.Where(x => x.ProcessName.Equals("git", StringComparison.OrdinalIgnoreCase)).Select(x => x.Id);
```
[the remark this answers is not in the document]
jeschu1: @derrickstolee can you confirm if 'git.exe' is the only process we care about here?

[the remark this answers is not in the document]
derrickstolee: Technically, the `git-<builtin>.exe` pattern exists for historical reasons. We still ship them with Git for Windows.

So yes, it is possible that someone is using these commands directly and we should be extra careful here by watching for `git*.exe`.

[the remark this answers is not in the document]
derrickstolee: Good catch! This is definitely an unlikely scenario, but worth guarding against.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -143,6 +147,45 @@ protected EventMetadata CreateEventMetadata(Exception e = null)
...
+        {
+            if (!this.Context.FileSystem.FileExists(this.LastRunTimeFilePath))
+            {
+                return true;
+            }
+
+            string lastRunTime = this.Context.FileSystem.ReadAllText(this.LastRunTimeFilePath);
+            if (!long.TryParse(lastRunTime, out long result))
+            {
+                this.Context.Tracer.RelatedError("Failed to parse long: {0}", lastRunTime);
```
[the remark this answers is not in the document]
jeschu1: Good question!

It will put an error in telemetry if the file exists and we can't read the timestamp.
Given this is a weird state I'd say skip the maintence, until we root cause (no worse than we have today).

@derrickstolee what are your thoughts on this?0

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,116 @@
...
+using GVFS.FunctionalTests.Tools;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixture]
+    [Category(Categories.FullSuiteOnly)]
```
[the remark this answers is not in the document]
derrickstolee: Or remove the category.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,87 @@
...
+            using (ITracer activity = this.Context.Tracer.StartActivity(this.Area, EventLevel.Informational, Keywords.Telemetry, metadata: null))
+            {
+                try
+                {
+                    // forceRun is only currently tree for functional tests
+                    if (!this.forceRun)
+                    {
+                        if (!this.EnoughTimeBetweenRuns())
+                        {
+                            activity.RelatedWarning("Skipping step due to not enough time between runs");
```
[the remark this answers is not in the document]
jeschu1: I'm OK with skipping it for now honestly.  We have telemetry to show when the last run was so we can determine it remotely.  If it we find the need for it we can always add it in the future.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: It should be a warning, good call!

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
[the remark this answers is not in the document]
derrickstolee: Tracing the fail to parse and running anyway should be fine. It will replace the file, we check for the git processes to avoid it being too soon.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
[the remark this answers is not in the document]
jeschu1: Spoke with @derrickstolee offline and we'll halt for anything that starts with "git". @alameenshah for reference

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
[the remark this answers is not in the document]
jeschu1: Updated to return true

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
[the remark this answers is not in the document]
jeschu1: OK Change of plans!  Too many things start with "git.." (i.e. for example GitServiceDaemon.exe)
We are checking for a weird case here.  And I'll try to explain best I can.

If a git command is running across LooseObjectsMaintenance that is fine.
If a git command is very long running and runs across two calls to LooseObjectsMaintenance, that is not fine.  This is because we may remove a loose object that is a new packfile that the process did not see.  Even if that happens the long running command may not fail.

Given how rare this case is and that the user should be using direct git commands we're sticking with git.exe.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,87 @@
...
+                        if (!this.EnoughTimeBetweenRuns())
+                        {
+                            activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to not enough time between runs");
+                            return;
+                        }
+
+                        IEnumerable<int> processIds = this.RunningGitProcessIds();
+                        if (processIds.Any())
+                        {
+                            activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to git pids {0}", string.Join(",", processIds), Keywords.Telemetry);
```
wilbaker: Can you avoid mixing the two different styles of string formatting or move `string.Join(",", processIds)` into an `EventMetadata`?

In other words, if you're sticking with `$"` formatting just use `pids {string.Join(",", processIds)}` rather than `pids {0}`.

GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -42,6 +42,11 @@ public enum DownloadAndSaveObjectResult
             Error
         }
 
+        public static bool IsLooseObjectsDirectory(string value)
+        {
+            return value.Length == 2 && !value.Any(c => !Uri.IsHexDigit(c));
```
wilbaker: Also update `SHA1Util.IsValidShaFormat` to use `IsHexDigit`?

GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            GitObjects.IsLooseObjectsDirectory("BB").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("bb").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("A7").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("55").ShouldEqual(true);
+            GitObjects.IsLooseObjectsDirectory("K7").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("A-").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("?B").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("BBB").ShouldEqual(false);
+            GitObjects.IsLooseObjectsDirectory("B-B").ShouldEqual(false);
```
wilbaker: Minor: Rather than `ShouldEqual(true)` and `ShouldEqual(false)` you can use `ShouldBeTrue()` and `ShouldBeFase()`

GVFS/GVFS.Common/Git/GitObjects.cs
jeschu1: I could, but does it seem weird for it to be in this PR now that I'm not using any SHAUtil functions.  I'm happy to do it if you are to approve :-)

GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
jeschu1: Ahh thanks for the tip!

GVFS/GVFS.Common/Git/GitObjects.cs
wilbaker: I am OK with whichever way you decide to go :) My thinking was that we might as well update the code to be consistent, but I am okay leaving it as-is as well.

kewillford: I would change this to use `All` so that it doesn't have double negatives.
```
return value.Length == 2 && value.All(c => Uri.IsHexDigit(c));
```

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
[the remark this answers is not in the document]
derrickstolee: Both `gitk` and `git-gui` are written in shell, so they get all of their information by executing `git.exe`.

Here is my understanding of the "worst that can happen" as we are deleting loose objects:

1. A process starts and loads pack-files.
2. We run our step and create a pack-file containing some of our loose objects.
3. We run our step _again_ and start deleting loose objects that we placed into pack-files the last time.
4. The process looks for a loose object, sees that it exists, then tries to read from the file (but it is deleted!). How the process fails at this point is unclear.

Note that if the process starts reading the loose object before we delete it, then the process is fine. If the process looks for existence after we delete it, then we will download the loose object again using the read-object hook.

I think this is an extremely rare error case.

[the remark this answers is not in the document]
wilbaker: @derrickstolee ok thanks for the details!

GVFS/GVFS.Common/InternalVerbParameters.cs
kewillford: Could this be an `enum` so that it is well defined?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -0,0 +1,87 @@
...
+                    }
+
+                    int beforeCount = this.CountLooseObjects();
+                    GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot));
+                    int afterCount = this.CountLooseObjects();
+
+                    EventMetadata metadata = new EventMetadata();
+                    metadata.Add("StartingCount", beforeCount);
+                    metadata.Add("EndingCount", afterCount);
+                    metadata.Add("RemovedCount", beforeCount - afterCount);
```
kewillford: Would the `GitObjectRoot` be useful here in the metadata?

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            MockDirectory hex2 = new MockDirectory(
+                Path.Combine(enlistment.GitObjectsRoot, "F1"), 
+                null, 
+                new List<MockFile>()
+                {
+                     new MockFile(Path.Combine(enlistment.GitObjectsRoot, "F1", "test1"), string.Empty),
+                     new MockFile(Path.Combine(enlistment.GitObjectsRoot, "F2", "test2"), string.Empty)
+                });
+
+            // Create NonHex Folder 2 with 2 Files
```
kewillford: Comment doesn't seem to match the code here.

GVFS/GVFS.Common/InternalVerbParameters.cs
jeschu1: I prefer to keep it a string at least for now.  Since it is used mostly for calling from function tests I think it's clearer to send in "LooseObjects" then say "0".  Let me know if I'm missing something.

GVFS/GVFS.UnitTests/Common/GitObjectsTests.cs
```diff
@@ -0,0 +1,24 @@
...
+using GVFS.Tests.Should;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class GitObjectsTests
+    {
+        [TestCase]
+        public void IsLooseObjectsDirectory()
```
kewillford: I would break this into 2 different test cases `IsLooseObjectsDirectory_ValidDirectories` and `IsLooseObjectsDirectory_InvalidDirectories` so a failure is more explicit about what didn't work.

GVFS/GVFS.Common/InternalVerbParameters.cs
kewillford: I was hoping that you could still pass in "LooseObjects" and the json deserialize would parse it into the enum correctly.  If not there are some checks in place so it should be okay.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: Great Idea!

Issue #607, opened 2018-12-11
Mac: Use $NOCANCEL variant of close()


wilbaker:
The errors returned by close() (and most likely fclose) are unreliable on Mac.

See discussions here:

https://github.com/rust-lang/libc/issues/595
https://bugs.chromium.org/p/chromium/issues/detail?id=269623

Both the user-mode ProjFS lib and VFS4G PInvokes should be updated.

Issue #606, opened 2018-12-11
Mac: xattrs and flags should be removed from a file when it is renamed out of a root


wilbaker:
Currently the xattrs are only removed when a file is converted to full (which is only happening when a file is modified).

PR #608, opened 2018-12-11
Mac: Cleanup while investigating converting files to full

### LINKED ISSUE #523: Mac: Confirm that two threads converting a file to full at the same time will not leave that file in a bad state
It's possible (likely?) that no changes will be required, but we should validate there aren't any races in this path that will leave a file in a bad state on disk.

wilbaker:
Resolves #523 

As best I could tell there are no problematic race conditions when converting a partial file to full, and the file will not be corrupted (converting a file to full only involves removes its xattrs).

During this investigation I filed two new issues:

#606 
#607 

And this PR has some updates to comments and some minor cleanup changes.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -89,9 +89,9 @@ static bool IsBitSetInFileFlags(const char* fullPath, uint32_t bit);
 
 static bool InitializeEmptyPlaceholder(const char* fullPath);
 template<typename TPlaceholder> static bool InitializeEmptyPlaceholder(const char* fullPath, TPlaceholder* data, const char* xattrName);
-static bool AddXAttr(const char* fullPath, const char* name, const void* value, size_t size);
-static bool GetXAttr(const char* fullPath, const char* name, size_t size, _Out_ void* value);
-static bool RemoveXAttr(const char* fullPath, const char* name, _Out_ int* error);
+static errno_t AddXAttr(const char* fullPath, const char* name, const void* value, size_t size);
+static bool TryGetXAttr(const char* fullPath, const char* name, size_t expectedSize, _Out_ void* value);
+static errno_t RemoveXAttrWithoutFollowingLinks(const char* fullPath, const char* name);
 
 static inline PrjFS_NotificationType KUMessageTypeToNotificationType(MessageType kuNotificationType);
 
@@ -551,7 +551,7 @@ PrjFS_Result PrjFS_DeleteFile(
     {
         // TODO(Mac): Determine if we need a similar check for directories as well
         PrjFSFileXAttrData xattrData = {};
-        if (!GetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData))
+        if (!TryGetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData))
         {
             *failureCause = PrjFS_UpdateFailureCause_FullFile;
             return PrjFS_Result_EVirtualizationInvalidOperation;
@@ -818,7 +818,7 @@ static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const
     CombinePaths(s_virtualizationRootFullPath.c_str(), relativePath, fullPath);
     
     PrjFSFileXAttrData xattrData = {};
-    if (!GetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData))
+    if (!TryGetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData))
     {
         return PrjFS_Result_EIOError;
     }
@@ -947,7 +947,7 @@ static PrjFS_Result HandleFileNotification(
 #endif
     
     PrjFSFileXAttrData xattrData = {};
-    bool partialFile = GetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData);
+    bool partialFile = TryGetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData);
 
     PrjFS_Result result = s_callbacks.NotifyOperation(
         0 /* commandId */,
@@ -964,10 +964,12 @@ static PrjFS_Result HandleFileNotification(
     {
         // PrjFS_NotificationType_FileModified is a post-modified FileOp event (that cannot be stopped
         // by the provider) and so there's no need to check the result of the call to NotifyOperation
-        int error;
-        if (!RemoveXAttr(fullPath, PrjFSFileXAttrName, &error))
+        errno_t result = RemoveXAttrWithoutFollowingLinks(fullPath, PrjFSFileXAttrName);
+        if (0 != result)
         {
-            // TODO(Mac) #395: Log error (if the error is something other than ENOATTR).
+            // TODO(Mac) #395: Log error
+            // Note that it's expected that RemoveXAttrWithoutFollowingLinks return ENOATTR if
+            // another thread has removed the attribute
         }
     }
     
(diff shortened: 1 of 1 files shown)


wilbaker:
@pmj, @kewillford, @nickgra, @jeschu1 please let me know if you agree with my assessment.  The relevant code is here:

https://github.com/Microsoft/VFSForGit/blob/bf44ddfcd281f6982c3ace02c1f6c5c0d1618542/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L963

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1120,10 +1124,13 @@ static bool GetXAttr(const char* fullPath, const char* name, size_t size, _Out_
     return false;
 }
 
-static bool RemoveXAttr(const char* fullPath, const char* name)
+static bool RemoveXAttr(const char* fullPath, const char* name, _Out_ int* error)
 {
+    *error = 0;
+    
     if (removexattr(fullPath, name, XATTR_NOFOLLOW))
```
jeschu1: Apologies in advance for the newbie question:

Is the right definition for removexattr https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/removexattr.2.html
i.e. it returns an int 0 for success, -1 for failure?

Once I have my mac I can check myself!

jeschu1: Thanks @nickgra for the verification!  I hear my mac is coming soon and I can't wait :-)
And that's what I was getting at, let's update the check.

jeschu1:
@wilbaker your changes make sense to me.

There are some questions i have on the existing code.  Since they are existing I wouldn't necessarily hold this PR up for.
Should setxattr, getxattr, removexattr be checked against -1 for failure?  For each would we want to do the error logging you are purposing for RemoveXattr?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: Thanks for the clarification @pmj!

Since the code is equivalent there isn't anything wrong here.  I am a fan of doing the "0 !=" as you suggest for clarity.  That said if the code already has multiple instances of this style, i'll leave it to @wilbaker to decide.

It's also a fair point to return errno_t.  Whatever we decide for this function perhaps we should consider for set/get as well.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -964,7 +964,11 @@ static PrjFS_Result HandleFileNotification(
     {
         // PrjFS_NotificationType_FileModified is a post-modified FileOp event (that cannot be stopped
         // by the provider) and so there's no need to check the result of the call to NotifyOperation
-        RemoveXAttr(fullPath, PrjFSFileXAttrName);
+        int error;
+        if (!RemoveXAttr(fullPath, PrjFSFileXAttrName, &error))
+        {
+            // TODO(Mac) #395: Log error (if the error is something other than ENOATTR).
```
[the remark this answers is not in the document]
wilbaker: > Is there a valid reason why we'd try to remove an attribute that doesn't exist?

This could happen if two threads are processing `PrjFS_NotificationType_FileModified` for the same file at the same time.

They would each see that the `xattr` exists:

```
bool partialFile = GetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData);
```

And then one of them would win and remove the xattr (and the other wouldn't find it).

I'll put a few more details in the comment here.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1111,7 +1115,7 @@ static bool GetXAttr(const char* fullPath, const char* name, size_t size, _Out_
 {
     if (getxattr(fullPath, name, value, size, 0, 0) == size)
```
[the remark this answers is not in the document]
wilbaker: I suspect the intention of this code was to perform a two part check:

1. Validate that `getxattr` succeeded
2. Validate that the size of the xattr returned by `getxattr` contains at least as much data as we requested

From the man page:

```
RETURN VALUES
    On success, the size of the extended attribute data is returned.  
    On failure, -1 is returned and the global variable errno is set as follows.
```

I can split this into to checks or add a comment to make this more readable.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: > Looking at the function as a whole though, I'm not quite clear on why we shouldn't just return an errno_t from the wrapper function directly instead of fiddling around with a boolean and an errno.

Two part answer:

- I went back and forth as to whether we should just return the errno.  Now that you're suggesting it as well that sounds like the better design 😄 

- I missed that we are using `errno_t` elsewhere in our code, so I'll make that change.  I went with `int` because I found that the headers define `errno` as an `int` rather than an `errno_t` (which I found surprising):

```
extern int * __error(void);
#define errno (*__error())
```

> Since the code is equivalent there isn't anything wrong here. I am a fan of doing the "0 !=" as you suggest for clarity. That said if the code already has multiple instances of this style, i'll leave it to @wilbaker to decide.

I am also not a fan of the implicit conversion so I'll make the change here and we can update other spots as we go.  (FWIW I'm also not a fan of our use of Yoda conditions, but I'm not sure where everyone else stands on that).

wilbaker:
> I think you're probably right about the question of converting files to full.

> I don't really like the new RemoveXAttr() wrapper function's semantics, if we care about the errno, we should just return that (or 0 in the success case). The bool return value doesn't add anything, it just makes the wrapper more complicated than the function it's wrapping.

@pmj thanks for taking a look.  I agree with your analysis of `RemoveXAttr`, I will have it simply return the `errno`. 

I might also rename the helper to something like `RemoveXAttrWithoutFollowingLinks` to make it a bit clearer what it's doing.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: I'm fine with the Yoda stuff if the left hand side is a very long expression (function call) so it's hard to see the comparison if it's all the way off to the right. I find it unnecessary on short expressions; the accidental assignment typo is caught by compilers these days anyway. (But I don't have strong feelings on this.)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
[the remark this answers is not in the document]
pmj: This is one place where the Yoda version would be nicer IMO. I'm not sure it's worth separating onto multiple lines, it shouldn't be too tricky to work out what's happening here.

wilbaker:
@jeschu1 

> Should setxattr, getxattr, removexattr be checked against -1 for failure?

I'll update these to use the `0 != ` approach discussed in a separate comment.

> For each would we want to do the error logging you are purposing for RemoveXattr?

That's a good idea, I'll check where these are called and stub them out for logging as needed.

wilbaker:
@pmj, @jeschu1, @nickgra thanks for the great feedback.  I just pushed a new set changes (with explicit `0 != <function call>`), please let me know if you have any other suggestions.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1035,10 +1041,16 @@ static bool InitializeEmptyPlaceholder(const char* fullPath, TPlaceholder* data,
         data->header.formatVersion = PlaceholderFormatVersion;
         
         static_assert(is_pod<TPlaceholder>(), "TPlaceholder must be a POD struct");
-        if (AddXAttr(fullPath, xattrName, data, sizeof(TPlaceholder)))
+        
+        errno_t result = AddXAttr(fullPath, xattrName, data, sizeof(TPlaceholder));
+        if (0 == result)
```
jeschu1: I'm happy with this as is.

If we wanted to go one step further we could define SUCCESS=0, and then check against that but I have no trouble understanding what's going on.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1097,37 +1109,37 @@ static bool IsBitSetInFileFlags(const char* fullPath, uint32_t bit)
...
     
-    return true;
+    return 0;
 }
 
-static bool GetXAttr(const char* fullPath, const char* name, size_t size, _Out_ void* value)
+static bool TryGetXAttr(const char* fullPath, const char* name, size_t expectedSize, _Out_ void* value)
 {
-    if (getxattr(fullPath, name, value, size, 0, 0) == size)
+    if (expectedSize != getxattr(fullPath, name, value, expectedSize, 0, 0))
```
jeschu1: The new variable name is much clearer thank you!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: > If we wanted to go one step further we could define SUCCESS=0, and then check against that but I have no trouble understanding what's going on.

It's unfortunate there's not already something defined for success being 0.  I thought about adding something but didn't want it to inadvertently get used in the wrong spot.  I think I'll leave this as-is for now and we can change it later if the 0s get too unwieldily.

PR #616, opened 2018-12-13
Update Git to include sparse push algorithm

derrickstolee:
See [this PR in microsoft/git](https://github.com/Microsoft/git/pull/88) for the changes here. Should speed up 'git push' by a significant amount.

Also includes Git for Windows v2.20.1(1).


derrickstolee:
I added `git push` to our large repo perf tests (with `--dry-run`, so it doesn't actually construct a pack and push it). But even better, I added the `git pack-objects` command that is run with push.

For a very small push (a few edits):

`master`: 3.8s
`sparse-push`: 0.3s

PR #620, opened 2018-12-13
Add unit tests for EpochConverter

derrickstolee:
We need EpochConverter to talk to the cache servers, and we are adding more dependencies on it with the maintenance steps. Add unit-level tests to ensure it is working correctly.

### DIFF
--- a/GVFS/GVFS.UnitTests/Common/EpochConverterTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/EpochConverterTests.cs
@@ -1,0 +1,55 @@
+﻿using GVFS.Common;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class EpochConverterTests
+    {
+        [TestCase]
+        public void DateToEpochToDate()
+        {
+            DateTime time = new DateTime(2018, 12, 18, 8, 12, 13, DateTimeKind.Utc);
+            DateTime converted = EpochConverter.FromUnixEpochSeconds(EpochConverter.ToUnixEpochSeconds(time));
+
+            time.ShouldEqual(converted);
+        }
+
+        [TestCase]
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.UnitTests/Common/EpochConverterTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        [TestCase]
+        public void FixedDates()
+        {
+            DateTime[] times = new DateTime[]
+            {
+                new DateTime(2018, 12, 13, 20, 53, 30, DateTimeKind.Utc),
+                new DateTime(2035, 1, 3, 5, 0, 59, DateTimeKind.Utc),
+                new DateTime(1989, 12, 31, 23, 59, 59, DateTimeKind.Utc)
```
jrbriggs: Should we also test that the subsequent second results in the same epoch?

derrickstolee: Mostly I was just making up an old date. Do you think the year boundary matters?

jrbriggs: You managed to make up a date within one second of a leap second, so i figured that was intentiona. I think we should have a unit test that we handle leap seconds correctly.

jrbriggs: Also, I think my previous comment was incorrect, the ambiguous datetime I think would be 1989/12/31 23:59:60 with 1990/01/01 00:00:00

derrickstolee: It's actually the 1990/1991 boundary, but `DateTime` doesn't register these as two seconds apart.

derrickstolee: https://en.wikipedia.org/wiki/Leap_second

PR #625, opened 2018-12-14
Remove trailing whitespace and enforce StyleCop rule SA1028

jamill:
@kewillford had suggested that we open issues to start driving down the number of rules that we have disable in our StyleCop ruleset - and this is a step in that direction. I wanted to see how people felt about this rule, fixing up violations in our code base, and seeing if this will cause disruption to work that is currently in flight. My hope is that the diff itself will be easy to look through, as it only contains whitespace differences for all but 1 file (`GVFS.ruleset`, where I remove the override for this rule).

Clean up existing instances of trailing whitespace in our codebase and
stop suppressing StyleCop rule SA1028 which enforces no trailing
whitespace:

    https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1028.md

These changes were generated by applying the StyleCop automatic fixup for SA1028.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# EditorConfig: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.cs]
+trim_trailing_whitespace = true
--- a/GVFS.sln
+++ b/GVFS.sln
@@ -4,6 +4,7 @@ VisualStudioVersion = 15.0.27428.2015
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{DCE11095-DA5F-4878-B58D-2702765560F5}"
 	ProjectSection(SolutionItems) = preProject
+		.editorconfig = .editorconfig
 		.gitattributes = .gitattributes
 		.gitignore = .gitignore
 		AuthoringTests.md = AuthoringTests.md
--- a/GVFS/FastFetch/CheckoutPrefetcher.cs
+++ b/GVFS/FastFetch/CheckoutPrefetcher.cs
@@ -28,7 +28,7 @@ namespace FastFetch
             int indexThreadCount,
             int checkoutThreadCount,
             bool allowIndexMetadataUpdateFromWorkingTree,
-            bool forceCheckout) 
+            bool forceCheckout)
                 : base(
                     tracer,
                     enlistment,
@@ -142,9 +142,9 @@ namespace FastFetch
                     if (!indexGen.HasFailures)
                     {
                         Index newIndex = new Index(
-                            this.Enlistment.EnlistmentRoot, 
+                            this.Enlistment.EnlistmentRoot,
                             this.Tracer,
-                            Path.Combine(this.Enlistment.DotGitRoot, GVFSConstants.DotGit.IndexName), 
+                            Path.Combine(this.Enlistment.DotGitRoot, GVFSConstants.DotGit.IndexName),
                             readOnly: false);
 
                         // Update from disk only if the caller says it is ok via command line
@@ -155,7 +155,7 @@ namespace FastFetch
                 }
             }
         }
-        
+
         /// <summary>
         /// * Updates local branch (N/A for checkout to detached HEAD)
         /// * Updates HEAD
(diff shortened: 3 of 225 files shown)


jeschu1:
So my concern is that for history a ton of stuff will have your name on it.  Maybe it's just a concern for @jamill  really :-).

I thought GitHub had a blame again feature but I'm not seeing it.

jamill:
> So my concern is that for history a ton of stuff will have your name on it

@jeschu1 - yes, this is a concern. This is (somewhat) mitigated by:

- you can use `git blame -w` to ignore whitespace changes (this is not obvious, or might not help until you actually run into a problem).

- On the github blame UI, you can view the previous blame.

I would be interested if people think this is worth the trade-off.

jamill:
The other trade off is we deal with non-functional whitespace changes as we encounter them in individual PRs. not as drastic of a single change, but the impact is spread out over time.

jamill:
> are there settings and\or extensions you can recommend for VS, VS for Mac, and VS Code that will strip trailing whitespace automatically?

Great question -

The StyleCop will flag trailing whitespace, and offer an option to fix it "everywhere", but it is a manual action.

There is an `.editorconfig` configuration we could use, that looks like it is supported in different editors, but I have not tried it everywhere. In playing around in Visual Studio (for windows), it looks like it will trim trailing whitespace when pressing "enter" on a line, but not necessarily trim it on save. I will play with this some more and see what other options are out there.

nickgra:
I use `.editorconfig` in other projects in VSCode and it works great. If we make this change (and I think we should), we should also set up `.editorconfig`. I've heard it also works in VS for Windows and Mac (docs for Mac here: https://docs.microsoft.com/en-us/visualstudio/mac/editorconfig). 

Notably, Xcode doesn't support .editorconfig so we'll need to keep in mind that our rules won't be applied to C++ code in ProjFS.Mac.

wilbaker:
> Notably, Xcode doesn't support .editorconfig so we'll need to keep in mind that our rules won't be applied to C++ code in ProjFS.Mac.

@jamill @nickgra I found this setting in Xcode:

image

In `XCode -> Preferences -> Text Editing`.

However, I'm guessing StyleCop won't apply to these files anyway.

> There is an `.editorconfig` configuration we could use, that looks like it is supported in different editors, but I have not tried it everywhere. In playing around in Visual Studio (for windows), it looks like it will trim trailing whitespace when pressing "enter" on a line, but not necessarily trim it on save. I will play with this some more and see what other options are out there

The `.editorconfig` sounds like a good option, can that file be checked into the repo?

nickgra:
@wilbaker Yep, in Xcode that's the feature we want turned on, but it won't light up automatically when someone is working in our repo with VSCode or VS2017/2019 like an `.editorconfig` would. I'll make sure to set that on my machine, but we can't enforce it like Stylecop does on our C# code.

Yes, the `.editorconfig` should be checked into the repo in the root. Here's an example of one in the VSCode repo: https://github.com/Microsoft/vscode/blob/master/.editorconfig

Since Jameson is out for the rest of the year, would anyone object to me adopting his commits, rebasing them, adding a .editorconfig file and creating a new PR to get this in? I run into this a fair bit and it frustrates me every time.

wilbaker:
> Since Jameson is out for the rest of the year, would anyone object to me adopting his commits, rebasing them, adding a .editorconfig file and creating a new PR to get this in? I run into this a fair bit and it frustrates me every time.

That's good with me, + @jrbriggs for his thoughts.

jrbriggs:
> That's good with me, + @jrbriggs for his thoughts.

@nickgra that'd be awesome, thank you!

jamill:
I rebased these changes onto latest and added an [EditorConfig](https://editorconfig.org) file.

Visual Studio has native support for EditorConfig ([documentation](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017))

VS Code does not have native support for EditorConfig, but there is a plug-in that can be installed: [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig).

A list of other editors that support EditorConfig can be found at [EditorConfig](https://editorconfig.org).

One other observation: While Visual Studio for Windows includes an automatic code fix option for the Roslyn Analyzer, I did not see this option for VS for Mac - which was a bit surprising.

/cc @nickgra

PR #626, opened 2018-12-14
Publish a mac build drop and consume it for functional test runs

jrbriggs:


Scripts/Mac/PrepFunctionalTests.sh
```diff
@@ -15,34 +15,3 @@ hdiutil attach $GITPUBLISH/*.dmg || exit 1
 GITPKG="$(find /Volumes/Git* -type f -name *.pkg)" || exit 1
 sudo installer -pkg "$GITPKG" -target / || exit 1
 hdiutil detach /Volumes/Git*
-
```
nickgra: Removing this will remove the ability for new devs/machines to install the current Java based GCM. This is fine for now since all of the machines are already configured correctly. Do we expect the new GCM to be coming in so soon that we want to remove this code now?

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -63,5 +63,15 @@ dotnet publish $VFS_SRCDIR/GVFS.sln /p:Configuration=$CONFIGURATION.Mac /p:Platf
 echo 'Copying Git installer to the output directory...'
 $VFS_SCRIPTDIR/PublishGit.sh $GITPATH || exit 1
 
+echo 'Installing shared data queue stall workaround...'
+# We'll generate a temporary project if and only if we don't find the correct dylib already in place.
+BUILDDIR=$VFS_OUTPUTDIR/GVFS.Build
+if [ ! -e $BUILDDIR/libSharedDataQueue.dylib ]; then
```
nickgra: Slick, this is a nice optimization! Do we want to do the same thing in the DownloadGVFSGit.sh script at this time?

Scripts/Mac/PrepFunctionalTests.sh
jrbriggs: I yanked installation here since I'm not certain that installing the cred manager is the responsibility of this script (named PrepFunctionalTests) now that our functional tests run against the anonymous endpoints.

I also expect the new GCM to be coming in soon for Mac.

If we want scripted installation of the GCM, should we consider creating a script to prep a dev environment? That will keep the functional test prep on ci machines quicker.

Scripts/Mac/BuildGVFSForMac.sh
jrbriggs: At this time, no, since it's not required for this change. I'll add it to my cleanup todo list.

Scripts/Mac/PrepFunctionalTests.sh
nickgra: Oh right, I forgot that we were testing against the anonymous endpoints now. Thanks for the reminder.

In that case, I think it's fine to remove this for now. Later when the new GCM is ready, we should make sure there's an easy way to install it on a dev machine (or maybe we rework our scripts to install Git and VFSForGit from the built installers like we do on Windows).

PR #628, opened 2018-12-14
Upgrade refactoring to enable Nuget upgrades

alameenshah:
## Refactor VFS4Git upgrade logic to support upgrade extensibility
This is the first PR of a series of changes to enable upgrade extensibility. In this one, UpgradeOrchestrator and ProductUpgrader are refactored so it becomes easy to incorporate Nuget upgrader in an upcoming PR.

### In this PR:
- Create a common interface for upgraders to implement (`IProductUpgrader`). `IProductUpgrader` encapsulates the API that components need to implement to download upgrade packages from specific feeds. A feed can be a GitHub release or a NuGet feed, for examples.

- `UpgradeOrchestrator`, which contains the logic to drive the upgrade process (including querying for updates, unmounting / remounting repositories, and orchestrating the upgrade process) works against the `IProductUpgrader` interface. This means the orchestration is agnostic to the specifics of where the upgrade packages are downloaded from and how they are installed.

- The existing `ProductUpgrader` class was renamed to `GitHubUpgrader`, as this component works against GitHub Releases.

- `GitHubUpgrader` implements `IProductUpgrader` interface.

 In an upcoming PR, a NugetUpgrader component (that also implements `IProductUpgrader` will be created. This component will know how to download and install upgrade packages from a NuGet feed.

- `ProductUpgrader` implements a factory method that creates an `IProductUpgrader` instance. It can be `GithubUpgrader` (or` NugetUpgrader` in an upcoming PR) based on Local GVFS Configuration.

- gvfs upgrade using existing Github/release upgrade works. UT also work.

### End-to-End
The [features/upgrade_extensibility](https://github.com/Microsoft/VFSForGit/tree/features/upgrade_extensibility) branch contains a preview of the end-to-end functionality. It includes a rough version of the upgrader that works with a NuGet feed. It can be used as a guide to see how we envision other upgraders to plug into the scaffolding being built in this PR.


GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -125,52 +118,52 @@ private bool LaunchInsideSpinner(Func<bool> method, string message)
...
             newVersion = null;
 
             Version newGVFSVersion = null;
-            GitVersion newGitVersion = null;
-            string errorMessage = null;
+            string error = null;
+            string consoleMessage;
+            bool isError;
+
+            if (!this.upgrader.CanRunUsingCurrentConfig(out isError, out consoleMessage, out error))
```
jrbriggs: Why do we need to have both an out string and an out bool here? Can we not always show `error` if it's not null or empty?

alameenshah: right, we don't need both. Will remove it.

jrbriggs:
> Is it possible to do that as a single commit so we see the diffs there easier?

@alameenshah thanks for rearranging the patch series. Other reviewers, please go commit by commit to avoid the rename noise.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -38,71 +38,139 @@ public partial class ProductUpgrader
...
+            {
+                string ring;
+                string error;
+                if (!this.Config.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, out ring, out error))
+                {
+                    ring = "invalid";
+                }
+
+                consoleMessage = null;
+                errorMessage = $"Invalid upgrade ring `{ring}` specified in gvfs config." + Environment.NewLine + GVFSConstants.UpgradeVerbMessages.SetUpgradeRingCommand;
```
jrbriggs: There are cases where this could read "Invalid upgrade ring `invalid` specified in gvfs config". Let's not use "gvfs" here, it's the product name: VFS for GIt. Also, I don't like "invalid ring invalid". Can we improve that error case?

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -0,0 +1,38 @@
...
+    public partial class ProductUpgrader
+    {
+        public static IProductUpgrader CreateUpgrader(ITracer tracer, out string error)
+        {
+            IProductUpgrader upgrader;
+            bool isEnabled;
+            bool isConfigured;
+            error = string.Empty;
+            
+            if ((upgrader = GitHubUpgrader.Create(tracer, out isEnabled, out isConfigured)) != null)
```
derrickstolee: this assignment inside the `if` is a bit messy. Pull the assignment out, then check `upgrader != null` in the `if` condition.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -411,6 +465,74 @@ private static bool TryCreateDirectory(string path, out Exception exception)
...
+        }
+
+        private bool TryRunGVFSInstaller(out bool installationSucceeded, out string error)
+        {
+            error = null;
+            installationSucceeded = false;
+
+            int exitCode = 0;
+            bool launched = this.TryRunInstallerForAsset(GVFSAssetId, out exitCode, out error);
+            installationSucceeded = exitCode == 0 || exitCode == RepoMountFailureExitCode;
```
jrbriggs: Aside from this line, the `TryRun*Installer` could be combined. Don't we run the VFS for Git installer with a parameter to not remount? Why do we need to special case that error code when the orchestrator would handle remounting?

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -57,7 +57,9 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="App.config" />
+    <None Include="App.config">
+      <SubType>Designer</SubType>
```
derrickstolee: I don't know the purpose of this change. Why is it here?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -73,14 +74,18 @@ private bool TryInitializeUpgrader()
 
                     this.tracer = jsonTracer;
                     this.prerunChecker = new InstallerPreRunChecker(this.tracer, this.Confirmed ? GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm : GVFSConstants.UpgradeVerbMessages.GVFSUpgrade);
-                    this.upgrader = new ProductUpgrader(ProcessHelper.GetCurrentProcessVersion(), this.tracer);
+                    this.upgrader = ProductUpgrader.CreateUpgrader(this.tracer, out error);
+                }
+                else
+                {
+                    error = null;
```
jrbriggs: Let's set this immediately on entering the method rather than having an else clause here.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -102,24 +106,25 @@ private bool TryRunProductUpgrade()
...
-                this.tracer.RelatedError($"{nameof(this.TryRunProductUpgrade)}: Could not load upgrade ring. {error}");
-                this.ReportInfoToConsole(GVFSConstants.UpgradeVerbMessages.InvalidRingConsoleAlert);
-                this.Output.WriteLine(errorOutputFormat, error);
-                return false;
-            }
-
-            if (ring == ProductUpgrader.RingType.None || ring == ProductUpgrader.RingType.NoConfig)
+            bool isError;
+            string consoleMessage;
+            if (!this.upgrader.CanRunUsingCurrentConfig(out isError, out consoleMessage, out error))
```
jrbriggs: I'm having a very hard time parsing this. `CanRunUsingCurrentConfig` returns bool, which I would expect to be whether or not we can run. But we also have an out param `isError`. What's an example of a time where we cannot run but it is not an error? Can we simplify this?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -55,29 +56,32 @@ public void Dispose()
         private void TimerCallback(object unusedState)
         {
             string errorMessage = null;
-
             InstallerPreRunChecker prerunChecker = new InstallerPreRunChecker(this.tracer, string.Empty);
-            ProductUpgrader productUpgrader = new ProductUpgrader(ProcessHelper.GetCurrentProcessVersion(), this.tracer);
-            if (prerunChecker.TryRunPreUpgradeChecks(out string _) && this.TryDownloadUpgrade(productUpgrader, out errorMessage))
+            IProductUpgrader productUpgrader = ProductUpgrader.CreateUpgrader(this.tracer, out errorMessage);
+
+            if (productUpgrader != null)
```
jrbriggs: Is it an error if the factory cannot instantiate an upgrader?

jrbriggs: Oh, I see. We're tracing later if anything has set `errorMessage`?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -105,5 +109,24 @@ private bool TryDownloadUpgrade(ProductUpgrader productUpgrader, out string erro
...
+
+        private void CleanupDownloadsDirectory(IProductUpgrader productUpgrader)
+        {
+            if (productUpgrader != null)
+            {
+                productUpgrader.CleanupDownloadDirectory();
+                return;
+            }
+
+            try
```
jrbriggs: It feels weird to delegate to the upgrader to clean up, but to also go whack the directory ourselves if for some reason we couldn't instantiate the upgrader. Why do we do that?

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
alameenshah: Ya... [looks like](https://github.com/Microsoft/msbuild/issues/2876) the build tools added it.  Manually removed it.

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: good point. I think this is a left over from initial days when installer itself used to do the re-mounting. Have removed it.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
wilbaker: This confused me as well.  Can this be updated to use the ` bool TryXXX` pattern (i.e. `TryCreateUpgrader`)

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: Couple times this can happen - when upgrade ring is `none` or user has not configured the ring yet. In both cases, upgrade cannot run but its not an error.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: >>Is it an error if the factory cannot instantiate an upgrader?

Yes.

>>We're tracing later if anything has set errorMessage?

Right.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: Directory cleanup code in ProductUpgradeTimer was not used anywhere. Removed it.

GVFS/GVFS.Common/IProductUpgrader.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// This can be used to report the beginning / end of each install step.
+    /// </summary>
+    /// <param name="method">The method to run inside wrapper</param>
+    /// <param name="message">The message to display</param>
+    /// <returns></returns>
+    public delegate bool InstallActionWrapper(Func<bool> method, string message);
+
+    public interface IProductUpgrader
+    {
+        bool CanRunUsingCurrentConfig(
```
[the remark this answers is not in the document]
alameenshah: On second thought having two separate parameters is more confusing than is helping. I reverted this back to a single parameter. Depending on whether isError is true or not, the caller would display message to user or log it as error.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -0,0 +1,749 @@
...
+            args = null;
+            return false;
+        }
+
+        private bool IsGVFSAsset(Asset asset)
+        {
+            return this.AssetInstallerNameCompare(asset, ProductUpgrader.GVFSInstallerFileNamePrefix, ProductUpgrader.VFSForGitInstallerFileNamePrefix);
+        }
+
+        private bool IsGitAsset(Asset asset)
```
wilbaker: Will some of this code be moved out when we support `NugetUpgrader`?  As best I can tell not all of this looks specific to GitHub.

GVFS/GVFS.Common/ProductUpgrader.Shared.cs
```diff
@@ -9,10 +9,10 @@ public partial class ProductUpgrader
         public const string UpgradeDirectoryName = "GVFS.Upgrade";
         public const string LogDirectory = "Logs";
         public const string DownloadDirectory = "Downloads";
+        public const string GVFSInstallerFileNamePrefix = "SetupGVFS";
+        public const string VFSForGitInstallerFileNamePrefix = "VFSForGit";
```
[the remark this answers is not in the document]
alameenshah: @jrbriggs - is there a date set for this

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockProductUpgrader.cs
```diff
@@ -7,15 +7,16 @@
 
 namespace GVFS.UnitTests.Windows.Mock.Upgrader
 {
-    public class MockProductUpgrader : ProductUpgrader
+    public class MockProductUpgrader : GitHubUpgrader
```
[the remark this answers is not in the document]
alameenshah: Right - have renamed to MockGitHubUpgrader. (I think we will need a separate PR for UT case refactoring & updates. Will discuss with @jamill and put that up once we have the details of NugetUpgrader available)

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
```diff
@@ -18,9 +18,9 @@ public override void Setup()
         public void UpgradeAvailableOnFastWhileOnLocalNoneRing()
         {
             this.SimulateUpgradeAvailable(
-                remoteRing: ProductUpgrader.RingType.Fast,
+                remoteRing: GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast,
```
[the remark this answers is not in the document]
alameenshah: All the Upgrader UT are today are geared towards GitHubUpgrader. I think we will need a separate PR for UT case refactoring & updates. Will discuss with @jamill and put that up once we have the details of NugetUpgrader available.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeTests.cs
```diff
@@ -19,19 +19,25 @@ public abstract class UpgradeTests
         protected MockTextWriter Output { get; private set; }
         protected MockInstallerPrerunChecker PrerunChecker { get; private set; }
         protected MockProductUpgrader Upgrader { get; private set; }
+        protected MockLocalGVFSConfig LocalConfig { get; private set; }
```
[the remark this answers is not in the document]
alameenshah: UpgradeTests is a generic class, although it knows GitHubUpgrader details today. As discussed above, UT refactoring can be done once we have NugetUpgrader details, in a separate PR.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -55,28 +56,37 @@ public void Dispose()
...
             {
-                return;
-            }
+                if (prerunChecker.TryRunPreUpgradeChecks(out string _) && this.TryDownloadUpgrade(productUpgrader, out errorMessage))
+                {
+                    return;
+                }
 
-            productUpgrader.CleanupDownloadDirectory();
+                productUpgrader.CleanupDownloadDirectory();
```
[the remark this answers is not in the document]
alameenshah: This is a merge from [this PR](https://github.com/Microsoft/VFSForGit/pull/533) and [this PR](https://github.com/Microsoft/VFSForGit/pull/601/files) and is not related to this refactoring. 

To summarize - download directory is cleaned up in 3 scenarios.
1. Upgrade is not installable on the PC (pre checks fail).
2. Download failed.
3. No new version is available.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -125,52 +118,53 @@ private bool LaunchInsideSpinner(Func<bool> method, string message)
...
+            string error = null;
+            bool isError;
+            if (!this.upgrader.CanRunUsingCurrentConfig(out isError, out error))
+            {
+                this.upgrader.CleanupDownloadDirectory();
+
+                if (isError)
+                {
+                    consoleError = error;
+                    this.tracer.RelatedError($"{nameof(this.TryRunUpgrade)}: Upgrade checks failed. {error}");
```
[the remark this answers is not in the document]
alameenshah: Only config settings. The error text doesn't make sense. Rephrased it.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -319,21 +282,21 @@ private bool TryCheckIfUpgradeAvailable(out Version newestVersion, out string co
...
-            
+
             return true;
         }
 
         private bool TryDownloadUpgrade(Version version, out string consoleError)
         {
             using (ITracer activity = this.tracer.StartActivity(
-                $"{nameof(this.TryDownloadUpgrade)}({version.ToString()})", 
+                $"{nameof(this.TryDownloadUpgrade)}({version.ToString()})",
```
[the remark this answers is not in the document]
alameenshah: Added it to metadata as well. Keeping it in the activityName is useful, it shows the version being downloaded in the logs even when there is no error. 

Like this -
```
[2018-12-19 11:46:11 -05:00] TryDownloadUpgrade(1.0.18297.1) (Start)
[2018-12-19 11:46:15 -05:00] Information {"Message":"Successfully downloaded version: 1.0.18297.1"}
[2018-12-19 11:46:15 -05:00] TryDownloadUpgrade(1.0.18297.1) (Stop) {"DurationMs":3300}
```

GVFS/GVFS.Common/ProductUpgrader.cs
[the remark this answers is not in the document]
alameenshah: GitHubUpgrader constructor now takes as GitHubUpgraderConfig as an argument. This new config class has been created so that in future NugetUpgrader can have its own NugetUpgraderConfig. The static Create method, helps keep the Upgrader specific Configuration objects private and within the respective Upgrader classes. The Factory method (`ProductUpgrader.TryCreateUpgrader()`) does not need to know details of GitHub or Nuget Upgrader Configs.

alameenshah:
>>Still trying to understand a lot of these changes but wanted to bring up one big question for me. Should there be a ProductUpgrader abstract base class instead of the interface and factory class? Seems like there is going to be shared code and this should make it trivial to create the NugetUpgrader.

In this PR the goal has been to focus on re-structuring existing code to facilitate easy incorporation of upcoming NugetUpgrader changes. NugetUpgrader is coming in the next PR. Once we know the details of NugetUpgrader, we can absolutely consider consolidating redundant code either in a common base class or use composition.

GVFS/GVFS.Common/LocalGVFSConfig.cs
```diff
@@ -13,16 +13,17 @@ public class LocalGVFSConfig
         private readonly PhysicalFileSystem fileSystem;
         private FileBasedDictionary<string, string> allSettings;
         
-        public LocalGVFSConfig()
+        public LocalGVFSConfig(FileBasedDictionary<string, string> settings = null)
```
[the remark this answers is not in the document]
alameenshah: right, it was not being set, removed it.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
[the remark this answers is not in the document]
wilbaker: @alameenshah sorry I should have been more clear in my suggestion.  `StartActivity` itself can take metadata, what does the logging look like if you pass the version in there (rather than as part of the activity name)?

wilbaker:
> [kewillford] Should there be a ProductUpgrader abstract base class instead of the interface and factory class? Seems like there is going to be shared code and this should make it trivial to create the NugetUpgrader.

> [alameenshah] In this PR the goal has been to focus on re-structuring existing code to facilitate easy incorporation of upcoming NugetUpgrader changes. NugetUpgrader is coming in the next PR. Once we know the details of NugetUpgrader, we can absolutely consider consolidating redundant code either in a common base class or use composition.

If we don't know the design\details of `NugetUpgrader` well enough yet to decide what should be moved to a base class should we hold off on re-structuring the existing code?  

Without an understanding of the design `NugetUpgrader` it's difficult to evaluate how well the changes in this PR are achieving their goal of making it easier to incorporate the upcoming `NugetUpgrader`.

jamill:
There is a prototype of the NugetUpgrader in the features/upgrade_extensibilty branch, in particular commit  fbc096. It should give an idea of the code common to both upgrades (currently in ProductUpgradetupgrader.cs, but as @alameenshah mentioned, maybe composition would be a better fit) and how another upgrader can be added with no (or minimal) changes to UpgradeOrchestrater. Hopefully that gives an idea of how everything can fit together.

The idea was this change focused on the interaction between UpgradeOrchestrator and the "upgrader" implementations. The next pr will look at the details of the ypgrader and how to share common code. If there is feedback on how we  could improve the API that the "Upgrader" components expose, I think that would be a valuable.

/cc @wilbaker @kewillford

wilbaker:
> The idea was this change focused on the interaction between UpgradeOrchestrator and the "upgrader" implementations. The next pr will look at the details of the ypgrader and how to share common code. If there is feedback on how we could improve the API that the "Upgrader" components expose, I think that would be a valuable.

@jamill @alameenshah ok thanks! It sounds like I wasn't focusing on the important part of this PR.  I will focus on the UpgradeOrchestrator <-> "upgrader" interactions.

alameenshah:
To give an idea of how much is common between Nuget and Github upgraders - adding a comparison table below of IProductUpgrader interface methods and how they are implemented in this PR vs in the Nuget prototype branch (`features/upgrade_extensibilty`).

| `IProductUpgrader`  method | `GitHubUpgrader` | `NugetUpgrader` | Commonality | 
| ------------- | ------------- | ------------- | ------------- |
| `CanRunUsingCurrentConfig`  | Checks upgrade.ring config to decide  | Checks feed URL, Credentials for Nuget & FeedName  | No commonality  |
| `Initialize`  | Empty placeholder  | Empty placeholder  | Unused by both, can possibly be removed  |
| `TryGetNewerVersion`  | Uses GitHub REST API to fetch new releases Json  | Uses Nuget Protocol (IPackageSearchMetadata) to get release info  | No commonality  |
| `TryGetGitVersion`  | Version info is read from Installer filename  | Version info contained in a Manifest file. (Don’t know if this implementation is final.)   | No commonality |
| `TryDownloadNewestVersion`  | Downloads Upgrade Assets using GitHub API  | Downloads package zip using Nuget API and extracts the zip file to final destination | No commonality  |
| `TryRunInstaller`  | Runs Git and GVFS installers only  | Iterates through all installers contained in the package  | Code to run installer exes can be isolated into a utility class and used by both Upgraders |
| `TryCleanup`  | Deletes the downloaded assets after successful upgrade| Not implemented  | Can potentially be common  |
| `CleanupDownloadDirectory`  | Deletes download directory  | Not implemented  | Possibly Common  |
| `TrySetupToolsDirectory`  | Copies upgrader tool and dependent assemblies to Temporary directory  | Copies upgrader tool and dependent assemblies to Temporary directory  | Common |

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -13,12 +13,12 @@ public class UpgradeVerb : GVFSVerb.ForNoEnlistment
     {
         private const string UpgradeVerbName = "upgrade";
         private ITracer tracer;
-        private ProductUpgrader upgrader;
+        private IProductUpgrader upgrader;
         private InstallerPreRunChecker prerunChecker;
         private ProcessLauncher processLauncher;
 
         public UpgradeVerb(
```
kewillford: Why is this class not using the `UpgradeOrchestrator`?  I thought the responsibility and purpose of the orchestrator is to handle what needs to be done for the upgrade but it seems this class is doing a lot of that work as well.  Can some of this code be moved there and see what common code we have?

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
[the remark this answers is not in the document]
alameenshah: oh ok. Here is how it looks -

```
[2018-12-20 12:40:48 -05:00] TryDownloadUpgrade(1.0.18297.1) (Start) {"Upgrade Step":"TryDownloadUpgrade","Version":"1.0.18297.1"}
[2018-12-20 12:40:54 -05:00] Information {"Message":"Successfully downloaded version: 1.0.18297.1"}
[2018-12-20 12:40:54 -05:00] TryDownloadUpgrade(1.0.18297.1) (Stop) {"DurationMs":6534}
```

Now version info gets logged twice. And your first comment to remove metadata from activityName makes sense, will remove it.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -411,6 +456,74 @@ private static bool TryCreateDirectory(string path, out Exception exception)
...
+            installationSucceeded = false;
+
+            int exitCode = 0;
+            bool launched = this.TryRunInstallerForAsset(GitAssetId, out exitCode, out error);
+            installationSucceeded = exitCode == 0;
+
+            return launched;
+        }
+
+        private bool TryRunGVFSInstaller(out bool installationSucceeded, out string error)
```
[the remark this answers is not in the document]
alameenshah: Cleaned it up.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: True. This is a cleanup that needs to be done, still there in my TODO list. It is not related to Nuget upgrader & this PR though.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -2,552 +2,39 @@
...
 using System.IO;
-using System.Net;
-using System.Net.Http;
-using System.Net.Http.Headers;
-using System.Runtime.Serialization;
-using System.Runtime.Serialization.Json;
 
 namespace GVFS.Common
 {
     public partial class ProductUpgrader
```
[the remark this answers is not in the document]
wilbaker: > Its only job now is to provide a Factory method for IProductUpgrader instances. How about ProductUpgraderFactory?

Maybe the next PR is the place to decide if the name should be changed?  If we anticipate removing `IProductUpgrader` and having an abstract `ProductUpgrader` base class then there's no need to rename this now.

Perhaps a more interesting question is whether this PR should be adding `IProductUpgrader` at all, or if we should just make `ProductUpgrader` the base class now and have `GitHubUpgrader` derive from it (we could still leave the task of moving common code here in a future PR).

Let me know what you think, I don't want to get this PR off-track, but if we anticipate removing `IProductUpgrader` in the next set of changes I'm not sure if it's worth adding it now.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
```diff
@@ -18,9 +18,9 @@ public override void Setup()
         public void UpgradeAvailableOnFastWhileOnLocalNoneRing()
         {
             this.SimulateUpgradeAvailable(
```
wilbaker: Should this class also be renamed?  It appears that it's now simulating\testing GitHub upgrades specifically.

wilbaker: Or if the plan is to make this generic in an upcoming PR then I think it's OK to leave the naming as-is.

GVFS/GVFS.Common/ProductUpgrader.cs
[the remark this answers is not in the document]
wilbaker: I just spoke with @jamill offline and I'm good with waiting for a future PR to decide how to best share code amongst the upgraders.  For now I'll focus on the `IProductUpgrader` interface and how the components interact.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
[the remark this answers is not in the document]
wilbaker: Ok thanks!

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -78,7 +78,7 @@ public void GVFSDownloadError()
             this.ConfigureRunAndVerify(
                 configure: () =>
                 {
-                    this.Upgrader.SetFailOnAction(MockProductUpgrader.ActionType.GVFSDownload);
+                    this.Upgrader.SetFailOnAction(MockGitHubUpgrader.ActionType.GVFSDownload);
```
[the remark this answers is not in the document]
alameenshah: Will create a separate PR for UT. This is what I have in mind -

**UpgradeOrchestrator & UpgradeVerb Tests**. Couple changes here.
1. What these classes test is high level upgrade flow, progress and error messaging. I don’t expect much changes in these tests. Currently the tests are a little polluted with details of Upgrade rings (which is specific to GitHubUpgrader) that probably will have to be moved out. 
2. The tests today are hard coded to use GitHubUpgrader instance only. New code has to be added so they could test the flow and messaging using IProductUpgrader instances. (The hard coding has to be removed and new code to test the flows using both GitHubUpgrader and NugetUpgrader instances has to be added.)

**ProductUpgrader Tests** - These are specific to GitHubUpgrader. The class has to be renamed. Probably consolidate test code spread across ProductUpgraderTests.cs and UpgradeTests.cs into one GitHubUpgraderTests.cs.

**New NugetUpgrader Tests** -  New tests to test specifics of Nuget upgrader.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
alameenshah: Agree. This and UpgradeTests.cs test GitHubUpgrader. Have to be renamed & consolidated. Please see my comment below.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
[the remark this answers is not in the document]
wilbaker: > Will create a separate PR for UT.

I was discussing this with @jamill a bit offline, but (at least for me) the state of the unit tests makes it harder to evaluate the interactions between the different components (e.g. the orchestrator, IProductUpgrader, etc.).

My understanding is that this PR's primary goal is to define the interaction between the components at a high level (without focusing on a specific `IProductUpgrader` implementation).  When I first saw `IProductUpgrader` I anticipated that there would be a pure “MockProductUpgrader” to test the interaction between the components.  As you pointed out though the product upgrader tests (and the `MockProductUpgrader`) are specific to `GitHubUpgrader`.  

For me, seeing all of the `GitHubUpgrader` specific code in the unit tests can make it tough to understand the high-level relationship between the components because the tests appear to be using GitHub specific concepts.  

For example, this test in `UpgradeVerbTests`:

```
[TestCase]
public void UpgradeAvailabilityReporting()
{
    this.ConfigureRunAndVerify(
        configure: () =>
        {
            this.SetUpgradeRing("Slow");
            this.Upgrader.PretendNewReleaseAvailableAtRemote(
                upgradeVersion: NewerThanLocalVersion,
                remoteRing: GitHubUpgrader.GitHubUpgraderConfig.RingType.Slow);
```

Uses GitHub specific upgrade rings, and reading it I'm not sure how the `UpgradeVerb` would work with non-GitHub `IProductUpgrader`.

> True. This is a cleanup that needs to be done, still there in my TODO list. It is not related to Nuget upgrader & this PR though

If there's still a TODO list for this feature to be production ready, should development be moved to a feature branch (with PRs into that feature branch) until the TODO list is complete?  

A feature branch would let you get some early feedback as you work through the different stages of this feature, and I would expect you'd be able to iterate more quickly (e.g. on `master` we should be making the required UT changes as we update the code, but they can be separate on a feature branch).

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -55,28 +56,37 @@ public void Dispose()
...
             InstallerPreRunChecker prerunChecker = new InstallerPreRunChecker(this.tracer, string.Empty);
-            ProductUpgrader productUpgrader = new ProductUpgrader(ProcessHelper.GetCurrentProcessVersion(), this.tracer);
-            if (prerunChecker.TryRunPreUpgradeChecks(out string _) && this.TryDownloadUpgrade(productUpgrader, out errorMessage))
+            IProductUpgrader productUpgrader;
+
+            if (ProductUpgrader.TryCreateUpgrader(out productUpgrader, this.tracer, out errorMessage))
             {
-                return;
-            }
+                if (prerunChecker.TryRunPreUpgradeChecks(out string _) && this.TryDownloadUpgrade(productUpgrader, out errorMessage))
```
wilbaker: Can\should `TryRunPreUpgradeChecks` be checked before creating an upgrader (`TryCreateUpgrader`)?

GVFS/GVFS.Common/IProductUpgrader.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// This can be used to report the beginning / end of each install step.
+    /// </summary>
+    /// <param name="method">The method to run inside wrapper</param>
+    /// <param name="message">The message to display</param>
+    /// <returns>success or failure return from the method run.</returns>
+    public delegate bool InstallActionWrapper(Func<bool> method, string message);
+
+    public interface IProductUpgrader
+    {
+        bool CanRunUsingCurrentConfig(out bool isConfigError, out string message);
```
wilbaker: Style suggestion:  It looks like this interface could be switched to a `TryXXX` style method, which would be more consistent with our other code for methods that may fail.

Perhaps something like:

```
/// <summary>
/// Checks if the current current upgrade configuration allows the upgrader to run
/// </summary>
/// <returns>
///    - true: If the configuration was valid and could be read
///    - false: if the configuration was invalid or could not be read
///</returns>
public bool TryGetConfigAllowsUpgrade(out bool configAllowsUpgrade, out string message)
```

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -55,28 +56,37 @@ public void Dispose()
...
 
-        private bool TryDownloadUpgrade(ProductUpgrader productUpgrader, out string errorMessage)
+        private bool TryDownloadUpgrade(IProductUpgrader productUpgrader, out string errorMessage)
         {
             using (ITracer activity = this.tracer.StartActivity("Checking for product upgrades.", EventLevel.Informational))
             {
                 Version newerVersion = null;
                 string detailedError = null;
+                bool isError;
+                if (!productUpgrader.CanRunUsingCurrentConfig(out isError, out errorMessage))
```
wilbaker: Why did we not need to check this before?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -48,21 +48,23 @@ protected override string VerbName
...
             get { return UpgradeVerbName; }
         }
-        
+
         public override void Execute()
         {
             ReturnCode exitCode = ReturnCode.Success;
-            if (!this.TryInitializeUpgrader() || !this.TryRunProductUpgrade())
+            string error;
+            if (!this.TryInitializeUpgrader(out error) || !this.TryRunProductUpgrade())
```
wilbaker: At a high level, what happens if a user has downloaded an upgrader from GitHub, and then their organization switches to using a NuGet feed?

Will the upgrade verb somehow know that it shouldn't run the installer downloaded from GitHub, and that it should run the NuGet installer?

In general the question is how does the upgrade verb know which installer to run if GVFS.Service has downloaded different upgraders from different locations.  (If this is in the code and I've missed it, please point me in the right direction 😄 )

jamill: Good question - the initial design does not do anything special if an organization switches to a different upgrader (e.g. switches to use a NuGet feed or vice versa). The upgrader does not uninstall any previous installations, it only runs the provided installer. I think this is reasonable for the initial version of this feature to keep complexity down, but I am interesting if there are opinions that this is insufficient.

In the current version of the NuGet upgrader change, If both upgraders are configured, the code currently prefers the NuGet upgrader. I could see this being an error condition (allowing the user to correct the ambiguous configuration).

wilbaker: > The upgrader does not uninstall any previous installations, it only runs the provided installer. I think this is reasonable for the initial version of this feature to keep complexity down

I agree it's better to go with the less complex option for now.

> In the current version of the NuGet upgrader change, If both upgraders are configured, the code currently prefers the NuGet upgrader. I could see this being an error condition (allowing the user to correct the ambiguous configuration)

What will happen for organizations that use the current version of VFS4G (that's configured to download upgrades from GitHub) when they switch to using Nuget?

If an update was downloaded from GitHub and then they switch to Nuget will VFS4G do the right thing (assuming version numbers keep increasing)?  Or would the old upgrader downloaded from GitHub cause problems?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: `TryCreateUpgrader` checks if the upgrade configuration is set correctly. It helps early exit with error (when config is incorrect) or info (when config is set to disable upgrade) message. The sequence of error/info messages that get displayed has gone through multiple rounds of reviews. The sequence is the same in `UpgradeVerb`, `UpgraderTool` and in `GVFS.Service` Upgrade timer. If we change the order to do `TryRunPreUpgradeChecks` first, then this message sequence would change as well. 

Example: After a fresh install of VFSForGit, `gvfs upgrade` (or service timer) might display pre-check specific error and exit, which might be incorrect because the User has not yet specified upgrade config (in this case the better behavior would be to give instruction to user on how to set upgrade config.)

GVFS/GVFS.Common/ProductUpgrader.cs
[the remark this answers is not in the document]
wilbaker: Re-opening this comment based on today's offline discussion that this should be renamed as part of this PR.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
alameenshah: The code was there before in `UpgradeVerb` & `UpgradeOrchestrator`, but not as a separate method. Now with upcoming Nuget upgrader, it made sense to make it as a separate interface method, so the verb and orchestrator don't need to know about Github release rings or Nuget feed URLs.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: >>If an update was downloaded from GitHub and then they switch to Nuget will VFS4G do the right thing (assuming version numbers keep increasing)? Or would the old upgrader downloaded from GitHub cause problems?

The old upgrader will not cause problems. 
1. Switching from GitHub to Nuget, while a previously downloaded GitHub release exists. Nuget upgrader downloads Nuget archive, un-zips it and uses the un-archived installer. Even if the older installer has the same name it will get overwritten (@jamill please correct me if I'm wrong.)
2. Switching from Nuget to GitHub, while a previously downloaded Nuget release exists. GitHubUpgrader always downloads new installer copy and uses it. It will overwrite any previously downloaded installers with the same name.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -179,14 +180,14 @@ private bool TryRunUpgrade(out Version newVersion, out string consoleError)
...
-                consoleError = errorMessage;
+                consoleError = error;
                 return false;
             }
 
             if (!this.LaunchInsideSpinner(
                 () =>
                 {
-                    if (!this.preRunChecker.TryUnmountAllGVFSRepos(out errorMessage))
+                    if (!this.preRunChecker.TryUnmountAllGVFSRepos(out error))
```
jamill: Reading through the code - it feels like this code to mount and unmount should be next to each other.

```
TryUnmount(...);
try
{
    DoWork(...);
}
finally
{
    TryMount(...);
}

File: GVFS/GVFS.Common/GitHubUpgrader.cs [hunk 30]
```diff
@@ -40,70 +40,129 @@ public partial class ProductUpgrader
...
+        public bool TryInitialize(out string errorMessage)
+        {
+            errorMessage = null;
+            return true;
+        }
 
-        public bool TryGetNewerVersion(
-            out Version newVersion,
-            out string errorMessage)
+        public bool TryGetConfigAllowsUpgrade(out bool upgradeAllowed, out string message)
```
jamill: I still get caught up on the different aspects this method is checking for, and the different return values...

1) Would `configuredForUpgrade` be a more accurate description of what this method is checking?
2) Could this method only be responsible for determining if it is properly configured for upgrade, and let another method be responsible determining whether there is a misconfiguration or not?

GVFS/GVFS.Common/IProductUpgrader.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <returns>success or failure return from the method run.</returns>
+    public delegate bool InstallActionWrapper(Func<bool> method, string message);
+
+    public interface IProductUpgrader
+    {
+        bool TryGetConfigAllowsUpgrade(out bool isConfigError, out string message);
+
+        bool TryInitialize(out string errorMessage);
+
+        bool TryGetNewerVersion(out Version newVersion, out string message);
```
jamill: I get caught up on `GetNewerVersion` and am not sure if "get" means we are going to download it, or if we are just going to report the latest version.

Would something like `QueryNewestVersion` be more descriptive?

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -40,70 +40,129 @@ public partial class ProductUpgrader
...
+                    gitHubUpgraderConfig);
+            }
+
+            return upgrader;
         }
 
-        public RingType Ring { get; protected set; }
+        public bool TryInitialize(out string errorMessage)
+        {
+            errorMessage = null;
```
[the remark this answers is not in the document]
alameenshah: It was created as a place holder, to hold all initialization code. But ended up un-necessary. Removed it.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -165,28 +224,48 @@ public bool TryDownloadNewestVersion(out string errorMessage)
...
+        public bool TryRunInstaller(InstallActionWrapper installActionWrapper, out string error)
         {
-            error = null;
-            installationSucceeded = false;
+            string localError;
 
-            int exitCode = 0;
-            bool launched = this.TryRunInstallerForAsset(GitAssetId, out exitCode, out error);
-            installationSucceeded = exitCode == 0;
+            this.TryGetGitVersion(out GitVersion newGitVersion, out localError);
```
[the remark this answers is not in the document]
alameenshah: The GitVersion is used only for messaging. If there is an installer asset with a known GitInstaller pre-fix name, installation will succeed.

GVFS/GVFS.Common/IProductUpgrader.cs
alameenshah: agree. renamed to `TryQueryNewestVersion`

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -40,70 +40,129 @@ public partial class ProductUpgrader
...
+            GitHubUpgraderConfig upgraderConfig)
+            : this(currentVersion, tracer)
+        {
+            this.Config = upgraderConfig;
+        }
+
+        public GitHubUpgrader(string currentVersion, ITracer tracer)
         {
             this.installedVersion = new Version(currentVersion);
             this.fileSystem = new PhysicalFileSystem();
```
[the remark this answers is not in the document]
alameenshah: Creation of `GitHubUpgrader` is the first step in upgrade process. So any failures in creation of `PhysicalFileSystem` would cause an early exit.

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: Thank you for pointing this out. I received comments from @jrbriggs also about this being difficult to follow. Made the following changes.

* Split the code into two. 1. Loading the config and 2. Deciding whether enough config info is available to do an actual upgrade.
* Updated `GitHubUpgrader.Create()` method to fail when config load fails. When this happens, `ProductUpgraderFactory.TryCreateUpgrader()` will return null with config error messaging. 

Now the callers (`UpgradeVerb`, `UpgradeOrchestrator`, `GVFS.Service`) don't need to check for config load errors. They can use `IProductUpgrader.UpgradeAllowed()` to continue with upgrade process.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
alameenshah: As discussed, this is un-related to current Nuget refactoring. Deferring to a future PR.

GVFS/GVFS.Common/IProductUpgrader.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        bool TryGetGitVersion(out GitVersion gitVersion, out string error);
+
+        bool TryDownloadNewestVersion(out string errorMessage);
+
+        bool TryRunInstaller(InstallActionWrapper installActionWrapper, out string error);
+
+        bool TryCleanup(out string error);
```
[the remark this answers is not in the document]
alameenshah: As discussed, moved `CleanupDownloadDirectory ` to outside of the interface. Renamed it to `DeleteAllInstallerDownloads` and temporarily (should be moved to Upgrader utilities class you are working on) kept it in `ProductUpgraderInfo`

GVFS/GVFS.Common/IProductUpgrader.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <param name="method">The method to run inside wrapper</param>
+    /// <param name="message">The message to display</param>
+    /// <returns>success or failure return from the method run.</returns>
+    public delegate bool InstallActionWrapper(Func<bool> method, string message);
+
+    public interface IProductUpgrader
+    {
+        bool TryGetConfigAllowsUpgrade(out bool isConfigError, out string message);
+
+        bool TryInitialize(out string errorMessage);
```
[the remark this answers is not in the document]
alameenshah: sure.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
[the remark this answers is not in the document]
alameenshah: renamed it to `UpgradeOrchestratorWithGitHubUpgraderTests.cs`

alameenshah:
> @alameenshah Now that #667 is merged, I've requeued your large repo build and it should pass. Sorry that it is blocking you.

Thank you @derrickstolee

Issue #518, opened 2018-11-16
Version is reported as 0.0.0.0


jrbriggs:
On macOS, the version is being reported incorrectly:
```
jrbriggs@jrbriggs-mac:~/test$ gvfs version
GVFS 0.0.0.0
```

wilbaker:
This might be the same underlying issue as #95

wilbaker:
See also:

https://stackoverflow.com/questions/42138418/equivalent-to-assemblyinfo-in-dotnet-core-csproj
https://stackoverflow.com/questions/39163558/do-i-need-assemblyinfo-while-working-with-net-core

One issue preventing the version from getting stamped on the Mac is that `GVFS.PreBuild` is not getting built (and it generates CommonAssemblyVersion.cs).

PR #632, opened 2018-12-17
Clean up Versioning and AssemblyInfo for .NET Core Projects

### LINKED ISSUE #95: Make sure copyright information is on all .NET Standard\Core assemblies
Some of the projects that have already been ported seem to have lost their copyright info

nickgra:
Fixes #95  #518 

When we started to port our codebase to .NET Core, we initially removed the AssemblyInfo.cs files and moved to the autogenerated AssemblyInfo that .NET Core provides. This is something we likely want to do long term, but since we're in a mixed state, let's restore the old AssemblyInfo.cs files and generate the CommonAssemblyVersion.cs file in our build script like we do on Windows in GVFS.Prebuild.

While we're here, update our Copyright string to be 2019 everywhere (as it was 2017 in some places and 2018 in others).

After this change, our assemblies have proper versions and have the right metadata stamped on them:
```
nickgra-mbp:Mac nickgra$ /Users/nickgra/Publish/fastfetch --version
FastFetch 0.2.173.2

nickgra-mbp:Mac nickgra$ /Users/nickgra/Publish/gvfs --version
GVFS 0.2.173.2

nickgra-mbp:Mac nickgra$ /Users/nickgra/Publish/gvfs.mount --version
GVFS 0.2.173.2
```

screen shot 2018-12-17 at 2 33 22 pm


GVFS/FastFetch/FastFetch.csproj
```diff
@@ -44,6 +44,12 @@
     </Compile>
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="$(BuildOutputDir)\CommonAssemblyVersion.cs">
+      <Link>CommonAssemblyVersion.cs</Link>
+    </Compile>
+  </ItemGroup>
```
kewillford: Can we put this in a `.props` file instead of adding it to each `.csproj`?

GVFS/GVFS.Common/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,22 @@
...
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+
+// The following GUID is for the ID of the typelib if this project is exposed to COM
+[assembly: Guid("9ea6ff63-6bb0-4440-9bfb-0ae79a8f9ba9")]
```
kewillford: Can some/most of this information be in the Common file?

jrbriggs: Nit: Also end with newline please.

nickgra: I think a lot of this can be outright deleted actually, if not moved into the Common generated file. Whenever we fully move to .NET Core, I'd like to again revisit this so I'd rather keep this as is for now since this gets us into a healthy state for Early Adopters. Does anyone feel strongly that we should consolidate this now?

I did fix the new lines on the newly created AssemblyInfo.cs files though, thanks for the good catch!

nickgra: I agree completely, I plan to come back to this before it's time to update the copyright string again though :)

PR #635, opened 2018-12-18
PackfileMaintenanceStep

derrickstolee:
Create a new maintenance step for cleaning up the packfiles in the shared object cache.

This is done in two steps:

1. `git multi-pack-index expire`: This deletes the pack-files whose objects appear in newer pack-files. The multi-pack-index prevents `git` from looking at these packs. Rewrites the `multi-pack-index` to no longer refer to these (deleted) packs.

2. `git multi-pack-index repack --batch-size=<size>` inspects packs covered by the multi-pack-index in modified-time order (ascending). Greedily selects a batch of packs whose file sizes are all less than "size", but that sum up to at least "size". Then generate a new pack-file containing the objects that are uniquely referenced by the `multi-pack-index`.

See [this PR in microsoft/git](https://github.com/Microsoft/git/pull/84) for details of the `git` implementation.

There are some important details:

* The `expire` command will not delete packs that are marked with a `.keep` file. So, we can keep our most-recent prefetch pack around by writing a `.keep` file. This avoids any change to the logic for the most-recent good prefetch timestamp.

* The `expire` command frequently fails to actually delete `.idx` files in a VFS for Git repo. This is due to something holding a handle to the files, as later we are able to delete them; we've seen this problem before with scripts that try to clean up the pack directory. We clean up `.idx` files that are not paired with a `.pack` after we run the `expire` step. This can generate warnings in the `multi-pack-index write` commands as it tries to add the packs based on their `.idx` files. The commands succeed, however.

* According to the [`git pack-objects` documentation](https://git-scm.com/docs/git-pack-objects#git-pack-objects---windowltngt), we should set the `pack.window` and `pack.depth` config settings to zero to avoid delta calculations. This saves significant computation time when our packs contain blobs, as the deltification algorithm can be very slow. If we were packing only commits and trees, the problem is not as significant. It is important that we don't set this globally, that way we actually compute deltas on push. Instead, just set the settings using the `-c` mechanism in Git.

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181219.1</GitPackageVersion>
+    <GitPackageVersion>2.20181221.1</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
--- a/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
+using System.Security;
 using System.Threading;
 
 namespace GVFS.Common.FileSystem
@@ -71,6 +72,27 @@ namespace GVFS.Common.FileSystem
         public virtual void WriteAllText(string path, string contents)
         {
             File.WriteAllText(path, contents);
+        }
+
+        public virtual bool TryWriteAllText(string path, string contents)
+        {
+            try
+            {
+                this.WriteAllText(path, contents);
+                return true;
+            }
+            catch (IOException)
+            {
+                return false;
+            }
+            catch (UnauthorizedAccessException)
+            {
+                return false;
+            }
+            catch (SecurityException)
+            {
+                return false;
+            }
         }
 
         public Stream OpenFileStream(string path, FileMode fileMode, FileAccess fileAccess, FileShare shareMode, bool callFlushFileBuffers)
--- a/GVFS/GVFS.Common/InternalVerbParameters.cs
+++ b/GVFS/GVFS.Common/InternalVerbParameters.cs
@@ -4,16 +4,22 @@ namespace GVFS.Common
 {
     public class InternalVerbParameters
     {
-        public InternalVerbParameters(string serviceName, bool startedByService, string maintenanceJob)
+        public InternalVerbParameters(
+            string serviceName = null,
+            bool startedByService = true,
+            string maintenanceJob = null,
+            string packfileMaintenanceBatchSize = null)
         {
             this.ServiceName = serviceName;
             this.StartedByService = startedByService;
             this.MaintenanceJob = maintenanceJob;
+            this.PackfileMaintenanceBatchSize = packfileMaintenanceBatchSize;
         }
 
         public string ServiceName { get; private set; }
         public bool StartedByService { get; private set; }
         public string MaintenanceJob { get; private set; }
+        public string PackfileMaintenanceBatchSize { get; private set; }
 
         public static InternalVerbParameters FromJson(string json)
         {
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
@@ -9,6 +9,10 @@ namespace GVFS.Common.Maintenance
     {
         private readonly TimeSpan looseObjectsDueTime = TimeSpan.FromMinutes(5);
         private readonly TimeSpan looseObjectsPeriod = TimeSpan.FromHours(6);
+
+        private readonly TimeSpan packfileDueTime = TimeSpan.FromMinutes(30);
+        private readonly TimeSpan packfilePeriod = TimeSpan.FromHours(12);
+
         private readonly TimeSpan prefetchPeriod = TimeSpan.FromMinutes(15);
 
         private List<Timer> stepTimers;
@@ -54,16 +58,22 @@ namespace GVFS.Common.Maintenance
             {
                 TimeSpan prefetchPeriod = TimeSpan.FromMinutes(15);
                 this.stepTimers.Add(new Timer(
-                    (state) => this.queue.TryEnqueue(new PrefetchStep(this.context, this.gitObjects, requireCacheLock: true)),
+                    (state) => this.queue.TryEnqueue(new PrefetchStep(this.context, this.gitObjects)),
                     state: null,
                     dueTime: this.prefetchPeriod,
                     period: this.prefetchPeriod));
 
                 this.stepTimers.Add(new Timer(
-                    (state) => this.queue.TryEnqueue(new LooseObjectsStep(this.context, requireCacheLock: true)),
+                    (state) => this.queue.TryEnqueue(new LooseObjectsStep(this.context)),
                     state: null,
                     dueTime: this.looseObjectsDueTime,
                     period: this.looseObjectsPeriod));
+
+                this.stepTimers.Add(new Timer(
+                    (state) => this.queue.TryEnqueue(new PackfileMaintenanceStep(this.context)),
+                    state: null,
+                    dueTime: this.packfileDueTime,
+                    period: this.packfilePeriod));
             }
         }
     }
--- a/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
@@ -1,0 +1,167 @@
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    /// <summary>
+    /// This step maintains the packfiles in the object cache.
+    ///
+    /// This is done in two steps:
+    ///
+    /// git multi-pack-index expire: This deletes the pack-files whose objects
+    /// appear in newer pack-files. The multi-pack-index prevents git from
+    /// looking at these packs. Rewrites the multi-pack-index to no longer
+    /// refer to these (deleted) packs.
+    ///
+    /// git multi-pack-index repack --batch-size= inspects packs covered by the
+    /// multi-pack-index in modified-time order(ascending). Greedily selects a
+    /// batch of packs whose file sizes are all less than "size", but that sum
+    /// up to at least "size". Then generate a new pack-file containing the
+    /// objects that are uniquely referenced by the multi-pack-index.
+    /// </summary>
+    public class PackfileMaintenanceStep : GitMaintenanceStep
+    {
+        public const string PackfileLastRunFileName = "pack-maintenance.time";
+        public const string DefaultBatchSize = "2g";
+        private readonly bool forceRun;
+        private readonly string batchSize;
+
+        public PackfileMaintenanceStep(
+            GVFSContext context,
+            bool requireObjectCacheLock = true,
+            bool forceRun = false,
+            string batchSize = DefaultBatchSize)
+            : base(context, requireObjectCacheLock)
+        {
+            this.forceRun = forceRun;
+            this.batchSize = batchSize;
+        }
+
+        public override string Area => nameof(PackfileMaintenanceStep);
+        protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", PackfileLastRunFileName);
+        protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(1);
+
+        // public only for unit tests
+        public void GetPackFilesInfo(out int count, out long size, out bool hasKeep)
+        {
+            count = 0;
+            size = 0;
+            hasKeep = false;
+
+            foreach (DirectoryItemInfo info in this.Context.FileSystem.ItemsInDirectory(this.Context.Enlistment.GitPackRoot))
+            {
+                string extension = Path.GetExtension(info.Name);
+
+                if (string.Equals(extension, ".pack", StringComparison.OrdinalIgnoreCase))
+                {
+                    count++;
+                    size += info.Length;
+                }
+                else if (string.Equals(extension, ".keep", StringComparison.OrdinalIgnoreCase))
+                {
+                    hasKeep = true;
+                }
+            }
+        }
+
+        // public only for unit tests
+        public List<string> CleanStaleIdxFiles(out int numDeletionBlocked)
+        {
+            List<DirectoryItemInfo> packDirContents = this.Context
+                                                          .FileSystem
+                                                          .ItemsInDirectory(this.Context.Enlistment.GitPackRoot)
+                                                          .ToList();
+
+            numDeletionBlocked = 0;
+            List<string> deletedIdxFiles = new List<string>();
+
+            // If something (probably VFS for Git) has a handle open to a ".idx" file, then
+            // the 'git multi-pack-index expire' command cannot delete it. We should come in
+            // later and try to clean these up. Count those that we are able to delete and
+            // those we still can't.
+
+            foreach (DirectoryItemInfo info in packDirContents)
+            {
+                if (string.Equals(Path.GetExtension(info.Name), ".idx", StringComparison.OrdinalIgnoreCase))
+                {
+                    string pairedPack = Path.ChangeExtension(info.FullName, ".pack");
+
+                    if (!this.Context.FileSystem.FileExists(pairedPack))
+                    {
+                        if (this.Context.FileSystem.TryDeleteFile(info.FullName))
+                        {
+                            deletedIdxFiles.Add(info.Name);
+                        }
+                        else
+                        {
+                            numDeletionBlocked++;
+                        }
+                    }
+                }
+            }
+
+            return deletedIdxFiles;
+        }
+
+        protected override void PerformMaintenance()
+        {
+            using (ITracer activity = this.Context.Tracer.StartActivity(this.Area, EventLevel.Informational, Keywords.Telemetry, metadata: null))
+            {
+                // forceRun is only currently true for functional tests
+                if (!this.forceRun)
+                {
+                    if (!this.EnoughTimeBetweenRuns())
+                    {
+                        activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to not enough time between runs");
+                        return;
+                    }
+
+                    IEnumerable<int> processIds = this.RunningGitProcessIds();
+                    if (processIds.Any())
+                    {
+                        activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
+                        return;
+                    }
+                }
+
+                this.GetPackFilesInfo(out int beforeCount, out long beforeSize, out bool hasKeep);
+
+                if (!hasKeep)
+                {
+                    activity.RelatedWarning(this.CreateEventMetadata(), "Skiping pack maintenance due to no .keep file.");
+                    return;
+                }
+
+                GitProcess.Result expireResult = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot));
+                List<string> staleIdxFiles = this.CleanStaleIdxFiles(out int numDeletionBlocked);
+                this.GetPackFilesInfo(out int expireCount, out long expireSize, out hasKeep);
+                GitProcess.Result repackResult = this.RunGitCommand((process) => process.MultiPackIndexRepack(this.Context.Enlistment.GitObjectsRoot, this.batchSize));
+                this.GetPackFilesInfo(out int afterCount, out long afterSize, out hasKeep);
+
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
+                metadata.Add("BatchSize", this.batchSize);
+                metadata.Add(nameof(beforeCount), beforeCount);
+                metadata.Add(nameof(beforeSize), beforeSize);
+                metadata.Add(nameof(expireCount), expireCount);
+                metadata.Add(nameof(expireSize), expireSize);
+                metadata.Add(nameof(afterCount), afterCount);
+                metadata.Add(nameof(afterSize), afterSize);
+                metadata.Add("ExpireOutput", expireResult.Output);
+                metadata.Add("ExpireErrors", expireResult.Errors);
+                metadata.Add("RepackOutput", repackResult.Output);
+                metadata.Add("RepackErrors", repackResult.Errors);
+                metadata.Add("NumStaleIdxFiles", staleIdxFiles.Count);
+                metadata.Add("NumIdxDeletionsBlocked", numDeletionBlocked);
+                activity.RelatedEvent(EventLevel.Informational, $"{this.Area}_{nameof(this.PerformMaintenance)}", metadata, Keywords.Telemetry);
+
+                this.SaveLastRunTimeToFile();
+            }
+        }
+    }
+}
--- a/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
@@ -133,7 +133,7 @@ namespace GVFS.Common.Maintenance
 
                 if (!hasKeep)
                 {
-                    activity.RelatedWarning(this.CreateEventMetadata(), "Skiping pack maintenance due to no .keep file.");
+                    activity.RelatedWarning(this.CreateEventMetadata(), "Skipping pack maintenance due to no .keep file.");
                     return;
                 }
 
--- a/GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
@@ -327,25 +327,22 @@ namespace GVFS.Common.Maintenance
 
             string newKeepFile = Path.ChangeExtension(info.FullName, ".keep");
 
-            try
+            if (!this.Context.FileSystem.TryWriteAllText(newKeepFile, string.Empty))
             {
-                this.Context.FileSystem.WriteAllText(newKeepFile, string.Empty);
-
-                foreach (string keepFile in this.Context
-                                         .FileSystem
-                                         .ItemsInDirectory(this.Context.Enlistment.GitPackRoot)
-                                         .Where(item => item.Name.EndsWith(".keep"))
-                                         .Select(item => item.FullName))
-                {
-                    if (!keepFile.Equals(newKeepFile))
-                    {
-                        this.Context.FileSystem.TryDeleteFile(keepFile);
-                    }
-                }
+                this.Context.Tracer.RelatedWarning(this.CreateEventMetadata(), $"Failed to create .keep file at {newKeepFile}");
+                return;
             }
-            catch (IOException e)
+
+            foreach (string keepFile in this.Context
+                                     .FileSystem
+                                     .ItemsInDirectory(this.Context.Enlistment.GitPackRoot)
+                                     .Where(item => item.Name.EndsWith(".keep"))
+                                     .Select(item => item.FullName))
             {
-                this.Context.Tracer.RelatedWarning(this.CreateEventMetadata(e), "Exception while managing '.keep' files");
+                if (!keepFile.Equals(newKeepFile))
+                {
+                    this.Context.FileSystem.TryDeleteFile(keepFile);
+                }
             }
         }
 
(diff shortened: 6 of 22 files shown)


GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -177,6 +177,9 @@ public bool TryWriteTempFileAndRename(string destinationPath, string contents, o
             handledException = null;
             string tempFilePath = destinationPath + ".temp";
 
+            string parentPath = Path.GetDirectoryName(tempFilePath);
```
derrickstolee: @jeschu1: this is to fix problems when the parent folder doesn't exist. It seems like a good place to put it, since we generally expect the folder to exist before we try to create a file inside. It doesn't recurse to higher levels, though.

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
```diff
@@ -9,6 +9,10 @@ public class GitMaintenanceScheduler : IDisposable
     {
         private readonly TimeSpan looseObjectsDueTime = TimeSpan.FromMinutes(5);
         private readonly TimeSpan looseObjectsPeriod = TimeSpan.FromHours(6);
+
+        private readonly TimeSpan packfileDueTime = TimeSpan.FromMinutes(10);
+        private readonly TimeSpan packfilePeriod = TimeSpan.FromHours(12);
```
jeschu1: Should we pick something that will not collide with the 6 hour timer for looseObjects?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,91 @@
...
+        protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(1);
+
+        public void CountPackFiles(out int count, out long size)
+        {
+            count = 0;
+            size = 0;
+
+            foreach (DirectoryItemInfo info in this.Context.FileSystem.ItemsInDirectory(this.Context.Enlistment.GitPackRoot))
+            {
+                if (info.Name.EndsWith(".pack"))
```
jeschu1: string.Equals(Path.GetExtension(file), "pack", System.StringComparison.OrdinalIgnoreCase) ?

GVFS/GVFS.Common/InternalVerbParameters.cs
```diff
@@ -4,16 +4,22 @@ namespace GVFS.Common
 {
     public class InternalVerbParameters
     {
-        public InternalVerbParameters(string serviceName, bool startedByService, string maintenanceJob)
+        public InternalVerbParameters(
+            string serviceName = null,
+            bool startedByService = true, 
+            string maintenanceJob = null, 
+            string batchSize = null)
```
jeschu1: Can you explain why we want to pass batchsize as a parameter.  This is only for testing correct?  I've tried to not add extra parameters here where I can.

jeschu1: Is this because the default data isn't big enough to be repacked?

derrickstolee: correct.

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
derrickstolee: they will be 5 minutes off due to the `dueTime` difference, but also it's fine if they both get added to the queue (they won't be running at the same time).

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            this.Enlistment.PackfileMaintenanceStep(batchSize: $"{maxSize + 1}");
+
+            this.GetPackSizes(out packCount, out maxSize, out totalSize);
+            packCount.ShouldEqual(3);
+
+            // We should expire all packs except the one we just created
+            this.Enlistment.PackfileMaintenanceStep();
+            List<string> packsAfter = this.GetPackfiles();
+            packsAfter.Count.ShouldEqual(1);
+            packsAfter[0].StartsWith("pack-").ShouldBeTrue();
```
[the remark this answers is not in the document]
derrickstolee: (this is no longer a todo)

GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20181217.4</GitPackageVersion>
+    <GitPackageVersion>2.20181217.5-pr</GitPackageVersion>
```
derrickstolee: Replace with release build

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
wilbaker: > It doesn't recurse to higher levels, though.

FYI - `Directory.CreateDirectory` will create all missing folders in the specified path (even those at higher levels)

GVFS/GVFS.Common/InternalVerbParameters.cs
jeschu1: It looks like we always use DefaultBatchSize = "2g"; in the tests.  Can we instead of passing in another param, use 2g as a default in DehydrateVerb?

InternalVerbParameters have to be passed on  the command line since it's serizlied json. It's tough to write the more we have.  Perhaps this is an argument for not using serialized json and going back to comma seperated field/value pairs which I originally did and was asked to change.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            GitProcess.InvokeProcess(
+                this.Enlistment.RepoRoot,
+                $"multi-pack-index write --object-dir={this.GitObjectRoot}");
+
+            this.GetPackSizes(out int afterPrefetchPackCount, out maxSize, out totalSize);
+
+            // Cannot be sure, as the prefetch uses parallel threads to get multiple packs
+            afterPrefetchPackCount.ShouldBeAtLeast(2);
+
+            this.Enlistment.PackfileMaintenanceStep(batchSize: totalSize - 1);
```
derrickstolee: @jeschu1: we use the size parameter here to ensure we actually repack ("2g" would be too large for the test repo to ever meaningfully run the repack logic).

jeschu1: Got it if it's a moving scale we gotta pass it in

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,122 @@
...
+            List<DirectoryItemInfo> packDirContents = this.Context
+                                                          .FileSystem
+                                                          .ItemsInDirectory(this.Context.Enlistment.GitPackRoot)
+                                                          .ToList();
+
+            foreach (DirectoryItemInfo info in packDirContents)
+            {
+                if (string.Equals(Path.GetExtension(info.Name), ".idx", StringComparison.OrdinalIgnoreCase))
+                {
+                    string pairedPack = info.FullName.Substring(0, info.FullName.Length - "idx".Length) + "pack";
```
jeschu1: info.FullName.replace(".idx", ".pack")?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,122 @@
...
+
+                    if (!this.Context.FileSystem.FileExists(pairedPack))
+                    {
+                        if (this.Context.FileSystem.TryDeleteFile(info.FullName))
```
jeschu1: If we can't delete the file should we log a warning?

GVFS/GVFS.Common/Maintenance/GitMaintenanceScheduler.cs
```diff
@@ -64,6 +68,12 @@ private void ScheduleRecurringSteps()
                     state: null,
                     dueTime: this.looseObjectsDueTime,
                     period: this.looseObjectsPeriod));
+
+                this.stepTimers.Add(new Timer(
+                    (state) => this.queue.TryEnqueue(new PackfileMaintenanceStep(this.context)),
```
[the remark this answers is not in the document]
derrickstolee: Probably. Makes it easier to use.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: Better: `Path.ChangeExtension()`.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: We log the number of stale IDX files in the metadata below, which is tracked by telemetry.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,122 @@
...
+using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.Maintenance
+{
+    public class PackfileMaintenanceStep : GitMaintenanceStep
```
jeschu1: You have an exellent explaination at the top of the PR.  Could we add that in comments here

> This is done in two steps:
> 
> git multi-pack-index expire: This deletes the pack-files whose objects appear in newer pack-files. The multi-pack-index prevents git from looking at these packs. Rewrites the multi-pack-index to no longer refer to these (deleted) packs.
> 
> git multi-pack-index repack --batch-size=<size> inspects packs covered by the multi-pack-index in modified-time order (ascending). Greedily selects a batch of packs whose file sizes are all less than "size", but that sum up to at least "size". Then generate a new pack-file containing the objects that are uniquely referenced by the multi-pack-index.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
jeschu1: Could this fail due to a something in GVFS having a handle to it?  Reading your explaination on the PR.  If there are expected fail cases it would be helpful to add comments here.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,122 @@
...
+            this.forceRun = forceRun;
+            this.batchSize = batchSize;
+        }
+
+        public override string Area => nameof(PackfileMaintenanceStep);
+        protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", PackfileLastRunFileName);
+        protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(1);
+
+        // public only for unit tests
+        public void CountPackFiles(out int count, out long size)
```
jeschu1: tiny nit: Since it does more than count, maybe PackFilesInfo or similar

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,122 @@
...
+                    if (processIds.Any())
+                    {
+                        activity.RelatedWarning($"Skipping {nameof(LooseObjectsStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
+                        return;
+                    }
+                }
+
+                this.CountPackFiles(out int beforeCount, out long beforeSize);
+                GitProcess.Result result = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot));
+                List<string> staleIdxFiles = this.CleanStaleIdxFiles().ToList();
```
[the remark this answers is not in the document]
jeschu1: From the explaination on the header it seems like expire should have removed the idx's but may not have since they could have a handle open. Is that correct.  A line here explaining would be great.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,122 @@
...
+                this.CountPackFiles(out int afterCount, out long afterSize);
+
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
+                metadata.Add("BatchSize", this.batchSize);
+                metadata.Add(nameof(beforeCount), beforeCount);
+                metadata.Add(nameof(beforeSize), beforeSize);
+                metadata.Add(nameof(expireCount), expireCount);
+                metadata.Add(nameof(expireSize), expireSize);
+                metadata.Add(nameof(afterCount), afterCount);
```
jeschu1: Out of curiosity will the repack save any us any space.  I understand it will reduce the number of files just curious is there are any other gains.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            this.GetPackSizes(out packCount, out maxSize, out totalSize);
+
+            packCount.ShouldEqual(afterPrefetchPackCount + 1, $"incorrect number of packs after second step: {packCount}");
+
+            // We should expire all packs except the one we just created,
+            // and the prefetch pack which is marked as ".keep"
+            this.Enlistment.PackfileMaintenanceStep();
+            List<string> packsAfter = this.GetPackfiles();
+            packsAfter.Count.ShouldBeAtMost(afterPrefetchPackCount - 1, $"incorrect number of packs after third step: {packCount}");
```
jeschu1: Getting a bit lost in the math here.

Previous to the PackfileMaintenanceStep we had afterPretchPackCount +1, not we should have at most afterPrefectCount -1, i.e. two packs should be gone?  Can you explain?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: It will save space because we will not blindly copy every object out of the packs, but only those referred to by the multi-pack-index. After our first 'expire' command, there are usually a lot of packs that have less than 50 referenced objects (but actually have thousands of duplicate objects).

While we have a 2 GB batch size, we frequently will write a pack that is 200 MB instead.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
derrickstolee: This `ShouldBeAtMost()` statement is out-dated! Now that we are able to specify `totalSize - 1`, we can ensure that we create one pack that contains all of the other packed objects. BUT we keep the most-recent prefetch pack (marked with ".keep"). Thus, we should be able to enforce _exactly two_ packs.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                this.Enlistment.RepoRoot,
+                $"multi-pack-index write --object-dir={this.GitObjectRoot}");
+
+            List<string> packsBefore = this.GetPackfiles();
+            packsBefore.Count.ShouldEqual(packCount);
+
+            this.Enlistment.PackfileMaintenanceStep();
+
+            // Should expire the pack-<hash>.pack file.
+            List<string> packsBetween = this.GetPackfiles();
```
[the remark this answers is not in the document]
derrickstolee: In this case, they do. We don't have a cache server, so we get a dynamically-computed pack containing all commits and trees as a prefetch pack.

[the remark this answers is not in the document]
derrickstolee: The `pack-<hash>.pack` is computed at clone time (to construct the `.git/index` file), so it is older.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            packsBetween.Count.ShouldEqual(1, $"incorrect number of packs after first step: {packsBetween.Count}");
+
+            Path.GetFileName(packsBetween[0])
+                .StartsWith("prefetch-")
+                .ShouldBeTrue($"packsBetween[0] should start with 'prefetch-': {packsBetween[0]}");
+            string prefetchPack = packsBetween[0];
+
+            // Create new pack(s) by prefetching blobs
+            this.Enlistment.Prefetch($"--folders {Path.Combine("GVFS", "GVFS")}");
```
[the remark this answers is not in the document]
wilbaker: Oh maybe this is because we don't have a cache server?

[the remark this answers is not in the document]
derrickstolee: This triggers the `BlobPrefetcher` which asks the server for a number of packs corresponding to some blob ids. It is different than trying to hydrate the directory by reading those files and folders.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+            this.GetPackSizes(out int afterPrefetchPackCount, out maxSize, out totalSize);
+
+            // Cannot be sure, as the prefetch uses parallel threads to get multiple packs
+            afterPrefetchPackCount.ShouldBeAtLeast(2);
+
+            this.Enlistment.PackfileMaintenanceStep(batchSize: totalSize - 1);
+            this.GetPackSizes(out packCount, out maxSize, out totalSize);
+
+            packCount.ShouldEqual(afterPrefetchPackCount + 1, $"incorrect number of packs after second step: {packCount}");
```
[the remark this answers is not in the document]
derrickstolee: When we repacked, we added a pack-file and expired none.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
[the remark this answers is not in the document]
wilbaker: Thanks! I had forgotten that it was possible to get pack files in response to loose object prefetch requests.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,153 @@
...
+                metadata.Add(nameof(afterSize), afterSize);
+                metadata.Add("ExpireOutput", expireResult.Output);
+                metadata.Add("ExpireErrors", expireResult.Errors);
+                metadata.Add("RepackOutput", repackResult.Output);
+                metadata.Add("RepackErrors", repackResult.Errors);
+                metadata.Add("NumStaleIdxFiles", staleIdxFiles.Count);
+                metadata.Add("NumIdxDeletionsBlocked", numDeletionBlocked);
```
jeschu1: I like this change, good to know if idx files are hanging around

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -291,6 +297,53 @@ private bool TrySchedulePostFetchJob(List<string> packIndexes)
...
+
+            DirectoryItemInfo info = this.Context
+                                         .FileSystem
+                                         .ItemsInDirectory(this.Context.Enlistment.GitPackRoot)
+                                         .Where(item => item.Name.StartsWith(prefix)
+                                                        && string.Equals(Path.GetExtension(item.Name), ".pack", StringComparison.OrdinalIgnoreCase))
+                                         .FirstOrDefault();
+            if (info == null)
+            {
+                return;
```
jeschu1: Sound this be logged as a warning?

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -291,6 +297,53 @@ private bool TrySchedulePostFetchJob(List<string> packIndexes)
...
+        /// <summary>
+        /// Ensure the prefetch pack with most-recent timestamp has an associated
+        /// ".keep" file. This prevents any Git command from deleting the pack.
+        ///
+        /// Delete the previous ".keep" file so that pack can be deleted when they
+        /// are not the most-recent pack.
+        /// </summary>
+        private void UpdateKeepPacks()
+        {
+            if (!this.TryGetMaxGoodPrefetchTimestamp(out long maxGoodTimeStamp, out string error))
```
jeschu1: Do we only except this to fail if there are no prefect packs?

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -291,6 +297,53 @@ private bool TrySchedulePostFetchJob(List<string> packIndexes)
...
+                                     .ItemsInDirectory(this.Context.Enlistment.GitPackRoot)
+                                     .Where(item => item.Name.EndsWith(".keep"))
+                                     .Select(item => item.FullName)
+                                     .FirstOrDefault();
+
+            string newKeepFile = Path.ChangeExtension(info.FullName, ".keep");
+
+            if (oldKeepFile == null || !newKeepFile.Equals(oldKeepFile))
+            {
+                this.Context.FileSystem.WriteAllText(newKeepFile, string.Empty);
```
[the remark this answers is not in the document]
jeschu1: Same with DeleteFile.  I noticed we have a TryDeleteFile that wraps exceptions.  Could consider adding a TryWriteAllText.  Putting it al in try / catch works too.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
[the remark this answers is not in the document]
jeschu1: It would be great to add as a comment

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PackfileMaintenanceStepTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            return this.fileSystem
+                       .EnumerateDirectory(this.PackRoot)
+                       .Split();
+        }
+
+        private List<string> GetPackfiles()
+        {
+            return this.GetPackDirContents()
+                       .Where(file => file.EndsWith(".pack"))
+                       .ToList();
```
jeschu1: return Directory.GetFiles(this.PackRoot, "*.pack")

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -291,6 +297,53 @@ private bool TrySchedulePostFetchJob(List<string> packIndexes)
...
         }
 
+        /// <summary>
+        /// Ensure the prefetch pack with most-recent timestamp has an associated
+        /// ".keep" file. This prevents any Git command from deleting the pack.
+        ///
+        /// Delete the previous ".keep" file so that pack can be deleted when they
+        /// are not the most-recent pack.
+        /// </summary>
+        private void UpdateKeepPacks()
```
jeschu1: Can we add a test for this new functionality.  Would it be an easy extention maybe of existing Prefetch functional tests?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -140,6 +147,8 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
                 { "merge.renames", "false" },
                 { "pack.useBitmaps", "false" },
                 { "pack.useSparse", "true" },
+                { "pack.depth", "0" },
+                { "pack.window", "0" },
```
jeschu1: If we do this change we can remove the additions I did in GitProcess.cs
$"pack-objects {packFilePath} --non-empty **--window=0 --depth=0** -q",

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -71,6 +76,8 @@ public string InternalParameters
 
         public string MaintenanceJob { get; set; }
 
+        public string BatchSize { get; set; }
```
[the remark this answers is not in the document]
jeschu1: I don't how far from convention this would be but PackFileMaintenance_BatchSize would be easier for me to read.

GVFS/GVFS/CommandLine/GVFSVerb.cs
derrickstolee: We can. There just wasn't a way to do it (currently) through the `multi-pack-index repack` command. It would require wiring up a lot of complexity that isn't worth it.

derrickstolee: **ACTUALLY...** this also shuts down delta calculation during `git push`. That's _not_ something we should do, as the server doesn't redeltify. We rely on clients sending decent deltas or the total space will blow up.

I'm going to see if running the process with `-c pack.depth=0 -c pack.window=0` carries those config settings through the `run_command()` method.

derrickstolee: (I noticed this because the `git pack-objects` command in the perf suite sped up.)

jeschu1: Nice Catch!

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -56,6 +60,8 @@ public bool TryPrefetchCommitsAndTrees(out string error, GitProcess gitProcess =
                     error = "Failed to download prefetch packs";
```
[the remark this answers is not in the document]
derrickstolee: I'll leave it as only under success, but we also will block the packfile maintenance until after we have a `.keep` file.

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
derrickstolee: There are a number of reasons why it can fail, it's pretty complicated. No prefetch packs is the most-typical case, but also if someone calls `Stop()`, it will terminate early.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -0,0 +1,167 @@
...
+                        activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
+                        return;
+                    }
+                }
+
+                this.GetPackFilesInfo(out int beforeCount, out long beforeSize, out bool hasKeep);
+
+                if (!hasKeep)
+                {
+                    activity.RelatedWarning(this.CreateEventMetadata(), "Skiping pack maintenance due to no .keep file.");
```
wilbaker: Typo: `skiping` -> `skipping`

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -291,6 +297,58 @@ private bool TrySchedulePostFetchJob(List<string> packIndexes)
...
+                                         .Where(item => item.Name.EndsWith(".keep"))
+                                         .Select(item => item.FullName))
+                {
+                    if (!keepFile.Equals(newKeepFile))
+                    {
+                        this.Context.FileSystem.TryDeleteFile(keepFile);
+                    }
+                }
+            }
+            catch (IOException e)
```
wilbaker: I noticed that `WriteAllFile` can also throw `UnauthorizedAccessException` and `SecurityException`, and there are other spots we're catching those exception types in the code.

It seems unlikely we'd hit them here (based on their description), but wanted to see if you thought it'd be worth catching them here as well.

wilbaker: *Edit:* I meant to say `WriteAllText` above

wilbaker: The general approach I've seen taken in the past is:

- Catch exceptions that we anticipate could happen at runtime that are not indicative of a code bug 
- Log exceptions and exit when there's:
    - An otherwise unhandled exception 
    - An exception that indicates that we probably have a bug in the code.  For example, `ArgumentNullException` or `NotSupportedException` could be prevented with code changes, but `SecurityException` probably could not.

I realize this is a bit fuzzy, as some exceptions (e.g. `DirectoryNotFoundException` might not fall cleaning into one of these categories).

I mentioned these three specifically because we've hit them at various times in the past and I'd imagine that if we were to hit them here it'd be better to fail the job than bring down the mount process.

> If we want to catch those three exactly, then I can instead go and build `TryWriteAllText()` to include that logic instead.

I think either approach (`TryWriteAllText()` or catching the exceptions) is good, so I'd say go with whichever you prefer.

PR #636, opened 2018-12-18
LooseObjectsStep (Part 2)

jeschu1:
- Pack loose objects into a Packfile

Update to let Stolee know how things are looking.  Don't worry about a review yet everyone else.

Next steps for me are to add the unit tests and make sure I'm happy with all the logging.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -373,6 +373,16 @@ namespace GVFS.Common.Git
                 null);
         }
 
+        public Result PackObjects(string filenamePrefix, string gitObjectsDirectory, Action<StreamWriter> packFileStream)
+        {
+            string packFilePath = Path.Combine(gitObjectsDirectory, "pack", filenamePrefix);
+            return this.InvokeGitAgainstDotGitFolder(
+                $"pack-objects {packFilePath} --non-empty",
+                packFileStream,
+                parseStdOutLine: null,
+                gitObjectsDirectory: gitObjectsDirectory);
+        }
+
         /// <summary>
         /// Write a new commit graph in the specified pack directory. Crawl the given pack-
         /// indexes for commits and then close under everything reachable or exists in the
@@ -596,6 +606,7 @@ namespace GVFS.Common.Git
                         }
 
                         writeStdIn?.Invoke(this.executingProcess.StandardInput);
+                        this.executingProcess.StandardInput.Close();
 
                         this.executingProcess.BeginOutputReadLine();
                         this.executingProcess.BeginErrorReadLine();
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -12,7 +12,8 @@ namespace GVFS.Common.Maintenance
     public class LooseObjectsStep : GitMaintenanceStep
     {
         public const string LooseObjectsLastRunFileName = "loose-objects.time";
-        private readonly bool forceRun;
+        public int LooseObjectsPutIntoPackFile = 0;
+        private readonly bool forceRun;        
 
         public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRun = false)
             : base(context, gitObjects: null, requireObjectCacheLock: requireCacheLock)
@@ -22,24 +23,61 @@ namespace GVFS.Common.Maintenance
 
         public override string Area => nameof(LooseObjectsStep);
 
+        public virtual int MaxLooseObjectsInPack => 1000;
+
         protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsLastRunFileName);
         protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(7);
 
         public int CountLooseObjects()
         {
             int count = 0;
-            
+
             foreach (string directoryPath in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
             {
                 string directoryName = directoryPath.TrimEnd(Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar).Last();
 
                 if (GitObjects.IsLooseObjectsDirectory(directoryName))
-                {                    
+                {
                     count += this.Context.FileSystem.GetFiles(Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath), "*").Count();
                 }
             }
 
             return count;
+        }
+
+        public void PackLooseObjects(StreamWriter streamWriter)
+        {
+            // Find looseObjects
+            foreach (string directoryPath in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
+            {
+                string directoryName = directoryPath.TrimEnd(Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar).Last();
+
+                if (GitObjects.IsLooseObjectsDirectory(directoryName))
+                {
+                    string[] looseObjectFileNamesInDir = this.Context.FileSystem.GetFiles(Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath), "*");
+
+                    foreach (string file in looseObjectFileNamesInDir)
+                    {
+                        if (this.LooseObjectsPutIntoPackFile == this.MaxLooseObjectsInPack)
+                        {
+                            return;
+                        }
+
+                        this.LooseObjectsPutIntoPackFile++;
+                        streamWriter.Write(this.GetLooseObjectGuid(directoryName, file) + "\n");
+                    }
+                }
+            }
+        }
+
+        public string GetLooseObjectGuid(string directoryName, string filePath)
+        {
+            return directoryName + Path.GetFileName(filePath);
+        }
+
+        public void CreateLooseObjectsPackFile()
+        {
+            GitProcess.Result result = this.RunGitCommand((process) => process.PackObjects("from-loose", this.Context.Enlistment.GitObjectsRoot, this.PackLooseObjects));
         }
 
         protected override void PerformMaintenance()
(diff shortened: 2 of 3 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -373,6 +373,16 @@ public Result UnpackObjects(Stream packFileStream)
                 null);
         }
 
+        public Result PackObjects(string filenamePrefix, string gitObjectsDirectory, Action<StreamWriter> packFileStream)
+        {
+            string packFilePath = Path.Combine(gitObjectsDirectory, "pack", filenamePrefix);
+            return this.InvokeGitAgainstDotGitFolder(
+                $"pack-objects {packFilePath} --non-empty",
```
derrickstolee: We may want a `-q` here to avoid progress output. But it depends on what the logs look like.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -12,7 +12,8 @@ namespace GVFS.Common.Maintenance
     public class LooseObjectsStep : GitMaintenanceStep
     {
         public const string LooseObjectsLastRunFileName = "loose-objects.time";
-        private readonly bool forceRun;
+        public int LooseObjectsPutIntoPackFile = 0;
```
derrickstolee: `{ get; private set; }`. Initializes to zero anyway, but want to avoid public setters.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -22,26 +23,63 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
 
         public override string Area => nameof(LooseObjectsStep);
 
+        public virtual int MaxLooseObjectsInPack => 1000;
```
derrickstolee: Some math to do here. How long does it take to pack 1000 objects? 10,000? 100,000? If we are running this only once a week, then we probably want this to be _at least_ 100,000 due to over-hydrating causes 2.1 million blob downloads. Unless there is a reason why that would be too many objects.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -73,6 +98,12 @@ private void DeleteFiles(List<string> filePaths)
             }
         }
 
+        private int CountPackFiles()
+        {
+            string[] files = Directory.GetFiles(this.PackRoot);
+            return files.Where(file => string.Equals(Path.GetExtension(file), ".pack", System.StringComparison.OrdinalIgnoreCase)).Count();
```
derrickstolee: split LINQ lines across multiple lines. No need to have the `files` variable, as we can chain LINQ:

```
return Directory.GetFiles(this.PackRoot)
                 .Where(file => string.Equals(Path.GetExtension(file), ".pack", System.StringComparison.OrdinalIgnoreCase))
                 .Count();
```

derrickstolee: and use `this.Context.FileSystem` instead of `Directory` directly?

jeschu1: Cool on the linq, since we're in a functional test I don't believe we have this.Context

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -596,6 +606,7 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
                         }
 
                         writeStdIn?.Invoke(this.executingProcess.StandardInput);
+                        this.executingProcess.StandardInput.Close();
```
jrbriggs: Why is this change necessary?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jrbriggs: Agreed. Also consider the size of a typical hourly prefetch pack or a daily prefetch pack. How many objects do we see there?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -22,26 +23,63 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
 
             return count;
         }
 
+        public void PackLooseObjects(StreamWriter streamWriter)
+        {
+            // Find looseObjects
+            foreach (string directoryPath in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
+            {
+                string directoryName = directoryPath.TrimEnd(Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar).Last();
```
jrbriggs: `System.IO.Directory.GetParent()`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -22,26 +23,63 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+
+                if (GitObjects.IsLooseObjectsDirectory(directoryName))
+                {
+                    string[] looseObjectFileNamesInDir = this.Context.FileSystem.GetFiles(Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath), "*");
+
+                    foreach (string file in looseObjectFileNamesInDir)
+                    {
+                        if (this.LooseObjectsPutIntoPackFile == this.MaxLooseObjectsInPack)
```
jrbriggs: Maybe test >= rather than exact equality, in case we decide one day we'll add in batches.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -22,26 +23,63 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+                        }
+
+                        this.LooseObjectsPutIntoPackFile++;
+                        streamWriter.Write(this.GetLooseObjectGuid(directoryName, file) + "\n");
+                    }
+                }
+            }
+        }
+
+        public string GetLooseObjectGuid(string directoryName, string filePath)
```
jrbriggs: This is an oid, not a guid, right?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -22,26 +23,63 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+        {
+            return directoryName + Path.GetFileName(filePath);
+        }
+
+        public void CreateLooseObjectsPackFile()
+        {
+            GitProcess.Result result = this.RunGitCommand((process) => process.PackObjects("from-loose", this.Context.Enlistment.GitObjectsRoot, this.PackLooseObjects));
```
jrbriggs: This could use some newlines.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -65,15 +103,18 @@ protected override void PerformMaintenance()
...
 
                     EventMetadata metadata = new EventMetadata();
                     metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
-                    metadata.Add("StartingCount", beforeCount);
-                    metadata.Add("EndingCount", afterCount);
-                    metadata.Add("RemovedCount", beforeCount - afterCount);
+                    metadata.Add("StartingCount", beforeLooseObjectsCount);
+                    metadata.Add("EndingCount", afterLooseObjectsCount);
+                    metadata.Add("RemovedCount", beforeLooseObjectsCount - afterLooseObjectsCount);
+                    metadata.Add(nameof(this.LooseObjectsPutIntoPackFile), this.LooseObjectsPutIntoPackFile);
```
jrbriggs: nit: let's not make this nameof here, so that refactoring doesn't change what we look for in the data exhaust streams.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: Updating to just Id as per @derrickstolee  that is most correct

GVFS/GVFS.Common/Git/GitProcess.cs
derrickstolee: `git pack-objects` listens to stdin until receiving an EOF. This is different from the other commands we've used in the past, which cared about an empty line.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: Keep in mind that the daily/hourly prefetch packs contain commits and trees, and we expect this step to create packs from blobs. The blobs will be much bigger.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: or rather, `Path.GetDirectory()`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: OID.

derrickstolee: right, ObjectId

GVFS/GVFS.Common/Git/GitProcess.cs
jrbriggs: Got it. It's a harmless change for other commands but necessary for `git pack-objects`. Should we have a comment on why the line is needed?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jrbriggs: Fair point. Either way I think 1000 is too small by an order of magnitude or two.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: Are you suggesting instead of Split(Path.DirectorySeparatorChar).Last(), substring using GetParent()?

jeschu1: If we do a substring @derrickstolee had the suggestion of 
directoryPath.TrimEnd(Path.DirectorySeparatorChar).Substring(directoryPath.LastIndexOf(separator) + 1)

Shame there isn't a builtin function to do this but I'm not really bothered with which permuation we go with.

jeschu1: @derrickstolee I think I can use that to simplify

jeschu1:
Sorry guys it looks like i made the original branch here instead of my fork.  I'll delete the branch in the MIcrosoft project and am opening a new PR.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: Went with 100,000 for now

PR #639, opened 2018-12-18
Fix VS IDE Builds post FastFetch and Versioning rework

### LINKED ISSUE #634: VFS4G builds failing in VS IDE on Windows
Sample error:

> error MSB4126: The specified solution configuration "|x64" is invalid. Please specify a valid solution configuration using the Configuration and Platform properties (e.g. MSBuild.exe Solution.sln /p:Configuration=Debug /p:Platform="Any CPU") or leave those properties blank to use the default solution configuration. [C:\github\VFSForGit\GVFS.sln]
1>C:\github\VFSForGit\GVFS\GVFS.Build\GVFS.PreBuild.csproj(185,5): error MSB3073: The command "call C:\github\VFSForGit\GVFS\GVFS.Build\..\..\Scripts\RestorePackages.bat" exited with code 1.

nickgra:
Fixes #634 

RestorePackages.bat now requires the Configuration to be passed into it for `dotnet restore` to be able to restore. For some reason, this works fine in the build script, but it doesn't work implicitly in the IDE. Pass the Configuration through from PreBuild into the script.

Additionally, now that the Versioning refactor is done, GVFS.Hooks.Mac has a dependency on PreBuild to get the CommonAssemblyVersion so we need to add a project dependency there.

### DIFF
--- a/Scripts/RestorePackages.bat
+++ b/Scripts/RestorePackages.bat
@@ -3,6 +3,10 @@ CALL %~dp0\InitializeEnvironment.bat || EXIT /b 10
 
 SETLOCAL
 
+IF "%1"=="" (SET "Configuration=Debug") ELSE (SET "Configuration=%1")
+ 
+SET SolutionConfiguration=%Configuration%.Windows
+
 SET nuget="%VFS_TOOLSDIR%\nuget.exe"
 IF NOT EXIST %nuget% (
   mkdir %nuget%\..
--- a/GVFS.sln
+++ b/GVFS.sln
@@ -180,6 +180,9 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GVFS.FunctionalTests.LockHolder", "GVFS\GVFS.FunctionalTests.LockHolder\GVFS.FunctionalTests.LockHolder.csproj", "{FA273F69-5762-43D8-AEA1-B4F08090D624}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GVFS.Hooks.Mac", "GVFS\GVFS.Hooks\GVFS.Hooks.Mac.csproj", "{4CC2A90D-D240-4382-B4BF-5E175515E492}"
+	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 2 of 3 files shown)


Scripts/RestorePackages.bat
```diff
@@ -3,6 +3,10 @@ CALL %~dp0\InitializeEnvironment.bat || EXIT /b 10
 
 SETLOCAL
 
+IF "%1"=="" (SET "Configuration=Debug") ELSE (SET "Configuration=%1")
```
wilbaker: I think this change looks good, but out of curiosity does the configuration matter when restoring nuget packages?  Or could we always restore the Release configuration?

nickgra: When I change the call in RestorePackages.bat to be this:
`dotnet restore %VFS_SRCDIR%\GVFS.sln /p:Configuration=%SolutionConfiguration% /p:VCTargetsPath="C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140" --packages %VFS_PACKAGESDIR% || exit /b 1`

I get this error:
> 1>C:\Users\nickgra\VFSForGit\GVFS.sln.metaproj : error MSB4126: The specified solution configuration "Debug|x64" is invalid. Please specify a valid solution configuration using the Configuration and Platform properties (e.g. MSBuild.exe Solution.sln /p:Configuration=Debug /p:Platform="Any CPU") or leave those properties blank to use the default solution configuration. [C:\Users\nickgra\VFSForGit\GVFS.sln]
1>C:\Users\nickgra\VFSForGit\GVFS\GVFS.Build\GVFS.PreBuild.csproj(185,5): error MSB3073: The command "call C:\Users\nickgra\VFSForGit\GVFS\GVFS.Build\..\..\Scripts\RestorePackages.bat Debug" exited with code 1.
3>------ Build started: Project: GVFS.GVFlt, Configuration: Debug x64 ------

This makes sense according to the docs, nothing implies a default of Debug and we don't have Debug as a configuration in our project.

The configuration doesn't matter here for what packages get restored (at least in our projects), but the Configuration does matter for what asset/metadata files get generated in BuildOutput\<Project Name>\obj\[Platform]\[Configuration]. We have to restore for whatever configuration we're building, we could restore both in this script (which would add more time to first time builds) or we could pass the platform in.

This entire script can be removed whenever we move to .NET Core completely so I think this is good enough for now.

PR #640, opened 2018-12-18
LooseObjectsStep (Part 2)

jeschu1:
- Pack loose objects into a Packfile

This repaces #636 . I accidentally made the branch not in my fork.  I'll leave old PR / branch up for 2 days and than delete the branch to cleanup.

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
@@ -177,6 +177,9 @@ namespace GVFS.Common.FileSystem
             handledException = null;
             string tempFilePath = destinationPath + ".temp";
 
+            string parentPath = Path.GetDirectoryName(tempFilePath);
+            this.CreateDirectory(parentPath);
+
             try
             {
                 using (Stream tempFile = this.OpenFileStream(tempFilePath, FileMode.Create, FileAccess.Write, FileShare.None, callFlushFileBuffers: true))
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -25,10 +25,12 @@ namespace GVFS.Common.Maintenance
 
         public override string Area => nameof(LooseObjectsStep);
         public int LooseObjectsPutIntoPackFile { get; private set; } = 0;
-        public int MaxLooseObjectsInPack { get; set; } = 100000;
+        
+        // 50,000 was found to be the optimal time taking ~5 minutes
+        public int MaxLooseObjectsInPack { get; set; } = 50000;
 
         protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsLastRunFileName);
-        protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(7);
+        protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(1);
 
         public int CountLooseObjects()
         {
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -49,7 +49,7 @@ namespace GVFS.Common.Maintenance
             return count;
         }
 
-        public void PackLooseObjects(StreamWriter streamWriter)
+        public void WriteLooseObjectIds(StreamWriter streamWriter)
         {
             this.LooseObjectsPutIntoPackFile = 0;
 
@@ -103,7 +103,7 @@ namespace GVFS.Common.Maintenance
                 (process) => process.PackObjects(
                     "from-loose",
                     this.Context.Enlistment.GitObjectsRoot,
-                    this.PackLooseObjects));
+                    this.WriteLooseObjectIds));
         }
 
         protected override void PerformMaintenance()
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -24,7 +24,6 @@ namespace GVFS.Common.Maintenance
         }
 
         public override string Area => nameof(LooseObjectsStep);
-        public int LooseObjectsPutIntoPackFile { get; private set; } = 0;
         
         // 50,000 was found to be the optimal time taking ~5 minutes
         public int MaxLooseObjectsInPack { get; set; } = 50000;
@@ -49,9 +48,14 @@ namespace GVFS.Common.Maintenance
             return count;
         }
 
-        public void WriteLooseObjectIds(StreamWriter streamWriter)
+        /// <summary>
+        /// Writes loose object Ids to streamWriter
+        /// </summary>
+        /// <param name="streamWriter">Writer to which SHAs are written</param>
+        /// <returns>The number of loose objects SHAs written to the stream</returns>
+        public int WriteLooseObjectIds(StreamWriter streamWriter)
         {
-            this.LooseObjectsPutIntoPackFile = 0;
+            int looseObjectsPutIntoPackFiles = 0;
 
             // Find loose Objects
             foreach (DirectoryItemInfo directoryItemInfo in this.Context.FileSystem.ItemsInDirectory(this.Context.Enlistment.GitObjectsRoot))
@@ -72,18 +76,20 @@ namespace GVFS.Common.Maintenance
                                 continue;
                             }
 
-                            this.LooseObjectsPutIntoPackFile++;
+                            looseObjectsPutIntoPackFiles++;
                             streamWriter.Write(objectId + "\n");
 
                             // Stop when we hit the limit.  The next run will pack more files.
-                            if (this.LooseObjectsPutIntoPackFile >= this.MaxLooseObjectsInPack)
+                            if (looseObjectsPutIntoPackFiles >= this.MaxLooseObjectsInPack)
                             {
-                                return;
+                                return looseObjectsPutIntoPackFiles;
                             }
                         }
                     }
                 }
             }
+
+            return looseObjectsPutIntoPackFiles;
         }
 
         public bool TryGetLooseObjectId(string directoryName, string filePath, out string objectId)
@@ -97,13 +103,21 @@ namespace GVFS.Common.Maintenance
             return true;
         }
 
-        public void CreateLooseObjectsPackFile()
+        /// <summary>
+        /// Creates a pack file from loose objects
+        /// </summary>
+        /// <returns>The number of loose objects added to the pack file</returns>
+        public int CreateLooseObjectsPackFile()
         {
+            int objectsAddedToPack = 0;
+
             GitProcess.Result result = this.RunGitCommand(
                 (process) => process.PackObjects(
                     "from-loose",
                     this.Context.Enlistment.GitObjectsRoot,
-                    this.WriteLooseObjectIds));
+                    (StreamWriter writer) => objectsAddedToPack = this.WriteLooseObjectIds(writer)));
+
+            return objectsAddedToPack;
         }
 
         protected override void PerformMaintenance()
@@ -133,14 +147,14 @@ namespace GVFS.Common.Maintenance
                     GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot));
                     int afterLooseObjectsCount = this.CountLooseObjects();
 
-                    this.CreateLooseObjectsPackFile();
+                    int objectsAddedToPack = this.CreateLooseObjectsPackFile();
 
                     EventMetadata metadata = new EventMetadata();
                     metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
                     metadata.Add("StartingCount", beforeLooseObjectsCount);
                     metadata.Add("EndingCount", afterLooseObjectsCount);
                     metadata.Add("RemovedCount", beforeLooseObjectsCount - afterLooseObjectsCount);
-                    metadata.Add("LooseObjectsPutIntoPackFile", this.LooseObjectsPutIntoPackFile);
+                    metadata.Add("LooseObjectsPutIntoPackFile", objectsAddedToPack);
                     activity.RelatedEvent(EventLevel.Informational, $"{this.Area}_{nameof(this.PerformMaintenance)}", metadata, Keywords.Telemetry);
                     this.SaveLastRunTimeToFile();
                 }
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -1,4 +1,5 @@
-﻿using GVFS.Common.Git;
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
 using GVFS.Common.Tracing;
 using System;
 using System.Collections.Generic;
@@ -8,7 +9,9 @@ using System.Linq;
 
 namespace GVFS.Common.Maintenance
 {
-    // Remove loose objects that appear in packfiles
+    // Performs LooseObject Maintenace
+    // 1. Removes loose objects that appear in packfiles
+    // 2. Packs loose objects into a packfile
     public class LooseObjectsStep : GitMaintenanceStep
     {
         public const string LooseObjectsLastRunFileName = "loose-objects.time";
@@ -21,6 +24,8 @@ namespace GVFS.Common.Maintenance
         }
 
         public override string Area => nameof(LooseObjectsStep);
+        public int LooseObjectsPutIntoPackFile { get; private set; } = 0;
+        public int MaxLooseObjectsInPack { get; set; } = 100000;
 
         protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsLastRunFileName);
         protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(7);
@@ -28,18 +33,61 @@ namespace GVFS.Common.Maintenance
         public int CountLooseObjects()
         {
             int count = 0;
-            
+
             foreach (string directoryPath in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
             {
                 string directoryName = directoryPath.TrimEnd(Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar).Last();
 
                 if (GitObjects.IsLooseObjectsDirectory(directoryName))
-                {                    
+                {
                     count += this.Context.FileSystem.GetFiles(Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath), "*").Count();
                 }
             }
 
             return count;
+        }
+
+        public void PackLooseObjects(StreamWriter streamWriter)
+        {
+            // Find loose Objects
+            foreach (DirectoryItemInfo directoryItemInfo in this.Context.FileSystem.ItemsInDirectory(this.Context.Enlistment.GitObjectsRoot))
+            {
+                if (directoryItemInfo.IsDirectory)
+                {
+                    string directoryName = directoryItemInfo.Name;
+
+                    if (GitObjects.IsLooseObjectsDirectory(directoryName))
+                    {
+                        string[] looseObjectFileNamesInDir = this.Context.FileSystem.GetFiles(Path.Combine(directoryItemInfo.FullName), "*");
+
+                        foreach (string file in looseObjectFileNamesInDir)
+                        {
+                            this.LooseObjectsPutIntoPackFile++;
+                            streamWriter.Write(this.GetLooseObjectId(directoryName, file) + "\n");
+
+                            // Stop when we hit the limit.  The next run will pack more files.
+                            if (this.LooseObjectsPutIntoPackFile == this.MaxLooseObjectsInPack)
+                            {
+                                return;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        public string GetLooseObjectId(string directoryName, string filePath)
+        {
+            return directoryName + Path.GetFileName(filePath);
+        }
+
+        public void CreateLooseObjectsPackFile()
+        {
+            GitProcess.Result result = this.RunGitCommand(
+                (process) => process.PackObjects(
+                    "from-loose",
+                    this.Context.Enlistment.GitObjectsRoot,
+                    this.PackLooseObjects));
         }
 
         protected override void PerformMaintenance()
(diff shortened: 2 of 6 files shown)


GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,72 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+                    {
+                        string[] looseObjectFileNamesInDir = this.Context.FileSystem.GetFiles(Path.Combine(directoryItemInfo.FullName), "*");
+
+                        foreach (string file in looseObjectFileNamesInDir)
+                        {
+                            this.LooseObjectsPutIntoPackFile++;
+                            streamWriter.Write(this.GetLooseObjectId(directoryName, file) + "\n");
+
+                            // Stop when we hit the limit.  The next run will pack more files.
+                            if (this.LooseObjectsPutIntoPackFile == this.MaxLooseObjectsInPack)
```
jeschu1: @jrbriggs you had a question here about changes this to <= for batches in teh future.  Not sure what you mean.

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -17,6 +17,7 @@ namespace GVFS.UnitTests.Maintenance
     public class LooseObjectStepTests
     {
         private const string PrunePackedCommand = "prune-packed -q";
+        private const string PackCommand = @"pack-objects mock:\path\.git\objects\pack\from-loose --non-empty -q";
         private MockTracer tracer;
```
jeschu1: The \'s won't work on apple.  Thanks unit tests!  I'll update tomorrow.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -65,15 +113,18 @@ protected override void PerformMaintenance()
...
+                    int afterLooseObjectsCount = this.CountLooseObjects();
+
+                    this.CreateLooseObjectsPackFile();
 
                     EventMetadata metadata = new EventMetadata();
                     metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
-                    metadata.Add("StartingCount", beforeCount);
-                    metadata.Add("EndingCount", afterCount);
-                    metadata.Add("RemovedCount", beforeCount - afterCount);
+                    metadata.Add("StartingCount", beforeLooseObjectsCount);
```
wilbaker: Should we also include a message in the metadata (key `TracingConstants.MessageKey.InfoMessage`) describing this event?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -65,15 +113,18 @@ protected override void PerformMaintenance()
...
+                    metadata.Add("EndingCount", afterLooseObjectsCount);
+                    metadata.Add("RemovedCount", beforeLooseObjectsCount - afterLooseObjectsCount);
+                    metadata.Add("LooseObjectsPutIntoPackFile", this.LooseObjectsPutIntoPackFile);
                     activity.RelatedEvent(EventLevel.Informational, this.Area, metadata, Keywords.Telemetry);
```
wilbaker: Minor (and not strictly related to these changes), if you make the `eventName` parameter of `RelatedEvent` more unique it might be easier to filter.

*Example:*

Instead of:
`this.Area`

Something like:
`$"{this.Area}_{nameof(this.PerformMaintenance)}"`

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,72 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+            return directoryName + Path.GetFileName(filePath);
+        }
+
+        public void CreateLooseObjectsPackFile()
+        {
+            GitProcess.Result result = this.RunGitCommand(
+                (process) => process.PackObjects(
+                    "from-loose",
+                    this.Context.Enlistment.GitObjectsRoot,
+                    this.PackLooseObjects));
```
wilbaker: What will happen if `PackLooseObjects` finds no loose objects?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,72 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+        public string GetLooseObjectId(string directoryName, string filePath)
+        {
+            return directoryName + Path.GetFileName(filePath);
+        }
+
+        public void CreateLooseObjectsPackFile()
+        {
+            GitProcess.Result result = this.RunGitCommand(
+                (process) => process.PackObjects(
+                    "from-loose",
```
wilbaker: What will happen if one (or more) of the loose objects is corrupt?  Is there anyway that we can detect that here and delete the corrupt object and try again?

(I would guess this would be a rare event based on how VFS4G downloads loose objects, but it would still be good to know what will happen)

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -61,21 +63,51 @@ public void LooseObjectsPassTimeRestriction()
...
             commands[0].ShouldEqual(PrunePackedCommand);
+            commands[1].ShouldEqual(PackCommand);
+        }
+
+        [TestCase]
+        public void LooseObjectsLimitPackCount()
+        {
+            this.TestSetup(DateTime.UtcNow.AddDays(-7));
+
+            // Verify no limit
```
wilbaker: Should this comment be updated to something like `Verify with default limit`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jrbriggs: I prefer to be safe with these stop conditions. If we ever change to adding multiple objects at once, we could skip over `this.MaxLooseObjectsInPack` entirely and pack all loose objects in the cache into one pack. Changing the stop condition to `this.LooseObjectsPutIntoPackFile >= this.MaxLooseObjectsInPack` prevents that. I think I used the wrong operator in the last comment, my apologies if that was the source of confusion.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: PackLooseObjects calls  $"pack-objects {packFilePath} --non-empty -q".  This means if there are no loose objects don't create a pack.  This case has been tested.  Would be a good idea to add a function test for it.  I'll make it happen!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: the git operation would complain.  I'll update the writer to skip correct objects.  This should never happen but it doesn't hurt to be proactive against errors.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -65,15 +113,18 @@ protected override void PerformMaintenance()
...
 
                     EventMetadata metadata = new EventMetadata();
                     metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
-                    metadata.Add("StartingCount", beforeCount);
-                    metadata.Add("EndingCount", afterCount);
-                    metadata.Add("RemovedCount", beforeCount - afterCount);
+                    metadata.Add("StartingCount", beforeLooseObjectsCount);
+                    metadata.Add("EndingCount", afterLooseObjectsCount);
+                    metadata.Add("RemovedCount", beforeLooseObjectsCount - afterLooseObjectsCount);
+                    metadata.Add("LooseObjectsPutIntoPackFile", this.LooseObjectsPutIntoPackFile);
```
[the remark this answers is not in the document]
jeschu1: Now that is an exellent find!  This would cause a problem for sure.  I'll add the reset.  
It's a member variable so can unit test it.  If you see an easier way to do that let me know!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jeschu1: I think with your great suggestion below to change the name eventName a InfoMessage isn't needed cause it should be self describing.  Let me know if you disagree!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,72 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+                    string directoryName = directoryItemInfo.Name;
+
+                    if (GitObjects.IsLooseObjectsDirectory(directoryName))
+                    {
+                        string[] looseObjectFileNamesInDir = this.Context.FileSystem.GetFiles(Path.Combine(directoryItemInfo.FullName), "*");
+
+                        foreach (string file in looseObjectFileNamesInDir)
+                        {
+                            this.LooseObjectsPutIntoPackFile++;
+                            streamWriter.Write(this.GetLooseObjectId(directoryName, file) + "\n");
```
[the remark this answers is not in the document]
jeschu1: Validation of SHA and SkipInvalidLooseObjects test added.

jeschu1:
Thanks for your great feedback @wilbaker !  I'm hoping to wrap this PR up today if possible since tomorrow is my last day in office of the year.  Let me know what you think of the changes :-)

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -177,6 +177,9 @@ public bool TryWriteTempFileAndRename(string destinationPath, string contents, o
             handledException = null;
             string tempFilePath = destinationPath + ".temp";
 
+            string parentPath = Path.GetDirectoryName(tempFilePath);
```
jeschu1: @derrickstolee I ported your changes over here since I was spewing out errors.

derrickstolee: Good. Best to get the fix in sooner than #635.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -73,6 +113,13 @@ private void DeleteFiles(List<string> filePaths)
             }
         }
 
+        private int CountPackFiles()
+        {
+            return Directory.GetFiles(this.PackRoot)
```
kewillford: Why did you not use a search pattern in this call like `*.pack`?  Is that not case insensitive?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -28,25 +29,64 @@ public LooseObjectStepTests()
         [TestCase]
         public void RemoveLooseObjectsInPackFiles()
         {
-            // Delete any starting loose objects and verify
+            // Delete/Move any starting loose objects and packfiles
             this.DeleteFiles(this.GetLooseObjectFiles());
+            this.MovePackFilesToTemp();
+            Assert.AreEqual(0, this.GetLooseObjectFiles().Count);
+            Assert.AreEqual(0, this.CountPackFiles());
```
kewillford: Create a `Setup` method that does these 4 lines for each test?

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -61,21 +63,72 @@ public void LooseObjectsPassTimeRestriction()
...
+            this.TestSetup(DateTime.UtcNow.AddDays(-7));
+
+            LooseObjectsStep step = new LooseObjectsStep(this.context, requireCacheLock: false, forceRun: false);
+            string directoryName = "AB";
+            string fileName = "830bb79cd4fadb2e73e780e452dc71db909001";
+            string objectId = step.GetLooseObjectId(directoryName, Path.Combine(this.context.Enlistment.GitObjectsRoot, directoryName, fileName));
+            objectId.ShouldEqual(directoryName + fileName);
+        }
+
         private void TestSetup(DateTime lastRun)
```
kewillford: I would have this return the string for the pack command instead of storing it in at member variable so one test can't affect another and it is easier in the test to know where the `packCommand` is getting set.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,81 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+                        {
+                            string objectId = this.GetLooseObjectId(directoryName, file);
+                            if (!SHA1Util.IsValidShaFormat(objectId))
+                            {
+                                this.Context.Tracer.RelatedWarning($"Skippin invalid ObjectId {objectId}", Keywords.Telemetry);
+                                continue;
+                            }
+
+                            this.LooseObjectsPutIntoPackFile++;
+                            streamWriter.Write(objectId + "\n");
```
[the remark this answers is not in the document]
derrickstolee: The count is just for the telemetry. The next run will also count how many loose objects were _deleted_, which will be at least the number that were packed the previous run (plus any commits and trees that were downloaded but then covered by prefetch packs).

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,81 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+
+                            // Stop when we hit the limit.  The next run will pack more files.
```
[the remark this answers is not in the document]
derrickstolee: We want the job to do it's work, then get out of the way. It picks a small piece to do, then comes back later to do some more.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,81 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        public string GetLooseObjectId(string directoryName, string filePath)
+        {
+            return directoryName + Path.GetFileName(filePath);
```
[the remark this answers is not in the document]
derrickstolee: This isn't a path, but a 40-character hex object id.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: It does hurt: we'll slow down a lot if VFS is opening all of these files, too. Better to let Git handle it. Just check to see what `git pack-objects` does on a corrupted object. (Running a test now)

derrickstolee: Here's what happens:

```

[2018-12-19 13:05:27 -05:00] LooseObjectsStep (Start)
[2018-12-19 13:05:28 -05:00] RunGitCommand (Start) {"Area":"LooseObjectsStep"}
[2018-12-19 13:05:57 -05:00] RunGitCommand (Stop) {"DurationMs":29171}
[2018-12-19 13:05:57 -05:00] RunGitCommand (Start) {"Area":"LooseObjectsStep"}
[2018-12-19 13:06:41 -05:00] Warning {"WarningMessage":"LooseObjectsStep: Git process failed with errors:error: corrupt loose object '2b7d64425d8c559c9424db6cadc7fa2043e2dc23'\nfatal: loose object 2b7d64425d8c559c9424db6cadc7fa2043e2dc23 (stored in C:\\.gvfsCache\\a7414878aa15429f9b63c1a753783c7a\\gitObjects/2b/7d64425d8c559c9424db6cadc7fa2043e2dc23) is corrupt\n"}
[2018-12-19 13:06:41 -05:00] RunGitCommand (Stop) {"DurationMs":44178}
[2018-12-19 13:06:41 -05:00] LooseObjectsStep {"GitObjectsRoot":"C:\\.gvfsCache\\a7414878aa15429f9b63c1a753783c7a\\gitObjects","StartingCount":344526,"EndingCount":244526,"RemovedCount":100000,"LooseObjectsPutIntoPackFile":10000}
[2018-12-19 13:06:42 -05:00] LooseObjectsStep (Stop) {"DurationMs":74395}
```

So, we get a warning, and `git` returns with failure (which makes sense, it can't pack that object!). It does mean that we can track the warning. We have enough data logging to see if users are stuck at a certain value (or only increasing) and investigate. Deleting the corrupt loose object fixes the problem.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
kewillford: Should SHA1Util.IsValidShaFormat be used in this method then? or a comment?  Also seem weird to have this a public method?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
jeschu1: *.pack should work i'll update

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
jeschu1: I'm updating to TryGetLooseObjectId and will do the validation there.  I'd rather not put the function in GitObjects unless we find another instance that want to use it.

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
jeschu1: We set quite a few variables values in TestSetup.  Originally i had them as out parameters but that was agreed to be ugly.  I think the user can determine where the packCommand is set.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
jeschu1: I actually had to change the test start, so its no longer the same 4 lines

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
wilbaker: That sounds good to me, thanks!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
derrickstolee: You could also return it from `CreateLooseObjectsPackFile()` and avoid the member.

We create a new instance of the step each time we queue it, so it's not a bug at the moment. (Could creep in if we changed that.)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/LooseObjectStepTests.cs
```diff
@@ -25,28 +26,59 @@ public LooseObjectStepTests()
...
+
+            Assert.AreEqual(0, this.GetLooseObjectFiles().Count);
+            Assert.AreEqual(1, this.CountPackFiles());
+        }
 
-            // Run Cleanup
+        [TestCase]
+        public void NoLooseObjectsDoesNothing()
+        {
             this.Enlistment.LooseObjectStep();
```
derrickstolee: I think you need to order your test cases explicitly to ensure this test runs after the previous tests. That way you know there are no loose objects. Otherwise, you'll want to delete all loose objects in this test before running the step.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
wilbaker: > It's a member variable so can unit test it. If you see an easier way to do that let me know!

I was thinking an approach like this would allow the member variable to be removed:

https://github.com/wilbaker/VFSForGit/commit/3405d9bd55786ae861a0004fb4b05ece017cf71e

(You could also use `out` parameters rather than returning the number of loose objects added to the pack).

The advantage I see to that approach is there's no risk making a bad decision based on the value of `this.LooseObjectsPutIntoPackFile`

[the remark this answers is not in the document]
wilbaker: Note: I confirmed the unit tests passed with my changes, but I did not validate the correct loose object count is being written to the log.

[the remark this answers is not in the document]
wilbaker: > We create a new instance of the step each time we queue it, so it's not a bug at the moment. (Could creep in if we changed that.)

Ahh good call, I missed that detail

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -21,27 +24,88 @@ public LooseObjectsStep(GVFSContext context, bool requireCacheLock, bool forceRu
...
-                {                    
+                {
                     count += this.Context.FileSystem.GetFiles(Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath), "*").Count();
                 }
             }
 
             return count;
         }
 
+        public void PackLooseObjects(StreamWriter streamWriter)
```
wilbaker: Minor naming comment:  This method does not actually pack loose objects.  Perhaps `WriteLooseObjectSHAs` would be a more accurate description? (I came across this while writing up the changes in https://github.com/wilbaker/VFSForGit/commit/3405d9bd55786ae861a0004fb4b05ece017cf71e).

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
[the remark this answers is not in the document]
jeschu1: I'll happily cherry-pick that commit over @wilbaker thanks!

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -61,21 +63,84 @@ public void LooseObjectsPassTimeRestriction()
...
+            this.tracer.RelatedWarningEvents.Count.ShouldEqual(0);
+
+            // Write an ObjectId file with an invalid name
+            this.context.FileSystem.WriteAllText(Path.Combine(this.context.Enlistment.GitObjectsRoot, "AA", "NOT_A_SHA"), string.Empty);
+
+            // Verify it wasn't added and a warning exists
+            step.PackLooseObjects(new StreamWriter(new MemoryStream()));
+            step.LooseObjectsPutIntoPackFile.ShouldEqual(3);
+            this.tracer.RelatedErrorEvents.Count.ShouldEqual(0);
+            this.tracer.RelatedWarningEvents.Count.ShouldEqual(1);
```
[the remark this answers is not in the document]
jeschu1: I thought about that.  Then I'd make the messages string const and that gets tricky for those that have variables.  Considering most unit tests don't do the tracer of level of detail and we verify all the outputs I think this hits the mark for test coverage.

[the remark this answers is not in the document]
derrickstolee: Let's leave that message as a TODO for #650.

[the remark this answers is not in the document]
jeschu1: @derrickstolee @wilbaker just to throw it out there I'm not sure how much we really want to do this.  There is an overhead with verifying individual messages.  I've done in the past and have mixed feelings on the benefit versus maintenance.  There are certain cases where it's worth doing but I wouldn't say do it as a blanket rule.

derrickstolee:
Since it doesn't appear in the list of checks, [here is the full functional test build happening now](https://dev.azure.com/gvfs/ci/_build/results?buildId=5466).

GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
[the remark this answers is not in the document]
wilbaker: > There are certain cases where it's worth doing but I wouldn't say do it as a blanket rule.

I agree, I don't think we have to do it here.  I was just throwing it out as an option to consider in case you thought it was one of those cases where it'd be worth it.

PR #641, opened 2018-12-19
Mac kext: Fix for vnode leak causing shutdown/reboot hang and fix for memory leak

### LINKED ISSUE #499: Mac: Investigate hangs during shutdown when virtualization root is attached

pmj:
This fixes #499, again. Or rather, this second manifestation, where shutdown/reboot was [blocked as a result of virtualisation root vnodes having an iocount](https://github.com/Microsoft/VFSForGit/issues/499#issuecomment-447991612). This iocount was leaked because `clientClose` is not called on the user clients if they are terminated as a result of the kext being unloaded as opposed to the client process explicitly disconnecting or dying. The solution is to do the cleanup in the `stop()` function, which the second commit in this series implements.

While investigating/fixing this problem, I also noticed that the virtualisation roots array memory leaks on kext unload, so the first commit in the series fixes that.

Finally, while adding a private function to the provider user client class, I noticed that some existing functions were marked public which actually do not need to be accessed from outside the class, so the third commit fixes that.

**Reviewer note:** I recommend reviewing each commit individually, this will be easier to follow than reviewing overall changes in each file.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
@@ -15,7 +15,7 @@ OSDefineMetaClassAndStructors(PrjFSProviderUserClient, IOUserClient);
 static const uint32_t ProviderMessageQueueCapacityBytes = 100 * 1024;
 
 
-static const IOExternalMethodDispatch ProviderUserClientDispatch[] =
+const IOExternalMethodDispatch PrjFSProviderUserClient::ProviderUserClientDispatch[] =
 {
     [ProviderSelector_RegisterVirtualizationRootPath] =
         {
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
@@ -80,6 +80,16 @@ CleanupAndFail:
     return false;
 }
 
+void PrjFSProviderUserClient::stop(IOService* provider)
+{
+    // On the code path of process dying or disconnecting, the following is unnecessary,
+    // but if the kext is unloaded with providers attached, clientClose() will never be
+    // called on them so we do the cleanup here.
+    this->cleanupProviderRegistration();
+    
+    this->super::stop(provider);
+}
+
 void PrjFSProviderUserClient::free()
 {
     OSSafeReleaseNULL(this->dataQueueMemory);
@@ -96,15 +106,20 @@ void PrjFSProviderUserClient::free()
 // the connection.
 IOReturn PrjFSProviderUserClient::clientClose()
 {
+    this->cleanupProviderRegistration();
+    
+    this->terminate(0);
+    return kIOReturnSuccess;
+}
+
+void PrjFSProviderUserClient::cleanupProviderRegistration()
+{
     VirtualizationRootHandle root = this->virtualizationRootHandle;
     this->virtualizationRootHandle = RootHandle_None;
     if (RootHandle_None != root)
     {
         ActiveProvider_Disconnect(root);
     }
-    
-    this->terminate(0);
-    return kIOReturnSuccess;
 }
 
 // Called when user process requests memory-mapping of kernel data
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
@@ -124,6 +124,13 @@ kern_return_t VirtualizationRoots_Init()
 
 kern_return_t VirtualizationRoots_Cleanup()
 {
+    if (s_virtualizationRoots != nullptr)
+    {
+        Memory_FreeArray(s_virtualizationRoots, s_maxVirtualizationRoots);
+        s_virtualizationRoots = nullptr;
+        s_maxVirtualizationRoots = 0;
+    }
+
     if (RWLock_IsValid(s_virtualizationRootsLock))
     {
         RWLock_FreeMemory(&s_virtualizationRootsLock);
--- a/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
@@ -126,6 +126,12 @@ kern_return_t VirtualizationRoots_Cleanup()
 {
     if (s_virtualizationRoots != nullptr)
     {
+        for (uint32_t i = 0; i < s_maxVirtualizationRoots; ++i)
+        {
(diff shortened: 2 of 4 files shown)


ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
```diff
@@ -80,6 +80,16 @@ bool PrjFSProviderUserClient::initWithTask(
     return false;
 }
 
+void PrjFSProviderUserClient::stop(IOService* provider)
```
wilbaker: I noticed a similar comment in `PrjFSLogUserClient::cleanUp()`, and it looks like in `PrjFSLogUserClient::cleanUp()` is called from `PrjFSLogUserClient::free()`.

Are there advantages to doing this cleanup in `stop(...)` instead of `free()` (or was using `free()` not an option here)?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -124,6 +124,19 @@ kern_return_t VirtualizationRoots_Init()
 
 kern_return_t VirtualizationRoots_Cleanup()
 {
+    if (s_virtualizationRoots != nullptr)
+    {
+        for (uint32_t i = 0; i < s_maxVirtualizationRoots; ++i)
+        {
+            // If there are still providers registered at this point, we will leak vnodes
```
wilbaker: For my own understanding, does `vfs_unmountall` block until all vnodes are released? (I am assuming this was the cause of the hang?)

ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
```diff
@@ -95,16 +105,21 @@ void PrjFSProviderUserClient::free()
 // Called when the user process explicitly or implicitly (process death) closes
 // the connection.
 IOReturn PrjFSProviderUserClient::clientClose()
+{
+    this->cleanupProviderRegistration();
```
wilbaker: Will `PrjFSProviderUserClient::stop` always be called?  

(If it is I'm wondering if we need to call `cleanupProviderRegistration()`)

ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
pmj: `free()` is effectively a destructor in the `OSObject` system, so that's conceptually more a place to free resources. For `IOService` subclasses (including `IOUserClient`) you've got this lifecycle:

 * Object creation with `new`
   * Object initialisation with `init…()`
     * Attaching service to provider object in IORegistry graph using `attach()`
       * Service start with `start()`
         * (useful life of the object)
         * Service termination request (`terminate()`)
       * Service shutdown with `stop()`
     * Detaching service from provider (`detach()`)
 * Object eventually hits retain count of 0 (hopefully), which causes `free()` to be called, and subsequent destruction/deallocation with `delete`

If any of `init`, `attach`, or `start` fail, it falls back out to the previous level of indentation in the above hierarchy. So `stop()` is only called if `start()` succeeded, etc. Similarly, `free()` will only be called once retain count hits 0, which could conceivably be a lot later than `stop()`. (Or never if there's a retain leak.)

So `stop()` is probably a better place for this, and if anything, the `KextLog_DeregisterUserClient()` call in `PrjFSLogUserClient` should be moved to its `stop()` function. (That one is less serious if it doesn't get called though.)

ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
pmj: Similarly to the `free` question - yes, we *could* wait for `stop()` but `clientClose()` will definitely be called in the context of the disconnecting or dying client, whereas I don't think there's any such guarantee about `stop()` on user clients. (Indeed, it wouldn't be called at all if we didn't `terminate(0)`.) This means that potentially, a client that disconnects and immediately reconnects to the same virtualisation root could find that its previous provider client connection hasn't been deregistered yet because `stop()` hasn't been called on it yet. The chances of this happening are minimal, but I'd rather avoid the situation. Plus, we'd have to guarantee that `sendMessage()` will be safe to call past `clientClose()`, which is just an unnecessary extra burden.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: I'd have to read the source code to say for sure, but I assume the answer is yes because vnodes have data attached to them that's private to the file system to which they belong. So it's potentially not safe to keep vnodes around after their file system has been unmounted (and the file system's kext potentially unloaded); with that background it makes sense to me that allowing an unmount to complete would require all associated vnodes to be recycled.

ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
wilbaker: Ok thanks for the details! Based on your description I agree `stop()` is the better place.

ProjFS.Mac/PrjFSKext/PrjFSKext/PrjFSProviderUserClient.cpp
```diff
@@ -80,6 +80,16 @@ bool PrjFSProviderUserClient::initWithTask(
     return false;
 }
 
+void PrjFSProviderUserClient::stop(IOService* provider)
+{
+    // On the code path of process dying or disconnecting, the following is unnecessary,
+    // but if the kext is unloaded with providers attached, clientClose() will never be
+    // called on them so we do the cleanup here.
```
jeschu1: I really appreciate great comments like this.  Makes it very clear whats going on!

jeschu1:
@pmj PR looks good sorry I didn't get to it yesterday (and it's my last day in the office until next year).  

I also appreciate you making the separate commits as it makes it very easy to review.

Issue #328, opened 2018-10-03
Mac: sporadic vnode I/O errors


pmj:
There seem to be a number of places in the kext's vnode/fileop handler where I/O (reading attributes, etc.) occasionally fails. In a few places there are `assert`s that even take down the system in case of such failures. We need to track down all the failure cases, some common root causes, and handle them correctly and gracefully in the kext instead of crashing the system.

Known failure cases:
 * [Failed assertions on reading file attributes](https://github.com/Microsoft/VFSForGit/pull/322/commits/fdd539e06deb26c7c066beac54757664c703b298)
 * [Panics in file close events](https://github.com/Microsoft/VFSForGit/pull/322#issuecomment-426387710)

Reports of more cases appreciated.

derrickstolee:
@pmj Is the following test case that failed on a Mac functional test related? Or should I create a separate issue?

```
1) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.MultithreadedReadWriteTests.CanReadVirtualFileInParallel()
  At least one of the reads failed
  Expected: null
  But was:  <System.IO.IOException: The process cannot access the file '/GVFS.FT/test/a6a14cb56232452c8d68/src/GVFS/GVFS.FunctionalTests/Tests/LongRunningEnlistment/GitMoveRenameTests.cs' because it is being used by another process.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize)
   at System.IO.File.InternalReadAllText(String path, Encoding encoding)
   at System.IO.File.ReadAllText(String path)
   at GVFS.FunctionalTests.FileSystemRunners.SystemIORunner.ReadAllText(String path) in /Users/vsts/agent/2.140.2/work/1/s/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs:line 70
[...]

PR #643, opened 2018-12-19
Mac kext: log more I/O errors

pmj:
This is a cleaned up and toned-down version of PR #555 in that (hopefully) all I/O calls in the kext are error checked and will log the error if it is unexpected. This is one part of issue #328 - as far as I'm aware, most of these errors should only very rarely be encountered (they're not turning up in my tests)  so I'm mainlining the logging of these remaining error conditions to try and get a better feel for if and when they do turn up.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -1100,8 +1100,8 @@ static bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t _Nonnull context, ui
     }
     
     assert(VATTR_IS_SUPPORTED(&attributes, va_flags));
-     *flags = attributes.va_flags;
-     return true;
+    *flags = attributes.va_flags;
+    return true;
 }
 
 static inline bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit)
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -15,6 +15,7 @@
 #include "Locks.hpp"
 #include "PrjFSProviderUserClient.hpp"
 #include "PerformanceTracing.hpp"
+#include "kernel-header-wrappers/mount.h"
 
 // Function prototypes
 static int HandleVnodeOperation(
@@ -345,9 +346,22 @@ static int HandleVnodeOperation(
         
         // Call vn_getpath first when the cache is hottest to increase the chances
         // of successfully getting the path
-        if (0 == vn_getpath(currentVnode, vnodePathBuffer, &vnodePathLength))
+        errno_t error = vn_getpath(currentVnode, vnodePathBuffer, &vnodePathLength);
+        if (0 == error)
         {
             vnodePath = vnodePathBuffer;
+        }
+        else
+        {
+            const char* name = vnode_getname(currentVnode);
+            mount_t mount = vnode_mount(currentVnode);
+            vfsstatfs* vfsStat = mount != nullptr ? vfs_statfs(mount) : nullptr;
+
+            KextLog_Error("HandleVnodeOperation: vn_getpath failed for vnode %p, error = %d, name '%s', recycled: %s, on mount point mounted at '%s'", currentVnode, error, name ?: "[NULL]", vnode_isrecycled(currentVnode) ? "yes" : "no", vfsStat ? vfsStat->f_mntonname : "[NULL]");
+            if (name != nullptr)
+            {
+                vnode_putname(name);
+            }
         }
     }
 
@@ -535,6 +549,7 @@ static int HandleFileOpOperation(
         errno_t toErr = vnode_lookup(newPath, 0 /* flags */, &currentVnode, context);
         if (0 != toErr)
         {
+            KextLog_Error("HandleFileOpOperation: vnode_lookup failed, errno %d for path '%s'", toErr, newPath);
             goto CleanupAndReturn;
         }
         
@@ -624,7 +639,12 @@ static int HandleFileOpOperation(
         bool fileFlaggedInRoot;
         if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
         {
-            KextLog_Info("Failed to read attributes when handling FileOp operation. Path is %s", path);
+            const char* vnode_name = vnode_getname(currentVnode);
+            KextLog_Error("KAUTH_FILEOP_CLOSE: checking file flags failed. Path = '%s' Vnode name: %s, type %d, being recycled: %s",
+                path, vnode_name ?: "[NULL]", vnode_vtype(currentVnode), vnode_isrecycled(currentVnode) ? "yes" : "no");
+            if (vnode_name)
+                vnode_putname(vnode_name);
+            
             goto CleanupAndReturn;
         }
         
@@ -1095,7 +1115,7 @@ static bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t _Nonnull context, ui
         //   - Logging this error
         //   - Falling back on vnode lookup (or custom cache) to determine if file is in the root
         //   - Assuming files are empty if we can't read the flags
-        
+        KextLog_FileError(vn, "ReadVNodeFileFlags: GetVNodeAttributes failed with error %d; vnode type: %d, recycled: %s", err, vnode_vtype(vn), vnode_isrecycled(vn) ? "yes" : "no");
         return false;
     }
     
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -353,15 +353,7 @@ static int HandleVnodeOperation(
         }
         else
         {
-            const char* name = vnode_getname(currentVnode);
-            mount_t mount = vnode_mount(currentVnode);
-            vfsstatfs* vfsStat = mount != nullptr ? vfs_statfs(mount) : nullptr;
-
-            KextLog_Error("HandleVnodeOperation: vn_getpath failed for vnode %p, error = %d, name '%s', recycled: %s, on mount point mounted at '%s'", currentVnode, error, name ?: "[NULL]", vnode_isrecycled(currentVnode) ? "yes" : "no", vfsStat ? vfsStat->f_mntonname : "[NULL]");
-            if (name != nullptr)
-            {
-                vnode_putname(name);
-            }
+            KextLog_ErrorVnodeProperties(currentVnode, "HandleVnodeOperation: vn_getpath failed, error = %d", error);
         }
     }
 
@@ -639,11 +631,7 @@ static int HandleFileOpOperation(
         bool fileFlaggedInRoot;
         if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
         {
-            const char* vnode_name = vnode_getname(currentVnode);
-            KextLog_Error("KAUTH_FILEOP_CLOSE: checking file flags failed. Path = '%s' Vnode name: %s, type %d, being recycled: %s",
-                path, vnode_name ?: "[NULL]", vnode_vtype(currentVnode), vnode_isrecycled(currentVnode) ? "yes" : "no");
-            if (vnode_name)
-                vnode_putname(vnode_name);
+            KextLog_ErrorVnodeProperties(currentVnode, "KAUTH_FILEOP_CLOSE: checking file flags failed. Path = '%s'", path);
             
             goto CleanupAndReturn;
         }
--- a/ProjFS.Mac/PrjFSKext/KextLog.hpp
+++ b/ProjFS.Mac/PrjFSKext/KextLog.hpp
@@ -4,6 +4,8 @@
 #include "public/PrjFSCommon.h"
 #include "PrjFSClasses.hpp"
 #include "public/PrjFSLogClientShared.h"
+#include "kernel-header-wrappers/vnode.h"
+#include "kernel-header-wrappers/mount.h"
 #include <os/log.h>
 
 extern os_log_t __prjfs_log;
@@ -22,7 +24,14 @@ void KextLog_Printf(KextLog_Level loglevel, const char* fmt, ...)  __printflike(
 
 // Helper macros/function for logging with file paths. Note that the path must
 // be the last % format code in the format string, but the vnode is the first
-// argument following the format string. An unfortunate implementation detail!
+// argument following the format string. An unfortunate implementation detail,
+// but not too problematic if you just use the macros and not the function directly.
+//
+// The reason for the helper function/macro split is that we want to encourage
+// the compiler to create a new stack frame to hold the path buffer in order to
+// avoid bloating the caller's stack frame; we can't do this in a macro.
+// We need the macros below for the format string concatenation; we can't do
+// this in a function.
 struct vnode;
 extern "C" int vn_getpath(struct vnode *vp, char *pathbuf, int *len);
 template <typename... args>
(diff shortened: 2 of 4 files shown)


ProjFS.Mac/PrjFSKext/PrjFSKext/VnodeUtilities.cpp
```diff
@@ -12,7 +13,11 @@ FsidInode Vnode_GetFsidAndInode(vnode_t vnode, vfs_context_t context)
     // TODO: check this is correct for hardlinked files
     VATTR_WANTED(&attrs, va_fileid);
 
-    vnode_getattr(vnode, &attrs, context);
+    int errno = vnode_getattr(vnode, &attrs, context);
+    if (0 != errno)
+    {
+        KextLog_FileNote(vnode, "Vnode_GetFsidAndInode: vnode_getattr failed with errno %d", errno);
```
wilbaker: How bad is it if we hit this error?  I am wondering if this should be logged at a higher severity (e.g. `Info` or `Error`)

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -625,6 +640,13 @@ static int HandleFileOpOperation(
         if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
         {
             KextLog_Info("Failed to read attributes when handling FileOp operation. Path is %s", path);
```
wilbaker: Can this `KextLog_Info` be removed now that we're logging an error on line 645?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -625,6 +640,13 @@ static int HandleFileOpOperation(
...
+
+            const char* vnode_name = vnode_getname(currentVnode);
+            KextLog_Error("KAUTH_FILEOP_CLOSE: checking file flags failed. Path = '%s' Vnode name: %s, type %d, being recycled: %s",
+                path, vnode_name ?: "[NULL]", vnode_vtype(currentVnode), vnode_isrecycled(currentVnode) ? "yes" : "no");
```
wilbaker: Should we also check if `path` is null?  I think you might have mentioned seeing a null path when the vnode had been recycled.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -144,7 +144,19 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
...
     VirtualizationRootHandle rootHandle = RootHandle_None;
     
-    vnode_get(vnode);
+    errno_t error = vnode_get(vnode);
+    if (error != 0)
+    {
+        const char* name = vnode_getname(vnode);
+        mount_t mount = vnode_mount(vnode);
+        vfsstatfs* vfsStat = mount != nullptr ? vfs_statfs(mount) : nullptr;
+        KextLog_FileNote(vnode, "VirtualizationRoot_FindForVnode: vnode_get() failed (error = %d) on vnode we'd expect to be live; name = '%s', recycling = %s, mount point mounted at path '%s'", error, name ?: "[NULL]", vnode_isrecycled(vnode) ? "yes" : "no", vfsStat ? vfsStat->f_mntonname : "[NULL]");
```
wilbaker: How severe do you think this scenario is?  Should we consider making it Info or Error level?

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -160,6 +172,10 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
         }
         
         vnode_t parent = vnode_getparent(vnode);
+        if (nullptr == parent)
```
wilbaker: Minor: Should we be comparing against `NULLVP` rather than `nullptr`? (I am assuming they have the same value, but I've seen us use `NULLVP` elsewhere).

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
wilbaker: Same question applies other spots that `KextLog_FileNote` is called.  How did you decide what was severe enough to warrant logging at the `Error` level?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -345,10 +346,23 @@ static int HandleVnodeOperation(
         
         // Call vn_getpath first when the cache is hottest to increase the chances
         // of successfully getting the path
-        if (0 == vn_getpath(currentVnode, vnodePathBuffer, &vnodePathLength))
+        int error = vn_getpath(currentVnode, vnodePathBuffer, &vnodePathLength);
```
jeschu1: Thoughts on 'result' or 'returnValue' instead of 'error'.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
```diff
@@ -345,10 +346,23 @@ static int HandleVnodeOperation(
...
+        else
+        {
+            const char* name = vnode_getname(currentVnode);
+            mount_t mount = vnode_mount(currentVnode);
+            vfsstatfs* vfsStat = mount != nullptr ? vfs_statfs(mount) : nullptr;
+
+            KextLog_Error("HandleVnodeOperation: vn_getpath failed for vnode %p, error = %d, name '%s', recycled: %s, on mount point mounted at '%s'", currentVnode, error, name ?: "[NULL]", vnode_isrecycled(currentVnode) ? "yes" : "no", vfsStat ? vfsStat->f_mntonname : "[NULL]");
+            if (name != nullptr)
+            {
+                vnode_putname(name);
```
jeschu1: For my own info, what happens in this case.  I see we call vnode_putname(name) here, what would be difference between this and going thru the normal flow.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -160,6 +172,10 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
...
+        {
+            KextLog_FileNote(vnode, "VirtualizationRoot_FindForVnode: vnode_getparent returned nullptr on vnode that is not root of a mount point");
+        }
         vnode_put(vnode);
         vnode = parent;
```
jeschu1: Could this create problems further on if parent was null?

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: The `vsprintf` implementation in the kernel already is hardened against `NULL` arguments for `%s`, so I don't see a need to redo this explicitly.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: I guess I picked "error" because this function returns `errno` style errors (`EINVAL`, `ENOSPC`, etc.). There's also no symbolic "success" constant for this type of result, so the check for that case, `(0 == error)` perhaps is a little clearer ("no error occurred") than `(0 == result)`.

Incidentally, I'll replace the declaration as `int` with `errno_t` to make the pool of possible cases clearer.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
wilbaker: > The vsprintf implementation in the kernel already is hardened against NULL arguments for %s, so I don't see a need to redo this explicitly.

How does that implementation behave when it encounters a NULL?  What prompted this question was the NULL included for `vnode_name`:

```vnode_name ?: "[NULL]"```

File: ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp (hunk 7 above)
Comment (pmj) [reply ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp]: Memory for vnode names is reference counted (and pooled, I believe?), so if you successfully call `vnode_getname()` as we do 7 lines up, you need to match it with a call to `vnode_putname()`. Not to be confused with `vnode_get()` and `vnode_put()` which operate on the vnode object itself, not the name.

File: ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp (hunk 8 above)
Comment (pmj) [reply ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp]: Yes; dealing with the fallout from this is something that still needs to be done and is out of scope of this PR. See [issue #338](https://github.com/Microsoft/VFSForGit/issues/338).

File: ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp [hunk 9]
```diff
@@ -1095,13 +1117,13 @@ static bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t _Nonnull context, ui
         //   - Logging this error
         //   - Falling back on vnode lookup (or custom cache) to determine if file is in the root
         //   - Assuming files are empty if we can't read the flags
-        
+        KextLog_FileError(vn, "ReadVNodeFileFlags: GetVNodeAttributes failed with error %d; vnode type: %d, recycled: %s", err, vnode_vtype(vn), vnode_isrecycled(vn) ? "yes" : "no");
         return false;
     }
     
     assert(VATTR_IS_SUPPORTED(&attributes, va_flags));
```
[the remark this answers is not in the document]
pmj: Kernel panic, if `MACH_ASSERT` is `#define`d as 1. (or something evaluating to an expression that corresponds to boolean true) We have it [defined as 1] for [all](https://github.com/Microsoft/VFSForGit/blob/687c159cc1e621a3976a97d390e541db50d60c57/ProjFS.Mac/PrjFSKext/PrjFSKext.xcodeproj/project.pbxproj#L349) [build](https://github.com/Microsoft/VFSForGit/blob/687c159cc1e621a3976a97d390e541db50d60c57/ProjFS.Mac/PrjFSKext/PrjFSKext.xcodeproj/project.pbxproj#L507) [configurations](https://github.com/Microsoft/VFSForGit/blob/687c159cc1e621a3976a97d390e541db50d60c57/ProjFS.Mac/PrjFSKext/PrjFSKext.xcodeproj/project.pbxproj#L533).

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
pmj: Ah yes, that's a little inconsistent; I just checked, and in practice, `path` [(arg1) will *never* be `NULL` for `KAUTH_FILEOP_CLOSE`](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/kern/kern_authorization.c#L561).

Incidentally, I believe the issue I noticed was that `vn_getpath()` was failing/returning an empty path when called during `KextLog_File*()` in an error message.

pmj: To answer your question: the built-in hardening [treats `NULL` strings as the empty string](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/osfmk/kern/printf.c#L443) which isn't exactly pretty and (to me) doesn't stand out when scanning error messages.

As it looks like path will never be NULL anyway, adding mitigation doesn't really make any sense, and likewise I think it might be useful to specifically call out the case of a NULL vnode name, so I'm inclined to stick with the current version of the code.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Sorry about that, looks like I didn't subject this series of commits to my usual level of scrutiny before opening the PR; probably because they're originally from the #555 branch which is at more of a prototype-level quality. I think they can all be errors, so I'll go through and change them shortly.

Incidentally, "note" is above "info" in severity; I've mapped it to `os_log`'s "default" level. For consistency's sake, we should probably rename our note level too, and deprecate the old name.

ProjFS.Mac/PrjFSKext/PrjFSKext/VirtualizationRoots.cpp
pmj: Yep, good catch. And yes, they're essentially the same, except with a bit of extra type safety.

ProjFS.Mac/PrjFSKext/PrjFSKext/KauthHandler.cpp
jrbriggs: Did you consider a RAII pattern to ensure we release the reference on `name`? I like letting the language ensure resource lifetime is tied to object lifetime, and it's a bit of extra safety if anyone introduces a new exit point when that reference is held.

wilbaker: @jrbriggs I like to use RAII when possible as well.

@pmj what are your thoughts in general when it comes to using the RAII pattern in kext code?

pmj: > what are your thoughts in general when it comes to using the RAII pattern in kext code?

I'm not generally opposed to it at all; I'm definitely a fan of getting the compiler to verify correctness as much as possible. I'd consider 2 major and 1 minor argument against wrapping this and some of the other resource releasing APIs, however:

1. The system APIs themselves don't return RAII types, so you need to either wrap all the kernel APIs with RAII-ified versions (and enforce using only them), *or* you need to RAII-ify correctly at each call site, which can itself become a potential hazard. `vnode_getname()` for example returns a `const char*`; this doesn't mean that every `const char*` can or should be wrapped in the same RAII class. For more complex APIs, such as those dealing with `vnode_t` themselves or `OSObject`, you have to be aware of whether a returned reference is retained or not (in the vnode case, the distinction between iocount and refcount makes it even more complicated).
2. Some things are best left explicit, particularly when interacting with locks/mutexes. There are certain APIs where it's very much relevant whether or not they're called inside or outside a held lock; when calling them explicitly, this intention is typically quite visible in the code. When they're called implicitly in a destructor, you have to infer the intention based on the scope of the RAII variable; sometimes your intention conflicts with C++ variable scoping rules, so you end up having to work around the scoping and perform the call explicitly in some way or another after all. If you're having to resort to that sort of thing, I'd consider that making things worse, not better.
3. *(This argument carries less weight.)* The RAII wrappers themselves need to be maintained and could themselves be sources of errors.

Item 2 could be counteracted by a kind of "RAII lite" where resource acquisition & release remain explicit, but are checked at runtime by assertions in the destructor - a.k.a. a hack for missing support for linear types in C++.

jeschu1: @pmj Thanks for the detailed explaination on vnode_putname and detailed thoughts!

You mentioned the get/put pairing is for reference counting.  Out of curiosity what would be the downfall if we forget to do the put?  Would it solely be a memory leak or could it be more destructive.  Since theses are vnodes I'm guessing just a memory leak would be pretty dangerous, just wondering what other consequences could happen for my own understanding.  

Finally would a wrapper function that pairs the get/put, be a simpler alternative to RAII in this case.  Just thinking it would be a more lightweight.  Something like:
```
char* GetName(vnode)
{
  const char* name = vnode_getname(vnode); 
  if (name != nullptr) 
  { 
     vnode_putname(name); 
  }
  return name;
}
```

pmj: > You mentioned the get/put pairing is for reference counting. Out of curiosity what would be the downfall if we forget to do the put? Would it solely be a memory leak or could it be more destructive. Since theses are vnodes I'm guessing just a memory leak would be pretty dangerous, just wondering what other consequences could happen for my own understanding.

I had to look it up in the source (bsd/vfs/vfs_cache.c in xnu), but it looks like it's a fairly straightforward memory leak. Obviously a memory leak in the kernel is worse than in user space because it's permanent until reboot, and it's wired memory, so can never be swapped out. Leaking a vnode itself (as opposed to its name) would arguably be worse, as we found out in #499.

> Finally would a wrapper function that pairs the get/put, be a simpler alternative to RAII in this case. Just thinking it would be a more lightweight. Something like:
> 
> ```
> char* GetName(vnode)
> {
>   const char* name = vnode_getname(vnode); 
>   if (name != nullptr) 
>   { 
>      vnode_putname(name); 
>   }
>   return name;
> }
> ```

It's a nice idea, but it's unfortunately not safe to dereference that pointer after you've relinquished your last refcount with `vnode_putname()`. For example, the vnode could be renamed or deleted by another thread while we're executing this code, which could cause the refcount on the name to drop to 0.

jeschu1: Thanks @pmj that makes sense!  I'm learning so much on these PRs :-).

I do think a pattern that enforces get/put are always together would a good idea for safety.

pmj: > I do think a pattern that enforces get/put are always together would be a good idea for safety.

I've now prototyped this [here](https://github.com/pmj/VFSForGit/commit/abd81ec931f0e636d5f17ed50d6b8341d1dcd28c).

I've not added it to this PR as I honestly don't think it's a sufficient improvement. You still have to remember to use the wrapper, you need to be aware of the wrapper's semantics (construction *transfers* ownership) and none of the instances where the code uses `vnode_getname()` is so complicated that the `vnode_putname()` call is likely to be missed out on some codepath or another.

Based on the observation that I've only used `vnode_getname()` for logging detailed vnode properties so far, I've gone in the opposite direction in the latest commit (524a78f) added to this PR. I've simply created logging wrapper macros along the lines of our existing `KextLog_File*` macros, but which output those extra details about the vnode. This cleans up the error site and moves the contentious `vnode_getname()/vnode_putname()` pair into a helper function where it's hopefully out of harm's way. (N.B. I still need to *test* this latest commit but barring any stupid typos it should work.)

ProjFS.Mac/PrjFSKext/PrjFSKext/KextLog.hpp
```diff
@@ -42,6 +44,53 @@ template <typename... args>
...
+    }
+
+template <typename... args>
+    void KextLog_PrintfVnodeProperties(KextLog_Level loglevel, struct vnode* vnode, const char* fmt, args... a)
+    {
+        const char* name = vnode_getname(vnode);
+        mount_t mount = vnode_mount(vnode);
+        vfsstatfs* vfsStat = mount != nullptr ? vfs_statfs(mount) : nullptr;
+        
+        KextLog_Printf(loglevel, fmt, a..., name ?: "[NULL]", vnode_vtype(vnode), vnode_isrecycled(vnode) ? "yes" : "no", vfsStat ? vfsStat->f_mntonname : "[NULL]");
```
wilbaker: Is there any clean way to move the format strings out of the macros and into `KextLog_PrintfVnodePathAndProperties` and `KextLog_PrintfVnodeProperties`?

I like the addition of the helpers, but the one thing that threw me off a bit was that the `KextLog_PrintfVnodePathAndProperties` and `KextLog_PrintfVnodeProperties` *must* be called with a particular `fmt`.

ProjFS.Mac/PrjFSKext/PrjFSKext/KextLog.hpp
```diff
@@ -42,6 +44,53 @@ template <typename... args>
...
+        
+        const char* name = vnode_getname(vnode);
+        mount_t mount = vnode_mount(vnode);
+        vfsstatfs* vfsStat = mount != nullptr ? vfs_statfs(mount) : nullptr;
+        
+        KextLog_Printf(loglevel, fmt, a..., vnodePath, name ?: "[NULL]", vnode_vtype(vnode), vnode_isrecycled(vnode) ? "yes" : "no", vfsStat ? vfsStat->f_mntonname : "[NULL]");
+ 
+        if (name != nullptr)
+        {
+            vnode_putname(name);
```
jeschu1: I like the way you've done this change :-).  Could you please add a small comment above the vnode_putname explaining why we need it.  I imagine it's redundant for folks that know kext well but for newbies like me it'd be helpful.

ProjFS.Mac/PrjFSKext/PrjFSKext/KextLog.hpp
pmj: > Is there any clean way to move the format strings out of the macros and into `KextLog_PrintfVnodePathAndProperties` and `KextLog_PrintfVnodeProperties`?

Essentially, no, at least not without runtime downsides. Let's look at the macro definition:
```
#define KextLog_ErrorVnodeProperties(vnode, format, ...) \
    ({ \
        _os_log_verify_format_str(format, ##__VA_ARGS__); \
        KextLog_PrintfVnodeProperties(KEXTLOG_ERROR, vnode, format " (vnode name: '%s', type: %d, recycling: %s, mount point mounted at path '%s')", ##__VA_ARGS__); \
    })
```

Note that the `format` macro parameter is placed directly next to the extra format string literal with only whitespace as separation:
```
format " (vnode name: '%s', type: %d, recycling: %s, mount point mounted at path '%s')"
```
When expanded, this simply places 2 string literals next to one another, which the compiler concatenates into a single string literal - this means it wouldn't won't work if you passed a symbolic format string variable or constant. (except of course a macro that itself expands to a string literal)

This literal concatenation of course doesn't work on function parameters, so it can't go inside the helper function. As far as I'm aware, you can't sensibly do this even with C++14's `constexpr` functions. There are some [awful hacks with templates](https://akrzemi1.wordpress.com/2017/06/28/compile-time-string-concatenation/), but for this use case it would make things much worse, not better.

> I like the addition of the helpers, but the one thing that threw me off a bit was that the `KextLog_PrintfVnodePathAndProperties` and `KextLog_PrintfVnodeProperties` _must_ be called with a particular `fmt`.

The other possible solution to that is of course to not use a helper function in the first place, and just put everything in the macro. I very much wanted to avoid that especially for the cases where we call `vn_getpath` with a stack-allocated string buffer. If we created one of those in an expanded macro, and we used that macro multiple times in a function, the compiler could elect to statically allocate space for those temporary buffers in the stack frame. We have a very small stack in kernel threads (a little less than 16KiB last time I checked), and the path string buffer is very large (1KiB) so allocating a bunch of them could
easily blow the stack, especially with nested function calls. Putting the buffer in a near-leaf helper function minimises that risk.

So I agree, it's not pretty, but it's the least bad solution I could think of.

pmj: I've just pushed a new version of this PR which adds this rationale as a comment for the `KextLogFile_Printf()` function, where the same reasoning applies.

ProjFS.Mac/PrjFSKext/PrjFSKext/KextLog.hpp
pmj: Done!

PR #644, opened 2018-12-19
Update Git to include commit-graph BUG statement

derrickstolee:
This prevents writing bad data to the commit-graph when the list of commits to write is corrupted in-memory.


derrickstolee:
See [this PR in microsoft/git](https://github.com/Microsoft/git/pull/97) for the details of this change.

PR #645, opened 2018-12-19
Enable FastFetch to lay out symlinks correctly

nickgra:
Provide a native UNIX implementation for creating symlinks so that repos that contain files with relative path symlinks (like Bundles: https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW13) can be created correctly on disk.

Since we don't support symlinks yet on Windows (in VFSForGit or FastFetch), I pulled the implementation into WriteFile as we can detect a symlink based on the mode of the path that we're trying to write out.

### DIFF
--- a/GVFS/FastFetch/NativeUnixMethods.cs
+++ b/GVFS/FastFetch/NativeUnixMethods.cs
@@ -14,19 +14,33 @@ namespace FastFetch
         public const int Create = 0x0200;
         public const int Truncate = 0x0400;
         private const int InvalidFileDescriptor = -1;
+        private const ushort SymLinkMode = 0xA000;
 
         public static unsafe void WriteFile(ITracer tracer, byte* originalData, long originalSize, string destination, ushort mode)
         {
             int fileDescriptor = InvalidFileDescriptor;
             try
             {
-                fileDescriptor = Open(destination, WriteOnly | Create | Truncate, mode);
-                if (fileDescriptor != InvalidFileDescriptor)
+                if (mode == SymLinkMode)
                 {
-                    IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
-                    if (result.ToInt32() == -1)
+                    string linkLocation = Marshal.PtrToStringUTF8(new IntPtr(originalData));
+                    int result = CreateSymLink(linkLocation, destination);
+                    if (result == -1)
                     {
+                        tracer.RelatedInfo("Failed to create symlink({0},{1})", linkLocation, destination);
                         throw new Win32Exception(Marshal.GetLastWin32Error());
+                    }
+                }
+                else
+                {
+                    fileDescriptor = Open(destination, WriteOnly | Create | Truncate, mode);
+                    if (fileDescriptor != InvalidFileDescriptor)
+                    {
+                        IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
+                        if (result.ToInt32() == -1)
+                        {
+                            throw new Win32Exception(Marshal.GetLastWin32Error());
+                        }
                     }
                 }
             }
@@ -79,6 +93,9 @@ namespace FastFetch
 
         [DllImport("libc", EntryPoint = "write", SetLastError = true)]
         private static unsafe extern IntPtr Write(int fileDescriptor, void* buf, IntPtr count);
+
+        [DllImport("libc", EntryPoint = "symlink", SetLastError = true)]
+        private static extern int CreateSymLink(string sourcePath, string destinationPath);
 
         private static NativeStat.StatBuffer StatFile(string fileName)
         {
--- a/GVFS/FastFetch/NativeUnixMethods.cs
+++ b/GVFS/FastFetch/NativeUnixMethods.cs
@@ -23,24 +23,25 @@ namespace FastFetch
             {
                 if (mode == SymLinkMode)
                 {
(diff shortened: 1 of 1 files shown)


GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -14,21 +14,35 @@ public class NativeUnixMethods
...
             int fileDescriptor = InvalidFileDescriptor;
             try
             {
-                fileDescriptor = Open(destination, WriteOnly | Create | Truncate, mode);
-                if (fileDescriptor != InvalidFileDescriptor)
+                if (mode == SymLinkMode)
                 {
-                    IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
-                    if (result.ToInt32() == -1)
+                    string linkLocation = Marshal.PtrToStringUTF8(new IntPtr(originalData));
```
wilbaker: Naming suggestion: `linkLocation` could be interpreted as the location at which the symlink file will be created.  Consider renaming it to something like `linkTarget`.

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -80,6 +94,9 @@ public static bool TryStatFileAndUpdateIndex(ITracer tracer, string path, Memory
         [DllImport("libc", EntryPoint = "write", SetLastError = true)]
         private static unsafe extern IntPtr Write(int fileDescriptor, void* buf, IntPtr count);
 
+        [DllImport("libc", EntryPoint = "symlink", SetLastError = true)]
+        private static extern int CreateSymLink(string sourcePath, string destinationPath);
```
wilbaker: Naming suggestion: The meanings of `sourcePath` and `destinationPath` are a bit vague when it comes to symlinks.  Perhaps something like `linkTarget` and `newLinkPath` instead?

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -14,21 +14,35 @@ public class NativeUnixMethods
...
-                if (fileDescriptor != InvalidFileDescriptor)
+                if (mode == SymLinkMode)
                 {
-                    IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
-                    if (result.ToInt32() == -1)
+                    string linkLocation = Marshal.PtrToStringUTF8(new IntPtr(originalData));
+                    int result = CreateSymLink(linkLocation, destination);
+                    if (result == -1)
                     {
+                        tracer.RelatedInfo("Failed to create symlink({0},{1})", linkLocation, destination);
```
wilbaker: Why are we writing a message when `CreateSymLink` fails but not when `Open` or `Write` fail (below)?

Assuming this exception is caught higher up the stack and logged as an error you could include a custom message in the exception itself:

```
throw new Win32Exception(
    Marshal.GetLastWin32Error(),
    $""Failed to create symlink({linkLocation},{destination})");
```

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -14,21 +14,35 @@ public class NativeUnixMethods
...
+                    fileDescriptor = Open(destination, WriteOnly | Create | Truncate, mode);
+                    if (fileDescriptor != InvalidFileDescriptor)
+                    {
+                        IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
+                        if (result.ToInt32() == -1)
+                        {
+                            throw new Win32Exception(Marshal.GetLastWin32Error());
+                        }
+                    }
+                }
```
wilbaker: Also throw if `Open` fails? (A bit unrelated to the changes in this PR, but I just noticed that in the `NativeWindowsMethods` implementation we throw if `OpenForWrite` fails).

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -14,21 +14,35 @@ public class NativeUnixMethods
...
                 }
+                else
+                {
+                    fileDescriptor = Open(destination, WriteOnly | Create | Truncate, mode);
+                    if (fileDescriptor != InvalidFileDescriptor)
+                    {
+                        IntPtr result = Write(fileDescriptor, originalData, (IntPtr)originalSize);
+                        if (result.ToInt32() == -1)
+                        {
+                            throw new Win32Exception(Marshal.GetLastWin32Error());
```
wilbaker: Include a message in this `Win32Exception ` that it was the `Write` that failed?

GVFS/FastFetch/NativeUnixMethods.cs
nickgra: This was mostly for my local testing, but now that this is working, I like your suggestion and will use it all over this WriteFile method and will rework how the exceptions are worded.

GVFS/FastFetch/NativeUnixMethods.cs
```diff
@@ -35,7 +50,7 @@ public static unsafe void WriteFile(ITracer tracer, byte* originalData, long ori
                 EventMetadata metadata = new EventMetadata();
                 metadata.Add("destination", destination);
                 metadata.Add("exception", e.ToString());
```
wilbaker: Include `mode` in the metadata as well?

PR #649, opened 2018-12-20
clone/mount point env configurable

kivikakk:
Use `PATH_TO_MIRROR` and `TEST_ROOT` env vars to override paths in Linux MirrorProvider scripts.

/cc @chrisd8088 -- this is for our Docker build tooling

### DIFF
--- a/MirrorProvider/Scripts/Linux/MirrorProvider_Mount.sh
+++ b/MirrorProvider/Scripts/Linux/MirrorProvider_Mount.sh
@@ -8,6 +8,6 @@ fi
 SCRIPTDIR=$(dirname ${BASH_SOURCE[0]})
 BUILDDIR=$SCRIPTDIR/../../../../BuildOutput/MirrorProvider.Linux/bin/$CONFIGURATION/x64/netcoreapp2.1
 
-TEST_ROOT="${TEST_ROOT:=~/TestRoot}"
+TEST_ROOT="${TEST_ROOT:-~/TestRoot}"
 
 dotnet $BUILDDIR/MirrorProvider.Linux.dll mount "$TEST_ROOT"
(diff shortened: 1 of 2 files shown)


MirrorProvider/Scripts/Linux/MirrorProvider_Mount.sh
```diff
@@ -8,4 +8,6 @@ fi
 SCRIPTDIR=$(dirname ${BASH_SOURCE[0]})
 BUILDDIR=$SCRIPTDIR/../../../../BuildOutput/MirrorProvider.Linux/bin/$CONFIGURATION/x64/netcoreapp2.1
 
-dotnet $BUILDDIR/MirrorProvider.Linux.dll mount ~/TestRoot
+TEST_ROOT="${TEST_ROOT:=~/TestRoot}"
```
derrickstolee: TIL about this `:=` trick. Thanks!

wilbaker: This is new to me as well 😄 

After reading up on `:=`, is there a reason to prefer it to `:-`? Or would they be equivalent with how we're using them here?

chrisd8088: I suspect we could use `:-` here since we're assigning the output of the quote-delimited string back to `TEST_ROOT`, so `:=` actually performs an extra unused assignment *within* the `${...}` block.  Quick demo:
```
DASH="${TEST:-foo}"
echo $DASH $TEST
EQUAL="${TEST:=bar}"
echo $EQUAL $TEST
```
Output:
```
foo
bar bar
```

The `man bash` page is truly a treatise and there's always something in there I've never seen before, even though I've read it hundreds of times.  This week's learning experience for me was `set -u`, among several others.

kivikakk: I'm absolutely no bash wizard (and pulled this from some random internal script which is configured with env vars). Happy to defer to the experts!

kivikakk:
It looks like the build has hung on "Uninstall VFSForGit"; should take 1~2s, but in this case has been sitting for 20 minutes and looks set to continue this way. I'll just amend my latest commit (noop) and force push to retrigger builds. (Was low-key hoping I could do that in Azure DevOps myself somehow, but looks like not.)

jeschu1:
@kivikakk the builds are green now if you want to merge.

PR #655, opened 2018-12-26
[WIP] Dispose of unused LibGit2Repos

derrickstolee:
The LibGit2RepoPool stores a collection of LibGit2Repo objects,
which are our way of interacting with the native libgit2 library
to discover object existence and local blob sizes. We currently
create a list of these objects and don't dispose of them until we
unmount.

This has some strange side-effects:

1. libgit2 doesn't use the multi-pack-index, so it creates handles
   to all of the .idx files, and sometimes the .pack files.

2. Since libgit2 has hold of these handles, the packfile maintenance
   step cannot delete those files, and hence causes a problem when
   that runs in the background. We have ways to clean up afterwards.

There are potentially other problems with the approach, such as
"how does libgit2 re-check the list of packfiles?" It is possible
(I have yet to verify) that libgit2 is stuck with its initial list
of packfiles and hence a long-running mount process would actually
be missing a lot of objects that are otherwise present via prefetch
packs.

The solution presented here is to dispose of these objects on a
regular cadence, assuming no one has interacted with the pool in
a while. Using the Timer, we can adjust the dueTime to delay the
disposal action. We can now also add repos back to the pool as
necessary.

**WIP NOTES:** I made the timeouts be very short on purpose, so I can test in a way that I know we will hit it.


jamill:
Is there any context / background for the original reason for the libgit2 repo pool? For instance, how much overhead does pooling these objects save.

derrickstolee:
@jamill I have no idea. I wasn't around when this was made. I also think that the pools were not cleaned up because back then the mount process didn't last very long.

GVFS/GVFS.Common/Git/LibGit2Repo.cs
```diff
@@ -256,7 +228,7 @@ public static class Repo
                 [DllImport(Git2NativeLibName, EntryPoint = "git_repository_open")]
                 public static extern uint Open(out IntPtr repoHandle, string path);
 
-                [DllImport(Git2NativeLibName, EntryPoint = "git_tree_free")]
+                [DllImport(Git2NativeLibName, EntryPoint = "git_repository_free")]
```
derrickstolee: This is a very old bug. It wasn't actually freeing the repository handle, because the method only freed tree handles. Of course, we never ran this method unless we were killing the process, anyway.

However, it meant that the fix I was trying to do didn't actually work. (It does with this change.)

The way to test it is:

1. Have packfiles that will be expired during the packfile maintenance step.
2. Mount your repo, and hydrate a new file.
3. Wait for the libgit2 pool to dispose of all repos.
4. Run the packfile maintenance step.
5. Look at the `NumStaleIdxFiles` and `NumIdxDeletionsBlocked` values in the `PackfileMaintenanceStep_PerformMaintenance` log entry.

Before this change, `NumStaleIdxFiles` was zero and `NumIdxDeletionsBlocked` was growing. With this change, we could delete all stale files when the libgit2 repos are disposed.

If the packfile maintenance step runs soon after someone hydrated a file through ProjFS (different than a `git checkout` would fill a file), then we can still get blocked deletions. This could be an argument to make the disposal schedule more aggressive. We can quickly ramp up the pool as needed.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -9,32 +9,46 @@ namespace GVFS.Common
...
-        
+        private const int TryTakeWaitForever = Timeout.Infinite;
+        private const int TryTakeWaitShort = 10;
+
         private readonly BlockingCollection<LibGit2Repo> pool;
+        private readonly Func<LibGit2Repo> repoFactory;
         private readonly ITracer tracer;
 
+        private readonly Timer repoDisposalTimer;
+        private readonly TimeSpan repoDisposalDueTime = TimeSpan.FromSeconds(15);
```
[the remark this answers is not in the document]
derrickstolee: (My latest push comes up with potentially good numbers, but this should be a big focus for discussion.)

GVFS/GVFS.Common/Git/LibGit2Repo.cs
jamill: Good find! Unfortunate that this went so long without being found - but I guess we only attempted to call it as we were killing the process in any case.

Besides your manual steps, I couldn't really think of a good way to add automated tests that we are calling this function correctly. I suppose you could open a repository, and then free it, and then check that handles to pack files were closed as expected, but I am not sure that is worth the effort.

Anyway - If you wanted to get this in sooner, we could target this fix against master by itself.

> This could be an argument to make the disposal schedule more aggressive.

Should the pack file maintenance step coordinate with the libgit2 repo pool? Maybe the pack file maintenance could indicate that it wants to run and have the mount process to clear out the libgit2 repo pool. It could still fail (if there is ongoing filesystem activity that requires libgit2 repositories), but it would remove 1 potential failure mode (existing libgit2 repos from previous activity).

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -9,24 +9,38 @@ namespace GVFS.Common
...
+        private const int TryTakeWaitShort = 10;
 
         private readonly BlockingCollection<LibGit2Repo> pool;
+        private readonly Func<LibGit2Repo> repoFactory;
         private readonly ITracer tracer;
 
+        private readonly Timer repoDisposalTimer;
+        private readonly TimeSpan repoDisposalDueTime = TimeSpan.FromMinutes(1);
+        private readonly TimeSpan repoDisposalPeriod = TimeSpan.FromSeconds(15);
+        private int numDisposedRepos;
```
jamill: What is `numDisposedRepos` tracking? I see we increment it when disposing repositories, but also increment it at times as well.

Did you consider keeping the original size of the pool, and the current number of "open" repositories? or maybe naming this `numAvailableRepos` (vs `numActiveRepos`) to capture the idea that this is the number of repository objects that we can spin up?

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -9,24 +9,38 @@ namespace GVFS.Common
     public class LibGit2RepoPool
     {
         private const int TryAddTimeoutMilliseconds = 10;
-        private const int TryTakeTimeoutMilliseconds = Timeout.Infinite;
+        private const int TryTakeWaitForever = Timeout.Infinite;
+        private const int TryTakeWaitShort = 10;
 
         private readonly BlockingCollection<LibGit2Repo> pool;
+        private readonly Func<LibGit2Repo> repoFactory;
```
jamill: Maybe name the function with what it exactly - `OpenRepository`, `CreateLibGit2Repo`?

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -67,8 +81,27 @@ public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult
...
+                return repo;
+            }
+
+            if (this.numDisposedRepos > 0)
+            {
+                if (Interlocked.Decrement(ref this.numDisposedRepos) >= 0)
+                {
+                    return this.repoFactory();
+                }
+                else
```
jamill: Why do we increment the number of disposed repos here?

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -90,6 +125,20 @@ private void ReturnToPool(LibGit2Repo repo)
             }
         }
 
+        private void ResetTimer()
```
jamill: suggestion: could make function name less generic to include which timer this is resetting in the method name. Someone just looking at this method in the `LibGit2RepoPool` class might not know which timer it was referring to.

GVFS/GVFS.Common/Git/LibGit2Repo.cs
derrickstolee: The packfile maintenance step has some things already in place to mitigate when there are open handles. Specifically, it cleans up stale IDX files on its own, including those from previous steps.

However, maybe that stale-IDX logic should be in the prefetch code, as it already has a lot of logic to clean up back packs.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -9,24 +9,38 @@ namespace GVFS.Common
     public class LibGit2RepoPool
     {
         private const int TryAddTimeoutMilliseconds = 10;
-        private const int TryTakeTimeoutMilliseconds = Timeout.Infinite;
+        private const int TryTakeWaitForever = Timeout.Infinite;
+        private const int TryTakeWaitShort = 10;
```
jamill: The other names include the unit of measure, should this one keep that? `TryTakeWaitShortMilliseconds`?

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: The tricky part is that the size of the collection isn't important, as we remove repos in use from the collection. We can't track the size, but instead track how many we have disposed (which means how many we can reallocate).

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: If the `Interlocked.Decrement()` is below zero, it means that another thread decremented it to zero between the top `if` and the conclusion of the `Interlocked.Decrement()` statement. We need to increment the value back to ensure the value is correct (as we are not actually instantiating a new repo.

This is essentially what happens when our pool is _almost full_ and multiple requests for a repo happen at the same time.

GVFS/GVFS.Common/LibGit2RepoPool.cs
jamill: > which means how many we can reallocate

This is more what I meant. Instead of tracking how many repositories have disposed, should we rather track how many repositories we can still open? This might just be a name change, as you said this already tracks how many we can (re)allocate, but I think it would more accurately reflect the concept it is tracking and is a bit clearer.

As is, I did ask myself if we cared about the number of times we have disposed a repository, or if cared about the net number of disposed repositories vs the number we have re-allocated. 

> The tricky part is that the size of the collection isn't important

Just to be clear - what I meant was the maximum number of open repos, which we specify when creating the Pool (but we do not keep around). It seems like a pool has the following properties (which may be implicit):

1) The maximum number of open repositories
2) The current number of open repositories
3) The number of repositories that we can open

Some of these numbers can be calculated from the others. With this change, you want to track property 3.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -67,8 +81,27 @@ public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult
...
+            if (this.pool.TryTake(out repo, TryTakeWaitShort))
+            {
+                return repo;
+            }
+
+            if (this.numDisposedRepos > 0)
+            {
+                if (Interlocked.Decrement(ref this.numDisposedRepos) >= 0)
+                {
+                    return this.repoFactory();
```
jamill: Does this created repository actually get added to a pool? It seems like it should, but currently does not?

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: Actually, we should make the short wait be _no wait_ with 0 milliseconds, which means the units don't matter.

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: It seems to be purposefully generic, but I'll rename it to what the constructor uses as a parameter.

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: It's related to how we actually _remove_ from the pool to lend a repo to the caller. The caller must then return the repo to the pool later, which is when it is added to the pool.

GVFS/GVFS.Common/LibGit2RepoPool.cs
jamill: Ah - thanks for the explanation.

I think the edge condition logic is sufficiently subtle enough that the entire block could be encapsulated in its own method `TryCreateRepository(...)`, or at least a comment on the logic.

Did you consider using a [Semaphore](https://docs.microsoft.com/en-us/dotnet/api/system.threading.semaphore?view=netframework-4.7.2) instead? I think this construct has the behavior you want, while avoiding some of need to re-increment on boundary conditions.

Unresolving this just to bring it back... If you decide to not address this, please resolve it again.

GVFS/GVFS.Common/LibGit2RepoPool.cs
jamill: Thanks again for the explanation.

GVFS/GVFS.Common/LibGit2RepoPool.cs
jamill: Another possibility - just do the book keeping under a regular lock. I think it would be more clear than the special case handling of the edge condition...

jamill: Semaphore has a [WaitOne(..)](https://docs.microsoft.com/en-us/dotnet/api/system.threading.waithandle.waitone?view=netframework-4.7.2#System_Threading_WaitHandle_WaitOne_System_Int32_System_Boolean_) with timeout.

We would just need to lock around the increment / decrement of the number of available repositories to allocate. 

i.e. 
1) take lock
2) test if we have available repositories to allocate, if so then decrement. 
3) release lock
4) create repo if we had available repositories to allocate.

I think these approaches could make the logic simpler - but I also think we could validate correctness with tests. Are we planning on including tests?

derrickstolee: I believe this use of `Interlocked` is a standard way of dealing with resources, and [interlocked is less resource-intensive than `lock()`](https://stackoverflow.com/a/154803/127088).

Tests are included in the latest push.

GVFS/GVFS.UnitTests/Common/LibGit2RepoPoolTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            pool.TryInvoke(repo => true, out bool invoked);
+
+            invoked.ShouldBeTrue();
+            disposalTriggers.TryTake(out object _, millisecondsTimeout: 50).ShouldBeTrue();
+            disposalTriggers.TryTake(out object _, millisecondsTimeout: 0).ShouldBeFalse();
+        }
+
+        private class MockLibGit2Repo : LibGit2Repo
+        {
+            private readonly BlockingCollection<object> disposalTriggers;
```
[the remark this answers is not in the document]
derrickstolee: It also allows us to halt the test thread as we wait for the disposals to happen, then immediately continue it after they trigger.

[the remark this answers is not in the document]
jamill: Makes sense - thanks!

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -9,24 +9,53 @@ namespace GVFS.Common
...
+            {
+                this.repoDisposalDueTime = repoDisposalDueTime.Value;
+            }
+
+            if (repoDisposalPeriod.HasValue)
+            {
+                this.repoDisposalPeriod = repoDisposalPeriod.Value;
+            }
+
+            this.repoDisposalTimer = new Timer(
```
wilbaker: Timer implements `IDisposable`, dispose the timer in `Dispose()`?

derrickstolee:
@wilbaker: It's good to check what is the cost for disposing these repos. The biggest cost comes in the form of startup time when a request come in, as the _throughput_ will not be a huge problem (we don't dispose after every request, just after no requests after a while).

I created a maintenance step in my local install that runs this code:

```
        protected override void PerformMaintenance()
        {
            const int N = 1000;

            using (ITracer activity = this.Context.Tracer.StartActivity(this.Area + "_SameRepo", EventLevel.Verbose))
            {
                Stopwatch sw = Stopwatch.StartNew();
                using (LibGit2Repo repo = new LibGit2Repo(activity, this.Context.Enlistment.WorkingDirectoryRoot))
                {
                    for (int i = 0; i < N; i++)
                    {
                        repo.IsBlob("9d4f5e6f3b7079cd7f61aa9f639a9195a4e07336");
                    }
                }

                activity.RelatedInfo("Time for {0} IsBlob calls: {1} ms", N, sw.ElapsedMilliseconds);
            }

            using (ITracer activity = this.Context.Tracer.StartActivity(this.Area + "_NewRepos", EventLevel.Verbose))
            {
                Stopwatch sw = Stopwatch.StartNew();
                for (int i = 0; i < N; i++)
                {
                    using (LibGit2Repo repo = new LibGit2Repo(activity, this.Context.Enlistment.WorkingDirectoryRoot))
                    {
                        repo.IsBlob("9d4f5e6f3b7079cd7f61aa9f639a9195a4e07336");
                    }
                }

                activity.RelatedInfo("Time for {0} IsBlob calls: {1} ms", N, sw.ElapsedMilliseconds);
            }
        }
```
and here are some performance numbers (they alternate as re-used vs allocated new every time):
```
[2019-01-02 15:02:27 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 28 ms"}
[2019-01-02 15:02:29 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1666 ms"}
[2019-01-02 15:02:52 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 7 ms"}
[2019-01-02 15:02:54 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1671 ms"}
[2019-01-02 15:03:17 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 6 ms"}
[2019-01-02 15:03:19 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1593 ms"}
[2019-01-02 15:03:42 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 6 ms"}
[2019-01-02 15:03:44 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1528 ms"}
[2019-01-02 15:04:07 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 5 ms"}
[2019-01-02 15:04:09 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1573 ms"}
[2019-01-02 15:04:32 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 6 ms"}
[2019-01-02 15:04:34 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1594 ms"}
[2019-01-02 15:04:57 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 6 ms"}
[2019-01-02 15:04:59 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1619 ms"}
[2019-01-02 15:05:22 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 6 ms"}
[2019-01-02 15:05:24 -05:00] Information {"Message":"Time for 1000 IsBlob calls: 1521 ms"}
```
Roughly, this means that every time we have a "pool miss" we will get a 15-16ms delay on each request until the pool is full (assuming that we are receiving a lot of parallel requests, such as during a build).

Let me know if you want me to tweak the test here, as it would be easy for me to re-run the computation.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -90,6 +140,20 @@ private void ReturnToPool(LibGit2Repo repo)
...
+        private void ResetRepoDisposalTimer()
+        {
+            this.repoDisposalTimer.Change(this.repoDisposalDueTime, this.repoDisposalPeriod);
+        }
+
+        private void TryDropARepo()
+        {
+            if (this.pool.TryTake(out LibGit2Repo repo, TryTakeNoWait))
+            {
+                repo.Dispose();
```
wilbaker: What happens if the packfile maintenance step fails because there are still some repos in the pool that have not been disposed?  Will VFS4G wait for the entire packfile maintenance period before it tries again?

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -90,6 +140,20 @@ private void ReturnToPool(LibGit2Repo repo)
             }
         }
 
+        private void ResetRepoDisposalTimer()
+        {
+            this.repoDisposalTimer.Change(this.repoDisposalDueTime, this.repoDisposalPeriod);
+        }
+
+        private void TryDropARepo()
```
wilbaker: High level question on the approach being used in the PR.

Based on my understanding of the changes it looks like the approach is:

- Periodically dispose `LibGit2Repo` that are not in use
- Assuming there is not disk I/O that requires the use of any repos, eventually the active count will drop to 0
- If the active repo count is 0 when the packfile maintenance job runs it won't run into problems from open libgit2 handles

Is that correct?

If that is the case, what are your thoughts on a slightly different approach?

- Keep all `LibGit2Repo` repos alive until the packfile maintenance job is ready to run
- As a first step of the packfile maintenance job, have it call a `TryDropAllRepos` that does the same thing as `TryDropARepo` except it does it for all the repos in the pool.

A potential benefit to the second approach would be that we avoid the overhead you mentioned in a separate comment:

> The biggest cost comes in the form of startup time when a request come in, as the throughput will not be a huge problem (we don't dispose after every request, just after no requests after a while).
> ... Roughly, this means that every time we have a "pool miss" we will get a 15-16ms delay on each request until the pool is full (assuming that we are receiving a lot of parallel requests, such as during a build). 

If the packfile maintenance job is the only thing that needs the repos to be disposed we can keep all of them hot and ready to go until the the packfile maintenance job needs to run.

If there is a desire to have a minimum 15 second cool-off period before disposing the repos this approach could be adjusted so that `TryDropAllRepos` would check if all of the repos are in the pool and if it's been at least 15 seconds since the last repo was returned to the pool.

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: The step doesn't actually _fail_, but it leaves some .idx files around without corresponding .pack files. The next time it is run, it will clean up those files (assuming we had some period where the libgit2 repo was disposed and released the handle).

When we delete the .pack associated with the .idx, new libgit2 repos will not hold a handle to the .idx files because they will notice there isn't a paired .pack.

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: I'm pretty sure there are side benefits to this approach.

* This will free native memory that is accumulating in the libgit2 repos. We've definitely seen some OOMs in the wild, and this may be part of the cause.

* There is some startup cost to these long-lived repos when we hit an object they don't know about because new packfiles were added by prefetch. The libgit2 code needs to re-scan the pack directory anyway, so that presents a cost not measured by my test.

* The startup cost is limited to a batch of operations that happen within a window. The biggest reason we have a pool is for things like running a build. This may add a small static cost, but it is paid once at the beginning of a build and the continued file access should reduce the cost. If the build is very rarely hydrating a file, then we pay at most 15 ms per 15 minutes.

The points above are to (hopefully) justify clearing the pool regularly without too much of a cost. To add a condition that the packfile maintenance step doesn't run unless the pool is empty is a good idea, but the idea should be reversed: the step should ask if it is already empty (like how we check that there are no running Git processes), but not try to empty the pool itself. By testing if we should run every 6 hours, we will likely find a time where the pool is empty every day to run our step.

wilbaker: > - This will free native memory that is accumulating in the libgit2 repos
> - The libgit2 code needs to re-scan the pack directory anyway, so that presents a cost not measured by my test

Both good points 👍 

> If the build is very rarely hydrating a file, then we pay at most 15 ms per 15 minutes.

Did you mean per 15 seconds here? (15s is the period for disposing repos).

One other related question, now that we can fill the pool up on demand, should we still be starting with `size` active repos when `LibGit2RepoPool` is created?  Or should we start with a subset and create more as needed?

GVFS/GVFS.Common/LibGit2RepoPool.cs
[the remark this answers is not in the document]
derrickstolee: @wilbaker (You may have missed this comment because it is out-dated... I don't actually want these values to be so short. Instead, we should try to find out good values that are reasonable.)

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: **Sorry!** You probably didn't see [this comment](https://github.com/Microsoft/VFSForGit/pull/655#discussion_r244025234) about how the current values are super-short on purpose for testing. We should find good values that are less aggressive.

I was thinking a 15 _minute_ due time and a 1 _minute_ period. This means we would not remove a single repo from the pool until 15 minutes after the latest file hydration, and we would only drop to none in the pool after ~16 more minutes (my machine registers as 8 cores, so it loads 16 to the pool).

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -67,8 +96,27 @@ public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult
...
+                {
+                    return this.createRepo();
+                }
+                else
+                {
+                    Interlocked.Increment(ref this.numAvailableRepoAllocations);
+                }
+            }
+
+            if (this.pool.TryTake(out repo, TryTakeWaitForever))
```
wilbaker: It looks like we a possible race conditions here if `LibGit2RepoPool` were ever created with a size of 1.

As best I can tell we'll always have at least 2 today, but I wanted to mention it and get your thoughts as to whether we should fix it (or update the code to require a minimum size of 2).

Scenario:

1) Thread 1 calls `GetRepoFromPool` and `numAvailableRepoAllocations` is 0
2) Timer thread files, and `TryDropARepo` has not yet called `TryTake`
3) Thread 2 returns the only repo to the pool
4) Timer thread runs (resuming`TryDropARepo`) and it removes the only repo from the pool
5) Thread 1 is stuck here waiting on `TryTake`

For what it's worth I think we could theoretically hit a similar problem if `size` > 1, but that would only be if the thread pool were so backed up that multiple timer callbacks were getting queued up (which seems unlikely).

Thoughts?

wilbaker: It's probably worth mentioning that a similar race would be possible if `size` were greater than 1 and all of the repos are in use.  

But I think the difference would be that in step 5 (above) Thread 1 would block until another thread returns a repo to the pool rather than blocking indefinitely (unless perhaps `TryDropARepo` crossed perfectly with `ReturnToPool` each time).

derrickstolee: You are right that this seems problematic. We currently try to avoid the `TryDropARepo` close to other calls by updating the `dueTime` on the timer. (Keep in mind, running `ResetRepoDisposalTimer()` means we will not fire the timer call until it has not been called for `dueTime`.)

It seems we need a number of "waiting" threads, and that number needs to be zero before we drop a repo.

wilbaker: > Keep in mind, running ResetRepoDisposalTimer() means we will not fire the timer call until it has not been called for dueTime.

That's a good point, and @jamill pointed out that it's being called form the start of `GetRepoFromPool` as well.  

There still could be a (admittedly unlikely) race, but it would only be if the timer fired *just* before `ResetRepoDisposalTimer` adjusted it.

> It seems we need a number of "waiting" threads, and that number needs to be zero before we drop a repo.

That sounds good, a "waiting" count should prevent this scenario from happening.

derrickstolee:
The latest test failure on Mac is _probably_ due to not waiting long enough in the unit test. Will investigate in a new push after more comments come in.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -9,28 +9,49 @@ namespace GVFS.Common
...
+        public LibGit2RepoPool(
+            ITracer tracer,
+                Func<LibGit2Repo> createRepo,
+                TimeSpan? repoDisposalDueTime = null,
+                TimeSpan? repoDisposalPeriod = null)
         {
-            if (size <= 0)
+            this.createRepo = createRepo;
+            this.tracer = tracer;
+            this.pool = new BlockingCollection<LibGit2Repo>();
```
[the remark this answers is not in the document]
derrickstolee: We could limit the size, no problem. You're right that we want the pool to drop. There could also be limits to our native memory space.

One thing to think about: should we just set `TryAddTimeoutMilliseconds` to zero? It prevents a delay in our `TryInvoke()` with the small downside that we may dispose of repos a little more aggressively.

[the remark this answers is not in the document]
wilbaker: I think setting `TryAddTimeoutMilliseconds` to 0 makes sense.  The current size that we're passing into to pool's constructor is 2x the number of background threads that hydrate files, and so it seems like we won't frequently be exceeding the max anyway.

GVFS/GVFS.UnitTests/Common/LibGit2RepoPoolTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            TimeSpan period = TimeSpan.FromMilliseconds(1);
+
+            BlockingCollection<object> threadTriggers = new BlockingCollection<object>();
+            BlockingCollection<object> disposalTriggers = new BlockingCollection<object>();
+
+            using (LibGit2RepoPool pool = new LibGit2RepoPool(tracer, () => new MockLibGit2Repo(disposalTriggers), size, dueTime, period))
+            {
+                for (int i = 0; i < size; i++)
+                {
+                    new Thread(() => pool.TryInvoke(repo => threadTriggers.TryTake(out object _, 5000), out bool result)).Start();
```
derrickstolee: I was having trouble with the unit test here, but figured out how to ensure that we create and dispose of _exactly_ `size` repos. This loop starts the threads, and then we only allow one to finish at a time by adding items to the `threadTriggers` collection.

PR #658, opened 2019-01-02
Verify commit-graph and multi-pack-index after writing

derrickstolee:
Run the `git commit-graph verify` and `git multi-pack-index verify` commands after any command that would change those files. If these fail, then delete the corrupt file and rewrite.

We've had issues with users data in the past, and this gives us a way to automatically detect and repair these scenarios. The immediate rewrite _should_ work since we will regenerate from the other Git data. Issues we've seen in the past are related to trusting the content in these files and carrying that data forward to future versions of the file.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -412,6 +412,12 @@ namespace GVFS.Common.Git
                 });
         }
 
+        public Result VerifyCommitGraph(string objectDir)
+        {
+            string command = "commit-graph verify --object-dir \"" + objectDir + "\"";
+            return this.InvokeGitInWorkingDirectoryRoot(command, useReadObjectHook: true);
+        }
+
         public Result IndexPack(string packfilePath, string idxOutputPath)
         {
             return this.InvokeGitAgainstDotGitFolder($"index-pack -o \"{idxOutputPath}\" \"{packfilePath}\"");
@@ -426,6 +432,12 @@ namespace GVFS.Common.Git
         {
             // We override the config settings so we keep writing the MIDX file even if it is disabled for reads.
             return this.InvokeGitAgainstDotGitFolder("-c core.multiPackIndex=true multi-pack-index write --object-dir=\"" + objectDir + "\"");
+        }
+
+        public Result VerifyMultiPackIndex(string objectDir)
+        {
+            // We override the config settings so we keep writing the MIDX file even if it is disabled for reads.
+            return this.InvokeGitAgainstDotGitFolder("-c core.multiPackIndex=true multi-pack-index verify --object-dir=\"" + objectDir + "\"");
         }
 
         public Result RemoteAdd(string remoteName, string url)
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -436,7 +436,6 @@ namespace GVFS.Common.Git
 
         public Result VerifyMultiPackIndex(string objectDir)
         {
-            // We override the config settings so we keep writing the MIDX file even if it is disabled for reads.
             return this.InvokeGitAgainstDotGitFolder("-c core.multiPackIndex=true multi-pack-index verify --object-dir=\"" + objectDir + "\"");
         }
 
--- a/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
@@ -140,8 +140,35 @@ namespace GVFS.Common.Maintenance
                 GitProcess.Result expireResult = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot));
                 List<string> staleIdxFiles = this.CleanStaleIdxFiles(out int numDeletionBlocked);
                 this.GetPackFilesInfo(out int expireCount, out long expireSize, out hasKeep);
+
+                GitProcess.Result verifyResult1 = this.RunGitCommand((process) => process.VerifyMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                if (verifyResult1.ExitCodeIsFailure)
+                {
+                    EventMetadata errorMetadata = this.CreateEventMetadata();
+                    errorMetadata["MultiPackIndexVerifyOutput"] = verifyResult1.Output;
+                    errorMetadata["MultiPackIndexVerifyErrors"] = verifyResult1.Errors;
+                    string multiPackIndexPath = Path.Combine(this.Context.Enlistment.GitPackRoot, "multi-pack-index");
+                    errorMetadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(multiPackIndexPath);
+                    activity.RelatedError(errorMetadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
+
+                    this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                }
+
                 GitProcess.Result repackResult = this.RunGitCommand((process) => process.MultiPackIndexRepack(this.Context.Enlistment.GitObjectsRoot, this.batchSize));
                 this.GetPackFilesInfo(out int afterCount, out long afterSize, out hasKeep);
+
+                GitProcess.Result verifyResult2 = this.RunGitCommand((process) => process.VerifyMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                if (verifyResult1.ExitCodeIsFailure)
+                {
+                    EventMetadata errorMetadata = this.CreateEventMetadata();
+                    errorMetadata["MultiPackIndexVerifyOutput"] = verifyResult1.Output;
+                    errorMetadata["MultiPackIndexVerifyErrors"] = verifyResult1.Errors;
+                    string multiPackIndexPath = Path.Combine(this.Context.Enlistment.GitPackRoot, "multi-pack-index");
+                    errorMetadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(multiPackIndexPath);
+                    activity.RelatedError(errorMetadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
+
+                    this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                }
 
                 EventMetadata metadata = new EventMetadata();
                 metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
@@ -154,8 +181,10 @@ namespace GVFS.Common.Maintenance
                 metadata.Add(nameof(afterSize), afterSize);
                 metadata.Add("ExpireOutput", expireResult.Output);
                 metadata.Add("ExpireErrors", expireResult.Errors);
+                metadata.Add("VerifyResult1", verifyResult1.ExitCode);
                 metadata.Add("RepackOutput", repackResult.Output);
                 metadata.Add("RepackErrors", repackResult.Errors);
+                metadata.Add("VerifyResult2", verifyResult2.ExitCode);
                 metadata.Add("NumStaleIdxFiles", staleIdxFiles.Count);
                 metadata.Add("NumIdxDeletionsBlocked", numDeletionBlocked);
                 activity.RelatedEvent(EventLevel.Informational, $"{this.Area}_{nameof(this.PerformMaintenance)}", metadata, Keywords.Telemetry);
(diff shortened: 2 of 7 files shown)


derrickstolee:
@jrbriggs pointed out that this would be better to do after _every write_ to these files. Limits the window where there could be problems.

jamill:
> @jrbriggs pointed out that this would be better to do after every write to these files. Limits the window where there could be problems.

I like this suggestion. It also keeps our maintenance tasks more deterministic, in that it will run in response to generating these files, instead of at a different set of (seemingly random) times.

Do we have any ideas on what is causing the corruption? For instance, are we writing out corrupt files or, or is something else corrupting them after we generate them? If we are generating corrupt files, then this approach should let us identify it right away. If something else is corrupting them, then we might not pick it up as quickly...

derrickstolee:
> Do we have any ideas on what is causing the corruption? For instance, are we writing out corrupt files or, or is something else corrupting them after we generate them? If we are generating corrupt files, then this approach should let us identify it right away. If something else is corrupting them, then we might not pick it up as quickly...

In the cases where I can get in contact with the user, the only ways the file could have been corrupted were on-disk bit flips or in-memory bit flips. We literally were validating the data right before sending it (via a buffer) to disk. Those users that ran memcheck found issues with their RAM.

The perf machine recently had an issue that was caught by the `git multi-pack-index repack` command, but the problem may have existed much earlier. I'm not sure what caused it, but we could have noticed right away had we verified right afterward.

jrbriggs:
If corruption occurs strictly after a write (e.g., midx verify would have passed immediately after the write, but would have failed a day later), I don't think it's within the scope of this project to fix. If we can detect corrupt data _was written_ (e.g., a bit flip in memory was persisted to disk), we should mitigate.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -428,6 +434,12 @@ public Result WriteMultiPackIndex(string objectDir)
             return this.InvokeGitAgainstDotGitFolder("-c core.multiPackIndex=true multi-pack-index write --object-dir=\"" + objectDir + "\"");
         }
 
+        public Result VerifyMultiPackIndex(string objectDir)
+        {
+            // We override the config settings so we keep writing the MIDX file even if it is disabled for reads.
```
derrickstolee: comment is incorrect.

jeschu1:
@derrickstolee Is it safe to assume the verify / write should be quick tasks?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -139,9 +139,36 @@ protected override void PerformMaintenance()
 
                 GitProcess.Result expireResult = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot));
                 List<string> staleIdxFiles = this.CleanStaleIdxFiles(out int numDeletionBlocked);
-                this.GetPackFilesInfo(out int expireCount, out long expireSize, out hasKeep);
+                this.GetPackFilesInfo(out int expireCount, out long expireSize, out hasKeep);
+
+                GitProcess.Result verifyResult1 = this.RunGitCommand((process) => process.VerifyMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
```
jrbriggs: Maybe `verifyAfterExpireResult`?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -139,9 +139,36 @@ protected override void PerformMaintenance()
...
+                    activity.RelatedError(errorMetadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
+
+                    this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                }
+
                 GitProcess.Result repackResult = this.RunGitCommand((process) => process.MultiPackIndexRepack(this.Context.Enlistment.GitObjectsRoot, this.batchSize));
-                this.GetPackFilesInfo(out int afterCount, out long afterSize, out hasKeep);
+                this.GetPackFilesInfo(out int afterCount, out long afterSize, out hasKeep);
+
+                GitProcess.Result verifyResult2 = this.RunGitCommand((process) => process.VerifyMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
```
jrbriggs: `verifyAfterRepackResult`?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -139,9 +139,36 @@ protected override void PerformMaintenance()
...
+                if (verifyResult1.ExitCodeIsFailure)
+                {
+                    EventMetadata errorMetadata = this.CreateEventMetadata();
+                    errorMetadata["MultiPackIndexVerifyOutput"] = verifyResult1.Output;
+                    errorMetadata["MultiPackIndexVerifyErrors"] = verifyResult1.Errors;
+                    string multiPackIndexPath = Path.Combine(this.Context.Enlistment.GitPackRoot, "multi-pack-index");
+                    errorMetadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(multiPackIndexPath);
+                    activity.RelatedError(errorMetadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
+
+                    this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
```
jrbriggs: This doesn't end with a validate of what was written. Should it?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -139,9 +139,36 @@ protected override void PerformMaintenance()
...
+                }
+
                 GitProcess.Result repackResult = this.RunGitCommand((process) => process.MultiPackIndexRepack(this.Context.Enlistment.GitObjectsRoot, this.batchSize));
-                this.GetPackFilesInfo(out int afterCount, out long afterSize, out hasKeep);
+                this.GetPackFilesInfo(out int afterCount, out long afterSize, out hasKeep);
+
+                GitProcess.Result verifyResult2 = this.RunGitCommand((process) => process.VerifyMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                if (verifyResult1.ExitCodeIsFailure)
+                {
```
jeschu1: I think you mean verifyResult2.ExitCodeIsFailure :-).  Since the two failure check blocks are similar should we move them to a function?

GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
```diff
@@ -27,6 +27,20 @@ protected override void PerformMaintenance()
...
+                if (verifyResult.ExitCodeIsFailure)
+                {
+                    EventMetadata metadata = this.CreateEventMetadata();
+                    metadata["MultiPackIndexVerifyOutput"] = verifyResult.Output;
+                    metadata["MultiPackIndexVerifyErrors"] = verifyResult.Errors;
+                    string multiPackIndexPath = Path.Combine(this.Context.Enlistment.GitPackRoot, "multi-pack-index");
+                    metadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(multiPackIndexPath);
+                    activity.RelatedError(metadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
+
+                    this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
```
jrbriggs: Same comment about writing the midx without validating. This would still leave enlistments broken if it was corrupted. Deletion without rewrite would leave the enlistment functional, but slow. What are your thoughts on how we should handle this?

jeschu1: What if we were a big more aggressive and had WriteMultiPackIndex do the verification / retry.  This way we can be sure the Write/Verify are always paired together.  

I agree it seems like it would be good to know if the rewrite was successful.  Would it make sense to make the Verification  a warning if we succeed on the second try?

GVFS/GVFS.UnitTests/Maintenance/PostFetchStepTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using GVFS.UnitTests.Mock.FileSystem;
+using GVFS.UnitTests.Mock.Git;
+using NUnit.Framework;
+using System.Collections.Generic;
+
+namespace GVFS.UnitTests.Maintenance
+{
+    [TestFixture]
+    public class PostFetchStepTests
+    {
```
jeschu1: @derrickstolee these unit tests are fantastic!

GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
jrbriggs: I guess the primary question to answer is if we think regenerating the file will fix the corruption?

derrickstolee: The examples we've seen all have the problem that the multi-pack-index or commit-graph files were corrupted, and the problem persisted only because we continued to trust that data when we write a new one. Deleting the file and rewriting (based on the .pack and .idx files) has always fixed the issue.

If the rewrite did not fix the issue, then we would see failures from the same enlistment multiple times in a row. I plan to set alerts for any user hitting this even once, so we would be ready to intervene if this happens repeatedly.

derrickstolee:
> @derrickstolee Is it safe to assume the verify / write should be quick tasks?

They take about as long as it would take to write them from scratch. I've measured 11-16 seconds.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -154,14 +167,28 @@ protected override void PerformMaintenance()
                 metadata.Add(nameof(afterSize), afterSize);
                 metadata.Add("ExpireOutput", expireResult.Output);
                 metadata.Add("ExpireErrors", expireResult.Errors);
+                metadata.Add("verifyAfterExpire", verifyAfterExpire.ExitCode);
```
jrbriggs: nit: capitalize for consistency, and maybe name this and `VerifyAfterRepack` `Verify*ExitCode`

GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
jrbriggs: Makes sense. So we expect the validate/rewrite cycle to fix the issue in all cases we have seen so far? With monitors in place, I'm happy with this pattern.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -154,14 +167,28 @@ protected override void PerformMaintenance()
...
+                metadata.Add("VerifyAfterRepack", verifyAfterRepack.ExitCode);
                 metadata.Add("NumStaleIdxFiles", staleIdxFiles.Count);
                 metadata.Add("NumIdxDeletionsBlocked", numDeletionBlocked);
                 activity.RelatedEvent(EventLevel.Informational, $"{this.Area}_{nameof(this.PerformMaintenance)}", metadata, Keywords.Telemetry);
 
                 this.SaveLastRunTimeToFile();
             }
         }
+
+        private void LogErrorAndRewrite(ITracer activity, GitProcess.Result result)
```
jeschu1: If we move this function up a level could PostFetchStep.cs use it?

GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
```diff
@@ -27,6 +27,13 @@ protected override void PerformMaintenance()
                 this.Context.FileSystem.TryDeleteFile(multiPackIndexLockPath);
 
                 this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+
+                GitProcess.Result verifyResult = this.RunGitCommand((process) => process.VerifyMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+
```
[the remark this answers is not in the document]
derrickstolee: _whitespace is your friend..._ (Mostly this is minor enough to not merit running the builds again.) Thanks for the detailed review!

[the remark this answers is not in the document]
jeschu1: It is minor, I just noticed the inconsistency here versus PackfileMaintenanceStep.cs.  I wouldn't hold up the build for it thought.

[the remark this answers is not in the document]
derrickstolee: well, I got a functional test failure (unrelated... creating an issue now) so I may as well do this to trigger new builds.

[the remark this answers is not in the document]
derrickstolee: Filed #663.

PR #661, opened 2019-01-03
Protoype: UpgradeOrchestrator tests with generalized Upgrader

jamill:
This is a spike on how we can test the `UpgradeOrchestrator` component with an arbitrary `ProductUpgrader`. It includes a skeleton of scenarios that we could test, along with a couple of examples of how we can test different scenarios. This also leverages the Moq framework (included as an example). The general testing outline does not depend on Moq, but we wanted to include an example of tests using that framework in this prototype. This PR shows examples of using Moq both with Interfaces as well as concrete classes.

### DIFF
--- a/GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
+++ b/GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
@@ -94,6 +94,7 @@
     <Compile Include="Windows\Service\RepoRegistryTests.cs" />
     <Compile Include="Windows\Upgrader\ProductUpgraderTests.cs" />
     <Compile Include="Windows\Upgrader\UpgradeOrchestratorTests.cs" />
+    <Compile Include="Windows\Upgrader\UpgradeOrchestratorUnitTests.cs" />
     <Compile Include="Windows\Upgrader\UpgradeTests.cs" />
     <Compile Include="Windows\Upgrader\UpgradeVerbTests.cs" />
     <Compile Include="Windows\Virtualization\ActiveEnumerationTests.cs" />
--- a/GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
+++ b/GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
@@ -35,8 +35,14 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
```diff
@@ -35,9 +35,15 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Castle.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc, processorArchitecture=MSIL">
```
derrickstolee: (just making a second note on this `Castle.Core` reference)

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorUnitTests.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.IO;
+
+namespace GVFS.UnitTests.Upgrader
+{
+    [TestFixture]
+    public class UpgradeOrchestratorUnitTests
+    {
+        private const string OlderThanLocalVersion = "1.0.17000.1";
+        private const string LocalGVFSVersion = "1.0.18115.1";
+        private const string NewerThanLocalVersion = "1.1.18115.1";
```
derrickstolee: Should `NewerThanLocalVersion` be different than `LocalGVFSVersion`?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorUnitTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        private Mock<IProductUpgrader> MockUpgrader { get; set; }
+
+        private UpgradeOrchestrator orchestrator { get; set; }
+
+        [SetUp]
+        public void Setup()
+        {
+            string message = string.Empty;
+
+            this.MockTracer = new Mock<ITracer>();
```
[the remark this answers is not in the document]
jamill: Yeah - good point. If we already have test objects that work for us, we do not need to generate a `Moq` version. This should probably use the existing `MockTracer` here.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorUnitTests.cs
jamill: that is a bit subtle... Maybe we can make the version numbers more distinct

PR #662, opened 2019-01-03
Update Git to include multi-pack-index expire fix and trace2 v3

derrickstolee:
* [PR 107 in microsoft/git](https://github.com/Microsoft/git/pull/107) includes a fix for the way we permute the set of packs during a `git multi-pack-index expire` command. Specifically, this fixes an issue when we expire packs and add packs to the multi-pack-index at the same time (and the order of those packs).

* [PR 100 in microsoft/git](https://github.com/Microsoft/git/pull/100) includes updates to trace2.


wilbaker:
Also run the full suite of functional tests before merging?

PR #664, opened 2019-01-03
Upgrade Microsoft.Data.Sqlite to prevent System.InvalidOperationException

### LINKED ISSUE #273: Upgrade Microsoft.Data.Sqlite to prevent System.InvalidOperationException: SqliteConnection does not support nested transactions.
The issue described here:

https://github.com/aspnet/Microsoft.Data.Sqlite/issues/532

could be causing the occasional SQLite related exception we've been seeing:

```
System.InvalidOperationException: SqliteConnection does not support nested transactions.
   at Microsoft.Data.Sqlite.SqliteConnection.BeginTransaction(IsolationLevel isolationLevel)
   at GVFS.Virtualization.BlobSize.BlobSizes.BlobSizesDatabaseWriter.TryAddSizes(ConcurrentQueue`1 sizes, Int32& errorCode, String& error)
   at GVFS.Virtualization.BlobSize.BlobSizes.FlushDbThreadMain()\",\"ErrorMessage\":\"FlushDbThreadMain caught unhandled exception, exiting process\"}"
```

If there is an internal rollback without the [fix mentioned above](https://github.com/aspnet/Microsoft.Data.Sqlite/issues/532) `SqliteTransaction.Complete()` will *not* be called leaving a non-null `Transaction` set on `BlobSizesDatabaseWriter.connection`

We should pick up the fix and see if the issue persists.

jeschu1:
This should hopefully resolve #273

The failure is pretty infrequent so I've never been able to repro but we see instances in Kusto.

### DIFF
--- a/GVFS/GVFS.FunctionalTests.Windows/GVFS.FunctionalTests.Windows.csproj
+++ b/GVFS/GVFS.FunctionalTests.Windows/GVFS.FunctionalTests.Windows.csproj
@@ -55,8 +55,8 @@
       <HintPath>..\..\..\packages\Microsoft.Database.Isam.1.9.4\lib\net40\Esent.Isam.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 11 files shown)


jrbriggs:
Closing this for now since it's not under active development

PR #666, opened 2019-01-04
Cleanups to upgrade feature

derrickstolee:
I wasn't paying close attention to the upgrade feature, so I checked out the feature branch and went through the code file-by-file. Instead of commenting the style fixes I'd like, I went and made them in commits.


jamill:
I would fix the following issue (after incorporating suggestions):

```
Windows\Upgrader\UpgradeTests.cs(34,21): error CS1729: 'GitHubUpgrader.GitHubUpgraderConfig' does not contain a constructor that takes 2 arguments [C:\agent\_work\9\s\GVFS\GVFS.UnitTests.Windows\GVFS.UnitTests.Windows.csproj]

Issue #385, opened 2018-10-15
GVFS upgrade should be cross platform


jrbriggs:
The initial implementation of upgrade was done for Windows only. We need to make this work on Mac as well.

PR #674, opened 2019-01-07
Nuget upgrader

jamill:
This change is a work in progress, but the general pieces are in place. I am cleaning up the code, adding tests, and documenting the details.

## Goal:
Implement an upgrade mechanism that makes the upgrade logic more flexible to enable customized steps to be run during an upgrade. Organizations might have specific requirements and need custom steps to deploy VFSForGit upgrades.

This introduces the NuGet upgrader, which allows for generalized upgrade packages. The upgrade package is distributed via NuGet feeds as a NuGet package. The package includes details on how to install itself, allowing organizations the ability to control the VFSForGit upgrade.

An overview of the process is described in #547 

### TODOs:
- [x] Tighten rules around loading NuGetUpgrader
  - [x] Add another property to indicate that we should load NuGet upgrader?
- [x] Tighten up documentation on ReleaseManifest JSON
- [x] Inspect interaction with NuGetProtocol (basically the details of NuGetFeedWrapper)
- [x] Flesh out unit tests
  - [x] ReleaseManifest
  - [x] NuGetUpgrader
- [ ] ~~Verify NuGet package~~ This will be deferred into a follow-up PR
- [x] Add additional tracing
### Notes

The following is a class diagram of the components involved in this change:

nugetupgrader-classdiagram


GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -25,44 +25,26 @@ public class GitHubUpgrader : IProductUpgrader
...
 
-        public GitHubUpgrader(string currentVersion, ITracer tracer)
-        {
             this.installedVersion = new Version(currentVersion);
             this.fileSystem = new PhysicalFileSystem();
             this.tracer = tracer;
 
-            string upgradesDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
-            this.fileSystem.CreateDirectory(upgradesDirectoryPath);
+            this.localUpgradeServices = new LocalUpgraderServices(tracer);
```
derrickstolee: You create a new `PhysicalFileSystem` above, but then also create a new one inside `LocalUpgraderServices()`. Perhaps supply one in the constructor?

GVFS/GVFS.Common/LocalUpgraderServices.cs
```diff
@@ -0,0 +1,152 @@
...
+            return Path.Combine(
+                ProductUpgraderInfo.GetUpgradesDirectoryPath(),
+                "InstallerTemp");
+        }
+
+        public static bool TryCreateDirectory(string path, out Exception exception)
+        {
+            try
+            {
+                Directory.CreateDirectory(path);
```
derrickstolee: why `static` and not using `this.fileSystem`?

GVFS/GVFS.Common/LocalUpgraderServices.cs
```diff
@@ -0,0 +1,152 @@
...
+        // upgrade installer that is downloaded and run will fail. This is because it cannot overwrite
+        // C:\ProgramFiles\GVFS\GVFS.Upgrader.exe that is running. Moving GVFS.Upgrader.exe along with
+        // its dependencies to a temporary location inside ProgramData and running GVFS.Upgrader.exe
+        // from this temporary location helps avoid this problem.
+        public virtual bool TrySetupToolsDirectory(out string upgraderToolPath, out string error)
+        {
+            string rootDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
+            string toolsDirectoryPath = Path.Combine(rootDirectoryPath, ToolsDirectory);
+            Exception exception;
+            if (TryCreateDirectory(toolsDirectoryPath, out exception))
```
derrickstolee: I usually prefer

```
if (!TryThing(out Exception exception))
{
    respond to error
    return error
}

do normal thing
return success
```

GVFS/GVFS.Common/LocalUpgraderServices.cs
derrickstolee: Hm... you have a non-static version lower in this file.

GVFS/GVFS.Common/LocalUpgraderServices.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                string currentPath = ProcessHelper.GetCurrentProcessLocation();
+                error = null;
```
derrickstolee: See [this discussion](https://github.com/Microsoft/VFSForGit/pull/675/files/7ecafcfa101197c323607e748611fa610fa014fc#r246064887) about setting out-params by default at the beginning (tl;dr don't. let the compiler help you).

GVFS/GVFS.Common/NuGetUpgrader/InstallManifest.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+
+namespace GVFS.Common
```
derrickstolee: Should this all be in `GVFS.Common` or something like `GVFS.Common.Upgrade`?

derrickstolee: And for file locations: Should we have an `Upgrade` folder with a `NuGet` folder inside?

(I see that we have a lot of upgrader code in the base of the project, which I wish I noticed earlier. It would be better to group those things into a folder.)

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+using System.IO.Compression;
+using System.Threading;
+
+namespace GVFS.Common
+{
+    public class NuGetUpgrader : IProductUpgrader
+    {
+        private static readonly string GitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+
+        private ITracer tracer;
```
derrickstolee: VS says this isn't used... it probably should be.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+                return false;
+            }
+
+            message = null;
+            return true;
+        }
+
+        public bool TryQueryNewestVersion(out Version newVersion, out string message)
+        {
+            newVersion = null;
```
derrickstolee: Delay setting out-params.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+            public string FeedUrl { get; private set; }
+            public string PackageFeedName { get; private set; }
+            public string FeedUrlForCredentials { get; private set; }
+            private ITracer Tracer { get; set; }
+            private LocalGVFSConfig LocalConfig { get; set; }
+
+            public bool TryLoad(out bool isEnabled, out bool isConfigured, out string error)
+            {
+                error = string.Empty;
+                isEnabled = false;
```
derrickstolee: It's ok that you are setting `error` here, as you are appending to it later. However, `isEnabled` and `isConfigured` are set unconditionally further down, so do not need to be set here.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+                installActionWrapper(
+                    () =>
+                    {
+                        string installerPath = Path.Combine(extractedPackagePath, "content", entry.RelativePath);
+                        this.localUpgradeServices.RunInstaller(installerPath, entry.Args, out installerExitCode, out localError);
+
+                        installSuccesesfull = installerExitCode == 0;
+
+                        // Just for initial experiment to make sure each step is displayed in UI
+                        Thread.Sleep(5 * 1000);
```
derrickstolee: Please remove before merging.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+                new NuGetFeedWrapper(config.FeedUrl, config.PackageFeedName, downloadFolder, personalAccessToken, tracer),
+                new LocalUpgraderServices(tracer))
+        {
+        }
+
+        public NuGetUpgrader(
+            string currentVersion,
+            ITracer tracer,
+            NugetUpgraderConfig config,
+            string downloadFolder,
```
derrickstolee: `downloadFolder` and `personalAccessToken` are not used.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+
+        public static IProductUpgrader Create(
+            ITracer tracer,
+            out string error)
+        {
+            NugetUpgraderConfig upgraderConfig = new NugetUpgraderConfig(tracer, new LocalGVFSConfig());
+            bool isConfigured;
+            bool isEnabled;
+
+            if (!upgraderConfig.TryLoad(out isEnabled, out isConfigured, out error))
```
derrickstolee: Should we use the results of `isEnabled` or `isConfigured` somehow?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+
+                if (highestVersion != null &&
+                    highestVersion.Identity.Version.Version > this.installedVersion)
+                {
+                    this.latestVersion = highestVersion;
+                }
+
+                newVersion = this.latestVersion?.Identity?.Version?.Version;
+
+                if (!(newVersion is null))
```
derrickstolee: `if (newVersion != null)`

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+                {
+                    message = $"New version {highestVersion.Identity.Version} is available.";
+                    return true;
+                }
+                else if (!(highestVersion is null))
```
derrickstolee: `if (highestVersion != null)`

GVFS/GVFS.Common/NuGetUpgrader/NuGetWrapper.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Handles interactions with a NuGet Feed.
+    /// </summary>
+    public class NuGetFeedWrapper
```
derrickstolee: Class doesn't match file name.

GVFS/GVFS.Common/NuGetUpgrader/NuGetWrapper.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        private ITracer tracer;
```
derrickstolee: Why the newlines between the members?

GVFS/GVFS.Common/ProductUpgraderFactory.cs
```diff
@@ -13,6 +13,12 @@ public class ProductUpgraderFactory
             ITracer tracer,
             out string error)
         {
+            newUpgrader = NuGetUpgrader.Create(tracer, out error);
```
derrickstolee: `newUpgrader = NuGetUpgrader.Create(tracer, out error) ?? GitHubUpgrader.Create(tracer, out error);` maybe?

GVFS/GVFS.UnitTests/Common/NuGetUpgraderTests.cs
```diff
@@ -0,0 +1,210 @@
...
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class NuGetUpgraderTests
```
derrickstolee: This class is in `GVFS.UnitTests`, but the tests from #675 are in `GVFS.UnitTests.Windows`. Should the `GitHubUpgrader` be tested in the universal-platform project?

jamill: Good question - currently, `UpgradeOrchestrator.cs` lives in the `GVFS.Upgrader` project, which is a .NET Framework project. As such, all tests that use `UpgradeOrchestrator` need to live in `GVFS.UnitTests.Windows`.

However, the actual upgraders live in `GVFS.Common`, which exists as netcore project, so they can be tested in `GVFS.UnitTests`.

So - yes, they should be in a universal-platform project (eventually).

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+                ProcessHelper.GetCurrentProcessVersion(),
+                tracer,
+                upgraderConfig,
+                ProductUpgraderInfo.GetAssetDownloadsPath(),
+                token);
+
+            return upgrader;
+        }
+
+        public bool UpgradeAllowed(out string message)
```
derrickstolee: This method does not appear to be covered by the unit tests.

derrickstolee: Neither is `Create()`, but since it is static it is harder to unit test.

Consider removing the static method in favor of some kind of factory pattern? (It is called by the static `ProductUpgraderFactory.TryCreateUpgrader()`, which I mention in #675 could be replaced with a factory instance.)

GVFS/GVFS.Common/LocalUpgraderServices.cs
```diff
@@ -0,0 +1,152 @@
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common
+{
+    public class LocalUpgraderServices
```
wilbaker: Naming suggestion:  From the name of this class it wasn't clear to me if `LocalUpgraderServices` was related to Windows services.

Maybe something like `CommonUpgraderTasks`?  Or move the code to a common base class?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
jamill: what is your reasoning here?

GVFS/GVFS.Common/NuGetUpgrader/NuGetWrapper.cs
jamill: I wasn't really happy "Wrapper" - I will go with `NuGetFeed`.

GVFS/GVFS.Common/LocalUpgraderServices.cs
jamill: I will think about the name some more - Instead of a "Common" class, I wanted a class that provided all of the local operations that upgraders depend on... the idea with the current name was that it provided local / client  "services" that upgraders use.

GVFS/GVFS.Common/NuGetUpgrader/ManifestEntry.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        public string Args { get; set; }
+
+        /// <summary>
+        /// User friendly name for the install action
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// The relative path to the install command
```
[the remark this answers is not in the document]
jamill: This is the path to the installer / script relative to a "root". Currently, the root is the content directory of the Nuget Package, as that is where the "custom" folder structure is rooted at. I will try to clarify in the comment.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+using NuGet.Protocol.Core.Types;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Threading;
+
+namespace GVFS.Common
+{
+    public class NuGetUpgrader : IProductUpgrader
```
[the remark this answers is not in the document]
jamill: I think this is fair feedback on the high level design, and is something that @kewillford and I also discussed. I think the `IProductUpgrader` interface we have now is an evolution from where the code started from, and it enables us to meet our primary goal (enable upgrade with different upgrade mechanism). I think we could decompose this further - but is a tradeoff between extra layers of abstraction and implementation. As you note - I think the class would pretty cleanly separate into two more focused classes.

I will think about this some more and at least do a bit more to see exactly what this would look like.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
jamill: Yeah - I have a TODO on my local notes to add more tracing. I added it to the PR Description TODO list as well.

GVFS/GVFS.Common/NuGetUpgrader/InstallManifest.cs
jamill: The namespace has been updated to match the folder structure.

The upgrader files could probably be more organized. Nesting too deeply comes with its own set of downsides. Maybe a good compromise would be 3 top level folders: `Upgrader`, `GitHubUpgrader`, and `NuGetUprader`, or maybe a single top level directory of `Upgrader`, or maybe we do add two levels of nesting...

GVFS/GVFS.Common/NuGetUpgrader/ReleaseManifest.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections.Generic;
+
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Details on the upgrade included in this package, including information
+    /// on what packages are included and how to install them.
+    /// </summary>
+    public abstract class ReleaseManifest
```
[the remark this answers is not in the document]
jamill: There is not a difference between upgrades and releases in the manifest... All of the contents of a manifest seem to be related to a particular release and how to install a particular release, rather than any logic for an "upgrade" (e.g. it doesn't check previous versions, etc...).

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
wilbaker: Out of curiosity is there a difference between `== null` and `is null`? 

I have not seen `is` used with `null` before.

jamill: > Out of curiosity is there a difference between == null and is null?

They can have different behavior depending on whether the equality operator is overloaded or not. The `is null` syntax makes it explicit what the code is doing. When using the equality operator, then it depends on what the equality operator does. There could also be a slight performance advantage to using the `is null` syntax when the equality operator is overloaded - it would depend on the implementation of the equality operator for that type.

Overall - I wouldn't expect an actual / noticeable difference here, but I think `is null` is the more technically appropriate syntax (at the cost of a bit more awkward syntax).

GVFS/GVFS.Common/LocalUpgraderServices.cs
jamill: This method was moved from `GitHubUpgrader` originally, but I like the suggested pattern and it was an easy enough change - thanks!

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
wilbaker: My preference is for the syntax that @derrickstolee suggested `if (newVersion != null)`.

That's the syntax we're using elsewhere in the codebase and I personally find it easier to follow.

GVFS/GVFS.Common/NuGetUpgrader/ManifestEntry.cs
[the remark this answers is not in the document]
wilbaker: Ahh okay, in that case even a small change to something like:

```
The relative path to the install script (or binary)
```

Would make the comment easier to follow

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <returns>List of packages that match query parameters</returns>
+        public virtual async Task<IList<IPackageSearchMetadata>> QueryFeed(string packageId)
+        {
+            SourceRepository sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
+            if (!string.IsNullOrEmpty(this.personalAccessToken))
+            {
+                sourceRepository.PackageSource.Credentials = this.Credentials;
+            }
+
+            var packageMetadataResource = await sourceRepository.GetResourceAsync<PackageMetadataResource>();
```
wilbaker: Out of curiosity, is there a non-async version of `GetResource`?

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            public void LogDebug(string data)
+            {
+                this.tracer.RelatedInfo($"NuGetPackageUpgrader (Debug): {data}");
+            }
+
+            public void LogError(string data)
+            {
+                this.tracer.RelatedInfo($"NuGetPackageUpgrader (Error): {data}");
```
wilbaker: Why are all of the Nuget log messages being mapped to `Info`? Should some any of them be mapped to warning or error?

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,169 @@
...
+            SourceRepository sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
+            if (!string.IsNullOrEmpty(this.personalAccessToken))
+            {
+                sourceRepository.PackageSource.Credentials = this.Credentials;
+            }
+
+            var packageMetadataResource = await sourceRepository.GetResourceAsync<PackageMetadataResource>();
+            var cacheContext = new SourceCacheContext();
+            cacheContext.DirectDownload = true;
+            cacheContext.NoCache = true;
```
wilbaker: What does the `NoCache` setting control?

GVFS/GVFS.Common/NuGetUpgrader/ReleaseManifest.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace GVFS.Common.NuGetUpgrader
+{
+    /// <summary>
+    /// Details on the upgrade included in this package, including information
+    /// on what packages are included and how to install them.
+    /// </summary>
+    public class ReleaseManifest
+    {
+        public static string WindowsPlatformKey = "Windows";
```
wilbaker: I only see this being used in the unit tests, can it be removed from this class?

jamill: There was a place in product code that used a string literal for this value - I updated that location to use this field (and turned it into a const value).

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -141,6 +122,10 @@ public bool TryDownloadNewestVersion(out string errorMessage)
         {
             bool downloadedGit = false;
             bool downloadedGVFS = false;
+
+            string upgradesDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
+            this.fileSystem.CreateDirectory(upgradesDirectoryPath);
```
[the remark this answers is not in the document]
jamill: Converted to using the `TryCreateDictionary` to match error handling, and tweaked error case to trace the actual error.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,376 @@
...
+                error = e.Message;
+                return false;
+            }
+
+            string extractedPackagePath = this.UnzipPackageToTempLocation();
+            this.releaseManifest = ReleaseManifest.FromJsonFile(Path.Combine(extractedPackagePath, "content", "install-manifest.json"));
+
+            this.fileSystem.CreateDirectory(upgradesDirectoryPath);
+
+            foreach (ManifestEntry entry in this.releaseManifest.PlatformInstallManifests[ReleaseManifest.WindowsPlatformKey].InstallActions)
```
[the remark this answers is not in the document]
jamill: I hardened the code to fail gracefully if there expected platform is not present. 

The manifest is structured to support multiple platforms. My thinking was that the install actions for multiple platforms could reference the same exe / script, but that each platform would need to specify all of the install actions. That is, there will not be a common install action section (there could be, if we found it valuable, but I wasn't going to include that initially).

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,376 @@
...
+        private static bool TryGetPersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer, out string token, out string error)
+        {
+            GitProcess gitProcess = new GitProcess(gitBinaryPath, null, null);
+            return gitProcess.TryGetCredentials(tracer, credentialUrl, out string username, out token, out error);
+        }
+
+        private string UnzipPackageToTempLocation()
+        {
+            string extractedPackagePath = this.localUpgradeServices.TempPath;
+            ZipFile.ExtractToDirectory(this.downloadedPackagePath, extractedPackagePath);
```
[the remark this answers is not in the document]
jamill: It is caught in `TryRunInstaller` (the calling code), and converted into an error flag.

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -260,6 +260,52 @@ public bool TryCopyToTempFileAndRename(string sourcePath, string destinationPath
...
+            try
+            {
+                Directory.CreateDirectory(path);
+            }
+            catch (IOException e)
+            {
+                exception = e;
+                return false;
+            }
+            catch (UnauthorizedAccessException e)
```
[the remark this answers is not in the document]
jamill: The general pattern is to catch the expected exceptions only. I tweaked the logic to catch the expected exceptions from the underlying method.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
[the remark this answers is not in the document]
jamill: I am going to resolve this conversation for now. I think what we have now, while it could be tweaked further, is good for our current goals. I will mark this a potential future improvement, though.

GVFS/GVFS.Common/LocalUpgraderServices.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                UpgraderToolName,
+                UpgraderToolConfigFile,
+                "GVFS.Common.dll",
+                "GVFS.Platform.Windows.dll",
+                "Microsoft.Diagnostics.Tracing.EventSource.dll",
+                "netstandard.dll",
+                "System.Net.Http.dll",
+                "Newtonsoft.Json.dll"
+            };
```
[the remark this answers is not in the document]
jamill: Let's punt on this for now and handle with #385. If there are blocking issues, then that would be good to raise them now. Please let me know if you disagree.

GVFS/GVFS.Common/NuGetUpgrader/InstallManifestPlatform.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class InstallManifestPlatform
+    {
+        public InstallManifestPlatform()
+        {
+            this.InstallActions = new List<ManifestEntry>();
+        }
+
+        public InstallManifestPlatform(IEnumerable<ManifestEntry> entries)
+        {
+            this.InstallActions = entries?.ToList();
```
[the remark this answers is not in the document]
alameenshah: Naming nit. Couple points. 1. The class is named `InstallManifestPlatform`, but at many places where it is used, the instance is named `platformInstallManifest`. This makes it look a little twisted - is this a platform or a manifest. 2. The word platform is also bit loaded with Mac and Windows specific code grouped into `GVFSPlatform` class, would be nice if we could find an alternate name. Also the class is used to hold a list of Manifest entries so it looks neither like a platform nor a manifest but a data source for manifests. 

One more question is do we need a container class, just to hold a list of manifest entries? I mean does having a separate class help in parsing, mocking or UT?

GVFS/GVFS.Common/NuGetUpgrader/InstallManifestPlatform.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace GVFS.Common.NuGetUpgrader
+{
+    public class InstallManifestPlatform
```
alameenshah: Naming nit. Couple points. 1. The class is named InstallManifestPlatform, but at many places where it is used, the instance is named platformInstallManifest. This makes it look a little twisted - is this a platform or a manifest. 2. The word platform is also bit loaded with Mac and Windows specific code grouped into GVFSPlatform class, would be nice if we could find an alternate name. Also the class is used to hold a list of Manifest entries so it looks neither like a platform nor a manifest but a data source for manifests.

One more question is do we need a container class, just to hold a list of manifest entries? I mean does having a separate class help in parsing, mocking or UT?

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,226 @@
...
+            }
+
+            return downloadPath;
+        }
+
+        /// <summary>
+        /// Implementation of logger used by NuGet library. It takes all output
+        /// and redirects it to the GVFS logger.
+        /// </summary>
+        public class Logger : ILogger
```
alameenshah: Should `Logger` class be moved out into GVFS.Common? If this is going to be there only temporarily during testing, then its ok.

jamill: This is the logger for NuGet (and is a nested class inside of NuGetFeed) - it has to be accessible for interacting with the NuGet APIs, so I kept it with this class.

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
jamill: There are some non-async versions of these methods, but not all. Looking at the code, the non-async versions seem to call through to the async versions. I think `GetResource` might have a non-async version, but `GetDownloadResulAsync` does not, for instance.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,368 @@
...
+            {
+                errorMessage = ex.Message;
+                return false;
+            }
+
+            errorMessage = null;
+            return true;
+        }
+
+        public bool TryCleanup(out string error)
```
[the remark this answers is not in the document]
jamill: It is not the most obvious at this point, but one method is meant to clean up after the installer runs, while the other is meant to cleanup the packages that `GVFS.Service` downloads in the background...

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+        public bool UpgradeAllowed(out string message)
+        {
+            if (string.IsNullOrEmpty(this.nugetUpgraderConfig.FeedUrl))
+            {
+                message = "Nuget Feed URL has not been configured";
+                return false;
+            }
+            else if (string.IsNullOrEmpty(this.nugetUpgraderConfig.PackageFeedName))
+            {
+                message = "URL to lookup credentials has not been configured";
```
wilbaker: Is this a copy paste error? 

I noticed this is the same message that we return in the ` else if (string.IsNullOrEmpty(this.nugetUpgraderConfig.FeedUrlForCredentials))`case.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+            }
+
+            return false;
+        }
+
+        public bool TryDownloadNewestVersion(out string errorMessage)
+        {
+            if (this.latestVersion == null)
+            {
+                errorMessage = "No new version to download. Query for latest version to ensure a new version is available before downloading.";
```
wilbaker: Is this a user facing error message?  If so, what action would\should a user take if they see this message?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                return false;
+            }
+
+            using (ITracer activity = this.tracer.StartActivity(nameof(this.TryDownloadNewestVersion), EventLevel.Informational))
```
wilbaker: This is not currently going to be reported via telemetry, just wanted to confirm if that is intentional.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+            {
+                try
+                {
+                    this.DownloadedPackagePath = this.nuGetFeed.DownloadPackage(this.latestVersion.Identity).GetAwaiter().GetResult();
+                }
+                catch (Exception ex)
+                {
+                    activity.RelatedError($"{nameof(this.TryDownloadNewestVersion)} - error encountered: ${ex.Message}");
+                    errorMessage = ex.Message;
```
wilbaker: Also log exception call stack?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+        public bool TryCleanup(out string error)
+        {
+            error = null;
+            Exception e;
+            bool success = this.fileSystem.TryDeleteDirectory(this.localUpgradeServices.TempPath, out e);
+
+            if (!success)
+            {
+                error = e?.Message;
+                this.tracer.RelatedError($"{nameof(this.TryCleanup)} - Error encountered: {error}");
```
wilbaker: Trace full exception stack?

wilbaker: (applies in a few spots in this class)

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                    Exception e;
+                    if (!this.fileSystem.TryDeleteDirectory(this.localUpgradeServices.TempPath, out e))
+                    {
+                        error = e?.Message;
+                        return false;
+                    }
+
+                    string extractedPackagePath = this.UnzipPackageToTempLocation();
+                    this.releaseManifest = ReleaseManifest.FromJsonFile(Path.Combine(extractedPackagePath, "content", "install-manifest.json"));
+                    InstallManifestPlatform platformInstallManifest = this.releaseManifest.PlatformInstallManifests[platformKey];
```
wilbaker: Use `TryGetValue` instead of `[platformKey]`? 

It looks like `Dictionary<TKey,TValue>.Item[TKey]` will throw a `KeyNotFoundException` if the key does not exist (https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.item?view=netframework-4.6.2) and so we won't ever hit the `if (platformInstallManifest == null)` case.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                        return false;
+                    }
+
+                    activity.RelatedInfo($"Extracted ReleaseManifest from JSON. InstallActions: {platformInstallManifest.InstallActions.Count}");
+
+                    this.fileSystem.CreateDirectory(upgradesDirectoryPath);
+
+                    foreach (ManifestEntry entry in platformInstallManifest.InstallActions)
+                    {
+                        string installerPath = Path.Combine(extractedPackagePath, "content", entry.InstallerRelativePath);
```
wilbaker: Minor: Use a named constant for `"content"`

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                            () =>
+                            {
+                                if (!this.dryRun)
+                                {
+                                    this.localUpgradeServices.RunInstaller(installerPath, entry.Args, out installerExitCode, out localError);
+                                }
+                                else
+                                {
+                                    // This is a temporary workaround to force the step to be written
+                                    // out to the console.
```
wilbaker: Should this temporary workaround be removed?  If not, what is the plan for removing it?

jamill: There are two reasons:

1. Without any delay, these actions are fast...
2. The logic to display messages to the console only displays messages if the action takes > 2 seconds. So, the install step would not be printed to the console.

Before we remove this, we need a way to have ensure the messages are printed to console for each step (otherwise, it is hard to see what steps are run...). I think we might even want this assurance for installation steps in general - we would want to display every installation step to the user.

I was going to look at a way to force the message to be displayed, even if the action takes < 2s. Or maybe this is already possible, but I overlooked it.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                this.FeedUrl = configValue;
+
+                bool credentialURLAvailable = false;
+                if (this.localConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedCredentialUrl, out configValue, out readError))
+                {
+                    credentialURLAvailable = !string.IsNullOrWhiteSpace(configValue);
+                }
+                else
+                {
+                    this.tracer.RelatedError(readError);
```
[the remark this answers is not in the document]
jamill: Yeah... the code tries to read all config values before returning, so it can build a more accurate message. At the end of the method, we can see that this method only returns true if *all* config entries have values. It is a bit verbose - would a comment up front help?

GVFS/GVFS.Common/NuGetUpgrader/InstallManifestPlatform.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace GVFS.Common.NuGetUpgrader
+{
+    public class InstallManifestPlatform
```
kewillford: I would remove this class because it is not giving any value right now.  If in the future we need something that does more we can add it back in.

GVFS/GVFS.UnitTests/Common/NuGetUpgraderTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            Version newVersion;
+            string message;
+            List<IPackageSearchMetadata> availablePackages = new List<IPackageSearchMetadata>()
+            {
+                this.GeneratePackageSeachMetadata(new Version(CurrentVersion)),
+                this.GeneratePackageSeachMetadata(new Version(NewerVersion)),
+                this.GeneratePackageSeachMetadata(new Version(NewerVersion2)),
+            };
+
+            this.mockNuGetFeed.Setup(foo => foo.QueryFeedAsync(It.IsAny<string>())).ReturnsAsync(new List<IPackageSearchMetadata>(availablePackages));
```
mjcheetham: Why are we essentially returning a `new List<..>(new List<..> { ..elements..}))`?

Could just do

```suggestion
            this.mockNuGetFeed.Setup(foo => foo.QueryFeedAsync(It.IsAny<string>())).ReturnsAsync(availablePackages);
```

Same comment applies elsewhere too.

GVFS/GVFS.Common/NuGetUpgrader/ManifestEntry.cs
```diff
@@ -0,0 +1,34 @@
+namespace GVFS.Common.NuGetUpgrader
+{
+    public class ManifestEntry
```
kewillford: This seems a little too generic for me.  I would name it something like `InstallerEntry`, `InstallerData`, `InstallerInfo`?

GVFS/GVFS.UnitTests/Common/NuGetUpgraderTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            {
+                this.GeneratePackageSeachMetadata(new Version(CurrentVersion)),
+                this.GeneratePackageSeachMetadata(new Version(NewerVersion)),
+            };
+
+            IPackageSearchMetadata newestAvailableVersion = availablePackages.Last();
+
+            string downloadPath = "c:\\test_download_path";
+            this.mockNuGetFeed.Setup(foo => foo.QueryFeedAsync(NuGetFeedName)).ReturnsAsync(new List<IPackageSearchMetadata>(availablePackages));
+            this.mockNuGetFeed.Setup(foo => foo.DownloadPackage(It.Is<PackageIdentity>(packageIdentity => packageIdentity == newestAvailableVersion.Identity))).Returns(Task.FromResult(downloadPath));
```
mjcheetham: `Returns(Task.FromResult(..))` can be replaced with `ReturnsAsync(..)`

GVFS/GVFS.UnitTests/Common/NuGetUpgraderTests.cs
```diff
@@ -0,0 +1,300 @@
...
+
+            bool success = this.upgrader.TryQueryNewestVersion(out actualNewestVersion, out message);
+
+            // Assert that no new version was returned
+            success.ShouldBeTrue($"Expecting TryQueryNewestVersion to have completed sucessfully. Error: {message}");
+            actualNewestVersion.ShouldEqual(newestAvailableVersion.Identity.Version.Version, "Actual new version does not match expected new version.");
+
+            bool downloadSuccessful = this.upgrader.TryDownloadNewestVersion(out message);
```
mjcheetham: What (if anything) can we assert about the `message`?

GVFS/GVFS.UnitTests/Common/NuGetUpgraderTests.cs
```diff
@@ -0,0 +1,300 @@
...
+             nuGetUpgrader = new NuGetUpgrader(
+                CurrentVersion,
+                this.tracer,
+                nuGetUpgraderConfig,
+                false,
+                this.mockFileSystem.Object,
+                this.mockNuGetFeed.Object,
+                new LocalUpgraderServices(this.tracer, this.mockFileSystem.Object));
+
+            nuGetUpgrader.UpgradeAllowed(out string _).ShouldBeFalse("Upgrade without FeedURL configured should not be allowed.");
```
mjcheetham: I'm surprised our StyleCop rules allow discards (`out string _`). Does it also permit type-less discards too?
I.e., just `out _`.

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,224 @@
...
+            this.nuGetLogger = new Logger(this.tracer);
+
+            this.sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
+            if (!string.IsNullOrEmpty(this.personalAccessToken))
+            {
+                this.sourceRepository.PackageSource.Credentials = this.Credentials;
+            }
+        }
+
+        private NuGet.Configuration.PackageSourceCredential Credentials
```
kewillford: Why don't we have a using for `Nuget.Configuration`?

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,224 @@
...
+            private ITracer tracer;
+
+            public Logger(ITracer tracer)
+            {
+                this.tracer = tracer;
+            }
+
+            public void Log(LogLevel level, string data)
+            {
+                string message = $"NuGet Logger: ({level}): {data}";
```
kewillford: Should the `data` be put in an `EventMetadata` field?  I don't think it matters too much.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
wilbaker: I think if you want to stick with the error message approach it could be simplified to: ""No new version to download" with a comment that provides the additional details.

If this will *only* fail due to programmer error than something like an InvalidOperationException is an option too.

jamill: @mjcheetham I think we agree on what the high level behavior is for this scenario (i.e. logging and stopping the operation). I don't know if introducing a second error handling pattern at this point is the way to achieve that, given our current patterns. An assert would achieve much of the same effect (crash in debug builds, handled more gracefully in product builds).

I am open to opinions on this though...

wilbaker: If a failure here results in an obvious failure for the user I'm not sure that the assert would add much value.  I think we could just stick with the error message.  (Either way if a dev sees it they'd investigate)

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
kewillford: Isn't there an option to change the delay to `0`?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+            public string FeedUrlForCredentials { get; private set; }
+
+            public bool TryLoad(out bool isEnabled, out bool isCorrectlyConfigured, out string error)
+            {
+                error = string.Empty;
+
+                string configValue;
+                string readError;
+
+                bool feedURLAvailable = false;
```
kewillford: `Available` especially when using URL to me means that the URL it up and you can make requests and get responses from it but I don't think that's what it means in this context right?  It only means we could get a string from the config?  Maybe it should be `haveFeedUrl`?

GVFS/GVFS.Common/ProductUpgraderFactory.cs
```diff
@@ -15,6 +16,12 @@ public class ProductUpgraderFactory
             bool dryRun = false,
             bool noVerify = false)
         {
+            newUpgrader = NuGetUpgrader.NuGetUpgrader.Create(tracer, dryRun, noVerify, out error);
+            if (newUpgrader != null)
```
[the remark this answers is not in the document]
kewillford: Same questions.  Why is the NuGetUpgrader first instead of the GitHubUpgrader?

GVFS/GVFS.Common/IProductUpgrader.cs
```diff
@@ -12,7 +12,7 @@ namespace GVFS.Common
     /// <returns>success or failure return from the method run.</returns>
     public delegate bool InstallActionWrapper(Func<bool> method, string message);
 
-    public interface IProductUpgrader
+    public interface IProductUpgrader : IDisposable
```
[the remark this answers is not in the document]
jamill: It is getting called in [UpgradeOrchestrator.cs](https://github.com/Microsoft/VFSForGit/pull/674/files#diff-ae998894e77b3b0dc63da871281616eaR121)

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
wilbaker: > Why not just let exceptions with appropriate error messages or types flow up the call stack, and then catch them, trace them and present the human actionable message at a very high level?

We've had a few reasons for generally preferring the TryXXX pattern (@kewillford might remember some reasons):

- Control flow via TryXXX is faster than exceptions
- We've needed to know exactly when\where a failures occur to know whether it's safe or VFS4G to continue.  If multiple methods can throw the same type exception and they're called in a try { } block it's difficult to know which failed (unless each call has its own try { } block).
- Catching exceptions at a high level was making it difficult to diagnose user issues because the logs would be missing details from the lower parts of the callstack.  The TryXXX approach allows for tracing the relevant parameters/details for the specific call that failed.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -376,10 +317,7 @@ protected virtual bool TryFetchReleases(out List<Release> releases, out string e
 
         protected virtual void RunInstaller(string path, string args, out int exitCode, out string error)
```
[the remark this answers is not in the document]
jamill: The GitHubUpgrader tests depend on overriding this method. It was not part of this original change, but we could come back and clean up some of these tests (either in a later PR or after I have addressed other feedback).

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,224 @@
...
+            ITracer tracer)
+        {
+            this.feedUrl = feedUrl;
+            this.feedName = feedName;
+            this.downloadFolder = downloadFolder;
+            this.personalAccessToken = personalAccessToken;
+            this.tracer = tracer;
+
+            // Configure the NuGet SourceCacheContext -
+            // - Direct download packages - do not download
```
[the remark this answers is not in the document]
jamill: It is set in `NullSourceCacheContext.Instance` I will update the comment to make this more clear.

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        /// <summary>
+        /// Download the specified packageId from the NuGet feed.
+        /// </summary>
+        /// <param name="packageId">PackageIdentity to download.</param>
+        /// <returns>Path to the downloaded package.</returns>
+        public virtual async Task<string> DownloadPackage(PackageIdentity packageId)
+        {
+            string downloadPath = Path.Combine(this.downloadFolder, $"{this.feedName}.zip");
```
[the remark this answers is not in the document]
jamill: Interesting thought. For now, I think it is ok for us to require feed names to be valid file names. I will double check on this.

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,224 @@
...
+                    includeUnlisted: true,
+                    sourceCacheContext: this.sourceCacheContext,
+                    log: this.nuGetLogger,
+                    token: CancellationToken.None);
+
+                return queryResults.ToList();
+            }
+            catch (Exception ex)
+            {
+                throw new Exception($"Failed to query the NuGet package feed due to error: {ex.Message}", ex);
```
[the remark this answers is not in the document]
jamill: yes - in some early testing, the error message we encountered if the user couldn't be authenticated was not very informative. This UX could still be tweaked some, but with this at least we know higher level operation failed...

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Threading;
+
+namespace GVFS.Common.NuGetUpgrader
+{
+    public class NuGetUpgrader : IProductUpgrader
+    {
+        private static readonly string GitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
```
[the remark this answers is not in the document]
jamill: It should be... but I can move this into the constructor.

GVFS/GVFS.Common/ProductUpgraderFactory.cs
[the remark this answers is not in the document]
jamill: > Why is the NuGetUpgrader

For this iteration - we will prefer to use the NuGetUpgrader if it is "enabled" (that is, if any of the config entries for NuGet Upgrader are populated, then we will assume the user means to use NuGet Upgrader instead of GitHub Upgrader).

As a followup item, we might refine this logic to having the user explicitly opt-in to NuGetUpgrader, possibly via the `upgrade.ring` setting (e.g. you would set upgrade.ring to NuGetUpgrader to get NuGet functionality).

> If Create could possibly return null and also has out string error then should it not be TryCreate

I could - the downside is yet another `out` variable, when we could go with the pattern that it will return a new object, or else return null + error string....

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -76,42 +76,49 @@ public void Execute()
...
+                }
+
+                if (this.ExitCode == ReturnCode.GenericError)
+                {
+                    error = Environment.NewLine + "ERROR: " + error;
+                    this.output.WriteLine(error);
+                }
+                else
+                {
+                    if (newVersion != null)
```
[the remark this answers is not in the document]
jamill: I can see why this might be a bit hard to follow - 

1) We are using the presence of `newVersion` to indicate whether upgrade was run or not
2) The string that is generated does included an embedded ternary operator that takes into account the `mountError`

However, the actual change being made here is to encapsulate this logic in a try/finally block to run dispose (you can confirm by ignoring whitespace differences). I think we should make these suggested cleanups in a separate PR (I can submit one that can be handled separately...).

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+
+            message = null;
+            return true;
+        }
+
+        public bool TryQueryNewestVersion(out Version newVersion, out string message)
+        {
+            try
+            {
+                IList<IPackageSearchMetadata> queryResults = this.nuGetFeed.QueryFeedAsync(this.nugetUpgraderConfig.PackageFeedName).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
jamill: Only the NuGetUpgrader version is asynchronous, the original method GitHubUpgrader is not. I would prefer to keep to the original contract, rather than make the original version match an async API. As none of our APIs are Async, I didn't see an advantage to making this expose an async implementation - thoughts?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                if (newVersion != null)
+                {
+                    this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - new version available: installedVersion: {this.installedVersion}, latestAvailableVersion: {highestVersion}");
+                    message = $"New version {highestVersion.Identity.Version} is available.";
+                    return true;
+                }
+                else if (highestVersion != null)
+                {
+                    this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - up-to-date");
+                    message = $"Latest available version is {highestVersion.Identity.Version}, you are up-to-date";
```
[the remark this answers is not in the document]
jamill: highest and latest are synonyms here... I will go through and consistently use `highest`, as we are using the version number.

GVFS/GVFS.Common/NuGetUpgrader/ManifestEntry.cs
jamill: `InstallActionInfo`?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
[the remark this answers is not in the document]
jamill: The NuGet config load code has been updated - hopefully it is now more clear...

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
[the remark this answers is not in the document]
jamill: Updated code to just query static instance on demand.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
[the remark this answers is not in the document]
kewillford: Sorry, what I meant was why does the `NuGetFeed` expose an async method instead of it having a synchronous call and hide that the NuGet API is async instead of here.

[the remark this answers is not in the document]
jamill: ah - I see. That is another option. I chose the NuGetUpgrader as the layer to absorb that detail, as it is the orchestrating these calls and is the main API that consumers will consume at this point.

GVFS/GVFS.UnitTests/Common/ReleaseManifestTests.cs
```diff
@@ -0,0 +1,117 @@
...
+@"
+{
+  ""Version"" : ""1"",
+  ""PlatformInstallManifests"" : {
+    ""Windows"": {
+      ""InstallActions"": [
+        {
+ 	    ""Name"" : ""Git"",
+        ""Version"" : ""2.19.0.1.34"",
+        ""InstallerRelativePath"" : ""Installers\\Windows\\G4W\\Git-2.19.0.gvfs.1.34.gc7fb556-64-bit.exe"",
```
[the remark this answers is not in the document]
jamill: Yes, that is what I would expect.

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,466 @@
...
+                new NuGetFeed(
+                    config.FeedUrl,
+                    config.PackageFeedName,
+                    downloadFolder,
+                    personalAccessToken,
+                    tracer))
+        {
+        }
+
+        public NuGetUpgrader(
```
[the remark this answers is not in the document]
jamill: I made it private for now... we can look at improvements in a follow up PR

GVFS/GVFS.Common/NuGetUpgrader/ReleaseManifest.cs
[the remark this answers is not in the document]
jamill: What do you think of `InstallManifest`?

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,531 @@
...
+
+        public override bool TryCleanup(out string error)
+        {
+            error = null;
+            Exception e;
+            bool success = this.fileSystem.TryDeleteDirectory(this.ExtractedInstallerPath, out e);
+
+            if (!success)
+            {
+                error = e?.Message;
```
wilbaker: Should an error message still be set when `e` is `null`?

wilbaker: (Same comment elsewhere in this class where we have `error = e?.Message`)

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -0,0 +1,531 @@
...
+                                    // the message for this install
+                                    // action is written to the
+                                    // console.  Even though the
+                                    // message is written with a delay
+                                    // of 0, the messages are not
+                                    // always written out.  If / when
+                                    // we can ensure that the message
+                                    // is written out to console, then
+                                    // we can remove this sleep.
+                                    Thread.Sleep(1500);
```
wilbaker: File a work item for this?  Assuming this is an issue with the console spinner I think all we'd need to do is add a reset even that ensures the `action()` isn't run until the initial message has been printed.

GVFS/GVFS.Common/LocalUpgraderServices.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            else
+            {
+                // Encountered error - do not set upgrade tool path and return failure.
+                upgraderToolPath = null;
+                return false;
+            }
+        }
+
+        public virtual void RunInstaller(string path, string args, out int exitCode, out string error)
```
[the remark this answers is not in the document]
jamill: Latest version has reworked this logic, so this is not needed in current version

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            this.TraceException(this.tracer, exception, method, message);
+        }
+
+        public void TraceException(ITracer tracer, Exception exception, string method, string message)
+        {
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("Method", method);
+            metadata.Add("Exception", exception.ToString());
+            tracer.RelatedError(metadata, message, Keywords.Telemetry);
```
wilbaker: It's not hurting anything by including it, but all `ReleatedError(...)` calls are recorded to telemetry automatically:

```
        public virtual void RelatedError(EventMetadata metadata, string message)
        {
            this.RelatedError(metadata, message, Keywords.Telemetry);
        }
```

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -0,0 +1,226 @@
...
+            this.sourceRepository = null;
+            this.sourceCacheContext?.Dispose();
+        }
+
+        /// <summary>
+        /// Query a NuGet feed for list of packages that match the packageId.
+        /// </summary>
+        /// <param name="packageId"></param>
+        /// <returns>List of packages that match query parameters</returns>
+        public virtual async Task<IList<IPackageSearchMetadata>> QueryFeedAsync(string packageId)
```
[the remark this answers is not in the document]
jamill: This needs to be public for Moq

PR #675, opened 2019-01-07
New UpgradeOrchestrator UT

alameenshah:
### Goal of the new TestFixture
- Verify UpgradeOrchestrator works works as expected with the new IProductUpgrader interface. 
- Test and also demonstrate state transitions inside UpgradeOrchestrator.

### Implementation details
This PR is based on this [prototype PR](https://github.com/Microsoft/VFSForGit/pull/661). At a high level, it creates a Moq Mock instance of IProductUpgrader interface. Initializes UpgradeOrchestrator with the Moq upgrader instance. Then drives the orchestrator through simulated error scenarios and verifies it behaves as expected.

### DIFF
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
@@ -1,0 +1,243 @@
+﻿using GVFS.Common;
+using GVFS.Common.Git;
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Mock.Common;
+using GVFS.UnitTests.Windows.Mock.Upgrader;
+using GVFS.Upgrader;
+using Moq;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+
+namespace GVFS.UnitTests.Upgrader
+{
+    [TestFixture]
+    public class UpgradeOrchestratorTests
+    {
+        private const string GVFSVersion = "1.1.18115.1";
+
+        private delegate void TryGetNewerVersionCallback(out System.Version version, out string message);
+        private delegate void UpgradeAllowedCallback(out string message);
+        private delegate void TryRunPreUpgradeChecksCallback(out string delegateMessage);
+        private delegate void TryDownloadNewestVersionCallback(out string message);
+        private delegate void TryRunInstallerCallback(InstallActionWrapper installActionWrapper, out string error);
+
+        private MockTracer Tracer { get; set; }
+        private MockTextWriter Output { get; set; }
+        private MockInstallerPrerunChecker PreRunChecker { get; set; }
+        private Mock<LocalGVFSConfig> MoqLocalConfig { get; set; }
+        private Mock<IProductUpgrader> MoqUpgrader { get; set; }
+
+        private UpgradeOrchestrator orchestrator { get; set; }
+
+        [SetUp]
+        public void Setup()
+        {
+            this.Tracer = new MockTracer();
+            this.Output = new MockTextWriter();
+            this.PreRunChecker = new MockInstallerPrerunChecker(this.Tracer);
+            this.PreRunChecker.Reset();
+            this.MoqUpgrader = this.DefaultUpgrader();
+            this.orchestrator = new UpgradeOrchestrator(
+                this.MoqUpgrader.Object,
+                this.Tracer,
+                this.PreRunChecker,
+                input: null,
+                output: this.Output);
+
+            this.SetUpgradeAvailable(new Version(GVFSVersion), error: null);
+        }
+
+        [TestCase]
+        public void ExecuteSucceedsWhenUpgradeAvailable()
+        {
+            this.orchestrator.Execute();
+
+            this.VerifyOrchestratorInvokes(
+                upgradeAllowed: true,
+                queryNewestVersion: true,
+                downloadNewestVersion: true,
+                installNewestVersion: true,
+                cleanup: true);
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.Success);
+        }
+
+        [TestCase]
+        public void ExecuteSucceedsWhenOnLatestVersion()
+        {
+            this.SetUpgradeAvailable(newVersion: null, error: null);
+
+            this.orchestrator.Execute();
+
+            this.VerifyOrchestratorInvokes(
+                upgradeAllowed: true,
+                queryNewestVersion: true,
+                downloadNewestVersion: false,
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.Success);
+        }
+
+        [TestCase]
+        public void ExecuteFailsWhenGetNewVersionFails()
+        {
+            string errorMessage = "Authentication error.";
+
+            this.SetUpgradeAvailable(newVersion: null, error: errorMessage);
+
+            this.orchestrator.Execute();
+
+            this.VerifyOrchestratorInvokes(
+                upgradeAllowed: true,
+                queryNewestVersion: true,
+                downloadNewestVersion: false,
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.VerifyOutput("ERROR: Authentication error.");
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.GenericError);
+        }
+
+        [TestCase]
+        public void ExecuteFailsWhenPrecheckFails()
+        {
+            this.PreRunChecker.SetReturnFalseOnCheck(MockInstallerPrerunChecker.FailOnCheckType.IsElevated);
+
+            this.orchestrator.Execute();
+
+            this.VerifyOrchestratorInvokes(
+                upgradeAllowed: true,
+                queryNewestVersion: false,
+                downloadNewestVersion: false,
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.GenericError);
+        }
+
+        [TestCase]
+        public void ExecuteFailsWhenDownloadFails()
+        {
+            this.MoqUpgrader.Setup(upgrader => upgrader.TryDownloadNewestVersion(out It.Ref<string>.IsAny))
+                .Callback(new TryDownloadNewestVersionCallback((out string delegateMessage) =>
+                {
+                    delegateMessage = "Download error.";
+                }))
+                .Returns(false);
+
+            this.orchestrator.Execute();
+
+            this.VerifyOrchestratorInvokes(
+                upgradeAllowed: true,
+                queryNewestVersion: true,
+                downloadNewestVersion: true,
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.VerifyOutput("ERROR: Download error.");
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.GenericError);
+        }
+
+        [TestCase]
+        public void ExecuteFailsWhenRunInstallerFails()
+        {
+            this.MoqUpgrader.Setup(upgrader => upgrader.TryRunInstaller(It.IsAny<InstallActionWrapper>(), out It.Ref<string>.IsAny))
+                .Callback(new TryRunInstallerCallback((InstallActionWrapper installActionWrapper, out string delegateMessage) =>
+                {
+                    delegateMessage = "Installer error.";
+                }))
+                .Returns(false);
+
+            this.orchestrator.Execute();
+
+            this.VerifyOrchestratorInvokes(
+                upgradeAllowed: true,
+                queryNewestVersion: true,
+                downloadNewestVersion: true,
+                installNewestVersion: true,
+                cleanup: true);
+
+            this.VerifyOutput("ERROR: Installer error.");
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.GenericError);
+        }
+
+        public Mock<IProductUpgrader> DefaultUpgrader()
+        {
+            Mock<IProductUpgrader> mockUpgrader = new Mock<IProductUpgrader>();
+
+            mockUpgrader.Setup(upgrader => upgrader.UpgradeAllowed(out It.Ref<string>.IsAny))
+                .Callback(new UpgradeAllowedCallback((out string delegateMessage) =>
+                {
+                    delegateMessage = string.Empty;
+                }))
+                .Returns(true);
+
+            string message = string.Empty;
+            mockUpgrader.Setup(upgrader => upgrader.TryDownloadNewestVersion(out It.Ref<string>.IsAny)).Returns(true);
+            mockUpgrader.Setup(upgrader => upgrader.TryRunInstaller(It.IsAny<InstallActionWrapper>(), out message)).Returns(true);
+            mockUpgrader.Setup(upgrader => upgrader.TryCleanup(out It.Ref<string>.IsAny)).Returns(true);
+
+            return mockUpgrader;
+        }
+
+        public void SetUpgradeAvailable(Version newVersion, string error)
+        {
+            bool upgradeResult = string.IsNullOrEmpty(error);
+
+            this.MoqUpgrader.Setup(upgrader => upgrader.TryQueryNewestVersion(out It.Ref<System.Version>.IsAny, out It.Ref<string>.IsAny))
+                .Callback(new TryGetNewerVersionCallback((out System.Version delegateVersion, out string delegateMessage) =>
+                {
+                    delegateVersion = newVersion;
+                    delegateMessage = error;
+                }))
+                .Returns(upgradeResult);
+        }
+
+        public void VerifyOrchestratorInvokes(
+            bool upgradeAllowed,
+            bool queryNewestVersion,
+            bool downloadNewestVersion,
+            bool installNewestVersion,
+            bool cleanup)
+        {
+            this.MoqUpgrader.Verify(
+                upgrader => upgrader.UpgradeAllowed(
+                    out It.Ref<string>.IsAny),
+                    upgradeAllowed ? Times.Once() : Times.Never());
+
+            this.MoqUpgrader.Verify(
+                upgrader => upgrader.TryQueryNewestVersion(
+                    out It.Ref<System.Version>.IsAny,
+                    out It.Ref<string>.IsAny),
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        private const string LocalGVFSVersion = "1.0.18115.1";
+        private const string NewerThanLocalVersion = "1.1.18115.1";
+
+        private delegate void TryGetNewerVersionCallback(out System.Version version, out string message);
+        private delegate void UpgradeAllowedCallback(out string message);
+        private delegate void TryRunPreUpgradeChecksCallback(out string delegateMessage);
+        private delegate void TryDownloadNewestVersionCallback(out string message);
+        private delegate void TryRunInstallerCallback(InstallActionWrapper installActionWrapper, out string error);
+
+        private MockTracer Tracer { get; set; }
```
jamill: One pattern I wanted to see if we could improve - we use the `Mock` prefix to indicate both a `mocked` object, as well as the `Mock<..>` container for the object. I thought it would be nice if we had some way to distinguish these when looking at the code, so we would know whether what type we are dealing with (and, do we need to pass in the `.Object` or just the `Object` itself.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
...
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.Success);
+        }
+
+        [TestCase]
+        public void GetNewVersionFails()
+        {
+            string errorMessage = "Authentication error.";
+
+            this.MockUpgrader.Setup(upgrader => upgrader.TryQueryNewestVersion(out It.Ref<System.Version>.IsAny, out It.Ref<string>.IsAny))
```
jamill: What would you think if we put some of the details behind a method with a descriptive name `SetTryQueryNewestVersionReturnValues(...)`.? I like the idea of being able to easily read the test case to get a high level idea of what the test is doing, and I feel the details of setting the Callback are a bit verbose and make it harder to see the intent...

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -152,19 +152,19 @@ private bool TryRunUpgrade(out Version newVersion, out string consoleError)
             if (!this.LaunchInsideSpinner(
                 () =>
                 {
-                    if (!this.TryCheckIfUpgradeAvailable(out newGVFSVersion, out error))
+                    if (!this.preRunChecker.TryRunPreUpgradeChecks(out error))
```
jamill: Nice to see that the unit tests caught an actual issue :)

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -262,14 +271,16 @@ private bool TryCheckIfUpgradeAvailable(out Version newestVersion, out string co
...
+                    consoleError = null;
+                    this.output.WriteLine(message);
+                    this.tracer.RelatedInfo($"No new upgrade releases available. {message}");
+                    return true;
                 }
 
                 activity.RelatedInfo("Successfully checked for new release. {0}", newestVersion);
             }
 
+            consoleError = null;
```
wilbaker: Minor style comment: For consistency you could set `consoleError` to `null` at the start of the method (like you do for `newestVersion`.  This would also allow you to remove ` consoleError = null;` on line 274

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -177,6 +177,13 @@ private bool TryRunUpgrade(out Version newVersion, out string consoleError)
                 return false;
             }
 
+            if (newGVFSVersion == null)
+            {
+                newVersion = null;
```
wilbaker: Why do we need to set `newVersion` to `null` here?

Similar to the comment in `TryCheckIfUpgradeAvailable` can all of `out` parameters be initialized to `null` at the start of the method and then only set as needed from that point on?  I have trouble following why some of these are being set to `null` multiple times.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
...
+            this.orchestrator = new UpgradeOrchestrator(
+                this.MockUpgrader.Object,
+                this.Tracer,
+                this.MockPreRunChecker,
+                input: null,
+                output: this.Output);
+        }
+
+        [TestCase]
+        public void UpgradesWhenUpgradeAvailable()
```
wilbaker: Naming suggestion:  Include a more explicit description of what's being tested and the the expected result in the name.  

For example, rather than `UpgradesWhenUpgradeAvailable` try something like `OrchestratorExecuteSucceedsWhenUpgradeAvailable`.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
wilbaker: Related, I see a lot of code like this being duplicated:

```
this.MockUpgrader.Verify(upgrader => upgrader.UpgradeAllowed(out It.Ref<string>.IsAny), Times.Once());
this.MockUpgrader.Verify(upgrader => upgrader.TryQueryNewestVersion(out It.Ref<System.Version>.IsAny, out It.Ref<string>.IsAny), Times.Once());
this.MockUpgrader.Verify(upgrader => upgrader.TryDownloadNewestVersion(out It.Ref<string>.IsAny), Times.Never());
```

Could all of the common `Verify` calls that are checking if a method has been called be moved to a common `VerifyOrchastratorExecuteBehavior` type method?  Each of the helper methods we expect to be called (or not) could be passed in.

Example:

```
this.VerifyOrchastratorExecuteBehavior(
    callsUpdatedAllowed: true, 
    callsTryQueryNewestVersion: true, 
    callsTryDownloadNewestVersion: false);
```

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            this.orchestrator.Execute();
+
+            this.MockUpgrader.Verify(upgrader => upgrader.UpgradeAllowed(out It.Ref<string>.IsAny), Times.Once());
+            this.MockUpgrader.Verify(upgrader => upgrader.TryQueryNewestVersion(out It.Ref<System.Version>.IsAny, out It.Ref<string>.IsAny), Times.Once());
+            this.MockUpgrader.Verify(upgrader => upgrader.TryDownloadNewestVersion(out It.Ref<string>.IsAny), Times.Once());
+            this.MockUpgrader.Verify(upgrader => upgrader.TryRunInstaller(It.IsAny<InstallActionWrapper>(), out It.Ref<string>.IsAny), Times.Once());
+            this.MockUpgrader.Verify(upgrader => upgrader.TryCleanup(out It.Ref<string>.IsAny), Times.Once());
```
wilbaker: Should we be checking that `TryCleanup` is (or is not called) in every `TestCase`?  If not, how did you decide when we should validate that it's been called?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
+﻿using GVFS.Common;
```
wilbaker: Why is this test in `GVFS/GVFS.UnitTests.Windows/` and not `GVFS/GVFS.UnitTests`?

Can we run these tests on both platforms?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
...
+using GVFS.Upgrader;
+using Moq;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+
+namespace GVFS.UnitTests.Upgrader
+{
+    [TestFixture]
+    public class UpgradeOrchestratorWithIProductUpgraderTests
```
wilbaker: Can this just be renamed to something like `UpgradeOrchestratorTests`?   Is there any functionality in `UpgradeOrchestrator` is not being covered by these tests?

(It's safe to assume we're mocking out the components used by `UpgradeOrchestrator`, there's no need to call out `IProductUpgrader` specifically)

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
derrickstolee: If you set an out-param at the beginning, then the compiler can't help you ensure you set it in every branching path. Even though he is setting it to `null` in two places, they are for different return conditions. We really want to ensure that it is set before each error return condition, and setting at the beginning loses that automated check.

I believe this is a case of function over style. Yes, the code would look cleaner, but we lose something in the process.

wilbaker: That's a great point, and for things like error messages I can see the benefit of setting them as close to the return as possible (rather that at the start of the method).

It does look like we still need a change here though, as `newVersion` is being set to `null` at the start of this method and so we're not getting the benefits you mentioned:

```
private bool TryRunUpgrade(out Version newVersion, out string consoleError)
{
    newVersion = null;
```

wilbaker: Doh! As soon as I posted my reply I realized this was about `newVersion` and not `consoleError`.

My main comment was really about consistency, if we need to make sure an `out` parameter is getting set in each branch then it shouldn't be set at the top of the method.  If we *are* initializing an `out` parameter at the top of a method there's no need to set it to that same value before returning.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
wilbaker: Disregard, see discussion regarding this style in the comment for `TryRunUpgrade`

cc: @derrickstolee

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
```diff
@@ -0,0 +1,202 @@
...
+    {
+        private const string OlderThanLocalVersion = "1.0.17000.1";
```
jeschu1: Are these three strings used?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
alameenshah: Jameson and I discussed about Mock vs Moq. I think Moq clear-cut. Using Moq.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
alameenshah: Added check for TryCleanup

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorWithIProductUpgraderTests.cs
alameenshah: Agree that `GVFS/GVFS.UnitTests` is the rightful place for upgrader tests. But right now Upgrade is supported only on Windows. Both the upgrade paths today (Github & Nuget) are Windows only. We can move the tests when we port upgrader to Mac.

jamill: Just a small clarification - `UpgradeOrchestrator` is part of `GVFS.Upgrader`, which is .NET framework (Windows) only.

The actual upgrader implementations are in `GVFS.Common` and those can be tested in `GVFS.UnitTests`, unless they are tested via the `UpgradeOrchestrator`

wilbaker: > Just a small clarification - UpgradeOrchestrator is part of GVFS.Upgrader, which is .NET framework (Windows) only.

> The actual upgrader implementations are in GVFS.Common and those can be tested in GVFS.UnitTests, unless they are tested via the UpgradeOrchestrator

Ok thanks!

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                queryNewestVersion: true,
+                downloadNewestVersion: true,
+                installNewestVersion: true,
+                cleanup: true);
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.Success);
+        }
+
+        [TestCase]
+        public void NewVersionNotAvailable()
```
wilbaker: Two minor naming suggestions:

- Like the above test case, start the method name with `ExecuteSucceedsWhen`
- Does `NewVersionNotAvailable` mean there is a new version but VFS4G fails to download it?  Or does it mean there is no new version on the server?  If it's the second, consider changing the name to something like `ExecuteSucceedsWhenOnLatestVersion`

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.VerifyOutput("ERROR: Authentication error.");
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.GenericError);
+        }
+
+        [TestCase]
+        public void FailPrecheck()
```
wilbaker: Maybe `ExecuteFailsWhenPrecheckFails`?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                queryNewestVersion: true,
+                downloadNewestVersion: false,
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.Success);
+        }
+
+        [TestCase]
+        public void GetNewVersionFails()
```
wilbaker: Another minor naming suggestion:

`ExecuteFailsWhenQueryingNewestVersionFails`

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                queryNewestVersion: false,
+                downloadNewestVersion: false,
+                installNewestVersion: false,
+                cleanup: true);
+
+            this.orchestrator.ExitCode.ShouldEqual(ReturnCode.GenericError);
+        }
+
+        [TestCase]
+        public void DownloadFails()
```
wilbaker: Similar naming suggestion: `ExecuteFailsWhenDownloadingNewestVersionFails`

wilbaker: (Same naming suggestion where appropriate on the other `TestCases`)

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeOrchestratorTests.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            string message = string.Empty;
+            mockUpgrader.Setup(upgrader => upgrader.TryDownloadNewestVersion(out It.Ref<string>.IsAny)).Returns(true);
+            mockUpgrader.Setup(upgrader => upgrader.TryRunInstaller(It.IsAny<InstallActionWrapper>(), out message)).Returns(true);
+            mockUpgrader.Setup(upgrader => upgrader.TryCleanup(out It.Ref<string>.IsAny)).Returns(true);
+
+            return mockUpgrader;
+        }
+
+        public void SetUpgradeAvailable(Version newVersion, string error)
```
wilbaker: What should `TryQueryNewestVersion` return if there is no upgrade available?  

I ask because it looks like this method is setting whether `TryQueryNewestVersion` succeeds or fails (rather than if it behaves like a new version is available).

alameenshah: When no upgrade is available - `TryQueryNewestVersion` returns `true`, `out newVersion` will be set to `null`, `out message` will be set to "`... you're all caught up on upgrades..`".

jamill: I do think the meaning of the return value for `TryQueryNewestVersion` method can be ambiguous or even initially unexpected, but I do think it follows the semantics for our `Try....` methods (I think...). We successfully queried the latest version - it just happens there is no newer version.

I had to read through this a couple of times to make sure I understood how to interpret the return value. Outside of this PR, maybe we could make this less confusing, as I do expect this to cause confusion in the future...

PR #676, opened 2019-01-08
Refactor PrjFS.xcworkspace into PrjFS.xcodeproj

nickgra:
@pmj and I in another PR (#637) discussed the shortcomings of our current ProjFS.Mac code structure. This PR attempts to refactor our code into a single xcodeproj with multiple targets, which is what I had in mind. This is just one approach we could take, I put this version together to get feedback and to see if others had a different solution in mind.

This PR puts a new .xcodeproj at the top level, deletes the two nested .xcodeproj files, removes an unneeded folder in PrjFSKext, updates headers to reflect the new directory structures and I manually copied as many of the customized xcodeproj settings from the old projects to this one.

Other things discovered as part of this PR:
 - @wilbaker and I noticed that we don't have warnings as errors turned on. I believe we've silenced a lot of the errors that came from the kernel headers, so we should be able to turn this on in a future PR.
- We've been sloppy about `PrjFS` vs `ProjFS`. I went with PrjFS to match the old name of the .xcworkspace
- I don't know where the Build/Products part of the BuildOutput path was coming from before, but it wasn't adding any value (they were directories with only one member) so I removed them and update all consumers that were looking in that path.
- We now build the user mode lib for the Profiling(Release) configuration. I think this is actually an improvement from the confusing old behavior. Some of my performance analysis work has shown the profiling to add a negligible overhead so (assuming the performance tests I'm formalizing agree with that statement) in a future PR, I'd like to just fold the profiling configuration into the Release configuration.


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -984,6 +984,8 @@ static bool TrySendRequestAndWaitForResponse(
         return false;
     }
     
+    KextLog_FileError(vnode, "vnode with path %s encountered", vnodePath);
```
wilbaker: Remove this?

nickgra:
@pmj I think this is what you were envisioning, if you're happy with this rework, I can merge this and you can rebase+merge your logging daemon change.

pmj:
Thanks for doing this! Sorry I only just got around to taking a look. Looks fine to me aside from the deployment target, which is set to macOS 10.14 for the project and 10.13 for the PrjFSLib target only. As far as I'm aware, we're targeting 10.13 in general, so I've fixed that in PR #679.

PR #677, opened 2019-01-08
Remove useless tracing and useless code

derrickstolee:
The `OnGetFileStream()` method is a very busy area of code when hydrating a repo. Don't bog it down with useless tracing that doesn't help anything.

While exploring the code base for this excessive tracing, I found some code that can be safely deleted because it isn't used by anything.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
@@ -934,8 +934,7 @@ namespace GVFS.Platform.Windows
                         length,
                         streamGuid,
                         sha,
-                        metadata,
-                        this.Context.Tracer),
+                        metadata),
                     () =>
                     {
                         cancellationSource.Dispose();
@@ -975,8 +974,7 @@ namespace GVFS.Platform.Windows
             uint length,
             Guid streamGuid,
             string sha,
-            EventMetadata requestMetadata,
-            ITracer activity)
+            EventMetadata requestMetadata)
         {
             if (cancellationToken.IsCancellationRequested)
             {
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -355,22 +355,19 @@ private bool TryGetSymLinkTarget(string sha, out string symLinkTarget)
                 metadata.Add(nameof(sha), sha);
                 metadata.Add(nameof(placeholderVersion), placeholderVersion);
                 metadata.Add(nameof(commandId), commandId);
-                ITracer activity = this.Context.Tracer.StartActivity("GetFileStream", EventLevel.Verbose, Keywords.Telemetry, metadata);
```
jamill: If this is a hot path - should we only create the event metadata if we are going to use it (i.e. inside the following if block)?

derrickstolee: Good idea. I can move it to a private method to avoid code duplication, too. Thanks!

derrickstolee: We can do that on the Mac, but not Windows (at least not easily), as there is an async handler that takes the existing metadata and adds to it if necessary.

derrickstolee:
We confirmed with the WinES team that they are no longer measuring performance around this activity.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -349,28 +349,20 @@ private bool TryGetSymLinkTarget(string sha, out string symLinkTarget)
                 string sha = GetShaFromContentId(contentId);
                 byte placeholderVersion = GetPlaceholderVersionFromProviderId(providerId);
 
-                EventMetadata metadata = this.CreateEventMetadata(relativePath);
```
derrickstolee: We need to undo these changes to the Mac platform, as we will want to trace this data carefully while we are still learning about the platform. @jeschu1 is on it. (I'll strip out the change in this PR, though.)

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -939,10 +935,9 @@ private HResult QueryFileNameHandler(string virtualPath)
                         streamGuid,
                         sha,
                         metadata,
-                        activity),
+                        this.Context.Tracer),
```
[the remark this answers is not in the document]
derrickstolee: Thanks, @jeschu1 for fixing this!

derrickstolee:
Since the comment is hidden now, we are not changing `GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs` as we actually want to trace that data while we are figuring out the mac platform. It is helpful to see which processes are hydrating the repo.

PR #678, opened 2019-01-09
Update paths that refer to kext that were missed as part of refactor

nickgra:
#676 changed the layout of the ProjFS.Native code in BuildOutput but didn't update the LoadPrjFSKext.sh script.

We seemingly got lucky on the functional test machine by having the build agent select a numbered slot in the _work folder that had been used before and contained a valid kext in that location. I'm investigating how common that is.

While I was here, I searched for all other locations of `Build/Products` in the codebase and updated them too if they were looking for the kext (the native hooks are fine, their structure hasn't been changed). @jrbriggs Is this still fine for your yaml pipelines work?


nickgra:
@jamill I noticed your change to add the 'CopyPrjFS' MSBuild variable was causing the libPrjFSLib.dylib not to be copied to the /Publish folder. I think .NET internally reruns the build when publishing, so since the variable isn't set, we don't have a reference to the dylib and thus don't copy it to the Publish directory. 

Commit 2 in this PR fixes this.

jrbriggs:
@nickgra the yaml pipeline is broken for other reasons. I'll fix it up shortly.

jeschu1:
@nickgra  this explains why I wasn't have libPrjFSLib.dylib copied over, thanks for the fix!

PR #680, opened 2019-01-10
LibGit2RepoPool OPTION 1: Use multiple repo objects, dispose as possible

derrickstolee:
In #655, we added logic to dispose `LibGit2Repo` objects by reworking `LibGit2RepoPool`.

Then, some large repo builds went red and we reverted the change in #667.

Now, I have been unable to reproduce the failures and have added extra functional tests for parallel hydration issues. This re-does the work from #655.

**Note:** We have another idea to simplify the pool by only creating one `LibGit2Repo` to share across multiple readers. If it works, then that is actually preferred because it simplifies the pool and reduces the native memory footprint. It depends on the thread-safety of libgit2's object cache, but I've heard that it's pretty good for read-only operations like the ones we have here.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
@@ -127,6 +127,12 @@ namespace GVFS.Common.Maintenance
                         activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
                         return;
                     }
+
+                    if (this.Context.Repository.HasActiveLibGit2Repos)
+                    {
+                        activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to active libgit2 repos", Keywords.Telemetry);
+                        return;
+                    }
                 }
 
                 this.GetPackFilesInfo(out int beforeCount, out long beforeSize, out bool hasKeep);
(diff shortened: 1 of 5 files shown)


jamill:
> Note: We have another idea to simplify the pool by only creating one LibGit2Repo to share across multiple readers.

Are you  investigating this alternative, instead of this change, currently? or is the plan to proceed with this PR, and if it turns out the other idea works, then just replace it when / if we re-visit this area?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -127,6 +127,12 @@ protected override void PerformMaintenance()
                         activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
                         return;
                     }
+
+                    if (this.Context.Repository.HasActiveLibGit2Repos)
+                    {
+                        activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to active libgit2 repos", Keywords.Telemetry);
```
jamill: Can you remind me when we the `Telemetry` keyword is appropriate? I have some similar code, and now I am wondering if I need this keyword...

derrickstolee:
@jamill I'm investigating the other method at the moment and will not merge this PR until we see that in a PR, too. I'll merge at most one of them ;)

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: We use it if we want to ensure it is sent to the telemetry stream. Other warnings are only written to logs. This way we can see if any users are always blocked on libgit2.

wilbaker:
@derrickstolee changes look good to me, I'll hold off on approving until you wrap up the investigating if we can move to a single libgit2 repo.

derrickstolee:
Closing since we want to go with option 2. The fact that the full functional tests are failing with this change is indicative of a problem when over-allocating libgit2 repos.

PR #681, opened 2019-01-10
Verify Installer signature before launching

alameenshah:
#### Goal
Verify code signature of downloaded upgrade installer exes before launching it.

#### Implementation
Signature verification: At a high level, the `upgrader` reads the certificate contained in the signed installer. It then tries to build the certificate chain. If it succeeds, then the `exe` is considered valid and `upgrader` goes on to verify that the certificate CN name matches expected values ("Microsoft Corporation" for VFSForGit and "Johannes Schindelin" for Git). Installer is launched only when signature verification succeeds. To prevent an attacker from modifying installer exe while signature verification is going on, `upgrader` holds an open reference to the exe that is later released.

### DIFF
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -29,6 +29,10 @@ namespace GVFS.Common
         private const string GitInstallerFileNamePrefix = "Git-";
         private const int RepoMountFailureExitCode = 17;
         private const string ToolsDirectory = "Tools";
+        private const string GVFSSigner = "Microsoft Corporation";
+        private const string GVFSCertIssuer = "Microsoft Code Signing PCA";
+        private const string GitSigner = "Johannes Schindelin";
+        private const string GitCertIssuer = "COMODO RSA Code Signing CA";
         private static readonly string UpgraderToolName = GVFSPlatform.Instance.Constants.GVFSUpgraderExecutableName;
         private static readonly string UpgraderToolConfigFile = UpgraderToolName + ".config";
         private static readonly string[] UpgraderToolAndLibs =
@@ -374,12 +378,33 @@ namespace GVFS.Common
             return false;
         }
 
-        protected virtual void RunInstaller(string path, string args, out int exitCode, out string error)
+        protected virtual void RunInstaller(string path, string args, string certCN, string issuerCN, out int exitCode, out string error)
         {
-            ProcessResult processResult = ProcessHelper.Run(path, args);
+            using (Stream stream = this.fileSystem.OpenFileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, false))
+            {
+                string expectedCNPrefix = $"CN={certCN}, ";
+                string expectecIssuerCNPrefix = $"CN={issuerCN}";
+                string subject;
+                string issuer;
+                if (!GVFSPlatform.Instance.TryVerifyAuthenticodeSignature(path, out subject, out issuer, out error))
+                {
+                    exitCode = -1;
+                    return;
+                }
 
-            exitCode = processResult.ExitCode;
-            error = processResult.Errors;
+                if (!subject.StartsWith(expectedCNPrefix) || !issuer.StartsWith(expectecIssuerCNPrefix))
+                {
+                    exitCode = -1;
+                    error = $"Installer {path} is signed by unknown signer.";
+                    this.tracer.RelatedError($"Installer {path} is signed by unknown signer. Signed by {subject}, issued by {issuer} expected signer is {certCN}, issuer {issuerCN}.");
+                    return;
+                }
+
+                ProcessResult processResult = ProcessHelper.Run(path, args);
+
+                exitCode = processResult.ExitCode;
+                error = processResult.Errors;
+            }
         }
 
         private bool TryGetGitVersion(out GitVersion gitVersion, out string error)
@@ -450,7 +475,26 @@ namespace GVFS.Common
                 {
                     string logFilePath = GVFSEnlistment.GetNewLogFileName(ProductUpgraderInfo.GetLogDirectoryPath(), Path.GetFileNameWithoutExtension(path));
                     string args = installerArgs + " /Log=" + logFilePath;
-                    this.RunInstaller(path, args, out installerExitCode, out error);
+                    string certCN = null;
+                    string issuerCN = null;
+                    switch (assetId)
+                    {
+                        case GVFSAssetId:
+                        {
+                            certCN = GVFSSigner;
+                            issuerCN = GVFSCertIssuer;
+                            break;
+                        }
+
+                        case GitAssetId:
+                        {
+                            certCN = GitSigner;
+                            issuerCN = GitCertIssuer;
+                            break;
+                        }
+                    }
+
+                    this.RunInstaller(path, args, certCN, issuerCN, out installerExitCode, out error);
 
                     if (installerExitCode != 0 && string.IsNullOrEmpty(error))
                     {
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -6,10 +6,12 @@ using GVFS.Platform.Windows.DiskLayoutUpgrades;
 using Microsoft.Win32;
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Pipes;
 using System.Linq;
+using System.Management.Automation;
 using System.Security.AccessControl;
 using System.Security.Principal;
 using System.ServiceProcess;
@@ -264,6 +266,30 @@ namespace GVFS.Platform.Windows
             }
 
             return true;
+        }
+
+        public override bool TryVerifyAuthenticodeSignature(string path, out string subject, out string issuer, out string error)
+        {
+            using (PowerShell powershell = PowerShell.Create())
+            {
+                powershell.AddScript($"Get-AuthenticodeSignature -FilePath {path}");
+
+                Collection<PSObject> results = powershell.Invoke();
+                if (powershell.HadErrors || results.Count <= 0)
+                {
+                    subject = null;
+                    issuer = null;
+                    error = $"Powershell Get-AuthenticodeSignature failed, could not verify authenticode for {path}.";
+                    return false;
+                }
+
+                Signature signature = results[0].BaseObject as Signature;
+                bool isValid = signature.Status == SignatureStatus.Valid;
+                subject = signature.SignerCertificate.SubjectName.Name;
+                issuer = signature.SignerCertificate.IssuerName.Name;
+                error = isValid == false ? signature.StatusMessage : null;
+                return isValid;
+            }
         }
 
         public override string GetCurrentUser()
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -51,6 +51,8 @@ namespace GVFS.Common
         public abstract bool TryGetGVFSHooksPathAndVersion(out string hooksPaths, out string hooksVersion, out string error);
         public abstract bool TryInstallGitCommandHooks(GVFSContext context, string executingDirectory, string hookName, string commandHookPath, out string errorMessage);
 
+        public abstract bool TryVerifyAuthenticodeSignature(string path, out string subject, out string issuer, out string error);
+
         public abstract InProcEventListener CreateTelemetryListenerIfEnabled(string providerName, string enlistmentId, string mountId);
 
         public abstract Dictionary<string, string> GetPhysicalDiskInfo(string path, bool sizeStatsOnly);
--- a/GVFS/GVFS.Platform.Mac/MacPlatform.cs
+++ b/GVFS/GVFS.Platform.Mac/MacPlatform.cs
@@ -56,6 +56,11 @@ namespace GVFS.Platform.Mac
             return true;
         }
 
+        public override bool TryVerifyAuthenticodeSignature(string path, out string subject, out string issuer, out string error)
+        {
+            throw new NotImplementedException();
+        }
+
         public override bool IsProcessActive(int processId)
         {
             return MacPlatform.IsProcessActiveImplementation(processId);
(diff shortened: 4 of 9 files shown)


GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -545,6 +568,43 @@ private bool AssetInstallerNameCompare(Asset asset, params string[] expectedFile
             this.tracer.RelatedEvent(EventLevel.Informational, message, metadata);
         }
 
+        private bool VerifyCodeSignature(string path, string certCN, out string error)
```
derrickstolee: Are we doing something similar with NuGet? Is there a way that we can generalize this method?

alameenshah: Tried verifying a Nuget package with this method - I get the exception " System.Security.Cryptography.CryptographicException: Cannot find the requested object."

Tried using the SignTool to verify nuget package - that also didn't work.
```
C:\Users\alsha>signtool verify /pa C:\Users\alsha\Downloads\nunit.3.11.0.nupkg
File: C:\Users\alsha\Downloads\nunit.3.11.0.nupkg
Index  Algorithm  Timestamp
========================================
SignTool Error: This file format cannot be verified because it is not
        recognized.

Number of errors: 1
```
Couldn't yet find an API/Tool that could be generic enough (Even Windows Explorer does not show certificate info for signed Nuget packages.) Will check with@jamill (he is looking into verifying nuget packages) and update this thread.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -545,6 +568,43 @@ private bool AssetInstallerNameCompare(Asset asset, params string[] expectedFile
...
+                    {
+                        RevocationFlag = X509RevocationFlag.EntireChain,
+                        RevocationMode = X509RevocationMode.Online,
+                        UrlRetrievalTimeout = new TimeSpan(0, 1, 0),
+                        VerificationFlags = X509VerificationFlags.NoFlag
+                    }
+                };
+                string expectedSubjectName = $"CN={certCN}";
+
+                if (certificateChain.Build(certificate2) && certificate2.SubjectName.Name.StartsWith(expectedSubjectName))
```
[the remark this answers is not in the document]
alameenshah: >> Can you elaborate on how we do the check here?

Am not an expert here, this is my understanding - The certificate used to sign Git is signed by an intermediate CA, and that intermediate CA can be signed by another, creating a chain of CAs.
screen shot 2019-01-10 at 6 24 06 pm

To validate Git certificate, it is required that a parent certificate in the CA chain is trusted (there are a set of trusted CAs on the system). CertificateChain.Build() traverses the chain and validates that.

>> Why do verify StartsWith(expectedSubjectName)? Is there a way to check the whole name.

Yes, we can check the whole name. This is the whole Subject name in the VFSForGit certificate for example - ```CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US```

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -364,12 +365,30 @@ protected virtual bool TryFetchReleases(out List<Release> releases, out string e
...
 
-        protected virtual void RunInstaller(string path, string args, out int exitCode, out string error)
+        protected virtual void RunInstaller(string path, string args, string certCN, out int exitCode, out string error)
         {
-            ProcessResult processResult = ProcessHelper.Run(path, args);
+            Stream stream = this.fileSystem.OpenFileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, false);
 
-            exitCode = processResult.ExitCode;
-            error = processResult.Errors;
+            if (stream == null)
```
derrickstolee: This stream check seems like something you should do as part of the `VerifyCodeSignature`, as you are not using it any more after this check.

Also, stick it in a `using()` statement so you are sure to close it even on exceptions.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -364,12 +365,30 @@ protected virtual bool TryFetchReleases(out List<Release> releases, out string e
...
+            {
+                exitCode = -1;
+                error = $"Could not launch installer {path}. Error reading file.";
+                return;
+            }
+
+            if (this.VerifyCodeSignature(path, certCN, out error))
```
derrickstolee: I generally prefer seeing `if (unusual situation) { return with error }` instead of `if (typical) { do stuff } else {return with error}`.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -459,7 +478,11 @@ private bool TryRunInstallerForAsset(string assetId, out int installerExitCode,
             {
                 string logFilePath = GVFSEnlistment.GetNewLogFileName(ProductUpgraderInfo.GetLogDirectoryPath(), Path.GetFileNameWithoutExtension(path));
                 string args = installerArgs + " /Log=" + logFilePath;
-                this.RunInstaller(path, args, out installerExitCode, out error);
+                string gvfsCN = "Microsoft Corporation";
+                string gitCN = "Johannes Schindelin";
+                string certCN = assetId.Equals(GVFSAssetId) ? gvfsCN : assetId.Equals(GitAssetId) ? gitCN : null;
```
derrickstolee: Please don't next conditional operators. You can use a `switch` statement with strings.

GVFS/GVFS.Common/GitHubUpgrader.cs
derrickstolee: Please do check, as I expect this crypto stuff to be useful for both.

GVFS/GVFS.Common/GitHubUpgrader.cs
derrickstolee: Should we split these into two `if` statements with different error messages?

GVFS/GVFS.Common/GitHubUpgrader.cs
derrickstolee: The other benefit to splitting it out preemptively is that you can inject a mock during unit tests to ensure the GitHubUpgrader responds correctly to incorrect certificate results.

alameenshah: >> Please do check, as I expect this crypto stuff to be useful for both.

Did some more digging. Found a few issues with using C# API's - 1. it doesn't validate Nuget packages. 2. it doesn't help validate that the signed binary has not been tampered with (it  only checks that the certificate used to sign the binary is valid). 3. C# API also don't work on the Mac. Couldn't see much value in using C# API (vs using more capable platform native).

Alternative option was to use platform specific APIs (WinTrustAPI on Windows and in future may be use `codesign` tool on Mac). There is a `PowerShell` wrapper for WinTrustAPI, `Get-AuthenticodeSignature()` which is easy to use. This method cannot be used to validate Nuget packages however, but it validates the signing certificate and checks binary tampering. `Nuget Upgrader` will have to use Nuget specific API to validate the container package, but once after the package has been unarchived to the `Downloads` directory, it can still use this method (`GVFSPlatform.Instance.TryVerifyAuthenticodeSignature()`) before launching the exes.

alameenshah:
> 1. Unit test that we respond correctly to an incorrect certificate.

Updated existing UT to check for authenticode failure behavior. The mock classes are not granular enough to simulate failure in `TryVerifyAuthenticodeSignature()` method, but I could do the next best thing which is to make `RunInstaller()` fake authentication error.

> 2. Functional test that the verification logic discovers bad data. A simple file corruption would be enough.

I had to do manual testing here. Couple issues with doing a Functional test. 1. We don't do full FT today to test upgrader (because it changes the state of the test machine). 2. Every time upgrader downloads a new installer package from server. It is hard (test would be fragile) to consistently tamper with the installer at the right time. I did do manual testing. This is how installer looks when I tampered with the downloaded git installer - 
screen shot 2019-01-14 at 6 11 32 pm

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -266,6 +268,29 @@ public override bool TryInstallGitCommandHooks(GVFSContext context, string execu
             return true;
         }
 
+        public override bool TryVerifyAuthenticodeSignature(PhysicalFileSystem fileSystem, string path, out string subject, out string error)
+        {
+            using (Stream stream = fileSystem.OpenFileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, false))
+            using (PowerShell powershell = PowerShell.Create())
+            {
+                powershell.AddScript($"Get-AuthenticodeSignature -FilePath {path}");
```
kewillford: Is powershell the only way to accomplish this?  Did you try using X509Certificate?  Is there anyway to use a self-signed certificate with a subject that is expected in this code to get around these checks and pass the validation?

mjcheetham: I'd imagine if you used a self-signed certificate to sign a package that had the correct subject name, and had that or one of it's parent certificates in the trust chain installed in the computer's trusted store, then you could fool it. That would be entirely on the user/computer admin however.

You could also get a real certificate, whose certification path included one in the trusted store, that had the subject name "Microsoft Corporation" that wasn't actually "Microsoft Corporation" if there are rogue signing authorities who are shirking their responsibility to verify the details of the thing they're signing (those who end up in the untrusted store).

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -266,6 +268,29 @@ public override bool TryInstallGitCommandHooks(GVFSContext context, string execu
...
         }
 
+        public override bool TryVerifyAuthenticodeSignature(PhysicalFileSystem fileSystem, string path, out string subject, out string error)
+        {
+            using (Stream stream = fileSystem.OpenFileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, false))
+            using (PowerShell powershell = PowerShell.Create())
+            {
+                powershell.AddScript($"Get-AuthenticodeSignature -FilePath {path}");
+
+                Collection<PSObject> results = powershell.Invoke();
```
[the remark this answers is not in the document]
alameenshah: Will take a look at this

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
alameenshah: @kewillford I did try `X509Certificate` - it does help verify whether the certificate is valid, but does not check whether the exe has been tampered with. Other options considered were pinvoke-ing into WinTrustAPI and Powershell command. Powershell was chosen because it is easier to use.

Tests that I've done - 
1. verify unsigned installers are not launched.
2. verify signed installers that have been tampered with are not launched.

>>I'd imagine if you used a self-signed certificate to sign a package that had the correct subject name, and had that or one of it's parent certificates in the trust chain installed in the computer's trusted store, then you could fool it. That would be entirely on the user/computer admin however.

Agree. Wondering if it would be an improvement to add a check for the issuer in addition to the signer. This would help avoid installing self-signed exes.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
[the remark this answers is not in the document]
alameenshah: Couldn't find an option in this API to specify `-NonInteractive -NoProfile` parameters. Wrote the following script to check whether PowerShell is launched interactive and also to find the profile that is loaded. Pasting the code used to test - 

```
public static void CheckInteractiveAndUserProfile()
        {
            RunCommand1($"Read-Host Please enter your name:");
            Console.WriteLine("");
            RunCommand1($"$profile");
        }

        public static void RunCommand1(string command)
        {
            using (PowerShell powershell = PowerShell.Create())
            {
                powershell.AddScript(command);
                Collection<PSObject> results = powershell.Invoke();
                if (powershell.HadErrors || results.Count <= 0)
                {
                    foreach (ErrorRecord error in powershell.Streams.Error)
                    {
                        Console.WriteLine($"{command} Error: {error.Exception.Message}");
                    }
                }
                else
                {
                    foreach (PSObject result in results)
                    {
                        string output = result == null ? "null" : result.ToString();
                        Console.WriteLine($"{command}: {output}");
                    }
                }
            }
        }
```

This is the output:
```
Read-Host Please enter your name: Error: A command that prompts the user failed because the host program or the command type does not support user interaction. The host was attempting to request confirmation with the following message:

$profile: null
```

While using `ProcessHelper.Run("powershell.exe")` method, for the same `Read-Host` and `$profile` commands I get similar results -
```
Read-Host Please enter your name: Error: Read-Host : Windows PowerShell is in NonInteractive mode. Read and Prompt functionality is not available.
At line:1 char:5
+ & { Read-Host Please enter your name: }
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Read-Host], PSInvalidOperationException
    + FullyQualifiedErrorId : InvalidOperation,Microsoft.PowerShell.Commands.ReadHostCommand



$profile: Output: C:\Users\alsha\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
```

Summary

| Launch Option | powershell.exe | PowerShell API |
| ------------- | ------------- |------------- |
| NonInteractive | No user interaction allowed | No user interaction allowed |
| NoProfile | Loads $HOME\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1 | Loads No profile |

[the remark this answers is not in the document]
wilbaker: > Loads $HOME\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1

How did you confirm powershell loaded the profile?  The documentation says: 

> -NoProfile
> Doesn't load the PowerShell profile.
https://docs.microsoft.com/en-us/powershell/scripting/components/console/powershell.exe-command-line-help?view=powershell-6

>THE NOPROFILE PARAMETER
>To start PowerShell without profiles, use the NoProfile parameter of PowerShell.exe, the program that starts PowerShell.
>https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-6

Is is possible the `$profile` variable is set even though the profile was not loaded?

Could you try adding something to your profile that would make it obvious it's being added?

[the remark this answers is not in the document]
wilbaker: > Could you try adding something to your profile that would make it obvious it's being added?

Note: The main concern I had was with the `PowerShell` in `System.Management.Automation`, there's no need to further investigate the behavior of powershell.exe for this PR.

Issue #529, opened 2018-11-16
Mac: create a telemetry listener


sanoursa:
Similar to the Windows implementation, this listener will write all telemetry events to a local log. It must not upload anything.

PR #682, opened 2019-01-11
Add out-of-proc telemetry daemon event listener

mjcheetham:
Add a new telemetry event listener which sends JSON messages to a named pipe for out-of-proc dispatch to ETW/AI.

The new listener can exist alongside any existing 'platform-specific' telemetry listeners (such as the ETW listener that is in-proc).

Events/messages sent by the listener are 'fire-and-forget'; we do not retry and we do not wait any length of time for the pipe to be ready.

The rationale here is that we do not want to introduce delay to the execution of the application for only telemetry events. Also if the current system does not have anyone listening on the telemetry named
pipe then every message would delay the application execution by the pipe connection timeout(!)

We continue to guard the telemetry listener's activation on the presence of the `gvfs.telemetry-id` global configuration entry, just like the existing in-proc ETW provider. The only difference is that now we don't
care what the value of that config entry is, just that it is set.
(The in-proc ETW listener uses the value for the ETW iKey which is stored in the daemon's logic instead now.)

Note: I renamed the `InProcEventListener` abstract class to just `EventListener` because nothing about it was specifically 'in proc' only, and now we have an instance which send data out-of-proc.

This PR aims to address #529, which has since been expanded to cover a single solution for posting telemetry across Windows and macOS (post telemetry to a local telemetry service/daemon).


GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,162 @@
...
+            try
+            {
+                var buffer = Encoding.UTF8.GetBytes(message);
+                this.pipeClient.Write(buffer, 0, buffer.Length);
+            }
+            catch (Exception)
+            {
+                // We can't send this message for some reason (e.g., broken pipe); we attempt no recovery or retry
+                // mechanism and drop this message. We will try to recreate/connect the pipe on the next message.
+                this.pipeClient = null;
```
jeffhostetler: do we need to dispose of pipeClient before nulling it it ?

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -40,6 +40,7 @@ public static void Register(GVFSPlatform platform)
         public abstract bool IsProcessActive(int processId);
         public abstract void IsServiceInstalledAndRunning(string name, out bool installed, out bool running);
         public abstract string GetNamedPipeName(string enlistmentRoot);
+        public abstract string GetTelemetryNamedPipeName();
```
jeffhostetler: I think this refers to the pathname of the named pipe to the my telemetry service.
We're using a named pipe on Windows, but I think we're going to use sockets or something (TBD) on the Mac.
I wonder if this can be hidden inside the platform code, or rather does it need to be visible in the shared layers?

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,162 @@
...
+            if (!result.TryParseAsString(out value, out error, defaultValue: string.Empty) || string.IsNullOrWhiteSpace(value))
+            {
+                result = GitProcess.GetFromGlobalConfig(gitBinRoot, configKey);
+                result.TryParseAsString(out value, out error, defaultValue: string.Empty);
+            }
+
+            return value.TrimEnd('\r', '\n');
+        }
+
+        private void SendMessage(string message)
```
jeffhostetler: This looks good for now.  (And I know I'm still working on the Mac version of the daemon, so some
of this is premature.)  But maybe (now or later) make this class abstract where it is here in the common
code and put this named-pipe SendMessage() routine in an override in Windows platform code.
(At least I think that is the way they do things in this code base.)

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: That's a good point. Most likely.. although I'm not sure what disposing of a broken pipe would do 🤔

mjcheetham: I've added a call to `Dispose` here too.

mjcheetham:
> Looks very good. I made a few notes on maybe isolating the named pipes information to
> the windows specific layers. But that can wait until we have a mac version to test against
> and refactor it later or we can do that now in anticipation of it. Your call.

Thanks for the review @jeffhostetler! I've fixed the missing `Dispose` call in `SendMessage`.

I was thinking it might be better to either hang fire with the PR until the Mac daemon is more realised and then make whatever changes are best to support both, _or_ just get this PR in (to get some mileage on the VFS4Git telemetry-out-of-proc idea on Windows) and then revisit later with the Mac daemon?

I'm not fussed either way 😃

jeffhostetler:
@mjcheetham Go ahead and merge this version as is.  we can refactor as necessary when the mac version is ready.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            // Create pipe if this is the first message, or if the last connection broke for any reason
+            if (this.pipeClient == null)
+            {
+                string pipeName = GVFSPlatform.Instance.GetTelemetryNamedPipeName();
```
kewillford: Okay now I see why you added this method to the platform interface.  What are you thoughts on having the method in the `GVFSPlatform` be 
```
void AddTelemetryListenersIfEnabled(List<EventListener> listeners, string providerName, string enlistmentId, string mountId)
```
and the platform can add what it wants.  For windows it would add both for now, mac nothing now but could only be the daemon listener.

Then the `CreateIfEnabled` could take the `pipeName` that the platform passes to it and store it for use here?

I hope that makes sense.  Not sure how much cleaner this would be.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            try
+            {
+                var buffer = Encoding.UTF8.GetBytes(message);
+                this.pipeClient.Write(buffer, 0, buffer.Length);
+            }
+            catch (Exception)
+            {
+                // We can't send this message for some reason (e.g., broken pipe); we attempt no recovery or retry
+                // mechanism and drop this message. We will try to recreate/connect the pipe on the next message.
```
kewillford: So if someone decides to turn off the daemon we will lose all messages?  And we will try the connect every time?  Do you know what the perf impact will be?  I know you have the timeout at `0` but there is still a cost at the very least at the exception being thrown.  Wondering if we should have a message queue with a limited size and try connecting and writing messages on a background thread?  I know that is added complexity but would get all this code out of the way in performance critical code paths.

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -51,7 +52,7 @@ public static void Register(GVFSPlatform platform)
         public abstract bool TryGetGVFSHooksPathAndVersion(out string hooksPaths, out string hooksVersion, out string error);
         public abstract bool TryInstallGitCommandHooks(GVFSContext context, string executingDirectory, string hookName, string commandHookPath, out string errorMessage);
 
-        public abstract InProcEventListener CreateTelemetryListenerIfEnabled(string providerName, string enlistmentId, string mountId);
+        public abstract EventListener CreatePlatformTelemetryListener(string providerName, string enlistmentId, string mountId);
```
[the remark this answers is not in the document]
mjcheetham: Sure thing, I'll add it back. I had cut the `IfEnabled` bit off given since ht method name was getting a little bit verbose IMO.

GVFS/GVFS.Common/GVFSPlatform.cs
mjcheetham: The pipe name (or socket name, config, or whatever) has been moved into the platform code and out of the public API surface.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -70,7 +77,7 @@ public bool HasLogFileEventListener
             }
         }
 
-        public void AddInProcEventListener(InProcEventListener listener)
+        public void AddInProcEventListener(EventListener listener)
```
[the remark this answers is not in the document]
mjcheetham: Renamed.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,163 @@
...
+                    ["enlistmentId"] = enlistmentId,
+                    ["mountId"] = mountId,
+                    ["json"] = payload,
+                },
+
+                // TODO: do we need these?
+                // ETW-only properties
+                ["etw.activityId"] = activityId.ToString("D"),
+                ["etw.parentActivityId"] = parentActivityId.ToString("D"),
+            };
```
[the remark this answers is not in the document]
mjcheetham: I've moved to a concrete class, and updated the test accordingly (to check the JSON dictionary the daemon/service is expecting is the same).

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,163 @@
...
+                ["eventLevel"] = ((int)level).ToString(),
+                ["eventOpcode"] = ((int)opcode).ToString(),
+                ["payload"] = new Dictionary<string, string>
+                {
+                    ["enlistmentId"] = enlistmentId,
+                    ["mountId"] = mountId,
+                    ["json"] = payload,
+                },
+
+                // TODO: do we need these?
```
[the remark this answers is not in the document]
mjcheetham: I'm not sure.. @jeffhostetler did we need to send the ETW activity and parent activity IDs?
The daemon code also leaves this as a 'TODO' currently.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: Yes that makes sense. I've tried your idea and it looks nicer IMO :+1:

I've replaced the `CreatePlatformTelemetryListener[IfEnabled]` method with one that can return possibly multiple `EventListener`s. These listeners (zero or many) are attached to the `JsonTracer` if any are returned.

```csharp
// GVFSPlatform.cs
IEnumerable<EventListener> CreateTelemetryListeners(...);

// JsonTracer.cs
IEnumerable<EventListener> listeners = GVFSPlatform.Instance.CreateTelemetryListeners(...);
this.listeners.AddRange(listeners);
```

Windows returns both (if enabled) and Mac will eventually return one (the telemetry daemon one).

Note: I went with a `Create___()` method and letting the `JsonTracer` add the returned listeners itself, rather than have the tracer have to hand over it's internal `List<>` and couple the platform code to the tracer's internal data structure.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: > So if someone decides to turn off the daemon we will lose all messages?

Yes. If the mount process (or service.. or whoever created the `JsonTracer`) was started when the `gvfs.telemetry-id` configuration entry was set then telemetry tracing is on for the lifetime of the process. We don't re-evaluate the setting on failure (say if they disabled it after mounting).

> And we will try the connect every time?

Yes, for every message (when telemetry is enabled at start). We _could_ recheck the telemetry configuration entry on failure to send a message, and then stop sending them until it was set again? But again checking the config every telemetry event would equally add some overhead (possibly even more as it would involve a git.exe process start).

> Do you know what the perf impact will be? I know you have the timeout at 0 but there is still a cost at the very least at the exception being thrown.

I did some very rough profiling and I could only measure 0-1 ms delay per message. The sending of a successful message was also between 0-1 ms per message however. I can and will dig deeper on these numbers however.

> Wondering if we should have a message queue with a limited size and try connecting and writing messages on a background thread? I know that is added complexity but would get all this code out of the way in performance critical code paths.

An interesting idea, but yes this would add more complexity. Do we think it's worth it? Do we want to risk breaking due to failing to send telemetry, or risk losing events if case of a problem?

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
```diff
@@ -25,6 +25,12 @@ public static string GetNamedPipeNameImplementation(string enlistmentRoot)
             return Path.Combine(enlistmentRoot, GVFSConstants.DotGVFS.Root, "GVFS_NetCorePipe");
         }
 
+        public static string GetTelemetryNamedPipeNameImplementation()
+        {
+            // TODO: determine the correct pipe/socket name for telemetry on macOS
```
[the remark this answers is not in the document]
mjcheetham: @jeffhostetler do you have a GitHub issue number or Azure Boards work item number tracking the Mac telemetry daemon work? I can add it to the comment.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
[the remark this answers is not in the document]
jrbriggs: I don't think we need to send these to the daemon, as the daemon will be responsible for its own configuration.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
jrbriggs: I'd like for us to trace once that we were not able to connect to the daemon if we expected to be able to, for supportability.

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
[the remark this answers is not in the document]
jeffhostetler: i don't have one for the mac version, sorry.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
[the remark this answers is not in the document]
jeffhostetler: I have no need for them.  There was an open question on the gvfs side if they were making use of the activity nesting feature (primarily for perfview and WPA).

I think we can omit them.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
jeffhostetler: cool.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
jeffhostetler: See John's comment about logging an error here.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,179 @@
...
+            public string ProviderName { get; set; }
+            [JsonProperty("eventName")]
+            public string EventName { get; set; }
+            [JsonProperty("eventLevel")]
+            public EventLevel EventLevel { get; set; }
+            [JsonProperty("eventOpcode")]
+            public EventOpcode EventOpcode { get; set; }
+            [JsonProperty("payload")]
+            public TelemetryDaemonMessagePayload Payload { get; set; }
+            [JsonProperty("etw.activityId")]
```
jeffhostetler: We can also get rid of the activity fields here.

GVFS/GVFS.UnitTests/Tracing/TelemetryDaemonEventListenerTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                ["eventName"] = eventName,
+                ["eventLevel"] = (int)level,
+                ["eventOpcode"] = (int)opcode,
+                ["payload"] = new Dictionary<string, string>
+                {
+                    ["enlistmentId"] = enlistmentId,
+                    ["mountId"] = mountId,
+                    ["json"] = payload,
+                },
+                ["etw.activityId"] = activityId.ToString("D"),
```
jeffhostetler: And here.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: > I'd like for us to trace once that we were not able to connect to the daemon if we expected to be able to, for supportability.

Hmm.. that'd be interesting to do, seeing as we're in the event listener tracing code itself.
I'd need to, as an `EventListener` have a reference to the current `ITracer`, and trace to it (in the hope that the log-file listener has been added.. then I'd need to ensure I don't get stuck in a recursive loop attempting and failing to trace the message to telemetry that the telemetry tracing didn't work 🤓

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
kewillford: Looks good thanks!

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
kewillford: If there isn't a noticeable performance impact then this should be fine but like John said it would be nice if we had some record that we are dropping messages.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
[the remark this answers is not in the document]
mjcheetham: Removed.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: @jrbriggs @kewillford @jeffhostetler, I've pushed another commit which now traces the first instance of an `EventListener` failure to all other listeners (and any new listeners that are registered after the failure).

Please have another look at that most recent commit.

mjcheetham:
> Looks good! The only other thought I had was having a more generic `Message` class that can be used to pass into all the `RecordMessage...` methods but that seems out of scope of these changes.

@kewillford I've created a `TraceEventMessage` class to group all those arguments like event name, level, payload, activity ID etc as you suggested.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -267,10 +291,45 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
...
+
+        private void MarkAndLogListenerFailure(EventListener failingListener, string errorMessage)
+        {
+            if (this.failedListeners.ContainsKey(failingListener))
+            {
+                // We've already logged that this listener has failed so there is no need to do it again
+                return;
+            }
+
+            this.failedListeners.Add(failingListener, errorMessage);
```
kewillford: When does the failed listener get removed from the failed list if the next message is successful?  For example the service daemon is running again.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -70,21 +71,28 @@ public bool HasLogFileEventListener
         public void AddEventListener(EventListener listener)
         {
             this.listeners.Add(listener);
+
+            // Tell the new listener about others who have previously failed
+            foreach (var kvp in this.failedListeners)
```
kewillford: We have not allowed `var` in the source code.  There are a couple other places where you are using it.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -118,7 +110,7 @@ private void SendMessage(string message)
                 catch (TimeoutException)
                 {
                     // We can't connect; we will try again with a new pipe on the next message
-                    return;
+                    throw;
```
kewillford: We have been using the `bool Try...` pattern but I think that can be propagated too much and make the code a lot less readable.  If we decide to keep the exception pattern here, I would just remove the try/catch here altogether.

mjcheetham: Will remove the `try..catch` here.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: Currently it does not, but logging the recovery (only once after a successful message on a previously failing listener) might be useful – I'll add it.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -270,10 +306,74 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
...
+            };
+
+            foreach (EventListener listener in this.listeners)
+            {
+                string errorMessage;
+                bool? success = listener.TryRecordMessage(message, out errorMessage);
+
+                if (success == null)
+                {
+                    // Listener was disabled for this message type
```
kewillford: I don't really like empty if blocks what about
```
// Disabled listeners for message type will not have a value
if (success.HasValue)
{
    if (success == true)
    {
        this.MarkAndLogListenerRecovery(listener);
    }
    else
    {
        this.MarkAndLogListenerFailure(listener, errorMessage);
    }
}
```

GVFS/GVFS.Common/Tracing/EventListener.cs
```diff
@@ -18,24 +18,26 @@ public virtual void Dispose()
...
+        public bool? TryRecordMessage(TraceEventMessage message, out string errorMessage)
         {
-            if (!this.IsEnabled(level, keywords))
+            if (this.IsEnabled(message.Level, message.Keywords))
             {
-                return;
+                try
+                {
+                    this.RecordMessageInternal(message);
+                }
```
kewillford: Don't we need to return `true` here?  Makes me think we don't have any testing around this code.  How hard would it be to add some unit tests?

mjcheetham: That's true! Thanks for spotting this. I'll add some unit tests around the `TryRecordMessage` method.

GVFS/GVFS.UnitTests/Tracing/TelemetryDaemonEventListenerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            const string vfsVersion = "test-vfsVersion";
+            const string providerName = "test-ProviderName";
+            const string eventName = "test-eventName";
+            const EventLevel level = EventLevel.Error;
+            const EventOpcode opcode = EventOpcode.Start;
+            const string enlistmentId = "test-enlistmentId";
+            const string mountId = "test-mountId";
+            const string payload = "test-payload";
+
+            var expectedDict = new Dictionary<string, object>
```
kewillford: Remove `var`

mjcheetham:
@jrbriggs are there anymore changes that you need from me? 😃

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            // Create pipe if this is the first message, or if the last connection broke for any reason
+            if (this.pipeClient == null)
+            {
+                var pipe = new NamedPipeClientStream(".", this.pipeName, PipeDirection.Out, PipeOptions.Asynchronous);
+
+                // Specify a instantaneous timeout because we don't want to hold up the rest of the
+                // application if the pipe is not available; we will just drop this event.
+                // We let any TimeoutExceptions bubble up and will try again with a new pipe on the next SendMessage call.
+                pipe.Connect(timeout: 0);
```
jrbriggs: Who is responsible for catching the `TimeoutException`? Does it risk bringing down the application if not caught?

mjcheetham: This is caught in the public method `EventListener.TryRecordMessage` which looks like this:

```csharp
public bool? TryRecordMessage(TraceEventMessage message, out string errorMessage)
{
    try
    {
        this.RecordMessageInternal(...);
    }
    catch (Exception ex)
    {
        errorMessage = ex.ToString();
        return false;
    }
}
```

The call stack is:
```
TelemetryDaemonEventListerner.SendMessage
TelemetryDaemonEventListerner.RecordMessageInternal
EventListener.TryRecordMessage
```

jrbriggs: Thanks!

PR #683, opened 2019-01-11
Use a single LibGit2Repo for multiple requests, dispose if left unused

derrickstolee:
We use libgit2 to get some information from our Git object database without loading a full `git.exe` process. This allows us to:

* Check for object existence (usually a commit and its tree).
* Get the data out of a blob so we can hydrate a file.

The `LibGit2RepoPool` created 2_N_ copies of a libgit2 repository object, where _N_ is the number of processors available. We then lent these out from the pool to the threads trying to hydrate the repo. This can cause significant memory pressure as the repository objects do not share memory caches.

Since libgit2 doesn't use the multi-pack-index file, it loads handles to all of the .idx files on its first use. It then also holds handles to the .pack files that it reads, which may include packs we want to expire in the `PackfileMaintenanceStep`.

Since the `LibGit2RepoPool` never disposed of the `LibGit2Repo` objects, we never reclaimed the native memory they allocated, and we never cleared the file handles. This meant we could never clear .idx files until unmounting. This may also be a source of out-of-memory errors.

This PR does multiple things to improve our situation:

1. Use only one libgit2 repo. This is safe because we are performing read-only operations on the object database, and libgit2 has mutexes around its object database structures. This is also tested by the full functional test `ParallelHydrationTests.HydrateRepoInParallel`, in addition to performance testing I did on my machine.

2. Dispose of the libgit2 repo if it has not been used for 15 minutes. This number is mostly large enough that we should not dispose during a build that is hydrating the filesystem, but will dispose soon enough to run background maintenance.

3. In `PackfileMaintenanceStep` we now check if there are running libgit2 repos, in which case we do not continue running the step.

Now that we are using only one libgit2 repo, it makes sense to rename `LibGit2RepoPool` to `LibGit2RepoInvoker`. This provides safe logic around accessing the `LibGit2Repo` object (initializing if necessary) and disposing when the timer activates. Also, move the file into the `Git` folder of the `GVFS.Common` project.

### DIFF
--- a/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
@@ -1,11 +1,10 @@
-﻿using GVFS.Common.Git;
-using GVFS.Common.Tracing;
+﻿using GVFS.Common.Tracing;
 using System;
 using System.Threading;
 
-namespace GVFS.Common
+namespace GVFS.Common.Git
 {
-    public class LibGit2RepoPool
+    public class LibGit2RepoInvoker : IDisposable
     {
         private static readonly TimeSpan DefaultRepositoryDisposalPeriod = TimeSpan.FromMinutes(15);
 
@@ -13,22 +12,24 @@ namespace GVFS.Common
         private readonly ITracer tracer;
         private readonly object sharedRepoLock = new object();
         private readonly TimeSpan sharedRepositoryDisposalPeriod;
-        private bool stopped;
+        private volatile bool stopped;
+        private volatile int activeCallers;
         private LibGit2Repo sharedRepo;
         private Timer sharedRepoDisposalTimer;
-        private int activeCallers;
 
-        public LibGit2RepoPool(ITracer tracer, Func<LibGit2Repo> createRepo, int size = 1, TimeSpan? disposalPeriod = null)
+        public LibGit2RepoInvoker(ITracer tracer, Func<LibGit2Repo> createRepo, TimeSpan? disposalPeriod = null)
         {
-            if (size <= 0)
-            {
-                throw new ArgumentException("ProcessPool: size must be greater than 0");
-            }
-
             this.tracer = tracer;
             this.createRepo = createRepo;
 
-            this.sharedRepositoryDisposalPeriod = disposalPeriod ?? DefaultRepositoryDisposalPeriod;
+            if (!disposalPeriod.HasValue || disposalPeriod.Value <= TimeSpan.Zero)
+            {
+                this.sharedRepositoryDisposalPeriod = DefaultRepositoryDisposalPeriod;
+            }
+            else
+            {
+                this.sharedRepositoryDisposalPeriod = disposalPeriod.Value;
+            }
 
             this.sharedRepoDisposalTimer = new Timer(
                 (state) => this.DisposeSharedRepo(),
@@ -37,9 +38,14 @@ namespace GVFS.Common
                 period: this.sharedRepositoryDisposalPeriod);
         }
 
+        public bool IsActive => this.sharedRepo != null;
+
         public void Dispose()
         {
             this.stopped = true;
+
+            this.sharedRepoDisposalTimer?.Dispose();
+            this.sharedRepoDisposalTimer = null;
 
             lock (this.sharedRepoLock)
             {
@@ -50,12 +56,6 @@ namespace GVFS.Common
 
         public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult result)
         {
-            if (this.stopped)
-            {
-                result = default(TResult);
-                return false;
-            }
-
             try
             {
                 Interlocked.Increment(ref this.activeCallers);
@@ -83,7 +83,7 @@ namespace GVFS.Common
 
         private LibGit2Repo GetRepoFromPool()
         {
-            this.sharedRepoDisposalTimer.Change(this.sharedRepositoryDisposalPeriod, this.sharedRepositoryDisposalPeriod);
+            this.sharedRepoDisposalTimer?.Change(this.sharedRepositoryDisposalPeriod, this.sharedRepositoryDisposalPeriod);
 
             lock (this.sharedRepoLock)
             {
--- a/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
@@ -12,7 +12,7 @@ namespace GVFS.Common.Git
         private readonly ITracer tracer;
         private readonly object sharedRepoLock = new object();
         private readonly TimeSpan sharedRepositoryDisposalPeriod;
-        private volatile bool stopped;
+        private volatile bool disposing;
         private volatile int activeCallers;
         private LibGit2Repo sharedRepo;
         private Timer sharedRepoDisposalTimer;
@@ -42,13 +42,13 @@ namespace GVFS.Common.Git
 
         public void Dispose()
         {
-            this.stopped = true;
-
-            this.sharedRepoDisposalTimer?.Dispose();
-            this.sharedRepoDisposalTimer = null;
+            this.disposing = true;
 
             lock (this.sharedRepoLock)
             {
+                this.sharedRepoDisposalTimer?.Dispose();
+                this.sharedRepoDisposalTimer = null;
+
                 this.sharedRepo?.Dispose();
                 this.sharedRepo = null;
             }
@@ -59,7 +59,7 @@ namespace GVFS.Common.Git
             try
             {
                 Interlocked.Increment(ref this.activeCallers);
-                LibGit2Repo repo = this.GetRepoFromPool();
+                LibGit2Repo repo = this.GetSharedRepo();
 
                 if (repo != null)
                 {
@@ -81,16 +81,16 @@ namespace GVFS.Common.Git
             }
         }
 
-        private LibGit2Repo GetRepoFromPool()
+        private LibGit2Repo GetSharedRepo()
         {
-            this.sharedRepoDisposalTimer?.Change(this.sharedRepositoryDisposalPeriod, this.sharedRepositoryDisposalPeriod);
-
             lock (this.sharedRepoLock)
             {
-                if (this.stopped)
+                if (this.disposing)
                 {
                     return null;
                 }
+
+                this.sharedRepoDisposalTimer?.Change(this.sharedRepositoryDisposalPeriod, this.sharedRepositoryDisposalPeriod);
 
                 if (this.sharedRepo == null)
                 {
@@ -105,14 +105,13 @@ namespace GVFS.Common.Git
         {
             lock (this.sharedRepoLock)
             {
-                if (this.stopped || this.activeCallers > 0)
+                if (this.disposing || this.activeCallers > 0)
                 {
                     return;
                 }
 
-                LibGit2Repo repo = this.sharedRepo;
+                this.sharedRepo?.Dispose();
                 this.sharedRepo = null;
-                repo?.Dispose();
             }
         }
     }
--- a/GVFS/GVFS.Common/LibGit2RepoPool.cs
+++ b/GVFS/GVFS.Common/LibGit2RepoPool.cs
@@ -1,20 +1,24 @@
 ﻿using GVFS.Common.Git;
 using GVFS.Common.Tracing;
 using System;
-using System.Collections.Concurrent;
 using System.Threading;
 
 namespace GVFS.Common
 {
     public class LibGit2RepoPool
     {
-        private const int TryAddTimeoutMilliseconds = 10;
-        private const int TryTakeTimeoutMilliseconds = Timeout.Infinite;
+        private static readonly TimeSpan DefaultRepositoryDisposalPeriod = TimeSpan.FromMinutes(15);
 
-        private readonly BlockingCollection<LibGit2Repo> pool;
+        private readonly Func<LibGit2Repo> createRepo;
         private readonly ITracer tracer;
+        private readonly object sharedRepoLock = new object();
+        private readonly TimeSpan sharedRepositoryDisposalPeriod;
+        private bool stopped;
+        private LibGit2Repo sharedRepo;
+        private Timer sharedRepoDisposalTimer;
+        private int activeCallers;
 
-        public LibGit2RepoPool(ITracer tracer, Func<LibGit2Repo> createRepo, int size)
+        public LibGit2RepoPool(ITracer tracer, Func<LibGit2Repo> createRepo, int size = 1, TimeSpan? disposalPeriod = null)
         {
             if (size <= 0)
             {
@@ -22,26 +26,41 @@ namespace GVFS.Common
             }
 
             this.tracer = tracer;
-            this.pool = new BlockingCollection<LibGit2Repo>();
-            for (int i = 0; i < size; ++i)
-            {
-                this.pool.Add(createRepo());
-            }
+            this.createRepo = createRepo;
+
+            this.sharedRepositoryDisposalPeriod = disposalPeriod ?? DefaultRepositoryDisposalPeriod;
+
+            this.sharedRepoDisposalTimer = new Timer(
+                (state) => this.DisposeSharedRepo(),
+                state: null,
+                dueTime: this.sharedRepositoryDisposalPeriod,
+                period: this.sharedRepositoryDisposalPeriod);
         }
 
         public void Dispose()
         {
-            this.pool.CompleteAdding();
-            this.CleanUpPool();
+            this.stopped = true;
+
+            lock (this.sharedRepoLock)
+            {
+                this.sharedRepo?.Dispose();
+                this.sharedRepo = null;
+            }
         }
 
         public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult result)
         {
-            LibGit2Repo repo = null;
+            if (this.stopped)
+            {
+                result = default(TResult);
+                return false;
+            }
 
             try
             {
-                repo = this.GetRepoFromPool();
+                Interlocked.Increment(ref this.activeCallers);
+                LibGit2Repo repo = this.GetRepoFromPool();
+
                 if (repo != null)
                 {
                     result = function(repo);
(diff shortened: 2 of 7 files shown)


GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
...
         private readonly ITracer tracer;
+        private readonly object sharedRepoLock = new object();
+        private readonly TimeSpan sharedRepositoryDisposalPeriod;
+        private bool stopped;
+        private LibGit2Repo sharedRepo;
+        private Timer sharedRepoDisposalTimer;
+        private int activeCallers;
 
-        public LibGit2RepoPool(ITracer tracer, Func<LibGit2Repo> createRepo, int size)
+        public LibGit2RepoPool(ITracer tracer, Func<LibGit2Repo> createRepo, int size = 1, TimeSpan? disposalPeriod = null)
```
mjcheetham:
```csharp
int size = 1
```

I thought this PR was making the pool only ever have a size of one? Are we not removing this from the constructor?
(Also are there existing places that set `size` to specific figures for a particular reason?)

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
...
-using System.Collections.Concurrent;
 using System.Threading;
 
 namespace GVFS.Common
 {
     public class LibGit2RepoPool
     {
-        private const int TryAddTimeoutMilliseconds = 10;
-        private const int TryTakeTimeoutMilliseconds = Timeout.Infinite;
+        private static readonly TimeSpan DefaultRepositoryDisposalPeriod = TimeSpan.FromMinutes(15);
```
mjcheetham: I'm guessing 15 minutes is just a 'it feels about right' number?

What sort of custom values for expiration do you expect some potential callers to pick, out of interest?

ethomson:
I hope that you didn't take my word for it and looked through things?  If not, I should probably make sure my crazy claims were actually correct.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
...
+            this.sharedRepositoryDisposalPeriod = disposalPeriod ?? DefaultRepositoryDisposalPeriod;
+
+            this.sharedRepoDisposalTimer = new Timer(
+                (state) => this.DisposeSharedRepo(),
+                state: null,
+                dueTime: this.sharedRepositoryDisposalPeriod,
+                period: this.sharedRepositoryDisposalPeriod);
         }
 
         public void Dispose()
```
mjcheetham: I see we have a `Dispose` method already (not `IDisposable.Dispose` though).
Should we not just implement `IDisposable` for good measure? Are the cases when it's legitimate to create an `LibGit2RepoPool` without calling `Dispose`?

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
...
-            this.pool = new BlockingCollection<LibGit2Repo>();
-            for (int i = 0; i < size; ++i)
-            {
-                this.pool.Add(createRepo());
-            }
+            this.createRepo = createRepo;
+
+            this.sharedRepositoryDisposalPeriod = disposalPeriod ?? DefaultRepositoryDisposalPeriod;
+
+            this.sharedRepoDisposalTimer = new Timer(
```
mjcheetham: Should this timer not be stopped in `Dispose` too?

derrickstolee:
I definitely read some of the code and saw mutexes around the packfile loading logic, which is what we need. We have some massively parallel tests to help check.

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
...
+            {
+                this.sharedRepo?.Dispose();
+                this.sharedRepo = null;
+            }
         }
 
         public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult result)
         {
-            LibGit2Repo repo = null;
+            if (this.stopped)
```
mjcheetham: Is reading/writing the value of `this.stopped` safe with multiple threads? Should we not be in a lock, or using `Interlocked`? (I know there is no `bool` overload for compare/exchange atomic operations, but we could just use an `int` for the Boolean in that case.)

GVFS/GVFS.UnitTests/Common/LibGit2RepoPoolTests.cs
```diff
@@ -0,0 +1,174 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class LibGit2RepoPoolTests
+    {
+        private readonly TimeSpan disposalPeriod = TimeSpan.FromMilliseconds(1);
```
mjcheetham: What happens if the disposal interval is zero/instantaneous? Should we be checking for a non-zero interval in the constructor of `LibGit2RepoPool`?

derrickstolee:
Here is how I compared the multi-threaded performance of this PR versus #680:

```

                Random seeds = new Random();

                List<Thread> threads = new List<Thread>();
                for (int i = 0; i < numThreads; i++)
                {
                    Random r = new Random(seeds.Next());
                    int localI = i;
                    threads.Add(new Thread(() =>
                    {
                        byte[] buf = new byte[20];
                        int count = 0;

                        for (int j = 0; j < objectQueries; j++)
                        {
                            r.NextBytes(buf);

                            try
                            {
                                string sha = this.blobIds[(localI + j) % this.blobIds.Count];
                                this.Context.Repository.ObjectExists(sha);
                                this.Context.Repository.TryGetIsBlob(sha, out bool isBlob);
                                this.Context.Repository.TryGetBlobLength(sha, out long size);
                                count++;
                            }
                            catch (Exception e)
                            {
                                activity.RelatedWarning(this.CreateEventMetadata(e), $"Exception in thread {i}");
                            }
                        }

                        activity.RelatedInfo("Thread {0} saw {1} blobs", localI, count);
                    }));
                }
```

I ran this in the mount process, where `blobIds` is a list of 100 real blob ids in the Windows repo.

This leads to these numbers:

| PR   | Repo Type | Time |
|------|-----------|------|
| #680 | Multiple  | 11 s |
| #683 | Single    | 6 s  |

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
...
 
-        private readonly BlockingCollection<LibGit2Repo> pool;
+        private readonly Func<LibGit2Repo> createRepo;
         private readonly ITracer tracer;
+        private readonly object sharedRepoLock = new object();
+        private readonly TimeSpan sharedRepositoryDisposalPeriod;
+        private bool stopped;
+        private LibGit2Repo sharedRepo;
+        private Timer sharedRepoDisposalTimer;
+        private int activeCallers;
```
wilbaker: Mark `activeCallers` as `volatile` (or use `Interlocked.CompareExchange` in `DisposeSharedRepo`)

wilbaker: Actually I see the comparison is inside of `lock (this.sharedRepoLock)`.  I don't know enough about the interaction of `lock` and `Interlocked` to know if the `lock` guarantees that `DisposeSharedRepo` will read the latest value.

derrickstolee: I can mark it as `volatile` for the purpose of exposing it publicly, but doesn't `Interlocked.Decrement()` do the right thing?

derrickstolee: We are only reading from `activeCallers` in `DisposeSharedRepo()`, but I can make it `volatile` to make that read safer.

wilbaker: > I can mark it as volatile for the purpose of exposing it publicly, but doesn't Interlocked.Decrement() do the right thing?

The `Interlocked.Increment` and `Interlocked.Decrement` will do the right thing when modifying the values, but based on what I've read they alone do not guarantee that reads of the variable from a different thread will see the latest value (see discussion [here](https://stackoverflow.com/questions/3290945/what-to-use-instead-of-interlocked-equals)).

I did just find another discussion [here](https://stackoverflow.com/a/12843092) that mentions the memory barrier created by the `lock` will accomplish the same as `volatile`, and so based on that answer it sounds like the current code is ok as-is (as mentioned above I missed the `lock` on my first pass).  What do you think?

GVFS/GVFS.Common/LibGit2RepoPool.cs
```diff
@@ -1,47 +1,66 @@
 ﻿using GVFS.Common.Git;
 using GVFS.Common.Tracing;
 using System;
-using System.Collections.Concurrent;
 using System.Threading;
 
 namespace GVFS.Common
 {
     public class LibGit2RepoPool
```
kewillford: If we are going with a single object can we get rid of this class and have the single object and this code live in the `GitRepo`?  I would at least rename this class to `LibGit2RepoInvoker` or something similar since it isn't a pool anymore.

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: I'll make it `volatile` for extra help, but this check is actually useless. We check `this.stopped` under a lock in `GetRepoFromPool()`.

GVFS/GVFS.Common/LibGit2RepoPool.cs
derrickstolee: Can't hurt to make it `volatile`. The `lock` makes it work, for sure.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
```diff
@@ -0,0 +1,118 @@
...
+        public void Dispose()
+        {
+            this.stopped = true;
+
+            this.sharedRepoDisposalTimer?.Dispose();
+            this.sharedRepoDisposalTimer = null;
+
+            lock (this.sharedRepoLock)
+            {
+                this.sharedRepo?.Dispose();
```
kewillford: Do we need to check the `this.activeCallers` like in `DisposeSharedRepo`?

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
```diff
@@ -0,0 +1,118 @@
...
+                this.tracer.RelatedWarning("Exception while invoking libgit2: " + e.ToString(), Keywords.Telemetry);
+                throw;
+            }
+            finally
+            {
+                Interlocked.Decrement(ref this.activeCallers);
+            }
+        }
+
+        private LibGit2Repo GetRepoFromPool()
```
kewillford: Rename to `GetRepo` or `GetSharedRepo`?

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
```diff
@@ -0,0 +1,118 @@
...
+                this.sharedRepo?.Dispose();
+                this.sharedRepo = null;
+            }
+        }
+
+        public bool TryInvoke<TResult>(Func<LibGit2Repo, TResult> function, out TResult result)
+        {
+            try
+            {
+                Interlocked.Increment(ref this.activeCallers);
```
jeschu1: Would it have been an option to instead do a using statement (i.e.):
using (LibGit2Repo repo = this.createRepo()) {}
To handle the disposal?  This would mean we can have more than 1 instance of course but it ties the dispose up nicely.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (!disposalPeriod.HasValue || disposalPeriod.Value <= TimeSpan.Zero)
+            {
+                this.sharedRepositoryDisposalPeriod = DefaultRepositoryDisposalPeriod;
+            }
+            else
+            {
+                this.sharedRepositoryDisposalPeriod = disposalPeriod.Value;
+            }
+
+            this.sharedRepoDisposalTimer = new Timer(
```
jeschu1: There is a 15 minute cleanup window here.

What happens if Expire background job runs before the dispose?  Would it be successful but potentially leave a .idx (or many)?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -127,6 +127,12 @@ protected override void PerformMaintenance()
                         activity.RelatedWarning($"Skipping {nameof(PackfileMaintenanceStep)} due to git pids {string.Join(",", processIds)}", Keywords.Telemetry);
                         return;
                     }
+
+                    if (this.Context.Repository.HasActiveLibGit2Repo)
```
jeschu1: Could this be a race condition?  I.e. it's false here and then turns true before GetPackFilesInfo

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
```diff
@@ -0,0 +1,182 @@
+﻿using GVFS.Common.Git;
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Mock.Common;
+using NUnit.Framework;
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
```
jeschu1: I get excited every time I see unit tests!

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Threading;
+
+namespace GVFS.UnitTests.Common
+{
+    [TestFixture]
+    public class LibGit2RepoInvokerTests
+    {
+        private readonly TimeSpan disposalPeriod = TimeSpan.FromMilliseconds(1);
+        private MockTracer tracer;
+        private LibGit2RepoInvoker pool;
```
jeschu1: nit: is pool an accurate name?

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
```diff
@@ -0,0 +1,118 @@
...
+{
+    public class LibGit2RepoInvoker : IDisposable
+    {
+        private static readonly TimeSpan DefaultRepositoryDisposalPeriod = TimeSpan.FromMinutes(15);
+
+        private readonly Func<LibGit2Repo> createRepo;
+        private readonly ITracer tracer;
+        private readonly object sharedRepoLock = new object();
+        private readonly TimeSpan sharedRepositoryDisposalPeriod;
+        private volatile bool stopped;
```
kewillford: I would rename this to `disposing`

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
derrickstolee: We specifically don't want multiple instances, and don't want to dispose when there are many requests coming in.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
derrickstolee: We could leave .idx files on disk if libgit2 has a handle. That gets reported as part of the logging for that step.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: We can't stop libgit2 from becoming active during our git process, so we could still get IDX files. We'll try to delete them, and the next time we come around we will clean them up (because libgit2 has disposed in the meantime and dropped the handles).

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
jeschu1: My honest question is why don't we want multiple instances?  Is it because it will take too long?

I think a comment at the top of the file explaining the problem, and why we solve it in this way would be super helpful!

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
jeschu1: +1 on GetSharedRepo

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
jeschu1: That makes sense, but it's a bit confusing to have a check that isn't bulletproof.

Maybe add a comment explaining or just remove the check?  As you said we'll recover on the next run.  There are ways to make the check more bulletproof but probably overkill for our case.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
jeschu1: Instead of doing the Interlocked.Increment/Decrement we could do the whole procedure in a lock.  I guess the problem here is that we want explicity want to allow multiple function runs concurrently.  Questing only being of the added complexity.

derrickstolee: We can't do it in a lock, because then we can't have parallel invocations of the method we are calling. It is possible to do a lock around the changes to `activeCallers`, but `Interlocked` is faster than `lock ()`.

We _want_ to allow parallel access, but we don't need independent native memory blocks to make that happen. And there is a real cost to creating and disposing the libgit2 repo with each invocation. And based on tests with #680, we break things if we have no limit to the number of native repos we create.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: The only way to guarantee we won't hit libgit2 issues is to block use of the `LibGit2RepoInvoker` while this step is running. That is unacceptable as the invoker is used to hydrate the file system.

This is similar to how we don't run if there are active `git` processes. It keeps us from running twice while the same libgit2 repo is open across both runs. That allows us to clean up any stale IDX files created in the previous run.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
derrickstolee: This comment is interesting, and I haven't done anything about it in the code yet.

This `Dispose()` method is only called when we are shutting down the process. So someone would need to be hydrating at the same time they are running `gvfs unmount`. I'm pretty sure that already has unclear behavior, but we don't want to make it worse.

Perhaps the safest thing to do is to skip disposing the shared repo if there are active callers? It will get cleaned up as the process shuts down, so it's a safer response to that situation than the current behavior.

Thoughts?

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
jeschu1: @derrickstolee  that makes sense.  A think a comment on the top of the class saying something like:

Here we create  single instance of a LibGit2Repo that we attempt to clean up every 15 minutes.  It would be expensive to create/dispose a LibGit2Repo for every call so we do it on a timer schedule.  The background cleanup allows us to free memory and lingering unnecessary handles on files.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
jeschu1: Ah I see it's another fine as long as it didn't run across two runs check.  

Perhaps add a short comment saying something to effect of:
"If LibGit2Repo is active it may be spanning two calls to PackfileMaintenceStep which would be problamatic.  This check verifies that condition hasn't occurred."

That would remove any confusion and make it clear the condition we're trying to guard against.  I really should have added a similar comment in LooseObjectsStep.cs.  I'll make sure I'll do that next time I'm in that code.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
kewillford: Yeah the one thing I wanted to avoid was a crash because Dispose was called but somehow there was an active request to invoke that runs and we get either a disposed or null object exception because we would probably get support issues from people that ran `gvfs unmount` and had a popup for the crash or would see it in the event viewer.

derrickstolee:
I resolved all remaining comments, but we will hold this PR for after the next release. Thanks, everyone!

jrbriggs:
Now that master targets M149, I'm merging this PR.

PR #684, opened 2019-01-11
Adding dry-run and no-verify

alameenshah:
#### Goal
Create a `--dry-run` `upgrade` arg that will help VFS4Git Developers to run `gvfs upgrade` without actually affecting the state of their PC. New `--no-verify` `upgrade` arg to prevent code signature verification of installers and Nuget packages.

#### Feature details
Dry run - Behaves like a regular `gvfs upgrade --confirm` command, except that the downloaded installer exe's will not be run.

No verify - This is arg is enabled but not implemented right now.

### DIFF
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -32,6 +32,7 @@ namespace GVFS.Common
             {
                 UpgraderToolName,
                 UpgraderToolConfigFile,
+                "CommandLine.dll",
                 "GVFS.Common.dll",
                 "GVFS.Platform.Windows.dll",
                 "Microsoft.Diagnostics.Tracing.EventSource.dll",
@@ -305,12 +306,6 @@ namespace GVFS.Common
 
         protected virtual bool TryDeleteDownloadedAsset(Asset asset, out Exception exception)
         {
-            if (this.dryRun)
-            {
-                exception = null;
-                return true;
-            }
-
             return this.fileSystem.TryDeleteFile(asset.LocalPath, out exception);
         }
 
@@ -320,7 +315,7 @@ namespace GVFS.Common
 
             string downloadPath = ProductUpgraderInfo.GetAssetDownloadsPath();
             Exception exception;
-            if (!this.dryRun && GitHubUpgrader.TryCreateDirectory(downloadPath, out exception))
+            if (!GitHubUpgrader.TryCreateDirectory(downloadPath, out exception))
             {
                 errorMessage = exception.Message;
                 this.TraceException(exception, nameof(this.TryDownloadAsset), $"Error creating download directory {downloadPath}.");
@@ -332,11 +327,7 @@ namespace GVFS.Common
 
             try
             {
-                if (!this.dryRun)
-                {
-                    webClient.DownloadFile(asset.DownloadURL, localPath);
-                }
-
+                webClient.DownloadFile(asset.DownloadURL, localPath);
                 asset.LocalPath = localPath;
             }
             catch (WebException webException)
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -15,6 +15,9 @@ namespace GVFS.Common
 {
     public class GitHubUpgrader : IProductUpgrader
     {
+        protected bool dryRun;
+        protected bool noVerify;
+
         private const string GitHubReleaseURL = @"https://api.github.com/repos/microsoft/vfsforgit/releases";
         private const string JSONMediaType = @"application/vnd.github.v3+json";
         private const string UserAgent = @"GVFS_Auto_Upgrader";
@@ -46,8 +49,6 @@ namespace GVFS.Common
         private Release newestRelease;
         private PhysicalFileSystem fileSystem;
         private ITracer tracer;
-        private bool dryRun;
-        private bool noVerify;
 
         public GitHubUpgrader(
             string currentVersion,
@@ -375,18 +376,10 @@ namespace GVFS.Common
 
         protected virtual void RunInstaller(string path, string args, out int exitCode, out string error)
         {
-            if (this.dryRun)
-            {
-                exitCode = 0;
-                error = null;
-            }
-            else
-            {
-                ProcessResult processResult = ProcessHelper.Run(path, args);
+            ProcessResult processResult = ProcessHelper.Run(path, args);
 
-                exitCode = processResult.ExitCode;
-                error = processResult.Errors;
-            }
+            exitCode = processResult.ExitCode;
+            error = processResult.Errors;
         }
 
         private static bool TryCreateDirectory(string path, out Exception exception)
@@ -474,13 +467,16 @@ namespace GVFS.Common
             string installerArgs;
             if (this.TryGetLocalInstallerPath(assetId, out path, out installerArgs))
             {
-                string logFilePath = GVFSEnlistment.GetNewLogFileName(ProductUpgraderInfo.GetLogDirectoryPath(), Path.GetFileNameWithoutExtension(path));
-                string args = installerArgs + " /Log=" + logFilePath;
-                this.RunInstaller(path, args, out installerExitCode, out error);
-
-                if (installerExitCode != 0 && string.IsNullOrEmpty(error))
+                if (!this.dryRun)
                 {
-                    error = assetId + " installer failed. Error log: " + logFilePath;
+                    string logFilePath = GVFSEnlistment.GetNewLogFileName(ProductUpgraderInfo.GetLogDirectoryPath(), Path.GetFileNameWithoutExtension(path));
+                    string args = installerArgs + " /Log=" + logFilePath;
+                    this.RunInstaller(path, args, out installerExitCode, out error);
+
+                    if (installerExitCode != 0 && string.IsNullOrEmpty(error))
+                    {
+                        error = assetId + " installer failed. Error log: " + logFilePath;
+                    }
                 }
 
                 installerIsRun = true;
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -45,14 +45,20 @@ namespace GVFS.Common
         private Release newestRelease;
         private PhysicalFileSystem fileSystem;
         private ITracer tracer;
+        private bool dryRun;
+        private bool noVerify;
 
         public GitHubUpgrader(
             string currentVersion,
             ITracer tracer,
-            GitHubUpgraderConfig upgraderConfig)
+            GitHubUpgraderConfig upgraderConfig,
+            bool dryRun = false,
+            bool noVerify = false)
             : this(currentVersion, tracer)
         {
             this.Config = upgraderConfig;
+            this.dryRun = dryRun;
+            this.noVerify = noVerify;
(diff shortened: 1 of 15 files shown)


GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -72,10 +86,13 @@ private bool TryInitializeUpgrader(out string error)
                     jsonTracer.AddLogFileEventListener(logFilePath, EventLevel.Informational, Keywords.Any);
 
                     this.tracer = jsonTracer;
-                    this.prerunChecker = new InstallerPreRunChecker(this.tracer, this.Confirmed ? GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm : GVFSConstants.UpgradeVerbMessages.GVFSUpgrade);
+                    this.prerunChecker = new InstallerPreRunChecker(
```
wilbaker: As discussed offline, display an error (and fail the `gvfs upgrade`) if both `--dry-run` and `--confirm` are passed in.

jamill:
> looks like the "Files changed" tab is showing a lot of changes unrelated to adding dry-run and no-verify.

This is my fault - I updated the target branch. Tomorrow I will work with @alameenshah to fix up the branches.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: Added in `UpgradeVerb.cs` line no. 70

kewillford:
I haven't kept up on these changes but is there anyway we can simplify these changes to not propagate the dry run variable through the code.  That just looks messy and error prone.  In the ideal world I see something like
```
ConfirmedRun
Step1
Step2
Step3
Step4
Step5

DryRun
Step1
Step2
Step3
Step4
```

Maybe that is not possible with how the code is currently organized though and there is more that would affect the state.  I would at least expect some testing around the code to make sure we are running the code we expect when dry run is passed.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -32,13 +34,21 @@ public virtual bool TryRunPreUpgradeChecks(out string consoleError)
                 {
                     consoleError = $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is not supported in unattended mode";
                     this.tracer.RelatedError($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
-                    return false;
+
+                    if (!this.dryRun)
```
kewillford: Just so I understand correctly `dryRun` will continue to run even in error conditions?  Is this so developers can see all the possible errors that might be hit when the upgrade is ran?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -44,13 +44,39 @@ public UpgradeVerb()
...
         {
+            if (this.DryRun)
+            {
+                if (this.Confirmed)
+                {
+                    this.ReportErrorAndExit(this.tracer, ReturnCode.GenericError, "--dry-run and --confirm arguments are not compatible.");
+                }
+                else
+                {
+                    this.Confirmed = true;
```
kewillford: I'm confused why we are setting this to `true`.  Will you explain this please?

jrbriggs: +1 -- let's figure out a cleaner way to do this.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -128,6 +138,12 @@ protected virtual bool IsUnattended()
 
         protected virtual bool IsBlockingProcessRunning(out HashSet<string> processes)
         {
+            if (this.dryRun)
```
jrbriggs: Why is this here? I think dry-run should fail if a blocking process exists. The way I think about this is dry-run should be semantically identical to the real thing except that we actually run the installers.  We _should_ have dry run fail if there's a process hanging around, since this will be how we do functional tests, and our functional tests should give us an indication of expected reliability.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
jrbriggs: Like I said below, I think the only distinction between dry run and real installer is that dry run should not run the installers. I think everything else should still error out so we have a good understanding of potential reliability issues.

alameenshah:
> I haven't kept up on these changes but is there anyway we can simplify these changes to not propagate the dry run variable through the code. That just looks messy and error prone. In the ideal world I see something like
> 
> ```
> ConfirmedRun
> Step1
> Step2
> Step3
> Step4
> Step5
> 
> DryRun
> Step1
> Step2
> Step3
> Step4
> ```
> Maybe that is not possible with how the code is currently organized though and there is more that would affect the state. I would at least expect some testing around the code to make sure we are running the code we expect when dry run is passed.

Have added couple UT cases for --dry-run.
1. Verify that `UpgradeVerb` behaves like during a regular install and launches upgrader tool.
2. Verify that `UpgradeOrchestrator` completes successfully but does not run the installer exes.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
alameenshah: Enabled pre-checks for dry-run.

GVFS/GVFS.Upgrader/App.config
```diff
@@ -1,6 +1,14 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <startup> 
         <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
     </startup>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
```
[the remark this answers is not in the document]
alameenshah: it is not needed. removed.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: agree. have tried to simplify please take a look.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockGitHubUpgrader.cs
```diff
@@ -56,6 +61,11 @@ public void ResetFailedAction()
             this.failActionTypes = ActionType.Invalid;
         }
 
+        public bool RunInstallerInvoked()
+        {
+            return this.InstallerArgs.Count > 0;
```
jamill: This seems like a bit indirect way to check this... on 1st glance, I wasn't sure what `InstallerArgs` had to do with checking whrether an installer was invoked. Maybe a comment that InstallerArgs captures the args from each of the install commands, or tweak the variable name?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -44,6 +44,20 @@ public UpgradeVerb()
...
+            Default = false,
+            Required = false,
+            HelpText = "Display progress and errors, but don't install GVFS")]
+        public bool DryRun { get; set; }
+
+        [Option(
+            "no-verify",
+            Default = false,
+            Required = false,
+            HelpText = "Don't verify authenticode signature of installers")]
```
jrbriggs: Please mark this as internal use only or hidden.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -72,10 +92,25 @@ private bool TryInitializeUpgrader(out string error)
...
+                    else if (this.DryRun)
+                    {
+                        commandToRerun = GVFSConstants.UpgradeVerbMessages.GVFSUpgradeDryRun;
+                    }
+                    else
+                    {
+                        commandToRerun = GVFSConstants.UpgradeVerbMessages.GVFSUpgrade;
+                    }
+
+                    this.prerunChecker = new InstallerPreRunChecker(this.tracer, commandToRerun);
```
jrbriggs: This seems pretty complicated to let the user know to re-run with --confirm. Do you think the distinction between --dry-run and --confirm is important here? Should we simplify by having the `InstallerPreRunChecker` always assume you want to end up with --confirm?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -214,7 +249,12 @@ private bool TryCopyUpgradeTool(out string upgraderExePath, out string consoleEr
 
             using (ITracer activity = this.tracer.StartActivity(nameof(this.TryCopyUpgradeTool), EventLevel.Informational))
             {
-                if (!this.upgrader.TrySetupToolsDirectory(out upgraderExePath, out consoleError))
+                if (this.DryRun)
```
jrbriggs: Why do we not copy the upgrade to the alternate directory in dry-run mode? Could this mask an issue during functional tests?

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -230,7 +270,8 @@ private bool TryLaunchUpgradeTool(string path, out string consoleError)
             using (ITracer activity = this.tracer.StartActivity(nameof(this.TryLaunchUpgradeTool), EventLevel.Informational))
             {
                 Exception exception;
-                if (!this.processLauncher.TryStart(path, out exception))
+                string args = string.Empty + (this.DryRun ? " --dry-run" : string.Empty) + (this.NoVerify ? " --no-verify" : string.Empty);
```
jrbriggs: Named constants for --dry-run and --no-verify, and use them throughout?

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockGitHubUpgrader.cs
alameenshah: right. replaced it with a more obvious field - `InstallerExeLaunched`.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: agree... reverted.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
alameenshah: yes possible...good catch.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -49,10 +53,14 @@ public class GitHubUpgrader : IProductUpgrader
         public GitHubUpgrader(
             string currentVersion,
             ITracer tracer,
-            GitHubUpgraderConfig upgraderConfig)
+            GitHubUpgraderConfig upgraderConfig,
+            bool dryRun = false,
+            bool noVerify = false)
```
kewillford: My only thought in looking through this is if we have more options we will want to refactor this so that we don't end up with a bunch of `bool` parameters but instead have an `UpgraderOptions` class that hold all of them including what the `GitHubUpgraderConfig` has but since we only have 2 right now I don't think we need that.

Issue #638, opened 2018-12-18
Mac: Install kext logging daemon and capture its logs with `gvfs diagnose`


wilbaker:
Once #637 is completed we'll need to install the daemon and capture its logs with `gvfs diagnose`.

PR #686, opened 2019-01-12
Add PrjFSKext log collection to diagnose verb

nickgra:
Part of #638 

The new PrjFSKextLogDaemon writes kext log messages out to the os log.
This change updates the diagnose verb to consume the os log via the
`log show` commandline.

This change does not install the daemon as part of the product.

Note that this effectively exports the logs as text rather than
the binary .logarchive format. This was chosen for ease of viewing
on any machine (Console.app and `log` are only on MacOS).

The questions here in my mind are about what args we should pass into `log`. It's very powerful and we could filter by type, content in the message, date, etc. I went with the simplest solution, put everything in the log file. I personally prefer logs structured like this rather than tools that produce separate log files for stdout/warnings/errors, but I'm open to new ideas :)

For more context on the unified logger, check https://developer.apple.com/documentation/os/logging or the manpage for `log`

### DIFF
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -14,7 +14,13 @@ namespace GVFS.Platform.Mac
 
         public bool EnumerationExpandsDirectories { get; } = true;
 
-        public string LogsFolderPath => throw new NotImplementedException();
+        public string LogsFolderPath
+        {
+            get
+            {
+                return Path.Combine(System.IO.Path.GetTempPath(), "PrjFSKext");
+            }
+        }
 
         public bool IsGVFSUpgradeSupported()
         {
@@ -42,9 +48,14 @@ namespace GVFS.Platform.Mac
             return true;
         }
 
-        public string FlushLogs()
+        public bool TryFlushLogs(out string error)
         {
-            throw new NotImplementedException();
+            Directory.CreateDirectory(this.LogsFolderPath);
+            ProcessResult logShowOutput = ProcessHelper.Run("log", args: "show --predicate \"subsystem contains \'org.vfsforgit\'\" --info", redirectOutput: true);
+            File.WriteAllText(Path.Combine(this.LogsFolderPath, "PrjFSKext.log"), logShowOutput.Output);
+            error = string.Empty;
+
+            return true;
         }
 
         public bool IsReady(JsonTracer tracer, string enlistmentRoot, out string error)
--- a/GVFS/GVFS.Common/FileSystem/IKernelDriver.cs
+++ b/GVFS/GVFS.Common/FileSystem/IKernelDriver.cs
@@ -8,7 +8,7 @@ namespace GVFS.Common.FileSystem
         bool EnumerationExpandsDirectories { get; }
         string LogsFolderPath { get; }
         bool IsSupported(string normalizedEnlistmentRootPath, out string warning, out string error);
-        string FlushLogs();
+        bool TryFlushLogs(out string errors);
         bool TryPrepareFolderForCallbacks(string folderPath, out string error, out Exception exception);
         bool IsReady(JsonTracer tracer, string enlistmentRoot, out string error);
         bool IsGVFSUpgradeSupported();
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -1,4 +1,4 @@
-﻿using GVFS.Common.FileSystem;
+﻿﻿using GVFS.Common.FileSystem;
 using GVFS.Common.Git;
 using GVFS.Common.Tracing;
 using System;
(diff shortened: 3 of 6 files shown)


GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -44,16 +50,21 @@ public bool IsSupported(string normalizedEnlistmentRootPath, out string warning,
 
         public string FlushLogs()
         {
-            throw new NotImplementedException();
+            Directory.CreateDirectory(this.LogsFolderPath);
+            ProcessResult logShowOutput = ProcessHelper.Run("log", args: "show --predicate \"subsystem contains \'org.vfsforgit\'\"", redirectOutput: true);
+            File.WriteAllText(Path.Combine(this.LogsFolderPath, "PrjFSKext.log"), logShowOutput.Output);
+
+            // Unlike FlushLogs on Windows, we will always produce a log file, even if it's empty if there is no results in the log subsystem.
+            return string.Empty;
```
jeschu1: What does this function always return string.empty, I guess a void isn't an option here?

pmj:
> The questions here in my mind are about what args we should pass into `log`. It's very powerful and we could filter by type, content in the message, date, etc. I went with the simplest solution, put everything in the log file. I personally prefer logs structured like this rather than tools that produce separate log files for stdout/warnings/errors, but I'm open to new ideas :)

Yeah, I think there's no point overcomplicating things here. You could include `--info` to get the more verbose messages. Unfortunately, there's no `--since-last-boot` flag, otherwise that would be a good way to reduce the amount of noise. I guess we could work out what to pass to `--start` based on uptime. As with any time & date calculations though: here be dragons. So probably not worth it.

Overall: Looks good to me. The only thing that I could imagine might be unexpected to users is that the `log show` command can take a long time to complete (minutes). So that's perhaps something the tool should point out to users, as it might just look like it's sitting there doing nothing.

nickgra:
> > The questions here in my mind are about what args we should pass into `log`. It's very powerful and we could filter by type, content in the message, date, etc. I went with the simplest solution, put everything in the log file. I personally prefer logs structured like this rather than tools that produce separate log files for stdout/warnings/errors, but I'm open to new ideas :)
> 
> Yeah, I think there's no point overcomplicating things here. You could include `--info` to get the more verbose messages. Unfortunately, there's no `--since-last-boot` flag, otherwise that would be a good way to reduce the amount of noise. I guess we could work out what to pass to `--start` based on uptime. As with any time & date calculations though: here be dragons. So probably not worth it.
> 
> Overall: Looks good to me. The only thing that I could imagine might be unexpected to users is that the `log show` command can take a long time to complete (minutes). So that's perhaps something the tool should point out to users, as it might just look like it's sitting there doing nothing.

Thanks for catching that I'm not including `info` messages, I forgot those weren't included by default. I'll make that change now.

One great thing about the logging subsystem is that if we do have a user who is having a complicated issue, we could hand craft a command to extract specifically what logs we need for investigation if the default option isn't enough (or is too noisy).

Diagnose already takes a bit of time and has a spinner showing that we're working so I don't think we need to be too worried about users being concerned that we're hung at this time. If it ends up being a support issue, we can take another change here.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -44,16 +50,21 @@ public bool IsSupported(string normalizedEnlistmentRootPath, out string warning,
 
         public string FlushLogs()
         {
-            throw new NotImplementedException();
+            Directory.CreateDirectory(this.LogsFolderPath);
+            ProcessResult logShowOutput = ProcessHelper.Run("log", args: "show --predicate \"subsystem contains \'org.vfsforgit\'\"", redirectOutput: true);
+            File.WriteAllText(Path.Combine(this.LogsFolderPath, "PrjFSKext.log"), logShowOutput.Output);
+
```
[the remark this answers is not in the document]
nickgra: Exactly! Phil's latest PR implements a daemon style logger that we'll set up as part of our installer. It'll route all messages from the kext into MacOS's built in logging subsystem. Once it's in there, we can easily query the information and put in a log file.

The big advantage here (in my mind) is that we'll always be logging so you'll have all the messages in the logging subsystem (depending on how they're classified and how volatile those messages are) so messages won't get lost if you forget to restart the logger between reloading the kext.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
nickgra: The FlushLogs method that we expose in the Kernel Driver interface is rather... strange, likely because it was designed around how PrjFlt handles logs on Windows. The Windows flavor of this method (in `ProjFSFilter.cs`) potentially returns a string of errors containing exceptions on why we failed to extract the logs from PrjFlt for consumption in Diagnose. 

I could rework the interface as part of this change, but the complexity around how PrjFlt logs will have to be handled somewhere so I went for the cheapest fix.

I'm happy to rework if people have opinions on this!

/cc: @wilbaker

jeschu1: @nickgra I think it's fine as it, if we can clarify a bit in the comment that would be great!

wilbaker: > I'm happy to rework if people have opinions on this!

@nickgra how about updating this to our usual TryXXX pattern:

```
public bool TryFlushLogs(out string error)
```

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -44,16 +50,21 @@ public bool IsSupported(string normalizedEnlistmentRootPath, out string warning,
 
         public string FlushLogs()
         {
-            throw new NotImplementedException();
+            Directory.CreateDirectory(this.LogsFolderPath);
+            ProcessResult logShowOutput = ProcessHelper.Run("log", args: "show --predicate \"subsystem contains \'org.vfsforgit\'\" --info", redirectOutput: true);
+            File.WriteAllText(Path.Combine(this.LogsFolderPath, "PrjFSKext.log"), logShowOutput.Output);
```
wilbaker: This might be a lot of text to store in memory.  Can we stream the log from stdOut to the log file?

wilbaker: @nickgra could you file a follow-up work item to stop using `File.WriteAllText`?

nickgra: #743

PR #688, opened 2019-01-15
NuGetUpgrader: do not report version for non-versioned entries

jamill:


GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -291,6 +291,10 @@ public bool TryRunInstaller(InstallActionWrapper installActionWrapper, out strin
                             $"Running install action: Name: {entry.Name}, Version: {entry.Version}" +
                             $"InstallerPath: {installerPath} Args: {entry.Args}");
 
+                        string progressMessage = string.IsNullOrWhiteSpace(entry.Version) ?
+                            $"Installing {entry.Name}" :
+                            $"Installing {entry.Name}, version {entry.Version}";
```
mjcheetham: Is there a reason to change the output to be `Installing foo, version bar` from `Installing foo Version: bar`?
Personally I'd say `Installing foo (version bar)` looks nicer (and just `Installing foo` when there is no version information) 😁

jamill: I like that suggestion - thanks!

I am also going to tweak this to `Running foo`, and then by convention the name of the install action will be the installer / script. I think this looks better for now, and I expect we might tweak this based on feedback.

PR #691, opened 2019-01-15
NuGetUpgrader: add ability to replace well known tokens in manifest entry

jamill:
Motivation:
We want the ability to have the installers write out logs to unique log files. The install-manifest does not know the path to write the log file when it is being generated, nor is it able to specify a unique name to write the log file out it.

This change enables that scenario by allowing the install-manifest to contain well known tokens in the args field that the installer will replace before it runs the install command.

The install-manifest would then have the following as part of its `Args` data:

`/LOG={logpath}\\VfsForGitUpgrader_GitInstaller_{timestamp}.log`

TODO:
- [x] Add unit tests


jamill:
The other option I had been considering was adding a `LogArgument` field to the install-manifest entry, which the NuGet Upgrader would look for, and if it exists, then append `/{LogArgument} path_to_log` to the argument string passed to the installer command

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -349,6 +358,16 @@ private static bool TryGetPersonalAccessToken(string gitBinaryPath, string crede
             return gitProcess.TryGetCredentials(tracer, credentialUrl, out string username, out token, out error);
         }
 
+        private string ReplaceTokens(string src)
+        {
+            string logFilePath = ProductUpgraderInfo.GetLogDirectoryPath();
+            string timeStamp = DateTime.Now.ToString("yyyyMMdd_HHmmss");
+
+            string dst = src.Replace("{logpath}", logFilePath)
```
wilbaker: Style suggestion:

Use named constants for `"{logpath}"` and `"{timestamp}"` (perhaps put them in a class that has the well known manifest tokens?)

PR #693, opened 2019-01-16
GitHubUpgrader: Use REMOUNTREPOS flag instead of MOUNTREPOS

jamill:
VFSForGit Setup is looking for REMOUNTREPOS instead of MOUNTREPOS.

### DIFF
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -22,7 +22,7 @@ namespace GVFS.Common
         private const string JSONMediaType = @"application/vnd.github.v3+json";
         private const string UserAgent = @"GVFS_Auto_Upgrader";
         private const string CommonInstallerArgs = "/VERYSILENT /CLOSEAPPLICATIONS /SUPPRESSMSGBOXES /NORESTART";
-        private const string GVFSInstallerArgs = CommonInstallerArgs + " /MOUNTREPOS=false";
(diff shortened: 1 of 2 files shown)


jamill:
Assuming this is the right change - I will fix up the unit tests.

derrickstolee:
I see that in Setup.iss we use `REMOUNTREPOS`, but not `MOUNTREPOS`, so the existing behavior is clearly incorrect. I do think we want to pass this option so users do not need to remount themselves.

@jamill: this isn't related to the background Git processes we saw, right? Because it looks like we unmount no matter what. This only controls mounting the repos again, right?

jamill:
> I see that in Setup.iss we use REMOUNTREPOS

yeah - I meant to include a link to these lines in Setup.iss, thanks for tracking it down and checking. I will update PR with a link to this a bit later.

> I do think we want to pass this option so users do not need to remount themselves.

When the upgrader is running the installation, we do not want the VFSForGit installer to remount the repositories, as the upgrader is going  to explicitly handle this in another step. The problems with the VFSForGit installer performing the mount itself are that:

 1. If remounting of any of the repositories fails, then the installation reports failure. We do not want a mount failure to fail the entire setup, as VFSForGit would otherwise be installed correctly.
 2. We would attempt remounting 2X - one by the VFSForGit installer, and one by the upgrader driving the process. 

> @jamill: this isn't related to the background Git processes we saw, right? Because it looks like we unmount no matter what. This only controls mounting the repos again, right?

Correct.

Thanks for taking a look!

PR #698, opened 2019-01-16
Mac/Linux: default MirrorProvider paths to $HOME and permit override

chrisd8088:
Bring the MirrorProvider scripts for macOS in sync with the changes from the Linux scripts (see commits f9fdb09, 1bd487e, a223566).  This adds support for non-default paths to be supplied
via environment variables when cloning or mounting the MirrorProvider.

Also use `$HOME` in the definition of the default ~/TestRoot and ~/PathToMirror paths instead of the tilde character, as Bash won't expand it when it's inside a quoted string.

### DIFF
--- a/MirrorProvider/Scripts/Linux/MirrorProvider_Clone.sh
+++ b/MirrorProvider/Scripts/Linux/MirrorProvider_Clone.sh
@@ -8,7 +8,7 @@ fi
 SCRIPTDIR=$(dirname ${BASH_SOURCE[0]})
 BUILDDIR=$SCRIPTDIR/../../../../BuildOutput/MirrorProvider.Linux/bin/$CONFIGURATION/x64/netcoreapp2.1
 
-PATH_TO_MIRROR="${PATH_TO_MIRROR:-~/PathToMirror}"
-TEST_ROOT="${TEST_ROOT:-~/TestRoot}"
+PATH_TO_MIRROR="${PATH_TO_MIRROR:-$HOME/PathToMirror}"
+TEST_ROOT="${TEST_ROOT:-$HOME/TestRoot}"
 
 dotnet $BUILDDIR/MirrorProvider.Linux.dll clone "$PATH_TO_MIRROR" "$TEST_ROOT"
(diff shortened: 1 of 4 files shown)


chrisd8088:
Hi @jrbriggs ... this is one of three small PRs I hope to push upstream for the Linux client work.  I'm not sure why the Windows functional tests aren't working, as we haven't touched that code at all.

Is it possible to re-run that check?  And/or, is it possible this features/linuxprototype branch which @sanoursa made some months ago has diverged enough from the Windows master branch that it no longer works against your CI infrastructure?

chrisd8088:
Digging into the failing logs, @jrbriggs, I see this:
```
Access Denied: The Personal Access Token used has expired.
```
My guess then is that we can pull in all the changes since the branch was made and that'll probably fix the problem.  Do you have a workflow preference here?  Shall I submit a PR which just pulls in changes from the master branch?

jrbriggs:
@chrisd8088 good find. Yep, we've moved to anonymous auth against the service and the feature branch is old enough not to have that change.

We generally rebase feature branches up to the latest version to avoid some wonky graph shapes, but a PR would suffice.

jrbriggs:
@chrisd8088 I rebased features/linuxprototype up to the latest in master. Can you take a look at the conflicts in the Linux scripts?

chrisd8088:
Yay!  Thanks, @jrbriggs -- I think this is looking all green, now.  🍏

chrisd8088:
Hey @jrbriggs, a caution ⚠️  -- before you consider merging this, though, I think perhaps I ought to redo it against the now-rebased Microsoft/VFSForGit `features/linuxprototype` branch.

GitHub is saying I want to merge 24 commits, but it's really only 2; the difference is due to prior commits in `features/linuxprototype` having been (I believe) replayed with different IDs:

https://github.com/Microsoft/VFSForGit/compare/master...github:mac-linux-mirror-homedir

So before you pull anything in, and unless you say otherwise, I may just redo these minor changes in a fresh PR.  Sound OK to you?

jrbriggs:
Sounds good to me @chrisd8088, thanks!

PR #708, opened 2019-01-17
First Kext Unit Tests as POC

jeschu1:
This is a very simple POC to show how we can start testing functions in the Kext.

It's running in the build now.


ProjFS.Mac/PrjFSKextTests/UtilitiesTests.m
```diff
@@ -0,0 +1,18 @@
...
+@end
+
+@implementation UtilitiesTests
+
+- (void)testActionBitIsSet {
+    XCTAssertTrue(ActionBitIsSet(1,1));
+    XCTAssertTrue(ActionBitIsSet(3,1));
+    XCTAssertTrue(ActionBitIsSet(5,1));
+    XCTAssertFalse(ActionBitIsSet(6,1));
+    XCTAssertTrue(ActionBitIsSet(6,2));
```
pmj: I realise the function under test is extremely basic and so this is more of a demo, but is there any particular reason you're just using integer literals rather than the symbolic `KAUTH_VNODE_*` constants it would actually be used with in practice? `#include <sys/kauth.h>` in userspace seems to work just fine and makes those constants available in the test method. It would be nice if the tests were at least loosely based in reality.

jeschu1: I was debating if we should keep to primitive types / mocks, but I agree the test looks old.  Further to that we use Utilities which includes <sys/kernel_types.h> which pulls in the constants anyways.  I've updated the test to bemore realistic, thanks!

jeschu1:
> This is certainly a decent start - the technique of pulling the function out into a separate file that doesn't include any "difficult" kernel header files will certainly get us _somewhere_ with testing simple leaf functions like that. I'm not sure we want to continue that trend too far though, as it seems like we'll unnecessarily pull functions out of their original context, plus it muddies the version history.

I have a feeling we will end up refactoring in our quest to provide better test coverage.  That will indeed muddy history so something to consider.

Including KauthHandler.cpp in it's current state for testing would have been quite difficult.  So I moved the function I wanted to test to a Utilities class which is then including.  @wilbaker what are your thoughts on this pattern and if it's a go can you suggest a naming convention?

Also as I'm sure everyone is aware generally only public methods are tested in unit tests, however given the complexity of calling the public methods I think we're going to have to get creative to test as much as we can (at least in the initial phases).

pmj:
> Also as I'm sure everyone is aware generally only public methods are tested in unit tests, however given the complexity of calling the public methods I think we're going to have to get creative to test as much as we can (at least in the initial phases).

Yeah, that rule is probably fine for OOP style, but I very much doubt it's helpful for our purposes. Functions currently have external linkage and are declared in the header function based on whether we need to call them from outside the compilation unit they're defined in. The only truly public functions we have are the kext start & stop functions, and the overridden public functions in our IOService classes, so that wouldn't be much to test.

I hinted at a pattern for extracting functions with local linkage (declared `static`) when you were screen sharing - basically hide the `static` specifier behind a macro that we define as, well, `static` in the real kext and nothing at all in the unit test build. We can then declare all these "testable" static functions in a special header file which the test suite can include and which isn't used in the kext itself.

But yeah, this relies on our ability to compile the kext cpp outside the kext. I'd like to *try* to make that happen at the very least because having glue code all over the place just to make the kext code testable is going to hurt readability.

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
```diff
@@ -934,6 +996,60 @@
...
+				CODE_SIGN_STYLE = Automatic;
+				COMBINE_HIDPI_IMAGES = YES;
+				DEVELOPMENT_TEAM = UBF8T346G9;
+				INFOPLIST_FILE = PrjFSKextTests/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/../Frameworks",
+					"@loader_path/../Frameworks",
+				);
+				MACOSX_DEPLOYMENT_TARGET = 10.14;
```
pmj: While playing around with this code, I noticed that the deployment target is explicitly set to 10.14 in the test suite target. That should probably just be inherited from the project, which uses 10.13.

pmj:
> But yeah, this relies on our ability to compile the kext cpp outside the kext. I'd like to _try_ to make that happen at the very least because having glue code all over the place just to make the kext code testable is going to hurt readability.

I appear to have made this approach work: https://github.com/pmj/VFSForGit/tree/kext_unit_test

This works by compiling the kext code (currently only KauthHandler.cpp) in "standalone" mode (no C/C++ standard library, like in the kernel) and using the Kernel.framework headers. The kext code is built into a static library, and I've linked your test bundle against it. The kext code depends on a lot of kernel functions which are currently not yet mocked, so I've enabled dead code stripping so that functions that aren't called in tests are removed (so the functions they depend on don't cause link errors). The implementation of `ActionBitIsSet()` can therefore remain in KauthHandler.cpp and still be called from the test suite. (I get around the `static inline` using a macro.)

Anyway, this approach isn't going to magically solve all our problems - getting the mock code and tests to compile is still going to be tricky, but at least the problem should be largely solved for the kext itself. (Obviously, this is just a suggestion, we don't have to go with this approach.)

wilbaker:
> I hinted at a pattern for extracting functions with local linkage (declared static) when you were screen sharing - basically hide the static specifier behind a macro that we define as, well, static in the real kext and nothing at all in the unit test build. We can then declare all these "testable" static functions in a special header file which the test suite can include and which isn't used in the kext itself.

@pmj thanks for this suggestion, and posting a proof of concept!  

I like that this approach does not require moving static functions out of their cpp files.

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -12,7 +12,19 @@ PACKAGES=$ROOTDIR/packages
...
 
-xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj build -scheme 'Build All' -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
+xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj  -scheme 'Build All' -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native build || exit 1
+
+if !(gem list --local | grep xcpretty); then
+   sudo gem install xcpretty
+fi
+
+# Run Tests and put output into a xml file
+xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj -scheme 'Build All' test | xcpretty -r junit --output $PROJFS/TestResultJunit.xml
```
jeschu1: @jrbriggs updated the build so the failed test appear in the Test category properly.

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -12,7 +12,19 @@ PACKAGES=$ROOTDIR/packages
 
 PROJFS=$SRCDIR/ProjFS.Mac
 
-xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj build -scheme 'Build All' -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native || exit 1
+xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj  -scheme 'Build All' -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Native build || exit 1
+
+if !(gem list --local | grep xcpretty); then
+   sudo gem install xcpretty
```
jeschu1: This may prompt for a password if required.  It would only do this once to install the gem, once installed no more prompts.

ProjFS.Mac/Scripts/Build.sh
jrbriggs: @jeschu1 thanks! can you look at why the functional test run isn't working the same way?

ProjFS.Mac/Scripts/Build.sh
pmj: If I ran this as a user and it asked me for a password, I'd start digging through the shell source to find out what it was trying to do. Perhaps we can echo something like "Attempting to run 'sudo gem install xcpretty'. This may ask you for your password to gain admin privileges."

ProjFS.Mac/Scripts/Build.sh
jrbriggs: @jeschu1 that's a windows build -- can you double-check?

jeschu1: https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=8278228&view=ms.vss-test-web.build-test-results-tab correct link :-)

ProjFS.Mac/PrjFSKext/public/PrjFSCommon.h
```diff
@@ -34,4 +34,14 @@ enum PrjFSServiceUserClientType
...
 };
 
+// When building the kext in user space for unit testing, we want some functions
+// to have external linkage in order to make them testable
+#ifdef KEXT_UNIT_TESTING
+#define KEXT_TESTABLE_STATIC
+#define KEXT_TESTABLE_STATIC_INLINE
+#else
+#define KEXT_TESTABLE_STATIC static
+#define KEXT_TESTABLE_STATIC_INLINE static inline
```
jrbriggs: What about just `KEXT_STATIC` and `KEXT_STATIC_INLINE`. I'm not sure the `TESTABLE` thoughout the codebase adds much extra value. It's the comment on the defines that tells me what we're making this a macro.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -0,0 +1,34 @@
+#import <XCTest/XCTest.h>
+#include "../PrjFSKext/KauthHandlerTestable.hpp"
+
+@interface KauthHandlerTests : XCTestCase
+@end
+
+@implementation KauthHandlerTests
+
+- (void)testActionBitIsSet {
+    XCTAssertTrue(ActionBitIsSet(KAUTH_VNODE_READ_DATA, KAUTH_VNODE_READ_DATA));
```
jrbriggs: Nit: Let's also add a few tests for multiple bits set.

pmj:
> I like the approach!

I think the real test (as it were) will be when we can start adding mock implementations for some KPI functions. I'm pretty optimistic about it though.

ProjFS.Mac/Scripts/Build.sh
jeschu1: @jrbriggs apologies bad link.  I purposely broke the build so you have look now.  Check out the failures.

ProjFS.Mac/PrjFSKext/public/PrjFSCommon.h
jeschu1: Yes!  @wilbaker and I actually talked about the same name change offline.  I'll make it happen.

ProjFS.Mac/PrjFSKextTests/PrjFSKextTests.exp
```diff
@@ -0,0 +1,6 @@
+# The explicit exported symbol list allows dead code stripping to work on the
+# test bundle binary.
+
+_OBJC_CLASS_$*
+_OBJC_METACLASS_$*
+
```
wilbaker: @pmj, what error(s) were you seeing without this file?

pmj: As the comment suggests, dead code stripping does nothing unless we explicitly specify the exported symbols. What's happening is:

 - The test target is a "bundle" executable. This is normally used for plugins etc. - essentially, the test bundle is dynamically loaded into the test runner.
 - Dynamic libraries export all symbols with external linkage by default, i.e. any functions not explicitly marked as having hidden or local (`static`) linkage.
 - This means dead code stripping can't really strip anything, as all functions are exported by default.
 - Many functions in KauthHandler.cpp in turn depend on various KPI functions and other function in the kext. These aren't (yet) defined in the test target bundle, so they produce link errors.

Exporting only the Objective-C classes (test case classes) means only the functions called by test methods survive dead code stripping, so we don't immediately depend on the world, only on the true dependencies of the tested functions.

wilbaker: Ahh that makes sense, thanks!

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
```diff
@@ -242,6 +277,7 @@
 				4391F88F21E42AC40008103C /* Info.plist */,
 				4391F89321E42AC40008103C /* KauthHandler.cpp */,
 				4391F8A421E42AC50008103C /* KauthHandler.hpp */,
+				F5E39C8121F11556006D65C2 /* KauthHandlerTestable.hpp */,
```
wilbaker: Can `KauthHandlerTestable.hpp` be moved into a unit testing specific folder (possibly `PrjFSKextsTests`)?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -16,6 +16,7 @@
 #include "PrjFSProviderUserClient.hpp"
 #include "PerformanceTracing.hpp"
 #include "kernel-header-wrappers/mount.h"
+#include "KauthHandlerTestable.hpp"
```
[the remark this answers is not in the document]
pmj: Your `#error` suggestion, the removal of the macros from the header and to some extent moving the header elsewhere depends on removing this `#include` so I won't comment on those separately.

We *can* remove this header, but I'm not convinced the upside makes up for the downside: by including it in the kext code, the compiler can check that the header and implementation have not diverged. To some extent, changes to the functions' signatures would be caught by the linker thanks to C++'s mangled names (although the error messages would not necessarily be clear on the source of the problem). If the return type changed in the implementation, and was not updated in the testable function header, that would go silently unnoticed.

It's maybe not a huge problem - in many cases, our unit cases would start failing or crashing. As far as I can tell, the advantage of decoupling the kext code completely is that we save an `#include` per tested kext cpp file, and perhaps a little bit of extra maintenance in the testable header to keep it compiling in the 3 different build environments. That doesn't seem a big enough advantage to give up that bit of compiler checking?

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -12,7 +12,20 @@ PACKAGES=$ROOTDIR/packages
...
+  sudo gem install xcpretty
+fi
+
+# Run Tests and put output into a xml file
+xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj -scheme 'Build All' test | xcpretty -r junit --output $PROJFS/TestResultJunit.xml
+
+#Check to see if the test results file contains a failure
+if grep -q '<failure' $PROJFS/TestResultJunit.xml; then
+  exit 1
+fi
```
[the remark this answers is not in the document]
jeschu1: Much better solution thanks @wilbaker!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
wilbaker: Thanks for the details @pmj, I had not considered that the compiler would help ensure that these had not diverged.

My primary concern had been protecting us from having unit test specific code leak into production builds.

Perhaps we can can achieve both of these goals by wrapping everything inside KauthHandlerTestable.hpp with a `#ifdef KEXT_UNIT_TESTING`.

That way the contents of the header will still be compiled when building the unit tests, but they will not be present when building the product.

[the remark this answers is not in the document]
jeschu1: @wilbaker  I updated it to be a #ifdef so we only include the file if we're testing

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
jeschu1: We could, but it honestly seemed a bit easier to find/understand having it as a single file along-side the others.  The file name is pretty self explanatory.  I could honestly go either way here.

wilbaker: Now that we're sticking with including the header in the .cpp I think it's fine to leave it side by side.

PR #710, opened 2019-01-18
File system common methods

jamill:
To reduce some of the code changes in the `NuGetUpgrader` code change, I am moving some of the encapsulated changes into their own PR. This PR moves / introduces a couple of helper methods around file system operations.

1) `TryCreateDirectory` is a general helper method that exists in `GitHubUpgrader` today, but seems like it is generally useful and could live in `PhysicalFileSystem`. The upcoming `NuGetUpgrader` change will leverage this method as well.

2) `TryDeleteDirectory` is not used today, but will be used by the `NuGetUpgrader` change. I wanted t

Also  removes an unused const while we are in here.

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
@@ -260,6 +260,51 @@ namespace GVFS.Common.FileSystem
             }
         }
 
+        public bool TryCreateDirectory(string path, out Exception exception)
+        {
+            try
+            {
+                Directory.CreateDirectory(path);
+            }
+            catch (Exception e) when (e is IOException ||
+                                      e is UnauthorizedAccessException ||
+                                      e is ArgumentException ||
+                                      e is NotSupportedException)
+            {
+                exception = e;
+                return false;
+            }
+
+            exception = null;
+            return true;
+        }
+
+        /// <summary>
+        /// Recursively deletes a directory and all contained contents.
+        /// </summary>
+        public bool TryDeleteDirectory(string path, out Exception exception)
+        {
+            try
+            {
+                this.DeleteDirectory(path);
+            }
+            catch (DirectoryNotFoundException)
+            {
+                // The directory does not exist - follow the
+                // convention of this class and report success
+            }
+            catch (Exception e) when (e is IOException ||
+                                      e is UnauthorizedAccessException ||
+                                      e is ArgumentException)
+            {
+                exception = e;
+                return false;
+            }
+
+            exception = null;
+            return true;
+        }
+
         /// <summary>
         /// Attempts to delete a file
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -260,6 +256,50 @@ public bool TryCopyToTempFileAndRename(string sourcePath, string destinationPath
...
+        /// </summary>
+        public bool TryDeleteDirectory(string path, out Exception exception)
+        {
+            try
+            {
+                this.DeleteDirectory(path);
+            }
+            catch (DirectoryNotFoundException)
+            {
+                // Swallow the exception
```
kewillford: I would explain why the code does this not what it is doing.

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -260,6 +256,50 @@ public bool TryCopyToTempFileAndRename(string sourcePath, string destinationPath
...
+            }
+            catch (Exception e) when (e is IOException ||
```
mjcheetham: Following the call chain here, to `this.DeleteDirectory -> RecursiveDelete > DirectoryInfo.Delete` it appears the `DirectoryNotFoundException` could also be [thrown](https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.delete?view=netframework-4.7.2).

Does this method `catch` all possible problems?

I also notice the `TryCreateDirectory` method above does not catch `PathTooLongException`.. is that something we consider valid to `catch`?

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -12,10 +12,6 @@ public class PhysicalFileSystem
     {
         public const int DefaultStreamBufferSize = 8192;
 
-        // https://msdn.microsoft.com/en-us/library/system.io.filesystemwatcher.internalbuffersize(v=vs.110).aspx:
-        // Max FileSystemWatcher.InternalBufferSize is 64 KB
-        private const int WatcherBufferSize = 64 * 1024;
```
mjcheetham: I assume this was removed as it was not being used?

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
mjcheetham: > `DirectoryNotFound` and `PathTooLong` exceptions are subclass of the `System.IO` exception, which is caught.

Ahh, good point I had missed that!

> I asked myself the same question, though. I could be explicit about the exceptions that are caught and just add them, to remove future confusion...

It probably isn't worth it?

Issue #705, opened 2019-01-17
Mac: Kext didn't autoload when in /Library/Extensions


nickgra:
From the customer's first impressions:
> Needed to manually call `sudo kextload /Library/Extensions/PrjFSKext.kext`

Maybe a reboot needs to occur for the kext to autoload? If so, maybe we need to load it as part of our postinstall script so users don't have this bad experience on their first run?

pmj:
> Maybe a reboot needs to occur for the kext to autoload?

Generally not. But the kext authorisation UI really messes with kext loading in my experience, with hard-to-reproduce one-off glitches like that.

There are a few things that can go wrong:
- The user hasn't approved the kext. To check for this, run `kextutil` or `kextload` on it, and if they report `System policy prevents loading the kernel extension.` that's the problem. The programmatic equivalent is a return value of `kOSKextReturnSystemPolicy` from the KextManager API. (Solution: direct the user to approve the kext.)
- The kext isn't added to the kext cache and IOKit matching. This can happen if there's a race between kextcache rescanning `/Library/Extensions` and the kext not being fully installed yet (and the kext thus being deemed not loadable by the kextcache scan because e.g. some files are missing or the signature doesn't validate), *or* if the mtime on `/Library/Extensions` is not updated when installing the kext. (E.g. when updating the kext using `sudo cp -r`) Neither should happen when installing using a .pkg, but I could swear kext authorisation mucks this one up occasionally. (Authorising too fast? Too slow? Who knows. I have no proof at the moment.) To check for this, again attempt to load the kext manually. If it loads, perform a `sudo touch /Library/Extensions` to force a kextcache rescan.

If explicit kext loading fails with some other error… there is some other problem. :-)

The only reliable recourse I'm aware of is to check for these conditions in the software that relies on the kext finds the kext hasn't been loaded. Loading the kext will need to be done with root privileges, so for example the log daemon could do it. However, that doesn't help for telling the user to approve the kext, plus the daemon could detect problems where there are none if it happens to be launched before the kext. So perhaps PrjFSLib could test for the error and send an IPC request to the log daemon to find out how to resolve it?

wilbaker:
@alameenshah, @nickgra, @kewillford, with Kevin's fix in #712 can we close this issue?

alameenshah:
> @alameenshah, @nickgra, @kewillford, with Kevin's fix in #712 can we close this issue?

[...]

Issue #584, opened 2018-12-05
Mac: Update xattr names, DNS bundle identifier, and magic number in header


sanoursa:
We need to drop the "gvfs" part of the xattr names anyway, might as well do it before we get any real users. 

And I'm not sure I have a strong opinion on what the magic number in the header ought to be. Is the current value good enough?

pmj:
There's also the reverse DNS bundle identifier for the kext and the corresponding class names which still use `io.gvfs` and `io_gvfs_`, respectively. (Doing a find-all in Xcode on those strings should bring up all instances.)

pmj:
I've noticed that some identifiers use `com.vfsforgit.` while others use `org.vfsforgit.` Which one are we standardising on?

(I'm having a go at implementing this fix as I've got a bit of time while waiting on PR feedback for all my other open tasks.)

pmj:
Thanks to PR #751, the only remains of the `io.gvfs.` prefix are in legacy-supporting/upgrade code which is on track to be removed somewhere down the line (#891).

PR #712, opened 2019-01-23
Try to load the kext if not loaded

kewillford:
When the kext is not loaded we can try and load it and check the return code to know if it was loaded successfully or if the user has not approved the certificate in Security & Privacy or it could not load for some other reason.

### DIFF
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -93,7 +93,7 @@ namespace GVFS.Platform.Mac
             ProcessResult loadKext = ProcessHelper.Run("sudo", "/sbin/kextload -b " + DriverName);
             if (loadKext.ExitCode == LoadKext_ExitCode_Success)
             {
-                tracer.RelatedWarning(metadata, $"{DriverName} was successfully loaded.", Keywords.Telemetry);
+                tracer.RelatedWarning(metadata, $"{DriverName} was successfully loaded but should have been autoloaded.", Keywords.Telemetry);
                 errorMessage = null;
                 return true;
             }
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -81,8 +81,9 @@ namespace GVFS.Platform.Mac
             return true;
         }
 
-        public bool TryLoad(ITracer tracer, string enlistmentRoot, out string errorMessage)
+        public bool TryLoad(ITracer tracer, string enlistmentRoot, TextWriter output, out string errorMessage)
         {
+            output?.WriteLine("Driver not loaded.  Attempting to load. You may be prompted for sudo password...");
             EventMetadata metadata = new EventMetadata();
             ProcessResult loadKext = ProcessHelper.Run("sudo", "/sbin/kextload -b " + DriverName);
             if (loadKext.ExitCode == LoadKext_ExitCode_Success)
@@ -104,7 +105,7 @@ System Preferences -> Security & Privacy";
                 metadata.Add("Errors", loadKext.Errors);
                 tracer.RelatedError(metadata, "Failed to load kext");
 
-                errorMessage = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
+                errorMessage = DriverName + " is not loaded. Make sure the kext is loaded and try again.";
             }
 
             return false;
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -84,7 +84,7 @@ namespace GVFS.Platform.Mac
         public bool TryLoad(ITracer tracer, string enlistmentRoot, out string errorMessage)
         {
             EventMetadata metadata = new EventMetadata();
-            ProcessResult loadKext = ProcessHelper.Run("sudo", "kextutil -b " + DriverName);
+            ProcessResult loadKext = ProcessHelper.Run("sudo", "/sbin/kextload -b " + DriverName);
             if (loadKext.ExitCode == LoadKext_ExitCode_Success)
             {
                 tracer.RelatedWarning(metadata, $"{DriverName} was successfully loaded.", Keywords.Telemetry);
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -12,6 +12,8 @@ namespace GVFS.Platform.Mac
     public class ProjFSKext : IKernelDriver
     {
         private const string DriverName = "io.gvfs.PrjFSKext";
+        private const int LoadKext_ExitCode_Success = 0;
+        private const int LoadKext_ExitCode_NotApproved = 27;
 
         public bool EnumerationExpandsDirectories { get; } = true;
 
@@ -62,9 +64,7 @@ namespace GVFS.Platform.Mac
         public bool IsReady(JsonTracer tracer, string enlistmentRoot, out string error)
         {
             error = null;
-            return
-                this.IsKextLoaded() ||
-                this.TryLoadKext(tracer, enlistmentRoot, out error);
+            return this.IsKextLoaded();
         }
 
         public bool TryPrepareFolderForCallbacks(string folderPath, out string error, out Exception exception)
@@ -81,63 +81,30 @@ namespace GVFS.Platform.Mac
             return true;
         }
 
-        private ProcessResult Bash(string cmd)
+        public bool TryLoad(ITracer tracer, string enlistmentRoot, out string errorMessage)
         {
-            string escapedArgs = cmd.Replace("\"", "\\\"");
-
-            ProcessStartInfo startInfo = new ProcessStartInfo
+            EventMetadata metadata = new EventMetadata();
+            ProcessResult loadKext = ProcessHelper.Run("sudo", "kextutil -b " + DriverName);
+            if (loadKext.ExitCode == LoadKext_ExitCode_Success)
             {
-                FileName = "/bin/bash",
-                Arguments = $"-c \"{escapedArgs}\"",
-                RedirectStandardOutput = true,
-                RedirectStandardError = true,
-                UseShellExecute = false,
-                CreateNoWindow = true,
-            };
-
-            using (Process process = new Process())
-            {
-                process.StartInfo = startInfo;
-                string errors = string.Empty;
-                process.ErrorDataReceived += (sender, args) =>
-                {
-                    if (args.Data != null)
-                    {
-                        errors += args.Data + "\r\n";
-                    }
-                };
-
-                process.Start();
-                process.BeginErrorReadLine();
-                string output = process.StandardOutput.ReadToEnd();
-                process.WaitForExit();
-                return new ProcessResult(output, errors, process.ExitCode);
-            }
-        }
-
-        private bool TryLoadKext(ITracer tracer, string enlistmentRoot, out string error)
-        {
-            Console.WriteLine("Kernel extension not loaded.  Attempting to load...");
-            error = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
-            ProcessResult loadKext = this.Bash("sudo kextutil /Library/Extensions/PrjFSKext.kext");
-            if (loadKext.ExitCode == 0)
-            {
-                error = null;
+                tracer.RelatedWarning(metadata, $"{DriverName} was successfully loaded.", Keywords.Telemetry);
+                errorMessage = null;
                 return true;
             }
-            else if (loadKext.ExitCode == 27)
+            else if (loadKext.ExitCode == LoadKext_ExitCode_NotApproved)
             {
-                tracer.RelatedWarning("Kext unable to load. Possibly has not been approved by the user");
-                error = DriverName + @" was unable to load.  Please check and make sure you have allowed the extension in
+                tracer.RelatedError("Kext unable to load. Not approved by the user");
+                errorMessage = DriverName + @" was unable to load.  Please check and make sure you have allowed the extension in
 System Preferences -> Security & Privacy";
             }
             else
             {
-                EventMetadata metadata = new EventMetadata();
                 metadata.Add("ExitCode", loadKext.ExitCode);
                 metadata.Add("Output", loadKext.Output);
                 metadata.Add("Errors", loadKext.Errors);
-                tracer.RelatedWarning(metadata, "Failed to load kext");
+                tracer.RelatedError(metadata, "Failed to load kext");
+
+                errorMessage = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
             }
 
             return false;
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -13,6 +13,9 @@ namespace GVFS.Platform.Mac
     {
         private const string DriverName = "io.gvfs.PrjFSKext";
         private const int LoadKext_ExitCode_Success = 0;
+
+        // This exit code was found in the following article
+        // https://developer.apple.com/library/archive/technotes/tn2459/_index.html
         private const int LoadKext_ExitCode_NotApproved = 27;
 
         public bool EnumerationExpandsDirectories { get; } = true;
--- a/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
+++ b/GVFS/GVFS.Platform.Mac/ProjFSKext.cs
@@ -3,6 +3,7 @@ using GVFS.Common.FileSystem;
 using GVFS.Common.Tracing;
 using PrjFSLib.Mac;
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 
@@ -60,15 +61,10 @@ namespace GVFS.Platform.Mac
 
         public bool IsReady(JsonTracer tracer, string enlistmentRoot, out string error)
         {
-            ProcessResult loadedKexts = ProcessHelper.Run("kextstat", args: "-b " + DriverName, redirectOutput: true);
-            if (loadedKexts.Output.Contains(DriverName))
-            {
-                error = null;
-                return true;
-            }
-
-            error = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
-            return false;
+            error = null;
+            return
+                this.IsKextLoaded() ||
+                this.TryLoadKext(tracer, enlistmentRoot, out error);
         }
 
         public bool TryPrepareFolderForCallbacks(string folderPath, out string error, out Exception exception)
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,73 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
+        {
+            Console.WriteLine("Kernel extension not loaded.  Attempting to load...");
+            error = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
+            ProcessResult loadKext = this.Bash("sudo kextutil /Library/Extensions/PrjFSKext.kext");
+            if (loadKext.ExitCode == 0)
+            {
+                error = null;
+                return true;
+            }
+            else if (loadKext.ExitCode == 27)
```
kewillford: Add const

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,73 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
+                process.WaitForExit();
+                return new ProcessResult(output, errors, process.ExitCode);
+            }
+        }
+
+        private bool TryLoadKext(ITracer tracer, string enlistmentRoot, out string error)
+        {
+            Console.WriteLine("Kernel extension not loaded.  Attempting to load...");
+            error = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
+            ProcessResult loadKext = this.Bash("sudo kextutil /Library/Extensions/PrjFSKext.kext");
```
kewillford: Is there a way to get this path other than hard coding?

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,73 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
+                process.BeginErrorReadLine();
+                string output = process.StandardOutput.ReadToEnd();
+                process.WaitForExit();
+                return new ProcessResult(output, errors, process.ExitCode);
+            }
+        }
+
+        private bool TryLoadKext(ITracer tracer, string enlistmentRoot, out string error)
+        {
+            Console.WriteLine("Kernel extension not loaded.  Attempting to load...");
```
kewillford: Added this message because the `sudo` will prompt for a password and I wanted some kind of message for the user to know why they were being prompted.

jeschu1: Is it worth adding "You may getting prompted as this calls sudo" or similar?

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,73 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
+            error = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
```
jeschu1: Do we need this line here?

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,73 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
+                error = DriverName + @" was unable to load.  Please check and make sure you have allowed the extension in
+System Preferences -> Security & Privacy";
+            }
+            else
+            {
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add("ExitCode", loadKext.ExitCode);
+                metadata.Add("Output", loadKext.Output);
+                metadata.Add("Errors", loadKext.Errors);
+                tracer.RelatedWarning(metadata, "Failed to load kext");
```
jeschu1: Set error

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
jeschu1: Not sure if we have a precident but I like seeing both the 0 and 27 hardcoded to consts.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
jrbriggs: I'm not sure we can rely on `Console.WriteLine` to do the right thing at this level. Should we send a failure back up to the MountVerb and allow the verb to interact with the user? My concern isn't with this PR -- it's when, in the future, we bring automount online.  We'd still `sudo` but the user would have no clue why they got the prompt. As part of enabling automount on mac work, we should teach GVFS.Service.UI to pop mac notifications, and then automount can go that route and mounts initiated from the command line will continue to interact on the terminal. Thoughts?

kewillford: Yeah this was kind of the quick and dirty way to get it up and working.  I will look into launching kextutil without bash and getting the sudo prompt so that it would work when we have automount.  Thanks!

alameenshah:
Would this still be needed if kext gets auto-loaded after install? Refer - [this issue](https://github.com/Microsoft/VFSForGit/issues/705)

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
pmj: Instead of a path, you can use the bundle identifier using the `-b` flag. (`sudo kextutil -b io.gvfs.PrjFSKext`, at least until we change it for #584.)

alameenshah: 1. can there be an alternative which does not need `sudo`? `GVFS.Service` would be running elevated, right? Can the service do the kext loading without a `sudo`?
2. can we use the full path to the tool here - like `/sbin/kextload`. Partial paths depend on runtime environment to construct the full path, and may not be the right thing to do while using `sudo`.
3. shouldn't it be `kextload`? rather than `kextutil`?

kewillford: @pmj 
How does it determine where the actual `.kext` file is if we switch to use the bundle identifier?

@alameenshah 
1. We don't have the service on Mac yet.  At least don't think we do.  Correct me if I am wrong.
2. Will you explain this more?  Couldn't someone could just move it?  If we hard code the full path can we ensure it will always be there?  When might it not be the right thing to do using the environment?
3. It was `kextutil` in the `LoadPrjFSKext.sh` so that is why I kept it using `kextutil` even though it is using `kextunload` in that script as well.  I will try `kextload` and make sure it uses the same exit codes.

pmj:
I'm not 100% sold on the approach, as it masks installation problems to some extent. Loading the kext on demand like this will just paper over such problems, it won't fix them. If there's some reason the kext isn't being auto-loaded on boot, then our offline virtualisation root code in the kext won't be running. (Which should normally protect against modifications of roots which aren't mounted.)

Note that instead of shelling out and processing the output of `kextutil`/`kextload`, there's also the KextManager API, e.g. [`KextManagerLoadKextWithIdentifier`](https://developer.apple.com/documentation/iokit/1562907-kextmanagerloadkextwithidentifie?language=objc). That will only be useful from a process already running as root though of course, and I don't know if there are any .NET wrappers for it.

See also [my comment in issue #705](https://github.com/Microsoft/VFSForGit/issues/705#issuecomment-455321838).

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
alameenshah: >>Couldn't someone could just move it? If we hard code the full path can we ensure it will always be there?

The path `/sbin/kextload` is protected by [SIP](https://support.apple.com/en-us/HT204899), so it will be there.

>> When might it not be the right thing to do using the environment?

We know the path to the tool that we want to use - `/sbin/kextload`. And by specifying the full path we don't depend on the shell to expand it to what we intended.

kewillford:
With SIP enabled and a signed kext the user still has to go into Security & Privacy and approve the use of the kext so it will never be loaded after install unless there is some kind of retry and the installer sits there until it knows the user has approved the kext.  

This is to help the user know that they haven't approved the kext loading yet and they need to go to the Security & Privacy to approve it.  I realize in the normal flow the user will approve during the install but I'm not sure we retry to load after initial failure.  Without this change and assuming we don't retry the load at install, I see the user experience as.

1. Install
2. During install prompt to go to approve kext
3. Approve the kext
4. Install finishes
5. Run gvfs clone
6. Error that kext isn't loaded
7. Contact us to know how to fix the error.

This will remove 6 and 7 assuming the install finished successfully.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
pmj: > 3\. shouldn't it be `kextload`? rather than `kextutil`?

`kextutil` can do anything `kextload` can; if you're showing its output to users, `kextutil` might be better, as it's intended for diagnosing problems.

pmj: > @pmj
> How does it determine where the actual `.kext` file is if we switch to use the bundle identifier?

It considers all kexts in `/System/Library/Extensions` and al that have passed validation in `/Library/Extensions`.

pmj:
> This will remove 6 and 7 assuming the install finished successfully.

I fully agree that we should have diagnostics for the kext-not-properly-installed/kext-not-loaded case. It's unfortunately a common enough occurrence. I'm just wondering if we should solve that in one central place, including code that attempts recovery of that situation, rather than various partial hacks in multiple places.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
pmj: > 1. can there be an alternative which does not need `sudo`? `GVFS.Service` would be running elevated, right? Can the service do the kext loading without a `sudo`?

Incidentally, we do have PrjFSKextLogDaemon running all the time; this could offer an IPC service for attempting to recover from common kext-not-loaded/kext-not-loadable situations.

I should probably mention that it's possible to mark a kext as loadable by non-root users using the [`OSBundleAllowUserLoad`Info.plist key](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KEXTConcept/Articles/infoplist_keys.html). However:

> `OSBundleAllowUserLoad`<br>
> The OSBundleAllowUserLoad property allows non-root users to load your kext. Using this property is not recommended.
> 
> I/O Kit drivers should never include this property, because they are loaded by the kernel when they are needed.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,40 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
             return true;
         }
+
+        public bool TryLoad(ITracer tracer, string enlistmentRoot, out string errorMessage)
+        {
+            EventMetadata metadata = new EventMetadata();
+            ProcessResult loadKext = ProcessHelper.Run("sudo", "/sbin/kextload -b " + DriverName);
+            if (loadKext.ExitCode == LoadKext_ExitCode_Success)
+            {
+                tracer.RelatedWarning(metadata, $"{DriverName} was successfully loaded.", Keywords.Telemetry);
```
wilbaker: What was the reason for making this a Warning?

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -381,6 +381,11 @@ public bool IsReady(JsonTracer tracer, string enlistmentRoot, out string error)
                 TryAttach(enlistmentRoot, out error);
         }
 
+        public bool TryLoad(ITracer tracer, string enlistmentRoot, out string errorMessage)
+        {
+            return TryAttach(enlistmentRoot, out errorMessage);
```
wilbaker: Something feels a bit off to me here when it comes to how we're using calling the platform's `KernelDriver` API from the mount verb.

- On Windows `KernelDriver.IsReady` will call `TryAttach` (and the Window's platform `TryLoad` method will never be called).

- On Mac, if `KernelDriver.IsReady` returns false `KernelDriver.TryLoad` will be called and attempt to load the kext.

I'm wondering if we should remove `TryLoad` from the `KernelDriver` interface and have the Mac's `KernelDriver.IsReady` attempt to load the kext if it's not already loaded (just as the Window's `IsReady` will attempt to attach the filter if it's not already attached).  

(Or alternatively, update Window's `IsReady` to not attach the filter and then call `TryLoad` in a place that it will be hit on both platforms)

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -11,6 +12,8 @@ namespace GVFS.Platform.Mac
     public class ProjFSKext : IKernelDriver
     {
         private const string DriverName = "io.gvfs.PrjFSKext";
+        private const int LoadKext_ExitCode_Success = 0;
+        private const int LoadKext_ExitCode_NotApproved = 27;
```
wilbaker: Where did you find these values?

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
```diff
@@ -73,5 +69,40 @@ public bool TryPrepareFolderForCallbacks(string folderPath, out string error, ou
...
+System Preferences -> Security & Privacy";
+            }
+            else
+            {
+                metadata.Add("ExitCode", loadKext.ExitCode);
+                metadata.Add("Output", loadKext.Output);
+                metadata.Add("Errors", loadKext.Errors);
+                tracer.RelatedError(metadata, "Failed to load kext");
+
+                errorMessage = DriverName + " is not loaded. Make sure the driver is loaded and try again.";
```
wilbaker: Error message suggestion: `sure the driver is loaded` -> `sure the kext is loaded`

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
kewillford: https://developer.apple.com/library/archive/technotes/tn2459/_index.html

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
kewillford: Yeah this was a little weird when I was coding it up because I originally had everything in the `IsReady`.  Changed it to this because Mac needs to display a message to the user before they get prompted for their password so they have some idea why and doing that in the `IsReady` was not the right place because I was using `Console.WriteLine` but that might not be available if the service or something else besides the CLI.  Windows does not need to do that.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
kewillford: I wanted this to be a little stronger than Informational because this would mean that the kext for some reason wasn't auto loaded by the OS.  But if this is too strong I can make it Informational and add some text to the message so that we know this usually shouldn't happen.  Maybe I should add the text in any case?

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
kewillford: I will look at refactoring the code so it is more the same on both platforms.

GVFS/GVFS.Platform.Mac/ProjFSKext.cs
wilbaker: >But if this is too strong I can make it Informational

Thanks for the explanation, I think it's OK to leave it as Warning.  

> Maybe I should add the text in any case?

Adding a bit more text for better context sounds good to me.

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
kewillford: Hmmm.  In looking at this, we have the following message currently output before running the `TryAttach`
```
this.Output.WriteLine("Driver not loaded.  Attempting to load. You may be prompted for sudo password...");
```
Windows will not need that message.  So in thinking about solutions, we could:
1. Add another property to the interface for whether or not the message is shown.  I hate to keep adding to that interface though.
2. Pass the `TextWriter` to the `TryLoad` so that it can output the message.  Windows would do nothing and Mac could output this message.  We would need a `null` check on the `TextWriter` so that if something like the service which might not have an output could still call the method.

What are your thoughts?

wilbaker: > I hate to keep adding to that interface though.

I agree.

I think Option 2 sounds like the best option.

GVFS/GVFS/CommandLine/MountVerb.cs
```diff
@@ -117,42 +117,45 @@ protected override void Execute(GVFSEnlistment enlistment)
 
                 if (!GVFSPlatform.Instance.KernelDriver.IsReady(tracer, enlistment.EnlistmentRoot, out errorMessage))
                 {
-                    if (GVFSPlatform.Instance.UnderConstruction.SupportsGVFSService)
+                    if (!GVFSPlatform.Instance.KernelDriver.TryLoad(tracer, enlistment.EnlistmentRoot, this.Output, out errorMessage))
```
wilbaker: We should always be calling `TryLoad` (at least on Windows), but with the latest changes we'll only call `TryLoad` if `IsReady` return false.

May we should do something like rename `IsReady` to `IsInstalled` and always call `TryLoad`? (or just always call `TryLoad`)

kewillford: Hmmm.  In looking at the code before:
```
IsServiceRunning(tracer) &&
IsNativeLibInstalled(tracer, new PhysicalFileSystem()) &&
TryAttach(enlistmentRoot, out error);
```
It would only run the `TryAttach` on windows if the other 2 would be `true` right?  So on windows it would only run the `TryLoad` if `IsReady` is true?  

Renaming to `IsInstalled` wouldn't make sense on Mac because it is checking if it is loaded and running.  Also on Mac if we call it everytime the user will get the sudo prompt every mount even if the kext is loaded which is not what we want.

I think switching this back to just having the `IsReady` and pass the `TextWriter` to that method and each platform can do the work needed is the best option here.

wilbaker: Your analysis looks correct to me, and switching back to IsReady sounds good to me as well.

GVFS/GVFS/CommandLine/MountVerb.cs
```diff
@@ -127,32 +127,32 @@ protected override void Execute(GVFSEnlistment enlistment)
...
-                            EventLevel.Informational,
+                            EventLevel.Error,
                             $"{nameof(MountVerb)}_{nameof(this.Execute)}",
                             new EventMetadata
                             {
-                                { "KernelDriver.IsReady_Error", errorMessage },
-                            });
-
-                        this.ReportErrorAndExit(tracer, ReturnCode.FilterError, errorMessage);
+                                { "KernelDriver.TryLoad_Error", errorMessage },
```
kewillford: Change this back.

PR #714, opened 2019-01-25
Create combined VFSForGit & Git installer

alameenshah:
GVFS.Installer.Mac project will now build an additional Meta package installer. This installer will contain both VFSForGit Mac installer and Git installer in a single package. The existing VFSForGit installer will continue to be produced.

### DIFF
--- a/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
+++ b/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
@@ -103,6 +103,47 @@ function CreateInstaller()
     eval $pkgBuildCommand || exit 1
 }
 
+function CreateMetaInstaller()
+{
+    GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
+    GITDMGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.dmg)" || exit 1
+    GITDMGNAME="${GITDMGPATH##*/}"
+    GITINSTALLERNAME="${GITDMGNAME%.dmg}"
+    GITVERSIONSTRING=`echo $GITINSTALLERNAME | cut -d"-" -f2`
+    
+    if [[ -z "$GITVERSION" || -z "$GITVERSIONSTRING" ]]; then
+        echo "Error creating metapackage: could not determine Git package version."
+        exit 1
+    fi
+    
+    if [ ! -f "$GITDMGPATH" ]; then
+        echo "Error creating metapackage: could not find Git disk image."
+        exit 1
+    fi
+    
+    mountCmd="/usr/bin/hdiutil attach \"$GITDMGPATH\""
+    echo "$mountCmd"
+    eval $mountCmd || exit 1
+    
+    MOUNTEDVOLUME=`/usr/bin/find /Volumes -type d -name "Git $GITVERSIONSTRING*"`
+    GITINSTALLERPATH=`/usr/bin/find "$MOUNTEDVOLUME" -type f -name "git-$GITVERSIONSTRING*.pkg"`
+    
+    if [ ! -f "$GITINSTALLERPATH" ]; then
+        echo "Error creating metapackage: could not find Git installer package."
+        exit 1
+    fi
+    
+    METAPACKAGENAME="$INSTALLERPACKAGENAME-Git.$GITVERSION.pkg"
+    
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -103,13 +103,54 @@ function CreateInstaller()
...
+        exit 1
+    fi
+    
+    mountCmd="/usr/bin/hdiutil attach \"$GITDMGPATH\""
+    echo "$mountCmd"
+    eval $mountCmd || exit 1
+    
+    MOUNTEDVOLUME=`/usr/bin/find /Volumes -type d -name "Git $GITVERSIONSTRING*"`
+    GITINSTALLERPATH=`/usr/bin/find "$MOUNTEDVOLUME" -type f -name "git-$GITVERSIONSTRING*.pkg"`
+    METAPACKAGENAME="$INSTALLERPACKAGENAME-Git.$GITVERSION.pkg"
```
nickgra: Could this be after the test for `-f "$GITINSTALLERPATH" `

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -103,13 +103,54 @@ function CreateInstaller()
     eval $pkgBuildCommand || exit 1
 }
 
+function CreateMetaInstaller()
+{
+    GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
+    GITDMGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.dmg)" || exit 1
```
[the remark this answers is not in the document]
wilbaker: What is the standard practice on Mac, and what are the pros/cons of using a dmg?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -103,13 +103,54 @@ function CreateInstaller()
...
+    MOUNTEDVOLUME=`/usr/bin/find /Volumes -type d -name "Git $GITVERSIONSTRING*"`
+    GITINSTALLERPATH=`/usr/bin/find "$MOUNTEDVOLUME" -type f -name "git-$GITVERSIONSTRING*.pkg"`
+    METAPACKAGENAME="$INSTALLERPACKAGENAME-Git.$GITVERSION.pkg"
+    
+    if [ ! -f "$GITINSTALLERPATH" ]; then
+        echo "Error creating metapackage: could not find Git installer package."
+        exit 1
+    fi
+    
+    buildMetapkgCmd="/usr/bin/productbuild --package \"$GITINSTALLERPATH\" --package \"${BUILDOUTPUTDIR}\"$INSTALLERPACKAGENAME.pkg \"${BUILDOUTPUTDIR}\"$METAPACKAGENAME"
```
wilbaker: What is the user experience like with this .pkg?  Does it walk them through the Git and VFS4G installers, or does it just package both together (and the user would run them individually)?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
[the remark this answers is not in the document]
alameenshah: > What is the standard practice on Mac, and what are the pros/cons of using a dmg?

We had a discussion about dmg vs pkg in [this PR](https://github.com/Microsoft/VFSForGit/pull/438). In the past `.pkg` used to be a [bundle](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1). `DMG`s are like Virtual disks and are useful to package and distribute such `pkg` bundles. It is also useful to distribute ReadMe, un-installers or other tools along with the installer. `DMG` could also be signed. The new `.pkg`s are flat files which are already compressed and signed. We didn't find much use in additionally wrapping a flat `.pkg` inside `.dmg`. 

`DMG` can be useful for `gvfs upgrade` though - we can put `VFSForGit` `pkg` in a read-only `DMG`, `gvfs upgrader` tool can mount the read-only virtual disk (`DMG`), verify the pkg and run it without having to worry about anyone tampering with the `pkg`.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
alameenshah: The user wouldn't know there are two separate installers. It will appear like a single installer that installs both VFSForGit and Git.

PR #718, opened 2019-01-28
MirrorProvider: handle mount error when enlistment missing

chrisd8088:
Currently, MirrorProvider will try to proceed to try to mount the enlistment even if the enlistment configuration file was not found, leading to errors, so instead we exit early on this condition.


jrbriggs:
Thanks for this, @chrisd8088 !

PR #719, opened 2019-01-28
Linux: add filesystem mount handle to placeholder write methods

chrisd8088:
Pass the Linux filesystem mount handle to the lower-level library when writing placeholder files and directories.

This allows the low-level Linux library to be able to locate its mount point and then execute internal file or directory write requests such that these internal writes don't trigger recursive enumeration or hydration requests back up to the C# calling code.


chrisd8088:
Thanks for merging #718, @jrbriggs!

Here's the last of my three changes for Linux, at least for now (there will be lots more in the future, obviously!  :-)  Basically we're just adding `mountHandle` to the arguments passed to the Linux support library when making enumeration or hydration requests.  There will be similar changes to other functions down the road, but this is all we need for the moment.

Thanks again for the merges!

jrbriggs:
This looks great, thanks @chrisd8088!

PR #724, opened 2019-01-30
Fix removal from modified paths when path should not be removed

### LINKED ISSUE #725: VFS4G will sometimes incorrectly remove files from ModifiedPaths.dat
Split from issue #721 

If a tracked file is deleted and then created again, VFS4G will incorrectly track it as a newly created file.

kewillford:
Fixes #725

For this fix I check to see if the path is in the modified paths before adding it to the newly created list.  This makes sure that if a file that was tracked was deleted it will not be added as a newly created list because it will be in the modified paths.

There will be some edge cases where a temp or newly created file will not get added to the list and should have been, for example modifying a file so it gets in the modified paths and committing then checking out a commit where the file does not exist.  In most cases I don't think these will matter too much with regards to keeping the modified paths clean.  

There is also the issue with an older enlistment that hasn't had their modified paths cleaned up and they have entries for temp files they are creating and deleting.  These will not get removed now.

Other thoughts were to keep track of deleted files and folders and check against that list but that is what the modified paths already is.  Also could check the index but that seemed like an expensive operation in this case.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
@@ -26,6 +26,19 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         }
 
         [TestCase]
+        public void DeleteThenCreateThenDeleteFile()
+        {
+            string srcPath = @"Readme.md";
+
+            this.DeleteFile(srcPath);
+            this.ValidateGitCommand("status");
+            this.CreateFile("Testing", srcPath);
+            this.ValidateGitCommand("status");
+            this.DeleteFile(srcPath);
+            this.ValidateGitCommand("status");
+        }
+
+        [TestCase]
         [Category(Categories.MacTODO.M4)]
         public void ModifyingAndDeletingRepositoryExcludeFileInvalidatesCache()
         {
--- a/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
+++ b/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
@@ -389,7 +389,7 @@ namespace GVFS.Virtualization
 
         public void OnFileCreated(string relativePath)
         {
-            this.newlyCreatedFileAndFolderPaths.Add(relativePath);
+            this.AddToNewlyCreatedList(relativePath, isFolder: false);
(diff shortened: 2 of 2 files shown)


GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -657,11 +657,10 @@ private FileSystemTaskResult ExecuteBackgroundOperation(FileSystemTask gitUpdate
                     // additional work is needed for any files\folders inside the folder being moved)
                     if (result == FileSystemTaskResult.Success && !string.IsNullOrEmpty(gitUpdate.VirtualPath))
                     {
+                        this.AddToNewlyCreatedList(gitUpdate.VirtualPath, isFolder: true);
                         result = this.TryAddModifiedPath(gitUpdate.VirtualPath, isFolder: true);
                         if (result == FileSystemTaskResult.Success)
                         {
-                            this.newlyCreatedFileAndFolderPaths.Add(gitUpdate.VirtualPath);
```
kewillford: Since the code now checks the modified paths to be able to add to the newly created list, this code needed to move to before the path is added to modified paths otherwise it would never add to the newly created list.

wilbaker:
@kewillford I updated the PR description to reference the new bug (#725)

GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
```diff
@@ -25,6 +25,19 @@ public void MoveFileIntoDotGitDirectory()
             this.ValidateGitCommand("status");
         }
 
+        [TestCase]
+        public void DeleteThenCreateThenDeleteFile()
+        {
+            string srcPath = @"Readme.md";
```
wilbaker: Should we use a different path here to avoid impacting the `MoveFileIntoDotGitDirectory` test?

kewillford: These tests are set to an enlistment per tests so it should never impact other tests right?

jeschu1: base(enlistmentPerTest: true

I would think that's fine actually

wilbaker: @kewillford  Ahh you're right, sorry about that!

PR #726, opened 2019-01-30
Update ProjFSLib to log Process Name when in debug mode

nickgra:
While debugging an issue on a customer's machine, we found that it would be helpful to have the name of the process who triggered the notification we're responding to.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -253,7 +253,10 @@ PrjFS_Result PrjFS_ConvertDirectoryToVirtualizationRoot(
     _In_    const char*                             virtualizationRootFullPath)
 {
 #ifdef DEBUG
-    cout << "PrjFS_ConvertDirectoryToVirtualizationRoot(" << virtualizationRootFullPath << ")" << endl;
+    cout
+        << "PrjFS_ConvertDirectoryToVirtualizationRoot("
+        << virtualizationRootFullPath
+        << ")" << endl;
 #endif
     
     if (nullptr == virtualizationRootFullPath)
@@ -286,7 +289,10 @@ PrjFS_Result PrjFS_WritePlaceholderDirectory(
     _In_    const char*                             relativePath)
 {
 #ifdef DEBUG
-    cout << "PrjFS_WritePlaceholderDirectory(" << relativePath << ")" << endl;
+    cout
+        << "PrjFS_WritePlaceholderDirectory("
+        << relativePath
+        << ")" << endl;
 #endif
     
     if (nullptr == relativePath)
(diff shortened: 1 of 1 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -708,7 +708,10 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
 static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* relativePath)
 {
 #ifdef DEBUG
-    cout << "PrjFSLib.HandleEnumerateDirectoryRequest: " << relativePath << endl;
+    cout << "PrjFSLib.HandleEnumerateDirectoryRequest: "
+    << relativePath
```
pmj: Style nit: perhaps we could vertically align the `<<` operators here. This is already the case on some of the other log expressions you edited, so it's probably best to follow that style.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -286,7 +289,10 @@ PrjFS_Result PrjFS_WritePlaceholderDirectory(
     _In_    const char*                             relativePath)
 {
 #ifdef DEBUG
-    cout << "PrjFS_WritePlaceholderDirectory(" << relativePath << ")" << endl;
+    cout
+        << "PrjFS_WritePlaceholderDirectory("
+        << relativePath
```
wilbaker: Minor: You could make the printing of `relativePath` more consistent.  Here (and in `PrjFS_ConvertDirectoryToVirtualizationRoot`) it's printed inside parenthesis that follow the function name.

However, I noticed in spots below it's printed after a `:` that follows the function name.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -809,7 +823,11 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
 static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const char* relativePath)
 {
 #ifdef DEBUG
-    cout << "PrjFSLib.HandleHydrateFileRequest: " << relativePath << endl;
+    cout
+        << "PrjFSLib.HandleHydrateFileRequest: "
+        << relativePath
+        << " Process name: " << request->procname
```
wilbaker: Also print the `pid` (here and everyone else we're printing `request->procname`)?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
nickgra: All of the Handle functions print the arguments with the `:` pattern while the methods that do the work use the `()` pattern. Given the nature of the arguments for the non Handle method, I think that pattern makes more sense there and the `:` pattern makes sense for Handle.

If you feel strongly, I can normalize these all though.

wilbaker: Ok thanks I hand't realized that pattern.  I think it's okay to leave it that way.

PR #727, opened 2019-01-30
Fix removal from modified paths when path should not be removed

kewillford:
Cherry picked from #724

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
@@ -26,6 +26,19 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         }
 
         [TestCase]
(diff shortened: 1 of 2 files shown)


wilbaker:
@kewillford there have been minimal changes to master since creating `releases/mac_ea` and so rather than cherry-picking this change to `releases/mac_ea` you could just point `releases/mac_ea` to the commit on master with this fix and force push it.

wilbaker:
>We can do that. Will need to turn off some of the branch policies for releases temporarily.

Ahh sorry, I didn't realize any policies were in place

> Long term do we expect to keep both branches? 

I expect that we'll keep the release/mac_ea branch until Mac releases are on the same cadence as Windows releases (so I would guess at least a few months).

> What is going to be our process for getting changes across branches?

We need to discuss with @jrbriggs, but I anticipate the process will be the same as the regular release branch, we just cut the mac_ea branch more frequently

Issue #717, opened 2019-01-28
Enable ability to customize VFSForGit Upgrade


jamill:
This issue tracks the different parts of the work to make VFSForGit upgrade extensible. This extensibility can be used by organizations to custom the VFSForGit upgrade process to meet their specific requirements.

## Design Doc
#547

## Initial changes to enable feature
#674 

## TODO - Required
- [ ] Verify Nuget Package signature
- [ ] Limit access to temp file directories
- [ ] Limit access to GVFS.Config
- [ ] Anonymous polling for available updates

## TODO - Nice to have
- [ ] Add concept of "minimum version allowed to upgrade from"

## Associated issues
#704

wilbaker:
Other nice to haves:

- Update unit tests to validate that VFS4G code that writes new version to a marker file is writing the expected version.  Currently it only validates that VFS4G has written something.

- Cleanup PhysicalFileSystem and IPlatformFileSystem.  Their interaction (and how they differ) is not clear, and the more we add to them the more confusing it becomes.

- Switch from calling `Exit` to letting unhandled exceptions bring down the process.  Some investigation is required to makes sure that all native\background threads with unhandled exceptions will crash the VFS4G process.

wilbaker:
- Cleanup `PhysicalFileSystem.DeleteDirectory` it currently just calls `RecusiveDelete`, so why have both?

wilbaker:
- Update upgrade functional tests to use Test.GVFS.Upgrade folder rather than GVFS.Upgrade

wilbaker:
Cleanup `gvfs log` so that file paths are all aligned in the same column.  Currently it looks like this:

```\E:\repo6>gvfs log
Most recent log files:
  clone     : E:\repo6\.gvfs\logs\gvfs_clone_20181024_094528.log
  mount     : E:\repo6\.gvfs\logs\gvfs_mount_verb_20190130_140733.log
  prefetch  : E:\repo6\.gvfs\logs\gvfs_prefetch_20181024_094548.log
  dehydrate : None
  repair    : E:\repo6\.gvfs\logs\gvfs_repair_20181105_163128.log
  service   : C:\ProgramData\GVFS\GVFS.Service\Logs\gvfs_service_20190207_143020.log
  productupgrade: C:\ProgramData\GVFS\GVFS.Upgrade\Logs\gvfs_productupgrade_process_20190207_144030.log

```

[...]

PR #728, opened 2019-01-30
[features/upgrade_extensibility] Restrict ACLs on VFS4G ProgramData and update verbs as needed

wilbaker:
Part of the work for #717

Updated GVFS.Service to set the ACLs on C:\ProgramData\GVFS as follows:

- Everyone has read access
- Administrators have read/modify/delete access

Also made the following tweaks:

- Updated GVFS.Service to log its version
- Updated `gvfs diagnose` to capture the `gvfs.config` file
- Updated `gvfs config` to require elevation when setting or deleting config values

Completed manual tests:
- Validated that `C:\ProgramData\GVFS` ACLs are set as expected when installing VFS43g fresh
- Validated that `C:\ProgramData\GVFS` ACLs are set as expected when upgrading from a previous version of VFS4G

Remaining manual tests:
- [ ] Test installation scenarios on Windows Server
- [ ] Test installation scenarios on a clean machine
- [ ] Test `gvfs upgrade` against GitHub and NuGet

### DIFF
--- a/GVFS/GVFS.Service/GvfsService.cs
+++ b/GVFS/GVFS.Service/GvfsService.cs
@@ -40,6 +40,10 @@ namespace GVFS.Service
         {
             try
             {
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add("Version", ProcessHelper.GetCurrentProcessVersion());
+                this.tracer.RelatedEvent(EventLevel.Informational, $"{nameof(GVFSService)}_{nameof(this.Run)}", metadata);
+
                 this.repoRegistry = new RepoRegistry(this.tracer, new PhysicalFileSystem(), this.serviceDataLocation);
                 this.repoRegistry.Upgrade();
                 this.productUpgradeTimer.Start();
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -348,19 +350,42 @@ private void LogExceptionAndExit(Exception e, string method)
...
+            this.serviceDataLocation = Paths.GetServiceDataRoot(this.serviceName);
+            string serviceDataRootPath = Path.GetDirectoryName(this.serviceDataLocation);
+
+            DirectorySecurity programDataSecurity = new DirectorySecurity();
+
+            // Protect the access rules from inheritance
+            programDataSecurity.SetAccessRuleProtection(isProtected: true, preserveInheritance: false);
+
+            // Everyone gets read access
+            SecurityIdentifier authenticatedUsers = new SecurityIdentifier(WellKnownSidType.WorldSid, null);
```
jamill: Is World Sid what we want here? Given the name is `authenticatedUsers`, I was expecting the `AuthenticatedUser` Sid?

wilbaker: Good point, I missed renaming the variable.  I actually switched from World to BuiltInUsers which are all users on the machine (I took a look at what other applications are doing that that seems to be pretty common).

PR #732, opened 2019-01-31
Add ShouldIgnoreVnodeType + FlaggedFileAsInRoot test

jeschu1:
ShouldIgnoreVnodeType is pretty straightforward

FlaggedFileAsInRoot shows the complexities when dealing with kernal types.  Here I overwrote the return types for TryReadVNodeFileFlags to facilitate testing.  Since TryReadVNodeFileFlags is a lookup it seems like a valid workaround.

### DIFF
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -1,10 +1,21 @@
 #import <XCTest/XCTest.h>
 #include "../PrjFSKext/KauthHandlerTestable.hpp"
+#include <sys/vnode.h>
 
 @interface KauthHandlerTests : XCTestCase
 @end
 
 @implementation KauthHandlerTests
+
+//Mock for TryReadVNodeFileFlags
+static uint32_t mockFileFlag;
+static bool mockReturnBool;
+bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t context, uint32_t* flags);
+bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t context, uint32_t* flags)
+{
+    *flags = mockFileFlag;
+    return mockReturnBool;
+}
 
 - (void)testActionBitIsSet {
     XCTAssertTrue(ActionBitIsSet(KAUTH_VNODE_READ_DATA, KAUTH_VNODE_READ_DATA));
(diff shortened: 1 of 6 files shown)


ProjFS.Mac/PrjFSKextTests/KextLogMock.h
```diff
@@ -0,0 +1,10 @@
+#ifndef KextLogMock_h
+#define KextLogMock_h
+
+inline void KextLog_Error(const char* fmt, ...) {}
```
jeschu1: Someday we could get fancy with testing logging, but I'll leave that as a future exercise.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -1,11 +1,22 @@
...
 
 @interface KauthHandlerTests : XCTestCase
 @end
 
 @implementation KauthHandlerTests
 
+//Mock for TryReadVNodeFileFlags
+static uint32_t mockFileFlag;
+static bool mockReturnBool;
+bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t context, uint32_t* flags);
```
pmj: If you need this extra prototype to silence warnings, there's some other problem - you've got a declaration in the header, so this one shouldn't be needed.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -32,4 +43,33 @@ - (void)testFileFlagsBitIsSet {
...
+    XCTAssertFalse(ShouldIgnoreVnodeType(VLNK, NULL));
+    XCTAssertFalse(ShouldIgnoreVnodeType(VSTR, NULL));
+    XCTAssertFalse(ShouldIgnoreVnodeType(VCPLX, NULL));
+}
+
+- (void)testFileFlaggedInRoot {
+    bool fileFlaggedInRoot;
+    mockReturnBool = false;
+    mockFileFlag = FileFlags_IsInVirtualizationRoot;
+    XCTAssertFalse(TryGetFileIsFlaggedAsInRoot(reinterpret_cast<vnode_t>(static_cast<uintptr_t>(1)), NULL, &fileFlaggedInRoot));
```
pmj: I still don't like this `reinterpret_cast` hack, although I think my previous comment about it was in one of @wilbaker's PRs, so I'll restate it here:

I'm fairly sure passing around pointers like this `reinterpret_cast<vnode_t>(static_cast<uintptr_t>(1))` is undefined behaviour, plus it's not exactly terribly readable. How about something like this at the top of the file:

```
// dummy struct definition
struct vnode
{
};

```
and then in the tests:

```
vnode test_vnode = {};
XCTAssertFalse(TryGetFileIsFlaggedAsInRoot(&test_vnode, NULL, &fileFlaggedInRoot));
```

jeschu1:
@pmj this is a great feedback thank you!  I knew this one would be controversial so definitely wanted to get thoughs on it.  Moving functionality to a helper is very unideal so anything that nixes that approach and gives us testability is a win.

I like the idea of just wrapping  "vnode_getattr" and that removes a lot complexity.  I need confirmation on that approach so I'll followup offline.  If we can't use that approach the swap technique you mentioned is a better alternative to what I have now.

PR #733, opened 2019-02-01
Mac: Support Version Number overrides

nickgra:
Currently, we don't have a solution for building a version of VFSForGit on MacOS that has a build number other than the dev number that is hardcoded into GVFS.props.

On Windows, we support taking the Version as an argument to the build script and plumb it through to MSBuild. The same approach works on MacOS but the way that we handle the version number (custom scripts instead of GVFS.PreBuild) makes this a little less straightforward. We'll repurpose `GenerateCommonAssemblyVersion.sh` to also update the `GVFS.props` file with the new version as it's consumed by other things than `CommonAssemblyVersion.cs` (The Installers are a good example)

Our build definitions will need to be updated to get a version number from Azure Pipelines and to pass that into BuildGVFSForMac.sh.

Also: Fix an issue with the meta package installer where a find command would dig through your entire drive hierarchy to find a folder mounted at /Volumes/Git <Number>. Set -maxdepth to 1 to get the intended behavior.

### DIFF
--- a/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
+++ b/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
@@ -125,7 +125,7 @@ function CreateMetaInstaller()
     echo "$mountCmd"
     eval $mountCmd || exit 1
     
-    MOUNTEDVOLUME=`/usr/bin/find /Volumes -type d -name "Git $GITVERSIONSTRING*"`
+    MOUNTEDVOLUME=`/usr/bin/find /Volumes -maxdepth 1 -type d -name "Git $GITVERSIONSTRING*"`
     GITINSTALLERPATH=`/usr/bin/find "$MOUNTEDVOLUME" -type f -name "git-$GITVERSIONSTRING*.pkg"`
     
     if [ ! -f "$GITINSTALLERPATH" ]; then
--- a/Scripts/Mac/GenerateCommonAssemblyVersion.sh
+++ b/Scripts/Mac/GenerateCommonAssemblyVersion.sh
@@ -1,11 +1,16 @@
 . "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
 
-GVFSPROPS=$VFS_SRCDIR/GVFS/GVFS.Build/GVFS.props
-VERSIONNUMBER="$(cat $GVFSPROPS | grep GVFSVersion | grep -Eo '[0-9.]+(-\w+)?')"
+if [ -z $1 ]; then
+  echo "Version Number not defined for CommonAssemblyVersion.cs"
+fi
 
+# Update the version number in GVFS.props for other consumers of GVFSVersion
+sed -i "" -E "s@<GVFSVersion>[0-9]+(\.[0-9]+)*</GVFSVersion>@<GVFSVersion>$1</GVFSVersion>@g" $VFS_SRCDIR/GVFS/GVFS.Build/GVFS.props
+
+# Then generate CommonAssemblyVersion.cs
 cat >$VFS_OUTPUTDIR/CommonAssemblyVersion.cs <<TEMPLATE
 using System.Reflection;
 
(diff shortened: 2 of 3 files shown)


Scripts/Mac/GenerateCommonAssemblyVersion.sh
```diff
@@ -1,11 +1,16 @@
 . "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
 
-GVFSPROPS=$VFS_SRCDIR/GVFS/GVFS.Build/GVFS.props
-VERSIONNUMBER="$(cat $GVFSPROPS | grep GVFSVersion | grep -Eo '[0-9.]+(-\w+)?')"
+if [ -z $1 ]; then
+  echo "Version Number not defined for CommonAssemblyVersion.cs"
+fi
 
+# Update the version number in GVFS.props for other consumers of GVFSVersion
+sed -i "" -E "s@<GVFSVersion>[0-9]+(\.[0-9]+)*</GVFSVersion>@<GVFSVersion>$1</GVFSVersion>@g" $VFS_SRCDIR/GVFS/GVFS.Build/GVFS.props
```
kewillford: Do we really need the regex for the version number?  Isn't it hard coded in the GVFS.props?  Should we just hard code it here too?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -125,7 +125,7 @@ function CreateMetaInstaller()
     echo "$mountCmd"
     eval $mountCmd || exit 1
     
-    MOUNTEDVOLUME=`/usr/bin/find /Volumes -type d -name "Git $GITVERSIONSTRING*"`
+    MOUNTEDVOLUME=`/usr/bin/find /Volumes -maxdepth 1 -type d -name "Git $GITVERSIONSTRING*"`
```
wilbaker: Do you know why we need to run find at all?  Does the script not know where to expect the installer? 

cc: @alameenshah

Scripts/Mac/GenerateCommonAssemblyVersion.sh
nickgra: I used a regex rather than hardcoding the build number for scenarios where the user runs a build with a custom build number multiple times in a row. If we hardcode the number, this scenario wouldn't work (we wouldn't match the GVFSVersion line once it's set to 1.0.0.0): 
```
./BuildGVFSForMac.sh Debug 1.0.0.0
./BuildGVFSForMac.sh Debug 1.0.0.1
```

I doubt anyone will ever set the version outside of CI/CD so maybe this is a moot point. I could clarify in the comment since Regex's tend to be hard to reason about.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
nickgra: The naming convention of the downloaded .nupkg, the .dmg inside of it, the name of the volume of the .dmg will mount and the .pkg that we'll actually install/bundle is not consistent. 

We have a bunch of code like this where we have to use find to figure out what the exact name/path of the next phase of our pipeline is going to be.

From the build script:
```
echo 'Downloading a VFS-enabled version of Git...'
$VFS_SCRIPTDIR/DownloadGVFSGit.sh || exit 1
GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
GITPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.dmg)" || exit 1
echo "Downloaded Git $GITVERSION"
# Now that we have a path containing the version number, generate GVFSConstants.GitVersion.cs
$VFS_SCRIPTDIR/GenerateGitVersionConstants.sh "$GITPATH" $BUILDDIR || exit 1
```

In the future, I'd love to rework how the Git installer works to reduce some of the complexity here.

Issue #721, opened 2019-01-29
Mac: 'git status' can miss newly created files whose file descriptor has not yet been closed


wilbaker:
Found this while running tests with `osacompile`.

It appears that the native library can receive `HandleNewFileInRootNotification` callbacks *after* git status runs for files that were created before `git status` was run:

**Commands**

```
~/GVFSTest/src/libraries>rm TestLib.scpt; osacompile -o TestLib.scpt -x -d TestLib.applescript
~/GVFSTest/src/libraries>git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
~/GVFSTest/src/libraries>git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   TestLib.scpt

no changes added to commit (use "git add" and/or "git commit -a")
```

**gvfs.mount logging**

```
PrjFSLib.HandleFileNotification: libraries/TestLib.scpt notificationType: PrjFS_NotificationType_PreDelete isDirectory: 0
[2019-01-28 15:35:34 -08:00] TryAcquireLockExternal {"LockRequest":"git status (57730)","IsElevated":false,"Result":"Accepted"}
HandleNewFileInRootNotification: .git/index.lock notificationType: PrjFS_NotificationType_NewFileCreated isDirectory: 0
PrjFSLib.HandleFileNotification: .git/index.lock notificationType: PrjFS_NotificationType_NewFileCreated isDirectory: 0
PrjFSLib.HandleFileNotification: .git/index.lock notificationType: PrjFS_NotificationType_PreDelete isDirectory: 0
[2019-01-28 15:35:34 -08:00] ReleaseExternalLockRequested
[2019-01-28 15:35:34 -08:00] ReleaseLockHeldByExternalProcess {"CurrentLockHolder":"git status (57730)","IsElevated":false,"EnlistmentId":"fd36b952a2ea415bb73249888c4ba054","Result":"Released","DurationMS":152,"LockHeldExternallyMS":152,"ParseGitIndexMS":0,"UpdatePlaceholdersMS":0,"BlobsDownloaded":0,"BlobDownloadTimeMS":0,"CommitsAndTreesDownloaded":0,"CommitsAndTreesDownloadTimeMS":0,"SizeQueries":0,"SizeQueryTimeMS":0}
[...]

Issue #209, opened 2018-08-23
Mac: Add PreConvertToFull notification that's driven by vnode events


wilbaker:
This notification could be sent when the KAUTH_VNODE_WRITE_DATA is set and the KAUTH_VNODE_ACCESS is *not* set.

Based on the discussion in #328 the current file modified notification is not always reliable and because status caching is not enabled on the Mac the file modified notification can be removed as part of this change.

Before merging the changes for this PR we should confirm there aren't too many entries being added to ModifiedPaths.dat.

wilbaker:
Related to #152

wilbaker:
Also see discussion in #643 on how the file modified notification is not always reliable

wilbaker:
This is also required so that VFS4G stops writing `D` entries into the Placeholder list every time a file is saved.

wilbaker:
There are actually two separate but related issues:

- The current file modified notification is not always reliable
- Calling `OnFileConvertedToFull` for every file modification results in the file being written to the PlaceholderList multiple times with a `D`

wilbaker:
If this is moved out of "macOS Next" a separate issue should be filed for:

- Calling OnFileConvertedToFull for every file modification results in the file being written to the PlaceholderList multiple times with a D

And fixed in "macOS Next"

PR #734, opened 2019-02-01
Add listener for OPEN event

jeschu1:
Previously we only listened to the CLOSE event and then checked for KAUTH_FILEOP_CLOSE_MODIFIED or NewFile (fileFlaggedInRoot == false) to determine if we should send an event.  With this change, we do the NewFile check on OPEN which allows us to get the event earlier.  

The underlying issue is a file handle can be opened and not closed by the process.

This addresses #721.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
@@ -4,6 +4,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Threading;
 
 namespace GVFS.FunctionalTests.FileSystemRunners
@@ -19,6 +20,17 @@ namespace GVFS.FunctionalTests.FileSystemRunners
         {
             File.Move(sourcePath, targetPath);
             return string.Empty;
+        }
+
+        public override void CreateFileWithoutClose(string path)
+        {
+            File.Create(path);
+        }
+
+        public override void OpenFileAndWriteWithoutClose(string path, string content)
+        {
+            StreamWriter file = new StreamWriter(path);
+            file.Write(content);
         }
 
         public override void MoveFileShouldFail(string sourcePath, string targetPath)
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
@@ -39,6 +39,23 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         }
 
         [TestCase]
+        public void CreateFileWithoutClose()
+        {
+            string srcPath = @"CreateFileWithoutClose.md";
+            this.CreateFileWithoutClose(srcPath);
+            this.ValidateGitCommand("status");
+        }
+
+        [Ignore("This test exposes a bug we don't yet have a fix for")]
+        [TestCase]
+        public void WriteWithoutClose()
+        {
+            string srcPath = @"Readme.md";
+            this.OpenFileAndWriteWithoutClose(srcPath, "More Stuff");
+            this.ValidateGitCommand("status");
+        }
+
+        [TestCase]
         [Category(Categories.MacTODO.M4)]
         public void ModifyingAndDeletingRepositoryExcludeFileInvalidatesCache()
         {
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -618,6 +618,66 @@ static int HandleFileOpOperation(
             }
         }
     }
+    else if (KAUTH_FILEOP_OPEN == action)
+    {
+        currentVnode = reinterpret_cast<vnode_t>(arg0);
+        putCurrentVnode = false;
+        const char* path = reinterpret_cast<const char*>(arg1);
+
+        if (vnode_isdir(currentVnode))
+        {
+            goto CleanupAndReturn;
+        }
+
+        UseMainForkIfNamedStream(currentVnode, putCurrentVnode);
+
+        bool fileFlaggedInRoot;
+        if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
+        {
+            KextLog_ErrorVnodeProperties(currentVnode, "KAUTH_FILEOP_OPEN: checking file flags failed. Path = '%s'", path);
+            
+            goto CleanupAndReturn;
+        }
+        
+        if (fileFlaggedInRoot)
+        {
+            goto CleanupAndReturn;
+        }
+        
+        VirtualizationRootHandle root = RootHandle_None;
+        FsidInode vnodeFsidInode;
+        int pid;
+        if (!ShouldHandleFileOpEvent(
+                                     &perfTracer,
+                                     context,
+                                     currentVnode,
+                                     action,
+                                     &root,
+                                     &vnodeFsidInode,
+                                     &pid))
+        {
+            goto CleanupAndReturn;
+        }
+        
+        char procname[MAXCOMLEN + 1];
+        proc_name(pid, procname, MAXCOMLEN + 1);
+        PerfSample fileCreatedSample(&perfTracer, PrjFSPerfCounter_FileOp_FileCreated);
+        int kauthResult;
+        int kauthError;
+        if (!TrySendRequestAndWaitForResponse(
+                                              root,
+                                              MessageType_KtoU_NotifyFileCreated,
+                                              currentVnode,
+                                              vnodeFsidInode,
+                                              path,
+                                              pid,
+                                              procname,
+                                              &kauthResult,
+                                              &kauthError))
+        {
+            goto CleanupAndReturn;
+        }
+    }
     else if (KAUTH_FILEOP_CLOSE == action)
     {
         currentVnode = reinterpret_cast<vnode_t>(arg0);
@@ -631,16 +691,8 @@ static int HandleFileOpOperation(
         }
         
         UseMainForkIfNamedStream(currentVnode, putCurrentVnode);
-
-        bool fileFlaggedInRoot;
-        if (!TryGetFileIsFlaggedAsInRoot(currentVnode, context, &fileFlaggedInRoot))
-        {
-            KextLog_ErrorVnodeProperties(currentVnode, "KAUTH_FILEOP_CLOSE: checking file flags failed. Path = '%s'", path);
-            
-            goto CleanupAndReturn;
-        }
         
-        if (fileFlaggedInRoot && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
+        if (KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
         {
             goto CleanupAndReturn;
         }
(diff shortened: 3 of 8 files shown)


jeschu1:
Is it worth changing
   if (fileFlaggedInRoot && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
      to
   if (fileFlaggedInRoot && (KAUTH_FILEOP_CLOSE == action && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)

Is it safe to assum we won't have clashes with KAUTH_FILEOP_CLOSE_MODIFIED on OPEN

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -618,7 +618,8 @@ static int HandleFileOpOperation(
             }
         }
     }
-    else if (KAUTH_FILEOP_CLOSE == action)
+    else if (KAUTH_FILEOP_CLOSE == action ||
+             KAUTH_FILEOP_OPEN == action)
```
wilbaker: Question related to this comment:

>Is it worth changing
if (fileFlaggedInRoot && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
to
if (fileFlaggedInRoot && (KAUTH_FILEOP_CLOSE == action && KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)
>
>Is it safe to assume we won't have clashes with KAUTH_FILEOP_CLOSE_MODIFIED on OPEN

With the current changes it's not clear if `KAUTH_FILEOP_CLOSE` or `KAUTH_FILEOP_OPEN` is responsible for sending the `MessageType_KtoU_NotifyFileCreated` message.  

Should it to be exclusively the responsibility of `KAUTH_FILEOP_OPEN`?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -618,7 +618,8 @@ static int HandleFileOpOperation(
...
     {
         currentVnode = reinterpret_cast<vnode_t>(arg0);
         putCurrentVnode = false;
```
wilbaker: If we remove `if (vnode_isdir(currentVnode))` from the `KAUTH_FILEOP_OPEN` path will it allow us to receive new file notifications for directories? (it would be awesome if it did 😃 )

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: > Should it to be exclusively the responsibility of `KAUTH_FILEOP_OPEN`?

My opinion is that new file notifications should exclusively be the responsibility of `KAUTH_FILEOP_OPEN`.  We know that `KAUTH_FILEOP_CLOSE` is sometimes too late, and restricting the callback to a specific notification will make that notification more predictable and easier to debug (if it's not delivered when expected).

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Looks like it.  You see the notifications when you try to access the directory (not on mkdir).  What are you thinking here?

wilbaker: > Looks like it. You see the notifications when you try to access the directory (not on mkdir). What are you thinking here?

Ahh, okay, I was wondering if we'd get it on mkdir.  Currently we detect new directories when a new file is created inside of them, and I wasn't sure if the `KAUTH_FILEOP_OPEN` notification would allow us to detect those new directories earlier.

I don't think it's something we need to change now, but we could consider changing when we trigger the notification for directories (so that they'll happen even without a file being created inside of them).

jeschu1: Sounds good!, @pmj mentioned we will get a KAUTH_VNODE_ADD_SUBDIRECTORY on the parent folder.  Thats another option we could look into.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: For *new file* notifications I agree we should stick with `KAUTH_FILEOP_OPEN` only. I don't think there's any situation where the open wouldn't catch it but the close event would.

For *modified* notifications note that `KAUTH_FILEOP_OPEN` is also delivered for files opened for *reading*, and there is no way for our code to know which kind of access it was opened for. So this will almost certainly generate a ton of false positives. The only way around that would be to correlate `KAUTH_VNODE_WRITE_DATA` events with `KAUTH_FILEOP_OPEN`. If we get a `KAUTH_VNODE_WRITE_DATA` event that is immediately followed by a `KAUTH_FILEOP_OPEN` from the same thread and PID, we can be fairly certain the open was for writing. The question there is if we really want to be keeping track of that kind of state. `KAUTH_FILEOP_CLOSE` of course has the disadvantage of potentially being delivered quite late if the handle is open for long.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Just to expand on that: the FILEOP open/close events will almost certainly only be delivered when the corresponding syscalls are made, and directories are only `open()`ed when *listing* directory contents. `mkdir` does not produce a file descriptor, and so does not act like an `open()` call in any way.

There is definitely a [`KAUTH_VNODE_ADD_SUBDIRECTORY` auth query](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/vfs_subr.c#L6399), but [no FILEOP event at all produced by `mkdir` syscalls](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/vfs_syscalls.c#L7997) - although it is [reported to FSEvents](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/vfs_syscalls.c#L8016). I don't know to what extent [FSEvents](https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/UsingtheFSEventsFramework/UsingtheFSEventsFramework.html) have been investigated for the change detection aspect of VFS4G; it obviously is no good for hydration etc. but should give fewer false positives than vnode listeners. They are of course entirely async, but should arrive in a more timely manner than `KAUTH_FILEOP_CLOSE` at any rate.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @wilbaker testing this out now, but I separated the OPEN into it's own flow as discussed.  There is some overlap with CLOSE (both do the directory check and ShouldProcess check), but I think their behavior is very easy to read now.

@pmj thanks for the extra info regarding modifications and this was actually what I was going to look at next.  It looks like if we get` KAUTH_VNODE_WRITE_DATA` we Hydrate a file.  Could we instead use that flag to say a file has been modified?

jeschu1: @pmj that makes sense thanks!  As long as we don't expect tons of false positives I would think that's better than missing the event.  I'll test it out and see how severe things look.  Sounds like the other alternative is pairing with `KAUTH_FILE_OPEN` which as you said becomes more involved.

wilbaker: > As long as we don't expect tons of false positives I would think that's better than missing the event. I'll test it out and see how severe things look. 

This would actually be more consistent with our behavior on Windows from the perspective of VFS4G.  ProjFS on Windows delivers a PreConvertToFull callback when a file is opened for write (even if no writes actually occur).

> Waiting for a response from user space on KAUTH_VNODE_WRITE_DATA could have a measurable performance impact. As an optimisation we could consider marking this and possibly some other notifications as not requiring a response from the provider, as I don't think we really need to wait in this case. This will require some not entirely trivial changes to both kext and lib.

To provide a bit more context here the scenario we're looking to fix is:

```
<modify placeholder file>
git status
```

The `git status` should report the placeholder file as modified, but today we detect the modification in the `KAUTH_FILEOP_CLOSE` event (which @jeschu1 has proven can be too late).

It's also worth noting that for correctness we only have to deliver this event (PreConvertToFull) for files that are marked as placeholders (and only once per-file).  Reducing the scope of the callback to these specific files will help reduce the performance impact of making blocking calls to the provider.

(Also as an FYI - #209 is tracking this work)

jeschu1: Since it's somewhat related to this change I added a functional test OpenFileAndWriteWithoutClose demonstrating the modify issue.  I've set it to Ignore since we currently don't have a fix.  I'm going to resolve this thread since it's an issue we'll tackle in another PR.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
```diff
@@ -36,6 +36,23 @@ public void DeleteThenCreateThenDeleteFile()
...
+
+        [TestCase]
+        public void CreateFileWithoutClose()
+        {
+            string srcPath = @"CreateFileWithoutClose.md";
+            this.CreateFileWithoutClose(srcPath);
+            this.ValidateGitCommand("status");
+        }
+
+        [Ignore("This test exposes a bug we don't yet have a fix for")]
```
jeschu1: The fix here will be addressed in a separate.  I wanted to validate the issue / my understanding which is why I'm adding the functional test now.

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
```diff
@@ -19,6 +20,17 @@ public override string MoveFile(string sourcePath, string targetPath)
         {
             File.Move(sourcePath, targetPath);
             return string.Empty;
+        }
+
+        public override void CreateFileWithoutClose(string path)
+        {
+            File.Create(path);
```
wilbaker: Any ideas on why there was trouble getting this to repro before with `File.Create`?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
wilbaker: You could include in the ignore message that https://github.com/Microsoft/VFSForGit/issues/209 should fix this issue.

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
pmj: My .NET experience is limited, but I don't see how this inhibits a `close()` on the file descriptor? Surely, `File.Create` returns a `FileStream` which is shortly after GC'd (as it's never stored) and thus closed? Are we relying on GC timing in this test?

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
```diff
@@ -19,6 +20,17 @@ public override string MoveFile(string sourcePath, string targetPath)
...
+
+        public override void CreateFileWithoutClose(string path)
+        {
+            File.Create(path);
+        }
+
+        public override void OpenFileAndWriteWithoutClose(string path, string content)
+        {
+            StreamWriter file = new StreamWriter(path);
+            file.Write(content);
```
pmj: Similarly to the `CreateFileWithoutClose` case, I'm not clear on how this inhibits a `close()` syscall; also, is the `StreamWriter` buffered by default? If the buffers aren't flushed, the write won't hit disk, so you wouldn't expect `git status` to pick them up?

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
jeschu1: @wilbaker to answer your question, user error is likely the reason why it didn't repro the first time :-) I tried several times yesterday with repos.

@pmj you are correct that Dispose may eventually get called, but it certainly isn't immediate.  I ran the test several times seeing it fail every time in the old kext.  The alternative of using the low-level windows/mac calls has a lot more overhead and IMHO not worth it since this simpler method causes a repro.

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
jeschu1: @pmj I validated the behavior by turning on the test and running a few times.  When I address the related bug I'll dig more.  If AuthFlush isn't true it's probably worth adding a .Flush for good measure.  Thank you!

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
pmj: OK, sounds good then.

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
pmj: Yeah, I couldn't work out from the docs if `AutoFlush` was true by default.

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
wilbaker: > you are correct that Dispose may eventually get called, but it certainly isn't immediate

@jeschu1 do you think we should have the helper method return the file reference so that the test can properly dispose it after running `git status`? (e.g. just have `CreateFile` and `OpenFileAndWrite` and have the test do the close)

That would also prevent the finalizer from kicking in too early in the test.

PR #735, opened 2019-02-02
[features/upgrade_extensibility] Additional ACL changes

wilbaker:
Part of the work required for #717 

After doing testing with the changes in #728 I found the following issues:

- The upgrader's logs directory could not be written to by a non-elevated process
- Download and Tools directory were not having the proper ACLs explicitly set

Changes in this PR:

- Explicitly remove any ACLs previously set on the `C:\ProgramData\GVFS` directory
- Explicitly remove any ACLs previously set on the `C\ProgramData\GVFS\GVFS.Upgrade\Downloads` directory
- Allow users to modify the `GVFS.Upgrade\Logs` directory so that `gvfs upgrade` can create log files when running without elevation
- Add\Update unit tests
- Fixed an issue where running the unit tests would attempt to delete the downloads directory on the dev machine (running the unit tests)
- Renamed `GVFS.Common.NuGetUpgrader` to `GVFS.Common.NuGetUpgrade` to avoid naming conflicts with the `NuGetUpgrader` class.
- Code cleanup

*Tagging as WIP as manual testing is still in progress*

### DIFF
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -49,17 +49,22 @@ namespace GVFS.Common
 
         public GitHubUpgraderConfig Config { get; private set; }
 
-        public static GitHubUpgrader Create(ITracer tracer, PhysicalFileSystem fileSystem, bool dryRun, bool noVerify, out string error)
-        {
-            return Create(new LocalGVFSConfig(), tracer, fileSystem, dryRun, noVerify, out error);
-        }
-
         public static GitHubUpgrader Create(
-            LocalGVFSConfig localConfig,
             ITracer tracer,
             PhysicalFileSystem fileSystem,
             bool dryRun,
             bool noVerify,
+            out string error)
+        {
+            return Create(tracer, fileSystem, dryRun, noVerify, new LocalGVFSConfig(), out error);
+        }
+
+        public static GitHubUpgrader Create(
+            ITracer tracer,
+            PhysicalFileSystem fileSystem,
+            bool dryRun,
+            bool noVerify,
+            LocalGVFSConfig localConfig,
             out string error)
         {
             GitHubUpgrader upgrader = null;
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -238,11 +238,9 @@ namespace GVFS.Common
             errorMessage = null;
 
             string downloadPath = ProductUpgraderInfo.GetAssetDownloadsPath();
-            Exception exception;
-            if (!this.fileSystem.TryCreateDirectory(downloadPath, out exception))
+            if (!this.fileSystem.DirectoryExists(downloadPath))
             {
-                errorMessage = exception.Message;
-                this.TraceException(exception, nameof(this.TryDownloadAsset), $"Error creating download directory {downloadPath}.");
+                errorMessage = $"Download directory `{downloadPath}` does not exist";
                 return false;
             }
 
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -35,10 +35,11 @@ namespace GVFS.Common
         public GitHubUpgrader(
             string currentVersion,
             ITracer tracer,
+            PhysicalFileSystem fileSystem,
             GitHubUpgraderConfig upgraderConfig,
             bool dryRun = false,
             bool noVerify = false)
-            : base(currentVersion, tracer, dryRun, noVerify, new PhysicalFileSystem())
+            : base(currentVersion, tracer, dryRun, noVerify, fileSystem)
         {
             this.Config = upgraderConfig;
 
@@ -48,12 +49,18 @@ namespace GVFS.Common
 
         public GitHubUpgraderConfig Config { get; private set; }
 
-        public static GitHubUpgrader Create(ITracer tracer, bool dryRun, bool noVerify, out string error)
+        public static GitHubUpgrader Create(ITracer tracer, PhysicalFileSystem fileSystem, bool dryRun, bool noVerify, out string error)
         {
-            return Create(new LocalGVFSConfig(), tracer, dryRun, noVerify, out error);
+            return Create(new LocalGVFSConfig(), tracer, fileSystem, dryRun, noVerify, out error);
         }
 
-        public static GitHubUpgrader Create(LocalGVFSConfig localConfig, ITracer tracer, bool dryRun, bool noVerify, out string error)
+        public static GitHubUpgrader Create(
+            LocalGVFSConfig localConfig,
+            ITracer tracer,
+            PhysicalFileSystem fileSystem,
+            bool dryRun,
+            bool noVerify,
+            out string error)
         {
             GitHubUpgrader upgrader = null;
             GitHubUpgraderConfig gitHubUpgraderConfig = new GitHubUpgraderConfig(tracer, localConfig);
@@ -72,6 +79,7 @@ namespace GVFS.Common
             upgrader = new GitHubUpgrader(
                     ProcessHelper.GetCurrentProcessVersion(),
                     tracer,
+                    fileSystem,
                     gitHubUpgraderConfig,
                     dryRun,
                     noVerify);
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -130,6 +130,12 @@ namespace GVFS.Common
 
         public override bool TryDownloadNewestVersion(out string errorMessage)
         {
+            if (!this.TryCreateAndConfigureDownloadDirectory(this.tracer, out errorMessage))
+            {
+                this.tracer.RelatedError($"{nameof(GitHubUpgrader)}.{nameof(this.TryCreateAndConfigureDownloadDirectory)} failed. {errorMessage}");
+                return false;
+            }
+
             bool downloadedGit = false;
             bool downloadedGVFS = false;
 
@@ -246,12 +252,6 @@ namespace GVFS.Common
             errorMessage = null;
 
             string downloadPath = ProductUpgraderInfo.GetAssetDownloadsPath();
-            if (!this.fileSystem.DirectoryExists(downloadPath))
-            {
-                errorMessage = $"Download directory `{downloadPath}` does not exist";
-                return false;
-            }
-
             string localPath = Path.Combine(downloadPath, asset.Name);
             WebClient webClient = new WebClient();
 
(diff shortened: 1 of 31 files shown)


wilbaker:
@jrbriggs @jamill @alameenshah I went ahead and created this PR so that I could get some early feedback on the changes.  

Please note that I am still testing these changes out and so I've tagged the PR as WIP.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -114,18 +115,15 @@ protected ProductUpgrader()
         /// Deletes any previously downloaded installers in the Upgrader Download directory.
         /// This can include old installers which were downloaded but never installed.
         /// </summary>
-        public static void DeleteAllInstallerDownloads(ITracer tracer = null)
+        public virtual void DeleteAllInstallerDownloads(ITracer tracer)
         {
             try
             {
-                PhysicalFileSystem.RecursiveDelete(ProductUpgraderInfo.GetAssetDownloadsPath());
+                this.fileSystem.RecursiveDelete(ProductUpgraderInfo.GetAssetDownloadsPath());
```
wilbaker: Without this change (and the corresponding fallout of having to pass `fileSystem` around) the unit tests were deleting the download path on my dev machine.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -164,7 +169,7 @@ private bool TryRunUpgrade(out Version newVersion, out string consoleError)
 
             if (!this.upgrader.UpgradeAllowed(out error))
             {
-                ProductUpgrader.DeleteAllInstallerDownloads();
+                this.upgrader.DeleteAllInstallerDownloads(this.tracer);
```
wilbaker: @jamill @alameenshah should we actually always be calling `DeleteAllInstallerDownloads` in this method?

I believe the idea is that the upgrader is always supposed to re-download the upgrades, but it wasn't clear to me what the NuGet and GitHub upgraders would do if the package that we want is already on disk in the downloads folder.

jamill: I think this is OK - if this check fails, it is most likely due to the fact that GVFS is not configured for upgrades. If something was downloaded already, then the most likely scenario is that someone changed the config after a package was downloaded...

@alameenshah - what do you think?

related - the proposed change in #736 no longer downloads the update in the background

wilbaker: > I think this is OK - if this check fails, it is most likely due to the fact that GVFS is not configured for upgrades. If something was downloaded already, then the most likely scenario is that someone changed the config after a package was downloaded...

I think I could have worded my initial comment a bit better.  To clarify I'm wondering if `DeleteAllInstallerDownloads()` should be moved out of the check and always called.

If a file already exists at the download location, will the NuGet and GitHub upgraders overwrite it?  Based on the offline conversations we've had I thought that we wanted to make sure that the installer would always run a freshly downloaded copy.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -238,11 +246,9 @@ protected virtual bool TryDownloadAsset(Asset asset, out string errorMessage)
             errorMessage = null;
 
             string downloadPath = ProductUpgraderInfo.GetAssetDownloadsPath();
-            Exception exception;
-            if (!this.fileSystem.TryCreateDirectory(downloadPath, out exception))
+            if (!this.fileSystem.DirectoryExists(downloadPath))
```
jamill: In this case - whose responsibility it is to create the download directory?

I had approached this as the Download method would ensure the download directory exists and create it if necessary.

wilbaker: I went back and forth on how\where to make this change.

With these changes it's the responsibility of the `ProductUpgradeTimer` (in GVFS.Service) and the `UpgradeOrchastrator` (in the `gvfs` verb).

My reasoning was:

- Creating the download directory is a common task (for both upgraders), and I wanted to create the directory in a common spot for both.

- Avoid growing the scope of `TryDownloadAsset` (now that we need to set ACLs as well)

Another alternative I considered was making the call in both upgrader's versions of `TryDownloadNewestVersion`.

@jamill I could not figure out where this directory was getting created for the NuGet upgrader, do you know if the nuget library was creating it if it didn't exist?

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -114,18 +115,15 @@ protected ProductUpgrader()
         /// Deletes any previously downloaded installers in the Upgrader Download directory.
         /// This can include old installers which were downloaded but never installed.
         /// </summary>
-        public static void DeleteAllInstallerDownloads(ITracer tracer = null)
+        public virtual void DeleteAllInstallerDownloads(ITracer tracer)
```
jamill: Should we use the class instance version of `tracer`, and not pass one in?

This is still a bit of a static-y method (only using the `filesystem` instance)... We don't need a GitHub or a NuGet upgrader to actually delete this directory. Would it be easier to just pass in the filesystem instance to use?

I don't feel strongly on this, either way :)

GVFS/GVFS.Common/GitHubUpgrader.cs
jamill: > @jamill I could not figure out where this directory was getting created for the NuGet upgrader, do you know if the nuget library was creating it if it didn't exist?

Yes, the NuGet library was creating it as part of downloading the package if it didn't exist.

I like having the `Upgraders` `TryDownloadNewestVersion` being responsible for creating the directory if it doesn't exist, as this will remove an interaction / expectation between the UpgradeOrchestrator and the Upgraders. The fact that the NuGet library was creating the directory was covering for the upgrader missing this logic in the 1st place.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
wilbaker: Per conversation with @jamill offline, the upgraders will delete any existing files if present

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -30,6 +32,7 @@ public class UpgradeVerb : GVFSVerb.ForNoEnlistment
         {
             this.upgrader = upgrader;
             this.tracer = tracer;
+            this.fileSystem = new PhysicalFileSystem();
```
alameenshah: Can we add this line in the default constructor as well. I see a crash with `System.NullReferenceException` without it.

GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -355,37 +355,103 @@ private void CreateAndConfigureProgramDataDirectories()
...
+            // Remove any rules already set for the directory
+            AuthorizationRuleCollection currentRules = serviceDataRootSecurity.GetAccessRules(includeExplicit: true, includeInherited: true, targetType: typeof(NTAccount));
+            foreach (AuthorizationRule authorizationRule in currentRules)
+            {
+                FileSystemAccessRule fileSystemRule = authorizationRule as FileSystemAccessRule;
+                if (fileSystemRule != null)
+                {
+                    serviceDataRootSecurity.RemoveAccessRule(fileSystemRule);
+                }
+            }
```
alameenshah: There is some code duplication between `WindowsFileSystem.cs` and the lines above. `Service` is trying to create a directory with a set of permissions; wouldn't it be better to create a generic function which does that inside the platform specific `WindowsFileSystem` and make `Service` use it.

GVFS/GVFS.Common/ProductUpgrader.cs
wilbaker: After giving both a try it looks a bit cleaner to not have to pass through the trace and filesystem.  I updated the method to use the class instance version of `tracer`.

GVFS/GVFS.Service/GvfsService.cs
wilbaker: Good suggestion, I've moved the common code into helper methods.  Once I push the latest changes let me know what you think.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
wilbaker: Yep nice catch, I am confirming this issue is caught by the full functional tests before pushing a fix.

GVFS/GVFS.Common/NuGetUpgrader/NuGetFeed.cs
```diff
@@ -57,6 +56,8 @@ public class NuGetFeed : IDisposable
             }
         }
 
+        public string DownloadFolder { get; }
```
wilbaker: There is no need to make this a public property

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -141,16 +139,14 @@ public virtual bool TrySetupToolsDirectory(out string upgraderToolPath, out stri
...
-            {
-                upgraderToolPath = null;
-                error = exception.Message;
-                this.TraceException(
-                    exception,
-                    nameof(this.TrySetupToolsDirectory),
-                    $"Error creating upgrade tools directory {toolsDirectoryPath}.");
-                return false;
+
+            if (!GVFSPlatform.Instance.FileSystem.TryCreateDirectoryWithAdminOnlyModify(
```
jamill: On first read through - I expected this functionality to be provided by the class `fileSystem` instance. I think it would be preferable if this class only had 1 file system abstraction to deal with, if possible.

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -1,16 +1,77 @@
...
+        }
+
+        /// <summary>
+        /// Adds a new FileSystemAccessRule granting read/exceute/modify/delete access for administrators.
+        /// </summary>
+        /// <param name="directorySecurity">DirectorySecurity to which a FileSystemAccessRule will be added.</param>
+        public static void AddAdminAccessRulesToDirectorySecurity(DirectorySecurity directorySecurity)
+        {
+            SecurityIdentifier administratorUsers = new SecurityIdentifier(WellKnownSidType.BuiltinAdministratorsSid, null);
+            directorySecurity.AddAccessRule(
```
jamill: A comment about why we chose these particular values might be useful - I know we spent a bit of time going through some of the combinations.

GVFS/GVFS.Common/ProductUpgrader.cs
wilbaker: > What is the distinction between `PhysicalFileSystem` and `IPlatformFileSystem`, and why did we chose to put the directory creation code in the latter.

The (not always perfect) distinction is that `IPlatformFileSystem` has any file system methods that need to be platform specific.  Typically this is because .NET does not provide a managed API for what we need to do, and so we need to PInvoke to the native API.

I do see now that in some other spots we moved the call to `GVFSPlatform.Instance.FileSystem` into `PhysicalFileSystem` and I think we could do the same here.

PR #736, opened 2019-02-03
ProductUpgrader: only write newest version available

jamill:
When an upgrade is available, GVFS will download the update in the
background, and git commands will periodically check for the existence
of the downloaded package in a well known directory to advertise that an
update is available.

However, this downloaded update is not used for the actual update.

This changes the logic to write out a file in a well known location
containing the version of the latest available update, and avoid
downloading the actual update.

While this change stands on its own, it will be even more applicable
with the NuGet Upgrader, which will still use GitHub releases to check
for available updates, but will download the actual update from another
location.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgraderInfo.cs
+++ b/GVFS/GVFS.Common/ProductUpgraderInfo.cs
@@ -5,37 +5,8 @@ using System.IO;
 
 namespace GVFS.Common
 {
-    public class ProductUpgraderInfo
+    public partial class ProductUpgraderInfo
     {
-        public const string UpgradeDirectoryName = "GVFS.Upgrade";
-        public const string LogDirectory = "Logs";
-        public const string DownloadDirectory = "Downloads";
-        public const string HighestAvailableVersionFileName = "HighestAvailableVersion";
-
-        protected const string RootDirectory = UpgradeDirectoryName;
-
-        public static bool IsLocalUpgradeAvailable(ITracer tracer)
-        {
-            try
-            {
-                return File.Exists(GetHighestAvailableVersionFilePath());
-            }
-            catch (Exception ex) when (
-                ex is IOException ||
-                ex is UnauthorizedAccessException ||
-                ex is NotSupportedException)
-            {
-                if (tracer != null)
-                {
-                    tracer.RelatedError(
-                        CreateEventMetadata(ex),
-                        "Exception encountered when determining if an upgrade is available.");
-                }
-            }
-
-            return false;
-        }
-
         public static void RecordHighestAvailableVersion(Version highestAvailableVersion)
         {
             string highestAvailableVersionFile = GetHighestAvailableVersionFilePath();
@@ -56,11 +27,6 @@ namespace GVFS.Common
         public static string CurrentGVFSVersion()
         {
             return ProcessHelper.GetCurrentProcessVersion();
-        }
-
-        public static string GetUpgradesDirectoryPath()
-        {
-            return Paths.GetServiceDataRoot(RootDirectory);
         }
 
         public static string GetLogDirectoryPath()
@@ -95,11 +61,6 @@ namespace GVFS.Common
             }
         }
 
-        public static string GetHighestAvailableVersionFilePath()
-        {
-            return Path.Combine(GetUpgradesDirectoryPath(), HighestAvailableVersionFileName);
-        }
-
         private static void RecursiveDelete(string path)
         {
             if (!Directory.Exists(path))
@@ -121,17 +82,6 @@ namespace GVFS.Common
             }
 
             directory.Delete();
-        }
-
-        private static EventMetadata CreateEventMetadata(Exception e)
-        {
-            EventMetadata metadata = new EventMetadata();
-            if (e != null)
-            {
-                metadata.Add("Exception", e.ToString());
-            }
-
-            return metadata;
         }
     }
 }
--- a/GVFS/GVFS.Common/ProductUpgraderInfo.cs
+++ b/GVFS/GVFS.Common/ProductUpgraderInfo.cs
@@ -1,3 +1,4 @@
+using GVFS.Common.FileSystem;
 using GVFS.Common.Tracing;
 using System;
 using System.Collections.Generic;
@@ -7,21 +8,13 @@ namespace GVFS.Common
 {
     public partial class ProductUpgraderInfo
     {
-        public static void RecordHighestAvailableVersion(Version highestAvailableVersion)
-        {
-            string highestAvailableVersionFile = GetHighestAvailableVersionFilePath();
+        private ITracer tracer;
+        private PhysicalFileSystem fileSystem;
 
-            if (highestAvailableVersion == null)
-            {
-                if (File.Exists(highestAvailableVersionFile))
-                {
-                    File.Delete(highestAvailableVersionFile);
-                }
-            }
-            else
-            {
-                File.WriteAllText(highestAvailableVersionFile, highestAvailableVersion.ToString());
-            }
+        public ProductUpgraderInfo(ITracer tracer, PhysicalFileSystem fileSystem)
+        {
+            this.tracer = tracer;
+            this.fileSystem = fileSystem;
         }
 
         public static string CurrentGVFSVersion()
@@ -46,42 +39,36 @@ namespace GVFS.Common
         /// This can include old installers which were downloaded, but user never installed
         /// using gvfs upgrade and GVFS is now up to date already.
         /// </summary>
-        public static void DeleteAllInstallerDownloads(ITracer tracer = null)
+        public void DeleteAllInstallerDownloads()
         {
             try
             {
-                RecursiveDelete(ProductUpgraderInfo.GetAssetDownloadsPath());
+                PhysicalFileSystem.RecursiveDelete(ProductUpgraderInfo.GetAssetDownloadsPath());
             }
             catch (Exception ex)
             {
-                if (tracer != null)
+                if (this.tracer != null)
                 {
-                    tracer.RelatedError($"{nameof(DeleteAllInstallerDownloads)}: Could not remove directory: {ProductUpgraderInfo.GetAssetDownloadsPath()}.{ex.ToString()}");
+                    this.tracer.RelatedError($"{nameof(this.DeleteAllInstallerDownloads)}: Could not remove directory: {ProductUpgraderInfo.GetAssetDownloadsPath()}.{ex.ToString()}");
                 }
             }
         }
 
-        private static void RecursiveDelete(string path)
+        public void RecordHighestAvailableVersion(Version highestAvailableVersion)
         {
-            if (!Directory.Exists(path))
+            string highestAvailableVersionFile = GetHighestAvailableVersionFilePath();
+
+            if (highestAvailableVersion == null)
             {
-                return;
+                if (this.fileSystem.FileExists(highestAvailableVersionFile))
+                {
+                    this.fileSystem.DeleteFile(highestAvailableVersionFile);
+                }
             }
-
-            DirectoryInfo directory = new DirectoryInfo(path);
-
-            foreach (FileInfo file in directory.GetFiles())
+            else
             {
-                file.Attributes = FileAttributes.Normal;
-                file.Delete();
+                this.fileSystem.WriteAllText(highestAvailableVersionFile, highestAvailableVersion.ToString());
             }
-
-            foreach (DirectoryInfo subDirectory in directory.GetDirectories())
-            {
-                RecursiveDelete(subDirectory.FullName);
-            }
-
-            directory.Delete();
         }
     }
 }
--- a/GVFS/GVFS.Common/ProductUpgraderInfo.cs
+++ b/GVFS/GVFS.Common/ProductUpgraderInfo.cs
@@ -10,39 +10,52 @@ namespace GVFS.Common
         public const string UpgradeDirectoryName = "GVFS.Upgrade";
         public const string LogDirectory = "Logs";
         public const string DownloadDirectory = "Downloads";
+        public const string HighestAvailableVersionFileName = "HighestAvailableVersion";
 
         protected const string RootDirectory = UpgradeDirectoryName;
 
-        private static readonly HashSet<string> GVFSInstallerFileNamePrefixCandidates = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        public static bool IsLocalUpgradeAvailable(ITracer tracer)
         {
-            "SetupGVFS",
-            "VFSForGit"
-        };
-
-        public static bool IsLocalUpgradeAvailable(string installerExtension)
-        {
-            string downloadDirectory = GetAssetDownloadsPath();
-            if (Directory.Exists(downloadDirectory))
+            try
             {
-                foreach (string file in Directory.EnumerateFiles(downloadDirectory, "*", SearchOption.TopDirectoryOnly))
+                return File.Exists(GetHighestAvailableVersionFilePath());
+            }
+            catch (Exception ex) when (
+                ex is IOException ||
+                ex is UnauthorizedAccessException ||
+                ex is NotSupportedException)
+            {
+                if (tracer != null)
                 {
-                    string[] components = Path.GetFileName(file).Split('.');
-                    int length = components.Length;
-                    if (length >= 2 &&
-                        GVFSInstallerFileNamePrefixCandidates.Contains(components[0]) &&
-                        installerExtension.Equals(components[length - 1], StringComparison.OrdinalIgnoreCase))
-                    {
-                        return true;
-                    }
+                    tracer.RelatedError(
+                        CreateEventMetadata(ex),
+                        "Exception encountered when determining if an upgrade is available.");
                 }
             }
 
             return false;
         }
 
-        public static IEnumerable<string> PossibleGVFSInstallerNamePrefixes()
+        public static void RecordHighestAvailableVersion(Version highestAvailableVersion)
         {
-            return GVFSInstallerFileNamePrefixCandidates;
+            string highestAvailableVersionFile = GetHighestAvailableVersionFilePath();
+
+            if (highestAvailableVersion == null)
+            {
+                if (File.Exists(highestAvailableVersionFile))
+                {
+                    File.Delete(highestAvailableVersionFile);
+                }
+            }
+            else
+            {
+                File.WriteAllText(highestAvailableVersionFile, highestAvailableVersion.ToString());
+            }
+        }
+
+        public static string CurrentGVFSVersion()
+        {
+            return ProcessHelper.GetCurrentProcessVersion();
         }
 
         public static string GetUpgradesDirectoryPath()
@@ -82,6 +95,11 @@ namespace GVFS.Common
             }
         }
 
+        public static string GetHighestAvailableVersionFilePath()
+        {
+            return Path.Combine(GetUpgradesDirectoryPath(), HighestAvailableVersionFileName);
+        }
+
         private static void RecursiveDelete(string path)
         {
             if (!Directory.Exists(path))
@@ -103,6 +121,17 @@ namespace GVFS.Common
             }
 
             directory.Delete();
+        }
+
+        private static EventMetadata CreateEventMetadata(Exception e)
+        {
+            EventMetadata metadata = new EventMetadata();
+            if (e != null)
+            {
+                metadata.Add("Exception", e.ToString());
+            }
+
+            return metadata;
         }
     }
 }
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
@@ -15,8 +15,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
     [Category(Categories.WindowsOnly)]
     public class UpgradeReminderTests : TestsWithEnlistmentPerFixture
     {
-        private const string GVFSInstallerName = "VFSForGit.1.0.18234.1.exe";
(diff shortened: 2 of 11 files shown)


GVFS/GVFS.Hooks/Program.cs
```diff
@@ -112,7 +112,7 @@ private static void RemindUpgradeAvailable()
             int reminderFrequency = 10;
             int randomValue = random.Next(0, 100);
 
-            if (randomValue <= reminderFrequency && ProductUpgraderInfo.IsLocalUpgradeAvailable(GVFSHooksPlatform.GetInstallerExtension()))
+            if (randomValue <= reminderFrequency && ProductUpgraderInfo.IsLocalUpgradeAvailable(tracer: null))
```
mjcheetham: The hooks don't have a tracer instance?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -23,7 +22,7 @@ public void Start()
         {
             if (!GVFSEnlistment.IsUnattended(this.tracer))
             {
-                TimeSpan startTime = TimeSpan.Zero;
+                TimeSpan startTime = TimeSpan.FromMinutes(1);
```
mjcheetham: What is the reason for this change?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
             {
-                Version newerVersion = null;
-                string detailedError = null;
-                if (!productUpgrader.UpgradeAllowed(out errorMessage))
+                try
                 {
-                    return false;
-                }
+                    // The upgrade check always goes against GitHub
+                    GitHubUpgrader productUpgrader = GitHubUpgrader.Create(
```
mjcheetham: Do we expect the `GitHubUpgrader` instance to be the same each time the timer fires? Could we not keep this instance around for the lifetime of the `ProductUpgraderTimer` rather than each `TimerCallback(..)`?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
-                    // Make sure there a no asset installers remaining in the Downloads directory. This can happen if user
-                    // upgraded by manually downloading and running asset installers.
-                    ProductUpgraderInfo.DeleteAllInstallerDownloads();
-                    errorMessage = null;
-                    return true;
-                }
+                        return;
+                    }
 
-                if (productUpgrader.TryDownloadNewestVersion(out detailedError))
```
mjcheetham: So when do we actually download the packages now? Also what would happen if the version number in the 'latest' file is different from the actual latest version on the server/feed? Do we download the version number as specified in the file or just 'latest'?

(Just thinking what would happen in the case we pulled an update file but a user already did the check and has a 'latest' file pointing to that now revoked version.)

GVFS/GVFS.Hooks/Program.cs
jamill: I will check again - but I did not see any tracers from the hooks. I don't recall us having any tracing from the hooks themselves for individual git commands...

GVFS/GVFS.Service/ProductUpgradeTimer.cs
jamill: Originally - I added a 1 minute delay before GVFS service checked for updates so that I had time to attach the debugger to the service. I tried following the basic tricks to debug service startup, but I was not successful - I suspect I was not getting interactive prompts, even though I had set the bit to allow the service to interact with the desktop....

However, I thought this might not be a bad change in general, as we delaying the initial check by a minute has minimal (no?) downside, and allows us to attach a debugger if needed. It could be a positive to wait a minute so we don't check while starting up.

I can put this back, if people object.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
jamill: > So when do we actually download the packages now?

We download the package when running the Upgrade command.

> Also what would happen if the version number in the 'latest' file is different from the actual latest version on the server/feed? Do we download the version number as specified in the file or just 'latest'?

We will always download the latest. If the `latest` version changes in between the version in the file being updated and running the Upgrade command, we will download and install the new latest. This has been brought up as something we might want to tweak, so that someone can download and install a specific version to avoid race conditions such as this, but we don't have that yet.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
jamill: I suppose the configured ring could change after the service has started. If we went this route, changing the GVFS config would require a service restart to pick up the new values...

GVFS/GVFS.Service/ProductUpgradeTimer.cs
mjcheetham: > We download the package when running the Upgrade command.

But from the changes I see in the PR so far the 'download' code has been removed, not moved. Are you saying the actual call to 'download' is already in the Upgrade command code, and previously we were downloading twice (once for the check, once for the actual upgrade cmd)?

> This has been brought up as something we might want to tweak, so that someone can download and install a specific version to avoid race conditions such as this, but we don't have that yet.

..or at least show a warning/message that "the latest version is now n+1".

GVFS/GVFS.Service/ProductUpgradeTimer.cs
mjcheetham: > If we went this route, changing the GVFS config would require a service restart to pick up the new values...

Right, I suspected this might have been the reason for not doing it. Thanks for confirming! 😃

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -10,39 +10,59 @@ public class ProductUpgraderInfo
...
+                    return false;
+                }
 
-        public static bool IsLocalUpgradeAvailable(string installerExtension)
-        {
-            string downloadDirectory = GetAssetDownloadsPath();
-            if (Directory.Exists(downloadDirectory))
+                Version currentVersion = new Version(CurrentGVFSVersion());
+                return highestAvailableVersion > currentVersion;
+            }
```
alameenshah: This looks a lot cleaner than searching for files matching version numbers.

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -104,5 +129,37 @@ private static void RecursiveDelete(string path)
...
+        {
+            Version highestAvailableVersion = null;
+            string highestAvailableVersionFile = GetHighestAvailableVersionFilePath();
+
+            if (!File.Exists(highestAvailableVersionFile))
+            {
+                return null;
+            }
+
+            string contents = File.ReadAllText(highestAvailableVersionFile).Trim();
```
alameenshah: Nit - Do we need to read the whole file? Should we stop at the first non-empty line. (Not a big issue since the highestAvailableVersionFile is not world writable.)

alameenshah:
Functional tests for Upgrade reminder messaging has to be updated to use the new `highestAvailableVersion` file.

GVFS/GVFS.Hooks/Program.cs
wilbaker: That's correct, the hooks don't have any tracing in them.

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -10,39 +10,59 @@ public class ProductUpgraderInfo
...
+                        CreateEventMetadata(ex),
+                        "Exception encountered when determining if an upgrade is available.");
                 }
             }
 
             return false;
         }
 
-        public static IEnumerable<string> PossibleGVFSInstallerNamePrefixes()
+        public static void RecordHighestAvailableVersion(Version highestAvailableVersion)
```
wilbaker: Use `PhysicalFileSystem` rather than `System.IO` so that we can unit test this method

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -10,39 +10,59 @@ public class ProductUpgraderInfo
...
+                {
+                    return false;
+                }
 
-        public static bool IsLocalUpgradeAvailable(string installerExtension)
-        {
-            string downloadDirectory = GetAssetDownloadsPath();
-            if (Directory.Exists(downloadDirectory))
+                Version currentVersion = new Version(CurrentGVFSVersion());
+                return highestAvailableVersion > currentVersion;
```
wilbaker: We previously weren't comparing the downloaded versions against our own process version.  Would it be enough to just test for the existence of a marker file here?

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -104,5 +129,37 @@ private static void RecursiveDelete(string path)
 
             directory.Delete();
         }
+
+        private static Version ReadHighestRecordedAvailableVersion()
```
wilbaker: Similar comment here, use `PhysicalFileSystem` and add unit tests.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
-                if (!productUpgrader.UpgradeAllowed(out errorMessage))
+                try
                 {
-                    return false;
-                }
+                    // The upgrade check always goes against GitHub
+                    GitHubUpgrader productUpgrader = GitHubUpgrader.Create(
+                        this.tracer,
+                        false,
+                        false,
```
wilbaker: Minor: Add named parameters here so that it's easier for readers to see what parameters are being set to `false`

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
-                if (newerVersion == null)
-                {
-                    // Already up-to-date
-                    // Make sure there a no asset installers remaining in the Downloads directory. This can happen if user
-                    // upgraded by manually downloading and running asset installers.
-                    ProductUpgraderInfo.DeleteAllInstallerDownloads();
-                    errorMessage = null;
-                    return true;
-                }
+                        return;
```
wilbaker: Should we also log something if `errorMessage` is `null`? How would we get in that situation?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
+                    }
 
-                if (productUpgrader.TryDownloadNewestVersion(out detailedError))
-                {
-                    errorMessage = null;
-                    return true;
+                    InstallerPreRunChecker prerunChecker = new InstallerPreRunChecker(this.tracer, string.Empty);
+                    if (!prerunChecker.TryRunPreUpgradeChecks(out _))
+                    {
+                        return;
```
wilbaker: Should we trace the output parameter?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
+                    }
+
+                    if (!productUpgrader.UpgradeAllowed(out errorMessage))
+                    {
+                        if (errorMessage != null)
+                        {
+                            this.tracer.RelatedError(errorMessage);
```
wilbaker: If `UpgradeAllowed` returns `false`, is it always an error?  I am wondering if any of these should actually be `RelatedWarning`.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
+                        return;
+                    }
+
+                    ProductUpgraderInfo.RecordHighestAvailableVersion(newerVersion);
                 }
-                else
+                catch (Exception ex) when (
+                    ex is IOException ||
+                    ex is UnauthorizedAccessException ||
+                    ex is NotSupportedException)
```
wilbaker: Should we be catching all exception types here?  

It looks like if some other exception was thrown the background thread would die but the service would keep running.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
+            errorMessage = null;
+            tracer.RelatedInfo("Querying server for latest version...");
+
+            if (!productUpgrader.TryQueryNewestVersion(out newVersion, out string detailedError))
+            {
+                errorMessage = "Could not fetch new version info. " + detailedError;
+                return false;
+            }
+
+            string logMessage = newVersion is null ? "No newer versions available." : $"Newer version available: {newVersion}.";
```
wilbaker: Minor: I think this came up in another PR as well, but elsewhere in the codebase we're using ` == null` to check if a variable is `null` and we should stay consistent.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +52,101 @@ public void Dispose()
...
+                            productUpgrader,
+                            out Version newerVersion,
+                            out errorMessage))
+                    {
+                        if (errorMessage != null)
+                        {
+                            this.tracer.RelatedError(errorMessage);
+                        }
+
+                        return;
```
wilbaker: In this case (we're on the latest version) should the `HighestAvailableVersion` file be deleted?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
wilbaker: I don't see a problem with waiting a minute.

wilbaker:
@jamill I just noticed this PR is into master, should it be into the feature branch so that we can test all of the upgrader changes together?

GVFS/GVFS.Common/ProductUpgraderInfo.cs
jamill: Probably - the benefit of reading the version here is we can confirm that we are out-of-date.

The window of when this file would exist, but the user is actually up-to-date is pretty small - in particular it could happen when a user manually installs a new version (not via upgrade), but the background process has not cleared the file yet.

That brings up a point I missed - upgrade should clear this file when it runs... Although as is, the user would not see incorrect results due to this check.

jamill:
> @jamill I just noticed this PR is into master

That was intentional - but I should have called it out. As this change stands on its own, I targeted master. I don't mind targeting the feature branch, but my thought was this keeps the feature branch smaller... On the flip side, it could introduce churn in the feature branch as it will need to absorb these changes, but I felt this was the easier path overall.

I can target either branch if people prefer otherwise.

GVFS/GVFS.Common/ProductUpgraderInfo.cs
mjcheetham: Would bringing in and initialising these dependencies increase the spin-up time for the hooks?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
jamill: I made most of these RelatedWarning

GVFS/GVFS.Common/ProductUpgraderInfo.cs
jamill: Good point - as the file is controlled by us, I didn't stop at the 1st line. If someone is tampering with the file, then we have bigger problems... @wilbaker suggested to not even read the contents of the file, which could also work...

GVFS/GVFS.Service/ProductUpgradeTimer.cs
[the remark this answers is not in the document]
jamill: I could - something like `TryCreateUpgraderForAnonymousPolling`?

GVFS/GVFS.Common/ProductUpgraderInfo.cs
wilbaker: If we decide to continue to read the contents of the file I think it's okay to read the entire file.  There should only be one line, and if we read in unexpected data it seems better to have `Version.TryParse` fail than silently ignore the unexpected extra data.

GVFS/GVFS.Common/ProductUpgraderInfo.cs
wilbaker: > Probably - the benefit of reading the version here is we can confirm that we are out-of-date.

That makes sense.  I wasn't sure how much extra overhead we're adding to the hooks with  open+read+parse vs a file existence check.

GVFS/GVFS.Common/ProductUpgraderInfo.cs
wilbaker: Rather than having hooks consume `ProductUpgraderInfo.cs` directly, we could move the method(s) the hooks need to a new `ProductUpgraderInfo.Shared.cs`, and then methods like this one could use `PhysicalFileSystem` without having to pull `PhysicalFileSystem` into the hooks project.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +54,102 @@ public void Dispose()
...
+            string logMessage = newVersion == null ? "No newer versions available." : $"Newer version available: {newVersion}.";
+            tracer.RelatedInfo(logMessage);
+
+            return true;
+        }
+
+        private bool TraceException(Exception ex)
+        {
+            EventMetadata metadata = this.CreateEventMetadata(ex);
+            this.tracer.RelatedError(metadata, message);
```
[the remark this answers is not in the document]
wilbaker: It looks like this comment crossed with your `fixup` commit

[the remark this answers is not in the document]
jamill: Yeah - unfortunately this code is not compiled on non windows platforms at the moment.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -77,15 +76,12 @@ private void EmptyDownloadDirectory()
...
-            string gitInstallerPath = Path.Combine(this.upgradeDirectory, GitInstallerName);
+            string gvfsUpgradeAvailableFilePath = Path.Combine(this.upgradeDirectory, HighestAvailableVersionFileName);
 
             this.EmptyDownloadDirectory();
 
-            this.fileSystem.CreateEmptyFile(gvfsInstallerPath);
-            this.fileSystem.CreateEmptyFile(gitInstallerPath);
-            this.fileSystem.FileExists(gvfsInstallerPath).ShouldBeTrue();
-            this.fileSystem.FileExists(gitInstallerPath).ShouldBeTrue();
+            this.fileSystem.CreateEmptyFile(gvfsUpgradeAvailableFilePath);
```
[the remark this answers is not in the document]
jamill: We do not have any code that reads in the contents of this file, only that it exists, so I was treating this as an implementation detail. If / when it becomes important, we could add a test then. We could alternatively just write a blank file...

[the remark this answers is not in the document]
wilbaker: > We could alternatively just write a blank file...

I think there's value in writing out the version from a support perspective.

> We do not have any code that reads in the contents of this file, only that it exists

That's a good point, we can always revisit this if we decide to start reading the value.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -53,70 +54,102 @@ public void Dispose()
...
+                    ex is NotSupportedException)
                 {
-                    errorMessage = null;
-                    return true;
+                    this.tracer.RelatedWarning(
+                        CreateEventMetadata(ex),
+                        "Exception encountered recording highest available version");
                 }
-                else
+                catch (Exception ex) when (this.TraceException(ex))
```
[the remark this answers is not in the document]
jamill: > Is there any reason we can't do that here?

We can do that here - I think the approach has several drawbacks which can make it harder to debug, but we can follow up another time. I will update the code to be consistent.

GVFS/GVFS.UnitTests/Common/ProductUpgraderInfoTests.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.mockFileSystem = null;
+            this.productUpgraderInfo = null;
+            this.tracer = null;
+        }
+
+        [TestCase]
+        public void RecordHighestVersion()
+        {
+            this.productUpgraderInfo.RecordHighestAvailableVersion(new Version("1.0.0.0"));
+            this.mockFileSystem.Verify(fileSystem => fileSystem.WriteAllText(this.expectedNewVersionExistsFilePath, It.IsAny<string>()), Times.Once());
```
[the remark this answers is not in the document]
jamill: I think this is a good idea - however, as we do not read the version from this file, I will put this on my list of items to follow up on, rather than include it in this PR.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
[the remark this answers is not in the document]
jamill: I like this suggestion - but was going to hold off on implementing it in this PR. We can come back after we have the different streams of work merged, to reduce churn / conflicts.

GVFS/GVFS.UnitTests/Common/ProductUpgraderInfoTests.cs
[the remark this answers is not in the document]
wilbaker: > I will put this on my list of items to follow up on, rather than include it in this PR.

Could you create a GitHub issue that has planned follow up work items?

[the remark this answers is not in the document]
wilbaker: You could also consider renaming the test for now to something like `CreatesHighestVersionFile` as the test is not validating that the highest version was recorded.

Issue #563, opened 2018-11-29
Mac: Write scripts\guide for capturing file system events when problems occur


wilbaker:
On Windows we can ask users to run [procmon](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon) to capture the I/O that's happening on their system when errors occur.

It sounds like `dtrace` *might* be able to get us something similar on Mac, but work needs to be done to determine how we should configure\run dtrace to get the information we're interested in.

The issue encompasses:

- Researching `dtrace` and determining what useful information can be captured and how to capture it
- Creating script(s) to start\stop `dtrace` with the options we need
- Writing up a guide on how to collect\analyze data from `dtrace`

pmj:
After out-of-band discussion, I understand the scope of this task to be:

 * Find out what useful information we can get with dtrace.
 * Check out the extent to which this is limited by SIP, and whether we can get good enough data for most purposes while keeping SIP enabled.
 * Potentially add comprehensive tracing to any actions taken by the kext. (e.g. log every DENY kauth vnode listener response to be able to trace errors back to our kext - or not.)
 * If dtrace with SIP isn't good enough, check the feasibility of building a special purpose tracing kext.
 * Write documentation & scripts to allow end users to enable/run such diagnostic traces.

wilbaker:
>  I understand the scope of this task to be:

@pmj that looks correct to me, thanks!

/ cc: @jrbriggs

wilbaker:
@pmj, I think there's actually one more item to call out as being within the scope of this task:

- If the amount of useful information we can get from dtrace depends on whether SIP is enabled\disabled, write scripts that can take advantage of the extra information available when SIP is disabled.  This can be either a separate script, or as an option for a single script (that can run in both SIP modes).  This will be helpful for diagnosing issues that we can reproduce on machines where SIP is disabled.

/ cc: @jrbriggs, @nickgra

pmj:
With #637 and #643 finally merged and with me back and refreshed(?) from my holiday break, I can now focus on this.

My initial plan was to attempt to write a DTrace script that will log any (failed?) file I/O syscalls from passed command's process and its subprocesses.

[...]

PR #737, opened 2019-02-04
Mac syscall tracing: Initial updates to dtruss and some basic documentation

pmj:
This is a first pass at tackling issue #563. At a bare minimum, it should allow us to collect a syscall trace of e.g. a failing build on the affected machine, which we can manually pick through to find out what's going through.

Stock `dtruss` wasn't really up to that job due to lack of special handling for many quirks of (modern versions of?) macOS. So the majority of this PR is bugfixing and adding features in `dtruss`.

 * When launching a command with dtruss, it's now possible to run the command as a non-root user.
 * Correct logging of execve() and posix_spawn() syscalls.
 * Improvements to output for syscalls with buffer or string parameters. (Better handling of NULL pointers, large buffers, etc.)
 * A new -F option for filtering out common but typically uninteresting syscalls. At the moment, this only filters out close() calls returning EBADF. Some programs seem to call close() on thousands of sequential integers, most of which are not valid file descriptors.
 * The mode for following child processes (-f) is much more reliable. There were so many bugs in the previous implementation of this it effectively didn't work at all most of the time.
 * I've added a short readme-style document for getting started with *using* `dtruss`.

The changes are largely contained and explained in individual commits as usual; I realise none of you have probably seen DTrace scripts before but it should be fairly readable to anyone familiar with C. Knowing about some of its oddities will make understanding easier:
 * There's no control flow apart from the `/ … /` conditional block at the start of each probe action, and the `?:` conditional operator. No if/else, no loops.
 * Variables are implicitly strongly typed based on the first assignment that occurs in the source code.
 * `self->foo` variables are thread-local, `this->bar` variables are local to action blocks.
 * Pointers can be in kernel or user space depending on context. User space pointers must be dereferenced with `copyin`/`copyinstr`.
Feel free to ask questions if you're curious about anything in particular. The macOS `dtrace` man page is a reference you might want to consult too. (`man 1 dtrace`)

Probably more important than reviewing the intricacies of the script's code: actually trying out the script to see what limitations and bugs still need to be fixed.

Not yet done:
 1. Limiting the tracing scope to file I/O syscalls only. For some programs this means we generate a lot of noise that's probably not useful for VFS4G diagnostics; plus it increases the likelihood of a trace buffer overflow.
 2. A trace is just a long list of barely human readable raw data. You have to keep track of which process is which by its PID yourself, likewise for file descriptors and such. There's no filtering beyond `grep` and friends, etc.
 3. There's no way to provide multiple PIDs or process names to watch; you can only trace one syscall or all of them. This might not be necessary in practice if follow-mode works well enough, and if we create predefined, hardcoded groups of syscalls such as the file I/O one.

Especially for item 2, I'd welcome discussion in #563 about what shape the post-processing might take.

### DIFF
--- a/Scripts/Mac/Tracing/Docs/cddl1.txt
+++ b/Scripts/Mac/Tracing/Docs/cddl1.txt
@@ -1,0 +1,385 @@
+
+COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL) Version 1.0
+
+
+      1. Definitions.
+
+            1.1. “Contributor” means each individual or entity that
+            creates or contributes to the creation of Modifications.
+
+            1.2. “Contributor Version” means the combination of the
+            Original Software, prior Modifications used by a
+            Contributor (if any), and the Modifications made by that
+            particular Contributor.
+
+            1.3. “Covered Software” means (a) the Original Software, or
+            (b) Modifications, or (c) the combination of files
+            containing Original Software with files containing
+            Modifications, in each case including portions thereof.
+
+            1.4. “Executable” means the Covered Software in any form
+            other than Source Code. 
+
+            1.5. “Initial Developer” means the individual or entity
+            that first makes Original Software available under this
+            License. 
+            
+            1.6. “Larger Work” means a work which combines Covered
+            Software or portions thereof with code not governed by the
+            terms of this License.
+
+            1.7. “License” means this document.
+
+            1.8. “Licensable” means having the right to grant, to the
+            maximum extent possible, whether at the time of the initial
+            grant or subsequently acquired, any and all of the rights
+            conveyed herein.
+            
+            1.9. “Modifications” means the Source Code and Executable
+            form of any of the following: 
+
+                  A. Any file that results from an addition to,
+                  deletion from or modification of the contents of a
+                  file containing Original Software or previous
+                  Modifications; 
+
+                  B. Any new file that contains any part of the
+                  Original Software or previous Modification; or 
+
+                  C. Any new file that is contributed or otherwise made
+                  available under the terms of this License.
+
+            1.10. “Original Software” means the Source Code and
+            Executable form of computer software code that is
+            originally released under this License. 
+
+            1.11. “Patent Claims” means any patent claim(s), now owned
+            or hereafter acquired, including without limitation,
+            method, process, and apparatus claims, in any patent
+            Licensable by grantor. 
+
+            1.12. “Source Code” means (a) the common form of computer
+            software code in which modifications are made and (b)
+            associated documentation included in or with such code.
+
+            1.13. “You” (or “Your”) means an individual or a legal
+            entity exercising rights under, and complying with all of
+            the terms of, this License. For legal entities, “You”
+            includes any entity which controls, is controlled by, or is
+            under common control with You. For purposes of this
+            definition, “control” means (a) the power, direct or
+            indirect, to cause the direction or management of such
+            entity, whether by contract or otherwise, or (b) ownership
+            of more than fifty percent (50%) of the outstanding shares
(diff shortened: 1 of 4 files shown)


jeschu1:
We've already used this script to identify an issue.  Nice job @pmj!

Scripts/Mac/Tracing/Docs/cddl1.txt
```diff
@@ -0,0 +1,385 @@
+
```
jeschu1: I'm not sure about licensing.  How did we arrive at this license?

Scripts/Mac/Tracing/Tracing.md
```diff
@@ -0,0 +1,78 @@
...
+For diagnosing issues potentially caused by VFS for Git, tracing syscalls to track down errors has proven to be a useful technique on Windows.
+On macOS, the DTrace script `dtruss` which ships with the OS can be used for tracing system calls made by processes.
+However, it has not been updated for newer OS versions for a while, making some of its features unreliable.
+We therefore recommend using our [updated version of the script](./dtruss), and the instructions below assume you are using this version.
+
+## TL;DR: This build command is failing/misbehaving and I've been asked to collect a trace
+
+To capture a trace of the syscalls made by the failing command, run the following:
+
+    ++path-to-vfs4g-source++/Scripts/Mac/Tracing/dtruss -d -e -F -f    ++command-to-run++   2> >(tee   ++trace-filename.txt++   >&2)
```
jeschu1: I used "2>&1 | tee log" as an alternative

Scripts/Mac/Tracing/dtruss
```diff
@@ -0,0 +1,1271 @@
+#!/bin/sh
+# #!/usr/bin/sh
+#
+# dtruss - print process system call time details.
+#          Written using DTrace (Solaris 10 3/05).
```
jeschu1: Where is this coming from?

Scripts/Mac/Tracing/Docs/cddl1.txt
pmj: Note the commit message - https://github.com/Microsoft/VFSForGit/pull/737/commits/a5176525b4b0a84e686a8df5a49ddb4137fbc861

`dtruss` itself originates in the DTrace toolkit scripts in Solaris, at the time owned by Sun, which in turn was later bought by Oracle. Dtrace and the various toolkit scripts including `dtruss` were ported to what was then known as Mac OS X by Apple around 2006-2007, and included in the Mac OS X 10.5 "Leopard" release in 2007. Sun released DTrace along with most of their Open Source code under the CDDL.

The version I'm working from is the one from Apple's dtrace-285.200.15 release, located at https://opensource.apple.com/source/dtrace/dtrace-284.200.15/DTTk/dtruss.auto.html and also included in recent versions of macOS at `/usr/bin/dtruss`.

I included the license text in its original location of Docs/cddl1.txt as that's also the location referenced in the dtruss source comments.

Scripts/Mac/Tracing/dtruss
pmj: See my comment on licensing https://github.com/Microsoft/VFSForGit/pull/737/commits/a5176525b4b0a84e686a8df5a49ddb4137fbc861#r254600223

Scripts/Mac/Tracing/Tracing.md
pmj: Note that while simpler, this also includes stdout in the log file, whereas the code I quoted only saves stderr to the file. This may help keep the output from the command itself separate from the trace data.

jeschu1:
@pmj, @jrbriggs and I chatted offline and I think we should take this PR as is since it's already successfully identified an issue.  Over time we can refine and make it more robust as needed.

PR #738, opened 2019-02-04
Use package feed URL to query for credentials

jamill:
Instead of explicitly providing the URL used to query Git Credential Helper for credentials, transform the URL used for the packaging feed to construct the URL to query Git Credential Helper with.

### DIFF
--- a/GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
@@ -121,9 +121,15 @@ namespace GVFS.Common.NuGetUpgrader
                 return false;
             }
 
+            string authUrl;
+            if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(upgraderConfig.FeedUrl, out authUrl, out error))
+            {
+                return false;
+            }
+
             if (!TryGetPersonalAccessToken(
                     GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath(),
-                    upgraderConfig.FeedUrlForCredentials,
+                    authUrl,
                     tracer,
                     out string token,
                     out string getPatError))
--- a/GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -144,6 +151,30 @@ public class NuGetUpgrader : ProductUpgrader
...
+            // and want to convert it to a URL of the form https://dev.azure.com/{org}
+            Regex packageUrlRegex = new Regex(
+                @"^https://pkgs.dev.azure.com/(?<org>.+?)/",
+                RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
+            Match urlMatch = packageUrlRegex.Match(packageFeedUrl);
+
+            if (!urlMatch.Success)
+            {
+                azureDevOpsUrl = null;
+                error = $"Input URL {packageFeedUrl} did not match expected format for an Azure Devops Package Feed URL";
```
wilbaker: Nit: `Devops` -> `DevOps`

GVFS/GVFS.Common/NuGetUpgrader/NuGetUpgrader.cs
```diff
@@ -144,6 +151,30 @@ public class NuGetUpgrader : ProductUpgrader
...
+                return false;
+            }
+
+            string org = urlMatch.Groups["org"].Value;
+
+            azureDevOpsUrl = urlMatch.Result($"https://dev.azure.com/{org}");
```
jamill: I tried this out on my actual machine, and it seems that the credential manager does not like this format. I can use `https://{org}.visualstudio.com`, but not `https://dev.azure.com/{org}`. I will take a closer look at what might be going on here.

/cc @mjcheetha, @jrbriggs m in case I am missing something obvious with URLs and GCM

mjcheetham: What do you mean by "does not like this format"? Are you seeing an error? Is this Windows GCM?

For Windows GCM we don't expect/work with the 'useHttpPath' option, but instead do that process walking logic to find the git-remote-helper/http process, which VFS for Git won't have as part of the process tree.

mjcheetham: You can use `https://{org}@dev.azure.com/{org}` for maximum compat.

jamill: ah - yeah, that is it. I forgot that was how the Windows GCM worked.

Issue #739, opened 2019-02-05
Enable the `TelemetryDaemonEventListener` on Windows


mjcheetham:
We added but disabled the new daemon-based telemetry event listener. We should enable this once the Mac version is also supported, and when we're happy with its stability.

PR #740, opened 2019-02-05
Disable the `TelemetryDaemonEventListener` on Windows

mjcheetham:
Disable the new daemon-based telemetry event listener on Windows until we have the Mac version working and that we're happy with the stability of the listener.

The work to re-enable this is tracked in #739

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -79,17 +79,20 @@ namespace GVFS.Platform.Windows
         {
             string gitBinRoot = this.GitInstallation.GetInstalledGitBinPath();
 
-            var etwListener = ETWTelemetryEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
+            ETWTelemetryEventListener etwListener = ETWTelemetryEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
             if (etwListener != null)
             {
                 yield return etwListener;
             }
 
-            var daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, pipeName: "vfs");
-            if (daemonListener != null)
-            {
-                yield return daemonListener;
-            }
+            // TODO: enable the daemon-based telemetry listener once we're happy.
+            // See GitHub issue: https://github.com/Microsoft/VFSForGit/issues/739
+            //
+            // TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, pipeName: "vfs");
+            // if (daemonListener != null)
+            // {
+            //     yield return daemonListener;
+            // }
         }
 
         public override void InitializeEnlistmentACLs(string enlistmentPath)


GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -85,6 +85,10 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
                 yield return etwListener;
             }
 
+            // TODO: enable the daemon-based telemetry listener once we're happy.
+            // See GitHub issue: https://github.com/Microsoft/VFSForGit/issues/739
+            yield break;
+
```
wilbaker: While you're in here could you swap out the `var` on line 82 and 92 with an explicit type?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -85,6 +85,10 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
                 yield return etwListener;
             }
 
+            // TODO: enable the daemon-based telemetry listener once we're happy.
+            // See GitHub issue: https://github.com/Microsoft/VFSForGit/issues/739
+            yield break;
```
wilbaker: Out of curiosity, what's the reason for using the `yield` syntax in this method?  

We know that callers of `CreateTelemetryListeners` will always want all of the listeners could we just return them in a `List`?

mjcheetham: Personally my brain is wired up to write the iterator pattern when an `IEnumerable<T>` method returns a variable number of elements. I think it's more readable and removes the collection/container boiler plate (admittedly `yield` involves code behind the scenes).

We could, but then we could also argue that we always know the exact type of everything in VFS4G and should never be using any of the collection interfaces like IEnumerable or ICollection. We could probably get away with using arrays most of the time also. I'd say stick with the nicer syntax unless it becomes a burden in a perf sense or a readability sense?.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -79,17 +79,20 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
...
-                yield return daemonListener;
-            }
+            // TODO: enable the daemon-based telemetry listener once we're happy.
+            // See GitHub issue: https://github.com/Microsoft/VFSForGit/issues/739
+            //
+            // TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, pipeName: "vfs");
+            // if (daemonListener != null)
+            // {
+            //     yield return daemonListener;
+            // }
```
kewillford: Just curious if StyleCop complained about the `//` without a space and if you had to add the space to each line. I think it does the last time I remember and what I usually do is add `////` by using Ctrl+E, C twice.

mjcheetham: Yes, StyleCop doens't like `//foo`, but wants `// foo`.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
wilbaker: > We could, but then we could also argue that we always know the exact type of everything in VFS4G and should never be using any of the collection interfaces like IEnumerable or ICollection. 

For the most part we don't use the collection interfaces in VFS4G.  The codebase is not meant to be an API for public consumption, and so I'm not sure we're gaining much by using interfaces like `ICollection` (unless we have a specific need for them).

> I think it's more readable and removes the collection/container boiler plate (admittedly `yield` involves code behind the scenes).

I think a lot of this comes down to personal preference.  I personally find `yield` more difficult to read\debug, and when I see it I assume there's a situation where there's a very large list (that we don't want to return all at once), the caller might not care about the whole list, or it's very computationally expensive (see [Proper use of 'yield return'](https://stackoverflow.com/a/410058) discussion on Stack Overflow).

> We could probably get away with using arrays most of the time also. I'd say stick with the nicer syntax unless it becomes a burden in a perf sense or a readability sense?

I guess I'm not convinced it's a nicer syntax 😄 

Returning an `IEnumerable` when we could return something like a `List` (or an `ICollection`) reduces the caller's ability to inspect the results (e.g. there is no `Count`) and encourages the use of Linq (whose performance is worse than using the properties of a more specific collection type).

The other concern I have using this syntax when it's not required is that it's inconsistent with the rest of the codebase.  I think that unless there's a strong argument for breaking with the established style we should side with consistency.

kewillford: >Returning an `IEnumerable` when we could return something like a `List` (or an `ICollection`) reduces the caller's ability to inspect the results (e.g. there is no `Count`)

I think this is the main argument for using it because if you don't want your caller to be able to change the results, you can use it to enforce this and prevent bugs by limiting what can be done.
But I don't think this is about the return type because I can have a `List<T>` internally and still return an `IEnumerable`.  

There are some areas of the codebase where we are using the `yield` syntax like the `FileBased..` classes and derived classes and the `PhysicalFileSystem` so it isn't inconsistent with the rest of the codebase necessarily.

For me it is just code so it would come down to when should `yield` syntax be used vs a `List` and that seems fuzzy and we don't have any solid guidelines for.  For example what is a very large list?  1000 items? 1,000,000 items?  So I think we need to come up with those guidelines and make sure we all agree and know what they are so we are using according to the codebase standards.

wilbaker: > There are some areas of the codebase where we are using the `yield` syntax like the `FileBased..` classes and derived classes and the `PhysicalFileSystem` so it isn't inconsistent with the rest of the codebase necessarily.

Good point, I did not realize there were already several spots we're using `yield`.

> I think this is the main argument for using it because if you don't want your caller to be able to change the results, you can use it to enforce this and prevent bugs by limiting what can be done.

`IEnumerable` doesn't protect against modifying the individual items though, correct?  It would only prevent things like inserting\deleting items into whatever collection is returned.

I see the value in using this approach when we're returning a reference to a class's member variable, as we don't want the caller to inadvertently grow\shrink the member collection.  In this case, however, we'd be generating a new collection and so there'd be no harm in the caller modifying the collection.  (And if we *really* wanted to prevent modification something like an `IReadOnlyList` would provide that level of protection and the additional functionality that comes from having a list).

> But I don't think this is about the return type because I can have a `List<T>` internally and still return an `IEnumerable`.

Yep you're right, what prompted the discussion was the usage of `yield` rather than the return type.

> For example what is a very large list? 1000 items? 1,000,000 items? 

That's a great point.  Historically when we've done work to reduce VFS4G's memory usage the threshold has been that the collection in question has a noticeable impact on VFS4G's overall memory consumption.  (I realize that's still a fuzzy answer)

> So I think we need to come up with those guidelines and make sure we all agree and know what they are so we are using according to the codebase standards.

That sounds good to me.  I think this discussion highlights that it's not clear what approach we should default when writing methods like this.

wilbaker: >> So I think we need to come up with those guidelines and make sure we all agree and know what they are so we are using according to the codebase standards.

>That sounds good to me. I think this discussion highlights that it's not clear what approach we should default when writing methods like this.

+ @jrbriggs for his thoughts on this as well

wilbaker: To close things out for this PR:  I think it's worth following up on this topic offline, but I don't think there's any need to block this PR on the discussion.

Issue #338, opened 2018-10-04
Mac: Investigate using fsid & inode to get paths in user-mode rather than vn_getpath


wilbaker:
`vn_getpath` can be unreliable.  Investigate if the user-mode lib can use `fsgetpath` to get the path instead (using fsid & inode).

pmj:
One thing I just remembered about inodes on HFS+ (or CNIDs as they're called there) is that each hard link has its own ID, along with an ID for the link target. I suspect the same is true for APFS for compatibility reasons. For purposes of getting a path, the link's ID is relevant, for purposes of checking a file's identity (and e.g. serialising access), the target ID is probably more useful.

pmj:
@wilbaker did I understand correctly from the conf call that you'd like me to give this a go for working around the `vn_getpath` and `vnode_getparent` failures?

wilbaker:
> @wilbaker did I understand correctly from the conf call that you'd like me to give this a go for working around the vn_getpath and vnode_getparent failures?

@pmj yep that would be great! I'll assign the issue over to you.  

Based on our discussion this morning it sounds like #563 and #396 are higher priority than this issue.

/ cc: @jrbriggs

pmj:
**Note to self:** We have largely eradicated `vn_getpath()` with #741, at least on any critical paths. Implementing #846 will confine it entirely to debugging/logging code. The fsid/inode technique still needs to be applied when `vnode_getparent()` calls fail, however, so I'm leaving this issue open until that's done.

Issue #648, opened 2018-12-20
Mac: Generate PrjFSKextVersion as part of the build


wilbaker:
To ensure that the the ProjFS kernel and user-mode components are always in sync, `PrjFSKextVersion` should be generated as part of the build.

Issue #337, opened 2018-10-04
Mac: Add both source and target path to ModifiedPaths.dat when hardlinks are created


wilbaker:
Currently the following scenario will not be handled correctly:

1. Create hardlink to partial file
2. Modify hardlink
 
`git status` will report the hardlink path as modified, but the path of the original partial file should be reported as well.

**UPDATE**

We may also want to convert all involved files to full when this happens.

wilbaker:
See also this discussion:

https://github.com/Microsoft/VFSForGit/pull/741#discussion_r260455605

> Seeing this change in your code got me wondering if there was something different I should be doing with hardlinks and the vnode cache.

Nothing fully-baked, but some thoughts that might give you some further ideas:
* Hard-linked files are easy to detect using the `va_nlink` attribute. If that value is 1, you have nothing to worry about. If it's not 1, `vnode_getparent()` will give you *a* parent, possibly not the one you want. `fsgetpath()` might give you the path inside the provider's root, or it might not.
* 3 central problems I can see with a multi-hardlinked file:
   1. *Identifying the root(s) where it's a member.* Even with the vnode cache, this is tricky as we don't have a way of iterating over all the links to the file, and vnodes get recycled. Perhaps we want to store a list of UUIDs of virtualisation roots in an xattr? With the vnode cache, we can at least cache this data so we don't take the hit for reading xattrs all the time.
   2. *How to handle various file operations in the case where it's a member of more than 1 root.* Is there any conceivable way such a file could sensibly be anything other than a "full" file? If not, we could *always* convert to full when a hardlink to outside the repo is made, and prevent hardlinking non-full files in offline roots. Unloading the prjfs kext still lets you mess with this of course but if a user does that they shouldn't expect things to run 100% smoothly.
   3. *In the provider, what to do when `fsgetpath()` (or a fileop handler arg) returns a path outside the root.* The best approach I can currently think of here is to track the inode numbers of all "full" files and look up inodes for multi-linked files there before using `fsgetpath()`.

PR #741, opened 2019-02-05
Mac PrjFS kext/lib: Use fsid/inode, not vn_getpath()

### LINKED ISSUE #271: Mac: vn_getpath failures result in HandleVnodeOperation denying operations
When `vn_getpath` fails (in `TrySendRequestAndWaitForResponse`) the kext will deny the vnode operation.

We've seen this cause unexpected git command failures:

```
2018-09-13T16:25:46.7927170Z => GVFS.FunctionalTests.Tests.GitCommands.GitCommandsTests.ChangeBranchAndRebaseOnAnotherBranch()
2018-09-13T16:25:46.7944470Z Test GVFS.FunctionalTests.Tests.GitCommands.GitCommandsTests.ChangeBranchAndRebaseOnAnotherBranch()
2018-09-13T16:25:46.7961340Z Failed at 4:25:46 PM taking 00:00:10.5196640
2018-09-13T16:25:46.7985500Z commit -m "Create for ChangeBranchAndRebaseOnAnotherBranch first branch" Errors Lines
2018-09-13T16:25:46.8017560Z commit -m "Create for ChangeBranchAndRebaseOnAnotherBranch first branch" Errors Lines counts do not match. was: 1 expected: 0
2018-09-13T16:25:46.8049320Z Extra: warning: unable to unlink '.git/objects/af/tmp_obj_E0B6hb': Permission denied
```

Because the `vn_getpath` for the tmp file that git it attempting to delete is being denied.

Wherever possible we need to:
- Eliminate our use of `vn_getpath` (e.g. through the use of notification masks)
- Anywhere that `vn_getpath` is used, ensure that we have the proper retry/fallback logic

pmj:
This is ~~a first pass at solving~~ resolves #271 aka #338.

**New approach:**

The key changes are:

 * On the user space side, if no path if received with a message from the kext, `fsgetpath()` is used to look it up from the FSID/inode pair.
 * On the kext side, this allows us to entirely remove the `vn_getpath()` calls from the kauth vnode listener, so no workarounds are needed for when that call fails.
 * The only other times `vn_getpath()` was called in the context of a kauth listener (other than in the KextLog macros, where failure is acceptable) was when discovering a previously unknown virtualisation root. By definition, this only is the case for roots with no active provider, and we only use the root's path to generate relative paths to send to the provider. So we don't actually need to get the path for offline roots, and we already insist on `vn_getpath()` succeeding when registering a provider. The path can be handy to have for offline roots for diagnostic purposes, so I've kept that code around for debug builds.
 * Use link inodes, not file inodes. This should preserve in-root file paths for files with multiple hard links.
 * Cleaned up some defunct message codes.

Removing the `vn_getpath()` call from every vnode listener callback should also improve performance, in case anyone wants to benchmark that on a Very Large Build™.

Possible further enhancements (discuss!):
 * For the remaining path-based kernel-user messages (in the Fileop handler), we're computing the relative path based on the root's path. This is then typically turned back into an absolute path in the provider. Instead, we could simply send the absolute path from the kext into the provider and generate the relative path where needed, or directly use the absolute path where that's the most convenient.
 * If we get rid of the relative path calculation, we can completely get rid of the path field from the virtualisation root struct.

**Previous attempt:**

(Code preserved at https://github.com/pmj/VFSForGit/tree/mac-inode-path-old )

~~This implements a drop-in replacement for vn_getpath that takes a roundtrip to user space. A new kernel-user message is added for querying the path of a file based on fsid+inode. The provider looks up the path using `fsgetpath()` and registers this path with the response to the message. The message is sent from the vnode kauth listener handler when vn_getpath fails, and a response is awaited.~~

~~It works fine - check the commit messages for each commit for some extra details. Each commit also implements some standalone functionality as usual.~~

~~I've also pushed a testing-only commit that deliberately causes vn_getpath to appear to fail about once in 16 times. This helps exercise the new code path.~~

~~However, there are some downsides to this approach:~~
 * ~~It causes extra roundtrips to user space. (Although `vn_getpath()` is now called much less frequently than before, and doesn't fail that often anyway, so it's probably an overall performance win.)~~
 * ~~What do we do when the handling provider is unreliable? There is essentially zero barrier to entry for a process registering as a PrjFS provider, so what do we do when a provider replies with incorrect paths?~~
 * ~~Handling races with renamed/moved/deleted vnodes is tricky, which makes detecting unreliable/untrustworthy providers more difficult.~~

~~So I think in practice I'll try to remove all uses of vn_getpath() in kauth handlers altogether, and instead of sending (relative) file paths with provider requests/notifications, instead rely entirely on fsid/inode and do all the path handling in the provider.~~

~~That still leaves one or two other ("legitimate") uses of `vn_getpath`, which I'll need to address.~~


ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -214,14 +214,21 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode,
         {
             // TODO: check xattr contents
             
-            char path[PrjFSMaxPath] = "";
-            int pathLength = sizeof(path);
-            errno_t error = vn_getpath(vnode, path, &pathLength);
+            const char* path = nullptr;
+#if DEBUG // Offline roots shouldn't need their path filled, and vn_getpath() may fail anyway.
```
nickgra: Why is this code in a Debug block now? If we don't need to fill paths for offline roots, can we just delete this code instead of only running it in Debug?

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -8,13 +8,9 @@
...
     MessageType_Invalid = 0,
-    
-    // Messages from user mode to kernel
-    MessageType_UtoK_StartVirtualizationInstance,
-    MessageType_UtoK_StopVirtualizationInstance,
-    
+	
     // Messages from kernel to user mode
-    MessageType_KtoU_EnumerateDirectory,
+    MessageType_KtoU_EnumerateDirectory = 3,
```
nickgra: Is there any reason why we need to set this to 3?

pmj: We don't have a proper API versioning story yet (there's a version check, but we're not incrementing the version), so not renumbering messages seemed a lower risk strategy, although I suppose as we're no longer guaranteeing paths on some messages that will only ensure backwards compatibility one way (new userlib with older kext) not both ways.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: I've found it helpful for diagnostics to be able to print out information relating to the root when doing development/debugging. Otherwise you've only got an index to go on, and an unretained vnode pointer. (which is not safe to pass to `KextLog_File*()`)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -960,24 +937,22 @@ static bool TrySendRequestAndWaitForResponse(
...
+    const char* relativePath = nullptr;
     
     OutstandingMessage message =
     {
         .receivedResult = false,
         .rootHandle = root,
     };
     
-    if (nullptr == vnodePath)
+    if (nullptr != vnodePath)
```
jeschu1: Should we still break if vnodePath == null?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1345,3 +1378,27 @@ static void ReturnFileMutexIterator(FileMutexMap::iterator lockIterator)
         s_fileLocks.erase(lockIterator);
     }
 }
+
+static const char* GetRelativePath(const char* path, const char* root)
```
jeschu1: This seems like function prime for unit testing

jeschu1: Also is this a copy of the implementation in VirtualizationRoots.cpp?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: No, this is now explicitly a supported situation. If no path to the vnode is available, that's fine, we'll just need to rely on the fsid/inode. See the `assert()` a few lines up.

jeschu1: @pmj that makes sense thanks!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: > Also is this a copy of the implementation in VirtualizationRoots.cpp?

Yes, aside from the log statements. With a bit of effort we can actually share that code between kext and userlib, but as mentioned in my introductory comment on the PR under possible further enhancements, I think we should get rid of the kernel side version altogether. We now don't always send a path from kernel to user space, and when we don't, user space generates an absolute path via the system API `fsgetpath()` which it then converts to a relative one. So I'd argue that we're best off sending the absolute path, when available, from the kext too and leave the relative path calculation to the user space lib.

> This seems like function prime for unit testing.

Definitely. So far I've been ~~too lazy for~~ holding off on adding some as it means creating a new target for testing userlib code etc. and I wanted to get the PR into a roughly-final shape first.

ProjFS.Mac/PrjFSKext/VnodeUtilities.cpp
```diff
@@ -10,8 +10,8 @@ FsidInode Vnode_GetFsidAndInode(vnode_t vnode, vfs_context_t context)
 {
     vnode_attr attrs;
     VATTR_INIT(&attrs);
-    // TODO: check this is correct for hardlinked files
-    VATTR_WANTED(&attrs, va_fileid);
+    // va_linkid is unique per hard link, va_fileid only identifies the content
+    VATTR_WANTED(&attrs, va_linkid);
```
wilbaker: For my own understanding, is it possible for the linkid of a vnode to change without the vnode's vid changing?

While working on the vnode cache changes I had to do the following:

```
    // Currently delete actions invalidate the cache entry to handle the hardlink+delete rename scenario.  If we do not invalidate
    // the entry we'll find the root for the newly created hardlink and we need to find the root of the path of the file being deleted.
    // Testing has shown that looking up the root again has consistently yielded the root of the file being deleted.
```

Do you know if this could have been avoided if I also stored/compared linkid when looking up an entry in the cache?

ProjFS.Mac/PrjFSKext/public/Message.h
wilbaker: Rather than renumbering could you manually bump `PrjFSKextVersion`?

We have #648 open to starting doing this automatically.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: @nickgra @pmj if we want to keep this around for debugging, what would you think of prepending path with something like "DEBUG:"?

That if something *did* accidentally take a dependency on having the offline root path we'd see it break when testing debug builds?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -960,24 +937,22 @@ static bool TrySendRequestAndWaitForResponse(
     int* kauthResult,
     int* kauthError)
 {
+    // To be useful, the message needs to either provide an FSID/inode pair or a path
+    assert(vnodePath != nullptr || (vnodeFsidInode.fsid.val[0] != 0 || vnodeFsidInode.fsid.val[1] != 0));
```
wilbaker: Is there anything about `vnodeFsidInode.inode` that we can/should assert on?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -97,6 +98,7 @@ static inline PrjFS_NotificationType KUMessageTypeToNotificationType(MessageType
 
 static bool IsVirtualizationRoot(const char* fullPath);
 static void CombinePaths(const char* root, const char* relative, char (&combined)[PrjFSMaxPath]);
+static const char* GetRelativePath(const char* path, const char* root);
```
wilbaker: Naming suggestion: `path` -> `fullPath` for consistency with the other helper functions that operate on full paths.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -638,8 +640,41 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
-    // At the moment, we expect all messages to include a path
-    assert(request.path != nullptr);
+    char pathBuffer[PrjFSMaxPath];
+    if (request.path == nullptr)
+    {
+        fsid_t fsid = request.messageHeader->fsidInode.fsid;
+        ssize_t pathSize = fsgetpath(pathBuffer, sizeof(pathBuffer), &fsid, request.messageHeader->fsidInode.inode);
+        if (pathSize < 0)
+        {
+            cout
```
wilbaker: This looks like somewhere that #395 would be helpful.

Would you mind adding a comment here with a TODO?  Something like:

```
// TODO(Mac): Add this message to PrjFSLib logging once available (#395)
```

/cc: @jrbriggs @jeschu1 @nickgra as FYI

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1345,3 +1378,27 @@ static void ReturnFileMutexIterator(FileMutexMap::iterator lockIterator)
...
     }
 }
+
+static const char* GetRelativePath(const char* path, const char* root)
+{
+    size_t rootLength = strlen(root);
+    size_t pathLength = strlen(path);
+    if (pathLength < rootLength || 0 != memcmp(path, root, rootLength))
+    {
+        fprintf(stderr, "GetRelativePath: root path '%s' is not a prefix of path '%s'\n", root, path);
```
wilbaker: This function is another spot we could use a `// TODO(Mac): Add this message to PrjFSLib logging once available (#395)`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: > So I'd argue that we're best off sending the absolute path, when available, from the kext too and leave the relative path calculation to the user space lib.

I agree it would be nice to consolidate this to user space if we can.  I've filed #846 for this.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: @wilbaker sorry, I don't understand your suggestion, can you please elaborate?

pmj: I think I see what you mean. We can remove the struct member outside of DEBUG builds entirely once the kext no longer needs to deal with relative paths. Do you think we'll still need to poison the string value in that case?

ProjFS.Mac/PrjFSKext/public/Message.h
pmj: Sure, can do.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: > I think I see what you mean. We can remove the struct member outside of DEBUG builds entirely once the kext no longer needs to deal with relative paths. Do you think we'll still need to poison the string value in that case?

I think if we can remove the member completely outside of debug builds then we probably don't need to poison the value.  

I'm thinking we shouldn't be adding so much complexity in `#if DEBUG` that we take a dependency on it for behaving properly.

ProjFS.Mac/PrjFSKext/VnodeUtilities.cpp
pmj: I realised my experience with the edge cases here were somewhat out of date, so I've just run through a few scenarios on both APFS and HFS+ on macOS 10.13.6.

On HFS+:
 * When you create a file, its linkid is equal to its fileid.
 * As soon as you create one or more additional links to a file, each hard link, including the initial one, gets its own separate linkid. So even the original link's ID is changed. The fileid is the same for all of them though.
 * Which link (and indeed linkid) a particular vnode is associated with is not fixed; there seems to be only one vnode for a multi-hardlinked file at a time, and the linkid will fluctuate over time. (This is what I apparently misremembered, or possibly what has changed since 10.8~10.9.)

On APFS:

 * The linkid is apparently always the same as the fileid, even with multiple hard links. The path (and parent vnode) of the vnode keeps changing.

I believe `fsgetpath()` will always get the "correct" path to the file if given the appropriate linkid on HFS+, but on APFS I guess there's no way to distinguish between them.

Overall, this isn't great news for us on either file system; I'm fairly sure there's no way to guarantee we'll always end up with the "correct" path we're after even on HFS+, let alone APFS. This is pretty much orthogonal to the changes in this PR, incidentally. But it does mean that correctly handling hard links is going to be a tricky problem.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I suspect we won't want to send messages for inode 0, but I'm not really willing to bet on it. In any case, this is intended as a basic check for programming errors when calling this function, nothing fancy.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Sounds good.

ProjFS.Mac/PrjFSKext/VnodeUtilities.cpp
wilbaker: > This is pretty much orthogonal to the changes in this PR, incidentally. But it does mean that correctly handling hard links is going to be a tricky problem.

Yep I agree thanks!  Seeing this change in your code got me wondering if there was something different I should be doing with hardlinks and the vnode cache.

pmj: > Seeing this change in your code got me wondering if there was something different I should be doing with hardlinks and the vnode cache.

Nothing fully-baked, but some thoughts that might give you some further ideas:
 * Hard-linked files are easy to detect using the `va_nlink` attribute. If that value is 1, you have nothing to worry about. If it's not 1, `vnode_getparent()` will give you *a* parent, possibly not the one you want. `fsgetpath()` might give you the path inside the provider's root, or it might not.
 * 3 central problems I can see with a multi-hardlinked file:
  1. *Identifying the root(s) where it's a member.* Even with the vnode cache, this is tricky as we don't have a way of iterating over all the links to the file, and vnodes get recycled. Perhaps we want to store a list of UUIDs of virtualisation roots in an xattr? With the vnode cache, we can at least cache this data so we don't take the hit for reading xattrs all the time.
  2. *How to handle various file operations in the case where it's a member of more than 1 root.* Is there any conceivable way such a file could sensibly be anything other than a "full" file? If not, we could *always* convert to full when a hardlink to outside the repo is made, and prevent hardlinking non-full files in offline roots. Unloading the prjfs kext still lets you mess with this.
  3. *In the provider, what to do when `fsgetpath()` (or a fileop handler arg) returns a path outside the root.* The best approach I can currently think of here is to track the inode numbers of all "full" files and look up inodes for multi-linked files there before using `fsgetpath()`.
  4.

wilbaker: @pmj thanks for the suggestions!  I've added a link to your comment to related issue #337

pmj:
@wilbaker Is there anything remaining in this PR you'd like me to change?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: > We can remove the struct member outside of DEBUG builds entirely once the kext no longer needs to deal with relative paths. Do you think we'll still need to poison the string value in that case?

Since the struct still has `path`, can we add the code to poison the path?  I checked and both the Mac functional tests and the Mac large repo build test run against the debug build of the kext.  

If a dependency on `path` is added I'm not sure that we'd catch it with out current tests.

pmj: Yep, if we're not testing release builds, then that does sound essential. I've added that in commit 31a854f now.

ProjFS.Mac/PrjFSKext/VnodeUtilities.cpp
pmj: The latest iteration of commit 3f19b35 in this series has been adjusted with what I learned from that linkid vs fileid investigation.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -867,7 +844,7 @@ static bool TryGetVirtualizationRoot(
 {
     PerfSample findRootSample(perfTracer, PrjFSPerfCounter_VnodeOp_GetVirtualizationRoot);
         
-    *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context);
+    *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context, true);
```
wilbaker: Could you add a small comment here (and in `ShouldHandleFileOpEvent` where we call `Vnode_GetFsidAndInode`) that describes why `true` is being passed in?

pmj: Done!

PR #742, opened 2019-02-05
Read upgrade ring config from feedurl

alameenshah:
This PR helps avoid user needing to specify upgrade ring info in two separate config settings (upgrade.ring and upgrade.feedurl).

Upgrade feedurl already contains upgrade ring specification as a substring. Github upgrader now attempts to parse feedurl and read the ring info specified. If a known ring info (Slow or Fast) is not found, then it would fallback on using the old upgrade.ring gvfs config.

Example: A feedurl in production might look like this "https://1essharedassets.pkgs.visualstudio.com/_packaging/GVFS@Fast/nuget/v3/index.json". The url contains the substring "@Fast/" meaning "Fast" ring has to be used for upgrade checks.

### DIFF
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -529,29 +529,65 @@ namespace GVFS.Common
                 this.UpgradeRing = RingType.NoConfig;
 
                 string ringConfig = null;
-                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, out ringConfig, out error))
-                {
-                    RingType ringType;
-                    if (Enum.TryParse(ringConfig, ignoreCase: true, result: out ringType) &&
-                        Enum.IsDefined(typeof(RingType), ringType) &&
-                        ringType != RingType.Invalid)
-                    {
-                        this.UpgradeRing = ringType;
-                    }
-                    else
-                    {
-                        if (!string.IsNullOrEmpty(ringConfig))
-                        {
-                            this.UpgradeRing = RingType.Invalid;
-                        }
-                    }
+                string upgradeFeedUrl = null;
+                string loadError = "Could not read GVFS Config." + Environment.NewLine + GVFSConstants.UpgradeVerbMessages.SetUpgradeRingCommand;
 
-                    return true;
+                // There are couple config settings that contain info about Upgrade rings.
+                // upgrade.ring and upgrade.feedURL. upgrade.feedURL will override upgrade.ring
+                // config. If upgrade.feedURL is set, but does not contain valid ring info then
+                // upgrade ring would be set to NoConfig and auto upgrade checks will fail.
+                if (!this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, out upgradeFeedUrl, out error))
+                {
+                    error = loadError;
+                    return false;
                 }
 
-                error = "Could not read GVFS Config." + Environment.NewLine;
-                error += GVFSConstants.UpgradeVerbMessages.SetUpgradeRingCommand;
-                return false;
+                // If upgradeFeedUrl is available then use it for the ring config, if not use regular
+                // upgrade.ring.
+                if (!string.IsNullOrEmpty(upgradeFeedUrl))
+                {
+                    this.UpgradeRing = RingType.None;
+
+                    string[] expectedRings = new string[] { RingType.Slow.ToString(), RingType.Fast.ToString() };
+                    foreach (string ring in expectedRings)
+                    {
+                        string ringSubstring = $"@{ring}/";
+                        if (upgradeFeedUrl.IndexOf(ringSubstring, StringComparison.OrdinalIgnoreCase) != -1)
+                        {
+                            ringConfig = ring;
+                            break;
+                        }
+                    }
+                }
+                else if (!this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, out ringConfig, out error))
+                {
+                    error = loadError;
+                    return false;
+                }
+
+                this.ParseUpgradeRing(ringConfig);
+                return true;
+            }
+
+            public void ParseUpgradeRing(string ringConfig)
+            {
+                if (string.IsNullOrEmpty(ringConfig))
+                {
+                    this.UpgradeRing = RingType.None;
+                    return;
+                }
+
+                RingType ringType;
+                if (Enum.TryParse(ringConfig, ignoreCase: true, result: out ringType) &&
+                    Enum.IsDefined(typeof(RingType), ringType) &&
+                    ringType != RingType.Invalid)
+                {
+                    this.UpgradeRing = ringType;
+                }
+                else
+                {
+                    this.UpgradeRing = RingType.Invalid;
+                }
             }
 
             public bool ConfigError()
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
@@ -80,6 +80,75 @@ namespace GVFS.UnitTests.Windows.Upgrader
                 expectedUpgradeVersion:UpgradeTests.NewerThanLocalVersion);
         }
 
+        [TestCase]
+        public void RingInNugetFeedURLOverridesUpgradeRing()
+        {
+            // Pretend there is an upgrade available in Fast ring. Set upgrade.ring
+            // to fast and verify that Upgrader returns the new version.
+            Version newVersion;
+            string error;
+            this.SetUpgradeRing(GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast.ToString());
+
+            // Replace pretend upgrade Release set by UpgradeTests.Setup() method
+            this.Upgrader.PretendNewReleaseAvailableAtRemote(UpgradeTests.NewerThanLocalVersion, GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast);
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldNotBeNull();
+            newVersion.ToString().ShouldEqual(UpgradeTests.NewerThanLocalVersion);
+
+            // Now add upgrade.feedurl with Slow ring info. The Slow ring in upgrade.feedurl should
+            // override the Fast that is set already (in the steps above) in upgrade.ring. Since
+            // there is no upgrade available in Slow, Verify that Upgrader returns Null upgrade
+            // this time.
+            string feedUrlWithSlowRing = "https://foo.bar.visualstudio.com/helloworld/_packaging/GVFS@Slow/nuget/v3/index.json";
+            this.LocalConfig.TrySetConfig("upgrade.feedurl", feedUrlWithSlowRing, out error);
+            this.Upgrader.Config.TryLoad(out error).ShouldBeTrue();
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldBeNull();
+            error.ShouldContain("Great news");
+        }
+
+        [TestCase]
+        public void FastUpgradeRingAndNoRingInNugetFeedURLReturnsNoUpgrade()
+        {
+            // Pretend there is an upgrade available in Fast ring. Set upgrade.ring
+            // to fast and verify that Upgrader returns the new version.
+            Version newVersion;
+            string error;
+            this.SetUpgradeRing(GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast.ToString());
+
+            // Replace pretend upgrade Release set by UpgradeTests.Setup() method
+            this.Upgrader.PretendNewReleaseAvailableAtRemote(UpgradeTests.NewerThanLocalVersion, GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast);
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldNotBeNull();
+
+            // Now add upgrade.feedurl with no ring info. The presence of upgrade.feedurl config
+            // should force upgrader to reset its ring to the one specified in upgrade.feedurl.
+            // But since there is no ring specified in upgrade.feedurl, upgrader should have no valid
+            // ring now. Verify that Upgrader returns Null upgrade this time.
+            string feedUrlWithNoRing = "https://foo.bar.visualstudio.com/helloworld/GVFS/nuget/v3/index.json";
+            this.LocalConfig.TrySetConfig("upgrade.feedurl", feedUrlWithNoRing, out error);
+            this.Upgrader.Config.TryLoad(out error).ShouldBeTrue();
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldBeNull();
+        }
+
+        [TestCase]
+        public void NoRingInNugetFeedURLReturnsNullUpgrade()
+        {
+            string error;
+            string feedUrlWithNoRing = "https://foo.bar.visualstudio.com/helloworld/_packaging/GVFS/nuget/v3/index.json";
+            this.LocalConfig.TrySetConfig("upgrade.feedurl", feedUrlWithNoRing, out error);
+            this.Upgrader.Config.TryLoad(out error).ShouldBeTrue();
+
+            Version newVersion;
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldBeNull();
+        }
+
         public override void NoneLocalRing()
         {
             throw new NotSupportedException();
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockLocalGVFSConfig.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockLocalGVFSConfig.cs
@@ -27,7 +27,8 @@ namespace GVFS.UnitTests.Windows.Mock.Upgrader
         {
             error = null;
 
-            return this.Settings.TryGetValue(name, out value);
+            this.Settings.TryGetValue(name, out value);
+            return true;
         }
 
         public override bool TrySetConfig(


GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -523,7 +523,27 @@ public bool TryLoad(out string error)
...
+                        string ringSubstring = $"@{ring}/";
+                        if (upgradeFeedUrl.IndexOf(ringSubstring, StringComparison.OrdinalIgnoreCase) != -1)
+                        {
+                            ringConfig = ring;
+                            break;
+                        }
+                    }
+                }
+
+                // 2. If feedurl didn't specify a valid ring, then fallback on old behavior (look for upgrade.ring)
```
wilbaker: Are there any valid reasons for the feedurl to not include a ring?

I'm wondering if we have to support falling back to the old config setting.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -523,7 +523,27 @@ public bool TryLoad(out string error)
                 this.UpgradeRing = RingType.NoConfig;
 
                 string ringConfig = null;
-                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, out ringConfig, out error))
+                string upgradeFeedUrl = null;
+
+                // 1. Try to read ring from upgrade.feedurl first.
```
wilbaker: Style suggestion:  I think the comment here (and the `2. If feedurl didn't ...` comment below) would be more useful as a single comment here that describes the rules for determining the ring.

As they are now they're not adding much detail beyond what's already self-documented by the code.

GVFS/GVFS.Common/GitHubUpgrader.cs
wilbaker: cc: @jrbriggs and @jamill for their thoughts as well

jrbriggs: The Feed@View syntax is (as far as I know) an Azure DevOps thing. Falling back to the ring would allow a ringed deployment model on other nuget hosters.

wilbaker: Ok thanks! 

And am I correct in assuming that someone could push official signed packages to their own feed? (I'm guessing that's one way we could get in this scenario?)

jrbriggs: Yeah. Admittedly it's a bit of a contrived scenario. I wouldn't be opposed to simplifying here and picking one way or another. I've gone back and forth on this several times.

wilbaker: > The Feed@View syntax is (as far as I know) an Azure DevOps thing. Falling back to the ring would allow a ringed deployment model on other nuget hosters.

Actually, if this is an Azure DevOps thing, @alameenshah why is this check happening in `GitHubUpgraderConfig`?  Is it because the GitHub upgrader is what's used to check for new versions?

(If this is the case, it would be helpful to include that detail in the comments)

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -523,7 +523,27 @@ public bool TryLoad(out string error)
...
+                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, out upgradeFeedUrl, out error) &&
+                    !string.IsNullOrEmpty(upgradeFeedUrl))
+                {
+                    string[] expectedRings = new string[] { RingType.Slow.ToString(), RingType.Fast.ToString() };
+                    foreach (string ring in expectedRings)
+                    {
+                        string ringSubstring = $"@{ring}/";
+                        if (upgradeFeedUrl.IndexOf(ringSubstring, StringComparison.OrdinalIgnoreCase) != -1)
+                        {
+                            ringConfig = ring;
```
wilbaker: Can some unit test(s) be added to validate the new functionality?

GVFS/GVFS.Common/GitHubUpgrader.cs
wilbaker: > Yeah. Admittedly it's a bit of a contrived scenario. I wouldn't be opposed to simplifying here and picking one way or another. I've gone back and forth on this several times.

@jrbriggs, @alameenshah  another option (and I'm going back and forth on how I feel about it) is that the ring setting, if present, will override whatever is picked up automatically from the feed name.  

Some potential advantages:

- A ring config of "None" will disable auto-upgrade (even if the feed has a ring name in it).

- If someone has a different naming scheme in their feed (e.g. "Lab", "Fast", "Canary") they can manually control which GitHub release should drive the auto-upgrade notifications.

> Admittedly it's a bit of a contrived scenario

My scenario might be even *more* contrived, but please let me know what you think.

jamill: For this initial release, let's keep this as simple as possible and treat these non-mainline scenarios as out-of-scope. I would rather explicitly treat these as not-handled / not-specified, rather than semi-support them and get locked into a behavior we didn't think through (or give the hint they are supported and then change the behavior in the future).

If `upgrade.feedurl` is specified, then we assume that the user intended to use `NuGetUpgrader`. If we cannot find a corresponding GitHub release ring, then prompt for auto-upgrade is disabled.

If the ring setting is present in addition to a feedurl, then we can either ignore it, or treat it as a misconfiguration (as we are not sure of the user's intent). I don't think we should fall back to GitHubUpgrader in this case, as a misconfigured NuGetUpgrader could then result in the user unexpectedly using the GitHubUpgrader.

> * A ring config of "None" will disable auto-upgrade (even if the feed has a ring name in it).

Now - this might be a state we want to support. Do we need to support a state where a NuGet upgrader is configured, but the prompt to upgrade is disabled?

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -523,7 +523,27 @@ public bool TryLoad(out string error)
                 this.UpgradeRing = RingType.NoConfig;
 
                 string ringConfig = null;
-                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, out ringConfig, out error))
+                string upgradeFeedUrl = null;
+
+                // 1. Try to read ring from upgrade.feedurl first.
+                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, out upgradeFeedUrl, out error) &&
```
jamill: As noted in my other comment - I don't think we should fall back to the GitHubUpgrader if there is an error to read the `UpgradeFeedUrl`, or it is non-empty. We should only load the GitHubUpgrader if `UpgraderFeedUrl` has been read, and it is empty (not present).

Alternatively, if we do not want to mix up `GitHubUpgrader` parsing `NuGetUpgrader` specific config settings, maybe we could read the config outside of `GitHubUpgrader`, and construct a `GitHubUpgrader` with an explicit ring... Someone has to read the config, but this might allow us to separate concerns a little bit.

GVFS/GVFS.Common/GitHubUpgrader.cs
jamill: There is some discussion about what the behavior should be - I think unit tests would be a great way to clarify the expected behavior under different conditions.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -523,7 +523,27 @@ public bool TryLoad(out string error)
...
+
+                // 1. Try to read ring from upgrade.feedurl first.
+                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, out upgradeFeedUrl, out error) &&
+                    !string.IsNullOrEmpty(upgradeFeedUrl))
+                {
+                    string[] expectedRings = new string[] { RingType.Slow.ToString(), RingType.Fast.ToString() };
+                    foreach (string ring in expectedRings)
+                    {
+                        string ringSubstring = $"@{ring}/";
+                        if (upgradeFeedUrl.IndexOf(ringSubstring, StringComparison.OrdinalIgnoreCase) != -1)
```
jamill: Can the @ symbol appear anywhere else in the URL? Is there any more checks / restrictions we should put on the URLs we consider to avoid any false hits?

I think this logic would be a great candidate to extract into its own method with its set of unit tests.

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: consolidated all comments into a single block.

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: Added couple UT to demonstrate 
1. feedURL config would override upgrade.ring config. 
2. Upgrade fails when feedURL config exists but does not contain ring info.

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: As discussed, keeping it simple for the time being. (No fallback to using `upgrade.ring` when `feedURL` does not contain `Fast`/`Slow` rings info.)

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: @jrbriggs correct me if I'm wrong - it is safe to assume that `@Fast/` or `@Slow/` will not occur more that once.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
```diff
@@ -80,6 +80,42 @@ public void UpgradeAvailableOnSlowWhileOnLocalFastRing()
...
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldNotBeNull();
+
+            string feedUrlWithSlowRing = "https://1essharedassets.pkgs.visualstudio.com/_packaging/GVFS@Slow/nuget/v3/index.json";
+            this.LocalConfig.TrySetConfig("upgrade.feedurl", feedUrlWithSlowRing, out error);
+            this.Upgrader.Config.TryLoad(out error).ShouldBeTrue();
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldBeNull();
```
wilbaker: What in this test is validating that the Feed URL has overridden the upgrade ring?

I can see that the test is validating that `TryQueryNewestVersion` returns true, but how do we know it's using the right ring?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
```diff
@@ -80,6 +80,42 @@ public void UpgradeAvailableOnSlowWhileOnLocalFastRing()
...
+            error.ShouldContain("Great news");
+        }
+
+        [TestCase]
+        public void NoRingInNugetFeedURLReturnsNullUpgrade()
```
wilbaker: Can you add a test where there is no ring set in the nuget feed URL, but there is a ring set in `upgrade.ring`?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
alameenshah: line no. 88 sets `upgrade.ring` to `Fast`. Line no. 89 to 92 sets an upgrade available in `Fast`ring  and verifies that Upgrader is able to query and return the new version. 

Now without changing anything (upgrade is still available in `Fast`), line no. 95 sets `upgrade.feedurl` with `Slow` ring config. When we query for newest version, upgrader now returns `Null` in line no. 98. Meaning the `Fast` `upgrade.ring` config got overridden by the `Slow` in `upgrade.feedurl`. We didn't set any upgrade to be available in `Slow`, so upgrader returned `Null` this time for new version.

Will add some comments explaining this behavior in the UT.

alameenshah: Updated the UT with comments to clarify the override behavior.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
alameenshah: Added. Please take a look.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
wilbaker: > line no. 88 sets `upgrade.ring` to `Fast`. Line no. 89 to 92 sets an upgrade available in `Fast`ring and verifies that Upgrader is able to query and return the new version.

The `Setup` method also makes an update available in the Slow ring:

```
        public virtual void Setup()
        {
            this.Tracer = new MockTracer();
            this.Output = new MockTextWriter();
            this.PrerunChecker = new MockInstallerPrerunChecker(this.Tracer);
            this.LocalConfig = new MockLocalGVFSConfig();

            this.Upgrader = new MockGitHubUpgrader(
                LocalGVFSVersion,
                this.Tracer,
                new GitHubUpgrader.GitHubUpgraderConfig(this.Tracer, this.LocalConfig));

            this.PrerunChecker.Reset();
            this.Upgrader.PretendNewReleaseAvailableAtRemote(
                upgradeVersion: NewerThanLocalVersion,
                remoteRing: GitHubUpgrader.GitHubUpgraderConfig.RingType.Slow);
            this.SetUpgradeRing("Slow");
        }
```

Line 92 (as best I can tell) is only validating that there was *some* newer version returned.  How is the test validating that it's the fast version (and not the slow version)?

> We didn't set any upgrade to be available in `Slow`, so upgrader returned `Null` this time for new version.

See code in `Setup`, it looks like something is available in `Slow`, why is this not getting picked up?  Did it get wiped out when we set it to fast?

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -523,7 +523,32 @@ public bool TryLoad(out string error)
...
                 string ringConfig = null;
-                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, out ringConfig, out error))
+                string upgradeFeedUrl = null;
+                bool feedURLConfigured = false;
+
+                // There are couple config settings that contain info about Upgrade rings.
+                // upgrade.ring and upgrade.feedURL. upgrade.feedURL will override upgrade.ring
+                // config. If upgrade.feedURL is set, but does not contain valid ring info then
+                // upgrade ring would be set to NoConfig and auto upgrade checks will fail.
+                if (this.LocalConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, out upgradeFeedUrl, out error) &&
```
wilbaker: If a feed is configured and it does not contain a valid ring this method will return `false` with the error "error = "Could not read GVFS Config."

Is that expected (we were able to read the GVFS config)?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
alameenshah: As discussed offline, added a comment in the UT to clarify the behavior of `this.Upgrader.PretendNewReleaseAvailableAtRemote`.

The flow:
1. `UpgradeTests.Setup()` sets the ring and a pretend-release applicable to most UT cases using `PretendNewReleaseAvailableAtRemote`.
2. When an individual UT calls `PretendNewReleaseAvailableAtRemote` again, it replaces (rather than adding to) what was set by `UpgradeTests.Setup()` method with new Release-Ring info that the UT wants to use.

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: good catch. we can provide more granular error messaging.

wilbaker: Is it expected that this method will return false in that scenario?  I think some of my confusion is coming from this method being used for two different purposes (determining the ring and reading the config when a user is configured to use the GitHub upgrader).

wilbaker: Actually, based on the unit tests it looks like `false` is expected?  I am not sure I understand when this method will return `true` vs `false`

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
```diff
@@ -80,6 +80,77 @@ public void UpgradeAvailableOnSlowWhileOnLocalFastRing()
...
+            this.Upgrader.PretendNewReleaseAvailableAtRemote(UpgradeTests.NewerThanLocalVersion, GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast);
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldNotBeNull();
+
+            // Now add upgrade.feedurl with Slow ring info. The Slow ring in upgrade.feedurl should
+            // override the Fast that is set already (in the steps above) in upgrade.ring. Since
+            // there is no upgrade available in Slow, Verify that Upgrader returns Null upgrade
+            // this time.
+            string feedUrlWithSlowRing = "https://1essharedassets.pkgs.visualstudio.com/_packaging/GVFS@Slow/nuget/v3/index.json";
```
wilbaker: Could you update the unit tests to use fake URLs?

GVFS/GVFS.Common/GitHubUpgrader.cs
alameenshah: On further digging, it turns out `TryLoad` returns `false` only when there is a file i/o error while attempting to read `gvfs.config`. Otherwise it should return `true`. I restructured the code a little to make this more obvious. Also removed the new code where I was incorrectly returning `false` on `feedurl` parse errors. Fixed couple UT as well where I was wrongly checking the `TryLoad` return code.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -529,29 +529,65 @@ public bool TryLoad(out string error)
...
+                    ringType != RingType.Invalid)
+                {
+                    this.UpgradeRing = ringType;
+                    return;
                 }
 
-                error = "Could not read GVFS Config." + Environment.NewLine;
-                error += GVFSConstants.UpgradeVerbMessages.SetUpgradeRingCommand;
-                return false;
+                if (string.IsNullOrEmpty(ringConfig))
```
wilbaker: Should we check `string.IsNullOrEmpty(ringConfig)` before calling `Enum.TryParse(ringConfig ...`?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/ProductUpgraderTests.cs
```diff
@@ -80,6 +80,74 @@ public void UpgradeAvailableOnSlowWhileOnLocalFastRing()
...
+            // to fast and verify that Upgrader returns the new version.
+            Version newVersion;
+            string error;
+            this.SetUpgradeRing(GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast.ToString());
+
+            // Replace pretend upgrade Release set by UpgradeTests.Setup() method
+            this.Upgrader.PretendNewReleaseAvailableAtRemote(UpgradeTests.NewerThanLocalVersion, GitHubUpgrader.GitHubUpgraderConfig.RingType.Fast);
+
+            this.Upgrader.TryQueryNewestVersion(out newVersion, out error).ShouldBeTrue();
+            newVersion.ShouldNotBeNull();
```
wilbaker: Minor: You could make this validation a bit more specific and verify that `newVersion` is the same as `UpgradeTests.NewerThanLocalVersion`.

wilbaker: (No need to block on this)

PR #746, opened 2019-02-06
Call git-credential approve on successful credential usage

### LINKED ISSUE #745: VFS4Git should call `git credential approve/reject` after `git credential fill`
From the [documentation about `git-credential`](https://git-scm.com/docs/git-credential):

> If the action is `fill`, git-credential will attempt to add "username" and "password" attributes to the description by reading config files, by contacting any configured credential helpers, or by prompting the user. The username and password attributes of the credential description are then printed to stdout together with the attributes already provided.
>
> If the action is `approve`, git-credential will send the description to any configured credential helpers, _**which may store the credential for later use**_.
>
> If the action is `reject`, git-credential will send the description to any configured credential helpers, which may erase any stored credential matching the description.
>
> If the action is `approve` or `reject`, no output should be emitted.

Currently we are only calling `git credential fill` without a subsequent `approve` or `reject` from inside of VFS for Git. This leads to a second authentication prompt when `git` operations are called because although we generated the PAT (for Azure Repos), we never instructed GCM to store it.

Due to quirky behaviour of the Windows GCM, credentials _are being stored on a call to `git credential fill`_ (which calls `git-credential-manager get`) without the calls to `approve/reject` as required by the Git usage documentation. The new GCM on macOS follows the `git-credential` API correctly and as such we are forgetting the credential after the initial prompt.

---

Aside: the credential helper API documentation can be found [here](https://git-scm.com/docs/api-credentials). It states that a helper should do the following things when called with each action:

> `get` (called by `git-credential fill`)
> Return a matching credential, if any exists.
>
> `store` (called by `git-credential approve`)
> Store the credential, if applicable to the helper.
>
> `erase` (called by `git-credential reject`)
> Remove a matching credential, if any, from the helper’s storage.

Windows GCM incorrectly combined `store` into the `get` call, and it's `store` (and `erase` in some instances) end up being no-ops.

mjcheetham:
After we have successfully used a generated credential we should inform Git that the credential is valid and should be stored, by calling `git credential approve`.

We make an effort to not call `approve` (which calls `gcm store`) if we've already called it once before for the currently cached credential to prevent a `store` per successfully auth'd network request.

Fixes #745

### DIFF
--- a/GVFS/GVFS.Common/Git/GitAuthentication.cs
+++ b/GVFS/GVFS.Common/Git/GitAuthentication.cs
@@ -20,7 +20,8 @@ namespace GVFS.Common.Git
         private int numberOfAttempts = 0;
         private DateTime lastAuthAttempt = DateTime.MinValue;
 
-        private string cachedAuthString;
+        private string cachedCredentialString;
+        private bool isCachedCredentialStringApproved = false;
 
         private bool isInitialized;
 
@@ -47,57 +48,90 @@ namespace GVFS.Common.Git
 
         private GitSsl GitSsl { get; }
 
-        public void ConfirmCredentialsWorked(string usedCredential)
+        public void ApproveCredentials(ITracer tracer, string credentialString)
         {
             lock (this.gitAuthLock)
             {
-                if (usedCredential == this.cachedAuthString)
+                // Don't reset the backoff if this is for a different credential than we have cached
+                if (credentialString == this.cachedCredentialString)
                 {
                     this.numberOfAttempts = 0;
                     this.lastAuthAttempt = DateTime.MinValue;
+
+                    // Tell Git to store the valid credential if we haven't already
+                    // done so for this cached credential.
+                    if (!this.isCachedCredentialStringApproved)
+                    {
+                        string username;
+                        string password;
+                        if (TryParseCredentialString(this.cachedCredentialString, out username, out password))
+                        {
+                            this.git.ApproveCredentials(this.repoUrl, username, password);
+                            this.isCachedCredentialStringApproved = true;
+                        }
+                        else
+                        {
+                            EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
+                            {
+                                ["RepoUrl"] = this.repoUrl,
+                                ["CredentialString"] = this.cachedCredentialString
+                            });
+                            tracer.RelatedError(metadata, "Failed to parse credential string for approval");
+                        }
+                    }
                 }
             }
         }
 
-        public void Revoke(string usedCredential)
+        public void RejectCredentials(ITracer tracer, string credentialString)
         {
             lock (this.gitAuthLock)
             {
-                if (usedCredential != this.cachedAuthString)
+                // Don't stomp a different credential
+                if (credentialString == this.cachedCredentialString && this.cachedCredentialString != null)
                 {
-                    // Don't stomp a different credential
-                    return;
-                }
+                    // If we can we should pass the actual username/password values we used (and found to be invalid)
+                    // to `git-credential reject` so the credential helpers can attempt to check if they're erasing
+                    // the expected credentials, if they so choose to.
+                    string username;
+                    string password;
+                    if (TryParseCredentialString(this.cachedCredentialString, out username, out password))
+                    {
+                        this.git.RejectCredentials(this.repoUrl, username, password);
+                    }
+                    else
+                    {
+                        // We failed to parse the credential string so instead (as a recovery) we try to erase without
+                        // specifying the particular username/password.
+                        EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
+                        {
+                            ["RepoUrl"] = this.repoUrl,
+                            ["CredentialString"] = this.cachedCredentialString
+                        });
+                        tracer.RelatedWarning(metadata, "Failed to parse credential string for rejection. Rejecting any credential for this repo URL.");
+                        this.git.RejectCredentials(this.repoUrl);
+                    }
 
-                if (this.cachedAuthString != null)
-                {
-                    this.cachedAuthString = null;
-                    this.git.RevokeCredential(this.repoUrl);
-
+                    this.cachedCredentialString = null;
+                    this.isCachedCredentialStringApproved = false;
                     this.UpdateBackoff();
                 }
             }
         }
 
-        public bool TryRefreshCredentials(ITracer tracer, out string errorMessage)
-        {
-            string authString;
-            return this.TryGetCredentials(tracer, out authString, out errorMessage);
-        }
-
-        public bool TryGetCredentials(ITracer tracer, out string gitAuthString, out string errorMessage)
+        public bool TryGetCredentials(ITracer tracer, out string credentialString, out string errorMessage)
         {
             if (!this.isInitialized)
             {
                 throw new InvalidOperationException("This auth instance must be initialized before it can be used");
             }
 
-            gitAuthString = this.cachedAuthString;
-            if (gitAuthString == null)
+            credentialString = this.cachedCredentialString;
+            if (credentialString == null)
             {
                 lock (this.gitAuthLock)
                 {
-                    if (this.cachedAuthString == null)
+                    if (this.cachedCredentialString == null)
                     {
                         if (this.IsBackingOff)
                         {
@@ -111,7 +145,7 @@ namespace GVFS.Common.Git
                         }
                     }
 
-                    gitAuthString = this.cachedAuthString;
+                    credentialString = this.cachedCredentialString;
                 }
             }
 
@@ -179,6 +213,28 @@ namespace GVFS.Common.Git
                 httpClientHandler.ClientCertificateOptions = ClientCertificateOption.Manual;
                 httpClientHandler.ClientCertificates.Add(cert);
             }
+        }
+
+        private static bool TryParseCredentialString(string credentialString, out string username, out string password)
+        {
+            if (credentialString != null)
+            {
+                byte[] data = Convert.FromBase64String(credentialString);
+                string rawCredString = Encoding.ASCII.GetString(data);
+
+                string[] usernamePassword = rawCredString.Split(':');
+                if (usernamePassword.Length == 2)
+                {
+                    username = usernamePassword[0];
+                    password = usernamePassword[1];
+
+                    return true;
+                }
+            }
+
+            username = null;
+            password = null;
+            return false;
         }
 
         private bool TryAnonymousQuery(ITracer tracer, Enlistment enlistment, out bool isAnonymous)
@@ -249,7 +305,8 @@ namespace GVFS.Common.Git
 
             if (!string.IsNullOrEmpty(gitUsername) && !string.IsNullOrEmpty(gitPassword))
             {
-                this.cachedAuthString = Convert.ToBase64String(Encoding.ASCII.GetBytes(gitUsername + ":" + gitPassword));
+                this.cachedCredentialString = Convert.ToBase64String(Encoding.ASCII.GetBytes(gitUsername + ":" + gitPassword));
+                this.isCachedCredentialStringApproved = false;
             }
             else
             {
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -155,11 +155,48 @@ namespace GVFS.Common.Git
             }
         }
 
-        public virtual void RevokeCredential(string repoUrl)
+        public virtual void RejectCredentials(string repoUrl, string username = null, string password = null)
         {
+            StringBuilder sb = new StringBuilder();
+            sb.AppendFormat("url={0}\n", repoUrl);
+
+            // Passing the username and password that we want to signal rejection for is optional.
+            // Credential helpers that support it can use the provided username/password values to
+            // perform a check that they're being asked to delete the same stored credential that
+            // the caller is asking them to erase.
+            if (username != null)
+            {
+                sb.AppendFormat("username={0}\n", username);
+            }
+
+            if (password != null)
+            {
+                sb.AppendFormat("password={0}\n", password);
+            }
+
+            sb.Append("\n");
+
+            string stdinConfig = sb.ToString();
+
             this.InvokeGitOutsideEnlistment(
                 "credential reject",
-                stdin => stdin.Write("url=" + repoUrl + "\n\n"),
+                stdin => stdin.Write(stdinConfig),
+                null);
+        }
+
+        public virtual void ApproveCredentials(string repoUrl, string username, string password)
+        {
+            StringBuilder sb = new StringBuilder();
+            sb.AppendFormat("url={0}\n", repoUrl);
+            sb.AppendFormat("username={0}\n", username);
+            sb.AppendFormat("password={0}\n", password);
+            sb.Append("\n");
+
+            string stdinConfig = sb.ToString();
+
+            this.InvokeGitOutsideEnlistment(
+                "credential approve",
+                stdin => stdin.Write(stdinConfig),
                 null);
         }
 
@@ -232,8 +269,8 @@ namespace GVFS.Common.Git
             using (ITracer activity = tracer.StartActivity("TryGetCredentials", EventLevel.Informational))
             {
                 Result gitCredentialOutput = this.InvokeGitAgainstDotGitFolder(
-                    "-c " + GitConfigSetting.CredentialUseHttpPath + "=true credential fill",
-                    stdin => stdin.Write("url=" + repoUrl + "\n\n"),
+                    $"-c {GitConfigSetting.CredentialUseHttpPath}=true credential fill",
+                    stdin => stdin.Write($"url={repoUrl}\n\n"),
                     parseStdOutLine: null);
 
                 if (gitCredentialOutput.ExitCodeIsFailure)
--- a/GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
+++ b/GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
@@ -1,4 +1,5 @@
-﻿using GVFS.Common.Git;
+using System.Linq;
+using GVFS.Common.Git;
 using GVFS.Tests;
 using GVFS.Tests.Should;
 using GVFS.UnitTests.Mock.Common;
@@ -33,15 +34,17 @@ namespace GVFS.UnitTests.Git
 
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true, "Failed to get initial credential");
 
-            dut.Revoke(authString);
-            dut.IsBackingOff.ShouldEqual(false, "Should not backoff after credentials initially revoked");
+            dut.RejectCredentials(tracer, authString);
+            dut.IsBackingOff.ShouldEqual(false, "Should not backoff after credentials initially rejected");
+            gitProcess.CredentialRejections["mock://repoUrl"].Count.ShouldEqual(1);
 
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true, "Failed to retry getting credential on iteration");
             dut.IsBackingOff.ShouldEqual(false, "Should not backoff after successfully getting credentials");
 
-            dut.Revoke(authString);
-            dut.IsBackingOff.ShouldEqual(true, "Should continue to backoff after revoking credentials");
+            dut.RejectCredentials(tracer, authString);
+            dut.IsBackingOff.ShouldEqual(true, "Should continue to backoff after rejecting credentials");
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(false, "TryGetCredential should not succeed during backoff");
+            gitProcess.CredentialRejections["mock://repoUrl"].Count.ShouldEqual(2);
         }
 
         [TestCase]
@@ -59,10 +62,12 @@ namespace GVFS.UnitTests.Git
             for (int i = 0; i < 5; ++i)
             {
                 dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true, "Failed to get credential on iteration " + i + ": " + error);
-                dut.Revoke(authString);
+                dut.RejectCredentials(tracer, authString);
                 dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true, "Failed to retry getting credential on iteration " + i + ": " + error);
-                dut.ConfirmCredentialsWorked(authString);
+                dut.ApproveCredentials(tracer, authString);
                 dut.IsBackingOff.ShouldEqual(false, "Should reset backoff after successfully refreshing credentials");
+                gitProcess.CredentialRejections["mock://repoUrl"].Count.ShouldEqual(i+1, $"Should have {i+1} credentials rejection");
+                gitProcess.CredentialApprovals["mock://repoUrl"].Count.ShouldEqual(i+1, $"Should have {i+1} credential approvals");
             }
         }
 
@@ -79,16 +84,18 @@ namespace GVFS.UnitTests.Git
             string error;
 
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true, "Failed to get initial credential");
-            dut.Revoke(authString);
+            dut.RejectCredentials(tracer, authString);
+            gitProcess.CredentialRejections["mock://repoUrl"].Count.ShouldEqual(1);
 
             gitProcess.ShouldFail = true;
 
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(false, "Succeeded despite GitProcess returning failure");
             dut.IsBackingOff.ShouldEqual(true, "Should continue to backoff if failed to get credentials");
 
-            dut.Revoke(authString);
+            dut.RejectCredentials(tracer, authString);
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(false, "TryGetCredential should not succeed during backoff");
             dut.IsBackingOff.ShouldEqual(true, "Should continue to backoff if failed to get credentials");
+            gitProcess.CredentialRejections["mock://repoUrl"].Count.ShouldEqual(1);
         }
 
         [TestCase]
@@ -108,8 +115,10 @@ namespace GVFS.UnitTests.Git
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true);
 
             // Simulate a 401 error on two threads
-            dut.Revoke(authString);
-            dut.Revoke(authString);
+            dut.RejectCredentials(tracer, authString);
+            dut.RejectCredentials(tracer, authString);
+            gitProcess.CredentialRejections["mock://repoUrl"].Count.ShouldEqual(1);
+            gitProcess.CredentialRejections["mock://repoUrl"][0].BasicAuthString.ShouldEqual(authString);
 
             // Both threads should still be able to get a PAT for retry purposes
             dut.TryGetCredentials(tracer, out authString, out error).ShouldEqual(true, "The second thread caused back off when it shouldn't");
@@ -126,7 +135,9 @@ namespace GVFS.UnitTests.Git
             dut.TryInitializeAndRequireAuth(tracer, out _);
 
             string thread1Auth;
+            string thread1AuthRetry;
             string thread2Auth;
+            string thread2AuthRetry;
             string error;
 
             // Populate an initial PAT on two threads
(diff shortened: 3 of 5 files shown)


GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -45,6 +46,16 @@ public void ConfirmCredentialsWorked(string usedCredential)
                 {
```
jeschu1: I realize this is existing but saw it was curious.  Why do we action only if (usedCredential == this.cachedAuthString).  Since this function is saying we've confirmed this credential works could just always save usedCredential?  Is there a case I'm missing here?

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -152,6 +164,28 @@ public bool TryInitializeAndRequireAuth(ITracer tracer, out string errorMessage)
...
         }
 
+        private bool TryParseCachedAuthString(out string username, out string password)
+        {
+            if (this.cachedAuthString != null)
+            {
+                byte[] data = Convert.FromBase64String(this.cachedAuthString);
+                string credentialString = Encoding.ASCII.GetString(data);
+
+                string[] foo = credentialString.Split(':');
```
jeschu1: nit: I think we use more descriptive names than foo generally :-)

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -168,17 +169,47 @@ public void TwoThreadsInterleavingFailuresShouldntStompASuccess()
...
+        {
+            MockTracer tracer = new MockTracer();
+            MockGitProcess gitProcess = this.GetGitProcess();
+
+            GitAuthentication dut = new GitAuthentication(gitProcess, "mock://repoUrl");
+            dut.TryInitializeAndRequireAuth(tracer, out _);
+
+            string authString;
+            dut.TryGetCredentials(tracer, out authString, out _).ShouldBeTrue();
+            dut.ConfirmCredentialsWorked(authString);
```
jeschu1: You may want to validate the case where we send in a string other than what was retrieved from Get

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -152,6 +164,28 @@ public bool TryInitializeAndRequireAuth(ITracer tracer, out string errorMessage)
             return false;
         }
 
+        private bool TryParseCachedAuthString(out string username, out string password)
```
jeschu1: Instead of this I think we could just store the username/password in addition to the cachedAuthString.  I don't feel strongly.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -168,17 +169,47 @@ public void TwoThreadsInterleavingFailuresShouldntStompASuccess()
...
+            gitProcess.StoredCredentials.Single().Key.ShouldEqual("mock://repoUrl");
+        }
+
         private MockGitProcess GetGitProcess()
         {
             MockGitProcess gitProcess = new MockGitProcess();
             gitProcess.SetExpectedCommandResult("config gvfs.FunctionalTests.UserName", () => new GitProcess.Result(string.Empty, string.Empty, GitProcess.Result.GenericFailureCode));
             gitProcess.SetExpectedCommandResult("config gvfs.FunctionalTests.Password", () => new GitProcess.Result(string.Empty, string.Empty, GitProcess.Result.GenericFailureCode));
 
+            int approvals = 0;
```
jeschu1: I don't see approvals (or revocations) used for validation.  What are the thoughts on this check?

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -221,6 +255,7 @@ private bool TryCallGitCredential(ITracer tracer, out string errorMessage)
             if (!string.IsNullOrEmpty(gitUsername) && !string.IsNullOrEmpty(gitPassword))
             {
                 this.cachedAuthString = Convert.ToBase64String(Encoding.ASCII.GetBytes(gitUsername + ":" + gitPassword));
+                this.storedCachedAuthString = false;
```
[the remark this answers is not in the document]
jeschu1: I believe there is no guarantee, the credentials here worked.  This is just what GCM returned.

GVFS/GVFS.Common/Git/GitAuthentication.cs
jeschu1: The line I'm referencing got cut off, but I was wondering about 

`                if (usedCredential == this.cachedAuthString)
`

in ConfirmCredentialsWorked.  Why can't we just always use the confirmed string?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -164,6 +164,22 @@ public virtual void RevokeCredential(string repoUrl)
...
 
+        public virtual void StoreCredential(string repoUrl, string username, string password)
+        {
+            var sb = new StringBuilder();
+            sb.AppendFormat("url={0}\n", repoUrl);
+            sb.AppendFormat("username={0}\n", username);
+            sb.AppendFormat("password={0}\n", password);
+            sb.Append("\n");
+
+            string stdinConfig = sb.ToString();
```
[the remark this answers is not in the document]
mjcheetham: This has to use \n by the Git protocol.

GVFS/GVFS.Common/Git/GitAuthentication.cs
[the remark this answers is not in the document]
mjcheetham: ah yes! my bad.

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
```diff
@@ -17,17 +15,33 @@ public MockGitProcess()
...
 
         public void SetExpectedCommandResult(string command, Func<Result> result, bool matchPrefix = false)
         {
             CommandInfo commandInfo = new CommandInfo(command, result, matchPrefix);
             this.expectedCommandInfos.Add(commandInfo);
         }
 
+        public override void StoreCredential(string repoUrl, string username, string password)
+        {
+            IList<Credential> repoCredentials;
```
nickgra: Is there a reason why this is `IList` vs `List`?

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
```diff
@@ -17,17 +15,33 @@ public MockGitProcess()
             : base(new MockGVFSEnlistment())
         {
             this.CommandsRun = new List<string>();
+            this.StoredCredentials = new Dictionary<string, IList<Credential>>(StringComparer.OrdinalIgnoreCase);
         }
 
         public List<string> CommandsRun { get; }
         public bool ShouldFail { get; set; }
+        public IDictionary<string, IList<Credential>> StoredCredentials { get; }
```
nickgra: Is there a reason why this is `IDictionary` vs `Dictionary`?

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -45,6 +46,16 @@ public void ConfirmCredentialsWorked(string usedCredential)
                 {
                     this.numberOfAttempts = 0;
                     this.lastAuthAttempt = DateTime.MinValue;
+
+                    // Tell Git to store the valid credential if we haven't already
+                    // done so for this cached credential.
+                    string username;
+                    string password;
+                    if (!this.storedCachedAuthString && this.TryParseCachedAuthString(out username, out password))
```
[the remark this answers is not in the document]
mjcheetham: Using a `SecureString` is usually more hassle than it's worth. A few reasons not to bother IMO:
1. `SecureString` isn't actually encrypted on Unix systems, only on Windows.
2. `SecureString` lives in unmanaged memory as a `BSTR`, meaning you need to jump through hoops to get access to it. I mean, this is kind of the point (to prevent accidental usage of the string in normal code).
3. If you're able to read managed memory to get access to a normal `System.String` then you can also just watch what GCM and Git are passing back and forth over standard input.
4. Anyone can just ask GCM to dump out their PATs knowing only the repo URL without any restrictions other than running as the current user (unless the user chooses to lock their own login keychain on Mac; Windows has no such protection).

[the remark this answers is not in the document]
mjcheetham: https://docs.microsoft.com/en-us/dotnet/api/system.security.securestring?view=netframework-4.7.2#how-secure-is-securestring

> Overall, SecureString is more secure than String because it limits the exposure of sensitive string data. However, those strings may still be exposed to any process or operation that has access to raw memory, such as a malicious process running on the host computer, a process dump, or a user-viewable swap file. **Instead of using SecureString to protect passwords, the recommended alternative is to use an opaque handle to credentials that are stored outside of the process\*.**

\* You could read the last one as only interfacing via GCM which reads/writes the credentials at rest from the OS credential store.

GVFS/GVFS.Common/Git/GitAuthentication.cs
mjcheetham: I did wonder about that, but then you've got to worry about keeping them in sync with each other.

GVFS/GVFS.Common/Git/GitAuthentication.cs
mjcheetham: Yeah.. I missed this one! Now you know how unimaginative my variable names are when doing WIP.
(Sometimes I also use `x`, `soup` or `fish`!)

GVFS/GVFS.Common/Git/GitAuthentication.cs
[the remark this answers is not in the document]
mjcheetham: Jessica is correct. We don't know if the credentials worked or not, just that we got some credentials back from Git/GCM. The user could have for example signed into the wrong account, which would return 'valid' credentials but not valid for accessing the particular Azure Repos repository/endpoint.

The `GitAuthentication` type here is actually following the same model as `git-credential` in that there is:
1. a way to get credentials `fill`/`TryGetCredentials(...)`,
2. a way to indicate credentials did _not_ work `reject`/`Revoke(...)`, and
3. a way to indicate credentials _did_ work `approve`/`ConfirmCredentialsWorked(...)`.

The `HttpRequestor` type (which is the only product-code caller of `GitAuthentication.TryGetCredentials(...)`) already calls all of the above methods, it's just that `GitAuthentication` itself never called through to `git-credential approve`.

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
mjcheetham: My default is to use `IFoo` rather than `Foo` when the member is `public`, even if that's only public to others in the assembly. No one else cares what the exact implementation is, and they don't need to use any members only exposed on `Foo` but not `IFoo`. If they do they can change it.

---

Aside:

I've noticed there seems to be a huge aversion in the team towards using interfaces or other abstract types, out of some fear I don't fully understand or buy\*.

Is "be liberal in what you accept and conservative in what you return" (even within in the same codebase) not a widely accepted ethos?

\* I know there is some opposition when it comes to the LINQ-based collection methods vs the concrete-types' O(1) properties, which is a valid concern when it comes to perf which I _do_ buy. However my counterargument to avoiding the `ICollection` etc interfaces is that:

1. Treating developers as 'dangerous' and introducing arbitrary rule/restrictions/commandments on usage or non-usage of language and framework constructs & patterns is silly. (It's why I'm also an opponent of some of the more pedantic StyleCop rules around comments and whitespace.. it's just more frustrating than anything for the perceived benefits.)
Each usage or non-usage should be taken in it's own context as to if it's appropriate, or not, or if it doesn't really matter.
2. Many of the LINQ calls actually have [fast paths](https://github.com/dotnet/corefx/blob/53ae9af48c5c133f4413e76c1469dfe6ccecd926/src/System.Linq/src/System/Linq/Count.cs#L19-L22) for the LINQ calls, for example `new List().Count()` actually calls through to the `.Count` property. Not that I'm advocating for using `.Count()` over `.Count` but any accidental usage isn't actually that impactful. 
3. Any good IDE, static analyser, and code review would notice "usage of `.Count()` when `.Count` is available". In fact if a code review misses `.Count()` vs `.Count` then it's also possible the same review would miss `IList` vs `List` too.

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
mjcheetham: See my answer to `IDictionary` vs `Dictionary`

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
mjcheetham: I did wonder this. I was wondering if really this `MockGitProcess` would be better being replaced with a `Moq.Mock` with calls to `Verify` to check the number of invocations of various methods.

I opted not to churn the code too much however. I would reconsider this decision however if there's support/consensus on making these unit tests more thorough?

GVFS/GVFS.Common/Git/GitAuthentication.cs
mjcheetham: I'm also not entirely sure why this check is done (this is place when code comments would have been useful!)
I didn't want to change things however because it seems to be some defence against cases when there are two different cached auth strings, and avoiding a reset of the backoff mechanism erroneously.

Still not entirely sure how or if such a situation could occur however. Perhaps it was a check that wasn't really needed. Thoughts?

GVFS/GVFS.Common/Git/GitAuthentication.cs
jeschu1: Hahahaha I love soup! :-)

GVFS/GVFS.Common/Git/GitAuthentication.cs
jeschu1: @mjcheetham which is exactly the other issue I was debating :-).

GVFS/GVFS.Common/Git/GitAuthentication.cs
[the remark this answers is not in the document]
jeschu1: We actually had a bit of debate here regarding, so pulling in @jamill for his feedback.

Am I correct in saying the new GCM will return whatever is stored if the string matches?  Aka no validation on GCM?  For *new* keys we would have had to validate we were able to get the PAT against the org.

GVFS/GVFS.Common/Git/GitAuthentication.cs
jeschu1: @mjcheetham I totally agree, it's hard to know the thought process without comments.  

To me the correct flow would be take the key that worked, break it apart and store.  I would think if GCM failed and got called again we may get different sets, but even still why not just take the working one?  If you need to get this PR in now maybe keep as is.  If it's going out in the next release I think it's worth fixing up to what we think the right behavior should be.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
jeschu1: I'm not a big fan of unused code, but I can understand the argument of keeping things consistent.  It seems like if we haven't used this to verify yet, not sure we ever will.  I wouldn't block a PR over this just pointing out something odd.

GVFS/GVFS.Common/Git/GitAuthentication.cs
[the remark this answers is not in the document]
mjcheetham: New GCM does zero validation of what is stored in the credential store/Keychain. We hand whatever we have straight back for the caller to determine if it's correct and valid. That's what the `approve` and `reject` actions are for on `git-credential`.

Encoding what is considered a 'valid' credential isn't as clear cut as you'd think and depends on what you're doing to do with it, hence new GCM doesn't. Windows GCM does do some  call to see if the repository 'upload_pack' endpoint can be reached.

If the credential is bad (say it has expired) then doing a `git fetch` will fail the first time, and the bad credential will be deleted. A subsequent call will result in a prompt (since there is now no credential stored, good or bad):

```
$ git fetch
fatal: Authentication failed for '<URL>'
$ git fetch
# prompt is shown & fetch is executed successfully
```

If we want to get into the business of verifying credentials before returning to Git then we should be very careful about thing. We can end up getting into states were we're being too smart, for example for NTLM/Kerberos or other federated tokens via ADAL where in Windows GCM we're saying the credentials are 'valid' for one purpose then then Git finds out they're not and we start ignoring requests to delete 'invalid' credentials.

Not to mention we should then be thinking about what constitutes 'valid' for VFS4Git vs normal Git operations, and other host providers like GitHub or on-prem TFS (and which versions?).

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
wilbaker: Another point I'd like to bring up is consistency.  StyleCop helps quite a bit in this area, but it cannot solve everything.

The approach we have taken in the rest of the code base is to prefer using the concrete types.  I don't have an issue with revisiting that decision, but for now we should strive to keep the code as consistent as possible.  

In my experience when there is a break in consistency developers assume there was an important reason for breaking with the paradigm that's already in place.  This, in turn, makes it more difficult to do things like refactor and get new developers up to speed.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -164,6 +164,22 @@ public virtual void RevokeCredential(string repoUrl)
                 null);
         }
 
+        public virtual void StoreCredential(string repoUrl, string username, string password)
+        {
+            var sb = new StringBuilder();
```
wilbaker: `var` -> `StringBuilder`

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
jrbriggs: @wilbaker brings up a good point. Let's aim to stay as consistent as we can in the codebase, especially modifying existing code. We can take up coding style changes as separate issues.

GVFS/GVFS.Common/Git/GitAuthentication.cs
mjcheetham: @jeschu1 so I think I know why we're checking first that the credentials we used are the same as the ones we have cached.

If the actual stored credentials in the credstore/keychain have been replaced with new ones (outside of our knowledge), but the ones we last returned in a call to `TryGetCredentials` (and we have cached) are different, we don't want to tell Git to erase 'the credentials for the repo' as this might erase the now good ones. Note that we don't actually tell Git "hey, credential X is bad" we can only say "the credentials you know about are bad".

mjcheetham: Of course this is all a problem of potentially having a stale cached value, and knowing when the cache is stale or if the stored credentials are bad. Git was never designed to be used this way.. all calls to `git credential fill` should be immediately followed by either `approve` or `reject`, and then `git-credential` should only be called by one process at a time per remote URL.

Interleaving calls to `fill` and `reject` for the same remote URL on a single system will lead to problems. There's an assumed static state after calling `fill` that no one has touched the credentials we just got back.

If `reject` also took the actual credential we want to say didn't work, then the helpers would be able to check if that's the actual credential value they are being asked to erase.

mjcheetham: @jeschu1 I've added a small comment here about not resetting the backoff for a non-cached credential string. I don't want to change this too much at this point.

However you got my thinking if new GCM could help prevent this sort of problem in the future. I think we can.
Git itself will call `reject` passing the username and password it got from the `fill` call, meaning GCM can be smarter about how it handles requests to delete.

_**If**_ the username and password were specified on a call to `reject/erase`, then we can double check that the stored credential we have matches the user/pass we were given, and not erase it if it's different from the one we were told about. If we're not given a user/pass then just delete the credential without checking if the user/pass match the stored values.

Since Git is always consistent in what it sends to `reject` (what it got back from `fill`) the normal path here for Git deleting an invalid cred is still OK. Older VFS for Git's don't send any username or password so GCM will continue to delete any old credential values for the given key.

One day we might be able to remove the check in 'Revoke()' and just always pass through the credential strings, ignoring checking the cached version, but we'll always need to check the cached value to ensure we're not resetting the backoff settings. :(

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -37,57 +38,62 @@ public bool IsBackingOff
...
 
-        public void Revoke(string usedCredential)
+        public void RejectCredentials(string credentialString)
         {
             lock (this.gitAuthLock)
-            {
-                if (usedCredential != this.cachedAuthString)
-                {
-                    // Don't stomp a different credential
-                    return;
```
mjcheetham: I inverted the `if (..x..) return; DoThing();` to `if (!..x..) { DoThing(); }` to be consistent with the way we do this check in the above `CredentialsWorked`/`Approve` method.

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -37,57 +38,62 @@ public bool IsBackingOff
...
+                        this.git.RejectCredentials(this.repoUrl);
 
-                    this.UpdateBackoff();
+                        this.UpdateBackoff();
+                    }
                 }
             }
         }
 
-        public bool TryRefreshCredentials(ITracer tracer, out string errorMessage)
```
mjcheetham: Refresh was not being used.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -168,17 +169,47 @@ public void TwoThreadsInterleavingFailuresShouldntStompASuccess()
             thread2Auth.ShouldEqual(thread1Auth, "The second thread stomp the first threads good auth string");
         }
 
+        [TestCase]
+        public void DontDoubleStoreExistingCredential()
```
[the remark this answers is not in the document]
mjcheetham: Not sure. I don't think the design of `GitAuthentication` lends itself to testing that unfortunately.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
mjcheetham: @jeschu1  I modified the `MockGitProcess` to track the number of approvals and rejections so we can assert the expected number of calls in the tests.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
mjcheetham: Added test.

GVFS/GVFS.Common/Git/GitAuthentication.cs
mjcheetham: I won't store the username/password as well as the basic auth string to avoid needing to keep them all in sync (and test for that). The deconstruction code here is only called once per mount process, and once per auth string, so it's not going to cause a noticeable perf problem to compute it rather than store the user/pass.

GVFS/GVFS.UnitTests/Mock/Git/MockGitProcess.cs
mjcheetham: Moved back to non-interface.

GVFS/GVFS.Common/Git/GitAuthentication.cs
jeschu1: Yay deleting code!

GVFS/GVFS.Common/Git/GitAuthentication.cs
jeschu1: @mjcheetham I think I see the point here.  So if another process replaces the credentials we wouldn't necessarily want to store/replace ours.  I guess that's fair enough.  The flow is a bit confusing, but it predates your work here.  Thanks for reasoning that out.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -154,36 +155,89 @@ public void TwoThreadsInterleavingFailuresShouldntStompASuccess()
...
+
+            string authString;
+            dut.TryGetCredentials(tracer, out authString, out _).ShouldBeTrue();
+            dut.ApproveCredentials(authString);
+            dut.ApproveCredentials(authString);
+            dut.ApproveCredentials(authString);
+            dut.ApproveCredentials(authString);
+            dut.ApproveCredentials(authString);
+
+            gitProcess.NumApprovals["mock://repoUrl"].ShouldEqual(1);
```
kewillford: Should we validate `NumRejections` here as well?  Should these validations of `NumApprovals` and `NumRejections` be added to any of the tests above?

GVFS/GVFS.Common/Git/GitAuthentication.cs
wilbaker: @mjcheetham  @jeschu1, I did some digging and this question was raised in the PR that initially added this check ("PR 223211: Improve GitAuthentication retry logic in multithreaded scenarios" in the old Azure Repos GVFS repo):

>Saeed Noursalehi
6/5/2017
In what scenario would this condition not be true?
>
>Forest Willard
6/5/2017
If a thread starts a long request as a PAT expires, then it will succeed with that PAT, but another thread may have already experienced a failure and changed the PAT.
>
>Saeed Noursalehi
6/5/2017
Removing this if doesn't cause any test failures. 

I have not tried removing the check but based on Saeed's comment I am assuming there is a test that fails without this check?

There was a lengthier discussion where the check was added in the Revoke/Reject method:

>Saeed Noursalehi
6/5/2017
Do we actually need this check? As I step through the race scenarios, I think the check below for "cachedAuthString != null" would also handle the race.
>
>The actual bug was that in the old code, the else path in this method would set authRetryBackoff to DateTime.MaxValue. Now that that line is gone, I think all of the rest of the logic works OK without having to pass around this extra usedCredential parameter.
>
>Thoughts?
>
>Forest Willard
6/5/2017
This handles a different race than the bug. If I remove this line, I'm pretty darn sure "TwoThreadsInterleavingFailuresShouldntStompASuccess" will start failing. The effects of it failing aren't so bad, though. It just means that we would create one extra PAT when we don't need to, given everything returns success. If two threads grab a failing PAT, then one retries before the second can fully fail, the second will revoke the good PAT the first retried with. Then we will renew it later.
>
>Actually, I think it'd be possible to get failed requests out of this race, but it's highly unlikely. When the second thread revokes the good PAT, there could be a brief period where the first thread hasn't reset the backoff counter, and anything else would fail without trying.
>
>Saeed Noursalehi
6/5/2017
I don't really follow what the race is, or how this check fixes it. The effect of this check here is that this method returns true and the caller will retry. If you remove this check, cachedAuthString will be null, and DateTime.Now will be less than the backoff time.
>
>Unless... if the first thread actually succeeds in recreating the PAT before the second thread even got to rejecting the old one, then cachedAuthString won't be null, and we revoke again. Is that the race?
>
>Forest Willard
6/5/2017
Yup, that's the race.
>
>Saeed Noursalehi
6/5/2017
OK makes sense, and I confirmed that removing this code causes a unit test failure, so that's good.

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -37,57 +38,62 @@ public bool IsBackingOff
...
-
-                if (this.cachedAuthString != null)
-                {
-                    this.cachedAuthString = null;
-                    this.git.RevokeCredential(this.repoUrl);
+            {
+                // Don't stomp a different credential
+                if (credentialString == this.cachedCredentialString)
+                {
+                    if (this.cachedCredentialString != null)
```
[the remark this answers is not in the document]
mjcheetham: > Combine this with the if above to be `if (this.cachedCredentialString != null && credentialString == this.cachedCredentialString)` now?

Sure!

> Wonder if we need the same `this.cachedCredentialString !=null` check above in `ApproveCredentials` as well?

I was wondering this as well. Need to check if the anonymous requests (which would have no auth string) play a role in this logic.

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
mjcheetham: Sure!

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -37,57 +38,62 @@ public bool IsBackingOff
...
                 {
                     this.numberOfAttempts = 0;
                     this.lastAuthAttempt = DateTime.MinValue;
+
+                    // Tell Git to store the valid credential if we haven't already
+                    // done so for this cached credential.
+                    string username;
+                    string password;
+                    if (!this.isCachedCredentialStringApproved &&
+                        TryParseCachedAuthString(this.cachedCredentialString, out username, out password))
```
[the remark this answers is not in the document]
mjcheetham: @wilbaker `GitAuthentication` does not have an `ITracer` except ones passed in for some methods.

Should I update `GitAuthentication` so it requires an `ITracer` in the constructor or just add `ITracer` to the `Approve` and `Reject` methods?

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -154,36 +155,89 @@ public void TwoThreadsInterleavingFailuresShouldntStompASuccess()
...
             thread2Auth.ShouldEqual(thread1Auth, "The second thread stomp the first threads good auth string");
         }
 
+        [TestCase]
+        public void DontDoubleStoreExistingCredential()
+        {
+            MockTracer tracer = new MockTracer();
+            MockGitProcess gitProcess = this.GetGitProcess();
+
+            GitAuthentication dut = new GitAuthentication(gitProcess, "mock://repoUrl");
```
jeschu1: Nit: since you use "mock://repoUrl" in several places you could assign it to a string

GVFS/GVFS.Common/Git/GitAuthentication.cs
mjcheetham: I've added a new [issue](https://github.com/Microsoft/Git-Credential-Manager-Core/issues/24) (and PR) on GCM Core to hopefully avoid this sort of problem (or at least push co-ordination of deleting credentials back to GCM). VFS will still need to do this sort of check however to update/reset its backoff.

GVFS/GVFS.Common/Git/GitAuthentication.cs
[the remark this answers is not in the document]
wilbaker: I think whichever option is easier.  If we already have an `ITracer` available when constructing `GitAuthentication` you could just add it to the constructor.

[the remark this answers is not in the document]
mjcheetham: Looking at passing an `ITracer` in via the constructor seemed like a lot of effort, since `GitAuthentication` is created inside the constructor of `Enlistment`, which would mean plumbing `ITracer` all the way through 😕 

I went with just taking `ITracer` in the `Approve` and `Reject` methods.

mjcheetham:
> I want to take this change, but let's target the release after the one that we're starting to prepare.

@jrbriggs am I OK to merge this into master now?

jrbriggs:
Yes, @mjcheetham -- master targets M149. Thanks!

PR #749, opened 2019-02-07
Update git to include trace2 V7

jamill:


wilbaker:
If there are any other changes in Git with this upgrade, could you add them to the PR description?

nickgra:
For some reason, the MacOS version of the Git package wasn't published to the MyGet feed. We'll need to get that figured out.

PR #751, opened 2019-02-07
Mac: Rename reverse-DNS xattr & kext identifiers

pmj:
Changes the various io.gvfs./io_gvfs_ identifiers to use the org.vfsforgit./org_vfsforgit_ prefix, as per #584. This includes kext bundle identifier, kext IOKit class prefixes and xattr names.


ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -210,6 +210,11 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode,
     {
         PrjFSVirtualizationRootXAttrData rootXattr = {};
         SizeOrError xattrResult = Vnode_ReadXattr(vnode, PrjFSVirtualizationRootXAttrName, &rootXattr, sizeof(rootXattr));
+        if (xattrResult.error == ENOATTR)
+        {
+            xattrResult = Vnode_ReadXattr(vnode, PrjFSVirtualizationRootXAttrNameOld, &rootXattr, sizeof(rootXattr));
+        }
+
```
pmj: Key question about this PR is whether to bother with fallbacks like this, or if we want any other kind of backwards compatibility facility, such as a script/tool that you can run over your VFS4G repos to rename all the xattrs.

nickgra: My 2c: We only have a few early adopters who are actively using VFS4G repos on MacOS. They're ready for some short term pain, so I don't think it's too unreasonable to ask them to push any local changes, delete their current clones and reclone with the new version whenever they pick this change up. So long as this is one of the only times we have to make a change like this, it seems fine to me.

If we do that, we can also avoid putting the old version of the xattr into the code. None of our automation relies on long lived repos, we clone a fresh one for every test.

I'm curious to hear what everyone else thinks.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1084,7 +1095,8 @@ static bool InitializeEmptyPlaceholder(const char* fullPath, TPlaceholder* data,
 static bool IsVirtualizationRoot(const char* fullPath)
 {
     PrjFSVirtualizationRootXAttrData data = {};
-    if (TryGetXAttr(fullPath, PrjFSVirtualizationRootXAttrName, sizeof(PrjFSVirtualizationRootXAttrData), &data))
+    if (TryGetXAttr(fullPath, PrjFSVirtualizationRootXAttrName, sizeof(PrjFSVirtualizationRootXAttrData), &data)
+        || TryGetXAttr(fullPath, PrjFSVirtualizationRootXAttrNameOld, sizeof(PrjFSVirtualizationRootXAttrData), &data))
```
jeschu1: @wilbaker what was the final decision on if we wanted backwards compat?

wilbaker: Since we don't have to support backwards compat right now, let's not take on maintenance costs of adding it.

What we could do, however, is bump the on-disk version such that on Windows there's a no-op upgrade and on Mac there is a message that a breaking change occurred and users must reclone.

(That could be its own PR as long as we don't ship before completing that work)

cc: @jrbriggs @nickgra

pmj: We don't even need to bump the on-disk version for this, given that the xattr name is different. As soon as we detect the old xattr name, we can output that error.

wilbaker: > We don't even need to bump the on-disk version for this, given that the xattr name is different. As soon as we detect the old xattr name, we can output that error.

Unfortunately once the repo is mounted VFS4G has no way to output errors to the user.

When we've had to make breaking changes like this in the past we've adjusted `RepoMetadata.DiskLayoutVersion.MinimumSupportedMajorVersion` so that when mounting a user will see:

>https://github.com/Microsoft/VFSForGit/blob/cf69c01dace22ddeef8e7bbc3a9322a1e73abfac/GVFS/GVFS.Common/DiskLayoutUpgrade.cs#L79) 

> As soon as we detect the old xattr name, we can output that error.


I don't think we should add the overhead of checking for the old xattr name.  If we make `MinimumSupportedMajorVersion` platform specific (and set it to a new value on Mac) we can prevent the old repos from mounting in the first place which eliminates the need to check for the old xattrs.

wilbaker: > I don't think we should add the overhead of checking for the old xattr name. If we make `MinimumSupportedMajorVersion` platform specific (and set it to a new value on Mac) we can prevent the old repos from mounting in the first place which eliminates the need to check for the old xattrs.

I should have included that I think that change and this change should get merged into master at the same time, @jrbriggs, is there someone on our side with the bandwidth to help make this change in the VFS4G product code (unless you'd like to make the change @pmj)?

jeschu1: @wilbaker, I think I can help with that if you are looking volunteers.  

Were you thinking just move MinimumSupportedMajorVersion to GVFSPlatform, and up to 8 for Mac?

wilbaker: > @wilbaker, I think I can help with that if you are looking volunteers.

Ok great!

> Were you thinking just move MinimumSupportedMajorVersion to GVFSPlatform, and up to 8 for Mac?

I was thinking that at a minimum we'll want to move `MinimumSupportedMajorVersion` into  `GVFSPlatform`, but I hadn't dug into what other changes we'd need to make.

Currently the on-disk version for both platforms is 17 (the original idea was to keep them in sync if we could), but perhaps we should allow them to diverge, and have separate Mac and Windows platform on-disk version numbers.  In our current situation, for example, we need to be able to increase both the `MinimumSupportedMajorVersion` and the current major version on Mac without impacting either of those version numbers on Windows.

I think it's worth spending a bit of time to dig into the options here so that we can avoid painting ourselves into a corner.  Feel free to catch me offline if you want some additional details on the current state of the on-disk version code.

pmj:
Thanks for the review everyone. I've just re-pushed this without the fallbacks to the old xattr names.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: > I should have included that I think that change and this change should get merged into master at the same time, @jrbriggs, is there someone on our side with the bandwidth to help make this change in the VFS4G product code (unless you'd like to make the change @pmj)?

I think @jeschu1 is no doubt better prepared for this task than I am; happy to help out wherever necessary/appropriate though!

@jeschu1 feel free to push to this PR branch so we can get both changes merged at the same time - I think that should work; if not, let me know.

wilbaker:
> Thanks for the review everyone. I've just re-pushed this without the fallbacks to the old xattr names.

All the changes for far look good to me.  I think the plan to have @jeschu1 push the disk version changes for Mac to this branch sounds good.  Thanks!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: @pmj, I added the commit to separate versioning on Mac/Windows.

jeschu1:
@wilbaker the versioning commit is added.

@pmj's changes look good to me so i'm unblocking, but we shouldn't merge until someone has a chance to review my new commit.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Thanks @jeschu1 - they look sane to me but I'm not really the right person to be signing them off so I'll leave that to those more experienced with the VFS4G userland code.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -18,7 +18,11 @@ public MacPlatform()
                 underConstruction: new UnderConstructionFlags(
                     supportsGVFSService: false,
                     supportsGVFSUpgrade: false,
-                    supportsGVFSConfig: false))
+                    supportsGVFSConfig: false),
+                diskLayoutVersion: new DiskLayoutVersion(
+                    currentMajorVersion: 18,
+                    currentMinorVersion: 0,
+                    minimumSupportedMajorVersion: 7))
```
wilbaker: `minimumSupportedMajorVersion` should be bumped to 18 as well so that `TryCheckDiskLayoutVersion` returns the correct error message.

GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -619,7 +619,7 @@ private void CheckNotInsideExistingRepo(string normalizedEnlistmentRootPath)
 
             try
             {
-                RepoMetadata.Instance.SaveCloneMetadata(tracer, enlistment);
+                RepoMetadata.Instance.SaveCloneMetadata(tracer, enlistment, GVFSPlatform.Instance.DiskLayout.CurrentMajorVersion, GVFSPlatform.Instance.DiskLayout.CurrentMinorVersion);
```
wilbaker: I don't think it hurts anything, but what was the reason for making the change here rather than using `GVFSPlatform.Instance.DiskLayout` directly in `RepoMetdata`?

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -149,6 +151,29 @@ public class UnderConstructionFlags
...
+            // version of GVFS.
+            public int CurrentMajorVersion { get; }
+
+            // The minor version should be bumped whenever there is an upgrade that can be safely ignored by older versions of GVFS.
+            // For example, this allows an upgrade step that sets a default value for some new config setting.
+            public int CurrentMinorVersion { get; }
+
+            // This is the last time GVFS made a breaking change that required a reclone. This should not
+            // be incremented ever again as all format changes should now be supported with an upgrade step.
+            public int MinimumSupportedMajorVersion { get; }
```
wilbaker: Minor, update comment from:

```
// This is the last time GVFS made a breaking change that required a reclone. This should not
// be incremented ever again as all format changes should now be supported with an upgrade step.
```

To something like:
```
// This is the last time GVFS made a breaking change that required a reclone. This should not
// be incremented on platforms that have released a v1.0 as all their format changes should be 
// supported with an upgrade step.
```

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -25,6 +26,7 @@ public GVFSPlatform(string executableExtension, string installerExtension, Under
 
         public GVFSPlatformConstants Constants { get; }
         public UnderConstructionFlags UnderConstruction { get; }
+        public DiskLayoutVersion DiskLayout { get; }
```
wilbaker: What are your thoughts on moving this into `IDiskLayoutUpgrade`?

Although it's not *strictly* related to the upgraders, perhaps they're close enough it's worth moving them?  (It might make the relationship between these versions and the upgraders a bit clearer).

/cc: @nickgra for his thoughts as well

GVFS/GVFS.Common/RepoMetadata.cs
```diff
@@ -91,10 +91,16 @@ public static void Shutdown()
 
         public string GetCurrentDiskLayoutVersion()
```
wilbaker: Rather than calling `TryGetOnDiskLayoutVersion` you can just use the values from `GVFSPlatform` (this method was previously just formatting the hardcoded version numbers).

GVFS/GVFS.Common/DiskLayoutUpgrade.cs
```diff
@@ -73,33 +73,33 @@ public static bool TryCheckDiskLayoutVersion(ITracer tracer, string enlistmentRo
             {
                 if (TryGetDiskLayoutVersion(tracer, enlistmentRoot, out majorVersion, out minorVersion, out error))
```
wilbaker: Could you add a functional test that validates that `gvfs mount` will fail when trying to mount a repo on Mac that's version 17?  (see `DiskLayoutUpgradeTests` for examples of how we test this on Windows)

GVFS/GVFS/CommandLine/CloneVerb.cs
jeschu1: Good point I should do the change directly in RepoMetaData

GVFS/GVFS.Common/RepoMetadata.cs
jeschu1: Can you elaborate on that?  Shouldn't it do check similar to what TryGetOnDiskLayoutVersion is doing?  Are there cases where a value could be missing and an error is more preferable?

GVFS/GVFS.Common/GVFSPlatform.cs
jeschu1: Do you mean IDiskLayoutUpgradeData?  I didn't see IDiskLayoutUpgrade.

I must admit I'm not super familiar the upgrade relationship myself, but since this just reads repo metadata I'm not sure I'd expect it with upgrade.

wilbaker: > Do you mean IDiskLayoutUpgradeData?

Yep sorry about that.

> but since this just reads repo metadata I'm not sure I'd expect it with upgrade.

The relationship I was thinking about was that the disk layout version is what drives the upgraders.

See this comment for `CurrentMajorVersion`

```
// The major version should be bumped whenever there is an on-disk format change that requires a one-way upgrade.
```

GVFS/GVFS.Common/RepoMetadata.cs
wilbaker: > Can you elaborate on that?

I wasn't sure what the motivation was for changing the behavior of `GetCurrentDiskLayoutVersion` as part of the changes in this PR.

`GetCurrentDiskLayoutVersion` is simply a helper method for formatting `CurrentMajorVersion` and `CurrentMinorVersion`, it doesn't check what's on disk.

(In fact, it's only called in one spot and could potentially be replaced with `$"{GVFSPlatform.Instance.DiskLayout.CurrentMajorVersion}.{GVFSPlatform.Instance.DiskLayout.CurrentMinorVersion}"`).

> Are there cases where a value could be missing and an error is more preferable?

The error conditions are already being detected by the mount verb and mount process using `TryGetOnDiskLayoutVersion`.

Perhaps there's a better name for this method (or we should simply remove it) to make it clearer what it's being used for?

GVFS/GVFS.Common/GVFSPlatform.cs
jeschu1: To be clear the proposed change would be to make:

    public interface IDiskLayoutUpgradeData
    {
        DiskLayoutUpgrade[] Upgrades { get; }
        DiskLayoutVersion Version { get; }
        bool TryParseLegacyDiskLayoutVersion(string dotGVFSPath, out int majorVersion);
    }

I can definitely do that if we think we should.   @wilbaker let me know your final verdict :-)

nickgra: I think the proposed change makes sense, the reasoning that the disk layout version drives the upgraders makes sense to me.

wilbaker: @jeschu1 I think the proposed interface change to `IDiskLayoutUpgradeData` looks.

GVFS/GVFS.Common/DiskLayoutUpgrade.cs
jeschu1: DiskLayoutVersionTests.cs added.  This swaps version based on Mac/Windows running on both.

jeschu1:
@wilbaker ready for another look

GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs
```diff
@@ -0,0 +1,54 @@
...
+using GVFS.FunctionalTests.Tools;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System.Runtime.InteropServices;
+
+namespace GVFS.FunctionalTests.Windows.Tests
+{
+    [TestFixture]
+    [Category(Categories.FullSuiteOnly)]
+    [Category(Categories.WindowsOnly)]
```
wilbaker: This shouldn't be tagged as `WindowsOnly`

GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            // Advance the minor version, mount should still work
+            this.Enlistment.UnmountGVFS();
+            GVFSHelpers.SaveDiskLayoutVersion(
+                this.Enlistment.DotGVFSRoot,
+                this.currentDiskMajorVersion.ToString(),
+                (CurrentDiskLayoutMinorVersion + 1).ToString());
+            this.Enlistment.TryMountGVFS().ShouldBeTrue("Mount should succeed because only the minor version advanced");
+
+            // Advance the major version, mount should fail
```
wilbaker: Could you also added a test that mount will fail if the version that's on disk is lower than the minimum supported version (7 on Windows and 18 on Mac)?

GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            // Advance the minor version, mount should still work
+            this.Enlistment.UnmountGVFS();
+            GVFSHelpers.SaveDiskLayoutVersion(
+                this.Enlistment.DotGVFSRoot,
+                this.currentDiskMinimumMajorVersion.ToString(),
+                (CurrentDiskLayoutMinorVersion).ToString());
+            this.Enlistment.TryMountGVFS().ShouldBeTrue("Mount should succeed because we are using minimum version");
+
+            // Advance the major version, mount should fail
```
wilbaker: This comment needs to be updated to something like "Mount should fail if on disk version is below minimum supported version"

GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs
```diff
@@ -0,0 +1,78 @@
...
+                this.Enlistment.DotGVFSRoot,
+                (this.currentDiskMajorVersion + 1).ToString(),
+                CurrentDiskLayoutMinorVersion.ToString());
+            this.Enlistment.TryMountGVFS().ShouldBeFalse("Mount should fail because the major version has advanced");
+        }
+
+        [TestCase]
+        public void MountFailsIfBeforeMinimumVersion()
+        {
+            // Advance the minor version, mount should still work
```
wilbaker: This comment also needs to be updated.

jeschu1:
@wilbaker it would be great if you can have another look today.  @jrbriggs suggested this goes out with the new drop we're doing shortly.

jeschu1:
@wilbaker I'm removing the 
// Mount should succeed if on disk version is the minimum supported version
portion of the test

This isn't passing because it requires writing auxillary files to do the upgrades.  I looked into creating them but we don't have hooks for file system writing here, it's involved, and it also duplicates the upgrader tests.

I've left the "Mount should fail if on disk version is below minimum supported version" test.

[2019-03-01 15:04:35 -05:00] Information {"Message":"Disk layout version is now: 13"}
[2019-03-01 15:04:35 -05:00] Information {"Message":"Upgrading from disk layout 13 to 14"}
[2019-03-01 15:04:35 -05:00] Information {"Message":"Copied no ESENT blob size entries. C:\\Repos\\GVFSFunctionalTests\\enlistment\\8d4032d1e6d64bb181fa\\.gvfs\\BlobSizes does not exist"}
[2019-03-01 15:04:35 -05:00] Information {"Message":"Set BlobSizesRoot: C:\\Repos\\GVFSFunctionalTests\\enlistment\\8d4032d1e6d64bb181fa\\.gvfs\\databases\\blobSizes"}
[2019-03-01 15:04:35 -05:00] Information {"Message":"Disk layout version is now: 14"}
[2019-03-01 15:04:35 -05:00] Information {"Message":"Upgrading from disk layout 14 to 15"}
[2019-03-01 15:04:35 -05:00] Error {"ErrorMessage":"IOException: System.IO.FileNotFoundException: Could not find file 'C:\\Repos\\GVFSFunctionalTests\\enlistment\\4032d1e6d64bb181fa\\src\\.git\\info\\sparse-checkout'.\r\nFile name: 'C:\\Repos\\GVFSFunctionalTests\\enlistment\\4032d1e6d64bb181fa\\src\\.git\

wilbaker:
> This isn't passing because it requires writing auxillary files to do the upgrades. I looked into creating them but we don't have hooks for file system writing here, it's involved, and it also duplicates the upgrader tests.

That sounds good to me, thanks!  When we start adding upgrade steps on Mac it will have its own set of upgrader steps and so I agree that portion of the test should be remove.

Issue #704, opened 2019-01-17
Upgrader fails to run due to git process


jeschu1:
This could be because
1. We don't wait for the process to exit and immediate run the next step causing the process check to fail
i.e. something like:
```
process.Kill();
if (!process.WaitForExit((int)timeout.TotalMilliseconds))
{
       //Send a warning we couldn’t kill the process
}
```
would fix.
2. The process is ignoring the kill signal

To test we should set up a long running background job and attempt to unmount.  

@jamill @derrickstolee  for visibility

PR #752, opened 2019-02-07
List blocking processes with pid

jeschu1:
This does a very small change to give both process and pid when the gvfs installer cannot run.
This will allow users to better see the git processes causing issues.
If we feel necessary we could go a step further and list the full command line, but that isn't on `Process `so we'd have to call lower level  windows/linux specific commands so I'm punting unless we fine we need it.

This
```
ERROR: Blocking processes are running.
Run `gvfs upgrade --confirm` again after quitting these processes - git
```
Becomes
```
ERROR: Blocking processes are running.
Run `gvfs upgrade --confirm` again after quitting these processes - git pid:9116, git pid:16396
```

Ideally we'd want to kill background processes when we unmount.  
Speaking with @dscho that unfortunately is non-trival.
What we have in code today sends process.Kill (Terminate) to the running git process.  
This isn't sufficient however, since git may have spawned other processes.

**Here are some options:**
1. Walk the processes tree calling Kill.  This is the easiest and also a cross-platform solution.  The issue here could be that we leave .lock files interfering with future runs causing user frustration.  
This would be something to double check with @derrickstolee , but @dscho believes we'll almost certainly have this issue.

2. Create a Windows and Unix specific solution
- On Windows
  - Ensure VFS / Git have the same bit version 
  - Enumerate the process tree filtering by original pid as (possibly transitive) parent, then send all of them a remote thread with ExitProcess()
- On Linux/Mac
  - kill(2)

3. Leverage Git For Windows for the Windows specific solution
- On Windows
  - Expose an existing git feature for killing [here](https://github.com/git-for-windows/git/blob/master/compat/win32/exit-process.h)

_For now I think it's best to see how big of an issue this is before planning reaction work._
#704

### DIFF
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -76,7 +76,7 @@ namespace GVFS.Upgrader
                 // actually quits.
                 this.tracer.RelatedInfo("Checking if GVFS or dependent processes are running.");
                 int retryCount = 10;
-                HashSet<string> processList = null;
+                List<string> processList = null;
                 while (retryCount > 0)
                 {
                     if (!this.IsBlockingProcessRunning(out processList))
@@ -126,11 +126,11 @@ namespace GVFS.Upgrader
             return GVFSEnlistment.IsUnattended(this.tracer);
         }
 
-        protected virtual bool IsBlockingProcessRunning(out HashSet<string> processes)
+        protected virtual bool IsBlockingProcessRunning(out List<string> processes)
         {
             int currentProcessId = Process.GetCurrentProcess().Id;
             Process[] allProcesses = Process.GetProcesses();
-            HashSet<string> matchingNames = new HashSet<string>();
+            List<string> matchingNames = new List<string>();
 
             foreach (Process process in allProcesses)
             {
@@ -139,7 +139,7 @@ namespace GVFS.Upgrader
                     continue;
                 }
 
-                matchingNames.Add(process.ProcessName);
+                matchingNames.Add(process.ProcessName + " pid:" + process.Id);
             }
 
             processes = matchingNames;
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockInstallerPreRunChecker.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockInstallerPreRunChecker.cs
@@ -72,9 +72,9 @@ namespace GVFS.UnitTests.Windows.Mock.Upgrader
             return this.FakedResultOfCheck(FailOnCheckType.UnattendedMode);
         }
 
-        protected override bool IsBlockingProcessRunning(out HashSet<string> processes)
+        protected override bool IsBlockingProcessRunning(out List<string> processes)
         {
-            processes = new HashSet<string>();
+            processes = new List<string>();
 
             bool isRunning = this.FakedResultOfCheck(FailOnCheckType.BlockingProcessesRunning);
             if (isRunning)


derrickstolee:
As far as the .lock files are concerned: the biggest one to worry about is index.lock. We already clean up the commit-graph.lock and multi-pack-index.lock files before calling the processes that rewrite them. We may also strand some tmp_XXXX files when writing .idx files or pack files.

I would focus on the commands we run in the background, as we can expect a user to help their own processes before running upgrade. Other tools may run read-only processes in the background (like status or rev-list) that will not be problematic to halt.

jeschu1:
> The bigger questions\concerns in my mind are how users will react when they see a message like this:
> 
> `Run `gvfs upgrade --confirm` again after quitting these processes - git pid:9116, git pid:16396`

Agree.  But of course this is an improvement from:
`Run `gvfs upgrade --confirm` again after quitting these processes - git`

> * Do we want users to kill these processes (assuming they are BG processes), or should they wait?
> * If users _do_ kill the BG git processes, can they restart again properly (given there might be stale lock files)?

When I previously discussed with @derrickstolee we said it was fine to force kill any background processes.  As part of the steps we clean up several lock files.  I want to double check `ALL `lock files from commands run from background maintenance jobs are cleaned up by maintenance steps.  If so the user could kill, but we should definitely just kill ourselves on unmount.

> * Will asking users to "quit" processes they did not start themselves cause confusion?

I bet it would :-)

jamill:
I think listing the process PID could be helpful to let the user know exactly which process is blocking the upgrade, and stands by itself.

I agree with @derrickstolee - I think the main issue we should focus on is with background git processes that VFSForGit kicks off. As we initiated those processes, I would expect VFSForGit to coordinate running upgrade with those processes and ensure that VFSForGit itself is not blocking upgrade. I would be less surprised if VFSForGit refused to upgrade because an outside git process is running.

Given this scope of focusing on process VFSForGit initiates, does that make the problem easier (can we possibly terminate our process that we know are safe to terminate, and clean up left behind files)?

Is there anything we can log in the unmount code path to indicate that the repository was unmounted, but it failed to terminate background git processes? Are there issues with the background git processes still running after the repository is unmounted?

jeschu1:
@jamill @wilbaker @derrickstolee we're all on the same page that scope here is background jobs.  My understanding is that all of the jobs we run in the background are safe to stop and that was what I was hoping to double check.  Speaking with @dscho he's concerned about option 1 (probably because folks have been burned several times by terminating jobs they shouldn't) so it's worth getting the details out there and verifying.

alameenshah:
Having a few questions...
1. After a forced kill, for vfsforgit processes to recover later (after the upgrade), would the user have to take further actions - like run `gvfs repair` or in the worst case, email support. Can the repair & recovery happen without user action?
2. If the recovery cannot happen automatically - should VFS4Git warn the user before it decides to kill its child processes. May be provide a `--force` switch in upgrade?
3. About the messaging (`Run `gvfs upgrade --confirm` again after quitting these processes - git pid:9116, git pid:16396`). 
3.1 Agree that we don’t want the user to quit processes launched by VFS4Git. Should probably rephrase the text to something like “Could not unmount all VFSForGit repositories. Manually unmount and re-run upgrade.” And don’t show VFS4Git children in the list (if needed, we can always capture detailed process listing in the upgrade logs.)
3.2 Also I’m not sure I understand the problem we are solving by showing `pid` numbers in the message. (If there is only one git instance running, Windows Task Manager would show just the one. If there are many git instances, the user would have to order the process listing by name and terminate all. In both cases `pid` is not being useful.)

jeschu1:
@alameenshah 
> 1. After a forced kill, for vfsforgit processes to recover later (after the upgrade), would the user have to take further actions - like run `gvfs repair` or in the worst case, email support. Can the repair & recovery happen without user action?
No need to run repair, these processes can be safely killed

> 2. If the recovery cannot happen automatically - should VFS4Git warn the user before it decides to kill its child processes. May be even provide a `--force` switch in upgrade?
Since they can be safely killed we're ok here.

> 3. About the messaging (`Run `gvfs upgrade --confirm` again after quitting these processes - git pid:9116, git pid:16396`).
>    3.1 Agree that we don’t want the user to quit processes launched by VFS4Git. When VFS4Git children are the ones blocking upgrade, may be rephrase the text to something like “Could not unmount all 
I've upgraded the latest PR to kill the background jobs so the problem won't exist.

VFSForGit repositories. Manually unmount and re-run upgrade”.
>    3.2 Also I’m not sure I understand the problem we are solving by showing `pid` numbers in the message. (If there is only one git instance running, Windows Task Manager would show just the one. If there are many git instances, the user would have to order the process listing by name and terminate all. In both cases `pid` is not being useful.)
We want to let the user know what exactly is blockly.  pid becomes important in case the processes disappears or in general to know how many instances are blocking.

Issue #769, opened 2019-02-09
Improve unsupported upgrade error message on RS1/Server


jrbriggs:
The error message when on RS1/Server could be improved.  Perhaps something like: 

> `gvfs upgrade` is only supported after the "Windows Projected File System" optional feature has been enabled by a manual upgrade of VFS for Git, and only on versions of Windows that support this feature. 
  
> Check your team's documentation for instructions on how to manually upgrade. 
 
The current message's use of "out of band" is a bit confusing on pre-RS4 machines, and users might think that by manually upgrading once they'll be able to start using the `gvfs upgrade` verb.

jrbriggs:
This is now in both branches.

PR #772, opened 2019-02-11
Rephrase ProjFS error messaging

alameenshah:
On Windows RS1/Server ProjFS is not inboxed. On those PCs the current `gvfs upgrade` pre-check error messaging is confusing. Updated the messaging.

Issue #769

### DIFF
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -190,7 +190,7 @@ namespace GVFS.Upgrader
             {
                 consoleError = string.Join(
                     Environment.NewLine,
-                    $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is not supported because you have previously installed an out of band ProjFS driver.",
+                    $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is only supported after the \"Windows Projected File System\" optional feature has been enabled by a manual installation of VFS for Git, and only on versions of Windows that support this feature.",
                     "Check your team's documentation for how to upgrade.");
                 return false;
             }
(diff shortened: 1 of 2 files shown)


jrbriggs:
This looks good. Please also cherry-pick this against milestones/M147.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeVerbTests.cs
```diff
@@ -142,12 +142,12 @@ public void ProjFSPreCheck()
                 expectedReturn: ReturnCode.GenericError,
                 expectedOutput: new List<string>
                 {
-                    "ERROR: `gvfs upgrade` is not supported because you have previously installed an out of band ProjFS driver.",
+                    "ERROR: `gvfs upgrade` is only supported after the \"Windows Projected File System\" optional feature",
```
nickgra: It looks like when we're comparing output/errors, we're testing if the expected output is a substring of the actual output (assuming this test passes). 

Before, we appear to be testing the full string. Is it intentional that expectedOutput behaves in this way? If we want this test to have the same behavior that it used to, we could update this string to be the full text from `InstallerPreRunChecker.cs`.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -188,9 +188,10 @@ private bool IsGVFSUpgradeAllowed(out string consoleError)
 
             if (!this.IsGVFSUpgradeSupported())
             {
+                string prjfsError = "is only supported after the \"Windows Projected File System\" optional feature has been enabled by a manual upgrade of VFS for Git, and only on versions of Windows that support this feature.";
```
jamill: manual upgrade? -> manual installation?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/UpgradeVerbTests.cs
jamill: after the optional feature has been enabled?

GVFS/GVFS.Common/InstallerPreRunChecker.cs
jamill: also - could string literal be inlined directly below (as it was before)? doesn't really matter, but if we are using it directly without modification, I think it is easier to read inline.

PR #773, opened 2019-02-11
Better feedback and fixes for Upgrade Scenario

### LINKED ISSUE #704: Upgrader fails to run due to git process

jeschu1:
This is a continuation of PR #752 , which was accidentally closed and can't be reopened.  Resolves #704

1. We kill the entire tree for background jobs.  

I used taskkill on windows and kill on mac.  I did try using the native libraries but in addition to being messier it didn't take out the entire tree.  If anyone is interested the branch with that approach is [here](https://github.com/jeschu1/VFSForGit/tree/upgrade_fail_native).

2. When running upgrade we do a seperate check to see if any git processes are running.  At this point the only clashers would be ones external to gvfs.

This
```
Downloading...Succeeded
Unmounting repositories...Failed
Mounting repositories...Succeeded

ERROR: Blocking processes are running.
Run `gvfs upgrade --confirm` again after quitting these processes - git
```
Becomes
```
Downloading...Succeeded
Unmounting repositories...Succeeded
Checking for running git processes....Failed
Mounting repositories...Succeeded

ERROR: Blocking processes are running.
Run `gvfs upgrade --confirm` again after quitting these processes - git pid:18360, git pid:46588`
```

Testing on Mac is still in progress...

### DIFF
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -58,6 +58,9 @@ namespace GVFS.Common
         public abstract Dictionary<string, string> GetPhysicalDiskInfo(string path, bool sizeStatsOnly);
 
         public abstract bool IsConsoleOutputRedirectedToFile();
+
+        public abstract bool TryKillProcessTree(int processId, out int exitCode, out string error);
+
         public abstract bool TryGetGVFSEnlistmentRoot(string directory, out string enlistmentRoot, out string errorMessage);
 
         public abstract bool IsGitStatusCacheSupported();
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -126,34 +126,33 @@ namespace GVFS.Common.Git
             return true;
         }
 
-        public bool WaitForRunningGitProcess()
+        /// <summary>
+        /// Tries to kill the run git process.  Make sure you only use this on git processes that can safely be killed!
+        /// </summary>
+        /// <param name="processName">Name of the running process</param>
+        /// <param name="exitCode">Exit code of the kill.  -1 means there was no running process.</param>
+        /// <param name="error">Error message of the kill</param>
+        /// <returns></returns>
+        public bool TryKillRunningProcess(out string processName, out int exitCode, out string error)
         {
             this.stopping = true;
+            processName = null;
+            exitCode = -1;
+            error = null;
 
-            try
+            lock (this.processLock)
             {
-                lock (this.processLock)
+                Process process = this.executingProcess;
+
+                if (process != null)
                 {
-                    Process process = this.executingProcess;
+                    processName = process.ProcessName;
 
-                    if (process != null)
-                    {
-                        process.WaitForExit();
-                    }
-
-                    return true;
+                    return GVFSPlatform.Instance.TryKillProcessTree(process.Id, out exitCode, out error);
                 }
-            }
-            catch (Win32Exception)
-            {
-                // Thrown when process is already terminating
-            }
-            catch (InvalidOperationException)
-            {
-                // Process already terminated
-            }
 
-            return false;
+                return true;
+            }
         }
 
         public virtual void RevokeCredential(string repoUrl)
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -126,7 +126,7 @@ namespace GVFS.Common.Git
             return true;
         }
 
-        public bool TryKillRunningProcess()
+        public bool WaitForRunningGitProcess()
         {
             this.stopping = true;
 
@@ -138,7 +138,7 @@ namespace GVFS.Common.Git
 
                     if (process != null)
                     {
-                        process.Kill();
+                        process.WaitForExit();
                     }
 
                     return true;
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -74,11 +74,9 @@ namespace GVFS.Upgrader
             return true;
         }
 
-        public virtual bool NoBlockingProcessCheck(out string consoleError)
+        public virtual bool IsInstallationBlockedByRunningProcess(out string consoleError)
         {
             consoleError = null;
-
-            this.tracer.RelatedInfo("Checking for running git processes.");
 
             // While checking for blocking processes like GVFS.Mount immediately after un-mounting,
             // then sometimes GVFS.Mount shows up as running. But if the check is done after waiting
@@ -87,7 +85,7 @@ namespace GVFS.Upgrader
             // actually quits.
             this.tracer.RelatedInfo("Checking if GVFS or dependent processes are running.");
             int retryCount = 10;
-            List<string> processList = null;
+            HashSet<string> processList = null;
             while (retryCount > 0)
             {
                 if (!this.IsBlockingProcessRunning(out processList))
@@ -105,7 +103,7 @@ namespace GVFS.Upgrader
                     Environment.NewLine,
                     "Blocking processes are running.",
                     $"Run {this.CommandToRerun} again after quitting these processes - " + string.Join(", ", processList.ToArray()));
-                this.tracer.RelatedError($"{nameof(this.TryUnmountAllGVFSRepos)}: {consoleError}");
+                this.tracer.RelatedError($"{nameof(this.IsInstallationBlockedByRunningProcess)}: {consoleError}");
                 return false;
             }
 
@@ -134,11 +132,11 @@ namespace GVFS.Upgrader
             return GVFSEnlistment.IsUnattended(this.tracer);
         }
 
-        protected virtual bool IsBlockingProcessRunning(out List<string> processes)
+        protected virtual bool IsBlockingProcessRunning(out HashSet<string> processes)
         {
             int currentProcessId = Process.GetCurrentProcess().Id;
             Process[] allProcesses = Process.GetProcesses();
-            List<string> matchingNames = new List<string>();
+            HashSet<string> matchingNames = new HashSet<string>();
 
             foreach (Process process in allProcesses)
             {
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -58,7 +58,6 @@ namespace GVFS.Upgrader
         public virtual bool TryUnmountAllGVFSRepos(out string consoleError)
         {
             consoleError = null;
-
             this.tracer.RelatedInfo("Unmounting any mounted GVFS repositories.");
 
             using (ITracer activity = this.tracer.StartActivity(nameof(this.TryUnmountAllGVFSRepos), EventLevel.Informational))
@@ -69,36 +68,45 @@ namespace GVFS.Upgrader
                     return false;
                 }
 
-                // While checking for blocking processes like GVFS.Mount immediately after un-mounting,
-                // then sometimes GVFS.Mount shows up as running. But if the check is done after waiting
-                // for some time, then eventually GVFS.Mount goes away. The retry loop below is to help
-                // account for this delay between the time un-mount call returns and when GVFS.Mount
-                // actually quits.
-                this.tracer.RelatedInfo("Checking if GVFS or dependent processes are running.");
-                int retryCount = 10;
-                HashSet<string> processList = null;
-                while (retryCount > 0)
-                {
-                    if (!this.IsBlockingProcessRunning(out processList))
-                    {
-                        break;
-                    }
-
-                    Thread.Sleep(TimeSpan.FromMilliseconds(250));
-                    retryCount--;
-                }
-
-                if (processList.Count > 0)
-                {
-                    consoleError = string.Join(
-                        Environment.NewLine,
-                        "Blocking processes are running.",
-                        $"Run {this.CommandToRerun} again after quitting these processes - " + string.Join(", ", processList.ToArray()));
-                    this.tracer.RelatedError($"{nameof(this.TryUnmountAllGVFSRepos)}: {consoleError}");
-                    return false;
-                }
-
                 activity.RelatedInfo("Successfully unmounted repositories.");
+            }
+
+            return true;
+        }
+
+        public virtual bool NoBlockingProcessCheck(out string consoleError)
+        {
+            consoleError = null;
+
+            this.tracer.RelatedInfo("Checking for running git processes.");
+
+            // While checking for blocking processes like GVFS.Mount immediately after un-mounting,
+            // then sometimes GVFS.Mount shows up as running. But if the check is done after waiting
+            // for some time, then eventually GVFS.Mount goes away. The retry loop below is to help
+            // account for this delay between the time un-mount call returns and when GVFS.Mount
+            // actually quits.
+            this.tracer.RelatedInfo("Checking if GVFS or dependent processes are running.");
+            int retryCount = 10;
+            List<string> processList = null;
+            while (retryCount > 0)
+            {
+                if (!this.IsBlockingProcessRunning(out processList))
+                {
+                    break;
+                }
+
+                Thread.Sleep(TimeSpan.FromMilliseconds(250));
+                retryCount--;
+            }
+
+            if (processList.Count > 0)
+            {
+                consoleError = string.Join(
+                    Environment.NewLine,
+                    "Blocking processes are running.",
+                    $"Run {this.CommandToRerun} again after quitting these processes - " + string.Join(", ", processList.ToArray()));
+                this.tracer.RelatedError($"{nameof(this.TryUnmountAllGVFSRepos)}: {consoleError}");
+                return false;
             }
 
             return true;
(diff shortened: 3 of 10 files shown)


GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -332,6 +332,12 @@ public override bool TryGetGVFSEnlistmentRoot(string directory, out string enlis
             return WindowsPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
         }
 
+        public override bool KillProcessTree(int processId)
```
jamill: Should we log that we are killing a process here (or in the caller)? Would the name of the process we are killing be useful if we have to look through the logs?

jeschu1: It's always going to 'git' so I wouldn't say the name is useful honestly.  Do log weather the kill was a success or not.

jrbriggs:
Fixes #704

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
jamill: This is in the `WindowsPlatform` code, which does not control what process is being killed, and so it could be any `pid`… The only known caller today might only do it for `git.exe`, but this could change in the future. My other thought was that it would be nice to know when we are killing a process - it shouldn't add too much overhead to the logging, as we don't do it very often.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -159,5 +159,11 @@ public override bool IsGitStatusCacheSupported()
         {
             return new MacFileBasedLock(fileSystem, tracer, lockPath);
         }
+
+        public override bool KillProcessTree(int processId)
+        {
+            ProcessResult result = ProcessHelper.Run("kill", $"{processId}");
```
wilbaker: How did you test this on Mac?

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -57,7 +57,10 @@ public static void Register(GVFSPlatform platform)
 
         public abstract Dictionary<string, string> GetPhysicalDiskInfo(string path, bool sizeStatsOnly);
 
-        public abstract bool IsConsoleOutputRedirectedToFile();
+        public abstract bool IsConsoleOutputRedirectedToFile();
+
+        public abstract bool KillProcessTree(int processId);
```
wilbaker: Make this `void` instead of `bool`? It doesn't look like any callers are checking the return value.

(If you leave it `bool` consider renaming to `TryKillProcessTree`)

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -138,7 +138,7 @@ public bool TryKillRunningProcess()
 
                     if (process != null)
                     {
-                        process.Kill();
+                        GVFSPlatform.Instance.KillProcessTree(process.Id);
```
alameenshah: The idea was to `kill` only background processes launched by VFS4Git, right? I don't see a check.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -69,36 +68,45 @@ public virtual bool TryUnmountAllGVFSRepos(out string consoleError)
...
+                activity.RelatedInfo("Successfully unmounted repositories.");
+            }
 
-                    Thread.Sleep(TimeSpan.FromMilliseconds(250));
-                    retryCount--;
-                }
+            return true;
+        }
+
+        public virtual bool NoBlockingProcessCheck(out string consoleError)
```
wilbaker: Naming suggestion:  Without reading the method implementation it was not clear to me what `NoBlockingProcessCheck` does.

Consider renaming to something like `IsInstallationBlockedByRunningProcess` or `IsBlockingProcessRunningAfterWait`

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -69,36 +68,45 @@ public virtual bool TryUnmountAllGVFSRepos(out string consoleError)
...
 
-                if (processList.Count > 0)
+            this.tracer.RelatedInfo("Checking for running git processes.");
+
+            // While checking for blocking processes like GVFS.Mount immediately after un-mounting,
+            // then sometimes GVFS.Mount shows up as running. But if the check is done after waiting
+            // for some time, then eventually GVFS.Mount goes away. The retry loop below is to help
+            // account for this delay between the time un-mount call returns and when GVFS.Mount
+            // actually quits.
+            this.tracer.RelatedInfo("Checking if GVFS or dependent processes are running.");
```
wilbaker: Nit:

Can one of these two be removed (maybe the first since we're not checking only for git)?

```
this.tracer.RelatedInfo("Checking for running git processes.");
this.tracer.RelatedInfo("Checking if GVFS or dependent processes are running.");
```

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -69,36 +68,45 @@ public virtual bool TryUnmountAllGVFSRepos(out string consoleError)
...
+                retryCount--;
+            }
+
+            if (processList.Count > 0)
+            {
+                consoleError = string.Join(
+                    Environment.NewLine,
+                    "Blocking processes are running.",
+                    $"Run {this.CommandToRerun} again after quitting these processes - " + string.Join(", ", processList.ToArray()));
+                this.tracer.RelatedError($"{nameof(this.TryUnmountAllGVFSRepos)}: {consoleError}");
```
wilbaker: `nameof(this.TryUnmountAllGVFSRepos)` -> `nameof(NoBlockingProcessCheck)`

GVFS/GVFS.Platform.Mac/MacPlatform.cs
alameenshah: This does not kill the tree, you wanna take a look at [`pkill`](https://unix.stackexchange.com/questions/124127/kill-all-descendant-processes).

wilbaker:
> 1. We kill the entire tree for background jobs.

@jeschu1 was GVFS.Mount already calling `GitProcess.TryKillRunningProcess` before this PR?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
jeschu1: I updated it to carry thru processName, exitCode, and error message.  This should give us all the info we need.

GVFS/GVFS.Common/Git/GitProcess.cs
jeschu1: @alameenshah the process is only called from GitMaintenanceStep.  This a buyer beware type of process.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
jeschu1: Thanks updated!

jeschu1: I left a comment on top saying "Testing in progress for Mac".  I actually had to switch to "pkill -P" to make this properly work.  

To answer your question though I switched the background job to push a large file (so it would take a while) and then ensured the push was killed on unmount.  Push spawns several children so it's a good test case.

jeschu1:
> > 1. We kill the entire tree for background jobs.
> 
> @jeschu1 was GVFS.Mount already calling `GitProcess.TryKillRunningProcess` before this PR?

Yes, but it only did a process.Kill();.  I.e. didn't kill the entire tree

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -219,6 +219,25 @@ private bool TryRunUpgrade(out Version newVersion, out string consoleError)
                 return false;
             }
 
+            if (!this.LaunchInsideSpinner(
```
jamill: Mainly curious - what do we get by breaking this out into its own step now? Is it just more granular messaging about what the upgrader is doing? Should it be more generic `Checking for blocking processes`, or even `Performing pre-upgrade checks`? alternatively should we keep with the existing top level messages and minimize change for 147? I do think it was a bit overloaded that we (used to) perform this check as part of unmounting repositories, so I think this fits better.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockInstallerPreRunChecker.cs
```diff
@@ -72,9 +72,9 @@ protected override bool IsUnattended()
             return this.FakedResultOfCheck(FailOnCheckType.UnattendedMode);
         }
 
-        protected override bool IsBlockingProcessRunning(out HashSet<string> processes)
+        protected override bool IsBlockingProcessRunning(out List<string> processes)
         {
-            processes = new HashSet<string>();
+            processes = new List<string>();
```
jamill: Is there a reason why you changed from `List` to `HashSet`? or was this just a general change?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -126,34 +126,26 @@ public static bool TryGetVersion(string gitBinPath, out GitVersion gitVersion, o
             return true;
         }
 
-        public bool TryKillRunningProcess()
+        public bool TryKillRunningProcess(out string processName, out int exitCode, out string error)
         {
             this.stopping = true;
+            processName = string.Empty;
```
jamill: I think other places in the project generally return `null` when error is not set. What does a default exit code of `-1` indicate?

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
jeschu1: `Checking for blocking processes` makes more sense than you!  If you look at the title we were putting blocking processes as a failure for unmounting repos.  That wasn't actually an accurate description of what was happening.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockInstallerPreRunChecker.cs
jeschu1: I can switch it back to HashSet.  Since we give the pid they're unique.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
jeschu1: To elaborate previously we told the user "Unmounting failed" when that isn't what happened (see summary at top for example).  That is likely to be confusing to a user.  Now we let the user know unmounting was successful but they have git processes that are blocking.  Since I moved existing code I think it's a safe change we should take.

GVFS/GVFS.Common/Git/GitProcess.cs
jeschu1: I was using it to indicate it hasn't been set.  The exit codes for the calling processes are >=0.  If you've got a better suggestion let me know!

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
jamill: Yeah - I think it makes more sense with your changes. Thanks!

GVFS/GVFS.Common/Git/GitProcess.cs
jamill: I think we use non-zero to indicate error in some places - at least it looks like at least some places in GitHubUpgrader do. I think `0` indicates no error, and so that is what I would have expected on initial read.

If we keep with `-1`, maybe add a method comment with how consumers should interpret the results from calling this method - they might only expect values >= 0?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -176,8 +186,8 @@ protected void SaveLastRunTimeToFile()
             {
                 this.Context.Tracer.RelatedError(this.CreateEventMetadata(handledException), "Failed to record run time");
             }
-        }
```
jamill: Are these CRLF changes?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -190,8 +200,8 @@ protected void LogErrorAndRewriteMultiPackIndex(ITracer activity, GitProcess.Res
             errorMetadata["RewriteResultExitCode"] = rewriteResult.ExitCode;
 
             activity.RelatedError(errorMetadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
-        }
-
```
jamill: CRLF changes?

GVFS/GVFS.UnitTests/Git/GitProcessTests.cs
```diff
@@ -12,7 +12,7 @@ public class GitProcessTests
         public void TryKillRunningProcess_NeverRan()
         {
             GitProcess process = new GitProcess(new MockGVFSEnlistment());
-            process.TryKillRunningProcess().ShouldBeTrue();
+            process.TryKillRunningProcess(out string processName, out int exitCode, out string error).ShouldBeTrue();
```
jamill: Is there anything we can assert about the out parameters here?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -332,6 +332,14 @@ public override bool TryGetGVFSEnlistmentRoot(string directory, out string enlis
             return WindowsPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
         }
 
+        public override bool TryKillProcessTree(int processId, out int exitCode, out string error)
+        {
+            ProcessResult result = ProcessHelper.Run("taskkill", $"/pid {processId} /f /t");
```
wilbaker: Do taskkill and pkill return success if the specified process ID is no longer running?

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -219,6 +219,25 @@ private bool TryRunUpgrade(out Version newVersion, out string consoleError)
...
 
+            if (!this.LaunchInsideSpinner(
+                () =>
+                {
+                    if (!this.preRunChecker.IsInstallationBlockedByRunningProcess(out error))
+                    {
+                        return false;
+                    }
+
+                    this.mount = true;
```
wilbaker: Why is `this.mount` being set to `true`?  Won't it already be set to `true` on line 211?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
jeschu1: They do not.  I can update them both to return success for those exit codes.   You agree that's worth changing?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -79,18 +79,28 @@ public void Stop()
...
+                                "{0}: killed background process {1} during {2}",
+                                this.Area,
+                                processName,
+                                nameof(this.Stop)),
                             metadata: null);
                     }
                     else
                     {
                         this.Context.Tracer.RelatedEvent(
                             EventLevel.Informational,
```
jamill: Should this be a warning level?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
wilbaker: Oh actually I think I misread the old code.  I thought it was treating already shutting down as success, but now that I take another look at `TryKillRunningProcess` it was returning false in that scenario:

```
            catch (Win32Exception)	
            {	
                // Thrown when process is already terminating	
            }	
            catch (InvalidOperationException)	
            {	
                // Process already terminated	
            }	


            return false;
```

So I think we're OK

jamill: Or maybe have the caller interpret the error code and report the process did not exist? (as opposed to a failure...)?

jeschu1: Considering we send the exitCode / error code back, i think it's fine to fail here.  This is part of shut down so the only difference would be that extra logging.  If we need something fancier in the future we can definitely do that.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: Not seeing why this is a diff :-(.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: Not seeing why this is a diff :-(.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: It wasn't before, and I don't think we need it necessarily in telemetry.

jamill: Do warnings get reported to telemetry by default? I thought that was only `RelatedError`...

I think `Warning` would be appropriate for events that might be unexpected and could indicate a problem.

wilbaker: > Do warnings get reported to telemetry by default? I thought that was only `RelatedError`

That's correct, warnings will not be reported to telemetry by default

jrbriggs:
Thanks @jeschu1

PR #775, opened 2019-02-11
Improve upgrader logging

jamill:
This includes several enhancements to the logging performed by the upgrader:

1) Logs are now written to an self contained directory per upgrade run. This groups all related log files together, and separates them from other runs.

2) The NuGet installer adds extra context to error messages when an install action fails. Previously, only the error text from the installer was included. This caused problems, especially when the install command did not generate error text.

3) When an upgrade fails, we now include instructions on where to find the log files.

Previous:

image

With changes:

image


GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -70,6 +70,8 @@ protected ProductUpgrader()
         {
         }
 
+        public string InstallerLogsDirectory { get; set; } = ProductUpgraderInfo.LogDirectory;
```
wilbaker: What is the reason for the `= ProductUpgraderInfo.LogDirectory;`?

`ProductUpgraderInfo.LogDirectory` is simply "Logs", and it appears that this property is supposed to be a path (you might even consider renaming to `InstallerLogsDirectoryPath` to make that clearer).

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
```diff
@@ -320,12 +320,13 @@ public void TestUpgradeAllowed()
         [TestCase]
         public void WellKnownArgumentTokensReplaced()
         {
+            string logDirectory = ProductUpgraderInfo.GetLogDirectoryPath();
```
wilbaker: Switch this to using a `mock:\` path instead of `ProductUpgraderInfo.GetLogDirectoryPath`? 

In my experience using mock paths helps prevent us from accidentally hitting the disk during unit tests (and helps find spot's we missed using `PhysicalFileSystem` in the product code).

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -434,8 +434,11 @@ public override bool TryRunInstaller(InstallActionWrapper installActionWrapper,
 
                 if (!installSuccessful)
                 {
-                    activity.RelatedError($"Could not complete all install actions. The following error was encountered: {localError}");
-                    error = localError;
+                    error = string.IsNullOrEmpty(localError) ?
+                        $"The {currentInstallAction.Name} failed, but no error message was provided by the failing command." :
```
wilbaker: If an exception is thrown before `currentInstallAction` is set this we'll have a null reference exception here.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -141,9 +151,13 @@ public void Execute()
 
         private JsonTracer CreateTracer()
         {
+            string upgradeInstanceId = DateTime.Now.ToString("yyyyMMdd_HHmmss");
+            this.logDirectory = Path.Combine(ProductUpgraderInfo.GetLogDirectoryPath(), upgradeInstanceId);
```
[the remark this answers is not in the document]
jamill: I tweaked this a bit to find the newest directory, and then the newest log file in the directory. However, the `UpgradeVerb` logs are now no longer considered in this search, as they are written to the root of the logs folder... I will come back to this.

GVFS/GVFS.Common/ProductUpgrader.cs
jamill: This was meant to be the full path to the log directory.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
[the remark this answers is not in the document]
jamill: I landed on a design that keeps all the logs in the same top level logs directory, but with the same "installation id". I think this is a good compromise between how the existing code expects logs to be laid out and letting the user find relevant logs.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -114,8 +119,16 @@ public void Execute()
...
-                    error = Environment.NewLine + "ERROR: " + error;
-                    this.output.WriteLine(error);
+                    StringBuilder sb = new StringBuilder();
+                    sb.AppendLine();
+                    sb.Append("ERROR: " + error);
+
+                    sb.AppendLine();
+                    sb.AppendLine();
+
+                    sb.AppendLine($"Installation logs can be found at: {this.logDirectory} with the installation id: {this.installationId}");
```
kewillford: Why can't this be the file name that information will be saved to?  Will the user even know what an "installation id" is and what is it used for and that they need to look for a file with that in the name?

jamill: The upgrader can generate multiple logs - in particular - each installer that is run can generate there own set of set of logs. My thought was the installation ID will be obvious enough that when a user goes and looks at this directory, they will be able to find the relevent files. Maybe we could just direct the user to the single top level log file, but then that doesn't include any hints about the other possible installer logs. 

The previous approach was to write all logs to a new directory for each upgrade. This had the advantage of pointing the user to a single directory with relevent logs, but the downside of the log related logic not being set up to deal with sub directories.

I think the current approach is a good compromise - but we can tweak this as needed.

wilbaker: I think Kevin brings up a good point that users might not know what an "installation id" is.

If we want to stick with this approach I think the experience could be improved by a small change to the message.  Rather than:

> with the installation id: {this.installationId}

If we said something like:

> with file names that end with {this.installationId}

Thoughts?

Alternatively, if you wanted to keep the directory structure perhaps `gvfs log` could search all files in the logs directories (including those in subfolders) and return the latest?

jamill: > with file names that end with installation id {this.installationId}

I prefer this approach for now (or a similar wordsmithing) rather than tweaking the directory structure of how we write out logs for now. `installation id` is kind of an arbitrary name. I could also do:

`file names with the suffix {this.installationId}`?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -186,11 +180,9 @@ public static bool TryCreateAzDevOrgUrlFromPackageFeedUrl(string packageFeedUrl,
         /// <param name="src">The unprocessed string to use as arguments to an install command</param>
         /// <param name="installationId">A unique installer ID to replace the installer_id token with.</param>
         /// <returns>The argument string with tokens replaced.</returns>
-        public static string ReplaceArgTokens(string src, string installationId)
+        public static string ReplaceArgTokens(string src, string installationId, string logsDirectory)
```
wilbaker: Is this change still needed? Now that we're back to putting everything in the same directory will `logsDirectory` go back to always being `ProductUpgraderInfo.GetLogDirectoryPath()`?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -186,11 +180,9 @@ public static bool TryCreateAzDevOrgUrlFromPackageFeedUrl(string packageFeedUrl,
         /// <param name="src">The unprocessed string to use as arguments to an install command</param>
         /// <param name="installationId">A unique installer ID to replace the installer_id token with.</param>
         /// <returns>The argument string with tokens replaced.</returns>
-        public static string ReplaceArgTokens(string src, string installationId)
+        public static string ReplaceArgTokens(string src, string installationId, string logsDirectory)
         {
-            string logDirectory = ProductUpgraderInfo.GetLogDirectoryPath();
-
-            string dst = src.Replace(NuGetUpgrader.replacementToken(InstallActionInfo.ManifestEntryLogDirectoryToken), logDirectory)
+            string dst = src.Replace(NuGetUpgrader.replacementToken(InstallActionInfo.ManifestEntryLogDirectoryToken), logsDirectory)
```
wilbaker: Not part of this PR, but what does dst stand for?

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
```diff
@@ -320,12 +320,13 @@ public void TestUpgradeAllowed()
         [TestCase]
         public void WellKnownArgumentTokensReplaced()
         {
+            string logDirectory = "mock://test_log_directory";
```
wilbaker: Nit: `mock://` -> `mock:\\` (Windows style paths)

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -14,6 +15,8 @@ public class UpgradeOrchestrator
         private const EventLevel DefaultEventLevel = EventLevel.Informational;
 
         private ProductUpgrader upgrader;
+        private string logDirectory = ProductUpgraderInfo.GetLogDirectoryPath();
```
wilbaker: Should we keep `logDirectory` around?  If it's always going to be `ProductUpgraderInfo.GetLogDirectoryPath()` we could go back to using that directly.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: src -> dst : source -> destination

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
jamill: Yeah - that is an option. My thinking was that instead of sprinkling the calls to `ProductUpgraderInfo.GetLogDirectoryPath` around in multiple places, I could consolidate this to a single instance variable, and reduce calls to static methods. It is not strictly needed for this change (but now we do use the `logDirectory` in multiple places.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: not technically needed - I thought it was a better fit, as now the caller is responsible for providing all the replacement tokens, rather than specifing 1, and having another one being handled automatically.

wilbaker: Ok sounds good, thanks!

PR #786, opened 2019-02-13
Listen to VNODE_WRITE to trigger modified files

jeschu1:
Moved the modified file trigger from KAUTH_FILEOP_CLOSE/KAUTH_FILEOP_CLOSE_MODIFIED to KAUTH_VNODE_WRITE_DATA.

This means whenever authorization to write is requested we'll trigger a modified event.
In testing this doesn't seem excessive, different IDEs act differently
- TextEdit requests authorization once on open (1 event)
- Sublime/VSCode requests authorization for each save (event per save)
- VS/XCode take the approach of renaming temp files and deleting (not affected by change)

This solves the issue where a file may be written to and the descriptor isn't closed by the process.

Address half of #209

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -499,6 +499,30 @@ static int HandleVnodeOperation(
                         kauthError))
                 {
                     goto CleanupAndReturn;
+                }                
+            }
+            
+            if(ActionBitIsSet(action, KAUTH_VNODE_WRITE_DATA))
+            {
+                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                {
+                    goto CleanupAndReturn;
+                }
+                
+                PerfSample fileModifiedSample(&perfTracer, PrjFSPerfCounter_VnodeOp_FileModified);
+                
+                if (!TrySendRequestAndWaitForResponse(
+                                                      root,
+                                                      MessageType_KtoU_NotifyFileModified,
+                                                      currentVnode,
+                                                      vnodeFsidInode,
+                                                      vnodePath,
+                                                      pid,
+                                                      procname,
+                                                      &kauthResult,
+                                                      kauthError))
+                {
+                    goto CleanupAndReturn;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -499,6 +499,30 @@ static int HandleVnodeOperation(
...
+                if (!TrySendRequestAndWaitForResponse(
+                                                      root,
+                                                      MessageType_KtoU_NotifyFileModified,
+                                                      currentVnode,
+                                                      vnodeFsidInode,
+                                                      vnodePath,
+                                                      pid,
+                                                      procname,
+                                                      &kauthResult,
+                                                      kauthError))
```
pmj: I'm very much nit-picking here, but this isn't the whitespace convention for multi-line function calls we've been following in the code so far. Rather than aligning arguments with the opening parenthesis, we've simply been indenting them by one or more regular indentation steps. As part of an `if` condition, we've been indenting the arguments by an extra 1-2 levels, depending on how it looks I guess. If it's not in an `if`, one indentation level will usually do.

I'm assuming Xcode added the whitespace for you. If I remember correctly, the setting to turn off for this is Preferences… -> Text Editing -> Indentation -> "Automatically indent based on syntax". I don't think you can turn it off on a per-project or per-file basis. (The setting works reasonably well with Objective-C code, but generates unreasonable amounts of whitespace in C and C++ code.)

jeschu1: Good catch and tip!  Yes I should definitely be sticking to convention.

jeschu1:
@pmj thanks for the quick review and feedback!  I had a conversation offline with @wilbaker and he gave me some more info to incorporate.  @jamill  for reference

On the windows side we have 2 events:
- PreconvertToFull
  - This gets called when we move a file from Placeholders to ModifiedFiles.dat
  - This is run when we ask for authorization to write and have a placeholder
- ModifiedFile
  - This triggers a StatusCache run
  - This is run on CLOSE if the file has changed

The problem becomes if we run StatusCache on authorization (before the file actually changed) it will see no changes and cache that.  By doing it on CLOSE we know the changes are saved on disk and therefore StatusCache will rerun and pick them up.

Right now we don't have the equivalent of PreconvertToFull on Mac and as a result we have files multiple times in Placeholders.dat.  I'm going to do the change to add this functionality matching windows.  This will cause it to show up when running git status.

Where can this go wrong?
- StatusCache is run showing no change for file A.txt
- A.txt is written to but the file descriptor is not closed (hence no new event sent)
- `git status` is run giving the cached result

Right now this is theoretical problem, that we believe could also happen on Windows.  If we go live with StatusCache on Mac and have issues this would be one to watch.

pmj:
Thanks for the explanation - makes sense to me. Another weird edge case is if a file descriptor is duplicated, either with dup() or due to a fork(). Then you’ll actually end up with 2 (or more) close notifications. I’m also not sure what happens if you open(), then mmap() some part of the file, call close() and then write to the memory-mapped region. Do we get the close notification when unmapping the memory region or already when calling close?

These edge cases seem like something we might be able to replicate via functional tests, then we could see what effect they have in practice.

jeschu1:
@pmj thanks for the additional info!  Can you elaborate on how FSEvents would help?

open(), mmap(), close(), mmap parts of the file seems like an interesting case.  I'm not sure what the OS does behind the scenes there but certainly a interesting one to test.  I think moving events as described makes sense, but there may be more movements needed to cover more cases.

pmj:
> Can you elaborate on how FSEvents would help?

As far as I'm aware, FSEvents is what underpins the built-in Time Machine mechanism - so that TM doesn't have to do a full file system rescan every time it makes a backup, it just sits there gathering modification events and uses that info to perform an incremental backup according to its schedule.

I don't think it tells us anything new, and doesn't help us with various other parts of VFS4G, but essentially, if we're finding that file modifications are slipping through the cracks when tracking them via kauth listeners, we should be able to get a rock-solid data set of modified files from FSEvents.

(There is also the woefully-underdocumented but *supported* (as in, Apple DTS has recommended using it) audit infrastructure, which may give us finer-grained event data.)

pmj:
> I'm not sure what the OS does behind the scenes there but certainly a interesting one to test.

From what I remember, it keeps the `fileproc` object alive until all the mappings are unmapped; so I guess the question is if that's what triggers the FILEOP or if it's the actual `close()` syscall. I don't even know which way I'd go if I had to guess, the kauth listeners are such a badly thought-out API, I wouldn't be surprised if it did the wrong thing in this case too. 🙃

jeschu1:
Closing as this will be replaced by a new PR

PR #787, opened 2019-02-13
Stop reading ring info from Nuget feedurl

alameenshah:
Reverting commit c4af8726f4ad02e0a7f414814f8051de7d8d023e. 

Service will only use ring info specified in upgrade.ring config setting. Service will not use ring info specified in nuget feed url (even if it contains one.) Also removed the associated UT cases.


jrbriggs:
@alameenshah @jamill do we need to make any changes in how we decide if the nuget upgrader is configured? The `IsReady` and `IsConfigured` checks?

alameenshah:
> @alameenshah @jamill do we need to make any changes in how we decide if the nuget upgrader is configured? The `IsReady` and `IsConfigured` checks?

No changes needed.

jamill:
I validated this works as expected as well

PR #788, opened 2019-02-14
Linux: remove unneeded virtualization root setup from API

chrisd8088:
We expect any Linux implementation to virtualize using a dedicated mounted filesystem, so we do not need to flag directories within the global file hierarchy as being virtualized.  Therefore we can remove the `ConvertDirectoryToVirtualizationRoot()` function from the Linux API, and just return true from any caller (e.g., in MirrorProvider).

We can also remove the pair of associated error code which pertain to the setup of a virtualization root.

### DIFF
--- a/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
+++ b/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
@@ -12,10 +12,8 @@ namespace MirrorProvider.Linux
 
         public override bool TryConvertVirtualizationRoot(string directory, out string error)
         {
-            Result result = VirtualizationInstance.ConvertDirectoryToVirtualizationRoot(directory);
-
-            error = result.ToString();
-            return result == Result.Success;
+            error = null;
+            return true;
         }
 
         public override bool TryStartVirtualizationInstance(Enlistment enlistment, out string error)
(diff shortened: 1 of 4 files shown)


chrisd8088:
Hi @jrbriggs -- I can't see any reason why these Linux-specific edits would affect the functional tests, so I *think* the failing check is due to an infrastructure timeout of some kind while the Windows VFS4Git installer is running.  (The Azure logs just say "GVFS-github-1 lost communication with the server. Verify the machine is running and has a healthy network connection.") 

Is there any way to dig a little further, or rerun that check?  Or is it possible that there's a recent change to the installer and/or pipeline setup such that we would need to rebase the `features/linuxprototype` branch against master again?  Any thoughts or advice appreciated!  Thanks very much,

Chris.

pmj:
> Is there any way to […] rerun that check?

I've just re-queued it for you. If you follow the 'details' link, then make sure you're signed into Azure dev, there's a "Rebuild" option under the "…" menu in the top-right corner.

I don't have any insight into the Windows CI side of things beyond that, so I can't answer any of your other questions.

MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -12,10 +12,8 @@ public class LinuxFileSystemVirtualizer : FileSystemVirtualizer
 
         public override bool TryConvertVirtualizationRoot(string directory, out string error)
         {
-            Result result = VirtualizationInstance.ConvertDirectoryToVirtualizationRoot(directory);
-
-            error = result.ToString();
-            return result == Result.Success;
+            error = Result.Success.ToString();
```
jrbriggs: For consistency, this should probably be `error = null`

chrisd8088: Gotcha, thanks!

chrisd8088:
Thanks, @pmj -- the key takeaway for me there was that I need to be logged into Azure dev, which of course makes sense!

chrisd8088:
I think this is OK now, @jrbriggs -- thanks for the careful review!

PR #791, opened 2019-02-14
Mac: Add PreConvertToFull when Write Access is requested

jeschu1:
Add PreConvertToFull Event when Write Access is requested
- This will mark the file as D in placeholders.dat and insert into modifiedPaths.dat
- This cleans up the existing ModifiedFile event which was also being used as the initial switch from placeholders.dat

Addresses #209

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -218,6 +218,7 @@ namespace GVFS.Platform.Mac
             this.virtualizationInstance.OnNewFileCreated = this.OnNewFileCreated;
             this.virtualizationInstance.OnFileRenamed = this.OnFileRenamed;
             this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
+            this.virtualizationInstance.OnFilePreConvertToFull = this.NotifyFilePreConvertToFull;
 
             uint threadCount = (uint)Environment.ProcessorCount * 2;
 
@@ -437,23 +438,17 @@ namespace GVFS.Platform.Mac
                 {
                     this.OnDotGitFileOrFolderChanged(relativePath);
                 }
-                else
-                {
-                    // TODO(Mac): As a temporary work around (until we have a ConvertToFull type notification) treat every modification
-                    // as the first write to the file
-                    bool isFolder;
-                    string fileName;
-                    bool isPathProjected = this.FileSystemCallbacks.GitIndexProjection.IsPathProjected(relativePath, out fileName, out isFolder);
-                    if (isPathProjected)
-                    {
-                        this.FileSystemCallbacks.OnFileConvertedToFull(relativePath);
-                    }
-                }
             }
             catch (Exception e)
             {
                 this.LogUnhandledExceptionAndExit(nameof(this.OnFileModified), this.CreateEventMetadata(relativePath, e));
             }
+        }
+
+        private Result NotifyFilePreConvertToFull(string relativePath)
+        {
+            this.OnFilePreConvertToFull(relativePath);
+            return Result.Success;
         }
 
         private Result OnPreDelete(string relativePath, bool isDirectory)
--- a/GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
@@ -277,6 +277,24 @@ namespace GVFS.Virtualization.FileSystem
             }
         }
 
+        protected void OnFilePreConvertToFull(string relativePath)
+        {
+            try
+            {
+                bool isFolder;
+                string fileName;
+                bool isPathProjected = this.FileSystemCallbacks.GitIndexProjection.IsPathProjected(relativePath, out fileName, out isFolder);
+                if (isPathProjected)
+                {
+                    this.FileSystemCallbacks.OnFileConvertedToFull(relativePath);
+                }
+            }
+            catch (Exception e)
+            {
+                this.LogUnhandledExceptionAndExit(nameof(this.OnFilePreConvertToFull), this.CreateEventMetadata(relativePath, e));
+            }
+        }
+
         protected EventMetadata CreateEventMetadata(
             Guid enumerationId,
             string relativePath = null,
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -17,6 +17,7 @@
 #include "kernel-header-wrappers/mount.h"
 #include "KextLog.hpp"
 #include "ProviderMessaging.hpp"
+#include "public/PrjFSXattrs.h"
 
 #ifdef KEXT_UNIT_TESTING
 #include "KauthHandlerTestable.hpp"
@@ -383,6 +384,38 @@ static int HandleVnodeOperation(
                     goto CleanupAndReturn;
                 }
             }
+            
+            if (ActionBitIsSet(action, KAUTH_VNODE_WRITE_DATA))
+            {
+                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                {
+                    goto CleanupAndReturn;
+                }
+                
+                PrjFSFileXAttrData rootXattr = {};
+                SizeOrError xattrResult = Vnode_ReadXattr(currentVnode, PrjFSFileXAttrName, &rootXattr, sizeof(rootXattr));
+                if (xattrResult.error == ENOATTR)
+                {
+                    // Only notify if the file is still a placeholder
+                    goto CleanupAndReturn;
+                }
+                
+                PerfSample preConvertToFullSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreConvertToFull);
+                
+                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                                                      root,
+                                                      MessageType_KtoU_NotifyFilePreConvertToFull,
+                                                      currentVnode,
+                                                      vnodeFsidInode,
+                                                      nullptr, // path not needed, use fsid/inode,
+                                                      pid,
+                                                      procname,
+                                                      &kauthResult,
+                                                      kauthError))
+                {
+                    goto CleanupAndReturn;
+                }
+            }
         }
     }
     
@@ -593,7 +626,7 @@ static int HandleFileOpOperation(
         {
             goto CleanupAndReturn;
         }
-        
+
         char procname[MAXCOMLEN + 1];
         proc_name(pid, procname, MAXCOMLEN + 1);
         PerfSample fileModifiedSample(&perfTracer, PrjFSPerfCounter_FileOp_FileModified);
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -701,6 +701,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
         case MessageType_KtoU_NotifyFileModified:
         case MessageType_KtoU_NotifyFilePreDelete:
         case MessageType_KtoU_NotifyDirectoryPreDelete:
+        case MessageType_KtoU_NotifyFilePreConvertToFull:
         {
             char fullPath[PrjFSMaxPath];
             CombinePaths(s_virtualizationRootFullPath.c_str(), request.path, fullPath);
@@ -1011,7 +1012,7 @@ static PrjFS_Result HandleFileNotification(
 #endif
     
     PrjFSFileXAttrData xattrData = {};
-    bool partialFile = TryGetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData);
+    bool placeholderFile = TryGetXAttr(fullPath, PrjFSFileXAttrName, sizeof(PrjFSFileXAttrData), &xattrData);
 
     PrjFS_Result result = s_callbacks.NotifyOperation(
         0 /* commandId */,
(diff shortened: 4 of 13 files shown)


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -501,6 +503,38 @@ static int HandleVnodeOperation(
...
+            if (ActionBitIsSet(action, KAUTH_VNODE_WRITE_DATA))
+            {
+                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                {
+                    goto CleanupAndReturn;
+                }
+                
+                PrjFSFileXAttrData rootXattr = {};
+                SizeOrError xattrResult = Vnode_ReadXattr(currentVnode, PrjFSFileXAttrName, &rootXattr, sizeof(rootXattr));
+                if (xattrResult.error == ENOATTR)
```
kewillford: For my own understanding, when does the extended attribute get removed from the file?

pmj: It's [removed when the provider receives a modified file notification](https://github.com/Microsoft/VFSForGit/blob/ed362aec65f790f26a2a1f1c593f7602687388f3/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L992).

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -501,6 +503,38 @@ static int HandleVnodeOperation(
...
+                {
+                    // Only notify if the file is still a placeholder
+                    goto CleanupAndReturn;
+                }
+                
+                PerfSample enumerateDirectorySample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreConvertToFull);
```
pmj: Nit-picking: `enumerateDirectorySample` is probably a copy-pasted name and not really appropriate here.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -501,6 +503,38 @@ static int HandleVnodeOperation(
                     goto CleanupAndReturn;
                 }
             }
+            
+            if (ActionBitIsSet(action, KAUTH_VNODE_WRITE_DATA))
```
pmj: I'm not 100% clear on when exactly we need the `MessageType_KtoU_NotifyPreConvertToFull` notification so this could be intentional, but I notice that if the file is initially empty and we get a `KAUTH_VNODE_WRITE_DATA` authorisation callback, we will go through the logic for hydrating the file (tell the provider `MessageType_KtoU_HydrateFile`) just above this new case. Would we in this case not *also* want to send the `MessageType_KtoU_NotifyPreConvertToFull` message once the file is hydrated? Currently that won't happen due to the `goto CleanupAndReturn;` above.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -985,7 +986,7 @@ static PrjFS_Result HandleFileNotification(
         notificationType,
         nullptr /* destinationRelativePath */);
     
-    if (partialFile && PrjFS_NotificationType_FileModified == notificationType)
+    if (partialFile && PrjFS_NotificationType_PreConvertToFull == notificationType)
     {
         // PrjFS_NotificationType_FileModified is a post-modified FileOp event (that cannot be stopped
```
pmj: More nit-picking: please update the comment to reflect the code change.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
kewillford: No, we don't because that will cause git to start tracking the file and in the case of only reading a file we don't want git to start tracking it as it will greatly impact the performance of git commands.
We only want git to start tracking and checking files for changes once a write has happened since that is a lot less common than reads and git will stay fast because of the limited number of files it has to go through.

pmj: > We only want git to start tracking and checking files for changes once a write has happened

Sure, but when we get `KAUTH_VNODE_WRITE_DATA`, the write is about to happen, so why not if the file was previously empty, only if it was already hydrated at that point?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @kewillford I'm really glad you brought this up, and I think it would good to make sure everyone is aware of how the process works (and confirm my own understanding).

When a new placeholder is written we set FileFlags_IsEmpty=true and set PrjFSFileXAttrName.  
When the placeholder becomes hydrated, FileFlag_isEmpty=false
When the hydrated file is modified we remove PrsFSFileXAttrName

@wilbaker is that correct?

PreConvertToFull means we should add the file to modifiedFiles.dat.  A Modified alert tells us to run the statusCache again (file would already be in modifiedFiles.dat).  If my understanding is correct the name PreCovertToFull seems a bit off, as does the partialFile variable name on  PrsFSFileXAttrName.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
kewillford: If I understand correctly, in the case of a write event and the file is empty, the code needs to hydrate the file and mark it as full, which is not happening currently with the `goto` above after the file is hydrated.  Isn't the `goto` only in error conditions?  Sorry I'm not yet familiar with this code.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: This matches my (possibly incorrect) understanding, and I too am confused by the apparent "placeholder" vs "partial" terminology mismatch. Is there a difference?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Ah, sorry - of course, the goto only happens in the error condition. My mistake.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Exellent nit!  The place I copied it from does use that var name and honestly I don't think it makes sense their either :-).  I'll wait for clarification on the states above update accordingly.

jeschu1:
@kewillford I found the editor flow suprising on Mac.  For many IDEs (VS, TextEdit, XCode) the flow is:

When I save aaa.txt it
1. Writes to aaa.txt.temp
2. Deletes aaa.txt
3. Rename aaa.txt.temp to aaa.txt

Which results in a serious of events on save such as:
PrjFSLib.HandleFileNotification: aaa.txt TextEdit PrjFS_NotificationType_PreDelete
HandleNewFileInRootNotification: aaa.txt TextEdit PrjFS_NotificationType_FileRenamed 
PrjFSLib.HandleFileNotification: aaa.txt TextEdit PrjFS_NotificationType_FileRenamed 

This means we'll get multiple D records in placeholders.dat for aaa.txt. Do you forsee any major issues with that flow?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: > When a new placeholder is written we set FileFlags_IsEmpty=true and set PrjFSFileXAttrName.
When the placeholder becomes hydrated, FileFlag_isEmpty=false
When the hydrated file is modified we remove PrsFSFileXAttrName

Yep that's correct.

> PreConvertToFull means we should add the file to modifiedFiles.dat.

That's right too.  When a file is converted to full (i.e. a plain\regular file without our flags or xattrs) then git becomes responsible for the file and it must be added to ModifiedPaths.dat.

It's also important to note that placeholders are removed from the placeholder list in response to this callback.

> If my understanding is correct the name PreConvertToFull seems a bit off

Which part of the name seems off?  This callback lets the provider know the placeholder is about to be converted to full.

In addition to the topic of naming, this comment brings up a great point.  I missed a detail of PreConvertToFull's behavior on Windows in our offline conversations. 

On Windows the flow is like this:

1. Placeholder file is being opened for write
2. Provider's PreConvertToFull callback is called, and the provider either returns HResult.OK, or an error code
3. If the provider returns OK, then the placeholder reparse data is removed from the file (and it's converted to full)
4. If the provider returns an error, the file is *not* converted to full, and the caller (trying to write to the file) sees an error

So on Windows the provider has a chance to deny the placeholder from being converted to full.  Although VFS4G does have a window of time where it does return an error, we have #793 open for eliminating that window (and so I'm not sure we need\want that feature in the kext).

> as does the partialFile variable name on PrsFSFileXAttrName.
> This matches my (possibly incorrect) understanding, and I too am confused by the apparent "placeholder" vs "partial" terminology mismatch. Is there a difference?

*Unofficial Glossary*

- "Placeholder" -> A file that is tagged with ProjFS attributes (reparse data on Windows and xattrs on Mac).  
- "Empty Placeholder" -> A placeholder that does not have its file contents
- "Hydrated Placeholder" or "Partial File" -> A placeholder whose file contents are on disk.  The Windows ProjFS API is designed such that in the future a provider could "partially" hydrate a file and only provide the parts of the file that the caller cares about, but VFS4G has no plans to use this feature.
- "Full file" -> A regular filesystem file that does not have any ProjFS related attributes attached to it.  The term "full file" is used to describe both placeholders that were converted to full, and newly created files (that were never projected by VFS4G).

I have no problem with eliminating the term "partial file" from the codebase if that makes the code easier to follow.

pmj:
> When I save aaa.txt it
> 
> 1. Writes to aaa.txt.temp
> 2. Deletes aaa.txt
> 3. Rename aaa.txt.temp to aaa.txt

Note that this is just the sequence of authorisation events. In practice, I can almost guarantee you that steps 2 & 3 are caused by a single `rename()` syscall. Doing it this way makes the file write atomic: under no circumstances (sudden crash etc.) can the file be in an inconsistent state. (Assuming the file system actually sticks to any synchronisation guarantees, which many don't.)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: > Which part of the name seems off? This callback lets the provider know the placeholder is about to be converted to full.

I was thinking PreConvertToFull was in relation to an empty placeholder being hydrated.  Now I understand the meaning is non-ritualized 'full' file.  Thanks for the glossary it all makes sense now!

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -217,7 +217,8 @@ protected override bool TryStart(out string error)
             this.virtualizationInstance.OnPreDelete = this.OnPreDelete;
             this.virtualizationInstance.OnNewFileCreated = this.OnNewFileCreated;
             this.virtualizationInstance.OnFileRenamed = this.OnFileRenamed;
-            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
+            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
+            this.virtualizationInstance.OnFilePreConvertToFull = this.OnFilePreConvertToFull;
```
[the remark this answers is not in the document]
jeschu1: Done, but please verify.  First time I've worked in that project.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -277,6 +277,31 @@ protected void OnHardLinkCreated(string relativeExistingFilePath, string relativ
...
 
+        protected void OnFilePreConvertToFull(string relativePath)
+        {
+            try
+            {
+                if (!this.FileSystemCallbacks.IsMounted)
+                {
+                    EventMetadata metadata = this.CreateEventMetadata(relativePath);
+                    metadata.Add(TracingConstants.MessageKey.InfoMessage, nameof(this.OnFilePreConvertToFull) + ": Mount has not yet completed");
+                    this.Context.Tracer.RelatedEvent(EventLevel.Informational, $"{nameof(this.OnFilePreConvertToFull)}_MountNotComplete", metadata);
```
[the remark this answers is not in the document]
jeschu1: Merged after `IsMounted `removed. Removed related `IsMounted` code.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
jeschu1: Probably hard to find the {}'s but this check is done regardless of `FileFlags_IsEmpty`.  If you view the full file it's probably easier to line the {}'s up.

We do in fact have a functional test covering the case too, WriteWithoutClose.

jeschu1:
@wilbaker this should be ready for another review, thanks!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
wilbaker: >Probably hard to find the {}'s but this check is done regardless of FileFlags_IsEmpty. If you view the full file it's probably easier to line the {}'s up.

Thanks! Yep that was my mistake, I must've jumped braces when seeing where this lined up.

> We do in fact have a functional test covering the case too, WriteWithoutClose.

The test in `WriteWithoutClose` appears to be covering the scenario where we write to an empty placeholder without having read from it first.  Will we get one VnodeOp callback in that scenario or two?

Assuming it's just one callback (and `FileFlags_IsEmpty` is set), it would be good to add a test opens/reads the file before writing to it so that we can test the scenario where the empty flag is not set when `KAUTH_VNODE_WRITE_DATA` is set.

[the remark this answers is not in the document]
jeschu1: @wilbaker As an FYI in the kext we do get a READ followed by WRITE for these operations
The open causes a READ and the write causes the WRITE.
So it does test the flow we want.

I can see how that can be confusing though.  I'll tidy the naming here and add an explicit read so there is no ambiguity.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1024,9 +1025,9 @@ static PrjFS_Result HandleFileNotification(
         notificationType,
         nullptr /* destinationRelativePath */);
     
-    if (partialFile && PrjFS_NotificationType_FileModified == notificationType)
+    if (result == 0 && placeholderFile && PrjFS_NotificationType_PreConvertToFull == notificationType)
     {
-        // PrjFS_NotificationType_FileModified is a post-modified FileOp event (that cannot be stopped
+        // PrjFS_NotificationType_PreConvertToFull is a event (that cannot be stopped
         // by the provider) and so there's no need to check the result of the call to NotifyOperation
```
wilbaker: This comment is out of date:

```
// PrjFS_NotificationType_PreConvertToFull is a event (that cannot be stopped
// by the provider) and so there's no need to check the result of the call to NotifyOperation
```

Now the provider can stop the event (and we are checking the result of `NotifyOperation`

PR #792, opened 2019-02-15
Enable telemetry daemon tracing on Windows and macOS

### LINKED ISSUE #529: Mac: create a telemetry listener

### LINKED ISSUE #739: Enable the `TelemetryDaemonEventListener` on Windows

mjcheetham:
Re-enable the telemetry daemon trace listener for Windows and enable it for macOS (using the byte-based transmission mode).

Resolves #739
Resolves #529


GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -91,16 +91,14 @@ public override NamedPipeServerStream CreatePipeByName(string pipeName)
...
-            // var daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, pipeName: "vfs");
-            // if (daemonListener != null)
-            // {
-            //     yield return daemonListener;
-            // }
-
-            yield break;
+            string gitBinRoot = this.GitInstallation.GetInstalledGitBinPath();
+
+            const string pipeName = "/usr/local/GitService/pipe/vfs-c780ac06-135a-4e9e-ab6c-d41e2d265baa";
```
jrbriggs: This should be read from config and not hardcoded in the product, similar to how the `GVFSTelemetryId` is managed.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -85,14 +85,12 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
...
 
-            // TODO: enable the daemon-based telemetry listener once we're happy.
-            // See GitHub issue: https://github.com/Microsoft/VFSForGit/issues/739
-            //
-            // TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, pipeName: "vfs");
-            // if (daemonListener != null)
-            // {
-            //     yield return daemonListener;
-            // }
+            const string pipeName = "vfs-c780ac06-135a-4e9e-ab6c-d41e2d265baa";
```
jrbriggs: Ditto on having this be a config setting.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
mjcheetham: Moved it to a new config setting `gvfs.telemetry-pipe` as not to interfere with the existing `telemetry-id` setting used by the ETW listener.

If `telemetry-pipe` is set to not-null then we'll take that as the pipe name and try sending to it, otherwise telemetry via pipe is disabled.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -112,6 +112,13 @@ private void SendMessage(string message)
 
             try
             {
+                // If we're in byte/stream transmission mode rather than message mode
+                // we should signal the end of each message with a line-feed (LF) character.
+                if (this.pipeClient.TransmissionMode == PipeTransmissionMode.Byte)
+                {
+                    message += '\n';
```
jrbriggs: If we're using a single '\n' for end of message, how much risk is there a message would contain a linefeed in the message payload?

jeffhostetler: The json formatter should backslash encode LF chars that are inside string fields.
The json formatter will only insert LFs in "indented/pretty" formatting mode.

So I think we're good here.  But I can't see the caller in this view, so we should test this.
I have stuff in the GItService to discard json fragments, so if I get one, we'll ignore the message.

jeffhostetler:
Do we have any idea about the maximum message size we might see?  (asking for a friend)

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: I did wonder about this too.. perhaps we should just play it safer by using something like two null-bytes? \0\0?

mjcheetham:
> Do we have any idea about the maximum message size we might see? (asking for a friend)

The sort of minimum payload size would be 100-150 bytes (version info, mount & enlistment IDs).
A simple `git status` results in a message of the order of 500-800 bytes.
Something like a single background maintenance step is about 220 bytes.
During mount we send at least one message that includes a bunch of information about the enlistment that is about 2000 bytes.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -85,14 +85,11 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
                 yield return etwListener;
```
kewillford: Will we eventually remove this listener?  Is there anything that will be preventing double entries in reporting?

mjcheetham: There's actually a chance of triple reporting here.
We will eventually remove the ETW listener on Windows, since the telemetry daemon on Windows can post to both ETW and AI any messages it gets.

@jrbriggs what do we want to do here? There are 3 telemetry paths on Windows:
1. ETW in-proc
2. ETW via pipe/service
3. AI via pipe/service

(on Mac we only have path 3)

Do we want to run the in-proc ETW listener on Windows until such a time that we're happy either the telemetry pipe is reliable, and/or the ETW publishing in the daemon is good, and/or the unified AI->kusto process works?

jrbriggs: I want to have VFS for Git work consistently on each platform. I think this means that we cut over to the daemon publishing. That will avoid duplicates going through the ETW pipeline.

Let's not publish the VFS4G stream to AI in this release.

mjcheetham: > Let's not publish the VFS4G stream to AI in this release.

But then we'd have nothing coming from Mac? Do you mean:

On Windows:

- disable ETW in-proc listener
- enable telemetry pipe listener
- configure telemetry service to post to ETW only

On Mac:

- enable telemetry pipe listener
- configure telemetry service to post to AI only

jeffhostetler: The 1.0.0.4 release of GitService that we're shipping in this cycle
does not have any of the AI nor VFS proxy code.  This version only
does ETW and only for Trace2 traffic.

We need to wait until next sprint to cut over.  (Or pull more code
over from my master-candidate branch into this maint branch.)

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
jeffhostetler: let's wait on this for now.  

let's do some testing where we build a dictionary with embedded LFs and use the
existing ETW call stack to send the dictionary and see what we get.

If we do manage to get a raw (un-escaped) LF onto the wire, then we can look at
making it a single NULL byte.  (A 2 byte sequence changes the message decoder
state machine (not a big deal, but just more work (that i'm not sure we need yet)).)

jeffhostetler:
WRT max message size.  That's fine.  I'm receiving into a 64k buffer and the socket layer is likely chunking traffic to me into 8k chunks.  So we should be fine.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
mjcheetham: I've removed the ETW in-proc listener completely.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: Just did a quick test; I added a bunch of LFs to the event name, and to the 'payload' dictionary (both key and value). The JSON payload is indeed escaped correctly and looks like this in the debug console for the GitService:

```
...\"\\n\\nLINE-FEED\\n\\n\":\"\\n\\n\\n\\n\\n\\n\"}"}}
```

The event name however seemed to be printed literally, with the line-feeds resulting in a new line as shown in the debug console:

```
Microsoft.Git.GVFS/EnlistmentInfo

: {"version":{"service":"1.0.1.0","vfs":"0.2.173.2"},...
```

whereas normal messages look like:

```
Microsoft.Git.GVFS/PreMount: {"version":{"service":"1.0.1.0"...
```

..but I think this is just aesthetic; the messages are being received and decoded correctly.

jeschu1: Would a size header make sense?

mjcheetham: @jeffhostetler what do you think

mjcheetham:
Closing and reopening to trigger a rebuild.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -79,20 +79,11 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
         {
             string gitBinRoot = this.GitInstallation.GetInstalledGitBinPath();
 
-            ETWTelemetryEventListener etwListener = ETWTelemetryEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
-            if (etwListener != null)
+            TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
```
wilbaker: Are there any plans to support multiple listeners in the future?  If not, we could simplify the method signature to 
```
EventListener CreateTelemetryListener(string providerName, string enlistmentId, string mountId)
```

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -91,16 +91,13 @@ public override NamedPipeServerStream CreatePipeByName(string pipeName)
...
 
-            // string gitBinRoot = this.GitInstallation.GetInstalledGitBinPath();
-            // var daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, pipeName: "vfs");
-            // if (daemonListener != null)
-            // {
-            //     yield return daemonListener;
-            // }
-
-            yield break;
+            TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
```
wilbaker: It looks like this code is identical to the code in `GVFSPlatform`.

Assuming that's the case, `CreateTelemetryListeners` can be removed from `GVFSPlatform` (`GVFSPlatform` is intended for code that needs to be platform specific).

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
mjcheetham: Removed this entirely from `GVFSPlatform`.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
jeffhostetler: let's not do a size header/prefix.  i think we're good with the LF.

mjcheetham:
@jrbriggs was there anything left on this PR that you wanted changing? (I believe your review vote is the only thing blocking this now :-) ).

PR #798, opened 2019-02-16
Mac PrjFS: Countermeasures for deadlocks with Antivirus software

### LINKED ISSUE #354: Mac: Investigate issues with PrjFSKext colliding with System Center Endpoint Protection (and other kexts and anti-virus software)
If our kext is loaded at the same time as the kext that ships with System Center Endpoint Protection, the machine will panic on next file access. I suspect this will occur with other AV kexts as well, we need to investigate what mitigations we can take in our kext to not collide with others on a machine.

pmj:
This change implements 2 strategies for reducing the likelihood of deadlocks with other software:

 * In the provider, report hydration success back to the kext without waiting for close() to return. This avoids delaying on other kexts' KAUTH_FILEOP_CLOSE callbacks.
 * In the kext, only allow the user who mounted the repo to cause files to be hydrated. Antivirus typically runs as root, while repos are owned by non-privileged users, so this simply denies AV services access to un-hydrated files.

This seems to fix #354 in the specific case of Avast AV for Mac.


jeschu1:
> * In the kext, only allow the user who mounted the repo to cause files to be hydrated. Antivirus typically runs as root, while repos are owned by non-privileged users, so this simply denies AV services access to un-hydrated files.

I don't think we can make that assumption. I believe there are many reasons a user may run as elevated.  Does the AV software have a processname we can put on the ignore list?

jeschu1:
`* In the provider, report hydration success back to the kext without waiting for close() to return. This avoids delaying on other kexts' KAUTH_FILEOP_CLOSE callbacks.`

What was the change to support this?

pmj:
> What was the change to support this?

[This one in `PrjFSLib.cpp`](https://github.com/Microsoft/VFSForGit/pull/798/commits/d564416f7e0656cdee976865d08cbfdd94093d7a#diff-5c944eca485016533d3f61ab623cb0a9). Note that the way I made the change was the path of least resistance rather than necessarily the best way to go about it.

It's not a terrible way, but whether it's the way to go longer term depends on whether other kernel-user messages need similar changes that can't easily be implemented in the same way, in which case we need a different pattern for sending the reply to the kext and then doing something else.

pmj:
> Does the AV software have a processname we can put on the ignore list?

It does, and we may want to do that regardless, but there are a lot of AV tools on the market, and trying to maintain an exhaustive list seems like a fool's errand. I'd like to try and come up with something that works pretty well even with AV(-like) software we haven't seen before; we can still aim to make the experience even better by blacklisting more software.

> I don't think we can make that assumption. I believe there are many reasons a user may run as elevated.

This is fair, to an extent at least. I'm not sure how common it would be to run an elevated process in a VFS4G repo that accesses files that won't have already been accessed by the unprivileged user, but it's certainly possible.

So perhaps we need to refine the criterion. A common case would be using `sudo` or perhaps a setuid-executable as part of a build process, right? So how about we look at the process's ancestry in addition to the user? If the process owner doesn't match the repo's mounter, we could walk up the process hierarchy and see if we find a process that *is* owned by the mounter. Then the chances are extremely high it's a user-initiated process and should be allowed to hydrate.

This breaks down for privileged helper tools launched via XPC. Identifying those could be tricky. Activity Monitor does it, but it uses private APIs to do it.

wilbaker:
(Note: I've only made it through the comments on this PR so far, I still need to review the changes)

> So how about we look at the process's ancestry in addition to the user? If the process owner doesn't match the repo's mounter, we could walk up the process hierarchy and see if we find a process that _is_ owned by the mounter. Then the chances are extremely high it's a user-initiated process and should be allowed to hydrate.

I think walking up the process hierarchy sounds like a good approach, but I don't think should limit hydration based on what user performed the mount (e.g. a user might log into a machine to inspect the state of a repo that was mounted by a build agent).

In offline discussions with other kext authors I've heard that the approach they've taken when deciding to hydrate is based on root vs. non-root rather than looking for specific users:

> 1. Non-root processes. Obviously these are ok.
> 2. Root processes with a non-root parent. These are processes that have been sudo'ed. These are ok.
> 3. Root processes with all parents also running as root. These are system processes and are blocked.

And from what I've heard this approach has worked well for them.  Could we try going with this same approach?

wilbaker:
> And from what I've heard this approach has worked well for them. Could we try going with this same approach?

Also, I see now that some of this discussion has been happening in #354, I will move conversation back over there :smile:

jeschu1:
@wilbaker that seems reasonable to me. :-)

pmj:
This latest push changes the process user check to walk the process hierarchy until it finds one with EUID <500, or hits process 1.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -889,28 +901,97 @@ static bool TryGetVirtualizationRoot(
...
+        {
+            break;
+        }
+        
+        proc_t parentProcess = proc_find(parentPID);
+        proc_rele(process);
+        process = parentProcess;
+        if (parentProcess == nullptr)
+        {
+            KextLog_Error("CurrentProcessIsOwnedOrWasSpawnedByRegularUser: Failed to locate ancestor process %d for current process %d\n", parentPID, proc_selfpid());
```
wilbaker: Is it possible that the oldest ancestor process has a PID between 2 and 499?

pmj:
> I (finally) got around to taking a look at this, I like it! My only concern is the same as William's with respect to having a helper that calls both `ShouldMakeProviderRequest` and `TryGetVirtualizationRoot` (or changing the way we call TryGetVirtualizationRoot`).

Yeah, the change of factoring bits out of `TryGetVirtualizationRoot` was perhaps a little half-hearted and at any rate didn't work out as well as hoped. This code definitely needs a cleanup though, so I've filed #900 to cover that. I'll change this PR to simply add the extra functionality in the existing functions.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -889,28 +901,97 @@ static bool TryGetVirtualizationRoot(
...
+}
+    
+    
+
+static bool ShouldMakeProviderRequest(
+    // In params:
+    PerfTracer* perfTracer,
+    VirtualizationRootHandle root,
+    pid_t pidMakingRequest,
+    bool restrictByUser,
```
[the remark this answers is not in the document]
pmj: Well, that's only strictly correct if we switch from uid<500 to uid 0.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: The ultimate parent is, as far as I'm aware, always `launchd` (PID 1), which runs as root (UID 0). Does that answer the question? If not, please elaborate!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -381,7 +390,8 @@ static int HandleVnodeOperation(
     
     if (isDeleteAction)
     {
-        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, &root, &vnodeFsidInode, &kauthResult, kauthError))
+        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, &root, &vnodeFsidInode, &kauthResult, kauthError)
+            || !ShouldMakeProviderRequest(&perfTracer, root, pid, false /* any user */, &kauthResult))
```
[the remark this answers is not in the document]
pmj: I've now moved the `ShouldMakeProviderRequest` functionality back into `TryGetVirtualizationRoot`, we can clean this code up separately. (see #900)

pmj:
I've incorporated the review feedback and rebased the patch set on the latest master branch. I've been using the branch for local testing (including with active AV) and it seems to be working pretty well, so I think it's ready to go in.

I still think the timeout would be worth adding eventually, and it would also not hurt to audit the kext->user notifications to see if some of them could do without feedback from the provider, or if the provider could unblock the kext sooner in some cases. I'd appreciate some help from someone more knowledgeable about the VFS4G core side for these additional changes. I'll continue tracking those in #354.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -46,7 +52,7 @@ static bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t _Nonnull context, ui
 KEXT_STATIC_INLINE bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
 static inline bool TryGetFileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t _Nonnull context, bool* flaggedInRoot);
 KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
-
+static bool CurrentProcessIsOwnedOrWasSpawnedByRegularUser();
```
nickgra: Naming nit: Is there a difference between owned and spawned as far as we're concerned or can we rename this to simply be Owned?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -21,6 +21,12 @@
 #include "KauthHandlerTestable.hpp"
 #endif
 
+enum ProviderCalloutUserPolicy
```
nickgra: Naming nit: Would `ProviderHydrationPolicy` be a clearer name? I think the use of callout here is uninituitive (to me at least).

pmj: The reason I didn't go for that kind of name is because for example when sending `MessageType_KtoU_NotifyDirectoryPreDelete` and `MessageType_KtoU_NotifyFilePreDelete` we need to use `UserPolicy_AllowAny` and this has nothing to do with hydration.

"ProviderRequest" doesn't make sense either, as again notifications aren't really requests, and "ProviderNotification" isn't quite right for hydration/enumeration requests. So I went with Callout because it's about the most generic word I could think of for making a roundtrip to the provider, be it for hydration/enumeration or notifications. I don't love the name either, so I'll easily be persuaded to change it. :-)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: "Owned" to me implies "UID is equal to", which is definitely not what this function checks. "Spawned" covers the ownership bit too though so we could reduce it down to `CurrentProcessWasSpawnedByRegularUser()` to make it less wordy.

wilbaker: > we could reduce it down to CurrentProcessWasSpawnedByRegularUser() to make it less wordy.

+1 for switching to `CurrentProcessWasSpawnedByRegularUser`

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -358,7 +365,7 @@ static int HandleVnodeOperation(
     
     if (isDeleteAction)
     {
-        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, &root, &vnodeFsidInode, &kauthResult, kauthError))
+        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, UserPolicy_AllowAny, &root, &vnodeFsidInode, &kauthResult, kauthError))
```
wilbaker: Could you add comments describing why we're using `UserPolicy_AllowAny` here and `UserPolicy_UserInitiatedOnly` below?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: > The ultimate parent is, as far as I'm aware, always launchd (PID 1), which runs as root (UID 0). Does that answer the question? If not, please elaborate!

Yep that answers my question thanks!  I was trying to understand the reason for the gap between 1 and 500 in the loop, but I see now that we just walk all the way up the process tree and bail if/when we find a user-spawned process (>= 500).

pmj:
Thanks for the feedback, I've just pushed a new version which implements all the suggestions.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -358,7 +365,8 @@ static int HandleVnodeOperation(
     
     if (isDeleteAction)
     {
-        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, &root, &vnodeFsidInode, &kauthResult, kauthError))
+        // Allow any user to delete, as this generally doesn't cause nested kauth callbacks.
+        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_AllowAny, &root, &vnodeFsidInode, &kauthResult, kauthError))
```
wilbaker: I don't think we have to make changes now, but I wanted to mention that this code looks a bit strange because we check `CallbackPolicy_AllowAny` here for PreDelete and then below for the PreDelete of directories we use `CallbackPolicy_UserInitiatedOnly`.

```
// Prevent system services from enumerating directories as this tends to cause deadlocks with the kauth listeners for Antivirus software
if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
{
```

I suspect this can get cleaned up once we have WILL_RENAME as we'll be able to differentiate between PreDelete for rename and PreDelete for delete.

We could consider updating this call to:

```isDirectory ? CallbackPolicy_UserInitiatedOnly : CallbackPolicy_AllowAny```

And including a comment that directory PreDeletes require expansion because they might be part of a rename.

Thoughts?

File: ProjFS.Mac/PrjFSKext/KauthHandler.cpp [hunk 8]
```diff
@@ -419,7 +428,8 @@ static int HandleVnodeOperation(
             }
             else if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                // Prevent system services from enumerating directories as this tends to cause deadlocks with the kauth listeners for Antivirus software
```
wilbaker: Minor comment suggestion:  

Here and in the comment above consider switching from the term "enumerating" to "expanding" (or even "expanding directories as part of enumeration").

Although the name of the callback includes "enumerate" it's the expansion that happens in the callback that causes deadlocks.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Good point, I've fixed that up and updated the comment to reflect it as well.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Done.

wilbaker:
@pmj thanks for the latest updates, I think this looks good to go!

Issue #515, opened 2018-11-15
Mac: Folder renames must hydrate children and convert them to full


wilbaker:
Currently folder renames expand all children, however, they are not hydrated nor converted to full.  While this is OK if the rename stays within the root, it will break if the rename is to a location outside of the root.

wilbaker:
Update: With the changes in #526 we'd have the option of blocking partial directory renames (which is how ProjFS behaves on Windows)  as we'll know if the PreDelete is for a rename or not.

wilbaker:
Per an offline discussion with @jrbriggs and @jeschu1 we'll aim for the same behavior that we have on Windows, and that is blocking renames of partial folders.

We think it *should* be possible to avoid any recursive walks when making the decision if a folder rename should be allowed assuming that the xattr data allows us to tell if the folder is one that was projected by VFS4G (i.e. it's either an expanded or unexpanded folder).  If a folder was created by a user the rename should be allowed.

pmj:
As far as I can tell, we currently don't set any xattr on directories. So to clarify, are you saying we should now do so to record whether a directory and all its children are partial versus full? In other words, we would probably need a number of expected children, as well as a counter of "full" children, and as long as the 2 do not match, the directory is partial and therefore cannot be renamed? (A subdirectory would increment its parent's counter only when its own counts match.)

I'm not sure this effort is worth it given the likely user confusion it will cause. Is there a way we can enable `git mv` but simply disable all other rename operations?

wilbaker:
> As far as I can tell, we currently don't set any xattr on directories. 

That is unfortunate.  It appears that to support this behavior we'd need an upgrade steps where VFS4G would scan its placeholder directory and add xattrs to the directories.

> So to clarify, are you saying we should now do so to record whether a directory and all its children are partial versus full? In other words, we would probably need a number of expected children, as well as a counter of "full" children, and as long as the 2 do not match, the directory is partial and therefore cannot be renamed? (A subdirectory would increment its parent's counter only when its own counts match.)

The idea was a bit simpler than that.  I had (incorrectly) assumed that directories had xattrs on them and the idea was that if the directory had the placeholder xattr on it then the rename of the directory would be blocked.  This block would happen regardless of whether the folder had been expanded.

> I'm not sure this effort is worth it given the likely user confusion it will cause.  Is there a way we can enable `git mv` but simply disable all other directory rename operations?

I *think* the only things that would be required to support this are:

[...]

PR #804, opened 2019-02-19
Remove isMountComplete check

### LINKED ISSUE #793: Remove isMountComplete check, and only start ProjFS once VFS4G is ready to service requests
Porting from old backlog (see User Story 1258754).

There should be no need to start the virtualization instance before VFS4G is ready.

jeschu1:
-  Only start ProjFS once VFS4G is ready to service requests
resolves #793

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
@@ -19,12 +19,25 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         private const int GVFSGenericError = 3;
         private const uint GenericRead = 2147483648;
         private const uint FileFlagBackupSemantics = 3355443;
+        private readonly int fileDeletedBackgroundOperationCode;
+        private readonly int directoryDeletedBackgroundOperationCode;
 
         private FileSystemRunner fileSystem;
 
         public MountTests()
         {
             this.fileSystem = new SystemIORunner();
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                this.fileDeletedBackgroundOperationCode = 16;
+                this.directoryDeletedBackgroundOperationCode = 17;
+            }
+            else
+            {
+                this.fileDeletedBackgroundOperationCode = 3;
+                this.directoryDeletedBackgroundOperationCode = 11;
+            }
         }
 
         [TestCaseSource(typeof(MountSubfolders), MountSubfolders.MountFolders)]
@@ -208,6 +221,31 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.Enlistment.MountGVFS();
 
             alternatesFilePath.ShouldBeAFile(this.fileSystem).WithContents(objectsRoot);
+        }
+
+        [TestCase]
+        public void MountCanProcessSavedBackgroundQueueTasks()
+        {
+            string deletedFileEntry = "Test_EPF_WorkingDirectoryTests/1/2/3/4/ReadDeepProjectedFile.cpp";
+            string deletedDirEntry = "Test_EPF_WorkingDirectoryTests/1/2/3/4/";
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, deletedFileEntry);
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, deletedDirEntry);
+            this.Enlistment.UnmountGVFS();
+
+            // Prime the background queue with delete messages
+            string deleteFilePath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4", "ReadDeepProjectedFile.cpp");
+            string deleteDirPath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4");
+            string persistedDeleteFileTask = $"A 1\0{this.fileDeletedBackgroundOperationCode}\0{deleteFilePath}\0";
+            string persistedDeleteDirectoryTask = $"A 2\0{this.directoryDeletedBackgroundOperationCode}\0{deleteDirPath}\0";
+            this.fileSystem.WriteAllText(
+                Path.Combine(this.Enlistment.EnlistmentRoot, ".gvfs", "databases", "BackgroundGitOperations.dat"),
+                $"{persistedDeleteFileTask}\r\n{persistedDeleteDirectoryTask}\r\n");
+
+            // Background queue should process the delete messages and modifiedPaths should show the change
+            this.Enlistment.MountGVFS();
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, deletedFileEntry);
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, deletedDirEntry);
         }
 
         [TestCaseSource(typeof(MountSubfolders), MountSubfolders.MountFolders)]
--- a/GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
+++ b/GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
@@ -72,32 +72,6 @@ namespace GVFS.UnitTests.Virtualization
         }
 
         [TestCase]
-        public void StartingAndStoppingSetsMountedState()
-        {
-            using (MockBackgroundFileSystemTaskRunner backgroundTaskRunner = new MockBackgroundFileSystemTaskRunner())
-            using (MockFileSystemVirtualizer fileSystemVirtualizer = new MockFileSystemVirtualizer(this.Repo.Context, this.Repo.GitObjects))
-            using (MockGitIndexProjection gitIndexProjection = new MockGitIndexProjection(new[] { "test.txt" }))
-            using (FileSystemCallbacks fileSystemCallbacks = new FileSystemCallbacks(
-                this.Repo.Context,
-                this.Repo.GitObjects,
-                RepoMetadata.Instance,
-                new MockBlobSizes(),
-                gitIndexProjection: gitIndexProjection,
-                backgroundFileSystemTaskRunner: backgroundTaskRunner,
-                fileSystemVirtualizer: fileSystemVirtualizer))
-            {
-                fileSystemCallbacks.IsMounted.ShouldBeFalse();
-
-                string error;
-                fileSystemCallbacks.TryStart(out error).ShouldBeTrue();
-                fileSystemCallbacks.IsMounted.ShouldBeTrue();
-
-                fileSystemCallbacks.Stop();
-                fileSystemCallbacks.IsMounted.ShouldBeFalse();
-            }
-        }
-
-        [TestCase]
         public void GetMetadataForHeartBeatDoesNotChangeEventLevelWhenNoPlaceholderHaveBeenCreated()
         {
             using (MockBackgroundFileSystemTaskRunner backgroundTaskRunner = new MockBackgroundFileSystemTaskRunner())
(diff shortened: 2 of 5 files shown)


GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
```diff
@@ -185,15 +159,6 @@ public void IsReadyForExternalAcquireLockRequests()
...
             {
                 string denyMessage;
-                fileSystemCallbacks.IsReadyForExternalAcquireLockRequests(
-                    new NamedPipeMessages.LockData(
-                        pid: 0,
-                        isElevated: false,
-                        checkAvailabilityOnly: false,
-                        parsedCommand: "git dummy-command"),
-                    out denyMessage).ShouldBeFalse();
-                denyMessage.ShouldEqual("Waiting for mount to complete");
```
wilbaker: Rather than removing this scenario from the test (trying to acquire the lock before `TryStart` has been called), can the validate be updated to match whatever `denyMessage` will be returned?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -180,7 +173,11 @@ public bool TryStart(out string error)
 
             this.backgroundFileSystemTaskRunner.Start();
```
wilbaker: I noticed that we don't have any functional tests for loading persisted background tests (and I've filed #816 to add some).  

Could you try running a few manual tests to validate that this still works as expected with this change? (I took a look through the background task code and it all looked like it should be OK to me).

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -180,7 +173,11 @@ public bool TryStart(out string error)
 
             this.backgroundFileSystemTaskRunner.Start();
 
-            this.IsMounted = true;
+            if (!this.fileSystemVirtualizer.TryStart(this, out error))
```
kewillford: We do have some file system calls in the background thread.  Will this cause issues if we start processing the background operations and potentially process one that make a file system call before this TryStart is finished?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: I was looking at this as well and the 2 background tasks that are doing file operations are the folder rename and the folder predelete.  If the file virtualization is not up when those run what will be the results?  

Looks like for the folder rename it would only not be removing entries from the modified paths which wouldn't be a big deal also the newly created list would be empty at this point anyways so I think we are okay there.

For the predelete if it didn't find the folder because the virtualization wasn't up yet and it removed it from the modified path would be an issue but could this not happen because if we are getting a predelete the placeholder is on disk and it would find it?

wilbaker: Great questions on those, please double check my thoughts on them for me 😄 

*Folder rename*

- We'll be enumerating a full folder (on Windows, folder rename will work that way once it's fixed on Mac, see #515) which should work fine if the virtualization root has not been started.  
- ModifiedPath.dat itself is outside the root so modifying it should be OK.

*Folder PreDelete*

- VFS4G will only check if the directory exists if the path is in `newlyCreatedFileAndFolderPaths`, and because VFS4G is just starting up `newlyCreatedFileAndFolderPaths` will be empty.
- Well, I guess that's not 100% guaranteed because another persisted task could have added the path(s) to `newlyCreatedFileAndFolderPaths`.
- However, any path that has been added to `newlyCreatedFileAndFolderPaths` should actually be on disk or not (i.e. it should not be pure virtual) and I would expect checking for existence to work properly even if the virtualization root is not started.  It would be worth confirming the behavior of `Directory.Exists` on Mac and Windows when the repo is not mounted.
- ModifiedPath.dat itself is outside the root so modifying it should be OK.

So for both of these scenarios I think we're okay now, but I could see accidentally breaking them if the background processing changes.  It might be worth adding a few comments near these background tasks as a reminder that they shouldn't assume the repo is mounted.

The other general thought I had was that ProjFS should not be triggering callbacks for I/O VFS4G initiates itself anyway, and so as long as the BG tasks aren't doing anything that ProjFS would block if we're not mounted I think we should be okay.

jeschu1: @wilbaker @kewillford  please let me know if I'm testing this properly.

I put a breakpoint so we halt before` if (!this.fileSystemVirtualizer.TryStart(this, out error))`

move UnitTests UnitTests2
The provider that supports file system virtualization is temporarily unavailable.

If I Delete the UnitTests folder however, it appears again when I continue the mount.  I.e. the delete isn't remembered.

wilbaker: @jeschu1 it sounds like you're trying to test a different scenario than I had in mind.  

To test loading BG tasks on mount the best approach I know is:

1. Temporarily modify the VFS4G source code so that it does not process BG tasks (e.g. make `BackgroundFileSystemTaskRunner.RunCallbackUntilSuccess` an infinite loop).

2. Perform the actions you'd like to have run on startup:
- Create a new directory and rename it
- Delete a directory (note that we only receive PreDelete for directories on Mac)

3. Stop GVFS.Mount.exe

4. There should now have a .gvfs\databases\BackgroundGitOperations.dat file on disk that has all of the unprocessed BG tasks.

5. Back out the change in 1 and mount the repo (it should process the BG tasks in the file)

The approach above could be used to add a functional test that validates that VFS4G can run the persisted BG tasks.  `BackgroundGitOperations.dat` could be checked into the repo and then moved into fresh VFS4G clones during the functional tests.

jeschu1: @wilbaker new background task test added as described

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -210,6 +210,21 @@ public void MountRegeneratesAlternatesFileWhenMissingFromDisk()
...
+        [TestCase]
+        public void MountCanProcessSavedBackgroundQueueTasks()
+        {
+            this.Enlistment.UnmountGVFS();
+
+            string deleteFilePath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4", "ReadDeepProjectedFile.cpp");
+            string deleteDirPath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4");
+            this.fileSystem.WriteAllText(
+                Path.Combine(this.Enlistment.EnlistmentRoot, ".gvfs", "databases", "BackgroundGitOperations.dat"),
+                "A 1\03\0{deleteFilePath}\0\r\nA 2\011\0{deleteDirPath}\0\r\n");
```
wilbaker: `"A 1\03\0{deleteFilePath}\0\r\nA 2\011\0{deleteDirPath}\0\r\n"`

This is missing the `$` and so `{deleteFilePath}` and `{deleteDirPath}` are not being substituted.

The fact that the test didn't fail makes me think it is worth implementing your suggestion of calling `git status` as well at the end of this test.  That would have caught that ModifiedPaths.dat was not updated as expected by VFS4G. when it ran the background tasks.

Also, minor style suggestion here:  It might be a bit easier to understand what's being written to "BackgroundGitOperations.dat" if each line is split out.  Something like:

```
string persistedDeleteFileTask = $"A 1\03\0{deleteFilePath}\0";
string persistedDeleteDirectoryTask = $"A 2\011\0{deleteDirPath}\0";

this.fileSystem.WriteAllText(
                Path.Combine(this.Enlistment.EnlistmentRoot, ".gvfs", "databases", "BackgroundGitOperations.dat"),
                $"{persistedDeleteFileTask}\r\n{persistedDeleteDirectoryTask}\r\n");
```

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -210,6 +210,21 @@ public void MountRegeneratesAlternatesFileWhenMissingFromDisk()
             alternatesFilePath.ShouldBeAFile(this.fileSystem).WithContents(objectsRoot);
         }
 
+        [TestCase]
+        public void MountCanProcessSavedBackgroundQueueTasks()
+        {
+            this.Enlistment.UnmountGVFS();
+
+            string deleteFilePath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4", "ReadDeepProjectedFile.cpp");
+            string deleteDirPath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4");
```
wilbaker: To more accurately simulate this scenario, you could also have the test do the following:

1. Read the file to be deleted before unmounting (to ensure it's on disk)
2. After unmounting, delete `deleteDirPath`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
wilbaker: This test should behave a little different on Mac and Windows.

On Windows we receive/persist the `OnFileDeleted` and `OnFolderDeleted` notifications (3 and 11), but on Mac we receive/persist `OnFilePreDelete` and `OnFolderPreDelete` (16 and 17).

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -210,6 +210,21 @@ public void MountRegeneratesAlternatesFileWhenMissingFromDisk()
             alternatesFilePath.ShouldBeAFile(this.fileSystem).WithContents(objectsRoot);
         }
 
+        [TestCase]
+        public void MountCanProcessSavedBackgroundQueueTasks()
+        {
+            this.Enlistment.UnmountGVFS();
+
+            string deleteFilePath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4", "ReadDeepProjectedFile.cpp");
```
wilbaker: Could you also validate the `OnFolderRenamed` (10) background task in this test?

That was the one other scenario that @kewillford called out in the comment on `FileSystemCallbacks.cs`.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
jeschu1: @wilbaker , in that case how would you recommend we handle it.  Would a switch like we did here for https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs, for OSPlatform.OSX be most appropriate?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
jeschu1: As discussed offline we are going to punt that test for now

jeschu1:
@wilbaker ready for another look :-)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -19,12 +19,25 @@ public class MountTests : TestsWithEnlistmentPerFixture
         private const int GVFSGenericError = 3;
         private const uint GenericRead = 2147483648;
         private const uint FileFlagBackupSemantics = 3355443;
+        private readonly int fileDeletedCode;
+        private readonly int directoryDeletedCode;
```
wilbaker: Naming suggestions:

`fileDeletedCode` -> `fileDeletedBackgroundOperationCode`
`directoryDeletedCode ` -> `directoryDeletedBackgroundOperationCode`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -213,16 +226,27 @@ public void MountRegeneratesAlternatesFileWhenMissingFromDisk()
         [TestCase]
         public void MountCanProcessSavedBackgroundQueueTasks()
         {
-            this.Enlistment.UnmountGVFS();
-
             string deleteFilePath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4", "ReadDeepProjectedFile.cpp");
             string deleteDirPath = Path.Combine("Test_EPF_WorkingDirectoryTests", "1", "2", "3", "4");
```
wilbaker: Minor style comment:  You could move `deleteFilePath` and `deleteDirPath` down below this comment:

`// Prime the background queue with delete messages`

That might make the relationship between `deletedFileEntry`/`deletedDirEntry ` and ModifiedPaths.dat more clear.

jeschu1: Sure.  I did see value in keeping the vars together too.  Aka, this is the file definition this is the entry.
I have moved it are per your suggestion, as it's the type of thing that could go either way (at least in my mind).

wilbaker: Ahh I can see that.  I think for me what threw me off a bit was whether "deletedFileEntry" meant an entry in the background tasks file or an entry in the modified paths file.  Either way is good with me if you prefer the original.

Issue #797, opened 2019-02-16
Mac: kernel panic on gvfs unmount (vnode_t over-release)


pmj:
I've just been made aware (out-of-band) of a kernel panic when unmounting a VFS4G root. The panic stack trace looks roughly like this:

```
*** Panic Report ***
panic(cpu 8 caller 0xffffff801a7f6b9f): "vnode_put(0xffffff8065522c98): iocount < 1"@/BuildRoot/Library/Caches/com.apple.xbs/Sources/xnu/xnu-4903.241.1/bsd/vfs/vfs_subr.c:4515
Backtrace (CPU 8), Frame : Return Address
0xffffff923319b8a0 : 0xffffff801a5aeb0d handle_debugger_trap (in kernel) + 1165
0xffffff923319b8f0 : 0xffffff801a6e8653 kdp_i386_trap (in kernel) + 339
0xffffff923319b930 : 0xffffff801a6da07a kernel_trap (in kernel) + 1274
0xffffff923319b9a0 : 0xffffff801a55bca0 0xffffff800035bca0 (in kernel) + 38
0xffffff923319b9c0 : 0xffffff801a5ae527 panic_trap_to_debugger (in kernel) + 407
0xffffff923319bae0 : 0xffffff801a5ae373 panic (in kernel) + 99
0xffffff923319bb50 : 0xffffff801a7f6b9f vnode_put_locked (in kernel) + 95
0xffffff923319bb80 : 0xffffff801a7f8736 vnode_put (in kernel) + 22
0xffffff923319bba0 : 0xffffff7f9c7aa95d (ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex))
0xffffff923319bbe0 : 0xffffff7f9c7ac3f0 (PrjFSProviderUserClient::clientClose())
0xffffff923319bc00 : 0xffffff801ac891fb iokit_client_died (in kernel) + 251
0xffffff923319bc30 : 0xffffff801a68e9cc iokit_notify (in kernel) + 188
0xffffff923319bc70 : 0xffffff801a5b4455 ipc_kobject_server (in kernel) + 405
0xffffff923319bcc0 : 0xffffff801a58ebe5 ipc_kmsg_send (in kernel) + 549
0xffffff923319bd40 : 0xffffff801a5b4b18 mach_msg_send_from_kernel_proper (in kernel) + 184
0xffffff923319bd80 : 0xffffff801a59a117 ipc_right_terminate (in kernel) + 1063
[...]

PR #805, opened 2019-02-19
Mac Kext: Extra virtualization root vnode assertions & logging

pmj:
I am slightly mystified by issue #797, and as yet do not know what is causing it. If it turns up again, it would be useful to have a log of the events that led up to it, or to catch a violation of the invariants as soon as possible.

The `rootVNode` in each `VirtualizationRoot` should hold an iocount while an active provider is connected, and should be allowed to be recycled (no iocount) when there is no provider connected. To that end, the iocount is reduced via `vnode_put()` in `ActiveProvider_Disconnect()`. The panic was caused by the fact that the iocount was already 0 when the user terminated the provider. I can only see 4 possible causes for this:

 1. When registering the provider in the first place, we did not retain an iocount on the root vnode. (In `VirtualizationRoot_RegisterProviderForPath`, the iocount is incremented by the `vnode_lookup` call, and in the code path where the provider is successfully registered, we arrange for `vnode_put()` *not* to be called. Upon close inspection of the code, I can't see a case where this is violated, but I guess it's possible.)
 2. For some reason, between provider registration and disconnection, `rootVNode` is reassigned to another vnode which does not have an iocount, so when we call `vnode_put()` on disconnecting, we're decrementing the iocount on the wrong vnode, so it goes bang. (The only times `rootVNode` is changed are during provider registration, disconnecting, and in `FindRootAtVnode_Locked()` if the vnode is recycled. Recycling should never happen for roots with an active provider, however, thanks to the nonzero iocount.)
 3. Some other, unrelated code makes an unmatched `vnode_put()` call to the `rootVNode`. (If this was routinely happening, the chances of hitting the iocount assertion in the kernel would surely be much higher for that unrelated code, as all our other functions that call `vnode_put()` run *much* more frequently than `ActiveProvider_Disconnect()`.)
 4. The array of virtualisation roots is getting corrupted somehow; the only place I can think of that happening is when inserting a new root, and for some reason the existing root is being overwritten.

I'm struggling to come up with scenarios where any of these 3 could actually happen. The closest I've got is finding that the recycled vnode refresh code in `FindRootAtVnode_Locked()` (see item 2 above) suffers from a race condition when it is called with only a shared lock, as happens (frequently) in `FindOrDetectRootAtVnode()`. However:

 * I'm struggling to come up with a case where that race condition would actually cause any problems, it seems to me that the worst that can happen is that the `rootVNode` is refreshed to the same new vnode by both racing threads.
 * The vnode of a root with an active provider shouldn't be recycled in the first place.

So, to try and narrow down where the problem is coming from, I've made the following changes:

* Added some more asserts that might catch programming errors in this area, and possibly the root cause of #797 too. Specifically, I'm trying to catch any cases where the vnode of a root with an active provider is recycled, or there is some mismatch between the vnodes.
* Adds verbose logging of changes to virtualisation root table entries: insertions, provider registration/disconnect, and stale root vnode updates.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -470,7 +470,7 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
     return VirtualizationRootResult { err, rootIndex };
 }
 
-void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex)
+void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex, PrjFSProviderUserClient* _Nonnull userClient)
 {
     assert(rootIndex >= 0);
     RWLock_AcquireExclusive(s_virtualizationRootsLock);
@@ -479,6 +479,8 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex)
 
         VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
         assert(nullptr != root->providerUserClient);
+        assertf(userClient == root->providerUserClient, "ActiveProvider_Disconnect: disconnecting provider IOUC %p for root index %d (%s), but expecting IOUC %p",
+            KextLog_Unslide(userClient), rootIndex, root->path, KextLog_Unslide(root->providerUserClient));
         
         assert(NULLVP != root->rootVNode);
         vnode_put(root->rootVNode);
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -328,6 +328,8 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
             assertf(rootEntry.providerUserClient == nullptr, "Finding root vnode based on FSID/inode equality but not vnode identity (recycled vnode) should only happen if no provider is active. Root index %d, provider PID %d, IOUC %p path '%s'",
                 i, rootEntry.providerPid, rootEntry.providerUserClient, rootEntry.path);
             // root vnode must be stale, update it
+            KextLog_FileNote(vnode, "FindRootAtVnode_Locked: virtualization root %d (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u has gone stale, refreshing to %p:%u",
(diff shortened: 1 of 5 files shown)


ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -470,8 +486,15 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex)
 
         VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
         assert(nullptr != root->providerUserClient);
+        assertf(userClient == root->providerUserClient, "ActiveProvider_Disconnect: disconnecting provider IOUC %p for root index %d (%s), but expecting IOUC %p",
```
wilbaker: Do the messages we print in `assertf` show up in kernel panic reports?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -470,8 +486,15 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex)
...
+            KextLog_Unslide(userClient), rootIndex, root->path, KextLog_Unslide(root->providerUserClient));
         
         assert(NULLVP != root->rootVNode);
+        
+        KextLog_FileNote(root->rootVNode, "ActiveProvider_Disconnect: disconnecting provider (PID %d, IOUC %p) for virtualization root %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u.",
```
wilbaker: Could you remind me, will `KEXTLOG_NOTE` level logging make it into the OS log?  Or is that only INFO level or higher?

pmj: * `KEXTLOG_INFO`, corresponding to  `os_log`'s `OS_LOG_TYPE_INFO`, doesn't hit disk unless followed by an ERROR.
* `KEXTLOG_NOTE`, corresponding to  `os_log`'s `OS_LOG_TYPE_DEFAULT` level always hits disk. (Same for errors, which additionally also commit the info ones, and are specially flagged as important.)

As there's been some confusion over this a few times now, I think I might change the `KEXTLOG_NOTE` level to `KEXTLOG_DEFAULT` to match `os_log` and additionally put some comments in the enum explaining the memory vs disk syslog policy.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Yes!

wilbaker: Great thanks! 😄

PR #807, opened 2019-02-19
Update RunFunctionalTests to not load kext when testing installed gvfs

nickgra:
We've recently made the infrastructure change to call the RunFunctionalTests.sh
script with '--test-gvfs-on-path' to ensure that we're testing the product
as it is installed on customer machines.

When we're testing in that scenario, we shouldn't call the developer convenience
script as it will load the locally built kext rather than the installed kext.

### DIFF
--- a/Scripts/Mac/RunFunctionalTests.sh
+++ b/Scripts/Mac/RunFunctionalTests.sh
@@ -10,5 +10,9 @@ fi
 sudo mkdir /GVFS.FT
 sudo chown $USER /GVFS.FT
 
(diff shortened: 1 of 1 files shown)


jrbriggs:
Thanks for this @nickgra . Are we now relying on installing the pkg to install and load the kext? Do we need to update docs on how to set up mac build agents?

nickgra:
Correct, we're relying on the pkg to install the kext and then for the mount process to load the kext (so now we're exercising the code that Kevin added recently). 
I checked the build agent set up documentation in our OneNote and it looks good, we always installed/uninstalled the kext between runs so this isn't a drastic change.

PR #811, opened 2019-02-20
Perform Upgrade Check only after PrjFlt is enabled

### LINKED ISSUE #758: [upgrade] Reconsider PhysicalFileSystem.DeleteDirectory
`PhysicalFileSystem.DeleteDirectory` currently just calls `RecusiveDelete`, so why have both?

### LINKED ISSUE #774: Upgrade reminder not showing on Windows Server PCs
On Windows Server PCs `ProjFS` is not inboxed. Installer pre-check fails in this scenario and causes `GVFS.Service` to ignore the available upgrade. No upgrade reminders get displayed by the git hooks.

The issue with only disabling the `ProjFS` pre-check - `VFS4Git` will start reminding the user that an upgrade is available and advise the user to install it. Now when the user gets around to install the upgrade, the product shows error saying upgrade cannot be installed because ProjFS is not inboxed. 

The goal is to define & implement a better user experience in this scenario.

alameenshah:
#### Perform Upgrade Check only after PrjFlt is enabled
On Windows server PCs, PrjFlt is not inboxed. In this scenario UpgradeCheck used to run in parallel with the Service code to enable PrjFlt. If UpgradeCheck wins, then installer pre-check detects PrjFlt is not enabled. Upgrade check would fail. In this PR, Upgrade check is scheduled to happen only after PrjFlt is enabled.

Fixes #774

#### Reconsider PhysicalFileSystem.DeleteDirectory
PhysicalFileSystem.DeleteDirectory is a single line wrapper around PhysicalFileSystem.RecursiveDelete. Removing DeleteDirectory.

Fixes #758

### DIFF
--- a/GVFS/GVFS.Installer.Windows/Setup.iss
+++ b/GVFS/GVFS.Installer.Windows/Setup.iss
@@ -616,6 +616,89 @@ begin
   Result := True;
 end;
 
+type
+  UpgradeRing = (urUnconfigured, urNone, urFast, urSlow);
+
+function GetConfiguredUpgradeRing(): UpgradeRing;
+var
+  ResultCode: integer;
+  ResultString: ansiString;
+begin
+  Result := urUnconfigured;
+  if ExecWithResult('gvfs.exe', 'config upgrade.ring', '', SW_HIDE, ewWaitUntilTerminated, ResultCode, ResultString) then begin
+    if ResultCode = 0 then begin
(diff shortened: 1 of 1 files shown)


jrbriggs:
@alameenshah let's keep these as separate changes, please. The reviews will be easier.

alameenshah:
> @alameenshah let's keep these as separate changes, please. The reviews will be easier.

Creating separate PRs.

PR #813, opened 2019-02-20
[upgrade] Cleanup gvfs log output

### LINKED ISSUE #760: [upgrade] Cleanup `gvfs log` so that file paths are all aligned in the same column.
Currently looks like:

```
Most recent log files:
  clone     : E:\repo6\.gvfs\logs\gvfs_clone_20181024_094528.log
  mount     : E:\repo6\.gvfs\logs\gvfs_mount_verb_20190130_140733.log
  prefetch  : E:\repo6\.gvfs\logs\gvfs_prefetch_20181024_094548.log
  dehydrate : None
  repair    : E:\repo6\.gvfs\logs\gvfs_repair_20181105_163128.log
  service   : C:\ProgramData\GVFS\GVFS.Service\Logs\gvfs_service_20190207_143020.log
  productupgrade: C:\ProgramData\GVFS\GVFS.Upgrade\Logs\gvfs_productupgrade_process_20190207_144030.log
```

alameenshah:
Cleanup `gvfs log` so that file paths are all aligned in the same column.

Fixes #760

#### New output
screen shot 2019-02-21 at 11 55 06 am

### DIFF
--- a/GVFS/GVFS/CommandLine/LogVerb.cs
+++ b/GVFS/GVFS/CommandLine/LogVerb.cs
@@ -1,5 +1,6 @@
 using CommandLine;
 using GVFS.Common;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
@@ -9,6 +10,7 @@ namespace GVFS.CommandLine
     public class LogVerb : GVFSVerb
     {
         private const string LogVerbName = "log";
+        private static readonly int LogNameConsoleOutputFormatWidth = GetMaxLogNameLength();
 
         [Value(
             0,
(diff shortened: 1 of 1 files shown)


GVFS/GVFS/CommandLine/LogVerb.cs
```diff
@@ -113,7 +113,7 @@ private void DisplayMostRecent(string logFolder, string logFileType)
         {
             string logFile = FindNewestFileInFolder(logFolder, logFileType);
             this.Output.WriteLine(
-                "  {0, -10}: {1}",
+                "  {0, -15}: {1}",
```
kewillford: It would be nice if we could find a way not to hard code this and use the known `logFileTypes`.  If not will you add a comment about this number, why it is `-15` so we will know how we came up with it in case we need to change it again.

alameenshah: Replaced the hardcoding with dynamic computation of width using `logFileTypes`.

GVFS/GVFS/CommandLine/LogVerb.cs
```diff
@@ -109,11 +111,27 @@ private static string GetLogFilePatternForType(string logFileType)
...
+                GVFSConstants.LogFileTypes.Clone,
+                GVFSConstants.LogFileTypes.MountPrefix,
+                GVFSConstants.LogFileTypes.Prefetch,
+                GVFSConstants.LogFileTypes.Dehydrate,
+                GVFSConstants.LogFileTypes.Repair,
+                GVFSConstants.LogFileTypes.Service,
+                GVFSConstants.LogFileTypes.UpgradePrefix,
+            };
+
+            return lognames.OrderByDescending(s => s.Length).First().Length + 1;
```
jrbriggs: What about lognames.Max(s => s.Length) + 1? That reads a little easier to me.

GVFS/GVFS/CommandLine/LogVerb.cs
```diff
@@ -109,11 +111,27 @@ private static string GetLogFilePatternForType(string logFileType)
             return "gvfs_" + logFileType + "_*.log";
         }
 
+        private static int MaxLogNameConsoleOutputWidth()
```
jrbriggs: I would consider naming this `GetMaxLogNameLength()`, since that's what we compute, even though we end up using to define a column width.

PR #814, opened 2019-02-20
[upgrade] Reconsider PhysicalFileSystem.DeleteDirectory

### LINKED ISSUE #758: [upgrade] Reconsider PhysicalFileSystem.DeleteDirectory

alameenshah:
PhysicalFileSystem.DeleteDirectory is a single line wrapper around for PhysicalFileSystem.RecursiveDelete. Removing redundant DeleteDirectory.

Fixes #758

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
@@ -12,7 +12,7 @@ namespace GVFS.Common.FileSystem
     {
         public const int DefaultStreamBufferSize = 8192;
 
-        public virtual void RecursiveDelete(string path)
+        public virtual void DeleteDirectory(string path, bool recursive = true)
         {
             if (!Directory.Exists(path))
             {
@@ -21,15 +21,18 @@ namespace GVFS.Common.FileSystem
 
             DirectoryInfo directory = new DirectoryInfo(path);
 
-            foreach (FileInfo file in directory.GetFiles())
+            if (recursive)
             {
-                file.Attributes = FileAttributes.Normal;
-                file.Delete();
-            }
+                foreach (FileInfo file in directory.GetFiles())
+                {
+                    file.Attributes = FileAttributes.Normal;
+                    file.Delete();
+                }
 
-            foreach (DirectoryInfo subDirectory in directory.GetDirectories())
-            {
-                this.RecursiveDelete(subDirectory.FullName);
+                foreach (DirectoryInfo subDirectory in directory.GetDirectories())
+                {
+                    this.DeleteDirectory(subDirectory.FullName);
+                }
             }
 
             directory.Delete();
@@ -124,11 +127,6 @@ namespace GVFS.Common.FileSystem
         public virtual bool TryCreateDirectoryWithAdminOnlyModify(ITracer tracer, string directoryPath, out string error)
         {
             return GVFSPlatform.Instance.FileSystem.TryCreateDirectoryWithAdminOnlyModify(tracer, directoryPath, out error);
-        }
-
-        public virtual void DeleteDirectory(string path, bool recursive = false)
-        {
-            this.RecursiveDelete(path);
         }
 
         public virtual bool IsSymLink(string path)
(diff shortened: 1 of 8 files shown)


GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -451,7 +451,7 @@ private void MigrateUpgradeLogsToDirectoryWithFreshACLs()
                 else
                 {
                     PhysicalFileSystem fileSystem = new PhysicalFileSystem();
-                    fileSystem.DeleteDirectory(tempUpgradeLogsPath, recursive: false);
+                    fileSystem.RecursiveDelete(tempUpgradeLogsPath);
```
wilbaker: This is actually one spot that I don't think we want `RecursiveDelete` because we want the delete to fail if there are any files left in the old Logs directory (it's currently a possible race condition).

Rather than removing `DeleteDirectory`, what would you think of updating it so that it respects `recursive:` and then making `RecursiveDelete` a private helper method?

alameenshah: Agree. Have renamed `RecursiveDelete` (because, most of its code can be used for non-recursive delete as well) to `DeleteDirectory`. It has an additional parameter `recursive` which is set to `true` by default. `DeleteDirectory` can now do both shallow and recursive deletes.

GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -27,9 +27,12 @@ public virtual void RecursiveDelete(string path)
                 file.Delete();
```
wilbaker: We shouldn't be deleting the files inside `path` if `recursive` if false.  We should just be calling `directory.Delete();`.

wilbaker: (We want to match the behavior of [Directory.Delete](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.delete?view=netframework-4.7.2#System_IO_Directory_Delete_System_String_System_Boolean_))

alameenshah: updated

GVFS/GVFS.UnitTests/Mock/FileSystem/MockFileSystem.cs
```diff
@@ -38,7 +38,7 @@ public MockFileSystem(MockDirectory rootDirectory)
         /// </summary>
         public bool DeleteNonExistentFileThrowsException { get; set; }
 
-        public override void RecursiveDelete(string path)
+        public override void DeleteDirectory(string path, bool recursive = true)
         {
             this.RootDirectory.DeleteDirectory(path);
```
wilbaker: Since you're in here, could you updating this to throw if `recursive` is false (since we don't currently have code to mock that out).  I am pretty sure we're not hitting that code path in the unit tests.

Something like:

```
public override void DeleteDirectory(string path, bool recursive = true)
{
    if (!recursive)
    {
        throw new NotImplementedException();
    }

    this.RootDirectory.DeleteDirectory(path);
}
```

GVFS/GVFS.UnitTests/Mock/FileSystem/MockFileSystemWithCallbacks.cs
```diff
@@ -48,11 +48,6 @@ public override void DeleteFile(string path)
         {
         }
 
-        public override void DeleteDirectory(string path, bool recursive = false)
```
wilbaker: Can this method stay now that we're keeping `DeleteDirectory`?

alameenshah: I've renamed `RecursiveDelete` to `DeleteDirectory(..., bool recursive)`. So we already have that (`DeleteDirectory`). The old `DeleteDirectory` is removed only becoz it didn't do anything more than redirecting to `RecursiveDelete`, I suggest we don't bring it back.

alameenshah: Missed it, good catch... Added the override back.

PR #815, opened 2019-02-20
[upgrade] Perform Upgrade Check after PrjFlt is enabled

### LINKED ISSUE #774: Upgrade reminder not showing on Windows Server PCs

alameenshah:
On Windows server PCs, PrjFlt is not inboxed. In this scenario UpgradeCheck used to run in parallel with the Service code to enable PrjFlt. If UpgradeCheck wins, then installer pre-check detects PrjFlt is not enabled. Upgrade check would fail. In this PR, Upgrade check is scheduled to happen only after PrjFlt is enabled.

Fixes #774

### DIFF
--- a/GVFS/GVFS.Service/GVFSMountProcess.cs
+++ b/GVFS/GVFS.Service/GVFSMountProcess.cs
@@ -27,7 +27,7 @@ namespace GVFS.Service
                 string error;
                 if (!EnableAndAttachProjFSHandler.TryEnablePrjFlt(this.tracer, out error))
                 {
-                    this.tracer.RelatedError($"{nameof(this.Mount)}: Unable to start the GVFS.exe process: {error}");
+                    this.tracer.RelatedError($"{nameof(this.Mount)}: Could not enable PrjFlt: {error}");
                 }
             }
 
--- a/GVFS/GVFS.Service/GvfsService.cs
+++ b/GVFS/GVFS.Service/GvfsService.cs
@@ -45,7 +45,6 @@ namespace GVFS.Service
 
                 this.repoRegistry = new RepoRegistry(this.tracer, new PhysicalFileSystem(), this.serviceDataLocation);
                 this.repoRegistry.Upgrade();
-                this.productUpgradeTimer.Start();
                 string pipeName = this.serviceName + ".Pipe";
                 this.tracer.RelatedInfo("Starting pipe server with name: " + pipeName);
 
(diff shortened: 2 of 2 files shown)


GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -61,6 +60,13 @@ public void Run()
 
                     this.serviceStopped.WaitOne();
                 }
+
+                // https://github.com/Microsoft/VFSForGit/issues/774.
+                // Start product upgrade timer only after attempting to enable prjflt.
+                // On Windows server (where PrjFlt is not inboxed) this helps avoid
+                // a race between TryEnablePrjFlt() and installer pre-check which is
+                // performed by UpgradeTimer in parallel.
+                this.productUpgradeTimer.Start();
```
kewillford: Do we need to check the error from the `TryEnablePrjFlt`?  Doesn't look like we were doing anything with it before which is suspicious.

GVFS/GVFS.Service/GVFSMountProcess.cs
```diff
@@ -27,7 +27,7 @@ public bool Mount(string repoRoot)
                 string error;
                 if (!EnableAndAttachProjFSHandler.TryEnablePrjFlt(this.tracer, out error))
                 {
-                    this.tracer.RelatedError($"{nameof(this.Mount)}: Unable to start the GVFS.exe process: {error}");
+                    this.tracer.RelatedError($"{nameof(this.Mount)}: Could not enable ProjFlt: {error}");
```
wilbaker: Super minor suggestion: `ProjFlt` -> `PrjFlt`

GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -55,12 +54,21 @@ public void Run()
...
+                        // Make a best-effort to enable PrjFlt. Continue even if it fails.
+                        // This will be tried again when user attempts to mount an enlistment.
                         string error;
                         EnableAndAttachProjFSHandler.TryEnablePrjFlt(activity, out error);
                     }
 
                     this.serviceStopped.WaitOne();
                 }
+
+                // https://github.com/Microsoft/VFSForGit/issues/774.
```
wilbaker: I don't think there's any need to include the issue number that prompted this change.  You've done a good job of explaining the scenario in the rest of the comment.

GVFS/GVFS.Service/GvfsService.cs
alameenshah: This is by design. Failure in `TryEnablePrjFlt` should not stop service startup. Service makes the best-effort, logs failure and proceeds. When user mounts an enlistment, `gvfs mount` will again try to enable `PrjFlt`. 

Added a comment to document this behavior. Also corrected the error messaging in `gvfs mount` when enable `PrjFlt` fails.

PR #819, opened 2019-02-21
Mac kext: Fix for race conditions in virtualization root lookup & provider client cleanup

pmj:
The function `FindRootAtVnode_Locked()` is called while holding a shared lock from `FindOrDetectRootAtVnode()`. However, very rarely, when a root's vnode is recycled and the new one is encountered, `FindRootAtVnode_Locked()` will refresh the registered vnode in the virtualization root. When only the shared lock is held, this could potentially cause a race between 2 or more threads. In theory, all threads involved should reach the same conclusion and update to the same vnode, so there should not be any actual data corruption.

This change fixes the race by not refreshing the vnode in `FindRootAtVnode_Locked()\ and instead performing the refresh in a new helper function which is only called when the lock is held exclusively.

**Update:** The second commit fixes the possible race condition [pointed out by @wilbaker ](https://github.com/Microsoft/VFSForGit/pull/819#issuecomment-466213421).

**Reviewer note:** I found this while trying to work out what was causing #797. ~~I *think* this race condition shouldn't be able to cause that bug, or indeed cause any other kind of malfunction because all threads would converge on the same result. Nevertheless, I'd appreciate a few extra sets of eyes & brains on analysing this race. I'm reluctant to merge this fix as long as we haven't worked out #797, but given that I might be wrong about this race being benign, perhaps someone can spot how it might have caused #797 after all, so I'm posting it now.~~

**Update:** Given that the cause for #797 is understood and a fix (#856) has been implemented, I think we should fix the discovered race conditions as well, even if they most probably are not causing any problems.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -43,6 +43,8 @@ static VirtualizationRoot* s_virtualizationRoots = nullptr;
 // Looks up the vnode/vid and fsid/inode pairs among the known roots
 static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t vid, FsidInode fileId);
 
+static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHandle, vnode_t vnode, uint32_t vid, FsidInode fileId);
+
 // Looks up the vnode and fsid/inode pair among the known roots, and if not found,
 // detects if there is a hitherto-unknown root at vnode by checking attributes.
 static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t vnode, const FsidInode& vnodeFsidInode);
@@ -210,10 +212,15 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode, 
     uint32_t vid = vnode_vid(vnode);
     
     VirtualizationRootHandle rootIndex;
+    bool rootVnodeStale = false;
     
     RWLock_AcquireShared(s_virtualizationRootsLock);
     {
         rootIndex = FindRootAtVnode_Locked(vnode, vid, vnodeFsidInode);
+        rootVnodeStale =
+            rootIndex >= 0
+            && (s_virtualizationRoots[rootIndex].rootVNode != vnode
+                || s_virtualizationRoots[rootIndex].rootVNodeVid != vid);
     }
     RWLock_ReleaseShared(s_virtualizationRootsLock);
     
@@ -254,6 +261,14 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode, 
         {
             KextLog_FileError(vnode, "FindOrDetectRootAtVnode: Vnode_ReadXattr/mac_vnop_getxattr failed with errno %d", xattrResult.error);
         }
+    }
+    else if (rootVnodeStale)
+    {
+        RWLock_AcquireExclusive(s_virtualizationRootsLock);
+        {
+            RefreshRootVnodeIfNecessary_Locked(rootIndex, vnode, vid, vnodeFsidInode);
+        }
+        RWLock_ReleaseExclusive(s_virtualizationRootsLock);
     }
     
     return rootIndex;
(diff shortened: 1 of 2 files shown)


wilbaker:
>I'd appreciate a few extra sets of eyes & brains on analysing this race.

@pmj I *might* have found another race, but I need your expertise because I can't find much documentation on the IOKit interface (if you know of any good documentation, please point me to it 😄 ).

Is it possible that `stop` and `clientClose` will be called at the same time on different threads?

```
void PrjFSProviderUserClient::stop(IOService* provider)
{
    // On the code path of process dying or disconnecting, the following is unnecessary,
    // but if the kext is unloaded with providers attached, clientClose() will never be
    // called on them so we do the cleanup here.
    this->cleanupProviderRegistration();
    
    this->super::stop(provider);
}
```

```
// Called when the user process explicitly or implicitly (process death) closes
// the connection.
IOReturn PrjFSProviderUserClient::clientClose()
{
    this->cleanupProviderRegistration();
    
    this->terminate(0);
    return kIOReturnSuccess;
}
```

I noticed that `cleanupProviderRegistration` is not thread safe, and so if it was running on two threads at once we could end up calling `ActiveProvider_Disconnect` twice for the same root:

```
void PrjFSProviderUserClient::cleanupProviderRegistration()
{
    VirtualizationRootHandle root = this->virtualizationRootHandle;
    this->virtualizationRootHandle = RootHandle_None;
    if (RootHandle_None != root)
    {
        ActiveProvider_Disconnect(root, this);
    }
}
```

The one thing that makes me doubt this is the root cause is that I would have expected `assert(nullptr != root->providerUserClient);` to fire in `ActiveProvider_Disconnect` rather than making all the way to `vnode_put(root->rootVNode);`

wilbaker:
> I might be wrong about this race being benign, perhaps someone can spot how it might have caused #797 after all, so I'm posting it now.

So far I've come to the same conclusion as you and not been able to find any evidence that the race would cause problems.

pmj:
> Is it possible that `stop` and `clientClose` will be called at the same time on different threads?

Hmm, interesting thought. I have to admit I'm not 100% sure, and a quick inspection of the code doesn't lead me to an immediate, obvious answer. I can say this:

 * There are thread-safe guards in place against `clientClose()` being called twice on the same user client.
 * `stop()` will in our case only be called on a separate thread from the one closing the client if the kext is unloaded with `kextunload`. I think we can rule that out in the case where the crash happened.

I don't have deep enough knowledge of the Mach mechanisms behind `IOServiceClose()` and such to say for sure whether the service terminations kicked off from `kextunload` are serialised with IOKit user calls, so I think it's worth making `cleanupProviderRegistration()` thread safe, so I'll follow up with a PR for that. I don't think the potential race there is the cause for #797, however.

pmj:
I've now rebased this on top of the actual fix for #797 (see #856) and pushed a second commit that fixes the possible race condition pointed out by @wilbaker in review.

pmj:
I've rebased this on latest master and #856 and fixed another pointer unslide in an assert. So this PR is only about commits b63b8f4 and 8437d6b, which are now ready for review.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -199,10 +212,15 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode,
     uint32_t vid = vnode_vid(vnode);
     
     VirtualizationRootHandle rootIndex;
+    bool rootVnodeStale = false;
     
     RWLock_AcquireShared(s_virtualizationRootsLock);
     {
         rootIndex = FindRootAtVnode_Locked(vnode, vid, vnodeFsidInode);
+        rootVnodeStale =
```
wilbaker: Should we (or do we) need to make this check every time we call `FindRootAtVnode_Locked`? I saw we're not on line 246.

pmj: We are holding an iocount on the vnode; so if `FindRootAtVnode_Locked()` on line 246 finds an entry, that means another thread has inserted the same vnode as a root between us dropping the shared lock on line 225 and reacquiring the exclusive lock on 243. Due to the iocount, the vnode cannot possibly have been recycled and gone stale in that time.

pmj: Incidentally, it's no big deal if the vnode does go stale and we *don't* refresh it. I'm contemplating removing the lookup by vnode altogether, and simply always using the fsid/inode pair to identify a root. In theory, with the vnode based lookup, we can avoid fetching the inode, but the way the code is currently structured, we always fetch the inode anyway. I've not done anything radical like that for now though as the vnode cache will cause major changes in this area again anyway.

Issue #380, opened 2018-10-12
PrefetchVerbTests.PrefetchCleansUpStalePrefetchLock fails intermittently


nickgra:
While refactoring code to light up 'prefetch --hydrate', I noticed that on Windows, this test fail intermittently. I have only seen it fail when the other LockHolder tests are failing (see #176), but often the LockHolder tests will fail without it.

See https://gvfs.visualstudio.com/ci/_build/results?buildId=2617&view=logs for one sample:

```
2018-10-12T12:36:50.1053529Z Errors, Failures and Warnings
2018-10-12T12:36:50.1060127Z 
2018-10-12T12:36:50.1060828Z 1) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.GitReadAndGitLockTests.GitAliasNamedAfterKnownCommandAcquiresLock()
2018-10-12T12:36:50.1061496Z   Expected substring 'Waiting for 'GVFS.FunctionalTests.LockHolder'' not found in ''
2018-10-12T12:36:50.1061932Z   Expected: True
2018-10-12T12:36:50.1062282Z   But was:  False
2018-10-12T12:36:50.1062896Z    at GVFS.Tests.Should.StringShouldExtensions.ShouldContain(String actualValue, String[] expectedSubstrings) in C:\agent\_work\2\s\GVFS\GVFS.Tests\Should\StringShouldExtensions.cs:line 19
2018-10-12T12:36:50.1064058Z    at GVFS.FunctionalTests.Tests.EnlistmentPerFixture.GitReadAndGitLockTests.GitAliasNamedAfterKnownCommandAcquiresLock() in C:\agent\_work\2\s\GVFS\GVFS.FunctionalTests\Tests\EnlistmentPerFixture\GitReadAndGitLockTests.cs:line 71
2018-10-12T12:36:50.1064701Z 
2018-10-12T12:36:50.1065616Z 2) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.GitReadAndGitLockTests.GitAliasInSubfolderNamedAfterKnownCommandAcquiresLock()
2018-10-12T12:36:50.1066354Z   Expected substring 'Waiting for 'GVFS.FunctionalTests.LockHolder'' not found in 'error: Failed to merge in the changes.
2018-10-12T12:36:50.1066862Z '
2018-10-12T12:36:50.1067173Z   Expected: True
2018-10-12T12:36:50.1067573Z   But was:  False
[...]

PR #820, opened 2019-02-21
Mac: Matches up KextLog levels with os_log levels

pmj:
The existing levels in the KextLog mechanism have repeatedly [caused confusion](https://github.com/Microsoft/VFSForGit/pull/805#discussion_r258728471) with regard to their relative importance and whether or not they hit the on-disk system log.

 * The "Note" level has been replaced with "Default" to match its os_log counterpart. The corresponding macros have been renamed to KextLog() and KextLog_File(), so "Default" need not be spelled out explicitly.
 * Comments have been added to the log level enum definition to explain the syslog policy.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/public/PrjFSLogClientShared.h
+++ b/ProjFS.Mac/PrjFSKext/public/PrjFSLogClientShared.h
@@ -24,11 +24,12 @@ enum PrjFSLogUserClientPortType
     LogPortType_MessageQueue,
 };
 
+// Log message levels, these correspond to their os_log counterparts
 enum KextLog_Level : uint32_t
 {
-    KEXTLOG_ERROR = 0,
-    KEXTLOG_INFO = 1,
-    KEXTLOG_NOTE,
+    KEXTLOG_ERROR = 0,   // For important failures, always gets logged and causes INFO messages to be committed too.
+    KEXTLOG_DEFAULT = 2, // Run-of-the-mill messages that shouldn't be too frequent as they always are logged to disk.
+    KEXTLOG_INFO = 1,    // Verbose logging that might help with diagnosing problems; these don't usually get logged to disk, except when an ERROR is loggod.
 };
 
 struct KextLog_MessageHeader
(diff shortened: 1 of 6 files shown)


ProjFS.Mac/PrjFSKext/public/PrjFSLogClientShared.h
```diff
@@ -24,11 +24,12 @@ enum PrjFSLogUserClientPortType
...
 
+// Log message levels, these correspond to their os_log counterparts
 enum KextLog_Level : uint32_t
 {
-    KEXTLOG_ERROR = 0,
-    KEXTLOG_INFO = 1,
-    KEXTLOG_NOTE,
+    KEXTLOG_ERROR = 0,   // For important failures, always gets logged and causes INFO messages to be committed too.
+    KEXTLOG_DEFAULT = 2, // Run-of-the-mill messages that shouldn't be too frequent as they always are logged to disk.
+    KEXTLOG_INFO = 1,    // Verbose logging that might help with diagnosing problems; these don't usually get logged to disk, except when an ERROR is loggod.
```
wilbaker: This helps clear things up, thanks!

ProjFS.Mac/PrjFSLib/prjfs-log/prjfs-log.cpp
```diff
@@ -145,7 +145,7 @@ static const char* KextLogLevelAsString(KextLog_Level level)
         return "Error";
     case KEXTLOG_INFO:
         return "Info";
-    case KEXTLOG_NOTE:
+    case KEXTLOG_DEFAULT:
         return "Note";
```
wilbaker: Should we change "Note" to "Default" as well?

pmj: Of course, thanks for catching that. Fixed in the new version.

wilbaker:
Looks like the Windows functional tests hit #380, I will kick them off again

Issue #826, opened 2019-02-21
gvfs fetch slow with specific file list


turbonaitis:
If we do `gvfs prefetch --files "*"`, it takes around 21s, downloads all 25k blobs.
If we run `gvfs prefetch --files-list <list of files>`, where the list of files contains only 4k blobs, it takes 42 minutes. 

Looking at the logs, in the quick case, the batching works as expected, and we're downloading in batches around 3k (+-200) in size, so it all happens in 8 batches. 

In the slow case, the batches are rarely larger than 1 and never over 10. Durations per batch are only up to 40ms, so even with no parallelization, and if no batch had more than 1 object in it, it would take only 160s, which seems to indicate, that the pipeline stages, that come before the download step are being slow.

turbonaitis:
Closing, as the PR was merged

PR #827, opened 2019-02-22
Switching to HashSet for the list of files to be prefetched to improv…

turbonaitis:
…e performance, when the list is long.

Addreses #826 

Before this change, if we were prefetching a list of ~4k files, the traversal of that list in `ShouldIncludeResult` method took ~70ms. After the change, the lookup in the HashSet is almost instant (testing showed it to take 0ms)

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
+++ b/GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
@@ -13,7 +13,8 @@ namespace GVFS.Common.Prefetch.Git
         private const string AreaPath = nameof(DiffHelper);
 
         private ITracer tracer;
-        private List<string> fileList;
+        private HashSet<string> exactFileList;
+        private List<string> patternList;
         private List<string> folderList;
         private HashSet<string> filesAdded = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -31,7 +32,8 @@ public DiffHelper(ITracer tracer, Enlistment enlistment, IEnumerable<string> fil
         public DiffHelper(ITracer tracer, Enlistment enlistment, GitProcess git, IEnumerable<string> fileList, IEnumerable<string> folderList, bool includeSymLinks)
         {
             this.tracer = tracer;
-            this.fileList = new List<string>(fileList);
+            this.exactFileList = new HashSet<string>(fileList.Where(x => !x.StartsWith("*")).Select(x => x.ToLower()));
```
kewillford: Should use `new HashSet<string>(fileList.Where(x => !x.StartsWith("*")), StringComparer.OrdinalIgnoreCase)` instead of having to use `ToLower()` here and in `Contains` below.

turbonaitis: Good point, changed.

jrbriggs:
Thanks for the contribution!

PR #829, opened 2019-02-22
Improved MaintanceStep Handling

jeschu1:
This change is different than the one on shipped, to better address the issue

- Stop MaintenceStep Process during shutdown
  - We throw an exception when we flip to Stopping.  I've tried to update the code to make the process more obvious.  We use an exception so we don't have put if checks everywhere.  There is a single place where we need to catch.
  - Thread.Abort is another option to using Stopping.  However Thread.Abort is quite a hammer, so we're gracefully trying to terminate.  I'm calling that out as something we may decide to change our mind on in the future.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -127,7 +127,7 @@ namespace GVFS.Common.Maintenance
                         metadata: null,
                         message: $"{this.Area}: Not launching Git process {gitCommand} because the mount is stopping",
                         keywords: Keywords.Telemetry);
-                    return null;
+                    throw new StoppingException();
                 }
 
                 GitProcess.Result result = work.Invoke(this.MaintenanceGitProcess);
@@ -234,7 +234,18 @@ namespace GVFS.Common.Maintenance
                 this.MaintenanceGitProcess = this.Context.Enlistment.CreateGitProcess();
             }
 
-            this.PerformMaintenance();
+            try
+            {
+                this.PerformMaintenance();
+            }
+            catch (StoppingException)
+            {
+                // Part of shutdown, skipped commands have already been logged
+            }
+        }
+
+        protected class StoppingException : Exception
+        {
         }
     }
 }
--- a/GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
@@ -153,7 +153,7 @@ namespace GVFS.Common.Maintenance
             maxGoodTimestamp = -1;
 
             int firstBadPack = -1;
-            for (int i = 0; !this.Stopping && i < orderedPacks.Count; ++i)
+            for (int i = 0; i < orderedPacks.Count; ++i)
             {
                 long timestamp = orderedPacks[i].Timestamp;
                 string packPath = orderedPacks[i].Path;
@@ -165,12 +165,6 @@ namespace GVFS.Common.Maintenance
                     metadata.Add("idxPath", idxPath);
                     metadata.Add("timestamp", timestamp);
                     GitProcess.Result indexResult = this.RunGitCommand(process => this.GitObjects.IndexPackFile(packPath, process), nameof(this.GitObjects.IndexPackFile));
-
-                    if (this.Stopping)
-                    {
-                        error = nameof(this.Stopping);
-                        return false;
-                    }
 
                     if (indexResult.ExitCodeIsFailure)
                     {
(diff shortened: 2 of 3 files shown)


GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -196,8 +199,10 @@ protected void LogErrorAndRewriteMultiPackIndex(ITracer activity, GitProcess.Res
             string multiPackIndexPath = Path.Combine(this.Context.Enlistment.GitPackRoot, "multi-pack-index");
             errorMetadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(multiPackIndexPath);
 
-            GitProcess.Result rewriteResult = this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+            GitProcess.Result rewriteResult = this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot), "WriteMultiPackIndex");
```
wilbaker: Minor: Here and below, switch from string constants (e.g. "WriteMultiPackIndex") to `nameof` (e.g. `nameof(process.WriteMultiPackIndex)`.

This comment applies in a few different spots/files.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
```diff
@@ -72,11 +72,13 @@ public CheckMethodStep(GVFSContext context)
 
             protected override void PerformMaintenance()
             {
-                this.RunGitCommand(process =>
-                {
-                    this.SawWorkInvoked = true;
-                    return null;
-                });
+                this.RunGitCommand(
```
wilbaker: Note: This comment is really for line 29, but I can't comment there ☹️ 

If any of the unit tests in this class will now hit `throw new StoppingException();` in the product code they should be tagged with `[Category(CategoryConstants.ExceptionExpected)]`

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -151,8 +153,13 @@ private bool TryGetMaxGoodPrefetchTimestamp(out long maxGoodTimestamp, out strin
             maxGoodTimestamp = -1;
 
             int firstBadPack = -1;
-            for (int i = 0; !this.Stopping && i < orderedPacks.Count; ++i)
+            for (int i = 0; i < orderedPacks.Count; ++i)
             {
+                if (this.Stopping)
```
wilbaker: Do we need this check here?  

If we're stopping `this.RunGitCommand` (on line 172) will throw a `StoppingException`, and if we skip that call to `RunGitCommand` we'll throw a `StoppingException` on line 198.

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -192,7 +193,12 @@ private bool TryGetMaxGoodPrefetchTimestamp(out long maxGoodTimestamp, out strin
                 }
             }
 
-            if (!this.Stopping && firstBadPack != -1)
+            if (this.Stopping)
+            {
+                throw new StoppingException();
```
wilbaker: For my own understanding, what would be the downside to not throwing any `StoppingException` here (or below) in this method?

Is the concern that `gvfs unmount` might take a long time if we need to retry deletes?  It looks like we don't launch any git processes from this point on.

jeschu1: @wilbaker , yes that was my concern.  

I am not convinced how necessary it is but wanted to stay consistent with the old code.  

We don't need to rush on it now but a more dangerous, but less verbose method of stopping would be to Abort the thread once we kill the GitProcess.

wilbaker: Ok thanks, I think it makes sense to stay consistent with how the old code was checking.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
jeschu1: Since the exception is catch (never leaks to the test) I don't I should add the tag right?

wilbaker: We've been tagging unit tests with `ExceptionExpected` when any of the code that runs in the test throws an exception.

We started doing that because running the unit tests with the debugger attached became an exercise in pressing continue to get through all of the expected exceptions.

The `ExceptionExpected` just prevents the test from being run by default when the debugger is attached.

PR #838, opened 2019-02-23
[releases/shipped] Use a new log directory for upgrade logs to avoid ACL issues

### LINKED ISSUE #837: GVFS.Service's renaming of GVFS.Upgrade\Logs fails when the upgrader process has an open handle to a log file inside of GVFS.Update\Logs
Steps to reproduce:

1. Run `gvfs upgrade` with an old version of VFS4G so that a non-admin is the owner of C:\ProgramData\GVFS.Upgrade\Logs
2. Run 'gvfs upgrade --confirm`
3. The upgrade process will create a log file inside of C:\ProgramData\GVFS.Upgrade\Logs, and that open handle will prevent GVFS.Service from renaming the directory (see https://github.com/Microsoft/VFSForGit/blob/d05248317c354f3bdb22c8cea55fa28063ffeaf4/GVFS/GVFS.Service/GvfsService.cs#L423).

Rather than trying to fix the ACLs on existing the existing "Logs" directory, it might be more reliable to simply create a new logs directory and have `gvfs diagnose` capture both.

wilbaker:
Fixes #837 

This address the issue in #837 by switching to a new directory for upgrader logs.  GVFS.Service will create the directory if it does not exist (with the proper ACLs) and prevent the directory from inheriting ACLs (when GVFS.Service updates them on C:\ProgramData\GVFS).

Note: There is no automatic healing if someone manually adjusts the ACLs on the directory.


wilbaker:
It looks like there is still a problem in this scenario:

1. Upgrade to a version with these changes
2. GVFS.Service never starts
3. User runs `gvfs upgrade` from an elevated CMD prompt
4. The new logs directory will not have the correct ACLs, and it will prevent non-elevated users from being able to create upgrade log files.

wilbaker:
Closing this PR and creating a new PR on top of 9d6e1bdd8bbd6a23bc1ab43e144c77e3df04ebce.

9d6e1bdd8bbd6a23bc1ab43e144c77e3df04ebce is the commit containing all of the ACL changes that was merged into both master and releases/shipped

PR #802, opened 2019-02-19
Add tests for ShouldIgnoreVnodeType

jeschu1:
Adds tests for ShouldIgnoreVnodeType

PR #839, opened 2019-02-24
Mac kext: Expanded test infrastructure, added virtualization root unit tests

pmj:
This is based on top of @jeschu1 's PR ~~#732~~ #802.

In my hunt for the source of #797, it seemed sensible to create unit tests for the functions likely to be the cause.

Those functions, `VirtualizationRoot_RegisterProviderForPath`, `ActiveProvider_Disconnect`, and `FindRootAtVnode_Locked` call out to a lot of functions including kernel APIs, so getting tests to build and run has required a fair bit of infrastructure work.

Specifically, this sequence of commits adds:

 * Basic mocking of `vnode`s and `mount` points.
 * Stub implementations for some memory allocation and lock functions. (These could get a lot fancier.)
 * Function call recording so we can verify that a mock function was or wasn't called by a test. (This currently does not support checking call order or argument values, but is in principle designed to be extended that way.)
 * 100% test coverage of `VirtualizationRoot_RegisterProviderForPath`
 * Test coverage recording. In Xcode, test coverage is shown for each code block on the right side of the editor window.

Still missing: tests for `ActiveProvider_Disconnect` and infrastructure required to get them working.

The tests found a few bugs, they weren't the cause for #797 though, and they only kick in when you hit 32768 virtualisation roots in the system. There's lots of useful testing infrastructure there though that made this a worthwhile effort.

**Reviewer notes:** This is a pretty large PR; for now I'm mostly interested in what you think of the general approach, rather than picking it apart line-by line.


ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
```diff
@@ -40,7 +40,7 @@
 		4391F8B621E42AC50008103C /* PrjFSKext.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 4391F89C21E42AC40008103C /* PrjFSKext.cpp */; };
 		4391F8B721E42AC50008103C /* PrjFSClasses.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 4391F89D21E42AC50008103C /* PrjFSClasses.hpp */; };
 		4391F8B821E42AC50008103C /* PrjFSLogUserClient.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 4391F89E21E42AC50008103C /* PrjFSLogUserClient.hpp */; };
-		4391F8B921E42AC50008103C /* PrjFSProviderUserClient.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 4391F89F21E42AC50008103C /* PrjFSProviderUserClient.hpp */; };
+		4391F8B921E42AC50008103C /* PrjFSProviderUserClientPrivate.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 4391F89F21E42AC50008103C /* PrjFSProviderUserClientPrivate.hpp */; };
```
alameenshah: haven't seen this pattern else where in `vfs4git`- what does Private specifier mean in a file name?

pmj: You're right, we haven't used this elsewhere yet, [except in PR 500 which has yet to be merged](https://github.com/Microsoft/VFSForGit/pull/500#discussion_r256141241). In the `PrjFSProviderUserClientPrivate.hpp` case, it's purely to signify that the definitions there are not-API (opaque implementation detail) for most of the rest of the kext. For `VirtualizationRootsPrivate.hpp`, in addition to being not-API for the rest of the kext, it allows the unit tests to access a private struct definition in order to be able to set up specific test scenarios.

pmj: After out-of-band discussion, it was decided we'll go ahead with the "…Private.hpp" approach, and have the private and testable headers side-by-side; testable will continue to contain the prototypes for functions that are private but need to be accessed from tests.

ProjFS.Mac/PrjFS.xcodeproj/xcshareddata/xcschemes/Build All.xcscheme
```diff
@@ -26,7 +26,25 @@
...
+            ReferencedContainer = "container:PrjFS.xcodeproj">
+         </BuildableReference>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "F5E39C7621F1118D006D65C2"
+            BuildableName = "PrjFSKextTests.xctest"
+            BlueprintName = "PrjFSKextTests"
+            ReferencedContainer = "container:PrjFS.xcodeproj">
+         </BuildableReference>
+      </CodeCoverageTargets>
```
pmj: We need to limit code coverage to specific targets, as coverage instrumentation is not supported for kexts and will screw up the build.

ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
```diff
@@ -1,7 +1,6 @@
 #include "public/PrjFSCommon.h"
 #include <sys/kernel_types.h>
-#include <sys/vnode.h>
```
pmj: From now on, kext & unit tests must never `#include <sys/vnode.h>` directly, and always go via the `kernel-header-wrappers/vnode.h` because the kernel version of the file is mutually exclusive with the user space one, and we need the kernel one for a bunch of declarations.

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.hpp
```diff
@@ -1,73 +1,9 @@
...
-        void* reference,
-        IOExternalMethodArguments* arguments);
-    IOReturn registerVirtualizationRoot(const char* rootPath, size_t rootPathSize, uint64_t* outError);
-
-    static IOReturn kernelMessageResponse(
-        OSObject* target,
-        void* reference,
-        IOExternalMethodArguments* arguments);
-    IOReturn kernelMessageResponse(uint64_t messageId, MessageType responseType);
-};
```
pmj: We do not want this class definition to leak into the testable build or unit tests.

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.hpp
```diff
@@ -1,73 +1,9 @@
...
+void ProviderUserClient_UpdatePathProperty(PrjFSProviderUserClient* userClient, const char* providerPath);
+void ProviderUserClient_Retain(PrjFSProviderUserClient* userClient);
+void ProviderUserClient_Release(PrjFSProviderUserClient* userClient);
+void ProviderUserClient_SendMessage(PrjFSProviderUserClient* userClient, const void* message, uint32_t size);
```
pmj: This means we need to use these wrapper functions to interact with opaque `PrjFSProviderUserClient` object pointers in all testable areas of the kext.

ProjFS.Mac/PrjFSKext/VirtualizationRootsPrivate.hpp
```diff
@@ -0,0 +1,25 @@
...
+    uint64_t                    rootInode;
+    
+    // TODO(Mac): this should eventually be entirely diagnostic and not used for decisions
+    char                        path[PrjFSMaxPath];
+};
+
+#ifdef KEXT_UNIT_TESTING
+extern uint16_t s_maxVirtualizationRoots;
+extern VirtualizationRoot* s_virtualizationRoots;
+#endif
```
pmj: TODO: move these to testable header. Struct definition stays here though.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
+#include <unordered_map>
+#include <unordered_set>
+
+class FunctionTypeCallRecorderBase
```
pmj: I struggled a bit with naming the classes and functions here, due to my limited exposure to extensive automated testing. I suspect there is existing terminology I should be applying here.

The idea is to be able to check whether specific mock functions were called or not by recording mock function calls in a central registry that can be interrogated after the test code has run. It currently doesn't support it, but the idea is to store the argument values too. ("Was function X called with an argument Y?") It already records calls with a sequence number; right now that number isn't used, but the idea is to be able to verify that e.g. a test calls F1() before F2().

ProjFS.Mac/PrjFSKextTests/TestLocks.cpp
```diff
@@ -0,0 +1,37 @@
+#include "../PrjFSKext/Locks.hpp"
+
```
pmj: In this file, I did the minimum required to get the tests to build & run. We could of course get fancy.

ProjFS.Mac/PrjFSKextTests/TestMemory.cpp
```diff
@@ -0,0 +1,11 @@
+#include "../PrjFSKext/Memory.hpp"
+
```
pmj: Much opportunity for fanciness here too, but that can be added later.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -0,0 +1,264 @@
...
+#include "../PrjFSKext/Memory.hpp"
+#include "../PrjFSKext/Locks.hpp"
+#include "../PrjFSKext/VnodeUtilities.hpp"
+#include "../PrjFSKext/KextLog.hpp"
+#include "KextMockUtilities.hpp"
+#include "MockVnodeAndMount.hpp"
+#include <vector>
+#include <string>
+
+class PrjFSProviderUserClient
```
pmj: Don't go near the real IOKit based implementation of this.

ProjFS.Mac/PrjFSKext/public/FsidInode.h
```diff
@@ -2,6 +2,7 @@
 #define FsidInode_h
 
 #include <sys/_types/_fsid_t.h>
+#include <stdint.h>
```
jeschu1: Why did we need to add this include?

jeschu1:
Can you refresh my memory on why we need both KEXT_UNIT_TESTING and TESTABLE_KEXT_TARGET?

jeschu1:
@pmj I really like the approach!  I'm going thru the code now and asking any questions I have.

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.hpp
jeschu1: Makes sense!  Wrappers for testing.
For my own understanding would mocks have been an option here or would it have been too involved/messy.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -250,7 +232,7 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode,
 
 static VirtualizationRootHandle FindUnusedIndex_Locked()
 {
-    for (VirtualizationRootHandle i = 0; i < s_maxVirtualizationRoots; ++i)
+    for (uint32_t i = 0; i < s_maxVirtualizationRoots; ++i)
```
jeschu1: :-) We were using a handle as int?

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
jeschu1: @Pmj if you're looking for inspiration [moq](https://github.com/moq/moq4) is a very popular c# testing framework.

I can understand what you've done so far.  I think it can grow organically over time as needed.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    return vnode->type;
+}
+
+mount_t vnode_mount(vnode_t vnode)
+{
+    return vnode->mountPoint.get();
+}
+
+int vnode_put(vnode_t vnode)
```
jeschu1: Asking to confirm my understanding of put/get.
put - removes
get - inserts

If we have a get unpaired with a put, what could happen?
if we have put without a get, what could happen?

ProjFS.Mac/PrjFSKextTests/TestLocks.cpp
jeschu1: I think this is an exellent first step and I wouldn't get fancy yet :-)

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -0,0 +1,264 @@
...
+    XCTAssertFalse(MockCalls::DidCallFunction(vfs_setauthcache_ttl));
+}
+
+- (void)testRegisterProviderForPath_DisallowedFileSystem
+{
+    const char* path = "/Volumes/USBStick/repo";
+    
+    fsid_t fatFsid = self->testMountPoint->GetFsid();
+    fatFsid.val[1]++;
+    MountPointer fatMount(mount::Create("msdos", fatFsid, rand()));
```
jeschu1: arguably this could remove the need to test VirtualizationRoot_VnodeIsOnAllowedFilesystem directly.

If we want to do that we'd need a test for both "hfs" and "apfs".

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.hpp
pmj: Doing anything else would have caused issues due to the way the testable kext target is compiled - that one pulls in the real kernel headers, so it would really have been an `IOUserClient` subclass. `IOUserClient` (inheritance chain: `IOUserClient : IOService : IORegistryEntry : OSObject : OSMetaClassBase`) literally has dozens of fields and even more member functions. Because the testable kext is built against the kernel headers, this means the in-memory layouts of the class itself and the vtable are all fixed from those. So any mock implementations would have to match those layouts 1:1. We'd have to provide dummy implementations for all those member functions, etc. Just far more effort than it's worth.

ProjFS.Mac/PrjFSKext/public/FsidInode.h
pmj: It was complaining about `uint64_t` not being defined when included from the unit tests, if I remember correctly.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Well, it's typedef’d to an `int16_t`. But yeah, turns out, not a good idea.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: `vnode_get` and `vnode_put` increment and decrement the iocount in the vnode. This roughly corresponds to opening and closing file descriptors.

vnodes can only become "dead" (`VBAD`) if the iocount is 0. This is a prerequisite for recycling, so if we were missing a `vnode_put`, we'd essentially be leaking vnodes. Worse, file system mount points (OS ones, not VFS4G ones) can only be unmounted if all vnodes for files on them are dead. So that prevents ejecting external volumes, etc.

If you call `vnode_put()` on a vnode with an iocount of 0, that causes a kernel panic, see #797.

jeschu1: @jrbriggs this is a very elegant explanation of the get/put requirement.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: I think we should definitely keep the unit tests for `VirtualizationRoot_VnodeIsOnAllowedFilesystem`. If anything, we should probably allow swapping out the `VirtualizationRoot_VnodeIsOnAllowedFilesystem` function itself for a test double to make this test here simpler.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
pmj: > if you're looking for inspiration [moq](https://github.com/moq/moq4) is a very popular c# testing framework.

Thanks, I'll check it out!

pmj:
> Can you refresh my memory on why we need both KEXT_UNIT_TESTING and TESTABLE_KEXT_TARGET?

In the unit test build target, we're using "normal" user space system headers. In the "PrjFSKextTestable" build target, we're using kernel headers so that the kext code compiles without modification. Both targets define `KEXT_UNIT_TESTING`, but we do sometimes need to know which one is the one currently being compiled, mostly so we can work around compile errors in system headers, such as in the `vnode.h` and `mount.h` wrappers.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -0,0 +1,78 @@
...
+    static VnodePointer Create(const MountPointer& mount, const char* path, vtype vnodeType = VREG);
+    static VnodePointer Create(const MountPointer& mount, const char* path, vtype vnodeType, uint64_t inode);
+    ~vnode();
+    
+    uint64_t GetInode() const { return this->inode; }
+    uint32_t GetVid() const { return this->vid; }
+    void SetGetPathError(errno_t error);
+    void StartRecycling();
+
+    friend int vnode_isrecycled(vnode_t vnode);
```
jeschu1: Should these be friend?

ProjFS.Mac/PrjFSKext/public/FsidInode.h
pmj: I've just checked, and yes, if you remove that `#include`, it complains about `uint64_t inode;` a few lines down when building the unit tests.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
pmj: Yeah, that ended up messy. Where it's sensible, I'll just make the fields they access in `struct vnode` public, and where it makes more sense, I'll add some public member functions that those mocked functions can call.

jeschu1:
@wilbaker I was able to extend this framework pretty easily for testing TryGetFileIsFlaggedAsInRoot.

POC here https://github.com/pmj/VFSForGit/pull/1/files

ProjFS.Mac/PrjFSKext/kernel-header-wrappers/vnode.h
```diff
@@ -1,7 +1,21 @@
 #pragma once
 
+#if defined(KEXT_UNIT_TESTING) && !defined(TESTABLE_KEXT_TARGET) // Building unit tests
+
+#include <time.h> // vnode.h below references struct timespec
+// Definitions to make the kernel vnode.h compile in user space
+typedef struct ipc_port* ipc_port_t;
+enum uio_rw {};
+enum uio_seg {};
```
jeschu1: @jrbriggs the lines is 'magic' required to use the Kernel/sys/vnode.h header.  If we hadn't done this we'd be looking at redefining kext objects.  @pmj feel free to elaborate more.

pmj: We're essentially cheating by pulling the kernel versions of *just* `sys/vnode.h` and `sys/mount.h`. There are cut-down user space versions of those files but they don't contain any prototypes for the in-kernel functions of course. (`vnode_getattr()` and such) So by pulling in the kernel versions of those headers we avoid having to duplicate the prototypes, which saves us some work but has an extra advantage in that we can't make mistakes in the signature of the mocked versions of those functions without the compiler yelling at us.

Of course, those kernel headers have dependencies for various types etc. themselves, but those aren't all met by the userspace versions of the headers they include. So we have to fake them sufficiently to get the compiler to stop complaining. This isn't a problem as we don't actually use any of the APIs that use these specific types.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
jeschu1: There is a train of though on unit tests that private functions should never be tested (not part of the interface).  Of course for the kext with the majority being private that really doesn't hold well.  

I'd be careful with mocking out`VirtualizationRoot_VnodeIsOnAllowedFilesystem`.  While it's true we actually testing two functions at the same time, it's also how it's used.  I've got mixed feelings there.  Once we get more converge this would be great to discuss in more detail :-)!

ProjFS.Mac/PrjFS.xcodeproj/xcshareddata/xcschemes/Build All.xcscheme
wilbaker: Where in Xcode is the coverage shown?

I'm on you're branch and I've confirmed that the code coverage is set to these two projects (in the GUI), but I've not been able to figure out where I need to go to see anything about the coverage.

wilbaker: @nickgra helped me figure this out.  I had to:

1. View->Navigators->Show Report Navigator
2. Run the tests
3. Select (in the tree-control on the left) Build All -> Test -> Coverage
4. Check "Show Test Bundles" in the top-right corner

wilbaker: In case someone else hits this, another setting that might need adjusting is Editor -> Show Code Coverage

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -0,0 +1,79 @@
...
+    class FunctionTypeCallRecorder : public FunctionTypeCallRecorderBase
+{
+    typedef R (*FunctionPointerType)(ARGS...);
+    struct FunctionCall
+    {
+        // TODO: add argument values
+        uint64_t callSequenceNumber;
+    };
+    
+    typedef std::unordered_multimap<FunctionPointerType, FunctionCall> RecordedCallMapType;
```
wilbaker: Another minor naming suggestion, I think you could drop the "Type" from `RecordedCallMapType` (i.e. `RecordedCallMapType` -> `RecordedCallMap`).

I saw that our `typedef std::unordered_map` typedefs don't use "Type" in the type name.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -0,0 +1,79 @@
...
+    virtual void Clear() override
+    {
+        this->recordedCalls.clear();
+    }
+    
+    friend class MockCalls;
+};
+
+template <typename R, typename... ARGS>
+    FunctionTypeCallRecorder<R, ARGS...> FunctionTypeCallRecorder<R, ARGS...>::functionTypeRegister;
```
wilbaker: I'm a bit confused about how this is working.  

What prevents each each compilation unit that includes KextMockUtilities.hpp from getting its own instance of `functionTypeRegister`? 

(I am assuming that MockVnodeAndMount.cpp and VirtualizationRootTests.mm are each individual units)

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "../PrjFSKext/kernel-header-wrappers/vnode.h"
+#include "../PrjFSKext/kernel-header-wrappers/mount.h"
+#include "../PrjFSKext/public/FsidInode.h"
+#include <memory>
+#include <string>
+
+typedef std::shared_ptr<mount> MountPointer;
+typedef std::shared_ptr<vnode> VnodePointer;
+typedef std::weak_ptr<vnode> VnodeWeakPointer;
```
wilbaker: Style suggestion: This might just be me, but I'm not sure we're gaining much readability by typedefing  `shared_ptr` and `weak_ptr`.

If we are going to keep the typedefs then I think we should include "Shared" in the names of the `shared_ptr` types (as we have "Weak" in the `weak_ptr` type).

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -0,0 +1,75 @@
...
+struct mount
+{
+private:
+    vfsstatfs statfs;
+    uint64_t nextInode;
+    
+public:
+    static MountPointer Create(const char* fileSystemTypeName, fsid_t fsid, uint64_t initialInode);
+    
+    inline fsid_t GetFsid() const { return this->statfs.f_fsid; }
```
wilbaker: Nit: There is no need for `inline` here and we're not using it for the functions below (in `vnode`)

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -0,0 +1,235 @@
...
+}
+
+vnode::~vnode()
+{
+    assert(this->ioCount == 0);
+}
+
+VnodePointer vnode::Create(const MountPointer& mount, const char* path, vtype vnodeType)
+{
+    VnodePointer result(new vnode(mount));
```
wilbaker: Minor: can we use `make_shared` here?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -0,0 +1,235 @@
...
+    if (lastSlash == string::npos)
+    {
+        this->name = this->path.c_str();
+    }
+    else
+    {
+        this->name = this->path.c_str() + lastSlash + 1;
+    }
+    
+    assert(!s_vnodesByPath[path].lock());
```
wilbaker: A comment here would be useful as well.  As best I can tell this `assert` is validating that this is the first `vnode` being recorded in `s_vnodesByPath` for this path, is that correct?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -0,0 +1,235 @@
...
+void vfs_setauthcache_ttl(mount_t mountPoint, int ttl)
+{
+    MockCalls::RecordFunctionCall(vfs_setauthcache_ttl, mountPoint, ttl);
+}
+
+void MockVnodes_CheckAndClear()
+{
+    for (WeakVnodeMap::const_iterator cur = s_allVnodes.begin(); cur != s_allVnodes.end(); ++cur)
+    {
+        VnodePointer strong = cur->second.lock();
```
wilbaker: A comment here would be useful too.  

Would this comment be accurate?

```
// All of the vnodes in s_allVnodes should destroyed by the time MockVnodes_CheckAndClear is called
```

ProjFS.Mac/PrjFSKextTests/TestMemory.cpp
pmj: I don't have a great solution for this yet, but I'd like to redefine `assert()` in test(able) code to call down to the XCTest equivalent. The XCTest asserts are a bit awkward because they expect `self` in the surrounding scope to refer to the test case object. I suspect we might want to store the current test in a thread-local variable or so and use that in some custom macros instead. Using regular `assert()` will currently just take down the whole process, which isn't quite as ergonomic with regard to pinpointing test failures.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: For `VirtualizationRoot_VnodeIsOnAllowedFilesystem` it probably doesn't matter too much, but I think we'll need to be able to swap out some of the other functions, as always testing the whole call subtree via its side effects is going to be awkward.

The public/private distinction isn't terribly helpful in our kext; the only truly public APIs are the kext start/stop functions, the callbacks we register with the kauth listener, and the IOKit userclient classes. We're also treating certain compilation units as class-like with APIs that are visible to the rest of the kext vs local static functions, but I think those only superficially map to class-based OOP design; we'd probably have more, smaller classes in practice if we were writing this in C# or Java or whatever.

ProjFS.Mac/PrjFSKext/Memory.hpp
```diff
@@ -1,7 +1,12 @@
 #ifndef Memory_h
 #define Memory_h
 
+#if !defined(KEXT_UNIT_TESTING) || defined(TESTABLE_KEXT_TARGET) // Building kext
 #include <kern/assert.h>
+#else // building unit tests/mocks
+#include <cassert>
+#include <mach/kern_return.h>
```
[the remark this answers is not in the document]
pmj: > Can/should we always include `<mach/kern_return.h>`?

Yes, that's fair enough, and in fact we might want to wrap the assert includes in a wrapper header that behaves differently for different builds anyway, particularly if we want to integrate with the XCTest assert mechanism.

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClientPrivate.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include "PrjFSProviderUserClient.hpp"
+#include "Locks.hpp"
+#include "public/Message.h"
+#include "VirtualizationRoots.hpp"
+#include <IOKit/IOUserClient.h>
+
+struct MessageHeader;
+struct VirtualizationRoot;
+class IOSharedDataQueue;
+class PrjFSProviderUserClient : public IOUserClient
```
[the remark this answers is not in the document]
pmj: Yes, I believe that should work, although you don't even need the `||`, `KEXT_UNIT_TESTING` is defined in the testable kext target too.

ProjFS.Mac/PrjFSKextTests/KextLogMock.h
```diff
@@ -1,11 +1,3 @@
-#ifndef KextLogMock_h
-#define KextLogMock_h
+#pragma once
```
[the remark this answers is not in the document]
pmj: Depends what we want to do with logging statements in tests.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "../PrjFSKext/kernel-header-wrappers/mount.h"
+#include "../PrjFSKext/public/FsidInode.h"
+#include <memory>
+#include <string>
+
+typedef std::shared_ptr<mount> MountPointer;
+typedef std::shared_ptr<vnode> VnodePointer;
+typedef std::weak_ptr<vnode> VnodeWeakPointer;
+
+struct mount
```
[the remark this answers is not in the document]
pmj: Yes, that's right - we don't get a choice about those names. I'll add the comment.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
pmj: > What prevents each each compilation unit that includes KextMockUtilities.hpp from getting its own instance of `functionTypeRegister`?

This is as far as I'm aware a guarantee made by the C++ standard - there will only be one instance of a static member for any given template instantiation. I suppose this is comparable to the fact that you don't get duplicate symbol link errors for template function instantiations even if those functions are not inlined. (And also the reason why linkers have to be C++-aware; pure C linkers will not handle C++ object files correctly.)

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: Almost. `s_vnodesByPath[path]` is a reference to a weak pointer; this should either be null (no entry for this path in the map) or fail to be turned into a strong pointer because any previous vnode for that path should have hit a reference count of 0.

So we specifically don't want to check `s_vnodesByPath.find(path) == s_vnodesByPath.end()` because that would incorrectly fail where a vnode used to exist for that path but simply hit a refcount of 0, but the weak pointer is still there.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
wilbaker: Thanks! I didn't realize that was the case with templates 👍

ProjFS.Mac/PrjFSKextTests/TestMemory.cpp
wilbaker: Ok thanks!

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
wilbaker: Thanks!

Is it more accurate to say that the assert is to validate that there's not already an active vnode for the specified path?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -0,0 +1,235 @@
...
+    return result;
+}
+
+void vnode::StartRecycling()
+{
+    s_vnodesByPath.erase(this->path);
+    this->path.clear();
+    this->type = VBAD;
+    this->vid++;
+    this->isRecycling = true;
```
wilbaker: With the current code, will `isRecycling` ever go back to `false`?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -0,0 +1,265 @@
...
+    
+    VnodePointer vnode = vnode::Create(fatMount, path);
```
wilbaker: Should this `vnode` be created as a `VDIR`?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -0,0 +1,265 @@
...
+    s_virtualizationRoots[rootIndex].inUse = true;
+    s_virtualizationRoots[rootIndex].rootVNode = oldVnode.get();
+    s_virtualizationRoots[rootIndex].rootVNodeVid = oldVnode->GetVid();
+    s_virtualizationRoots[rootIndex].rootFsid = self->testMountPoint->GetFsid();
+    uint64_t inode = oldVnode->GetInode();
+    s_virtualizationRoots[rootIndex].rootInode = inode;
+    
+    oldVnode->StartRecycling();
+    
+    VnodePointer newVnode = vnode::Create(self->testMountPoint, path, VDIR, oldVnode->GetInode());
```
wilbaker: How does the recycling of `oldVnode` have an impact on `newVnode`?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: Yes definitely, thanks!

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: Currently no, but we'll probably want that feature for other tests, e.g. for verifying that our code correctly handles the vid.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: It doesn't really, but we definitely don't want to have 2 vnodes for the same path, so the old one must be recycled first.

ProjFS.Mac/PrjFS.xcodeproj/xcshareddata/xcschemes/Build All.xcscheme
pmj: Apologies, all that seemed to be necessary my end was to actually run the tests, but perhaps that was down to some previous Xcode state.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: The constructor for `vnode` is private, so [not without a workaround](https://stackoverflow.com/questions/8147027/how-do-i-call-stdmake-shared-on-a-class-with-only-protected-or-private-const). We can of course make the constructor public, but the reason it's private is because the `s_allVnodes` check won't work if someone accidentally creates an instance without going via `::Create()`.

wilbaker: Missed that detail, thanks!

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -0,0 +1,75 @@
...
+private:
+    VnodeWeakPointer weakSelfPointer;
+    MountPointer mountPoint;
+    
+public:
+    bool isRecycling = false;
+    vtype type = VREG;
+
+private:
+    uint64_t inode;
```
jeschu1: For my own understanding can you explain how we layout the types here.  I see we go private/public/private/public.  I want to ensure I understand the standard we have.

pmj:
> I think this is a phenomenal start!

Thanks!

> I'm sure as we get more test coverage we'll make changes over time.
> 
> Please also wait for @wilbaker's approval before merging

I've been going through and incorporating feedback; I'm planning to complete that later this evening and push a new version of the PR so feel free to await that before approving.

pmj:
> I'm planning to complete that later this evening

I'm not quite done with this just yet, I'll try and get this done ASAP though.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
pmj: I've been grouping the member variables logically rather than by access permissions. The 2 remaining public ones are probably better off private and accessed via functions, though.

pmj:
I've just pushed an overhauled version of this PR without making any significant changes to the overall functionality to avoid overloading this PR. I'm hoping we can merge this soonish and make major improvements in future PRs. Please re-review at least briefly! Thanks.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
pmj: Thanks, I've incorporated those naming suggestions now.

ProjFS.Mac/PrjFSKext/PrjFSService.cpp
```diff
@@ -1,7 +1,7 @@
 #include "public/PrjFSCommon.h"
 #include "public/PrjFSProviderClientShared.h"
 #include "PrjFSService.hpp"
-#include "PrjFSProviderUserClient.hpp"
+#include "PrjFSProviderUserClientPrivate.hpp"
```
wilbaker: No need to change this now (and I don't like adding yet another naming scheme) but it looks a bit weird to include the "Private.cpp" of `PrjFSProviderUserClient` in `PrjFSService.cpp`.

Elsewhere the "Private.hpp" naming scheme has meant that the contents of the .hpp should only be known to the .cpp with the same name.

Perhaps something like "PrjFSProviderUserClientKernel.hpp"?

cc: @jeschu1

PR #840, opened 2019-02-25
[releases/shipped] Use a new log directory for upgrade logs to avoid ACL issues

### LINKED ISSUE #837: GVFS.Service's renaming of GVFS.Upgrade\Logs fails when the upgrader process has an open handle to a log file inside of GVFS.Update\Logs

wilbaker:
Fixes #837 

This address the issue in #837 by switching to a new directory for upgrader logs.  GVFS.Service will create the directory if it does not exist (with the proper ACLs) and prevent the directory from inheriting ACLs (when GVFS.Service updates them on C:\ProgramData\GVFS).

Additionally, `gvfs upgrade` will use the correct ACLs when creating the logs directory if for some reason the service has not already created it.

Note: There is no automatic healing if someone manually adjusts the ACLs on the directory.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
@@ -122,7 +122,35 @@ namespace GVFS.Platform.Windows
             return false;
         }
 
-        public bool TryCreateDirectoryWithAdminOnlyModify(ITracer tracer, string directoryPath, out string error)
+        public bool TryCreateDirectoryWithAdminAndUserModifyPermissions(string directoryPath, out string error)
+        {
+            try
+            {
+                DirectorySecurity directorySecurity = new DirectorySecurity();
+
+                // Protect the access rules from inheritance and remove any inherited rules
+                directorySecurity.SetAccessRuleProtection(isProtected: true, preserveInheritance: false);
+
+                // Add new ACLs for users and admins.  Users will be granted write permissions.
+                AddUsersAccessRulesToDirectorySecurity(directorySecurity, grantUsersModifyPermissions: true);
+                AddAdminAccessRulesToDirectorySecurity(directorySecurity);
+
+                Directory.CreateDirectory(directoryPath, directorySecurity);
+            }
+            catch (Exception e) when (e is IOException ||
+                                      e is UnauthorizedAccessException ||
+                                      e is PathTooLongException ||
+                                      e is DirectoryNotFoundException)
+            {
+                error = $"Exception while creating directory `{directoryPath}`: {e.Message}";
+                return false;
+            }
+
+            error = null;
+            return true;
+        }
+
+        public bool TryCreateOrUpdateDirectoryToAdminModifyPermissions(ITracer tracer, string directoryPath, out string error)
         {
             try
             {
@@ -153,7 +181,7 @@ namespace GVFS.Platform.Windows
             {
                 EventMetadata metadata = new EventMetadata();
                 metadata.Add("Exception", e.ToString());
-                tracer.RelatedError(metadata, $"{nameof(this.TryCreateDirectoryWithAdminOnlyModify)}: Exception while creating/configuring directory");
+                tracer.RelatedError(metadata, $"{nameof(this.TryCreateOrUpdateDirectoryToAdminModifyPermissions)}: Exception while creating/configuring directory");
 
                 error = e.Message;
                 return false;
--- a/GVFS/GVFS.Service/GvfsService.cs
+++ b/GVFS/GVFS.Service/GvfsService.cs
@@ -384,86 +384,18 @@ namespace GVFS.Service
         private void CreateAndConfigureUpgradeLogDirectory()
         {
             string upgradeLogsPath = ProductUpgraderInfo.GetLogDirectoryPath();
-            DirectorySecurity upgradeLogsSecurity = this.GetUpgradeLogsDirectorySecurity(upgradeLogsPath);
-            Directory.CreateDirectory(upgradeLogsPath, upgradeLogsSecurity);
-            try
-            {
-                // Call SetAccessControl in case the directory already existed
-                // (in which case the above CreateDirectory was a no-op)
-                Directory.SetAccessControl(upgradeLogsPath, upgradeLogsSecurity);
-            }
-            catch (UnauthorizedAccessException e)
-            {
-                // UnauthorizedAccessException can occur when the upgrade logs directory was
-                // created by a non-elevated user running 'gvfs upgrade'.  Only the owner
-                // is allowed to modify the ACLs, and if the logs directory was created by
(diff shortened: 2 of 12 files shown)


wilbaker:
It looks like there is still a problem in this scenario:

1. Upgrade to a version with these changes
2. GVFS.Service never starts
3. User runs `gvfs upgrade` from an elevated CMD prompt
4. The new logs directory will not have the correct ACLs, and it will prevent non-elevated users from being able to create upgrade log files.

GVFS/GVFS.Service/GvfsService.cs
```diff
@@ -386,85 +386,6 @@ private void CreateAndConfigureUpgradeLogDirectory()
             string upgradeLogsPath = ProductUpgraderInfo.GetLogDirectoryPath();
             DirectorySecurity upgradeLogsSecurity = this.GetUpgradeLogsDirectorySecurity(upgradeLogsPath);
             Directory.CreateDirectory(upgradeLogsPath, upgradeLogsSecurity);
```
wilbaker: It looks like there is still a problem in this scenario:

1. Upgrade to a version with these changes
2. GVFS.Service never starts
3. User runs `gvfs upgrade` from an elevated CMD prompt
4. The new logs directory will not have the correct ACLs, and it will prevent non-elevated users from being able to create upgrade log files.

wilbaker: Addressed this by making sure the logs directory is created with the correct ACLs when running from an elevated CMD prompt

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -132,12 +133,20 @@ protected override void Execute(GVFSEnlistment enlistment)
                         if (GVFSPlatform.Instance.UnderConstruction.SupportsGVFSUpgrade)
                         {
                             // upgrader
+                            this.CopyAllFiles(
```
jeschu1: So now we'll have:
UpgradeDirectoryName/Logs
UpgradeDirectoryName/UpgradeLogs

We should only look at UpgradeLogs

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -94,6 +94,15 @@ private bool TryInitializeUpgrader(out string error)
                 error = null;
                 if (this.upgrader == null)
                 {
+                    string createDirectoryError;
+                    if (!this.fileSystem.TryCreateDirectoryWithAdminAndUserModifyPermissions(ProductUpgraderInfo.GetLogDirectoryPath(), out createDirectoryError))
```
jeschu1: @alameenshah this may be related to the issue you wanted to look at.  Should we exit and write error if we fail due to permissions?

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
wilbaker: In the future we'll probably only look at `UpgradeDirectoryName/UpgraderLogs`.  

If an upgrade from the last GA release hit an issue, that would be in `UpgradeDirectoryName/Logs`.

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
```diff
@@ -122,7 +122,35 @@ public bool IsSocket(string fileName)
...
+        {
+            try
+            {
+                DirectorySecurity directorySecurity = new DirectorySecurity();
+
+                // Protect the access rules from inheritance and remove any inherited rules
+                directorySecurity.SetAccessRuleProtection(isProtected: true, preserveInheritance: false);
+
+                // Add new ACLs for users and admins.  Users will be granted write permissions.
+                WindowsFileSystem.AddUsersAccessRulesToDirectorySecurity(directorySecurity, grantUsersModifyPermissions: true);
```
jamill: nit: below, we use the unqualified access to `AddUsersAccessRulesToDirectorySecurity` and friends

GVFS/GVFS/CommandLine/UpgradeVerb.cs
jamill: Can this be run from a non-admin command prompt?

wilbaker: > Can this be run from a non-admin command prompt?

Under normal circumstances yes.  

GVFS.Service will create the `UpgraderLogs` directory with the correct ACLs when it starts up.

However, if something during an install\upgrade went wrong the the service failed to start the `UpgraderLogs` directory would not be created by the service and it's possible that a non-elevated user would be unable to create the directory.

GVFS/GVFS.Platform.Windows/WindowsFileSystem.cs
wilbaker: Nice catch!

wilbaker:
> I think the PR has some good discussion about why we are making this change - could we bring some of it into the commit message(s)? I think it would be helpful for people looking through the commit history, where they might not have as ready access to the PR discussion.

Sure, I can add more details there.

GVFS/GVFS/CommandLine/UpgradeVerb.cs
wilbaker: I will add a comment to the code with these details.

wilbaker:
Next push will squash commits and provide more descriptive commit message

GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -94,6 +94,19 @@ private bool TryInitializeUpgrader(out string error)
                 error = null;
                 if (this.upgrader == null)
                 {
+                    // Under normal circumstances ProductUpgraderInfo.GetLogDirectoryPath will have already been created by GVFS.Service.  If for some reason it
+                    // does not (e.g. the service failed to start), we need to create ProductUpgraderInfo.GetLogDirectoryPath() explicity to ensure that
+                    // it has the correct ACLs (so that both admin and non-admin users can create log files).
+                    // If the logs directory does not already exist, this call could fail when running an a non-elevated user.
```
kewillford: >running an a non-elevated user.

running as a non-elevated user.

PR #842, opened 2019-02-25
[releases/shipped] Enhance git command logging

jeschu1:
- Limit error message output to 1000 chars
- Include name of GitCommand
- Remove Output logging from commands.  This can be looked up in Trace2.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -22,7 +22,7 @@ namespace GVFS.Common.Maintenance
         protected virtual TimeSpan TimeBetweenRuns { get; }
         protected virtual string LastRunTimeFilePath { get; set; }
         protected GVFSContext Context { get; }
-        protected GitProcess GitProcess { get; private set; }
+        protected GitProcess MaintenanceGitProcess { get; private set; }
         protected bool Stopping { get; private set; }
         protected bool RequireObjectCacheLock { get; }
         protected GitProcessChecker GitProcessChecker { get; }
@@ -75,7 +75,7 @@ namespace GVFS.Common.Maintenance
             {
                 this.Stopping = true;
 
-                GitProcess process = this.GitProcess;
+                GitProcess process = this.MaintenanceGitProcess;
 
                 if (process != null)
                 {
@@ -114,13 +114,16 @@ namespace GVFS.Common.Maintenance
         /// </summary>
         protected abstract void PerformMaintenance();
 
-        protected GitProcess.Result RunGitCommand(Func<GitProcess, GitProcess.Result> work)
+        protected GitProcess.Result RunGitCommand(Func<GitProcess, GitProcess.Result> work, string gitCommand)
         {
-            using (ITracer activity = this.Context.Tracer.StartActivity("RunGitCommand", EventLevel.Informational, Keywords.Telemetry, metadata: this.CreateEventMetadata()))
+            EventMetadata metadata = this.CreateEventMetadata();
+            metadata.Add("gitCommand", gitCommand);
+
+            using (ITracer activity = this.Context.Tracer.StartActivity("RunGitCommand", EventLevel.Informational, Keywords.Telemetry, metadata))
             {
                 if (this.Stopping)
                 {
-                    string message = this.Area + "Not launching Git process because the mount is stopping";
+                    string message = $"{this.Area}: Not launching Git process {gitCommand} because the mount is stopping";
                     this.Context.Tracer.RelatedWarning(
                         metadata: null,
                         message: message,
@@ -128,13 +131,20 @@ namespace GVFS.Common.Maintenance
                     return new GitProcess.Result(message, string.Empty, GitProcess.Result.ExitDueToShutDownCode);
                 }
 
-                GitProcess.Result result = work.Invoke(this.GitProcess);
+                GitProcess.Result result = work.Invoke(this.MaintenanceGitProcess);
 
                 if (!this.Stopping && result?.ExitCodeIsFailure == true)
                 {
+                    string errorMessage = result?.Errors == null ? string.Empty : result.Errors;
+                    if (errorMessage.Length > 1000)
+                    {
+                        // For large error messages, we show the first and last 500 chars
+                        errorMessage = $"beginning: {errorMessage.Substring(0, 500)} ending: {errorMessage.Substring(errorMessage.Length - 500)}";
+                    }
+
                     this.Context.Tracer.RelatedWarning(
                         metadata: null,
-                        message: this.Area + ": Git process failed with errors:" + result.Errors,
+                         message: $"{this.Area}: Git process {gitCommand} failed with errors: {errorMessage}",
                         keywords: Keywords.Telemetry);
                     return result;
                 }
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -174,7 +176,6 @@ private bool TryGetMaxGoodPrefetchTimestamp(out long maxGoodTimestamp, out strin
                     {
                         firstBadPack = i;
 
-                        metadata.Add("Errors", indexResult.Errors);
```
wilbaker: Why was this line removed?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -132,9 +135,16 @@ protected GitProcess.Result RunGitCommand(Func<GitProcess, GitProcess.Result> wo
 
                 if (!this.Stopping && result?.ExitCodeIsFailure == true)
                 {
+                    string errorMessage = result.Errors;
```
wilbaker: Is `result.Errors` guaranteed to be non-null?

GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
```diff
@@ -26,12 +26,12 @@ protected override void PerformMaintenance()
                 string multiPackIndexLockPath = Path.Combine(this.Context.Enlistment.GitPackRoot, MultiPackIndexLock);
                 this.Context.FileSystem.TryDeleteFile(multiPackIndexLockPath);
 
-                this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot));
+                this.RunGitCommand((process) => process.WriteMultiPackIndex(this.Context.Enlistment.GitObjectsRoot), "WriteMultiPackIndex");
```
wilbaker: Minor: For these string literals you could switch to `nameof(<method>)` to make future refactoring a bit easier.

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
jeschu1: To keep things simple I've updated all processes to rely on RunGitCommand to print errors.  In there we truncate based on size.

For this specific error, it won't be large since we make it ourselves.  However it would be a duplicate error in telemetry and I'd like the behavior to be consistent.  A more intrusive change that could be considered (for the future) would be to only return an exitCode from RunGitCommand.

wilbaker: Ok thanks!

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: Ha, no, nice catch.

We creatively call this which will result in null:

```
            this.RunGitCommand(
                process =>
                {
                    this.TryPrefetchCommitsAndTrees(out error, process);
                    return null;
                },
                "TryPrefetchCommitsAndTrees");
```

wilbaker: > In that case you'll never get into the if.

I noticed that too.  Unfortunately the GitProcess.Result class itself didn't have any hints about whether Errors would always be non-null and I agree it's better to be safe.

GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
jeschu1: Done.  I couldn't nameof(GitProcess.WriteMultiPackIndex) out of the box since we used the same name for a local var and the class.  I renamed the local var for ease.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -118,7 +118,8 @@ public int CreateLooseObjectsPackFile()
                 (process) => process.PackObjects(
                     "from-loose",
                     this.Context.Enlistment.GitObjectsRoot,
-                    (StreamWriter writer) => objectsAddedToPack = this.WriteLooseObjectIds(writer)));
+                    (StreamWriter writer) => objectsAddedToPack = this.WriteLooseObjectIds(writer)),
+                "PackObjects");
```
wilbaker: "PackObjects" -> nameof(GitProcess.PackObjects)

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -138,23 +138,23 @@ protected override void PerformMaintenance()
                     return;
                 }
 
-                GitProcess.Result expireResult = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot));
+                GitProcess.Result expireResult = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot), "MultiPackIndexExpire");
```
wilbaker: "MultiPackIndexExpire" -> nameof(GitProcess.MultiPackIndexExpire)

GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -89,11 +89,13 @@ protected override void PerformMaintenance()
...
-                this.TryPrefetchCommitsAndTrees(out error, process);
-                return null;
-            });
+            this.RunGitCommand(
+                process =>
+                {
+                    this.TryPrefetchCommitsAndTrees(out error, process);
+                    return null;
+                },
+                "TryPrefetchCommitsAndTrees");
```
wilbaker: "TryPrefetchCommitsAndTrees" -> nameof(this.TryPrefetchCommitsAndTrees)

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -114,27 +114,37 @@ public void Stop()
...
+                GitProcess.Result result = work.Invoke(this.MaintenanceGitProcess);
 
                 if (!this.Stopping && result?.ExitCodeIsFailure == true)
                 {
+                    string errorMessage = result?.Errors == null ? string.Empty : result.Errors;
+                    if (errorMessage.Length > 1000)
+                    {
+                        // For large error messages, we show the first and last 500 chars
+                        errorMessage = $"beginning: {errorMessage.Substring(0, 500)} ending: {errorMessage.Substring(errorMessage.Length - 500)}";
+                    }
```
kewillford: Should this be moved to the tracer or do we only expect large error messages from git?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -189,29 +199,25 @@ protected void SaveLastRunTimeToFile()
...
 
             activity.RelatedError(errorMetadata, "multi-pack-index is corrupt after write. Deleting and rewriting.");
         }
 
-        protected void LogErrorAndRewriteCommitGraph(ITracer activity, GitProcess.Result result, List<string> packs)
+        protected void LogErrorAndRewriteCommitGraph(ITracer activity, List<string> packs)
         {
             EventMetadata errorMetadata = this.CreateEventMetadata();
-            errorMetadata["CommitGraphVerifyOutput"] = result.Output;
-            errorMetadata["CommitGraphVerifyErrors"] = result.Errors;
```
kewillford: We are no longer capturing the `Output`?  Are there cases where we want that in the tracing?

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: We only expect the large error messages from git.

If we think we could have an epidemic of large messages we could limit message size coming from the tracer directly the same way. I would punt that to the future though, since that affects all messages and will require more thought.

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jeschu1: For consistency we're only printing errrors for maitence jobs.  We should be able to look it up any error/output from git using Trace2 if necessary.  So yes not needed.

jamill:
You might want to include the fact that we are no longer logging the output of the Git Command in the list of things this PR changes (in addition to the other 2 bullet points you already have)

GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
jamill: Would we want the errors to be in the logs written out to file? We might not always have access to the trace2 data, and I could see value in being able to look up the error in the local log file. Do we trace enough other data to some information about the error in the local log file?

jeschu1: @jamill we don't log the output for other git commands, and I really don't think it's a necessity.   The errors will be in the log.   If we need to look up more info we can in Trace2.  

I've done a lot of error looksups myself already and I find Trace2 to have vastly more info and be super useful.  

Of course we can change our mind on this in the future.  But so far with my investigations this is my recommended way to go.

jamill: >We will do it in a central spot now in RunGitProcess so we can limit message size.

Ah! I see that now.

Issue #818, opened 2019-02-21
[Linux] Rebase of branch


kivikakk:
We've done a rebase of the `features/linuxprototype` branch on top of VFSForGit's current `master` -- feel free to reset your branch pointer to https://github.com/github/VFSForGit/commit/ebec5f12d668025463f08e93fee47b411d6e0c8a if you'd like to take advantage of this!

chrisd8088:
Awesome, thank you @kivikakk, as always!  ✨ ⭐️

jrbriggs:
I've reset `features/linuxprototype` to ebec5f1. Thanks @kivikakk

kivikakk:
Thank you! :heart:

PR #800, opened 2019-02-18
Linux: use stat to get filemode

kivikakk:
Create placeholders with the correct permissions. It's "just" the mirror provider, but it was bugging me enough that I wanted to fix it.

/cc @chrisd8088 fyi

PR #848, opened 2019-02-26
[Linux] call projfs directly, don't use intermediary projfs_vfsapi

kivikakk:
👋  We've been developing `libprojfs` as a library which offers its own notifications and callbacks, and then built a VFS API on top of that which connects `libprojfs` to a .NET provider. The VFS API exports the `PrjFS_*` symbols used by `PrjFSLib.Linux.Managed`.

Right now, the VFS API lives in the same tree as `libprojfs` -- see [`projfs_vfsapi.h`](https://github.com/github/libprojfs/blob/cf89b2043b5da08871e77c28f4f7589045d65005/include/projfs_vfsapi.h) as of time of writing to see what it exposes.

We want to reduce the coupling of `libprojfs` to VFS for Git. I had first thought to simply pull the VFS API code into this repo and have a separate build step (with two `.so`s), but that's quite a lot of overhead for a thin wrapper. See https://github.com/github/libprojfs/issues/48 where I spike that out and @chrisd8088 lays out the land.

We resolved to try option #​3. This PR retargets `PrjFSLib.Linux.Managed` -- it now consumes `libprojfs` directly. This removes a layer of callbacks and forwarding, simplifying the mental model significantly.

There's probably a whole bunch of style issues to be addressed here. (At the very least it's StyleCop clean.) Likewise, `ProjFS.Linux/Scripts/GenerateConstants.sh` is pretty hacky, but does only need to be run when `projfs_notify.h` changes.

There's a companion libprojfs PR at https://github.com/github/libprojfs/pull/53 which exposes a required entrypoint.


kivikakk:
This PR has been rebased (as of the reset in #818).

chrisd8088:
Hi @jrbriggs -- would you be able to take a look at this PR in the next few days?  
Getting this into features/linuxprototype would help us move along a bit.

One thing I suspect you'll flag is the use of some Mono classes.  If you would prefer to avoid any dependency on Mono, maybe we can look into using the `NativeMethods` technique I found in [MacFileBasedLock.cs](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs#L99).

Let me know what you think and if you see any other issues in this PR.  Thanks very much, and have a great weekend!

jrbriggs:
@chrisd8088 I would prefer to avoid the mono dependency in the mac codebase, and since we're trying to fold linux and mac together in the POSIX layer, can we try pinvoking?

Thanks!

/cc @kivikakk

chrisd8088:
Hi @jrbriggs and @wilbaker!

I'm hoping you can take a look at this PR again and whether we could get it merged into `features/linuxprototype`, as that would help us simplify the number of outstanding interdependent branches we have on the go.  :-)

While I have removed the use of Mono in `ProjFS.Linux` itself, and also tuned up the original `ProjFS.Linux/Scripts/GenerateConstants.sh` in conjunction with that, you may notice that there's still a dependency on Mono in `MirrorProvider` -- but that [already exists](https://github.com/Microsoft/VFSForGit/blob/features/linuxprototype/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs#L108) prior to this PR in the `features/linuxprototype` branch, so we're not adding Mono anywhere in this PR now beyond what's already in place (i.e., in the Linux `MirrorProvider`).

---
Now I did also manage to fully remove Mono from the `MirrorProvider.Linux`; see my [`vfs-api-nomono-wip`](https://github.com/github/VFSForGit/compare/vfs-api...github:vfs-api-nomono-wip) branch for an initial version of that work.  However, this comes with significant caveats, so I don't want it to block this PR.  It was worth thinking about, though, because it relates to the [issues with `stat(2)` calls](https://github.com/Microsoft/VFSForGit/pull/800#discussion_r257842423) which @kivikakk had, and will absolutely crop up again as we look at migrating the GVFS code proper, particularly the use of a call to `stat(2)` in [`GVFS.Platform.Mac/MacFileSystem.cs`](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Mac/MacFileSystem.cs#L115), and also pertains to @jrbriggs's [question about native calls](https://github.com/Microsoft/VFSForGit/pull/800#discussion_r257746146).

#### A. Issues with the `stat(2)` family
The immediate challenge (and again, this doesn't apply to this PR per se, but looking ahead a bit) is that GNU libc [doesn't provide](https://sourceware.org/glibc/wiki/FAQ#Looking_through_the_shared_libc_file_I_haven.27t_found_the_functions_.60stat.27.2C_.60lstat.27.2C_.60fstat.27.2C_and_.60mknod.27_and_while_linking_on_my_Linux_system_I_get_error_messages.__How_is_this_supposed_to_work.3F) the `stat(2)` family of calls in its dynamic `libc.so`, because it's working around the many historical and architecture-dependent variations on the layout of fields in `struct stat`.  Instead, there's a small static `libc_nonshared.a` library with those calls, which are [wrappers](https://sourceware.org/git/?p=glibc.git;a=blob;f=io/lstat64.c;#l37) around the [`__xstat()` functions](http://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic.html#BASELIB---XSTAT64) in the dynamic library, and unfortunately the .NET runtime can't use that static library, so it can't P/Invoke things like `lstat(2)` directly.

Because the `__xstat()` workarounds are part of the Linux Standard Base spec, at least some alternate libc implementations like [musl include it](https://git.musl-libc.org/cgit/musl/tree/src/stat/__xstat.c), but I don't think that's true for uClibc-ng, for instance.  Even if we assume the use of GNU libc or an ABI-compatible libc, we face the issue of variant `struct stat` layouts; my experimental [WIP branch uses](https://github.com/github/VFSForGit/compare/vfs-api...github:vfs-api-nomono-wip#diff-81f4d84eb19b3a88b61b0f2ded1526b5R285) the [one for x86-64](http://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-AMD64/LSB-Core-AMD64.html#AEN5763) but there are [others](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86/bits/stat.h#l117) depending on the system architecture.

The Mono system gets around this by not trying to P/Invoke directly into libc; instead it builds a 
`libMonoPosixHelper.so` library which [wraps](https://github.com/mono/mono/blob/master/support/sys-stat.c#L149) the various `stat(2)` syscalls and [maps their `stat` structures](https://github.com/mono/mono/blob/master/support/sys-stat.c#L75) to a standard C# structure.  I strongly suspect we will have to do something similar for Linux (if we're not using Mono itself), while you can probably stick with a [hard-coded mapping](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Mac/MacFileSystem.cs#L125) on macOS where your system architecture is more nailed down.

#### B. Handling Unix filenames and UTF-8
The other thing Mono provides is a more comprehensive solution to the business of handling Unix filenames via its [`UnixEncoding`](http://docs.go-mono.com/monodoc.ashx?link=T%3AMono.Unix.UnixEncoding) and custom [`FileNameMarshaler`](https://github.com/mono/mono/blob/master/mcs/class/Mono.Posix/Mono.Unix.Native/FileNameMarshaler.cs), which is used to translate C# `string`s when passing filenames to something like [`open(2)`](https://github.com/mono/mono/blob/master/mcs/class/Mono.Posix/Mono.Unix.Native/Syscall.cs#L3028).  This is in contrast to the way it's done in [`GVFS.Platform.Mac/MacFileSystem.cs`](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Mac/MacFileSystem.cs#L180) which relies, I believe, on the default [`CharSet.Ansi`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.charset?view=netframework-4.7.2#remarks) transform done by .NET.

That may work on macOS (although I'm [not totally sure](https://blog.kutulu.org/2012/04/marshaling-utf-8-harder-than-it-ought.html) about that), but on Linux I suspect, even if we make the assumption that filenames are almost always encoded in UTF-8 (not an absolute given, alas), doing UTF-8 conversions inline (as I'm [doing now](https://github.com/github/VFSForGit/compare/vfs-api...github:vfs-api-nomono-wip#diff-81f4d84eb19b3a88b61b0f2ded1526b5R107) in my [`vfs-api-nomono-wip`](https://github.com/github/VFSForGit/compare/vfs-api...github:vfs-api-nomono-wip) test branch with a small [`GetBytesUTF8() helper`]( https://github.com/github/VFSForGit/compare/vfs-api...github:vfs-api-nomono-wip#diff-81f4d84eb19b3a88b61b0f2ded1526b5R268)) will get ugly fast.

So I think we may want to borrow some of the Mono code from [`UnixMarshal`](https://github.com/mono/mono/blob/master/mcs/class/Mono.Posix/Mono.Unix/UnixMarshal.cs#L162)  and [`UnixEncoding`](https://github.com/mono/mono/blob/master/mcs/class/Mono.Posix/Mono.Unix/UnixEncoding.cs#L160), as called from `FileNameMarshaler`, at least enough to -- as they say in the [`UnixEncoding` FAQ](http://docs.go-mono.com/monodoc.ashx?link=T%3AMono.Unix.UnixEncoding) -- be able to simply pass and receive `string`s with file paths to our P/Invoke syscall wrappers and let the custom marshaler figure out the UTF-8 (or other) conversion from `string` to `char*` and back.  Then our `DllImport`s could be [similar to theirs](https://github.com/mono/mono/blob/master/mcs/class/Mono.Posix/Mono.Unix.Native/Syscall.cs#L3028) for syscalls like `open(2)` and `readlink(2)`, etc., and our invocations of them would be clean too.  (Note that `MPH` in their code refers to their `libMonoPosixHelper.so` library, which as I said above, wraps things like the arch-and-libc-specific `stat(2)` calls.)

I believe Mono is licensed under the MIT license, so hopefully we could just pull some of that code into VFSForGit if you're keen not to have an outright dependency on a pre-built Mono package.

---
Whew.  OK, so, this PR doesn't actually get deeply involved in any of those musings; that's all for the (near) future.  What this PR does do is implement @kivikakk's work to call our "native" libprojfs API, and avoid using Mono in the `ProjFS.Linux` code itself, although it retains such a dependency in `MirrorProvider.Linux` for at least the time being.  Please do take a look over it and let me know if anything else needs a fixup before it could be merged in `features/linuxprototype`.  And thanks very much for all the help!

/cc @kivikakk, @wrighty

chrisd8088:
Oh, and I should add, @jrbriggs, that the CI test failure is (I'm virtually certain) unrelated to this PR.  I've bumped into this transient CI failure last week, and it went away when the failing test ran a second time.  So I think it must be a bug in the CI tests which was introduced recently.

The specific error message is:
```
Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.PrefetchVerbTests.PrefetchCleansUpPackDir()
  C:\Repos\GVFSFunctionalTests\enlistment\..\.gvfsCache is expected to have only one folder. Actual folders: C:\Repos\GVFSFunctionalTests\.gvfsCache\5aa3c69dc0ac42d39714896f33f37ff0,C:\Repos\GVFSFunctionalTests\.gvfsCache\97dc78053d8f4e79aeb15ed21a4a08db
```
and at a wild guess I'm wondering if it has something to do with the changes from #941?  Anyway, I'm pretty certain it's nothing to do with the Linux work.  :-)

jrbriggs:
@chrisd8088 This is merged, if you want to rebase up. Thanks!

chrisd8088:
Thank you!  And I'll rebase soon; that would be great as it sounds like it'll fix the CI issue.

Much appreciated!

PR #850, opened 2019-02-26
Upgrader cleanup tweaks

### LINKED ISSUE #833: Downloads directory is not deleted after running `gvfs upgrade` and subsequent non-elevated `gvfs upgrade` commands report an error
1. Upgrade VFS4G via `gvfs upgrade --confirm`
2. Open a non-elevated CMD prompt and run `gvfs upgrade`
3. The follow error is logged and reported via telemetry:

>[2019-02-22 13:38:23 -08:00] Error {"ErrorMessage":"DeleteAllInstallerDownloads: Could not remove directory: C:\\ProgramData\\GVFS\\GVFS.Upgrade\\Downloads.System.ArgumentException: Access to the path is denied.\r\n   at System.IO.FileSystemInfo.set_Attributes(FileAttributes value)\r\n   at GVFS.Common.FileSystem.PhysicalFileSystem.RecursiveDelete(String path)\r\n   at GVFS.Common.ProductUpgraderInfo.DeleteAllInstallerDownloads()"}

Two issues here:

- The downloads directory should have been deleted after the upgrade
- `gvfs upgrade` should not report an error if `DeleteAllInstallerDownloads` fails when running non-elevated (in fact it should probably not try to delete the directory at all).

jamill:
Streamline when upgrade cleans up old downloads

Previously, the upgrade notification logic checked for previous downloads of the upgrade packages in order to drive whether to notify the user about whether an upgrade is available. There were cases where the user updated outside of the upgrader logic, where the downloaded upgrade packages where not cleaned up, resulting in spurious upgrade notifications.

Since then, we have changed the logic to look at a sentinel file to indicate whether an upgrade is available, and stopped downloading packages in the background.

The NuGet upgrader did not clean up the downloaded packages, only the extracted packages.

This change:
1) UpgradeVerb and UpgradeOrchestrator no longer opportunistically deletes the download folder containing upgrade packages
2) UpgradVerb deletes the upgrade available sentinel file if run elevated
3) Upgrader deletes the download folder after running

### Questions:
Q1) Do we want to always delete the downloaded packages after upgrade runs (if upgrade fails for some reason, it might be useful to have the package around for diagnostic purposes)

A) We can look at this as a future improvement

Q2) Do we need to be more proactive when running `GVFS upgrade` to remove the sentinel file if not update is available? One issue is that this file can only be modified by a process with elevated permissions.

A) We will delete the sentinel file if run as an admin. We can look at improving this more in the future based on feedback.

Fixes #833

### DIFF
--- a/GVFS/GVFS.Common/GitHubUpgrader.cs
+++ b/GVFS/GVFS.Common/GitHubUpgrader.cs
@@ -226,38 +226,6 @@ namespace GVFS.Common
             return true;
         }
 
-        public override bool TryCleanup(out string error)
-        {
-            error = string.Empty;
-            if (this.newestRelease == null)
-            {
-                return true;
-            }
-
-            foreach (Asset asset in this.newestRelease.Assets)
-            {
-                Exception exception;
-                if (!this.TryDeleteDownloadedAsset(asset, out exception))
-                {
-                    error += $"Could not delete {asset.LocalPath}. {exception.ToString()}." + Environment.NewLine;
-                }
-            }
-
-            if (!string.IsNullOrEmpty(error))
-            {
-                error.TrimEnd(Environment.NewLine.ToCharArray());
-                return false;
-            }
-
-            error = null;
-            return true;
-        }
-
-        protected virtual bool TryDeleteDownloadedAsset(Asset asset, out Exception exception)
-        {
-            return this.fileSystem.TryDeleteFile(asset.LocalPath, out exception);
-        }
-
         protected virtual bool TryDownloadAsset(Asset asset, out string errorMessage)
         {
             errorMessage = null;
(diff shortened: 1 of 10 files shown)


alameenshah:
Wouldn't this re-open #502. How do we mitigate this.

jamill:
> Wouldn't this re-open #502. How do we mitigate this.

I think the opportunity for this to happen is smaller now. When restarting the GVFS.Service, we will refresh the sentinel file (as opposed to before, where restarting the service would not refresh the upgrade notification). I think the cases of someone switching rings in between installs will be less, but that is a hole.

Currently, running as a non-admin will also have issues, as the non-elevated command will not be able to clear the sentinel file. Running from an elevated prompt will clear the sentinel file (with this change). One mitigation could be to have the user run as an admin (maybe we could include this in the instructions if we detect that the sentinel file is present, but no upgrade is available and the command is run non-elevated)

derrickstolee:
@jamill do you still need this PR?

jamill:
> @jamill do you still need this PR?

This is a bit outdated, I will close this for now.

PR #855, opened 2019-02-26
[upgrade] Background upgrade check fix

alameenshah:
Service was not checking for upgrades in the background. `UpgradeTimer.Start()` was incorrectly placed after `service.WaitOne()`. Meaning the timer was getting scheduled just when the Service is about to stop. Fixed and manually tested and verified that timer is getting scheduled correctly on Service launch. Added FT to verify Upgrader get scheduled on Service start.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
@@ -67,9 +67,19 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         public void UpgradeTimerScheduledOnServiceStart()
         {
             this.RestartService();
-            Thread.Sleep(TimeSpan.FromSeconds(15));
 
-            this.ServiceLogContainsUpgradeMessaging().ShouldBeTrue();
+            bool timerScheduled = false;
+            for (int trialCount = 0; trialCount < 15; trialCount++)
+            {
+                Thread.Sleep(TimeSpan.FromSeconds(1));
+                if (this.ServiceLogContainsUpgradeMessaging())
+                {
+                    timerScheduled = true;
+                    break;
+                }
+            }
+
+            timerScheduled.ShouldBeTrue();
         }
 
         private bool ServiceLogContainsUpgradeMessaging()
@@ -81,22 +91,17 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
                 GVFSServiceProcess.TestServiceName,
                 "Logs");
             DirectoryInfo logsDirectory = new DirectoryInfo(serviceLogFolder);
-            IEnumerable<FileInfo> files = logsDirectory.GetFiles()
+            FileInfo logFile = logsDirectory.GetFiles()
                 .OrderByDescending(f => f.LastWriteTime)
-                .Take(2);
+                .FirstOrDefault();
 
-            foreach (FileInfo nextFile in files)
+            if (logFile != null)
             {
-                using (StreamReader fileStream = new StreamReader(File.Open(nextFile.FullName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)))
+                using (StreamReader fileStream = new StreamReader(File.Open(logFile.FullName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)))
                 {
-                    for (int lineCount = 0; lineCount < 20; lineCount++)
+                    string nextLine = null;
+                    while ((nextLine = fileStream.ReadLine()) != null)
                     {
-                        string nextLine = fileStream.ReadLine();
-                        if (string.IsNullOrEmpty(nextLine))
-                        {
-                            break;
-                        }
-
                         if (nextLine.Contains(upgradeTimerMessage))
                         {
                             return true;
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -62,6 +63,51 @@ public void NoReminderForLeftOverDownloads()
...
+            string upgradeTimerMessage = "Checking for product upgrades. (Start)";
+            string serviceLogFolder = Path.Combine(
+                Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData),
+                "GVFS",
+                GVFSServiceProcess.TestServiceName,
+                "Logs");
+            DirectoryInfo logsDirectory = new DirectoryInfo(serviceLogFolder);
+            IEnumerable<FileInfo> files = logsDirectory.GetFiles()
+                .OrderByDescending(f => f.LastWriteTime)
+                .Take(2);
```
wilbaker: Why take 2?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -62,6 +63,51 @@ public void NoReminderForLeftOverDownloads()
             this.VerifyUpgradeVerbStopsReminder();
         }
 
+        [TestCase]
+        public void UpgradeTimerScheduledOnServiceStart()
+        {
+            this.RestartService();
+            Thread.Sleep(TimeSpan.FromSeconds(15));
```
wilbaker: Rather than always waiting 15 seconds, could we do something like check for the message every 1 second a max of 15 times?  That way we might be able to get through this test faster.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -62,6 +63,51 @@ public void NoReminderForLeftOverDownloads()
...
+            DirectoryInfo logsDirectory = new DirectoryInfo(serviceLogFolder);
+            IEnumerable<FileInfo> files = logsDirectory.GetFiles()
+                .OrderByDescending(f => f.LastWriteTime)
+                .Take(2);
+
+            foreach (FileInfo nextFile in files)
+            {
+                using (StreamReader fileStream = new StreamReader(File.Open(nextFile.FullName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)))
+                {
+                    for (int lineCount = 0; lineCount < 20; lineCount++)
```
wilbaker: Why limit the search to the first 20 lines?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
alameenshah: Yes, that is a better approach. Incorporated it.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
alameenshah: It usually appears in the first 20. But on second thought, I think this approach could make the test unnecessarily fragile. Also since we are restarting the service in the beginning of the test case, the log file itself wouldn't be very large. Reading the whole file now (only in the worst case when the upgrade check message string is missing.)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
alameenshah: that was by mistake. Corrected. Using `FirstOrDefault()` now.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -62,6 +63,56 @@ public void NoReminderForLeftOverDownloads()
...
+                    break;
+                }
+            }
+
+            timerScheduled.ShouldBeTrue();
+        }
+
+        private bool ServiceLogContainsUpgradeMessaging()
+        {
+            string upgradeTimerMessage = "Checking for product upgrades. (Start)";
```
jamill: maybe call this `expectedTimerMessage`, and include a comment about where it comes from?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -62,6 +63,56 @@ public void NoReminderForLeftOverDownloads()
...
+                GVFSServiceProcess.TestServiceName,
+                "Logs");
+            DirectoryInfo logsDirectory = new DirectoryInfo(serviceLogFolder);
+            FileInfo logFile = logsDirectory.GetFiles()
+                .OrderByDescending(f => f.LastWriteTime)
+                .FirstOrDefault();
+
+            if (logFile != null)
+            {
+                using (StreamReader fileStream = new StreamReader(File.Open(logFile.FullName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite)))
```
jamill: If this file is not expected to be very long and file access is not an issue, a higher level method may require less code (e.g. either `File.ReadAllText` or `File.ReadLines`)

alameenshah: The higher level `File.Read` apis seem to open the file with read write permissions. Which throws exception, because service is still holding on to the log file.

PR #856, opened 2019-02-27
Mac kext: Fix for vnode_put() panic bug, tighter invariant assertion & improved root search logic.

### LINKED ISSUE #797: Mac: kernel panic on gvfs unmount (vnode_t over-release)

pmj:
The first commit in this series fixes #797, the second one strengthens one of the related `assert()`s and the third further cleans up the function that contained the bug.

 1. I explained the nature & cause of the bug [in this comment](https://github.com/Microsoft/VFSForGit/issues/797#issuecomment-467618364) so I won't repeat that here. The fix is of course to use the inode of each directory level to try to match a virtualisation root as we walk up the hierarchy.
 2. In addition to the new asserts I added in PR #805, this tightens up a much older assert in a way that would have caught this bug in action. If the vnode/vid pair matches a root vnode/vid pair, their fsid/inode must also match, or something has gone badly wrong.
 3. There has so far been no proper error handling for the `vnode_get()` call in `VirtualizationRoot_FindForVnode()`. I don't think that call *can* fail in that situation, but just in case, I've now changed the logic such that the iocount remains correct. In cases where the incoming vnode is not a directory however, it actually makes more sense to start the search with the parent directory, so we can call `vnode_getparent()` in that case instead.

I am planning to write some unit tests for `VirtualizationRoot_FindForVnode()` that would have caught this bug. However, these will depend on the substantial upgrades to the unit testing infrastructure I added in PR #839, and as that PR is still some way away from being merged, I won't delay the fix until then.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -316,7 +316,10 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
         
         if (rootEntry.rootVNode == vnode && rootEntry.rootVNodeVid == vid)
         {
-            assert(fileId.fsid.val[0] == rootEntry.rootFsid.val[0]);
+            assertf(
+                FsidsAreEqual(fileId.fsid, rootEntry.rootFsid) && fileId.inode == rootEntry.rootInode,
+                "FindRootAtVnode_Locked: matching root vnode/vid but not fsid/inode? vnode %p:%u. rootEntry fsid 0x%x:%x, inode 0x%llx, searching for fsid 0x%x:%x, inode 0x%llx",
+                KextLog_Unslide(vnode), vid, rootEntry.rootFsid.val[0], rootEntry.rootFsid.val[1], rootEntry.rootInode, fileId.fsid.val[0], fileId.fsid.val[1], fileId.inode);
             return i;
         }
         else if (rootEntry.rootVNode == vnode)
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -151,7 +151,7 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
     PrjFSPerfCounter functionCounter,
     PrjFSPerfCounter innerLoopCounter,
     vnode_t _Nonnull vnode,
-    const FsidInode& vnodeFsidInode)
+    vfs_context_t context)
 {
     PerfSample findForVnodeSample(perfTracer, functionCounter);
     
@@ -167,6 +167,8 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
     while (RootHandle_None == rootHandle && NULLVP != vnode && !vnode_isvroot(vnode))
     {
         PerfSample iterationSample(perfTracer, innerLoopCounter);
+
+        FsidInode vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context);
 
         rootHandle = FindOrDetectRootAtVnode(vnode, vnodeFsidInode);
         
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -156,11 +156,20 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
     PerfSample findForVnodeSample(perfTracer, functionCounter);
     
     VirtualizationRootHandle rootHandle = RootHandle_None;
+    errno_t error = 0;
     
-    errno_t error = vnode_get(vnode);
-    if (error != 0)
+    if (vnode_isdir(vnode))
     {
-        KextLog_ErrorVnodePathAndProperties(vnode, "VirtualizationRoot_FindForVnode: vnode_get() failed (error = %d) on vnode we'd expect to be live", error);
+        error = vnode_get(vnode);
+        if (error != 0)
+        {
+            KextLog_ErrorVnodePathAndProperties(vnode, "VirtualizationRoot_FindForVnode: vnode_get() failed (error = %d) on vnode %p:%u which we'd expect to be live", error, KextLog_Unslide(vnode), vnode_vid(vnode));
+            return RootHandle_None;
+        }
+    }
+    else
+    {
+        vnode = vnode_getparent(vnode);
     }
     
     // Search up the tree until we hit a known virtualization root or THE root of the file system
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -873,7 +873,7 @@ static bool TryGetVirtualizationRoot(
         PrjFSPerfCounter_VnodeOp_FindRoot,
         PrjFSPerfCounter_VnodeOp_FindRoot_Iteration,
         vnode,
-        *vnodeFsidInode);
+        context);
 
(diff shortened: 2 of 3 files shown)


wilbaker:
> I am planning to write some unit tests for `VirtualizationRoot_FindForVnode()` that would have caught this bug. However, these will depend on the substantial upgrades to the unit testing infrastructure I added in PR #839, and as that PR is still some way away from being merged, I won't delay the fix until then.

I agree we shouldn't delay the fix, I just filed issue #862 for adding unit tests

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -314,7 +325,10 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
         
         if (rootEntry.rootVNode == vnode && rootEntry.rootVNodeVid == vid)
         {
-            assert(fileId.fsid.val[0] == rootEntry.rootFsid.val[0]);
+            assertf(
```
wilbaker: Will these asserts trigger a KP in release builds, or are they only in effect for debug builds?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -314,7 +325,10 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
...
+                FsidsAreEqual(fileId.fsid, rootEntry.rootFsid) && fileId.inode == rootEntry.rootInode,
+                "FindRootAtVnode_Locked: matching root vnode/vid but not fsid/inode? vnode %p:%u. rootEntry fsid 0x%x:%x, inode 0x%llx, searching for fsid 0x%x:%x, inode 0x%llx",
+                vnode, vid, rootEntry.rootFsid.val[0], rootEntry.rootFsid.val[1], rootEntry.rootInode, fileId.fsid.val[0], fileId.fsid.val[1], fileId.inode);
```
wilbaker: Should we `KextLog_Unslide` this `vnode`?  Or is that not required because this is an alert?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: They are currently active in release builds too, the macro controlling this is `MACH_ASSERT=1`.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Good catch! It doesn't matter for security reasons in an assert, but if we want to match the message from the assert up with kernel logs, the slide is helpful. I'll fix that up.

jeschu1:
> I agree we shouldn't delay the fix, I just filed issue #862 for adding unit tests

@wilbaker I'm fine with that but let's prioritize that for next week, unless there are objections.  It would definitely give us a sense of security if test for cases like this in the future.
@pmj nice job again tracking this down!

wilbaker:
> @wilbaker I'm fine with that but let's prioritize that for next week, unless there are objections.

I completely agree, adding the unit tests is highest priority (after merging this in).

/ cc: @pmj @jrbriggs

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: Great thanks!

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: This is now fixed.

pmj:
> I completely agree, adding the unit tests is highest priority (after merging this in).

Sounds good. Perhaps, @wilbaker you could take a high-to-mid-level look at #839 in that case please, as I'm reluctant to head further down that path with the tests for `VirtualizationRoot_FindForVnode()` unless I'm reasonably sure I'm not going to have to rework it all in a major way.

#819, the fix for the race condition related to bug #797 is ready to go too (assuming all the tests pass) and should be a fairly quick review if anyone would like to jump on that. Thanks!

wilbaker:
>Sounds good. Perhaps, @wilbaker you could take a high-to-mid-level look at #839 in that case please, as I'm reluctant to head further down that path with the tests for `VirtualizationRoot_FindForVnode()` unless I'm reasonably sure I'm not going to have to rework it all in a major way.

>#819, the fix for the race condition related to bug #797 is ready to go too (assuming all the tests pass) and should be a fairly quick review if anyone would like to jump on that. Thanks!

@pmj sure thing! I will look at #819 and #839 next.

PR #857, opened 2019-02-27
GitStatusCache: logging tweaks

### LINKED ISSUE #206: GitStatusCache logging improvements
Some things I noticed in my testing:
* When the GitStatusCache remains disabled due to a required reboot, it does not log anything into the mount log to indicate that it is disabled. In fact, all I saw was a message that the status cache config was loaded, implying that it was enabled.
* The "Generating new Status Cache" message is always accompanied by a message of "delayed for 2.0Xs". I think we should reserve that delay message for cases where we actually backed off. In these cases, there was no backoff occurring.
* The log output is currently very verbose. Every time I do some IO, I see two lines printed into the mount log ("Generating new Status Cache" and "Done generating status"). And the next call to `git status` also outputs another line that the cache is ready. I think we could trim this down quite a bit without losing any real diagnostic info.

jamill:
This includes the following logging tweaks:

1) Log when status cache is not enabled.
This makes it more explicit whether the cache is enabled or not.

```
[2019-02-27 10:30:43 -05:00] Information {"Message":"Git status cache is not enabled"}
```

2) Reduce the number of trace statements generated by the status cache
The current logging is a bit verbose, and at this point has not been used for diagnostics. This change reduces the number of lines we trace, but adds a bit more information to the final trace statement after a background status command is run.

```
[2019-02-27 10:27:51 -05:00] Information {"Message":"GitStatusCache.RebuildStatusCacheIfNeeded: Done generating status. Cache state: Clean. Status scan time: 24.69s."}
[2019-02-27 10:28:37 -05:00] Information {"Message":"GitStatusCache.RebuildStatusCacheIfNeeded: Done generating status. Cache state: Clean. Status scan time: 2.17s. Status scan was delayed for: 6.01s."}
```

Fixes #206

### DIFF
--- a/GVFS/GVFS.Common/GitStatusCache.cs
+++ b/GVFS/GVFS.Common/GitStatusCache.cs
@@ -23,6 +23,10 @@ namespace GVFS.Common
 
         private readonly TimeSpan backoffTime;
 
+        // arbitrary value used when deciding whether to print
+        // a message about a delayed status scan.
+        private readonly TimeSpan delayThreshold = TimeSpan.FromSeconds(0.5);
+
         private string serializedGitStatusFilePath;
 
         /// <summary>
@@ -161,9 +165,8 @@ namespace GVFS.Common
             if (!shouldAllowExternalRequest)
             {
                 this.statistics.RecordBlockedRequest();
+                this.context.Tracer.RelatedWarning("GitStatusCache.IsReadyForExternalAcquireLockRequests: request blocked");
             }
-
-            this.context.Tracer.RelatedInfo("GitStatusCache.IsReadyForExternalAcquireLockRequests: isCacheReady: {0}, shouldAllowRequest: {1}", isCacheReady, shouldAllowExternalRequest);
 
             return shouldAllowExternalRequest;
         }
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/GitStatusCache.cs
```diff
@@ -311,20 +314,20 @@ private void RebuildStatusCacheIfNeeded(bool ignoreBackoff)
...
-
                 this.statistics.RecordBackgroundStatusScanRun();
 
                 bool rebuildStatusCacheSucceeded = this.TryRebuildStatusCache();
 
-                this.context.Tracer.RelatedInfo("GitStatusCache.RebuildStatusCacheIfNeeded: Done generating status. Cache is now: {0}", this.cacheState);
+                TimeSpan delayedTime = startTime - this.initialDelayTime;
+                TimeSpan statusRunTime = DateTime.UtcNow - startTime;
+
+                string message = string.Format("GitStatusCache.RebuildStatusCacheIfNeeded: Done generating status. Cache state: {0}. Status scan time: {1:0.##}s.", this.cacheState, statusRunTime.TotalSeconds);
```
kewillford: Use `$""` since that is what you are using below?

jamill: Thanks - I went and used `String.Format` in both places to be consistent. I looked at using interpolation, but then the line was pretty long.

Issue #489, opened 2018-11-08
Mount Failure when Status Cache can't run


jeschu1:
Perhaps we should issue a warning and not take down the process when Status Cache can't run?

MountId - a3d2edc96e50439eb28592076419bd1f

System.IO.DirectoryNotFoundException: Could not find a part of the path 'G:\o1\.gvfs\gitStatusCache'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.InternalCreateDirectory(String fullPath, String path, Object dirSecurityObj, Boolean checkHost)
   at System.IO.Directory.InternalCreateDirectoryHelper(String path, Boolean checkHost)
   at GVFS.Common.GitStatusCache.TryRebuildStatusCache()
   at GVFS.Common.GitStatusCache.RebuildStatusCacheIfNeeded(Boolean ignoreBackoff)
   at GVFS.Common.GitStatusCache.SerializeStatusMainThread()

jeschu1:
As dicussed offline the affected user switched his drive from G to D.  This isn't supported in VFS.  

Up to you if you want to make the StatusCache more bulletproof or punt the bug since we know it's an unsupported scenario.

jeschu1:
Thinking about this more if we were going to do anything here, i'd vote for checking if the drive changed on startup and have a error message there we can filter in telemetry.  So far I've seen two users change their drive letter and when that happens it would be great to filter that out as potential bugs on our side since the feature is unsupported.

PR #859, opened 2019-02-27
GitStatusCache: handle errors creating directory

jamill:
An error creating the status cache directory should not bring down the mount process

Fixes: #489

### DIFF
--- a/GVFS/GVFS.UnitTests/Common/GitStatusCacheTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/GitStatusCacheTests.cs
@@ -8,6 +8,7 @@ using GVFS.UnitTests.Mock.FileSystem;
 using GVFS.UnitTests.Mock.Git;
 using NUnit.Framework;
 using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace GVFS.UnitTests.Common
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.UnitTests/Common/GitStatusCacheTests.cs
```diff
@@ -154,5 +154,31 @@ public void CanRefreshCache()
                 statusCache.Shutdown();
             }
         }
+
+        [TestCase]
+        public void HandlesExceptionCreatingDirectory()
```
kewillford: Do we need to test different exception types since we are catching two different types `IOException` and `UnauthorizedAccessException` and all others would not be handled?

jamill: Interesting question - I think we could test different exceptions, and it would help verify the types of exceptions we are capturing. As is, we test that the exceptions we handle are being handled correctly (which I think provided a good bit of value on its own).

wilbaker: > As is, we test that the exceptions we handle are being handled correctly

The code also handles `UnauthorizedAccessException` should we test that it's being handled correctly (as Kevin suggested)?

One might be to change `CreateDirectoryThrowsException` to something like ` Exception ExceptionThrownByDirectory { get; set; }` and allow the test to set the specific exception it wants `CreateDirectory` to throw.

PR #861, opened 2019-02-27
[upgrade] No telemetry for non-elevated upgrade

### LINKED ISSUE #831: Upgrade run in a non-admin command prompt reports error
Running `gvfs upgrade --confirm` from a non-admin prompt *correctly* reports error on console. But its not worth sending this as an error event to live site support server.

alameenshah:
When user runs `gvfs upgrade --confirm` from a non-elevated command prompt, VFS4Git displays error on Console and triggers a Telemetry error event. This PR removes the telemetry (no other change). Rather than generating pre-check telemetry errors generally at the higher `UpgradeVerb` level, `InstallerPrerunChecker` now decides whether to log Warning or Error events depending on specific pre-check failures.

Updated events.
1. Unattended mode - Log error locally. No reporting to server.
2. Non-elevated upgrade - Log error locally. No reporting to server.
3. PrjFlt not inboxed & enabled - Report to server.
4. Service not running - Log warming. Report warning to server.
5. Installer blocked by running processes - Log error locally.

Fixes #831

### DIFF
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -198,7 +198,7 @@ namespace GVFS.Upgrader
                     Environment.NewLine,
                     $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is only supported after the \"Windows Projected File System\" optional feature has been enabled by a manual installation of VFS for Git, and only on versions of Windows that support this feature.",
                     "Check your team's documentation for how to upgrade.");
-                this.tracer.RelatedError($"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}");
+                this.tracer.RelatedWarning(metadata: null, message:$"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}", keywords: Keywords.Telemetry);
                 return false;
             }
 
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -31,7 +31,7 @@ namespace GVFS.Upgrader
                 if (this.IsUnattended())
                 {
                     consoleError = $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is not supported in unattended mode";
-                    this.tracer.RelatedError($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
+                    this.tracer.RelatedWarning($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
                     return false;
                 }
 
@@ -102,7 +102,7 @@ namespace GVFS.Upgrader
                     Environment.NewLine,
                     "Blocking processes are running.",
                     $"Run {this.CommandToRerun} again after quitting these processes - " + string.Join(", ", processList.ToArray()));
-                this.tracer.RelatedError($"{nameof(this.IsInstallationBlockedByRunningProcess)}: {consoleError}");
+                this.tracer.RelatedWarning($"{nameof(this.IsInstallationBlockedByRunningProcess)}: {consoleError}");
                 return false;
             }
 
@@ -209,7 +209,7 @@ namespace GVFS.Upgrader
                     Environment.NewLine,
                     "GVFS Service is not running.",
                     adviceText);
-                this.tracer.RelatedError($"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}");
+                this.tracer.RelatedWarning($"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}");
                 return false;
             }
 
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -37,7 +37,6 @@ namespace GVFS.Upgrader
 
                 if (!this.IsGVFSUpgradeAllowed(out consoleError))
                 {
-                    this.tracer.RelatedError($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
                     return false;
                 }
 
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -198,6 +198,7 @@ private bool IsGVFSUpgradeAllowed(out string consoleError)
                     Environment.NewLine,
                     $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is only supported after the \"Windows Projected File System\" optional feature has been enabled by a manual installation of VFS for Git, and only on versions of Windows that support this feature.",
                     "Check your team's documentation for how to upgrade.");
+                this.tracer.RelatedError($"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}");
```
jamill: According to my understanding, these should probably not be errors either. Do we want this event reported via telemetry?

//cc @wilbaker

wilbaker: @jamill I agree this event shouldn't be an error, and I'm not even sure it should be a warning.  For users running pre-RS4 Windows (or Server 2016) this message is expected and not indicative of any problems.

(I think the telemetry question is a separate one, @jrbriggs do you think this is something we'll want to track?)

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -208,6 +209,7 @@ private bool IsGVFSUpgradeAllowed(out string consoleError)
                     Environment.NewLine,
                     "GVFS Service is not running.",
                     adviceText);
+                this.tracer.RelatedError($"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}");
```
jeschu1: Same feedback here, should this be an error logged to telemetry?

jeschu1:
@alameenshah  Can you also change this line
`                this.tracer.RelatedError($"{nameof(this.IsInstallationBlockedByRunningProcess)}: {consoleError}");`
to a Warning?

I'm the original author and I shouldn't have put it as error.  The change is inline with others you have here.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
alameenshah: As discussed, changing this to warning.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
alameenshah: Updated behavior -
* Unattended mode - Log error locally. No reporting to server.
* Non-elevated upgrade - Log error locally. No reporting to server.
* PrjFlt not inboxed & enabled - Report to server.
* Service not running - Log error locally. No reporting to server.
* Installer blocked by running processes - Log error locally.

wilbaker: > PrjFlt not inboxed & enabled - Report to server

@alameenshah apologies I missed in our offline discussion that we'd be logging this as an error even in the scenarios where it's expected.

What's the reason for logging this as an error on pre-RS4 machines?

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -31,13 +31,12 @@ public virtual bool TryRunPreUpgradeChecks(out string consoleError)
...
                 {
                     consoleError = $"{GVFSConstants.UpgradeVerbMessages.GVFSUpgrade} is not supported in unattended mode";
-                    this.tracer.RelatedError($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
+                    this.tracer.RelatedWarning($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
                     return false;
                 }
 
                 if (!this.IsGVFSUpgradeAllowed(out consoleError))
                 {
-                    this.tracer.RelatedError($"{nameof(this.TryRunPreUpgradeChecks)}: {consoleError}");
```
jeschu1: Does this change the user experience?  I.e. previously they would have gotten the message on their screen.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
jeschu1: when we spoke we decided to keep the same logging level as the same what we'd get on mount.  Is it error there @alameenshah ?

GVFS/GVFS.Common/InstallerPreRunChecker.cs
alameenshah: No it does not change user experience.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
alameenshah: In mount this is treated as error.

GVFSMountProcess.cs line 27
```
if (!EnableAndAttachProjFSHandler.TryEnablePrjFlt(this.tracer, out error))
{
    this.tracer.RelatedError($"{nameof(this.Mount)}: Could not enable PrjFlt: {error}");
}
```

On pre-RS4 PCs, since this is the default PrjFlt state, I'm turning this into a warning, with live-site reporting enabled.

wilbaker: > In mount this is treated as error.

@alameenshah FYI the code you referenced (`GVFSMountProcess`) is run by GVFS.Service when auto-mounting, it is not part of `gvfs mount`.

That code (in `GVFSMountProcess`) is _not_ expected to fail on pre-RS4 machines.

wilbaker: You might be looking for the mount code here:

https://github.com/Microsoft/VFSForGit/blob/4baa37df6bde2c9a9e1917fc7ce5debd653777c0/GVFS/GVFS/CommandLine/MountVerb.cs#L120

Please note that `gvfs mount` does not care whether or not ProjFS is inbox, it only cares that it can be enabled and attached to the volume.

`gvfs upgrade` might be the only verb that behaves differently (from the user's perspective) based on whether or not they're pre-RS4).

Issue #863, opened 2019-02-27
`gvfs upgrade` can fail if its background processes are launched inside a repo


wilbaker:
We've seen a report of `PostInstall.bat` failing when it calls out to `git`:

```
---------------------------
pre-command.exe - Application Error
---------------------------
The application was unable to start correctly (0xc000ce01). Click OK to close the application. 
---------------------------
```

0xc000ce01 is returned by ProjFS when trying to read a non-existent file from inside the repo, and this is what I saw happening the ProjFS logs:

```
[2]1440.2060::02/27/2019-10:09:10.070 [prjflt] ERROR: NTSTATUS=C000CE01: [<root>\src\MSVCP140.dll] Options = 0x00200000 0x00000001(OPEN)
[3]1440.2254::02/27/2019-10:09:10.070 [prjflt] ERROR: NTSTATUS=C000CE01: [<root>\src\VCRUNTIME140.dll] Options = 0x00200000 0x00000001(OPEN)
```

To avoid this issue the upgrade processes should be run from a well known location outside of the repos (e.g. in `InvokeGitOutsideEnlistment` we use Environment.SystemDirectory, see https://github.com/Microsoft/VFSForGit/blob/2917f18d524d95fbb9454a299e0ba5a6f43591b6/GVFS/GVFS.Common/Git/GitProcess.cs#L801)

jrbriggs:
Reactivating to track port to M147

PR #866, opened 2019-02-28
Upgrade: run upgrader in system diretory

jamill:
Run the upgrade tool from the system directory to avoid potential
interactions with the local directory. This could happen if the
installer attempts to run a batch script, which might pick environment
from the local location (e.g. if it runs a git command, it might try and
run the locally configured git hooks).

Fixes: #863


jamill:
> Today upgrader tool is standalone, but may be in future if it references scripts or other resources packaged with it, it would become easier for it to find them in current working directory.

I think this is an interesting idea - for now, I will keep with this simpler change, but we might need to do something like this in the future.

PR #867, opened 2019-02-28
Upgrade: handle not finding git installation

### LINKED ISSUE #864: Upgrader: handle  not being able to find git installation more gracefully
Upgrader should more gracefully handle errors when creating a tracer. Currently, if the tracer is unable to find the local git installation, it will throw an error which will crash the upgrade command.

This scenario should be handled more gracefully.

jamill:
Fixes an issue where the upgrader would crash if it was unable to find the local git installation. With this change, we will handle this scenario better.

Previously, the tracer would throw an exception if it was unable to find the git installation. This change:

1) The tracer will handle this by assuming it is not configured to report telemetry

2) The NuGet Upgrader includes a check to make sure it can find the git installation, as it is required for acquiring credentials.

When running the NuGet upgrader with this change, but in an environment where it can't find the git installation, it will report the following error:

```
>gvfs upgrade
ERROR: NuGetUpgrader: Unable to locate git installation. Ensure git is installed and try again.
```

Fixes #864

### DIFF
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -119,6 +119,13 @@ namespace GVFS.Common.NuGetUpgrade
                 return false;
             }
 
+            string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+            if (string.IsNullOrEmpty(GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath()))
+            {
+                error = $"NuGetUpgrader: Unable to locate git installation. Ensure git is installed and try again.";
+                return false;
+            }
+
             string authUrl;
             if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(upgraderConfig.FeedUrl, out authUrl, out error))
             {
@@ -126,7 +133,7 @@ namespace GVFS.Common.NuGetUpgrade
             }
 
             if (!TryGetPersonalAccessToken(
-                    GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath(),
+                    gitBinPath,
                     authUrl,
                     tracer,
                     out string token,
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -79,6 +79,13 @@ namespace GVFS.Platform.Windows
         {
             string gitBinRoot = this.GitInstallation.GetInstalledGitBinPath();
 
+            // If we do not have a git binary, then we cannot check if we should set up telemetry
+            // We also cannot log this, as we are setting up tracer.
+            if (string.IsNullOrEmpty(gitBinRoot))
+            {
+                yield break;
+            }
+
             ETWTelemetryEventListener etwListener = ETWTelemetryEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
             if (etwListener != null)
             {


GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -79,6 +79,13 @@ public override IEnumerable<EventListener> CreateTelemetryListeners(string provi
         {
             string gitBinRoot = this.GitInstallation.GetInstalledGitBinPath();
 
+            // If we do not have a git binary, then we cannot check if we should set up telemetry
+            // We also cannot log this, as we are setting up tracer.
+            if (string.IsNullOrEmpty(gitBinRoot))
```
jeschu1: To confirm this means no log file correct?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -117,6 +117,13 @@ public class NuGetUpgrader : ProductUpgrader
             if (!upgraderConfig.IsReady(out error))
             {
                 return false;
+            }
+
+            string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+            if (string.IsNullOrEmpty(GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath()))
+            {
+                error = $"NuGetUpgrader: Unable to locate git installation. Ensure git is installed and try again.";
```
jeschu1: Should we add a timestamp?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
jamill: This means to telemetry. We will still get local logs.

jeschu1: In that case probably don't need a timestamp per my other comment

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: I thought about this - but we don't usually include timestamps with our error messages, and since we will have local logs, the timestamps will be present there. If we expected to not have logs, then I could see the making an exception from our general pattern to include a timestamp... Open to other thoughts on this, though.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
jeschu1: Does this change the behavior for all telemetry listeners?

Are we ok with that change for other paths?  I'd expect most would break shortly after anyways, since git commands would fail.

jamill: > Are we ok with that change for other paths? I'd expect most would break shortly after anyways, since git commands would fail.

Yes - the change I am proposing here is to handle this generally (with the degradation that it will not report telemetry). From looking at how several commands construct the tracer, I think today we will generally crash if the tracer fails to find the git installation, and degrading to at least not crash seemed preferable.

If we want to do more around this, (I could see maybe we want to log when the tracer is not able to look up the telemetry switch), I propose we do that in a followup issue. This was our first report of this behavior, so it doesn't seem like a wide spread problem.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -117,6 +117,13 @@ public class NuGetUpgrader : ProductUpgrader
             if (!upgraderConfig.IsReady(out error))
             {
                 return false;
+            }
+
+            string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+            if (string.IsNullOrEmpty(GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath()))
```
wilbaker: Do we need to add a similar check in verbs and/or GVFS.Mount.exe?

I'm not sure if looking for git is part of our usual start up process or if we've been depending on the tracer throwing an exception to prevent the mount from succeeding.

jamill: As we haven't had reports of this except for during upgrade, I was addressing the immediate specific issue.

From looking at the code, we don't have error handling around creating the tracer, so I don't think we depended on this behavior.. otherwise, I would have expected us to handle this error and exited gracefully.

If we need larger follow-up changes, I was suggesting we do that separately.

wilbaker: > I was addressing the immediate specific issue

I should clarify my comment, I think the change in `NuGetUpgrader` is good.  The concern I had was with the more general change in `CreateTelemetryListeners`.

Today if a user tries to run VFS4G processes without git installed we get some automatic protection by virtue of the tracer requiring that `gitBinRoot` be set.  With the change to `CreateTelemetryListeners`, how (if at all) does the user experience change if they run VFS4G verbs (or GVFS.Mount.exe) without git installed?

If those processes don't fail right away there's a risk that GVFS.Mount.exe might stay mounted until some random time in the future when it ends up needing git to be present for whatever work it's doing (e.g. background tasks).

jeschu1: We could also CreateTelemetryListeners to take a flag, 'failIfMissingGit'.  If we only want to change behavior for the upgrader.

jamill: All Verbs that derive from [GVFSVerb.ForExistingEnlistment](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS/CommandLine/GVFSVerb.cs#L1083), check that there is a git installation. This covers many, but not all verbs.

The other verbs are:


 Verb | Notes 
------ | ----------
`RepairVerb` | [Calls ](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS/CommandLine/RepairVerb.cs#L42)`CreateWithoutRepoUrlFromDirectory`, which will [error ](https://github.com/jamill/VFSForGit/blob/d3b3c661ecb71851c53e7f0cc9e1ab2228d4c241/GVFS/GVFS.Common/Enlistment.cs#L18-L21)if it can't find git installation.
`CloneVerb` | Has a [built in check ](https://github.com/Microsoft/VFSForGit/blob/c9785b365a783db6fdf65429be1034e6ac7ee1a6/GVFS/GVFS/CommandLine/CloneVerb.cs#L296)
`LogVerb` and `ConfigVerb` and `UnmountVerb` |  do not construct a tracer.

wilbaker: @jamill thanks for doing the leg work to confirm that VFS4G will check elsewhere that git is installed.

PR #868, opened 2019-02-28
Upgrade: run upgrader in system diretory

jamill:
Run the upgrade tool from the system directory to avoid potential
interactions with the local directory. This could happen if the
installer attempts to run a batch script, which might pick environment
from the local location (e.g. if it runs a git command, it might try and
run the locally configured git hooks).

(cherry picked from commit 714867872d55d7d3a6a9fcb2a7d47763c0dd234b)


jrbriggs:
Closes #863

PR #871, opened 2019-02-28
Update git to include midx verify performance fix

jamill:
Update git to include the following fix:
https://github.com/Microsoft/git/pull/124


wilbaker:
@jamill can you update the description to include the Git issue (or PR) numbers for the MIDX perf fix?

PR #874, opened 2019-03-01
Linux/Mac: remove unused NotificationMapping class

chrisd8088:
The `NotificationMapping` class is used only on Windows to fill out the [`PRJ_NOTIFICATION_MAPPING` structure](https://docs.microsoft.com/en-us/windows/desktop/api/projectedfslib/ns-projectedfslib-prj_notification_mapping) supported by the PrjFlt filter driver.

This allows the Windows implementation to defer filtering of file operation notifications to the filter driver so that, for example, the [`NotifyPreDeleteHandler`](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs#L1241) can assume it will only be called for events on the Git index file.

By contrast, the macOS and Linux ProjFS implementations do not support this functionality in the lower-level libraries, and instead filter events by file path in the handlers, e.g., by checking for the Git index file path in the [`OnPreDelete`](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L495) handler.

Therefore we can remove their respective versions of this class.

(As a future possibility, the Linux libprojfs library could support such functionality, in a manner akin to -- or by actually using -- the [inotify(7)](http://man7.org/linux/man-pages/man7/inotify.7.html) interface. If we choose to pursue that option, we can re-introduce a dedicated Linux class or structure to represent the paths and event masks which is more aligned with the underlying Linux APIs.)

/cc @jrbriggs @kivikakk

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/NotificationMapping.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/NotificationMapping.cs
@@ -1,8 +1,0 @@
-﻿namespace PrjFSLib.Linux
-{
-    public class NotificationMapping
-    {
-        public NotificationType NotificationMask { get; set; }
-        public string NotificationRelativeRoot { get; set; }
-    }
-}
--- a/ProjFS.Mac/PrjFSLib.Mac.Managed/NotificationMapping.cs
+++ b/ProjFS.Mac/PrjFSLib.Mac.Managed/NotificationMapping.cs
@@ -1,8 +1,0 @@
-﻿namespace PrjFSLib.Mac
-{
-    public class NotificationMapping
-    {
-        public NotificationType NotificationMask { get; set; }
-        public string NotificationRelativeRoot { get; set; }
-    }
-}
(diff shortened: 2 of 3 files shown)


chrisd8088:
Hi @jrbriggs ... would you have time to take a look at this small PR?  I think it should be relatively uncontroversial; I think I recall @sanoursa saying he wanted to remove this class from the Mac side, and we don't need it on the Linux side either.  Thoughts?

wilbaker:
@chrisd8088 @jrbriggs I think it's OK to remove the code now (it is dead code).

I suspect that in the future we'll want to add support for kernel side filtering on the Mac for performance, and we can always re-add this code then.

chrisd8088:
Thanks for looking at this, @wilbaker and @jrbriggs ... by all means merge into `features/linuxprototype` whenever you get a chance, and have a great weekend!

chrisd8088:
Thanks @jrbriggs!  Will update github/VFSForGit now to match ... :-)

PR #875, opened 2019-03-01
[upgrade] Test reliability

### LINKED ISSUE #852: NoReminderForLeftOverDownloads test occasionally fails
```
Errors, Failures and Warnings

  Expected: False
  But was:  True
```

alameenshah:
The issue is that when gvfs upgrade verb uses Nuget server for upgrades when Nuget config is available. The failing test sets the upgrade ring to "None". Then creates dummy upgrade marker file (C:\ProgramData\GVFS\GVFS.Upgrade\highestAvailableVersion). Then launches upgrade verb. Since the upgrade ring is None, the test expects upgrade verb to delete the marker file. This used to work. But now, upgrade verb will use Nuget upgrader when Nuget config is availble. Nuget upgrader ignores upgrade ring and checks for upgrade. It will not delete the marker file, if it finds one. Updated the test to delete Nuget config before running.

Fixes #852

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
@@ -60,7 +60,16 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         public void NoReminderForLeftOverDownloads()
         {
             this.VerifyServiceRestartStopsReminder();
+
+            // This test should not use Nuget upgrader because it will usually find an upgrade
+            // to download.  The "None" ring config doesn't stop the Nuget upgrader from checking
+            // its feed for updates, and the VFS4G binaries installed during functional test
+            // runs typically have a 0.X version number (meaning there will always be a newer
+            // version of VFS4G available to download from the feed).
+            this.ReadNugetConfig(out string feedUrl, out string feedName);
+            this.DeleteNugetConfig();
             this.VerifyUpgradeVerbStopsReminder();
+            this.WriteNugetConfig(feedUrl, feedName);
         }
 
         [TestCase]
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -82,6 +89,35 @@ public void UpgradeTimerScheduledOnServiceStart()
             timerScheduled.ShouldBeTrue();
         }
 
+        private void ReadNugetConfig(out string feedUrl, out string feedName)
+        {
+            ProcessResult result = ProcessHelper.Run("gvfs", "config upgrade.feedurl");
```
wilbaker: Rather than adding the calls to `gvfs config` in this test fixture, could you add helper method(s) to [GVFSProcess.cs](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs)?

That's where we've been adding the code for calling the other VFS4G verbs during the functional tests.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -60,7 +60,14 @@ public void RemindWhenUpgradeAvailable()
         public void NoReminderForLeftOverDownloads()
         {
             this.VerifyServiceRestartStopsReminder();
+
+            // Nuke the nuget config before running VerifyUpgradeVerbStopsReminder.
+            // This test should not use Nuget upgrader, because None ring config
+            // is ignored by Nuget upgrader.
```
wilbaker: Could you add a bit more detail to the comment regarding why we can't use the Nuget upgrader during this test?

Specifically, it'd be good to capture why the Nuget upgrader will always find something to download.

Maybe something like:

```
// This test should not use Nuget upgrader because it will usually find an upgrade
// to download.  The "None" ring config doesn't stop the Nuget upgrader from checking
// its feed for updates, and the VFS4G binaries installed during functional test
// runs typically have a 0.X version number (meaning there will always be a newer
// version of VFS4G available to download from the feed).
```

PR #876, opened 2019-03-02
[Testing] Update PrjFSLib to create sparse files

nickgra:
Go for the simplest approach of seeking to the end and writing '\0' to start.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -382,13 +382,8 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
         goto CleanupAndFail;
     }
     
-    // Expand the file to the desired size
-    if (ftruncate(fileno(file), fileSize))
-    {
-        result = PrjFS_Result_EIOError;
-        goto CleanupAndFail;
-    }
-    
+    fseek(file, fileSize-1, SEEK_SET);
+    fputs("\0", file);
     fclose(file);
     file = nullptr;
     


pmj:
I think this should work for APFS, but does `ftruncate()` not have the same effect there?

HFS+ doesn't support sparse files, my suggestion there would be to write a compressed file with all zeroes, which should fit into a single 4KiB xattr extent for all but the most gigantic files.

alameenshah:
Can we also add a test to verify this? (I think `stat` on the file can be used to read the size as would be seen by user and `du` could be used to verify actual disk usage of the file.)

pmj:
> Can we also add a test to verify this? (I think `stat` on the file can be used to read the size as would be seen by user and `du` could be used to verify actual disk usage of the file.)

On Linux, this is certainly true, on macOS I'm not having much luck with either `stat` or `du`:

```
$ rm -f ./sparse
$ dd oseek=1000 count=1 if=/dev/zero of=./sparse
1+0 records in
1+0 records out
512 bytes transferred in 0.000031 secs (16519105 bytes/sec)
$ stat -f "%z %b" sparse
512512 1008
$ du sparse 
1008	sparse
```

This is on APFS on macOS 10.14.3, so in theory sparse files should be supported.

alameenshah:
> > Can we also add a test to verify this? (I think `stat` on the file can be used to read the size as would be seen by user and `du` could be used to verify actual disk usage of the file.)
> 
> On Linux, this is certainly true, on macOS I'm not having much luck with either `stat` or `du`:
> 
> ```
> $ rm -f ./sparse
> $ dd oseek=1000 count=1 if=/dev/zero of=./sparse
> 1+0 records in
> 1+0 records out
> 512 bytes transferred in 0.000031 secs (16519105 bytes/sec)
> $ stat -f "%z %b" sparse
> 512512 1008
> $ du sparse 
> 1008	sparse
> ```
> This is on APFS on macOS 10.14.3, so in theory sparse files should be supported.

Tried this on on 10.13.6 APFS (am using `if=/dev/null`) -
```
$ rm -r ./sparse
$ dd oseek=10000 count=1 if=/dev/null of=./sparse
0+0 records in
0+0 records out
0 bytes transferred in 0.000047 secs (0 bytes/sec)
$ stat -f "%z %b" sparse
5120000 0
$ du sparse
0	sparse
```

alameenshah:
> HFS+ doesn't support sparse files, my suggestion there would be to write a compressed file with all zeroes, which should fit into a single 4KiB xattr extent for all but the most gigantic files.

How does this work? Do you mean keeping a compressed all-zero template file, that could be used for placeholder creation?

pmj:
> How does this work? Do you mean keeping a compressed all-zero template file, that could be used for placeholder creation?

Compressed files on HFS+ (“decmpfs”) consist of a header struct followed by the compressed payload written to the special xattr. There are a bunch of compression modes available, including zlib/deflate. So you can just send an appropriate number of zero bytes through deflate() and write the result to the xattr to give you an all-0 file with that size. If you know the deflate bitstream format, you can also just keep explicitly emitting commands for blocks of repeated zero bytes to avoid the cost of the whole compression algorithm. Alternatively there are probably cheaper/simpler algorithms you can choose which are just as good at repeating 0 bytes.

wilbaker:
@nickgra @jeschu1 can this PR be closed now that we've gone with the approach of writing empty files?

jeschu1:
> @nickgra @jeschu1 can this PR be closed now that we've gone with the approach of writing empty files?
> (See #1130 for that change)

I'd thinks so but @nickgra is the boss of this PR :-)

nickgra:
> > @nickgra @jeschu1 can this PR be closed now that we've gone with the approach of writing empty files?
> > (See #1130 for that change)
> 
> I'd thinks so but @nickgra is the boss of this PR :-)

Thanks for the reminder, closing this stale PR.

Issue #757, opened 2019-02-08
[upgrade] Switch from calling `Exit` to letting unhandled exceptions bring down the process


jrbriggs:
Some investigation is required to makes sure that all native/background threads with unhandled exceptions will crash the VFS4G process.

wilbaker:
If this change is made, the contribution guidelines should be updated (see #879)

PR #879, opened 2019-03-04
Add contributing guidelines to VFSForGit repo

### LINKED ISSUE #537: Add coding style guide for contributors to follow.

wilbaker:
Resolves #537.

Add contributing guidelines to the VFSForGit repo.

The guidelines I've included have been collected from PR feedback and discussions since VFS4G development started, there shouldn't be any surprises.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,301 @@
+# Contributing to VFS for Git
+
+Thank you for taking the time to contribute!
+
+## Guidelines
+
+* [Code of Conduct](#code-of-conduct)
+* [Design Reviews](#design-reviews)
+* [Platform Specific Code](#platform-specific-code)
+* [Tracing and Logging](#tracing-and-logging)
+* [Error Handling](#error-handling)
+* [Background Threads](#background-threads)
+* [Coding Conventions](#coding-conventions)
+* [Testing](#testing)
+
+## Code of Conduct
+
+This project has adopted the [Microsoft Open Source Code of
+Conduct](https://opensource.microsoft.com/codeofconduct/).
+For more information see the [Code of Conduct
+FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or
+contact [opencode@microsoft.com](mailto:opencode@microsoft.com)
+with any additional questions or comments.
+
+## Design Reviews
+
+Architectural changes and new features should start with a design review.  It's easier and wastes less time to incorporate feedback at this stage.
+
+The design review process is as follows:
+
+1. Create a pull request that contains a design document in Markdown (.md) format for the proposed change.  Assign the `design-doc` label to the pull request.
+2. Use the pull request for design feedback and for iterating on the design.
+3. Once the design is approved, create a new issue with a description that includes the final design document.  Include a link to the pull request that was used for discussion.
+4. Close (without merging!) the pull request used for the design discussion.
+
+## Platform Specific Code
+
+- *Prefer cross-platform code to platform-specific code*
+
+  Cross-platform code is more easily reused.  Reusing code reduces the amount of code that must be written, tested, and maintained.
+
+- *Platform specific code, and only platform specific code, should go in `GVFSPlatform`*
+
+  When platform specific code is required, it should be placed in `GVFSPlatform` or one of the platforms it contains (e.g. `IKernelDriver`)
+
+## Tracing and Logging
+
+- *The "Error" logging level is reserved for non-retryable errors that result in I/O failures or the VFS4G process shutting down*
+
+  The expectation from our customers is that when VFS4G logs an "Error" level message in its log file either:
+    * VFS4G had to shut down unexpectedly
+    * VFS4G encountered an issue severe enough that user-initiated I/O would fail.
+
+- *Log full exception stacks*
+
+  Full exception stacks (i.e. `Exception.ToString`) provide more details than the exception message alone (`Exception.Message`). Full exception stacks make root-causing issues easier.
+  
+- *Do not display full exception stacks to users*
+
+  Exception call stacks are not usually actionable for the user.  Users frequently (sometimes incorrectly) assume that VFS4G has crashed when shown a full stack.  The full stack *should* be included in VFS4G logs, but *should not* be displayed as part of the error message provided to the user.
+
+- *Include relevant details when logging exceptions*
+
+  Sometimes an exception call stack alone is not enough to root cause failures in VFS4G.  When catching (or throwing) exceptions, log relevant details that will help diagnose the issue.  As a general rule, the closer an exception is caught to where it's thrown, the more relevant details there will be to log.
+
+  Example:
+  ```
+  catch (Exception e)
+  {
+    EventMetadata metadata = new EventMetadata();
+    metadata.Add("Area", "Mount");
+    metadata.Add(nameof(enlistmentHookPath), enlistmentHookPath);
+    metadata.Add(nameof(installedHookPath), installedHookPath);
+    metadata.Add("Exception", e.ToString());
+    context.Tracer.RelatedError(metadata, $"Failed to compare {hook.Name} version");
+  }
+  ```
+
+## Error Handling
+
+- *Fail fast: An error or exception that risks data loss or corruption should shut down VFS4G immediately*
+
+  Preventing data loss and repository corruption is critical.  If an error or exception occurs that could lead to data loss, it's better to shut down VFS4G than keep the repository mounted and risk corruption.
+
+- *Do not catch exceptions that are indicative of a programmer error (e.g. `ArgumentNullException`)*
+
+  Any exceptions that result from programmer error (e.g. `ArgumentNullException`) should be discovered as early in the development process as possible.  Avoid `catch` statements that would hide these errors (e.g. `catch(Exception)`).
+
+  The only exception to this rule is for [unhandled exceptions in background threads](#bgexceptions)
+
+- *Do not use exceptions for normal control flow*
+
+  Prefer writing code that does not throw exceptions.  The `TryXXX` pattern, for example, avoids the performance costs that come with using exceptions.  Additionally, VFS4G typically needs to know exactly where errors occur and handle the errors there.  The `TryXXX` pattern helps ensure errors are handled in that fashion.
+
+  Example: Handle errors where they occur (good):
+
+  ```
+  bool TryDoWorkOnDisk(string fileContents, out string error)
+  {
+    if (!TryCreateReadConfig())
+    {
+      error = "Failed to read config file";
+      return false;
+    }
+    
+    if (!TryCreateTempFile(fileContents))
+    {
+      error = "Failed to create temp file";
+      return false;
+    }
+  
+    if (!TryRenameTempFile())
+    {
+      error = "Failed to rename temp file";
+      if (!TryDeleteTempFile())
+      {
+        error += ", and failed to cleanup temp file";
+      }
+    
+      return false;
+    }
+    
+    error = null;
+    return true;
+  }
+  ```
+
+  Example: Handle errors in `catch` without knowing where they came from (bad):
+
+  ```
+  bool TryDoWorkOnDisk(string fileContents, out string error)
+  {
+    try
+    {
+      CreateReadConfig();
+      CreateTempFile(fileContents);
+      RenameTempFile();
+    }
+    catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
+    {
+      error = "Something went wrong doing work on disk";
+      
+      try
+      {
+        if (TempFileExists())
+        {
+          DeleteTempFile();
+        }
+      }
+      catch (Exception e) when (e is IOException || e is UnauthorizedAccessException)
+      {
+        error += ", and failed to cleanup temp file";
+      }
+    
+      return false;
+    }
+  
+    error = null;
+    return true;
+  }
+  ```
+
+- *Provide the user with user-actionable messages whenever possible*
+
+  Don't tell a user what went wrong.  Help the user fix the problem.
+
+  Example:
+  > `"You can only specify --hydrate if the repository is mounted. Run 'gvfs mount' and try again."`
+
+## Background Threads
+
+- *Avoid using the thread pool (and avoid using async)*
+
+  `HttpRequestor.SendRequest` makes a [blocking call](https://github.com/Microsoft/VFSForGit/blob/4baa37df6bde2c9a9e1917fc7ce5debd653777c0/GVFS/GVFS.Common/Http/HttpRequestor.cs#L135) to `HttpClient.SendAsync`.  That blocking call consumes a thread from the managed thread pool.  Until that design changes, the rest of VFS4G must avoid using the thread pool unless absolutely necessary.  If the thread pool is required, any long running tasks should be moved to a separate thread managed by VFS4G itself (see [GitMaintenanceQueue](https://github.com/Microsoft/VFSForGit/blob/4baa37df6bde2c9a9e1917fc7ce5debd653777c0/GVFS/GVFS.Common/Maintenance/GitMaintenanceQueue.cs#L19) for an example).
+
+  Long-running or blocking work scheduled on the managed thread pool can prevent the normal operation of VFS4G.  For example, it could prevent downloading file sizes, loose objects, or file contents in a timely fashion.
+
+- <a id="bgexceptions"></a>*Catch all exceptions on long-running tasks and background threads*
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Error Handling
+
+- *Fail fast: An error or exception that risks data loss or corruption should shut down VFS4G immediately*
+
+  Preventing data loss and repository corruption is critical.  If an error or exception occurs that could lead to data loss it's better to shut down VFS4G than keep the repository mounted and risk corruption.
+
+- *Do not catch exceptions that are indicative of a programmer error (e.g. `ArgumentNullException`)*
+
+  Any exceptions that result from programmer error (e.g. `ArgumentNullException`) should be caught as early in the development process as possible.  Avoid `catch` statements that would hide these errors (e.g. `catch(Exception)`).
```
kewillford: Using `caught` here makes it sound like you are talking about catching in the code when I think you mean found, discovered, or exposed right?

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+  Long-running or blocking work scheduled on the managed thread pool can prevent the normal operation of VFS4G.  For example, it could prevent downloading file sizes, loose objects, or file contents in a timely fashion.
+
+- <a id="bgexceptions"></a>*Catch all exceptions on long-running tasks and background threads*
+
+  Wrap all code that runs in the background thread in a top-level `try/catch(Exception)`.  Any exceptions caught by this handler should be logged, and then force VFS4G to terminate with `Environment.Exit`.  It's not saft to allow VFS4G to continue to run after an exception that would have been unhandled in a long-running task (created with `TaskCreationOption.LongRunning`) or in a background thread.
```
kewillford: >It's not saft to allow 

It's not safe to allow

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+  An example of this pattern can be seen [here](https://github.com/Microsoft/VFSForGit/blob/4baa37df6bde2c9a9e1917fc7ce5debd653777c0/GVFS/GVFS.Virtualization/Background/BackgroundFileSystemTaskRunner.cs#L233) in `BackgroundFileSystemTaskRunner.ProcessBackgroundTasks`.
```
kewillford: Attach the code link to something other than `here` if possible like:

```
can be seen in [`BackgroundFileSystemTaskRunner.ProcessBackgroundTasks`](https://github.com/Microsoft/VFSForGit/blob/4baa37df6bde2c9a9e1917fc7ce5debd653777c0/GVFS/GVFS.Virtualization/Background/BackgroundFileSystemTaskRunner.cs#L233)
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Coding Conventions
+
+- *Most C# coding style guidelines are covered by StyleCop*
+
+  Fix any StyleCop issues reported in changed code. When adding new projects to VFS4G, be sure that StyleCop is analyzing them as part of the build.
+
+- *Prefer explicit types to interfaces and implicitly typed variables*
+
+  Avoid the use of `var` (C#) and `auto` (C++), and prefer concrete/explicit types to interface (e.g. prefer `List` to `IList`).
```
kewillford: Is `dynamic` something we should add to this list?

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+- *Add new interfaces when it makes sense for the product, not simply for unit testing*
+
+  When a class needs to be mocked (or have a subset of its behavior mocked), prefer using virtual methods instead of adding a new interface.  VFS4G uses interfaces when multiple implementations of the interface exist in the product code.
+
+- *Check for `null` using the equality (`==`) and inequality (`!=`) operators rather than `is`*
+
+  A corollary to this guideline is that equality/inequality operators that break `null` checks should not be added (see [this post](https://stackoverflow.com/questions/40676426/what-is-the-difference-between-x-is-null-and-x-null) for an example).
+
+- *Use `nameof(...)` rather than hardcoded strings*
```
kewillford: Do we need to mention that in some cases a hard-coded string is preferred for reporting to make sure we don't have to update reports when the name changes?

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+  Comprehensive tests are essential for maintaining the health and quality of the product.  For more details on writing tests see [Authoring Tests](https://github.com/Microsoft/VFSForGit/blob/master/AuthoringTests.md).
+
+- *Functional tests are black-box tests and should not build against any VFS4G product code*
+
+  Keeping the code separate helps ensure that bugs in the product code do not compromise the integrity of the functional tests.
+
+### Unit Tests
+
+- *Add `ExceptionExpected` to unit tests that are expected to have exceptions*
```
kewillford: I am thinking we need to change this name because when I see it I think `ExpectedException` that NUnit has.  Maybe this should be something like `SkipBreakOnException` or `DoNotBreakOnExceptions`?  Something to think about at least.

CONTRIBUTING.md
```diff
@@ -0,0 +1,222 @@
...
+* [Coding Conventions](#coding-conventions)
+* [Testing](#testing)
+
+## Design Reviews
+
+Architectural changes and new features should start with a design review.  It's easier and wastes less time to incorporate feedback at this stage.
+
+The design review process is as follows:
+
+1. Create a pull request that contains a design document for the proposed change.  Assign the `design-doc` label to the pull request.
```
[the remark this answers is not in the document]
nickgra: I think we should specify Markdown here vs something like a docx that won't render in GitHub's web UI.

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+  ```
+  catch (Exception e)
+  {
+    EventMetadata metadata = new EventMetadata();
+    metadata.Add("Area", "Mount");
+    metadata.Add(nameof(enlistmentHookPath), enlistmentHookPath);
+    metadata.Add(nameof(installedHookPath), installedHookPath);
+    metadata.Add("Exception", e.ToString());
+    context.Tracer.RelatedError(metadata, $"Failed to compare {hook.Name} version");
+  ```
```
nickgra: nit: Missing } in the close block

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Error Handling
+
+- *Fail fast: An error or exception that risks data loss or corruption should shut down VFS4G immediately*
+
+  Preventing data loss and repository corruption is critical.  If an error or exception occurs that could lead to data loss it's better to shut down VFS4G than keep the repository mounted and risk corruption.
```
nickgra: I think this flows better with a comma?

If an error or exception occurs that could lead to data loss **,** it's better to shut down VFS4G than keep the repository mounted and risk corruption.

CONTRIBUTING.md
```diff
@@ -0,0 +1,232 @@
...
+- *Do not use exceptions for normal control flow*
+
+  Prefer to write code that avoids exceptions being thrown (e.g. the `TryXXX` pattern).  There are performance costs to using exceptions for control flow, and VFS4G code typically needs to handle errors where (and as soon as) they occur.
+   
+- *Provide the user with user-actionable messages whenever possible*
+
+  Don't tell a user what went wrong.  Help the user fix the problem.
+
+  Example:
+  > "You can only specify --hydrate if the repository is mounted. Run 'gvfs mount' and try again."
```
nickgra: Code quotes `` here instead of ''?

CONTRIBUTING.md
nickgra: FWIW, I also always read this as `ExpectedException` and have to re-read it until my brain processes it correctly :)

CONTRIBUTING.md
turbonaitis: Why are you discouraging the use of interfaces? Seems to go against a few of the SOLID principles.

CONTRIBUTING.md
wilbaker: Filed #883 for this

CONTRIBUTING.md
wilbaker: That's a good idea, thanks!

CONTRIBUTING.md
wilbaker: I'll add both

CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+
+  Long-running or blocking work scheduled on the managed thread pool can prevent the normal operation of VFS4G.  For example, it could prevent downloading file sizes, loose objects, or file contents in a timely fashion.
+
+- <a id="bgexceptions"></a>*Catch all exceptions on long-running tasks and background threads*
+
+  Wrap all code that runs in the background thread in a top-level `try/catch(Exception)`.  Any exceptions caught by this handler should be logged, and then force VFS4G to terminate with `Environment.Exit`.  It's not safe to allow VFS4G to continue to run after an exception that would have been unhandled in a long-running task (created with `TaskCreationOption.LongRunning`) or in a background thread.
```
jamill: Were there details / discussion to include regarding why `Environment.Exit` is preferred over allowing the unhandled exception to propagate?

CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+
+- *Do not catch exceptions that are indicative of a programmer error (e.g. `ArgumentNullException`)*
+
+  Any exceptions that result from programmer error (e.g. `ArgumentNullException`) should be discovered as early in the development process as possible.  Avoid `catch` statements that would hide these errors (e.g. `catch(Exception)`).
+
+  The only exception to this rule is for [unhandled exceptions in background threads](#bgexceptions)
+
+- *Do not use exceptions for normal control flow*
+
+  Prefer to write code that avoids exceptions being thrown (e.g. the `TryXXX` pattern).  There are performance costs to using exceptions for control flow, and VFS4G code typically needs to handle errors where (and as soon as) they occur.
```
jamill: ```VFS4G code typically needs to handle errors where (and as soon as) they occur.```

I am not sure what this means in this context...

CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+
+  Cross-platform code is more easily reused.  Reusing code reduces the amount of code that must be written, tested, and maintained.
+
+- *Platform specific code, and only platform specific code, should go in `GVFSPlatform`*
+
+  When platform specific code is required, it should be placed in `GVFSPlatform` or one of the platforms it contains (e.g. `IKernelDriver`)
+
+## Tracing and Logging
+
+- *The "Error" logging level is reserved for non-retryable errors that result in I/O failures or the VFS4G process shutting down*
```
jamill: This seems to be a common issue that people stumble over. When an "unexpected condition" is encountered (exception, or other problem), it seems people use the Error helper methods. Maybe we could do something to make it easier for developers do the right thing here.

CONTRIBUTING.md
wilbaker: > Were there details / discussion to include regarding why Environment.Exit is preferred over allowing the unhandled exception to propagate?

I will add some details on this point to the doc.

We've been using `Environment.Exit` because we know that it will consistently bring down the process regardless of how the background thread was started (e.g. native thread, `new Thread(...)`, or `Task.Factory.StartNew(..., TaskCreationOptions.LongRunning)`.

What we had run into is that for some background thread types the unhandled exception would not bring down the process.  I suspect in the case of `TaskCreationOptions.LongRunning` the exception is being stored in the `Task` returned by `Task.Factory.StartNew`.

I think that once the investigation/changes for #757 are complete we can update this guideline as needed (e.g. we might land at something like "Propagate background thread exceptions and do not use `TaskCreationOptions.LongRunning` to create background threads").

I've left a note in #757 that the guidelines should be updated when that change is made.

CONTRIBUTING.md
wilbaker: I did not word that well.

What I was trying to get across is that for many I/O errors VFS4G cares about exactly where the error occurred in a sequence of operations, and that determining that failure point be a bit clunky with exceptions.

I'll work on a better description for what I mean, I've included an example below, let me know if you have any suggestions on how to re-word things.

_Example with TryXXX_

```
bool TryDoWorkOnDisk(string fileContents, out string error)
{
  if (!TryCreateReadConfig())
  {
    error = "Failed to read config file";
    return false;
  }

  if (!TryCreateTempFile(fileContents))
  {
    error = "Failed to create temp file";
    return false;
  }

  if (!TryRenameTempFile())
  {
    error = "Failed to rename temp file";
    if (!TryDeleteTempFile())
    {
        error += ", and failed to cleanup temp file";
    }

    return false;
  }

  error = null;
  return true;
}
```

_Example 2: Same behavior using exceptions_

```
bool TryDoWorkOnDisk(string fileContents, out string error)
{
  try
  {
    CreateReadConfig();
  }
  catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
  {
    error = "Failed to read config file";
    return false;
  }

  try
  {
    CreateTempFile(fileContents);
  }
  catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
  {
    error = "Failed to create temp file";
    return false;
  }

  try
  {
    RenameTempFile();
  }
  catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
  {
    error = "Failed to rename temp file";

    try
    {
      DeleteTempFile();
    }
    catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
    {
        error += ", and failed to cleanup temp file";
    }

    return false;
  }

  error = null;
  return true;
}
```

_Example 3: Catch needs to figure out what went wrong_

```
bool TryDoWorkOnDisk(string fileContents, out string error)
{
  try
  {
    CreateReadConfig();
    CreateTempFile(fileContents);
    RenameTempFile();
  }
  catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException)
  {
    error = "Something went wrong doing work on disk";

    try
    {
       if (TempFileExists())
       {
           DeleteTempFile();
        }
    }
    catch (Exception e) when (e is IOException || e is UnauthorizedAccessException)
    {
        error += ", and failed to cleanup temp file";
    }

    return false;
  }

  error = null;
  return true;
}
```

CONTRIBUTING.md
wilbaker: I completely agree, I will file a work item for this.

CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+
+   Example:
+   ```
+   // Functions declared in VirtualizationRoots.h have "VirtualizationRoot_" prefix
+   bool VirtualizationRoot_IsOnline(VirtualizationRootHandle rootHandle);
+
+   // Static helper function in VirtualizationRoots.cpp has no prefix
+   static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t vnode, const FsidInode& vnodeFsidInode);
+   ```
+
```
pmj: We had a slightly longer, more comprehensive list for C++ style guidelines on a wiki somewhere prior to the move to GitHub. I won't claim to know where that ended up…

CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+## Testing
+
+- *Add new unit and functional tests when making changes*
+
+  Comprehensive tests are essential for maintaining the health and quality of the product.  For more details on writing tests see [Authoring Tests](https://github.com/Microsoft/VFSForGit/blob/master/AuthoringTests.md).
+
```
pmj: It might be worth mentioning that this applies largely to C#-side tests. For the Mac PrjFS native code tests we could probably mention that they're currently rapidly evolving, but implemented via the XCTest framework and to refer to the existing code for inspiration until things have settled down in that area.

CONTRIBUTING.md
wilbaker: @pmj thanks for the tip, I'll see if I can dig those up.

wilbaker: @pmj I believe I've found the style guide you mentioned in a OneNote.  

I've filed #894 to move those guidelines over here and will do that in a follow-up PR.

CONTRIBUTING.md
wilbaker: Filed #895 for this

CONTRIBUTING.md
wilbaker: For now I'm labeling this C# unit tests and have filed #897

wilbaker:
Next push will squash commits

Issue #828, opened 2019-02-22
Placeholder update during git checkout takes a long time


turbonaitis:
```
[2019-02-20 21:29:15 +02:00] ClearNegativePathCache_ClearedCache {"Area":"GitIndexProjection","Message":"ClearNegativePathCache: Cleared negative path cache","totalEntryCount":0,"negativePathCacheUpdatedForGitCount":0,"clearCacheResult.Result":"GVFS.Virtualization.FileSystem.FileSystemResult","clearCacheResult.RawResult":1}
[2019-02-20 21:29:22 +02:00] UpdatePlaceholders (Start) {"File placeholder count":117135,"Folder placeholders count":29978}
[2019-02-20 21:29:25 +02:00] Warning {"WarningMessage":"WaitToAcquireGVFSLock: BackgroundFileSystemTaskRunner unable to acquire lock, retrying"}
[2019-02-20 21:29:26 +02:00] UpdatePlaceholders (Stop) {"DurationMs":4103}
```
Loading PlaceholderListDatabase took 7s, updating placeholders took another 4s. This is just one sample, but this behavior is reproducable with this amount of hydrated files.

turbonaitis:
@jrbriggs are you guys seeing any such behavior on your side, when the number of placeholders is large? Or is that not a concern for your use cases?

jrbriggs:
We haven't seen this perf problem yet @turbonaitis, but we have some suspicions about what might cause this, and it will likely apply to our large Mac repos too.

jrbriggs:
Related: #873

turbonaitis:
Not sure - #873 mostly talks about the placeholder files themselves, if I'm reading it correctly, while in this case the bottleneck seems to be more related to the implementation of `PlaceholderListDatabase`. I'm working on a PR to address this, hope to submit in the next couple of days (but essentially I'm switching from text format to binary format for `PlaceholderListDatabase`). We can discuss the solution, once the code is there.

turbonaitis:
Also, it would seem to me, that performance should be just as bad on Windows, as the code is shared.

jrbriggs:
Closed with #1140

Issue #442, opened 2018-10-30
Harden placeholder list persistence, and persist before writing placeholder


wilbaker:
(Copying from old backlog)

Currently VFS4G can lose track of a placeholder if either:

1) There is a power loss before VFS4G attempts to write to the placeholder list (but after it has written the placeholder)
2) There is a power loss before the placeholder list is flushed to disk

To address these issues:

1) VFS4G should write to the placeholder list before writing the placeholder
2) VFS4G should flush the file buffers for the placeholder list file

PR #880, opened 2019-03-05
Improving performance of the placeholder list database

turbonaitis:
This pull request will eventually address #828 

This first commit demonstrates, how the current implementation is performing poorly by adding a unit test, that fails.

### DIFF
--- a/GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
+++ b/GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Threading;
 
@@ -11,6 +12,7 @@ namespace GVFS.Common.FileBasedCollections
 {
     public abstract class BinaryFileBasedCollection<TEntry> : IDisposable
     {
+        public static readonly byte[] EntryTerminator = new byte[] { 0, 0, 0, 0 };
         protected const byte AddEntryPrefix = 1 << 0;
         protected const byte RemoveEntryPrefix = 1 << 1;
         private const string EtwArea = nameof(FileBasedCollection);
@@ -272,6 +274,12 @@ namespace GVFS.Common.FileBasedCollections
                                 error = string.Format("{0} is corrupt on line {1}: Invalid Prefix '{2}'", this.GetType().Name, lineCount, prefix);
                                 return false;
                             }
+
+                            byte[] terminatingBytes = reader.ReadBytes(4);
+                            if (terminatingBytes.Length != 4 || !Array.TrueForAll(terminatingBytes, b => b == 0))
+                            {
+                                error = string.Format("{0} is corrupt on line {1}: Invalid entry suffix '{2}'", this.GetType().Name, lineCount, terminatingBytes.Aggregate(string.Empty, (s, b) => s+= b + " ", s => s.TrimEnd()));
+                            }
                         }
                         catch (EndOfStreamException)
                         {
@@ -393,8 +401,7 @@ namespace GVFS.Common.FileBasedCollections
             {
                 using (BinaryWriter w = new BinaryWriter(this.dataFileHandle, Encoding.UTF8, true))
                 {
-                    w.Write(prefix);
-                    this.serializeEntry(w, element);
+                    this.WriteEntry(w, prefix, element);
                 }
 
                 this.dataFileHandle.Flush();
@@ -406,6 +413,26 @@ namespace GVFS.Common.FileBasedCollections
         /// </summary>
         private void RemoveLastEntryIfInvalid()
         {
+            if (this.dataFileHandle.Length > EntryTerminator.Length)
+            {
+                this.dataFileHandle.Seek(0, SeekOrigin.End);
+                this.dataFileHandle.Seek(EntryTerminator.Length * -1, SeekOrigin.Current);
+                byte[] buffer = new byte[4];
+                if (this.dataFileHandle.Read(buffer, 0, EntryTerminator.Length) != EntryTerminator.Length || !buffer.SequenceEqual(EntryTerminator))
+                {
+                    long lastTerminatorPosition = 0;
+                    while (this.dataFileHandle.Position - (EntryTerminator.Length + 1) > 0)
+                    {
+                        this.dataFileHandle.Seek((EntryTerminator.Length + 1) * -1, SeekOrigin.Current);
+                        if (this.dataFileHandle.Read(buffer, 0, EntryTerminator.Length) == EntryTerminator.Length && buffer.SequenceEqual(EntryTerminator))
+                        {
+                            lastTerminatorPosition = this.dataFileHandle.Position;
+                        }
+                    }
+
+                    this.dataFileHandle.SetLength(lastTerminatorPosition);
+                }
+            }
         }
 
         /// <summary>
@@ -426,8 +453,7 @@ namespace GVFS.Common.FileBasedCollections
                 {
                     foreach ((byte prefix, TEntry element) in getDataElemenets())
                     {
-                        writer.Write(prefix);
-                        this.serializeEntry(writer, element);
+                        this.WriteEntry(writer, prefix, element);
                     }
 
                     tempFile.Flush();
@@ -445,6 +471,13 @@ namespace GVFS.Common.FileBasedCollections
                 handledException = e;
                 return false;
             }
+        }
+
+        private void WriteEntry(BinaryWriter writer, byte prefix, TEntry entry)
+        {
+            writer.Write(prefix);
+            this.serializeEntry(writer, entry);
+            writer.Write(EntryTerminator);
         }
     }
 }
--- a/GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
+++ b/GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
@@ -1,0 +1,450 @@
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.Text;
+using System.Threading;
+
+namespace GVFS.Common.FileBasedCollections
+{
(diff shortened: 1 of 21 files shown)


turbonaitis:
So it turns out loading 100k placeholders takes only 836ms on WIndows, while it takes 4650ms on Mac.

turbonaitis:
After the changes it takes 143ms on Mac and 156ms on Windows to load 100k placeholder list from file.

turbonaitis:
The only remaining question now is - with Mac and Windows disk layout versions diverged, how do we bump them? Should we increment both by +1? I assume we only write migrations for windows layouts, as Mac is not yet being used?

GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
```diff
@@ -0,0 +1,450 @@
...
+                }
+
+                this.dataFileHandle.Flush();
+            }
+        }
+
+        /// <summary>
+        /// Reads entries from dataFileHandle, removing any data after the last NewLine ("\r\n"). Requires fileLock.
+        /// </summary>
+        private void RemoveLastEntryIfInvalid()
```
kewillford: Why is this method empty?  The reason for this method was to keep the file from becoming corrupted by a power failure or a BSOD.  When a write was not flushed completely to disk, zeros would be at the end of the file which would cause issues loading the list.  To make this change resilient to BSODs then it would need to have an end marker for each entry so that this code find that last good entry and "fix" the file. 

Did you do any profiling to see where the time was spent when loading the placeholders?  Perhaps this method is the offender and it was searching for the last `\r\n` when it didn't need to?  I will try and find some time to profile and see if this is the case.

One other thought is if we should use a SQLite database for this instead of a custom binary format.  We are using one for the BlobSizes and have discussed moving other "databases" to use it.  Using it of course would be a question of performance and making sure transactions get committed and the database doesn't become corrupt during power loss or BSOD.

turbonaitis: Yep, you're right, for power failure (and other critical failure) scenarios, we need to implement this method. 

I didn't do any proper profiling (couldn't find a dotTrace equivalent for Mac), but this method, implemented properly, should read just the last few bytes (EOL marker) from the end of file, so should be fast. Working with strings is not.

Yes, we could standardize on SQLite, that would be a more maintainable format. I didn't see, that it's used elsewhere and didn't want to introduce new dependencies. 

How are you dealing with a potential corruption of the SQLite, that's being used for BlobSizes?

kewillford: @wilbaker Did the work around the BlobSizes which can be found [here](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Virtualization/BlobSize/BlobSizes.cs).  The main difference with BlobSizes is that if the write doesn't make it to the database it will download the information again and try to write again.  The placeholders database should be more fault tolerant although we are losing a placeholder right now with the current implementation in critical failure scenarios so we would be no worse off than today.  We haven't had any cases where the SQLite database itself becomes corrupt so it seems very resilient to failure.  It has [extensive testing](https://sqlite.org/testing.html) around these scenarios which is good.

wilbaker: > The placeholders database should be more fault tolerant although we are losing a placeholder right now with the current implementation in critical failure scenarios so we would be no worse off than today.

What @kewillford mentions here is key.  

In the case of the BlobSizes database we don't *really* care if entries never make it to the database, as that database is just a local cache of file sizes.  If writes are not completely flushed we can always request the file sizes again (this is what allows us to make the performance optimization of writing sizes in batches in a background thread).  

If we decide to use SQLite for the placeholders we'll need to be more strict.  From the perspective of the thread that's adding a placeholder, the writes need to be synchronous and flushed to disk before the thread returns from its callback.

> we are losing a placeholder right now with the current implementation in critical failure scenarios so we would be no worse off than today.

Please also note that the issue @kewillford mentioned here can be fixed by writing to the database before writing the placeholder (#442 is open to fix this).  

We should make sure that whatever option we go with for improving perf is at least as reliable as what we have today and does not prevent us from fixing the power-loss scenario mentioned above.

turbonaitis: I've implemented RemoveLastEntryIfInvalid method for BinaryFileBasedCollection. Unfortunately, trying it out, whether SQLite would perform adequately, is a bit harder than, I originally thought and I don't have the time to do this now. 

Are you guys open to merging this now, to benefit from the performance gains and then perhaps standardise on SQLite later? If so, we need to discuss disk layout migrations.

turbonaitis:
> @turbonaitis: I started reviewing this last week and got side-tracked. One thing that is difficult to know is how much is re-used code from the previous collection and how much is new. Is there any way you can share some code across the two collections?

I've compared the two files and it seems refactoring for them to be able to share a lot of code shouldn't be hard - I'll try to do that.

GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+        public void ForceFlush()
+        {
+            if (this.dataFileHandle != null)
+            {
+                FileStream fs = this.dataFileHandle as FileStream;
+                if (fs != null)
+                {
+                    fs.Flush(flushToDisk: true);
```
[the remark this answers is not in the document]
turbonaitis: This is a copy from FileBasedCollection.

GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
```diff
@@ -0,0 +1,483 @@
...
+
+                    bool tmpFileCreated = false;
+                    int tmpFileCreateAttempts = 0;
+
+                    bool tmpFileMoved = false;
+                    int tmpFileMoveAttempts = 0;
+
+                    Exception lastException = null;
+
+                    while (!tmpFileCreated || !tmpFileMoved)
```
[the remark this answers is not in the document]
turbonaitis: This whole method is a almost an exact copy from FileBasedCollection.

turbonaitis:
@derrickstolee I've refactored the code, so that both old and new database classes inherrit from the same base class. Please have another look.

GVFS/GVFS.Common/FileBasedCollections/BinaryFileBasedCollection.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+
+namespace GVFS.Common.FileBasedCollections
+{
+    public abstract class BinaryFileBasedCollection<TEntry> : GenericFileBasedCollection<TEntry, Tuple<byte, TEntry>>
```
derrickstolee: I almost always prefer using a custom `struct` over `Tuple<>` since we can provide meaningful names to the entries instead of `Item1`, `Item2`. This is particularly troubling inside BinaryPlaceholderListDatabase.GenerateDataLines()`.

Best to make a `readonly` struct:

```
public struct PrefixAndEvent
{
    public readonly byte Prefix;
    public readonly PlaceholderEvent PlaceholderEvent;

    public PrefixAndEvent(byte prefix, PlaceholderEvent event)
    {
        this.Prefix = prefix;
        this.PlaceholderEvent = event;
    }
}
```

GVFS/GVFS.Common/FileBasedCollections/BinaryPlaceholderListDatabase.cs
```diff
@@ -0,0 +1,358 @@
...
+                        }
+                    });
+            }
+            catch (Exception e)
+            {
+                throw new FileBasedCollectionException(e);
+            }
+        }
+
+        private bool TryParseAddLine(BinaryReader reader, out string key, out PlaceholderEvent value, out string error)
```
derrickstolee: This ties us tightly to the underlying format. Is this not something we can put into the collection so we don't need to be aware of the format here?

GVFS/GVFS.Common/FileBasedCollections/BinaryPlaceholderListDatabase.cs
```diff
@@ -0,0 +1,358 @@
...
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Common.FileBasedCollections
+{
+    public class BinaryPlaceholderListDatabase : BinaryFileBasedCollection<PlaceholderEvent>
```
derrickstolee: I compared this file to `PlaceholderListDatabase.cs` and found the changes mostly related to the new type structure, with a little bit of logic around the "remove" events.

While the changes looked appropriate, I can't shake a feeling that this "copy and update types" strategy is smelly and likely to include bugs. However, sometimes we need a big change because the old code is old and poorly typed.

The best thing I could think to do is to create an interface between the placeholder list database logic and the file based collection logic, and that can use the new type structure around `PlaceholderEvent`. Then, create an adapter class that "upgrades" the old `FileBasedCollection` to create the new types. Then, we don't need to have two classes that are mostly duplicates here, and don't need to `[Obsolete]` the other class.

One obstruction I see to this plan is that we have some coupling in this class to the underlying file format. That's already something I'd like to remedy, and that would make the classes more independent. In general, I prefer to have only one class know about a file format, and everything else be independent.

kewillford:
@turbonaitis I am currently looking into sqlite and doing some performance and memory testing on a spike of using that.  I also queued a performance build on your changes here as well to see how it compares to sqlite.

jrbriggs:
I'm closing this PR because #1107 starts the conversion of the placeholder database to SQLite. In local testing we've found this has better read performance than the improvements in this PR as well as significantly better write performance. Thanks for pointing us in this direction @turbonaitis

turbonaitis:
@derrickstolee thanks for your thorough review! All of your suggestions, if implemented, would have made this code better! I'm also glad you're pointing out the value of programming to explicitly defined interfaces, something that was lacking before. 

@jrbriggs looking forward to see the SQLite version implemented!

PR #884, opened 2019-03-05
Adding Distribution.xml & installation-check

### LINKED ISSUE #881: Mac: VFS4G Installer should validate that all VFS4G related processes aren't running before installation
While testing a new build for early adopters, I noticed that I was allowed to have `GVFS.Mount` processes running while upgrading my version of VFS4G. While this caused some other issues down the line, the bug here seems to be allowing the upgrade to proceed while we have providers alive. 

On Windows, we ban a whole set of processes related to G4W, but I think we can simply look for these processes on macOS for now:

- GVFS.Mount
- git
- gvfs

We should update the installer to test for these running and present the user with a helpful error message on what they need to fix to get unblocked.

alameenshah:
On the Mac, the GVFS standalone installer and the combined GVFS & Git meta installer, would check if any GVFS of git processes are running. If they are running, then installer would display error messaging and will not continue.

1. Added [Distribution XML](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html) for both GVFS standalone and combined meta installers.
2. Updated `GVFS.Installer.Mac::CreateInstaller` script to create distributions for both GVFS Standalone and combined meta installers.
3. Created a new `Packages` staging directory where component packages are kept. (The distributable installers is kept in the same location as before.)
4. Updated `preflight` installer script to automatically unload `PrjFlt.kext` if it is loaded at install time.
5. Installation pre-check will block on these processes - `"gvfs", "gvfs.mount", "git", "gitk", "wish"`

### DIFF
--- a/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
+++ b/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
@@ -18,7 +18,7 @@ if [ -z $PACKAGEVERSION ]; then
   exit 1
 fi
 
-BUILDOUTPUTDIR=$4
+BUILDOUTPUTDIR=${4%/}
 if [ -z $BUILDOUTPUTDIR ]; then
   echo "Error: Build output directory not specified"
   exit 1
@@ -34,7 +34,8 @@ if [ -z $VFS_PUBLISHDIR ]; then
   exit 1
 fi
 
-STAGINGDIR=$BUILDOUTPUTDIR"Staging"
+STAGINGDIR="${BUILDOUTPUTDIR}/Staging"
+PACKAGESTAGINGDIR="${BUILDOUTPUTDIR}/Packages"
 VFSFORGITDESTINATION="usr/local/vfsforgit"
 DAEMONPLISTDESTINATION="Library/LaunchDaemons"
 LIBRARYEXTENSIONSDESTINATION="Library/Extensions"
@@ -42,6 +43,7 @@ INSTALLERPACKAGENAME="VFSForGit.$PACKAGEVERSION"
 INSTALLERPACKAGEID="com.vfsforgit.pkg"
 UNINSTALLERPATH="${SOURCEDIRECTORY}/uninstall_vfsforgit.sh"
 SCRIPTSPATH="${SOURCEDIRECTORY}/scripts"
+DIST_FILE_NAME="Distribution.updated.xml"
 
 function CheckBuildIsAvailable()
 {
@@ -61,6 +63,9 @@ function CreateInstallerRoot()
 {
     mkdirVfsForGit="mkdir -p \"${STAGINGDIR}/$VFSFORGITDESTINATION\""
     eval $mkdirVfsForGit || exit 1
+    
+    mkdirPkgStaging="mkdir -p \"${PACKAGESTAGINGDIR}\""
+    eval $mkdirPkgStaging || exit 1
 
     mkdirBin="mkdir -p \"${STAGINGDIR}/usr/local/bin\""
     eval $mkdirBin || exit 1
@@ -98,7 +103,7 @@ function CopyBinariesToInstall()
     copyPrjFS="cp -Rf \"${VFS_OUTPUTDIR}/ProjFS.Mac/Native/$CONFIGURATION\"/PrjFSKext.kext \"${STAGINGDIR}/${LIBRARYEXTENSIONSDESTINATION}/.\""
     eval $copyPrjFS || exit 1
     
-    copyPrjFS="cp -Rf \"${VFS_OUTPUTDIR}/ProjFS.Mac/Native/$CONFIGURATION\"/org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist \"${STAGINGDIR}/${DAEMONPLISTDESTINATION}/.\""
+    copyPrjFS="cp -Rf \"${VFS_OUTPUTDIR}/ProjFS.Mac/Native/$CONFIGURATION/org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist\" \"${STAGINGDIR}/${DAEMONPLISTDESTINATION}/.\""
     eval $copyPrjFS || exit 1
     
     currentDirectory=`pwd`
@@ -108,19 +113,59 @@ function CopyBinariesToInstall()
     cd $currentDirectory
 }
 
-function CreateInstaller()
+function CreateVFSForGitInstaller()
 {
-    pkgBuildCommand="/usr/bin/pkgbuild --identifier $INSTALLERPACKAGEID --scripts \"${SCRIPTSPATH}\" --root \"${STAGINGDIR}\" \"${BUILDOUTPUTDIR}\"$INSTALLERPACKAGENAME.pkg"
+    pkgBuildCommand="/usr/bin/pkgbuild --identifier $INSTALLERPACKAGEID --scripts \"${SCRIPTSPATH}\" --root \"${STAGINGDIR}\" \"${PACKAGESTAGINGDIR}/$INSTALLERPACKAGENAME.pkg\""
     eval $pkgBuildCommand || exit 1
 }
 
-function CreateMetaInstaller()
+function UpdateDistributionFile()
+{
+    VFSFORGIT_PKG_VERSION=$PACKAGEVERSION
+    VFSFORGIT_PKG_NAME="$INSTALLERPACKAGENAME.pkg"
+    GIT_PKG_NAME=$1
+    GIT_PKG_VERSION=$2
+        
+    /usr/bin/sed -e "s|VFSFORGIT_VERSION_PLACHOLDER|$VFSFORGIT_PKG_VERSION|g" "$SCRIPTSPATH/Distribution.xml" > "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
+    /usr/bin/sed -i.bak "s|VFSFORGIT_PKG_NAME_PLACEHOLDER|$VFSFORGIT_PKG_NAME|g" "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
+    
+    if [ ! -z "$GIT_PKG_NAME" ] && [ ! -z "$GIT_PKG_VERSION" ]; then
+        GIT_CHOICE_OUTLINE_ELEMENT_TEXT="<line choice=\"com.git.pkg\"/>"
+        GIT_CHOICE_ID_ELEMENT_TEXT="<choice id=\"com.git.pkg\" visible=\"false\"> <pkg-ref id=\"com.git.pkg\"/> </choice>"
+        GIT_PKG_REF_ELEMENT_TEXT="<pkg-ref id=\"com.git.pkg\" version=\"$GIT_PKG_VERSION\" onConclusion=\"none\">$GIT_PKG_NAME</pkg-ref>"
+    else
+        GIT_CHOICE_OUTLINE_ELEMENT_TEXT=""
+        GIT_CHOICE_ID_ELEMENT_TEXT=""
+        GIT_PKG_REF_ELEMENT_TEXT=""
+    fi
+    
+    /usr/bin/sed -i.bak "s|GIT_CHOICE_OUTLINE_PLACEHOLDER|$GIT_CHOICE_OUTLINE_ELEMENT_TEXT|g" "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
+    /usr/bin/sed -i.bak "s|GIT_CHOICE_ID_PLACEHOLDER|$GIT_CHOICE_ID_ELEMENT_TEXT|g" "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
+    /usr/bin/sed -i.bak "s|GIT_PKG_REF_PLACEHOLDER|$GIT_PKG_REF_ELEMENT_TEXT|g" "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
+    
+    /bin/rm -f "${BUILDOUTPUTDIR}/$DIST_FILE_NAME.bak"
+}
+
+function CreateVFSForGitDistribution()
+{
+    # Update distribution file(removes Git info from template.)
+    UpdateDistributionFile "" ""
+    
+    buildVFSForGitDistCmd="/usr/bin/productbuild --distribution \"${BUILDOUTPUTDIR}/Distribution.updated.xml\" --package-path \"$PACKAGESTAGINGDIR\" \"${BUILDOUTPUTDIR}/$INSTALLERPACKAGENAME.pkg\""
+    echo $buildVFSForGitDistCmd
+    eval $buildVFSForGitDistCmd || exit 1
+    
+    /bin/rm -f "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
+}
+
+function CreateMetaDistribution()
 {
     GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
     GITDMGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.dmg)" || exit 1
     GITDMGNAME="${GITDMGPATH##*/}"
     GITINSTALLERNAME="${GITDMGNAME%.dmg}"
     GITVERSIONSTRING=`echo $GITINSTALLERNAME | cut -d"-" -f2`
+    GITPKGNAME="$GITINSTALLERNAME.pkg"
     
     if [[ -z "$GITVERSION" || -z "$GITVERSIONSTRING" ]]; then
         echo "Error creating metapackage: could not determine Git package version."
@@ -144,11 +189,17 @@ function CreateMetaInstaller()
         exit 1
     fi
     
-    METAPACKAGENAME="$INSTALLERPACKAGENAME-Git.$GITVERSION.pkg"
+    copyGitPkgCmd="/bin/cp -Rf \"${GITINSTALLERPATH}\" \"${PACKAGESTAGINGDIR}/.\""
+    eval $copyGitPkgCmd
+
+    UpdateDistributionFile "$GITPKGNAME" "$GITVERSIONSTRING"
     
-    buildMetapkgCmd="/usr/bin/productbuild --package \"$GITINSTALLERPATH\" --package \"${BUILDOUTPUTDIR}\"$INSTALLERPACKAGENAME.pkg \"${BUILDOUTPUTDIR}\"$METAPACKAGENAME"
+    METAPACKAGENAME="$INSTALLERPACKAGENAME-Git.$GITVERSION.pkg"
+    buildMetapkgCmd="/usr/bin/productbuild --distribution \"${BUILDOUTPUTDIR}/Distribution.updated.xml\" --package-path \"$PACKAGESTAGINGDIR\" \"${BUILDOUTPUTDIR}/$METAPACKAGENAME\""
     echo $buildMetapkgCmd
     eval $buildMetapkgCmd || exit 1
+    
+    /bin/rm -f "${BUILDOUTPUTDIR}/$DIST_FILE_NAME"
     
     unmountCmd="/usr/bin/hdiutil detach \"$MOUNTEDVOLUME\""
     echo "$unmountCmd"
@@ -161,8 +212,9 @@ function Run()
     CreateInstallerRoot
     CopyBinariesToInstall
     SetPermissions
-    CreateInstaller
-    CreateMetaInstaller
+    CreateVFSForGitInstaller
+    CreateVFSForGitDistribution
+    CreateMetaDistribution
 }
 
 Run
--- a/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
+++ b/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<installer-gui-script minSpecVersion="1">
+    <pkg-ref id="com.vfsforgit.pkg"/>
+    <pkg-ref id="com.git.pkg"/>
+    <options customize="never" allow-external-scripts="true"/>
+    <choices-outline>
+        <line choice="default">
+            <line choice="com.vfsforgit.pkg"/>
+            GIT_CHOICE_OUTLINE_PLACEHOLDER
+        </line>
+    </choices-outline>
+    <choice id="default"/>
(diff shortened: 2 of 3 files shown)


GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -35,6 +35,7 @@ if [ -z $VFS_PUBLISHDIR ]; then
 fi
 
 STAGINGDIR=$BUILDOUTPUTDIR"Staging"
+PACKAGESTAGINGDIR="${BUILDOUTPUTDIR}Packages"
```
wilbaker: Should this have a `/` in it after `${BUILDOUTPUTDIR}`? --> `"${BUILDOUTPUTDIR}/Packages"`

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -110,10 +114,17 @@ function CopyBinariesToInstall()
 
 function CreateInstaller()
 {
-    pkgBuildCommand="/usr/bin/pkgbuild --identifier $INSTALLERPACKAGEID --scripts \"${SCRIPTSPATH}\" --root \"${STAGINGDIR}\" \"${BUILDOUTPUTDIR}\"$INSTALLERPACKAGENAME.pkg"
+    pkgBuildCommand="/usr/bin/pkgbuild --identifier $INSTALLERPACKAGEID --scripts \"${SCRIPTSPATH}\" --root \"${STAGINGDIR}\" \"${PACKAGESTAGINGDIR}/$INSTALLERPACKAGENAME.pkg\""
     eval $pkgBuildCommand || exit 1
 }
 
+function CreateGVFSDistribution()
```
wilbaker: Naming suggestion -> `CreateGVFSDistribution` -> `CreateVFSForGitDistribution`

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -110,10 +114,17 @@ function CopyBinariesToInstall()
...
+{
+    buildGVFSDistCmd="/usr/bin/productbuild --distribution \"$SCRIPTSPATH/Distribution_GVFS.xml\" --package-path \"$PACKAGESTAGINGDIR\" \"${BUILDOUTPUTDIR}$INSTALLERPACKAGENAME.pkg\""
```
wilbaker: Another spot: `buildGVFSDistCmd` -> `buildVFSForGitDistCmd`

(Mentioning these because from what I've seen we're using "VFSForGit" naming scheme in this file)

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
```diff
@@ -0,0 +1,51 @@
...
+            }
+            
+            my.result.message = "No blocking processes running.";
+            my.result.type = "Info";
+            return true;
+        }
+        
+        function GetBlockingProcesses()
+        {
+            var watchList = [ "Calculator", "TextEdit", "Preview" ];
```
wilbaker: Will these be updated to the VFS4G/Git process names we care about?

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
```diff
@@ -0,0 +1,51 @@
...
+            var blockingProcesses = new Array();
+            for (var process of watchList)
+            {
+                var isRunningCmd = "[ `ps -ef | grep -w '" + process + "' | grep -v grep | wc -l` -eq 0 ]";
+                // var isRunningCmd = "exit 1;";
```
wilbaker: Remove this commented out line?

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
```diff
@@ -0,0 +1,51 @@
...
+                var status = system.run("/bin/bash", "-c", isRunningCmd);
+                system.log("isRunningCmd: " + isRunningCmd + ". Status: " + status);
```
wilbaker: Where is `system.log` written?

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
```diff
@@ -0,0 +1,51 @@
...
+    <choices-outline>
+        <line choice="default">
+            <line choice="com.vfsforgit.pkg"/>
+        </line>
+    </choices-outline>
+    <choice id="default"/>
+    <choice id="com.vfsforgit.pkg" visible="false">
+        <pkg-ref id="com.vfsforgit.pkg"/>
+    </choice>
+    <pkg-ref id="com.vfsforgit.pkg" version="1.2.3" onConclusion="none">VFSFORGITPACKAGEFILENAME</pkg-ref>
```
wilbaker: `version="1.2.3"` -> Where did this version number come from?

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
```
wilbaker: Rename file `Distribution_GVFS.xml` -> `Distribution_VFSForGit.xml`

GVFS/GVFS.Installer.Mac/scripts/Distribution_Meta.xml
```diff
@@ -0,0 +1,57 @@
...
+                my.result.type = 'Fatal';
+                return false;
+            }
+            
+            my.result.message = "No blocking processes running.";
+            my.result.type = "Info";
+            return true;
+        }
+        
+        function GetBlockingProcesses()
```
wilbaker: Same comments for `GetBlockingProcesses` as left in `Distribution_GVFS.xml`

GVFS/GVFS.Installer.Mac/scripts/Distribution_Meta.xml
```diff
@@ -0,0 +1,57 @@
...
+    <choice id="default"/>
+    <choice id="com.vfsforgit.pkg" visible="false">
+        <pkg-ref id="com.vfsforgit.pkg"/>
+    </choice>
+    <pkg-ref id="com.vfsforgit.pkg" version="0" onConclusion="none">VFSForGit.0.2.173.2.pkg</pkg-ref>
+    <choice id="com.git.pkg" visible="false">
+        <pkg-ref id="com.git.pkg"/>
+    </choice>
+    <pkg-ref id="com.git.pkg" version="2.20.1.vfs.1.1.102.gdb3f8ae2d1" onConclusion="none">git-2.20.1.vfs.1.1.102.gdb3f8ae2d1-intel-universal-snow-leopard.pkg</pkg-ref>
+    <installation-check script="InstallationCheck()"/>
```
wilbaker: How does this XML file work?  Is `InstallationCheck()` a special well known function?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
alameenshah: `${BUILDOUTPUTDIR}` already has the trailing `/`. I can remove it if it looks confusing.

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
alameenshah: absolutely. updated to `"gvfs", "gvfs.mount", "git", "gitk"`

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
alameenshah: Consolidated the two separate `xml`s into one. The new one has placeholders for `VFSForGit` and `Git` package info. The placeholders are substituted with actual values (like `VFSForGit` version, package name etc) during build time.

GVFS/GVFS.Installer.Mac/scripts/Distribution_Meta.xml
alameenshah: How this XML file is used by the tools during packaging is not very well documented, there is a reference for its scheme [here](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html). In our case, I used `productbuild --synthesize` command which reads constituent packages info and automatically creates Distribution.xml. I manually then replaced version numbers and package names with placeholders. Our build script will substitute the place holders with actual dynamic values during build time.

About `InstallationCheck()` - any javascript function can be specified in `<installation-check>`, naming is not important. [This page](https://matthew-brett.github.io/docosx/flat_packages.html) has some info on sample `InstallationCheck()` implementation.

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
alameenshah: That shows up in the Installer logs. Like this - 
 
screen shot 2019-03-06 at 11 38 23 am

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
alameenshah: that was just a placeholder. Updated the build script to update `Distribution.xml` with actual values computed during build time.

GVFS/GVFS.Installer.Mac/scripts/Distribution_GVFS.xml
wilbaker: Thanks, is that log also saved to disk somewhere?  Or is it only present while the installer is running?

alameenshah: User can choose `File->Save` from the installer menu and save it to disk.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
wilbaker: Ahh I missed that.  Related question, why does `STAGINGDIR=$BUILDOUTPUTDIR"Staging"` not use the `{ }` braces like `PACKAGESTAGINGDIR="${BUILDOUTPUTDIR}Packages"` does?

Can we make these consistent (I am not familiar with bash script and so I'm not sure which is preferable)?

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
```diff
@@ -0,0 +1,51 @@
...
+                my.result.type = 'Fatal';
+                return false;
+            }
+            
+            return true;
+        }
+        
+        function GetBlockingProcesses()
+        {
+            var watchList = [ "gvfs", "gvfs.mount", "git", "gitk" ];
```
wilbaker: Have you tested this with `gitk`?  I believe the application name is actually "wish" and not "gitk".

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -108,13 +112,34 @@ function CopyBinariesToInstall()
...
+	/usr/bin/sed -i.bak "s|VFSFORGIT_PKG_NAME_PLACEHOLDER|$INSTALLERPACKAGENAME.pkg|g" "${BUILDOUTPUTDIR}/Distribution.updated.xml"
+	
+	/bin/rm -f "${BUILDOUTPUTDIR}/Distribution.updated.xml.bak"
+	
+    buildVFSForGitDistCmd="/usr/bin/productbuild --distribution \"${BUILDOUTPUTDIR}/Distribution.updated.xml\" --package-path \"$PACKAGESTAGINGDIR\" \"${BUILDOUTPUTDIR}/$INSTALLERPACKAGENAME.pkg\""
+    echo $buildVFSForGitDistCmd
+    eval $buildVFSForGitDistCmd || exit 1
+    
+    /bin/rm -f "${BUILDOUTPUTDIR}/Distribution.updated.xml"
+}
```
kewillford: This method looks like it shares a lot of code from the `CreateMetaDistribution` method below.  Can we have a helper method that takes some parameters and shares most of this code?

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
alameenshah: You are right `gitk` does launch `Wish`, will add it. (Our check still succeeds because it finds `gitk` as well, when running `ps` command.)
```
ps -ax | grep gitk
29744 ttys004    0:00.01 grep gitk
29673 ttys005    0:00.01 /bin/sh /usr/bin/wish /usr/local/git/bin/gitk --
29675 ttys005    0:02.82 /usr/bin/../../System/Library/Frameworks/Tk.framework/Versions/8.5/Resources/Wish.app/Contents/MacOS/Wish /usr/local/git/bin/gitk --```

File: GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh (hunk 1 above)
Comment (alameenshah) [reply GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh]: have clean it up a little bit, removed the trailing `/` in `BUILDOUTPUTDIR` and other inconsistencies.

File: GVFS/GVFS.Installer.Mac/scripts/Distribution.xml (hunk 11 above)
Comment (alameenshah) [reply GVFS/GVFS.Installer.Mac/scripts/Distribution.xml]: Added `wish`

File: GVFS/GVFS.Installer.Mac/scripts/preinstall [hunk 13]
```diff
@@ -4,3 +4,12 @@ if [ -f /Library/LaunchDaemons/org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist ]; t
     echo "Unloading PrjFSKextLogDaemon: '$unloadCmd'..."
     eval $unloadCmd || exit 1
 fi
+
+KEXTBUNDLEID="org.vfsforgit.PrjFSKext"
+isKextLoadedCmd="/usr/sbin/kextstat -l -b $KEXTBUNDLEID | wc -l"
+isKextLoaded=$(eval $isKextLoadedCmd)
+if [ "$isKextLoaded" -gt 0 ]; then
+	unloadCmd="sudo /sbin/kextunload -b $KEXTBUNDLEID"
```
kewillford: Where will the prompt for the sudo password be when this is ran?

alameenshah: Installer only runs elevated - so it wouldn't need to display the prompt.

pmj: Why bother with `sudo` at all?

pmj: Also, we should probably check for the legacy kext identifier here too, at least for now?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -144,12 +169,33 @@ function CreateMetaInstaller()
...
+
+	# 1.  Update Distribution.xml
+	# 1.1 Substitute Git placeholders with actual values
+    /usr/bin/sed -e "s|GIT_CHOICE_OUTLINE_PLACEHOLDER|$GIT_CHOICE_OUTLINE_ELEMENT_TEXT|g" "$SCRIPTSPATH/Distribution.xml" > "${BUILDOUTPUTDIR}/Distribution.updated.xml"
+	/usr/bin/sed -i.bak "s|GIT_CHOICE_ID_PLACEHOLDER|$GIT_CHOICE_ID_ELEMENT_TEXT|g" "${BUILDOUTPUTDIR}/Distribution.updated.xml"
+	/usr/bin/sed -i.bak "s|GIT_PKG_REF_PLACEHOLDER|$GIT_PKG_REF_ELEMENT_TEXT|g" "${BUILDOUTPUTDIR}/Distribution.updated.xml"
+	
+	# 1.2 Substitute VFSForGit placeholders with actual values
+	/usr/bin/sed -i.bak "s|VFSFORGIT_VERSION_PLACHOLDER|$PACKAGEVERSION|g" "${BUILDOUTPUTDIR}/Distribution.updated.xml"
+	/usr/bin/sed -i.bak "s|VFSFORGIT_PKG_NAME_PLACEHOLDER|$INSTALLERPACKAGENAME.pkg|g" "${BUILDOUTPUTDIR}/Distribution.updated.xml"
```
pmj: Generating templated XML this way seems like it's taking shell scripting a bit too far if you ask me, but I guess I'm not the one maintaining this script. 🙃

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
```diff
@@ -0,0 +1,51 @@
...
+            return true;
+        }
+        
+        function GetBlockingProcesses()
+        {
+            var watchList = [ "gvfs", "GVFS.Mount", "git", "gitk", "Wish" ];
+            var blockingProcesses = new Array();
+            for (var process of watchList)
+            {
+                var isRunningCmd = "[ `ps -ef | grep -w '" + process + "' | grep -v grep | wc -l` -eq 0 ]";
```
pmj: Would the output of `ps -Ac -o command` not be easier/less error-prone to parse than `ps -ef`? Also, `grep`'s exit code is 0 if one or more lines were matched, and 1 if not, so you shouldn't need to use `wc`.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
nickgra: Do you have a better suggestion for handling the version numbers here? I also am a little scared of this, but I don't have any other ideas off the top of my head.

GVFS/GVFS.Installer.Mac/scripts/preinstall
nickgra: Since we have so few early adopter customers, I'd rather tell them just to fully uninstall VFS4G and then reinstall before shipping our next release rather than bake in the backwards compatibility.

Thoughts?
cc: @wilbaker @jrbriggs

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
pmj: As far as I'm aware, `xsltproc` ships with macOS and can be used for some basic XML processing. There's also scripting languages like python, ruby, and perl where the xml library packages are probably also included.

Or maybe put together a very small .NET program to generate the XML? That would probably fit best with the rest of the project so far… (Is JScript still a thing?)

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
alameenshah: `ps -Ac -o command` is definitely better. Keeping `wc` - without it, looks like the following `-eq 0` check uses the output of `grep -v grep` which will be always `0` and breaks our logic.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
alameenshah: removed the duplication.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
alameenshah: Have removed some duplicate code - it is slightly easier to follow now.

GVFS/GVFS.Installer.Mac/scripts/preinstall
alameenshah: Removed `sudo`.

wilbaker: > I'd rather tell them just to fully uninstall VFS4G and then reinstall

If we are going to hold the next release to EA for this PR then I agree with Phil's suggestion.  All the EA users currently have the legacy kext identifier, and so for these changes to work correctly for those users we'll need to check the legacy kext identifier.  

I agree we don't need this long term though, and so we could create a work item (or even a PR) to remove it for the next release.

cc: @jrbriggs for his thoughts

jrbriggs: +1 for Phil's suggestion and I agree with the follow-up work to remove it.

alameenshah: Added the check for legacy kext as well. Created [work item](https://github.com/Microsoft/VFSForGit/issues/891) to remove this check in future.

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
pmj: I don’t think you need the `grep -v grep` step anymore as the command arguments aren’t part of the output when using `-c`.

alameenshah: That works, Thank you!

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
nickgra: This is much smaller now so I'm happy with this relatively cheap solution versus introducing another scripting language or writing a mini .NET program. If this gets more complicated, I think that's probably the way we should go though.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -18,7 +18,7 @@ if [ -z $PACKAGEVERSION ]; then
   exit 1
 fi
 
-BUILDOUTPUTDIR=$4
+BUILDOUTPUTDIR=${4%/}
```
wilbaker: Why do we need the `/}` for `BUILDOUTPUTDIR` but not for `SOURCEDIRECTORY`?

alameenshah: `GVFS.Installer.Mac.csproj` uses build variable `OutputPath` to set `BUILDOUTPUTDIR`. `OutputPath` is originally set elsewhere, I think in `GVFS.cs.pros` line no. 7 with a trailing `/`.

In other places that use `OutputPath`, like in `GVFS.Installer.Windows.csproj` line no. 35, line no. 51, the same pattern exists.

wilbaker: It seems error prone to have scripts expect directory path parameters in different formats.  I've filed #892 to clean this up.

cc: @jrbriggs for his thoughts on if this cleanup should happen in this PR or if it can wait.

jrbriggs: I think we can defer that cleanup.

PR #885, opened 2019-03-06
Updates for new ProjFS API

jeschu1:
**Several updates to support the changes to the new API**
- The new api is released under the package Microsoft.Windows.ProjFS.1.0.19079.1, which just contains the managed API
- GVFS.ProjFS.2018.823.2 has been released with the native API.  Previously this package was released with both native and managed libraries.  Since we now have a dedicated feed for managed, that portion has been removed.


GVFS/GVFS.Platform.Windows/GVFS.Platform.Windows.csproj
```diff
@@ -45,8 +45,8 @@
       <HintPath>..\..\..\packages\Microsoft.Database.Isam.1.9.4\lib\net40\Esent.Isam.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="ProjectedFSLib.Managed, Version=10.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=AMD64">
-      <HintPath>$(PackagesDir)\$(ProjFSPackage)\lib\x64\ProjectedFSLib.Managed.dll</HintPath>
+    <Reference Include="ProjectedFSLib.Managed, Version=1.0.118.0, Culture=neutral, processorArchitecture=AMD64">
+      <HintPath>..\..\..\packages\GVFS.ProjFS.2019.211.1-pr\lib\x64\ProjectedFSLib.Managed.dll</HintPath>
```
jeschu1: @wilbaker note the hint path difference here.  The surrounding assemblies follow the "..\..\.." convention so I wanted to double check what we'd prefer here.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
```diff
@@ -174,6 +180,30 @@ public void WaitForCreateWriteBuffer()
             this.waitForCreateWriteBuffer.WaitOne();
         }
 
+        public HResult CompleteCommand(int commandId, NotificationType newNotificationMask)
```
jeschu1: Not for myself check when each Complete Command is called.

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -134,29 +139,26 @@ private HResult EndDirectoryEnumeration(Guid enumerationId)
...
-                    if (result == HResult.Ok)
+                    if (addResult == true)
                     {
                         entryAdded = true;
                         activeEnumeration.MoveNext();
                     }
                     else
                     {
-                        if (result == HResult.InsufficientBuffer)
+                        if (entryAdded)
```
jeschu1: @wilbaker we no longer have the option of checking for InsufficentBuffer.

Can you advise on this change?

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -504,46 +507,35 @@ private HResult EndDirectoryEnumerationHandler(Guid enumerationId)
...
                         changeTime: properties.LastWriteTimeUTC);
 
-                    if (result == HResult.Ok)
+                    if (addResult == true)
                     {
                         entryAdded = true;
                         activeEnumeration.MoveNext();
                     }
-                    else if (result == HResult.InsufficientBuffer)
+                    else
```
jeschu1: @wilbaker @cgallred was this the one we discussed.  I.e. it only returns false on InsufficentBuffer?

jeschu1:
The functional tests and unit tests are green whipee!

@wilbaker this is ready for a first pass.  I'm sure there will be lots of feedback. 
@cgallred I turned off 2 tests that I can turn back on once we have a version with IWriteBuffer

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
wilbaker: Yep there was a slight change in behavior here.  The code can be simplified to:

```
                if (results.Add(...))
                {
                    entryAdded = true;
                    enumeration.MoveNext();
                }
                else
                {
                    if (entryAdded)
                    {
                        hr = HResult.Ok;
                    }
                    else
                    {
                        hr = HResult.InsufficientBuffer;
                    }

                    break;
                }
```

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
wilbaker: We should be able to simplify this to:

```
                if (results.Add(...))
                {
                    entryAdded = true;
                    enumeration.MoveNext();
                }
                else
                {
                    if (entryAdded)
                    {
                        hr = HResult.Ok;
                    }
                    else
                    {
                        hr = HResult.InsufficientBuffer;
                    }

                    break;
                }
```

wilbaker: More details on the return value for `Add`:

```
    /// <returns>
    ///     <para>
    ///     <c>true</c> if the entry was successfully added to the enumeration buffer, <c>false</c> otherwise.
    ///     </para>
    ///     <para>
    ///     If this method returns <c>false</c>, the provider returns <see cref="HResult::Ok"/> and waits for
    ///     the next <c>GetDirectoryEnumerationCallback</c>.  Then it resumes filling the enumeration with
    ///     the entry it was trying to add when it got <c>false</c>. 
    ///     </para>
    ///     <para>
    ///     If the function returns <c>false</c> for the first file or directory in the enumeration, the
    ///     provider returns <see cref="HResult::InsufficientBuffer"/> from the <c>GetDirectoryEnumerationCallback</c>
    ///     method.
    ///     </para>
    /// </returns>
```

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
jeschu1: Small but i simplified the 'if' to 
result = entryAdded ? HResult.Ok : HResult.InsufficientBuffer;
                        break;

GVFS/ProjectedFSLib.NativeBinaries.props
```diff
@@ -1,10 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <ItemGroup>
-        <None Condition="Exists('$(PackagesDir)\$(ProjFSPackage)\lib\x64\ProjectedFSLib.dll')" Include="$(PackagesDir)\$(ProjFSPackage)\lib\x64\ProjectedFSLib.dll">
+        <None Condition="Exists('$(PackagesDir)\$(ProjFSNativePackage)\lib\x64\ProjectedFSLib.dll')" Include="$(PackagesDir)\$(ProjFSNativePackage)\lib\x64\ProjectedFSLib.dll">
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </None>
-        <None Condition="Exists('$(PackagesDir)\$(ProjFSPackage)\lib\x64\ProjectedFSLib.pdb')" Include="$(PackagesDir)\$(ProjFSPackage)\lib\x64\ProjectedFSLib.pdb">
+        <None Condition="Exists('C:\Users\username\gvfs3\packages\GVFS.ProjFS.2018.823.2\lib\x64\ProjectedFSLib.dll')" Include="C:\Users\username\gvfs3\packages\GVFS.ProjFS.2018.823.2\lib\x64\ProjectedFSLib.dll">
```
jeschu1: remove these lines

jeschu1:
@wilbaker it's all building now (wahoo!).  

Ready for review.  I'd recommend grabbing a coffee and scheduling a relaxing time shot.  It's a long read.

GVFS/GVFS.Platform.Windows/ActiveEnumeration.cs
```diff
@@ -125,7 +125,7 @@ private void SaveFilter(string filter)
             {
                 this.filterString = filter;
 
-                if (ProjFS.Utils.DoesNameContainWildCards(this.filterString))
+                if (Microsoft.Windows.ProjFS.Utils.DoesNameContainWildCards(this.filterString))
```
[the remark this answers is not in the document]
jeschu1: Sure can i'll update

GVFS/GVFS.Platform.Windows/DiskLayoutUpgrades/DiskLayout12to13Upgrade_FolderPlaceholder.cs
```diff
@@ -81,30 +81,32 @@ private static IEnumerable<string> GetFolderPlaceholdersFromDisk(ITracer tracer,
...
-                        }
-                        else if (result != HResult.FileNotFound)
-                        {
-                            // FileNotFound is returned for tombstones when the filter is attached to the volume so we want to
-                            // just skip those folders.  Any other HResults may cause valid folder placeholders not to be written
-                            // to the placeholder database so we want to error out on those.
-                            throw new InvalidDataException($"Error getting on disk file state. HResult = {result} for {directory}");
+                            else
+                            {
+                                // May cause valid folder placeholders not to be written
```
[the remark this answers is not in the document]
jeschu1: Basically we now throw exceptions where previously "HResult.FileNotFound" would not have.
@wilbaker can you confirm the change in the API and why this was the right move.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -817,10 +769,6 @@ private HResult QueryFileNameHandler(string virtualPath)
             {
                 EventMetadata metadata = this.CreateEventMetadata(virtualPath, e);
                 metadata.Add("commandId", commandId);
-                metadata.Add("desiredAccess", desiredAccess);
```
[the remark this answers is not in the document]
jeschu1: If you follow the function up we used to these fields as part of the function call and now we do not.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1111,65 +1056,39 @@ private HResult QueryFileNameHandler(string virtualPath)
...
-                            {
-                                this.FileSystemCallbacks.OnFileSuperseded(virtualPath);
-                            }
-
-                            break;
-
-                        default:
-                            break;
-                    }
+                    this.FileSystemCallbacks.OnFileOverwritten(virtualPath);
```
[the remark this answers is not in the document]
jeschu1: If you follow the code old paths, the same things was done for both cases.

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -135,31 +140,22 @@ private HResult EndDirectoryEnumeration(Guid enumerationId)
...
                     {
-                        if (result == HResult.InsufficientBuffer)
-                        {
-                            if (entryAdded)
-                            {
-                                result = HResult.Ok;
-                            }
-                        }
-
+                        result = entryAdded ? HResult.Ok : HResult.InsufficientBuffer;
```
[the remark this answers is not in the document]
jeschu1: I actually asked the same question, not sure why the comment didn't come over.

@wilbaker can you confirm why this was the right move?

GVFS/GVFS.Platform.Windows/DiskLayoutUpgrades/DiskLayout12to13Upgrade_FolderPlaceholder.cs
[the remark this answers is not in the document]
wilbaker: @kewillford it looks like you added this check in PR 336155 (in Azure Repos).  Do you remember how we were getting in a state where `fileSystem.EnumerateDirectories(path)` was returning a tombstone?

@cgallred what should `Utils.TryGetOnDiskFileState` return for a path that's a tombstone when the filter is attached?

MirrorProvider/MirrorProvider.Windows/WindowsFileSystemVirtualizer.cs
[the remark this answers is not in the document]
wilbaker: > Is the change in behavior intentional as a result of the new API? Could Add fail for anything other than InsufficientBuffer?

The managed API forwards the request to [PrjFillDirEntryBuffer](https://docs.microsoft.com/en-us/windows/desktop/api/projectedfslib/nf-projectedfslib-prjfilldirentrybuffer) which should only return `S_OK` or `HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER)`.

The managed API used to just pass the HRESULT back up to the caller, but it now translates the return value to `true` or `false`.

More details here:

_Documentation_
>    ///     If this method returns false, the provider returns HResult::Ok and waits for
    ///     the next GetDirectoryEnumerationCallback.  Then it resumes filling the enumeration with
    ///     the entry it was trying to add when it got false. 
    ///     
    ///     If the function returns false for the first file or directory in the enumeration, the
    ///     provider returns HResult::InsufficientBuffer from the GetDirectoryEnumerationCallback 
    ///     method.

https://github.com/Microsoft/ProjFS-Managed-API/blob/9ac27a91830cd3597f5f6249ffd18ff0eba141bf/ProjectedFSLib.Managed.API/DirectoryEnumerationResults.h#L68

_Sample Provider_
https://github.com/Microsoft/ProjFS-Managed-API/blob/9ac27a91830cd3597f5f6249ffd18ff0eba141bf/simpleProviderManaged/SimpleProvider.cs#L406

GVFS/GVFS.Platform.Windows/DiskLayoutUpgrades/DiskLayout12to13Upgrade_FolderPlaceholder.cs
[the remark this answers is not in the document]
kewillford: I don't remember exactly but in reading through this maybe the previous comment should have been "when the filter is **not** attached" because you are right that the enumeration should not even return the directories that are tombstones but without the filter attached it might have been since this is potentially running before the filter has been attached to the volume.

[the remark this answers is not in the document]
kewillford: Found the email that addressed this and it was a bug in prjflt that was fixed and this was the quick fix for the bug so I think we should be good here.  See Bug 17097791 in the Windows Azure workitems.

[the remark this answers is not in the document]
wilbaker: @kewillford thanks for the details!

> See Bug 17097791 in the Windows Azure workitems.

It looks like 17097791 was fixed in RS5.  

@jrbriggs @jeschu1, I suspect that if we don't do something special here then a user could have an issue if:

- They are still running on Windows RS4
- They have an old repo that has not been upgraded since this upgrade step was added.

My guess it's unlikely anyone will hit this, but I still wanted to mention it so that we can decide on the risk vs maintenance cost.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
```diff
@@ -174,12 +149,61 @@ public void WaitForCreateWriteBuffer()
...
+            return HResult.Ok;
+        }
+
+        public HResult CompleteCommand(int commandId, IDirectoryEnumerationResults results)
+        {
+            this.commandCompleted.Set();
+            return HResult.Ok;
+        }
+
+        HResult IVirtualizationInstance.CompleteCommand(int commandId, HResult completionResult)
```
wilbaker: Why did you need to include the `IVirtualizationInstance.` here and below?

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
```diff
@@ -174,12 +149,61 @@ public void WaitForCreateWriteBuffer()
...
             this.Dispose(true);
             GC.SuppressFinalize(this);
         }
 
+        public HResult WriteFileData(Guid dataStreamId, IWriteBuffer buffer, ulong byteOffset, uint length)
+        {
+            return this.WriteFileReturnResult;
+        }
+
+        IWriteBuffer IVirtualizationInstance.CreateWriteBuffer(ulong byteOffset, uint length, out ulong alignedByteOffset, out uint alignedLength)
```
wilbaker: Where is this version of `CreateWriteBuffer` called? (If it's not getting called can we just have it throw `NotImplementedException`)

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
```diff
@@ -174,12 +149,61 @@ public void WaitForCreateWriteBuffer()
             this.waitForCreateWriteBuffer.WaitOne();
         }
 
+        public HResult CompleteCommand(int commandId, NotificationType newNotificationMask)
+        {
```
wilbaker: I think the only version of `CompleteCommand` we're calling is

`CompleteCommand(int commandId, HResult completionResult)`

Can the others just throw `NotImplementedException`?

GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
```diff
@@ -182,9 +185,9 @@ public void OnStartDirectoryEnumerationReturnsPendingWhenResultsNotInMemory()
 
                     Guid enumerationGuid = Guid.NewGuid();
                     gitIndexProjection.EnumerationInMemory = false;
-                    mockVirtualization.OnStartDirectoryEnumeration(1, enumerationGuid, "test").ShouldEqual(HResult.Pending);
+                    mockVirtualization.requiredCallbacks.StartDirectoryEnumerationCallback(1, enumerationGuid, "test", 1, "test").ShouldEqual(HResult.Pending);
```
wilbaker: Pass in `TriggeringProcessId` and `TriggeringProcessImageFileName` here instead of `1, "test"`?

GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
```diff
@@ -526,7 +529,13 @@ public void MoveFileIntoDotGitDirectory()
...
                     NotificationType notificationType = NotificationType.UseExistingMask;
-                    mockVirtualization.OnNotifyFileRenamed("test.txt", Path.Combine(".git", "test.txt"), isDirectory: false, notificationMask: ref notificationType);
+                    mockVirtualization.OnNotifyFileRenamed(
+                        "test.txt",
+                        Path.Combine(".git", "test.txt"),
+                        isDirectory: false,
+                        triggeringProcessId: TriggeringProcessId,
+                        triggeringProcessImageFileName: TriggeringProcessImageFileName,
+                        notificationMask: out notificationType);
                     fileSystemCallbacks.OnIndexFileChangeCallCount.ShouldEqual(0);
```
wilbaker: Add validation of `notificationType`, e.g. `notificationType.ShouldEqual(NotificationType.UseExistingMask)`?

GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
```diff
@@ -535,7 +544,13 @@ public void MoveFileIntoDotGitDirectory()
...
                     // We don't expect something to rename something from outside the .gitdir to the .git\index, but this
                     // verifies that we behave as expected in case that happens
-                    mockVirtualization.OnNotifyFileRenamed("test.txt", Path.Combine(".git", "index"), isDirectory: false, notificationMask: ref notificationType);
+                    mockVirtualization.OnNotifyFileRenamed(
+                        "test.txt",
+                        Path.Combine(".git", "index"),
+                        isDirectory: false,
+                        triggeringProcessId: TriggeringProcessId,
+                        triggeringProcessImageFileName: TriggeringProcessImageFileName,
+                        notificationMask: out notificationType);
```
wilbaker: Validate `notificationType` here as well.

I see there are a few more spots below.  Now that `notificationType` is an `out` parameter we should validate it everywhere we expect it to be set.

GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
```diff
@@ -846,7 +885,7 @@ public void OnGetFileStreamHandlesWriteFailure()
 
         [TestCase]
         [Category(CategoryConstants.ExceptionExpected)]
-        public void OnGetFileStreamHandlesFileClosed()
+        public void OnGetFileStreamHandlesNewResult()
```
wilbaker: Naming suggestion: `OnGetFileStreamHandlesNewResult` -> `OnGetFileStreamHandlesHResultHandleResult`

GVFS/GVFS.Platform.Windows/WindowsRequiredCallbacks.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Windows.ProjFS;
+
+namespace GVFS.Platform.Windows
+{
+    public class WindowsRequiredCallbacks : IRequiredCallbacks
```
kewillford: This class seems like it is not needed.  Why not just have `WindowsFileSystemVirtualizer` implement `IRequiredCallbacks`?  This class just creates a bunch of stuff so that methods in `WindowsFileSystemVirtualizer` will get called anyways.  This class would completely go away and you would call
`this.virtualizationInstance.StartVirtualizing(this)` instead.

MirrorProvider/MirrorProvider.Windows/WindowsRequiredCallbacks.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Microsoft.Windows.ProjFS;
+using System;
+
+namespace MirrorProvider.Windows
+{
+    public class WindowsRequiredCallbacks : IRequiredCallbacks
```
kewillford: Same as above, this class really is not needed.

GVFS/GVFS.Platform.Windows/WindowsRequiredCallbacks.cs
jeschu1: That seems like a fair suggestion.  @wilbaker any objection with that approach?

wilbaker: No objections from me, that approach sounds good.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -306,7 +308,7 @@ private bool TryCompleteCommand(int commandId, HResult result)
 
         // TODO: Need ProjFS 13150199 to be fixed so that GVFS doesn't leak memory if the enumeration cancelled.
         // Currently EndDirectoryEnumerationHandler must be called to remove the ActiveEnumeration from this.activeEnumerations
-        private HResult StartDirectoryEnumerationHandler(int commandId, Guid enumerationId, string virtualPath)
+        private HResult StartDirectoryEnumerationHandler(int commandId, Guid enumerationId, string virtualPath, uint triggeringProcessId, string triggeringProcessImageFileName)
         {
             try
```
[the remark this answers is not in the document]
jeschu1: I have some questions on that one, so i'd prefer to defer it for now

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1003,15 +951,15 @@ private HResult QueryFileNameHandler(string virtualPath)
                                             // If we hit this in GVFS, the most common explanation is that we're calling WriteFile after the ProjFS thread waiting on the respose
                                             // from GetFileStream has already timed out
                                             {
-                                                requestMetadata.Add(TracingConstants.MessageKey.InfoMessage, $"{nameof(this.virtualizationInstance.WriteFile)} returned StatusObjectNameNotFound");
+                                                requestMetadata.Add(TracingConstants.MessageKey.InfoMessage, $"{nameof(this.virtualizationInstance.WriteFileData)} returned StatusObjectNameNotFound");
```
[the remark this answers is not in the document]
jeschu1: ping @cgallred, trying to wrap up this PR

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
jeschu1: Because I was being silly!

GVFS/GVFS.UnitTests.Windows/Windows/Virtualization/WindowsFileSystemVirtualizerTests.cs
jeschu1: Done locally!  Note, this was existing, but since it's a small change I incorporated it.

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockWriteBuffer.cs
```diff
@@ -0,0 +1,45 @@
...
+        public long Length
+        {
+            get;
+            set;
+        }
+
+        public void Dispose()
+        {
+            this.Stream.Dispose();
+            this.Stream = null;
```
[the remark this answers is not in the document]
jeschu1: Please double check the way I've done this, not convinced i've done it properly

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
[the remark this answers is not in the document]
cgallred: @wilbaker @jeschu1 sorry, I didn't notice the earlier notification on this.  That's correct.  `WriteFileData` shouldn't be able to return `FileNotFound` anymore.

GVFS/GVFS.Platform.Windows/DiskLayoutUpgrades/DiskLayout12to13Upgrade_FolderPlaceholder.cs
```diff
@@ -81,30 +81,32 @@ private static IEnumerable<string> GetFolderPlaceholdersFromDisk(ITracer tracer,
...
                     {
-                        OnDiskFileState fileState = OnDiskFileState.Full;
-                        HResult result = Utils.GetOnDiskFileState(directory, out fileState);
-                        if (result == HResult.Ok)
+                        if (Directory.Exists(directory))
                         {
-                            if (IsPlaceholder(fileState))
+                            OnDiskFileState fileState = OnDiskFileState.Full;
+                            bool result = Utils.TryGetOnDiskFileState(directory, out fileState);
+                            if (result == true)
```
kewillford: Get rid of the `result` variable and change to 
`if (Utils.TryGetOnDiskFileState(directory, out fileState))`

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockWriteBuffer.cs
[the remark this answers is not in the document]
wilbaker: I think this works, but you could consider using the [`Dispose` pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) for classes that have finalizers.

```
     ~MockWriteBuffer()
    {
        this.Dispose(false);
    }

    public void Dispose()
    { 
       Dispose(true);
       GC.SuppressFinalize(this);           
    }

    protected virtual void Dispose(bool disposing)
    {
        if (disposing)
        {   
            if (this.Stream != null)
            {
                this.Stream.Dispose();
                this.Stream = null;
            }
        }

         if (this.memIntPtr != IntPtr.Zero)
         {
             Marshal.FreeHGlobal(this.memIntPtr);
             this.memIntPtr = IntPtr.Zero;
        }
    }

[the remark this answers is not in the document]
Comment (kewillford) [reply GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockWriteBuffer.cs]: See this article for how to implement the dispose pattern https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

File: GVFS/GVFS.Platform.Windows/DiskLayoutUpgrades/DiskLayout12to13Upgrade_FolderPlaceholder.cs [hunk 24]
```diff
@@ -81,30 +81,32 @@ private static IEnumerable<string> GetFolderPlaceholdersFromDisk(ITracer tracer,
                 {
                     if (!directory.EndsWith(Path.DirectorySeparatorChar + GVFSConstants.DotGit.Root))
                     {
-                        OnDiskFileState fileState = OnDiskFileState.Full;
-                        HResult result = Utils.GetOnDiskFileState(directory, out fileState);
-                        if (result == HResult.Ok)
+                        if (Directory.Exists(directory))
```
wilbaker: I might have missed this in the earlier conversation; why was this `if (Directory.Exists(directory))` check added?

GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockWriteBuffer.cs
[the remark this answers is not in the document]
jeschu1: Thanks @wilbaker @kewillford 

The comment "this should probably go in a finalizer" where I got confused.

The code above is the standard pattern and that makes sense.

GVFS/GVFS.Platform.Windows/DiskLayoutUpgrades/DiskLayout12to13Upgrade_FolderPlaceholder.cs
jeschu1: We discussed this in one of our meetings with @cgallred.

Previously Utils.`GetOnDiskFileState `returned a result.  Now it has been turned into a bool.  
The only case we checked the HResult for previously was when `result != HResult.FileNotFound`.  In that instance we would not throw an exception.  We can simulate the same check by doing Directory.Exists upfront.

Please let me know if that makes sense or if I missed something!

wilbaker: > The only case we checked the HResult for previously was when `result != HResult.FileNotFound`

I thought in our discussion with @cgallred we concluded that this check was only in the code to work around a  bug in RS4 where `TryGetOnDiskFileState` would return `HResult.FileNotFound` for tombstones and that moving forward VFS4G wasn't going to try to work around the Windows bug here.

> We can simulate the same check by doing Directory.Exists upfront.

If the directory is returned by `fileSystem.EnumerateDirectories(path)` would `Directory.Exists(directory)` return false?

If the `if (Directory.Exists(directory))` is adding value then I think we should include a comment explaining what the check is for.  If the `if (Directory.Exists(directory))` does not functionally change what we're doing here then it should be removed to reduce the complexity of this method.

kewillford: Yes, the only reason we were checking `HResult.FileNotFound` before was as a work around so I don't think we need this check unless we are continuing to work around that issue on a few machines that might continue to have the bug.

jeschu1: @wilbaker @kewillford, check removed

PR #886, opened 2019-03-06
Mac kext: Don't try to message offline providers, fixes missing root assert

pmj:
This fixes #882.

1. The first commit adds an assertion that a root's path is not poisoned when attempting to use it. This means the bug will also be caught in debug builds.
2. The second commit fixes the actual bug - attempting to send a message to a root that doesn't have a provider from the fileop handler. This is done by checking if the root has an online provider in `ShouldHandleFileOpEvent()`.
3. The third commit fixes some kextlog messages which I noticed while debugging this.

PR #888, opened 2019-03-06
Merge master to releases/mac_ea

jrbriggs:
Update releases/mac_ea to pick up two fixes that we want for Early Adopters:
#886 which was causing kernel panics
#884 to ensure that all VFS4G related processes are shutdown before installing/upgrading


wilbaker:
Are we going to wait for #884 to be merged in before merging this PR?

jrbriggs:
I think we should, yes.

PR #898, opened 2019-03-07
Upgrader: reacquire credentials on auth failure

jamill:
The NuGet upgrader can get stale credentials. When the credentials fail (due to a 401 response), we want to clear the credentials in the credential manager and re-acquire fresh credentials.

This change will be ported to the current release, so I would like to keep it structured in a way that is amenable to porting.

### DIFF
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
@@ -53,12 +53,7 @@ namespace GVFS.Common.NuGetUpgrade
             this.sourceCacheContext.NoCache = true;
 
             this.nuGetLogger = new Logger(this.tracer);
-
-            this.sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
-            if (!string.IsNullOrEmpty(this.personalAccessToken))
-            {
-                this.sourceRepository.PackageSource.Credentials = BuildCredentialsFromPAT(this.personalAccessToken);
-            }
+            this.SetSourceRepository();
         }
 
         public void Dispose()
@@ -68,6 +63,13 @@ namespace GVFS.Common.NuGetUpgrade
             this.sourceCacheContext = null;
         }
 
+        public void SetCredentials(string credential)
+        {
+            this.personalAccessToken = credential;
+
+            this.SetSourceRepository();
+        }
+
         /// <summary>
         /// Query a NuGet feed for list of packages that match the packageId.
         /// </summary>
@@ -75,25 +77,16 @@ namespace GVFS.Common.NuGetUpgrade
         /// <returns>List of packages that match query parameters</returns>
         public virtual async Task<IList<IPackageSearchMetadata>> QueryFeedAsync(string packageId)
         {
-            try
-            {
-                PackageMetadataResource packageMetadataResource = await this.sourceRepository.GetResourceAsync<PackageMetadataResource>();
-                IEnumerable<IPackageSearchMetadata> queryResults = await packageMetadataResource.GetMetadataAsync(
-                    packageId,
-                    includePrerelease: false,
-                    includeUnlisted: false,
-                    sourceCacheContext: this.sourceCacheContext,
-                    log: this.nuGetLogger,
-                    token: CancellationToken.None);
+            PackageMetadataResource packageMetadataResource = await this.sourceRepository.GetResourceAsync<PackageMetadataResource>();
+            IEnumerable<IPackageSearchMetadata> queryResults = await packageMetadataResource.GetMetadataAsync(
+                packageId,
+                includePrerelease: false,
+                includeUnlisted: false,
+                sourceCacheContext: this.sourceCacheContext,
+                log: this.nuGetLogger,
+                token: CancellationToken.None);
 
-                return queryResults.ToList();
-            }
-            catch (Exception ex)
-            {
-                EventMetadata data = CreateEventMetadata(ex);
-                this.tracer.RelatedWarning(data, "Error encountered when querying NuGet feed");
-                throw new Exception($"Failed to query the NuGet package feed due to error: {ex.Message}", ex);
-            }
+            return queryResults.ToList();
         }
 
         /// <summary>
@@ -166,6 +159,15 @@ namespace GVFS.Common.NuGetUpgrade
                 "PersonalAccessToken",
                 personalAccessToken,
                 storePasswordInClearText: false);
+        }
+
+        private void SetSourceRepository()
+        {
+            this.sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
+            if (!string.IsNullOrEmpty(this.personalAccessToken))
+            {
+                this.sourceRepository.PackageSource.Credentials = BuildCredentialsFromPAT(this.personalAccessToken);
+            }
         }
 
         /// <summary>
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -222,7 +222,7 @@ namespace GVFS.Common.NuGetUpgrade
         {
             try
             {
-                IList<IPackageSearchMetadata> queryResults = this.nuGetFeed.QueryFeedAsync(this.nuGetUpgraderConfig.PackageFeedName).GetAwaiter().GetResult();
+                IList<IPackageSearchMetadata> queryResults = this.QueryFeed(firstAttempt: true);
 
                 // Find the package with the highest version
                 IPackageSearchMetadata highestVersionAvailable = null;
@@ -455,10 +455,40 @@ namespace GVFS.Common.NuGetUpgrade
             }
         }
 
+        protected static EventMetadata CreateEventMetadata(Exception e = null)
+        {
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("Area", nameof(NuGetFeed));
+            if (e != null)
+            {
+                metadata.Add("Exception", e.ToString());
+            }
+
+            return metadata;
+        }
+
         private static bool TryGetPersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer, out string token, out string error)
         {
             GitProcess gitProcess = new GitProcess(gitBinaryPath, null, null);
             return gitProcess.TryGetCredentials(tracer, credentialUrl, out string username, out token, out error);
+        }
+
+        private static void ErasePersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer)
+        {
+            GitProcess gitProcess = new GitProcess(gitBinaryPath, null, null);
+            gitProcess.RejectCredentials(credentialUrl);
+        }
+
+        private static bool TryReacquirePersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer, out string token, out string error)
+        {
+            ErasePersonalAccessToken(gitBinaryPath, credentialUrl, tracer);
+
+            if (!TryGetPersonalAccessToken(gitBinaryPath, credentialUrl, tracer, out token, out error))
+            {
+                return false;
+            }
+
+            return true;
         }
 
         private static string replacementToken(string tokenString)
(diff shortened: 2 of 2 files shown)


GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,66 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+            {
+                try
+                {
+                    queryResults = this.nuGetFeed.QueryFeedAsync(this.nuGetUpgraderConfig.PackageFeedName).GetAwaiter().GetResult();
+                    return queryResults;
+                }
+                catch (Exception ex) when (ex is FatalProtocolException &&
+                                           ex.InnerException != null &&
+                                           ex.InnerException is System.Net.Http.HttpRequestException &&
+                                           ex.InnerException.Message.Contains("401 (Unauthorized)") &&
```
wilbaker: Would there be any downsides to just checking for "401"?  I wasn't sure what the chances are that the exception message could change.

jrbriggs: Additionally, should we check for 403 in case we have a PAT scope issue that could be resolved by reacquiring?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,66 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+                    return false;
+                }
+
+                NuGetUpgrader.TryReacquirePersonalAccessToken(gitBinPath, authUrl, this.tracer, out string token, out error);
+                this.nuGetFeed.SetCredentials(token);
+                return true;
+            }
+            catch (Exception ex)
+            {
+                error = ex.Message;
```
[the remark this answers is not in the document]
jrbriggs: Along the same line, we'll still get the error into the telemetry streams in the event of a failure, correct?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,66 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+                catch (Exception ex) when (ex is FatalProtocolException &&
+                                           ex.InnerException != null &&
+                                           ex.InnerException is System.Net.Http.HttpRequestException &&
+                                           ex.InnerException.Message.Contains("401 (Unauthorized)") &&
+                                           retryCount == 0)
+                {
+                    EventMetadata data = CreateEventMetadata(ex);
+                    this.tracer.RelatedWarning(data, "Failed to query feed due to unauthorized error. Re-acquiring new credentials and trying again.");
+                    retryCount++;
+                    if (!this.TryRefreshCredentials(out string error))
```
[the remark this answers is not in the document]
jrbriggs: I think it would be relatively infrequent. It would indicate an error with permissions at that point, I think, such as lack of read access to the feed.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -453,12 +453,42 @@ public override bool TryRunInstaller(InstallActionWrapper installActionWrapper,
...
         private static bool TryGetPersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer, out string token, out string error)
         {
             GitProcess gitProcess = new GitProcess(gitBinaryPath, null, null);
             return gitProcess.TryGetCredentials(tracer, credentialUrl, out string username, out token, out error);
+        }
+
+        private static void ErasePersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer)
+        {
+            GitProcess gitProcess = new GitProcess(gitBinaryPath, null, null);
+            gitProcess.RejectCredentials(credentialUrl);
```
jeschu1: @jamill did you double check this erases using windows GCM?  It absolutely should, but I remember there is some weird logic there.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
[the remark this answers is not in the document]
jamill: I expect the `TryRefreshCredentials` to generally succeed, even if there is an initial failure in `QueryFeedAsync`.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: I tested this with revoking my credentials, and GCM erased the credentials as requested.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: Sounds good - I will check just for the status code, and will check for 403 as well (and check if there are other status codes to retry on as well).

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,71 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+            {
+                string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+
+                string authUrl;
+                if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
+                {
+                    return false;
+                }
+
+                NuGetUpgrader.TryReacquirePersonalAccessToken(gitBinPath, authUrl, this.tracer, out string token, out error);
```
jrbriggs: Check return value?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,71 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+                }
+                catch (Exception ex)
+                {
+                    EventMetadata data = CreateEventMetadata(ex);
+                    string message = $"Error encountered when querying NuGet feed. Number of attempts: {retryCount + 1}";
+                    this.tracer.RelatedWarning(data, message);
+                    throw new Exception($"Failed to query the NuGet package feed due to error: {ex.Message}", ex);
+                }
+            }
+            while (true);
```
kewillford: Are we retrying forever if something doesn't go as expected?  I would prefer some sort of retry limit.

jamill: We will only retry 1 time - I can restructure to make this more explicit / obvious

kewillford: Yeah I would make a method and check a return code and call the method again if it failed rather than a loop.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,75 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+        }
+
+        private bool TryRefreshCredentials(out string error)
+        {
+            try
+            {
+                string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+
+                string authUrl;
+                if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
```
kewillford: Just a note at this point that these should probably not be Nuget classes since we have tied implementation to Azure DevOps and things would not work if I had the package on some other Nuget feed.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,75 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+                        throw new Exception($"Failed to query the feed for upgrade packages due to: {ex.Message}, and was not able to re-acquire new credentials due to: {error}", ex);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    EventMetadata data = CreateEventMetadata(ex);
+                    string message = $"Error encountered when querying NuGet feed. Number of attempts: {retryCount + 1}";
+                    this.tracer.RelatedWarning(data, message);
+                    throw new Exception($"Failed to query the NuGet package feed due to error: {ex.Message}", ex);
+                }
```
kewillford: Do we even need this catch if we are rethrowing and it will only happen on a retry.  The call stack should give us the information we need when it is caught higher up in the call stack right?

GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
```diff
@@ -168,6 +161,15 @@ private static PackageSourceCredential BuildCredentialsFromPAT(string personalAc
                 storePasswordInClearText: false);
         }
 
+        private void SetSourceRepository()
+        {
+            this.sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
```
kewillford: Are there any affects of resetting the `sourceRepository` on the retry when we set the credentials?  Why could we not just set the `this.sourceRepository.PackageSource.Credentials` on the retry?  Is the connection bad after the failed attempt?

kewillford:
Are there any tests that we can write for this logic?

GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
jamill: > Why could we not just set the this.sourceRepository.PackageSource.Credentials on the retry?

I tried this originally, but it did not appear to actual result in the new credentials being used. I have it on my list to dig into further.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: The excepton is caught here because the underlying exception can be quite ambiguous and unhelpful to a user. The idea was put some context around the unhelpful error (e.g. "Cannot Query the NuGet Feed"). 

In practice, I think what I have still falls short, but I thought I would tackle that separately.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jeschu1: The version that wouldn't properly delete is prior to 1.18.4.  Credit to @mjcheetham for finding the commit https://github.com/Microsoft/Git-Credential-Manager-for-Windows/commit/d8b3b320c253cea8594fcc3f3f57d76b0abc1d79#diff-52aaad27aeab0754e3c8c6e4ee388026

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,91 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+
+        private bool IsAuthRelatedException(Exception ex)
+        {
+            // In observation, we have seen either an HttpRequestException directly, or
+            // a FatalProtocolException wrapping an HttpRequestException when we are not able
+            // to auth against the NuGet feed.
+            System.Net.Http.HttpRequestException httpRequestException = null;
+            if (ex is System.Net.Http.HttpRequestException)
+            {
+                httpRequestException = ex as System.Net.Http.HttpRequestException;
```
mjcheetham: You can do a hard-cast here rather than a soft one because you've already checked the type.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,91 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+            }
+            else if (ex is FatalProtocolException &&
+                ex.InnerException is System.Net.Http.HttpRequestException)
+            {
+                httpRequestException = ex.InnerException as System.Net.Http.HttpRequestException;
```
mjcheetham: Same comment about using a hard-cast here.
Alternatively you could also use the new pattern matching syntax to avoid a double type check:

```csharp
else if (ex is FatalProtocolException &&
         ex.InnerException is System.Net.Http.HttpRequestException innerHttpEx)
{
    httpRequestException = innerHttpEx;
}
```

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -492,6 +522,91 @@ private bool TryRecursivelyDeleteInstallerDirectory(out string error)
...
+            }
+
+            if (httpRequestException != null &&
+                (httpRequestException.Message.Contains("401") || httpRequestException.Message.Contains("403")))
```
mjcheetham: It's a shame the best we can do here is check for the strings "401" or "403" in the `Message` 😞

jamill:
> Are there any tests that we can write for this logic?

The tests I had thought about writing were to validate that we make the expected calls into GitProcess to reject / re-acquire the credentials. I don't think we have a great pattern / mechanism to test this at the moment.

Did you have thoughts on other aspects we could test here? Otherwise, given that we will need some plumbing work, and this PR is targeting release, I propose we proceed with the current change and I will followup to make the version in master testable.

GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
jrbriggs: Seems like, if anything, it's safer to dump the whole `this.sourceRepository` to prevent more subtle bugs from cropping up.

PR #901, opened 2019-03-08
[releases/shipped] Use a thread-safe collection for the failed listeners in JsonTracer

### LINKED ISSUE #893: NullReferenceException in GetPlaceholderInformationAsyncHandler
Version 1.0.19060.1 logged a NullReferenceException

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at GVFS.Common.Tracing.JsonTracer.MarkAndLogListenerFailure(EventListener failedListener, String errorMessage)
   at GVFS.Common.Tracing.JsonTracer.WriteEvent(String eventName, EventLevel level, Keywords keywords, EventMetadata metadata, EventOpcode opcode)
   at GVFS.Common.Tracing.JsonTracer.RelatedEvent(EventLevel level, String eventName, EventMetadata metadata, Keywords keyword)
   at GVFS.Common.Tracing.JsonTracer.RelatedError(EventMetadata metadata, String message, Keywords keywords)
   at GVFS.Platform.Windows.WindowsFileSystemVirtualizer.GetPlaceholderInformationAsyncHandler(CancellationToken cancellationToken, BlobSizesConnection blobSizesConnection, Int32 commandId, String virtualPath, UInt32 desiredAccess, UInt32 shareMode, UInt32 createDisposition, UInt32 createOptions, UInt32 triggeringProcessId, String triggeringProcessImageFileName) in E:\A\_work\586\s\GVFS\GVFS.Platform.Windows\WindowsFileSystemVirtualizer.cs:line 803
```

wilbaker:
Fixes #893 

`failedListeners` can be modified on multiple threads, and so we should use a `ConcurrentDictionary` rather than a `Dictionary`.

### DIFF
--- a/GVFS/GVFS.Common/Tracing/JsonTracer.cs
+++ b/GVFS/GVFS.Common/Tracing/JsonTracer.cs
@@ -1,5 +1,6 @@
 ﻿using Newtonsoft.Json;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -11,7 +12,7 @@ namespace GVFS.Common.Tracing
         public const string NetworkErrorEventName = "NetworkError";
 
         private readonly List<EventListener> listeners;
-        private readonly Dictionary<EventListener, string> failedListeners = new Dictionary<EventListener, string>();
+        private readonly ConcurrentDictionary<EventListener, string> failedListeners = new ConcurrentDictionary<EventListener, string>();
 
         private readonly string activityName;
         private readonly Guid parentActivityId;
@@ -339,32 +340,33 @@ namespace GVFS.Common.Tracing
 
         private void MarkAndLogListenerRecovery(EventListener recoveredListener)
         {
+            // Check ContainsKey first (rather than always calling TryRemove) because ContainsKey
+            // is lock-free and recoveredListener should rarely be in failedListeners
             if (!this.failedListeners.ContainsKey(recoveredListener))
             {
-                // This listener has not failed since the last time it was called, so no need to do anything
+                // This listener has not failed since the last time it was called, so no need to log recovery
                 return;
             }
 
-            this.failedListeners.Remove(recoveredListener);
-
-            TraceEventMessage message = CreateListenerRecoveryMessage(recoveredListener);
-
-            // Only log that the listener has recovered to the other good listeners
-            foreach (EventListener listener in this.listeners.Except(this.failedListeners.Keys))
+            if (this.failedListeners.TryRemove(recoveredListener, out _))
             {
-                listener.TryRecordMessage(message, out _);
+                TraceEventMessage message = CreateListenerRecoveryMessage(recoveredListener);
+
+                // Only log that the listener has recovered to the other good listeners
+                foreach (EventListener listener in this.listeners.Except(this.failedListeners.Keys))
+                {
+                    listener.TryRecordMessage(message, out _);
+                }
             }
         }
 
         private void MarkAndLogListenerFailure(EventListener failedListener, string errorMessage)
         {
-            if (this.failedListeners.ContainsKey(failedListener))
+            if (!this.failedListeners.TryAdd(failedListener, errorMessage))
             {
                 // We've already logged that this listener has failed so there is no need to do it again
                 return;
             }
-
-            this.failedListeners.Add(failedListener, errorMessage);
 
             TraceEventMessage message = CreateListenerFailureMessage(failedListener, errorMessage);
 


wilbaker:
Full functional tests running here (I kicked them off manually):

https://gvfs.visualstudio.com/ci/_build/results?buildId=7909

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -339,40 +340,39 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
 
         private void MarkAndLogListenerRecovery(EventListener recoveredListener)
         {
+            // Check ContainsKey first (rather than always calling TryRemove) because ContainsKey
+            // is lock-free and recoveredListener should rarely be in failedListeners
             if (!this.failedListeners.ContainsKey(recoveredListener))
```
jeschu1: @wilbaker and I talked offline and it's fine if the ContainsKey check returns true when it's turned to false (no-lock).  The TryRemove would then catch it.

What about the case when Contains key is false but becomes true.  Is that a case that could happen and if so would it be handled?

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -339,40 +340,39 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
...
+                {
+                    listener.TryRecordMessage(message, out _);
+                }
             }
         }
 
         private void MarkAndLogListenerFailure(EventListener failedListener, string errorMessage)
         {
-            if (this.failedListeners.ContainsKey(failedListener))
+            // If we've already logged that this listener has failed there is no need to do it again
```
wilbaker: I took the `ContainsKey` check out here because the failure scenario shouldn't be a hot path (and if it is we have bigger problems!)

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -339,40 +340,39 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
...
+                foreach (EventListener listener in this.listeners.Except(this.failedListeners.Keys))
+                {
+                    listener.TryRecordMessage(message, out _);
+                }
             }
         }
 
         private void MarkAndLogListenerFailure(EventListener failedListener, string errorMessage)
         {
-            if (this.failedListeners.ContainsKey(failedListener))
```
jeschu1: For anyone else reviewing the Contains key is removes here cause there is no perf win and it's unnecessary with TryAdd.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
wilbaker: If `ContainsKey` returns `false` but another thread *just* added `recoveredListener` it means the listener is failing and succeeding on two different threads at the same time.

That means the thread that just added the listener to `failedListeners` will log the failure and this thread (where `ContainsKey` returned false) will exit early and not return anything.

Then the next time `MarkAndLogListenerRecovery` is called `ContainsKey` will return true and recovery will be logged.

Note that this race could happen even if we removed the call to `ContainsKey` and relied on `TryRemove`.

Thread 1: Enters `MarkAndLogListenerFailure`
Thread 2: Enters `MarkAndLogListenerRecovery`
Thread 2: Calls `TryRemove` and it returns false (just as `ContainsKey` would)
Thread 1: Calls `TryAdd` and it returns true
Thread 1: Logs listener error
Later: `MarkAndLogListenerRecovery` is called and recovery is logged

The only change introduced by `ContainsKey` is that if it crosses with "Thread 1: Calls `TryAdd`" `ContainsKey` could return either `true` or `false`.

In general if we're getting failures and successes at the same time on two different threads, there's not a guarantee on how the failure/recovery messages will be logged.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -339,40 +340,39 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
...
+            if (this.failedListeners.TryRemove(recoveredListener, out _))
             {
-                listener.TryRecordMessage(message, out _);
+                TraceEventMessage message = CreateListenerRecoveryMessage(recoveredListener);
+
+                // Only log that the listener has recovered to the other good listeners
+                foreach (EventListener listener in this.listeners.Except(this.failedListeners.Keys))
+                {
+                    listener.TryRecordMessage(message, out _);
+                }
```
kewillford: Could have a method for recording messages for non failed listeners so that we can share the code with `MarkAndLogListenerFailure`?

GVFS/GVFS.Common/Tracing/JsonTracer.cs
jamill: We could keep the previous structure (and avoid adding a level of nesting) by doing something like:

```
if (!this.failedListeners.TryAdd(….))
{
    return;
}
```

GVFS/GVFS.Common/Tracing/JsonTracer.cs
wilbaker: Sounds good, I'll make that change when I port this to master

PR #905, opened 2019-03-08
[upgrade] Upgrade Functional Test failure fix

### LINKED ISSUE #852: NoReminderForLeftOverDownloads test occasionally fails

alameenshah:
When Nuget settings (`upgrade.feedurl` & `upgrade.feedpackagename`) are not available in `gvfs config`, then an attempt to read (or write) those missing values will cause `gvfs config read/write` to exit with error code `3`. `UpgradeReminder` FT tries to 1. read existing nuget config, 2. delete it, 3. execute tests and then 4. restore previously read Nuget config. The failure used to happen in step 1. while attempting to read a config key which is not available in `gvfs config` file.

Fixes #852

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
@@ -108,9 +108,9 @@ namespace GVFS.FunctionalTests.Tools
             return this.CallGVFS("service " + argument, failOnError: true);
         }
 
-        public string ReadConfig(string key)
+        public string ReadConfig(string key, bool failOnError)
         {
-            return this.CallGVFS($"config {key}", failOnError: true).TrimEnd('\r', '\n');
+            return this.CallGVFS($"config {key}", failOnError).TrimEnd('\r', '\n');
         }
 
         public void WriteConfig(string key, string value)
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
```diff
@@ -110,7 +110,10 @@ public string RunServiceVerb(string argument)
 
         public string ReadConfig(string key)
         {
-            return this.CallGVFS($"config {key}", failOnError: true).TrimEnd('\r', '\n');
+            // failOnError is set to false because gvfs config read can exit with
+            // GenericError when the key-value is not available in config file. That
+            // is normal.
+            return this.CallGVFS($"config {key}", failOnError: false).TrimEnd('\r', '\n');
```
wilbaker: I recommend passing `bool failOnError` in as a parameter like we do for `Prefetch`.  That way the caller is forced to declare whether or not a failure is allowed/expected (it's generally safe to assume that `failOnError: true` is okay, but I don't think we should do that for `failOnError: false`).

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
```diff
@@ -108,9 +108,12 @@ public string RunServiceVerb(string argument)
...
         }
 
-        public string ReadConfig(string key)
+        public string ReadConfig(string key, bool failOnError)
         {
-            return this.CallGVFS($"config {key}", failOnError: true).TrimEnd('\r', '\n');
+            // failOnError is set to false because gvfs config read can exit with
+            // GenericError when the key-value is not available in config file. That
+            // is normal.
+            return this.CallGVFS($"config {key}", failOnError: failOnError).TrimEnd('\r', '\n');
```
wilbaker: Nit: There's no need for `failOnError: ` since we're passing in the named `failOnError` now.

wilbaker: Also, this comment:

```
            // failOnError is set to false because gvfs config read can exit with
            // GenericError when the key-value is not available in config file. That
            // is normal.
```

Should be moved to where `ReadConfig` is getting called.

PR #906, opened 2019-03-08
Catch exceptions when trying to set Visual Studio registry key

kewillford:
There are some cases when the registry key that is being written to
is volatile and will get the error message:

Cannot create a stable subkey under a volatile parent key.

This change will catch exceptions and continue to mount since the
registry key is not required for VFSForGit to work.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -193,27 +193,37 @@ namespace GVFS.Platform.Windows
 
         public override void ConfigureVisualStudio(string gitBinPath, ITracer tracer)
         {
-            const string GitBinPathEnd = "\\cmd\\git.exe";
-            string[] gitVSRegistryKeyNames =
+            try
             {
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -193,27 +193,37 @@ public override string GetNamedPipeName(string enlistmentRoot)
...
 
-            string regKeyValue = gitBinPath.Substring(0, gitBinPath.Length - GitBinPathEnd.Length);
-            foreach (string registryKeyName in gitVSRegistryKeyNames)
+                string regKeyValue = gitBinPath.Substring(0, gitBinPath.Length - GitBinPathEnd.Length);
+                foreach (string registryKeyName in gitVSRegistryKeyNames)
+                {
+                    Registry.SetValue(registryKeyName, GitVSRegistryValueName, regKeyValue);
+                }
+            }
+            catch (Exception ex)
```
alameenshah: is it worth replacing this catch-all handler with more specific one related to the issue you are fixing (writing to volatile registry keys)

kewillford: Thought about that and I didn't want any exception to stop VFSForGit from mounting since this is purely a helpful thing that we are doing for Visual Studio.

Issue #862, opened 2019-02-27
Mac: Add unit tests for VirtualizationRoot_FindForVnode


wilbaker:
Unit tests should cover the issue fixed in #856

pmj:
I've taken a look at what it will take to create some unit tests for

pmj:
Fat finger 🙄

pmj:
In the tests I implemented for this in #908 it turned out that keeping the genuine kext implementation of `FindOrDetectRootAtVnode()` was fine, so I ended up not implementing the function swap-out functionality just yet. I suspect we'll need it for something else soon enough.

PR #908, opened 2019-03-08
Mac kext: Unit tests for VirtualizationRoot_FindForVnode

pmj:
This sequence of patches builds up enough kext unit testing infrastructure to implement #862.

- [x] Tests calling `VirtualizationRoot_FindForVnode` successfully build & link.
- [x] `VirtualizationRoot_FindForVnode` expects to be able to walk a directory tree of vnodes.
- [x] `VirtualizationRoot_FindForVnode` expects to be able to call `Vnode_ReadXattr()` on each directory.
- [x] Tests with recycled vnodes
- [x] Tests that would have caught #797.
- [x] 100% test coverage in `VirtualizationRoot_FindForVnode`*

<span>*</span> Not quite 100% as we actually don't handle the `vnode_getparent()` failure case satisfactorily yet (#338), and I don't want to add tests that verify *incorrect* behaviour.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -35,7 +35,7 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
 static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t vnode, const FsidInode& vnodeFsidInode);
 
 static VirtualizationRootHandle FindUnusedIndex_Locked();
-static VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path);
+KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path);
 
 ActiveProviderProperties VirtualizationRoot_GetActiveProvider(VirtualizationRootHandle rootHandle)
 {
@@ -377,7 +377,7 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
 }
 
 // Returns negative value if it failed, or inserted index on success
-static VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path)
+KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path)
 {
     VirtualizationRootHandle rootIndex = FindUnusedIndexOrGrow_Locked();
     
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
@@ -111,7 +111,7 @@ shared_ptr<vnode> vnode::Create(const shared_ptr<mount>& mount, const char* path
     shared_ptr<vnode> result(new vnode(mount));
     s_allVnodes.insert(make_pair(result.get(), weak_ptr<vnode>(result)));
     result->weakSelfPointer = result;
-    result->SetPath(path);
+    result->SetAndRegisterPath(path);
     result->type = vnodeType;
     return result;
 }
@@ -138,7 +138,7 @@ void vnode::SetGetPathError(errno_t error)
     this->getPathError = error;
 }
 
-void vnode::SetPath(const string& path)
+void vnode::SetAndRegisterPath(const string& path)
 {
     s_vnodesByPath.erase(this->path);
 
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
@@ -8,6 +8,9 @@ using std::string;
 using std::unordered_map;
 using std::shared_ptr;
 using std::weak_ptr;
+using std::pair;
+using std::vector;
+using std::min;
 
 typedef unordered_map<string, weak_ptr<vnode>> PathToVnodeMap;
 typedef unordered_map<vnode_t, weak_ptr<vnode>> WeakVnodeMap;
@@ -309,11 +312,32 @@ void MockVnodes_CheckAndClear()
     s_allVnodes.clear();
 }
 
+vnode::BytesOrError vnode::ReadXattr(const char* xattrName)
+{
+    // TODO: add support for explicit error mocking
+    XattrMap::const_iterator found = this->xattrs.find(xattrName);
+    if (found == this->xattrs.end())
+    {
+        return BytesOrError{ .error = ENOATTR };
+    }
+    else
+    {
+        return BytesOrError{ .error = 0, .bytes = found->second };
+    }
+}
 
 SizeOrError Vnode_ReadXattr(vnode_t _Nonnull vnode, const char* _Nonnull xattrName, void* _Nullable buffer, size_t bufferSize)
 {
-    assert(false); // TODO: implement
-    return SizeOrError{};
+    vnode::BytesOrError xattrResult = vnode->ReadXattr(xattrName);
+    if (xattrResult.error == 0)
+    {
+        memcpy(buffer, xattrResult.bytes.data(), min(xattrResult.bytes.size(), bufferSize));
+        return SizeOrError { xattrResult.bytes.size(), 0 };
+    }
+    else
+    {
+        return SizeOrError { 0, xattrResult.error };
+    }
 }
 
 vnode_t vnode_getparent(vnode_t vnode)
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
@@ -142,13 +142,6 @@ shared_ptr<vnode> vnode::Create(const shared_ptr<mount>& mount, const char* path
     return result;
 }
 
-shared_ptr<vnode> vnode::Create(const shared_ptr<mount>& mount, const char* path, vtype vnodeType, uint64_t inode)
-{
-    shared_ptr<vnode> result = Create(mount, path, vnodeType);
-    result->inode = inode;
-    return result;
-}
-
 void vnode::StartRecycling()
 {
     s_vnodesByPath.erase(this->path);
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
@@ -38,6 +38,15 @@ vnode::vnode(const shared_ptr<mount>& mount) :
 {
 }
 
+vnode::vnode(const std::shared_ptr<mount>& mount, VnodeCreationProperties properties) :
+    mountPoint(mount),
+    name(nullptr),
+    inode(properties.inode == UINT64_MAX ? mount->nextInode++ : properties.inode),
+    type(properties.type),
+    parent(properties.parent)
+{
+}
+
 vnode::~vnode()
 {
     assert(this->ioCount == 0);
@@ -107,6 +116,20 @@ shared_ptr<vnode> mount::CreateVnodeTree(const string& path, vtype vnodeType)
     }
     
     return fileVnode;
+}
+
+std::shared_ptr<vnode> mount::CreateVnode(std::string path, VnodeCreationProperties properties)
+{
+    uint64_t inode = properties.inode;
+    if (inode == UINT64_MAX)
+    {
+        inode = this->nextInode++;
+    }
+    shared_ptr<vnode> newVnode(new vnode(this->weakSelfPointer.lock(), properties));
+    s_allVnodes.insert(make_pair(newVnode.get(), weak_ptr<vnode>(newVnode)));
+    newVnode->weakSelfPointer = newVnode;
+    newVnode->SetAndRegisterPath(path);
+    return newVnode;
 }
 
 shared_ptr<vnode> vnode::Create(const shared_ptr<mount>& mount, const char* path, vtype vnodeType)
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
@@ -18,10 +18,12 @@ static WeakVnodeMap s_allVnodes;
 shared_ptr<mount> mount::Create(const char* fileSystemTypeName, fsid_t fsid, uint64_t initialInode)
 {
     shared_ptr<mount> result(new mount{});
+    result->weakSelfPointer = result;
     assert(strlen(fileSystemTypeName) + 1 < sizeof(result->statfs.f_fstypename));
     result->statfs.f_fsid = fsid;
     result->nextInode = initialInode;
     strlcpy(result->statfs.f_fstypename, fileSystemTypeName, sizeof(result->statfs.f_fstypename));
+    
     return result;
 }
 
@@ -36,6 +38,72 @@ vnode::vnode(const shared_ptr<mount>& mount) :
 vnode::~vnode()
 {
     assert(this->ioCount == 0);
+}
+
+static string ParentPathString(const string& path)
+{
+    assert(path.length() > 0);
+    size_t lastSlashPos = path.find_last_of('/');
+    assert(lastSlashPos != string::npos);
+    if (lastSlashPos == 0)
+    {
+        return "/";
+    }
+    else if (lastSlashPos == path.length() - 1) // path ends in "/"
+    {
+        lastSlashPos = path.find_last_of('/', lastSlashPos - 1);
+    }
+    
+    return path.substr(0, lastSlashPos);
+}
+
+// This creates a vnode at the given path below the mount point, and ensures
+// that a hierarchy of directory vnodes exists all the way to the root.
+//
+// For example:
+// mount->CreateVnodeTree("/path/to/file", VREG) will produce:
+// vnode VREG "/path/to/file"
+// vnode VDIR "/path/to"
+// vnode VDIR "/path"
+// vnode VDIR "/"
+// While walking up the path, we stop creating vnodes if one already exists for
+// a particular directory. For example, if the example above is followed by:
+// mount->CreateVnodeTree("/path/for/other/file", VREG) will only generate:
+// vnode VREG "/path/for/other/file"
+// vnode VDIR "/path/for/other"
+// vnode VDIR "/path/for"
+// Because the vnode generated for "/path" in the previous example will be
+// re-used as the parent for the "/path/for" vnode.
+shared_ptr<vnode> mount::CreateVnodeTree(const string& path, vtype vnodeType)
+{
+    assert(path[0] == '/'); // Only absolute paths allowed
+    shared_ptr<vnode> parentVnode;
+    
+    if (path != "/")
+    {
+        string parentPath = ParentPathString(path);
+        PathToVnodeMap::const_iterator found = s_vnodesByPath.find(parentPath);
+    
+        if (found != s_vnodesByPath.end())
+        {
+            parentVnode = found->second.lock();
+        }
+        
+        if (!parentVnode)
+        {
+            parentVnode = this->CreateVnodeTree(parentPath, VDIR);
+        }
+    }
+    
+    shared_ptr<vnode> fileVnode = vnode::Create(this->weakSelfPointer.lock(), path.c_str(), vnodeType);
+    fileVnode->parent = parentVnode;
+    
+    if (path == "/")
+    {
+        this->rootVnode = fileVnode;
+    }
+    
+    return fileVnode;
 }
 
 shared_ptr<vnode> vnode::Create(const shared_ptr<mount>& mount, const char* path, vtype vnodeType)
@@ -61,6 +129,7 @@ void vnode::StartRecycling()
     this->path.clear();
     this->type = VBAD;
     this->vid++;
+    this->parent.reset();
     this->isRecycling = true;
 }
 
@@ -238,4 +307,27 @@ void MockVnodes_CheckAndClear()
     }
     
     s_allVnodes.clear();
+}
+
+
+SizeOrError Vnode_ReadXattr(vnode_t _Nonnull vnode, const char* _Nonnull xattrName, void* _Nullable buffer, size_t bufferSize)
+{
+    assert(false); // TODO: implement
+    return SizeOrError{};
+}
+
+vnode_t vnode_getparent(vnode_t vnode)
+{
+    shared_ptr<struct vnode> parentVnode = vnode->GetParentVnode();
+    if (parentVnode)
+    {
+        parentVnode->RetainIOCount();
+    }
+    
+    return parentVnode.get();
+}
+
+int vnode_isvroot(vnode_t vnode)
+{
+    return vnode->GetMountPoint()->GetRootVnode().get() == vnode;
 }
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
@@ -8,6 +8,13 @@
 #include <vector>
 #include <unordered_map>
 
+struct VnodeCreationProperties
+{
+    vtype type = VREG;
+    uint64_t inode = UINT64_MAX; // auto-generate
+    std::shared_ptr<vnode> parent;
+};
+
 // The struct names mount and vnode are dictated by mount_t and vnode_t's
 // definitions as (opaque/forward declared) pointers to those structs.
 // As the (testable) kext treats them as entirely opaque, we can implement
@@ -25,6 +32,7 @@ public:
     static std::shared_ptr<mount> Create(const char* fileSystemTypeName, fsid_t fsid, uint64_t initialInode);
     
     std::shared_ptr<vnode> CreateVnodeTree(const std::string& path, vtype vnodeType = VREG);
+    std::shared_ptr<vnode> CreateVnode(std::string path, VnodeCreationProperties properties);
     
     fsid_t GetFsid() const { return this->statfs.f_fsid; }
     std::shared_ptr<vnode> GetRootVnode() const { return this->rootVnode.lock(); }
@@ -58,7 +66,8 @@ private:
     void SetAndRegisterPath(const std::string& path);
 
     explicit vnode(const std::shared_ptr<mount>& mount);
-    
+    explicit vnode(const std::shared_ptr<mount>& mount, VnodeCreationProperties properties);
+
     vnode(const vnode&) = delete;
     vnode& operator=(const vnode&) = delete;
     
--- a/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
@@ -8,10 +8,14 @@
 #include <vector>
 #include <unordered_map>
 
+// An aggregate for setting properties of newly created mock vnodes with sensible defaults.
 struct VnodeCreationProperties
 {
     vtype type = VREG;
-    uint64_t inode = UINT64_MAX; // auto-generate
+    // UINT64_MAX is a special value that causes the associated mount point to
+    // automatically assign an inode. Set this field to something else to
+    // explicitly choose an inode number.
+    uint64_t inode = UINT64_MAX;
     std::shared_ptr<vnode> parent;
 };
 
@@ -32,7 +36,10 @@ public:
     static std::shared_ptr<mount> Create(const char* fileSystemTypeName, fsid_t fsid, uint64_t initialInode);
     
     std::shared_ptr<vnode> CreateVnodeTree(const std::string& path, vtype vnodeType = VREG);
-    std::shared_ptr<vnode> CreateVnode(std::string path, VnodeCreationProperties properties);
+    // By default, CreateVnode() will create a regular file with an
+    // auto-assigned inode and no existing parent vnode.
+    // See struct VnodeCreationProperties for changing this.
+    std::shared_ptr<vnode> CreateVnode(std::string path, VnodeCreationProperties properties = VnodeCreationProperties{});
     
     fsid_t GetFsid() const { return this->statfs.f_fsid; }
     std::shared_ptr<vnode> GetRootVnode() const { return this->rootVnode.lock(); }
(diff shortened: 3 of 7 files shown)


pmj:
I've just pushed an extra commit, 4be8d4f, which effectively reverts the behaviour to the one causing bug #797, and as you can see, this fails the unit tests. It's not quite as simple as reverting commit 0b841d8 as the fix involved an API change, and the unit tests are of course written against the new API. Obviously the "reverting" commit is only for demonstration purposes, the tests all pass for ffa164c which is the one I'd be merging.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -144,7 +144,9 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
     
     VirtualizationRootHandle rootHandle = RootHandle_None;
     errno_t error = 0;
-    
+
+    FsidInode vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context, false);
```
jeschu1: Quick marker that this was to demonstrate the unit test failure and will be moved back.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,68 @@ vnode::vnode(const shared_ptr<mount>& mount) :
 {
 }
 
+vnode::vnode(const std::shared_ptr<mount>& mount, VnodeCreationProperties properties) :
```
jeschu1: We're pretty confident this will grow?  Looks like creation properties are only 3 parameters right now.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,68 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+
 vnode::~vnode()
 {
     assert(this->ioCount == 0);
 }
 
+static string ParentPathString(const string& path)
+{
+    assert(path.length() > 0);
+    size_t lastSlashPos = path.find_last_of('/');
```
jeschu1: assert(lastSlashPos != -1);

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,68 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+    
+    return path.substr(0, lastSlashPos);
+}
+
+shared_ptr<vnode> mount::CreateVnodeTree(string path, vtype vnodeType)
+{
+    assert(path[0] == '/'); // Only absolute paths allowed
+    string parentPath = ParentPathString(path);
+    PathToVnodeMap::const_iterator found = s_vnodesByPath.find(parentPath);
+    shared_ptr<vnode> parentVnode = path == "/" ? nullptr : (found != s_vnodesByPath.end() ? found->second.lock() : nullptr) ?: this->CreateVnodeTree(parentPath, VDIR);
```
jeschu1: While the above line is very concise, it's not easy (for me at least) to read.  Thoughts on making it more verbose for clarity?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,68 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+    }
+    else if (lastSlashPos == path.length() - 1) // path ends in "/"
+    {
+        lastSlashPos = path.find_last_of('/', lastSlashPos - 1);
+    }
+    
+    return path.substr(0, lastSlashPos);
+}
+
+shared_ptr<vnode> mount::CreateVnodeTree(string path, vtype vnodeType)
```
jeschu1: It would be super helpful if we had an example in comments.  I.e. you send in path x, we create a tree that looks y.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -239,3 +302,45 @@ void MockVnodes_CheckAndClear()
...
+    else
+    {
+        return make_pair(0, found->second);
+    }
+}
+
+SizeOrError Vnode_ReadXattr(vnode_t _Nonnull vnode, const char* _Nonnull xattrName, void* _Nullable buffer, size_t bufferSize)
+{
+    pair<errno_t, vector<uint8_t>> xattrResult = vnode->ReadXattr(xattrName);
+    if (xattrResult.first == 0)
```
jeschu1: Making sure I follow here... does .first == 0 mean no error so copy attribute value and return?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,116 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
 
 }
 
+- (void)testFindForVnode_FileInRoot
+{
+    const char* repoPath = "/Users/test/code/Repo";
+    const char* filePath = "/Users/test/code/Repo/file";
```
jeschu1: Worth adding a case for a file with a deeper path?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,116 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
...
+
+- (void)testFindForVnode_FileInUndetectedRoot
+{
+    const char* repoPath = "/Users/test/code/Repo";
+    const char* filePath = "/Users/test/code/Repo/file";
+    
+    shared_ptr<vnode> repoRootVnode = self->testMountPoint->CreateVnodeTree(repoPath, VDIR);
+    shared_ptr<vnode> testFileVnode = self->testMountPoint->CreateVnodeTree(filePath);
+
+    PrjFSVirtualizationRootXAttrData rootXattr = {};
```
jeschu1: Is this to test the case when InsertVirtualizationRoot_Locked was not called?  When would that happen?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: I have a few issues with the existing approach:

 * There's not a good way to pick & mix the arguments you care about; default arguments only work right-to-left. The combinations I ran into were pretty arbitrary however.
 * With lots of arguments, it becomes non-obvious which one means what.
 * As you imply, it doesn't scale very well.

The C99 style named struct field initialisers effectively give us named function arguments, which makes it obvious what each one does. For an example [look here](https://github.com/Microsoft/VFSForGit/pull/908/files#diff-16401a67e162da73f89646a0727af85dR350).

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: Yes; perhaps I should define a struct instead of using `pair<>` here.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: Yes, essentially the case where a particular virtualisation root hasn't been mounted (no active provider) and is discovered by the kext for the first time based on its special xattr.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: I feel like the other tests, such as `testFindForVnode_DirectoryInUndetectedRoot` already cover that case.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: I've removed that commit in the latest push.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: Yeah, that was a rats' nest. I've cleaned it up.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: Done.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: I've solved this with a struct now, please take a look and let me know if that makes it clearer.

pmj:
With the latest push, this is now ready for another round of review.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,100 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+    shared_ptr<vnode> parentVnode;
+    
+    if (path != "/")
+    {
+        string parentPath = ParentPathString(path);
+        PathToVnodeMap::const_iterator found = s_vnodesByPath.find(parentPath);
+    
+        if (found != s_vnodesByPath.end())
+        {
+            parentVnode = found->second.lock();
```
jeschu1: Under what conditions would we find the path, but vnode would null?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,100 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+        this->rootVnode = fileVnode;
+    }
+    
+    return fileVnode;
+}
+
+std::shared_ptr<vnode> mount::CreateVnode(std::string path, VnodeCreationProperties properties)
+{
+    uint64_t inode = properties.inode;
+    if (inode == UINT64_MAX)
```
jeschu1: if they pass in UINT64_MAX we should use nextInode?  Is UINIT64_MAX the default value for inode?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,119 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
...
+    vector<uint8_t> rootXattrData(sizeof(rootXattr), 0x00);
+    memcpy(rootXattrData.data(), &rootXattr, rootXattrData.size());
+    repoRootVnode->xattrs.insert(make_pair(PrjFSVirtualizationRootXAttrName, rootXattrData));
+
+    VirtualizationRootHandle foundRoot = VirtualizationRoot_FindForVnode(&self->dummyTracer, PrjFSPerfCounter_VnodeOp_FindRoot, PrjFSPerfCounter_VnodeOp_FindRoot_Iteration, testSubdirVnode.get(), self->dummyVFSContext);
+    
+    XCTAssertTrue(VirtualizationRoot_IsValidRootHandle(foundRoot));
+    if (VirtualizationRoot_IsValidRootHandle(foundRoot))
```
jeschu1: We've already asserted this is true, so no 'if' necessary

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,119 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
...
+    {
+        XCTAssertEqual(s_virtualizationRoots[foundRoot].rootInode, repoRootVnode->GetInode());
```
jeschu1: Should we also double check the rootInode on the File version of this test? Should we do that on any other tests?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: When the last strong reference (`shared_ptr`) to the vnode goes away, the vnode is destroyed. Any remaining weak references (`weak_ptr`) to that vnode will return null references when attempting to convert them to a strong reference with `lock()`.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: [There is a comment for this in the `VnodeCreationProperties` struct definition](https://github.com/Microsoft/VFSForGit/pull/908/files#diff-511f4078d7e454d8d396f5970eded57bR14). Too cryptic?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: Unfortunately, it is. If `XCTAssert*`s fail, they don't immediately terminate the current test, but keep going. If `foundRoot` here is not a valid handle (e.g. `RootHandle_None`, i.e. -1) we must not use it to look up entries in the array of roots because we'll go out of bounds. Hence the `if` block.

jeschu1: Ah interesting!  Generally tests stop once they hit an assert.  I wonder if that a purposeful change in `XCTAssert `or if there is another method for it.  Either way in that case the 'if' makes sense.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: The recycling test, which uses a file, [also checks this](https://github.com/Microsoft/VFSForGit/pull/908/files#diff-16401a67e162da73f89646a0727af85dR349) - I think that pretty much covers our bases. I'm not sure there's much value in asserting the kitchen sink in each test, my philosophy here is to stick to what's necessary to make the tests easier to read - when looking at the test I'd like it to be reasonably clear what failure case(s) it's protecting us from, which gets muddied when we test all possible invariants.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: Yeah, I don't know why they designed it like that. I'm not aware of a non-continuing assertion in XCTest.framework, and didn't immediately spot any in `XCTAssertions.h`.

wilbaker: Not something for this PR, but do you think it would be worth defining our own assert for unit tests that does return immediately if the assert fails?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
wilbaker: > essentially the case where a particular virtualisation root hasn't been mounted (no active provider) and is discovered by the kext for the first time based on its special xattr.

Could you add this in a comment at the start of the test?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,119 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
...
+
+    XCTAssertTrue(VirtualizationRoot_IsValidRootHandle(foundRoot));
+    
+    if (VirtualizationRoot_IsValidRootHandle(foundRoot))
+    {
+        XCTAssertEqual(s_virtualizationRoots[foundRoot].rootInode, repoRootVnode->GetInode());
+        
+        repoRootVnode->StartRecycling();
+        
+        shared_ptr<vnode> newRepoRootVnode = self->testMountPoint->CreateVnode(repoPath, VnodeCreationProperties { .inode = repoRootVnode->GetInode(), .type = VDIR });
```
wilbaker: Why is `newRepoRootVnode` using the same inode as `repoRootVnode` with a different path?  Is this to test a hardlink scenario? (In general I'm not sure what how/why recycling `repoRootVnode` has any impact on `newRepoRootVnode`).

On a somewhat related note, I noticed that `Vnode_GetFsidAndInode` is ignoring the `useLinkIDForInode` parameter we're passing in.  Do you remember if there were plans to change what value is used for inode based on `useLinkIDForInode` ?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,119 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
...
+    shared_ptr<vnode> testFileVnode = self->testMountPoint->CreateVnodeTree(filePath);
+
+    PrjFSVirtualizationRootXAttrData rootXattr = {};
+    vector<uint8_t> rootXattrData(sizeof(rootXattr), 0x00);
+    memcpy(rootXattrData.data(), &rootXattr, rootXattrData.size());
+    repoRootVnode->xattrs.insert(make_pair(PrjFSVirtualizationRootXAttrName, rootXattrData));
+
+    VirtualizationRootHandle foundRoot = VirtualizationRoot_FindForVnode(&self->dummyTracer, PrjFSPerfCounter_VnodeOp_FindRoot, PrjFSPerfCounter_VnodeOp_FindRoot_Iteration, testFileVnode.get(), self->dummyVFSContext);
+    
+    XCTAssertTrue(VirtualizationRoot_IsValidRootHandle(foundRoot));
```
wilbaker: This might be outside the scope of this PR, but could we also call `VirtualizationRoot_GetActiveProvider` and validate that `isOnline` is false?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -264,4 +271,119 @@ - (void)testRegisterProviderForPath_ExistingRecycledRoot
...
+    
+    shared_ptr<vnode> repoRootVnode = self->testMountPoint->CreateVnodeTree(repoPath, VDIR);
+    shared_ptr<vnode> testSubdirVnode = self->testMountPoint->CreateVnodeTree(subdirPath, VDIR);
+
+    XCTAssertTrue(vnode_isdir(testSubdirVnode.get()));
+
+    PrjFSVirtualizationRootXAttrData rootXattr = {};
+    vector<uint8_t> rootXattrData(sizeof(rootXattr), 0x00);
+    memcpy(rootXattrData.data(), &rootXattr, rootXattrData.size());
+    repoRootVnode->xattrs.insert(make_pair(PrjFSVirtualizationRootXAttrName, rootXattrData));
```
wilbaker: Minor, I noticed that this block of code:

```
    PrjFSVirtualizationRootXAttrData rootXattr = {};
    vector<uint8_t> rootXattrData(sizeof(rootXattr), 0x00);
    memcpy(rootXattrData.data(), &rootXattr, rootXattrData.size());
    repoRootVnode->xattrs.insert(make_pair(PrjFSVirtualizationRootXAttrName, rootXattrData));
```

Is repeated in a few tests, should it be moved to a helper function?

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: I think that does fall outside the scope of this PR, I've created #934 so I remember.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: > Why is `newRepoRootVnode` using the same inode as `repoRootVnode` with a different path?

As far as I can tell, and as far as intended, the paths are the same (`repoPath` in both instances):

```
shared_ptr<vnode> repoRootVnode = self->testMountPoint->CreateVnodeTree(repoPath, VDIR);
//…
shared_ptr<vnode> newRepoRootVnode = self->testMountPoint->CreateVnode(repoPath, VnodeCreationProperties { .inode = repoRootVnode->GetInode(), .type = VDIR });
```

> On a somewhat related note, I noticed that Vnode_GetFsidAndInode is ignoring the useLinkIDForInode parameter we're passing in. Do you remember if there were plans to change what value is used for inode based on useLinkIDForInode?

I'm planning to switch to the "real" `Vnode_GetFsidAndInode()` eventually and allow that to call down to `vnode_getattr()`. The distinction has so far not been important to the tests, and I've been doing the minimum reasonably possible to support the existing tests rather than building too far ahead and making these test PRs even bigger than they already are.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,100 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+    
+    if (path == "/")
+    {
+        this->rootVnode = fileVnode;
+    }
+    
+    return fileVnode;
+}
+
+std::shared_ptr<vnode> mount::CreateVnode(std::string path, VnodeCreationProperties properties)
```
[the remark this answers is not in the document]
pmj: In hindsight, I think `mount::CreateVnode` is probably the better pattern, so I'm thinking I'll eventually get rid of the (public) `vnode::Create` functions.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -16,11 +25,17 @@ struct mount
 private:
     vfsstatfs statfs;
     uint64_t nextInode;
-    
+    std::weak_ptr<mount> weakSelfPointer;
+    std::weak_ptr<vnode> rootVnode;
```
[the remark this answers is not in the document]
pmj: > What's the reason for using a `weak_ptr` for `rootVnode` rather than a `shared_ptr`?

This way, we can have a mock mount point as a test class instance variable and don't have to manually clear out the `rootVnode` after every test.

Generally, the idea is that mock `vnode` objects are all destroyed automatically by the end of each test because their `shared_ptr` references are all: (1) local to the test or (2) transitive (the `parent` reference in `vnode`).

If they aren't, that should indicate a bug, and is immediately caught by the assertion in `MockVnodes_CheckAndClear()`.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
jeschu1: That makes sense to me!  

For my understanding why do we need the check here and not in testFindForVnode_FileInUndetectedRoot?  Or is it just a matter of this one does the check so the other one shouldn't need to verify it.

pmj: * `testFindForVnode_FileInUndetectedRoot` Generally checks that we can discover a previously unknown root by walking up the directory hierarchy and encountering the root, which has a special xattr.
 * `testFindForVnode_DetectRootRecycleThenFindRootForOtherFile` specifically tests against a regression to #797 - that bug manifested itself as the wrong inode being assigned to a root when discovering a previously unknown root via the xattr. So that's why that test checks the inode is correct. Additionally it checks that the inode-based detection works after the original root vnode was recycled. (Which is the most common situation in which #797 actually became noticeable.)
 * `testFindForVnode_DirectoryInUndetectedRoot` Is kind of a variation on parts of both of the above tests with a directory vnode rather than a regular file. It's specifically designed to cover the `if (vnode_isdir(vnode))` branch in `VirtualizationRoot_FindForVnode`. We probably don't *need* to check the inode here (#797 affected directories and regular files equally), but it doesn't hurt either, just in case taking that branch (which was newly introduced with the #797 fix) had any effect on the inode in particular.

If anything, `testFindForVnode_FileInUndetectedRoot` is superfluous, because the other two tests would fail for anything that might cause that test to fail. But I like having tests of escalating complexity for a lack of better term for it. Test some really basic scenarios as an initial sanity check, then test some very specific failure scenarios or variations.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: I've filed #936 for this and a bunch of other, similar `assert()` related problems.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: I've now added comments to all the new tests explaining what they're actually trying to test.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,11 +38,100 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+{
+    uint64_t inode = properties.inode;
+    if (inode == UINT64_MAX)
+    {
+        inode = this->nextInode++;
+    }
+    shared_ptr<vnode> newVnode(new vnode(this->weakSelfPointer.lock(), properties));
+    s_allVnodes.insert(make_pair(newVnode.get(), weak_ptr<vnode>(newVnode)));
+    newVnode->weakSelfPointer = newVnode;
+    newVnode->SetPath(path);
```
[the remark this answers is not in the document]
pmj: I've renamed it in the latest push.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
pmj: Thanks for the suggestion, I've done exactly that in this latest push.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
pmj: Commit 6718456 adds a better explanation for the function and the property struct.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -5,6 +5,15 @@
 #include "../PrjFSKext/public/FsidInode.h"
 #include <memory>
 #include <string>
+#include <vector>
+#include <unordered_map>
+
+struct VnodeCreationProperties
+{
+    vtype type = VREG;
+    uint64_t inode = UINT64_MAX; // auto-generate
```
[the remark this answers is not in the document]
pmj: Commit 6718456 adds a better explanation for the creation function and the property struct.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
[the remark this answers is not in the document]
pmj: Commit 6718456 now removes one of the old `vnode::Create()` overloads and replaces all of its uses with `mount::CreateVnode`. I think I'll create another overload of `mount::CreateVnode` which will be a little less verbose for the other, much more common `vnode::Create()` variant, but I'll do that in a future PR.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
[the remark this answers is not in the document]
wilbaker: Thanks all that makes sense!

If you don't mind, I think documenting this could be as simple as added a comment to the top of MockVnodeAndMount.hpp that includes what you've described here.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
wilbaker: >As far as I can tell, and as far as intended, the paths are the same

D'oh, I must have missed that in my earlier review, thanks!

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
wilbaker: I see how `testFindForVnode_DirectoryInUndetectedRoot ` is a variation of `testFindForVnode_FileInUndetectedRoot`, but I'm not sure what part of it is a variation of `testFindForVnode_DetectRootRecycleThenFindRootForOtherFile`.

Is it because of this check?

```
XCTAssertEqual(s_virtualizationRoots[foundRoot].rootInode, repoRootVnode->GetInode());
```

Or is it that in both of the above tests we're discovering the root by its xattr?

(I think the test looks good, my question is just about how the test is related to `testFindForVnode_DetectRootRecycleThenFindRootForOtherFile`).

pmj: > Is it because of this check?

Yes, like `testFindForVnode_DetectRootRecycleThenFindRootForOtherFile` it checks that `VirtualizationRoot_FindForVnode()` sets the correct inode in the detected root, but it checks this on the `if (vnode_isdir(vnode))` path of the function. I didn't think the rest of `testFindForVnode_DetectRootRecycleThenFindRootForOtherFile` (the recycling bit) really warranted repeating.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,34 +38,117 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+// vnode VDIR "/"
+// While walking up the path, we stop creating vnodes if one already exists for
+// a particular directory. For example, if the example above is followed by:
+// mount->CreateVnodeTree("/path/for/other/file", VREG) will only generate:
+// vnode VREG "/path/for/other/file"
+// vnode VDIR "/path/for/other"
+// vnode VDIR "/path/for"
+// Because the vnode generated for "/path" in the previous example will be
+// re-used as the parent for the "/path/for" vnode.
+shared_ptr<vnode> mount::CreateVnodeTree(string path, vtype vnodeType)
```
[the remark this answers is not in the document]
pmj: Good point, I forgot that C++ would implicitly create temporary `std::string` rvalues that could  be treated as `const std::string&` when passing a `const char*` argument. I've fixed this locally now, it'll be included in the next (and hopefully final) revision of this PR.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -33,34 +38,117 @@ vnode::vnode(const shared_ptr<mount>& mount) :
...
+        inode = this->nextInode++;
+    }
+    shared_ptr<vnode> newVnode(new vnode(this->weakSelfPointer.lock(), properties));
+    s_allVnodes.insert(make_pair(newVnode.get(), weak_ptr<vnode>(newVnode)));
+    newVnode->weakSelfPointer = newVnode;
+    newVnode->SetAndRegisterPath(path);
+    return newVnode;
+}
+
 shared_ptr<vnode> vnode::Create(const shared_ptr<mount>& mount, const char* path, vtype vnodeType)
```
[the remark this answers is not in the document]
pmj: Yep, you're right; I'm going to leave it for `vnode::Create` for now as I want to replace that function with a `mount::CreateVnode()` overload anyway, but I'll do that in the next PR with vnode based unit tests.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -25,6 +27,13 @@ void ProviderUserClient_UpdatePathProperty(PrjFSProviderUserClient* userClient,
     MockCalls::RecordFunctionCall(ProviderUserClient_UpdatePathProperty, userClient, providerPath);
 }
 
+static void MockVnodeSetRootXattrData(shared_ptr<vnode> vnode)
```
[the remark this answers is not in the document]
pmj: Done in this latest version!

PR #909, opened 2019-03-10
Mac kext: Move provider messaging to separate source file

pmj:
I'm about to do some work in this area for the remaining (`vnode_getparent`) aspect of #338, so this seemed about as good a time as any to sort out something that's been bugging me for a while.

KauthHandler.cpp has thus far covered 2 main areas of responsibility:

 * The handler callback functions for kauth listeners (as the name suggests)
 * Sending & waiting for messages to providers

The 2 blocks of functionality are almost entirely independent from one another, and as KauthHandler.cpp is a large file, it makes more sense to move the messaging aspect to a separate source file.

There are no functional changes. A few functions have had their prefix changed or turned from static to global.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -16,6 +16,7 @@
 #include "PerformanceTracing.hpp"
 #include "kernel-header-wrappers/mount.h"
 #include "KextLog.hpp"
+#include "ProviderMessaging.hpp"
 
 #ifdef KEXT_UNIT_TESTING
 #include "KauthHandlerTestable.hpp"
@@ -49,18 +50,7 @@ KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mas
 
 KEXT_STATIC bool IsFileSystemCrawler(const char* procname);
 
-static void Sleep(int seconds, void* channel, Mutex* _Nullable mutex);
-static bool TrySendRequestAndWaitForResponse(
-    VirtualizationRootHandle root,
-    MessageType messageType,
-    const vnode_t vnode,
-    const FsidInode& vnodeFsidInode,
-    const char* vnodePath,
-    int pid,
-    const char* procname,
-    int* kauthResult,
-    int* kauthError);
-static void AbortAllOutstandingEvents();
+static void WaitForListenerCompletion();
 KEXT_STATIC bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode);
 
 static bool ShouldHandleVnodeOpEvent(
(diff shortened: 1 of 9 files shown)


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -1081,19 +869,10 @@ static void AbortAllOutstandingEvents()
     // https://developer.apple.com/library/archive/samplecode/KauthORama/Listings/KauthORama_c.html#//apple_ref/doc/uid/DTS10003633-KauthORama_c-DontLinkElementID_3
     do
     {
-        Sleep(1, nullptr, nullptr);
+        Mutex_Sleep(1, nullptr, nullptr);
```
jeschu1: For my own understanding can you explain this change?

jeschu1: I like the new name here :-)

wilbaker:
> @wilbaker could you give it a quick scan and see if you find any issues.

Sure I'll take a look at this next

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -964,113 +857,8 @@ static bool ShouldHandleFileOpEvent(
...
-        
-        OutstandingMessage* outstandingMessage;
-        LIST_FOREACH(outstandingMessage, &s_outstandingMessages, _list_privates)
-        {
-            wakeup(outstandingMessage);
-        }
-    }
-    Mutex_Release(s_outstandingMessagesMutex);
-    
     // ... and wait until all kauth events have noticed and returned.
```
wilbaker: Minor: The comment starting with `// ... and wait` doesn't make as much sense as before.  Maybe updated it to simply "Wait until all kauth events have noticed and returned."

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
```diff
@@ -0,0 +1,234 @@
...
+                *kauthResult = KAUTH_RESULT_DENY;
+            }
+            else if (MessageType_Response_Success == message.result)
+            {
+                *kauthResult = KAUTH_RESULT_DEFER;
+                result = true;
+            }
+            else
+            {
+                KextLog_Error("TrySendRequestAndWaitForResponse; message.result = %d", message.result);
```
wilbaker: (This `KextLog_Error` looks new)

If my understanding is correct, this scenario is when the provider returns an error code from a callback?

If that's the case, I don't think we should be logging this as an error because it's a supported/expected scenario.

pmj: Ah, good catch, looks like a local debugging commit snuck in there. I'll make sure to pull that back out.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yep, you're right, that doesn't make much sense. Now fixed as suggested.

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
pmj: Just to confirm I've just done a diff of the old `KauthHandler.cpp` and the new `ProviderMessaging.cpp` and that was the only change that snuck into the moved functions apart from the function name changes.

PR #910, opened 2019-03-11
[TESTING/WIP/RFC] Mac kext: Improve vnode_getparent reliability using path based approach

pmj:
The idea is to provide the final pieces of the #338 puzzle.

If `vnode_getparent()` fails, this means the corresponding vnode has dropped out of the system, presumably because it hasn't been used for a while, or because the file has actually been unlinked.

In the former case, we can work around the problem by attempting to obtain the path to the vnode, most likely by calling out to a provider which will call `fsgetpath()` using the fsid/inode pair.

**Sub-tasks:**

 - [x] Enable provider user clients to send arbitrary data along with responses to kernel messages.
 - [x] Implement the path-based vnode parent lookup logic.
 - [x] Implement callout to the provider to use `fsgetpath()` from user space when `vn_getpath()` fails in the kernel.
 - [ ] Unit tests covering the new functionality
 - [ ] Integration tests verifying that the new functionality works in the context of virtualisation root lookup.
 - [x] Fix root lookup for case where virtualisation root coincides with the filesystem root.
 - [ ] Unit tests for the filesystem-root-as-virtualisation-root case.

**Reviewer notes:**

 * If some of this code looks familiar, I reused and updated the provider-based path lookup and messaging code from the first iteration of #741.
 * This PR builds on the test infrastructure added with #908, and the refactoring in #909, so the combined diff shown by GitHub will not be helpful. Only commits past c6540e8 are actually covered by this PR.


pmj:
One problem with this approach in general is that it doesn't work if no provider is connected. We could turn the log daemon into a "dummy provider" which only services path queries. Thoughts?

pmj:
The functional tests ended up stalling with the injected errors; the logs weren't helpful, but after running the functional tests on a local system, I figured out that this was caused by the confluence of 2 circumstances:

 * `fsgetpath()` was failing because the file in question was being deleted and we were racing with that.
 * Somewhere between the provider and `TrySendRequestAndWaitForResponse`, the responses to the failed path requests are getting stuck. I'll need to debug where exactly this is going wrong and will follow up with a fix.

Manual testing didn't catch this because I guess the builds I tested didn't delete much.

pmj:
Looks like that issue is a deadlock resulting from `fsgetpath()`'s vnode auth `KAUTH_VNODE_SEARCH` callback causing a directory enumeration run. I'll need to prevent *any* provider process from causing enumerations.

wilbaker:

PR #913, opened 2019-03-12
Additional Kext Tests

jeschu1:
Added Tests for 
- TryGetFileIsFlaggedAsInRoot
- ShouldHandleVnodeOpEvent
- VirtualizationRoot_VnodeIsOnAllowedFilesystem
- VirtualizationRoot_IsValidRootHandle

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
@@ -12,5 +12,6 @@
 
 KEXT_STATIC_INLINE bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
 KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
+KEXT_STATIC_INLINE bool TryGetFileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t context, bool* flaggedInRoot);
 KEXT_STATIC bool IsFileSystemCrawler(const char* procname);
 KEXT_STATIC bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode);
--- a/ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
@@ -1,6 +1,7 @@
 #include "public/PrjFSCommon.h"
 #include <sys/kernel_types.h>
 #include "../PrjFSKext/kernel-header-wrappers/vnode.h"
+#include "../PrjFSKext/PerformanceTracing.hpp"
 
 #ifndef __cplusplus
 #error None of the kext code is set up for being called from C or Objective-C; change the including file to C++ or Objective-C++
@@ -15,3 +16,17 @@ KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mas
 KEXT_STATIC_INLINE bool TryGetFileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t context, bool* flaggedInRoot);
 KEXT_STATIC bool IsFileSystemCrawler(const char* procname);
 KEXT_STATIC bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode);
+KEXT_STATIC bool ShouldHandleVnodeOpEvent(
+    // In params:
+    PerfTracer* perfTracer,
+    vfs_context_t context,
+    const vnode_t vnode,
+    kauth_action_t action,
+
+    // Out params:
+    vtype* vnodeType,
+    uint32_t* vnodeFileFlags,
+    int* pid,
+    char procname[MAXCOMLEN + 1],
+    int* kauthResult,
+    int* kauthError);
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -58,4 +58,29 @@ using std::shared_ptr;
     XCTAssertFalse(ShouldIgnoreVnodeType(static_cast<vtype>(1000), testVnode.get()));
 }
 
+- (void)testFileFlaggedInRoot {
+    bool fileFlaggedInRoot;
+    MountPointer testMount = mount::Create("hfs", fsid_t{}, 0);
+    VnodePointer testVnode = vnode::Create(testMount, "/foo");
+    
+    testVnode->SetAttr(FileFlags_IsInVirtualizationRoot);
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertTrue(fileFlaggedInRoot);
+    
+    testVnode->SetAttr(FileFlags_IsEmpty);
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->SetAttr(FileFlags_Invalid);
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->SetAttr(100);
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->SetGetAttrReturnCode(-1);
+    XCTAssertFalse(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+}
+
 @end
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -60,26 +60,26 @@ using std::shared_ptr;
 
 - (void)testFileFlaggedInRoot {
     bool fileFlaggedInRoot;
-    MountPointer testMount = mount::Create("hfs", fsid_t{}, 0);
-    VnodePointer testVnode = vnode::Create(testMount, "/foo");
+    shared_ptr<mount> testMount = mount::Create();
+    shared_ptr<vnode> testVnode = vnode::Create(testMount, "/foo");
     
-    testVnode->SetAttr(FileFlags_IsInVirtualizationRoot);
+    testVnode->values.getattr = FileFlags_IsInVirtualizationRoot;
     XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
     XCTAssertTrue(fileFlaggedInRoot);
     
-    testVnode->SetAttr(FileFlags_IsEmpty);
+    testVnode->values.getattr = FileFlags_IsEmpty;
     XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
     XCTAssertFalse(fileFlaggedInRoot);
     
-    testVnode->SetAttr(FileFlags_Invalid);
+    testVnode->values.getattr = FileFlags_Invalid;
     XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
     XCTAssertFalse(fileFlaggedInRoot);
     
-    testVnode->SetAttr(100);
+    testVnode->values.getattr = 100;
     XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
     XCTAssertFalse(fileFlaggedInRoot);
-    
-    testVnode->SetGetAttrReturnCode(-1);
+
+    testVnode->errors.getattr = EBADF;
     XCTAssertFalse(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
 }
 
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -1,7 +1,10 @@
 #include "../PrjFSKext/kernel-header-wrappers/vnode.h"
-#import <XCTest/XCTest.h>
 #include "../PrjFSKext/KauthHandlerTestable.hpp"
+#include "../PrjFSKext/PerformanceTracing.hpp"
+#import <XCTest/XCTest.h>
+#import <sys/stat.h>
 #include "MockVnodeAndMount.hpp"
+#include "MockProc.hpp"
 
 using std::shared_ptr;
 
@@ -42,7 +45,7 @@ using std::shared_ptr;
 }
 
 - (void)testShouldIgnoreVnodeType {
-    shared_ptr<mount> testMount = mount::Create("hfs", fsid_t{}, 0);
+    shared_ptr<mount> testMount = mount::Create();
     shared_ptr<vnode> testVnode = testMount->CreateVnode("/foo");
     XCTAssertTrue(ShouldIgnoreVnodeType(VNON, testVnode.get()));
     XCTAssertTrue(ShouldIgnoreVnodeType(VBLK, testVnode.get()));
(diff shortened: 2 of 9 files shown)


ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -1,10 +1,33 @@
...
+    return 1;
+}
+
+extern "C" void proc_name(int pid, char * buf, int size);
+char * procname;
+void proc_name(int pid, char * buf, int size)
+{
+    buf = procname;
+}
+
```
jeschu1: I don't love having the definitions and 3 functions here, but i'm reluctant to move them unless we think we'll use them elsewhere.  
If folks think we should move please also suggest a place/filename.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -1,10 +1,33 @@
 #include "../PrjFSKext/kernel-header-wrappers/vnode.h"
 #import <XCTest/XCTest.h>
 #include "../PrjFSKext/KauthHandlerTestable.hpp"
+#include "../PrjFSKext/PerformanceTracing.hpp"
 #include "MockVnodeAndMount.hpp"
 
 using std::shared_ptr;
 
+#define KAUTH_RESULT_ALLOW    (1)
```
jeschu1: These could also be enums.  Not sure what we gain/lose by one versus another.  If we do make it an enum note to self to initialize at 1.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +81,179 @@ - (void)testShouldIgnoreVnodeType {
...
+    vtype vnodeType;
+    uint32_t vnodeFileFlags;
+    int pid;
+    char procname[MAXCOMLEN + 1] = "";
+    int kauthResult;
+    int kauthError;
+
+    
+    // Test Success Case
+    XCTAssertTrue(ShouldHandleVnodeOpEvent(
```
jeschu1: I tested evey category of failure for this function.  Side note it is super cool that I can do that :-).  Big thanks to @pmj for helping us get this far :-D.

We may decide to test every scenario in every category at a later date.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +81,179 @@ - (void)testShouldIgnoreVnodeType {
...
+    
+    testVnode->values.getattr = FileFlags_IsEmpty;
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->values.getattr = FileFlags_Invalid;
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->values.getattr = 100;
```
pmj: Given this is a bit field, a hex literal might be more appropriate.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +81,179 @@ - (void)testShouldIgnoreVnodeType {
...
+    uint32_t vnodeFileFlags;
+    int pid;
+    char procname[MAXCOMLEN + 1] = "";
+    int kauthResult;
+    int kauthError;
+
+    
+    // Test Success Case
+    XCTAssertTrue(ShouldHandleVnodeOpEvent(
+                                            &perfTracer,
```
pmj: Nit: the indentation is a bit much here, 1 level ought to be enough.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +81,179 @@ - (void)testShouldIgnoreVnodeType {
...
+                                           &vnodeType,
+                                           &vnodeFileFlags,
+                                           &pid,
+                                           procname,
+                                           &kauthResult,
+                                           &kauthError));
+    XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
+
+    
+    // Test invalid File System
```
pmj: What's invalid about this file system?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +81,179 @@ - (void)testShouldIgnoreVnodeType {
...
+    testVnode->values.getattr = FileFlags_IsInVirtualizationRoot;
+    PerfTracer perfTracer;
+    vfs_context_t _Nonnull context = vfs_context_create(NULL);
+    kauth_action_t action = KAUTH_VNODE_READ_DATA;
+    
+    // Out Parameters
+    vtype vnodeType;
+    uint32_t vnodeFileFlags;
+    int pid;
+    char procname[MAXCOMLEN + 1] = "";
```
pmj: This shadows the global variable of the same name, is that intentional?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -1,10 +1,33 @@
...
+int proc_pid(proc_t)
+{
+    return 1;
+}
+
+extern "C" void proc_name(int pid, char * buf, int size);
+char * procname;
+void proc_name(int pid, char * buf, int size)
+{
+    buf = procname;
```
pmj: This essentially a no-op, `buf` is passed by value. You'll need to use `strlcpy()` to write to the supplied buffer. (But then `procname` is never initialised; is this function called by the current tests? If not, perhaps just put an `assert(false)` in here until we actually need to implement it properly.)

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +81,179 @@ - (void)testShouldIgnoreVnodeType {
...
+                                            &vnodeFileFlags,
+                                            &pid,
+                                            procname,
+                                            &kauthResult,
+                                            &kauthError));
+    XCTAssertEqual(kauthResult, KAUTH_RESULT_DENY);
+
+    
+    // Test with finder trying to populate an empty file
+    strcpy(procname, "finder");
```
pmj: Nit: The `"Finder"` process is capitalised, but it shouldn't matter for the purposes of the test.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: It doesn't really matter either way, except if we move it to a header, you might need to surround the definitions with `#ifdef KAUTH_RESULT_ALLOW` to avoid clashing with the kernel headers when building the testable kext.

ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
```diff
@@ -12,5 +13,20 @@
 
 KEXT_STATIC_INLINE bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
 KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
+KEXT_STATIC_INLINE bool TryGetFileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t context, bool* flaggedInRoot);
 KEXT_STATIC bool IsFileSystemCrawler(const char* procname);
 KEXT_STATIC bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode);
+KEXT_STATIC bool ShouldHandleVnodeOpEvent(
+                                          // In params:
```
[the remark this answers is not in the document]
jeschu1: Thanks for bringing this up.

When I autoformat in XCode (Cntrl-I), it does this level of indenting.
I really don't like fighting autoformatting of editors in new code.

Am I correct that this is the autoformat of XCode?  If not, please advise me on how to fix.  If so, can we consider moving to autoformat standards for new code?  @wilbaker

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: > The thing to bear in mind with having them defined here is that they equally apply to other test suites because they're all linked into a single bundle.

Thats a good point to remember.  Moving to MockProc.  We'll get more specific / move in the future if needbe.

ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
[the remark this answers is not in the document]
pmj: Xcode's default indentation rules are unfortunately optimised for Objective-C's named function parameters and tend to cause excessive indentation for C and C++ style function calls. I personally have "Automatically indent based on syntax" disabled (Preferences -> Text Editing -> Indentation); as far as I'm aware, this is the only way to turn off the crazy levels of function argument indentation. Unfortunately, unlike indentation level rules (spaces/tabs, number of spaces) this isn't something you can set on a per-file or per-project basis.

I don't think that level of indentation is in any way helpful, so I'm not keen on changing our style guidelines simply because Xcode isn't designed particularly well for C or C++ editing.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: No it wasn't, fixing..

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: "hfs"

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.cpp
```diff
@@ -104,6 +99,12 @@ void vnode_putname(const char* name)
     // TODO: track name reference counts
 }
 
+int vnode_getattr(vnode_t vp, struct vnode_attr *vap, vfs_context_t ctx)
+{
+    vap->va_flags = vp->values.getattr;
```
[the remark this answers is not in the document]
jeschu1: Great tips here thank you!

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: I switched them to enums so we don't have to worry about clashes

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -18,14 +18,25 @@ struct mount
...
 
+struct VnodeMockErrors
+{
+    errno_t getpath = 0;
+    errno_t getattr = 0;
+};
+
+struct VnodeMockValues
+{
+    int getattr = 0;
```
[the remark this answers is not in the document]
jeschu1: Definitely prefer this as a cleaner approach.  It was not really an int :-), uint32_t as you may have guessed.

I updated for va_flags.

Moving some of the existing fields to attrValues, seems like a meatier change (or i'm missing something).  Can you elaborate?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -1,10 +1,17 @@
...
-#import <XCTest/XCTest.h>
 #include "../PrjFSKext/KauthHandlerTestable.hpp"
+#include "../PrjFSKext/PerformanceTracing.hpp"
+#import <XCTest/XCTest.h>
 #include "MockVnodeAndMount.hpp"
+#include "MockProc.hpp"
 
 using std::shared_ptr;
 
+proc_t vfs_context_proc(vfs_context_t ctx)
```
jeschu1: @pmj this was annoying but when I move this function to MockProc, the linker complained about a dual definition.  Do you see any obvious problems here?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +65,189 @@ - (void)testShouldIgnoreVnodeType {
...
+    XCTAssertTrue(
+        ShouldHandleVnodeOpEvent(
+            &perfTracer,
+            context,
+            testVnode.get(),
+            action,
+            &vnodeType,
+            &vnodeFileFlags,
+            &pid,
+            procname,
```
jeschu1: How vigelent should we be abount freeing memory?
We aren't freeing procname here, but the memory will be returned when the test is over correct?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: Hmm, I don't know what might have been the problem there. I just tried it and it worked straight away:

https://github.com/pmj/VFSForGit/commit/98d80aa41262d47efcb05ffa700f0552951c1675

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: That's [one of the 2 supported file systems](https://github.com/Microsoft/VFSForGit/blob/98acb8c7eaf92b26d514f68bf3f9dc88095ad475/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp#L596) though?

FWIW, this test fails when I run it locally.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: `procname` is a local variable of array type, so it's allocated on the stack and automatically disappears when the function returns, so in this case there's nothing to worry about.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -0,0 +1,23 @@
...
+{
+    if (return_proc_name != nullptr)
+    {
+        strcpy(buf, return_proc_name);
+    }
+}
+
+void SetProcName(char* procName)
+{
+    return_proc_name = strdup(procName);
```
pmj: We should `free()` the previous memory before overwriting the pointer. Or just use a `std::string` for the variable holding the mock value and don't worry about memory.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+void proc_name(int pid, char * buf, int size)
+{
+    if (return_proc_name != nullptr)
+    {
+        strcpy(buf, return_proc_name);
+    }
+}
+
+void SetProcName(char* procName)
```
pmj: This parameter should be a `const char*` so we can use literals.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -0,0 +1,23 @@
+#include "MockProc.hpp"
+#include <string.h>
+
+int proc_pid(proc_t)
+{
+    return 1;
+}
+
+char* return_proc_name;
```
pmj: I don't think we need this variable outside this compilation unit, so it can be `static`. I'm not sure what our naming convention for global variables is, but for file-level `static` variables it's `s_camelCase`.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+void proc_name(int pid, char * buf, int size)
+{
+    if (return_proc_name != nullptr)
+    {
+        strcpy(buf, return_proc_name);
```
pmj: Please always use `strlcpy`, or in some exceptional circumstances `strncpy`. `strcpy` is almost never appropriate, as it doesn't check for enough space in the destination.

[The real `proc_name` implementation](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/kern/kern_proc.c#L874) uses `strlcpy`, so let's definitely use that please.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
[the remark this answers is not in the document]
pmj: > Moving some of the existing fields to attrValues, seems like a meatier change (or i'm missing something). Can you elaborate?

Yes, that'd be a bit bigger. We can fix that up on demand, no need to lump it in with this change. The new version looks good to me!

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: Thanks for that, you are correct here.  Updating.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: Thanks for trying that @pmj, yeah it seems totally fine now.  Perhaps I was working too late :)

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: I copied the pattern you had for Path.  Using string is a nice way to not have to worry about free-ing.

jeschu1:
@pmj thanks for the quick turnaround and pointers here.  My C is rusty for sure, so I definitely appreciate the tips :-).  I addressed your comments in a separate commit for easy review.  I'll squash before merging.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -1,23 +1,24 @@
...
 
-char* return_proc_name;
-
 void proc_name(int pid, char * buf, int size)
 {
-    if (return_proc_name != nullptr)
-    {
-        strcpy(buf, return_proc_name);
-    }
+    strlcpy(buf, s_procName.c_str(), MIN(size, MAXPATHLEN));
```
pmj: Checking for `MAXPATHLEN` isn't really appropriate here. There is a `MAXCOMLEN` limitation in the kernel, but for our purposes I think we can ignore that for the sake of simplicity:

```
strlcpy(buf, s_procName.c_str(), size);
```

jeschu1: Sigh yes :-), copy/paste error

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +60,187 @@ - (void)testShouldIgnoreVnodeType {
...
+    
+    testVnode->attrValues.va_flags = FileFlags_IsEmpty;
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->attrValues.va_flags = FileFlags_Invalid;
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+    
+    testVnode->attrValues.va_flags = 0x00000100;
```
wilbaker: Does `0x00000100` have special meaning? Or was it just something other than `FileFlags_IsInVirtualizationRoot`?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +60,187 @@ - (void)testShouldIgnoreVnodeType {
...
+    XCTAssertTrue(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+    XCTAssertFalse(fileFlaggedInRoot);
+
+    testVnode->errors.getattr = EBADF;
+    XCTAssertFalse(TryGetFileIsFlaggedAsInRoot(testVnode.get(), NULL, &fileFlaggedInRoot));
+}
```
wilbaker: Consider also adding a test that sets `va_flags` to `FileFlags_IsInVirtualizationRoot` OR-ed (`|`) with something else and assert that `fileFlaggedInRoot` is true.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +60,187 @@ - (void)testShouldIgnoreVnodeType {
...
+
+- (void)testShouldHandleVnodeOpEvent {
+    // In Parameters
+    shared_ptr<mount> testMount = mount::Create();
+    shared_ptr<vnode> testVnode = vnode::Create(testMount, "/foo");
+    testVnode->attrValues.va_flags = FileFlags_IsInVirtualizationRoot;
+    PerfTracer perfTracer;
+    vfs_context_t _Nonnull context = vfs_context_create(NULL);
```
wilbaker: Nit: `NULL` -> `nullptr`

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +60,187 @@ - (void)testShouldIgnoreVnodeType {
...
+            &vnodeType,
+            &vnodeFileFlags,
+            &pid,
+            procname,
+            &kauthResult,
+            &kauthError));
+    XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
+    
+    
+    // Test Invalid Action Bit
```
wilbaker: Minor:  When I read "Invalid Action Bit" it sounded like we'd be passing in a bit that doesn't correspond to a well known `KAUTH_VNODE_` value.

Consider updating this comment to something like:

`KAUTH_VNODE_ACCESS should not be handled`

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -58,4 +60,187 @@ - (void)testShouldIgnoreVnodeType {
...
+    XCTAssertFalse(
+        ShouldHandleVnodeOpEvent(
+            &perfTracer,
+            context,
+            testVnode.get(),
+            KAUTH_VNODE_ACCESS,
```
wilbaker: Also add a test for `KAUTH_VNODE_ACCESS` OR-ed with something we do otherwise care about?

e.g. `KAUTH_VNODE_ACCESS | KAUTH_VNODE_DELETE`

ProjFS.Mac/PrjFSKextTests/MockProc.hpp
```diff
@@ -0,0 +1,14 @@
...
+    KAUTH_RESULT_ALLOW = 1,
+    KAUTH_RESULT_DENY,
+    KAUTH_RESULT_DEFER
+};
+
+extern "C" int proc_pid(proc_t);
+extern "C" void proc_name(int pid, char * buf, int size);
+extern "C" proc_t vfs_context_proc(vfs_context_t ctx);
+
+void SetProcName(const char* procName);
```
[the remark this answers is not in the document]
jeschu1: Forgive my rusty C++, but if we don that can we call "SetProcName("Finder")"?

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
```diff
@@ -72,6 +72,12 @@ struct mount
 // This way, we can have a mock mount point as a test class instance variable
 // and don't have to manually clear out the rootVnode after every test.
 //
+struct VnodeMockErrors
+{
+    errno_t getpath = 0;
+    errno_t getattr = 0;
```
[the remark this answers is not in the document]
jeschu1: VnodeMockErrors (Struct Name) signifies they are errors for the listed function calls.
With the struct name in mind, the error seems a bit redundant.
All that would be in this class are return codes for the specified function names.

ProjFS.Mac/PrjFSKextTests/MockProc.hpp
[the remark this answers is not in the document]
wilbaker: Yep, if we make the parameter `const std::string&` we can still call the function with `char*`.

This is allowed because `std::string` has a [converting constructor](https://stackoverflow.com/questions/15077466/what-is-a-converting-constructor-in-c-what-is-it-for) for [char*](http://www.cplusplus.com/reference/string/string/string/).  A temporary `std::string` will be automatically constructed for us from the `char*`.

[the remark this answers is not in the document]
pmj: Yeah, `const std::string&` is OK here. It's definitely less efficient, but we don't care about that much in the tests, and allows us to pass either `const char*` or `std::string` objects in.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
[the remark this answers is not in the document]
pmj: I agree with @jeschu1 here; actual uses of this struct in test code (e.g. elsewhere in this PR) currently look like this:

```
testVnode->errors.getattr = EBADF;
```

That makes perfect sense to me, including sticking with the capitalisation of function `vnode_getattr` vs. the camelCase convention we'd be applying if we were coming up with the name from scratch. I don't think the alternatives,

```
testVnode->errors.getAttrError = EBADF;
// or
testVnode->errors.getAttrResult = EBADF;
```

are an improvement.

If we're going to change anything here, the variant I'd be open to is:

```
testVnode->errors.vnode_getattr = EBADF;
```

I like this despite it flagrantly disregarding our naming conventions - it makes it even more obvious we're providing the mock error code for `vnode_getattr()`.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: FWIW, possible OS-defined values for flags are:

 * `UF_NODUMP`
 * `UF_IMMUTABLE`
 * `UF_APPEND`
 * `UF_OPAQUE`
 * `UF_HIDDEN`
 * `SF_ARCHIVED`
 * `SF_IMMUTABLE`
 * `SF_APPEND`

See the manpage for `chflags(2)`. These values should be defined in `<sys/stat.h>`.

ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
```diff
@@ -12,5 +13,20 @@
 
 KEXT_STATIC_INLINE bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
 KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
+KEXT_STATIC_INLINE bool TryGetFileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t context, bool* flaggedInRoot);
```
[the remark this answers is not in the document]
pmj: The mock KPIs currently don't use the context pointer for anything, but we could change that to check it gets passed through correctly and consistently. (some APIs silently accept `nullptr` contexts in some circumstances while KPing in others, for example)

So I'd agree, we should probably try to write our tests consistently. Where we don't really care about vfs contexts, we can just create one in `setUp` and destroy it in `tearDown` and use it in each of the test methods. That's not significantly more work than passing `nullptr`.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: Just something different

ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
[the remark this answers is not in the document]
jeschu1: @wilbaker since KauthHandlerTestable is a header file if we add _Nonnull for one paramter we must add it for every parameter.  

KauthHandler.cpp does not have this requirement since it isn't a header.

So to get the signatures to match we'd either need to
1. Add a nonnullable/nullable tag to every param in Kauthhandler.cpp which creates a lot of noise.  Don't do that.
2. Use " #pragma clang assume_nullable begin/end" which @pmj has advised against.

For now I'll complete this PR without adding `_Nonnull `but we can pick the discussion up at a later time.

ProjFS.Mac/PrjFSKextTests/MockVnodeAndMount.hpp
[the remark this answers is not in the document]
wilbaker: @jeschu1 @pmj that sounds good to me, thanks for the detailed explanations!

ProjFS.Mac/PrjFSKext/KauthHandlerTestable.hpp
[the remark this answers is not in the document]
wilbaker: > @wilbaker since KauthHandlerTestable is a header file if we add _Nonnull for one paramter we must add it for every parameter.

> KauthHandler.cpp does not have this requirement since it isn't a header.

Ahh I didn't realize there was a difference in the requirements for these specifiers (between headers and source).

> Add a nonnullable/nullable tag to every param in Kauthhandler.cpp which creates a lot of noise. Don't do that.

I actually think tagging all of the pointer params isn't such a bad idea.  It would get us some additional protection from the compiler and help convey the expectations for those parameters.  

However, I agree there's no need to clean that up now (or block this PR on it).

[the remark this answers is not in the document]
wilbaker: > Ahh I didn't realize there was a difference in the requirements for these specifiers (between headers and source).

This would also explain why I didn't hit this issue with the vnode cache, I believe I specified the nullability for all pointer parameters.

PR #917, opened 2019-03-12
PrjFs: use builtins for detecting overflow to protect memory allocations

jrbriggs:


ProjFS.Mac/PrjFSKext/KextLog.cpp
```diff
@@ -81,7 +81,9 @@ void KextLog_Printf(KextLog_Level loglevel, const char* fmt, ...)
     va_start(args, fmt);
     int messageLength = vsnprintf(message.logString, sizeof(message.logString), fmt, args);
     va_end (args);
-    int messageSize = sizeof(KextLog_MessageHeader) + messageLength + 1 /* null terminator */;
+    int messageSize;
+    if (__builtin_add_overflow(sizeof(KextLog_MessageHeader), messageLength, &messageSize)) return;
+    if (__builtin_add_overflow(messageLength, 1, &messageSize)) return;
```
jrbriggs: bad bug.

jrbriggs: Should be 
```suggestion
    if (__builtin_add_overflow(messageSize, 1, &messageSize)) return;
```

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -287,10 +287,10 @@ static VirtualizationRootHandle FindUnusedIndexOrGrow_Locked()
     if (RootHandle_None == rootIndex)
     {
         // No space, resize array
-        uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
-        if (newLength <= s_maxVirtualizationRoots)
+        uint16_t newLength;
+        if (__builtin_mul_overflow(s_maxVirtualizationRoots, 2, &newLength))
```
jrbriggs: @pmj I think this is equivalent but please confirm your intent with the old code.

ProjFS.Mac/PrjFSKext/Memory.hpp
```diff
@@ -18,21 +18,25 @@ void Memory_Free(void* buffer, uint32_t size);
 template <typename T>
 T* Memory_AllocArray(uint32_t arrayLength)
```
jeschu1: Unit Test?

ProjFS.Mac/PrjFSKext/Memory.hpp
```diff
@@ -18,21 +18,25 @@ void Memory_Free(void* buffer, uint32_t size);
...
         return nullptr;
     }
-    
-    return static_cast<T*>(Memory_Alloc(static_cast<uint32_t>(allocBytes)));
+
+    return static_cast<T*>(Memory_Alloc(allocBytes));
 }
 
 template <typename T>
 void Memory_FreeArray(T* array, uint32_t arrayLength)
```
jeschu1: Unit Test?

ProjFS.Mac/PrjFSKext/Memory.hpp
```diff
@@ -18,21 +18,25 @@ void Memory_Free(void* buffer, uint32_t size);
...
 {
-    size_t arrayBytes = arrayLength * sizeof(T);
-    assert(arrayBytes <= UINT32_MAX);
-    Memory_Free(array, static_cast<uint32_t>(arrayBytes));
+    uint32_t arrayBytes;
+    if (__builtin_umul_overflow(arrayLength, sizeof(T), &arrayBytes))
+    {
+        assert(!"Overflow detected");
```
pmj: `panic()` or `assertf()` might be more appropriate for explicitly crashing with a message. `assertf()` in particular would allow us to put the values for `arrayLength` and `sizeof(T)` in the panic message, so we have a chance of working out where the bug is coming from.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Not quite, the handles are signed integers hence the limit of `INT16_MAX + 1u` (= 32768), whereas the max value of `newLength` is `UINT16_MAX` (= 65535).

ProjFS.Mac/PrjFSKext/Memory.hpp
jrbriggs: Done, though this required changing the signature. The other alternative would be to provide a mocked implementation of `assert()`. @pmj @jeschu1 any preferences?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -287,10 +287,10 @@ static VirtualizationRootHandle FindUnusedIndexOrGrow_Locked()
...
         {
-            assert(newLength > 0);
```
pmj: This check had nothing to do with overflow, but catching the situation where `s_maxVirtualizationRoots` was not initialised. Happened to me in unit tests, which caused things to screw up. Please don't remove the assert.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -287,10 +287,10 @@ static VirtualizationRootHandle FindUnusedIndexOrGrow_Locked()
     if (RootHandle_None == rootIndex)
     {
         // No space, resize array
-        uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
-        if (newLength <= s_maxVirtualizationRoots)
```
pmj: This is checking if the `newLength` is actually the same as the existing length, in which case there is no need to reallocate memory. Nothing to do with overflow.

ProjFS.Mac/PrjFSKext/KextLog.cpp
```diff
@@ -81,7 +81,9 @@ void KextLog_Printf(KextLog_Level loglevel, const char* fmt, ...)
     va_start(args, fmt);
     int messageLength = vsnprintf(message.logString, sizeof(message.logString), fmt, args);
     va_end (args);
-    int messageSize = sizeof(KextLog_MessageHeader) + messageLength + 1 /* null terminator */;
+    int messageSize;
+    if (__builtin_add_overflow(sizeof(KextLog_MessageHeader), messageLength, &messageSize)) return;
+    if (__builtin_add_overflow(messageSize, 1, &messageSize)) return;
```
wilbaker: Style nit:

Please break each of these `return` statements onto their own lines.

ProjFS.Mac/PrjFSKext/KextLog.cpp
```diff
@@ -81,7 +81,9 @@ void KextLog_Printf(KextLog_Level loglevel, const char* fmt, ...)
     va_start(args, fmt);
     int messageLength = vsnprintf(message.logString, sizeof(message.logString), fmt, args);
     va_end (args);
-    int messageSize = sizeof(KextLog_MessageHeader) + messageLength + 1 /* null terminator */;
+    int messageSize;
```
pmj: This should probably be a `uint32_t`.

ProjFS.Mac/PrjFSKext/Memory.hpp
```diff
@@ -18,21 +18,27 @@ void Memory_Free(void* buffer, uint32_t size);
...
     }
-    
-    return static_cast<T*>(Memory_Alloc(static_cast<uint32_t>(allocBytes)));
+
+    return static_cast<T*>(Memory_Alloc(allocBytes));
 }
 
 template <typename T>
-void Memory_FreeArray(T* array, uint32_t arrayLength)
+bool Memory_FreeArray(T* array, uint32_t arrayLength)
```
wilbaker: Can we leave this as return type `void`?

`Memory_FreeArray` should only be called for memory that was allocated with `Memory_AllocArray`, and `Memory_AllocArray` already validated that there was no overflow when allocating.

What about something like:

```
void Memory_FreeArray(T* array, uint32_t arrayLength)
{
    uint32_t arrayBytes;
    if (__builtin_umul_overflow(arrayLength, sizeof(T), &arrayBytes)) 
    {
        // Overflow should never occur as Memory_AllocArray should have prevented it
        assert(false);
    }
    else
    {
        Memory_Free(array, arrayBytes);
    }
}
```

ProjFS.Mac/PrjFSKextTests/MemoryTests.mm
```diff
@@ -0,0 +1,34 @@
+//
+//  MemoryTests.m
+//  PrjFSKextTests
+//
+//  Created by John Briggs on 3/12/19.
+//  Copyright © 2019 Microsoft Corporation. All rights reserved.
```
wilbaker: This automatic copyright should be removed (perhaps replaced with boilerplate MIT license comment?)

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
jrbriggs: Walk me through how that would happen? I must be missing something.

jrbriggs: Oh, wait, I see. We're mixing signed and unsigned types. That begs the question -- why are we mixing signed and unsigned types?

ProjFS.Mac/PrjFSKext/Memory.hpp
jrbriggs: I like this idea, let me revise the PR.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: The negative values for root handles are the various special `RootHandle_None`, `RootHandle_Indeterminate`, etc. values. We could move those to the top end of the `uint16_t` range instead, but I suspect that would make things worse, not better. So I guess the solution would be to limit `s_maxVirtualizationRoots` to `INT16_MAX` instead of `INT16_MAX + 1` and use signed types throughout.

ProjFS.Mac/PrjFSKext/Memory.hpp
pmj: Yes, I agree, let's not change the return type. The in-kext asserts are currently badly integrated with the unit tests, but this is something I'd like to fix.

The size calculations in `Memory_AllocArray` and `Memory_FreeArray` are identical, so perhaps we can pull those out into a helper function and unit test that. You could also leave a TODO comment for testing `Memory_FreeArray` once we can handle in-kext asserts in unit tests.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Bonus points: as it's evidently not obvious why the `assert` is there, perhaps turn it into an `assertf` with a suitable message.

ProjFS.Mac/PrjFSKext/Memory.hpp
jrbriggs: That's a good point. I've incorporated that feedback.

ProjFS.Mac/PrjFSKextTests/MemoryTests.mm
jrbriggs: Removed. I'll make a pass through at a later date to get correct headers on each file.

ProjFS.Mac/PrjFSKext/KextLog.cpp
jrbriggs: Restructured into one if statement.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
jrbriggs: I convinced myself the math would never overflow, removed the overflow checking, and made the assert more descriptive.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -279,10 +279,16 @@ static VirtualizationRootHandle FindUnusedIndexOrGrow_Locked()
...
         {
-            assert(newLength > 0);
+            assertf(newLength > 0, "s_maxVirtualizationRoot was likely not initialized");
             return RootHandle_None;
         }
-        
+
+        if (newLength <= s_maxVirtualizationRoots)
+        {
+            // Already at max size, nothing to do.
```
jrbriggs: This block is duplicated, need to remove.

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
```diff
@@ -1187,8 +1191,8 @@
 		F5E39C7C21F1118D006D65C2 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				CODE_SIGN_STYLE = Automatic;
 				CLANG_ENABLE_CODE_COVERAGE = YES;
+				CODE_SIGN_STYLE = Automatic;
```
jeschu1: Nit: Not ideal that CODE_SIGN_STYLE moved.  Perhaps we always get small changes like this and I never noticed.  @wilbaker?

ProjFS.Mac/PrjFSKext/KextLog.cpp
```diff
@@ -81,7 +81,13 @@ void KextLog_Printf(KextLog_Level loglevel, const char* fmt, ...)
     va_start(args, fmt);
     int messageLength = vsnprintf(message.logString, sizeof(message.logString), fmt, args);
     va_end (args);
-    int messageSize = sizeof(KextLog_MessageHeader) + messageLength + 1 /* null terminator */;
+    uint32_t messageSize;
+    if (__builtin_uadd_overflow(sizeof(KextLog_MessageHeader), messageLength, &messageSize) ||
+        __builtin_uadd_overflow(messageSize, 1, &messageSize))
```
jeschu1: Here we'd skip logging a message to Kext.  Is it worth trying to truncate and write something or an error?

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
wilbaker: I thought that maybe Xcode decided to reorder these when the source files were reordered?

jeschu1: Maybe? It's equivalent just noise.
Not sure if it's the move or Xcode version / machine dependent
Not worth blocking over.  I don't recall seeing a change like this previously.

wilbaker: It looks like Xcode wants these values to be alphabetized.

When @jrbriggs added `CLANG_STATIC_ANALYZER_MODE` in new spots here:

https://github.com/Microsoft/VFSForGit/commit/c21ad3e5f8fc414104e654970a0704b08d3e51ab#diff-e0c3dea4367db12e2eec645c616be25b

They were ordered alphabetically:

```
CLANG_STATIC_ANALYZER_MODE = deep; 
CODE_SIGN_STYLE = Automatic;
```

And this change is to make the other CLANG_STATIC_ANALYZER_MODE, CODE_SIGN_STYLE entries alphabetical as well.

(It looks like the re-ordered entries were first added here https://github.com/Microsoft/VFSForGit/commit/4342ef86f3fa46cc3197b0a9086cc6f613dfa8c7 and they were not in alphabetical order, I'm not sure if those were added by hand).

I was discussing this with @nickgra offline and he recommended that we not fight Xcode and allow it to re-order things when it wants to.

jeschu1: Thanks for verifying @wilbaker that makes sense to me!

jrbriggs: Sorry, I couldn't stand not having the items in xcode alphabetized :)

ProjFS.Mac/PrjFSKext/KextLog.cpp
jeschu1: Talked to @jrbriggs offline and that would be a change for a separate PR.

PR #810, opened 2019-02-20
[WIP] initial build of GVFS provider for Linux

chrisd8088:
Please **DO NOT MERGE** -- this is a WIP branch to debug issues getting the GVFS provider to build and run on Linux.

(Note to @jrbriggs et al: I'm opening this PR because I want to see if this passes the CI checks on Mac and Windows; there may well be additional things I need to tweak in the tooling before it does.  Thanks and apologies!)

This is largely a duplicate of the macOS build tooling and platform-specific code, with some changes to ensure we build successfully on Linux (e.g., testing for `'$(OS)' != 'Windows_NT'` is no longer sufficient to determine that the platform is OS X).

Note that for now the `BuildGVFSForLinux.sh` script exits early, without checking if a native libprojfs library is available, performing any installation or packaging, etc.

The `GVFS.sln` configuration changes should ensure that the `ProjFS.Linux.Managed` and `GVFS.Platform.Linux` subprojects are built on all platforms (and the Mac and Windows equivalents
are built on Linux), so as to allow the `GVFS.UnitTests` project to build on Linux, as it seems to do multi-platform unit testing.

However, we don't yet require the `GVFS.Mount.Linux` project to build on other platforms, and the `GVFS.FunctionalTests` project's configuration and dependencies will also need to be revisited
once the Linux code is actually functional.

/cc @kivikakk @wrighty

PR #926, opened 2019-03-14
[Mac, Linux] refactor Mac platform into POSIX base

kivikakk:
Here's a refactor of the Mac platform into a POSIX base. (On my end, I did the full refactor of Mac and Linux platforms, then rebased away the Linux changes.) This PR does so without introducing the Mono dependency; just moving code around.

/cc @jrbiggs, #810
/cc @chrisd8088 fyi!


GVFS/GVFS.Platform.POSIX/POSIXFileBasedLock.cs
```diff
@@ -33,91 +33,91 @@ public override bool TryAcquireLock()
...
+            return metadata;
+        }
+
+        private static class NativeMethods
+        {
+            // #define O_WRONLY    0x0001      /* open for writing only */
+            public const int OpenWriteOnly = 0x0001;
+
+            // #define O_CREAT     0x0200      /* create if nonexistant */
+            public const int OpenCreate = 0x0200;
```
chrisd8088: One problem here is that these constants aren't the same on Linux as on BSD/macOS.

In particular, this should be `OpenCreate = 0x0040` on Linux, and below should have `EWouldBlock = 11`, and use `uint` and `ToUInt32` for the `mode_t` values.  See b02e4a3845 for what I believe are the relevant correct values, but that could use a double-check from someone else, for sure.

As a result, I expect this class should be divvied up into Linux and Mac variants, perhaps as a base common POSIX class (with things like `public const int InvalidFileDescriptor = -1`) and then the platform-specific values defined in separate classes.

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
```diff
@@ -1,42 +1,42 @@
...
+using GVFS.Platform.POSIX;
 
 namespace GVFS.Hooks.HooksPlatform
 {
     public static class GVFSHooksPlatform
     {
         public static string GetInstallerExtension()
         {
-            return MacPlatform.InstallerExtension;
+            return "dmg";
```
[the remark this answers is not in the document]
chrisd8088: Not sure, and @kivikakk is out for a stretch now.  I suspect `MacPlatform.InstallerExtension` would be fine; perhaps this is an unintended consequence of refactoring or rebasing.

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.Shared.cs
```diff
@@ -1,13 +1,13 @@
...
-        }
-    }
-}
+namespace GVFS.Platform.POSIX
+{
+    public partial class POSIXFileSystem
+    {
+        public static bool TryGetNormalizedPathImplementation(string path, out string normalizedPath, out string errorMessage)
+        {
+            // TODO(POSIX): Properly determine normalized paths (e.g. across links)
```
wilbaker: Do you expect that all of the TODO(POSIX) will apply to the Mac platform?

We've been using "TODO(Mac)" as a key when searching for TODOs in the code that we still have for Mac.  I'm wondering if perhaps instead of updating these to TODO(POSIX) we should use TODO(Mac), TODO(Linux)?

@jrbriggs any thoughts/preferences here?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -1,157 +1,22 @@
...
-using GVFS.Common.Tracing;
-using System;
-using System.Collections.Generic;
-using System.IO;
-using System.IO.Pipes;
 
 namespace GVFS.Platform.Mac
 {
-    public partial class MacPlatform : GVFSPlatform
+    public class MacPlatform : POSIX.POSIXPlatform
```
wilbaker: Minor: Change `POSIX.POSIXPlatform` to `POSIXPlatform` and add a `using` for the POSIX namespace.

GVFS/GVFS.Platform.POSIX/NetCoreMethods.cs
```diff
@@ -1,101 +1,101 @@
...
-                }
-
-                // And then the array itself
-                Marshal.FreeHGlobal((IntPtr)arr);
-            }
-        }
-
-    }
-}
+// This file isn't generated, but this comment is necessary to exclude it from StyleCop analysis.
```
wilbaker: This comment applies in a few places, but I see in some files there are a lot of whitespace changes.  Unfortunately the line endings in the repo are kind of messy right now.  Could you try to leave whatever line endings are already being used in a file as-is rather than updating them all?

GVFS/GVFS.Platform.POSIX/POSIXDiskLayoutUpgradeData.cs
```diff
@@ -1,27 +1,27 @@
...
-            return false;
-        }
-    }
-}
+using GVFS.Common;
+using GVFS.DiskLayoutUpgrades;
+
+namespace GVFS.Platform.POSIX
+{
+    public class POSIXDiskLayoutUpgradeData : IDiskLayoutUpgradeData
```
wilbaker: I think the `IDiskLayoutUpgradeData` implementations will need to stay platform specific.

Especially as Linux and Mac are being built up I expect that we'll need to update the disk layout on one platform without impacting the other (e.g. we just had to bump Mac from 17 to 18 while leaving Windows at 17).

If there ends up being common `DiskLayoutUpgrade` steps those can be shared, but I think that the major/minor disk layout versions will need to be kept separate.

GVFS/GVFS.Platform.POSIX/NetCoreMethods.cs
jrbriggs: Good catch. I had whitespace diffs disabled.

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.Shared.cs
jrbriggs: I can see pros and cons for each. TODO(Mac), TODO(Linux) would let us knock one off at a time and perhaps leave the other as a reminder to test the implementation on that platform. It would be a bit more verbose. I'm supportive of either if removing TODO(POSIX) means the implementation has been proven correct on both platforms.

wilbaker: That sounds good to me, as long as were good about TODO(POSIX) meaning "needs to be done on Mac and Linux" I think we can stick with TODO(POSIX).

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
```diff
@@ -1,42 +1,42 @@
-using GVFS.Platform.Mac;
+using GVFS.Platform.POSIX;
```
kewillford: Not sure if we have established a hard rule but using `Posix` instead of `POSIX` reads better to me.  What are you thoughts @wilbaker

wilbaker: I agree, I prefer `Posix`

jrbriggs: I'll be the dissenting opinion here. We have precedent for not camel-casing acronyms in GVFS. We should probably stick with POSIX for the same reason.

wilbaker: > We have precedent for not camel-casing acronyms in GVFS. We should probably stick with POSIX for the same reason.

That's a good point, we use `GVFS` in the code rather than `Gvfs`.  I'm good with sticking with `POSIX`.

GVFS/GVFS.Platform.POSIX/NetCoreMethods.cs
chrisd8088: I can try to fix this while @kivikakk is out; it likely stems from my using LF on the Linux-exclusive files and then having those converted to the POSIX versions, instead of going from the Mac originals to the POSIX variant.

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
```diff
@@ -1,202 +1,202 @@
...
-                public long[] QSpare;     /* RESERVED: DO NOT USE! */
-            }
-        }
-
-        private class NativeFileReader
-        {
-            public const int ReadOnly = 0x0000;
-            public const int WriteOnly = 0x0001;
-
-            public const int Create = 0x0200;
```
chrisd8088: We have the same problem here as in the file-lock code; this is `0x0040` on Linux, and the various `chmod`-related calls should use an `uint`.  There may be other Linux vs. BSD changes here as well, so this probably needs to be refactored into a POSIX base with Linux and Mac extensions.

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
chrisd8088: One additional thought here is that we should presumably also rename this file to `GVFSHooksPlatform.POSIX.cs` and adjust the various `*.csproj` files to use that instead.

GVFS/GVFS.Platform.POSIX/NetCoreMethods.cs
chrisd8088: OK ... as a pure whitespace fixup, I've added f6aac4a which puts the line endings back as they were in the Mac originals, but makes no other edits.

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
wilbaker: >There may be other Linux vs. BSD changes here as well, so this probably needs to be refactored into a POSIX base with Linux and Mac extensions.

@kivikakk, @chrisd8088, are you planning to make this change in this PR, or in a follow up PR?

wilbaker:
@kivikakk @chrisd8088, @kewillford and I have been unable to kick off the Mac functional tests and we suspect it may be due to the merge conflict in `GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs`.

Could you resolve this conflict and we'll try kicking off the tests again?

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
chrisd8088: @kivikakk is away for a bit and as this is their work, I'm not looking to make any major changes to it right now.  My own feeling at the moment is summarized by in my [review comment](https://github.com/Microsoft/VFSForGit/pull/926#discussion_r268354587):

> To that end, the #810 PR (still a WIP) has a number of these modifications in it now, so using that as a basis for refactoring (maybe refactoring after that code has landed?) might be useful.

But I'll defer to @kivikakk to decide if they want to try to get the POSIX changes in first.  For my part, I'm focused on looking at blockers to running `gvfs mount` on Linux, which is somewhat orthogonal work to this refactoring.

chrisd8088:
OK @wilbaker, I've rebased on the latest `master` and resolved the conflicts and changes which can with that, and all the CI tests pass except the one which seems "stuck".

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
wilbaker: @chrisd8088 ok thanks!

Regarding the status of this PR, I think it's close being ready to merge, there are just a few open comments left from the first round of review.

/cc: @jrbriggs @kivikakk

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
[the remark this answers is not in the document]
kivikakk: `MacPlatform.InstallerExtension` ceased to exist as of this rebase, but it's better left in. Readjusted in 0d1ba4ce.

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
kivikakk: There'll likely need to be one each for Mac and Linux since `GetInstallerExtension` will differ for both. It's a bit awkward; any thoughts? Could dump everything except `GetInstallerExtension` into `GVFSHooksPlatform.POSIX`, make the class partial, and then have a `.Mac` and `.Linux` variant with only `GetInstallerExtension`?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -1,157 +1,22 @@
...
-        public override string GetCurrentUser()
-        {
-            throw new NotImplementedException();
-        }
 
         public override string GetOSVersionInformation()
         {
             ProcessResult result = ProcessHelper.Run("sw_vers", args: string.Empty, redirectOutput: true);
             return string.IsNullOrWhiteSpace(result.Output) ? result.Errors : result.Output;
-        }
```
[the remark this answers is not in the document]
kivikakk: Indeed; this is remaining in the `MacPlatform` class, not in the common base. The method remains abstract in the base and will be implemented by `LinuxPlatform `when introduced.

GVFS/GVFS.Platform.POSIX/POSIXDiskLayoutUpgradeData.cs
kivikakk: Reverted in e67c4efb.

GVFS/GVFS.Platform.POSIX/POSIXFileBasedLock.cs
kivikakk: Super helpful commit link, thank you! It'll be easier to know what to refactor when it does come time to add the Linux platform on top of these commits -- for now I'll mark POSIXFileBasedLock as abstract and add an empty implementation class for Mac so we can shift specific stuff into it when we refactor later.

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
kivikakk: I'd like to perform these changes later when we add the Linux platform in -- at that point we should refactor so that e.g. there's only one definition of any given native method for a given platform, whereas right now we have `open` and `O_CREAT` defined twice (in FileSystem and FileBasedLock).

It's pretty hard to get a refactor right when there's only one consumer of the code, and right now that's the case (Mac platform code is the single consumer of the POSIX platform base code). We'll maybe find more cases like this where Linux varies vs. what we've pre-emptively abstracted, but we should handle them as we implement the Linux platform in the tree, and not try to get this PR 100% correct before (imo). More important is that we get the initial refactor down asap so we don't get increasingly large deltas as time presses on.

kivikakk:
All test passing. Let me know if this looks okay; cc @wilbaker @jrbriggs.

kivikakk:
Looks like Mac Functional Tests timed out after running for more than 3 hours (but passed as far as it got?).

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
chrisd8088: Sounds like a reasonable plan!

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -1,133 +1,17 @@
...
+    {
+        public MacFileBasedLock(
+            PhysicalFileSystem fileSystem,
+            ITracer tracer,
+            string lockPath)
+            : base(fileSystem, tracer, lockPath)
+        {
+        }
+    }
+}
```
jrbriggs: nit: newline before EOF

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -1,133 +1,17 @@
...
-            public static extern int Close(int fd);
-
-            [DllImport("libc", EntryPoint = "flock", SetLastError = true)]
-            public static extern int FLock(int fd, int operation);
-        }
-    }
-}
+namespace GVFS.Platform.Mac
+{
+    public class MacFileBasedLock : POSIXFileBasedLock
```
jrbriggs: Seems like, at least right now, this is just a `POSIXFileBasedLock`. Do you know of something coming shortly that will require a distinction?

kivikakk: Indeed; see https://github.com/Microsoft/VFSForGit/pull/926#discussion_r271140642. There's no strict reason to abstract it at all just yet, so I'll undo it. That can come later.

jrbriggs: I'm fine leaving `MacFileBasedLock` as you have it here if we already know why we need it.

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
kivikakk: Line endings /o\

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
kivikakk: 👍  I'd prefer to start more correct, which I think this way is. There's unlikely to be churn here, too, so it's not an issue.

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
wilbaker: > Could dump everything except GetInstallerExtension into GVFSHooksPlatform.POSIX, make the class partial, and then have a .Mac and .Linux variant with only GetInstallerExtension?

That sounds good to me.  I am okay with leaving the current code as-is and making that change in a follow up PR if you prefer.

wilbaker: @kivikakk I just realized nothing is calling `GetInstallerExtension` anymore.  Could you just remove it from this class (and the .Windows.cs version of this class)?

GVFS/GVFS.Platform.Mac/MacFileBasedLock.cs
```diff
@@ -86,7 +86,7 @@ public override void Dispose()
         private EventMetadata CreateEventMetadata(int errno = 0)
```
wilbaker: Minor: Could you revert the whitespace changes in this file?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -1,157 +1,33 @@
...
     {
         public MacPlatform()
             : base(
-                executableExtension: string.Empty,
-                installerExtension: ".dmg",
-                underConstruction: new UnderConstructionFlags(
-                    supportsGVFSService: false,
-                    supportsGVFSUpgrade: false,
-                    supportsGVFSConfig: false))
+                installerExtension: "." + InstallerExtension)
```
wilbaker: It looks like this is now the only spot in the code where `InstallerExtension` is being used (after removing `GetInstallerExtension` from `GVFSHooksPlatform`).  

Could you remove `InstallerExtension`  or update the extension string to include the `.`?

kivikakk:
Suggestions taken up 👍

kivikakk:
Curious if the following will work for me:

GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
```diff
@@ -1,42 +1,37 @@
...
```
wilbaker: Rename this file now? 

`GVFSHooksPlatform.Mac.cs` -> `GVFSHooksPlatform.POSIX.cs`

kivikakk: Done.

kivikakk:
Thanks for your patience with this! :heart:

PR #935, opened 2019-03-15
Make `TelemetryDaemonEventListener` and `JsonTracer` thread-safe and async

### LINKED ISSUE #903: TelemetryDaemonEventListener is not thread safe
See discussion in #893 for more details.

Multiple threads might try to use `this.pipeClient` at the same time and we don't have any locking to protect it.

Related, although we've not seen any issues with `JsonTracer.listeners` it's worth confirming that VFS4G's usage of it won't run into any thread-safety issues.

mjcheetham:
Ensure that modifications and enumeration of the `JsonTracer::listeners` collection is safe with multiple threads adding/removing listeners at the same time as we're writing messages. A `ConcurrentBag` is used to achieve this aim.

The 'dispose' flag of the `JsonTracer` has been updated to be read and compared/written atomically to prevent threading issues with listeners being added after disposal has started.

Also ensure the `TelemetryDaemonEventListener` does not hold up calling threads posting messages over the pipe by queuing up messages to be sent on a single background thread.

Fixes #903

### DIFF
--- a/GVFS/GVFS.Common/Tracing/EventListener.cs
+++ b/GVFS/GVFS.Common/Tracing/EventListener.cs
@@ -7,37 +7,32 @@ namespace GVFS.Common.Tracing
     {
         private readonly EventLevel maxVerbosity;
         private readonly Keywords keywordFilter;
+        private readonly IEventListenerEventSink eventSink;
 
-        protected EventListener(EventLevel maxVerbosity, Keywords keywordFilter)
+        protected EventListener(EventLevel maxVerbosity, Keywords keywordFilter, IEventListenerEventSink eventSink)
         {
             this.maxVerbosity = maxVerbosity;
             this.keywordFilter = keywordFilter;
+            this.eventSink = eventSink;
         }
 
         public virtual void Dispose()
         {
         }
 
-        public bool? TryRecordMessage(TraceEventMessage message, out string errorMessage)
+        public void RecordMessage(TraceEventMessage message)
         {
             if (this.IsEnabled(message.Level, message.Keywords))
             {
                 try
                 {
                     this.RecordMessageInternal(message);
-
-                    errorMessage = null;
-                    return true;
                 }
                 catch (Exception ex)
                 {
-                    errorMessage = ex.ToString();
-                    return false;
+                    this.RaiseListenerFailure(ex.ToString());
                 }
             }
-
-            errorMessage = null;
-            return null;
         }
 
         protected abstract void RecordMessageInternal(TraceEventMessage message);
@@ -66,6 +61,16 @@ namespace GVFS.Common.Tracing
             return this.keywordFilter != Keywords.None &&
                 this.maxVerbosity >= level &&
                 (this.keywordFilter & keyword) != 0;
+        }
+
+        protected void RaiseListenerRecovery()
+        {
+            this.eventSink?.OnListenerRecovery(this);
+        }
+
+        protected void RaiseListenerFailure(string errorMessage)
+        {
+            this.eventSink?.OnListenerFailure(this, errorMessage);
         }
     }
 }
--- a/GVFS/GVFS.Common/Tracing/JsonTracer.cs
+++ b/GVFS/GVFS.Common/Tracing/JsonTracer.cs
@@ -4,14 +4,15 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 
 namespace GVFS.Common.Tracing
 {
-    public class JsonTracer : ITracer
+    public class JsonTracer : ITracer, IEventListenerEventSink
     {
         public const string NetworkErrorEventName = "NetworkError";
 
-        private readonly List<EventListener> listeners;
+        private readonly ConcurrentBag<EventListener> listeners;
         private readonly ConcurrentDictionary<EventListener, string> failedListeners = new ConcurrentDictionary<EventListener, string>();
 
         private readonly string activityName;
@@ -22,8 +23,8 @@ namespace GVFS.Common.Tracing
         private readonly EventLevel startStopLevel;
         private readonly Keywords startStopKeywords;
 
+        private bool isDisposed = false;
         private bool stopped = false;
-        private bool disposed;
 
         public JsonTracer(string providerName, string activityName, bool disableTelemetry = false)
             : this(providerName, Guid.Empty, activityName, enlistmentId: null, mountId: null, disableTelemetry: disableTelemetry)
@@ -54,7 +55,7 @@ namespace GVFS.Common.Tracing
                     return;
                 }
 
-                TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId);
+                TelemetryDaemonEventListener daemonListener = TelemetryDaemonEventListener.CreateIfEnabled(gitBinRoot, providerName, enlistmentId, mountId, this);
                 if (daemonListener != null)
                 {
                     this.listeners.Add(daemonListener);
@@ -62,9 +63,9 @@ namespace GVFS.Common.Tracing
             }
         }
 
-        private JsonTracer(List<EventListener> listeners, Guid parentActivityId, string activityName, EventLevel startStopLevel, Keywords startStopKeywords)
+        private JsonTracer(ConcurrentBag<EventListener> listeners, Guid parentActivityId, string activityName, EventLevel startStopLevel, Keywords startStopKeywords)
         {
-            this.listeners = listeners ?? new List<EventListener>();
+            this.listeners = listeners ?? new ConcurrentBag<EventListener>();
             this.parentActivityId = parentActivityId;
             this.activityName = activityName;
             this.startStopLevel = startStopLevel;
@@ -92,47 +93,58 @@ namespace GVFS.Common.Tracing
 
         public void AddEventListener(EventListener listener)
         {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(JsonTracer));
+            }
+
             this.listeners.Add(listener);
 
             // Tell the new listener about others who have previously failed
             foreach (KeyValuePair<EventListener, string> kvp in this.failedListeners)
             {
                 TraceEventMessage failureMessage = CreateListenerFailureMessage(kvp.Key, kvp.Value);
-                listener.TryRecordMessage(failureMessage, out _);
+                listener.RecordMessage(failureMessage);
             }
         }
 
         public void AddDiagnosticConsoleEventListener(EventLevel maxVerbosity, Keywords keywordFilter)
         {
-            this.AddEventListener(new DiagnosticConsoleEventListener(maxVerbosity, keywordFilter));
+            this.AddEventListener(new DiagnosticConsoleEventListener(maxVerbosity, keywordFilter, this));
         }
 
         public void AddPrettyConsoleEventListener(EventLevel maxVerbosity, Keywords keywordFilter)
         {
-            this.AddEventListener(new PrettyConsoleEventListener(maxVerbosity, keywordFilter));
+            this.AddEventListener(new PrettyConsoleEventListener(maxVerbosity, keywordFilter, this));
         }
 
         public void AddLogFileEventListener(string logFilePath, EventLevel maxVerbosity, Keywords keywordFilter)
         {
-            this.AddEventListener(new LogFileEventListener(logFilePath, maxVerbosity, keywordFilter));
+            this.AddEventListener(new LogFileEventListener(logFilePath, maxVerbosity, keywordFilter, this));
         }
 
         public void Dispose()
         {
+            if (this.isDisposed)
+            {
+                // This instance has already been disposed
+                return;
+            }
+
             this.Stop(null);
 
             // If we have no parent, then we are the root tracer and should dispose our eventsource.
             if (this.parentActivityId == Guid.Empty)
             {
-                foreach (EventListener listener in this.listeners)
+                // Empty the listener bag and dispose of the instances as we remove them.
+                EventListener listener;
+                while (this.listeners.TryTake(out listener))
                 {
                     listener.Dispose();
                 }
-
-                this.listeners.Clear();
             }
 
-            this.disposed = true;
+            this.isDisposed = true;
         }
 
         public virtual void RelatedEvent(EventLevel level, string eventName, EventMetadata metadata)
@@ -288,6 +300,35 @@ namespace GVFS.Common.Tracing
             this.WriteEvent(this.activityName, this.startStopLevel, keywords, metadata, EventOpcode.Start);
         }
 
+        void IEventListenerEventSink.OnListenerRecovery(EventListener listener)
+        {
+            // Check ContainsKey first (rather than always calling TryRemove) because ContainsKey
+            // is lock-free and recoveredListener should rarely be in failedListeners
+            if (!this.failedListeners.ContainsKey(listener))
+            {
+                // This listener has not failed since the last time it was called, so no need to log recovery
+                return;
+            }
+
+            if (this.failedListeners.TryRemove(listener, out _))
+            {
+                TraceEventMessage message = CreateListenerRecoveryMessage(listener);
+                this.LogMessageToNonFailedListeners(message);
+            }
+        }
+
+        void IEventListenerEventSink.OnListenerFailure(EventListener listener, string errorMessage)
+        {
+            if (!this.failedListeners.TryAdd(listener, errorMessage))
+            {
+                // We've already logged that this listener has failed so there is no need to do it again
+                return;
+            }
+
+            TraceEventMessage message = CreateListenerFailureMessage(listener, errorMessage);
+            this.LogMessageToNonFailedListeners(message);
+        }
+
         private static string GetCategorizedErrorEventName(Keywords keywords)
         {
             switch (keywords)
@@ -332,11 +373,8 @@ namespace GVFS.Common.Tracing
         {
             string jsonPayload = metadata != null ? JsonConvert.SerializeObject(metadata) : null;
 
-            if (this.disposed)
+            if (this.isDisposed)
             {
-                Console.WriteLine("Writing to disposed tracer");
-                Console.WriteLine(jsonPayload);
-
                 throw new ObjectDisposedException(nameof(JsonTracer));
             }
 
@@ -351,53 +389,12 @@ namespace GVFS.Common.Tracing
                 Payload = jsonPayload
             };
 
+            // Iterating over the bag is thread-safe as the enumerator returned here
+            // is of a snapshot of the bag.
             foreach (EventListener listener in this.listeners)
             {
-                string errorMessage;
-                bool? success = listener.TryRecordMessage(message, out errorMessage);
-
-                // Try to mark success or failure for this listener only if it was enabled for this message type
-                if (success.HasValue)
-                {
-                    if (success == true)
-                    {
-                        this.MarkAndLogListenerRecovery(listener);
-                    }
-                    else
-                    {
-                        this.MarkAndLogListenerFailure(listener, errorMessage);
-                    }
-                }
+                listener.RecordMessage(message);
             }
-        }
-
-        private void MarkAndLogListenerRecovery(EventListener recoveredListener)
-        {
-            // Check ContainsKey first (rather than always calling TryRemove) because ContainsKey
-            // is lock-free and recoveredListener should rarely be in failedListeners
-            if (!this.failedListeners.ContainsKey(recoveredListener))
-            {
-                // This listener has not failed since the last time it was called, so no need to log recovery
-                return;
-            }
-
-            if (this.failedListeners.TryRemove(recoveredListener, out _))
-            {
-                TraceEventMessage message = CreateListenerRecoveryMessage(recoveredListener);
-                this.LogMessageToNonFailedListeners(message);
-            }
-        }
-
-        private void MarkAndLogListenerFailure(EventListener failedListener, string errorMessage)
-        {
-            if (!this.failedListeners.TryAdd(failedListener, errorMessage))
-            {
-                // We've already logged that this listener has failed so there is no need to do it again
-                return;
-            }
-
-            TraceEventMessage message = CreateListenerFailureMessage(failedListener, errorMessage);
-            this.LogMessageToNonFailedListeners(message);
         }
 
         private void LogMessageToNonFailedListeners(TraceEventMessage message)
@@ -405,7 +402,7 @@ namespace GVFS.Common.Tracing
             foreach (EventListener listener in this.listeners.Except(this.failedListeners.Keys))
             {
                 // To prevent infinitely recursive failures, we won't try and log that we failed to log that a listener failed :)
-                listener.TryRecordMessage(message, out _);
+                listener.RecordMessage(message);
             }
         }
     }
--- a/GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
+++ b/GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
@@ -1,0 +1,186 @@
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.IO.Pipes;
+using System.Text;
+using System.Threading;
+
+namespace GVFS.Common.Tracing
+{
+    public enum QueuedPipeStringWriterState
+    {
+        Unknown = 0,
+        Stopped = 1,
+        Failing = 2,
+        Healthy = 3,
+    }
+
+    /// <summary>
+    /// Accepts string messages from multiple threads and dispatches them over a named pipe from a
+    /// background thread.
+    /// </summary>
+    public class QueuedPipeStringWriter : IDisposable
+    {
+        private const int DEFAULT_MAX_QUEUE_SIZE = 256;
+
+        private readonly Func<NamedPipeClientStream> createPipeFunc;
+        private readonly IQueuedPipeStringWriterEventSink eventSink;
+        private readonly BlockingCollection<string> queue;
+
+        private Thread writerThread;
+        private NamedPipeClientStream pipeClient;
+        private QueuedPipeStringWriterState state = QueuedPipeStringWriterState.Unknown;
+        private bool isDisposed;
+
+        public QueuedPipeStringWriter(Func<NamedPipeClientStream> createPipeFunc, IQueuedPipeStringWriterEventSink eventSink, int maxQueueSize = DEFAULT_MAX_QUEUE_SIZE)
+        {
+            this.createPipeFunc = createPipeFunc;
+            this.eventSink = eventSink;
+            this.queue = new BlockingCollection<string>(new ConcurrentQueue<string>(), boundedCapacity: maxQueueSize);
+        }
+
+        public void Start()
+        {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(QueuedPipeStringWriter));
+            }
+
+            if (this.writerThread != null)
+            {
+                return;
+            }
+
+            this.writerThread = new Thread(this.BackgroundWriterThreadProc)
+            {
+                Name = nameof(QueuedPipeStringWriter),
+                IsBackground = true,
+            };
+
+            this.writerThread.Start();
+        }
+
+        public bool TryEnqueue(string message)
+        {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(QueuedPipeStringWriter));
+            }
+
+            return this.queue.TryAdd(message);
+        }
+
+        public void Stop()
+        {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(QueuedPipeStringWriter));
+            }
+
+            if (this.queue.IsAddingCompleted)
+            {
+                return;
+            }
+
+            // Signal to the queue draining thread that it should drain once more and then terminate.
+            this.queue.CompleteAdding();
+            this.writerThread.Join();
+
+            Debug.Assert(this.queue.IsCompleted, "Message queue should be empty after being stopped");
+        }
+
+        public void Dispose()
+        {
+            if (!this.isDisposed)
+            {
+                this.Stop();
+
+                this.pipeClient?.Dispose();
+                this.pipeClient = null;
+
+                this.writerThread = null;
+
+                this.queue.Dispose();
+            }
+
+            this.isDisposed = true;
+        }
+
+        private void RaiseStateChanged(QueuedPipeStringWriterState newState, Exception ex)
+        {
+            if (this.state != newState)
+            {
+                this.state = newState;
+                this.eventSink?.OnStateChanged(this, newState, ex);
+            }
+        }
+
+        private void BackgroundWriterThreadProc()
+        {
+            // Drain the queue of all messages currently in the queue.
+            // TryTake() using an infinite timeout will block until either a message is available (returns true)
+            // or the queue has been marked as completed _and_ is empty (returns false).
+            string message;
+            while (this.queue.TryTake(out message, Timeout.Infinite))
+            {
+                if (message != null)
+                {
+                    this.WriteMessage(message);
+                }
+            }
+
+            this.RaiseStateChanged(QueuedPipeStringWriterState.Stopped, null);
+        }
+
+        private void WriteMessage(string message)
+        {
+            // Create pipe if this is the first message, or if the last connection broke for any reason
+            if (this.pipeClient == null)
+            {
+                try
+                {
+                    // Create a new pipe stream instance using the provided factory
+                    NamedPipeClientStream pipe = this.createPipeFunc();
+
+                    // Specify a instantaneous timeout because we don't want to hold up the
+                    // background thread loop if the pipe is not available; we will just drop this event.
+                    // The pipe server should already be running and waiting for connections from us.
+                    pipe.Connect(timeout: 0);
+
+                    // Keep a hold of this connected pipe for future messages
+                    this.pipeClient = pipe;
+                }
+                catch (Exception ex)
+                {
+                    this.RaiseStateChanged(QueuedPipeStringWriterState.Failing, ex);
+                    return;
+                }
+            }
+
+            try
+            {
+                // If we're in byte/stream transmission mode rather than message mode
+                // we should signal the end of each message with a line-feed (LF) character.
+                if (this.pipeClient.TransmissionMode == PipeTransmissionMode.Byte)
+                {
+                    message += '\n';
+                }
+
+                byte[] data = Encoding.UTF8.GetBytes(message);
+                this.pipeClient.Write(data, 0, data.Length);
+                this.pipeClient.Flush();
+
+                this.RaiseStateChanged(QueuedPipeStringWriterState.Healthy, null);
+            }
+            catch (Exception ex)
+            {
+                // We can't send this message for some reason (e.g., broken pipe); we attempt no recovery or retry
+                // mechanism and drop this message. We will try to recreate/connect the pipe on the next message.
+                this.pipeClient.Dispose();
+                this.pipeClient = null;
+                this.RaiseStateChanged(QueuedPipeStringWriterState.Failing, ex);
+                return;
+            }
+        }
+    }
+}
--- a/GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
+++ b/GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
@@ -1,45 +1,47 @@
 ﻿using System;
 using System.IO.Pipes;
-using System.Text;
 using GVFS.Common.Git;
 using Newtonsoft.Json;
 
 namespace GVFS.Common.Tracing
 {
-    public class TelemetryDaemonEventListener : EventListener
+    public class TelemetryDaemonEventListener : EventListener, IQueuedPipeStringWriterEventSink
     {
-        private readonly string pipeName;
         private readonly string providerName;
         private readonly string enlistmentId;
         private readonly string mountId;
         private readonly string vfsVersion;
 
-        private NamedPipeClientStream pipeClient;
+        private QueuedPipeStringWriter pipeWriter;
 
         private TelemetryDaemonEventListener(
             string providerName,
             string enlistmentId,
             string mountId,
-            string pipeName)
-            : base(EventLevel.Verbose, Keywords.Telemetry)
+            string pipeName,
+            IEventListenerEventSink eventSink)
+            : base(EventLevel.Verbose, Keywords.Telemetry, eventSink)
         {
-            this.pipeName = pipeName;
             this.providerName = providerName;
             this.enlistmentId = enlistmentId;
             this.mountId = mountId;
             this.vfsVersion = ProcessHelper.GetCurrentProcessVersion();
+
+            this.pipeWriter = new QueuedPipeStringWriter(
+                () => new NamedPipeClientStream(".", pipeName, PipeDirection.Out, PipeOptions.Asynchronous),
+                this);
+            this.pipeWriter.Start();
         }
 
         public string GitCommandSessionId { get; set; }
 
-        public static TelemetryDaemonEventListener CreateIfEnabled(string gitBinRoot, string providerName, string enlistmentId, string mountId)
+        public static TelemetryDaemonEventListener CreateIfEnabled(string gitBinRoot, string providerName, string enlistmentId, string mountId, IEventListenerEventSink eventSink)
         {
             // This listener is disabled unless the user specifies the proper git config setting.
-
             string telemetryPipe = GetConfigValue(gitBinRoot, GVFSConstants.GitConfig.GVFSTelemetryPipe);
             if (!string.IsNullOrEmpty(telemetryPipe))
             {
-                return new TelemetryDaemonEventListener(providerName, enlistmentId, mountId, telemetryPipe);
+                return new TelemetryDaemonEventListener(providerName, enlistmentId, mountId, telemetryPipe, eventSink);
             }
             else
             {
@@ -49,14 +51,42 @@ namespace GVFS.Common.Tracing
 
         public override void Dispose()
         {
-            this.pipeClient?.Dispose();
+            if (this.pipeWriter != null)
+            {
+                this.pipeWriter.Stop();
+                this.pipeWriter.Dispose();
+                this.pipeWriter = null;
+            }
+
             base.Dispose();
+        }
+
+        void IQueuedPipeStringWriterEventSink.OnStateChanged(
+            QueuedPipeStringWriter writer,
+            QueuedPipeStringWriterState state,
+            Exception exception)
+        {
+            switch (state)
+            {
+                case QueuedPipeStringWriterState.Failing:
+                    this.RaiseListenerFailure(exception?.ToString());
+                    break;
+                case QueuedPipeStringWriterState.Healthy:
+                    this.RaiseListenerRecovery();
+                    break;
+            }
         }
 
         protected override void RecordMessageInternal(TraceEventMessage message)
         {
-            string json = this.SerializeMessage(message);
-            this.SendMessage(json);
+            string pipeMessage = this.CreatePipeMessage(message);
+
+            bool dropped = !this.pipeWriter.TryEnqueue(pipeMessage);
+
+            if (dropped)
+            {
+                this.RaiseListenerFailure("Pipe delivery queue is full. Message was dropped.");
+            }
         }
 
         private static string GetConfigValue(string gitBinRoot, string configKey)
@@ -74,16 +104,16 @@ namespace GVFS.Common.Tracing
             return value.TrimEnd('\r', '\n');
         }
 
-        private string SerializeMessage(TraceEventMessage message)
+        private string CreatePipeMessage(TraceEventMessage message)
         {
-            var pipeMessage = new TelemetryMessage
+            var pipeMessage = new PipeMessage
             {
                 Version = this.vfsVersion,
                 ProviderName = this.providerName,
                 EventName = message.EventName,
                 EventLevel = message.Level,
                 EventOpcode = message.Opcode,
-                Payload = new TelemetryMessage.TelemetryMessagePayload
+                Payload = new PipeMessage.PipeMessagePayload
                 {
                     EnlistmentId = this.enlistmentId,
                     MountId = this.mountId,
@@ -97,45 +127,7 @@ namespace GVFS.Common.Tracing
             return pipeMessage.ToJson();
         }
 
-        private void SendMessage(string message)
-        {
-            // Create pipe if this is the first message, or if the last connection broke for any reason
-            if (this.pipeClient == null)
-            {
-                var pipe = new NamedPipeClientStream(".", this.pipeName, PipeDirection.Out, PipeOptions.Asynchronous);
-
-                // Specify a instantaneous timeout because we don't want to hold up the rest of the
-                // application if the pipe is not available; we will just drop this event.
-                // We let any TimeoutExceptions bubble up and will try again with a new pipe on the next SendMessage call.
-                pipe.Connect(timeout: 0);
-
-                // Keep a hold of this connected pipe for future messages
-                this.pipeClient = pipe;
-            }
-
-            try
-            {
-                // If we're in byte/stream transmission mode rather than message mode
-                // we should signal the end of each message with a line-feed (LF) character.
-                if (this.pipeClient.TransmissionMode == PipeTransmissionMode.Byte)
-                {
-                    message += '\n';
-                }
-
-                var buffer = Encoding.UTF8.GetBytes(message);
-                this.pipeClient.Write(buffer, 0, buffer.Length);
-            }
-            catch (Exception)
-            {
-                // We can't send this message for some reason (e.g., broken pipe); we attempt no recovery or retry
-                // mechanism and drop this message. We will try to recreate/connect the pipe on the next message.
-                this.pipeClient.Dispose();
-                this.pipeClient = null;
-                throw;
-            }
-        }
-
-        public class TelemetryMessage
+        public class PipeMessage
         {
             [JsonProperty("version")]
             public string Version { get; set; }
@@ -148,11 +140,11 @@ namespace GVFS.Common.Tracing
             [JsonProperty("eventOpcode")]
             public EventOpcode EventOpcode { get; set; }
             [JsonProperty("payload")]
-            public TelemetryMessagePayload Payload { get; set; }
+            public PipeMessagePayload Payload { get; set; }
 
-            public static TelemetryMessage FromJson(string json)
+            public static PipeMessage FromJson(string json)
             {
-                return JsonConvert.DeserializeObject<TelemetryMessage>(json);
+                return JsonConvert.DeserializeObject<PipeMessage>(json);
             }
 
             public string ToJson()
@@ -160,7 +152,7 @@ namespace GVFS.Common.Tracing
                 return JsonConvert.SerializeObject(this);
             }
 
-            public class TelemetryMessagePayload
+            public class PipeMessagePayload
             {
                 [JsonProperty("enlistmentId")]
                 public string EnlistmentId { get; set; }
--- a/GVFS/GVFS.UnitTests/Tracing/EventListenerTests.cs
+++ b/GVFS/GVFS.UnitTests/Tracing/EventListenerTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using GVFS.Common.Tracing;
+using Moq;
 using NUnit.Framework;
 
 namespace GVFS.UnitTests.Tracing
@@ -8,53 +9,29 @@ namespace GVFS.UnitTests.Tracing
     public class EventListenerTests
     {
         [TestCase]
-        public void EventListener_TryRecordMessage_ExceptionThrownInternally_ReturnsFalseAndErrorMessage()
+        public void EventListener_RecordMessage_ExceptionThrownInternally_RaisesFailureEventWithErrorMessage()
         {
-            string expectedErrorMessage = "test error message";
+            string expectedErrorMessage = $"test error message unique={Guid.NewGuid():N}";
+
+            Mock<IEventListenerEventSink> eventSink = new Mock<IEventListenerEventSink>();
 
             TraceEventMessage message = new TraceEventMessage { Level = EventLevel.Error, Keywords = Keywords.None };
-            TestEventListener listener = new TestEventListener(EventLevel.Informational, Keywords.Any)
+            TestEventListener listener = new TestEventListener(EventLevel.Informational, Keywords.Any, eventSink.Object)
             {
                 RecordMessageInternalCallback = _ => throw new Exception(expectedErrorMessage)
             };
 
-            string actualErrorMessage;
-            bool? actualResult = listener.TryRecordMessage(message, out actualErrorMessage);
+            listener.RecordMessage(message);
 
-            Assert.IsFalse(actualResult);
-            Assert.IsTrue(actualErrorMessage.Contains(expectedErrorMessage));
-        }
-
-        [TestCase]
-        public void EventListener_TryRecordMessage_ListenerNotEnabledForMessage_ReturnsNull()
-        {
-            TraceEventMessage message = new TraceEventMessage { Level = EventLevel.Informational, Keywords = Keywords.None };
-            TestEventListener listener = new TestEventListener(EventLevel.Critical, Keywords.Telemetry);
-
-            string actualErrorMessage;
-            bool? actualResult = listener.TryRecordMessage(message, out actualErrorMessage);
-
-            Assert.IsNull(actualResult);
-            Assert.IsNull(actualErrorMessage);
-        }
-
-        [TestCase]
-        public void EventListener_TryRecordMessage_MessageSentSuccessfully_ReturnsTrue()
-        {
-            TraceEventMessage message = new TraceEventMessage { Level = EventLevel.Error, Keywords = Keywords.None };
-            TestEventListener listener = new TestEventListener(EventLevel.Informational, Keywords.Any);
-
-            string actualErrorMessage;
-            bool? actualResult = listener.TryRecordMessage(message, out actualErrorMessage);
-
-            Assert.IsTrue(actualResult);
-            Assert.IsNull(actualErrorMessage);
+            eventSink.Verify(
+                x => x.OnListenerFailure(listener, It.Is<string>(msg => msg.Contains(expectedErrorMessage))),
+                times: Times.Once);
         }
 
         private class TestEventListener : EventListener
         {
-            public TestEventListener(EventLevel maxVerbosity, Keywords keywordFilter)
-                : base(maxVerbosity, keywordFilter)
+            public TestEventListener(EventLevel maxVerbosity, Keywords keywordFilter, IEventListenerEventSink eventSink)
+                : base(maxVerbosity, keywordFilter, eventSink)
             {
             }
 
--- a/GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
+++ b/GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
@@ -1,0 +1,273 @@
+using System;
+using System.IO.Pipes;
+using System.Threading;
+using GVFS.Common.Tracing;
+using GVFS.Tests.Should;
+using Moq;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Tracing
+{
+    [TestFixture]
+    public class QueuedPipeStringWriterTests
+    {
+        [TestCase]
+        public void Stop_RaisesStateStopped()
+        {
+            // Capture event invocations
+            Mock<IQueuedPipeStringWriterEventSink> eventSink = new Mock<IQueuedPipeStringWriterEventSink>();
+
+            // createPipeFunc returns `null` since the test will never enqueue any messaged to write
+            QueuedPipeStringWriter writer = new QueuedPipeStringWriter(
+                () => null,
+                eventSink.Object);
+
+            // Try to write some dummy data
+            writer.Start();
+            writer.Stop();
+
+            eventSink.Verify(
+                x => x.OnStateChanged(writer, QueuedPipeStringWriterState.Stopped, null),
+                Times.Once);
+        }
+
+        [TestCase]
+        public void MissingPipe_RaisesStateFailing()
+        {
+            const string inputMessage = "FooBar";
+
+            // Capture event invocations
+            Mock<IQueuedPipeStringWriterEventSink> eventSink = new Mock<IQueuedPipeStringWriterEventSink>();
+
+            QueuedPipeStringWriter writer = new QueuedPipeStringWriter(
+                () => throw new Exception("Failing pipe connection"),
+                eventSink.Object);
+
+            // Try to write some dummy data
+            writer.Start();
+            bool queueOk = writer.TryEnqueue(inputMessage);
+            writer.Stop();
+
+            queueOk.ShouldBeTrue();
+            eventSink.Verify(
+                x => x.OnStateChanged(writer, QueuedPipeStringWriterState.Failing, It.IsAny<Exception>()),
+                Times.Once);
+            eventSink.Verify(
+                x => x.OnStateChanged(writer, QueuedPipeStringWriterState.Stopped, It.IsAny<Exception>()),
+                Times.Once);
+        }
+
+        [TestCase]
+        public void GoodPipe_WritesDataAndRaisesStateHealthy()
+        {
+            const string inputMessage = "FooBar";
+            byte[] expectedData =
+            {
+                0x46, 0x6F, 0x6F, 0x42, 0x61, 0x72, (byte)'\n', // "FooBar\n"
+                0x46, 0x6F, 0x6F, 0x42, 0x61, 0x72, (byte)'\n', // "FooBar\n"
+                0x46, 0x6F, 0x6F, 0x42, 0x61, 0x72, (byte)'\n', // "FooBar\n"
+            };
+
+            string pipeName = Guid.NewGuid().ToString("N");
+
+            // Capture event invocations
+            Mock<IQueuedPipeStringWriterEventSink> eventSink = new Mock<IQueuedPipeStringWriterEventSink>();
+
+            QueuedPipeStringWriter writer = new QueuedPipeStringWriter(
+                () => new NamedPipeClientStream(".", pipeName, PipeDirection.Out),
+                eventSink.Object);
+
+            using (TestPipeReaderWorker pipeWorker = new TestPipeReaderWorker(pipeName, PipeTransmissionMode.Byte))
+            {
+                // Start the pipe reader worker first and wait until the pipe server has been stood-up
+                // before starting the pipe writer/enqueuing messages because the writer does not wait
+                // for the pipe to be ready to accept (it returns and drops messages immediately).
+                pipeWorker.Start();
+                pipeWorker.WaitForReadyToAccept();
+                writer.Start();
+
+                // Try to write some dummy data
+                bool queueOk1 = writer.TryEnqueue(inputMessage);
+                bool queueOk2 = writer.TryEnqueue(inputMessage);
+                bool queueOk3 = writer.TryEnqueue(inputMessage);
+
+                // Wait until we've received all the sent messages before shuting down
+                // the pipe worker thread.
+                pipeWorker.WaitForRecievedBytes(count: expectedData.Length);
+                pipeWorker.Stop();
+                writer.Stop();
+
+                queueOk1.ShouldBeTrue();
+                queueOk2.ShouldBeTrue();
+                queueOk3.ShouldBeTrue();
+
+                byte[] actualData = pipeWorker.GetReceivedDataSnapshot();
+                CollectionAssert.AreEqual(expectedData, actualData);
+
+                // Should only receive one 'healthy' state change per successfully written message
+                eventSink.Verify(
+                    x => x.OnStateChanged(writer, QueuedPipeStringWriterState.Healthy, It.IsAny<Exception>()),
+                    Times.Once);
+                eventSink.Verify(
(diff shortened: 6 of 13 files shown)


GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -4,13 +4,15 @@
...
+using System.Threading;
 
 namespace GVFS.Common.Tracing
 {
     public class JsonTracer : ITracer
     {
         public const string NetworkErrorEventName = "NetworkError";
 
+        private readonly ReaderWriterLockSlim listenersLock = new ReaderWriterLockSlim();
         private readonly List<EventListener> listeners;
```
wilbaker: Is it an option to use a thread-safe collection (e.g. `ConcurrentBag`) rather than introducing a lock?  

If not, what are the scenario(s) where a thread-safe collection would be insufficient?

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -47,7 +48,11 @@ public static TelemetryDaemonEventListener CreateIfEnabled(string gitBinRoot, st
 
         public override void Dispose()
         {
-            this.pipeClient?.Dispose();
+            lock (this.pipeLock)
+            {
+                this.pipeClient?.Dispose();
```
wilbaker: Also set `this.pipeClient` to null after disposing?

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -96,39 +101,42 @@ private string SerializeMessage(TraceEventMessage message)
...
-                var pipe = new NamedPipeClientStream(".", this.pipeName, PipeDirection.Out, PipeOptions.Asynchronous);
-
-                // Specify a instantaneous timeout because we don't want to hold up the rest of the
-                // application if the pipe is not available; we will just drop this event.
-                // We let any TimeoutExceptions bubble up and will try again with a new pipe on the next SendMessage call.
-                pipe.Connect(timeout: 0);
+                // Create pipe if this is the first message, or if the last connection broke for any reason
+                if (this.pipeClient == null)
+                {
+                    var pipe = new NamedPipeClientStream(".", this.pipeName, PipeDirection.Out, PipeOptions.Asynchronous);
```
wilbaker: Nit: `var` -> `NamedPipeClientStream` (I see the code was like this before too, but might as well clean it up while we're in here)

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -96,39 +101,42 @@ private string SerializeMessage(TraceEventMessage message)
...
+                try
+                {
+                    // If we're in byte/stream transmission mode rather than message mode
+                    // we should signal the end of each message with a line-feed (LF) character.
+                    if (this.pipeClient.TransmissionMode == PipeTransmissionMode.Byte)
+                    {
+                        message += '\n';
+                    }
+
+                    var buffer = Encoding.UTF8.GetBytes(message);
```
wilbaker: `var` -> `byte[]`

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: There's a problem in `Dispose` where we enumerate `listeners` (calling `Dispose` on them in turn) and then clear the collection.

```csharp
foreach (EventListener listener in this.listeners)
{
    listener.Dispose();
}

this.listeners.Clear();
```


The enumeration with `foreach` is [fine because `GetEnumerator` returns a snapshot](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1.getenumerator?redirectedfrom=MSDN&view=netstandard-2.0#remarks), but when we later call `Clear` it's possible that a new listener has been added that is not going to be `Dispose`d of.

I could replace this with a loop doing `TryTake` while the collection is not empty as below..

```csharp
EventListener listener;
while (this.listeners.TryTake(out listener))
{
    listener.Dispose();
}
```

..but of course the failure of `TryTake` could be for reasons other than 'nothing left to take'/'collection is empty'.

This also raises the question about what exactly the point of `private bool disposed` is for in this class; we're not actually doing anything with it. Really we should be throwing `ObjectDisposedException` when attempting to interact with the `JsonTracer` (such as writing messages, or adding new listeners). To do that we'd also want to make the `this.disposed = true;` atomic with the clearing of the bag – we could do that by moving the `this.disposed = true` _before_ the `while(TryTake)`.

(In this case we'd also want to switch `private bool disposed` to `private int disposed` and use `Interlocked.CompareExchange` because there is no `bool` overload for CAS actions in .NET)

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -96,39 +102,42 @@ private string SerializeMessage(TraceEventMessage message)
 
         private void SendMessage(string message)
         {
-            // Create pipe if this is the first message, or if the last connection broke for any reason
-            if (this.pipeClient == null)
+            lock (this.pipeLock)
```
kewillford: What is the performance impact of this lock?  Especially in things like updating placeholders, enumerating directories, parsing the index, etc.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
wilbaker: > The enumeration with foreach is fine because GetEnumerator returns a snapshot, but when we later call Clear it's possible that a new listener has been added that is not going to be Disposed of.

I'm not sure that the lock is actually protecting us against missing a listener when we're disposing.

A call to add a new listener could execute immediately after the locked code block:

```
using (this.listenersLock.Read())
{
    foreach (EventListener listener in this.listeners)
    {
        listener.Dispose();
     }

    this.listeners.Clear();
}
```
And the new listener would not be disposed.

I have the same thoughts as @kewillford regarding performance, and unless we can show that we have a problem that can only be solved with locking I'd rather avoid the complexity introduced by adding locks.

I also think that with how specialized/critical our tracing code is, we'd be OK setting limitations on how callers are allowed to use `JsonTracer`.  For example, if either:

- `Dispose` is always called on the same thread as the thread adding listeners
-or-
- All listeners are added before sharing the `JsonTracer` with another thread

Then we might not need the complexity to handle the add/dispose race.

If we find we *do* need to handle that race though, then perhaps we could use a combination of locking in Add+Dispose and a thread-safe collection? 

That would allow lock-free reads of the collection and protection against crossing Adds and Dispose.

wilbaker: I just saw that @kewillford is actually referring to a different lock 😄

wilbaker: Sorry I left out an important part.  I do agree that the race condition you mentioned is best fixed with locking 👍

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: Unknown. I've not looked at it yet (this is still a draft PR for a reason).

It's also important to bear in mind that this event listener is only logging _telemetry_ information. I seriously hope we're not writing a trace message that's tagged at the 'telemetry' level for each and every file action.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: > I'm not sure that the lock is actually protecting us against missing a listener when we're disposing.

This is exactly what I called out in my comment:

> The enumeration with foreach is fine because GetEnumerator returns a snapshot, **_but when we later call Clear it's possible that a new listener has been added that is not going to be Disposed of._**
>
> ...
>
> This also raises the question about what exactly the point of private bool disposed is for in this class; we're not actually doing anything with it. Really we should be throwing ObjectDisposedException when attempting to interact with the JsonTracer (such as writing messages, or adding new listeners). **_To do that we'd also want to make the this.disposed = true; atomic with the clearing of the bag – we could do that by moving the this.disposed = true before the while(TryTake)._**

The solution I call out first prevents new listeners from being added by setting the disposed flag (with the various `Add` methods also checking `!disposed` using `Interlocked.CompareExchange`), then it's safe to enumerate the listener collection disposing and removing them.

> I have the same thoughts as @kewillford regarding performance, and unless we can show that we have a problem that can only be solved with locking I'd rather avoid the complexity introduced by adding locks.
>
> I also think that with how specialized/critical our tracing code is, we'd be OK setting limitations on how callers are allowed to use JsonTracer. 

Let's just ignore `JsonTracer` being thread *un*safe for now if this is too difficult.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: Plus even without locking, the fact we'd be writing a message over a pipe/socket for each and every file operation would most likely result in performance problems itself. We shouldn't be logging micro-events to telemetry.. only summary information.

kewillford: I agree.  This was more of a note to make sure we haven't introduced any regressions in performance with this change because with locking there is always that possibility.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
wilbaker: > The enumeration with foreach is fine because GetEnumerator returns a snapshot, but when we later call Clear it's possible that a new listener has been added that is not going to be Disposed of.

Ahh sorry, I misread that point as something that is addressed with the current changes (that requires locking).

> This also raises the question about what exactly the point of private bool disposed is for in this class; we're not actually doing anything with it.

I saw that we are using `disposed` to detect if anyone is writing to a disposed tracer:

```
        private void WriteEvent(string eventName, EventLevel level, Keywords keywords, EventMetadata metadata, EventOpcode opcode)
        {
            string jsonPayload = metadata != null ? JsonConvert.SerializeObject(metadata) : null;

            if (this.disposed)
            {
                Console.WriteLine("Writing to disposed tracer");
                Console.WriteLine(jsonPayload);

                throw new ObjectDisposedException(nameof(JsonTracer));
            }
```

And I agree with the point you've raised here:

> In this case we'd also want to switch private bool disposed to private int disposed and use Interlocked.CompareExchange because there is no bool overload for CAS actions in .NET

`disposed` is not currently atomic and so threads might read the wrong value.

And I also agree that we're not consistently checking `disposed` in all the places we could.

> Let's just ignore `JsonTracer` being thread unsafe for now if this is too difficult

I think it's great that you called out that `listeners` is not thread safe, and I think it's worth either confirming that our usage of it is currently safe or fixing it as needed.

> this is still a draft PR for a reason

I missed that this is a draft PR ☹️ Apologies for jumping in with comments that were not at the right level  for a draft.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
alameenshah: Is addressing performance issues in scope for this PR? We could cache the message and unblock the caller immediately. The cached messages can then be sent asynchronously.

mjcheetham: > We could cache the message and unblock the caller immediately. The cached messages can then be sent asynchronously.

At the minute we just drop failed messages on the floor upon failure.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: A `ConcurrentBag` has been added in lieu of a lock here for `JsonTracer`. The `disposed` flag has been updated to now be read/written atomically.

wilbaker:
@mjcheetham I have been holding off on reviewing this because it's still marked as Draft.  

Let me know when you're ready for me to take a look and I'll be happy to, thanks!

mjcheetham:
> @mjcheetham I have been holding off on reviewing this because it's still marked as Draft.
> 
> Let me know when you're ready for me to take a look and I'll be happy to, thanks!

Hey @wilbaker I've updated the PR to non-draft and changed to a lock-free approach in `JsonTracer`. The `TelemetryDaemonEventListener` also now does not hold up the calling threads sending the messages over the pipe – this is handled in a lock-avoiding background thread.

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: I've updated this to create a background thread and lock-free queue for posting the messages without holding up the calling threads.

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.IO.Pipes;
+using System.Threading;
+
+namespace GVFS.Common.NamedPipes
+{
+    public class QueuedPipeWriter : IDisposable
+    {
+        private static readonly TimeSpan PollingInterval = TimeSpan.FromSeconds(30);
```
mjcheetham: When the queue goes to sleep, there are two ways it gets woken up again:

1. a new message is posted,
2. 30 seconds pass.

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
```diff
@@ -0,0 +1,209 @@
...
+            {
+                throw new ObjectDisposedException(nameof(QueuedPipeWriter));
+            }
+
+            if (!this.isRunning)
+            {
+                throw new InvalidOperationException($"{nameof(QueuedPipeWriter)} is not started");
+            }
+
+            this.queue.Enqueue(data);
```
mjcheetham: No locking is required here as this is a `ConcurrentQueue<>` 😃

GVFS/GVFS.Common/Tracing/EventListener.cs
```diff
@@ -14,30 +14,27 @@ protected EventListener(EventLevel maxVerbosity, Keywords keywordFilter)
...
+        public event EventHandler ListenerRecovery;
+
+        public event EventHandler<string> ListenerFailure;
+
         public virtual void Dispose()
         {
         }
 
-        public bool? TryRecordMessage(TraceEventMessage message, out string errorMessage)
+        public void RecordMessage(TraceEventMessage message)
```
mjcheetham: Because the posting of events over the pipe is now disconnected from actual tracer message write, failure (and recovery) is now communicated using events.

GVFS/GVFS.Common/Tracing/EventListener.cs
```diff
@@ -14,30 +14,27 @@ protected EventListener(EventLevel maxVerbosity, Keywords keywordFilter)
             this.keywordFilter = keywordFilter;
         }
 
+        public event EventHandler ListenerRecovery;
```
mjcheetham: Because the posting of events over the pipe is now disconnected from actual tracer message write, failure (and recovery) is now communicated using events.

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
mjcheetham: Also note that I've set _no upper limit to the size of this queue_! This means that if many hundreds of threads send messages we'll have a lot of messages to churn through. Should I set an upper limit? Opinions for a magic number if we want this?

Note: we always dequeue messages if there are any, and if we fail to deliver them over the pipe we never requeue it.. in other words the queue will never grow forever.. it is always attempting to be drained.
(Plus if the `gvfs.telemetry-pipe` configuration variable isn't set we're never even executed to start with!)

kewillford: I think we should test this and see what the memory pressure is at how many messages.  If the dequeuing thread dies for some reason we would be queuing and it would grow forever.  Raises the question whether we should have a `try...catch` in the `ThreadProc`?

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+
+            if (!this.isRunning)
+            {
+                return;
+            }
+
+            // Signal to the queue draining thread that it should drain once more and then terminate.
+            // Setting isRunning to false will additionally prevent any new messages from being enqueued.
+            this.isRunning = false;
```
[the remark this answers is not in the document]
mjcheetham: > This looks like it will throw a `InvalidOperationException` if any messages are enqueued while waiting to stop right? Is that what we want to do?

Do you think we should just drop the messages instead (`return;`) rather than throw?

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -91,48 +101,25 @@ private string SerializeMessage(TraceEventMessage message)
                 // Other TraceEventMessage properties are not used
             };
 
-            return pipeMessage.ToJson();
+            string json = pipeMessage.ToJson();
+
+            return Encoding.UTF8.GetBytes(json);
```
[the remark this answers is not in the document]
mjcheetham: We could do that, yes. I might want to then move this `QueuedPipeWriter` into the `Tracing` namespace though since it becomes a little less generic at that point. It is currently just a 'binary/byte' `*Writer'.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -112,18 +128,27 @@ public void Dispose()
         {
             this.Stop(null);
 
+            if (Interlocked.CompareExchange(ref this.isDisposed, DISPOSED_TRUE, DISPOSED_FALSE) == DISPOSED_TRUE)
```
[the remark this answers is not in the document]
mjcheetham: No, because `Stop` writes an event via `WriteEvent` which throws when `this.isDisposed` is true.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -4,14 +4,17 @@
...
 using System.Linq;
+using System.Threading;
 
 namespace GVFS.Common.Tracing
 {
     public class JsonTracer : ITracer
     {
         public const string NetworkErrorEventName = "NetworkError";
+        private const int DISPOSED_FALSE = 0;
+        private const int DISPOSED_TRUE  = 1;
```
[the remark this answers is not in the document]
mjcheetham: I built on Mac and Windows for me.. so I'm assuming yes?

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Threading;
+
+namespace GVFS.Common.NamedPipes
+{
+    public class QueuedPipeWriter : IDisposable
+    {
+        private static readonly TimeSpan PollingInterval = TimeSpan.FromSeconds(30);
+
+        private readonly Func<NamedPipeClientStream> pipeFactory;
+        private readonly ConcurrentQueue<byte[]> queue = new ConcurrentQueue<byte[]>();
```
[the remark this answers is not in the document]
mjcheetham: I did not know it existed! Haha. I'll look at replacing it.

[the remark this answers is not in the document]
mjcheetham: Hmm.. replacing it with `BlockingCollection` seems a little more involved that first appears. More timing and sequencing bugs (for shutdown) are reappearing now and I'd rather not have to spend even more time hardening this.. :(

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
```diff
@@ -0,0 +1,164 @@
...
+                Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Stopped, eventCalls[1].State);
+            }
+        }
+
+        private class TestPipeWorker : IDisposable
+        {
+            private readonly string pipeName;
+            private readonly PipeTransmissionMode transmissionMode;
+            private readonly Thread thread;
+            private readonly List<byte[]> allData = new List<byte[]>();
```
jamill: Mainly a question - do we ever use the individual chunks of byte data read? Would it be less bookkeeping to just have a single list of bytes that we append to?

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -83,13 +90,22 @@ public bool HasLogFileEventListener
 
         public void AddEventListener(EventListener listener)
         {
+            if (Volatile.Read(ref this.isDisposed) == DISPOSED_TRUE)
```
jamill: This is the only place we have a `Volatile.XXX` operation - do other operations on this variable need to consider threading concerns (i.e. should all reads / writes to this variable be volatile)?

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
```diff
@@ -0,0 +1,164 @@
...
+                // Try to write some dummy data
+                writer.Start();
+                writer.Enqueue(inputMessage);
+                writer.Stop();
+
+                byte[] actualData = pipeWorker.GetAllData();
+                CollectionAssert.AreEqual(expectedData, actualData);
+
+                Assert.AreEqual(2, eventCalls.Count);
+                Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Healthy, eventCalls[0].State);
```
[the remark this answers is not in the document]
mjcheetham: There is logic already in JsonTracer to only fire 'recovery' and 'failure' once per change.
I could move this logic, or rather duplicate it here in the `QueuedPipeStringWriter` too, so that state change events only fire once per change.
The logic in the `JsonTracer` would need to remain to stop reporting repeated recovery/failure for other tracers.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
mjcheetham: Nope, we do not. But this is only test code, and it makes multiple debugging messages easier.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: All `isDisposed` operations are using `Volatile` or `Interlocked` in `JsonTracer`.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,211 @@
...
+                // we should signal the end of each message with a line-feed (LF) character.
+                if (this.pipeClient.TransmissionMode == PipeTransmissionMode.Byte)
+                {
+                    message += '\n';
+                }
+
+                byte[] data = Encoding.UTF8.GetBytes(message);
+                this.pipeClient.Write(data, 0, data.Length);
+
+                this.RaiseStateChanged(QueuedPipeStringWriterState.Healthy, message, null);
```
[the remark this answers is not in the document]
mjcheetham: Yes, it does.. but in `JsonTracer` we de-dup any changes to overall `EventListener` health.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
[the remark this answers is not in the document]
wilbaker: I was surprised by this too, I wonder if this is due to the changes in #572 

@jamill, do you know if there is any GUI available to easily see what all of the rules IDs correspond to in `GVFS.ruleset`?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,211 @@
...
+                return;
+            }
+
+            // Signal to the queue draining thread that it should drain once more and then terminate.
+            // Setting isRunning to false will additionally prevent any new messages from being enqueued.
+            this.isRunning = false;
+            this.queuePulse.Set();
+            this.writerThread.Join();
+
+            Debug.Assert(this.queue.IsEmpty, "Message queue should be empty after being stopped");
```
wilbaker: If this is an actual error/bad condition, could we throw an exception instead of adding a `Debug.Assert`?

If this assert were to start firing I'm not sure that we'd catch it (given our release/validation process) before shipping VFS4G (in general the code relies on unit tests and exceptions rather than asserts to ensure that expected conditions are being enforced).

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,211 @@
...
+        public void Start()
+        {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(QueuedPipeStringWriter));
+            }
+
+            if (this.isRunning)
+            {
+                return;
```
wilbaker: Should we throw something like an `InvalidOperationException` here rather that returning?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,211 @@
...
+        public void Stop()
+        {
+            if (this.isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(QueuedPipeStringWriter));
+            }
+
+            if (!this.isRunning)
+            {
+                return;
```
wilbaker: Throw an exception here too? Or are there scenarios where we expect `Stop` to be called multiple times (or on a `QueuedPipeStringWriter` that was never started)?

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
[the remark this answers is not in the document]
wilbaker: > Hmm.. replacing it with BlockingCollection seems a little more involved that first appears. More timing and sequencing bugs (for shutdown) are reappearing now and I'd rather not have to spend even more time hardening this

What sort of sequencing issues are you running into?  I had the same thought as @kewillford when I saw this code, `BlockingCollection` should allow you to remove `PollingInterval` and `queuePulse`.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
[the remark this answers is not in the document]
kewillford: `Stop` is also using the `this.duration` which may have already been disposed.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
wilbaker: Could we mark `this.isDisposed` as `volatile` instead of using `Volatile.Read`?  (That's what we've done elsewhere in the code where we need volatile reads, and I wasn't sure how/if `Volatile.Read` was different)

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
jamill: > it makes multiple debugging messages easier.

Thanks - that was mainly my question - I didn't know if there was a reason for it to be implemented this way.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
jamill: You are right - I got mixed up with the dispose variables in different files. Thanks!

jamill: Another question - we have the `Volatile` / `Interlocked` actions on this variable, but I think we can still have race conditions. What if we:

1) Enter `AddEventListener(...)`
2)    Pass the check for `isDisposed` (and we are not disposed), but immediately after is interrupted 
3) Another method disposes the Tracer, and completes the dispose method
4) The rest this method executes - but the object is disposed.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
wilbaker: Actually, this `Debug.Assert` could be removed completely if we switched to using `BlockingCollection` as this scenario would be impossible.

I'm happy to help if you've run into problems with `BlockingCollection`, as I've found it can really simply scenarios like this.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -340,28 +365,18 @@ private void WriteEvent(string eventName, EventLevel level, Keywords keywords, E
...
-                    if (success == true)
-                    {
-                        this.MarkAndLogListenerRecovery(listener);
-                    }
-                    else
-                    {
-                        this.MarkAndLogListenerFailure(listener, errorMessage);
-                    }
-                }
+                listener.RecordMessage(message);
```
jamill: I have mixed opinions on using events to signal that a problem occurred... I liked how the previous pattern handled these errors inline in the the caller, or it least it seemed simpler to follow then the event pattern. As it is, are there other subscribers to the success / failure events?

(I do like how the new pattern avoids the `TryXXX` pattern for this method, as we expect it to generally succeed).

GVFS/GVFS.Common/Tracing/JsonTracer.cs
wilbaker: Are `AddEventListener` and `Dispose` ever called on different threads?

If not, I'm not sure that we need to make those methods thread safe.

It seems like in most places the pattern is:

```
using (JsonTracer tracer = new JsonTracer(...))
{
    // Add event listeners

    // Do stuff
}
```

Have you found spots in the code where we're not doing this?  If there are spots that are not thread safe, would it be simpler to update that code to be safe (rather than making `AddEventListener` and `Dispose` thread-safe)?

wilbaker: > You are right - I got mixed up with the dispose variables in different files. Thanks!

I'm going to unresolve this comment as I'm still wondering about `volatile` vs `Volatile.Read`

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: Do you have another suggestion how to report failures asynchronously/from the other thread/detached from the initial method call that enqueued the data?

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: > Have you found spots in the code where we're not doing this? If there are spots that are not thread safe, would it be simpler to update that code to be safe (rather than making AddEventListener and Dispose thread-safe)?

@wilbaker I'll just back out changes in `JsonTracer` regarding Add/Dispose.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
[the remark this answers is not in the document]
mjcheetham: What do you suggest?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: What's the harm in calling `Start()` twice, with the 2nd being a no-op? (other than it being a technical coding error)

Given your argument regarding `JsonTracer`..

> Are AddEventListener and Dispose ever called on different threads?
>
> If not, I'm not sure that we need to make those methods thread safe.
>
> It seems like in most places the pattern is:
>
> ```
> using (JsonTracer tracer = new JsonTracer(...))
> {
>     // Add event listeners
> 
>     // Do stuff
> }
> ```
>
> Have you found spots in the code where we're not doing this? If there are spots that are not thread safe, would it be simpler to update that code to be safe (rather than making AddEventListener and Dispose thread-safe)?

..we 'know' we're not calling `Start()` twice in the codebase, and it's not a problem if we just no-op.

If we should be throwing `InvalidOperationException` here for a double start (to catch a would-be coding error), then why is it OK to ignore capturing the introduction of a coding error in `JsonTracer` (adding listeners interleaved with dispose/other ops)?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: Same question as this [comment](https://github.com/Microsoft/VFSForGit/pull/935/files#r275741684).

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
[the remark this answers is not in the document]
mjcheetham: I'll also add basic de-duping here in `QueuedPipeStringWriter`. It won't be thread safe since the only place that does a `RaiseStateChanged(..)` is the single background thread.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
jamill: Could the caller collect and handle the failures?

Maybe I missed it (I will try and look again when I am back at computer), but was there a deficiency in the original pattern where this was handled in the loop that delivered messages to the listeners?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
jamill: I think it is ok to no-op the double call unless - the object is already started so the desired end state is fulfilled. Maybe consider including an assert to catch on debug builds if we don’t expect double calls and want to be notified of this in test builds...

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: That write loop is still happening, but now it's invoked from the 'failure' event.
Previously it came in from the _return value_ of the `TryRecordMessage` call.. But since we don't want to hold this call up, we don't know the outcome of the invocation until some indeterminate time later.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: > Maybe consider including an assert to catch on debug builds if we don’t expect double calls and want to be notified of this in test builds...

This is how I see assertions too! 😃
A check for something that should not happen, but it would be OK if it did (or there's a mitigation for it.. for example doing a no-op). The assertion code itself is also a form a part-enforcing documentation about what assumptions have been made by the developer at the time.
Unless developers are running release builds all the time, assertions will be run.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
```diff
@@ -0,0 +1,213 @@
...
+            writer.Stop();
+
+            Assert.IsTrue(queueOk);
+            Assert.AreEqual(2, eventCalls.Count);
+            Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Failing, eventCalls[0].State);
+            Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Stopped, eventCalls[1].State);
+        }
+
+        [TestCase]
+        public void GoodPipe_WritesDataAndRaisesStateHealthy()
```
mjcheetham: @jeschu1 this is the test that is failing on macOS but not in Windows.
My current theory is that the bug is in the test code and the `TestPipeWorker` or pipe-writer is not waiting for each other to be connected. Probably need to gate `TestPipeWorker::GetAllData` on the worker thread completing the train of the pipe.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
wilbaker: > Unless developers are running release builds all the time, assertions will be run.

Do you know if assertions cause a failure of the functional test runs on the build machines if they fire?

I ask because I would not be surprised if developers run release builds more frequently than debug builds.  Unless I'm debugging a new feature I'm working on I typically install the official releases and rely on the functional tests (on the build machines) to validate my changes.

>  What's the harm in calling Start() twice, with the 2nd being a no-op? (other than it being a technical coding error)

I probably should have included what prompted this comment 😃 

I saw that we are throwing an exception in `Start`/`Stop` when `this.isDisposed` is `true`:

```
            if (this.isDisposed)
            {
                throw new ObjectDisposedException(nameof(QueuedPipeStringWriter));
            }
```

And so I assumed the idea was that if any of our expected pre-conditions are not met we'd throw an exception.

I can see that calling `Start`/`Stop` twice is worse than calling Start`/`Stop` after `Dispose`, and so I don't think this is a big deal.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
mjcheetham: Actually.. it's because my `TestPipeWorker` needs to wait for the pipe to be drained before calling `GetAllData`.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
wilbaker: And regarding the topic of assertions in general:

> A check for something that should not happen, but it would be OK if it did (or there's a mitigation for it.. for example doing a no-op).

Personally I've run into the following issues with assertions in practice:

- Assertions can slip by because they aren't noticed in release builds (perhaps we could add our own version that would fire a telemetry event and trigger an alert?)
- Although it might be OK for something to happen from the perspective of one component (e.g. calling `Start` twice in a row) it's difficult to know if the bigger issue that triggered that condition is safe/OK.  I'd rather make those errors as noticeable/obvious as possible to reduce the odds of shipping a bad bug.
- Unfortunately I've seen that if an assertion can be ignored it's not uncommon for a developer to ignore it (especially if it's unrelated to whatever the developer is working on), and that debt keeps getting carried until there's enough of a problem that it has to get paid down. 

I see your point that we don't validate everything we could (e.g. validating that start/stop/dispose are called on the same thread), and I don't think we want to get into that business.  If a condition is easy to check for (and easy enough that we can add an assertion) my own preference is that we make the issue more obvious by using an exception.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
jeschu1: @mjcheetham, a fun one I missed!  So I guess it just drains faster on windows for whatever reason.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
jamill: It sounds like you are against the idea of assertions in general. Personally, I have found them helpful in asserting conditions during development :) If not used appropriately, then there can be problems... but that is true to just about anything.

> If a condition is easy to check for (and easy enough that we can add an assertion) my own preference is that we make the issue more obvious by using an exception.

I think there is a balance with coding defensively where appropriate. Assertions can also help clarify intent of the code as well.

Is this included in the coding guidelines?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            Unknown = 0,
+            Stopped = 1,
+            Failing = 2,
+            Healthy = 3,
+        }
+
+        public void Start()
+        {
+            if (this.isDisposed)
```
jeschu1: What is the driver is checking for this.isDisposed in every call?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
jamill: > Do you know if assertions cause a failure of the functional test runs on the build machines if they fire?

If not, it might be good to run tests with debug / extra assertion checking enabled.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            int prevState = Interlocked.Exchange(ref this.state, (int)newState);
+
+            if (prevState != (int)newState)
+            {
+                this.StateChanged?.Invoke(this, new QueuedPipeStringWriterStateEventArgs(newState,  ex));
+            }
+        }
+
+        private void ThreadProc()
```
jeschu1: Nit: Maybe BackgroundThreadWritter or something more descriptive?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            // Drain the queue of all messages currently in the queue
+            do
+            {
+                try
+                {
+                    // Take() is a blocking call that will return a message when there is one available
+                    // or will throw an InvalidOperationException when the queue has been marked as completed.
+                    string message = this.queue.Take();
+                    if (message != null)
```
jeschu1: Should null messages ever be put on the queue?  I guess since it's part of the logger itself it would hard to log that condition.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,216 @@
...
+
+            if (!this.isRunning)
+            {
+                return;
+            }
+
+            // Signal to the queue draining thread that it should drain once more and then terminate.
+            this.isRunning = false;
+            this.queue.CompleteAdding();
+            this.writerThread.Join();
```
jeschu1: Why is it important to drain whatever is one the queue?  Would we refuse new requests?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
wilbaker: > It sounds like you are against the idea of assertions in general. 
> Is this included in the coding guidelines?

No this is something that was missed when the initial guidelines were written.  There are a few spots in VFS4G that we have asserts, but historically the guidance/rule has been to avoid them.
> with debug / extra assertion checking enabled

How do you enable this setting?  That sounds like it will be useful to turn on.

> If not used appropriately, then there can be problems... but that is true to just about anything.

Good point 😄 

> I think there is a balance with coding defensively where appropriate. Assertions can also help clarify intent of the code as well.

That's a good point as well, and I think the key is getting the balance right.

What would be the guidelines that you'd recommend?

wilbaker: To close this discussion out, I am okay with us just returning here (or doing an assert + return).

What prompted this comment was seeing that we were being so aggressive with checking `this.isDisposed`, and wondering if we should be more aggressive when checking if `Start`/`Stop` were called multiple times.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,216 @@
...
+                        this.WriteMessage(message);
+                    }
+                }
+                catch (InvalidOperationException)
+                {
+                    // The queue is both empty *and* no longer accepting new messages
+                    Debug.Assert(this.queue.IsCompleted, "Queue should be marked as completed");
+                }
+            }
+            while (this.queue.Count > 0 || this.isRunning);
```
kewillford: Should this be `while (!this.queue.IsCompleted)` and remove the `isRunning` member? `TryAdd` will still throw when the  `CompleteAdding` is called so can we get away with using that as the flag?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: It would be odd for the following to be true:

```csharp
using (var writer = new QueuedPipeStringWriter(..))
{
    writer.Start();
    writer.Enqueue("message1");
    writer.Enqueue("message2");
    writer.Stop();
}

// Pipe gets message1, but not message2 because it had not been sent before Stop().
```

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: That's just going to be a limitation I'm enforcing here. It makes no sense to log null. Think about what we'd send over the pipe? Zero bytes.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: It's good measure. You should not interact with an `IDisposable` instance once it's been disposed. Especially something like `Start`/`Stop`.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
jeschu1: That's fair enough, I have some questions on handling when we can't add.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: > What prompted this comment was seeing that we were being so aggressive with checking `this.isDisposed`, and wondering if we should be more aggressive when checking if `Start`/`Stop` were called multiple times.

Coding errors around disposal I treat as very serious; they're more serious than a redundant call. Take for example the two pieces of bad C code:

```c
char *str = (char *) malloc(7);
strcpy(str, "foobar");
free(str);
puts(str); // use-after-free
```

```c
void *buf = (void *) malloc(32);
memset(p, 0, 16);
memset(p, 0, 16); // redundant call
```

The latter isn't critical and is in practice a no-op. The former is a serious bug.

Failing to dispose, or interacting with an object after disposal in the managed world is a big no-no(!)

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
```diff
@@ -47,14 +46,27 @@ public static TelemetryDaemonEventListener CreateIfEnabled(string gitBinRoot, st
...
         {
-            string json = this.SerializeMessage(message);
-            this.SendMessage(json);
+            string pipeMessage = this.CreatePipeMessage(message);
+
+            bool dropped = !this.pipeWriter.TryEnqueue(pipeMessage);
+
+            if (dropped)
+            {
+                this.RaiseListenerFailure("Pipe delivery queue is full. Message was dropped.");
```
jeschu1: Didn't find great documentation on when TryAdd would return false.  Seems kinda scary we can fill the queue and drop messages.  Is this the instance where we return false?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: True, but we've no way now to prevent `TryEnqueue` when we've not yet called `Start` which sets-up and starts the draining thread. Perhaps that would be a desirable feature? (the ability to populate a queue of messages before dispatching them)

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
jeschu1: We could call dispose right after this 'if' check right?  I guess how much protection does this give us?  Are there cases we could hit it?

GVFS/GVFS.Common/Tracing/TelemetryDaemonEventListener.cs
mjcheetham: > Didn't find great documentation on when TryAdd would return false.

[`TryAdd(T)`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1.tryadd?view=netframework-4.7.2#System_Collections_Concurrent_BlockingCollection_1_TryAdd__0_)

> **Remarks**
> If the collection is a bounded collection, and is full, this method immediately returns false without adding the item.

--

> Seems kinda scary we can fill the queue and drop messages. Is this the instance where we return false?

Yes. In the old implementation we didn't have any throttling since we are synchronously sending (attempting to send) messages over the pipe. Now we're sending messages in the background it is possible we might be getting more messages in faster than can be drained, so we need to prevent an explosion of memory/forever increasing message queue.

This is only used in the telemetry event listener so the events we're happy with dropping - they're not critical. We're already saying dropping a message on send _failure_ is OK.
At least here we'll be surfacing (to the other event listeners) that the message queue is too full.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: > We could call dispose right after this 'if' check right? 

You are correct. I'm saying (similar to the argument made in `JsonTracer`) that we're going to assume disposal and initalisation are single-threaded and 'correct', meaning we're going to gloss over it 😄 .. we only care about the TryEnqueue part.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
mjcheetham: Fixed this up by adding more signalling between the test and the test pipe worker thread, to ensure the test pipe server is actually ready to accept messages.

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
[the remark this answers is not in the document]
mjcheetham: I've replaced this with `BlockingCollection<T>`.

GVFS/GVFS.Common/NamedPipes/QueuedPipeWriter.cs
mjcheetham: There is now an upper limit to the queue of 256 messages. Given in testing, messages are almost immediately dispatched (and each message is typically only about 400-800 _bytes_ in size), I don't expect there to be a backlog of unsent (or not-yet-dropped) messages.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
kewillford: Thinking about this and another option would be for the JsonTracer to implement an interface like:
```
public interface IListenerHealthHandler
{
    void Recovered(EventListener listener);
    void Failed(EventListener listener, string errorMessage);
}
```

And set that on each listener.  Pretty much the same as an event handler but more explicit and there could be only one.

jamill: > But since we don't want to hold this call up, we don't know the outcome of the invocation until some indeterminate time later.

Thanks for the clarification!

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
```diff
@@ -0,0 +1,265 @@
...
+                CollectionAssert.AreEqual(expectedData, actualData);
+
+                // Should only receive one 'healthy' state change per successfully written message
+                Assert.AreEqual(2, eventCalls.Count);
+                Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Healthy, eventCalls[0].State);
+                Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Stopped, eventCalls[1].State);
+            }
+        }
+
+        private class TestPipeReaderWorker : IDisposable
```
kewillford: This seems like a lot of code to have in a unit test.  Can we instead abstract the `pipeClient` into an interface that is returned by `createPipeFunc` instead of the `NamedPipeClientStream` and use a mock of the interface?  Not sure how feasible this is or how much work it would be vs saved code here.  Let me know what you think.

kewillford:
I think this is looking really good.  Only had a couple questions on if we should swap the events and handlers for an interface also if we can simplify the unit tests with more abstraction.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
```diff
@@ -0,0 +1,265 @@
...
+            // Capture event invocations
+            List<QueuedPipeStringWriter.QueuedPipeStringWriterStateEventArgs> eventCalls = new List<QueuedPipeStringWriter.QueuedPipeStringWriterStateEventArgs>();
+            writer.StateChanged += (_, e) => eventCalls.Add(e);
+
+            // Try to write some dummy data
+            writer.Start();
+            writer.Stop();
+
+            Assert.AreEqual(1, eventCalls.Count);
+            Assert.AreEqual(QueuedPipeStringWriter.QueuedPipeStringWriterState.Stopped, eventCalls[0].State);
```
wilbaker: Comment applies to all the `Assert` checks in the file:

Switch from using `Assert` syntax to using the `Should` syntax that's used elsewhere in the unit tests (you'll need to add a `using GVFS.Tests.Should`)

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,208 @@
...
+                    {
+                        this.WriteMessage(message);
+                    }
+                }
+                catch (InvalidOperationException) when (this.queue.IsCompleted)
+                {
+                    // The queue is both empty *and* no longer accepting new messages
+                }
+            }
+            while (!this.queue.IsCompleted);
```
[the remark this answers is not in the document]
wilbaker: A bit more background here, `Take()` is actually implemented [with a call to `TryTake`](https://referencesource.microsoft.com/#System/sys/system/collections/concurrent/BlockingCollection.cs,550):

```
        public T Take()
        {
            T item;
 
            if (!TryTake(out item, Timeout.Infinite, CancellationToken.None))
            {
                throw new InvalidOperationException(SR.GetString(SR.BlockingCollection_CantTakeWhenDone));
            }
 
            return item;
        }
```

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,208 @@
...
+
+            if (prevState != (int)newState)
+            {
+                this.StateChanged?.Invoke(this, new QueuedPipeStringWriterStateEventArgs(newState,  ex));
+            }
+        }
+
+        private void ThreadProc()
+        {
+            // Drain the queue of all messages currently in the queue
```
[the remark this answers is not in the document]
mjcheetham: > [..] with a call to Environment.Exit in the catch.

I don't think we want to bring the mount process down for this sort of failure in telemetry?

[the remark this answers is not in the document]
wilbaker: >I don't think we want to bring the mount process down for this sort of failure in telemetry?

With the current code what would happen to the mount process if there were an unhandled exception in `ThreadProc`?

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -83,13 +86,22 @@ public bool HasLogFileEventListener
 
         public void AddEventListener(EventListener listener)
         {
+            if (this.isDisposed)
+            {
+                Console.WriteLine("Adding listener to disposed tracer");
```
jamill: In the "common" case, the mount process will not have a console...  I think Json tracer might already write to console, but is this what we want?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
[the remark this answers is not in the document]
mjcheetham: I mean, yes it would bring down the process as is. I agree with want the `try/catch`.
What I'm not sure about is the `Environment.Exit(1)`.. I don't think we should terminate the entire mount process because the telemetry background thread hit a problem.

This raises another question about recovery there then too.. should this thread attempt recovery?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,208 @@
...
+
+                this.writerThread = null;
+            }
+
+            this.isDisposed = true;
+        }
+
+        private void RaiseStateChanged(QueuedPipeStringWriterState newState, Exception ex)
+        {
+            int prevState = Interlocked.Exchange(ref this.state, (int)newState);
```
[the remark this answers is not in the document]
mjcheetham: Correct! Will make it an enum.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,208 @@
...
+                    }
+                }
+                catch (InvalidOperationException) when (this.queue.IsCompleted)
+                {
+                    // The queue is both empty *and* no longer accepting new messages
+                }
+            }
+            while (!this.queue.IsCompleted);
+
+            this.RaiseStateChanged(QueuedPipeStringWriterState.Stopped, null);
```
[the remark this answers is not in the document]
mjcheetham: That is correct. I should move the un-subscription to after `Stop`, to mirror the subscription that is prior to `Start`.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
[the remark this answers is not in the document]
mjcheetham: > A bit more background here, Take() is actually implemented with a call to TryTake:

Aha! That's useful to know. Thanks. I'll make the suggested change.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: Will remove this. I added it to be consistent with the other 'is disposed' test that already existed in the file. Looks like that was debugging code that shouldn't have been in there in the first place. Will remove this.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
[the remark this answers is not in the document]
wilbaker: \+ @jrbriggs for his thoughts on this as well.

In general how much should VFS4G explode if it expects telemetry to be working and finds that it's not?

GVFS/GVFS.Common/Tracing/EventListener.cs
mjcheetham: We're using an event-sink object pattern rather than events now.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: @kewillford I've updated the PR to replace events with an event sink object & interface as suggested. I went with the term 'event sink' rather than 'handler'.

GVFS/GVFS.Common/Tracing/EventListener.cs
```diff
@@ -8,36 +8,41 @@ public abstract class EventListener : IDisposable
...
 
+        private IEventListenerEventSink eventSink;
+
         protected EventListener(EventLevel maxVerbosity, Keywords keywordFilter)
         {
             this.maxVerbosity = maxVerbosity;
             this.keywordFilter = keywordFilter;
         }
 
+        public void RegisterEventSink(IEventListenerEventSink eventSink)
```
kewillford: Can we have the interface come through the constructor instead of a separate method.  `AddEventListener` is the only place it is used where it doesn't have the construction of the listener but that method is only used in tests and could be `private` if we didn't have the tests which I don't think we should remove so I think we need to keep it but can we skip the `RegisterEventSink` there and leave the responsibility to the constructor of the event listener?

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,195 @@
...
+        private QueuedPipeStringWriterState state = QueuedPipeStringWriterState.Unknown;
+        private bool isDisposed;
+
+        public QueuedPipeStringWriter(Func<NamedPipeClientStream> createPipeFunc, int maxQueueSize = DEFAULT_MAX_QUEUE_SIZE)
+        {
+            this.createPipeFunc = createPipeFunc;
+            this.queue = new BlockingCollection<string>(new ConcurrentQueue<string>(), boundedCapacity: maxQueueSize);
+        }
+
+        public void RegisterEventSink(IQueuedPipeStringWriterEventSink eventSink)
```
kewillford: Again can this interface come in the constructor so we don't need this method or the check for already being registered?

GVFS/GVFS.UnitTests/Tracing/EventListenerTests.cs
```diff
@@ -8,47 +9,24 @@ namespace GVFS.UnitTests.Tracing
...
 
             TraceEventMessage message = new TraceEventMessage { Level = EventLevel.Error, Keywords = Keywords.None };
             TestEventListener listener = new TestEventListener(EventLevel.Informational, Keywords.Any)
             {
                 RecordMessageInternalCallback = _ => throw new Exception(expectedErrorMessage)
             };
 
-            string actualErrorMessage;
-            bool? actualResult = listener.TryRecordMessage(message, out actualErrorMessage);
+            Moq.Mock<IEventListenerEventSink> eventSink = new Moq.Mock<IEventListenerEventSink>();
```
kewillford: Don't need the `Moq.` since there is a `using`.  Also need to pass in `MockBehavior.Strict` with at `Setup` and call `VerifyAll` at the end.  This should be for all the `Mock` objects unless there is some reason not to follow those standards?

GVFS/GVFS.Common/Tracing/EventListener.cs
```diff
@@ -8,36 +8,41 @@ public abstract class EventListener : IDisposable
...
+        {
+            if (this.eventSink != null)
+            {
+                throw new Exception($"Already registered an event sink for event listener {this.GetType().Name}");
```
jamill: Would `InvalidOperationException` be more appropriate? Not that it really matters...

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,195 @@
...
+
+                    // Specify a instantaneous timeout because we don't want to hold up the
+                    // background thread loop if the pipe is not available; we will just drop this event.
+                    // The pipe server should already be running and waiting for connections from us.
+                    pipe.Connect(timeout: 0);
+
+                    // Keep a hold of this connected pipe for future messages
+                    this.pipeClient = pipe;
+                }
+                catch (Exception ex)
```
jamill: Is there a more specific set of expected exceptions we could catch here?

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
jamill: > This seems like a lot of code to have in a unit test.

Is the concern that this test is a bit large for a unit test? or that the test itself is not appropriate? If I understand correctly, it seems nice that we have a test that is exercising a bit more of the real components, even if it is a bit larger scope than desired for a unit test (i.e. would you prefer the test live "somewhere else" - not sure where that would be...).

GVFS/GVFS.Common/Tracing/EventListener.cs
mjcheetham: Moved to passing the sinks in via the constructors.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: Done!

GVFS/GVFS.UnitTests/Tracing/EventListenerTests.cs
mjcheetham: > Don't need the Moq. since there is a using.
Yep!

> Also need to pass in MockBehavior.Strict
Yep!

> with at Setup and call VerifyAll at the end. This should be for all the Mock objects unless there is some reason not to follow those standards?
We'd like to verify the number of times these events were raised (to ensure we only get events raised when the state _changed_) which is not possible using the `Setup && VerifyAll` pattern, only with the `Verify(expr, times)` call.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: The mantra of the telemetry-over-pipe tracing was to ensure we _never ever_ brought down the mount process just because of a problem with telemetry. Yes, we expect only 'TimeoutException' or maybe 'BrokenPipe' (or maybe some permission/security exceptions too).

The comment raised by @wilbaker about the `ThreadProc` method and calling `Environment.Exit(1)` on `catch` to bring down the entire process is an open question here, waiting on @jrbriggs's input about if the "don't crash at all costs" idea is still valid 😃

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
kewillford: I guess my question would be what are we trying to test?  I looks like we are testing that the state change gets raised and the message gets "somewhere".  Right now the "somewhere" is a pipe and we have a bunch of code with a thread and locks and reset events which are all complex things to get right and understand but would it make it easy and less code if the "somewhere" was an interface so it could be a pipe class or a mock or something else? I'm not going to hold up the PR for this just something to think about because too many abstractions is also an issue that we don't want to get into.

GVFS/GVFS.Common/Tracing/JsonTracer.cs
```diff
@@ -288,6 +300,35 @@ public void WriteStartEvent(EventMetadata metadata, Keywords keywords)
             this.WriteEvent(this.activityName, this.startStopLevel, keywords, metadata, EventOpcode.Start);
         }
 
+        void IEventListenerEventSink.OnListenerRecovery(EventListener listener)
+        {
+            // Check ContainsKey first (rather than always calling TryRemove) because ContainsKey
+            // is lock-free and recoveredListener should rarely be in failedListeners
+            if (!this.failedListeners.ContainsKey(listener))
```
jamill: One last comment - I spent some time trying to understand whether this code could be called on multiple threads at the same time, or how it would interact with logic running asynchronously elsewhere. I think it can be, but I didn't see anythink that could obviously go wrong (i.e. - could we add a listener to the failed list, but the listener has already recovered so that it would stay on the failed list....)

Anyway - I think this all works, but I don't know if a comment around how we expect (or why we expect) that this is correct would be helpful.

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
mjcheetham: > I guess my question would be what are we trying to test? I looks like we are testing that the state change gets raised and the message gets "somewhere".

Correct.

> [..] if the "somewhere" was an interface so it could be a pipe class or a mock or something else? I'm not going to hold up the PR for this just something to think about because **too many abstractions** is also an issue that we don't want to get into.

So in this case I'd need to wrap `NamedPipeClientStream` which includes construction, connection, and sending/writing functions. This feels a little bit like too many abstractions.
Ideally I could use a `Stream` (and then pass in a `MemoryStream` or something in testing.. but only the pipe streams have the notion of connection).

The `TestPipeReaderWorker` is mocking out the telemetry service behaviour which is to be running ahead of mount, waiting for a connection and draining the pipe until it gets shutdown.

kewillford: Okay sounds good. We might want to eventually move this outside of this class so that other tests that deal with pipes can use it.

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
```diff
@@ -0,0 +1,186 @@
...
+            if (!this.isDisposed)
+            {
+                this.Stop();
+
+                this.pipeClient?.Dispose();
+                this.pipeClient = null;
+
+                this.writerThread = null;
+
+                this.queue.Dispose();
```
wilbaker: I noticed we're setting `this.pipeClient` to `null`, but not `this.queue`.  Should `this.queue` be set to `null` as well?

mjcheetham: `this.queue` is currently `readonly` so it cannot be set to `null` without removing this.

In practice the GC will still collect the queue object just as quickly even though this reference still exists since once the reference to the `QueuedPipeStringWriter` is gone. Also given this will be used by the `TelemetryDaemonEventListener` which lives for the lifetime of a `JsonTracer`, which itself lives for the lifetime of the process means these are all going to be Generation 2 objects so are less likely to be reclaimed until a full collection happens.. plus if we're being disposed it means the process is being shutdown in this case.

wilbaker: > this.queue is currently readonly so it cannot be set to null without removing this.

Thanks I missed that detail!

GVFS/GVFS.Common/Tracing/JsonTracer.cs
mjcheetham: > i.e. - could we add a listener to the failed list, but the listener has already recovered so that it would stay on the failed list....

Correct, and on the next successful message from that listener we'd then detect and log recovery (just one message later). Unless we have listeners that are always going up/down/up/down between each message this will be a non-issue (and even then the issue would only be we didn't always log the failures and/or recoveries).

GVFS/GVFS.UnitTests/Tracing/QueuedPipeStringWriterTests.cs
mjcheetham: Absolutely. At the moment the test class has only been designed with these use-cases in mind so I'd rather not pull it up-and-out yet until it's needed. Then we can worry about if it's general purpose. 😁

GVFS/GVFS.Common/Tracing/QueuedPipeStringWriter.cs
mjcheetham: Updated!

Issue #928, opened 2019-03-14
Prompts for authentication for each new repo


bk2204:
I have multiple repositories using VFSforGit stored on the same server. Each time I try to clone a new repository with `gvfs clone`, I'm prompted to log in again and must enter my username, password, and a 2FA code.

Instead, VFSforGit should prompt once to acquire a personal access token and store it in the credential manager for reuse across all repositories on that server, as upstream Git does. If the user requires different credentials for different repositories, then they can set the `credential.useHttpPath` option in their configuration to specify this, and VFSforGit can then prompt them each time.

jamill:
Hi!

Great feedback. We effectively force `credential.useHttpPath` for all of our VFS4G repositories currently, as some remotes require it to authenticate as expected - in particular internal repositories that we support. This is unnecessary (and has negative consequences) in general.

I think one option is for our team to handle setting the `useHttpPath` as an internal concern (e.g. in our internal setup) rather than pushing this behavior into the general behavior for VFSForGit.

jamill:
Addressed with #939

PR #939, opened 2019-03-15
Fix credential interactions

### LINKED ISSUE #916: VFSForGit does not erase credentials on Authentication failures
VFSForGit has logic to erase and re-acquire crendentials if VFSForGit fails to authenticate against the service, but this appears to not be working.

Initial investigation indicates the problem is that VFSForGit made a change to use the full URL when acquiring credentials `usehttppath=true`, but this same setting is not used when erasing credentials.

jamill:
In testing, there were 2 changes necessary to reject credentials:

1) Make sure the `usehttppath` option is specified, so that the credential lookup and rejection use the same context

2) The call to reject credentials with username / password was not rejecting the credential (with GCM 1.18.5).

Additionally, since `usehttppath` only needs to be applied to `https://dev.azure.com` urls, we scope this setting. This should improve the experience when other remote URLs are used.

Fixes #916, #928  
Related: #949

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -179,7 +179,7 @@ namespace GVFS.Common.Git
             string stdinConfig = sb.ToString();
 
             this.InvokeGitOutsideEnlistment(
-                "credential reject",
+                GenerateCredentialVerbCommand("reject"),
                 stdin => stdin.Write(stdinConfig),
                 null);
         }
@@ -195,7 +195,7 @@ namespace GVFS.Common.Git
             string stdinConfig = sb.ToString();
 
             this.InvokeGitOutsideEnlistment(
-                "credential approve",
+                GenerateCredentialVerbCommand("approve"),
                 stdin => stdin.Write(stdinConfig),
                 null);
         }
@@ -269,7 +269,7 @@ namespace GVFS.Common.Git
             using (ITracer activity = tracer.StartActivity("TryGetCredentials", EventLevel.Informational))
             {
                 Result gitCredentialOutput = this.InvokeGitAgainstDotGitFolder(
-                    $"-c {GitConfigSetting.CredentialUseHttpPath}=true credential fill",
+                    GenerateCredentialVerbCommand("fill"),
                     stdin => stdin.Write($"url={repoUrl}\n\n"),
                     parseStdOutLine: null);
 
@@ -758,6 +758,11 @@ namespace GVFS.Common.Git
             {
                 this.executingProcess = null;
             }
+        }
+
+        private static string GenerateCredentialVerbCommand(string verb)
+        {
+            return $"-c {GitConfigSetting.CredentialUseHttpPath}=true credential {verb}";
         }
 
         private static string ParseValue(string contents, string prefix)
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -179,7 +179,7 @@ public virtual void RejectCredentials(string repoUrl, string username = null, st
             string stdinConfig = sb.ToString();
 
             this.InvokeGitOutsideEnlistment(
-                "credential reject",
+                GenerateCredentialVerbCommand("reject"),
```
jeschu1: How easy/hard would it  be to add unit tests in this space?

jeschu1: @mjcheetham hahaha perfect :-)

jamill:
@turbonaitis - this should address the issues you brought up in #949

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -760,6 +755,11 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
             }
         }
 
+        private static string GenerateCredentialVerbCommand(string verb)
+        {
+            return $"-c {GitConfigSetting.CredentialUseHttpPath}=true credential {verb}";
```
kewillford: Why are we not just setting this in the git config for the repo?  It shouldn't affect other git commands right? And should only be used for `dev.azure.com`?  Maybe we should check the repo url on clone and mount and set it in the config?  How will this affect clones done against GitHub?

jamill: Thanks Kevin -

> Why are we not just setting this in the git config for the repo?

1) It is set on mount
2) Not all config commands that deal with credentials are run in the context of a repository.

> And should only be used for dev.azure.com

It is scoped to dev.azure.com, so will only affect dev.azure.com urls. Maybe tweaking the variable name would make this more clear? There could be a discussion about whether we should even set this if we know the URL is not a dev.azure.com URL, but:

1) It doesn't hurt to always set this (git will take it and only apply it to the expected urls)
2) It is arguably more consistent to always apply this in VFS4G
3) It is closer to where we want to end eventually

kewillford: Okay thanks Jameson.  Can we just make add this to the const and not have this helper method like so
`public const string CredentialUseHttpPath = "-c credential.\"https://dev.azure.com\".useHttpPath=true";`
And then use a
`$"{CredentialUseHttpPath} credential fill"` etc in the invocations of git?

GVFS/GVFS.UnitTests/Git/GitAuthenticationTests.cs
```diff
@@ -262,19 +262,19 @@ private MockGitProcess GetGitProcess()
             int approvals = 0;
             int rejections = 0;
             gitProcess.SetExpectedCommandResult(
-                "-c credential.useHttpPath=true credential fill",
+                "-c credential.\"https://dev.azure.com\".useHttpPath=true credential fill",
```
kewillford: Add `const` for `-c credential.\"https://dev.azure.com\".useHttpPath=true` to use in all the tests?

GVFS/GVFS.Common/Git/GitProcess.cs
jamill: I like this suggestion, but we also use the the setting name directly in `TrySetRequiredGitConfigSettings`:

```
{ GitConfigSetting.CredentialUseHttpPath, "true" },
```

kewillford: Then how about another `const` for `CommandLineDevAzureCredentialUseHttpPath` that is
`$"-c {CredentialUseHttpPath}=true";` or similar?

Issue #150, opened 2018-08-09
Port GVFS.FunctionalTests.LockHolder to .NET Core


wilbaker:

wilbaker:
Resolved by #941

PR #941, opened 2019-03-18
.Net core porting - GVFS.FunctionalTests.LockHolder

### LINKED ISSUE #701: Mac: Detection of dead git processes (during lock acquire\release) does not work
```
~/BugBashRepos/Office.Prototype/src>git checkout - .
^C
~/BugBashRepos/Office.Prototype/src>git checkout -- .
Waiting for 'git checkout - .' to release the lock.../
```

I confirmed that that the ^C did kill the git process.

alameenshah:
Porting FunctionalTests.LockHolder to .netcore
- Migrated the project file to new format 
- Removed GVFS.Platform.Windows dependency 
- Directly using WindowsFileSystem.Shared.cs 
- Updated GitHelper to launch GVFS.FunctionalTests.LockHolder.dll using dotnet command. Previously it was an exe and launched directly 
- Replaced WindowsPlatform code (GVFSPlatform.Instance.GetNamedPipeName) with private implementation 
- Removed Categories.MacTODO.NeedsLockHolder 7.Implemented IsProcessActiveImplementation on MacPlatform.

Fixes #701, #150

### DIFF
--- a/GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
+++ b/GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
@@ -1,9 +1,11 @@
 ﻿using CommandLine;
 using GVFS.Common;
 using GVFS.Common.NamedPipes;
+using GVFS.Platform.Mac;
 using GVFS.Platform.Windows;
 using System;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 
 namespace GVFS.FunctionalTests.LockHolder
 {
@@ -20,20 +22,14 @@ namespace GVFS.FunctionalTests.LockHolder
 
         public void Execute()
         {
-            string normalizedCurrentDirectory;
             string errorMessage;
-            if (!WindowsFileSystem.TryGetNormalizedPathImplementation(Environment.CurrentDirectory, out normalizedCurrentDirectory, out errorMessage))
-            {
-                throw new Exception("Unable to get normalized path: " + errorMessage);
-            }
-
             string enlistmentRoot;
-            if (!WindowsPlatform.TryGetGVFSEnlistmentRootImplementation(Environment.CurrentDirectory, out enlistmentRoot, out errorMessage))
+            if (!TryGetGVFSEnlistmentRootImplementation(Environment.CurrentDirectory, out enlistmentRoot, out errorMessage))
             {
                 throw new Exception("Unable to get GVFS Enlistment root: " + errorMessage);
             }
 
-            string enlistmentPipename = GVFSPlatform.Instance.GetNamedPipeName(enlistmentRoot);
+            string enlistmentPipename = GetNamedPipeNameImplementation(enlistmentRoot);
 
             AcquireLock(enlistmentPipename);
 
@@ -43,6 +39,43 @@ namespace GVFS.FunctionalTests.LockHolder
             {
                 ReleaseLock(enlistmentPipename, enlistmentRoot);
             }
+        }
+
+        private static bool TryGetGVFSEnlistmentRootImplementation(string directory, out string enlistmentRoot, out string errorMessage)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return MacPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
+            }
+
+            // Not able to use WindowsPlatform here - because of its dependency on WindowsIdentity (and also kernel32.dll).
+            enlistmentRoot = null;
+
+            string finalDirectory;
+            if (!WindowsFileSystem.TryGetNormalizedPathImplementation(directory, out finalDirectory, out errorMessage))
+            {
+                return false;
+            }
+
+            enlistmentRoot = Paths.GetRoot(finalDirectory, GVFSConstants.DotGVFS.Root);
+            if (enlistmentRoot == null)
+            {
+                errorMessage = $"Failed to find the root directory for {GVFSConstants.DotGVFS.Root} in {finalDirectory}";
+                return false;
+            }
+
+            return true;
+        }
+
+        private static string GetNamedPipeNameImplementation(string enlistmentRoot)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return MacPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
+            }
+
+            // Not able to use WindowsPlatform here - because of its dependency on WindowsIdentity (and also kernel32.dll).
+            return "GVFS_" + enlistmentRoot.ToUpper().Replace(':', '_');
         }
 
         private static void AcquireLock(string enlistmentPipename)
--- a/GVFS/GVFS.FunctionalTests.LockHolder/GVFS.FunctionalTests.LockHolder.csproj
+++ b/GVFS/GVFS.FunctionalTests.LockHolder/GVFS.FunctionalTests.LockHolder.csproj
@@ -1,87 +1,42 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(SolutionDir)\GVFS\GVFS.Build\GVFS.cs.props" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\GVFS.Build\GVFS.cs.props" />
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">x64</Platform>
-    <ProjectGuid>{FA273F69-5762-43D8-AEA1-B4F08090D624}</ProjectGuid>
     <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <Platforms>x64</Platforms>
+    <!-- see https://github.com/NuGet/Home/issues/4837 for reference to CopyLocalLockFileAssemblies -->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
+    <RuntimeIdentifiers>win-x64;osx-x64</RuntimeIdentifiers>
+  </PropertyGroup>
+  
+  <PropertyGroup>
     <RootNamespace>GVFS.FunctionalTests.LockHolder</RootNamespace>
     <AssemblyName>GVFS.FunctionalTests.LockHolder</AssemblyName>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <NuGetPackageImportStamp>
-    </NuGetPackageImportStamp>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
-    <PlatformTarget>x64</PlatformTarget>
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x64' ">
+    <Version>$(GVFSVersion)</Version>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|x64' ">
-    <PlatformTarget>x64</PlatformTarget>
-    <DebugType>pdbonly</DebugType>
-    <Optimize>true</Optimize>
-    <DefineConstants>TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x64' ">
+    <Version>$(GVFSVersion)</Version>
   </PropertyGroup>
+  
   <ItemGroup>
-    <Reference Include="CommandLine, Version=2.0.275.0, Culture=neutral, PublicKeyToken=de6f01bd326f8c32, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\..\packages\CommandLineParser.2.1.1-beta\lib\net45\CommandLine.dll</HintPath>
-      <Private>True</Private>
-    </Reference>
-    <Reference Include="System" />
-    <Reference Include="System.ComponentModel.Composition" />
-    <Reference Include="System.Console, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\System.Console.4.0.0\lib\net46\System.Console.dll</HintPath>
-    </Reference>
-    <Reference Include="System.Core" />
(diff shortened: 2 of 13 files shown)


wilbaker:
@alameenshah can you also try removing `NeedsLockHolder` from `Categories.MacTODO` (and all of the code that uses it)?

The tests that require the LockHolder should pass on Mac.  If they're failing, we'll need a new MacTODO category and work item to track fixing the test(s).  

If the failures are something minor, e.g. using Windows paths instead of `Path.Combine` you might as well fix them now.

wilbaker:
Resolves #150

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
```diff
@@ -45,6 +41,45 @@ public void Execute()
...
 
+        private static bool TryGetGVFSEnlistmentRootImplementation(string directory, out string enlistmentRoot, out string errorMessage)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return MacPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
+            }
+            else
+            {
+                enlistmentRoot = null;
```
alameenshah: Not able to use WindowsPlatform here - because of its dependency on WindowsIdentity. Let me know if there is a better approach.

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
```diff
@@ -45,6 +41,45 @@ public void Execute()
...
+
+        private static string GetNamedPipeNameImplementation(string enlistmentRoot)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return MacPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
+            }
+            else
+            {
+                return "GVFS_" + enlistmentRoot.ToUpper().Replace(':', '_');
```
alameenshah: Not able to use WindowsPlatform here - because of its dependency on WindowsIdentity (and also kernel32.dll). Let me know if there is an alternate approach.

Scripts/Mac/CleanupFunctionalTests.sh
```diff
@@ -3,6 +3,7 @@
 pkill -9 -l GVFS.FunctionalTests
 pkill -9 -l git
 pkill -9 -l gvfs
+pkill -9 -l GVFS.Mount
```
alameenshah: This is an unrelated change - Functional tests were failing when a previous test was cancelled.

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
wilbaker: The only alternative that jumps out to me is moving `TryGetGVFSEnlistmentRootImplementation` and `GetNamedPipeNameImplementation` into a new `WindowsPlatform.Shared.NetStandard.cs` file that has only code that can compile into .NET Standard.

With how limited this is though I think you'd be OK just adding a comment here that describes why the code was copied rather than calling `WindowsPlatform`.  As general rule the functional test code doesn't consume product code anyway, this project is the one exception.

wilbaker: Once everything moves to .NET Core for the WindowsIdentity issue we can look at the compatibility pack described here: https://devblogs.microsoft.com/dotnet/announcing-the-windows-compatibility-pack-for-net-core/

GVFS/GVFS.FunctionalTests.LockHolder/GVFS.FunctionalTests.LockHolder.csproj
```diff
@@ -1,87 +1,42 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\GVFS.Build\GVFS.cs.props" />
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">x64</Platform>
-    <ProjectGuid>{FA273F69-5762-43D8-AEA1-B4F08090D624}</ProjectGuid>
     <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <Platforms>x64</Platforms>
+	<!-- see https://github.com/NuGet/Home/issues/4837 for reference to CopyLocalLockFileAssemblies -->
```
nickgra: nit: ws around the comment

GVFS/GVFS.FunctionalTests.LockHolder/GVFS.FunctionalTests.LockHolder.csproj
```diff
@@ -1,87 +1,42 @@
...
     <RootNamespace>GVFS.FunctionalTests.LockHolder</RootNamespace>
     <AssemblyName>GVFS.FunctionalTests.LockHolder</AssemblyName>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <NuGetPackageImportStamp>
-    </NuGetPackageImportStamp>
   </PropertyGroup>
+
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x64' ">
```
nickgra: Can we keep the spacing consistent in the PropertyGroups for all of our SDK style projects? VSMac has been sensitive to this in the past.
GVFS.Common has
`  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">`

GVFS/GVFS.FunctionalTests.LockHolder/GVFS.FunctionalTests.LockHolder.csproj
```diff
@@ -1,87 +1,42 @@
...
-    <ProjectReference Include="..\GVFS.Platform.Windows\GVFS.Platform.Windows.csproj">
-      <Project>{4CE404E7-D3FC-471C-993C-64615861EA63}</Project>
-      <Name>GVFS.Platform.Windows</Name>
-    </ProjectReference>
-    <ProjectReference Include="..\GVFS.Virtualization\GVFS.Virtualization.csproj">
-      <Project>{f468b05a-95e5-46bc-8c67-b80a78527b7d}</Project>
-      <Name>GVFS.Virtualization</Name>
-    </ProjectReference>
+    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
+	<ProjectReference Include="..\GVFS.Virtualization\GVFS.Virtualization.csproj" />
```
nickgra: ws on this line

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
```diff
@@ -27,7 +28,19 @@ private static string LockHolderCommandPath
         {
             get
             {
-                return Path.Combine(Settings.Default.CurrentDirectory, LockHolderCommand);
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
nickgra: Can you leave a comment here about why we need this change? (the differences between Windows and Mac and how we run the functional tests)

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
nickgra: I agree with William; let's leave a comment here on why we're doing it this way (and that we're intentionally duplicating code). Probably best to leave this comment as a TODO(Mac) so we'll find it whenever we end up moving everything to .NET Core.

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
```diff
@@ -15,7 +17,19 @@ public static bool IsElevatedImplementation()
...
-            // TODO(Mac): Implement proper check
+            try
+            {
+                Process process = Process.GetProcessById(processId);
+            }
+            catch (ArgumentException)
+            {
+                return false;
+            }
+            catch (InvalidOperationException)
```
wilbaker: Minor: On the Windows platform we [don't catch](https://github.com/Microsoft/VFSForGit/blob/6dca004e2361927f96f2ebdbff3c351b4695bd38/GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs#L61) `InvalidOperationException`, and I am guessing because this appears to be [a bug in the documentation](https://stackoverflow.com/questions/52519030/net-getprocessbyid-when-is-invalidoperationexception-thrown).

As best I can tell this won't get thrown in [the .NET Core version on Mac](https://github.com/dotnet/corefx/blob/71122bb4921e0239ea679dd275d53695b94e52f4/src/System.Diagnostics.Process/src/System/Diagnostics/Process.cs#L998) either, and so I don't think we need to catch it here.

wilbaker: (If you'd prefer to match the docs I recommend updating the Windows code to match instead)

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
```diff
@@ -45,6 +41,45 @@ public void Execute()
             }
         }
 
+        private static bool TryGetGVFSEnlistmentRootImplementation(string directory, out string enlistmentRoot, out string errorMessage)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return MacPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
+            }
+            else
```
wilbaker: Minor: There's no need for this `else` because of the earlier `return`.

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
alameenshah: removed the un-necessary catch

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
```diff
@@ -45,6 +41,43 @@ public void Execute()
...
+
+        private static string GetNamedPipeNameImplementation(string enlistmentRoot)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return MacPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
+            }
+
+            // Not able to use WindowsPlatform here - because of its dependency on WindowsIdentity (and also kernel32.dll).
+            return "GVFS_" + enlistmentRoot.ToUpper().Replace(':', '_');
```
kewillford: There is now code that we have to keep in sync.  Add a comment here pointing to the code in `WindowsPlatform`  and a comment in `WindowsPlatform` pointing to this code so that if we ever change one we will know to change the other.

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
```diff
@@ -18,7 +19,7 @@ public static class GitHelpers
         /// GVFS.FunctionalTests.LockHolder program.
         /// </summary>
         private const string LockHolderCommandName = @"GVFS.FunctionalTests.LockHolder";
-        private const string LockHolderCommand = @"GVFS.FunctionalTests.LockHolder.exe";
+        private const string LockHolderCommand = @"GVFS.FunctionalTests.LockHolder.dll";
```
kewillford: Aren't we going to run a `dotnet publish` for this which will make it a `.exe` on Windows and no extension on Mac?

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
```diff
@@ -126,7 +145,20 @@ public static void CheckGitCommandAgainstGVFSRepo(string virtualRepoRoot, string
...
+            string args = LockHolderCommandPath;
+            if (skipReleaseLock)
+            {
+                args += " --skip-release-lock";
+            }
+
+            return RunCommandWithWaitAndStdIn(
+                enlistment,
+                resetTimeout,
+                "dotnet",
```
kewillford: So we are running the `dotnet` command and passing the lock holder as an arg?  Why can't we just run the lockholder?

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
```diff
@@ -15,7 +17,15 @@ public static bool IsElevatedImplementation()
 
         public static bool IsProcessActiveImplementation(int processId)
         {
-            // TODO(Mac): Implement proper check
+            try
+            {
+                Process process = Process.GetProcessById(processId);
```
kewillford: Do we need to check that the `process` is not null or match the processId that we passed in?  Is an exception the only indication that the process is not active?

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
wilbaker: (I'l jump in here since this was my recommendation in offline discussions with @alameenshah)

The Windows build does not currently have a `dotnet publish` step, and so it was simpler to have the tests call `dotnet` for the lock holder (like we do for the [unit tests](https://github.com/Microsoft/VFSForGit/blob/8c8927ff421d8181a12b7c53de76c1defd80092c/Scripts/RunUnitTests.bat#L9) and [functional tests](https://github.com/Microsoft/VFSForGit/blob/8c8927ff421d8181a12b7c53de76c1defd80092c/Scripts/RunFunctionalTests.bat#L42) on Windows).

We *do* have a publish step on Mac (and could launch the lock holder there without `dotnet`) but we can reduce the amount of Mac and Windows specific code by using `dotnet` on both.

CC: @nickgra for his thoughts on this approach

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
wilbaker: This is because of the conversation in the other comment, Windows does not have a `dotnet publish` step.

GVFS/GVFS.FunctionalTests.LockHolder/AcquireGVFSLock.cs
alameenshah: Agree. We discussed this - please see @wilbaker's comments above. 

(With how limited this is though I think you'd be OK just adding a comment here that describes why the code was copied rather than calling WindowsPlatform. As general rule the functional test code doesn't consume product code anyway, this project is the one exception.

Once everything moves to .NET Core for the WindowsIdentity issue we can look at the compatibility pack described here: https://devblogs.microsoft.com/dotnet/announcing-the-windows-compatibility-pack-for-net-core/)

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
alameenshah: the extension would be `.dll` on both Mac and Windows. Even after Publish.

kewillford: Are there any gotchas or potential differences in running one way or the other?

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
alameenshah: after the migration to .netcore, the extension now is `.dll` and became easier to launch it using `dotnet`

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
alameenshah: No need to check `process`. We already follow this pattern in  `WindowsPlatform.Shared::IsProcessActiveImplementation` and also in `GitReadAndGitLockTests::OrphanedGVFSLockIsCleanedUp()`)

GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs
nickgra: In my mind, the risk of running the publish binaries has always been that if you're iterating in the IDE on some changes, you're only updating the BuildOutput version of the binaries, not the ones that live in Publish. We could publish as part of every build, but that'd slow down our inner loop.

Maybe that's a risk that is imagined, but I can see myself making that mistake at least :)

PR #949, opened 2019-03-21
Credential http path fix

### LINKED ISSUE #948: Approving/Rejecting credentials is not passing useHttpPath config
When we call `git credential fill`, we add `-c credential.useHttpPath=true`
https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Common/Git/GitProcess.cs#L272
While we simply call `git credential approve` and `git credential reject`, without this parameter
https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Common/Git/GitProcess.cs#L198
https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Common/Git/GitProcess.cs#L182

Because of this, git removes the http path part, before calling the git credential helper
https://github.com/Microsoft/git/blob/vfs-2.20.1/credential.c#L97

This causes interesting behavior, depending on configured credential helper.
This could cause #916 

In general, we should probably not always force `useHttpPath` to `true`, but either always respect whatever is configured in the git config, or at least give a configurable option to respect that.

turbonaitis:
Fixes #948

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -179,7 +179,7 @@ namespace GVFS.Common.Git
             string stdinConfig = sb.ToString();
 
             this.InvokeGitOutsideEnlistment(
-                "credential reject",
+                $"-c {GitConfigSetting.CredentialUseHttpPath}=true credential reject",
                 stdin => stdin.Write(stdinConfig),
                 null);
         }
@@ -195,7 +195,7 @@ namespace GVFS.Common.Git
             string stdinConfig = sb.ToString();
 
             this.InvokeGitOutsideEnlistment(
-                "credential approve",
+                $"-c {GitConfigSetting.CredentialUseHttpPath}=true credential approve",
                 stdin => stdin.Write(stdinConfig),
                 null);
         }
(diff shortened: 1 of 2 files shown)


turbonaitis:
Actually, probably the right thing to do here is to stop explicitly setting `credential.useHttpPath`, as this was introduced in this commit:
https://github.com/Microsoft/VFSForGit/commit/9d6ec5352cd783f14be8b4193b6468885300151f
which clearly addresses only VSTS, but not the rest of the world. Also, the setting is still available via git config.

Let me know, if you agree with the above and I'll amend the PR.

jamill:
Thank you for this - it is similar (in parts) to #939 (your way may be more succinct).

> Actually, probably the right thing to do here is to stop explicitly setting credential.useHttpPath

I agree - this was necessary to support Azure DevOps repositories. The challenge we had was that not all of our users accessing Azure DevOps repositories will have the `credential.useHttpPath` set in their git config. We now have an alternative mechanism where VFS4G does not need to explicitly set this, but this requires some reaction work on our team's side (See #928 as well). 

My proposal - instead of changing the code to explicitly set `useHttpPath` to all credential operations, I will make the necessary changes on our side so VFS4G doesn't have to explicitly set it, and then we can remove this logic from VFS4G. I will try to make those changes this week. I can also make the corresponding changes in the VFS4G side and let you know when they are up (unless you wanted to make the changes...)

turbonaitis:
As #939 addresses the root issue + reduces the scope of the setting, I'm closing this PR.

PR #950, opened 2019-03-21
NuGetUpgrader: defer credential acquisition

jamill:
The goal of this change is to enable NuGet Upgrader to defer credential acquisition until it performs an operation that requires authentication. This will enable the functionality of the NuGet Upgrader that does not require authentication to be used without having a credential.

The initial design of NuGet upgrader was that it would have the credentials passed into it, and it was not involved in credential management. Later, NuGetUpgrader became responsible for re-acquiring credentials if they were no longer valid, but this logic was not covered by tests. With this change, NuGet Upgrader will be responsible for the entire credential management lifecycle, and this will be more testable.

### Changes
This change introduces an `ICredentialStore` interface, which includes just the methods required to interact with the credentials on a system. This allows tests to pass in this dependency to NuGet upgrader during tests.

NuGet Upgrader now defers getting credentials until it performs an operation that requires credentials

### DIFF
--- a/GVFS/GVFS.Common/Git/GitAuthentication.cs
+++ b/GVFS/GVFS.Common/Git/GitAuthentication.cs
@@ -66,7 +66,7 @@ namespace GVFS.Common.Git
                         string password;
                         if (TryParseCredentialString(this.cachedCredentialString, out username, out password))
                         {
-                            this.git.ApproveCredentials(this.repoUrl, username, password);
+                            this.git.TryStoreCredential(tracer, this.repoUrl, username, password, out string error);
                             this.isCachedCredentialStringApproved = true;
                         }
                         else
@@ -97,7 +97,7 @@ namespace GVFS.Common.Git
                     string password;
                     if (TryParseCredentialString(this.cachedCredentialString, out username, out password))
                     {
-                        this.git.RejectCredentials(this.repoUrl, username, password);
+                        this.git.TryDeleteCredential(tracer, this.repoUrl, username, password, out string error);
                     }
                     else
                     {
@@ -109,7 +109,7 @@ namespace GVFS.Common.Git
                             ["CredentialString"] = this.cachedCredentialString
                         });
                         tracer.RelatedWarning(metadata, "Failed to parse credential string for rejection. Rejecting any credential for this repo URL.");
-                        this.git.RejectCredentials(this.repoUrl);
+                        this.git.TryDeleteCredential(tracer, this.repoUrl, null, null, out string error);
                     }
 
                     this.cachedCredentialString = null;
@@ -297,7 +297,7 @@ namespace GVFS.Common.Git
         {
             string gitUsername;
             string gitPassword;
-            if (!this.git.TryGetCredentials(tracer, this.repoUrl, out gitUsername, out gitPassword, out errorMessage))
+            if (!this.git.TryGetCredential(tracer, this.repoUrl, out gitUsername, out gitPassword, out errorMessage))
             {
                 this.UpdateBackoff();
                 return false;
--- a/GVFS/GVFS.Common/Git/GitAuthentication.cs
+++ b/GVFS/GVFS.Common/Git/GitAuthentication.cs
@@ -79,7 +79,6 @@ namespace GVFS.Common.Git
                             EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
                             {
                                 ["RepoUrl"] = this.repoUrl,
-                                ["CredentialString"] = this.cachedCredentialString
                             });
                             tracer.RelatedError(metadata, "Failed to parse credential string for approval");
                         }
@@ -115,7 +114,6 @@ namespace GVFS.Common.Git
                         EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
                         {
                             ["RepoUrl"] = this.repoUrl,
-                            ["CredentialString"] = this.cachedCredentialString
                         });
                         tracer.RelatedWarning(metadata, "Failed to parse credential string for rejection. Rejecting any credential for this repo URL.");
                         this.credentialStore.TryDeleteCredential(tracer, this.repoUrl, username: null, password: null, error: out string error);
--- a/GVFS/GVFS.Common/Git/GitAuthentication.cs
+++ b/GVFS/GVFS.Common/Git/GitAuthentication.cs
@@ -14,7 +14,7 @@ namespace GVFS.Common.Git
         private const double MaxBackoffSeconds = 30;
 
         private readonly object gitAuthLock = new object();
-        private readonly GitProcess git;
+        private readonly ICredentialStore credentialStore;
         private readonly string repoUrl;
 
         private int numberOfAttempts = 0;
@@ -27,7 +27,7 @@ namespace GVFS.Common.Git
 
         public GitAuthentication(GitProcess git, string repoUrl)
         {
-            this.git = git;
+            this.credentialStore = git;
             this.repoUrl = repoUrl;
 
             if (git.TryGetConfigUrlMatch("http", this.repoUrl, out Dictionary<string, GitConfigSetting> configSettings))
@@ -66,7 +66,12 @@ namespace GVFS.Common.Git
                         string password;
                         if (TryParseCredentialString(this.cachedCredentialString, out username, out password))
                         {
-                            this.git.TryStoreCredential(tracer, this.repoUrl, username, password, out string error);
+                            if (!this.credentialStore.TryStoreCredential(tracer, this.repoUrl, username, password, out string error))
+                            {
+                                // Storing credentials is best effort attempt - log failure, but do not fail
+                                tracer.RelatedWarning("Failed to store credential string: {0}", error);
+                            }
+
                             this.isCachedCredentialStringApproved = true;
                         }
                         else
@@ -97,7 +102,11 @@ namespace GVFS.Common.Git
                     string password;
                     if (TryParseCredentialString(this.cachedCredentialString, out username, out password))
                     {
-                        this.git.TryDeleteCredential(tracer, this.repoUrl, username, password, out string error);
+                        if (!this.credentialStore.TryDeleteCredential(tracer, this.repoUrl, username, password, out string error))
+                        {
+                            // Deleting credentials is best effort attempt - log failure, but do not fail
+                            tracer.RelatedWarning("Failed to delete credential string: {0}", error);
+                        }
                     }
                     else
                     {
@@ -109,7 +118,7 @@ namespace GVFS.Common.Git
                             ["CredentialString"] = this.cachedCredentialString
                         });
                         tracer.RelatedWarning(metadata, "Failed to parse credential string for rejection. Rejecting any credential for this repo URL.");
-                        this.git.TryDeleteCredential(tracer, this.repoUrl, null, null, out string error);
+                        this.credentialStore.TryDeleteCredential(tracer, this.repoUrl, username: null, password: null, error: out string error);
                     }
 
                     this.cachedCredentialString = null;
@@ -297,7 +306,7 @@ namespace GVFS.Common.Git
         {
             string gitUsername;
             string gitPassword;
-            if (!this.git.TryGetCredential(tracer, this.repoUrl, out gitUsername, out gitPassword, out errorMessage))
+            if (!this.credentialStore.TryGetCredential(tracer, this.repoUrl, out gitUsername, out gitPassword, out errorMessage))
             {
                 this.UpdateBackoff();
                 return false;
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -155,7 +155,7 @@ namespace GVFS.Common.Git
             }
         }
 
-        public virtual void RejectCredentials(string repoUrl, string username = null, string password = null)
+        public virtual bool TryDeleteCredential(ITracer tracer, string repoUrl, string username, string password, out string errorMessage)
         {
             StringBuilder sb = new StringBuilder();
             sb.AppendFormat("url={0}\n", repoUrl);
@@ -173,13 +173,24 @@ namespace GVFS.Common.Git
 
             string stdinConfig = sb.ToString();
 
-            this.InvokeGitOutsideEnlistment(
+            Result result = this.InvokeGitOutsideEnlistment(
                 GenerateCredentialVerbCommand("reject"),
                 stdin => stdin.Write(stdinConfig),
                 null);
+
+            if (result.ExitCodeIsFailure)
+            {
+                tracer.RelatedWarning("Git could not reject credentials: {0}", result.Errors);
+
+                errorMessage = result.Errors;
+                return false;
+            }
+
+            errorMessage = null;
+            return true;
         }
 
-        public virtual void ApproveCredentials(string repoUrl, string username, string password)
+        public virtual bool TryStoreCredential(ITracer tracer, string repoUrl, string username, string password, out string errorMessage)
         {
             StringBuilder sb = new StringBuilder();
             sb.AppendFormat("url={0}\n", repoUrl);
@@ -189,10 +200,21 @@ namespace GVFS.Common.Git
 
             string stdinConfig = sb.ToString();
 
-            this.InvokeGitOutsideEnlistment(
+            Result result = this.InvokeGitOutsideEnlistment(
                 GenerateCredentialVerbCommand("approve"),
                 stdin => stdin.Write(stdinConfig),
                 null);
+
+            if (result.ExitCodeIsFailure)
+            {
+                tracer.RelatedWarning("Git could not approve credentials: {0}", result.Errors);
+
+                errorMessage = result.Errors;
+                return false;
+            }
+
+            errorMessage = null;
+            return true;
         }
 
         /// <summary>
@@ -250,7 +272,7 @@ namespace GVFS.Common.Git
             }
         }
 
-        public virtual bool TryGetCredentials(
+        public virtual bool TryGetCredential(
             ITracer tracer,
             string repoUrl,
             out string username,
@@ -261,7 +283,7 @@ namespace GVFS.Common.Git
             password = null;
             errorMessage = null;
 
-            using (ITracer activity = tracer.StartActivity("TryGetCredentials", EventLevel.Informational))
+            using (ITracer activity = tracer.StartActivity(nameof(this.TryGetCredential), EventLevel.Informational))
             {
                 Result gitCredentialOutput = this.InvokeGitAgainstDotGitFolder(
                     GenerateCredentialVerbCommand("fill"),
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -10,7 +10,7 @@ using System.Text;
 
 namespace GVFS.Common.Git
 {
-    public class GitProcess
+    public class GitProcess : ICredentialStore
     {
         private const int HResultEHANDLE = -2147024890; // 0x80070006 E_HANDLE
 
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -21,6 +21,8 @@ namespace GVFS.Common.NuGetUpgrade
         private InstallManifest installManifest;
         private IPackageSearchMetadata highestVersionAvailable;
         private NuGetFeed nuGetFeed;
+        private ICredentialStore credentialStore;
+        private bool isNuGetFeedInitialized;
 
         public NuGetUpgrader(
             string currentVersion,
@@ -30,7 +32,7 @@ namespace GVFS.Common.NuGetUpgrade
             bool noVerify,
             NuGetUpgraderConfig config,
             string downloadFolder,
-            string personalAccessToken)
+            ICredentialStore credentialStore)
             : this(
                 currentVersion,
                 tracer,
@@ -42,8 +44,9 @@ namespace GVFS.Common.NuGetUpgrade
                     config.FeedUrl,
                     config.PackageFeedName,
                     downloadFolder,
-                    personalAccessToken,
-                    tracer))
+                    null,
+                    tracer),
+                credentialStore)
         {
         }
 
@@ -54,7 +57,8 @@ namespace GVFS.Common.NuGetUpgrade
             bool noVerify,
             PhysicalFileSystem fileSystem,
             NuGetUpgraderConfig config,
-            NuGetFeed nuGetFeed)
+            NuGetFeed nuGetFeed,
+            ICredentialStore credentialStore)
             : base(
                 currentVersion,
                 tracer,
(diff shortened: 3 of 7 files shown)


GVFS/GVFS.Common/Git/CredentialManager.cs
```diff
@@ -0,0 +1,29 @@
+﻿using GVFS.Common.Tracing;
+
+namespace GVFS.Common.Git
+{
+    public class CredentialManager : ICredentialManager
+    {
+        private GitProcess gitProcess;
+
+        public CredentialManager(GitProcess gitProcess)
```
kewillford: Since all the methods of this class take an `ITracer`, I would move that to the constructor, that will most likely cause other classes like `GitAuthentication` to take it in the constructor as well.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -155,7 +155,7 @@ public bool TryKillRunningProcess(out string processName, out int exitCode, out
             }
         }
 
-        public virtual void RejectCredentials(string repoUrl, string username = null, string password = null)
+        public virtual bool TryRejectCredentials(ITracer tracer, string repoUrl, string username, string password, out string errorMessage)
```
kewillford: Most of the methods in `GitProcess` return the `Result` and let the caller decide what to do.  What are you thoughts on using that here and letting the `CredentialManager` handle the result?

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
```diff
@@ -54,20 +57,26 @@ public void SetUp()
...
+
             this.mockFileSystem = new MockFileSystem(
                 new MockDirectory(
                     Path.GetDirectoryName(this.downloadDirectoryPath),
                     new[] { new MockDirectory(this.downloadDirectoryPath, null, null) },
                     null));
 
+            this.mockCredentialManager = new Mock<ICredentialManager>();
+            string credentialManagerString = "value";
+            this.mockCredentialManager.Setup(foo => foo.TryGetCredential(It.IsAny<ITracer>(), It.IsAny<string>(), out credentialManagerString, out credentialManagerString, out credentialManagerString)).Returns(true);
```
kewillford: Seems like with this mock we could now test failure on the first `TryGetCredential` and make sure the delete and store get called and it is retried properly?

GVFS/GVFS.Common/Git/CredentialManager.cs
jamill: This is actually what I attempted initially. The problem that I ran into was that not only does `GitAuthentication` need to take an `ITracer`, but also the classes that build `GitAuthentication`, such as `Enlistment`, and `GVFSEnlistment`, all the way to `GVFSVerb` ([here](https://github.com/microsoft/VFSForGit/blob/master/GVFS/GVFS/CommandLine/GVFSVerb.cs#L844))

GVFS/GVFS.Common/Git/CredentialManager.cs
```diff
@@ -0,0 +1,29 @@
+﻿using GVFS.Common.Tracing;
+
+namespace GVFS.Common.Git
+{
+    public class CredentialManager : ICredentialManager
```
wilbaker: Minor: Unless we plan on having multiple `CredentialManager` implementations in the product, can you remove the `ICredentialManager` interface and make the methods you want to override for unit testing virtual?  That's the approach we're using elsewhere.

wilbaker: Naming suggestion: The name `CredentialManager` made me think this class interacted with GCM, when it looks like a more accurate description is that it's a collection of methods to help interact with git's credential CLI.

Would a name like `GitCredential`, `GitCredentialVerb`, or `GitCredentialCLI` be a closer match to the purpose/function of this class?  

You're more familiar with this area so feel free to take/leave this suggestion.

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
```diff
@@ -23,7 +25,7 @@ public class NuGetUpgraderTests
         private const string NewerVersion = "1.6.1185.0";
         private const string NewerVersion2 = "1.7.1185.0";
 
-        private const string NuGetFeedUrl = "feedUrlValue";
+        private const string NuGetFeedUrl = "https://pkgs.dev.azure.com/contoso/packages";
```
wilbaker: This looks like a real URL, is there a fake one we can use to ensure our tests don't accidentally hit the real one? (e.g. something like "mock://pkgs.dev.azure.com/vfs4gunittest/packages")

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -124,25 +129,9 @@ public class NuGetUpgrader : ProductUpgrader
...
-                    out string token,
-                    out string getPatError))
-            {
-                error = $"NuGetUpgrader was not able to acquire Personal Access Token to access NuGet feed. Error: {getPatError}";
-                tracer.RelatedError(error);
-                return false;
-            }
+            }
+
+            CredentialManager credentialManager = new CredentialManager(new GitProcess(gitBinPath, null, null));
```
wilbaker: This might fall under one of your TODOs, but it looks like this method is eventually hit in the `UpgradeOrchastratorTests`, and so you might want to make `credentialManager` a parameter so that the unit tests can pass in a mock `CredentialManager` (or a real `CredentialManager` created with a mock `GitProcess`).

1.  Unit test calls`this.orchestrator.Execute()`
2. `Execute` calls `UpgradeOrchestrator.TryInitialize()`
3. `TryInitialize` calls `ProductUpgrader.TryCreateUpgrader`
4. `TryCreateUpgrader` calls this method (`NuGetUpgrader.TryCreate`)

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -465,35 +460,33 @@ protected static EventMetadata CreateEventMetadata(Exception e = null)
...
+        {
+            if (!this.credentialManager.TryDeleteCredential(this.tracer, credentialUrl, null, null, out error))
             {
+                token = null;
                 return false;
             }
 
-            return true;
-        }
+            if (!this.TryGetPersonalAccessToken(credentialUrl, tracer, out token, out error))
```
wilbaker: Minor, this could be simplified to:

```
return this.TryGetPersonalAccessToken(credentialUrl, tracer, out token, out error);
```

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -178,13 +178,27 @@ public virtual void RejectCredentials(string repoUrl, string username = null, st
...
 
-            this.InvokeGitOutsideEnlistment(
+            Result result = this.InvokeGitOutsideEnlistment(
                 "credential reject",
                 stdin => stdin.Write(stdinConfig),
                 null);
+
+            if (result.ExitCodeIsFailure)
+            {
+                EventMetadata errorData = new EventMetadata();
```
wilbaker: Should we be adding anything to this `EventMetadata`?  If not, you could remove it and call the version of `RelatedWarning` that takes a `string` as its only parameter.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -194,10 +208,24 @@ public virtual void ApproveCredentials(string repoUrl, string username, string p
...
 
-            this.InvokeGitOutsideEnlistment(
+            Result result = this.InvokeGitOutsideEnlistment(
                 "credential approve",
                 stdin => stdin.Write(stdinConfig),
                 null);
+
+            if (result.ExitCodeIsFailure)
+            {
+                EventMetadata errorData = new EventMetadata();
```
wilbaker: Same comment here, this `EventMetadata` is not needed if we're not adding anything to it.

GVFS/GVFS.Common/Git/CredentialManager.cs
kewillford: Right now this class isn't really doing anything but passing data on to the `GitProcess` so is really not needed.  You could have the `GitProcess` be an `ICredentialManager` and remove this class all together.  I would only have this class if it was doing something besides being a pass through.

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -24,12 +24,15 @@ public class GitAuthentication
         private bool isCachedCredentialStringApproved = false;
 
         private bool isInitialized;
+        private ICredentialManager credentialManager;
 
         public GitAuthentication(GitProcess git, string repoUrl)
```
kewillford: At this point we can have this class take a `ICredentialManager` in the constructor as a dependency instead of the `GitProcess` class since it is no longer used.

GVFS/GVFS.Common/Git/ICredentialManager.cs
```diff
@@ -0,0 +1,13 @@
+﻿using GVFS.Common.Tracing;
+
+namespace GVFS.Common.Git
+{
+    public interface ICredentialManager
```
kewillford: To me the simpler the interface the better.  I would imagine in a perfect world it would be something more like a database CRUD:
```
public class CredentialData
{
    public string Url { get; set; }
    public string Username { get; set; }
    public string Password { get; set; }
}

public interface ICredentialManager
{
    CredentialData GetCredential();
    void StoreCredential(CredentialData data);
    void DeleteCredential(CredentialData data);
}
```

And you could use exceptions or return `bool` instead of `void` but this is just me off in a perfect world.

kewillford: After writing that it makes sense to name the interface `ICredentialStore`

jamill: I really like this API as an end goal

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
jamill: Yep! see `QueryNewestVersionReacquiresCredentialsOnAuthFailure`

GVFS/GVFS.Common/Git/CredentialManager.cs
jamill: Yeah - I was worried about that. @kewillford suggested `CredentialStore`, and I liked that.

jamill: My end goal was to move the functions out of `GitProcess`, so that `CredentialManager` / `CredentialStore` would just call into GitProcess to invoke Git, but the setup / processing of the return data would be done by `CredentialManager`. This would let us test the `CredentialManager` by controlling the `GitProcess` dependency.

For now, I could remove this intermediate class until / when / if we bring more of the implementation into this class.

GVFS/GVFS.Common/Git/GitAuthentication.cs
jamill: That would be nice - alas, we still have one other call to `GitProcess` in this class:

```
git.TryGetConfigUrlMatch("http", this.repoUrl, out Dictionary<string, GitConfigSetting> configSettings)
```

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
jamill: I like the suggestion - one wrinkle is that the code being testing is looking for a real URL (that starts with `https://`). I could tweak the URL parsing code to accept `mock://` as well, but I didn't want to make the product code change just to support using `mock://` URLs in test code. Maybe there is another way to handle this... Thoughts?

GVFS/GVFS.Common/Git/GitProcess.cs
jamill: I could do that - would that be enough (along with `CredentialManager` keeping an `ITracer` member variable) to justify a `CredentialManager` class?

I liked the API you proposed for `ICredentialManager`, and while what we have now with `Try*` methods isn't quite there, I think it is closer to the ideal goal than returning `GitProcess.Result` (i.e. - If we are going to remove the `CredentialManager` class, I would rather keep this API than transforming it to return a `GitProcess.Result`).

GVFS/GVFS.Common/Git/CredentialManager.cs
mjcheetham: We have only one implementation of `ITracer` which is `JsonTracer`?
I'm all for the interface in my vote.

GVFS/GVFS.Common/Git/GitAuthentication.cs
kewillford: Maybe that call is the responsibility of the creator of this class so that this one becomes:
`public GitAuthentication(ICredentialStore store, Dictionary<string, GitConfigSetting> configSettings)`
or
`public GitAuthentication(ICredentialStore store, GitSsl gitSsl)`

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
```diff
@@ -54,20 +66,27 @@ public void SetUp()
...
                 null,
                 this.tracer);
+
             this.mockFileSystem = new MockFileSystem(
                 new MockDirectory(
                     Path.GetDirectoryName(this.downloadDirectoryPath),
                     new[] { new MockDirectory(this.downloadDirectoryPath, null, null) },
                     null));
 
+            this.mockCredentialManager = new Mock<ICredentialStore>();
```
kewillford: What are our standards with using Mock?  I have always used [`MockBehavior.Strict`](https://github.com/moq/moq4/wiki/Quickstart) so that you know exactly what the expected calls for your test are.

GVFS/GVFS.Common/Git/CredentialManager.cs
wilbaker: > We have only one implementation of ITracer which is JsonTracer?
> I'm all for the interface in my vote.

You're right about `ITracer/JsonTracer`.  It predates the guideline to avoid interfaces whose only purpose is unit testing and no one went back to make that class/interface consistent.  

The other interfaces in the code are either for cross-platform support, or are used when we need to expose a small surface area of an underlying class (e.g. `IProfilerOnlyIndexProjection`, `IHeartBeatMetadataProvider`).  

I think that @kewillford's suggestion for `ICredentialManager` is a better fit for the style used elsewhere in VFS4G:  

> You could have the GitProcess be an ICredentialManager and remove this class all together. I would only have this class if it was doing something besides being a pass through.

The [current guideline](https://github.com/Microsoft/VFSForGit/blob/master/CONTRIBUTING.md) is to avoid adding interfaces whose only purpose is unit testing.  If there's agreement the guideline should change then there should be a separate PR to update the guideline (and source code to match the new guideline).

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
wilbaker: > one wrinkle is that the code being testing is looking for a real URL (that starts with https://).

Ahh, good point.

> Maybe there is another way to handle this... Thoughts?

We could include characters that are invalid (e.g. it sounds like [TAB is invalid](https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid)), but I'm not sure if that would actually stop clients from sending a request to `https://pkgs.dev.azure.com`.

Maybe there's not much more we can do here?  (But if contoso is a real package feed, perhaps we should switch to one that belongs to VFS4G instead?)

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
wilbaker: > What are our standards with using Mock? I have always used MockBehavior.Strict so that you know exactly what the expected calls for your test are.

I don't think we currently have any, but after reading about `MockBehavior.Strict` it sounds like a good guideline.

Unless someone has objections I think it makes sense to update our current `Mock` code to use it (if we're not already) and adding it to the coding guidelines (either in this PR or a separate PR).

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
kewillford: In looking at the code seems like we need a `NuGetFeedUrlToAzDevOrgUrlConverter` that has an interface so that we can have other implementations like a `NuGetFeedUrlToGitHubUrlConverter`, etc?  Then we could have one to pass in for tests that verifies `mock://` instead of `https://`?  No idea how much work or how disruptive this would be though.

GVFS/GVFS.Common/Git/CredentialManager.cs
jamill: This change has been scoped down a bit - instead of a new `CredentialStore` class, the functionality will be implemented in `GitProcess`

GVFS/GVFS.Common/Git/CredentialManager.cs
jamill: > I think that @kewillford's suggestion for ICredentialManager is a better fit for the style used elsewhere in VFS4G

That is the approach taken in now.

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
jamill: I like this idea, but was going to defer actually making it

GVFS/GVFS.Common/Git/CredentialManager.cs
jamill: > Right now this class isn't really doing anything but passing data on to the GitProcess so is really not needed

I removed this class and left the implementation in `GitProcess`

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/NuGetUpgraderTests.cs
jamill: > either in this PR or a separate PR)

Let's do this in a separate PR.

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -66,15 +66,20 @@ public void ApproveCredentials(ITracer tracer, string credentialString)
...
                         {
-                            EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
+                            EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
                             {
-                                ["RepoUrl"] = this.repoUrl,
-                                ["CredentialString"] = this.cachedCredentialString
+                                ["RepoUrl"] = this.repoUrl,
+                                ["CredentialString"] = this.cachedCredentialString
                             });
                             tracer.RelatedError(metadata, "Failed to parse credential string for approval");
```
alameenshah: The variable name `CredentialString` suggests it is holding sensitive information. I know we were doing this before, is it encrypted? & okay to send it to LiveSite?

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -97,19 +102,23 @@ public void RejectCredentials(ITracer tracer, string credentialString)
...
-                        });
-                        tracer.RelatedWarning(metadata, "Failed to parse credential string for rejection. Rejecting any credential for this repo URL.");
-                        this.git.RejectCredentials(this.repoUrl);
+                        EventMetadata metadata = new EventMetadata(new Dictionary<string, object>
+                        {
+                            ["RepoUrl"] = this.repoUrl,
+                            ["CredentialString"] = this.cachedCredentialString
+                        });
+                        tracer.RelatedWarning(metadata, "Failed to parse credential string for rejection. Rejecting any credential for this repo URL.");
+                        this.credentialStore.TryDeleteCredential(tracer, this.repoUrl, null, null, out string error);
```
alameenshah: nit: can we add the label for `null` parameters.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -126,23 +131,7 @@ public class NuGetUpgrader : ProductUpgrader
...
-                    authUrl,
-                    tracer,
-                    out string token,
-                    out string getPatError))
-            {
-                error = $"NuGetUpgrader was not able to acquire Personal Access Token to access NuGet feed. Error: {getPatError}";
-                tracer.RelatedError(error);
-                return false;
-            }
+            ICredentialStore credentialStore = new GitProcess(gitBinPath, null, null);
```
alameenshah: nit: labels for null params

GVFS/GVFS.Common/Git/GitAuthentication.cs
jamill: Will remove

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -20,7 +20,9 @@ public class NuGetUpgrader : ProductUpgrader
 
         private InstallManifest installManifest;
         private IPackageSearchMetadata highestVersionAvailable;
-        private NuGetFeed nuGetFeed;
+        private NuGetFeed nuGetFeed;
+        private ICredentialStore credentialManager;
```
mjcheetham: Update the name of the field to match the new interface name `credentialStore`

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -467,33 +462,26 @@ protected static EventMetadata CreateEventMetadata(Exception e = null)
             return metadata;
         }
 
-        private static bool TryGetPersonalAccessToken(string gitBinaryPath, string credentialUrl, ITracer tracer, out string token, out string error)
+        private static string replacementToken(string tokenString)
```
mjcheetham: ??? Lower case function name? this is also a pretty small method. Why not inline?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -605,6 +591,29 @@ private bool TryRefreshCredentials(out string error)
                 this.TraceException(ex, nameof(this.TryRefreshCredentials), "Failed to refresh credentials.");
                 return false;
             }
+        }
+
+        private bool EnsureNuGetFeedInitialized(out string error)
+        {
+            if (!this.isNuGetFeedInitialized)
```
mjcheetham: Should this not be an error case when initialising twice?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: Good catch - this was a lower case string already, and is just showing up in the diff due to things moving around... I will update the casing, but in a different change.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: We are making this call each time we attempt to query the remote endpoint, so it is just to ensure that we are initialized already. It is an internal method, and the parent method could potentially be called multiple times...

PR #952, opened 2019-03-21
Test HandleVnodeOperation

jeschu1:
This incorporates
- Updates by @pmj to allow parameter testing
- Tests for HandleVnodeOperation


ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
```diff
@@ -0,0 +1,514 @@
...
+    repoRootVnode.reset();
+    testFileVnode.reset();
+    testDirVnode.reset();
+    VirtualizationRoots_Cleanup();
+    vfs_context_rele(context);
+    MockVnodes_CheckAndClear();
+    MockCalls::Clear();
+}
+
+- (void) testEmptyFileHydrates {
```
jeschu1: @wilbaker can you please do a quick scan of this file and let me know if you think we're missing any good cases.  I tried to test many relevant permutations.

ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
```diff
@@ -0,0 +1,514 @@
...
+            MockCalls::DidCallFunction(
+                ProviderMessaging_TrySendRequestAndWaitForResponse,
+                _,
+                MessageType_KtoU_HydrateFile,
+                testFileVnode.get(),
+                _,
+                _,
+                _,
+                _,
+                _,
```
jeschu1: @pmj I asked this on IM but if I stick 
reinterpret_cast<int*>(KAUTH_RESULT_DEFER),
in this parameter I get a failure.  I'm guessing ref compare instead of value compare.
How do you think we should handle this?

ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
```diff
@@ -0,0 +1,514 @@
...
+        HandleVnodeOperation(
+            nullptr,
+            nullptr,
+            actions[i],
+            reinterpret_cast<uintptr_t>(context),
+            reinterpret_cast<uintptr_t>(testFileVnode.get()),
+            0,
+            0);
+        XCTAssertFalse(
+            MockCalls::DidCallFunction(
```
jeschu1: @pmj I'm thinking a call count would be handy here to verify.

MockCalls::CallFunctionCount(ProviderMessaging_TrySendRequestAndWaitForResponse)

ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
pmj: Are you trying to check the input value or simulate setting the output parameter?

Neither is currently supported, and I'm not sure what syntax would really work for this type of thing. The best I can think of is to put the check or output parameter setting straight in the mock function itself.

ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
pmj: You can skip the arguments altogether with `DidCallFunction`, but I can definitely add a call count check as well.

ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
jeschu1: @pmj that makes sense.  I'm not sure we need the to test for that value explicitly since we validated the message was sent.  What do you think?
We can certainly test it in the mock if absolutely necessary.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,553 @@
...
+- (void) testOpen {
+    // A file that is not yet in the virtual root, should recieve a creation request
+    char pathBuffer[PrjFSMaxPath] = "";
+    int pathLength = sizeof(pathBuffer);
+    vn_getpath(testFileVnode.get(), pathBuffer, &pathLength);
+
+    HandleFileOpOperation(
+        nullptr,
+        nullptr,
+        1,
```
jeschu1: Instead of 1 I want t use KAUTH_FILEOP_OPEN
I see this in Kernel/sys/kauth.h

@pmj whats the best way to import it into the unit tests here.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+    {
+        HandleVnodeOperation(
+            nullptr,
+            nullptr,
+            actions[i],
+            reinterpret_cast<uintptr_t>(context),
+            reinterpret_cast<uintptr_t>(testFileVnode.get()),
+            0,
+            0);
+        XCTAssertTrue(
```
wilbaker: Also validate that `KAUTH_VNODE_` actions that should not hydrate do not trigger `MessageType_KtoU_HydrateFile`? (e.g. `KAUTH_VNODE_READ_SECURITY`)

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+            MessageType_KtoU_NotifyFilePreDelete,
+            testFileVnode.get(),
+            _,
+            _,
+            _,
+            _,
+            _,
+            nullptr));
+    XCTAssertTrue(
+       MockCalls::DidCallFunction(
```
wilbaker: Possible enhancement here (and same idea below in `testDeleteDir`):

The `FunctionCall` struct stored in `RecordedCallMap` includes `callSequenceNumber`.  

If we extracted the `callSequenceNumber`we could validate that `MessageType_KtoU_NotifyFilePreDelete` was called before `MessageType_KtoU_HydrateFile`.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+- (void) testEmptyDirectoryEnumerates {
+    testDirVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
+    const int actionCount = 5;
+    kauth_action_t actions[actionCount] =
+    {
+        KAUTH_VNODE_LIST_DIRECTORY,
+        KAUTH_VNODE_SEARCH,
+        KAUTH_VNODE_READ_SECURITY,
+        KAUTH_VNODE_READ_ATTRIBUTES,
+        KAUTH_VNODE_READ_EXTATTRIBUTES
```
wilbaker: Similar to comment in `testEmptyFileHydrates`, add testing for other `KAUTH_VNODE_` that we don't expect to trigger enumeration callbacks?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+        vnodePath,
+        pid,
+        procname,
+        kauthResult,
+        kauthError);
+    
+    return true;
+}
+
+static void SetFileXattrData(shared_ptr<vnode> vnode)
```
wilbaker: Naming suggestion: `SetFileXattrData` -> `SetPrjFSFileXattrData`

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+    testFileVnode.reset();
+    testDirVnode.reset();
+    VirtualizationRoots_Cleanup();
+    vfs_context_rele(context);
+    MockVnodes_CheckAndClear();
+    MockCalls::Clear();
+}
+
+- (void) testEmptyFileHydrates {
+    testFileVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
```
wilbaker: Also call `SetFileXattrData`? It would be set on an empty placeholder file.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+                _,
+                _,
+                _,
+                _,
+                nullptr));
+        MockCalls::Clear();
+    }
+}
+
+- (void) testNonEmptyFileDoesNotHydrate {
```
wilbaker: Also test after calling `SetFileXattrData`.  That will allow for covering hydrated placeholder and full file scenarios.

(I am guessing the kext doesn't care, but it would be nice to have the coverage so that if that assumption ever changes in the future we can catch it)

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+            _,
+            MessageType_KtoU_HydrateFile,
+            testFileVnode.get(),
+            _,
+            _,
+            _,
+            _,
+            _,
+            nullptr));
+    XCTAssertTrue(
```
wilbaker: Similar comment as above, can we check `callSequenceNumber`?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+- (void) testEventsAreIgnored {
+    testDirVnode->attrValues.va_flags = FileFlags_IsInVirtualizationRoot | FileFlags_IsEmpty;
+    const int actionCount = 5;
+    kauth_action_t actions[actionCount] =
+    {
+        KAUTH_VNODE_APPEND_DATA,
+        KAUTH_VNODE_ADD_SUBDIRECTORY,
+        KAUTH_VNODE_WRITE_SECURITY,
+        KAUTH_VNODE_TAKE_OWNERSHIP,
+        KAUTH_VNODE_ACCESS
```
wilbaker: Can we give `KAUTH_VNODE_ACCESS` its own test?

The kext should be ignoring all events with `KAUTH_VNODE_ACCESS` set, even those it would otherwise care about (`KAUTH_VNODE_ACCESS` is a flag that is set with other access bits (e.g. `KAUTH_VNODE_DELETE` when `HandleVnodeOperation` is called).

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+            actions[i],
+            reinterpret_cast<uintptr_t>(context),
+            reinterpret_cast<uintptr_t>(testDirVnode.get()),
+            0,
+            0);
+        XCTAssertFalse(MockCalls::DidCallFunction(ProviderMessaging_TrySendRequestAndWaitForResponse));
+    }
+}
+    
+-(void) testWriteFile {
```
wilbaker: Minor naming suggestion:

If the some of the more generic test names were slightly more detailed it might make it easier to find test(s) of interest when looking for specific coverage:

_Examples_

- `testWriteFile ` -> `testWriteDataVnodeOpEmptyFile` 
- `testWriteFileFull` -> `testWriteDataVnodeOpFullFile`

What got me thinking about this was when I read the implementation of `testWriteFileFull` I was surprised to see that it wasn't checking that the `MessageType_KtoU_NotifyFileModified` message was sent.  After realizing that these first tests were only for vnode operations it made sense, but it got me wondering if the test names should include VnodeOp or FileOp (or perhaps the tests should be in different classes?).

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+            _,
+            MessageType_KtoU_NotifyFileCreated,
+            testFileVnode.get(),
+            _,
+            _,
+            _,
+            _,
+            _,
+            _));
+}
```
wilbaker: Could we check `nextCallSequenceNumber` to ensure that the above callback was the only callback called (and that it was only called once)?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+        nullptr,
+        nullptr,
+        KAUTH_FILEOP_CLOSE,
+        reinterpret_cast<uintptr_t>(testFileVnode.get()),
+        reinterpret_cast<uintptr_t>(filePath),
+        0,
+        0);
+    
+    XCTAssertFalse(MockCalls::DidCallFunction(ProviderMessaging_TrySendRequestAndWaitForResponse));
+}
```
wilbaker: Can you add tests for `KAUTH_FILEOP_RENAME` and `KAUTH_FILEOP_LINK`? 

(It's also OK with me if you're saving more `HandleFileOpOperation` tests for a future PR).

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+            _,
+            _,
+            _,
+            _,
+            _,
+            _));
+}
+
+- (void) testOpenInVirtualizationRoot {
+    // A file that is already in the virtual root, should not recieve a creation request
```
wilbaker: This comment doesn't quite match the code (now that we have the more generic `XCTAssertFalse(MockCalls::DidCallFunction(ProviderMessaging_TrySendRequestAndWaitForResponse))` check), maybe update it to something like:

```
// KAUTH_FILEOP_OPEN  should not trigger any callbacks for files that are already flagged as in the virtualization root.
```

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -3,6 +3,14 @@
 #include <unordered_map>
 #include <unordered_set>
 
+namespace KextMock
+{
+    struct PlaceholderValue
+    {};
+    
+    extern PlaceholderValue _;
```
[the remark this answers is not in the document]
pmj: Sure, that commit was thrown together as quickly as possible so that @jeschu1 had something to work with, I think there are a few things there that could be improved. I'll try and get that done soon.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -218,6 +219,49 @@ - (void)testRegisterProviderForPath_InsertionSucceeded
...
+    XCTAssertTrue(VirtualizationRoot_IsValidRootHandle(result.root));
+
+    PrjFSProviderUserClient dummyClient2;
+    VirtualizationRootResult result2 = VirtualizationRoot_RegisterProviderForPath(&dummyClient2, 1000, path2);
+    XCTAssertEqual(result2.error, 0);
+    XCTAssertTrue(VirtualizationRoot_IsValidRootHandle(result2.root));
+
+    XCTAssertNotEqual(result.root, result2.root);
+
+    if (VirtualizationRoot_IsValidRootHandle(result.root))
```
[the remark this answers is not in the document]
pmj: > Out of curiosity, what would happen if we just let the code inside the `if(...)` run?

`s_virtualizationRoots[result.root].providerUserClient` on the line below would use a negative index into a 0-based array, which we *really* don't want to do even if the test is failing.

`XCTAssert` macros do not exit the test when they fail. And yes, this is a bit weird compared to languages where exceptions are safe to use. (Objective-C exceptions are de facto deprecated because they're almost impossible to get right.)

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
```diff
@@ -218,6 +219,49 @@ - (void)testRegisterProviderForPath_InsertionSucceeded
...
+
+    if (VirtualizationRoot_IsValidRootHandle(result2.root))
+    {
+        XCTAssertEqual(s_virtualizationRoots[result2.root].providerUserClient, &dummyClient2);
+
+        XCTAssertTrue(MockCalls::DidCallFunction(vfs_setauthcache_ttl, secondMountPoint.get(), 0));
+        XCTAssertTrue(MockCalls::DidCallFunction(ProviderUserClient_UpdatePathProperty, &dummyClient2, _));
+    
+        s_virtualizationRoots[result2.root].providerUserClient = nullptr;
+        vnode_put(s_virtualizationRoots[result2.root].rootVNode);
```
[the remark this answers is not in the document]
pmj: Personally, I was hoping to replace these with calls to `ActiveProvider_Disconnect()` but last I checked we couldn't call that function from tests yet, and I haven't got around to sorting that out yet.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: This is on my list of features to add. I'll try to get that in for this PR…

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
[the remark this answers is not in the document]
jeschu1: If you get pressed for time @pmj I can do it

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: It seems that `#include <Kernel/sys/kauth.h>` works with a bit of trickery. I've pushed a commit [to this branch](https://github.com/pmj/VFSForGit/tree/more_testing_handlevnodeop) that does that and avoids the need for our own `#define`s.

jeschu1: @pmj Nice Thank You!

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -0,0 +1,569 @@
...
+- (void) testCloseWithModifed {
+    testFileVnode->attrValues.va_flags = FileFlags_IsInVirtualizationRoot;
+
+    HandleFileOpOperation(
+        nullptr,
+        nullptr,
+        KAUTH_FILEOP_CLOSE,
+        reinterpret_cast<uintptr_t>(testFileVnode.get()),
+        reinterpret_cast<uintptr_t>(filePath),
+        KAUTH_FILEOP_CLOSE_MODIFIED,
```
pmj: I just noticed in the code this is testing that `KAUTH_FILEOP_CLOSE_MODIFIED` is in theory a bit field, but `HandleFileOpOperation()` tests for equality. There currently aren't any other possible flags, but if Apple were to add some in future, our code could break. So (1) we should replace `if (KAUTH_FILEOP_CLOSE_MODIFIED != closeFlags)` with `if ((KAUTH_FILEOP_CLOSE_MODIFIED & closeFlags) == 0)` (or helper function call to that effect) and (2) we should add a test case that deliberately invents an extra bit flag that we or (`|`) with `KAUTH_FILEOP_CLOSE_MODIFIED` and make sure the test still passes. (Which it currently won't, but should after the fix.)

ProjFS.Mac/PrjFSKextTests/MockProc.hpp
```diff
@@ -14,6 +14,14 @@ extern "C"
     int proc_pid(proc_t);
     void proc_name(int pid, char* buf, int size);
     proc_t vfs_context_proc(vfs_context_t ctx);
+    proc_t proc_self(void);
+    kauth_cred_t kauth_cred_proc_ref(proc_t procp);
+    uid_t kauth_cred_getuid(kauth_cred_t _cred);
+    void kauth_cred_unref(kauth_cred_t *_cred);
```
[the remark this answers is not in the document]
pmj: The kauth credential and proc functions are, to be fair, quite strongly related, so putting them in the same file does make sense. For example, I can imagine creating mock `proc` objects which contain a mock `kauth_cred_t` which we then pass around, only for the kext code to extract the corresponding `proc` back out. Splitting the mock code for this across 2 files seems unnecessarily awkward.

But if we're keeping them together, then I agree the file name should probably change so it's obvious that's what we've done.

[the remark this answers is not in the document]
wilbaker: Cool thanks for the explanation.  I think even updating the comment might be enough.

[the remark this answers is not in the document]
jeschu1: I'll update the comment for now.  

When we add unit tests for CurrentProcessWasSpawnedByRegularUser (which manipulates these functions more) we can get fancier.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -22,3 +22,43 @@ proc_t vfs_context_proc(vfs_context_t ctx)
...
+}
+
+proc_t proc_find(int pid)
+{
+    return NULL;
+}
+
+int proc_rele(proc_t p)
+{
+   return 1;
```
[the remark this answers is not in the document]
jeschu1: Any number will actually due since we don't test the result in code.
The results here will seem a bit odd, but hopefully make a lot more sense when we unit tests CurrentProcessWasSpawnedByRegularUser.

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
```diff
@@ -124,6 +124,7 @@ void ProviderMessaging_AbortOutstandingEventsForProvider(VirtualizationRootHandl
     Mutex_Release(s_outstandingMessagesMutex);
 }
 
+#ifndef KEXT_UNIT_TESTING
```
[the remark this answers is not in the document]
jeschu1: Thats a better option, updating.
Previously since the messaging was in KAuthHandler we couldn't pull it apart.  Another advanage to the refactoring @pmj did here.

ProjFS.Mac/PrjFSKextTests/VirtualizationRootsTests.mm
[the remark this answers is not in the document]
jeschu1: I would save that change for that rework.  The new test follows the existing convention and isn't the focus of this PR.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -60,20 +73,67 @@ class MockCalls
     static MockCalls singleton;
     
 public:
-    template <typename R, typename... ARGS>
-        static void RecordFunctionCall(R (*fn)(ARGS...), ARGS... args)
+    template <typename R, typename... ARGS, typename... ACTUALARGS>
+        static void RecordFunctionCall(R (*fn)(ARGS...), ACTUALARGS... args)
```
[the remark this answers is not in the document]
jeschu1: I like the new name suggestion so i'll update

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -60,20 +73,67 @@ class MockCalls
...
+}
+
+template <typename ARG_T>
+bool CheckArgument(ARG_T& argument, KextMock::PlaceholderValue)
+{
+    return true;
+}
+
+template <typename TUPLE_T, typename CHECK_TUPLE_T, size_t... INDICES>
+bool CheckArguments(TUPLE_T& arguments, CHECK_TUPLE_T check, std::index_sequence<INDICES...>)
```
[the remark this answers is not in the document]
jeschu1: Updated to ArgumentsAreEqual

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: Would it make sense to reuse ActionBitIsSet for this check?

I think this is a good suggestion for another PR.  I entered an issue to track https://github.com/Microsoft/VFSForGit/issues/971

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: This exactly!  I'm hoping to get this PR in then
1. Add Code Coverage to the Build
2. Finish off any needed coverage

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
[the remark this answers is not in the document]
jeschu1: Comment added :-)

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: This is a good point.  I haven't fully decided on seperate files or the same files for both types.  My plan is to revisit decision when I go full force on FileOp tests (seperate PR).  Hence for now I'd like to punt the naming to a future (but not distance future PR).

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -1,3 +1,4 @@
+#include "../PrjFSKext/kernel-header-wrappers/kauth.h"
```
wilbaker: Out of curiosity, why did this header need to get added if nothing else in `KauthHandlerTests.mm` changed?

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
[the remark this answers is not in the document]
wilbaker: Apologies for bringing this up again, but could you add a few more details (and possibly rename `PlaceholderValue`)?  Someone new to the code might think that `PlaceholderValue` is related to file placeholders based on its name.

Maybe something like:

```
namespace KextMock
{
    struct ParameterPlaceholderValue
    {};

    // Matches with any parameter value
    // ParameterPlaceholderValue can be used:
    //   - When calling kext functions, for parameters values that do not matter to the test
    //   - When checking if a function was called (i.e. DidCallFunction).  ParameterPlaceholderValue is 
    //     a match for all parameter values.
    extern ParameterPlaceholderValue _;
```

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: I see this comment was marked as resolved, but it looks like the comment was not updated.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: If the `callSequenceNumber` work is planned for a follow up PR, could you file a work item for it?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: Cause I put the comment on the wrong place.  I'm updated with 2 superior comments.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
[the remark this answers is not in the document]
jeschu1: Updating with suggested comment

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: I've added this functionality will update with a commit shortly

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: I added a sequence check.  It does start at 2.  Previous to this call other mocks are called which up the counter.
We may change our minds on how we want this.  
If we did it function based then we would lose the overall ordering of calls which might be unideal.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: Nice Catch, the change was unnecessary here!

ProjFS.Mac/PrjFSKextTests/HandleVnodeOperationTests.mm
jeschu1: I added a call count, so need to worry about it @pmj

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -8,7 +8,11 @@ namespace KextMock
     struct PlaceholderValue
     {};
     
-    // Allow any value to match
+    // Matches with any parameter value
+    // ParameterPlaceholderValue can be used:
```
wilbaker: Unless you're renaming the class to `ParameterPlaceholderValue` all of the `ParameterPlaceholderValue` in the comment should be replaced with `PlaceholderValue`.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: Instead of validating the specific sequence number, what would you think of an approach like this:

```
int GetFunctionCallSequenceNum(FunctionPointerType function, CHECK_ARGS... checkArgs);
```

And we could have a special value (e.g. -1) defined to be NOT_CALLED.

Then the test could have something like:

```
int AaCallSequence = GetFunctionCallSequenceNum(A, a); // Get sequence number of call to A(a)
int AbCallSequence = GetFunctionCallSequenceNum(A, b); // Get sequence number of call to A(b)
assert (AaCallSequence != NOT_CALLED);
assert (AbCallSequence != NOT_CALLED);
assert (AaCallSequence < AbCallSequence);
```

That way we don't break all the tests if we add another function call in the setup (or somewhere else) and we're still validating the ordering of the functions we care about.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -120,21 +139,40 @@ bool ArgumentsAreEqual(TUPLE_T& arguments, CHECK_TUPLE_T check, std::index_seque
...
+{
+    return DidCallFunction(function, -1, checkArgs...);
+}
+
+template <typename R, typename... ARGS>
+int SpecificFunctionCallRecorder<R, ARGS...>::CallCount(FunctionPointerType function)
+{
+    int callCount = 0;
+    std::pair<typename RecordedCallMap::const_iterator, typename RecordedCallMap::const_iterator> foundCalls = this->recordedCalls.equal_range(function);
+    for (typename RecordedCallMap::const_iterator foundCall = foundCalls.first; foundCall != foundCalls.second; ++foundCall)
```
wilbaker: Rather than writing the loop yourself you can use [std::distance](http://www.cplusplus.com/reference/iterator/distance/) to get the count:

```
return distance(foundCalls.first, foundCalls.second);
```

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: \+ @pmj for his thoughts on this as well

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: I'm pretty sure we needed it because the enum declaration from `MockProc.hpp` for `KAUTH_RESULT_ALLOW` and such is no longer around. I'm not sure `#include "../PrjFSKext/kernel-header-wrappers/kauth.h"` is appropriate inside `MockProc.hpp` as nothing there requires any of the stuff in that header.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: I think/agree that checking specific sequence number values is much too fragile. I also don't think changing the semantics of the existing `DidCallFunction()` itself is helpful; a lot of the time we don't care about order, so that extra argument will just get in the way in those cases.

My original plan for this feature was something along the following lines:

    XCTAssertTrue(
       MockCalls::DidCallFunctionsInOrder(
            ProviderMessaging_TrySendRequestAndWaitForResponse,
            make_tuple(
                _,
                MessageType_KtoU_NotifyDirectoryPreDelete,
                testDirVnode.get(),
                _,
                _,
                _,
                _,
                _,
                nullptr),
            ProviderMessaging_TrySendRequestAndWaitForResponse,
            make_tuple(
                _,
                MessageType_KtoU_RecursivelyEnumerateDirectory,
                testDirVnode.get(),
                _,
                _,
                _,
                _,
                _,
                nullptr)));

This would assert relative order, but not require *all* called functions to be listed.

jeschu1: Ohhh I really like that @pmj.  This was along the lines of a similar solution @wilbaker pinged me.

ProjFS.Mac/PrjFSKextTests/KextMockUtilities.hpp
```diff
@@ -120,21 +139,40 @@ bool ArgumentsAreEqual(TUPLE_T& arguments, CHECK_TUPLE_T check, std::index_seque
...
+    bool SpecificFunctionCallRecorder<R, ARGS...>::DidCallFunction(FunctionPointerType function, CHECK_ARGS... checkArgs)
+{
+    return DidCallFunction(function, -1, checkArgs...);
+}
+
+template <typename R, typename... ARGS>
+int SpecificFunctionCallRecorder<R, ARGS...>::CallCount(FunctionPointerType function)
+{
+    int callCount = 0;
+    std::pair<typename RecordedCallMap::const_iterator, typename RecordedCallMap::const_iterator> foundCalls = this->recordedCalls.equal_range(function);
```
pmj: I know some of my code has declarations like this as well, but I'm thinking it's less readable than `auto` at this point. We don't have to go for `auto`, there are a few in-between options which I think would also be preferable over the status quo:

 * I think the latest C++ standard allows partial type declarations, so in this case, I believe `std::pair foundCalls = …;` should work. This at least makes it obvious we're getting a `pair` - and pretty much all operations on STL containers return iterators in some form, so not spelling those out in detail seems sensible.
 * Locally Typedef'ing the iterator type to something much shorter.

wilbaker: > I think the latest C++ standard allows partial type declarations, so in this case, I believe std::pair foundCalls = …; should work. This at least makes it obvious we're getting a pair - and pretty much all operations on STL containers return iterators in some form, so not spelling those out in detail seems sensible.

Cool I did not know about this feature 👍 

> I know some of my code has declarations like this as well, but I'm thinking it's less readable than auto at this point. 

I agree on this point, and template programming is one spot where things can get messy without `auto`.  

If the above approach does not work out I'm good with using `auto` here with some descriptive variable names.

PR #954, opened 2019-03-22
Add timings to different parts of updating placeholders

kewillford:


GVFS/GVFS.Common/GVFSLock.cs
```diff
@@ -284,9 +288,17 @@ public void RecordReleaseExternalLockRequested()
                 this.lockHeldExternallyTimeMs = this.lockAcquiredTime.ElapsedMilliseconds;
             }
 
-            public void RecordUpdatePlaceholders(long durationMs)
+            public void RecordUpdatePlaceholders(long durationMs, long updateMs, long deleteEmptyMs, long writeAndFlushMs)
             {
-                this.placeholderUpdateTimeMs = durationMs;
+                this.placeholderTotalUpdateTimeMs = durationMs;
+                this.placeholderUpdateTimeMs = updateMs;
```
jrbriggs: We're changing the definition of this telemetry point. Do you know if we have any reporting looking at this right now?

kewillford: None that I know of.  The reports I know of were looking at modified paths count and William just asked about the placeholder count which noone said they were using.  I will send a email to verify.

GVFS/GVFS.Common/GVFSLock.cs
```diff
@@ -319,7 +331,11 @@ public void AddStatsToTelemetry(EventMetadata metadata)
                 metadata.Add("DurationMS", this.lockAcquiredTime.ElapsedMilliseconds);
                 metadata.Add("LockHeldExternallyMS", this.lockHeldExternallyTimeMs);
                 metadata.Add("ParseGitIndexMS", this.parseGitIndexTimeMs);
+                metadata.Add("UpdatePlaceholdersTotalMS", this.placeholderTotalUpdateTimeMs);
                 metadata.Add("UpdatePlaceholdersMS", this.placeholderUpdateTimeMs);
+                metadata.Add("DeleteEmptyFolderPlaceholdersMS", this.placeholderEmptyFolderDeleteTimeMs);
```
wilbaker: On Mac this time will also include re-expanding folders.  Should the name reflect that?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1167,6 +1176,10 @@ private void UpdatePlaceholders()
                     }
                 }
 
+                stopwatch.Stop();
+                long millisecondsRemovingEmptyFolders = stopwatch.ElapsedMilliseconds;
```
wilbaker: Should we also include a count of how many folders were successfully deleted? (I'm not sure if this should go in `GVFSLock.Stats` though or be recorded to telemetry somewhere else.

GVFS/GVFS.Common/GVFSLock.cs
```diff
@@ -319,7 +331,11 @@ public void AddStatsToTelemetry(EventMetadata metadata)
                 metadata.Add("DurationMS", this.lockAcquiredTime.ElapsedMilliseconds);
                 metadata.Add("LockHeldExternallyMS", this.lockHeldExternallyTimeMs);
                 metadata.Add("ParseGitIndexMS", this.parseGitIndexTimeMs);
+                metadata.Add("UpdatePlaceholdersTotalMS", this.placeholderTotalUpdateTimeMs);
                 metadata.Add("UpdatePlaceholdersMS", this.placeholderUpdateTimeMs);
```
wilbaker: Should `"UpdatePlaceholdersMS"` be changed to something like `"UpdateExistingPlaceholdersMS"` to avoid any possible confusion with its old meaning (of total time)?

GVFS/GVFS.Common/GVFSLock.cs
kewillford: I'm hesitant to name it based on something that only happens on one platform.  Maybe it should be named `CleanupFolderPlaceholdersMS` to give it a general name that would indicate multiple things could have happened?

wilbaker: I think that's a good idea.  

Even sticking with the "Update" theme (e.g. `UpdateFolderPlaceholdersMS`) might be accurate enough because this is the code that actually processes them (as far as I'm aware).

(e.g. `UpdatePlaceholdersMS` includes deleting file placeholders)

Any general name works for me.

PR #961, opened 2019-03-25
Pass session id through to mount process for correlation with git trace2

kewillford:
git trace2 creates a session id for the git command and by passing this
through to the mount process, logging will be able to have this correlation
id to be able to join the git trace2 data with any logged data from the mount
process.

### DIFF
--- a/GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
+++ b/GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
@@ -235,9 +235,9 @@ namespace GVFS.Common.NamedPipes
                 this.RequestData = LockData.FromBody(messageBody);
             }
 
-            public LockRequest(int pid, bool isElevated, bool checkAvailabilityOnly, string parsedCommand)
+            public LockRequest(int pid, bool isElevated, bool checkAvailabilityOnly, string parsedCommand, string gitCommandSessionId)
             {
-                this.RequestData = new LockData(pid, isElevated, checkAvailabilityOnly, parsedCommand);
+                this.RequestData = new LockData(pid, isElevated, checkAvailabilityOnly, parsedCommand, gitCommandSessionId);
             }
 
             public LockData RequestData { get; }
@@ -250,15 +250,18 @@ namespace GVFS.Common.NamedPipes
 
         public class LockData
         {
-            public LockData(int pid, bool isElevated, bool checkAvailabilityOnly, string parsedCommand)
+            public LockData(int pid, bool isElevated, bool checkAvailabilityOnly, string parsedCommand, string gitCommandSessionId)
             {
                 this.PID = pid;
+                this.GitCommandSessionId = gitCommandSessionId;
                 this.IsElevated = isElevated;
                 this.CheckAvailabilityOnly = checkAvailabilityOnly;
                 this.ParsedCommand = parsedCommand;
             }
 
             public int PID { get; set; }
+
+            public string GitCommandSessionId { get; set; }
 
             public bool IsElevated { get; set; }
 
@@ -317,15 +320,30 @@ namespace GVFS.Common.NamedPipes
 
                     // ParsedCommandLength should be the length of the string at the end of the message
                     // Add the length of the previous parts, plus delimiters
-                    int startingSpot = dataParts[0].Length + dataParts[1].Length + dataParts[2].Length + dataParts[3].Length + 4;
-                    if ((startingSpot + parsedCommandLength) != body.Length)
+                    int commandStartingSpot = dataParts[0].Length + dataParts[1].Length + dataParts[2].Length + dataParts[3].Length + 4;
+                    if ((commandStartingSpot + parsedCommandLength) >= body.Length)
                     {
                         throw new InvalidOperationException(string.Format("Invalid lock message. The parsedCommand is an unexpected length, got: {0} from message: '{1}'", parsedCommandLength, body));
                     }
 
-                    parsedCommand = body.Substring(startingSpot, parsedCommandLength);
+                    parsedCommand = body.Substring(commandStartingSpot, parsedCommandLength);
 
-                    return new LockData(pid, isElevated, checkAvailabilityOnly, parsedCommand);
+                    string sessionIdSubString = body.Substring(commandStartingSpot + parsedCommandLength + 1);
+                    string[] sessionIdParts = sessionIdSubString.Split(MessageSeparator);
+                    if (!int.TryParse(sessionIdParts[0], out int sessionIdLength))
+                    {
+                        throw new InvalidOperationException(string.Format("Invalid lock message. Expected session id length, got: {0} from message: '{1}'", sessionIdParts[0], body));
+                    }
+
+                    int sessionIdStartingSpot = commandStartingSpot + parsedCommandLength + sessionIdParts[0].Length + 2;
+                    if ((sessionIdStartingSpot + sessionIdLength) != body.Length)
+                    {
+                        throw new InvalidOperationException(string.Format("Invalid lock message. The sessionId is an unexpected length, got: {0} from message: '{1}'", sessionIdLength, body));
+                    }
+
+                    string sessionId = body.Substring(sessionIdStartingSpot, sessionIdLength);
+
+                    return new LockData(pid, isElevated, checkAvailabilityOnly, parsedCommand, sessionId);
                 }
 
                 return null;
@@ -333,7 +351,15 @@ namespace GVFS.Common.NamedPipes
 
             internal string ToMessage()
             {
-                return string.Join(MessageSeparator.ToString(), this.PID, this.IsElevated, this.CheckAvailabilityOnly, this.ParsedCommand.Length, this.ParsedCommand);
+                return string.Join(
+                    MessageSeparator.ToString(),
+                    this.PID,
+                    this.IsElevated,
+                    this.CheckAvailabilityOnly,
+                    this.ParsedCommand.Length,
+                    this.ParsedCommand,
+                    this.GitCommandSessionId.Length,
+                    this.GitCommandSessionId);
             }
         }
 
--- a/GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
+++ b/GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
@@ -297,9 +297,9 @@ namespace GVFS.Common.NamedPipes
                     bool checkAvailabilityOnly = false;
                     string parsedCommand = null;
 
-                    if (dataParts.Length < 5)
+                    if (dataParts.Length < 7)
                     {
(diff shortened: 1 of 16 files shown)


GVFS/GVFS.Common/Tracing/ITracer.cs
```diff
@@ -9,6 +9,8 @@ public interface ITracer : IDisposable
         ITracer StartActivity(string activityName, EventLevel level, EventMetadata metadata);
         ITracer StartActivity(string activityName, EventLevel level, Keywords startStopKeywords, EventMetadata metadata);
 
+        void SetGitCommandSessionId(string sessionId);
```
kewillford: Adding this here seems like a hack to me but when the lock request comes in we only have the `ITracer` object.  We could cast to a `JsonTracer` so we don't have to add this here but that to me would also be a hack.  There didn't seem to be any clean way to get the `GitCommandSessionId` to the `TelemetryDaemonEventListener` so if anyone has some brilliant ideas please share.

GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
```diff
@@ -317,23 +320,46 @@ internal static LockData FromBody(string body)
...
+                    int commandStartingSpot = dataParts[0].Length + dataParts[1].Length + dataParts[2].Length + dataParts[3].Length + 4;
+                    if ((commandStartingSpot + parsedCommandLength) >= body.Length)
                     {
                         throw new InvalidOperationException(string.Format("Invalid lock message. The parsedCommand is an unexpected length, got: {0} from message: '{1}'", parsedCommandLength, body));
                     }
 
-                    parsedCommand = body.Substring(startingSpot, parsedCommandLength);
+                    parsedCommand = body.Substring(commandStartingSpot, parsedCommandLength);
+
+                    string sessionIdSubString = body.Substring(commandStartingSpot + parsedCommandLength + 1);
```
kewillford: Need to add some comments about these calculations.  This it ugly and I wish we were simply using `JSON` for the message data especially when both sides are `C#`.  Not sure of the performance impact though.

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -340,6 +341,11 @@ private void HandleReleaseLockRequest(string messageBody, NamedPipeServer.Connec
             }
 
             NamedPipeMessages.ReleaseLock.Response response = this.fileSystemCallbacks.TryReleaseExternalLock(request.RequestData.PID);
+            if (response.Result == NamedPipeMessages.ReleaseLock.SuccessResult)
+            {
+                this.tracer.SetGitCommandSessionId(string.Empty);
```
wilbaker: Do we also need to call `SetGitCommandSessionId(string.Empty)` in `GVFSLock.ReleaseLockForTerminatedProcess`?

wilbaker: (or in `ReleaseExternalLock`)?

wilbaker:
> logging will be able to have this correlation
id to be able to join the git trace2 data with any logged data from the mount
process.

Is the goal just to be able to tell what VFS4G logged while a given git command was running?  

Because VFS4G may log things during a git command that are not necessarily related to that git command (e.g. a git command that does not require the lock could call the read-object hook during another git command that does acquire the lock).

GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
```diff
@@ -317,23 +320,46 @@ internal static LockData FromBody(string body)
...
+                        throw new InvalidOperationException(string.Format("Invalid lock message. Expected session id length, got: {0} from message: '{1}'", sessionIdParts[0], body));
+                    }
+
+                    int sessionIdStartingSpot = commandStartingSpot + parsedCommandLength + sessionIdParts[0].Length + 2;
+                    if ((sessionIdStartingSpot + sessionIdLength) != body.Length)
+                    {
+                        throw new InvalidOperationException(string.Format("Invalid lock message. The sessionId is an unexpected length, got: {0} from message: '{1}'", sessionIdLength, body));
+                    }
+
+                    string sessionId = body.Substring(sessionIdStartingSpot, sessionIdLength);
```
jeffhostetler: I'm having a little trouble following the above, but I think it is expecting a "%d %s" past the end of the previous
message for the SID-length and the SID-string itself.  I think this is fine as long as we treat the content of the
SID-string as an opaque string and not try to parse the content of it.

GVFS/GVFS.Hooks/Program.cs
```diff
@@ -456,5 +459,17 @@ private static bool IsAlias(string command)
...
             return !string.IsNullOrEmpty(result.Output);
         }
+
+        private static string GetGitCommandSessionId()
+        {
+            try
+            {
+                return Environment.GetEnvironmentVariable("GIT_TR2_PARENT_SID", EnvironmentVariableTarget.Process);
+            }
+            catch (SecurityException)
```
jeffhostetler: Should we have a catchall here too?

GVFS/GVFS.UnitTests/Mock/Common/MockTracer.cs
```diff
@@ -132,6 +132,10 @@ public TimeSpan Stop(EventMetadata metadata)
             return TimeSpan.Zero;
         }
 
+        public void SetGitCommandSessionId(string sessionId)
+        {
```
jeffhostetler: Just asking.  Should we do something here?

jeffhostetler:
When Git.exe runs it exports a SID uniquely identifying that invocation of git.exe.  The pre- and post-command hooks inherit that SID in their process environment.  If those hooks pass the SID to the mount
process over the pipe, then the hope was that we could denote any work that the mount process did
for that git.exe process with that SID.

So, for example, if the mount process blocks my pre- hook waiting for a lock, it could annotate the
lock telemetry and include my SID to say that it is locking it for me.

WRT a git command invoking the read-object hook, I already know on the git.exe side how long
it takes for the read-object hook to respond and how many requests were made.  (I'm assuming
here that the mount process is not involved in read-object hook. -- or rather, that the read-object
hook does the work itself and doesn't need to ask the mount process to do the fetch for it.)

Since the mount process is concurrently working for multiple client processes, the SID should
remain attached to the pipe/connection with the client and not put in some global variable.

I guess that means that requests from the kernel driver to dynamically resolve something
will not and should not have a SID (because the faulting process may just be the compiler
or an editor or something).

@wilbaker Hopefully this helps.

wilbaker:
@jeffhostetler 

>I'm assuming here that the mount process is not involved in read-object hook. -- or rather, that the read-object hook does the work itself and doesn't need to ask the mount process to do the fetch for it.

The read-object hook is simply a pass-through to GVFS.Mount.exe.  It connects to the mount process and asks the mount process to download an object on its behalf.

> Since the mount process is concurrently working for multiple client processes, the SID should remain attached to the pipe/connection with the client and not put in some global variable.

As best I can tell the changes in this PR are putting the SID in a global variable when the GVFSLock is acquired, and then removing them from that global variable when the lock is released.  All messages logged during that time will be attributed to the SID (even if they're for unrelated git commands or background maintenance tasks). 

> I guess that means that requests from the kernel driver to dynamically resolve something
will not and should not have a SID (because the faulting process may just be the compiler
or an editor or something).

That's the sort of scenario that prompted my question.  I was concerned that we might incorrectly assume assume that events tagged with a SID were triggered by the git.exe that generated that SID.  As it is right now, the SID would only mean that that work ran concurrently in GVFS.Mount.exe (it could be completely unrelated to the git command).

kewillford:
@wilbaker Yes I discussed this with John and we decided to put it in as a global because the other option would be to pass through the sid to all calls and all hooks would have to handle the sid and pass it through and all telemetry logging would need to include it when they log.  And once we get the process for the file system events we would need to store the sid for each git command and then look it up when the files system event happens.  All of that seemed like overkill for now but if we truly want to match the git command with what exactly VFSForGit did in response there will need to be a lot more changes than this.

kewillford:
@jeffhostetler 
>Since the mount process is concurrently working for multiple client processes, the SID should
remain attached to the pipe/connection with the client and not put in some global variable.

The issue with that is the pre-command hook will run and open a pipe to the mount process but then go away while the git command has a bunch of other work to do, so it cannot be attached to the pipe/connection.  This would work better if we had a wrapper process around git and not the pre/post command hooks.

jeffhostetler:
@kewillford You're right. I always forget that the pre- and post- are 2 separate processes and there's nothing binding the git process to the mount process between the 2.

wilbaker:
>@wilbaker Yes I discussed this with John and we decided to put it in as a global because the other option would be to pass through the sid to all calls and all hooks would have to handle the sid and pass it through and all telemetry logging would need to include it when they log. And once we get the process for the file system events we would need to store the sid for each git command and then look it up when the files system event happens. All of that seemed like overkill for now but if we truly want to match the git command with what exactly VFSForGit did in response there will need to be a lot more changes than this.

@kewillford thanks for confirming! I wanted to make sure we're all on the same page 😃, and you explanation for going with this approach makes sense to me.

jeffhostetler:
(I think I'm getting confused here, but admitting you have a problem is the first step, right?)

So, if we globally record the SID of the process who requested and was granted the lock,
then all we can say is in any telemetry events that the mount process generates is
"SID locked X",  "we think SID is still alive and using X" and when unlocking it "we unlocked
X that we think SID was holding".

So mount process events related to X can name the SID.

And any mount process events unrelated to lock X can either have no SID or can
reference my SID as an "oh, by the way SID is holding X".

So if I'm understanding all this, I think I'm OK with this.

kewillford:
Yep and my one big issue with this is that "sometimes" VFSForGit will not get the post-command hook and will not get "unlocked".  In which case the SID would still be getting placed on events until another command tries to get it and VFSForGit will then recognize that the previous git command that had the lock is gone and release the lock to let the new git command get the lock.  So it might not always all line up.  Maybe we could use the timestamp of the trace2 event to filter those out?

GVFS/GVFS.Common/NamedPipes/LockNamedPipeMessages.cs
kewillford: Yes that is right and I don't like all the formatting and parsing that we are doing ourselves here but it is basically a `|` delimited string but there are strings that are length "prefixed" so that the string itself can have `|`s in it.  So it is something like:
`123|true|false|13|parsedCommand|9|sessionId`

I will add a comment on the format so that this is a little easier to understand.

GVFS/GVFS.UnitTests/Mock/Common/MockTracer.cs
kewillford: Let me look to see if there are some tests where we could validate that this is getting called and set the the right thing.

kewillford: I was able to add to the `GVFSLockTests.cs` but used Moq instead since it was easier.  The other place would be for the `InProcessMount` class which doesn't have any unit tests for it currently so would be a little bigger undertaking to get that one.  So I am going to leave this as is for now since nothing else needs it for testing.  Thanks!

PR #962, opened 2019-03-25
Fix GVFS.FunctionalTests.Windows so that they can be run from the IDE

wilbaker:
The `xcopy` step was overwriting .NET Framework assemblies with .NET Core assemblies, and as a result Test.GVFS.Service was unable to start when running the functional tests locally.

The tests were passing on the build machine because they are run against the installed version of VFS4G rather than the version in the functional tests output directory.

I am still not sure how this code was working:

```
// On Windows, FT is run from the Output directory of GVFS.FunctionalTest project.
// LockHolder is a .netcore assembly and can be found inside netcoreapp2.1
// subdirectory of GVFS.FunctionalTest Output directory.
return Path.Combine(
    Settings.Default.CurrentDirectory,
    "netcoreapp2.1",
    LockHolderCommand);
```

Because as best as I could tell the "netcoreapp2.1" folder was not getting created.


wilbaker:
@kewillford can you confirm if this commit fixes the issue you were seeing?

wilbaker:
@alameenshah can you look into this question?

> I am still not sure how this code was working:

If the "netcoreapp2.1" was not getting created, how was the LockHolder running on Windows?

wilbaker:
> If the "netcoreapp2.1" was not getting created, how was the LockHolder running on Windows?

I discussed this offline with @alameenshah.  This works on the build machine because they run the functional test script, which passes `--windows-only` to `GVFS.FunctionalTests.Windows.exe`, meaning that the LockHolder tests only run in the .NET Core version of `GVFS.FunctionalTests` which does have a "netcoreapp2.1" folder.

wilbaker:
https://gvfs.visualstudio.com/ci/_build/results?buildId=8542

PR #963, opened 2019-03-26
Service porting to Mac

alameenshah:
#### In this PR
1. GVFS.Service project has been separated into GVFS.Service.Mac and GVFS.Service.Windows projects.
2. GVFS.Service.Mac does not do anything useful (Windows specific `PrjFlt, auto-repository mount and background upgrade checks have been removed.) It is just a .netcore port of GVFS.Service Framework project. The assembly gets picked up by the installer and is installed on the Mac. The Service does not get auto-launched.

#### Upcoming Sprint
Mac installer changes that would install Service as a LaunchAgent. The Service will be automatically launched when a user logs in.

#### Future Sprints
1. Auto-mount repos & 2. Background upgrade checks.


GVFS/GVFS.Service/Mac/CurrentUser.cs
```diff
@@ -0,0 +1,59 @@
+﻿using GVFS.Common.Tracing;
```
alameenshah: This is just a placeholder class for now. Its there because `NotificationHandler` class which is included in the project references it.

alameenshah: This is just a placeholder class for now. Its there because NotificationHandler class which is included in the project references it.

GVFS/GVFS.Service/Mac/GVFSMountProcess.cs
```diff
@@ -0,0 +1,60 @@
+﻿using GVFS.Common;
```
alameenshah: Removed `PrjFlt` related code from here. This class is referenced from `RepoRegistration`.

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,122 @@
...
+            NamedPipeMessages.Message message = NamedPipeMessages.Message.FromString(request);
+            if (string.IsNullOrWhiteSpace(message.Header))
+            {
+                return;
+            }
+
+            using (ITracer activity = this.tracer.StartActivity(message.Header, EventLevel.Informational, new EventMetadata { { "request", request } }))
+            {
+                switch (message.Header)
+                {
```
alameenshah: Ignores all messages for the time being

GVFS/GVFS.Service/Mac/CurrentUser.cs
```diff
@@ -0,0 +1,59 @@
...
+        public CurrentUser(ITracer tracer, int sessionId)
+        {
+            this.tracer = tracer;
+            this.token = GetCurrentUserToken(this.tracer, sessionId);
+            this.Identity = null;
+        }
+
+        public UserIdentity Identity { get; }
+
+        public bool RunAs(string processName, string args)
```
nickgra: Some of these methods (like RunAs) seem like they're only going to be relevant on Windows. If we want to start with a skeleton for now, should we just make these classes empty rather than putting in stub code now, only to delete it in the next PR?

GVFS/GVFS.Service/Mac/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace GVFS.Service
+{
+    public static class Program
+    {
+        public static void Main(string[] args)
+        {
+            GVFSPlatformLoader.Initialize();
+
+            AppDomain.CurrentDomain.UnhandledException += UnhandledExceptionHandler;
```
nickgra: I think AppDomain is Framework only and didn't make it into .NET Core?

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public void RunWithArgs(string[] args)
+        {
+            if (this.serviceThread != null)
+            {
+                throw new InvalidOperationException("Cannot start service twice in a row.");
+            }
+
+            // TODO: 865304 Used for functional tests and development only. Replace with a smarter appConfig-based solution
```
jrbriggs: Let's strike the numerical references to the old internal work items.

GVFS/GVFS.Service/Mac/Program.cs
jrbriggs: I believe it made it over in .NET Core 2.0: https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.unhandledexception?view=netcore-2.0

GVFS/GVFS.Service/GVFS.Service.Mac.csproj
```diff
@@ -0,0 +1,48 @@
...
+    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
+  </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.1.1-beta" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.0.2">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+</Project>
```
jrbriggs: Nit: newline at EOF.

GVFS/GVFS.Service/Mac/CurrentUser.cs
alameenshah: Removed all Stubs

GVFS/GVFS.Service/Mac/GVFSMountProcess.cs
alameenshah: Removed all Stubs

alameenshah: Removed all Stubs

GVFS/GVFS.Service/Mac/CurrentUser.cs
alameenshah: Removed all the un-used Handler classes and the stubs they needed. 

Looks much cleaner now -

Screen Shot 2019-03-27 at 3 26 11 PM

GVFS/GVFS.Service/Mac/Program.cs
alameenshah: Right - its available in the [dotnet core sources](https://github.com/dotnet/corefx/find/master).

We are also using it in `GVFS.Mount.Mac::InProcessMountVerb::Execute()`

GVFS/GVFS.Service/GVFS.Service.Mac.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <Version>$(GVFSVersion)</Version>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <Version>$(GVFSVersion)</Version>
+  </PropertyGroup>
+  <ItemGroup>
```
nickgra: Do we need to explicitly add the None glob here if we have `<EnableDefaultCompileItems>false</EnableDefaultCompileItems>`?

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,115 @@
+using GVFS.Common;
+using GVFS.Common.NamedPipes;
+using GVFS.Common.Tracing;
+using System;
+using System.Linq;
+using System.Threading;
+
+namespace GVFS.Service
```
wilbaker: Minor: Namespace should match folder structure `namespace GVFS.Service` -> `namespace GVFS.Service.Mac`

GVFS/GVFS.Service/Mac/Program.cs
```diff
@@ -0,0 +1,32 @@
+﻿using GVFS.Common;
+using GVFS.Common.Tracing;
+using GVFS.PlatformLoader;
+using System;
+
+namespace GVFS.Service
```
wilbaker: `GVFS.Service` -> `GVFS.Service.Mac`

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,115 @@
...
+    public class GVFSService
+    {
+        private const string ServiceNameArgPrefix = "--servicename=";
+        private const string EtwArea = nameof(GVFSService);
+
+        private JsonTracer tracer;
+        private Thread serviceThread;
+        private ManualResetEvent serviceStopped;
+        private string serviceName;
+        private string serviceDataLocation;
```
wilbaker: `serviceDataLocation` is never used and can be removed

GVFS/GVFS.Service/Mac/Program.cs
```diff
@@ -0,0 +1,32 @@
+﻿using GVFS.Common;
+using GVFS.Common.Tracing;
+using GVFS.PlatformLoader;
+using System;
+
+namespace GVFS.Service
+{
+    public static class Program
```
wilbaker: To confirm, is this the machine-wide launchd service, or will this be the per-user service?

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            string nameArg = args.FirstOrDefault(arg => arg.StartsWith(ServiceNameArgPrefix, StringComparison.OrdinalIgnoreCase));
+            if (nameArg != null)
+            {
+                this.serviceName = nameArg.Substring(ServiceNameArgPrefix.Length);
+            }
+
+            try
+            {
+                this.serviceDataLocation = Paths.GetServiceDataRoot(this.serviceName);
```
wilbaker: What will `Paths.GetServiceDataRoot(this.serviceName)` be on Mac?

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        private void StartServiceThread()
+        {
+            if (this.serviceStopped != null)
+            {
+                return;
+            }
+
+            this.serviceStopped = new ManualResetEvent(false);
```
wilbaker: Rather than having this code here:

```
if (this.serviceStopped != null)
{
    return;
}

this.serviceStopped = new ManualResetEvent(false);
```

Can we just construct the `ManualResetEvent` in the constructor?

(I am not quite sure why the `if (this.serviceStopped != null)` check is there right now)

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,115 @@
...
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add("Version", ProcessHelper.GetCurrentProcessVersion());
+                this.tracer.RelatedEvent(EventLevel.Informational, $"{nameof(GVFSService)}_{nameof(this.ServiceThreadMain)}", metadata);
+
+                string pipeName = this.serviceName + ".Pipe";
+                this.tracer.RelatedInfo("Starting pipe server with name: " + pipeName);
+
+                using (NamedPipeServer pipeServer = NamedPipeServer.StartNewServer(pipeName, this.tracer, this.HandleRequest))
+                {
+                    this.serviceStopped.WaitOne();
```
wilbaker: Call `Dispose` on `serviceStopped` before exiting `ServiceThreadMain`?

GVFS/GVFS.Service/Mac/GVFSService.cs
```diff
@@ -0,0 +1,115 @@
...
+            try
+            {
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add("Version", ProcessHelper.GetCurrentProcessVersion());
+                this.tracer.RelatedEvent(EventLevel.Informational, $"{nameof(GVFSService)}_{nameof(this.ServiceThreadMain)}", metadata);
+
+                string pipeName = this.serviceName + ".Pipe";
+                this.tracer.RelatedInfo("Starting pipe server with name: " + pipeName);
+
+                using (NamedPipeServer pipeServer = NamedPipeServer.StartNewServer(pipeName, this.tracer, this.HandleRequest))
```
wilbaker: Do we need/want to start the named pipe server in this iteration of GVFS.Service on Mac?

Also, once we know where GVFS.Service will store it's config files we should probably switch to using [a rooted pipe name](https://github.com/Microsoft/VFSForGit/blob/d02c62832d512ab3ef91aeed5d0457b4675934da/GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs#L35) to make the pipe more easily discoverable.

GVFS/GVFS.Service/Mac/GVFSService.cs
wilbaker: Do we expect any messages at this point?

GVFS/GVFS.Service/GVFS.Service.Mac.csproj
```diff
@@ -0,0 +1,42 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\GVFS.Build\GVFS.cs.props" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>GVFS.Service</RootNamespace>
+    <AssemblyName>GVFS.Service</AssemblyName>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <Platforms>x64</Platforms>
+    <RunTimeIdentifiers>osx-x64</RunTimeIdentifiers>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```
wilbaker: `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` is set in `..\GVFS.Build\GVFS.cs.props`:

https://github.com/Microsoft/VFSForGit/blob/d02c62832d512ab3ef91aeed5d0457b4675934da/GVFS/GVFS.Build/GVFS.cs.props#L15

And so I think we should be able to remove it here, @nickgra would you agree?

GVFS/GVFS.Service/Mac/GVFSService.cs
alameenshah: it returns `/usr/share/GVFS/GVFS.Service`

GVFS/GVFS.Service/Mac/GVFSService.cs
alameenshah: Moved the creation to constructor.

On Windows I think this check `if (this.serviceStopped != null)` helps avoid the service thread from being started multiple times. This pattern does not apply on the Mac - since launchd only starts the service once (there is no suspend/resume)

GVFS/GVFS.Service/Mac/GVFSService.cs
alameenshah: We don't need it now. Can add it later with rooted pipe name.

GVFS/GVFS.Service/GVFS.Service.Mac.csproj
alameenshah: removed it.

GVFS/GVFS.Service/Mac/GVFSService.cs
alameenshah: No. Removed it.

Issue #183, opened 2018-08-16
Mac: Return deny reason from kext when VFSForGit fails a request


wilbaker:
Currently the kext is ignoring the specific reason that VFSForGit failed a request.  See if it's possible to bubble this detail up higher.

PR #964, opened 2019-03-26
Log when a path can't be looked up from fsid

### LINKED ISSUE #959: Mac: Null path was delivered to OnPreDelete callback, causing VFS4G to crash
It looks like this can happen if `fsgetpath` fails in user-mode [before delivering a callback](https://github.com/Microsoft/VFSForGit/blob/cbcec0ee144b2340112eec72321efc5a8d4ae84b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L650).

>[2019-03-21 21:14:52 -07:00] Error {"Area":"MacFileSystemVirtualizer","Exception":"System.NullReferenceException: Object reference not set to an instance of an object.\n   at GVFS.Virtualization.FileSystemCallbacks.IsPathInsideDotGit(String relativePath) in /Users/vsts/agent/2.148.0/work/1/s/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs:line 163\n   at GVFS.Platform.Mac.MacFileSystemVirtualizer.OnPreDelete(String relativePath, Boolean isDirectory) in /Users/vsts/agent/2.148.0/work/1/s/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs:line 489","isDirectory":false,"ErrorMessage":"OnPreDelete caught unhandled exception, exiting process"}
[2019-03-21 21:14:52 -07:00] Error {"Area":"MacFileSystemVirtualizer","Exception":"System.NullReferenceException: Object reference not set to an instance of an object.\n   at GVFS.Virtualization.FileSystemCallbacks.IsPathInsideDotGit(String relativePath) in /Users/vsts/agent/2.148.0/work/1/s/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs:line 163\n   at GVFS.Platform.Mac.MacFileSystemVirtualizer.OnPreDelete(String relativePath, Boolean isDirectory) in /Users/vsts/agent/2.148.0/work/1/s/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs:line 489","isDirectory":false,"ErrorMessage":"OnPreDelete caught unhandled exception, exiting process"}

jeschu1:
- Remove the crash
- Add logging around this case

resolves #959

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -431,10 +431,9 @@ namespace GVFS.Platform.Mac
             return Result.EIOError;
         }
 
-        private Result OnLogError(string errorMessage)
+        private void OnLogError(string errorMessage)
         {
             this.Context.Tracer.RelatedError($"{nameof(MacFileSystemVirtualizer)}::{nameof(this.OnLogError)}: {errorMessage}");
-            return Result.Success;
         }
 
         private void OnFileModified(string relativePath)
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -213,6 +213,7 @@ namespace GVFS.Platform.Mac
             // Callbacks
             this.virtualizationInstance.OnEnumerateDirectory = this.OnEnumerateDirectory;
             this.virtualizationInstance.OnGetFileStream = this.OnGetFileStream;
+            this.virtualizationInstance.OnLogError = this.OnLogError;
             this.virtualizationInstance.OnFileModified = this.OnFileModified;
             this.virtualizationInstance.OnPreDelete = this.OnPreDelete;
             this.virtualizationInstance.OnNewFileCreated = this.OnNewFileCreated;
@@ -428,6 +429,12 @@ namespace GVFS.Platform.Mac
             }
 
             return Result.EIOError;
+        }
+
+        private Result OnLogError(string errorMessage)
+        {
+            this.Context.Tracer.RelatedError($"{nameof(MacFileSystemVirtualizer)}::{nameof(this.OnLogError)}: {errorMessage}");
+            return Result.Success;
         }
 
         private void OnFileModified(string relativePath)
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -45,11 +45,11 @@ using std::map;
 using std::move;
 using std::mutex;
 using std::oct;
+using std::ostringstream;
 using std::pair;
 using std::queue;
 using std::set;
 using std::shared_ptr;
-using std::stringstream;
 using std::stack;
 using std::string;
 
@@ -655,7 +655,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
         if (pathSize < 0)
         {
             // TODO(Mac): Add this message to PrjFSLib logging once available (#395)
-            stringstream ss;
+            ostringstream ss;
             ss
                 << "MessageType: " << requestHeader->messageType << " "
                 << "PrjFSLib.HandleKernelRequest: fsgetpath failed for fsid 0x"
@@ -669,11 +669,8 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
             string errorMessage = ss.str();
 
             s_callbacks.LogError(errorMessage.c_str());
-            cout << errorMessage << endl;
-
-            free(messageMemory);
             result = PrjFS_Result_EPathNotFound;
-            return;
+            goto CleanupAndReturn;
         }
         else
         {
@@ -751,6 +748,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     // async callbacks are not yet implemented
     assert(PrjFS_Result_Pending != result);
     
+CleanupAndReturn:
     if (PrjFS_Result_Pending != result)
     {
         MessageType responseType =
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -669,7 +669,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
             string errorMessage = ss.str();
 
             s_callbacks.LogError(errorMessage.c_str());
-            result = PrjFS_Result_EPathNotFound;
+            result = PrjFS_Result_Success;
             goto CleanupAndReturn;
         }
         else
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -18,6 +18,8 @@
 #include <IOKit/IODataQueueClient.h>
 #include <mach/mach_port.h>
 #include <CoreFoundation/CFNumber.h>
+#include <string>
+#include <sstream>
 
 #include "stdlib.h"
 
@@ -47,6 +49,7 @@ using std::pair;
 using std::queue;
 using std::set;
 using std::shared_ptr;
+using std::stringstream;
 using std::stack;
 using std::string;
 
@@ -163,13 +166,15 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
         << callbacks.EnumerateDirectory << ", "
         << callbacks.GetFileStream << ", "
         << callbacks.NotifyOperation << ", "
+        << callbacks.LogError << ","
         << poolThreadCount << ")" << endl;
 #endif
     
     if (nullptr == virtualizationRootFullPath ||
         nullptr == callbacks.EnumerateDirectory ||
         nullptr == callbacks.GetFileStream ||
-        nullptr == callbacks.NotifyOperation)
+        nullptr == callbacks.NotifyOperation ||
+        nullptr == callbacks.LogError)
     {
         return PrjFS_Result_EInvalidArgs;
     }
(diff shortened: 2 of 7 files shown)


jeschu1:
@nickgra I would greatly appreciate it if you can pull/build/test this branch against your repro.  It will also give us insight into which messageType is failing.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs2
```diff
@@ -0,0 +1,679 @@
+using GVFS.Common;
```
wilbaker: Was this file added by mistake?

jeschu1: Yup just repushed to remove

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -657,6 +666,14 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
                 << errno
                 << "(" << strerror(errno) << ")"
                 << endl;
+            const char* errorMessage = ss.str().c_str();
```
pmj: I'm not convinced this call sequence is safe. As far as I'm aware `stringstream::str()` returns a temporary `string` object, so the string buffer returned by `c_str()` on that temp object will presumably also be transient. The solution would be to make `errorMessage` a `string` storing the result of `ss.str()` so that doesn't go out of scope until `s_callbacks.LogError(errorMessage.c_str())` returns.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -639,16 +644,20 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
     if (request.path == nullptr)
     {
         fsid_t fsid = request.messageHeader->fsidInode.fsid;
         ssize_t pathSize = fsgetpath(pathBuffer, sizeof(pathBuffer), &fsid, request.messageHeader->fsidInode.inode);
+
         if (pathSize < 0)
         {
             // TODO(Mac): Add this message to PrjFSLib logging once available (#395)
-            cout
+            stringstream ss;
```
pmj: Very minor: consider using an `ostringstream`.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: @pmj let me know if the update looks good

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -428,6 +429,12 @@ private bool TryGetSymLinkTarget(string sha, out string symLinkTarget)
             }
 
             return Result.EIOError;
+        }
+
+        private Result OnLogError(string errorMessage)
```
wilbaker: Minor: Return type can be `void` instead of `Result` as we're always returning `Success` and the caller isn't checking the return value.

wilbaker: If you'd like, this change can be left for #395

jeschu1:
> Also, are we still maintaining the mirror provider? If so, I think the change to _PrjFS_Callbacks will break that; the mirror provider will need to implement the OnLogError callback as well.

I could use more info on the MirrorProvider impact here.  I would expect it to run, but not do anything for this event.  Considering we're doing this as a patch and it may be a short lived function, I would think it's OK to skip MirrorProvider for now.  Am I missing anything?  @wilbaker

wilbaker:
> I could use more info on the MirrorProvider impact here. I would expect it to run, but not do anything for this event. Considering we're doing this as a patch and it may be a short lived function, I would think it's OK to skip MirrorProvider for now. Am I missing anything? @wilbaker

@pmj is correct that we will break MirrorProvider with this change if we don't update it.

Since the update to MirrorProvider is very minor (it just needs to implement `OnLogError` and register it as a callback) I think we should update it as well.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -212,7 +212,8 @@ protected override bool TryStart(out string error)
 
             // Callbacks
             this.virtualizationInstance.OnEnumerateDirectory = this.OnEnumerateDirectory;
-            this.virtualizationInstance.OnGetFileStream = this.OnGetFileStream;
+            this.virtualizationInstance.OnGetFileStream = this.OnGetFileStream;
```
wilbaker: All we need for the mirror provider is this line:

```
this.virtualizationInstance.OnLogError = this.OnLogError;
```

And an implementation of `private Result OnLogError(string errorMessage)` (it could just write the error to the console).

jeschu1: I'll update

pmj:
> I could use more info on the MirrorProvider impact here. I would expect it to run, but not do anything for this event.

Well, `PrjFS_StartVirtualizationInstance()` [explicitly fails if `OnLogError` is null](https://github.com/Microsoft/VFSForGit/pull/964/commits/e341ecfaafa6a33f7c647aeda93c1a0a9a550cb6#diff-5c944eca485016533d3f61ab623cb0a9R177). If you want it to fall back to doing nothing, you'll need to remove that check and instead guard for null when calling the callback.

> Considering we're doing this as a patch and it may be a short lived function, I would think it's OK to skip MirrorProvider for now. Am I missing anything? @wilbaker

Yeah, I can't comment on the status of where this patch is ending up, all I'm seeing is that this PR is attempting to merge into master, so I assumed it was going to stick around.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
jeschu1: We might as well update it now, I'll do that.  We don't check the result and the tracer function we call is a void.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -657,6 +666,14 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
                 << errno
                 << "(" << strerror(errno) << ")"
                 << endl;
+            string errorMessage = ss.str();
+
+            s_callbacks.LogError(errorMessage.c_str());
+            cout << errorMessage << endl;
+
+            free(messageMemory);
+            result = PrjFS_Result_EPathNotFound;
```
wilbaker: We need to call `SendKernelMessageResponse(requestHeader->messageId, responseType);` with this result or the kext will hang waiting on our response.

Rather than an early return adding a `goto` and a `label` (where we do this below) might be the cleaner options.

pmj: Good catch! Make sure we don't double-free the `messageMemory`. (One of these days I want to come up with a good RAII way of handling that memory, but that's not for now.)

wilbaker: > with this result

Actually, it looks like we're currently mapping the result to Success/Failure (see below).

#183 is open for sending back the the more detailed error code.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -657,6 +666,14 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
                 << errno
                 << "(" << strerror(errno) << ")"
                 << endl;
+            string errorMessage = ss.str();
+
+            s_callbacks.LogError(errorMessage.c_str());
+            cout << errorMessage << endl;
```
wilbaker: Now that we're sending this message to the provider, should we still be writing it to the console?

jeschu1: I was debating that too.  Since we don't need it i'll remove the cout.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: `PrjFS_Result_EPathNotFound`

@jeschu1 before merging, can you check with @nickgra how returning either success or failure in this situation impacts his scenario?

I'm wondering if we have two requests racing for to delete the same file, will returning Failure here cause the second delete to fail when perhaps it would have otherwise succeeded (or resulted in an error of "No such file or directory")?

If anything this might raise the importance of #183 as we'd be able to return a more describe error code here.

cc: @pmj for his thoughts on this as well.

jeschu1: @wilbaker , this is a good question.

I guess whats the harm in us returning success?  We'll have the entry in logs / telemetry so we don't really need to fail.

jeschu1: We talked offline and decided always returning success is the best approach for now.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -657,6 +666,10 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
                 << errno
                 << "(" << strerror(errno) << ")"
                 << endl;
+            string errorMessage = ss.str();
+
+            s_callbacks.LogError(errorMessage.c_str());
+            goto CleanupAndReturn;
```
wilbaker: Just to be explicit about how we're handling, set `result` here before going to `CleanupAndReturn`?

wilbaker: @nickgra @jeschu1 I think the changes look good and ready to test.  @nickgra can you give them a try (current changes return failure, could you test out returning both failure and success)?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: @nickgra actually maybe start running tests where we return success?  I think we still want both data points thought.

pmj:
FWIW, latest round of changes looks good to me.

PR #965, opened 2019-03-26
[Port from releases/shipped] Turn off Commit Graph Verify

jeschu1:
- temporarily remove until performance issues are addressed

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
@@ -48,12 +48,15 @@ namespace GVFS.Common.Maintenance
 
                 this.RunGitCommand((process) => process.WriteCommitGraph(this.Context.Enlistment.GitObjectsRoot, this.packIndexes), nameof(GitProcess.WriteCommitGraph));
 
+                // Turning off Verify for commit graph due to performance issues.
+                /*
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
```diff
@@ -48,12 +48,15 @@ protected override void PerformMaintenance()
 
                 this.RunGitCommand((process) => process.WriteCommitGraph(this.Context.Enlistment.GitObjectsRoot, this.packIndexes), nameof(GitProcess.WriteCommitGraph));
 
+                // Turning off Verify for commit graph due to performance issues.
+                /*
                 GitProcess.Result verifyResult = this.RunGitCommand((process) => process.VerifyCommitGraph(this.Context.Enlistment.GitObjectsRoot), nameof(GitProcess.VerifyCommitGraph));
 
                 if (verifyResult.ExitCodeIsFailure)
```
derrickstolee: @jeschu1 This `if` statement needs to be updated to check if we are stopping the step! The process can fail if we halt due to an unmount command.

derrickstolee: (I'll make a new PR that does that, and does the same for the multi-pack-index verify command.)

PR #966, opened 2019-03-26
Credentials: consistently use usehttppath for dev.azure.com scope

jamill:
The git credential commands are not consistent in whether they set
usehttppath or not. This can lead to inconsistencies in how credentials
are generatated and cleared. VFSForGit might retrieve / create
credentials using the full http path, but clear with the only the host
name. This leads to the incorrect behavior where VFSForGit is unable to
erase bad / stale credentials.

Additionally, only dev.azure.com URLs need to set the useHttpPath
setting - so scope it to these URLs.


jamill:
Covered by #939

PR #972, opened 2019-03-27
Install GVFS.Service as a Launch Agent on Mac

alameenshah:
#### Install GVFS.Service as a Launch Agent on Mac

* Added new Launchd plist for GVFS.Service. 
* Updated Mac installer to auto-load the Service as a User agent for all logged in users. 
* Updated Uninstaller to unload the service. 
* Also added title text to Mac installer.

### DIFF
--- a/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
+++ b/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <installer-gui-script minSpecVersion="1">
+    <title>VFS For Git</title>
     <pkg-ref id="com.vfsforgit.pkg"/>
     <pkg-ref id="com.git.pkg"/>
     <options customize="never" allow-external-scripts="true"/>
--- a/GVFS/GVFS.Installer.Mac/scripts/postinstall
+++ b/GVFS/GVFS.Installer.Mac/scripts/postinstall
@@ -2,3 +2,17 @@
 loadCmd="sudo launchctl load -w /Library/LaunchDaemons/org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist"
 echo "Loading PrjFSKextLogDaemon: '$loadCmd'..."
 eval $loadCmd || exit 1
+
+# Load Service in all active User sessions
+# There will be one loginwindow instance for each logged in user, 
+# get its uid (this will correspond to the logged in user's id.) 
+# Then use launchctl bootstrap gui/uid to auto load the Service 
+# for each user.
+servicePlist="/Library/LaunchAgents/org.vfsforgit.service.plist"
+if [ -f "${servicePlist}" ]; then
+    for uid in $(ps -Ac -o uid,command | grep -iw "loginwindow" | awk '{print $1}'); do
+        loadCmd="launchctl bootstrap gui/$uid ${servicePlist}"
+        echo "Loading Service: '$loadCmd'..."
+        eval $loadCmd || exit 1
+    done    
+fi
--- a/GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
+++ b/GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
@@ -4,7 +4,9 @@ KEXTFILENAME="PrjFSKext.kext"
 VFSFORDIRECTORY="/usr/local/vfsforgit"
 PRJFSKEXTDIRECTORY="/Library/Extensions"
 LAUNCHDAEMONDIRECTORY="/Library/LaunchDaemons"
+LAUNCHAGENTDIRECTORY="/Library/LaunchAgents"
 LOGDAEMONLAUNCHDFILENAME="org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist"
+SERVICEAGENTLAUNCHDFILENAME="org.vfsforgit.service.plist"
 GVFSCOMMANDPATH="/usr/local/bin/gvfs"
 UNINSTALLERCOMMANDPATH="/usr/local/bin/uninstall_vfsforgit.sh"
 INSTALLERPACKAGEID="com.vfsforgit.pkg"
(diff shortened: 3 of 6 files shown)


GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <installer-gui-script minSpecVersion="1">
+	<title>VFSForGit Mac</title>
```
alameenshah: This is a placeholder text, please suggest alternate title. It appears in the Mac installer Wizard, like this -

Screen Shot 2019-03-27 at 12 08 19 PM

GVFS/GVFS.Installer.Mac/scripts/postinstall
```diff
@@ -1,4 +1,16 @@
 #!/bin/bash
 loadCmd="sudo launchctl load -w /Library/LaunchDaemons/org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist"
 echo "Loading PrjFSKextLogDaemon: '$loadCmd'..."
-eval $loadCmd || exit 1
\ No newline at end of file
+eval $loadCmd || exit 1
+
+# Load Service in all active User sessions
+servicePlist="/Library/LaunchAgents/org.vfsforgit.service.plist"
```
alameenshah: Gets the `loginwindow` `pid` and `uid` for each logged in user. Then use `launchctl bsexec` to auto load the Service for each user.

GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
```diff
@@ -43,6 +45,20 @@ function UnInstallVFSForGit()
         eval $rmCmd || exit 1
     fi
     
+    if [ -f "${LAUNCHAGENTDIRECTORY}/$SERVICEAGENTLAUNCHDFILENAME" ]; then
```
alameenshah: Same approach as above.

GVFS/GVFS.Installer.Mac/scripts/postinstall
nickgra: This is good information, can we put this into the comment in the shell script? When it comes to shell scripts, I'd rather have the behavior clearly spelled out rather than trying to infer it.

GVFS/GVFS.Service/Mac/org.vfsforgit.service.plist
```diff
@@ -0,0 +1,25 @@
...
+<dict>
+	<key>Label</key>
+	<string>org.vfsforgit.service</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>/usr/local/share/dotnet/dotnet</string>
+		<string>/usr/local/vfsforgit/GVFS.Service.dll</string>
+	</array>
+	<key>WorkingDirectory</key>
+    <string>/usr/local/vfsforgit</string>
```
nickgra: nit: ws

GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
nickgra: Can we leave a comment here too?

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <installer-gui-script minSpecVersion="1">
+	<title>VFSForGit</title>
```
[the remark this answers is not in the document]
jrbriggs: Let's try this:
```suggestion
    <title>VFS For Git</title>
```

GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
```diff
@@ -43,6 +45,23 @@ function UnInstallVFSForGit()
...
+    # Unloading Service LaunchAgent for each user
+    # There will be one loginwindow instance for each logged in user, 
+    # get its uid (this will correspond to the logged in user's id.) 
+    # Then use launchctl bootstrap gui/uid to auto load the Service 
+    # for each user.
+    if [ -f "${LAUNCHAGENTDIRECTORY}/$SERVICEAGENTLAUNCHDFILENAME" ]; then
+        for uid in $(ps -Ac -o uid,command | grep -iw "loginwindow" | awk '{print $1}'); do
+            unloadCmd="sudo launchctl bootout gui/$uid ${LAUNCHAGENTDIRECTORY}/$SERVICEAGENTLAUNCHDFILENAME"
+            echo "$unloadCmd..."
+            eval $unloadCmd || exit 1
```
jrbriggs: We don't have to block on this, since I think we already have this bug on Windows, but what happens if another user has an active mount process/GVFS.Service when I uninstall VFS for Git? Do we cleanly shut things down? Is there an opportunity to handle this more gracefully on mac?

GVFS/GVFS.Service/Mac/org.vfsforgit.service.plist
```diff
@@ -0,0 +1,25 @@
...
+    <key>ProgramArguments</key>
+    <array>
+        <string>/usr/local/share/dotnet/dotnet</string>
+        <string>/usr/local/vfsforgit/GVFS.Service.dll</string>
+    </array>
+    <key>WorkingDirectory</key>
+    <string>/usr/local/vfsforgit</string>
+    <key>ProcessType</key>
+    <string>Interactive</string>
+    <key>Disabled</key>
```
jrbriggs: I don't know what's more common on macOS, but I've seen a few example plists like:

```xml
    <key>Name1</key>    <string>Value1</string>
    <key>Name2</key>    <true/>
```
That's a bit easier for me to read over, but definitely not worth blocking the PR for.

GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
alameenshah: `GVFS.Service` is gracefully unloaded. Created an [issue](https://github.com/Microsoft/VFSForGit/issues/978) for more graceful shutdown of `GVFS.Mount`.

GVFS/GVFS.Service/Mac/org.vfsforgit.service.plist
alameenshah: will take care of this in next PR. Tracking [issue](https://github.com/Microsoft/VFSForGit/issues/979).

PR #976, opened 2019-03-29
[Linux] Rebase Linux prototype against current master

chrisd8088:
Hi @jrbriggs (or @wilbaker) ... as I don't have permissions to rebase the `features/linuxprototype` branch in your repo, so could I ask you to do it, in the hope it will the [transient CI issue](https://github.com/Microsoft/VFSForGit/pull/848#issuecomment-476828805) we discussed?

Also, it's also good for us to catch up to where `master` is at!  :-)

I created this branch as a rebase of `features/linuxprototype` on top of the latest upstream `master`; I hope that might help.

And as soon as that's done, I have the tiny follow-up in #977 which syncs us up with the small set of changes which affect our `ProjFS.Linux` project and `MirrorProvider.Linux`.

Thanks very much, and let me know if there's anything I can do to help!  Thanks and cheers, Chris.

PR #977, opened 2019-03-29
Linux rebase callback sync

chrisd8088:
@jrbriggs ... This is the small followup to #976 which syncs the Linux implementations of `MirrorProvider` and `VirtualizationInstance` against the recent changes to the Mac equivalents, as introduced in cada100.


chrisd8088:
There's the [CI bug](https://github.com/Microsoft/VFSForGit/pull/848#issuecomment-476828805) again; hopefully the rebase in #976 will get rid of that!

chrisd8088:
Hi @jrbriggs -- well, this is up-to-date and ready, I think, despite the CI failure; this time it appears to be due to a timeout from the GVFS cache server:
```
Unable to query /gvfs/config
Server returned error code 503 (ServiceUnavailable)
```
So as before, I'm confident this has nothing to do with the changes in this PR, at least.

PR #983, opened 2019-03-29
[Linux] Remove Mono dependency from Linux MirrorProvider

chrisd8088:
@jrbriggs and @wilbaker -- after the rebase in #976, would you be able to look at this PR which removes the Mono dependency in the Linux `MirrorProvider`?

The one question mark I have is that I've tied the implementation to systems with GNU libc-compatible ABIs for `libc.so` and also to those with `struct stat` field layouts as found on x86 64-bit architectures.  See my ["item A" comments](https://github.com/Microsoft/VFSForGit/pull/848#issuecomment-476827046) in #848 for more background, but it basically boils down to the heterogeneity and history of implementations of `stat(2)` and friends.  If this seems too fragile to you (or you, @kivikakk) for the longer-term, we could definitely add a small wrapper `.so` library around `stat()`, `lstat()`, etc.

/cc @kivikakk

### DIFF
--- a/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
+++ b/MirrorProvider/MirrorProvider.Linux/LinuxFileSystemVirtualizer.cs
@@ -106,18 +106,21 @@ namespace MirrorProvider.Linux
                     else
                     {
                         string childRelativePath = Path.Combine(relativePath, child.Name);
+                        string childFullPathInMirror = this.GetFullPathInMirror(childRelativePath);
                         NativeMethods.StatBuffer statBuffer = new NativeMethods.StatBuffer();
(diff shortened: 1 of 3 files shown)


chrisd8088:
OK, this has been rebased and is ready for consideration and comment, I think!

/cc @jrbriggs, @wilbaker, @kivikakk

chrisd8088:
Good catch, @kivikakk -- yes, there's a NuGet package dependency we can also drop!  👍

chrisd8088:
@jrbriggs / @wilbaker -- if this looks good to you also, I think it's ready to merge.

chrisd8088:
I think this one is ready, @jrbriggs, if it looks OK to you ...

Issue #846, opened 2019-02-25
Move all relative path calculations out of the kext and into user mode


wilbaker:
Follow up work item from discussion in #741

* For the remaining path-based kernel-user messages (in the Fileop handler), we're computing the relative path based on the root's path. This is then typically turned back into an absolute path in the provider. Instead, we could simply send the absolute path from the kext into the provider and generate the relative path where needed, or directly use the absolute path where that's the most convenient.
* If we get rid of the relative path calculation, we can completely get rid of the path field from the virtualisation root struct.

wilbaker:
@pmj was this work item completely resolved by #984? 

Or is some of the work in #985 related to this work item?

pmj:
Yeah, this was all sorted in #984.

Issue #356, opened 2018-10-08
Mac: Deliver oldPath to VFS for Git for rename notifications


wilbaker:
This is now required for removing old entries from ModifiedPaths.dat (see #207 for details on the ModifiedPaths.dat change)

wilbaker:
@jeschu1 can this issue be closed now?

jeschu1:
yes :-)

PR #984, opened 2019-03-30
Mac ProjFS: Moves all relative path calculations to user space.

pmj:
This change removes the relative path calculation from the kext and changes the native ProjFS library to expect absolute paths in messages from the kernel. The absolute path is passed through to event handlers where in many cases the logic for re-combining the relative path into an absolute one could be removed as well.

This change resolves task #846.

Although not strictly a prerequisite for my work on #337 and #356, I did this now as it makes those tasks vastly easier.


ProjFS.Mac/PrjFSKext/public/PrjFSCommon.h
```diff
@@ -10,7 +10,7 @@
 #define PrjFSServiceClass       "org_vfsforgit_PrjFS"
 
 // TODO: move this to an autogenerated header. (#648)
-#define PrjFSKextVersion "0.1.1"
+#define PrjFSKextVersion "0.1.2"
```
jeschu1: What circumstances do we up this number?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -690,24 +694,29 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
 #endif
         }
     }
+    else
+    {
+        absolutePath = request.path;
```
jeschu1: Under what cases would we expect a request.path?  (FileOps?)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -646,6 +646,9 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     Message request = ParseMessageMemory(messageMemory, messageSize);
     const MessageHeader* requestHeader = request.messageHeader;
     
+    const char* absolutePath = nullptr;
+    const char* relativePath = nullptr;
```
jeschu1: I very much like the naming change here to keep it as absolutePath/relativePath all the way thru.  Very clear!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -762,7 +767,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     free(messageMemory);
 }
 
-static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* relativePath)
+static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* fullPath, const char* relativePath)
 {
 #ifdef DEBUG
     cout
```
jeschu1: As part of this debug, would we want to print out the new full Path?  This view cuts it off but you can see the debug on the full file.  This question for all the affected functions.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -762,7 +767,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     free(messageMemory);
 }
 
-static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* relativePath)
+static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* fullPath, const char* relativePath)
```
jeschu1: In the signature we use absolutePath not fullPath.
The existing convention for the file appears to be fullPath.  I prefer absolutePath myself but may be worth sticking to fullPath to avoid larger changes in this PR.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -838,7 +841,7 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
         
         CombinePaths(s_virtualizationRootFullPath.c_str(), directoryRelativePath.c_str(), path);
     
-        PrjFS_Result result = HandleEnumerateDirectoryRequest(request, directoryRelativePath.c_str());
+        PrjFS_Result result = HandleEnumerateDirectoryRequest(request, path, directoryRelativePath.c_str());
```
jeschu1: I realize this is existing but directoryAbsolutePath would be clearer than 'path'

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -873,7 +876,7 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
     return result;
 }
 
-static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const char* relativePath)
+static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const char* fullPath, const char* relativePath)
```
jeschu1: Same comment about fullPath not matching signature name

ProjFS.Mac/PrjFSKext/public/PrjFSCommon.h
pmj: When the kext/userlib interface api changes. (In this case, message format.)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Yep, FileOps will always send a path along.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Yeah, I feel `absolutePath` is clearer ("full" also is a possible state for a VFS4G-tracked file) but I've kept `fullPath` where there already was a variable of that name.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: It would be great if we can add a small comment where appropriate in this function
"FileOp operations contain full path.  Vnode operations must look up full path from fsid"

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: @pmj my vote would be to use fullPath for this PR, enter an issue to update all fullPath's to absolute path, and we can address in a separate PR.  Am I correct in saying every place we use 'fullPath' in this file we could use 'absolutePath' instead?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Done in this latest version.

pmj:
Latest push fixes the merge conflict, no functional changes.

Issue #900, opened 2019-03-08
Mac: Create unit/integration tests covering kauth listener callbacks, then refactor


pmj:
Issue #882 and its fix #886 highlighted that none of us are particularly happy with the code for `HandleFileOpOperation` and `HandleVnodeOperation`, as well as the helper functions `TryGetVirtualizationRoot`, `ShouldHandleFileOpEvent`, and friends. The code is hard to read, contains a lot of duplication with subtle differences, and thus is also error-prone.

We should first aim for good test coverage in these areas of code (currently it is essentially zero) and then use them to make structural changes to the code with confidence.

jeschu1:
@pmj Once we merge the referenced PR's what work do you think is left for this issue?

pmj:
Some things that immediately spring to mind, in no particular order, and not necessarily all easily fixable, or worth fixing as part of this issue:

 * The `ProviderMessaging_TrySendRequestAndWaitForResponse` calls in both `HandleFileOpOperation()` and `HandleVnodeOperation()` take up a lot of space visually, and are not in the slightest obvious in terms of what differs between them - not that there's much. There's got to be a better way of doing them, possibly including changing the function's signature a little.
 * Even if we reduce the above, `HandleFileOpOperation()` will be loooong for no good reason and can be split into separate functions for each `KAUTH_FILEOP_*` case.
 * I am not a fan of the `TryGetVirtualizationRoot()` helper function. It does a lot more than its name suggests, and can probably be split up into a more logical arrangement. `HandleVnodeOperation()` also calls it more than once on some code paths, which doesn't seem right at all.
 * The semantics of the many output function parameters on the helper functions in this area are a bit dubious in many cases. (Some are in/out, some are always-out, some are sometimes-out.) I'd like to tighten that up, as output parameters are a common source of undefined behaviour due to uninitialised variables. Maybe some functions would be better returning structs of multiple values rather than using output parameters.
 * This is on the unit testing side, not requiring any changes in the kext's code itself, but the `s_cleanupRootsAfterRequest` thing should probably be fixed sooner rather than later, considering I bumped into its rough edges while rebasing the tests from #985 on it. Which leads me to…
 * Full test coverage for `HandleFileOpOperation()` would be rather good to have before implementing the first few items on this list.
 * I'm not convinced we should be testing the `ShouldHandle…()` helper functions directly. A bunch of those tests would probably be better as tests on the outer `Handle*Operation()`.

jeschu1:
[...]

PR #985, opened 2019-03-30
Mac ProjFS: Two-path provider messages

pmj:
This covers the implementation of the Kext and ProjFS native lib based facilities needed for #337 and #356.

 * [x] #846 Move all relative path calculations to user space. The "to" or "from" path for rename and hardlink notifications can lie outside the provider's virtualisation root, and sending relative paths for these would quickly become awkward, so this is a good time to be making this change.
 * [x] Change the kernel -> user message format and send/receive code to allow an arbitrary number of strings in the data section. Previously, only 0 or 1 path could be sent.
 * [x] Update the Fileop handler for renames and hardlink events to check the source path as well as the destination and look up the source path's provider in case it's different. Send both paths to user space for those events.
 * [x] Cover as much of the new functionality as possible with unit tests.

Additionally, the second commit in the series (266e501) implements a change (allow responses to kernel messages containing arbitrary data) that isn't a prerequisite for the above changes, but are included in the PR anyway because it's needed for PR #910 and the subsequent changes here would have caused annoying merge conflicts.

**Review notes:**

 * Reviewing one commit at a time is recommended.
 * Review the first commit in PR #984 please, as it's a simple, stand-alone improvement so most likely can be merged before the rest of this PR.
 * Please check if the implemented functionality is enough for someone else to take over implementing #337 and #356 on the user space side of things. I think those changes should be done in a separate PR, so this one doesn't stay open for too long and start going stale.
 * The final commit possibly overshoots what's needed for #337/#356. I figured we probably wanted this for the future, but if not it's easy to disable. Specifically, the Fileop handler for `KAUTH_FILEOP_RENAME`/`KAUTH_FILEOP_LINK` now sends the notification message to both the providers responsible for the source and the destination, respectively, assuming they are different. I'm assuming a provider probably wants to know if a file was moved out from it as part of #356, but for hard-link events I believe we were thinking about transitioning hardlinked files to the "full" state, so the notification to the original location could be useful. In both cases, `libPrjFSLib` doesn't actually do anything with the extra information.


ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -553,35 +555,32 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex, PrjFSProvider
     RWLock_ReleaseShared(s_virtualizationRootsLock);
 }
 
-errno_t ActiveProvider_SendMessage(VirtualizationRootHandle rootIndex, const Message message)
+static PrjFSProviderUserClient* GetRetainedProvider(VirtualizationRootHandle rootHandle)
```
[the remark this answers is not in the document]
pmj: I've been following the convention of using "index" when it's known to be a real virtualisation root index (>= 0) or "handle" when it might be one of the "special" handle values. (`RootHandle_None` etc.)

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
```diff
@@ -74,6 +79,9 @@ void ProviderMessaging_HandleKernelMessageResponse(VirtualizationRootHandle prov
                         outstandingMessage->result = responseType;
                         outstandingMessage->receivedResult = true;
                         
+                        memcpy(outstandingMessage->resultDataBuffer, resultData, MIN(resultDataSize, outstandingMessage->resultDataBufferSize));
+                        outstandingMessage->resultDataSize = resultDataSize;
```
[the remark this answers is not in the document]
pmj: Returning `resultDataSize` indicates to the sender of the original message that the reply got truncated. How exactly this is handled is of course a matter of taste.

pmj:
> This opens us up to easily pass more info in the future. I wanted to double check the approach here. Would we want to try to have actual fields for the extra members or is an array the right approach. Downside to the array is you need to know what each spot represents. I'm honestly not sure the best approach just wanted to call it out. @wilbaker

In the case of strings/paths, if you were to stick it all into a struct, you'd have to send the maximum possible size every time, rather than just the length of the actual strings.

I'd say put small (non-string/non-binary-blob) fields in `MessageHeader` and put anything variable-length in the general data area. For now we only need (a variable number of) strings, but in principle you could send other data down there too.

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -47,25 +47,38 @@ struct MessageHeader
...
+    uint16_t            dataSizeBytes;
 };
 
 // Description of a decomposed, in-memory message header plus variable length string field
 struct Message
 {
     const MessageHeader* messageHeader;
-    const char*    path;
+    const void*          data;
+    uint16_t             dataSizeBytes;
```
wilbaker: >>This opens us up to easily pass more info in the future. I wanted to double check the approach here. Would we want to try to have actual fields for the extra members or is an array the right approach. Downside to the array is you need to know what each spot represents. I'm honestly not sure the best approach just wanted to call it out. @wilbaker

>In the case of strings/paths, if you were to stick it all into a struct, you'd have to send the maximum possible size every time, rather than just the length of the actual strings.

>I'd say put small (non-string/non-binary-blob) fields in MessageHeader and put anything variable-length in the general data area. For now we only need (a variable number of) strings, but in principle you could send other data down there too.

Moving this conversation here to make threaded conversation a bit easier.

I like @jeschu1's idea of having explicit fields for each path, and I also agree with @pmj that variable length data should go in the general data area.

Can we combine both of these ideas and go with something like this?

```
struct MessageHeader
{
    ...
    // Sizes of the flexible-length, nul-terminated paths following the message body, including the nul character.	   
    uint16_t            pathSizeBytes;
    uint16_t            fromPathSizeBytes;
```

```
struct Message
{
    const MessageHeader* messageHeader;
    const char*    path;
    const char*    fromPath;
}
```

We would be losing some flexibility, but in exchange we'd gain:

- More explicit validation (e.g. `assert(vnodePath != nullptr` rather than `assert(dataStringCount > 0`)
- Immediate access to all message fields (i.e. no need to parse `data` to split strings)
- Easier to view path values when debugging

@jeschu1, @pmj, what do you think?

jeschu1: @jeschu1 I like the idea of keeping it more explicit for now.  I *think* this should make for a slightly easier PR.  @pmj thoughts on trying approach @wilbaker suggests above for comparison.

Note the first commit in this batch is approved in a separate PR, so the number of changes here will drop.

pmj: Yep, we can do that. I'm going to experiment with putting the sizes in a `struct` inside a `union` to share space with an array, so that we don't have to write special-case code for each field, and can just loop over them with an index when encoding/decoding. (And yes, this is valid C & C++ as long as the types match.) So that would be:

```
struct MessageHeader
{
   …
   union
   {
      uint16_t sizesBytes[2];
      struct
      {
          uint16_t            pathSizeBytes;
          uint16_t            fromPathSizeBytes;
      };
   };
};
```

"Experiment" because I'm not 100% sure yet if the code will be simpler this way or not. If not, I'll go with the "raw" approach as suggested.

pmj: I've reworked the message format change commit as above now. (48ac789) The `union` with the array does make the code less repetitive, obviously this will compound if we go beyond 2 variable-length fields.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -468,67 +477,87 @@ KEXT_STATIC int HandleFileOpOperation(
         
         putCurrentVnode = true;
         
-        VirtualizationRootHandle root = RootHandle_None;
+        VirtualizationRootHandle toRoot = RootHandle_None, fromRoot = RootHandle_None;
         FsidInode vnodeFsidInode = {};
-        int pid;
-        if (!ShouldHandleFileOpEvent(
+        int pid = -1;
```
jeschu1: why are we adding the -1 assignment now?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -637,6 +669,7 @@ KEXT_STATIC int HandleFileOpOperation(
                 currentVnode,
                 vnodeFsidInode,
                 path,
+                nullptr, // TODO
```
jeschu1: what does the TODO mean?

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.cpp
```diff
@@ -29,7 +29,7 @@ const IOExternalMethodDispatch PrjFSProviderUserClient::ProviderUserClientDispat
         {
             .function =                 &PrjFSProviderUserClient::kernelMessageResponse,
             .checkScalarInputCount =    2, // message id, response type
-            .checkStructureInputSize =  0,
+            .checkStructureInputSize =  kIOUCVariableStructureSize, // Arbitrary data buffer, contents depends on kernel request
```
jeschu1: If this change isn't needed for this PR can we please put it up in a separate PR?  I don't mind small PR for any changes, just want to keep this PR to task at hand if possible.

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.cpp
```diff
@@ -212,12 +212,14 @@ IOReturn PrjFSProviderUserClient::kernelMessageResponse(
...
     return static_cast<PrjFSProviderUserClient*>(target)->kernelMessageResponse(
         arguments->scalarInput[0],
-        static_cast<MessageType>(arguments->scalarInput[1]));
+        static_cast<MessageType>(arguments->scalarInput[1]),
+        arguments->structureInput,
+        arguments->structureInputSize);
 }
 
-IOReturn PrjFSProviderUserClient::kernelMessageResponse(uint64_t messageId, MessageType responseType)
+IOReturn PrjFSProviderUserClient::kernelMessageResponse(uint64_t messageId, MessageType responseType, const void* resultData, size_t resultDataSize)
```
jeschu1: Same comment about resultData, is it used in this PR?

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -47,17 +47,38 @@ struct MessageHeader
...
     const MessageHeader* messageHeader;
-    const char*    path;
+    union
+    {
+        struct
+        {
+            const char* path;
+            const char* sourcePath;
+        };
+        const char*     strings[2];
```
jeschu1: Whats the advantch here over:
const char* path;
const char* sourcePath;

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -536,7 +572,11 @@ -(void) testWriteFileHydrated {
             _,
             _,
             _,
-            nullptr));
+            _,
+            nullptr,
+            _,
+            _,
+            _));
```
jeschu1: Sigh, one disadvantage fo tests and having to do maintance like this when the signatures change.  Thanks for keeping i up @pmj.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -898,16 +922,24 @@ static bool ShouldHandleFileOpEvent(
...
+    const char* fromPath, // used only for hard-link & rename events, nullptr otherwise
 
     // Out params:
-    VirtualizationRootHandle* root,
+    VirtualizationRootHandle* vnodeRoot,
+    VirtualizationRootHandle* fromProvider, // when fromPath is non-null, the handle of the provider where the file was originally located is written to here
     FsidInode* vnodeFsidInode,
     int* pid)
 {
+    bool pidIsValid = false;
```
jeschu1: pidIsSet?  I take it pidIsValid means we've made the pid call and want to avoid doing it twice

ProjFS.Mac/PrjFSKext/public/Message.h
jeschu1: @pmj I totally see how that works!  In my head I'm thinking the extra field may be simplier code-wise.  Any chance you could know that up for comparison?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -621,31 +622,50 @@ PrjFS_Result PrjFS_WriteFileContents(
 static Message ParseMessageMemory(const void* messageMemory, uint32_t size)
 {
     const MessageHeader* header = static_cast<const MessageHeader*>(messageMemory);
-    if (header->pathSizeBytes + sizeof(*header) != size)
+    if (size < sizeof(*header))
```
wilbaker: Can we still check that (header + path sizes) should equal size?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -621,31 +622,50 @@ PrjFS_Result PrjFS_WriteFileContents(
...
+            // Path string should fit exactly in reserved memory, with nul terminator in end position
+            assert(strnlen(string, stringSize) == stringSize - 1);
+            messagePosition += stringSize;
+            messageBytesRemain -= stringSize;
+            
+            parsedMessage.strings[i] = string;
+        }
     }
-    return Message { header, path };
+
```
wilbaker: `assert` that `messageBytesRemain` is 0?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -716,30 +742,53 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
         
-        case MessageType_KtoU_NotifyFileCreated:
         case MessageType_KtoU_NotifyFileRenamed:
         case MessageType_KtoU_NotifyDirectoryRenamed:
         case MessageType_KtoU_NotifyFileHardLinkCreated:
         {
-            char fullPath[PrjFSMaxPath];
-            CombinePaths(s_virtualizationRootFullPath.c_str(), request.path, fullPath);
+            const char* relativeFromPath = GetRelativePath(request.sourcePath, s_virtualizationRootFullPath.c_str());
+            cout << "PrjFSLib.HandleKernelRequest: " << (requestHeader->messageType == MessageType_KtoU_NotifyFileHardLinkCreated ? "hard-linked " : "renamed ") << request.sourcePath << " -> " << absolutePath << " (absolute), ";
```
wilbaker: Wrap `cout` (and the `if` blocks below that only `cout`) with a `#if DEBUG`

wilbaker:
@jeschu1, do you think you'll use this PR to make the user-mode changes, or were you planning to do those changes in a follow-up PR?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I can't quite spot the case I was thinking of specifically at the time right now but it was to avoid undefined behaviour when reading the variable and we hadn't written the PID to it.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -553,33 +555,35 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex, PrjFSProvider
...
```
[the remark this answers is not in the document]
pmj: One on which `->retain()` has been called to increment its reference count, i.e. an owned reference.

ProjFS.Mac/PrjFSKext/public/Message.h
pmj: I've just rebased the PR on latest master and (hopefully) fixed the conflicts with the vnode cache changes. I've also pushed a series of commits that remove the union and just use the struct fields. I don't like this code in either of the 2 iterations. I think if we really must remove the union, I'd prefer to stick with the array part of it and use named (enum) array indices instead.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Leftovers from the first iteration of this PR, when we still had the fully flexible arbitrary strings.

ProjFS.Mac/PrjFSKext/PrjFSProviderUserClient.cpp
pmj: I was expecting #910 to go in before this PR, so this would have disappeared in a rebase. I've removed it now and fixed up the merge conflicts, as #910 is bogged down in other merge conflicts and I don't have the bandwidth to keep both that PR and this one up to date right now. I'll revisit #910 as soon as this one is merged.

ProjFS.Mac/PrjFSKext/public/Message.h
pmj: See [my comment here](https://github.com/Microsoft/VFSForGit/pull/985#discussion_r275348890) on how this turned out.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1255,14 +1304,16 @@ static inline PrjFS_NotificationType KUMessageTypeToNotificationType(MessageType
     }
 }
 
-static errno_t SendKernelMessageResponse(uint64_t messageId, MessageType responseType)
+static errno_t SendKernelMessageResponse(uint64_t messageId, MessageType responseType, const void* resultData, size_t resultDataSize)
 {
     const uint64_t inputs[] = { messageId, responseType };
-    IOReturn callResult = IOConnectCallScalarMethod(
+    IOReturn callResult = IOConnectCallMethod(
```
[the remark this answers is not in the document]
pmj: `IOConnectCallScalarMethod` is a shorthand variant of `IOConnectCallMethod` you can use when both input & output struct arguments are zero length.

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -67,6 +67,7 @@ while read line; do
 	 [[ $line != *"FindOrDetectRootAtVnode"* ]] &&                   #SHOULD ADD COVERAGE
 	 [[ $line != *"FindUnusedIndexOrGrow_Locked"* ]] &&              #SHOULD ADD COVERAGE
 	 [[ $line != *"FindRootAtVnode_Locked"* ]] &&                    #SHOULD ADD COVERAGE
+	 [[ $line != *"DirectoryContainsPath"* ]] &&                     #TODO: Why does this get reported as not having coverage when it does?
```
jeschu1: You probably know this already but after generating a code coverage run, xcode has a really nice display for seeing sections it believes are not covered.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -482,68 +492,89 @@ KEXT_STATIC int HandleFileOpOperation(
...
-                &root,
+                fromPath,
+                &toRoot,
+                &fromRoot,
                 &vnodeFsidInode,
-                &pid))
+                &pid);
+        bool messageSourceProvider = VirtualizationRoot_IsValidRootHandle(fromRoot);
+        
+        if (!messageDestinationProvider && !messageSourceProvider)
```
jeschu1: To be clear if ShouldHandleFileOpEvent is false, but fromRoot is a valid root handle we continue?
We have a number of checks in ShouldHandleFileOpEvent (ex check the pid to avoid deadlocks).  Are we sure this is a safe change?

ProjFS.Mac/Scripts/Build.sh
pmj: The coverage counter column in the code view? Yes, this was showing nonzero on all branches, and I just wanted to run the functional tests rather than leaving it as I didn’t have time to investigate the discrepancy last night.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.hpp
```diff
@@ -30,6 +30,9 @@ VirtualizationRootHandle VirtualizationRoot_FindForVnode(
     vnode_t _Nonnull vnode,
     vfs_context_t _Nonnull context);
 
+VirtualizationRootHandle ActiveProvider_FindForPath(
```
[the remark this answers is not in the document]
pmj: It will currently only find *active* roots. So even if the “from” path is in an offline root, it will return a “none” handle.

We can in principle re-instate tracking the path of offline roots but this will be even less reliable in the face of moves/renames than for active providers.

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -47,17 +47,26 @@ struct MessageHeader
...
 struct Message
 {
     const MessageHeader* messageHeader;
-    const char*    path;
+
+    const char* path;
+    const char* fromPath;
 };
 
+#if defined(KERNEL) || defined(KEXT_UNIT_TESTING)
```
jeschu1: Can you remind of why we need this defined check?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -748,17 +765,47 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
             break;
         }
         
-        case MessageType_KtoU_NotifyFileCreated:
         case MessageType_KtoU_NotifyFileRenamed:
         case MessageType_KtoU_NotifyDirectoryRenamed:
         case MessageType_KtoU_NotifyFileHardLinkCreated:
         {
+            const char* relativeFromPath = GetRelativePath(request.fromPath, s_virtualizationRootFullPath.c_str());
```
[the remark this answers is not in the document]
jeschu1: You can move it to the `#if DEBUG`.  We can easily move it out of there when we need it.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -553,7 +555,7 @@ void ActiveProvider_Disconnect(VirtualizationRootHandle rootIndex, PrjFSProvider
     RWLock_ReleaseShared(s_virtualizationRootsLock);
 }
 
-errno_t ActiveProvider_SendMessage(VirtualizationRootHandle rootIndex, const Message message)
+errno_t ActiveProvider_SendMessage(VirtualizationRootHandle rootIndex, const Message& message)
```
jeschu1: Why did this change to a pass by reference?

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
```diff
@@ -33,7 +35,57 @@ void Message_Init(
...
+}
+
+uint32_t Message_Encode(void* buffer, const uint32_t bufferSize, const Message& message)
+{
+    uint8_t* bufferPosition = static_cast<uint8_t*>(buffer);
+    uint32_t bufferBytesRemain = bufferSize;
+    
+    assert(bufferSize >= sizeof(*message.messageHeader));
+    memcpy(bufferPosition, message.messageHeader, sizeof(*message.messageHeader));
+    bufferPosition +=    sizeof(*message.messageHeader);
```
jeschu1: nit: spacing

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
```diff
@@ -33,7 +35,57 @@ void Message_Init(
...
+    bufferBytesRemain -= sizeof(*message.messageHeader);
+    
+    {
+        uint16_t stringSize = message.messageHeader->pathSizeBytes;
+        if (stringSize > 0)
+        {
+            assert(bufferSize >= stringSize);
```
jeschu1: bufferBytesRemain for here and below?

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
```diff
@@ -33,7 +35,57 @@ void Message_Init(
...
+{
+    return sizeof(*message.messageHeader) + message.messageHeader->pathSizeBytes + message.messageHeader->fromPathSizeBytes;
+}
+
+uint32_t Message_Encode(void* buffer, const uint32_t bufferSize, const Message& message)
+{
+    uint8_t* bufferPosition = static_cast<uint8_t*>(buffer);
+    uint32_t bufferBytesRemain = bufferSize;
+    
+    assert(bufferSize >= sizeof(*message.messageHeader));
```
jeschu1: We do a number of asserts for sizing here, which we didn't previously.  

I can see the value, in case an sizing error creeps in, but it does become more verbose.   @wilbaker

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
```diff
@@ -33,7 +35,57 @@ void Message_Init(
...
+    memcpy(bufferPosition, message.messageHeader, sizeof(*message.messageHeader));
+    bufferPosition +=    sizeof(*message.messageHeader);
+    bufferBytesRemain -= sizeof(*message.messageHeader);
+    
+    {
```
jeschu1: For my own info, why is this done in a set of {'s

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -482,68 +492,89 @@ KEXT_STATIC int HandleFileOpOperation(
...
                     vnodeFsidInode,
                     newPath,
+                    fromPath,
                     pid,
                     procname,
                     &kauthResult,
                     &kauthError))
             {
-                goto CleanupAndReturn;
+                KextLog_Error("HandleFileOpOperation: Request %u to destination provider %d failed, kauthResult = %u, kauthError = %u",
```
jeschu1: should this also CleanupAndReturn?  Same question below

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -482,68 +492,89 @@ KEXT_STATIC int HandleFileOpOperation(
...
-
+        }
+        else
+        {
+            perfCounter = PrjFSPerfCounter_FileOp_HardLinkCreated;
+            
+            messageType = MessageType_KtoU_NotifyFileHardLinkCreated;
+        }
+        
+        if (messageDestinationProvider)
```
jeschu1: It looks like we don't actually send messages if !messageDestinationProvider.  Is there a chance a message will be sent later on for this case?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -573,13 +575,10 @@ errno_t ActiveProvider_SendMessage(VirtualizationRootHandle rootIndex, const Mes
     
     if (nullptr != userClient)
     {
-        uint32_t messageSize = sizeof(*message.messageHeader) + message.messageHeader->pathSizeBytes;
```
jeschu1: As you've nicely done, seperating the size / encoding into separate functions requires the need to do several checks for sizing (the functions are independent now).

The advantage would be we can unit test.  I *think* this is not being picked up by code coverage because this class isn't part of the unit tests project.  I.e. we mock out sending the request.

Would it make sense to add the new functions to unit tests?  Does anyone else have feelings on the approach? @wilbaker

jeschu1:
@pmj thanks for iterating on this.  Your changes are making a lot of sense!

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
jeschu1: Spoke with @wilbaker offline and we like your approach better than what was there previously.  It would be great if we can add tests for the new functions.

ProjFS.Mac/PrjFSKext/public/Message.h
pmj: We don't define `Message_Init()` in user space, and this file is shared between user space and kernel. We should actually probably just move the kernel-only function prototypes to a kernel-only file instead.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: As `Message` becomes larger and larger, it'll be implicitly passed by reference anyway, but it'll be a reference to a copy. No need to copy, as we don't modify it anyway, so explicit readonly reference is in theory slightly more efficient.

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
pmj: I guess I find it helpful to make it extra clear that we're incrementing the position and decrementing the remaining byte count by the same amount by aligning the amounts.

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
pmj: Re: all the questions/comments about this function - the code of this function is still in flux, this is a prototyping iteration on the original loop-based approach. I don't like the current approach with the repeated code, so I'm going to find a better solution to the overall structure of the code before picking apart the style.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I don't understand the question, can you please be more specific about the case you're thinking about? The code sends a message to the source provider if `messageSourceProvider` is true a few lines further down, i.e. we send a message to either the source or the destination provider. We don't want to message the destination provider if it is e.g. offline or the destination is outside a root.

Whether this is the behaviour we want is another interesting question of course, but as far as I can tell this is the original inspiration for the change, so I think the answer is yes.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: I think we should debate unit testing once we have a fundamental approach we can agree on. But yeah, unit testing serialisation/deserialisation code is definitely a good idea I think.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've looked over this now, and I'm not sure what the specific trigger was, possibly a complaint by the static analysis step, or possibly a debug output statement that printed `pid` in cases where it might not have been set.

In any case, passing around pointers or references to variables that haven't been initialised or may not have been initialised is something I consider something of a code smell, it's just really hard to reason about and explicitly initialising it costs almost nothing.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -911,45 +952,80 @@ static bool ShouldHandleFileOpEvent(
...
         }
     }
+    
+    if (shouldHandle || shouldHandleFromProvider)
+    {
+        *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context, true /* the inode is used for getting the path in the provider, so use linkid */);
+    }
 
-    *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context, true /* the inode is used for getting the path in the provider, so use linkid */);
+    return shouldHandle;
```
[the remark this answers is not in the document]
pmj: If you check the relevant rename/link call site for this in `HandleFileOpOperation`, you'll find that the return value is treated as wether "we should message the destination provider":

```
bool messageDestinationProvider = ShouldHandleFileOpEvent(
```

whereas "we should message the source provider" is determined by whether we got a valid handle for the "from" root.

```
bool messageSourceProvider = VirtualizationRoot_IsValidRootHandle(fromRoot);
```

We could certainly return `shouldHandle || shouldHandleFromProvider` but then we need to additionally convey whether or not to message the destination provider.

I think the calling semantics of `ShouldHandleFileOpEvent` are awful (hence #900) and this PR certainly makes things worse rather than better. Fixing them is out of scope here, and I'm not sure what would be a better approach as a stopgap. Returning a struct of 2 booleans is going to make things worse for all the 1-root call sites, and your proposed return value would imply we need to either change the semantics of `VirtualizationRootHandle* vnodeRoot` so that a `None` handle is returned if the destination provider *shouldn't* be messaged, (I guess this should be safe, and our now-excellent unit test coverage should tell us if not?) or we need an extra output argument.

What's your preferred option?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -911,45 +952,80 @@ static bool ShouldHandleFileOpEvent(
...
         {
-            perfTracer->IncrementCount(PrjFSPerfCounter_FileOp_ShouldHandle_OriginatedByProvider);
-            return false;
+            perfTracer->IncrementCount(PrjFSPerfCounter_FileOp_ShouldHandle_NoRootFound);
         }
     }
+    
+    if (shouldHandle || shouldHandleFromProvider)
+    {
+        *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context, true /* the inode is used for getting the path in the provider, so use linkid */);
```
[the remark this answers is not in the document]
pmj: The inode and fsid are sent along with the rename message to the source provider in the [in root] -> [out of root] rename/hardlink case. I don't think it's strictly necessary as we're guaranteed correct paths. I think making sure that fsid+inode are always correct when the function returns with some kind of call to messaging action is reasonable though. Perhaps we can find a better solution as part of #900 however.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -911,45 +952,80 @@ static bool ShouldHandleFileOpEvent(
...
+    }
 
-    *vnodeFsidInode = Vnode_GetFsidAndInode(vnode, context, true /* the inode is used for getting the path in the provider, so use linkid */);
+    return shouldHandle;
+}
 
+static bool ShouldMessageProviderWithHandle(PerfTracer* perfTracer, VirtualizationRootHandle root, pid_t pid)
```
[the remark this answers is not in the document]
pmj: This is a case of habit following Apple's naming conventions that are used especially in Objective-C method names but also in some of their C++ APIs including IOKit. Basically, the "with" part indicates the "decisive" arguments for that function. I can remove it.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Hmm, I guess in the case where a provider process is moving (or hardlinking) a file from one virtualisation root to another, I'm not sure we should be messaging the provider for the other virtualisation root. Thoughts? I mean it's not exactly a likely scenario. And to be clear, the code as it stands should never self-message (i.e. sending a message to the provider that caused the event.)

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
pmj: Good catch!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: As far as I can tell, the only reason `ProviderMessaging_TrySendRequestAndWaitForResponse` would be returning false here is that the provider has gone offline since we checked that it was online, or that the provider itself returned an error message. Neither seems a good reason *not* to send a message to the source provider if that's what we were originally going to do.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Am I correct that the case we're trying to catch is:
A is in the Virtualization Root
B is out of the Virtualization Root

`ln A B` is called.  Now when B is changed A would be updated too, but without any reference to A.  Hence we want to make sure A gets on our watch (modified paths) list.

If so I would think ShouldHandleFileOpEvent should cover it all.  

Or are we talking about a case where
A is in 1 Virtualization Root
B is in a 2nd Virtualization Root

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: We generally call CleanupAndReturn which does a put if TrySendrequestAndWaitForResponse fails.  

My question is why would we not need it in this case.

If I'm missing something please let me know!

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
jeschu1: i've never held up a PR over spacing, so I wouldn't do that here.
To me it looks more like off spacing than pointing out the +/-.
Just 1 person's opinion :-)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: There's no `return` prior to the `CleanupAndReturn:`, so we call `vnode_put(currentVnode)` if necessary in all call cases, so that should not have any bearing on the decision. So as far as I can tell the only difference with or without the `goto` is that without, we message the "from" provider even if messaging the "to" provider didn't succeed for whatever reason, assuming both must be messaged. Failure to message "to" isn't great, obviously, but I don't see how then *also* dropping the message to "from" would be helpful. (We can't fail the kauth call or anything like that, as FILEOP calls must always succeed.)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Looks like we're in agreement on what *should* happen, and that's certainly what the code so far should be doing. (If you've spotted a discrepancy, please point it out.)

The `ShouldHandle…` functions are suboptimal in a lot of ways to say the least. My original thought was that fixing them in this PR would be too much. I'm sure I can be persuaded otherwise though. :-)

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -32,6 +32,14 @@ typedef enum
     
 } MessageType;
 
+enum MessagePathField
+{
+    MessagePath_Target = 0,
+    MessagePath_From,
+    
+    MessagePath_Count,
+};
```
pmj: So this is the second option for removing the `union` from the original approach - stick with the array and name each element in the array using this `enum`.

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -47,25 +55,19 @@ struct MessageHeader
     int32_t             pid;
     char                procname[MAXCOMLEN + 1];
 
-    // Size of the flexible-length, nul-terminated path following the message body, including the nul character.
-    uint16_t            pathSizeBytes;
+    // Sizes of the flexible-length, nul-terminated paths following the message body.
+    // Sizes include the nul characters but can be 0 to indicate total absence.
+    uint16_t            pathSizesBytes[MessagePath_Count];
```
pmj: This is now an array instead of 2 scalars or a union of an array and a struct with 2 scalars.

ProjFS.Mac/PrjFSKext/public/Message.h
```diff
@@ -47,25 +55,19 @@ struct MessageHeader
...
 };
 
 // Description of a decomposed, in-memory message header plus variable length string field
 struct Message
 {
     const MessageHeader* messageHeader;
-    const char*    path;
+    const char* paths[MessagePath_Count];
```
pmj: Analogous to the `MessageHeader` struct.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -632,22 +632,36 @@ PrjFS_Result PrjFS_WriteFileContents(
...
     }
-            
-    const char* path = nullptr;
-    if (header->pathSizeBytes > 0)
+    
+    Message parsedMessage = { header };
+    
+    const char* messagePosition = static_cast<const char*>(messageMemory) + sizeof(*header);
+    uint32_t messageBytesRemain = size - sizeof(*header);
+    for (unsigned i = 0; i < extent<decltype(parsedMessage.paths)>::value; ++i)
```
pmj: We can still loop over the paths with the array.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -664,7 +678,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     // whereas messages originating in the kext's vnode handler will only fill
     // the fsid/inode, so we need to look up the path below.
     char pathBuffer[PrjFSMaxPath];
-    if (request.path == nullptr)
+    if (request.paths[MessagePath_Target] == nullptr)
```
pmj: Each element in the array is named so the code is pretty readable.

ProjFS.Mac/PrjFSKext/Message_Kernel.hpp
```diff
@@ -0,0 +1,17 @@
...
+    MessageHeader* header,
+    uint64_t messageId,
+    MessageType messageType,
+    const FsidInode& fsidInode,
+    int32_t pid,
+    const char* procname,
+    const char* path,
+    const char* fromPath);
+
+uint32_t Message_Encode(void* buffer, uint32_t bufferSize, const Message& message);
```
pmj: I've now moved the kernel-only function prototypes into a kernel-only header file.

ProjFS.Mac/PrjFSKext/Message_Shared.cpp
```diff
@@ -0,0 +1,18 @@
...
+uint32_t Message_EncodedSize(const MessageHeader* messageHeader)
+{
+    uint32_t size = sizeof(*messageHeader);
+    for (unsigned i = 0; i < MessagePath_Count; ++i)
+    {
+        bool overflow = __builtin_uadd_overflow(messageHeader->pathSizesBytes[i], size, &size);
+        assert(!overflow); // small constant plus small number of uint16_t should never overflow a uint32_t
+    }
+    return size;
+}
```
pmj: Also, this function can be (and is) called from both user and kernel mode code, so I've put it in a separate .cpp file.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Looking into the simple 2-way split suggestion after discussing it out of band, I don't see a super simple way to do this:

 * We don't want to send a message to either the target vnode's provider or the `fromPath`'s provider if the `ShouldIgnoreVnodeType()` or `VirtualizationRoot_VnodeIsOnAllowedFilesystem()` checks fail, and the original version of `ShouldHandleFileOpEvent()` made no distinction between those checks failing versus the vnode not being in a virtualisation root or the PID matching the provider of the root it's in. So we'd either need to do those checks twice or change the API of the function somehow after all, in which case we may as well stick with what we've got and refactor it properly later.

pmj:
I've now added tests that call `HandleFileOpOperation()` to simulate the different rename scenarios.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @wilbaker and I talked offline about how we can make ShouldHandleFileOpEvent easiest to understand / least complex.  As you've pointed out there is so much we want to refactor about ShouldHandleFileOpEvent.

What we were thinking would be do something like this:
```
   bool shouldHandleTo = ShouldHandleFileOpEvent(other params…, vnode, nullptr, other params..)
   bool shouldHandleFrom = ShouldHandleFileOpEvent(other params…, vnode, frompath, other params..)
```
Where the only difference in `ShouldHandleFileOpEvent` versus what was originally there would be an `if `to choose between the vnodeCache and ActiveProvider_FindForPath for finding the virtualization root.    
This does mean we'd do some duplicate checks, but I'd happily take that if it's easier to follow.

I started that approach [here](https://github.com/Microsoft/VFSForGit/pull/1080/commits/1411818767122fafc46e1b7c90109061d6a49901).  Let me know what you think.  Obviously this commit is a big WIP.


I *think* and correct me if I'm wrong please, what we have now works because we say:
```
        bool messageDestinationProvider = ShouldHandleFileOpEvent( 
                &perfTracer,
		context, 
                currentVnode, 
                action, 
                isDirectory, 
                &root, 
                fromPath, 
                &toRoot, 
                &fromRoot, 
                &vnodeFsidInode, 
                &pid); 
bool messageSourceProvider = VirtualizationRoot_IsValidRootHandle(fromRoot);
```
If we have an invalid vnode, fromRoot would be null, and hence the `VirtualizationRoot_IsValidRootHandle` would return false.  I think that works just trying to think of ways to make the checks here clearer.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Nevermind, pulling the branch it all makes sense now.

pmj:
In this latest code push I've:
 * updated the changes so that only hardlink events take into account the "from" path and send both paths to the provider(s), not renames ([as per this comment](https://github.com/microsoft/VFSForGit/pull/1103#issuecomment-489747960)).
 * made the requested changes to `ShouldHandleFileOpEvent()`
 * added unit tests that call `HandleFileOpOperation()` in a way that exercises all the new hardlink message functionality.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -756,7 +844,7 @@ KEXT_STATIC bool ShouldHandleVnodeOpEvent(
         }
     }
 
-    *pid = GetPid(context);
+    *pid = vfs_context_pid(context);
```
jeschu1: I think this is just cleanup.  Can you confirm?

ProjFS.Mac/PrjFSKext/public/Message.h
jeschu1: I personally like the explicit from/to, but I know you prefer it the other way.  Can we get a third opinion from @wilbaker?  Both ways certainly work :-), I prefer explicit fields for 2 variables rather than an array.  I think it's safer.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes, I noticed we were pointlessly replicating the kernel’s built-in `vfs_context_pid()` with our `GetPid` helper.

jeschu1: Let's definitely take the change here, but if we can make separate small PRs for stuff like this it just makes reviewing easier.  And if we're slow at reviewing super small PRs with cleanup please do yell at us / me :-).

jeschu1:
I really like where this PR is at.  @pmj thanks for taking the time to make a complicated problem easy to understand.  I also very much appreciate you going back and making the commits relative to the work.

What I'm thinking is the next steps are:
1. Let's get (1117)[https://github.com/microsoft/VFSForGit/pull/1117] merged (hopefully today with @wilbaker's help)
2. Then let's give this a final review and merge.
I forsee only small nits from my side.

wilbaker:
@pmj apologies for falling behind on this PR, it (and #1117) are next on my list.

pmj:
> @pmj apologies for falling behind on this PR, it (and #1117) are next on my list.

No worries, iterating on it has kept me busy anyway. #1117 would be the one to review first as #985 now builds on top of it.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -515,6 +519,7 @@ KEXT_STATIC int HandleFileOpOperation(
                     currentVnode,
                     vnodeFsidInode,
                     newPath,
+                    nullptr, // TODO
```
jeschu1: nit: is this a TODO?

ProjFS.Mac/PrjFSKextTests/MessageTests.mm
```diff
@@ -0,0 +1,44 @@
...
+    srand(0);
+    for (unsigned i = 0; i < 10000; ++i)
+    {
+        uint16_t targetSize = testHeader.pathSizesBytes[MessagePath_Target] = rand() & UINT16_MAX;
+        uint16_t fromSize = testHeader.pathSizesBytes[MessagePath_From] = rand() & UINT16_MAX;
+        uint32_t size = Message_EncodedSize(&testHeader);
+        XCTAssertGreaterThanOrEqual(size, sizeof(testHeader));
+        XCTAssertLessThanOrEqual(size, maxSize);
+        XCTAssertGreaterThan(size, targetSize);
+        XCTAssertGreaterThan(size, fromSize);
```
wilbaker: Can we also have an `XCTAssertEqual` for `size` (or perhaps in place of the other checks)?

```
XCTAssertEqual(size, sizeof(testHeader) + testHeader.pathSizesBytes[MessagePath_Target] + testHeader.pathSizesBytes[MessagePath_From]);
```

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -523,24 +528,104 @@ KEXT_STATIC int HandleFileOpOperation(
...
-                goto CleanupAndReturn;
+                int kauthResult;
+                int kauthError = 0;
+                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                        targetRoot,
+                        MessageType_KtoU_NotifyFileHardLinkCreated,
+                        currentVnode,
+                        vnodeFsidInode,
+                        newPath,
+                        fromPath,
```
jeschu1: I *think* we only want to send fromPath here if the providers match.  Otherwise we're sending a message for another Virtualization Root.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -598,3 +598,51 @@ bool VirtualizationRoot_VnodeIsOnAllowedFilesystem(vnode_t vnode)
         0 == strncmp("hfs", vfsStat->f_fstypename, sizeof(vfsStat->f_fstypename))
         || 0 == strncmp("apfs", vfsStat->f_fstypename, sizeof(vfsStat->f_fstypename));
 }
+
+KEXT_STATIC bool DirectoryContainsPath(const char* directoryPath, const char* path)
```
jeschu1: PathContainsDirectory?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -632,22 +632,36 @@ PrjFS_Result PrjFS_WriteFileContents(
 static Message ParseMessageMemory(const void* messageMemory, uint32_t size)
 {
     const MessageHeader* header = static_cast<const MessageHeader*>(messageMemory);
-    if (header->pathSizeBytes + sizeof(*header) != size)
+    if (size != Message_EncodedSize(header))
     {
-        fprintf(stderr, "ParseMessageMemory: invariant failed, bad message? PathSizeBytes = %u, message size = %u, expecting %zu\n",
-            header->pathSizeBytes, size, header->pathSizeBytes + sizeof(*header));
+        fprintf(stderr, "ParseMessageMemory: invariant failed, bad message? message size = %u, expecting minimum of %zu\n",
+            size, sizeof(*header));
```
wilbaker: Can we pass this message to `s_callbacks.LogError` instead of (or in addition to) printing to the console?

I suspect `s_callbacks.LogError` did not exist when the old code was written.

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
```diff
@@ -33,7 +35,47 @@ void Message_Init(
...
+{
+    uint8_t* bufferPosition = static_cast<uint8_t*>(buffer);
+    uint32_t bufferBytesRemain = bufferSize;
+    
+    assert(bufferSize >= sizeof(*message.messageHeader));
+    memcpy(bufferPosition, message.messageHeader, sizeof(*message.messageHeader));
+    bufferPosition +=    sizeof(*message.messageHeader);
+    bufferBytesRemain -= sizeof(*message.messageHeader);
+    
+    for (unsigned i = 0; i < 2; ++i)
```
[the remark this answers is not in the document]
wilbaker: Or `extent<decltype(...)>::value` (but I'm not sure we can use that in the kernel)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -748,10 +762,42 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
+                cout << "from this root (relative path " << relativeFromPath << ") ";
+            }
+            if (relativePath != nullptr)
+            {
+                cout << "into this root (relative path " << relativePath << ")";
+            }
+            cout << endl;
+#endif
+            
+            bool isDirectory = requestHeader->messageType == MessageType_KtoU_NotifyDirectoryRenamed;
```
wilbaker: There is no need to check `requestHeader->messageType == MessageType_KtoU_NotifyDirectoryRenamed` because in this case we're handing `MessageType_KtoU_NotifyFileHardLinkCreated`

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: What's needed for this TODO?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -523,24 +528,104 @@ KEXT_STATIC int HandleFileOpOperation(
...
-                    pid,
-                    procname,
-                    &kauthResult,
-                    &kauthError))
+        // Don't expect named stream here as they can't be directly hardlinked or renamed, only the main fork can
+        assert(!vnode_isnamedstream(currentVnode));
+        
+        putCurrentVnode = true;
+        
+        bool isDirectory = (0 != vnode_isdir(currentVnode));
```
wilbaker: Is it possible to hard-link directories?

If so, @jeschu1 we might need more reactionary work in the provider to do the right thing.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -523,24 +528,104 @@ KEXT_STATIC int HandleFileOpOperation(
...
+            PerfSample sample(&perfTracer, PrjFSPerfCounter_FileOp_HardLinkCreated);
+
+            if (messageTargetProvider)
             {
-                goto CleanupAndReturn;
+                int kauthResult;
+                int kauthError = 0;
+                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                        targetRoot,
+                        MessageType_KtoU_NotifyFileHardLinkCreated,
```
wilbaker: Related to above question, if this is for a directory do we need a new `MessageType_KtoU_NotifyDirectoryHardLinkCreated` message?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: Or maybe `PathInsideDirectory`?

ProjFS.Mac/PrjFSKextTests/MessageTests.mm
pmj: The question then becomes what exactly we are testing. The equals test is something of a tautological `XCTAssert(a + b == a + b);` kind of test where the implementation under test is the same as the test itself, which doesn't really say much.

What I'm trying to do with the inequality checks is testing various properties of the result that must not be violated, without actually assuming a specific implementation.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: > Is it possible to hard-link directories?

Yes, and macOS's Time Machine backup system actively uses this feature, so it's definitely something our kext needs to be able to handle gracefully.

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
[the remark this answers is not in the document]
pmj: `std::extent` is part of the C++ standard library, which is not available when building kexts. But yes, we can implement our own version of it; I so far have opted not to here because templated code in previous PRs has met stiff resistance. Any suggestions for naming the file where we'd gather such utility templates?

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
```diff
@@ -166,7 +171,7 @@
 		4391F89021E42AC40008103C /* kernel-header-wrappers */ = {isa = PBXFileReference; lastKnownFileType = folder; path = "kernel-header-wrappers"; sourceTree = "<group>"; };
 		4391F89121E42AC40008103C /* KextLog.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = KextLog.cpp; sourceTree = "<group>"; };
 		4391F89221E42AC40008103C /* KextLog.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = KextLog.hpp; sourceTree = "<group>"; };
-		4391F89321E42AC40008103C /* KauthHandler.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = KauthHandler.cpp; sourceTree = "<group>"; };
+		4391F89321E42AC40008103C /* KauthHandler.cpp */ = {isa = PBXFileReference; fileEncoding = 4; indentWidth = 4; lastKnownFileType = sourcecode.cpp.cpp; path = KauthHandler.cpp; sourceTree = "<group>"; usesTabs = 0; };
```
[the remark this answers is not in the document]
pmj: That's probably because I seem to run into an Xcode indentation bug a lot and actively have to reset the indentation settings whenever I change git branch. I'll remove this change.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @pmj let me know if @wilbaker's comment here makes sense.  i think we want a separation notification message then for directory renames (similar to renames).
I can pick this work up too if you want, not sure if crosses too much into non-kext side work.

jeschu1: @wilbaker  this does make me wonder if hard links are allowed on windows directories

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: As long as we don't save a `vfs_context_t` for use outside the kauth listener callback, and we likewise don't use any previously-saved VFS contexts in the callback, it doesn't really matter either way in practice. `proc_self` is implicit state of the execution context, while the `vfs_context_t` is an explicit reference to it, unless it's `nullptr`, in which case most functions fall back to `proc_self` etc.

One way in which the VFS context may matter is when performing I/O (including metadata I/O) on vnodes, because the context carries the permissions of the process it describes. So if the userspace process is for whatever reason not permitted to access a file in the way the kext tries to do, the kext will also be denied permission if it uses the standard context. I assume we haven't run into any edge cases where this matters yet, but in theory, we should look out for this, especially when populating the vnode cache - e.g. if we were denied permission to read the xattr of a directory to check if it's a root, we shouldn't then assume that it's not a root and save that state. The other option is to perform I/O with the kernel's root VFS context, which shouldn't have the permissions problem, but then you have to be careful about opening up information leakage security bugs.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: @jeschu1 Windows does not support directory hardlinks:

https://docs.microsoft.com/en-us/windows/desktop/FileIO/hard-links-and-junctions

I think it makes sense to have a separate PR for directory hardlinks (how we handle them might require a larger design discussion).  

I filed #1152 for directly hardlink support.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: Filed #1152 for this.  

I think we'll want separate messages for File/Directory hardlink messages to user-mode (like we have for renames) but I think we can scope this PR to just the file hardlink message.

ProjFS.Mac/PrjFSKextTests/MessageTests.mm
wilbaker: > What I'm trying to do with the inequality checks is testing various properties of the result that must not be violated, without actually assuming a specific implementation.

In that case would it be fair to also check:

```
XCTAssertGreaterThanOrEqual(size, sizeof(testHeader) + targetSize + fromSize);
```

That would ensure the encoded size is allowing enough room for the header and the paths.

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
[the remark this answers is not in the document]
wilbaker: I actually added this helper at one point during the vnode cache implementation and didn't end up needing it and it was removed from the PR.  

If you're interested you could pull the implementation and unit tests back in:

https://github.com/microsoft/VFSForGit/blob/846c64df880160abb6cf8d2c85a590ba1c519db1/ProjFS.Mac/PrjFSKext/ArrayUtilities.hpp

https://github.com/microsoft/VFSForGit/blob/846c64df880160abb6cf8d2c85a590ba1c519db1/ProjFS.Mac/PrjFSKextTests/ArrayUtilitiesTests.mm

I wasn't thrilled with the name `ArrayUtilities` but thought it was preferable to something even more generic (e.g. UtilityFunctions.h).  If we add more helper templates we could split them up by what they're used with.

What do you think?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @pmj let's leave everything as here regarding directories per @wilbaker  suggestion.  We can create a new message type in a seperate PR.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: > Currently the provider will fail to obtain a root-relative path for it, so the relativePath variable will be a nullptr and behave as expected

Yes, but we will print an error https://github.com/microsoft/VFSForGit/blob/57db0f04fa75be0ebb5b74a0d08ccada7c1bc9de/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L1428

If we just null it out here no error

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Thanks, missed that leftover when I removed the "from" provider messaging for on rename.

ProjFS.Mac/PrjFSKext/Message_Kernel.cpp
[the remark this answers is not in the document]
pmj: I had a `KextCppUtilities.hpp` way back on an early iteration of #500 which had some bits & pieces like this in it.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -357,6 +357,8 @@ - (void)testShouldHandleFileOpEvent {
     shared_ptr<mount> testMountNone = mount::Create("none", fsid_t{}, 0);
     shared_ptr<vnode> testVnodeNone = vnode::Create(testMountNone, "/none");
     shared_ptr<vnode> testVnodeUnsupportedType = vnode::Create(testMount, "/foo", VNON);
+    std::string fromFilePath = repoPath + "/fromFile.txt";
+    std::string fromDirectoryPath = repoPath + "/fromDirectory";
```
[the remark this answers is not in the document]
pmj: Nope, those have moved to `HandleFileOpOperationTests.mm` in later iterations instead and have since been further modified; I've removed this leftover change.

pmj:
This latest push should address most or all of the outstanding feedback, please re-review!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @pmj, this is the last comment i'm waiting on before approving.  Let me know what you think!

pmj: I've just pushed a new version that doesn't send any out-of-root paths with hardlink messages, whether that be "from" or "to" paths/vnodes - we'll see if it passes the functional tests or if that change will require some further work.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -524,24 +529,105 @@ KEXT_STATIC int HandleFileOpOperation(
...
+                        currentVnode,
+                        vnodeFsidInode,
+                        newPath,
+                        (messageFromProvider && targetRoot == fromRoot) ? fromPath : nullptr, // Only send "from" path if in the same root.
```
jeschu1: Just an fyi, things may fail later on when "null" is sent down to the mount.  If your tests are passing I wouldn't worry about this.  I think we can discuss in a later PR if we should switch to "" here or just before sending the message in the lib layer.

pmj: There's a null check in PrjFSLib, so we never send a `PrjFS_NotificationType_HardLinkCreated` with a `nullptr` to `GVFS.Mount` itself.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -524,24 +529,105 @@ KEXT_STATIC int HandleFileOpOperation(
...
+        const char* fromPath = reinterpret_cast<const char*>(arg0);
+        const char* newPath = reinterpret_cast<const char*>(arg1);
+        
+        // TODO(Mac): We need to handle failures to lookup the vnode.  If we fail to lookup the vnode
+        // it's possible that we'll miss notifications
+        errno_t toErr = vnode_lookup(newPath, 0 /* flags */, &currentVnode, context);
+        if (0 != toErr)
         {
-            PerfSample hardLinkSample(&perfTracer, PrjFSPerfCounter_FileOp_HardLinkCreated);
+            KextLog_Error("HandleFileOpOperation: vnode_lookup failed, errno %d for path '%s'", toErr, newPath);
```
wilbaker: Now that this can be logged in two spots (here and line 476) it would be helpful to include the `action` that the kext was processing in the error message.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -593,15 +680,16 @@ KEXT_STATIC int HandleFileOpOperation(
...
-                                              pid,
-                                              procname,
-                                              &kauthResult,
-                                              &kauthError))
+                root,
+                MessageType_KtoU_NotifyFileCreated,
+                currentVnode,
+                vnodeFsidInode,
+                path,
+                nullptr,
```
wilbaker: Minor: Add comment next to `nullptr` describing its parameter

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -652,6 +741,7 @@ KEXT_STATIC int HandleFileOpOperation(
                 currentVnode,
                 vnodeFsidInode,
                 path,
+                nullptr,
```
wilbaker: Add comment here too

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -524,24 +529,105 @@ KEXT_STATIC int HandleFileOpOperation(
...
-                    &kauthResult,
-                    &kauthError))
+        // Don't expect named stream here as they can't be directly hardlinked or renamed, only the main fork can
+        assert(!vnode_isnamedstream(currentVnode));
+        
+        putCurrentVnode = true;
+        
+        bool isDirectory = (0 != vnode_isdir(currentVnode));
+        
+        VirtualizationRootHandle targetRoot, fromRoot;
```
wilbaker: Only run the rest of the `KAUTH_FILEOP_LINK` code if `isDirectory` is false? 

That would guarantee `MessageType_KtoU_NotifyFileHardLinkCreated` is only being delivered for files, and the `TODO: may yet be a directory` comment could be removed from PrjFSLib.cpp.

pmj: I've also added a `KextLog_Info()` for this case so it's detectable if we go looking for issues.

pmj:
I've been seeing a few Mac functional test failures due to timeout. I haven't been able to repro this locally yet, so I can't realistically diagnose it at this time. (Would need machine access.) I can't think of a way this could have been caused by this change, but if these timeouts haven't been happening anywhere else but this branch, please let me know and I won't merge it until we've tracked the problem down.

pmj:
The latest version addresses the remaining feedbacks, and it seems that the occasional functional test timeouts are not limited to this PR branch, e.g. https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9493066&view=results so I'm finally merging this now.

jeschu1:
@pmj yes we've been seeing hangs in functional tests.  I'm sorry if I didn't give you a heads up there.  I think the "timeout/diagnose hangs" task we talked about in the meeting may get some priority in the nearer term.  Thanks for getting this change in!

PR #986, opened 2019-03-30
[Linux] Build GVFS provider on Linux

chrisd8088:
This is a refreshed take on the WIP #810, which I think is closer to readiness, with a few caveats.

~1. First, after the `features/linuxprototype` branch is rebased (see #976 and #977), there are likely a few new entries in `GVFS.sln` and other changes which will need to be incorporated into this PR so that it builds fully on Linux.  For instance, 2d2a0e19df and c2421cacb4 from my [gvfs-latest-wip](https://github.com/github/VFSForGit/tree/linux-latest-wip/) branch, which has everything except the POSIX refactoring (see item 3 below) included.~  DONE!  :-)
2. Second, this PR avoids using Mono in the same manner as #983, by limiting our support to x86-64 GNU libc systems with a specific `struct stat` layout.  The relevant invocation of `stat(2)` is in `GVFS/GVFS.Platform.Linux/LinuxFileSystem.cs`.  I think the 64-bit limitation is OK, at least, given that all the `*.csproj` files specify an `x64` build target already.
3. Third, if the POSIX refactoring work @kivikakk has done in #926 is completed and merged first, we should redo this PR to utilize that work.  On the other hand, there are some elements of that PR which should incorporate the differences in `DllImport` calls to `libc`, as illustrated in this PR.
4. And fourth, it should be noted that while this PR allows us to build most of the GVFS provider code, including the `gvfs`, `GVFS.Mount`, and Git hook binaries, and does permit the `gvfs clone` operation to complete (assuming you have libprojfs installed), the `gvfs mount` command is *not* functional yet.


kivikakk:
@chrisd8088 I'm curious -- is there a reason we don't just correct what we send to the provider, rather than modifying everything we receive in 19ca466?

chrisd8088:
Good question!  This was a fast patch on my WIP branch, just to see if it worked (and it did!), so we can definitely do something tidier than 19ca466, which was just a quick belt-and-suspenders patch!  :-)

For one thing, we don't need to pull off `./` because libprojfs never sends that; it only uses `.` for events on the root of the mount point itself.

There's no particular consistency among our potential models -- libfuse's high-level API prepends '/' to everything, which we [trim off](https://github.com/github/libprojfs/blob/master/lib/projfs.c#L437).

That puts the libprojfs API more in line with the [inotify API](http://man7.org/linux/man-pages/man7/inotify.7.html), but there too we diverge because inotify only reports events directly within a directory (not subdirectories thereof), and leaves the `name` field empty in its event structure when the event refers to the watched directory itself and not a file within it.

I'd be OK with returning an empty string for events on the mount point, but I have an inkling that will break some other assumptions in the libprojfs code which will also have to be adjusted.  As a first step, though, I'll remove the `./` trimming here because that, at least, is not actually required.

kivikakk:
Rebased `features/linuxprototype` onto master -- @jrbriggs, could you reset `features/linuxprototype` to 5e73fef9d24ee5065d41d187c52c202f0ff6317e?

jrbriggs:
@kivikakk done

kivikakk:
@jrbiggs inasmuch as we want to move toward a place where we have a Linux CI build to prevent regressions, this branch has a clean build (`Scripts/Linux/BuildGVFSForLinux.sh`) even if we don't yet have functional tests running. If you'd be happy with something like that in master, we should merge this and then add a CI pipeline (? unclear on terminology with Azure DevOps) to run the build script.

PR #988, opened 2019-03-31
Mac kext: Replaces use of OSAtomic with standard C11 atomic operations

pmj:
The OSAtomic APIs are not available in that form in user space, and so caused errors when attempting to call functions (`ProviderMessaging_TrySendRequestAndWaitForResponse()`!) that use them in the kext testing target. The similar user space equivalents are already deprecated in any case, so replacing the kext versions' use with the standardised C11 operations is the sensible choice.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -1,9 +1,7 @@
 #include <kern/debug.h>
 #include <sys/kauth.h>
 #include <sys/proc.h>
-#include <libkern/OSAtomic.h>
 #include <kern/assert.h>
-#include <stdatomic.h>
 
 #include "public/PrjFSCommon.h"
 #include "public/PrjFSPerfCounter.h"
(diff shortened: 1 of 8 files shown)


nickgra:
IIRC, the OSAtomics being transitioned to C11 atomics is what caused our problem with IOSharedDataQueue. While we believe that to be fixed, I'm going to run a large repo build on top of this PR just to be sure.

https://mseng.visualstudio.com/AzureDevOps/_build/results?buildId=9158349

pmj:
> IIRC, the OSAtomics being transitioned to C11 atomics is what caused our problem with IOSharedDataQueue. While we believe that to be fixed, I'm going to run a large repo build on top of this PR just to be sure.

Running the large repo build is definitely a good idea; the problem with `IOSharedDataQueue` was that the developer who ported that to C11 atomics for 10.13.6 got fancy with relaxing memory ordering to (I guess?) eke out a tiny extra bit of performance. They got it wrong though and relaxed it too much. I've not used the `atomic_*_explicit()` function variants here, just the `atomic_*()` which uses the default, sequentially consistent ordering, which is the most stringent and should be equivalent to the `OSAtomic` ordering guarantees.

pmj:
Looks like the large repo build completed without a hitch, so I'm going to go ahead and merge this.

PR #997, opened 2019-04-02
Added Code Coverage to build

jeschu1:
We include code coverage reports in the build.  You can see this in the logs.
 
        WaitForListenerCompletion()                                                                                                                                                                                                                            0.00% (0/14)       
        GetPid(vfs_context*)                                                                                                                                                                                                                                   100.00% (4/4)      
        GetVNodeAttributes(vnode*, vfs_context*, vnode_attr*)                                                                                                                                                                                                  100.00% (6/6)      
        TryReadVNodeFileFlags(vnode*, vfs_context*, unsigned int*)                                                                                                                                                                                             100.00% (19/19)    
        FileFlagsBitIsSet(unsigned int, unsigned int)                                                                                                                                                                                                          100.00% (4/4)      
        TryGetFileIsFlaggedAsInRoot(vnode*, vfs_context*, bool*)                                                                                                                                                                                               100.00% (11/11)    
        ActionBitIsSet(int, int)                                                                                                                                                                                                                               100.00% (3/3)      
        IsFileSystemCrawler(char const*)                                                                                                                                                                                                                       100.00% (13/13)
        CurrentProcessWasSpawnedByRegularUser()                                                                                                                                                                                                                53.66% (22/41)     
        ShouldHandleFileOpEvent(PerfTracer*, vfs_context*, vnode*, int, short*, FsidInode*, int*)                                                                                                                                                              68.42% (39/57)     
        WaitForListenerCompletion()                                                                                                                                                                                                                            0.00% (0/14)       
        GetPid(vfs_context*)               

We will fail on Kext functions not on the exclusion list and is not at 100%.

We hope to move to the json format and a prettier output in the future, once we have full ansci support on the build logs.

### DIFF
--- a/ProjFS.Mac/Scripts/Build.sh
+++ b/ProjFS.Mac/Scripts/Build.sh
@@ -9,6 +9,7 @@ SCRIPTDIR=$(dirname ${BASH_SOURCE[0]})
 SRCDIR=$SCRIPTDIR/../..
 ROOTDIR=$SRCDIR/..
 PACKAGES=$ROOTDIR/packages
+COVERAGEDIR=$ROOTDIR/BuildOutput/ProjFS.Mac/Coverage
 
 PROJFS=$SRCDIR/ProjFS.Mac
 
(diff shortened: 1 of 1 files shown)


ProjFS.Mac/Scripts/Build.sh
```diff
@@ -21,9 +21,71 @@ fi
...
+fi
+
+xcrun xccov view "${COVERAGE_FILES[(COVERAGE_FILE_COUNT - 1)]}" --json | xcperfect 
+
+# Example the latest code coverage file
+xcrun xccov view "${COVERAGE_FILES[(COVERAGE_FILE_COUNT - 1)]}" > $PROJFS/CoverageResult.txt 
+
+# Fail on any line that doesn't show %100 coverage and isn't on the exclusion list or hpp/cpp
+while read line; do
+  if [[ $line != *"100.00%"* ]] &&
```
jeschu1: We need all these exclusions since we're including too many files in the test report.  

@pmj do you know how we could switch the code coverage report to just be on KauthHandler.cpp && VirtualizationRoots.cpp.

Alternatively we could parse the json and remove the files we don't care about in that report.  @mjcheetham know a fancy json parser for MacOS as a backup?

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -21,9 +21,71 @@ fi
 
 # Run Tests and put output into a xml file
 set -o pipefail
-xcodebuild -configuration $CONFIGURATION -project $PROJFS/PrjFS.xcodeproj -scheme 'Build All' test | xcpretty -r junit --output $PROJFS/TestResultJunit.xml || exit 1
+xcodebuild -configuration $CONFIGURATION -enableCodeCoverage YES -project $PROJFS/PrjFS.xcodeproj -derivedDataPath $ROOTDIR/BuildOutput/ProjFS.Mac/Coverage -scheme 'Build All' test | xcpretty -r junit --output $PROJFS/TestResultJunit.xml || exit 1
```
mjcheetham: nit: Do we not want to extract a local variable for the `derivedDataPath` argument value?

ProjFS.Mac/Scripts/Build.sh
mjcheetham: You could use `jq` which is very small, lightweight and installable by Homebrew.. it's like "sed/awk for JSON"

ProjFS.Mac/Scripts/Build.sh
jeschu1: Good point yes!

ProjFS.Mac/Scripts/Build.sh
jeschu1: Thanks Matt!  I had to turn off json since the build output didn't support the output from tool.  Once this is supported I think the aim should be to move json full stop and jq looks like a great tool for that.

wilbaker: > Alternatively we could parse the json

@jeschu1 for my own understanding, which of the above commands is producing the json?

jeschu1: @wilbaker the "--json" option does it

`#xcrun xccov view "$COVERAGE_FILE" --json `

PR #999, opened 2019-04-03
use "which" to locate git installation

kivikakk:
Address this TODO -- the same code will work for Linux later. I added a test to exercise this -- it should pass on any VFSForGit running system as git should always be installed, but I have a feeling the unit test will run on Windows too and so will fail. How should I restrict a unit test to Mac(+Linux) only?

/cc @chrisd8088
/cc https://github.com/github/VFSForGit/issues/10


chrisd8088:
Sweet!  👍

GVFS/GVFS.UnitTests/Platform.Mac/MacGitInstallationTests.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace GVFS.UnitTests.Platform.Mac
+{
+    [TestFixture]
+    public class MacGitInstallationTests
+    {
+        [TestCase]
+        public void GetInstalledGitBinPathWorks()
+        {
+            string gitBinPath = new MacGitInstallation().GetInstalledGitBinPath();
```
wilbaker: > I added a test to exercise this -- it should pass on any VFSForGit running system as git should always be installed, but I have a feeling the unit test will run on Windows too and so will fail. How should I restrict a unit test to Mac(+Linux) only?

VFS4G unit tests should not require that Git be installed and they should be able to run on all platforms (with the exception of the Windows unit tests that require .NET Framework).

If we *really* wanted to unit test the Mac's `GetInstalledGitBinPath` implementation we'd need to mock out the call to `which`.  However, `GetInstalledGitBinPath` is hit quite frequently when cloning/mounting and so I think we're getting sufficient coverage of the method from the functional tests.

chrisd8088:
For my WIP branch, I confess I just removed the unit tests but kept the code change in 215357a0, because it was really useful to not have the hard-coded reliance on `/usr/local/bin/git` on my dev system.  :-)

kivikakk:
> To make sure my earlier comments are clear, I think if the unit test is removed from this PR then it's good to go :)

Thanks for this! The unit test has been removed.

We presumably want some kind of exercise of the platform-specific functionality _somewhere_, though, I'd imagine. Where should those (eventually) go?

wilbaker:
> We presumably want some kind of exercise of the platform-specific functionality somewhere, though, I'd imagine. Where should those (eventually) go?

Currently the platform specific code (like `GetInstalledGitBinPath`) is exercised by the functional tests.

kivikakk:
@wilbaker Thanks for the review and clarification there!

PR #1000, opened 2019-04-03
Fetching all sizes before index update

turbonaitis:
If we have ~100k placeholders, getting their sizes one-by-one during index update (so during every checkout, even if you're checking out the same revision, that you're on) from the on-disk size DB takes 2+s. If we preload all sizes from DB into an in-memory dictionary, that takes around 800ms.

There's a risk though, that with time, as the SQLite DB of sizes grows, the proportion of sizes needed during index update will become lower. However, I believe that this falls into its own class of problems, that VFS4G has yet to address - evicting data from on-disk caches (that includes sizes and blobs), as the repository moves on and that data is not needed anymore.

### DIFF
--- a/GVFS/GVFS.Common/Git/Sha1Id.cs
+++ b/GVFS/GVFS.Common/Git/Sha1Id.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Runtime.InteropServices;
 
 namespace GVFS.Common.Git
@@ -87,6 +88,11 @@ namespace GVFS.Common.Git
                     shaBytes17Through20 = *(uint*)thirdChunk;
                 }
             }
+        }
+
+        public static string ShaStringFromBuffer(byte[] shaBuffer)
+        {
+            return new string(shaBuffer.SelectMany(b => new[] { GetHexValue(b / 16),  GetHexValue(b % 16) }).ToArray());
         }
 
         public void ToBuffer(byte[] shaBuffer)
--- a/GVFS/GVFS.Virtualization/BlobSize/BlobSizes.cs
+++ b/GVFS/GVFS.Virtualization/BlobSize/BlobSizes.cs
@@ -292,6 +292,7 @@ namespace GVFS.Virtualization.BlobSize
             // the prepared SQLite statement can be reused
             private SqliteConnection connection;
             private SqliteCommand querySizeCommand;
+            private SqliteCommand getAllDataCommand;
             private SqliteParameter shaParam;
 
             private byte[] shaBuffer;
@@ -333,6 +334,11 @@ namespace GVFS.Virtualization.BlobSize
                     this.querySizeCommand.CommandText = "SELECT size FROM BlobSizes WHERE sha = (@sha);";
                     this.querySizeCommand.Parameters.Add(this.shaParam);
                     this.querySizeCommand.Prepare();
+
+                    this.getAllDataCommand = this.connection.CreateCommand();
+
+                    this.getAllDataCommand.CommandText = "SELECT sha, size FROM BlobSizes;";
+                    this.getAllDataCommand.Prepare();
                 }
                 catch (Exception e)
                 {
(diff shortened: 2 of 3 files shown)


GVFS/GVFS.Virtualization/BlobSize/BlobSizes.cs
```diff
@@ -333,6 +334,11 @@ public BlobSizesConnection(BlobSizes blobSizes, string connectionString)
                     this.querySizeCommand.CommandText = "SELECT size FROM BlobSizes WHERE sha = (@sha);";
                     this.querySizeCommand.Parameters.Add(this.shaParam);
                     this.querySizeCommand.Prepare();
+
+                    this.getAllDataCommand = this.connection.CreateCommand();
+
+                    this.getAllDataCommand.CommandText = "SELECT sha, size FROM BlobSizes;";
```
jrbriggs: This has the possibility of returning several million rows in what I believe would not be a very uncommon case in some of Microsoft's internal repositories. We need to limit this query to only what we need to project the index, but pulling that data out in a batch, or several, might be a feasible optimization. Have you explored that?

turbonaitis: Not yet - went with the simple option first and in my case (~100k placeholders) it worked well. Let me see how this would work, if we specified the sha's of the items, that we're interested in.

jrbriggs: Thanks! I think this could certainly help performance, if we can tune the batch size correctly.

GVFS/GVFS.Virtualization/BlobSize/BlobSizes.cs
```diff
@@ -380,6 +392,22 @@ public virtual bool TryGetSize(Sha1Id sha, out long length)
...
             }
 
+            public virtual Dictionary<string, long> GetAllSizes()
+            {
+                Dictionary<string, long> result = new Dictionary<string, long>();
+                using (SqliteDataReader reader = this.getAllDataCommand.ExecuteReader())
+                {
+                    while (reader.Read())
+                    {
+                        var blob = (byte[])reader.GetValue(0);
```
derrickstolee: We don't use `var`.

GVFS/GVFS.Common/Git/Sha1Id.cs
```diff
@@ -89,6 +90,11 @@ public static bool TryParse(string sha, out Sha1Id sha1, out string error)
             }
         }
 
+        public static string ShaStringFromBuffer(byte[] shaBuffer)
+        {
+            return new string(shaBuffer.SelectMany(b => new[] { GetHexValue(b / 16),  GetHexValue(b % 16) }).ToArray());
```
derrickstolee: Is this run in a tight loop? This use of Linq including a `new [] {}` in the middle of your loop is likely to be inefficient:

1. You create and destroy a lot of string arrays with the `new [] {}` in the middle.
2. `.ToArray()` is notoriously bad, since it cannot guess the size of the array at the start, so dynamically sizes a few arrays before truncating at the end. However, we know the length that `shaBuffer` should be.
3. Even the `.SelectMany()` is instantiating an enumerator that is unnecessary.

I think it is worth rewriting this method using [`StringBuilder` with a specified capacity](https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.-ctor?view=netframework-4.8#System_Text_StringBuilder__ctor_System_Int32_) and a simple `for (int i = 0; i < shaBuffer.Length; i++)` loop.

Finally, please add a simple unit test for this method.

turbonaitis:
@derrickstolee was this issue addressed in another PR?

derrickstolee:
@turbonaitis I was just doing some housekeeping. I should have left a note that this was closed for inactivity, as it seemed to not have any progress in a few months.

turbonaitis:
Ok, I'll then keep this on my list and might reopen, once I have time to
work on this particular issue, thanks!

derrickstolee:
It looks like @wilbaker is improving the perf here, anyway, in #1398.

turbonaitis:
oh, that change looks even better! Looking forward to that change being
merged.

wilbaker:
> oh, that change looks even better! Looking forward to that change being
merged.

FYI #1398 is for a different size download perf issue (expanding folders during enumeration).  

However, I am starting on #1397 now which will address this scenario (updating placeholders after index update).

wilbaker:
>  I am starting on #1397 now which will address this scenario (updating placeholders after index update). Note that although the issue description mentions downloading sizes, VFS4G will not look them up at all when updating placeholders.

Changes for this are now in a PR here: #1403

Issue #781, opened 2019-02-12
Sporadic failure in DehydrateTests.DehydrateShouldBackupFiles


wilbaker:
I've seen this issue once so far in this run: https://gvfs.visualstudio.com/ci/_build/results?buildId=6815

```
2019-02-12T16:34:05.9122172Z => GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.DehydrateShouldBackupFiles()
2019-02-12T16:34:05.9123160Z Test GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.DehydrateShouldBackupFiles()
2019-02-12T16:34:05.9124207Z Failed at 4:34:05 PM taking 00:00:16.3815202
2019-02-12T16:34:05.9124527Z   Expected substring 'The repo was successfully dehydrated and remounted' not found in 'Running git status before dehydrating to make sure you don't have any pending changes.
2019-02-12T16:34:05.9125196Z If this takes too long, you can abort and run dehydrate with --no-status to skip this safety check.
2019-02-12T16:34:05.9125812Z 
2019-02-12T16:34:05.9126355Z 
2019-02-12T16:34:05.9127004Z Running git status...—
2019-02-12T16:34:05.9127450Z Running git status...\
2019-02-12T16:34:05.9127919Z Running git status...\
2019-02-12T16:34:05.9128338Z Running git status...|
2019-02-12T16:34:05.9128768Z Running git status...|
2019-02-12T16:34:05.9129190Z Running git status.../
2019-02-12T16:34:05.9129791Z Running git status.../
2019-02-12T16:34:05.9130974Z Running git status...—
2019-02-12T16:34:05.9131495Z Running git status...—
2019-02-12T16:34:05.9132011Z Running git status...\
2019-02-12T16:34:05.9132520Z Running git status...\
2019-02-12T16:34:05.9132943Z Running git status...|
2019-02-12T16:34:05.9133505Z Running git status...Succeeded
2019-02-12T16:34:05.9134179Z 
2019-02-12T16:34:05.9134812Z Starting dehydration. All of your existing files will be backed up in C:\Repos\GVFSFunctionalTests\enlistment\9a2e2a23dcf448bc9770\dehydrate_backup\20190212_163353
2019-02-12T16:34:05.9135315Z WARNING: If you abort the dehydrate after this point, the repo may become corrupt
2019-02-12T16:34:05.9135658Z 
[...]

PR #1002, opened 2019-04-03
Add more logging when test gets access denied from moving src folder

kewillford:
This is to hopefully get more information when a test hits #781 so that we will be able to find out what process has the handle open and is causing the access denied from the move.

This adds the dependency on `handle.exe` but the code is catching all exceptions when running it so if it isn't found it will only output that it wasn't found.  It will also only be ran in the very specific case of moving the `src` folder failing on access right now. So it will not cause a test to fail or succeed, only output information.


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -200,5 +207,24 @@ private ProcessResult RunDehydrateProcess(bool confirm, bool noStatus)
 
             return ProcessHelper.Run(processInfo);
         }
+
+        private string RunHandleProcess(string path)
+        {
+            try
+            {
+                ProcessStartInfo processInfo = new ProcessStartInfo("handle.exe");
```
jrbriggs: Do all the agents already have handle.exe installed?

kewillford: Yep just put it on all of them.

kewillford:
Closes #781

PR #1004, opened 2019-04-03
Test reliability issue #852

alameenshah:
About the test failure - when Upgrade reminder marker file is created by the `Service`, `git` hooks should start showing reminder messaging. That does not seem to happen sometimes. It is not clear whether it was  a failure in the `Service` (the `Service` was not able to create the marker file) or if it was the `git` hooks which failed to display the reminder. 

Added better logging - Log details on failures. Updated Service to log message when upgrade reminder marker file is created or deleted.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgraderInfo.cs
+++ b/GVFS/GVFS.Common/ProductUpgraderInfo.cs
@@ -58,11 +58,21 @@ namespace GVFS.Common
                 if (this.fileSystem.FileExists(highestAvailableVersionFile))
                 {
                     this.fileSystem.DeleteFile(highestAvailableVersionFile);
+
+                    if (this.tracer != null)
+                    {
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -58,11 +58,21 @@ public void RecordHighestAvailableVersion(Version highestAvailableVersion)
                 if (this.fileSystem.FileExists(highestAvailableVersionFile))
                 {
                     this.fileSystem.DeleteFile(highestAvailableVersionFile);
+
+                    if (this.tracer != null)
+                    {
+                        this.tracer.RelatedInfo($"{nameof(this.RecordHighestAvailableVersion)}: Deleted Upgrade reminder marker file");
```
jrbriggs: nit: unnecessary capitalization on "upgrade". Same applies to next instance.

PR #1007, opened 2019-04-03
Do not attempt to approve credentials for anonymous access

jamill:
he current logic has an issue where it will attempt to approve
credentials when communicating with anonymous access, even though there
are no credentials to approve. The other logic in this method guards
credential operations to non-anonymous auth scenarios. Update logic to
only approve / store credentials to non-anonymous auth scenarios.

This can be observed in the functional test logs that connect with anonymous auth.
These tests generate error logs indicating that it was unable to parse the credential string:

`Error "Failed to parse credential string for approval"`

### DIFF
--- a/GVFS/GVFS.Common/Http/HttpRequestor.cs
+++ b/GVFS/GVFS.Common/Http/HttpRequestor.cs
@@ -99,7 +99,7 @@ namespace GVFS.Common.Http
 
             request.Headers.UserAgent.Add(this.userAgentHeader);
 
-            if (!string.IsNullOrEmpty(authString))
+            if (!this.authentication.IsAnonymous)
             {
                 request.Headers.Authorization = new AuthenticationHeaderValue("Basic", authString);
             }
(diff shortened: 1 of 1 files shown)


jamill:
I will look at testing options here...

GVFS/GVFS.Common/Http/HttpRequestor.cs
```diff
@@ -99,7 +99,7 @@ public void Dispose()
 
             request.Headers.UserAgent.Add(this.userAgentHeader);
 
-            if (!string.IsNullOrEmpty(authString))
+            if (!this.authentication.IsAnonymous)
```
wilbaker: This is definitely cleaner than the old code, but for my own understanding is `authString` always null or empty when `authentication.IsAnonymous` is true?

jamill: > is authString always null or empty 

Yes - a couple of lines above we only attempt to get credentials if this is *not* anonymous access. It is a bit removed and I asked myself the same thing. 

I had wondered if we could do a bit broader cleanup / refactoring to improve readability around some of the authentication classes to help address questions like this, but that is a different topic :)

jamill:
@jrbriggs - updated PR description with more details on the motivation for the change.

jrbriggs:
> @jrbriggs - updated PR description with more details on the motivation for the change.

Thank you!

PR #1008, opened 2019-04-03
Mac uninstaller launchd issue

### LINKED ISSUE #1005: Uninstaller should remove launchd plist file
VFSForGit Mac un-install script deletes `/usr/local/vfsforgit` directory before unloading the Service. If (for some reason) Service unload fails, then un-installer quits immediately, without deleting `/Library/LaunchAgents/org.vfsforgit.service.plist`.) Now launchd continues to think `GVFS.Service.dll` exists and try to load it. Each load results in failure, and by design repeatedly failing launchd agents get disabled. Once `GVFS.Service` is disabled, subsequent VFSForGit installers will fail (because `Service` was disabled and loading it from the installer will always fail

alameenshah:
Un-installer was not deleting launchd plist files when LaunchAgent unload fails. This was causing a dangling launchd plist file that was pointing to VFSForGit assemblies that no longer exist on disc. Launchd disables the VFSForGit services, causing future VFSForGit installs to fail. Updated un-installer to remove the launchd plist file even when the unloading fails. When such a failure happens, user will have to manually logout and re-login for the un-install to be complete.

Fixes #1005

### DIFF
--- a/GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
+++ b/GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
@@ -24,48 +24,59 @@ function UnloadKext()
 
 function UnInstallVFSForGit()
 {
-    if [ -d "${VFSFORDIRECTORY}" ]; then
-        rmCmd="sudo /bin/rm -Rf ${VFSFORDIRECTORY}"
-        echo "$rmCmd..."
-        eval $rmCmd || exit 1
-    fi
-    
     if [ -d "${PRJFSKEXTDIRECTORY}/$KEXTFILENAME" ]; then
         rmCmd="sudo /bin/rm -Rf ${PRJFSKEXTDIRECTORY}/$KEXTFILENAME"
         echo "$rmCmd..."
-        eval $rmCmd || exit 1
+        eval $rmCmd || { echo "Error: Could not delete ${PRJFSKEXTDIRECTORY}/$KEXTFILENAME. Delete it manually."; exit 1; }
     fi
     
     if [ -f "${LAUNCHDAEMONDIRECTORY}/$LOGDAEMONLAUNCHDFILENAME" ]; then
-        unloadCmd="sudo launchctl unload -w ${LAUNCHDAEMONDIRECTORY}/$LOGDAEMONLAUNCHDFILENAME"
-        echo "$unloadCmd..."
-        eval $unloadCmd || exit 1
+        # Check if the daemon is loaded. Unload only if necessary.
+        isLoadedCmd="sudo launchctl kill SIGCONT system/org.vfsforgit.prjfs.PrjFSKextLogDaemon"
+        echo "$isLoadedCmd"
+        if $isLoadedCmd; then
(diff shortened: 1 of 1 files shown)


nickgra:
Should we try unloading the service before we delete the `/usr/local/vfsforgit` directory to ensure correctness?

GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
```diff
@@ -54,18 +54,18 @@ function UnInstallVFSForGit()
         for uid in $(ps -Ac -o uid,command | grep -iw "loginwindow" | awk '{print $1}'); do
             unloadCmd="sudo launchctl bootout gui/$uid ${LAUNCHAGENTDIRECTORY}/$SERVICEAGENTLAUNCHDFILENAME"
             echo "$unloadCmd..."
-            eval $unloadCmd || exit 1
+            eval $unloadCmd || echo "Error: Could not unload ${LAUNCHAGENTDIRECTORY}/$SERVICEAGENTLAUNCHDFILENAME. Try logout and login to finish un-install."
```
nickgra: If we reach these echo statements, will we still return a non-zero error code?

jrbriggs: I would expect a non-zero error code if we didn't completely succeed, I think.

alameenshah: how about we cache the error. But rather than exiting immediately, continue with remaining un-install. In the end exit with the cached error code.

alameenshah: Updated to log & early exit in case of errors.

alameenshah:
> Should we try unloading the service before we delete the `/usr/local/vfsforgit` directory to ensure correctness?

makes sense, moved it.

GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
```diff
@@ -24,48 +24,57 @@ function UnloadKext()
...
         echo "$rmCmd..."
-        eval $rmCmd || exit 1
+        eval $rmCmd || { echo "Error: Could not delete ${LAUNCHDAEMONDIRECTORY}/$LOGDAEMONLAUNCHDFILENAME. Delete it manually."; exit 1; }
     fi
-    
+        
     # Unloading Service LaunchAgent for each user
     # There will be one loginwindow instance for each logged in user, 
     # get its uid (this will correspond to the logged in user's id.) 
     # Then use launchctl bootstrap gui/uid to auto load the Service
```
nickgra: Update comment to describe what we expect the isLoadedCmd to do? Also, this is old but we should fix the copy/paste error from the preinstall script:
> Then use launchctl bootstrap gui/uid to auto load the Service 

That should be bootout.

alameenshah: rephrased

PR #1009, opened 2019-04-04
Remove Code Coverage Exclusion for ShouldIgnoreVnodeType

jeschu1:
ShouldIgnoreVnodeType can be removed from the exclusion list since it has code coverage at 100%

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -1057,13 +1057,14 @@ KEXT_STATIC bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode)
 	case VREG:
     case VDIR:
     case VLNK:
-        return false;
+        break;
     case VSTR:
     case VCPLX:
         KextLog_FileInfo(vnode, "vnode with type %s encountered", vnodeType == VSTR ? "VSTR" : "VCPLX");
-        return false;
+        break;
     default:
         KextLog_FileInfo(vnode, "vnode with unknown type %d encountered", vnodeType);
-        return false;
     }
+    
+    return false;
 }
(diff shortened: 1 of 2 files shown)


wilbaker:
@jeschu1 it looks like the Mac builds are failing with this PR but I can't figure out why.  Any ideas?

jamill:
Can you add more context to the commit message / PR description?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -1057,13 +1057,14 @@ KEXT_STATIC bool ShouldIgnoreVnodeType(vtype vnodeType, vnode_t vnode)
...
     case VCPLX:
         KextLog_FileInfo(vnode, "vnode with type %s encountered", vnodeType == VSTR ? "VSTR" : "VCPLX");
-        return false;
+        break;
     default:
         KextLog_FileInfo(vnode, "vnode with unknown type %d encountered", vnodeType);
-        return false;
     }
+    
+    return false;
```
jeschu1: @nickgra XCode 9x's code coverage was confused by the previous function and thought the last 2 curly braces were never tested.  This was fixed in XCode10.

The force coverage for this function, I reworked it slightly.  I think its a fair tradeoff, although not something we want to do on a regular basis.

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -83,7 +82,7 @@ while read line; do
 	 [[ $line != *".xctest"* ]] && 
 	 [[ $line != *".cpp"* ]] && 
 	 [[ $line != *".hpp"* ]]; then
-       printf "\nError: Not at 100% Code Coverage: $line"
+       echo "Error: not at 100% Code Coverage $line"
```
jeschu1: The bash on the build machine as not able to execute the printf, so it's been updated to echo.  Locally of course it all ran fine.

PR #1011, opened 2019-04-05
[Linux] Pin callback delegates and simplify marshalling in ProjFS.Linux

chrisd8088:
Hi @kivikakk, @jrbriggs, and @wilbaker -- I'll be interested in your comments on this one!

It might best be reviewed by looking at the individual commits; the overall summary is that this fixes several issues in our `ProjFS.Linux` `VirtualizationInstance` implementation, and tidies up the marshalling of UTF-8 strings.

1. We need to retain class references to our callback delegates in the [same way](https://github.com/Microsoft/VFSForGit/blob/master/ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs#L38) the Mac `VirtualizationInstance` does or else we get crashes on callbacks when testing with the GVFS provider.  (3e2147f)

2. We weren't retrieving our extended attributes correctly after a file rename.  At first I thought my quick fix in 5e250f3 had an obvious race against other file operations, but now I'm not convinced of that, since libprojfs will not return from its FUSE `rename` operation until the event notification to the C# provider is complete, and so FUSE won't have returned success to the Linux VFS.  So it really depends on the exact `rename()` locking semantics within the Linux kernel.  If there is a race condition, we'll need a more comprehensive fix which will involve changes to libprojfs, so I'd like to get this into place now as at least a stopgap before we deal with it more comprehensively.

3. Based on my [stumbling](https://github.com/Microsoft/VFSForGit/pull/848#issuecomment-476827046) around the question of UTF-8 marshalling it appears (and testing appears to confirm) that on recent .NET Core platforms, encoding to/from UTF-8 is the default when marshalling unmanaged strings, so we can simplify a fair bit of our handling of those.  The only downside is that we lose the clarity of having `byte*` in our `Event` structure; it becomes a more generic `IntPtr`.  @kivikakk, do you (or others) may have particular concerns about that?  (1e6e827)

4. I moved the write-bytes-to-file loop down into a `NativeFileWriter` class, which I think [parallels](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.Platform.Mac/MacFileSystem.cs#L153) the `NativeFileReader` found in `GVFS.Platform.Mac.MacFileSystem`.  That also tidies up the main class in `VirtualizationInstance` a bit.  (d6c3e12)

5. Ashe fixed a problem where I was inadvertently relying on GNU awk extensions in one of our build scripts.  (1a331b8)

6. And lastly I don't think we need the `NotifyOperationCallback` delegate at all on Linux, given how we've implemented other callbacks.  (Also in 5e250f3)

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -319,8 +319,16 @@ namespace PrjFSLib.Linux
 
                 if (!isDirectory)
                 {
+                    string currentRelativePath = relativePath;
+
+                    // TODO(Linux): can other intermediate file ops race us here?
+                    if (nt == NotificationType.FileRenamed)
+                    {
+                        currentRelativePath = PtrToStringUTF8(ev.TargetPath);
+                    }
+
                     result = this.projfs.GetProjAttrs(
-                        relativePath,
+                        currentRelativePath,
                         providerId,
                         contentId);
                 }
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -259,36 +259,36 @@ namespace PrjFSLib.Linux
                 relativePath = PtrToStringUTF8(ev.Path);
             }
 
-                if ((ev.Mask & ProjFS.Constants.PROJFS_ONDIR) != 0)
+            if ((ev.Mask & ProjFS.Constants.PROJFS_ONDIR) != 0)
+            {
+                result = this.OnEnumerateDirectory(
+                    commandId: 0,
+                    relativePath: relativePath,
+                    triggeringProcessId: ev.Pid,
+                    triggeringProcessName: triggeringProcessName);
+            }
+            else
+            {
+                byte[] providerId = new byte[PlaceholderIdLength];
+                byte[] contentId = new byte[PlaceholderIdLength];
+
+                result = this.projfs.GetProjAttrs(
+                    relativePath,
+                    providerId,
+                    contentId);
+
+                if (result == Result.Success)
                 {
-                    result = this.OnEnumerateDirectory(
+                    result = this.OnGetFileStream(
                         commandId: 0,
                         relativePath: relativePath,
+                        providerId: providerId,
+                        contentId: contentId,
                         triggeringProcessId: ev.Pid,
-                        triggeringProcessName: triggeringProcessName);
+                        triggeringProcessName: triggeringProcessName,
+                        fd: ev.Fd);
                 }
-                else
-                {
-                    byte[] providerId = new byte[PlaceholderIdLength];
-                    byte[] contentId = new byte[PlaceholderIdLength];
-
-                    result = this.projfs.GetProjAttrs(
-                        relativePath,
-                        providerId,
-                        contentId);
-
-                    if (result == Result.Success)
-                    {
-                        result = this.OnGetFileStream(
-                            commandId: 0,
-                            relativePath: relativePath,
-                            providerId: providerId,
-                            contentId: contentId,
-                            triggeringProcessId: ev.Pid,
-                            triggeringProcessName: triggeringProcessName,
-                            fd: ev.Fd);
-                    }
-                }
+            }
 
             return -result.ToErrno();
         }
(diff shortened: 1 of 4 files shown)


kivikakk:
I was wary about the change in 1e6e827, but manual testing confirmed Unicode paths look okay, and I eventually found the bit where the non-Windows-specific behaviour of `PtrToStringAnsi` is defined: https://github.com/dotnet/coreclr/blob/56043b6c896417e792e64a69692fed3251ed3adc/src/System.Private.CoreLib/shared/System/String.cs#L234

So I'm satisfied! We should be able to drop the `PtrToStringUTF8` wrapper entirely -- there's a null check in Marshal which returns null if the zero pointer is given.

chrisd8088:
Indeed -- I was likewise dubious about the UTF-8 marshalling and went through the a similar process (see my ["item B" comments](https://github.com/Microsoft/VFSForGit/pull/848#issuecomment-476827046)) of testing and digging around in docs and code before being convinced.  :-)

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -231,74 +221,65 @@ private static string GetProcCmdline(int pid)
         }
 
         // TODO(Linux): replace with netstandard2.1 Marshal.PtrToStringUTF8()
```
wilbaker: Is there anything left to do for this TODO?

`// TODO(Linux): replace with netstandard2.1 Marshal.PtrToStringUTF8()`

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -307,36 +288,40 @@ private int HandleNonProjEvent(ref Interop.ProjFS.Event ev, bool perm)
...
-                        isDirectory: isDirectory,
-                        notificationType: nt);
-                }
+                result = this.projfs.GetProjAttrs(
+                    currentRelativePath,
+                    providerId,
+                    contentId);
+            }
+
+            if (result == Result.Success)
```
wilbaker: Under what conditions will `GetProjAttrs` return false?

This is a bit unrelated to this PR but I wanted to mention that VFS4G needs notifications for all files inside of its root.  If `GetProjAttrs` returns false for newly created files inside the repo (i.e. those that it never projected), then VFS4G will be missing notifications it needs.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -404,13 +389,32 @@ private int HandlePermEvent(ref Interop.ProjFS.Event ev)
             return Result.ENotYetImplemented;
         }
 
-        private static unsafe class NativeMethods
+        private static unsafe class NativeFileWriter
         {
-            [DllImport("libc", EntryPoint = "strlen")]
-            public static extern ulong Strlen(byte* s);
+            public static bool TryWrite(int fd, byte[] bytes, uint byteCount)
```
wilbaker: > I moved the write-bytes-to-file loop down into a NativeFileWriter class, which I think parallels the NativeFileReader found in GVFS.Platform.Mac.MacFileSystem. That also tidies up the main class in VirtualizationInstance a bit.

I think that `NativeFileReader` and `NativeFileWriter` are probably being used for different things.  `NativeFileReader` is used during `gvfs prefetch` (when `--hydrate` is specified) to force files to be hydrated.

I am guessing that `NativeFileWriter` is used to actually do the hydration in the GetFileStream callback?  If that's the case it looks like the parallel to [PrjFS_WriteFileContents ](https://github.com/Microsoft/VFSForGit/blob/e3a7dee707ac847e9ab169a87c632829e711af7c/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L589).

(I don't think any change is needed here, I just wanted to make sure we're all on the same page).

chrisd8088: Indeed, that's all correct -- I apologize, I didn't mean to imply a functional parallel with the `NativeFileReader`, only that the implementation (i.e., a `NativeFile*` helper class) was similar.

The macOS `PrjFS_WriteFileContents()` can just call `fwrite()` natively in C++, but we do a bit more work in the Linux `GetFileStream` callback because we're in C#.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: Not really ... I mused over whether to just call `Marshal.PtrToStringAnsi()` in the three places we're calling our wrapper, and left it in place for now only because I think (and may be wrong!) that if we ever build on .NET Standard 2.1, we could explicitly call `Marshal.PtrToStringUTF8()`, which would be a little cleaner to read.

It looks like the `PtrToStringUTF8()` was only [introduced for .NET Standard 2.1](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.ptrtostringutf8?view=netstandard-2.1), but we're using only [.NET Standard 2.0](https://github.com/Microsoft/VFSForGit/blob/features/linuxprototype/ProjFS.Linux/PrjFSLib.Linux.Managed/PrjFSLib.Linux.Managed.csproj#L3) right now.

wilbaker: I don't have a strong preference here, I just wasn't sure with the change to call a `Marshal` method if you still wanted to leave the `TODO` comment in place.  

If it's a reminder to change methods when we eventually update to 2.1 I'm good with leaving it.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: Good question.  We should be returning notifications for all files, regardless of whether they have projection xattrs, because we treat that condition specially.

So `GetProjAttrs()` would only return a non-`Success` `Result` error code if something happened like the file was missing.

If an event is triggered for a file which lacks projection attributes (e.g., a file which VFS4G did not create), that wouldn't produce an error here, because we treat `ENOATTR` specially in libprojfs.  In this case the `providerId` and `contentId` byte arrays will be left unchanged; I believe C# allocates bytes as `0` so the arrays will just have all-zero values.  Does that seem OK?

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: I'm also sort of hoping/expecting that the C# compiler or runtime will optimize away the wrapper.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
wilbaker: That makes sense, thanks!

chrisd8088:
OK, I think this is set to merge unless you have any additional thoughts, @jrbriggs.

PR #1013, opened 2019-04-05
WIP: Make ProductUpgradeTimer testable

jamill:
This is an early iteration on making the ProductUpgradeTimer testable.

This class has several layers of dependencies, and several options for where to introduce test objects.

## Goals:
- Get Product Upgrade Timer into a test harness, so we can continue to iterate on testability improvements.

- Example of changes to get existing code into our unit test harness

- Enable testability of ProductUpgradeTimer with specific Upgraders to enable integration testing without hitting network.

## Todo / open questions:
- [ ] Does `ProductUpgradeTimer.TimerCallback(...)` need to be public?
- [ ] Make Interface / Abstract class for `ProductUpgradeInfo`
- [ ] Test ProductUpgradeTimer in isolation from specific Upgraders?
- [ ] Interface vs Abstract classes for contract definition


GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -283,18 +291,16 @@ protected virtual bool TryDownloadAsset(Asset asset, out string errorMessage)
...
+            this.httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(JSONMediaType));
+            this.httpClient.DefaultRequestHeaders.Add("User-Agent", UserAgent);
 
             releases = null;
             errorMessage = null;
 
             try
             {
-                Stream result = client.GetStreamAsync(GitHubReleaseURL).GetAwaiter().GetResult();
+                Stream result = this.httpClient.GetStreamAsync(GitHubReleaseURL).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
jamill: There seems to be a lot of discussion in general about how to mock this class - a particular lengthy thread can be found here:

https://github.com/dotnet/corefx/issues/1624

GVFS/GVFS.UnitTests.Windows/Windows/Service/ProductUpgradeTimerTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        private interface IHttpMessageHandlerProtectedMembers
+        {
+            Task<HttpResponseMessage> SendAsync(HttpRequestMessage message, CancellationToken token);
+        }
+
+        [TestCase]
+        public void QueriesGitHubForUpgrade()
+        {
+            MockTracer tracer = new MockTracer();
+            Mock<PhysicalFileSystem> fileSystemMock = new Mock<PhysicalFileSystem>();
```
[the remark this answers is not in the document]
jamill: I am not opposed to this in cases where we set up expectations exactly (or maybe we can always set up expectations exactly). It might get a little harder with some methods that have more complex signatures with out params (which is a lot of our code).

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -4,21 +4,35 @@
...
+        private ITracer tracer;
         private PhysicalFileSystem fileSystem;
+        private LocalGVFSConfig gvfsConfig;
+        private HttpClient httpClient;
         private Timer timer;
+        private InstallerRunPreCheckerBase installerPreRunChecker;
+        private ProductUpgraderInfo productUpgradeInfo;
 
-        public ProductUpgradeTimer(JsonTracer tracer)
+            public ProductUpgradeTimer(ITracer tracer, PhysicalFileSystem fileSystem, LocalGVFSConfig gvfsConfig, HttpClient httpClient, InstallerRunPreCheckerBase installerPreRunChecker, ProductUpgraderInfo productUpgradeInfo)
```
alameenshah: nit: can we add line breaks.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -4,21 +4,35 @@
...
-            this.fileSystem = new PhysicalFileSystem();
+            this.fileSystem = fileSystem;
+            this.gvfsConfig = gvfsConfig;
+            this.httpClient = httpClient;
+            this.installerPreRunChecker = installerPreRunChecker;
+            this.productUpgradeInfo = productUpgradeInfo;
+        }
+
+        public ProductUpgradeTimer(ITracer tracer)
+                : this(tracer, new PhysicalFileSystem(), new LocalGVFSConfig(), new HttpClient(), new InstallerPreRunChecker(tracer, string.Empty), new ProductUpgraderInfo(tracer, new PhysicalFileSystem()))
```
alameenshah: nit: line breaks?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -55,33 +69,20 @@ public void Dispose()
...
 
             using (ITracer activity = this.tracer.StartActivity("Checking for product upgrades.", EventLevel.Informational))
             {
                 try
                 {
-                    ProductUpgraderInfo info = new ProductUpgraderInfo(
-                        this.tracer,
-                        this.fileSystem);
-
                     // The upgrade check always goes against GitHub
```
alameenshah: This would change to checking against well known server with anonymous access, right? Can we test this (which server gets used for upgrade checks)?

GVFS/GVFS.UnitTests.Windows/Windows/Service/ProductUpgradeTimerTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            productUpgradeInfoMock.Setup(m => m.RecordHighestAvailableVersion(null)).Verifiable();
+
+            using (ProductUpgradeTimer upgradeChecker = new ProductUpgradeTimer(tracer, fileSystemMock.Object, gvfsConfig, httpClient, installerPreRunChecker.Object, productUpgradeInfoMock.Object))
+            {
+                upgradeChecker.TimerCallback(null);
+            }
+
+            productUpgradeInfoMock.VerifyAll();
+        }
+
```
alameenshah: Can we also add a positive test case where the upgrade check succeeds

GVFS/GVFS.Common/InstallerRunPreCheckerBase.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace GVFS.Common
+{
+    public abstract class InstallerRunPreCheckerBase
```
alameenshah: Why did you need to create a base class? Is that because of any limitation with `Mock`?

GVFS/GVFS.Common/GitHubUpgrader.cs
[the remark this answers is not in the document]
kewillford: In my mind if you can't mock it then abstract it away so for example this class would have a dependency of 
```
public interface IReleaseFetcher
{
    List<Release> GetReleases();
}
```

Then you could at least mock that and have another class that handles the http call and have separate tests for that class which of course you would have the same issues with mocking `HttpClient` but would hopefully limit the scope.

alameenshah:
The testing approach looks good to me. 

One comment (not related to this PR though) - the decision to use Github (or another server) for upgrade checks seem to exist in the `ProductUpgradeTimer` class. It would be a cleaner separation of responsibilities if `ProductUpgrader` could own that decision; `ProductUpgradeTimer` could then become a blind consumer of `ProductUpgrader`.

GVFS/GVFS.Common/InstallerRunPreCheckerBase.cs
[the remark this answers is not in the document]
jamill: I am not against this...

PR #1014, opened 2019-04-06
Check the projection for the folder before attempting to delete

### LINKED ISSUE #245: Mac: Handle failures writing and deleting placeholders when re-expanding folders as part of a projection change
Follow up work item for #239, see TODO's in code for specific spots that need to be addressed

### LINKED ISSUE #480: Aggressive removal of folder placeholders can cause a significant perf hit when changing branches
*Steps to reproduce:*

1. `gvfs clone <large repo with a lot of folders>`
2. `git checkout -b <new branch>`
3. `dir /s` inside `src`
4. `git checkout master`

**Observe** VFS4G will take a long time in step 4.  If after step 4 is complete you checkout `<new branch>` again the checkout runs quickly.

*Additional details*
It appears that the issue is VFS4G's aggressive removal of folder placeholders.  When there are a lot of them (e.g. 600k+) it can take minutes for VFS4G to delete all of the folder placeholders (which it's doing even though the folders are still in the projection).

Perhaps on Windows VFS4G should only attempt to remove folders when they are no longer in the projection?

kewillford:
Because attempting to delete a folder placeholder is an expensive operation
especially when there are hundreds of thousands, we want to reduce the
number that we need to try to delete.  There are a few criteria for attempting
the delete.

First if the folder is in the `folderPlaceholdersToKeep` list it will not be
deleted.  Then if one of the following is true it will try to delete the folder placeholder.

1. If the path is no longer in the projection
2. If the path is in the projection but is a file and not a folder
3. If the path no longer exists on the file system - this is in the case
of tombstones on windows.

Fixes #480 
Closes #245 

Testing with ~325,000 folder placeholder the delete was taking a little over 60 seconds.  With these changes it is only taking ~20 seconds.

### DIFF
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1163,20 +1163,25 @@ namespace GVFS.Virtualization.Projection
                         bool keepFolder = true;
                         if (!folderPlaceholdersToKeep.Contains(folderPlaceholder.Path))
                         {
-                            this.GetChildNameAndParentKey(folderPlaceholder.Path, out string childName, out string parentKey);
-                            FolderData folderData = this.GetProjectedFolderEntryData(CancellationToken.None, blobSizesConnection, availableSizes, childName, parentKey, out string _) as FolderData;
+                            bool isProjected = this.IsPathProjected(folderPlaceholder.Path, out string fileName, out bool isFolder);
 
                             // Check the projection for the folder to determine if the folder needs to be deleted
                             // The delete will be attempted if one of the following is true
                             // 1. not in the projection anymore
                             // 2. in the projection but is not a folder in the projection
                             // 3. Folder no longer exist on disk - this is to remove the tombstone
-                            if (folderData == null ||
-                                !folderData.IsFolder ||
-                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path)))
+                            if (!isProjected ||
+                                !isFolder ||
+                                (!GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories &&
+                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path))))
                             {
                                 keepFolder = !this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder);
                                 ++deleteFolderPlaceholderAttempted;
+                            }
+
+                            if (keepFolder)
+                            {
+                                this.AddParentFoldersToListToKeep(folderPlaceholder.Path, folderPlaceholdersToKeep);
                             }
                         }
 
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1157,12 +1157,35 @@ namespace GVFS.Virtualization.Projection
                     //  1. Ensures child folders are deleted before their parents
                     //  2. Ensures that folders that have been deleted by git (but are still in the projection) are found before their
                     //     parent folder is re-expanded (only applies on platforms where EnumerationExpandsDirectories is true)
+                    Dictionary<string, long> availableSizes = new Dictionary<string, long>();
                     foreach (PlaceholderListDatabase.PlaceholderData folderPlaceholder in placeholderFoldersListCopy.OrderByDescending(x => x.Path))
                     {
-                        // Remove folder placeholders before re-expansion to ensure that projection changes that convert a folder to a file work
-                        // properly
-                        if (!this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder, addPlaceholderToUpdatedPlaceholders, folderPlaceholdersToKeep))
+                        bool keepFolder = true;
+                        if (!folderPlaceholdersToKeep.Contains(folderPlaceholder.Path))
                         {
+                            this.GetChildNameAndParentKey(folderPlaceholder.Path, out string childName, out string parentKey);
+                            FolderData folderData = this.GetProjectedFolderEntryData(CancellationToken.None, blobSizesConnection, availableSizes, childName, parentKey, out string _) as FolderData;
+
+                            // Check the projection for the folder to determine if the folder needs to be deleted
+                            // The delete will be attempted if one of the following is true
+                            // 1. not in the projection anymore
+                            // 2. in the projection but is not a folder in the projection
+                            // 3. Folder no longer exist on disk - this is to remove the tombstone
+                            if (folderData == null ||
+                                !folderData.IsFolder ||
+                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path)))
+                            {
+                                keepFolder = !this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder);
+                                ++deleteFolderPlaceholderAttempted;
+                            }
+                        }
+
+                        if (keepFolder)
+                        {
+                            addPlaceholderToUpdatedPlaceholders(folderPlaceholder);
+
+                            // Remove folder placeholders before re-expansion to ensure that projection changes that convert a folder to a file work
+                            // properly
                             if (GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories && folderPlaceholder.IsExpandedFolder)
                             {
                                 if (updatedPlaceholderDictionary == null)
@@ -1173,10 +1196,6 @@ namespace GVFS.Virtualization.Projection
 
                                 this.ReExpandFolder(blobSizesConnection, folderPlaceholder.Path, updatedPlaceholderDictionary, folderPlaceholders);
                             }
-                        }
-                        else
-                        {
-                            ++deleteFolderPlaceholderAttempted;
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1157,12 +1157,35 @@ private void UpdatePlaceholders()
...
                     {
-                        // Remove folder placeholders before re-expansion to ensure that projection changes that convert a folder to a file work
-                        // properly
-                        if (!this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder, addPlaceholderToUpdatedPlaceholders, folderPlaceholdersToKeep))
+                        bool keepFolder = true;
+                        if (!folderPlaceholdersToKeep.Contains(folderPlaceholder.Path))
                         {
+                            this.GetChildNameAndParentKey(folderPlaceholder.Path, out string childName, out string parentKey);
+                            FolderData folderData = this.GetProjectedFolderEntryData(CancellationToken.None, blobSizesConnection, availableSizes, childName, parentKey, out string _) as FolderData;
+
```
wilbaker: It looks like `GetProjectedFolderEntryData` may attempt to download size(s) if the path we pass in is for a file.

Should/could we do one of the following instead:

- Use `IsPathProjected` here instead of `GetChildNameAndParentKey` + `GetProjectedFolderEntryData`
- Pass in `null` for `blobSizesConnection` to prevent `GetProjectedFolderEntryData` from downloading sizes

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1157,12 +1157,35 @@ private void UpdatePlaceholders()
...
+                            // 1. not in the projection anymore
+                            // 2. in the projection but is not a folder in the projection
+                            // 3. Folder no longer exist on disk - this is to remove the tombstone
+                            if (folderData == null ||
+                                !folderData.IsFolder ||
+                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path)))
+                            {
+                                keepFolder = !this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder);
+                                ++deleteFolderPlaceholderAttempted;
+                            }
```
wilbaker: If `keepFolder` is true (after making it to this point), should we call `AddParentFoldersToListToKeep` for the parent(s) of the folder?

I think that would save us from projection+disk lookups for all ancestors of the folder (e.g. in the scenarios where there are a lot of folder placeholders but not many file placeholders).

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: Yeah looks like `IsPathProjected` would be best.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: I don't think that would hurt anything.

wilbaker: Do you think there would be a benefit? 

(I thought maybe there would be, but was hoping you'd double check my thoughts).

I was thinking if there was a directory tree like this with no file placeholders in it:

`src\a\b\c\d\e\f\g`

The current code would check each of these directories (assuming it decides to keep g):

`src\a\b\c\d\e\f\g`
`src\a\b\c\d\e\f`
`src\a\b\c\d\e`
`src\a\b\c\d`
`src\a\b\c`
`src\a\b`
`src\a`

And that we could skip all those checks by calling `AddParentFoldersToListToKeep`.

Let me know what you think.

kewillford: Okay let me think about the scenarios.  The code gets here and `keepFolder` is `true` when it is still in the projection and it is a folder and it exists on disk.  If any of those are not the case we will try to delete the folder and if we get an error or directory not empty when trying to delete, `keepFolder` will be true.

1. folder no longer in the projection
Attempt the delete
2. folder still in the projection as a folder and on disk
Can keep all parents
3. folder still in the projection but doesn't exist on disk
Attempt the delete
4. folder path still in the projection as a file
Attempt the delete

Attempting to delete the folder:
1. deleting folder gets directory not empty - `keepFolder` will be `true`
Can keep all parents because they will get not empty as well.
2. deleting folder gets error - `keepFolder` will be `true`
Keep the folder placeholder and assume deletes on parents will fail as well so skip them
3. deleting folder gets success or not found - `keepFolder` will be `false`
Need to continue deleting parents

So from my thinking we would be okay.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1157,12 +1157,35 @@ private void UpdatePlaceholders()
...
+                            FolderData folderData = this.GetProjectedFolderEntryData(CancellationToken.None, blobSizesConnection, availableSizes, childName, parentKey, out string _) as FolderData;
+
+                            // Check the projection for the folder to determine if the folder needs to be deleted
+                            // The delete will be attempted if one of the following is true
+                            // 1. not in the projection anymore
+                            // 2. in the projection but is not a folder in the projection
+                            // 3. Folder no longer exist on disk - this is to remove the tombstone
+                            if (folderData == null ||
+                                !folderData.IsFolder ||
+                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path)))
```
wilbaker: One other thought (that we discussed offline briefly) is that if `DirectoryExists` is the big perf hit, could we keep a list of all directories that have been deleted and check it before calling `DirectoryExists`? (in some ways similar to how we have the `newlyCreatedFileAndFolderPaths` collection)

As the code is now we don't care who/what deleted directories, we're only trying to remove possible tombstones and I would guess that deleting directories is a relatively infrequent operation.

kewillford: That is a good idea but would be more invasive of a change so I would like to do that in a separate PR.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1157,12 +1157,40 @@ private void UpdatePlaceholders()
                     //  1. Ensures child folders are deleted before their parents
                     //  2. Ensures that folders that have been deleted by git (but are still in the projection) are found before their
                     //     parent folder is re-expanded (only applies on platforms where EnumerationExpandsDirectories is true)
+                    Dictionary<string, long> availableSizes = new Dictionary<string, long>();
```
kewillford: Remove.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1157,12 +1157,40 @@ private void UpdatePlaceholders()
...
+                            bool isProjected = this.IsPathProjected(folderPlaceholder.Path, out string fileName, out bool isFolder);
+
+                            // Check the projection for the folder to determine if the folder needs to be deleted
+                            // The delete will be attempted if one of the following is true
+                            // 1. not in the projection anymore
+                            // 2. in the projection but is not a folder in the projection
+                            // 3. Folder no longer exist on disk - this is to remove the tombstone
+                            if (!isProjected ||
+                                !isFolder ||
+                                (!GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories &&
```
wilbaker: Out of curiosity, why did the `EnumerationExpandsDirectories` check need to be added?

I would've thought that the `DirectoryExists` check wouldn't matter on on Mac.

wilbaker: (I think it's the right change, I just wasn't sure why it would functionally break on Mac)

kewillford: I was trying to fix the `CheckoutBranchWithDirectoryNameSameAsFileWithWrite` test which is still failing in the tear down when validating the working directory.  Now there are 2 tests failing so this wasn't the right fix.  In the test it is checking out a branch with a file in file.txt/file.txt and edits that file then commits the change and checks out a branch with the file in the root at file.txt.  So it is the case of a folder changing to a file in the root directory and seems like the root directory is not re-expanding.

You wrote the expand on enumerate code so do you have any ideas what the issue might be?  Does it need to re-expand when folder changes to a file?  How and when was the root directory re-expanded in the previous code?

wilbaker: > Does it need to re-expand when folder changes to a file?

Yes it will need to re-expand.

Previously I would have expected that it works like this:

1. git deletes file.txt/file.txt
2. VFS4G deletes file.txt (directory)
3. VFS4G re-expands root directory
4. VFS4G writes file.txt

But so far I've not found how your changes would have altered this behavior.

When I was debugging issues like this before (when implementing the re-expand logic).  I found it helpful to add a bunch of logging in this method so that VFS4G would record exactly what folders it's deleting (or not) and record what it does when re-expanding.

I'll spend some more time looking at your change and see if anything jumps out at me.

wilbaker: > How and when was the root directory re-expanded in the previous code?

The root directory was expanded because it was in `placeholderFoldersListCopy`.  The expansion happened here:

`this.ReExpandFolder(blobSizesConnection, folderPlaceholder.Path, updatedPlaceholderDictionary, folderPlaceholders);`

On like 1202.

The root would have been expanded last as the folders are processed bottom up.

kewillford:
@jrbriggs The issue with adding it to the large repo perf tests is the amount of time it takes to create a sufficient number of folder placeholders to get performance numbers.  Running a `dir /s` on my machine would take minutes when there were not any sizes to download and hours when it had to download all the sizes.  I will try it on some of the main folders and see if there is one that will have enough folders but not take too long.

PR #1015, opened 2019-04-08
Test Coverage for ProcesswasSpawnedByRegularUser

jeschu1:
CurrentProcessWasSpawnedByRegularUser Test Coverage
- There were some intricacies getting this to work that I've highlighted in comments
- Glad I embarked on this task as it allowed me to fully understand the flow!

### DIFF
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -4,6 +4,8 @@
 #include "../PrjFSKext/VirtualizationRootsTestable.hpp"
 #import <XCTest/XCTest.h>
 #import <sys/stat.h>
+#include "KextLogMock.h"
+#include "KextMockUtilities.hpp"
 #include "MockVnodeAndMount.hpp"
 #include "MockProc.hpp"
 #include "VnodeCacheEntriesWrapper.hpp"
@@ -21,15 +23,21 @@ class org_vfsforgit_PrjFSProviderUserClient
 @implementation KauthHandlerTests
 {
     VnodeCacheEntriesWrapper cacheWrapper;
+    vfs_context_t context;
 }
 
 - (void) setUp {
     kern_return_t initResult = VirtualizationRoots_Init();
     XCTAssertEqual(initResult, KERN_SUCCESS);
     self->cacheWrapper.AllocateCache();
+    context = vfs_context_create(nullptr);
+    MockProcess_AddContext(context, 501 /*pid*/);
+    MockProcess_SetSelfPid(501);
+    MockProcess_AddProcess(501 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "test" /*name*/);
 }
 
 - (void) tearDown {
+    MockProcess_Reset();
     self->cacheWrapper.FreeCache();
     MockVnodes_CheckAndClear();
     VirtualizationRoots_Cleanup();
@@ -247,7 +255,10 @@ class org_vfsforgit_PrjFSProviderUserClient
 
     // Test with file crawler trying to populate an empty file
     testVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
-    SetProcName("mds");
+    MockProcess_Reset();
+    MockProcess_SetSelfPid(501);
+    MockProcess_AddContext(context, 501 /*pid*/);
+    MockProcess_AddProcess(501 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "mds" /*name*/);
     XCTAssertFalse(
         ShouldHandleVnodeOpEvent(
             &perfTracer,
@@ -261,10 +272,12 @@ class org_vfsforgit_PrjFSProviderUserClient
             &kauthResult,
             &kauthError));
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DENY);
-
     
     // Test with finder trying to populate an empty file
-    SetProcName("Finder");
+    MockProcess_Reset();
+    MockProcess_SetSelfPid(501);
+    MockProcess_AddContext(context, 501 /*pid*/);
+    MockProcess_AddProcess(501 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "Finder" /*name*/);
     XCTAssertTrue(
         ShouldHandleVnodeOpEvent(
             &perfTracer,
@@ -278,6 +291,49 @@ class org_vfsforgit_PrjFSProviderUserClient
             &kauthResult,
             &kauthError));
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
+}
+
+- (void)testCurrentProcessWasSpawnedByRegularUser {
+    // Defaults should pass for all tests
+    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
+    MockProcess_Reset();
+
+    // Process is a service user and does not have a parent
+    MockProcess_AddContext(context, 500 /*pid*/);
+    MockProcess_SetSelfPid(500);
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddProcess(500 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
+    MockProcess_Reset();
+
+    // Test a process with a service UID, valid parent pid, but proc_find fails to find parent pid
+    MockCalls::Clear();
+    MockProcess_AddContext(context, 500 /*pid*/);
+    MockProcess_SetSelfPid(500);
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddProcess(500 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
+    XCTAssertTrue(MockCalls::DidCallFunction(KextMessageLogged, KEXTLOG_ERROR));
+    MockProcess_Reset();
+
+    // 'sudo' scenario: Root process with non-root parent
+    MockProcess_AddContext(context, 502 /*pid*/);
+    MockProcess_SetSelfPid(502);
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddCredential(2 /*credentialId*/, 501 /*UID*/);
+    MockProcess_AddProcess(502 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    MockProcess_AddProcess(501 /*pid*/, 2 /*credentialId*/, 1 /*ppid*/, "test" /*name*/);
+    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
+    MockProcess_Reset();
+
+    // Process and it's parent are service users
+    MockProcess_AddContext(context, 502 /*pid*/);
+    MockProcess_SetSelfPid(502);
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddCredential(2 /*credentialId*/, 2 /*UID*/);
+    MockProcess_AddProcess(502 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    MockProcess_AddProcess(501 /*pid*/, 2 /*credentialId*/, 1 /*ppid*/, "test" /*name*/);
+    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
 }
 
 - (void)testUseMainForkIfNamedStream {
@@ -303,7 +359,6 @@ class org_vfsforgit_PrjFSProviderUserClient
 
 - (void)testShouldHandleFileOpEvent {
     PerfTracer perfTracer;
-    vfs_context_t _Nonnull context = vfs_context_create(nullptr);
     shared_ptr<mount> testMount = mount::Create();
     std::string repoPath = "/Users/test/code/Repo";
     shared_ptr<vnode> repoRootVnode = testMount->CreateVnodeTree(repoPath, VDIR);
--- a/ProjFS.Mac/PrjFSKextTests/MockProc.cpp
+++ b/ProjFS.Mac/PrjFSKextTests/MockProc.cpp
@@ -1,56 +1,130 @@
 #include "MockProc.hpp"
+#include <map>
 #include <string>
 
-static std::string s_procName;
+using std::make_pair;
+using std::map;
+using std::string;
 
-int proc_pid(proc_t)
+static map<uintptr_t /*credential ID*/, int /*UID*/> s_credentialMap;
+static map<vfs_context_t /*context*/, int /*pid*/> s_contextMap;
+static map<int /*process Id*/, proc> s_processMap;
+static int s_selfPid;
+
+void MockProcess_Reset()
 {
-    return 1;
+    s_processMap.clear();
+    s_credentialMap.clear();
+    s_contextMap.clear();
+}
+
+void MockProcess_SetSelfPid(int selfPid)
+{
+    s_selfPid = selfPid;
+}
+
+int proc_pid(proc_t proc)
+{
+    return proc->pid;
 }
 
 void proc_name(int pid, char* buf, int size)
 {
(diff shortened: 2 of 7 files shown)


ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -2,22 +2,50 @@
...
 {
-    return 1;
+    s_processPid = 1;
+    s_proc_find_result = nullptr;
+    s_kauth_cred_getuid_count = 0;
+    s_servicePidOnFirstCall = false;
 }
 
-void proc_name(int pid, char* buf, int size)
+void ReturnServicePidOnFirstCall()
```
jeschu1: This was the easiest way to get a service pic on the first (and only the first) call.

jeschu1:
@pmj / @wilbaker for CurrentProcessWasSpawnedByRegularUser it looks like we ignore processes whos parent process is null or <=1.  Is it worth logging anything there, or is this an expected case where we want to hault.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
wilbaker: I had some trouble following how the mock process code works, as well as understanding exactly what these helpers do and when to call them:

```
void ResetMockProc();
void SetProcName(const std::string& procName);
void SetPpid(int pid);
void SetDummyProcFind();
void ReturnServicePidOnFirstCall();
```

What would you think if rather than controlling the results of each mock function we built up a collection of mock processes/credentials (similar to what we're doing for mock vnodes)?

The interface could then be something like:

```
MockProc_AddCredential(int credentialId, uid_t UID);
MockProc_AddProcess(int pid, int credentialId, int ppid);
MockProc_SetSelfPid(int pid); // Optional, we could also have a well known SELF_PID value
MockProc_Reset();
```

And the test would look like:

```
    // Fail the test for a service pid (processUID >= 500), and get a non valid Parent Pid ( pid <= 1)
    MockProc_AddCredential(1 /*credentialId*/, 1 /*UID*/);
    MockProc_AddProcess(5 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/);
    MockProc_SetSelfPid(5);
    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
    MockProc_Reset();

...

   // Fail the test for a service pid (processUID >= 500), and get a valid Parent  Pid / Process with a non-service Pid
    MockProc_AddCredential(1 /*credentialId*/, 1 /*UID*/);
    MockProc_AddCredential(2 /*credentialId*/, 501 /*UID*/);
    MockProc_AddProcess(5 /*pid*/, 1 /*credentialId*/, 6 /*ppid*/); 
    MockProc_AddProcess(6 /*pid*/, 2 /*credentialId*/, 0 /*ppid*/);
    MockProc_SetSelfPid(5);
    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
    MockProc_Reset();
```

Which would also allow for larger process hierarchies:

```
    MockProc_AddCredential(1 /*credentialId*/, 1 /*UID*/);
    MockProc_AddCredential(2 /*credentialId*/, 501 /*UID*/);

    // Force CurrentProcessWasSpawnedByRegularUser to walk several levels
    MockProc_AddProcess(5 /*pid*/, 1 /*credentialId*/, 6 /*ppid*/); 
    MockProc_AddProcess(6 /*pid*/, 1 /*credentialId*/, 7 /*ppid*/);
    MockProc_AddProcess(7 /*pid*/, 1 /*credentialId*/, 8 /*ppid*/);
    MockProc_AddProcess(8 /*pid*/, 2 /*credentialId*/, 0 /*ppid*/);
    MockProc_SetSelfPid(5);
    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
    MockProc_Reset();
```

Internally we could just keep collections of the credentials and processes:

```
struct proc
{
    int pid;
    int credentialId;
    int ppid;
}

static map<int /*credential ID*/, int /* UID */> credentialMap;
static map<int /*process ID*/, proc> processMap;
static int selfPid;
```

wilbaker: `MockProc_AddProcess(int pid, int credentialId, int ppid);` could have another parameter for process name as well

jeschu1: I agree that's clearer I'll update accordingly.

wilbaker:
> @pmj / @wilbaker for CurrentProcessWasSpawnedByRegularUser it looks like we ignore processes whos parent process is <=1. Is it worth logging anything there, or is this an expected case where we want to hault.

@jeschu1 I found this comment from @pmj in the PR that added this function:

> The ultimate parent is, as far as I'm aware, always launchd (PID 1), which runs as root (UID 0). Does that answer the question? If not, please elaborate!

(see discussion here https://github.com/Microsoft/VFSForGit/pull/798#discussion_r259547764)

And so as best I can tell this is an expected case, but perhaps we should log if the PID is less than 0? (@pmj is more of the expert here)

jeschu1:
> And so as best I can tell this is an expected case, but perhaps we should log if the PID is less than 0? (@pmj is more of the expert here)

Thats what I was wondering!  We are talking about a case that should never happen.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
jeschu1: @wilbaker I've updated per your suggestion please let me know what you think!

ProjFS.Mac/PrjFSKextTests/MockProc.hpp
```diff
@@ -17,4 +21,17 @@ extern "C"
...
+struct proc {
+    int pid;
+    uintptr_t credentialId;
+    int ppid;
+    std::string name;
+};
+
+static map<uintptr_t /*credential ID*/, int /* UID */> credentialMap;
+static map<int /*process Id*/, proc> processMap;
+static int selfPid;
```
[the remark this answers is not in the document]
wilbaker: (or can we start setting/using it ... see discussion in `proc_self`)

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -10,12 +17,14 @@ int proc_pid(proc_t)
 
 void proc_name(int pid, char* buf, int size)
 {
-    strlcpy(buf, s_procName.c_str(), size);
-}
+    // For now this returns the first entry in the map.  If we need to get fancier in the future we can
```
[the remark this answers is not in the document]
jeschu1: @wilbaker, the issue is we'd have to alter a bunch of other functions in the chain so the right pid gets called in here.  IMHO it's not worth doing at this time, but we certainly have that as an option for the future.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -261,10 +261,10 @@ - (void)testShouldHandleVnodeOpEvent {
             &kauthResult,
             &kauthError));
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DENY);
-
+    MockProcess_Reset();
```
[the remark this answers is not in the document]
jeschu1: We don't I was keeping it consistent.  Perhaps it's confusing like that, i can remove it.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -278,6 +278,33 @@ - (void)testShouldHandleVnodeOpEvent {
             &kauthResult,
             &kauthError));
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
+    MockProcess_Reset();
+}
+
+- (void)testCurrentProcessWasSpawnedByRegularUser {
+    // Defaults should pass for all tests
+    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
```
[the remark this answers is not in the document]
jeschu1: With the defaults the test passed.  This basically verifies the defaults pass.

I can add a test with proc values we know should pass (and match the defaults) as a second test.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -25,32 +34,81 @@ proc_t vfs_context_proc(vfs_context_t ctx)
 
 proc_t proc_self(void)
 {
-    return nullptr;
+    // Since this function takes no parameters, it will return the first entry in the map, if one exists
```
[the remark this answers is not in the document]
jeschu1: Since it's an ordered map, it's not arbitrary.  The user would have added it in the test order.  

We do need the ability to change this value for testing.

I don't love the SetSelfPid option either honestly.  We could do it, if we think it adds clarity.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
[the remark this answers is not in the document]
jeschu1: Since we alter the map above (AddProcess), resetting it here.  The test below will setup again.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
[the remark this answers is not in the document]
jeschu1: The required 'adds' are now added as part of setup

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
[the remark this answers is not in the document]
jeschu1: Updated so eventhing is done thru 'add's and we lookup in a map

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -25,32 +34,81 @@ proc_t vfs_context_proc(vfs_context_t ctx)
...
 }
 
 uid_t kauth_cred_getuid(kauth_cred_t _cred)
 {
-    // Values over 500 are non-system processes
-    return 501;
+    map<uintptr_t /*credential ID*/, int /* UID */>::const_iterator found = credentialMap.find(reinterpret_cast<uintptr_t>(_cred));
+    if (found == credentialMap.end())
+    {
+        return 501;
```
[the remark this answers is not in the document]
jeschu1: I've added asserts for that case.

jeschu1:
@wilbaker this is ready for another look.

There were so many changes I did merge the commits to make sure I didn't have unnecessary additions.  Hopefully still easy to read thru.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -1,56 +1,130 @@
...
 {
-    return nullptr;
+    map<vfs_context_t /*context ID*/, int /* UID */>::const_iterator foundPidId = s_contextMap.find(ctx);
+    if (foundPidId == s_contextMap.end())
+    {
+        assert("ctx not found in s_contextMap");
+        return nullptr;
+    }
+
+    return proc_find(foundPidId->second);
```
wilbaker: `proc_find` takes a `pid`, but it looks like we're passing in a `UID`.  Should `s_contextMap` be a map from `vfs_context_t` to `PID`?

(This comment is just about naming and comments, it looks like we're only using `s_contextMap` here)

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -1,56 +1,130 @@
...
 }
 
-void SetProcName(const std::string& procName)
+void proc_name(int pid, char* buf, int size)
 {
-    s_procName = procName;
+    map<int /*process Id*/, proc>::const_iterator found = s_processMap.find(pid);
+    if (found == s_processMap.end())
+    {
+        assert("pid not found in s_processMap");
```
wilbaker: Out of curiosity, do these asserts do anything if they fire during a test?

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
wilbaker: Actually, I see that in the tests we refer to the second parameter as "pid":

```
   MockProcess_AddContext(context, 501 /*pid*/);
```

And so it's probably just this file and the header that need to be updated.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -280,6 +293,49 @@ - (void)testShouldHandleVnodeOpEvent {
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
 }
 
+- (void)testCurrentProcessWasSpawnedByRegularUser {
+    // Defaults should pass for all tests
+    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
+    MockProcess_Reset();
+
+    // Process tree with all UIDs less than 500 were not spawned by regular users
```
wilbaker: Should we have more than one process in this tree?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -280,6 +293,49 @@ - (void)testShouldHandleVnodeOpEvent {
...
+    MockProcess_SetSelfPid(1);
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddProcess(1 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "test" /*name*/);
+    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
+    MockProcess_Reset();
+
+    // Test a process with a service UID, valid parent pid, but proc_find fails to find parent pid
+    MockCalls::Clear();
+    MockProcess_AddContext(context, 1 /*pid*/);
+    MockProcess_SetSelfPid(1);
```
wilbaker: Consider using something other than 1 for PID here.  My understanding is 1 is typically the root of all process trees, and so the intent of the test might be clearer if you used some other random PID instead (with a parent of 501).

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -280,6 +293,49 @@ - (void)testShouldHandleVnodeOpEvent {
...
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddProcess(1 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
+    XCTAssertTrue(MockCalls::DidCallFunction(KextMessageLogged, KEXTLOG_ERROR));
+    MockProcess_Reset();
+
+    // 'sudo' scenario: Root process with non-root parent
+    MockProcess_AddContext(context, 1 /*pid*/);
+    MockProcess_SetSelfPid(1);
```
wilbaker: Same comment here:  Consider making the process tree with 1 as the top-most parent:

```
   1 (UID: 1)
   |
  501 (UID: 501)
   |
  502 (self, UID: 1)

File: ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm [hunk 13]
```diff
@@ -280,6 +293,49 @@ - (void)testShouldHandleVnodeOpEvent {
...
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddCredential(2 /*credentialId*/, 501 /*UID*/);    
+    MockProcess_AddProcess(1 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    MockProcess_AddProcess(501 /*pid*/, 2 /*credentialId*/, 502 /*ppid*/, "test" /*name*/);
+    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
+    MockProcess_Reset();
+
+    // Process and it's parent are service users
```
wilbaker: Nit:  In this comment you use the term "service users" and in the above comment you use the term "non-root parent".  Could you make these consistently one way or the other ("service" is the term used the code so maybe use that here too)?

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
jeschu1: Doesn't look like it :-(, but we have a bunch of them in MockVnodeAndMount.  @pmj were you able to have them turned on during debugging?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
jeschu1: I'll update the comment.  This should test for process wiht Id under 500, not having a parent

PR #1016, opened 2019-04-08
Test Coverage for UseMainForkIfNamedStream

jeschu1:
Add test coverage for UseMainForkIfNamedStream

### DIFF
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -263,4 +263,25 @@ using std::shared_ptr;
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
 }
 
+- (void)testUseMainForkIfNamedStream {
+    shared_ptr<mount> testMount = mount::Create();
+    const char* filePath = "/Users/test/code/Repo/file";
+    shared_ptr<vnode> testFileVnode = testMount->CreateVnodeTree(filePath);
+
+    // Out Parameters to test
+    bool putVnodeWhenDone;
+    vnode_t testVnode = testFileVnode.get();
+
+    testVnode->namedStream = false;
+    UseMainForkIfNamedStream(testVnode, putVnodeWhenDone);
+    XCTAssertFalse(putVnodeWhenDone);
+    XCTAssertTrue(testVnode == testFileVnode.get());
+
+    testVnode->namedStream = true;
+    UseMainForkIfNamedStream(testVnode, putVnodeWhenDone);
+    XCTAssertTrue(putVnodeWhenDone);
+    XCTAssertTrue(testVnode == testFileVnode->GetParentVnode().get());
+    vnode_put(testVnode);
+}
+
 @end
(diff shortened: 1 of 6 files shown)


ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -263,4 +263,25 @@ - (void)testShouldHandleVnodeOpEvent {
     XCTAssertEqual(kauthResult, KAUTH_RESULT_DEFER);
 }
 
+//KEXT_STATIC void UseMainForkIfNamedStream(vnode_t& vnode, bool& putVnodeWhenDone);
```
wilbaker: Minor: We don't have comments like this:

```//KEXT_STATIC void UseMainForkIfNamedStream(vnode_t& vnode, bool& putVnodeWhenDone);```

On the other tests.  I think it can probably be removed.

File: ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm [hunk 2]
```diff
@@ -263,4 +263,25 @@ - (void)testShouldHandleVnodeOpEvent {
...
+- (void)testUseMainForkIfNamedStream {
+    shared_ptr<mount> testMount = mount::Create();
+    const char* filePath = "/Users/test/code/Repo/file";
+    shared_ptr<vnode> testFileVnode = testMount->CreateVnodeTree(filePath);
+
+    // Out Parameters to test
+    bool result;
```
wilbaker: Naming suggestion: `result` made me think this value represents if the call succeeded or failed, when it actually corresponds to the `putVnodeWhenDone` parameter.

Maybe rename `result ` to `putVnodeWhenDone`?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -263,4 +263,25 @@ - (void)testShouldHandleVnodeOpEvent {
...
+    vnode_t testVnode = testFileVnode.get();
+
+    testVnode->namedStream = false;
+    UseMainForkIfNamedStream(testVnode, result);
+    XCTAssertFalse(result);
```
wilbaker: Also `XCTAssertTrue` that `testVnode` equals `testFileVnode.get()`?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -263,4 +263,25 @@ - (void)testShouldHandleVnodeOpEvent {
...
+    
+    testVnode->namedStream = true;
+    UseMainForkIfNamedStream(testVnode, result);
```
wilbaker: I think this test is good for now, but I'd be interested to hear if @pmj has any ideas on how to better mock named streams.

As it is now this call to `UseMainForkIfNamedStream` returns the parent directory vnode ("/Users/test/code/Repo") while in production I would not expect `UseMainForkIfNamedStream` to ever return a directory vnode.

jeschu1: Under what circumstances would `UseMainForkIfNamedStream `.  Would a link do it?

wilbaker: I believe that on the real file system this would happen for named/resource forks (I saw a discussion on how to create them here: https://apple.stackexchange.com/questions/228444/how-do-i-create-a-named-fork-and-store-data-in-it).  

There were also a few (somewhat vague) details here: https://support.apple.com/en-us/HT202021

@nickgra, do you have any tricks for created files that use named streams?

PR #1017, opened 2019-04-08
POSIX: Change default location of .gvfsCache folder from /.gvfsCache to ~/.gvfsCache

### LINKED ISSUE #519: Mac: Determine a better default location for the local cache
Currently we try to place the cache at /.gvfsCache. It should go to inside the user's home folder instead.

wilbaker:
Resolves #519 

Update the behavior of `gvfs clone` so that on Mac the default location of the .gvfsCache directory is ~/.gvfsCache instead of /.gvfsCache.  The default location on Windows is not impacted by these changes.


GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -124,6 +124,30 @@ public override bool TryGetGVFSEnlistmentRoot(string directory, out string enlis
             return POSIXPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
         }
 
+        public override bool TryGetDefaultLocalCacheRoot(string enlistmentRoot, out string localCacheRoot, out string localCacheRootError)
+        {
+            string homeDirectory = Environment.GetEnvironmentVariable("HOME");
```
jrbriggs: This *could* throw a `SecurityException` if user doesn't have permissions to perform the operation.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -124,6 +124,30 @@ public override bool TryGetGVFSEnlistmentRoot(string directory, out string enlis
             return POSIXPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
         }
 
+        public override bool TryGetDefaultLocalCacheRoot(string enlistmentRoot, out string localCacheRoot, out string localCacheRootError)
+        {
+            string homeDirectory = Environment.GetEnvironmentVariable("HOME");
+            if (string.IsNullOrEmpty(homeDirectory))
+            {
+                localCacheRoot = null;
+                localCacheRootError = "Failed to read HOME environment variable";
```
jrbriggs: I think this message is correct when we catch the `SecurityException`, but is it correct if `HOME` is unset (for whatever reason). Should we distinguish between the two failure cases?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -325,6 +325,14 @@ public override bool TryGetGVFSEnlistmentRoot(string directory, out string enlis
             return WindowsPlatform.TryGetGVFSEnlistmentRootImplementation(directory, out enlistmentRoot, out errorMessage);
         }
 
+        public override bool TryGetDefaultLocalCacheRoot(string enlistmentRoot, out string localCacheRoot, out string localCacheRootError)
+        {
+            string pathRoot = Path.GetPathRoot(enlistmentRoot);
```
jrbriggs: Should the `Path.GetPathRoot` and `Path.Combine` calls be in a  `try { } catch (ArgumentException e) { }` block?

GVFS/GVFS/CommandLine/VerbConstants.POSIX.cs
```diff
@@ -0,0 +1,10 @@
+namespace GVFS.CommandLine
+{
+    public static class VerbConstants
+    {
+        public static class CloneVerb
+        {
+            public const string LocalCacheRootHelpText = "Use this option to override the path for the local GVFS cache. The default location is ~/.gvfsCache";
+        }
+    }
+}
```
jrbriggs: nit: newline at EOF

GVFS/GVFS/CommandLine/VerbConstants.Windows.cs
```diff
@@ -0,0 +1,10 @@
+namespace GVFS.CommandLine
+{
+    public static class VerbConstants
+    {
+        public static class CloneVerb
+        {
+            public const string LocalCacheRootHelpText = "Use this option to override the path for the local GVFS cache. The default location is the .gvfsCache folder in the root of the volume.";
+        }
+    }
+}
```
jrbriggs: nit: newline at EOF

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
kewillford: `HOME` seems like a really generic name for an environment variable should it be more specific?

GVFS/GVFS/CommandLine/VerbConstants.POSIX.cs
```diff
@@ -0,0 +1,10 @@
+namespace GVFS.CommandLine
+{
+    public static class VerbConstants
```
kewillford: Is there anyway we can add this to the GVFSPlatform instead since that it where differences in the platforms are?  I would hate to start having `.POSIX.cs`, `.Mac.cs`, `.Unix.cs`, and `.Windows.cs` files all over the code base.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
wilbaker: `HOME` is the environment variable that stores the location of the home directory on POSIX systems.  According to [this post](https://unix.stackexchange.com/questions/123858/is-the-home-environment-variable-always-set-on-a-linux-system) it will always be set.

Do you think a comment here would be helpful?

GVFS/GVFS/CommandLine/VerbConstants.POSIX.cs
wilbaker: > Is there anyway we can add this to the GVFSPlatform instead since that it where differences in the platforms are?

I don't think this is possible because we need this for an attribute that is set at compile time:

```
[Option(
  "local-cache-path",
   Required = false,
   HelpText = VerbConstants.CloneVerb.LocalCacheRootHelpText)]
```

The only alternative that comes to mind is that rather than adding the sources files in this folder I could add them in the `GVFS.Platform.Windows` and `GVFS.Platform.POSIX` folders, but the csprojs here would need to link the files in directly (similar to how GVFS.Hooks links in shared files).

Do you think that would be cleaner?

@nickgra, do you have any suggestions for cleaning this up?

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
wilbaker: Good point, I will catch that and set an appropriate error message.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
wilbaker: > Should we distinguish between the two failure cases?

I think it's best to provide as much detail as we can in the error message, and so I will separate out the failure cases.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
wilbaker: That's a good idea.  We weren't doing that before, but I think it's better to be consistent with the POSIX implementation.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
wilbaker: One other detail, when you're using the terminal [~ is expanded to $HOME by Bash](https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html) and so we can't use `~` in the path ourselves.

kewillford: Are there any cases when the user would want to keep the `HOME` environment variable and have something else for the default VFSForGit cache root or if a user wants to do that they have to specify the cache location when cloning?

GVFS/GVFS/CommandLine/VerbConstants.POSIX.cs
kewillford: I just seems to me like we are ending up with multiple ways to have platform specific code but maybe that is just me and there is some standard convention we are following?

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
wilbaker: > Are there any cases when the user would want to keep the HOME environment variable and have something else for the default VFSForGit cache root or if a user wants to do that they have to specify the cache location when cloning?

With these changes if the user wants the cache somewhere other than `~/.gvfsCache` they'd need to specify it when cloning.

We could consider adding our own environment variable to control this (e.g. `VFS4G_CLONE_CACHE_LOCATION`) but that work hasn't been done yet.

@jrbriggs let me know if you'd like me to file an issue for that feature and/or make that change in this PR.

GVFS/GVFS/CommandLine/VerbConstants.POSIX.cs
wilbaker: The general approach we've been using:

- Compile time dependency (e.g. this PR) -> Include platform specific code in csproj
- Project cannot consume GVFS.Common assembly (e.g. GVFS.Hooks) -> Include platform specific code in csproj
- Everything else -> Use GVFSPlatform

@nickgra please double check me here.

I think it's a little more gray in the service work that @alameenshah is working on.  Based on the approach described above we probably should be using GVFSPlatform there for the platform-specific code.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
jrbriggs: Let's treat this PR as the changing of the default. If we get user feedback we can add our own environment variable to control the cache location.

GVFS/GVFS/CommandLine/VerbConstants.POSIX.cs
nickgra: I tried some things locally which didn't work out so I don't have any ideas or suggestions on ways to clean this up.

I agree with the theme behind Kevin's comment, things are getting complicated with multiple ways to do platform specific code. I believe this is the first PR in which we've implemented a compile time dependency like this (FastFetch does something similar, but it's a different program), but I don't think we have an option with attributes. Maybe there's another way to specify arguments for CommandLineParser that doesn't use attributes?

Long term, I think we want to see the Hooks go away so that would remove that method.

nickgra: Another (cheap) option would be to put both the default for Windows and the default for POSIX (macOS and Linux) in a string that is shared across both platforms. I don't think that's too poor of a user experience.

wilbaker: > Another (cheap) option would be to put both the default for Windows and the default for POSIX (macOS and Linux) in a string that is shared across both platforms. I don't think that's too poor of a user experience.

I didn't initially go with that option because of UX concerns.

@jrbriggs what are your thoughts on this option?

wilbaker: \+ @kewillford for his thoughts on this option as well

jrbriggs: A crazy idea I had was `<DefineConstants>` in the csproj per-platform, and then just ifdef the attribute. Not sure that's much better though :)

wilbaker: Yeah I had considered the `#ifdef` route as well, but as a general rule we've been trying to avoid adding `#ifdef` ... looks like there is no perfect option here 🙁

wilbaker: @jrbriggs do you if is possible to extract the value of project constants in the source code?

If that was the case then we could add these strings to the csprojs themselves and not need to add new source files for them.

jrbriggs: C# doesn't support preprocessor-like operations. While it can check whether or not a token was defined, it can't do any replacements like it's C bretheren.

kewillford: Yeah looks like there isn't really a good option.  If the `CommandLineParser` supported localization of the HelpText we could use that but it doesn't.  How important is showing the default location is the question because without that we could have the same text for both platforms.

wilbaker: > How important is showing the default location is the question because without that we could have the same text for both platforms.

That's a good point, @kewillford, @jrbriggs, @nickgra what would you think if I just shortened the help text to:

> "Use this option to override the default path for the local GVFS cache."

I doubt that users have been relying on the help text to know where the default location is.

nickgra: > "Use this option to override the default path for the local GVFS cache."

I like this suggestion!

jrbriggs: I like it.

wilbaker: Sounds good, thanks everyone! I will clean this up.

PR #1018, opened 2019-04-09
Mac: Check the return value of stat in PrjFS_DeleteFile

### LINKED ISSUE #1019: Mac: PrjFS_DeleteFile does not check if stat was successful
`PrjFS_DeleteFile` is not checking if its call to `stat` was successful, and this can result in `PrjFS_DeleteFile` returning that a delete failed due to dirty data, when in fact the file/directory did not exist.

wilbaker:
Fixes #1019

`PrjFS_DeleteFile` is incorrectly returning `PrjFS_Result_EVirtualizationInvalidOperation` when it's called for files/directories that do not exist.

This results in a lot of misleading messages being recorded in the GVFS mount process log files.

_Example log message_

>[2019-04-08 13:18:54 -07:00] RemoveFolderPlaceholderIfEmpty_DeleteFileFailure {"Area":"GitIndexProjection","Folder Path":"Test_EPF_GitCommandsTestOnlyFileFolder","result.Result":"VirtualizationInvalidOperation","result.RawResult":536871936,"UpdateFailureCause":"DirtyData"}

Prior to this PR the functional tests recorded `RemoveFolderPlaceholderIfEmpty_DeleteFileFailure` 129 times.  With this change the message is not recorded at all during the functional test runs.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -549,7 +549,18 @@ PrjFS_Result PrjFS_DeleteFile(
     CombinePaths(s_virtualizationRootFullPath.c_str(), relativePath, fullPath);
     
     struct stat path_stat;
-    stat(fullPath, &path_stat);
+    if (0 != stat(fullPath, &path_stat))
+    {
+        switch(errno)
+        {
+            case ENOENT:  // A component of fullPath does not exist
+            case ENOTDIR: // A component of fullPath is not a directory
+                return PrjFS_Result_Success;
+            default:
+                return PrjFS_Result_EIOError;
+        }
+    }
+    
     if (!(S_ISREG(path_stat.st_mode) || S_ISDIR(path_stat.st_mode)))
     {
         // Only files and directories can be deleted with PrjFS_DeleteFile


wilbaker:
@kewillford I noticed this issue when looking at the logs for your #1014 PR.

I noticed this was getting logged a lot in your test run (as well as successful test runs) and the noise of this getting logged all of the time made it difficult to understand if there was a problem.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -549,7 +549,18 @@ PrjFS_Result PrjFS_DeleteFile(
     CombinePaths(s_virtualizationRootFullPath.c_str(), relativePath, fullPath);
     
     struct stat path_stat;
-    stat(fullPath, &path_stat);
+    if (0 != stat(fullPath, &path_stat))
+    {
+        switch(errno)
+        {
+            case ENOENT:  // A component of fullPath does not exist
+            case ENOTDIR: // A component of fullPath is not a directory
```
kewillford: Do you know when it would hit this case? How does that stat know that you are looking for a directory?  What if this was called for a file?  I don't think in VFSForGit we ever do try to delete a file though.

wilbaker: > Do you know when it would hit this case?

I would expect the `ENOTDIR` case to be hit if you called `stat` on a path like /a/b/c/d.txt and a/b/c is a file.

> How does that stat know that you are looking for a directory?
> What if this was called for a file?

`stat` doesn't know what you're looking for, these errors indicate that the file that you're looking for is not on disk (either because part of the path does not exist, or part of the path is a file).

> I don't think in VFSForGit we ever do try to delete a file though.

In VFS4G we delete files when changing to a projection where some of the file placeholders no longer exist.

PR #1023, opened 2019-04-10
Testable product upgrader

jamill:
One area that I will be working on is which `ProductUpgrader` implementation to choose based on the GVFS config settings.

This is some prepatory changes in order to enable testing of this area.

### DIFF
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -4,6 +4,7 @@ using GVFS.Common.Tracing;
 using NuGet.Protocol.Core.Types;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
 using System.Text.RegularExpressions;
@@ -96,6 +97,7 @@ namespace GVFS.Common.NuGetUpgrade
             ITracer tracer,
             PhysicalFileSystem fileSystem,
             LocalGVFSConfig gvfsConfig,
+            ICredentialStore credentialStore,
             bool dryRun,
             bool noVerify,
             out NuGetUpgrader nuGetUpgrader,
@@ -124,15 +126,6 @@ namespace GVFS.Common.NuGetUpgrade
             {
                 return false;
             }
-
-            string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
-            if (string.IsNullOrEmpty(gitBinPath))
-            {
-                error = $"NuGetUpgrader: Unable to locate git installation. Ensure git is installed and try again.";
-                return false;
-            }
-
-            ICredentialStore credentialStore = new GitProcess(gitBinPath, workingDirectoryRoot: null, gvfsHooksRoot: null);
 
             nuGetUpgrader = new NuGetUpgrader(
                 ProcessHelper.GetCurrentProcessVersion(),
@@ -598,6 +591,13 @@ namespace GVFS.Common.NuGetUpgrade
         {
             if (!this.isNuGetFeedInitialized)
             {
+                Debug.Assert(this.credentialStore != null, "Attempting to access nuget feed without a credential store configured.");
+                if (this.credentialStore == null)
+                {
+                    error = "Program error - no Credential Store configured";
+                    return false;
+                }
+
                 string authUrl;
                 if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
                 {
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -95,13 +95,14 @@ namespace GVFS.Common.NuGetUpgrade
         public static bool TryCreate(
             ITracer tracer,
             PhysicalFileSystem fileSystem,
+            LocalGVFSConfig gvfsConfig,
             bool dryRun,
             bool noVerify,
             out NuGetUpgrader nuGetUpgrader,
             out bool isConfigured,
             out string error)
         {
-            NuGetUpgraderConfig upgraderConfig = new NuGetUpgraderConfig(tracer, new LocalGVFSConfig());
+            NuGetUpgraderConfig upgraderConfig = new NuGetUpgraderConfig(tracer, gvfsConfig);
             nuGetUpgrader = null;
             isConfigured = false;
 
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -591,11 +591,9 @@ namespace GVFS.Common.NuGetUpgrade
         {
             if (!this.isNuGetFeedInitialized)
             {
-                Debug.Assert(this.credentialStore != null, "Attempting to access nuget feed without a credential store configured.");
                 if (this.credentialStore == null)
                 {
-                    error = "Program error - no Credential Store configured";
-                    return false;
+                    throw new InvalidOperationException("Attempted to call method that requires authentication but no CredentialStore is configured.");
                 }
 
                 string authUrl;
--- a/GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
@@ -1,0 +1,183 @@
+﻿using GVFS.Common;
+using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.NuGetUpgrade;
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Mock.Common;
+using Moq;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Common
+{
+    public class TryCreateProductUpgradeTests
+    {
+        private static string defaultUpgradeFeedPackageName = "package";
+        private static string defaultUpgradeFeedUrl = "https://pkgs.dev.azure.com/contoso";
+        private static string defaultRing = "slow";
+
+        private MockTracer tracer;
+        private Mock<PhysicalFileSystem> fileSystemMock;
+        private Mock<ICredentialStore> credentialStoreMock;
+
+        [SetUp]
+        public void Setup()
+        {
+            this.tracer = new MockTracer();
+
+            // It is important that creating a new Upgrader does not
+            // require credentials.  We must be able to create an
+            // upgrader to query / check upgrade preconditions without
+            // requiring authorization.  We create these mocks with
+            // strict behavior to validate methods on them are called
+            // unnecessarily.
+            this.credentialStoreMock = new Mock<ICredentialStore>(MockBehavior.Strict);
+            this.fileSystemMock = new Mock<PhysicalFileSystem>(MockBehavior.Strict);
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            this.credentialStoreMock.VerifyAll();
+            this.fileSystemMock.VerifyAll();
+        }
+
+        [TestCase]
+        public void CreatesNuGetUpgraderWhenConfigured()
+        {
+            MockLocalGVFSConfig gvfsConfig = this.ConstructDefaultMockNuGetConfigBuilder()
+                .Build();
+
+            bool success = ProductUpgrader.TryCreateUpgrader(
+                this.tracer,
+                this.fileSystemMock.Object,
+                gvfsConfig,
+                this.credentialStoreMock.Object,
+                false,
+                false,
+                out ProductUpgrader productUpgrader,
+                out string error);
+
+            success.ShouldBeTrue();
+            productUpgrader.ShouldNotBeNull();
+            productUpgrader.ShouldBeOfType<NuGetUpgrader>();
+            error.ShouldBeNull();
+        }
+
+        [TestCase]
+        public void CreatesNuGetUpgraderWhenConfiguredWithNoRing()
+        {
+            MockLocalGVFSConfig gvfsConfig = this.ConstructDefaultMockNuGetConfigBuilder()
+                .WithNoUpgradeRing()
+                .Build();
+
+            bool success = ProductUpgrader.TryCreateUpgrader(
+                this.tracer,
+                this.fileSystemMock.Object,
+                gvfsConfig,
+                this.credentialStoreMock.Object,
+                false,
+                false,
+                out ProductUpgrader productUpgrader,
+                out string error);
+
+            success.ShouldBeTrue();
+            productUpgrader.ShouldNotBeNull();
+            productUpgrader.ShouldBeOfType<NuGetUpgrader>();
+            error.ShouldBeNull();
+        }
+
+        [TestCase]
+        public void CreatesGitHubUpgraderWhenConfigured()
+        {
+            MockLocalGVFSConfig gvfsConfig = this.ConstructDefaultGitHubConfigBuilder()
+                .Build();
+
+            this.fileSystemMock.Setup(m => m.CreateDirectory(It.IsAny<string>()));
+
+            bool success = ProductUpgrader.TryCreateUpgrader(
+                this.tracer,
+                this.fileSystemMock.Object,
+                gvfsConfig,
+                this.credentialStoreMock.Object,
+                false,
+                false,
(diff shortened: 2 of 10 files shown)


GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -597,6 +591,13 @@ private bool EnsureNuGetFeedInitialized(out string error)
         {
             if (!this.isNuGetFeedInitialized)
             {
+                Debug.Assert(this.credentialStore != null, "Attempting to access nuget feed without a credential store configured.");
```
jamill: This would be a programming error to call this method wihtout having ICredentialStore set. It is unfortunate that this check has to be here (and calling code needs to know this implicit contract) - I think this could indicate that methods that require authentication could be split from the methods that do not.... but I think that is a larger change

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using GVFS.Common.NuGetUpgrade;
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Mock.Common;
+using GVFS.UnitTests.Windows.Mock.Upgrader;
+using Moq;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Windows.Windows.Upgrader
+{
+    public class NewProductUpgradeTests
```
jamill: I don't like the name, but there is an existing test class for `ProductUpgraderTests` whose testing style I did not want to mix with these tests, as they have different setups.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        [TestCase]
+        public void CreatesNuGetUpgraderWhenConfigured()
+        {
+            this.gvfsConfig = this.BuildNuGetGvfsConfig();
+            ProductUpgrader.TryCreateUpgrader(this.tracer, this.fileSystemMock.Object, this.gvfsConfig, this.credentialStoreMock.Object, false, false, out ProductUpgrader productUpgrader, out string error);
+
+            productUpgrader.ShouldNotBeNull();
+            productUpgrader.ShouldBeOfType<NuGetUpgrader>();
```
kewillford: Do we need to check the return value or the `error` value?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        private MockLocalGVFSConfig BuildNuGetConfigMissingUpgradeFeed()
+        {
+            MockLocalGVFSConfig gvfsConfig = new MockLocalGVFSConfig();
+            gvfsConfig.TrySetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, "slow", out _);
+            gvfsConfig.TrySetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedPackageName, "package", out _);
+            return gvfsConfig;
+        }
+
+        private MockLocalGVFSConfig BuildNuGetConfigMissingPackageName()
```
kewillford: Do we need to test missing `UpgradeRing`?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
kewillford: I don't think we have been using `Debug.Assert` in the code base up to this point.  What value does this bring?  What will the user see when they hit this situation?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
jamill: Not at this time - Upgrade ring is not used by `NuGetUpgrader`.  In the main workflow, we expect it to be present, but ignored by the `NuGetUpgrader` logic. `NuGetUpgrader` logic should work, even if it is not present.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: > I don't think we have been using Debug.Assert in the code base up to this point. What value does this bring? What will the user see when they hit this situation?

I could throw an exception here instead... Users should never see this, as it indicates a programming error. If they do see it, the will see the error from the line below.

I guess the question is - do we want to throw in the release builds, or throw in just debug builds and have this error message propagated in release builds...

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
kewillford: `TryCreateUpgraderTests`? or will you be adding more tests to this later?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
wilbaker: >I could throw an exception here instead... Users should never see this, as it indicates a programming error. If they do see it, the will see the error from the line below.
>
>I guess the question is - do we want to throw in the release builds, or throw in just debug builds and have this error message propagated in release builds...

Elsewhere in the code that we've added checks like this (i.e. invalid states that indicate programmer error) we've thrown an `InvalidOperationException`.  I think the same would be appropriate here.

wilbaker: @jamill actually, is it possible to create a `NuGetUpgrader` without using one of the methods/constructors that takes a `credentialStore` as a parameter?

If not, could those methods be updated to throw an `ArgumentNullException` if `credentialStore` is null rather than checking if `this.credentialStore` is null here?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            MockLocalGVFSConfig gvfsConfig = new MockLocalGVFSConfig();
+            gvfsConfig.TrySetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, "slow", out _);
+            gvfsConfig.TrySetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, "https://pkgs.dev.azure.com/contoso/", out _);
+            return gvfsConfig;
+        }
+
+        private MockLocalGVFSConfig BuildGitHubtGvfsConfig()
```
wilbaker: Typo: `Hubt`.   The other methods don't include "Gvfs" in the name, so you could consider renaming this method to `BuildGitHubConfig`.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
wilbaker: >`NuGetUpgrader` logic should work, even if it is not present.

Could/should we validate that by having a test with no `UpgradeRing`?

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using GVFS.UnitTests.Windows.Mock.Upgrader;
+using Moq;
+using NUnit.Framework;
+
+namespace GVFS.UnitTests.Windows.Windows.Upgrader
+{
+    public class NewProductUpgradeTests
+    {
+        private MockTracer tracer;
+        private MockLocalGVFSConfig gvfsConfig;
```
wilbaker: Minor: Since ever test ends up creating its own `gvfsConfig` I'm not sure there's any value in having `gvfsConfig` as a member variable.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -182,8 +182,17 @@ private bool TryInitialize(out string errorMessage)
         {
             if (this.upgrader == null)
             {
+                string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
+                if (string.IsNullOrEmpty(gitBinPath))
+                {
+                    errorMessage = $"NuGetUpgrader: Unable to locate git installation. Ensure git is installed and try again.";
```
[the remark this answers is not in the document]
jamill: yep - you are right. left over copy / paste issue from original code.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        private Mock<PhysicalFileSystem> fileSystemMock;
+        private Mock<ICredentialStore> credentialStoreMock;
+
+        [SetUp]
+        public void Setup()
+        {
+            this.tracer = new MockTracer();
+            this.fileSystemMock = new Mock<PhysicalFileSystem>();
```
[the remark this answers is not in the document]
jamill: We do end up making a call to file system:

`Moq.MockException : PhysicalFileSystem.CreateDirectory("/usr/share/GVFS/GVFS.Upgrade") invocation failed with mock behavior Strict.`

I can look into this more and see if this is expected...

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: > is it possible to create a NuGetUpgrader without using one of the methods/constructors that takes a credentialStore as a parameter?

With the current design, this is possible (unfortunately). Instead of refactoring to remove this behavior at this time, I was going to keep this bit of awkwardness and will throw an exception if we call into this incorrectly...

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -182,8 +182,17 @@ private bool TryInitialize(out string errorMessage)
...
+                {
+                    errorMessage = $"nameof(this.TryInitialize): Unable to locate git installation. Ensure git is installed and try again.";
+                    return false;
+                }
+
+                ICredentialStore credentialStore = new GitProcess(gitBinPath, workingDirectoryRoot: null, gvfsHooksRoot: null);
+
                 ProductUpgrader upgrader;
-                if (!ProductUpgrader.TryCreateUpgrader(this.tracer, this.fileSystem, this.DryRun, this.NoVerify, out upgrader, out errorMessage))
+                if (!ProductUpgrader.TryCreateUpgrader(this.tracer, this.fileSystem, new LocalGVFSConfig(), credentialStore, this.DryRun, this.NoVerify, out upgrader, out errorMessage))
```
kewillford: Is there a way we can share this code and use it here and in the UpgradeVerb?

GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            // It is important that creating a new Upgrader does not
+            // require credentials.  We must be able to create an
+            // upgrader to query / check upgrade preconditions without
+            // requiring authorization.  We create these mocks with
+            // strict behavior to validate methods on them are called
+            // unnecessarily.
+            this.credentialStoreMock = new Mock<ICredentialStore>(MockBehavior.Strict);
+            this.fileSystemMock = new Mock<PhysicalFileSystem>();
+        }
```
kewillford: Do we need a `TearDown` that will call `VerifyAll()` for these mocks?

GVFS/GVFS.UnitTests/Mock/Common/MockLocalGVFSConfigBuilder.cs
```diff
@@ -0,0 +1,57 @@
+using GVFS.Common;
+using System.Collections.Generic;
+
+namespace GVFS.UnitTests.Mock.Common
+{
+    public class MockLocalGVFSConfigBuilder
```
kewillford: I like this pattern, there are a couple of things that I would do differently.
1. set the default values from the test here in the constructor and put them in the `entries` since that is the default case happy path good config.
2. Do not use default values in your `With...` methods because I don't know from looking at the test what the value of it will be.
3. You can use `WithNo...` to remove the default value from the entries.

I think this will make it more explicit what you are testing so the name of the test matches what your builder is calling. 

Like this
```
        public void NoUpgraderWhenNuGetFeedMissing()
        {
            MockLocalGVFSConfig gvfsConfig = this.ConstructMockLocalGVFSConfigBuilder()
                .WithNoUpgradeFeedUrl()
                .Build();
```
instead of 
```
        public void NoUpgraderWhenNuGetFeedMissing()
        {
            MockLocalGVFSConfig gvfsConfig = this.ConstructMockLocalGVFSConfigBuilder()
                .WithUpgradeRing()
                .WithUpgradeFeedPackageName()
                .Build();
```

GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
jamill: With the `Behavior.Strict`, it will throw an error if there is a call to a method that is not expected (I don't need `VerifyAll` for this). For the `fileSystemMock`, I am using it more to support the test and didn't want to add any verification on it at this time. From another comment, there is an existing call in one of the GitHubUpgrader code paths that might not be necessary, but we still have to account for it now....

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
jamill: Probably :) I hadn't looked at it too closely, but I suspect the answer is yes.... but outside the scope of this PR.

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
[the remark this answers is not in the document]
wilbaker: > I can look into this more and see if this is expected...

Assuming this is expected, could we add a Mock for `CreateDirectory`?

(I'm also not sure if we should test is succeeding/failing, of if that would fall under a different set of unit tests)

GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
wilbaker: > For the fileSystemMock, I am using it more to support the test and didn't want to add any verification on it at this time.

As discussed offline for a sanity check you could consider mocking `CreateDirectory` and setting `Behavior.Strict` so that we can catch if new dependencies are introduced.

GVFS/GVFS.UnitTests/Mock/Common/MockLocalGVFSConfigBuilder.cs
jamill: Thanks!

I incorporated some of these ideas. In the test setup, I create the "default" / "baseline" config for NuGet or GitHub, and then in the specific test cases I remove entries as appropriate with the `WithNo...` pattern.

GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
jamill: Done!

GVFS/GVFS.UnitTests.Windows/Windows/Upgrader/NewProductUpgradeTests.cs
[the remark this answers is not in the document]
jamill: Done!

GVFS/GVFS.UnitTests/Mock/Common/MockLocalGVFSConfigBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+            return this.WithNo(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl);
+        }
+
+        public MockLocalGVFSConfigBuilder With(string key, string value)
+        {
+            this.entries.Add(key, value);
+            return this;
+        }
+
+        public MockLocalGVFSConfigBuilder WithNo(string key)
```
kewillford: `With` and `WithNo` can be `private` since they are never used outside this class.

PR #1026, opened 2019-04-11
Track tombstones so that we know when DeleteFile needs to be called

kewillford:
Currently we are calling `Directory.Exists` when checking if we need to call `DeleteFile` on a folder placeholder.  We do this because we need to delete tombstones when a git command is running and the projection is updated but this is very expensive when there are 100K+ folder placeholders and no tombstones the need to be deleted.  Every folder is paying the price of checking for existence.

We can do better by tracking the tombstones that are created when a folder is deleted during a git command and save that information in the placeholder database to use when the projection is updated.

FYI on performance:

checkout w/ ~100K folder placeholders|time
--- | ---: |
before #1014|30s|
after #1014|22s|
with this PR|11s|

Todo:
- [x] Add tests to the PlaceholderDatabaseTests
- [x] See if there are other tests we can add to test a git command being canceled or crashing and making sure the tombstones still get deleted the next projection change.
- [x] Update the major disk layout version


GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -202,6 +202,12 @@ protected void OnWorkingDirectoryFileOrFolderDeleteNotification(string relativeP
                         this.FileSystemCallbacks.OnFolderDeleted(relativePath);
                     }
                 }
+                else if (!isPreDelete)
```
wilbaker: A few suggestions/considerations, let me know what you think:


1. It could be a bit misleading/confusing to call a method named `OnTombstoneFolderCreated` here.  When `OnWorkingDirectoryFileOrFolderDeleteNotification` is called on Windows there's no guarantee that a tombstone has been created, it might have simply been a full folder that was deleted.

    One way to address this would be to add another parameter (something like `triggeredByGit`) to `OnFolderDeleted`, call `OnFolderDeleted` instead, and then push the decision of how to handle deletes by git into `OnFolderDeleted`.

2. Whether it be here or in `FileSystemCallbacks.OnFolderDeleted`, should we only record the folder delete when the platform supports tombstones?  If Linux ends up supporting post-delete notifications and doesn't use tombstones there would be no need to record the delete in the placeholder database.  Perhaps rather than driving this decision based on `isPreDelete` we should check `GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories` (or maybe add a new `GVFSPlatform.Instance.KernelDriver.SupportsTombstones`).

3. Related to 1, because there's no guarantee that this path is a tombstone, should we record something other than `TOMBSTONE FOLDER` in the placeholder database (e.g. `PSBL_TOMBSTONE FOLDER`)?  An alternative would be to have a separate database of deleted folders that we check against when updating folder placeholders, but that seemed potentially more complicated (and slower).

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1172,7 +1177,7 @@ private void UpdatePlaceholders()
                             if (!isProjected ||
                                 !isFolder ||
                                 (!GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories &&
-                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path))))
+                                folderPlaceholder.IsTombstoneFolder))
```
wilbaker: If we decide to only record possible tombstones when the platform supports it, can/should we remove the `EnumerationExpandsDirectories` check?

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -13,6 +13,7 @@ public class PlaceholderListDatabase : FileBasedCollection
         // - Not be a valid SHA-1 value (to avoid collisions with files)
         public const string PartialFolderValue = "                          PARTIAL FOLDER";
         public const string ExpandedFolderValue = "                         EXPANDED FOLDER";
+        public const string TombstoneFolderValue = "                        TOMBSTONE FOLDER";
```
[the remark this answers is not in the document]
kewillford: Yeah I was also wondering if this needed a version bump and an upgrader to get the all the tombstones from disk and add them to the placeholder database but since this should only matter during a git command I didn't.  What will bumping the minor version do as opposed to the major version?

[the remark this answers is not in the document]
wilbaker: Oh I am glad you asked, I was wrong and it looks like this requires a major version bump.

```
        // The major version should be bumped whenever there is an on-disk format change that requires a one-way upgrade.
        // Increasing this version will make older versions of GVFS unable to mount a repo that has been mounted by a newer
        // version of GVFS.
        public int CurrentMajorVersion { get; }

        // The minor version should be bumped whenever there is an upgrade that can be safely ignored by older versions of GVFS.
        // For example, this allows an upgrade step that sets a default value for some new config setting.
        public int CurrentMinorVersion { get; }
```

If we wanted to avoid the major version bump then I think we'd have to go with the approach of a new database (that old VFS4G versions don't care about).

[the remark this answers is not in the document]
wilbaker: >and an upgrader to get the all the tombstones from disk and add them to the placeholder database but since this should only matter during a git command I didn't.

I think we would only need this if we decided to handle the scenario of an interrupted git command followed by immediate upgrade of VFS4G.

That seems a bit obscure, but we should get some other people to weigh in as well.

/cc @jrbriggs

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
kewillford: >push the decision of how to handle deletes by git into OnFolderDeleted

I don't think we would have enough information there either unless we try to read the directory attributes to find the tombstone attribute.  So I don't see a difference in where the decision is made.  
>only record the folder delete when the platform supports tombstones

What do you think about having the `FileSystemVirtualizer` have a method that is virtual and doesn't do anything something like `OnPossibleTombstoneFolder` and override that in the `WindowsFileSystemVirtualizer` then we wouldn't need anything extra in the Platform.

I don't really want another database to keep this data since it is about a placeholder and the placeholder database is where we keep that information.  Adding more information into the `SHA` field is fine.  We have 40 characters. :)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: That should be fine since there shouldn't ever be any folder placeholder marked as possible tombstones.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
wilbaker: > What do you think about having the FileSystemVirtualizer have a method that is virtual and doesn't do anything something like OnPossibleTombstoneFolder and override that in the WindowsFileSystemVirtualizer then we wouldn't need anything extra in the Platform.

I can get on board with that if the MacFileSystemVirtualizer has a no-op implementation with a comment explaining why it's not doing anything.  However, it still feels a bit wrong to me to call `OnPossibleTombstoneFolder` on a platform that doesn't support it, and so let me throw another alternative out there and let me know what you think.

We could remove the shared `OnWorkingDirectoryFileOrFolderDeleteNotification` method and move the "isPreDelete" code into `MacFileSystemVirtualizer` and the "post delete + tombstone" code to `WindowsFileSystemVirtualizer`. 

`OnWorkingDirectoryFileOrFolderDeleteNotification` was first added because the delete code on the two platforms was so similar, but perhaps it's diverging enough now it shouldn't be shared anymore?

kewillford: I will give that a try and see how it looks.  They are different in that they each will call different `FileSystemCallbacks` but they share the logic for when to call them but this change is kind of breaking that.

kewillford: @wilbaker I reworked this to try and keep the logic and have the platform specific code handle the difference in what `FileSystemCallbacks` was called.  Let me know what you think and if there are any tweaks to the method names I should make.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1248,7 +1268,7 @@ private HResult NotifyPreDeleteHandler(string virtualPath, bool isDirectory)
                     }
                     else
                     {
-                        this.OnWorkingDirectoryFileOrFolderDeleteNotification(virtualPath, isDirectory, isPreDelete: false);
+                        this.OnWorkingDirectoryFileOrFolderDeleteNotification(virtualPath, isDirectory);
```
wilbaker: I like that this change keeps the call to `OnPossibleTombstoneFolderCreated` in `WindowsFileSystemVirtualizer`, but I'm not sure about all of the back and forth that we know have between FileSystemCallbacks and the Mac/WindowsFileSystemVirtualizer.cs:

- `ProjFS` -> `WindowsFileSystemVirtualizer` -> `FileSystemCallbacks` -> `WindowsFileSystemVirtualizer` -> `FileSystemCallbacks`

What if this call to `OnWorkingDirectoryFileOrFolderDeleteNotification` was replaced with something like the following (which could always go in a new `WindowsFileSystemVirtualizer` if that would be cleaner):

```
            if (isDirectory)
            {
                GitCommandLineParser gitCommand = new GitCommandLineParser(this.Context.Repository.GVFSLock.GetLockedGitCommand());
                if (!gitCommand.IsValidGitCommand)
                {
                    this.FileSystemCallbacks.OnFolderDeleted(relativePath);
                }
                else
                {
                    // During a git command if it deletes a folder we need to track that as a tombstone
                    // So that we can delete them if the projection changes
                    this.FileSystemCallbacks.OnPossibleTombstoneFolderCreated(relativePath);
                }
            }
            else
            {
                this.FileSystemCallbacks.OnFileDeleted(relativePath);
            }

            this.FileSystemCallbacks.InvalidateGitStatusCache();
```

And we removed `OnWorkingDirectoryFileOrFolderDeleteNotification` and the new `OnFileDeleteNotification` and `OnFolderDeleteNotification` methods?

Similarly in `MacFileSystemVirtualizer.cs` the call to `OnWorkingDirectoryFileOrFolderDeleteNotification` could be replaced with:

```
            if (isDirectory)
            {
                GitCommandLineParser gitCommand = new GitCommandLineParser(this.Context.Repository.GVFSLock.GetLockedGitCommand());
                if (!gitCommand.IsValidGitCommand)
                {
                    this.FileSystemCallbacks.OnFolderPreDelete(relativePath);
                }
            }
            else
            {
                this.FileSystemCallbacks.OnFilePreDelete(relativePath);
            }

            this.FileSystemCallbacks.InvalidateGitStatusCache();
```

If we wanted to share a bit of code we could always add a common helper method like this:

```
bool IsGitCommandRunning()
{
    GitCommandLineParser gitCommand = 
        new GitCommandLineParser(this.Context.Repository.GVFSLock.GetLockedGitCommand());
    return gitCommand.IsValidGitCommand;
}
```

This does duplicate the decision of whether OnFolderDelete should be called (based on if the git command is running), but it avoids having to add the new `OnFileDeleteNotification` and `OnFolderDeleteNotification` methods.

I think given the choice between the current changes and the other alternative you suggested (no-op `OnPossibleTombstoneFolderCreated` in Mac) my preference is the latter, as I think it might be tougher to understand why Mac and Windows call different methods in their `OnFileDeleteNotification` and `OnFolderDeleteNotification` overrides.

kewillford: I originally did refactor it like you are suggesting but the issue I saw was that the were both making the same decisions and it was only the last call to `FileSystemCallbacks` that was platform specific.  So if we needed to add something like the invalidate of the git status cache we would need to remember to add it in both places which I didn't like because sooner or later when you have to remember, you will forget until there is an issue with a user.  

I will go with the no-op then because it really is code that I think should be shared if at all possible which we can to avoid bugs because we forgot to change something in multiple places.

wilbaker: > I will go with the no-op then because it really is code that I think should be shared if at all possible which we can to avoid bugs because we forgot to change something in multiple places.

Ok thanks that sounds good.

wilbaker:
@kewillford the latest changes look good to me.  I will hold off on approving until the TODO list is complete.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -371,6 +377,19 @@ public bool IsExpandedFolder
                     return this.Sha == ExpandedFolderValue;
                 }
             }
+
+            public bool IsPossibleTombstoneFolder
+            {
+                get
+                {
+                    return this.Sha == PossibleTombstoneFolderValue;
```
jamill: While not something this project seems to always follow, I like when we explicitly define which string comparison strategy to use instead of the `==` method.

What string comparison strategy is most appropriate here? (Ordinal Case Sensitive)? Same for `IsShaAFolder`

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -202,6 +202,13 @@ protected void OnWorkingDirectoryFileOrFolderDeleteNotification(string relativeP
                         this.FileSystemCallbacks.OnFolderDeleted(relativePath);
                     }
                 }
+                else
+                {
+                    // During a git command if it deletes a folder we need to track that as a tombstone
```
jamill: nit: I might consider re-organizing these blocks so the positive case comes 1st... but this might just fit better with how I read code - as it, it works fines.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -202,6 +202,13 @@ protected void OnWorkingDirectoryFileOrFolderDeleteNotification(string relativeP
                         this.FileSystemCallbacks.OnFolderDeleted(relativePath);
                     }
                 }
+                else
+                {
+                    // During a git command if it deletes a folder we need to track that as a tombstone
+                    // So that we can delete them if the projection changes. This will be a no-op on platforms
+                    // that don't override OnPossibleTombstoneFolderCreated
+                    this.OnPossibleTombstoneFolderCreated(relativePath);
```
jamill: One question I had - why is this a `possibleTombstoneFolder`? Here we have a notification that the directory was deleted, so we should "know" that the folder was deleted, and is a tombstone? When would we delete the directory here, but the directory would not be a tombstone?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1168,11 +1173,10 @@ private void UpdatePlaceholders()
                             // The delete will be attempted if one of the following is true
                             // 1. not in the projection anymore
                             // 2. in the projection but is not a folder in the projection
-                            // 3. Folder no longer exist on disk - this is to remove the tombstone
+                            // 3. Folder placeholder is a possible tombstone
                             if (!isProjected ||
                                 !isFolder ||
-                                (!GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories &&
-                                !this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, folderPlaceholder.Path))))
+                                folderPlaceholder.IsPossibleTombstoneFolder)
```
jamill: Before we checked if the file exists before attempting to delete it. If this is a `PossibleTombstone`, do we always delete the folder? Do we need to check if this "possible" tombstone is actually a tombstone?

GVFS/GVFS.Common/PlaceholderListDatabase.cs
kewillford: Yeah I was following what the code in this file was already doing so I will leave this for another PR.  Do you know what the performance difference in using `==` vs using a comparison strategy would be?

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
kewillford: Yeah sometimes that causes the diff to be uglier so I will leave as is for now.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
kewillford: The folder could have been a full folder and we would still get a notification for the delete but there would not be a tombstone for the folder.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: We just attempt the delete in `RemoveFolderPlaceholderIfEmpty` and it will get a `FSResult.FileOrPathNotFound` if it isn't a tombstone which saves a check by calling the directory exists or trying to read the file attributes and check for tombstones.

GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/WindowsTombstoneTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            placeholderListContent = string.Join(Delimiter, placeholderListContent.Split(new[] { Delimiter }, StringSplitOptions.RemoveEmptyEntries).Where(x => !x.StartsWith($"A {folderToDelete}")));
+            this.fileSystem.WriteAllText(placeholderListFile, placeholderListContent + Delimiter);
+            this.fileSystem.AppendAllText(placeholderListFile, $"A {folderToDelete}\0               POSSIBLE TOMBSTONE FOLDER{Delimiter}");
+
+            this.Enlistment.MountGVFS();
+            directoryToDelete.ShouldNotExistOnDisk(this.fileSystem);
+
+            // checkout branch to remove tombstones and project the folder again
+            GitHelpers.InvokeGitAgainstGVFSRepo(this.Enlistment.RepoRoot, "checkout -f HEAD");
+            directoryToDelete.ShouldBeADirectory(this.fileSystem);
```
wilbaker: Can we also validate that `A {folderToDelete}\0               POSSIBLE TOMBSTONE FOLDER` has been removed from `placeholderListFile` after the checkout completes?

GVFS/GVFS.Common/PlaceholderListDatabase.cs
jamill: > Do you know what the performance difference in using == vs using a comparison strategy would be?

I think the `==` operator implicity [calls the static Equals(String, String) method](https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netframework-4.7.2#System_String_Equals_System_String_System_String_),  which [performs an ordinal (case-sensitive and culture-insensitive) comparison.](https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netframework-4.7.2#System_String_Equals_System_String_System_String_)

TLDR: I think the equality operator is using the string comparison method we want, so there shouldn't be any perf difference. The only difference would be that it would be explicitly obvious in code and force developers to make an explicit choice when writing code.

PR #1027, opened 2019-04-12
[POSIX] rename common Mac hook code to share with Linux

chrisd8088:
Rename the `common.mac.cpp` source file to `common.posix.cpp` so it can be used as a basis for Linux hook programs as well.

Also replace the BSD-specific `strlcat()` and `strlcpy()` with generic C string functions from the `<cstring>` standard library header.

### DIFF
--- a/GVFS/GVFS.NativeHooks.Common/common.h
+++ b/GVFS/GVFS.NativeHooks.Common/common.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cstring>
+
 #ifdef __APPLE__
 typedef std::string PATH_STRING;
 typedef int PIPE_HANDLE;
--- a/GVFS/GVFS.NativeHooks.Common/common.posix.cpp
+++ b/GVFS/GVFS.NativeHooks.Common/common.posix.cpp
@@ -31,22 +31,22 @@ PATH_STRING GetGVFSPipeName(const char *appName)
     }
     
     PATH_STRING finalRootPath(GetFinalPathName(enlistmentRoot));
-    size_t resultLength = strlcpy(enlistmentRoot, finalRootPath.c_str(), sizeof(enlistmentRoot));
-    if (resultLength >= sizeof(enlistmentRoot))
+    size_t enlistmentRootLength = strlen(finalRootPath.c_str());
+    if (enlistmentRootLength + 2 > sizeof(enlistmentRoot))
     {
         die(ReturnCode::PipeConnectError,
-            "Could not copy finalRootPath: %ls, insufficient buffer. resultLength: %d, sizeof(enlistmentRoot): %d\n",
+            "Could not copy finalRootPath: %ls, insufficient buffer. enlistmentRootLength: %d, sizeof(enlistmentRoot): %d\n",
             finalRootPath.c_str(),
-            resultLength,
+            enlistmentRootLength,
             sizeof(enlistmentRoot));
     }
     
-    size_t enlistmentRootLength = strlen(enlistmentRoot);
+    memcpy(enlistmentRoot, finalRootPath.c_str(), enlistmentRootLength);
     if ('/' != enlistmentRoot[enlistmentRootLength - 1])
     {
-        strlcat(enlistmentRoot, "/", sizeof(enlistmentRoot));
-        enlistmentRootLength++;
+        enlistmentRoot[enlistmentRootLength++] = '/';
     }
+    enlistmentRoot[enlistmentRootLength] = '\0';
     
     // Walk up enlistmentRoot looking for a folder named .gvfs
     char* lastslash = enlistmentRoot + enlistmentRootLength - 1;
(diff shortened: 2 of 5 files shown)


GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -31,22 +31,22 @@ PATH_STRING GetGVFSPipeName(const char *appName)
     }
     
     PATH_STRING finalRootPath(GetFinalPathName(enlistmentRoot));
-    size_t resultLength = strlcpy(enlistmentRoot, finalRootPath.c_str(), sizeof(enlistmentRoot));
-    if (resultLength >= sizeof(enlistmentRoot))
+    size_t enlistmentRootLength = strlen(finalRootPath.c_str());
```
jrbriggs: I know you're not the one to introduce this, but can you please use `strnlen` or similar, as we know (at least right now) this cannot be larger than `MAX_PATH`, and we need to move away from `strlen`

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -116,17 +116,19 @@ PIPE_HANDLE CreatePipeToGVFS(const PATH_STRING& pipeName)
     memset(&socket_address, 0, sizeof(struct sockaddr_un));
     
     socket_address.sun_family = AF_UNIX;
-    size_t resultLength = strlcpy(socket_address.sun_path, pipeName.c_str(), sizeof(socket_address.sun_path));
-    
-    if (resultLength >= sizeof(socket_address.sun_path))
+    size_t pathLength = strlen(pipeName.c_str());
```
jrbriggs: Same comment on `strnlen`, since it seems to be bounded by `sizeof(socket_address.sun_path)`

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -31,22 +31,22 @@ PATH_STRING GetGVFSPipeName(const char *appName)
     }
     
     PATH_STRING finalRootPath(GetFinalPathName(enlistmentRoot));
-    size_t resultLength = strlcpy(enlistmentRoot, finalRootPath.c_str(), sizeof(enlistmentRoot));
-    if (resultLength >= sizeof(enlistmentRoot))
+    size_t enlistmentRootLength = strlen(finalRootPath.c_str());
+    if (enlistmentRootLength + 2 > sizeof(enlistmentRoot))
```
wilbaker: Could you add a comment explaining why we're adding 2?

I'm assuming it's because we might have to add a `/` below.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
wilbaker: `finalRootPath` is actually an `std::string` and so we can simply call `finalRootPath.length()` 

We could switch from `PATH_STRING` to `std::string` above on line 33, but I'm not sure for readability which is better.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
wilbaker: We can do the same here, `std::string` already knows its length.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
chrisd8088: Yes, exactly.

GVFS/GVFS.NativeHooks.Common/common.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <cstring>
+
 #ifdef __APPLE__
```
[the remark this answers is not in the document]
chrisd8088: Right; the `#elif` clause for Linux is in another commit in the [`linux-gvfs-build`](https://github.com/Microsoft/VFSForGit/pull/986/commits/de6b32a0270d433b20142d0607aa4a7bb516aebf) branch.

This change to add the `#include <cstring>` here is because there are calls to `strcmp()` and friends in the various hook programs' [`main.cpp`](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.PostIndexChangedHook/main.cpp#L18) files, and at least with my GCC compiler these fail unless `<cstring>` is included; it has nothing to do with the various string calls in `common.posix.cpp` elsewhere in this PR.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
chrisd8088: And I notice on a second look that I wasn't leaving enough space for the trailing null, either.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
chrisd8088: Done!

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -31,22 +31,23 @@ PATH_STRING GetGVFSPipeName(const char *appName)
...
             finalRootPath.c_str(),
-            resultLength,
+            enlistmentRootLength,
             sizeof(enlistmentRoot));
     }
     
-    size_t enlistmentRootLength = strlen(enlistmentRoot);
-    if ('/' != enlistmentRoot[enlistmentRootLength - 1])
+    memcpy(enlistmentRoot, finalRootPath.c_str(), enlistmentRootLength);
+    if (enlistmentRootLength == 0 || enlistmentRoot[enlistmentRootLength - 1] != '/')
```
chrisd8088: On self-review, I think the [existing code](https://github.com/Microsoft/VFSForGit/blob/master/GVFS/GVFS.NativeHooks.Common/common.mac.cpp#L45) and my first try at this both could fail if `enlistmentRootLength` is zero as they test `enlistmentRoot[enlistmentRootLength - 1]`.  It's possible the path can't be empty, but I figure it doesn't hurt to be careful here.

wilbaker:
@chrisd8088 FYI there was break to the Extra functional tests merged into master.  

Until #1049 is completed you will probably see "PR - Windows - Extra" fail in `GVFS.FunctionalTests.Tests.EnlistmentPerFixture.UpgradeReminderTests.UpgradeTimerScheduledOnServiceStart`

chrisd8088:
OK, rebased on master to get the latest CI all working ... thanks for the tip!

@jrbriggs, does this look OK to you now?

chrisd8088:
Thanks very much @wilbaker, @jrbriggs -- please merge at your pleasure!

chrisd8088:
Hi @wilbaker or @jrbriggs ... would you be able to merge this into master, if there are no further changes required?

chrisd8088:
@jrbriggs, @wilbaker -- I updated this against the current master, although there were no conflicts.  I'm unable to merge to master myself, however.

PR #1029, opened 2019-04-12
exclude .a files from code coverage

jeschu1:
We want to see coverage failures on a per function basis, not a full file (.cpp/.hpp./.a).  

The .a (like a .cpp/.hpp) gives a code coverage summary for a set of functions.  

The reason why this caused a failure is that one of the functions in the .a summary report is on the exclusion list.  

However it is a mystery why the .a file is only now being generated.  Perhaps something changed in the build causing this to happen, or the Xcode verison on the machine is slightly different.

### DIFF
--- a/ProjFS.Mac/Scripts/Build.sh
+++ b/ProjFS.Mac/Scripts/Build.sh
@@ -83,6 +83,7 @@ while read line; do
 	 [[ $line != *"RWLock_DropExclusiveToShared"* ]] && 
 	 [[ $line != *".xctest"* ]] && 
 	 [[ $line != *".cpp"* ]] && 
+	 [[ $line != *".a"* ]] && 
 	 [[ $line != *".hpp"* ]]; then
        echo "Error: not at 100% Code Coverage $line"
        exit 1


jamill:
Thanks for taking a look on this so quickly!

> The reason why this caused a failure is that one of the functions in the .a summary report is on the exclusion list.

How does this work? There is an exclusion list that indicates what files need to be checked?

How likely is it that other changes (outside of our control) will cause this test to start failing? Were you planning on investigating what changed here?

jeschu1:
> How does this work? There is an exclusion list that indicates what files need to be checked?

Yes, thats correct.  We have a list of text exclusions.  .cpp/.hpp were already excluded

> How likely is it that other changes (outside of our control) will cause this test to start failing? Are we going to see what changed here to cause this failure?

I've asked @nickgra if he can get us in touch with a code coverage expert from office.
If there are changes, we should see them fail on the PR not the master branch.  I'm wondering if we have different versions of XCode installed on the machines that allowed this to happen.  If so, we may want to stabilize on one version of XCode.

PR #1030, opened 2019-04-12
Add more time for disposal

jeschu1:
I saw a failure in a unit test, and the new logging helped me diagnose the issue :-)
- The failure was "Did not dispose object in time"

Unfortunately after requeue I no longer see that build, so I can't provide a link.  I'll investigate that in a separate thread.

The current tests waits 100ms for dispose to be called.
This ups the time to 500ms.
If you look at the test class, there is a background timer every 1ms that attempts to dispose.  

My vote is to keep the test on with the 500ms.  If it doesn't dispose in that amount of time, i'd be more wary there is an issue.

### DIFF
--- a/GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
@@ -147,7 +147,7 @@ namespace GVFS.UnitTests.Common
             result.ShouldEqual(true, string.Format("Unexcepted function result from first call to TryInvoke: {0} ReturnCode: {1}", result, tryInvokeResult1));
             this.numConstructors.ShouldEqual(1);
 
-            this.DisposalTriggers.TryTake(out object _, (int)this.disposalPeriod.TotalMilliseconds * 100).ShouldBeTrue("Did not dispose object in time");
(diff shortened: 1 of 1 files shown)


jeschu1:
> If this is testing that `LibGit2RepoInvoker` can correctly dispose when there are no active callers, we could consider adding a mechanism that allows the cleanup outside of a timer driven event. This would remove the race condition in the unit test and ensure the test runs fast (I think 500 ms would be one of our longest running unit tests)...

Can you elaborate on how the test would look for that?  I think the test also tests the timer driven disposal mechanism.  So we don't necessarily want to to remove that.

derrickstolee:
>  (I think 500 ms would be one of our longest running unit tests)...

This should only take 500 ms if it is going to fail. Since we are pulling from a `BlockingCollection`, the test will continue as soon as the disposal is registered. If this takes too long, then it is due to something with the `Timer` that triggers the disposal.

If we are running a lot of unit tests in parallel, could that cause the timer thread to be delayed by as much as 500 ms?

jeschu1:
> If we are running a lot of unit tests in parallel, could that cause the timer thread to be delayed by as much as 500 ms?
This test failing isn't common with the 100ms we have in place.  If a machine is overloaded maybe we could hit that 500ms mark, but that is a signal something is wrong.

Another great idea @Jameson had was to split the test into 2 tests.  1 for the timer and 1 for the dispose.  If we still see failures (and deem an overloaded machine is the issue) that's a good route to consider.

PR #1034, opened 2019-04-15
Refactor NuGetUpgrader to better align with latest design

jamill:
Early draft of NuGetUpgrader refactor to better align with the latest design.

## Changes:
Defines the following interfaces:

- `IQueryGVFSVersion`
- 'IDownloadGVFSVersion`

which can be implemented by classes that talk to a NuGet feed or an org info server. The `NuGetUpgrader` now takes these as input, and a single class now supports a scenario where the upgrader should query an org info server for the version to download, and download from a NuGet feed, or should only work against a NuGet feed.


GVFS/GVFS.Common/VersionResponse.cs
```diff
@@ -0,0 +1,15 @@
+using Newtonsoft.Json;
+
+namespace GVFS.Common
+{
+    public class VersionResponse
+    {
+        public string Version { get; set; }
```
mjcheetham: Does this not need to have `[JsonProperty("version")]` for good measure?

GVFS/GVFS.Common/NuGetUpgrade/ModernNugetUpgrader.cs
```diff
@@ -0,0 +1,204 @@
...
+            {
+                return false;
+            }
+
+            if (!upgraderConfig.IsReady(out error))
+            {
+                return false;
+            }
+
+            if (!TryParseOrgFromNugetFeedUrl(upgraderConfig.FeedUrl, out string orgName))
```
mjcheetham: Why is this still needed? I thought the config would now contain the org URL?

GVFS/GVFS.Common/OrgInfoApiClient.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Class that handles communication with a server that contains version information.
+    /// </summary>
+    public class OrgInfoApiClient
```
mjcheetham: What is the rationale for having `OrgInfoApiClient` be a different component from `QueryGVFSVersionFromOrgInfo` given that both only do one thing.. query an endpoint for a version.

Can't `OrgInfoApiClient` be the only component and have an `IOrgInfoApiClient` interface, removing the `QueryGVFSVersionFromOrgInfo` and `IQueryGVFSVersionFromOrgInfo` types?

jamill: This is something I had been thinking about...

The reason why they are separate classes is that `OrgInfoApiClient` is more flexible, and you can use it to query the API (e.g. multiple queries for different query params) with the same instance.

`IQueryGVFSVersionFromOrgInfo` is a wrapper / adapter that is more specific to how the Upgrader wants to call it, and is a bit more restricted... The constructed instance can only query 1 set of parameters.

GVFS/GVFS.Common/NuGetUpgrade/ModernNugetUpgrader.cs
jamill: The config contains the URL to the endpoint to query for available VFSForGit versions, but it does not (currently) know which organization it is querying for. This is a parameter to the query... it doesn't have to come from the NuGetFeed URL, but that is where it currently lives...

I can add more documentation on how the pieces fit tother here.

GVFS/GVFS.Common/QueryGVFSVersionFromOrgInfo.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Net.Http;
+
+namespace GVFS.Common
+{
+    public class QueryGVFSVersionFromOrgInfo : IQueryGVFSVersion
```
wilbaker: I think this might be related to @mjcheetham's comment in OrgInfoApiClient.cs.

I am having trouble understanding why we have both `QueryGVFSVersionFromOrgInfo` and `OrgInfoApiClient`.  As best I can tell all this class is doing is passing through a call to `QueryNewestVersion` (saving the caller from having to pass in the values it used when it constructed the `QueryGVFSVersionFromOrgInfo`).

Is it possible to reduce the level of abstractions here?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -81,6 +83,8 @@ public class NuGetUpgrader : ProductUpgrader
 
         public string DownloadedPackagePath { get; private set; }
 
+        public override bool SupportsAnonymousVersionQuery { get => false; }
```
wilbaker: Why do we need to support the `SupportsAnonymousVersionQuery` is `false` scenario?

Could we make it a requirement that if your organization wants to upgrade via NuGet you must support an anonymous version query endpoint?

jamill:
#1039 is based off the core commits in this change - but includes more details / tests around querying the org info server, and does not include `IQueryGVFSVersion`.

PR #1036, opened 2019-04-15
Normalize line endings

jamill:
Normalize line endings to CRLF


jamill:
I was also going to look into if we can use editor.config to help with this, but I wanted to fix these line endings to help with some PRs in flight.

nickgra:
> I was also going to look into if we can use editor.config to help with this, but I wanted to fix these line endings to help with some PRs in flight.

A few months ago, I looked at turning that EditorConfig feature on, and it seemed to work fine with the editors that I use, but I didn't have the time to validate it on all of the editors that I know people on our team use. Even if it doesn't work perfectly with all of the editors, it seems like forward progress to turn it on.

jamill:
> How are we going to avoid this going forward?

Good question - I think editor config would help (for editors that support this feature).

> working on mac will convert them back to LF and we will continue to have a bunch of different line endings.

I think we have set the .gitattributes to remove the text attribute from our files, so hopefully git will not normalize on the macOS platform.

PR #1038, opened 2019-04-16
Close directory fds when walking directory hierarchy in PrjFSLib

### LINKED ISSUE #1037: GVFS.Mount leaks file descriptors in HandleRecursivelyEnumerateDirectoryRequest and RecursivelyMarkAllChildrenAsInRoot
While standing up the Large Repo Perf Tests on macOS, I noticed that my machine was running out of file descriptors after running a test where thousands of placeholders (files and directories) are created and then deleted. GVFS.Mount had a bunch of file descriptors open for folders that were already deleted. 

We have the same bug in two code paths, `HandleRecursivelyEnumerateDirectoryRequest` and `RecursivelyMarkAllChildrenAsInRoot` both walk a directory structure but only call `closedir` on the parent directory.

nickgra:
Fixes #1037 

We have two spots in PrjFSLib where we traverse a directory structure, calling `opendir` on directories encountered, but we never call `closedir` on those directories before moving onto the next directory.

Now both HandleRecursivelyEnumerateDirectoryRequest and RecursivelyMarkAllChildrenAsInRoot properly close directory file descriptors when walking the directory hierarchy.

This was detected while standing up the Large Repo Perf Tests on macOS. I noticed that my machine was running out of file descriptors after running a test where thousands of placeholders (files and directories) are created and then deleted. Using `lsof -a -p <PID>` on the PID of GVFS.Mount, I saw that we had a bunch of file descriptors open for folders that were already deleted.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -881,6 +881,8 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
             
             dirEntry = readdir(directory);
         }
+        
+        closedir(directory);
     }
     
 CleanupAndReturn:
@@ -1346,6 +1348,8 @@ static PrjFS_Result RecursivelyMarkAllChildrenAsInRoot(const char* fullDirectory
             
             dirEntry = readdir(directory);
         }
+        
+        closedir(directory);
     }
     
 CleanupAndReturn:


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -881,6 +881,11 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
             
             dirEntry = readdir(directory);
         }
+        
+        if (directory != nullptr)
```
wilbaker: We already have a `nullptr` check for this on line 867 and so I don't think we need another one here (`CleanupAndReturn` checks for `nullptr` because it's possible we jump there with a null `directory`).

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1346,6 +1351,11 @@ static PrjFS_Result RecursivelyMarkAllChildrenAsInRoot(const char* fullDirectory
...
```
wilbaker: Same comment here.  Since we're already checking for `nullptr` on line 1326 I don't think we need to check again here (let me know if you see a scenario I'm missing that requires this check).

PR #1039, opened 2019-04-16
Org info server

jamill:
Consume an endpoint that provides indicates what version of VFS for Git should be downloaded based on Organization, platform, and ring.

This is a slightly different take than #1034, and focuses on the org info server implementation more than refining the `Upgrader` logic.

If we can require the upgrade logic to always be driven off of the org info server, we could probably collapse the `ModernNuGetUpgrader` logic into `NuGetUpgrader`.

### DIFF
--- a/GVFS/GVFS.Common/AzDevOpsOrgFromNuGetFeed.cs
+++ b/GVFS/GVFS.Common/AzDevOpsOrgFromNuGetFeed.cs
@@ -1,0 +1,51 @@
+using System.Text.RegularExpressions;
+
+namespace GVFS.Common
+{
+    public class AzDevOpsOrgFromNuGetFeed
+    {
+        /// <summary>
+        /// Given a URL for a NuGet feed hosted on Azure DevOps,
+        /// return the organization that hosts the feed.
+        /// </summary>
+        public static bool TryParseOrg(string packageFeedUrl, out string orgName)
+        {
+            // We expect a URL of the form https://pkgs.dev.azure.com/{org}
+            // and want to convert it to a URL of the form https://{org}.visualstudio.com
+            Regex packageUrlRegex = new Regex(
+                @"^https://pkgs.dev.azure.com/(?<org>.+?)/",
+                RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
+            Match urlMatch = packageUrlRegex.Match(packageFeedUrl);
+
+            if (!urlMatch.Success)
+            {
+                orgName = null;
+                return false;
+            }
+
+            orgName = urlMatch.Groups["org"].Value;
+            return true;
+        }
+
+        /// <summary>
+        /// Given a URL for a NuGet feed hosted on Azure DevOps,
+        /// return a URL that Git Credential Manager can use to
+        /// query for a credential that is valid for use with the
+        /// NuGet feed.
+        /// </summary>
+        public static bool TryCreateCredentialQueryUrl(string packageFeedUrl, out string azureDevOpsUrl, out string error)
+        {
+            if (!TryParseOrg(packageFeedUrl, out string org))
+            {
+                azureDevOpsUrl = null;
+                error = $"Input URL {packageFeedUrl} did not match expected format for an Azure DevOps Package Feed URL";
+                return false;
+            }
+
+            azureDevOpsUrl = $"https://{org}.visualstudio.com";
+            error = null;
+
+            return true;
+        }
+    }
+}
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -25,6 +25,7 @@ namespace GVFS.Common
 
         public GVFSPlatformConstants Constants { get; }
         public UnderConstructionFlags UnderConstruction { get; }
+        public abstract string Name { get; }
 
         public static void Register(GVFSPlatform platform)
         {
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -1,3 +1,4 @@
+using GVFS.Common;
 using GVFS.Common.FileSystem;
 using GVFS.Common.Git;
 using GVFS.Common.Tracing;
@@ -8,7 +9,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
-using System.Text.RegularExpressions;
 using System.Threading;
 
 namespace GVFS.Common.NuGetUpgrade
@@ -140,30 +140,6 @@ namespace GVFS.Common.NuGetUpgrade
                 upgraderConfig,
                 ProductUpgraderInfo.GetAssetDownloadsPath(),
                 credentialStore);
-
-            return true;
-        }
-
-        public static bool TryCreateAzDevOrgUrlFromPackageFeedUrl(string packageFeedUrl, out string azureDevOpsUrl, out string error)
-        {
-            // We expect a URL of the form https://pkgs.dev.azure.com/{org}
-            // and want to convert it to a URL of the form https://{org}.visualstudio.com
-            Regex packageUrlRegex = new Regex(
-                @"^https://pkgs.dev.azure.com/(?<org>.+?)/",
-                RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
-            Match urlMatch = packageUrlRegex.Match(packageFeedUrl);
-
-            if (!urlMatch.Success)
-            {
-                azureDevOpsUrl = null;
-                error = $"Input URL {packageFeedUrl} did not match expected format for an Azure DevOps Package Feed URL";
-                return false;
-            }
-
-            string org = urlMatch.Groups["org"].Value;
-
-            azureDevOpsUrl = urlMatch.Result($"https://{org}.visualstudio.com");
-            error = null;
 
             return true;
         }
@@ -600,7 +576,7 @@ namespace GVFS.Common.NuGetUpgrade
             try
             {
                 string authUrl;
-                if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
+                if (!AzDevOpsOrgFromNuGetFeed.TryCreateCredentialQueryUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
                 {
                     return false;
                 }
@@ -631,7 +607,7 @@ namespace GVFS.Common.NuGetUpgrade
                 }
 
                 string authUrl;
-                if (!TryCreateAzDevOrgUrlFromPackageFeedUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
+                if (!AzDevOpsOrgFromNuGetFeed.TryCreateCredentialQueryUrl(this.nuGetUpgraderConfig.FeedUrl, out authUrl, out error))
                 {
                     return false;
                 }
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -81,6 +81,8 @@ namespace GVFS.Common.NuGetUpgrade
 
         public string DownloadedPackagePath { get; private set; }
 
+        public override bool SupportsAnonymousVersionQuery { get => false; }
+
         /// <summary>
         /// Path to unzip the downloaded upgrade package
         /// </summary>
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -1,6 +1,7 @@
 using GVFS.Common.FileSystem;
 using GVFS.Common.Git;
 using GVFS.Common.Tracing;
+using NuGet.Packaging.Core;
 using NuGet.Protocol.Core.Types;
 using System;
 using System.Collections.Generic;
@@ -14,13 +15,14 @@ namespace GVFS.Common.NuGetUpgrade
 {
     public class NuGetUpgrader : ProductUpgrader
     {
+        protected readonly NuGetUpgraderConfig nuGetUpgraderConfig;
+        protected Version highestVersionAvailable;
+
         private const string ContentDirectoryName = "content";
         private const string InstallManifestFileName = "install-manifest.json";
         private const string ExtractedInstallerDirectoryName = "InstallerTemp";
-        private readonly NuGetUpgraderConfig nuGetUpgraderConfig;
 
         private InstallManifest installManifest;
-        private IPackageSearchMetadata highestVersionAvailable;
         private NuGetFeed nuGetFeed;
         private ICredentialStore credentialStore;
         private bool isNuGetFeedInitialized;
@@ -216,33 +218,33 @@ namespace GVFS.Common.NuGetUpgrade
                 IList<IPackageSearchMetadata> queryResults = this.QueryFeed(firstAttempt: true);
 
                 // Find the package with the highest version
-                IPackageSearchMetadata highestVersionAvailable = null;
+                IPackageSearchMetadata newestPackage = null;
                 foreach (IPackageSearchMetadata result in queryResults)
                 {
-                    if (highestVersionAvailable == null || result.Identity.Version > highestVersionAvailable.Identity.Version)
+                    if (newestPackage == null || result.Identity.Version > newestPackage.Identity.Version)
                     {
-                        highestVersionAvailable = result;
+                        newestPackage = result;
                     }
                 }
 
-                if (highestVersionAvailable != null &&
-                    highestVersionAvailable.Identity.Version.Version > this.installedVersion)
+                if (newestPackage != null &&
+                    newestPackage.Identity.Version.Version > this.installedVersion)
                 {
-                    this.highestVersionAvailable = highestVersionAvailable;
+                    this.highestVersionAvailable = newestPackage.Identity.Version.Version;
                 }
 
-                newVersion = this.highestVersionAvailable?.Identity?.Version?.Version;
+                newVersion = this.highestVersionAvailable;
 
                 if (newVersion != null)
                 {
                     this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - new version available: installedVersion: {this.installedVersion}, highestVersionAvailable: {newVersion}");
-                    message = $"New version {highestVersionAvailable.Identity.Version} is available.";
+                    message = $"New version {newestPackage.Identity.Version} is available.";
                     return true;
                 }
-                else if (highestVersionAvailable != null)
+                else if (newestPackage != null)
                 {
                     this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - up-to-date");
-                    message = $"highest version available is {highestVersionAvailable.Identity.Version}, you are up-to-date";
+                    message = $"highest version available is {newestPackage.Identity.Version}, you are up-to-date";
                     return true;
                 }
                 else
@@ -277,6 +279,11 @@ namespace GVFS.Common.NuGetUpgrade
                 return false;
             }
 
+            if (!this.EnsureNuGetFeedInitialized(out errorMessage))
+            {
+                return false;
+            }
+
             if (!this.TryCreateAndConfigureDownloadDirectory(this.tracer, out errorMessage))
             {
                 this.tracer.RelatedError($"{nameof(NuGetUpgrader)}.{nameof(this.TryCreateAndConfigureDownloadDirectory)} failed. {errorMessage}");
@@ -287,7 +294,15 @@ namespace GVFS.Common.NuGetUpgrade
             {
                 try
                 {
-                    this.DownloadedPackagePath = this.nuGetFeed.DownloadPackageAsync(this.highestVersionAvailable.Identity).GetAwaiter().GetResult();
+                    PackageIdentity packageId = this.GetPackageForVersion(this.highestVersionAvailable);
+
+                    if (packageId == null)
+                    {
+                        errorMessage = $"The specified version {this.highestVersionAvailable} was not found in the NuGet feed. Please check with your administrator to make sure the feed is set up correctly.";
+                        return false;
+                    }
+
+                    this.DownloadedPackagePath = this.nuGetFeed.DownloadPackageAsync(packageId).GetAwaiter().GetResult();
                 }
                 catch (Exception ex)
                 {
@@ -463,6 +478,23 @@ namespace GVFS.Common.NuGetUpgrade
             return "{" + tokenString + "}";
         }
 
+        private PackageIdentity GetPackageForVersion(Version version)
+        {
+            IList<IPackageSearchMetadata> queryResults = this.QueryFeed(firstAttempt: true);
+
+            IPackageSearchMetadata packageForVersion = null;
+            foreach (IPackageSearchMetadata result in queryResults)
+            {
+                if (result.Identity.Version.Version == version)
+                {
+                    packageForVersion = result;
+                    break;
+                }
+            }
+
+            return packageForVersion?.Identity;
+        }
+
         private bool TryGetPersonalAccessToken(string credentialUrl, ITracer tracer, out string token, out string error)
         {
             error = null;
@@ -619,8 +651,8 @@ namespace GVFS.Common.NuGetUpgrade
 
         public class NuGetUpgraderConfig
         {
-            private readonly ITracer tracer;
-            private readonly LocalGVFSConfig localConfig;
+            protected readonly ITracer tracer;
+            protected readonly LocalGVFSConfig localConfig;
 
             public NuGetUpgraderConfig(ITracer tracer, LocalGVFSConfig localGVFSConfig)
             {
@@ -648,7 +680,7 @@ namespace GVFS.Common.NuGetUpgrade
             /// NuGetUpgrader is considered ready if all required
             /// config settings are present.
             /// </summary>
-            public bool IsReady(out string error)
+            public virtual bool IsReady(out string error)
             {
                 if (string.IsNullOrEmpty(this.FeedUrl) ||
                     string.IsNullOrEmpty(this.PackageFeedName))
@@ -667,7 +699,7 @@ namespace GVFS.Common.NuGetUpgrade
             /// <summary>
             /// Check if the NuGetUpgrader is configured.
             /// </summary>
-            public bool IsConfigured(out string error)
+            public virtual bool IsConfigured(out string error)
             {
                 if (string.IsNullOrEmpty(this.FeedUrl) &&
                     string.IsNullOrEmpty(this.PackageFeedName))
@@ -686,7 +718,7 @@ namespace GVFS.Common.NuGetUpgrade
             /// <summary>
             /// Try to load the config for a NuGet upgrader. Returns false if there was an error reading the config.
             /// </summary>
-            public bool TryLoad(out string error)
+            public virtual bool TryLoad(out string error)
             {
                 string configValue;
                 if (!this.localConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.UpgradeFeedUrl, out configValue, out error))
--- a/GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
@@ -1,0 +1,262 @@
+using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
+using GVFS.Common.Tracing;
+using System;
+using System.Net.Http;
+using System.Runtime.Serialization;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace GVFS.Common.NuGetUpgrade
+{
+    public class OrgNuGetUpgrader : NuGetUpgrader
+    {
+        private HttpClient httpClient;
+        private string platform;
+
+        public OrgNuGetUpgrader(
+           string currentVersion,
+           ITracer tracer,
+           PhysicalFileSystem fileSystem,
+           HttpClient httpClient,
+           bool dryRun,
+           bool noVerify,
+           OrgNuGetUpgraderConfig config,
+           string downloadFolder,
+           string platform,
+           ICredentialStore credentialStore)
+           : base(
+               currentVersion,
+               tracer,
+               fileSystem,
+               dryRun,
+               noVerify,
+               config,
+               downloadFolder,
+               credentialStore)
+        {
+            this.httpClient = httpClient;
+            this.platform = platform;
+        }
+
+        public OrgNuGetUpgrader(
+           string currentVersion,
+           ITracer tracer,
+           PhysicalFileSystem fileSystem,
+           HttpClient httpClient,
+           bool dryRun,
+           bool noVerify,
+           OrgNuGetUpgraderConfig config,
+           string platform,
+           NuGetFeed nuGetFeed,
+           ICredentialStore credentialStore)
+           : base(
+               currentVersion,
+               tracer,
+               dryRun,
+               noVerify,
+               fileSystem,
+               config,
+               nuGetFeed,
+               credentialStore)
+        {
+            this.httpClient = httpClient;
+            this.platform = platform;
+        }
+
+        public override bool SupportsAnonymousVersionQuery { get => true; }
+
+        private OrgNuGetUpgraderConfig Config { get => this.nuGetUpgraderConfig as OrgNuGetUpgraderConfig;  }
+        private string OrgInfoServerUrl { get => this.Config.OrgInfoServer; }
+        private string Ring { get => this.Config.UpgradeRing; }
+
+        public static bool TryCreate(
+            ITracer tracer,
+            PhysicalFileSystem fileSystem,
+            LocalGVFSConfig gvfsConfig,
+            HttpClient httpClient,
+            ICredentialStore credentialStore,
+            bool dryRun,
+            bool noVerify,
+            out OrgNuGetUpgrader upgrader,
+            out string error)
+        {
+            OrgNuGetUpgraderConfig upgraderConfig = new OrgNuGetUpgraderConfig(tracer, gvfsConfig);
+            upgrader = null;
+
+            if (!upgraderConfig.TryLoad(out error))
+            {
+                upgrader = null;
+                return false;
+            }
+
+            if (!upgraderConfig.IsConfigured(out error))
+            {
+                return false;
+            }
+
+            if (!upgraderConfig.IsReady(out error))
+            {
+                return false;
+            }
+
+            string platform = GVFSPlatform.Instance.Name;
+
+            upgrader = new OrgNuGetUpgrader(
+                ProcessHelper.GetCurrentProcessVersion(),
+                tracer,
+                fileSystem,
+                httpClient,
+                dryRun,
+                noVerify,
+                upgraderConfig,
+                ProductUpgraderInfo.GetAssetDownloadsPath(),
+                platform,
+                credentialStore);
+
+            return true;
+        }
+
+        public override bool TryQueryNewestVersion(out Version newVersion, out string message)
+        {
+            newVersion = null;
+
+            if (!TryParseOrgFromNuGetFeedUrl(this.Config.FeedUrl, out string orgName))
+            {
+                message = "OrgNuGetUpgrader is not able to parse org name from NuGet Package Feed URL";
+                return false;
+            }
+
+            OrgInfoApiClient infoServer = new OrgInfoApiClient(this.httpClient, this.OrgInfoServerUrl);
+
+            try
+            {
+                this.highestVersionAvailable = infoServer.QueryNewestVersion(orgName, this.platform, this.Ring);
+            }
+            catch (Exception exception) when (exception is HttpRequestException ||
+                                              exception is TaskCanceledException)
+            {
+                // GetStringAsync can also throw a TaskCanceledException to indicate a timeout
+                // https://github.com/dotnet/corefx/issues/20296
+                message = string.Format("Network error: could not connect to server ({0}). {1}", this.OrgInfoServerUrl, exception.Message);
+                this.TraceException(exception, nameof(this.TryQueryNewestVersion), "Error connecting to server.");
+
+                return false;
+            }
+            catch (SerializationException exception)
+            {
+                message = string.Format("Parse error: could not parse response from server({0}). {1}", this.OrgInfoServerUrl, exception.Message);
+                this.TraceException(exception, nameof(this.TryQueryNewestVersion), "Error parsing response from server.");
+
+                return false;
+            }
+            catch (Exception exception) when (exception is ArgumentException ||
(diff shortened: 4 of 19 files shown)


GVFS/GVFS.Common/AzDevOpsFromNuGetFeedUrlParser.cs
```diff
@@ -0,0 +1,41 @@
+using System.Text.RegularExpressions;
```
jamill: Much of the logic for this class was moved from `NuGetUpgrader`

GVFS/GVFS.Common/OrgInfoApiClient.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            this.client = client;
+            this.baseUrl = baseUrl;
+        }
+
+        private string VersionUrl
+        {
+            get
+            {
+                return this.baseUrl + "/api/vfsforgitversion";
```
jamill: todo: update to new endpoint

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -104,6 +105,50 @@ private void TimerCallback(object unusedState)
...
+                            }                            
+                        }
+                        
+                        string message = string.Format(
+                            "{0}.{1}: Configured Product Upgrader does not support anonymous version queries.",
+                            nameof(ProductUpgradeTimer),
+                            nameof(this.TimerCallback),
+                            errorMessage);
+
+                        info.RecordHighestAvailableVersion(highestAvailableVersion: null);
```
jamill: TODO: return here

GVFS/GVFS.UnitTests/Common/AzDevOpsFromNuGetFeedUrlParserTests.cs
```diff
@@ -0,0 +1,35 @@
+using GVFS.Common;
```
jamill: Test cases moved from `NuGetUpgraderTests`

GVFS/GVFS.Common/AzDevOpsFromNuGetFeedUrlParser.cs
```diff
@@ -0,0 +1,41 @@
+using System.Text.RegularExpressions;
+
+namespace GVFS.Common
+{
+    public class AzDevOpsFromNuGetFeedUrlParser
+    {
+        public static bool TryParseOrgFromNuGetFeedUrl(string packageFeedUrl, out string orgName)
```
kewillford: Since these are `static` and will have the class name with the method name, I think we can make these a little shorter and remove the redundancy of `FromNuGetFeedUrl` so you would have  `AzDevOpsFromNuGetFeedUrlParser.TryParseOrg` and `AzDevOpsFromNuGetFeedUrlParser.TryCreate`.  Also the `TryCreate` doesn't feel like it fits if this class is a "parser".

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
```diff
@@ -0,0 +1,256 @@
...
+        public override bool SupportsAnonymousVersionQuery { get => true; }
+
+        private ModernNuGetUpgraderConfig Config { get => this.nuGetUpgraderConfig as ModernNuGetUpgraderConfig;  }
+        private string OrgInfoServerUrl { get => this.Config.OrgInfoServer; }
+        private string Ring { get => this.Config.UpgradeRing; }
+        private string OrgName
+        {
+            get
+            {
+                if (!AzDevOpsFromNuGetFeedUrlParser.TryParseOrgFromNuGetFeedUrl(this.Config.FeedUrl, out string orgName))
```
kewillford: This will run the regex everytime the code accesses the `OrgName` which is bad IMO.  Can we move this to the constructor so it only happens once?  Or will this be changing and we have to do this each time the `FeedUrl` changes?

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
```diff
@@ -0,0 +1,256 @@
...
+           string currentVersion,
+           ITracer tracer,
+           PhysicalFileSystem fileSystem,
+           HttpClient httpClient,
+           bool dryRun,
+           bool noVerify,
+           ModernNuGetUpgraderConfig config,
+           string downloadFolder,
+           string platform,
+           ICredentialStore credentialStore)
```
kewillford: The number of parameters for this class is a little out of control.  Can we have something like a `NuGetUpgraderOptions` that has the currentVersion, dryRun, noVerify, downloadFolder, platform, and maybe the values from the config to trim this down?

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
```diff
@@ -0,0 +1,256 @@
...
+            if (this.highestVersionAvailable != null &&
+                this.highestVersionAvailable > this.installedVersion)
+            {
+                newVersion = this.highestVersionAvailable;
+            }
+
+            if (newVersion != null)
+            {
+                this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - new version available: installedVersion: {this.installedVersion}, highestVersionAvailable: {newVersion}");
+                message = $"New version {this.highestVersionAvailable} is available.";
```
kewillford: Shouldn't this be using `newVersion` since that is what is getting the null check?  How do we know `this.highestVersionAvailable` isn't null here?

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
```diff
@@ -0,0 +1,256 @@
...
+            public override bool TryLoad(out string error)
+            {
+                if (!base.TryLoad(out error))
+                {
+                    return false;
+                }
+
+                if (!this.localConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.OrgInfoServerUrl, out string orgInfoServerUrl, out error))
+                {
+                    return false;
```
kewillford: Do we need to log anything here or when returning `false`?  Looks suspicious only logging in the one case.

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
```diff
@@ -0,0 +1,256 @@
...
+                    this.tracer.RelatedError(error);
+                    return false;
+                }
+
+                this.UpgradeRing = upgradeRing;
+
+                return true;
+            }
+
+            public override bool IsReady(out string error)
```
kewillford: Can we do better with this method have have it only output the required settings that were missing.  Something like:
`public override bool IsReady(out List<string> missingSettings)` and have the caller build the error string?  Then you could have the specific fields that were missing instead of saying "One or more".

GVFS/GVFS.Common/OrgInfoApiClient.cs
```diff
@@ -0,0 +1,70 @@
...
+                { "Ring", ring },
+            };
+
+            string responseString = this.client.GetStringAsync(this.ConstructRequest(this.VersionUrl, queryParams)).GetAwaiter().GetResult();
+            VersionResponse versionResponse = VersionResponse.FromJsonString(responseString);
+
+            return new Version(versionResponse.Version);
+        }
+
+        private string ConstructRequest(string baseUrl, Dictionary<string, string> queryParams)
```
kewillford: Since the `queryParams` are all hard coded it don't seem like we need this method and just use a `string.Format` or `$""`?

GVFS/GVFS.Common/OrgInfoApiClient.cs
```diff
@@ -0,0 +1,70 @@
...
+        public Version QueryNewestVersion(string orgName, string platform, string ring)
+        {
+            Dictionary<string, string> queryParams = new Dictionary<string, string>()
+            {
+                { "Organization", orgName },
+                { "Platform", platform },
+                { "Ring", ring },
+            };
+
+            string responseString = this.client.GetStringAsync(this.ConstructRequest(this.VersionUrl, queryParams)).GetAwaiter().GetResult();
```
kewillford: Do we need to escape anything that will be in the Url?

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -83,37 +87,71 @@ protected ProductUpgrader()
...
+                        out error))
+                    {
+                        // We were successfully able to load a NuGetUpgrader - use that.
+                        newUpgrader = modernNuGetUpgrader;
+                        return true;
+                    }
+                    else
+                    {
+                        tracer.RelatedError($"{nameof(TryCreateUpgrader)}: Could not create upgrader. {error}");
+                        newUpgrader = null;
```
kewillford: Should this fallback to the NugetUpgrader or since there was the OrgInfo key we want to fail?

GVFS/GVFS.Common/VersionResponse.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace GVFS.Common
+{
+    public class VersionResponse
+    {
+        public string Version { get; set; }
+
+        public static VersionResponse FromJsonString(string jsonString)
+        {
+            JsonSerializer serializer = new JsonSerializer();
+            return JsonConvert.DeserializeObject<VersionResponse>(jsonString);
```
kewillford: `serializer` isn't used and I would remove this method and just call the `JsonConvert.DeserializeObject<VersionResponse>` inline.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -104,6 +105,50 @@ private void TimerCallback(object unusedState)
...
+                                info.RecordHighestAvailableVersion(highestAvailableVersion: null);
+                                return;
+                            }                            
+                        }
+                        
+                        string message = string.Format(
+                            "{0}.{1}: Configured Product Upgrader does not support anonymous version queries.",
+                            nameof(ProductUpgradeTimer),
+                            nameof(this.TimerCallback),
+                            errorMessage);
```
kewillford: Where is this `message` used?

GVFS/GVFS.Service/ProductUpgradeTimer.cs
```diff
@@ -78,19 +78,20 @@ private void TimerCallback(object unusedState)
...
                         error: out errorMessage);
 
                     if (productUpgrader == null)
                     {
                         string message = string.Format(
-                            "{0}.{1}: GitHubUpgrader.Create failed to create upgrader: {2}",
+                            "{0}.{1}: failed to create upgrader: {2}",
                             nameof(ProductUpgradeTimer),
                             nameof(this.TimerCallback),
                             errorMessage);
```
kewillford: All of these formats using `nameof(ProductUpgradeTimer), nameof(this.TimerCallback)` make me think we need a helper method to contruct this message.  They are all very similar.  Maybe we should be using the `EventMetadata` and message should be something simple like "Failed to create upgrader"?

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
jamill: It could be a lazy property / cached. It is a private property that is only accessed 1 time, so is not actually an issue. It could also be a method (`GetOrgName()`) or something...

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
jamill: yeah - `newVersion` is probably more clear.

> How do we know this.highestVersionAvailable isn't null here?

 if `newVersion` is not null, then `newVersion == highestVersionAvailable`..

I will update to `newVersion`

GVFS/GVFS.Common/OrgInfoApiClient.cs
jamill: Yeah - as is, the params are all hard coded / unchanging. There are also some optional parameters that have been discussed, that are not included in this PR.

GVFS/GVFS.Common/ProductUpgrader.cs
jamill: The more I think about it - I think the current NuGetUpgrader will go away (either with this PR or soon after), and `ModernNuGetUpgrader` will just become `NuGetUpgrader`.

I do want to fail, and not fall back. In the case where the user "intends" to use one version of the upgrader, I didn't want to fall back to a different version "unexpectedly... and, I expect the fallback option to go away :)

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
```diff
@@ -0,0 +1,256 @@
...
+using GVFS.Common.Tracing;
+using System;
+using System.Net.Http;
+using System.Runtime.InteropServices;
+using System.Runtime.Serialization;
+using System.Threading.Tasks;
+
+namespace GVFS.Common.NuGetUpgrade
+{
+    public class ModernNuGetUpgrader : NuGetUpgrader
```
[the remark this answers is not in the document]
jamill: I like `OrgNugetUpgrader` - Modern was more of a placeholder.

GVFS/GVFS.Common/AzDevOpsFromNuGetFeedUrlParser.cs
jamill: I wanted something a little more specific / better than `AzDevOpsNuGetFeedUrlUtils`... `AzDevOrgFromNuGetUrlExtractor`? `AzDevOrgUrlCreator`?

GVFS/GVFS.Common/OrgInfoApiClient.cs
jamill: yes, we do

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        private interface IHttpMessageHandlerProtectedMembers
+        {
+            Task<HttpResponseMessage> SendAsync(HttpRequestMessage message, CancellationToken token);
+        }
+
+        public static IEnumerable<Exception> NetworkFailureCases()
+        {
+            yield return new HttpRequestException("Response status code does not indicate success: 401: (Unauthorized)");
+            yield return new TaskCanceledException("Task canceled");
```
wilbaker: Is it also possible to get a `WebException`? (I noticed we're catching that exception in `HttpRequestor.SendRequest`, but I'm not sure it could get thrown in this code path)

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -83,37 +87,71 @@ protected ProductUpgrader()
...
+                {
+                    if (NuGetUpgrader.TryCreate(tracer, fileSystem, gvfsConfig, credentialStore, dryRun, noVerify, out NuGetUpgrader nuGetUpgrader, out bool isConfigured, out error))
+                    {
+                        // We were successfully able to load a NuGetUpgrader - use that.
+                        newUpgrader = nuGetUpgrader;
+                        return true;
+                    }
+                    else
+                    {
+                        tracer.RelatedError($"{nameof(TryCreateUpgrader)}: Could not create upgrader. {error}");
```
wilbaker: Make this `RelatedError` text and the text above (on line 123) different to guarantee we know which error case was hit?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -275,6 +255,11 @@ public override bool TryDownloadNewestVersion(out string errorMessage)
                 return false;
             }
 
+            if (!this.EnsureNuGetFeedInitialized(out errorMessage))
```
wilbaker: We are calling `EnsureNuGetFeedInitialized` above on line 188 as well.  Do we need to call this twice?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -285,7 +270,15 @@ public override bool TryDownloadNewestVersion(out string errorMessage)
             {
                 try
                 {
-                    this.DownloadedPackagePath = this.nuGetFeed.DownloadPackageAsync(this.highestVersionAvailable.Identity).GetAwaiter().GetResult();
+                    PackageIdentity packageId = this.GetPackageForVersion(this.highestVersionAvailable);
+
+                    if (packageId == null)
+                    {
+                        errorMessage = "Could not find package for version. This indicates the package feed is out of sync.";
```
wilbaker: > the package feed is out of sync

Out of sync with what?

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,244 @@
...
+            {
+                return false;
+            }
+
+            if (!upgraderConfig.IsReady(out error))
+            {
+                return false;
+            }
+
+            string platform = RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "macOS" : "windows";
```
wilbaker: We've tried to avoid adding calls to `RuntimeInformation.IsOSPlatform` in VFS4G.  Rather than calling `IsOSPlatform` could you add a new `GetPlatformName` type method to `GVFSPlatform`?

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        public override bool TryQueryNewestVersion(out Version newVersion, out string message)
+        {
+            newVersion = null;
+
+            if (!AzDevOpsOrgFromNuGetFeed.TryParseOrg(this.Config.FeedUrl, out string orgName))
+            {
+                message = "OrgNuGetUpgrader is not able to parse org name from NuGet Package Feed URL";
+                return false;
```
wilbaker: Should we trace anything here, or will this issue get traced somewhere else?

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,244 @@
...
+            OrgInfoApiClient infoServer = new OrgInfoApiClient(this.httpClient, this.OrgInfoServerUrl);
+
+            try
+            {
+                this.highestVersionAvailable = infoServer.QueryNewestVersion(orgName, this.platform, this.Ring);
+            }
+            catch (HttpRequestException exception)
+            {
+                message = string.Format("Network error: could not connect to server ({0}). {1}", this.OrgInfoServerUrl, exception.Message);
+                this.TraceException(exception, nameof(this.TryQueryNewestVersion), $"Error connecting to server.");
```
wilbaker: Not a big deal, but here and in some of the catches below we are using the `$` string formatter with no variables in the string.  Not sure if you want to leave that in.

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,244 @@
...
+            if (newVersion != null)
+            {
+                this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - new version available: installedVersion: {this.installedVersion}, highestVersionAvailable: {newVersion}");
+                message = $"New version {newVersion} is available.";
+                return true;
+            }
+            else if (this.highestVersionAvailable != null)
+            {
+                this.tracer.RelatedInfo($"{nameof(this.TryQueryNewestVersion)} - up-to-date");
+                message = $"highest version available is {this.highestVersionAvailable}, you are up-to-date";
```
wilbaker: Nit: The other messages start with a capital letter

GVFS/GVFS.UnitTests/Common/OrgInfoApiClientTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            Task<HttpResponseMessage>  SendAsync(HttpRequestMessage message, CancellationToken token);
+        }
+
+        [TestCaseSource("TestOrgInfo")]
+        public void QueryNewestVersionWithParams(OrgInfo orgInfo)
+        {
+            Mock<HttpMessageHandler> handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
+
+            handlerMock.Protected().As<IHttpMessageHandlerProtectedMembers>()
```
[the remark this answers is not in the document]
jamill: This gives us access to protected members of the type being mocked.

GVFS/GVFS.UnitTests/Common/OrgInfoApiClientTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                .Setup(m => m.SendAsync(It.Is<HttpRequestMessage>(request => this.UriMatches(request.RequestUri, this.baseUrl, orgInfo.OrgName, orgInfo.Platform, orgInfo.Ring)), It.IsAny<CancellationToken>()))
```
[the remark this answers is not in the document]
jamill: > What is the impact of it returning true or false?

This indicates whether a call matches the "setup". If a call is made to `SendAsync` that does not match the expected parameters, then this setup will not apply. This is to verify that the `SendAsync` method is made with the expected parameters.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: With the `OrgNuGetUpgrader`, it is possible to call "TryDownloadNewestVersion" without calling the  "TryQueryNewestVersion" implementation in this method.

I have a refactoring change coming up that should make this a bit clearer.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
jamill: I will clarifying this message... but, it means that the version it has been instructed to download does not exist in the feed (e.g., the version returned by the `orgInfoServer` is out of sync with what is available in the feed)

GVFS/GVFS.UnitTests/Common/OrgInfoApiClientTests.cs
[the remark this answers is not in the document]
wilbaker: > then this setup will not apply

If that happens is an exception thrown, or is our mock a no-op?

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
jamill: > Rather than calling IsOSPlatform could you add a new GetPlatformName type method to GVFSPlatform?

I can move this logic, however - my concern is that `GVFSPlatform` introduces some questions with testing. Ideally, I would like to be able to test with different platforms. Is that possible with the static GVFSPlatform instance?

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
wilbaker: Oops I totally missed that we were in a different method here

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
wilbaker: Currently the unit tests register `MockPlatform` as the active `GVFSPlatform`:

```
    [SetUpFixture]
    public class Setup
    {
        [OneTimeSetUp]
        public void SetUp()
        {
            GVFSPlatform.Register(new MockPlatform());
        }
    }
```

And so you should be able to add a `SetCurrentPlatform` type method to `MockPlatform` to allow you to set different platforms for testing.

wilbaker: (This code is in GVFS/GVFS.UnitTests/Setup.cs)

As an alternative/enhancement we could add another method to `GVFSPlatform` that acts like `Register` but can be called multiple times (for unit testing only) so that we could swap in different mocks.

jamill: > And so you should be able to add a SetCurrentPlatform type method to MockPlatform to allow you to set different platforms for testing.

Does this get cleared / reset between each test? or wpuld I run the risk of polluting / affecting other tests?

wilbaker: Based on this:

> The OneTimeSetUp method in a SetUpFixture is executed once before any of the fixtures contained in its namespace. The OneTimeTearDown method is executed once after all the fixtures have completed execution.

https://github.com/nunit/docs/wiki/SetUpFixture-Attribute

I would expect it's a one time thing.  However, until we have something like this:

> As an alternative/enhancement we could add another method to GVFSPlatform that acts like Register but can be called multiple times (for unit testing only) so that we could swap in different mocks.

We could use a per-test Teardown in the fixture(s) that call `SetCurrentPlatform` to ensure we leave the platform name with its expected value.

GVFS/GVFS.UnitTests/Common/OrgInfoApiClientTests.cs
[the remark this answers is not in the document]
jamill: If that happens, it is a no-op. However, the test will fail as:

1) We have configured the mock with `Behavior.Strict`, so test will failDue to an unexpected method call,
2) If this expectation that we are not setting up is not met, the test should fail (unless there was another call that did match the args)

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
jamill: Good question - according to the msdn documentation, this is not one of the exceptions thrown by HttpClient. It looks like `WebClient.Download` could throw this exception (e.g. as used in `GitHubUpgrader`, but not `HttpClient.GetStringAsync`...

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,244 @@
...
+                this.highestVersionAvailable = infoServer.QueryNewestVersion(orgName, this.platform, this.Ring);
+            }
+            catch (HttpRequestException exception)
+            {
+                message = string.Format("Network error: could not connect to server ({0}). {1}", this.OrgInfoServerUrl, exception.Message);
+                this.TraceException(exception, nameof(this.TryQueryNewestVersion), $"Error connecting to server.");
+
+                return false;
+            }
+            catch (TaskCanceledException exception)
```
kewillford: Can we combine this with the `HttpRequestExceptino since both handlers are exactly the same?

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,244 @@
...
+                    return false;
+                }
+
+                if (!this.localConfig.TryGetConfig(GVFSConstants.LocalGVFSConfig.OrgInfoServerUrl, out string orgInfoServerUrl, out error))
+                {
+                    this.tracer.RelatedError(error);
+                    return false;
+                }
+
+                this.OrgInfoServer = orgInfoServerUrl;
```
kewillford: Can we have a method in the base class that is something like:
```
protected bool TrySetConfig(string key, Action<string> setValue, out string error)
{
	if (!this.localConfig.TryGetConfig(key, out string value, out error))
	{
		this.tracer.RelatedError(error);
		return false;
	}

	setValue(value);
}
```

Then this method could have something like:
```
   return base.TryLoad(out error) &&
        this.TrySetConfig(GVFSConstants.LocalGVFSConfig.OrgInfoServerUrl, x => this.OrgInfoServer = x, out error) &&
        this.TrySetConfig(GVFSConstants.LocalGVFSConfig.UpgradeRing, x => this.UpgradeRing= x, out error) &&
        etc...
```

The base class could do the same thing.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -83,37 +87,71 @@ protected ProductUpgrader()
...
+                        newUpgrader = null;
+                        return false;
+                    }
+                }
+                else
+                {
+                    if (NuGetUpgrader.TryCreate(tracer, fileSystem, gvfsConfig, credentialStore, dryRun, noVerify, out NuGetUpgrader nuGetUpgrader, out bool isConfigured, out error))
```
kewillford: multiple lines so it matches the `TryCreate` above.

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
```diff
@@ -0,0 +1,175 @@
...
+{
+    [TestFixture]
+    public class OrgNuGetUpgraderTests
+    {
+        private const string CurrentVersion = "1.5.1185.0";
+        private const string NewerVersion = "1.6.1185.0";
+
+        private const string DefaultUpgradeFeedPackageName = "package";
+        private const string DefaultUpgradeFeedUrl = "https://pkgs.dev.azure.com/contoso/";
+        private const string DefaultOrgInfoServerUrl = "https://www.contoso.com";
```
kewillford: Why do these need to be real URLs in a unit test?

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
```diff
@@ -0,0 +1,175 @@
...
+            Version newVersion;
+            string message;
+
+            this.httpMessageHandlerMock.Protected().As<IHttpMessageHandlerProtectedMembers>()
+                .Setup(m => m.SendAsync(It.IsAny<HttpRequestMessage>(), It.IsAny<CancellationToken>()))
+                .ReturnsAsync(new HttpResponseMessage()
+                {
+                    StatusCode = HttpStatusCode.OK,
+                    Content = new StringContent(this.ConstructResponseContent(NewerVersion))
+                });
```
kewillford: Why do we need to set this up the exact same as in the Setup?  Can we remove one?

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            bool success = this.upgrader.TryQueryNewestVersion(out newVersion, out message);
+
+            success.ShouldBeTrue();
+            newVersion.ShouldNotBeNull();
+            newVersion.ShouldEqual<Version>(new Version(NewerVersion));
+            message.ShouldNotBeNull();
```
kewillford: Can we compare against an exact message that we expect?

GVFS/GVFS.UnitTests/Common/OrgInfoApiClientTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+                queryParams.Add(key, value);
+            }
+
+            if (queryParams.Count != 3)
+            {
+                return false;
+            }
+
+            if (!queryParams.TryGetValue("Organization", out string orgName) || !string.Equals(orgName, expectedOrgName, StringComparison.OrdinalIgnoreCase))
```
kewillford: Can this just use a `ShouldEqual` and have the test fail?

GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
```diff
@@ -132,6 +155,50 @@ public void NoUpgraderWhenNuGetFeedMissing()
...
+                gvfsConfig,
+                this.credentialStoreMock.Object,
+                false,
+                false,
+                out ProductUpgrader productUpgrader,
+                out string error);
+
+            success.ShouldBeFalse();
+            productUpgrader.ShouldBeNull();
+            error.ShouldNotBeNull();
```
kewillford: Do we need to validate what the value of `error` is?

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
jamill: sure - I can do something like:

```
catch (Exception ex) where (ex is TaskCanceledExcption || ex is HttpRequestException)
{
...
}
```

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
jamill: `DefaultUpgradeFeedUrl` is parsed and interpreted, so it needs to follow some requirements. Right now, this pattern is hardcoded, but is something we could look at relaxing.

`DefaultOrgInfoServerUrl` could probably be anything.

GVFS/GVFS.UnitTests/Common/TryCreateProductUpgraderTests.cs
jamill: I am not against it, but I think most of the value is captured by testing that there is an error. If I test for the exact value of the error message, it makes it harder to change the details of the message.

GVFS/GVFS.UnitTests/Common/OrgInfoApiClientTests.cs
jamill: This is used in determining whether a Moq setup matches a call. If the setup expectations do not match, then the test will fail elsewhere.

kewillford: But will we know elsewhere that it was because the orgName didn't match the expectedOrgName or will it be some obscure failure that will be hard to determine what failed?

jamill: We will know that the method was not called with the expected parameters, which I think is pretty close. In my experience, it also prints out the parameters from the actual calls (although, I am not sure exactly it prints out in this case).

GVFS/GVFS.Common/AzDevOpsOrgFromNuGetFeed.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace GVFS.Common
+{
+    public class AzDevOpsOrgFromNuGetFeed
+    {
+        /// <summary>
+        /// Given a URL for a NuGet feed hosted on Azure DevOps,
+        /// return the organization that hosts the feed.
+        /// </summary>
+        public static bool TryParseOrg(string packageFeedUrl, out string orgName)
```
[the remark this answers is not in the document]
jamill: Interesting point. I would rather not add the responsibility of parsing these strings to the `NuGetUpgraderConfig` class.

I can:
1) Make these instance methods on the `AzDevOpsOrgFromNuGetFeed` class (until I find a better name for this class). This would address the question about calling the regex multiple times.

2) Bring these back to static methods on the `NuGetUpgrader` class, or even work it to make it only called 1 time (possible as a property of NuGetUpgrader...)

However, This is not being called in a perf critical code currently, and I don't think this is likely to be called multiple times in a perf critical section...

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
jamill: I went and verified that this will get traced by callers.

GVFS/GVFS.UnitTests/Common/NuGetUpgrade/OrgNuGetUpgraderTests.cs
jamill: Yes - thanks!

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
jamill: I think longer term we should revisit this pattern, as it makes it harder to test, and consider alternate patterns that are easier to test. For now, I went ahead and added properties off of the GVFS platform.

GVFS/GVFS.Service/ProductUpgradeTimer.cs
jamill: I like this suggestion - I might hold off on making this change as this code should change at some point (as we transition to deprecate the existing nuget upgrader logic)

GVFS/GVFS.Common/NuGetUpgrade/ModernNuGetUpgrader.cs
jamill: I agree - I have some thoughts on how to tame this, but I would like to tackle that in a seperate PR.

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
```diff
@@ -0,0 +1,242 @@
...
+                return false;
+            }
+            catch (Newtonsoft.Json.JsonReaderException exception)
+            {
+                message = string.Format("Parse error: could not parse response from server ({0}). {1}", this.OrgInfoServerUrl, exception.Message);
+                this.TraceException(exception, nameof(this.TryQueryNewestVersion), "Error parsing response from server.");
+
+                return false;
+            }
+            catch (Exception exception) when (exception is ArgumentException ||
```
jamill: I am not super fond of this error handling logic here, and have some ideas on how to clean this up moving forward, but I think this is sufficient for the current state of things.

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
jamill: I like this suggestion - but, I think I want to hold off on this change:

1) I want to make a bit larger change on this logic

2) This is not an area that we expect end users to configure directly. I want to make it more user friendly, but I think I can follow up with this work.

GVFS/GVFS.Common/NuGetUpgrade/OrgNuGetUpgrader.cs
wilbaker: > I think longer term we should revisit this pattern, as it makes it harder to test, and consider alternate patterns that are easier to test. For now, I went ahead and added properties off of the GVFS platform.

I'm happy to do some brainstorming in this area with you for a future PR.  

If there is a way we can keep runtime platform checks out of the VFS4G code and make the code easier to test that sounds like a win.

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -88,6 +89,9 @@ public bool TryGetNormalizedPathRoot(string path, out string pathRoot, out strin
 
         public class GVFSPlatformConstants
         {
+            public const string WindowsPlatformName = "Windows";
+            public const string MacPlatformName = "macOS";
```
wilbaker: Can you move these into `WindowsPlatform` and `MacPlatform`. 

`GVFSPlatform` should not know/care which platforms implement it.

(If you want the property to live in `GVFSPlatformConstants` you could take the same approach used for `installerExtension` and have the value passed in via the `GVFSPlatform` constructor)

PR #1041, opened 2019-04-16
Update PrjFSKextLogDaemon to monitor and record vnode cache health

### LINKED ISSUE #973: Mac: Periodically poll the state of the vnode cache and record the results
See discussion in #715 

Some interesting data points:

- Cache utilization %
- Collision rate
- Cache invalidations

wilbaker:
Resolves #973 

Updated PrjFSKextLogDaemon to poll the kext every 30 minutes and request information about the health/status of the vnode cache.  This information is then recorded in the macOS logs (where it can be queried by `gvfs diagnose`).

Also updated the vnode cache unit tests to make use of the new stats to perform additional validation.

**Remaining Work**
- [X] Additional local testing of the changes
- [X] Compare large build times with and without this change

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VnodeCache.cpp
+++ b/ProjFS.Mac/PrjFSKext/VnodeCache.cpp
@@ -5,7 +5,7 @@
 #include "Memory.hpp"
 #include "KextLog.hpp"
 #include "../PrjFSKext/public/PrjFSCommon.h"
-#include "../PrjFSKext/public/PrjFSHealthData.h"
+#include "../PrjFSKext/public/PrjFSVnodeCacheHealth.h"
 
 #include "VnodeCachePrivate.hpp"
 
@@ -230,7 +230,7 @@ void VnodeCache_InvalidateCache(PerfTracer* _Nonnull perfTracer)
 
 IOReturn VnodeCache_ExportHealthData(IOExternalMethodArguments* _Nonnull arguments)
 {
-    PrjFSHealthData healthData =
+    PrjFSVnodeCacheHealth healthData =
     {
         .cacheCapacity = s_entriesCapacity,
         .cacheEntries = s_cacheStats.cacheEntries, // cacheEntries is reset to 0 when VnodeCache_InvalidateCache is called
@@ -267,7 +267,7 @@ IOReturn VnodeCache_ExportHealthData(IOExternalMethodArguments* _Nonnull argumen
         return result;
     }
 
-    if (arguments->structureOutput == nullptr || arguments->structureOutputSize != sizeof(PrjFSHealthData))
+    if (arguments->structureOutput == nullptr || arguments->structureOutputSize != sizeof(PrjFSVnodeCacheHealth))
     {
         KextLog("VnodeCache_ExportHealthData: structure output size %u, expected %lu\n", arguments->structureOutputSize, sizeof(healthData));
         return kIOReturnBadArgument;
--- a/ProjFS.Mac/PrjFSKext/VnodeCache.cpp
+++ b/ProjFS.Mac/PrjFSKext/VnodeCache.cpp
@@ -280,7 +280,7 @@ IOReturn VnodeCache_ExportHealthData(IOExternalMethodArguments* _Nonnull argumen
 KEXT_STATIC_INLINE void InvalidateCache_ExclusiveLocked()
 {
     memset(s_entries, 0, s_entriesCapacity * sizeof(VnodeCacheEntry));
-    atomic_exchange_explicit(&s_cacheStats.cacheEntries, 0U, memory_order_relaxed);
+    atomic_store_explicit(&s_cacheStats.cacheEntries, 0U, memory_order_relaxed);
 }
 
 KEXT_STATIC_INLINE uint32_t ComputePow2CacheCapacity(int expectedVnodeCount)
@@ -509,11 +509,11 @@ KEXT_STATIC bool TryInsertOrUpdateEntry_ExclusiveLocked(
 
 KEXT_STATIC_INLINE void InitCacheStats()
 {
-    atomic_exchange_explicit(&s_cacheStats.cacheEntries, 0U, memory_order_relaxed);
+    atomic_store_explicit(&s_cacheStats.cacheEntries, 0U, memory_order_relaxed);
     
     for (int32_t i = 0; i < VnodeCacheHealthStat_Count; ++i)
     {
-        atomic_exchange_explicit(&s_cacheStats.healthStats[i], 0ULL, memory_order_relaxed);
+        atomic_store_explicit(&s_cacheStats.healthStats[i], 0ULL, memory_order_relaxed);
     }
 }
 
@@ -530,6 +530,6 @@ KEXT_STATIC_INLINE void AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat healt
                 VnodeCacheHealthStatNames[healthStat],
                 statValue);
         
-        atomic_exchange_explicit(&s_cacheStats.healthStats[healthStat], 0ULL, memory_order_relaxed);
+        atomic_store_explicit(&s_cacheStats.healthStats[healthStat], 0ULL, memory_order_relaxed);
     }
 }
--- a/ProjFS.Mac/PrjFSKext/VnodeCache.cpp
+++ b/ProjFS.Mac/PrjFSKext/VnodeCache.cpp
@@ -1,9 +1,11 @@
 #include <string.h>
+#include <IOKit/IOUserClient.h>
 #include "Locks.hpp"
 #include "VnodeCache.hpp"
 #include "Memory.hpp"
 #include "KextLog.hpp"
 #include "../PrjFSKext/public/PrjFSCommon.h"
+#include "../PrjFSKext/public/PrjFSHealthData.h"
 
 #include "VnodeCachePrivate.hpp"
 
@@ -53,12 +55,17 @@ KEXT_STATIC bool TryInsertOrUpdateEntry_ExclusiveLocked(
     bool forceRefreshEntry,
     VirtualizationRootHandle rootHandle);
 
+KEXT_STATIC_INLINE void InitCacheStats();
+KEXT_STATIC_INLINE void AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat healthStat, uint64_t value);
+
 KEXT_STATIC uint32_t s_entriesCapacity;
 KEXT_STATIC VnodeCacheEntry* s_entries;
 
 // s_entriesCapacity will always be a power of 2, and so we can compute the modulo
 // using (value & s_ModBitmask) rather than (value % s_entriesCapacity);
 KEXT_STATIC uintptr_t s_ModBitmask;
+
+KEXT_STATIC VnodeCacheStats s_cacheStats;
 
 static RWLock s_entriesLock;
 
@@ -86,6 +93,8 @@ kern_return_t VnodeCache_Init()
     }
     
     memset(s_entries, 0, s_entriesCapacity * sizeof(VnodeCacheEntry));
+    
+    InitCacheStats();
     
     PerfTracing_RecordSample(PrjFSPerfCounter_CacheCapacity, 0, s_entriesCapacity);
     
@@ -126,10 +135,13 @@ VirtualizationRootHandle VnodeCache_FindRootForVnode(
     if (TryGetVnodeRootFromCache(vnode, vnodeHashIndex, vnodeVid, rootHandle))
     {
         perfTracer->IncrementCount(cacheHitCounter, true /*ignoreSampling*/);
+        AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalFindRootForVnodeHits, 1ULL);
         return rootHandle;
     }
     
     perfTracer->IncrementCount(cacheMissCounter, true /*ignoreSampling*/);
+    AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalFindRootForVnodeMisses, 1ULL);
+    
     FindVnodeRootFromDiskAndUpdateCache(
         perfTracer,
         cacheMissFallbackFunctionCounter,
@@ -158,6 +170,8 @@ VirtualizationRootHandle VnodeCache_RefreshRootForVnode(
     uint32_t vnodeVid = vnode_vid(vnode);
     
     perfTracer->IncrementCount(cacheMissCounter, true /*ignoreSampling*/);
+    AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalRefreshRootForVnode, 1ULL);
+    
     FindVnodeRootFromDiskAndUpdateCache(
         perfTracer,
         cacheMissFallbackFunctionCounter,
@@ -186,6 +200,8 @@ VirtualizationRootHandle VnodeCache_InvalidateVnodeRootAndGetLatestRoot(
     uint32_t vnodeVid = vnode_vid(vnode);
     
     perfTracer->IncrementCount(cacheMissCounter, true /*ignoreSampling*/);
+    AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalInvalidateVnodeRoot, 1ULL);
+    
     FindVnodeRootFromDiskAndUpdateCache(
         perfTracer,
         cacheMissFallbackFunctionCounter,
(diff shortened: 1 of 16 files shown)


ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -203,6 +219,8 @@ VirtualizationRootHandle VnodeCache_InvalidateVnodeRootAndGetLatestRoot(
 void VnodeCache_InvalidateCache(PerfTracer* _Nonnull perfTracer)
 {
     perfTracer->IncrementCount(PrjFSPerfCounter_CacheInvalidateCount, true /*ignoreSampling*/);
+    AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_InvalidateEntireCacheCount, 1ULL);
+    atomic_exchange(&s_cacheStats.cacheEntries, 0U);
```
wilbaker: Move this `atomic_exchange` inside the exclusive lock?

wilbaker:
Large builds show no perf hit from these changes.

**Large Build**
*Master*
https://mseng.visualstudio.com/AzureDevOps/_build/results?buildId=9270493
*With PR Changes*
https://mseng.visualstudio.com/AzureDevOps/_build/results?buildId=9269317

**Large Build Prefetch + Hydrate**
*Master*
https://mseng.visualstudio.com/AzureDevOps/_build/results?buildId=9276907
*With PR Changes*
https://mseng.visualstudio.com/AzureDevOps/_build/results?buildId=9278800

ProjFS.Mac/PrjFSKext/public/PrjFSHealthData.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+struct PrjFSHealthData
```
jeschu1: Right now we only have vnode cache data.  Do we expect to expand it to other data?  If not we may want to make the name more specific.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -211,9 +228,58 @@ void VnodeCache_InvalidateCache(PerfTracer* _Nonnull perfTracer)
...
+    }
+
+    memcpy(arguments->structureOutput, &healthData, sizeof(healthData));
+    return kIOReturnSuccess;
+}
+
 KEXT_STATIC_INLINE void InvalidateCache_ExclusiveLocked()
 {
     memset(s_entries, 0, s_entriesCapacity * sizeof(VnodeCacheEntry));
+    atomic_exchange(&s_cacheStats.cacheEntries, 0U);
```
jeschu1: Does this need to be atomic_exchange if it's under an exclusive lock?

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -428,3 +505,30 @@ KEXT_STATIC bool TryInsertOrUpdateEntry_ExclusiveLocked(
     
     return false;
 }
+
+KEXT_STATIC_INLINE void InitCacheStats()
+{
+    atomic_exchange(&s_cacheStats.cacheEntries, 0U);
```
jeschu1: Is this function called multi-threaded?  If not does it have to be an assignment?
Are there any perf impacts from using atomic_exchange?

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -367,13 +433,17 @@ KEXT_STATIC bool TryFindVnodeIndex_Locked(
     /* out parameters */
```
jeschu1: Food for thought, it would be nice to know this is only called under a shared lock.  Not part of this PR, but throwing it out there for readability in the future.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -367,13 +433,17 @@ KEXT_STATIC bool TryFindVnodeIndex_Locked(
...
     uintptr_t& vnodeIndex)
 {
+    AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalCacheLookups, 1ULL);
```
jeschu1: Since we're under a shared lock I understand the driver for making this operation atomic.
Where we are under an exclusive lock do we need to be there atomic too?

ProjFS.Mac/PrjFSLib/prjfs-log/prjfs-log.cpp
```diff
@@ -79,14 +79,8 @@ static void ProcessLogMessagesOnConnection(io_connect_t connection, io_service_t
     dispatch_source_set_event_handler(logState->dataQueue.dispatchSource, ^{
         DataQueue_ClearMachNotification(logState->dataQueue.notificationPort);
         
-        while(true)
+        while(IODataQueueEntry* entry = DataQueue_Peek(logState->dataQueue.queueMemory))
```
jeschu1: I *think* this is unrelated code cleanup.  As a general rule I like new PRs for that sort of stuff and am happy to jump on those quickly!

Also to double check, entry = nullptr will break the loop

wilbaker:
@jeschu1 I decided to push the commit that uses `memory_order_relaxed` so you can take a look at that option and let me know what you think.  More info on this ordering options is available here:

https://en.cppreference.com/w/c/language/atomic
https://en.cppreference.com/w/c/atomic/memory_order

ProjFS.Mac/PrjFSLib/prjfs-log/prjfs-log.cpp
wilbaker: > I think this is unrelated code cleanup. As a general rule I like new PRs for that sort of stuff and am happy to jump on those quickly!

Yep it's code cleanup and I am generally in the same boat 😄 

In this instance I was getting confused over why the main loop of prjfs-log and the logging daemon looked so different.  After determining they were doing the same thing I updated prjfs-log to use the same pattern as the daemon, please let me know if you'd rather I back that change out.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
wilbaker: I'm glad you pointed out this line as we should be using `atomic_store` here rather than `atomic_exchange`.

Based on what I could find online `=` is equivalent to `atomic_store`, and so if we switch to using `atomic_store_exchange` with weak ordering it might be a bit faster that using `=`.  (I would also guess/hope the compiler would see that we're in an exclusive lock and do the right thing here).

References:
https://stackoverflow.com/questions/34915926/is-assignment-equivalent-to-load-store-for-stdatomicbool
https://en.cppreference.com/w/c/language/atomic

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
wilbaker: See related comment, I can update these to using `atomic_store_explicit` with `memory_order_relaxed` which should be a cheaper operation than `=`.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
wilbaker: This function is actually called under both shared and exclusive locks:

```
KEXT_STATIC bool TryInsertOrUpdateEntry_ExclusiveLocked(
    vnode_t _Nonnull vnode,
    uintptr_t vnodeHashIndex,
    uint32_t vnodeVid,
    bool forceRefreshEntry,
    VirtualizationRootHandle rootHandle)
{
    uintptr_t vnodeIndex;
    if (TryFindVnodeIndex_Locked(vnode, vnodeHashIndex, /*out*/ vnodeIndex))
    {
```

Do you think a longer name (e.g. _SharedOrExclusiveLocked) would be better?

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
wilbaker: See related discussion in earlier comment, because the stats have atomic types `++` and `+=` will also be atomic.

I went with `atomic_fetch_add` to make it explicit that we expect these to be atomic types.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
jeschu1: That would be helpful for me yes, but toally scope creep for this PR :-).  I can do it as a small separate PR if you're cool with it.

ProjFS.Mac/PrjFSKext/public/PrjFSHealthData.h
wilbaker: > Do we expect to expand it to other data? If not we may want to make the name more specific.

I went with the more generic name because I thought we might expand it in the future.  I'm not aware of any plans to add more stats in the near future though so I can make this more specific.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
wilbaker: > I can do it as a small separate PR if you're cool with it.

Sure if you don't mind that would be great 👍

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -428,3 +506,30 @@ KEXT_STATIC bool TryInsertOrUpdateEntry_ExclusiveLocked(
...
     return false;
 }
+
+KEXT_STATIC_INLINE void InitCacheStats()
+{
+    atomic_exchange_explicit(&s_cacheStats.cacheEntries, 0U, memory_order_relaxed);
+    
+    for (int32_t i = 0; i < VnodeCacheHealthStat_Count; ++i)
+    {
+        atomic_exchange_explicit(&s_cacheStats.healthStats[i], 0ULL, memory_order_relaxed);
```
jeschu1: Adding @jrbriggs in case he has any thoughts on the atomic usage.  John it would be great if you can do a scan of the atomic operations in the file and let us know if you have any thoughts.

For instance this isn't a multi-threaded scenario, but there's concern that if we don't init it this way, it can cause problems later on.

@wilbaker  feel free to add any more thoughts.

wilbaker: > but there's concern that if we don't init it this way, it can cause problems later on.

From what I've found online `atomic_store` and `=` are equivalent for atomics.  If we use `atomic_store_explicit` here we can relax the memory ordering requirements a bit by specifying `memory_order_relaxed`.

In other words, I'm not sure there is a more efficient way to initialize the atomics.

ProjFS.Mac/PrjFSLib/prjfs-log/prjfs-log.cpp
jeschu1: I wouldn't bother backing it out.  Maybe do a tiny separate PR next time, but not worth the trouble here.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
jeschu1: @wilbaker i'm honestly not super familar in this space either.  Considering you've tested the build performance and didn't notice any change, what we have should be fine.  If @jrbriggs  has experience in this area it would be good to hear his thoughts.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -428,3 +506,30 @@ KEXT_STATIC bool TryInsertOrUpdateEntry_ExclusiveLocked(
     
     return false;
 }
+
+KEXT_STATIC_INLINE void InitCacheStats()
```
jrbriggs: Nit: do we need atomic access here during initialization, or can this be a memeset? Definitely does not block the PR, just curious.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
jrbriggs: It's been a while since I've been thinking about lockless programming in C++, but the patterns here make sense to me.

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -383,10 +454,12 @@ KEXT_STATIC bool TryFindVnodeIndex_Locked(
         if (vnodeIndex == vnodeHashIndex)
         {
             // Looped through the entire cache and didn't find an empty slot or the vnode
+            AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalLookupCollisions, totalSteps);
```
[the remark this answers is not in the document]
wilbaker: > In what case is this not just
> AtomicFetchAddCacheHealthStat(VnodeCacheHealthStat_TotalLookupCollisions, s_entriesCapacity);

`s_entriesCapacity` is the total capacity of the vnode cache.  `totalSteps` is the number of steps we had to take after our hash function produces a collision.  

Usually these will be different because we won't have to linearly probe the entire cache to find the entry we're looking for (or an empty slot depending on if we're looking up an entry or inserting).

> Could we drop totalSteps and the additional preincrement in the loop?

We could, but it would mean that we'd potentially introduce more atomic operations.  By using `totalSteps` we ensure that we'll only atomically adjust `VnodeCacheHealthStat_TotalLookupCollisions` once per call to `TryFindVnodeIndex_Locked`.

(I can add a comment to totalSteps explaining the motivation if that would be helpful)

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
wilbaker: > Nit: do we need atomic access here during initialization, or can this be a memeset? Definitely does not block the PR, just curious.

Things get tricky because these are not plain old datatypes.

When running the unit tests these variables are `std::atomic<T>`:

```
#ifdef KEXT_UNIT_TESTING
#include <atomic>
#define _Atomic(X) std::atomic< X >
```

And in general memsetting STL data structures is not good (and as discussed in a separate comment, we could use `=` here, but that would be no different than calling `atomic_store` ourselves).

I would not be surprised if there's a risk from memsetting [C atomic types](https://en.cppreference.com/w/c/language/atomic), but because this is the initialization path I'd rather take whatever small perf hit (if any) comes from using `atomic_store`

ProjFS.Mac/PrjFSLib/prjfs-log/prjfs-log.cpp
wilbaker: >Maybe do a tiny separate PR next time, but not worth the trouble here.

Will do, thanks!

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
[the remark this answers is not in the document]
jrbriggs: Ah, I see where I was misreading this -- I completely missed the add on ln461.I was confused why we needed to keep track of the variable when, _at this exit branch_, it should always report `s_entriesCapacity` since we probed the entire table and didn't find room. No change necessary, reviewer problem :)

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
```diff
@@ -211,9 +228,59 @@ void VnodeCache_InvalidateCache(PerfTracer* _Nonnull perfTracer)
...
+{
+    PrjFSVnodeCacheHealth healthData =
+    {
+        .cacheCapacity = s_entriesCapacity,
+        .cacheEntries = s_cacheStats.cacheEntries, // cacheEntries is reset to 0 when VnodeCache_InvalidateCache is called
+        .invalidateEntireCacheCount = atomic_exchange_explicit(
+            &s_cacheStats.healthStats[VnodeCacheHealthStat_InvalidateEntireCacheCount], 0ULL, memory_order_relaxed),
+        .totalCacheLookups = atomic_exchange_explicit(&s_cacheStats.healthStats[VnodeCacheHealthStat_TotalCacheLookups], 0ULL, memory_order_relaxed),
+        .totalLookupCollisions = atomic_exchange_explicit(&s_cacheStats.healthStats[VnodeCacheHealthStat_TotalLookupCollisions], 0ULL, memory_order_relaxed),
+        .totalFindRootForVnodeHits = atomic_exchange_explicit(&s_cacheStats.healthStats[VnodeCacheHealthStat_InvalidateEntireCacheCount], 0ULL, memory_order_relaxed),
```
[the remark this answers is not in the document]
wilbaker: > Why are we exporting VnodeCacheHealthStat_InvalidateEntireCacheCount as totalFindRootForVnodeHits?

This was definitely a bug and fix is in https://github.com/Microsoft/VFSForGit/pull/1041/commits/16bca67d6ded272e68fcf13453cbbff2dc00c875

I came across this same issue yesterday while doing some final testing because hit count was always being reported as 0 which seemed very suspicious

ProjFS.Mac/PrjFSKext/VnodeCache.cpp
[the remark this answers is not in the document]
wilbaker: Oh sorry I missed that this comment was for this case:

`// Looped through the entire cache and didn't find an empty slot or the vnode`

In that case I would expect `totalSteps` to equal `s_entriesCapacity` (I went with `totalSteps` because at least for me it was easier to follow the code when we used `totalSteps` here and on ln461

PR #1042, opened 2019-04-16
[Linux] Ignore provider-generated events

chrisd8088:
We want to ignore events triggered by the provider process itself, and we also should handle the case where we can't map a pid to a process command line because the process has already exited.

/cc @kivikakk

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -1,4 +1,6 @@
 using System;
+using System.Diagnostics;
+using System.IO;
 using System.Runtime.InteropServices;
 using PrjFSLib.Linux.Interop;
 using static PrjFSLib.Linux.Interop.Errno;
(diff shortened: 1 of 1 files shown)


ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -213,10 +216,18 @@ public virtual void StopVirtualizationInstance()
 
         private static string GetProcCmdline(int pid)
         {
-            using (var stream = System.IO.File.OpenText(string.Format("/proc/{0}/cmdline", pid)))
+            try
             {
-                string[] parts = stream.ReadToEnd().Split('\0');
-                return parts.Length > 0 ? parts[0] : string.Empty;
+                using (var stream = File.OpenText(string.Format("/proc/{0}/cmdline", pid)))
```
kivikakk:
```suggestion
                using (var stream = File.OpenText($"/proc/{pid}/cmdline"))
```

I need to start getting in the habit of using string interpolation in C# (wasn't a thing when I was using it last). Mind committing this for me while you're here?

chrisd8088:
Updated to pick up latest CI checks.

kivikakk:
Looks good for merge.

chrisd8088:
@jrbriggs or @wilbaker -- would you have some time to take a look at this PR for the Linux branch?

chrisd8088:
Thanks, @wilbaker!

PR #1046, opened 2019-04-17
Add category for extra tests to run in PRs

kewillford:
This adds another category to the tests to allow the extra tests that are currently only ran during to the full suite to be ran so that we can add a PR build for just them without having to run the full suite since the other tests are already ran during a PR build.


GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -85,6 +85,12 @@ public static void Main(string[] args)
                 excludeCategories.Add(Categories.MacOnly);
             }
 
+            if (runner.HasCustomArg("--extra-only"))
+            {
+                includeCategories.Add(Categories.ExtraCoverage);
+                excludeCategories.Remove(Categories.FullSuiteOnly);
```
wilbaker: Do we need a new `ExtraCoverage` category?  

I think if you updated this code to:

```
if (runner.HasCustomArg("--extra-only"))
{                
    includeCategories.Add(Categories. FullSuiteOnly);
```

It would only run the tests marked as `FullSuiteOnly` (this is how the `--windows-only` flag works).

Unfortunately the behavior of include/exclude is a bit weird.  I believe as soon as you set an include category only tests from the included categories will be run.

GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -85,6 +85,12 @@ public static void Main(string[] args)
                 excludeCategories.Add(Categories.MacOnly);
             }
 
+            if (runner.HasCustomArg("--extra-only"))
```
wilbaker: Should this check be in an `else if` after this check?

```
if (runner.HasCustomArg("--full-suite"))
```

Also, I've started adding a `Console.WriteLine` to describe what will be run (e.g. `Console.WriteLine("Running the full suite of tests");` above).

Could you do the same here?  I've found it to be a useful sanity check when looking at functional test run output.

GVFS/GVFS.FunctionalTests/Program.cs
kewillford: Yeah but `FullSuiteOnly` is also in the `excludeCategories` so we would need to keep the `Remove` here otherwise I'm not sure how it would handle the same category in both.

wilbaker: > Yeah but FullSuiteOnly is also in the excludeCategories so we would need to keep the Remove here otherwise I'm not sure how it would handle the same category in both.

If you put this case here could you avoid it being in the excludeCategories?

```
            if (runner.HasCustomArg("--full-suite") || runner.HasCustomArg("--extra-only"))
            {
                if (runner.HasCustomArg("--extra-only"))
                {
                    includeCategories.Add(Categories. FullSuiteOnly);
                }
                else
                {
                    Console.WriteLine("Running the full suite of tests");
                }

                GVFSTestConfig.GitRepoTestsValidateWorkTree = DataSources.AllBools;

                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                {
                    GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.AllWindowsRunners;
                }
                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
                {
                    GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.AllMacRunners;
                }
            }
            else
            {
                GVFSTestConfig.GitRepoTestsValidateWorkTree =
                    new object[]
                    {
                        new object[] { true }
                    };

                excludeCategories.Add(Categories.FullSuiteOnly);
                GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.DefaultRunners;
            }
```

(Or alternatively give it its own `else if` block?)

wilbaker: Also, FWIW now that these tests are run on their own I like the `ExtraCoverage` name more than `FullSuiteOnly`, so maybe the category to keep is the new one you've added?

kewillford: Do we want the `GVFSTestConfig.GitRepoTestsValidateWorkTree`  to be all bools and with all the runners?  I was thinking to add it in the else like so:
```
            else
            {
                GVFSTestConfig.GitRepoTestsValidateWorkTree =
                    new object[]
                    {
                        new object[] { true }
                    };
               
                if (runner.HasCustomArg("--extra-only"))
                {
                    includeCategories.Add(Categories. FullSuiteOnly);
                }
                else
                {
                    excludeCategories.Add(Categories.FullSuiteOnly);
                }

                GVFSTestConfig.FileSystemRunners = FileSystemRunners.FileSystemRunner.DefaultRunners;
            }
```

I think we just need to run the tests so I'm not sure how important it is to make sure we have both for validate the work tree and all the runners.  What do you think?

wilbaker: > I think we just need to run the tests so I'm not sure how important it is to make sure we have both for validate the work tree and all the runners. What do you think?

Ahh good point.  For regular PRs I don't think we need all the runners and work tree validation scenarios.

I think we should still run those as part of the release process though (cc: @jrbriggs)

kewillford: There is still the CI build that will run the full functional tests so we still run everything with every merge into master.

wilbaker:
@kewillford latest changes look good to me

kewillford:
@jamill I don't remember any official discussion on the tests that were added to this group so I would have to go through them and see.  For now I like them as simply extra coverage tests and we can sort out why and if they need a different name some other time.

wilbaker:
> This makes sense. My only question is around naming - are these tests in the "full test suite only" / "extra-coverage" because they are long running? If so, maybe name this level of tests "long-running". Or are these in this category because they are lower priority? or...

@jamill @kewillford here is the description/motivation from the PR in Azure Repos (302803):

>This PR does the following:
>
> - Add the ability to mark a test with Category([Categories.FullSuiteOnly]) to indicate that it should not run in normal PR tests, only in the full suite of tests
> - Marked a few test fixtures with that category. Tried to mark the tests that are slow, and also unlikely to catch errors in core scenarios, so that it's relatively safe to skip running them in all PRs. Folks who want full coverage should still run the optional full test suite in their PRs.
> - Cleaned up the test outputs so that start/end messages are not interleaved with other tests.
> - After the tests complete, print out stats on how long each fixture and each test case took to run, so that we can evaluate which ones are worth investigating for performance
Along with these changes, I also plan to make the full suite of tests run after every merge into master.
>
> Another possibility to explore (as future work) is to divide the tests into various categories of roughly equal runtime, and then create a build definition per category and make them all required. The advantage of that approach is that if the build machines are all free, you can complete an individual PR much faster by spreading the tests over multiple machines.

PR #1055, opened 2019-04-19
Fix race condition in MockVirtualizationInstance.CompleteCommand

### LINKED ISSUE #1054: Windows: GetFileStream unit tests are occasionally failing
Two examples:

```
1) Failed : GVFS.UnitTests.Windows.Virtualization.WindowsFileSystemVirtualizerTests.OnGetFileStreamHandlesHResultHandleResult()
 Expected: Handle
 But was: Ok
 at GVFS.Tests.Should.ValueShouldExtensions.ShouldEqual[T](T actualValue, T expectedValue, String message)
 at GVFS.UnitTests.Windows.Virtualization.WindowsFileSystemVirtualizerTests.OnGetFileStreamHandlesHResultHandleResult() in D:\a\1\s\GVFS\GVFS.UnitTests.Windows\Windows\Virtualization\WindowsFileSystemVirtualizerTests.cs:line 938
```

```
1) Failed : GVFS.UnitTests.Windows.Virtualization.WindowsFileSystemVirtualizerTests.OnGetFileStreamHandlesWriteFailure()
  Expected: InternalError
  But was:  Ok
   at GVFS.Tests.Should.ValueShouldExtensions.ShouldEqual[T](T actualValue, T expectedValue, String message)
   at GVFS.UnitTests.Windows.Virtualization.WindowsFileSystemVirtualizerTests.OnGetFileStreamHandlesWriteFailure() in D:\a\1\s\GVFS\GVFS.UnitTests.Windows\Windows\Virtualization\WindowsFileSystemVirtualizerTests.cs:line 883
```

I have had luck reproducing this locally by running the unit tests 1000 times.  It appears that there's some race between the main thread and the background threads in these tests.

wilbaker:
Fixes #1054 

Two fixes:

- Updated `completionResult` and `WriteFileReturnResult` to use `volatile` to ensure that the test and background thread read the most up-to-date values (and removed public `CompletionResult` as it was not needed)
- Fixed a race in `CompleteCommand` where `commandCompleted` was being set too early.

### DIFF
--- a/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
@@ -2,7 +2,6 @@
 using GVFS.UnitTests.Windows.Windows.Mock;
 using Microsoft.Windows.ProjFS;
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 
@@ -28,12 +27,6 @@ namespace GVFS.UnitTests.Windows.Mock
             this.waitForCreateWriteBuffer = new ManualResetEvent(true);
 
             this.WriteFileReturnResult = HResult.Ok;
-        }
-
-        public HResult CompletionResult
-        {
-            get { return this.completionResult; }
-            set { this.completionResult = value; }
         }
 
         public ConcurrentHashSet<string> CreatedPlaceholders { get; private set; }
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.UnitTests.Windows/Windows/Mock/MockVirtualizationInstance.cs
```diff
@@ -27,7 +30,11 @@ public MockVirtualizationInstance()
             this.WriteFileReturnResult = HResult.Ok;
         }
 
-        public HResult CompletionResult { get; set; }
+        public HResult CompletionResult
+        {
+            get { return this.completionResult; }
+            set { this.completionResult = value; }
```
kewillford: It looked like this can be set as `private`.  Not sure how many other properties could be as well but this is a mock for testing so probably doesn't matter too much.

wilbaker: Agreed, and since it can be private I'll just remove the property and we can use `this.completionResult` directly in the mock

Issue #94, opened 2018-06-11
Update .NET Framework projects to use PackageReference rather than packages.config


sanoursa:
Use PackageReferences in all projects, to avoid mixing the two patterns.

wilbaker:
@jrbriggs FYI the latest ProjFS drops adjusted the layout of their Nuget packages and so I don't there's anything blocking us from picking up this work.

wilbaker:
As part of this work item we should also see if we can combine csproj files that are currently split (e.g. `GVFS.Mac` and `GVFS.Windows`).

/cc: @jrbriggs, @nickgra, @jamill

PR #1059, opened 2019-04-22
GVFS.Upgrader: migrate to netcore project 

jamill:
GVFS.Upgrader will be run on both Windows and non-Windows platforms.
This enables a netcore version of the project to be built along side a
netframework version.

Update tests to run in the base unit test project to follow the
implementation.


GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -1,137 +1,56 @@
...
+    <Platforms>x64</Platforms>
     <PlatformTarget>x64</PlatformTarget>
-    <ErrorReport>prompt</ErrorReport>
-    <Prefer32Bit>false</Prefer32Bit>
+    <RootNamespace>GVFS.Upgrader</RootNamespace>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
+    <RuntimeIdentifiers>osx-x64</RuntimeIdentifiers>
```
jamill: This `RuntimeIdentifier` probably does not belong in this common PropertyGroup that is included for both `net461` and `netcoreapp2.1`.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -1,137 +1,56 @@
...
+
+    <Platforms>x64</Platforms>
     <PlatformTarget>x64</PlatformTarget>
-    <ErrorReport>prompt</ErrorReport>
-    <Prefer32Bit>false</Prefer32Bit>
+    <RootNamespace>GVFS.Upgrader</RootNamespace>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
```
jamill: TODO: what behavior does the `TargetLatestRuntimePatch` control.

GVFS/GVFS.GVFlt/GVFS.GVFlt.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
-  <Import Project="$(SolutionDir)\GVFS\GVFS.Build\GVFS.cs.props" />
+  <Import Project="..\GVFS.Build\GVFS.cs.props" />
```
alameenshah: wondering why this path is changed.

GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
```diff
@@ -113,21 +113,14 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="Windows\Mock\MockInstallerPreRunChecker.cs" />
```
alameenshah: These are now moved to GVFS.UnitTests project?

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -1,137 +1,63 @@
...
-    <Analyzer Include="..\..\..\packages\StyleCop.Analyzers.1.0.2\analyzers\dotnet\cs\StyleCop.Analyzers.CodeFixes.dll" />
-    <Analyzer Include="..\..\..\packages\StyleCop.Analyzers.1.0.2\analyzers\dotnet\cs\StyleCop.Analyzers.dll" />
+    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
+    <ProjectReference Include="..\GVFS.Virtualization\GVFS.Virtualization.csproj" /> 
   </ItemGroup>
+
+  <!-- ItemGroup Conditions are not supported on VS for Mac and Choose/When must be
+       Used instead, see https://github.com/mono/monodevelop/issues/7417 -->
+  <Choose>
+    <When Condition="'$(OS)' == 'Windows_NT'">
```
alameenshah: The existing pattern to have separate `csproj` for Mac and Windows (like in `GVFS.Mount` project) helps avoid the need to do platform checks like this. But your approach of keeping a single project file for both platforms looks cleaner. Also now we end up having 2 different patterns (existing pattern of separate platform specific `csproj` files and this new common `csproj` with platform specific source code inclusions.)
 
 It would be good if we could discuss and decide on a single pattern that we could follow consistently.  @nickgra, @wilbaker, @kewillford, @jrbriggs - thoughts?

alameenshah:
The changes look good. A few things to consider before merging.
1. Weigh the pros & cons of single `csproj` that includes source files based on platform checks vs existing pattern of keeping separate `csproj`s.
2. Should we need to keep different sets of dependent assemblies (in `ProductUpgrader.UpgraderToolAndLibs`) for `netcore` and `framework` targets.
3. Is `ProductUpgrader.TrySetupToolsDirectory()` required at all on the Mac? The original problem was that upgrader could not over-write installed assemblies that are open on `Windows`. This might not be a problem on `Mac`.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
jamill: See #1132 for details

GVFS/GVFS.GVFlt/GVFS.GVFlt.csproj
mjcheetham: Most likely because Visual Studio like to do things like this 😀 😢

jamill: This is because SolutionDir is not defined when building the project, only when building the solution.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
mjcheetham: You can set the identifier conditionally on the TF.

```xml
<TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
<RuntimeIdentifiers>win-x64;osx-x64;linux-x64</RuntimeIdentifiers>
<RuntimeIdentifier Condition="'$(TargetFramework)'=='net461'">win-x64</RuntimeIdentifier>
```

Here when building the TF as `net461` you're also forcing the RID to be `win-x64`.
For TF `netcoreapp2.1` you'll need to specify which OS you're targeting still.

GVFS/GVFS.GVFlt/GVFS.GVFlt.csproj
mjcheetham: One thing I've done in the past is to define an MSBuild property `<RepoRoot>` in a Directory.Build.props file in the root of the repository.

Projects can use `$(RepoRoot)` if they want to use an absolute path (rather than a maybe more brittle relative one), without needing the Solution build to set `$(SolutionDir)`.

GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
jamill: yep - as these are no longer Windows specific.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
jamill: If we can have a single file for both targets, I think that is preferable in the long run. If we can make it work now, I view this as a step in the right direction. At this point, I don't see any blockers in having everything as a single file (debugging on VS for Mac works, I will look again and make sure everything works on VS for Windows).

If there is a problem with using a single file, then my plan was to fall back to having separate projects.

wilbaker: > It would be good if we could discuss and decide on a single pattern that we could follow consistently. @nickgra, @wilbaker, @kewillford, @jrbriggs - thoughts?

When @nickgra updated FastFetch to be cross-platform he was able to use with a single csproj (although I believe in that case it's .NET Core on both platforms).  

I believe that historically we couldn't do this because the ProjFS nuget package was incompatible with `PackageReference`.  Now that the layout has been fixed we should be able to start switching to using single projects (see also #94).

cc: @nickgra for his thoughts

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -1,137 +1,63 @@
...
+    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
+
+    <Platforms>x64</Platforms>
     <PlatformTarget>x64</PlatformTarget>
-    <ErrorReport>prompt</ErrorReport>
-    <Prefer32Bit>false</Prefer32Bit>
+    <RootNamespace>GVFS.Upgrader</RootNamespace>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
wilbaker: As best I can tell `AllowUnsafeBlocks` was not set before, can this line be removed?

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -1,137 +1,63 @@
...
+        <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Windows.cs">
+          <Link>PlatformLoader.Windows.cs</Link>
+        </Compile>
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <ProjectReference Include="..\GVFS.Platform.Mac\GVFS.Platform.Mac.csproj" />
+        <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Mac.cs">
+          <Link>PlatformLoader.Mac.cs</Link>
```
wilbaker: I guess this is the one downside to having a single csproj, it means that each platform can only build their own version (at least until we can find an alternative to `Condition="'$(OS)'`).

I'm thinking that might be preferable to doubling the number of projects though.

@jamill @nickgra what are your thoughts?

wilbaker:
@nickgra can you take a look at this PR when you have a few minutes?  There aren't too many changes and you have a lot of familiarity in this area. Thanks!

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
wilbaker: > I guess this is the one downside to having a single csproj, it means that each platform can only build their own version (at least until we can find an alternative to Condition="'$(OS)').

One other thought here, the long term goal is to get everything on .NET Core, and I think switching to single projects is a step in that direction.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
jamill: I had thought it was needed when building the project, but apparently not. I will remove it.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
jamill: > I guess this is the one downside to having a single csproj, it means that each platform can only build their own version (at least until we can find an alternative to Condition="'$(OS)').

Another option I was thinking about (not for this PR) might be to condition on the RunTime identifier (or a combination of target framework and runtime identifier).

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
nickgra: What @mjcheetham proposed seems like it should work, but I bet VSMac won't be able to parse it. 

It sounds like most of us are moving towards using VSCode going forward, so maybe it's alright to take this proposed change if VSMac can't handle this MSBuild structure.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
nickgra: I think long term we want to move everything to .NET Core and moving to a single project is a good step in that direction. IIRC, `PackageReference` was what was stopping us from doing this. 

If this works fine on VSMac, I'm happy with this solution for now, although it does have the con of making us unable to build Mac on Windows as William called out in another comment.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
wilbaker: > It sounds like most of us are moving towards using VSCode going forward, so maybe it's alright to take this proposed change if VSMac can't handle this MSBuild structure.

Please don't break VSMac until we've decided that it's the preferred editor for Mac development and have documented how to configure VSCode to build/debug/etc. VFS4G.

jamill: > Please don't break VSMac

This was my goal. If I couldn't get this to work with a single csproj file, I was going to fall back to having 2 separate csproj files.

jamill: > Please don't break VSMac

This was my goal. If I couldn't get this to work with a single csproj file, I was going to fall back to having 2 separate csproj files.

PR #1060, opened 2019-04-23
Fix FastFetch build in VS for Mac

wilbaker:
The `ItemGroup` conditions are not working properly in VS for Mac (I noticed this [reported on StackOverflow](https://stackoverflow.com/questions/28851338/why-the-condition-attribute-doesnt-work-for-the-itemgroup-element) for VS 7.5.3 and so we don't appear to be the only ones having this issue).

To work around this limitation I've switched the project to use. `Choose`/`Where`

### DIFF
--- a/GVFS/FastFetch/FastFetch.csproj
+++ b/GVFS/FastFetch/FastFetch.csproj
@@ -19,30 +19,30 @@
     <Version>$(GVFSVersion)</Version>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(MSBuildRuntimeType)' == 'Core'">
-    <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'true'">true</IsOSX>
-    <IsLinux Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">true</IsLinux>
-  </PropertyGroup>
-  
   <ItemGroup>
     <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
     <ProjectReference Include="..\GVFS.Virtualization\GVFS.Virtualization.csproj" />
   </ItemGroup>
   
-  <ItemGroup Condition="'$(IsOSX)' == 'true'">
-    <ProjectReference Include="..\GVFS.Platform.Mac\GVFS.Platform.Mac.csproj" />
-    <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Mac.cs">
-      <Link>PlatformLoader.Mac.cs</Link>
-    </Compile>
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(OS)' == 'Windows_NT'">
-    <ProjectReference Include="..\GVFS.Platform.Windows\GVFS.Platform.Windows.csproj" />
-    <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Windows.cs">
-      <Link>PlatformLoader.Windows.cs</Link>
-    </Compile>
-  </ItemGroup>
-
+  <Choose>
+    <When Condition="'$(OS)' == 'Windows_NT'">
+      <ItemGroup>
+        <ProjectReference Include="..\GVFS.Platform.Windows\GVFS.Platform.Windows.csproj" />
+        <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Windows.cs">
+          <Link>PlatformLoader.Windows.cs</Link>
+        </Compile>
+      </ItemGroup>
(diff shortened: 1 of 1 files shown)


nickgra:
In the Stack Overflow post, I don't see a link to an issue in VSMac for this. Should we open an issue on their side (in monodevelop?) for this to track removing this workaround?

wilbaker:
>Should we open an issue on their side (in monodevelop?) for this to track removing this workaround?

That's a good idea, do you know where I can submit an issue to them?

nickgra:
> > Should we open an issue on their side (in monodevelop?) for this to track removing this workaround?
> 
> That's a good idea, do you know where I can submit an issue to them?

Historically, we've filed issues directly in their Github issues (see: https://github.com/mono/monodevelop/issues/4837). This also seems like a 'Project Model' issue to me.

GVFS/FastFetch/FastFetch.csproj
```diff
@@ -19,30 +19,30 @@
...
+      <ItemGroup>
+        <ProjectReference Include="..\GVFS.Platform.Windows\GVFS.Platform.Windows.csproj" />
+        <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Windows.cs">
+          <Link>PlatformLoader.Windows.cs</Link>
+        </Compile>
+      </ItemGroup>
+    </When>
+    <Otherwise>
+      <ItemGroup>
+        <ProjectReference Include="..\GVFS.Platform.Mac\GVFS.Platform.Mac.csproj" />
```
nickgra: Whenever Linux comes online, is there a way we can rework this code to keep VSMac working but load the Linux platform instead of the Mac one?

wilbaker: >Whenever Linux comes online, is there a way we can rework this code to keep VSMac working but load the Linux platform instead of the Mac one?

I found that the conditionals did work when using the build script, and so I suspect we could something like:

```
<When Condition="'$(OS)' == 'Windows_NT'">
    // Windows stuff
</When>
<When Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">
    // Linux stuff
</When>
<Otherwise>
    // Mac stuff
</Otherwise>
```

I was also discussing this with @jamill a bit and it would be nice if the condition was driven by the active configuration (e.g. Debug.Windows|x64) rather than the current platform that we're building on.  So far I have not figured out how to do that, do you know if it's possible?

nickgra: Frankly, your proposed approach does not look too bad to me so long as we leave a comment on why it's like that.

> So far I have not figured out how to do that, do you know if it's possible?

I have not seen any documentation that indicates picking up the active configuration is possible, although it would be very useful :) Even if it's supported on Windows, I wouldn't trust the VSMac .csproj parser to understand it.

jamill: I did a bit of searching on this, but could not find a great solution for controlling this via the solution configuration.

One thing I was considering:

1) Could we define a property in `GVFS.props` that specifies which target OS we are building for, and put all the logic for detecting this in there? That way, all other projects could key off the value of this property. (e.g. `TargetOSPlatform` or something?

2) While not accessible from the IDE, we could build for a specific platform from MSBuild (or dotnet build) via `/p:TargetOSPlatform=macOS` or `/p:TargetOSPlatform=Windows`..., or possibly via an environment variable?

wilbaker:
> Historically, we've filed issues directly in their Github issues (see: mono/monodevelop#4837). This also seems like a 'Project Model' issue to me.

Thanks! I've filed https://github.com/mono/monodevelop/issues/7417 for this.

GVFS/FastFetch/FastFetch.csproj
wilbaker: > so long as we leave a comment on why it's like that.

Comments added!

> Could we define a property in GVFS.props that specifies which target OS we are building for, and put all the logic for detecting this in there?

I think that sounds good.  If/when we need to add checks like this in a second project we should see if we can get this logic in a common spot.

> we could build for a specific platform from MSBuild (or dotnet build) via /p:TargetOSPlatform=macOS or /p:TargetOSPlatform=Windows..., or possibly via an environment variable?

It looks like we're pretty close to this today:

_Windows_
```
SET SolutionConfiguration=%Configuration%.Windows

%msbuild% %VFS_SRCDIR%\GVFS.sln /p:GVFSVersion=%GVFSVersion% /p:Configuration=%SolutionConfiguration% /p:Platform=x64 || exit /b 1
```

_Mac_
```
dotnet build $VFS_SRCDIR/GVFS.sln --runtime osx-x64 --framework netcoreapp2.1 --configuration $CONFIGURATION.Mac -p:CopyPrjFS=true /maxcpucount:1 /warnasmessage:MSB4011 || exit 1
```

Fortunately the command line builds haven't had any issues, it's been with the VS for Mac IDE specifically.

PR #1061, opened 2019-04-23
Update coding guidelines to promote more testable code

jamill:


CONTRIBUTING.md
```diff
@@ -224,9 +224,19 @@ The design review process is as follows:
...
 
-  When a class needs to be mocked (or have a subset of its behavior mocked), prefer using virtual methods instead of adding a new interface.  VFS4G uses interfaces when multiple implementations of the interface exist in the product code.
+  When writing new code, aim to follow existing coding guidelines. However, when exiting guidelines and testability goals of a design conflict, optimize for testability.
+
+- *When writing tests, prefer a well defined boundary for what dependencies are test classes and what dependencies are product classes.*
+
+  When a test needs to control a dependency, try to implement test class from a contract instead of extending a product implementation class. This will ensure that we understand what behaviors are test classes and what are product classes.
+
+  When a class might be substituted for a test class in other componenents, consider defining the public API of the class explicity, either deriving from an abstract base class or as an interface as appropriate. The ability to substite a test class that implements an API is more important than following the style of existing code.
```
mjcheetham: 📎 says:
> When a class might be substituted for a test class in other ~componenents~components, consider defining the public API of the class ~explicity~explicitly, either deriving from an abstract base class or as an interface as appropriate. The ability to ~substite~substitute a test class that implements an API is more important than following the style of existing code.

CONTRIBUTING.md
```diff
@@ -224,9 +224,19 @@ The design review process is as follows:
   if (GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories)
   ```
 
-- *Add new interfaces when it makes sense for the product, not simply for unit testing*
+- *Testability of a class should be a priority in the design of a class and during code reviews.*
```
mjcheetham: 📎 says:
Consider revising this sentence to avoid specifying the subject twice:
> - *Testability ~of a class~ should be a priority in the design of a class and during code reviews.*

CONTRIBUTING.md
```diff
@@ -224,9 +224,19 @@ The design review process is as follows:
...
+  When writing new code, aim to follow existing coding guidelines. However, when exiting guidelines and testability goals of a design conflict, optimize for testability.
+
+- *When writing tests, prefer a well defined boundary for what dependencies are test classes and what dependencies are product classes.*
+
+  When a test needs to control a dependency, try to implement test class from a contract instead of extending a product implementation class. This will ensure that we understand what behaviors are test classes and what are product classes.
+
+  When a class might be substituted for a test class in other componenents, consider defining the public API of the class explicity, either deriving from an abstract base class or as an interface as appropriate. The ability to substite a test class that implements an API is more important than following the style of existing code.
+
+  The Moq framework used by the project works best when paired with explicit contracts.
```
kewillford: I think this should be moved to the `Testing` section or also added to the `Testing` section as information about the framework that we are using for mocking and adding some information on using `Setup`, `MockBehavior.Strict`, `VerifyAll`, and any other information on mocking that might be helpful.

CONTRIBUTING.md
```diff
@@ -224,9 +224,19 @@ The design review process is as follows:
...
+
+- *If code need to be refactored to enable new functionality to be testable, make the refactoring changes in a self contained set of commits.*
```
kewillford:
```suggestion
- *If code needs to be refactored to enable the testability of new functionality, make the refactoring changes in a self contained set of commits.*
```

CONTRIBUTING.md
kewillford: Some examples would be nice with bad and good code.  Like instead of … do this ...

CONTRIBUTING.md
```diff
@@ -224,9 +224,19 @@ The design review process is as follows:
   if (GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories)
   ```
 
-- *Add new interfaces when it makes sense for the product, not simply for unit testing*
+- *Testability of a class should be a priority in the design of a class and during code reviews.*
 
-  When a class needs to be mocked (or have a subset of its behavior mocked), prefer using virtual methods instead of adding a new interface.  VFS4G uses interfaces when multiple implementations of the interface exist in the product code.
+  When writing new code, aim to follow existing coding guidelines. However, when exiting guidelines and testability goals of a design conflict, optimize for testability.
```
wilbaker: > when exiting guidelines and testability goals of a design conflict

- `exiting` -> `existing`
- Does "existing coding guidelines" here refer to other guidelines that might conflict with testability?
- Would it be fair to reword this to something like "When guidelines conflict, optimize for testability goals.  The only exception is that testability changes should not negatively impact performance critical paths."

CONTRIBUTING.md
```diff
@@ -224,9 +224,19 @@ The design review process is as follows:
...
+
+- *When writing tests, prefer a well defined boundary for what dependencies are test classes and what dependencies are product classes.*
+
+  When a test needs to control a dependency, try to implement test class from a contract instead of extending a product implementation class. This will ensure that we understand what behaviors are test classes and what are product classes.
+
```
wilbaker: Is there anything we can add to encourage developers to add both "true unit tests" and "integration" unit tests?

I think it makes sense to always recommend that the true unit tests be added, and with possible/sensible also add unit tests that exercise method calls across the components of the product.

PR #1065, opened 2019-04-24
Add Full Coverage for HandleVnodeOperation

jeschu1:
Add Full Coverage for HandleVnodeOperation

### DIFF
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -14,6 +14,7 @@
 #include "KextMockUtilities.hpp"
 #include "MockVnodeAndMount.hpp"
 #include "MockProc.hpp"
+#include "ProviderMessagingMock.hpp"
 #include "VnodeCacheEntriesWrapper.hpp"
 #include <tuple>
 
@@ -78,6 +79,11 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     MockProcess_AddContext(context, 501 /*pid*/);
     MockProcess_SetSelfPid(501);
     MockProcess_AddProcess(501 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "test" /*name*/);
+    
+    ProvidermessageMock_ResetResultCount();
+    ProviderMessageMock_SetDefaultRequestResult(true);
+    ProviderMessageMock_SetSecondRequestResult(true);
+    ProviderMessageMock_SetCleanupRootsAfterRequest(false);
 }
 
 - (void) tearDown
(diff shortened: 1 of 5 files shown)


ProjFS.Mac/PrjFSKextTests/ProviderMessagingMock.cpp
```diff
@@ -1,12 +1,18 @@
...
 #include "KextMockUtilities.hpp"
 
 #include <sys/proc.h>
 #include <libkern/OSAtomic.h>
 #include <sys/kauth.h>
+#include "ProviderMessagingMock.hpp"
 
+bool static s_requestResult = true;
+bool static s_secondRequestResult = false;
+bool static s_cleanupRoots = false;
```
wilbaker: Minor: Something like `s_cleanupRootsAfterRequest` might be a more descriptive name.

ProjFS.Mac/PrjFSKextTests/ProviderMessagingMock.cpp
```diff
@@ -18,6 +24,22 @@ void ProviderMessaging_Cleanup()
...
+
+void ProviderMessageMock_SetCleanupRootsAfterRequest(bool cleanupRoots)
+{
+    s_cleanupRoots = cleanupRoots;
+}
+
+void ProviderMessageMock_SetSecondRequestResult(bool secondRequestResult)
+{
+    s_secondRequestResult = secondRequestResult;
+    s_requestCount = 0;
```
wilbaker: Why does setting `s_secondRequestResult` reset `s_secondRequestResult`?

ProjFS.Mac/PrjFSKextTests/ProviderMessagingMock.cpp
```diff
@@ -49,7 +71,15 @@ bool ProviderMessaging_TrySendRequestAndWaitForResponse(
...
-    return true;
+    
+    if (s_cleanupRoots)
+    {
+        VirtualizationRoots_Cleanup();
+    }
+    
+    s_requestCount++;
+    
+    return s_requestResult && !(s_secondRequestResult == true && s_requestCount == 2);
```
wilbaker: Why does `s_requestResult` impact the what is returned when `s_requestCount` is 2?

Based on how it's used, would it be accurate to rename `s_requestResult` (and its setter) to include "default" in the name, e.g. `s_defaultRequestResult`, and then update this code to:

```
if (s_requestCount == 2)
{
    return s_secondRequestResult;
}
else
{
    return s_defaultRequestResult;
}

File: ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm [hunk 4]
```diff
@@ -706,5 +707,182 @@ - (void) testCloseWithoutModifed {
...
+        nullptr,
+        nullptr,
+        KAUTH_VNODE_DELETE,
+        reinterpret_cast<uintptr_t>(context),
+        reinterpret_cast<uintptr_t>(testFileVnode.get()),
+        0,
+        0);
+
+   XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 1);
+   ProviderMessageMock_SetRequestResult(true);
```
wilbaker: Could we update `(void) setUp` to always call:
- `ProviderMessageMock_SetRequestResult(true)`
- `ProviderMessageMock_SetSecondRequestResult(true)`
- `ProviderMessageMock_SetCleanupRootsAfterRequest(false)`

That way each test case won't have to worry about resetting their state.

ProjFS.Mac/PrjFSKextTests/ProviderMessagingMock.cpp
```diff
@@ -1,12 +1,18 @@
...
 #include "KextLogMock.h"
 #include "KextMockUtilities.hpp"
 
 #include <sys/proc.h>
 #include <libkern/OSAtomic.h>
 #include <sys/kauth.h>
+#include "ProviderMessagingMock.hpp"
 
+bool static s_requestResult = true;
+bool static s_secondRequestResult = false;
```
wilbaker: Why is the second result defaulted to `false`, is `false` the error scenario?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -706,5 +707,182 @@ - (void) testCloseWithoutModifed {
...
+        reinterpret_cast<uintptr_t>(testDirVnode.get()),
+        0,
+        0);
+
+    XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 1);
+    ProviderMessageMock_SetCleanupRootsAfterRequest(false);
+}
+
+
+- (void) testDeleteDirectoryWithSecondRequestResultFalse {
```
wilbaker: Naming suggestion: `testDeleteDirectoryWithSecondRequestResultFalse` -> `testDeleteDirectoryWhenSecondSendRequestFails`

(Same comment elsewhere, consider switching to "SecondSendRequestFails" instead of "SecondRequestResultFalse")

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -706,5 +707,182 @@ - (void) testCloseWithoutModifed {
...
+        reinterpret_cast<uintptr_t>(context),
+        reinterpret_cast<uintptr_t>(testFileVnode.get()),
+        0,
+        0);
+
+    XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 2);
+    ProviderMessageMock_SetSecondRequestResult(false);
+}
+
+- (void) testForkedPath {
```
wilbaker: Naming suggestion: `testForkedPath` -> `testWriteDataToForkedPath`

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -706,5 +707,182 @@ - (void) testCloseWithoutModifed {
...
+    testFileVnode->namedStream = true;
+    testFileVnode->GetParentVnode()->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
+    SetPrjFSFileXattrData(testFileVnode);
```
wilbaker: Could you add a comment here explaining why `HandleVnodeOperation` will think this is a named fork?

ProjFS.Mac/PrjFSKextTests/ProviderMessagingMock.cpp
jeschu1: Updating the logic to be more intuitive

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -706,5 +712,176 @@ - (void) testCloseWithoutModifed {
...
+        reinterpret_cast<uintptr_t>(context),
+        reinterpret_cast<uintptr_t>(testFileVnode.get()),
+        0,
+        0);
+
+   XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 1);
+}
+
+- (void) testDeleteDirectoryWithNoVirtualizationRoot {
+    ProviderMessageMock_SetCleanupRootsAfterRequest(true);
```
wilbaker: Should we be calling `VirtualizationRoots_Cleanup();` here instead of ` ProviderMessageMock_SetCleanupRootsAfterRequest(true)`?

I noticed that's what we're doing in the `testDeleteWithNoVirtualizationRoot` and `testReadAttributesDirectoryWithNoVirtualizationRoot` tests.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -706,5 +712,176 @@ - (void) testCloseWithoutModifed {
...
+    HandleVnodeOperation(
+        nullptr,
+        nullptr,
+        KAUTH_VNODE_READ_ATTRIBUTES,
+        reinterpret_cast<uintptr_t>(context),
+        reinterpret_cast<uintptr_t>(testDirVnode.get()),
+        0,
+        0);
+
+    XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 1);
```
wilbaker: Is there anything else we can/should validate happens when a request fails? (e.g. should the kext be logging an error?)

This comment applies to all of the tests for the failure scenario.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -706,5 +712,176 @@ - (void) testCloseWithoutModifed {
...
+        reinterpret_cast<uintptr_t>(context),
+        reinterpret_cast<uintptr_t>(testFileVnode.get()),
+        0,
+        0);
+
+    XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 1);
+}
+
+- (void) testWriteWithNoVirtualizationRoot {
+    ProviderMessageMock_SetCleanupRootsAfterRequest(true);
```
wilbaker: Why are we calling `ProviderMessageMock_SetCleanupRootsAfterRequest ` instead of `VirtualizationRoots_Cleanup`?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: Under the covers two calls to TryGetVirtualizationRoot are made in the code path.

To get full coverage we want the first call to success and the second to fail.  This is why we can't just call 
`VirtualizationRoots_Cleanup();`.  That would make the first call fail never testing the second.

I did think about weather we needed to do this check twice but:
1. Since we're multi-threaded I *think* the state of a VirtualizationRoot can change while processing
2. Perhaps refactoring is a good candidate for a separate PR once we have all the tests in place.

Let me know what you think!

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: I was thinking that too.  It can validly fail because it's shutting down (probably don't want to log that). Another case would be if the userClient became null.  In `ActiverProvider_SendMessage` we kernal panic if the virtualization root is invalid.  

Perhaps it wouldn't hurt to add a call saying the send failed, just curious if we thought it would be extra noise.  It does seem like this would drop a message.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: Please see above comment regarding.  Thank you!

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: >Under the covers two calls to TryGetVirtualizationRoot are made in the code path.
>To get full coverage we want the first call to success and the second to fail.

Ahh okay, in that case could you rename the test?  "WithNoVirtualizationRoot" made me think there'd be no root from the start, perhaps something like `testDeleteDirectoryWithDisappearingRoot`?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: See discussion in other comment, rename this method as well?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
wilbaker: > Is there anything else we can/should validate happens when a request fails? 

> I was thinking that too

In my earlier pass over the changes I missed that we're not validating that `HandleVnodeOperation` is returning the result we expect it to:


_From HandleVnodeOperation_
```
CleanupAndReturn:
    if (putVnodeWhenDone)
    {
        vnode_put(currentVnode);
    }
    
    atomic_fetch_sub(&s_numActiveKauthEvents, 1);
    return kauthResult;
```

Can you add that validation to the `HandleVnodeOperation` tests?  That will allow us to verify that `HandleVnodeOperation` is properly passing through the result returned by `ProviderMessaging_TrySendRequestAndWaitForResponse`.

(It looks like the `ProviderMessaging_TrySendRequestAndWaitForResponse` mock will need to be updated to set `kauthResult` and `kauthError` properly)

jeschu1: Testing kauthResult/kauthError for `TryGetVirtualizationRoot ` is straightforward and since we are calling the actual function there we can easily validate.

Failures of `ProviderMessaging_TrySendRequestAndWaitForResponse ` are a bit more hairy.  In the mock I can make the fail cases return an error and we can validate on that.  

In the actual function shutting down for instance on `ProviderMessaging_TrySendRequestAndWaitForResponse ` would return false and kauthResult/kauthError would be unchanged.  Perhaps these should be set?

wilbaker: > Testing kauthResult/kauthError for TryGetVirtualizationRoot is straightforward and since we are calling the actual function there we can easily validate.

That sounds good.

> Failures of ProviderMessaging_TrySendRequestAndWaitForResponse are a bit more hairy. In the mock I can make the fail cases return an error and we can validate on that.

That sounds good to me too, and if you think it will be a major change I'm good with spinning out a follow up issue/PR for enhancing the `ProviderMessaging` mock and our validation of `ProviderMessaging` failure scenarios.

> In the actual function shutting down for instance on ProviderMessaging_TrySendRequestAndWaitForResponse would return false and kauthResult/kauthError would be unchanged. Perhaps these should be set?

I agree it would be best to set kauthResult/kauthError in that scenario.  Fixing up `ProviderMessaging_TrySendRequestAndWaitForResponse` seems outside of the scope of this PR though and so I think that work could be deferred to a follow up issue.

PR #1070, opened 2019-04-25
Revise and expand Linux event handling

chrisd8088:
These changes are necessary to match the changes in github/libprojfs#66 and to build on those to implement several new event notifications on Linux, specifically, the `OnHardLinkCreated` and `OnPreConvertToFull` events.

I suspect the only thing which may catch the eye of reviewers is that e16f446 removes some unused event notification types on both Linux and Mac, with the goal of keeping the two managed-code ProjFS libraries in sync wherever possible (like we already removed `NotificationMapping` on both platoforms in fc85024).  But it would be fine to make this change Linux-only, or skip it altogether.

@jrbriggs, @wilbaker, if you're able to review this PR (and #1042) fairly soon, that would be great, because until then we have a mismatch between libprojfs master and VFSForGit features/linuxprototype.  Thanks very much!


chrisd8088:
@wilbaker, @jrbriggs -- updated with merge conflicts from #1042 fixed!

chrisd8088:
Hi @jrbriggs, @wilbaker, @derrickstolee -- may I ask that someone take a look at this as soon as possible?  The issue is now that we're out-of-sync between libprojfs's master and features/linuxprototype, and the sooner we can merge this PR into features/linuxprototype, the better.

As I noted in the PR description, if you're not thrilled with e16f446 and the changes to remove unused notification types on the Mac, that's more than fine; I can easily pull that out.  I just wanted to keep the Linux and Mac code in alignment, where possible.

ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/ProjFS.Constants.cs
```diff
@@ -9,11 +9,12 @@ internal static class Constants
             public const ulong PROJFS_ALLOW = 0x01;
             public const ulong PROJFS_DENY = 0x02;
             public const ulong PROJFS_CLOSE_WRITE = 0x00000008;
-            public const ulong PROJFS_OPEN = 0x00000020;
-            public const ulong PROJFS_DELETE_SELF = 0x00000400;
-            public const ulong PROJFS_MOVE_SELF = 0x00000800;
+            public const ulong PROJFS_MOVE = 0x000000C0;
```
wilbaker: Why is `PROJFS_MOVE` `0x000000C0`?  (Wondering about the use of `C` vs a single bit)

wilbaker:
> verify there are no plans to need those on Mac.

We have had some discussions about adding post-delete notification, but I'm good with removing the dead code for now.

ProjFS.Linux/PrjFSLib.Linux.Managed/Interop/ProjFS.Constants.cs
chrisd8088: Well spotted!  Because Linux inotify delivers one or two messages for `rename()` (i.e., move) operations, [`IN_MOVE_FROM` and `IN_MOVE_TO`](https://github.com/torvalds/linux/blob/a4ccb5f9dc6c4fb4d4c0a9d73a911986f20ec88a/include/uapi/linux/inotify.h#L36), and then provides an [`IN_MOVE`](https://github.com/torvalds/linux/blob/a4ccb5f9dc6c4fb4d4c0a9d73a911986f20ec88a/include/uapi/linux/inotify.h#L50) convenience flag which combines the two.

The same is also true for fanotify [`FAN_MOVE`](https://github.com/torvalds/linux/blob/a4ccb5f9dc6c4fb4d4c0a9d73a911986f20ec88a/include/uapi/linux/fanotify.h#L34), because inotify and fanotify are aligned, and since we're trying to align libprojfs's events too -- in case we can potentially replace our FUSE libprofjs "shim" with some combination of OverlayFS and inotify/fanotify -- hence the matching definition here.

chrisd8088:
Thanks for the approvals -- will merge shortly.

PR #1071, opened 2019-04-25
Teach Enlistment.GetNewGVFSLogFileName to take FileSystem parameter

jamill:
This will enable the Enlistment.GetNewGVFSLogFileName method to interact with
the file system through the file system abstraction layer, enabling us to unit
test this method and methods that call this method.

With this change, the GitHubUpgrader tests should no longer create actual directories when running the unit tests.

### DIFF
--- a/GVFS/GVFS.Common/Enlistment.cs
+++ b/GVFS/GVFS.Common/Enlistment.cs
@@ -1,4 +1,5 @@
-﻿using GVFS.Common.Git;
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Git;
 using System;
 using System.IO;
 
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Common/Enlistment.cs
```diff
@@ -65,13 +66,17 @@ public abstract class Enlistment
 
         public GitAuthentication Authentication { get; }
 
-        public static string GetNewLogFileName(string logsRoot, string prefix, string logId = null)
+        public static string GetNewLogFileName(
+            PhysicalFileSystem fileSystem,
+            string logsRoot,
+            string prefix,
+            string logId = null)
```
kewillford: I don't like how this caused a bunch of `new PhysicalFileSystem()` calls to be added to that callers especially with what we were talking about yesterday with working towards having only one on the context or platform.  I haven't look at all the call sites but could we instead change this to not be a static call and pass the file system into the constructor of the Enlistment?  Not sure if all the call sites have an enlistment to use though.  But that would eventually allow us to mock the `GetNewLogFileName` if needed.  Statics make it really hard for unit testing and mocking.

jamill: >  I haven't look at all the call sites but could we instead change this to not be a static call and pass the file system into the constructor of the Enlistment?

Not all of the call sites have a `GVFSEnlistment` instance :) For example, the upgrader is not run in the context of an enlistment, and neither is the service callsites - which maybe hints that the method is not in the most appropriate location.

> Statics make it really hard for unit testing and mocking.

Yep - the motiviation for this change is around making the callers of this method more testable :)

kewillford: Okay thanks!

jamill:
> One suggestion. Using new PhysicalFileSystem() in the GetNewLogFileName method will make this change much smaller.

I like the idea of reducing the impact on call sites. I tweaked the method to optionally take in a `PhysicalFileSystem`, but not require it.

PR #1072, opened 2019-04-26
GitMaintenanceStep: Throw if stopped while Git is running

derrickstolee:
The GitMaintenanceStep throws a StoppingException if the
RunGitCommand() method is run after a Stop() call. However,
it doesn't throw when the Stop() call happens during the
Git process. This requires extra handling of the Stopping
boolean after receiving a response from RunGitCommand().
Such extra handling was added in #1064, but would not
be necessary with this change.

The better thing to do is to throw the StoppingException
after the Git process completes.

Add a unit test that verifies we automatically skip the
remainder of a PerformMaintenance() method in a subclass.

Looking through the error logs, all logged instances of a
failure during 'commit-graph verify' or 'multi-pack-index verify'
are actually due to this issue. Otherwise we would see errors
from GitMaintenanceStep saying "Git process {command} failed with errors:"
and those messages do no appear in the logs.

### DIFF
--- a/GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
+++ b/GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
@@ -15,15 +15,23 @@ namespace GVFS.UnitTests.Maintenance
     {
         private GVFSContext context;
 
+        public enum WhenToStop
+        {
+            Never,
+            BeforeGitCommand,
+            DuringGitCommand
+        }
+
         [TestCase]
         public void GitMaintenanceStepRunsGitAction()
         {
             this.TestSetup();
 
-            CheckMethodStep step = new CheckMethodStep(this.context);
+            CheckMethodStep step = new CheckMethodStep(this.context, WhenToStop.Never);
             step.Execute();
 
             step.SawWorkInvoked.ShouldBeTrue();
+            step.SawEndOfMethod.ShouldBeTrue();
         }
 
         [TestCase]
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
```diff
@@ -27,7 +27,6 @@ public void GitMaintenanceStepRunsGitAction()
         }
 
         [TestCase]
-        [Category(CategoryConstants.ExceptionExpected)]
```
jeschu1: As discussed the naming here isn't great, but we want this message in.  ExceptionExpected means don't run in debug mode.

GVFS/GVFS.UnitTests/Maintenance/GitMaintenanceStepTests.cs
```diff
@@ -108,5 +122,32 @@ protected override void PerformMaintenance()
                     nameof(this.SawWorkInvoked));
             }
         }
+
+        public class CheckStopDuringGitStep : GitMaintenanceStep
```
kewillford: I wonder if we should just have one class that has properties on when to call `Stop()` since these three class seem to all be the same except with if/when stop is called

derrickstolee: Good idea!

derrickstolee: [Done!](https://github.com/Microsoft/VFSForGit/pull/1072/commits/4a49bf20239919f619087ba8f104554e26b21b9e)

PR #1073, opened 2019-04-26
GVFS.Service should copy ProjectedFSLib.dll when it's missing and the installed ProjFS version matches the packaged version

### LINKED ISSUE #127: GVFS.Service should not proceed with installing ProjFS via INF if Get-WindowsOptionalFeature fails
GVFS.Service's current behavior makes things worse, and makes automatic recovery more difficult after a restart

### LINKED ISSUE #423: Uninstalling VFS For Git doesn't clean up filter driver
Running the VFS For Git uninstaller doesn't cleanup the projfs filter driver and screws up the following VFS For Git installation.

We've hit this issue while trying to upgrade several Azure agent pool machines.

wilbaker:
Fixes #127, fixes #423 

Updated GVFS.Service so that if the native ProjFS library is missing *and* the installed version of prjflt matches the version that ships with VFS4G, GVFS.Service will copy the native ProjFS library into `C:\Program Files\GVFS`.

_Manual testing_

- Installed the non-inbox ProjFS that ships with VFS4G and removed ProjectedFSLib.dll.  Restarted GVFS.Service and verified that it copied `ProjectedFSLib.dll` from `C:\Program Files\GVFS\ProjFS` to `C:\Program Files\GVFS`. 

- Installed an older version of the ProjFS driver and removed ProjectedFSLib.dll.  Restarted GVFS.Service and verified that it did not copy `ProjectedFSLib.dll` because the packaged prjflt version didn't match the installed prjflt version.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
+++ b/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
@@ -253,13 +253,6 @@ namespace GVFS.Platform.Windows
                     return true;
                 }
 
-                if (!isProjFSFeatureAvailable)
-                {
-                    // If enabling ProjFS failed because we were unable to find the optional feature, fallback
-                    // on installing ProjFS via the INF
-                    return TryInstallProjFSViaINF(tracer, fileSystem);
-                }
-
                 return false;
             }
 
--- a/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
+++ b/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
@@ -5,6 +5,7 @@ using Microsoft.Win32;
 using Microsoft.Windows.ProjFS;
 using System;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
@@ -26,6 +27,7 @@ namespace GVFS.Platform.Windows
         private const string PrjFltAutoLoggerStartValue = "Start";
 
         private const string System32LogFilesRoot = @"%SystemRoot%\System32\LogFiles";
+        private const string System32DriversRoot = @"%SystemRoot%\System32\drivers";
 
         // From "Autologger" section of prjflt.inf
         private const string FilterLoggerGuid = "ee4206ff-4a4d-452f-be56-6bd0ed272b44";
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
```diff
@@ -87,9 +87,26 @@ public static bool TryEnablePrjFlt(ITracer tracer, out string error)
                 isNativeLibInstalled = ProjFSFilter.IsNativeLibInstalled(tracer, new PhysicalFileSystem());
                 if (!isNativeLibInstalled)
                 {
-                    string missingNativeLibMessage = "Native library is not installed";
-                    error = string.IsNullOrEmpty(error) ? missingNativeLibMessage : $"{error}. {missingNativeLibMessage}";
-                    tracer.RelatedError($"{nameof(TryEnablePrjFlt)}: {missingNativeLibMessage}");
+                    if (isRunning)
+                    {
+                        tracer.RelatedInfo($"{nameof(TryEnablePrjFlt)}: Native ProjFS library is not installed, attempting to copy packaged version");
```
jeschu1: Would it be interesting to see this in telemetry?  I don't believe Info logged there by default.

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -286,6 +281,90 @@ public static bool IsNativeLibInstalled(ITracer tracer, PhysicalFileSystem fileS
             return existsInSystem32 || existsInAppDirectory;
         }
 
+        public static bool TryCopyNativeLibIfDriverVersionsMatch(ITracer tracer, PhysicalFileSystem fileSystem, out string copyNativeDllError)
```
jeschu1: I think the logic makes sense here, how were you able to test?

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
mjcheetham: I would say this should be a warning, since it's about an undesirable state that would cause problems, but we might be able to mitigate (copying the library file).

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
wilbaker: I've included some details in the "Manual testing" section of the PR description.  

I did all of the testing on my RS5+ machine, but I manually disabled the inbox ProjFS and then manually installed the ProjFS we package by running `C:\Program Files\GVFS\Filter\prjflt.inf`.

I used the same approach to verify the service catches version mismatches (by using an older version of prjflt.inf/prjflt.sys).

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
```diff
@@ -87,9 +87,26 @@ public static bool TryEnablePrjFlt(ITracer tracer, out string error)
                 isNativeLibInstalled = ProjFSFilter.IsNativeLibInstalled(tracer, new PhysicalFileSystem());
                 if (!isNativeLibInstalled)
                 {
-                    string missingNativeLibMessage = "Native library is not installed";
-                    error = string.IsNullOrEmpty(error) ? missingNativeLibMessage : $"{error}. {missingNativeLibMessage}";
-                    tracer.RelatedError($"{nameof(TryEnablePrjFlt)}: {missingNativeLibMessage}");
+                    if (isRunning)
```
jamill: Not part of your change, but as part of reviewing this I found the `isRunning` name ambigious - and what does it refer to. Looking up here, I see this refers to whether the service is running?

wilbaker: Thanks for pointing this out, I will update the name to something more descriptive.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
jamill: suggestion: `attempting to copy packaged version` => `attempting to copy version packaged with VFS for Git`

'packaged version` can be ambiguous? but this is just for debugging purposes, so maybe it will be clear to us...

wilbaker: > Would it be interesting to see this in telemetry? I don't believe Info logged there by default.

\+ @jrbriggs for his thoughts on this.  I can certainly update this code to send telemetry, but I'm not sure how we'd react to that data and I think we're trying to be frugal with the telemetry we're sending.

> I would say this should be a warning

I'll update this to a warning.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
```diff
@@ -87,9 +87,26 @@ public static bool TryEnablePrjFlt(ITracer tracer, out string error)
                 isNativeLibInstalled = ProjFSFilter.IsNativeLibInstalled(tracer, new PhysicalFileSystem());
```
jamill: also not your change - but we already have a `PhysicalFileSystem` instance

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
```diff
@@ -87,9 +87,26 @@ public static bool TryEnablePrjFlt(ITracer tracer, out string error)
...
+                        tracer.RelatedInfo($"{nameof(TryEnablePrjFlt)}: Native ProjFS library is not installed, attempting to copy packaged version");
+                        string copyNativeDllError;
+                        if (ProjFSFilter.TryCopyNativeLibIfDriverVersionsMatch(tracer, new PhysicalFileSystem(), out copyNativeDllError))
+                        {
+                            tracer.RelatedInfo($"{nameof(TryEnablePrjFlt)}: Successfully copied packaged ProjFS native library");
+                            isNativeLibInstalled = true;
+                        }
+                        else
+                        {
+                            tracer.RelatedError($"{nameof(TryEnablePrjFlt)}: Failed to copy packaged ProjFS native library {copyNativeDllError}");
```
jamill: Do we need to set the error message in this code path? It seems like before we set an error if we got to this point, but were missing the native ProjFS library.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
```diff
@@ -87,9 +87,26 @@ public static bool TryEnablePrjFlt(ITracer tracer, out string error)
...
+                    }
+                    else
+                    {
+                        string missingNativeLibMessage = "Native ProjFS library is not installed";
+                        error = string.IsNullOrEmpty(error) ? missingNativeLibMessage : $"{error}. {missingNativeLibMessage}";
+                        tracer.RelatedError($"{nameof(TryEnablePrjFlt)}: {missingNativeLibMessage}");
+                    }
                 }
 
                 bool isAutoLoggerEnabled = ProjFSFilter.IsAutoLoggerEnabled(tracer);
```
jamill: Do we want to run this logic if native prjflt library is missing?

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -286,6 +281,90 @@ public static bool IsNativeLibInstalled(ITracer tracer, PhysicalFileSystem fileS
...
+            {
+                copyNativeDllError = $"{system32PrjfltDriverPath} not found, unable to validate that packaged driver matches installed driver";
+                return false;
+            }
+
+            FileVersionInfo packagedDriverVersion;
+            FileVersionInfo system32DriverVersion;
+            try
+            {
+                packagedDriverVersion = FileVersionInfo.GetVersionInfo(installedPrjfltDriverPath);
```
jamill: Should these methods exist on `PhysicalFileSystem` to make this method more testable?

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
jrbriggs: I doubt this particular point would generate _too much_ data when compared with the rest of the telemetry exhaust, so I don't worry about the data volume perspective.

The reason to dispatch this is so we could understand how many (would-have-been) mount failures this logic is mitigating. However, I don't think the event has any value beyond that.

wilbaker: @jrbriggs thanks! I will update this to be sent via telemetry.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
wilbaker: > Do we want to run this logic if native prjflt library is missing?

There's no need to prevent enabling the auto-logger if ProjectedFSLib.dll is missing.  

Enabling the auto-logger will allow us to capture filter driver log messages.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
```diff
@@ -87,9 +87,26 @@ public static bool TryEnablePrjFlt(ITracer tracer, out string error)
...
-                    error = string.IsNullOrEmpty(error) ? missingNativeLibMessage : $"{error}. {missingNativeLibMessage}";
-                    tracer.RelatedError($"{nameof(TryEnablePrjFlt)}: {missingNativeLibMessage}");
+                    if (isRunning)
+                    {
+                        tracer.RelatedInfo($"{nameof(TryEnablePrjFlt)}: Native ProjFS library is not installed, attempting to copy packaged version");
+                        string copyNativeDllError;
+                        if (ProjFSFilter.TryCopyNativeLibIfDriverVersionsMatch(tracer, new PhysicalFileSystem(), out copyNativeDllError))
+                        {
+                            tracer.RelatedInfo($"{nameof(TryEnablePrjFlt)}: Successfully copied packaged ProjFS native library");
+                            isNativeLibInstalled = true;
```
wilbaker: Does this mean we'll need to adjust `isInstalled` as well?

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
wilbaker: Next commit will add this to `PhysicalFileSystem`, and I will look into adding unit test(s) in a subsequent commit.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
wilbaker: @jamill, I renamed several of this variables let me know what you think.

GVFS/GVFS.Service/Handlers/EnableAndAttachProjFSHandler.cs
wilbaker: Adjusted the logic here to remove `isInstalled` completely.

GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -251,13 +253,6 @@ public static bool TryEnableAutoLogger(ITracer tracer)
                     return true;
                 }
```
jamill: Is the method name - `TryEnableOrInstallDriver` - still accurate? I don't think it tries to install the driver anymore.

wilbaker: > Is the method name - TryEnableOrInstallDriver - still accurate? I don't think it tries to install the driver anymore.

Yep this is still accurate, we will install the driver when ProjFS is not inbox (see line 259):

`return TryInstallProjFSViaINF(tracer, fileSystem);`

jamill: ah - that was below the fold, and I missed it. Thanks!

GVFS/GVFS.UnitTests.Windows/Windows/Platform/ProjFSFilterTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        {
+            this.mockFileSystem.VerifyAll();
+        }
+
+        [TestCase]
+        public void IsNativeLibInstalled_ReturnsTrueWhenLibInSystem32()
+        {
+            this.mockFileSystem.Setup(fileSystem => fileSystem.FileExists(this.system32NativeLibPath)).Returns(true);
+            this.mockFileSystem.Setup(fileSystem => fileSystem.FileExists(this.nonInboxNativeLibInstallPath)).Returns(false);
+            ProjFSFilter.IsNativeLibInstalled(new MockTracer(), this.mockFileSystem.Object).ShouldBeTrue();
```
jamill: As every test includes a new `MockTracer`, you might consider setting this up in the `SetUp` method.

PR #1074, opened 2019-04-29
Separate LocalStorageRoot from WorkingDirectoryRoot

kivikakk:
On platforms where a layered filesystem is used (i.e. Linux), the `.git` directory must be accessed on the lower filesystem directly, and not through the projection as that would cause a call back into the provider.

This PR introduces the concept of a "local storage root", defined as the directory where actual stored objects are located. On platforms where filters are used (Windows and Mac), this is the same as the working directory root -- the hydrated files (and the `.git` directory) exist physically in place. On layered filesystems, the objects are stored in a separate directory (currently `.gvfs/lower` relative to the enlistment root). We need to be sure to make accesses to `.git` in this directory, not via the projection.

/cc @chrisd8088 @jrbriggs @wilbaker

### DIFF
--- a/GVFS/GVFS.Common/Enlistment.cs
+++ b/GVFS/GVFS.Common/Enlistment.cs
@@ -10,6 +10,7 @@ namespace GVFS.Common
         protected Enlistment(
             string enlistmentRoot,
             string workingDirectoryRoot,
+            string localStorageRoot,
             string repoUrl,
             string gitBinPath,
             string gvfsHooksRoot,
@@ -23,7 +24,8 @@ namespace GVFS.Common
 
             this.EnlistmentRoot = enlistmentRoot;
             this.WorkingDirectoryRoot = workingDirectoryRoot;
-            this.DotGitRoot = Path.Combine(this.WorkingDirectoryRoot, GVFSConstants.DotGit.Root);
+            this.LocalStorageRoot = localStorageRoot;
+            this.DotGitRoot = Path.Combine(this.LocalStorageRoot, GVFSConstants.DotGit.Root);
             this.GitBinPath = gitBinPath;
             this.GVFSHooksRoot = gvfsHooksRoot;
             this.FlushFileBuffersForPacks = flushFileBuffersForPacks;
@@ -54,6 +56,7 @@ namespace GVFS.Common
 
         public string EnlistmentRoot { get; }
         public string WorkingDirectoryRoot { get; }
+        public string LocalStorageRoot { get; }
         public string DotGitRoot { get; private set; }
         public abstract string GitObjectsRoot { get; protected set; }
         public abstract string LocalObjectsRoot { get; protected set; }
--- a/GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
+++ b/GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
@@ -82,7 +82,7 @@ namespace GVFS.Common.Prefetch.Git
         {
             string targetTreeSha;
             string headTreeSha;
-            using (LibGit2Repo repo = new LibGit2Repo(this.tracer, this.enlistment.WorkingDirectoryRoot))
+            using (LibGit2Repo repo = new LibGit2Repo(this.tracer, this.enlistment.LocalStorageRoot))
             {
                 targetTreeSha = repo.GetTreeSha(targetCommitSha);
                 headTreeSha = repo.GetTreeSha("HEAD");
@@ -125,7 +125,7 @@ namespace GVFS.Common.Prefetch.Git
                     GitProcess.Result result = this.git.DiffTree(
                         sourceTreeSha,
                         targetTreeSha,
-                        line => this.EnqueueOperationsFromDiffTreeLine(this.tracer, this.enlistment.WorkingDirectoryRoot, line));
+                        line => this.EnqueueOperationsFromDiffTreeLine(this.tracer, this.enlistment.LocalStorageRoot, line));
 
                     if (result.ExitCodeIsFailure)
                     {
@@ -202,7 +202,7 @@ namespace GVFS.Common.Prefetch.Git
 
         private void EnqueueOperationsFromLsTreeLine(ITracer activity, string line)
         {
-            DiffTreeResult result = DiffTreeResult.ParseFromLsTreeLine(line, this.enlistment.WorkingDirectoryRoot);
+            DiffTreeResult result = DiffTreeResult.ParseFromLsTreeLine(line, this.enlistment.LocalStorageRoot);
             if (result == null)
             {
                 this.tracer.RelatedError("Unrecognized ls-tree line: {0}", line);
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -192,11 +192,6 @@ namespace GVFS.Platform.POSIX
                 get { return string.Empty; }
             }
 
-            public override string InstallerExtension
-            {
-                get { return ".deb"; }
-            }
-
             public override string GVFSExecutableName
             {
                 get { return "gvfs"; }
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -22,7 +22,6 @@ namespace GVFS.Platform.POSIX
 
         public override IGitInstallation GitInstallation { get; } = new POSIXGitInstallation();
         public override IPlatformFileSystem FileSystem { get; } = new POSIXFileSystem();
-        public override GVFSPlatformConstants Constants { get; } = new POSIXPlatformConstants();
         public override void ConfigureVisualStudio(string gitBinPath, ITracer tracer)
         {
         }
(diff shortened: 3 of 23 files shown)


derrickstolee:
@kivikakk Could you include some reasons for why this change is necessary? The code change itself doesn't seem like it explains the issue at hand.

chrisd8088:
The issue here, @derrickstolee, is that the Linux implementation is uses a pass-through projecting filesystem ([libprojfs](https://github.com/github/libprojfs)) which we're integrating with VFSForGit through the use of a lower storage directory in `.gvfs/lower/`, and then `src/` is a Linux mount point which virtualizes the lower storage, adding the projected files and directories.

So our startup code requires that we pass in the additional path; see [GVFS.Platform.Linux.LinuxFileSystemVirtualizer.TryStart()](https://github.com/Microsoft/VFSForGit/blob/d3591f0eb895c13053ed06aee52f962e46e55fac/GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs#L231) in #986 and [PrjFSLib.Linux.VirtualizationInstance.StartVirtualizationInstance()](https://github.com/Microsoft/VFSForGit/blob/6a487f577252a49bd63573254673dc80b9aacb64/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs#L32) in the `features/linuxprototype` branch.

We've learned that if we mount "as usual" with the existing `GVFS` provider code, we stall during the mount because the provider is trying to access `src/.git/...` but `src/` is waiting for its initial hydration.  See our github/VFSForGit#15 issue for the details.

So @kivikakk has been working on disentangling the GVFS provider's accesses to the `.git` folder from the mount point (i.e., the working directory under `src/`).  The **client** would continue to use `src/.git` as normal, but the provider process itself should look "under the hood" at the storage location.  On Mac and Windows, those are the same, i.e., `src/.git/` but on Linux, they will be different.

Does that help at all in understanding our context?

kivikakk:
@derrickstolee Sorry for throwing you in the deep end here! There's an email chain ongoing with John and William with the context, so I only cc'd them. My colleague Chris has explained the situation well -- in short, on Windows and Mac, the location of the directory containing `.git` is always equal to the working directory, but on Linux it needs to be different to avoid the provider calling back into itself. Hence this PR, which separates out the concept of the `.git`-containing-directory into `StorageRoot`.

derrickstolee:
@chrisd8088, @kivikakk: I appreciate the details. Specifically, I'm hoping some of those details (including a link to the issue) make it into the PR description so anyone can come to the PR and find at least some context before diving into a review. The PR description then becomes the commit message for the merge commit (when we remember to copy/paste it during merge time) so we can track the reason for the change in the commit history.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -202,7 +202,7 @@ private void FlushStagedQueues()
 
         private void EnqueueOperationsFromLsTreeLine(ITracer activity, string line)
         {
-            DiffTreeResult result = DiffTreeResult.ParseFromLsTreeLine(line, this.enlistment.WorkingDirectoryRoot);
+            DiffTreeResult result = DiffTreeResult.ParseFromLsTreeLine(line, this.enlistment.WorkingDirectoryRoot); // XXX ?
```
derrickstolee: Please resolve whatever is the issue here.

GVFS/GVFS.Common/Enlistment.cs
```diff
@@ -53,6 +55,7 @@ public abstract class Enlistment
 
         public string EnlistmentRoot { get; }
         public string WorkingDirectoryRoot { get; }
+        public string StorageRoot { get; }
```
derrickstolee: This name is ambiguous. What are we storing? I was initially going to recommend `GitStorageRoot`, but I see later that we add the `.git` path to the working directory to get the `.git` folder. Whatever name we use here needs to contrast with things like the shared object cache.

Perhaps `LocalStorageRoot`? That can give some information that this is information limited to the enlistment.

Or, do you plan to put anything other than the `.git` folder in this location? Why not put the `.git` folder inside `EnlistmentRoot` and then point `DotGitRoot` there?

We have an opportunity here to think carefully about all of our options here, and later we will be stuck with our decisions. I just want to make sure we've considered all options.

chrisd8088: To add a bit of context, our libprojfs filesystem works much like any other stackable filesystem on Linux.  We've got a [design document](https://github.com/github/libprojfs/blob/master/docs/design.md#stackable-file-systems) which gets into some of the details as well as a our long-term plans; for now, we're using FUSE and a user-space implementation to thrash out issues which turn up getting VFSForGit to run -- exactly like this one!  :-)

One difference from a filter-based approach, like the PrjFlt NTFS filter or the Mac kauth filter, is that **everything** actually lives in the lower storage directory, not just `.git/` but the entire contents of the working directory (minus those files and directories which have not yet been hydrated).  If a user does `touch src/foo`, the new `foo` file is actually created as `.gvfs/lower/foo`, and is also visible as `src/foo`.  And you can technically read the contents of that file at either location, or modify it in either place, etc.

But the difference is that if you directly modify files in `.gvfs/lower/`, our libprojfs filesystem doesn't see those actions, and sends no events to the VFSForGit provider.  Moreover, if you try to read a not-yet-hydrated file in `.gvfs/lower/`, you just see an empty file; nothing happens to hydrate it.  You only get the hydration and projection actions if you work in `src/`.

This is a pretty typical thing on Linux, familiar to anyone using OverlayFS or eCryptFS.  But it does introduce a whole second "lower storage" directory.  And we've learned that we need to point the GVFS provider at that directory to avoid stalls when it goes looking at the contents of `.git/` before the initial directory is hydrated.

I'd be fine with `LocalStorageRoot` or whatever you think best in terms of naming.  To answer your two specific questions:
> Or, do you plan to put anything other than the .git folder in this location?

Yes, everything visible in the `src/` folder (i.e., the entire working directory's contents) actually is stored in `.gvfs/lower`, including the `.git/` folder.

> Why not put the .git folder inside EnlistmentRoot and then point DotGitRoot there?

I think, off the top of my head, that would break the general functionality of the `src/` folder, which is where the Git working directory is expected to be.  I don't think VFSForGit in general would accommodate such a change, but I might be mistaken!

derrickstolee: Thanks for the details! Since you are having a local copy of every hydrated file _and_ the `.git` folder in this directory, I think `LocalStorageRoot` is a good enough descriptor. It would not be appropriate to only update `DotGitRoot`.

kivikakk:
I think this should now be up to scratch.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -15,9 +15,10 @@ public abstract partial class POSIXPlatform : GVFSPlatform
     {
         private const string GVFSBinPath = "/usr/local/vfsforgit";
 
-        public POSIXPlatform(string installerExtension) : base(
+        public POSIXPlatform(string installerExtension, string localStorageRoot) : base(
```
derrickstolee: @kivikakk Are we going to need a `LinuxPlatform` class? Is that something you have in your feature branch, and can we add it to master? Or is there an existing plan for the feature integration? cc @wilbaker

chrisd8088: Just FYI, yes, there's a [`LinuxPlatform`](https://github.com/Microsoft/VFSForGit/tree/d3591f0eb895c13053ed06aee52f962e46e55fac/GVFS/GVFS.Platform.Linux) class coming soon to features/linuxprototype; see #986 and some of our other [WIP branches](https://github.com/github/VFSForGit/branches) in our dev fork.

derrickstolee: Thanks! I look forward to seeing different platform-specific pieces show up in small bites.

kivikakk:
@derrickstolee Thank you for your careful attention! :bow-woman: I'll keep an eye on that PR and ping here again when it's merged.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -186,7 +185,7 @@ public override bool TryKillProcessTree(int processId, out int exitCode, out str
         [DllImport("libc", EntryPoint = "getuid", SetLastError = true)]
         private static extern int Getuid();
 
-        public class POSIXPlatformConstants : GVFSPlatformConstants
+        public abstract class POSIXPlatformConstants : GVFSPlatformConstants
```
kivikakk: There should be no need to instantiate `POSIXPlatformConstants`, only subclasses which specify total behaviour. Accordingly I've marked it abstract and removed the `NotImplementedException` implementations for those which are overrided.

derrickstolee:
Looks like #1087 will no longer take any changes from `master`, so I'll go ahead and merge this. Thanks for the contribution, @kivikakk!

wilbaker:
@kivikakk @jrbriggs @derrickstolee apologies for being a bit late to the game on this PR.

The only feedback I have is that the `LocalStorageRoot` name is pretty generic doesn't describe what is being stored (git objects? working directory files? downloaded sizes?).  The name also made me think there might be "remote storage" as well (which I don't think we'll have here?).  Unfortunately we have several "roots" floating around already  (e.g. `LocalObjectsRoot`) and I could see there being confusion about additional root paths if we're not careful with the naming.

Based on its purpose, what would you think if we renamed `LocalStorageRoot` to something like: 
- `WorkingDirectoryBackingRoot` 
- `WorkingDirectoryBackingLayerRoot`
- `WorkingDirectoryStorageRoot`

Adding a comment describing the purpose/usage of the new path would be helpful as well.

Thoughts?

I'm happy to create a new PR with the changes I've proposed and continue the discussion there.

PR #1085, opened 2019-04-30
refresh AuthoringTests.md

kivikakk:
Fix Markdown formatting and some old references.

[Rendered](https://github.com/kivikakk/VFSForGit/blob/test-doc-refresh/AuthoringTests.md).

### DIFF
--- a/AuthoringTests.md
+++ b/AuthoringTests.md
@@ -4,55 +4,56 @@
 
 ### 1. Running the functional tests
 
-Our functional tests are in the GVFS.FunctionalTests project.  They are built on NUnit 3, which is available as a set of NuGet packages.
+Our functional tests are in the `GVFS.FunctionalTests` project.  They are built on NUnit 3, which is available as a set of NuGet packages.
 
 To run the functional tests:
 1. Open GVFS.sln in Visual Studio
 2. Build all, which will download the NUnit framework and runner
 3. You have three options for how to run the tests, all of which are equivalent.
-  a. Run the GVFS.FunctionalTests project.  Even better, set it as the default project and hit F5.
-  b. Use the command line runner.  After building, execute ```Scripts\RunFunctionalTests.bat```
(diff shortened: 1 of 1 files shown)


AuthoringTests.md
```diff
@@ -4,55 +4,56 @@
...
+
+    For any test fixture (a fixture is the same as a class in NUnit) that derives from this class, we create an enlistment and mount VFS for Git before running
 	any of the tests in the fixture, and then we unmount and delete the enlistment after all tests are done (but before any other fixture runs).  If you need
 	to write a sequence of tests that manipulate the same repo, this is the right base class.
 
-3. TestsWithEnlistmentPerTestCase
-	Derive from this class if you need a brand new enlistment per test case.  This is the most reliable, but also most expensive option.
+3. `TestsWithEnlistmentPerTestCase`
+
+   Derive from this class if you need a brand new enlistment per test case.  This is the most reliable, but also most expensive option.
```
kewillford: One other thing that would should add here is the use of `GitRepoTests` and use it if you want to test a git command and have the results of a GVFS repo compared against a regular git clone repo.

derrickstolee: Unfortunately, the repo containing `GitRepoTests` is not public, right? (It contains specific branch and commit names for private, large repos.) So, that is something for our team wiki and not the public docs.

kewillford: The `GitRepoTests` is an `abstract` class in the functional tests that is used to run git commands and compare the output of the git command on GVFS and on a regular git clone.  [This is all public](https://github.com/Microsoft/VFSForGit/blob/655ba9de69843c3b64e20cecaaca2c43bb4466fa/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L14) All the tests that are in the `Tests/GitCommands` folder use this base class.

derrickstolee: Neat! Consider me corrected. I was not familiar with that.

kewillford: We should probably rename it to something like `TestWithGitControlRepo` to match the other base class names and avoid the confusion.

PR #1090, opened 2019-05-01
Replace use of Count on FileSystemTaskQueue with IsEmpty

kewillford:
Using `Count` on `ConcurrentQueue` carries a significant performance cost when using .Net Core.  This should be fixed in an upcoming version but in the meantime `IsEmpty` is much more performant and recommended.  This change replaces the use of `Count` in performance critical paths for the Background operations.

See [Extremely poor performance of ConcurrentQueue.Count](https://github.com/dotnet/corefx/issues/29759)

### DIFF
--- a/GVFS/GVFS.UnitTests/Mock/Virtualization/Background/MockBackgroundTaskManager.cs
+++ b/GVFS/GVFS.UnitTests/Mock/Virtualization/Background/MockBackgroundTaskManager.cs
@@ -17,6 +17,8 @@ namespace GVFS.UnitTests.Mock.Virtualization.Background
 
         public List<FileSystemTask> BackgroundTasks { get; private set; }
 
+        public override bool IsEmpty => this.BackgroundTasks.Count == 0;
+
         public override int Count
         {
             get
--- a/GVFS/GVFS.Virtualization/Background/FileSystemTaskQueue.cs
+++ b/GVFS/GVFS.Virtualization/Background/FileSystemTaskQueue.cs
@@ -22,6 +22,18 @@ namespace GVFS.Virtualization.Background
         {
         }
 
+        public bool IsEmpty
+        {
+            get { return this.data.IsEmpty; }
+        }
+
+        /// <summary>
+        /// Gets the count of tasks in the queue
+        /// </summary>
+        /// <remarks>
+        /// This is an expensive call on .net core and you should avoid calling in performance critical paths.
+        /// Use the IsEmpty property when checking if the queue has any items instead of Count.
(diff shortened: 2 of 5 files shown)


GVFS/GVFS.Virtualization/Background/FileSystemTaskQueue.cs
```diff
@@ -22,6 +22,18 @@ private FileSystemTaskQueue(ITracer tracer, PhysicalFileSystem fileSystem, strin
...
+        {
+            get { return this.data.IsEmpty; }
+        }
+
+        /// <summary>
+        /// Gets the count of items in the queue
+        /// </summary>
+        /// <remarks>
+        /// This is an expensive call on .net core and you should avoid calling
+        /// in performance critical paths.  Use the IsEmpty property instead.
```
jamill: nit: `Use the IsEmpty property instead` - obviously this suggestion is only valid when you want to know if the queue is empty, and is not a replacement for all usages of `Count` :)

GVFS/GVFS.UnitTests/Mock/Virtualization/Background/MockBackgroundTaskManager.cs
```diff
@@ -17,6 +17,8 @@ public MockBackgroundFileSystemTaskRunner()
 
         public List<FileSystemTask> BackgroundTasks { get; private set; }
 
+        public override bool IsEmpty => this.BackgroundTasks.Count == 0;
```
jamill: consider using`!this.BackgroundTasks.Any()` - which I think is more idiomatic... but doesn't matter in practice

derrickstolee: Since it uses `List<T>`, `.Count == 0` is faster than `.Any()`.

jamill: > Since it uses List<T>, .Count == 0 is faster than .Any().

I doubt you will find any discernable difference in this test code :)

jamill: What made me think of this was @derrickstolee's comment on initial his confusion here. I think the code is fine either way :)

kewillford: I will leave it as is since there will be little difference and the confusion was mostly in the context of this PR and not the code in general.

PR #1093, opened 2019-05-02
[DO NOT MERGE] Update build to use VS2019

derrickstolee:
If you happen to uninstall Visual Studio 2017, the current build scripts will not work!

Hopefully this commit can unblock you, but I don't anticipate it being something we can take into master yet.

### DIFF
--- a/GVFS/GVFS.NativeTests/GVFS.NativeTests.vcxproj
+++ b/GVFS/GVFS.NativeTests/GVFS.NativeTests.vcxproj
@@ -14,7 +14,7 @@
     <ProjectGuid>{3771C555-B5C1-45E2-B8B7-2CEF1619CDC5}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>GVFSNativeTests</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.10240.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <Import Project="$(SolutionDir)\GVFS\GVFS.Build\ProjFS.props" />
@@ -22,13 +22,13 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>NotSet</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
     <WholeProgramOptimization>true</WholeProgramOptimization>
     <CharacterSet>NotSet</CharacterSet>
   </PropertyGroup>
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.NativeTests/GVFS.NativeTests.vcxproj
```diff
@@ -14,21 +14,21 @@
...
+    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <Import Project="$(SolutionDir)\GVFS\GVFS.Build\ProjFS.props" />
   <Import Project="$(SolutionDir)\GVFS\GVFS.Build\GVFS.cpp.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v142</PlatformToolset>
```
jeffhostetler: Changing from v141 to v142 will probably break VS2017 users, since I don't think it is available.
The alternative is to add the v141 package to your VS2019 installation.

I'm not sure how the build machines are configured.

derrickstolee: Oh, this commit will _definitely_ break 2017 users.

GVFS/GVFS.PostIndexChangedHook/GVFS.PostIndexChangedHook.Windows.vcxproj
```diff
@@ -14,7 +14,7 @@
     <ProjectGuid>{24D161E9-D1F0-4299-BBD3-5D940BEDD535}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>GVFSPostIndexChangedHook</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.10240.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
```
mjcheetham: 16299 is RS3. This should be OK?

wilbaker: > 16299 is RS3. This should be OK?

We still need to support RS1.  If we switch to RS3 are we risking the newer Windows SDK calling function(s) that aren't present in RS1?

wilbaker: >  If we switch to RS3 are we risking the newer Windows SDK calling function(s) that aren't present in RS1?

*Full disclosure:*  I'm not sure how large of a risk this is, or what the official guidance is on what Windows version to target.  Historically the concern has been that we might accidentally break RS1/Server2016.

GVFS/GVFS.NativeTests/GVFS.NativeTests.vcxproj
wilbaker: Is there any impact on VS for Mac?

jamill: You should be able to install the v141 build tools in VS 2019, under the "Desktop development with C++" components.

It might be a little soon to require everyone to update to VS 2019, but that is another option.

GVFS/GVFS.PostIndexChangedHook/GVFS.PostIndexChangedHook.Windows.vcxproj
jamill: Another option is to install the 10240 SDK manually if you only have VS2019:

https://developer.microsoft.com/en-us/windows/downloads/sdk-archive

It might be interesting to see what the requirements are around supporting specific versions of Windows.

jamill:
A slightly different approach (which assumes we you can install the older MSVC tool chains via the options, and the older Windows 10 SDKs) is spiked here:

https://github.com/jamill/VFSForGit/commit/6f1598ada17bef097d3f1b5fd5e398a402f00493

It basically tweaks the conditions in the VS version that it is looking for, and performs a manual check that the Windows 10 SDK version is installed at an expected location...

There is probably better ways to do what I am trying to do there...

jrbriggs:
We'll take #1097 instead

PR #1094, opened 2019-05-02
Delete corrupt loose objects during LooseObjectsStep

### LINKED ISSUE #1079: LooseObjectStep may fail due to corrupt loose object
During the `LooseObjectStep`, we can hit an issue where our loose object is corrupt. We don't respond well, and will continue retrying on every execution:

```
error: inflate: data stream error (unknown compression method)
error: unable to unpack {oid} header
error: inflate: data stream error (unknown compression method)
error: unable to unpack {oid} header
fatal: loose object {oid} (stored in {path}\gitObjects/{??}/{oid}) is corrupt
```

We could try to respond to this error by parsing the output and deleting the specific loose object. Perhaps we could make Git more resilient and have it delete the object and continue if corrupt. This would be a behavior change, as normally `git pack-objects` really _should_ fail if a given object can't be added to the pack.

derrickstolee:
If we have a corrupt loose object, the `git pack-objects` command will fail as it cannot put that object into a pack-file. This stops all future iterations of the `LooseObjectsStep` from succeeding, so users will be stuck in a bad situation.

* Refactor `WriteLooseObjectIds()` to rely on a new `LooseObjectsBatch()` method.

* On failure to create a pack, check all objects in the batch (using `LooseObjectsBatch()` and libgit2) and delete those that fail.

* Start disposing the repo in the dehydrate verb for safety.

* Create a functional test that verifies this behavior. Move the `LooseObjectStepTests` to have a new enlistment for each test, since we are messing with the object store. Part of the issue is that .NET doesn't allow us to move files that are marked as read-only, which becomes problematic as tests create pack-indexes and other files.

Resolves #1079

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -162,6 +162,13 @@ namespace GVFS.Common.Maintenance
 
             foreach (string objectId in this.LooseObjectsBatch(this.MaxLooseObjectsInPack))
             {
+                // Check if the loose object is a valid blob. Several cases can make
+                // `isBlob` be false:
+                //
+                // 1. The object is corrupt. This is the case that is blocking the pack-objects call.
+                // 2. The object does not exist. This is unlikely, since we see the file in the object directory.
+                // 3. The object is not a blob. Deleting a commit or tree here is a low cost, since we
+                //    will probably get those objects in prefetch packs. Re-downloading will be a low cost.
                 if (!this.Context.Repository.TryGetIsBlob(objectId, out bool isBlob))
                 {
                     numTryGetIsBlobFailures++;
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -118,18 +118,50 @@ namespace GVFS.Common.Maintenance
         /// Creates a pack file from loose objects
         /// </summary>
         /// <returns>The number of loose objects added to the pack file</returns>
-        public int CreateLooseObjectsPackFile()
+        public bool TryCreateLooseObjectsPackFile(out int objectsAddedToPack)
         {
-            int objectsAddedToPack = 0;
+            int localObjectCount = 0;
 
             GitProcess.Result result = this.RunGitCommand(
                 (process) => process.PackObjects(
                     "from-loose",
                     this.Context.Enlistment.GitObjectsRoot,
-                    (StreamWriter writer) => objectsAddedToPack = this.WriteLooseObjectIds(writer)),
+                    (StreamWriter writer) => localObjectCount = this.WriteLooseObjectIds(writer)),
                 nameof(GitProcess.PackObjects));
 
-            return objectsAddedToPack;
+            if (result.ExitCodeIsSuccess)
+            {
+                objectsAddedToPack = localObjectCount;
+                return true;
+            }
+            else
+            {
+                objectsAddedToPack = 0;
+                return false;
+            }
+        }
+
+        public int ClearCorruptLooseObjects()
+        {
+            int numBadObjects = 0;
+            foreach (string objectId in this.LooseObjectsBatch(this.MaxLooseObjectsInPack))
+            {
+                if (!this.Context.Repository.TryGetIsBlob(objectId, out bool isBlob) ||
+                    !isBlob)
+                {
+                    string objectFile = Path.Combine(
+                                                this.Context.Enlistment.GitObjectsRoot,
+                                                objectId.Substring(0, 2),
+                                                objectId.Substring(2, GVFSConstants.ShaStringLength - 2));
+
+                    if (this.Context.FileSystem.TryDeleteFile(objectFile))
+                    {
+                        numBadObjects++;
+                    }
+                }
+            }
+
+            return numBadObjects;
         }
 
         protected override void PerformMaintenance()
@@ -159,7 +191,7 @@ namespace GVFS.Common.Maintenance
                     GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot), nameof(GitProcess.PrunePacked));
                     int afterLooseObjectsCount = this.CountLooseObjects();
 
-                    int objectsAddedToPack = this.CreateLooseObjectsPackFile();
+                    bool createPack = this.TryCreateLooseObjectsPackFile(out int objectsAddedToPack);
 
                     EventMetadata metadata = new EventMetadata();
                     metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
@@ -167,6 +199,13 @@ namespace GVFS.Common.Maintenance
                     metadata.Add("EndingCount", afterLooseObjectsCount);
                     metadata.Add("RemovedCount", beforeLooseObjectsCount - afterLooseObjectsCount);
                     metadata.Add("LooseObjectsPutIntoPackFile", objectsAddedToPack);
+
+                    if (!createPack)
+                    {
+                        int removedObjects = this.ClearCorruptLooseObjects();
+                        metadata.Add("RemovedCorruptObjects", removedObjects);
+                    }
+
                     activity.RelatedEvent(EventLevel.Informational, $"{this.Area}_{nameof(this.PerformMaintenance)}", metadata, Keywords.Telemetry);
                     this.SaveLastRunTimeToFile();
                 }
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -30,7 +30,7 @@ namespace GVFS.Common.Maintenance
         {
             Succeess,
             UnknownFailure,
-            CorruptBlob
+            CorruptObject
         }
 
         public override string Area => nameof(LooseObjectsStep);
@@ -147,7 +147,7 @@ namespace GVFS.Common.Maintenance
 
                 if (result.Errors.Contains("is corrupt"))
                 {
-                    return CreatePackResult.CorruptBlob;
+                    return CreatePackResult.CorruptObject;
                 }
 
                 return CreatePackResult.UnknownFailure;
@@ -225,7 +225,7 @@ namespace GVFS.Common.Maintenance
                     metadata.Add("LooseObjectsPutIntoPackFile", objectsAddedToPack);
                     metadata.Add("CreatePackResult", createPackResult.ToString());
 
-                    if (createPackResult == CreatePackResult.CorruptBlob)
+                    if (createPackResult == CreatePackResult.CorruptObject)
                     {
                         this.ClearCorruptLooseObjects(metadata);
                     }
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -154,6 +154,14 @@ namespace GVFS.Common.Maintenance
             }
         }
 
+        public string GetLooseObjectFileName(string objectId)
+        {
+            return Path.Combine(
+                            this.Context.Enlistment.GitObjectsRoot,
+                            objectId.Substring(0, 2),
+                            objectId.Substring(2, GVFSConstants.ShaStringLength - 2));
+        }
+
         public int ClearCorruptLooseObjects(EventMetadata metadata)
         {
             int numBadObjects = 0;
@@ -177,10 +185,7 @@ namespace GVFS.Common.Maintenance
 
                 if (!isBlob)
                 {
-                    string objectFile = Path.Combine(
-                                                this.Context.Enlistment.GitObjectsRoot,
-                                                objectId.Substring(0, 2),
-                                                objectId.Substring(2, GVFSConstants.ShaStringLength - 2));
+                    string objectFile = this.GetLooseObjectFileName(objectId);
 
                     if (this.Context.FileSystem.TryDeleteFile(objectFile))
                     {
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -26,6 +26,13 @@ namespace GVFS.Common.Maintenance
             this.forceRun = forceRun;
         }
 
+        public enum CreatePackResult
+        {
+            Succeess,
+            UnknownFailure,
+            CorruptBlob
+        }
+
         public override string Area => nameof(LooseObjectsStep);
 
         // 50,000 was found to be the optimal time taking ~5 minutes
@@ -51,7 +58,7 @@ namespace GVFS.Common.Maintenance
             return count;
         }
 
-        public IEnumerable<string> LooseObjectsBatch(int count)
+        public IEnumerable<string> LooseObjectsBatch(int batchSize)
         {
             // Find loose Objects
             foreach (DirectoryItemInfo directoryItemInfo in this.Context.FileSystem.ItemsInDirectory(this.Context.Enlistment.GitObjectsRoot))
@@ -72,10 +79,10 @@ namespace GVFS.Common.Maintenance
                                 continue;
                             }
 
-                            count--;
+                            batchSize--;
                             yield return objectId;
 
-                            if (count <= 0)
+                            if (batchSize <= 0)
                             {
                                 yield break;
                             }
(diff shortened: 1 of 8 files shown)


derrickstolee:
In addition to the functional test, I tried this on my machine. The object gets deleted as expected, along with additional logging.

Good news: if the corrupt loose object already exists in a pack, then the `git prune-packed` command will delete it without checking its contents.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +118,50 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+                return false;
+            }
+        }
+
+        public int ClearCorruptLooseObjects()
+        {
+            int numBadObjects = 0;
+            foreach (string objectId in this.LooseObjectsBatch(this.MaxLooseObjectsInPack))
+            {
+                if (!this.Context.Repository.TryGetIsBlob(objectId, out bool isBlob) ||
```
jeschu1: Can you confirm how we get into this state?
1. The loose object is missing? (TryGetIsBlob fails)
2. The object is corrupt? (fails the !isBlob)

Some comments here would be useful

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +118,50 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+                    !isBlob)
+                {
+                    string objectFile = Path.Combine(
+                                                this.Context.Enlistment.GitObjectsRoot,
+                                                objectId.Substring(0, 2),
+                                                objectId.Substring(2, GVFSConstants.ShaStringLength - 2));
+
+                    if (this.Context.FileSystem.TryDeleteFile(objectFile))
```
jeschu1: Is it worth logging when this fails?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +118,50 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+            int numBadObjects = 0;
+            foreach (string objectId in this.LooseObjectsBatch(this.MaxLooseObjectsInPack))
+            {
+                if (!this.Context.Repository.TryGetIsBlob(objectId, out bool isBlob) ||
+                    !isBlob)
+                {
+                    string objectFile = Path.Combine(
+                                                this.Context.Enlistment.GitObjectsRoot,
+                                                objectId.Substring(0, 2),
+                                                objectId.Substring(2, GVFSConstants.ShaStringLength - 2));
```
jeschu1: Is it worth making the objectFile into a utility function?  

I saw this in another area of code:
            string blobPath = Path.Combine(
                this.enlistment.GitObjectsRoot,
                blobSha.Substring(0, 2),
                blobSha.Substring(2));

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -51,15 +51,8 @@ public int CountLooseObjects()
             return count;
         }
 
-        /// <summary>
-        /// Writes loose object Ids to streamWriter
-        /// </summary>
-        /// <param name="streamWriter">Writer to which SHAs are written</param>
-        /// <returns>The number of loose objects SHAs written to the stream</returns>
-        public int WriteLooseObjectIds(StreamWriter streamWriter)
+        public IEnumerable<string> LooseObjectsBatch(int count)
```
jeschu1: If this is easy to add a unit test for (I think it should be), I would.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +118,50 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+                    (StreamWriter writer) => localObjectCount = this.WriteLooseObjectIds(writer)),
                 nameof(GitProcess.PackObjects));
 
-            return objectsAddedToPack;
+            if (result.ExitCodeIsSuccess)
+            {
+                objectsAddedToPack = localObjectCount;
+                return true;
+            }
+            else
```
jeschu1: Is a corrupt object the only reason why we wouldn't return ExitCodeIsSuccess?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: The way we get into this state is unclear, but the issue is that the file exists and is corrupt, so Git can't do anything about it.

Most likely, these are corruptions from many versions ago before we wrote loose objects safely. (See `GitObjects.WriteLooseObject()` for the details on our current writes.) But it could also be from disk problems.

jeschu1: Makes sense, perhaps `isValidBlob` rather than just `isBlob `may be helpful.  
A one line comment would also do the trick.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: `WriteLooseObjectIds()` is covered by unit tests, so this method is covered by proxy.

`batchSize` _is_ better, thanks!

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: Maybe not? I'll scan the error message first.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
derrickstolee: I'm not ignoring this, just saving it for a later push.

derrickstolee: The previous use in `GitObjects` does too much to split out nicely (it creates the directory if needed and returns the folder AND file name). Hopefully it is enough that I made a method in `GitLooseObjectsStep` and created a unit test.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +125,92 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+
+        public int ClearCorruptLooseObjects(EventMetadata metadata)
+        {
+            int numBadObjects = 0;
+            int numTryGetIsBlobFailures = 0;
+            int numFailedDeletes = 0;
+
+            foreach (string objectId in this.LooseObjectsBatch(this.MaxLooseObjectsInPack))
+            {
+                // Check if the loose object is a valid blob. Several cases can make
```
derrickstolee: On second thought, this is a bad idea. I'll adapt `TryGetIsBlob` to create `TryGetIsValidObject()` instead. Will avoid all of these problems.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -51,15 +58,8 @@ public int CountLooseObjects()
             return count;
         }
 
-        /// <summary>
-        /// Writes loose object Ids to streamWriter
-        /// </summary>
-        /// <param name="streamWriter">Writer to which SHAs are written</param>
-        /// <returns>The number of loose objects SHAs written to the stream</returns>
-        public int WriteLooseObjectIds(StreamWriter streamWriter)
+        public IEnumerable<string> LooseObjectsBatch(int batchSize)
```
kewillford: Rename `GetBatchOfLooseObjects`

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +125,85 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+
+        public int ClearCorruptLooseObjects(EventMetadata metadata)
+        {
+            int numBadObjects = 0;
+            int numTryGetIsValidFailures = 0;
+            int numFailedDeletes = 0;
+
+            foreach (string objectId in this.LooseObjectsBatch(this.MaxLooseObjectsInPack))
+            {
+                if (!this.Context.Repository.TryGetIsValidObject(objectId, out bool isBlob))
```
kewillford: Rename `isBlob` to `isValidObject`?

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +125,85 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+                        numBadObjects++;
+                    }
+                    else
+                    {
+                        numFailedDeletes++;
+                    }
+                }
+            }
+
+            if (numTryGetIsValidFailures > 0)
```
kewillford: I would just put this in the `metadata` all the time even when zero.  Same with the `numFailedDeletes`.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -148,17 +230,24 @@ protected override void PerformMaintenance()
                     }
 
                     int beforeLooseObjectsCount = this.CountLooseObjects();
-                    GitProcess.Result result = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot), nameof(GitProcess.PrunePacked));
+                    GitProcess.Result gitResult = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot), nameof(GitProcess.PrunePacked));
                     int afterLooseObjectsCount = this.CountLooseObjects();
 
-                    int objectsAddedToPack = this.CreateLooseObjectsPackFile();
+                    CreatePackResult result = this.TryCreateLooseObjectsPackFile(out int objectsAddedToPack);
```
kewillford: `result` -> `createPackResult`

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -148,17 +230,24 @@ protected override void PerformMaintenance()
...
                     metadata.Add("GitObjectsRoot", this.Context.Enlistment.GitObjectsRoot);
                     metadata.Add("StartingCount", beforeLooseObjectsCount);
                     metadata.Add("EndingCount", afterLooseObjectsCount);
                     metadata.Add("RemovedCount", beforeLooseObjectsCount - afterLooseObjectsCount);
                     metadata.Add("LooseObjectsPutIntoPackFile", objectsAddedToPack);
+                    metadata.Add("CreatePackResult", result.ToString());
+
+                    if (result == CreatePackResult.CorruptBlob)
+                    {
+                        int removedObjects = this.ClearCorruptLooseObjects(metadata);
```
kewillford: Do we need this method to return anything if we aren't going to use it?  We already added the information to the `metadata`.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
jrbriggs: Agreed, especially since we only would add any metadata if something was corrupt.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -110,18 +125,85 @@ public bool TryGetLooseObjectId(string directoryName, string filePath, out strin
...
+                    continue;
+                }
+
+                if (!isBlob)
+                {
+                    string objectFile = this.GetLooseObjectFileName(objectId);
+
+                    if (this.Context.FileSystem.TryDeleteFile(objectFile))
+                    {
+                        numBadObjects++;
```
jrbriggs: nit: numDeletedObjects

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/LooseObjectStepTests.cs
```diff
@@ -25,33 +25,24 @@ public LooseObjectStepTests()
 
         private string GitObjectRoot => this.Enlistment.GetObjectRoot(this.fileSystem);
         private string PackRoot => this.Enlistment.GetPackRoot(this.fileSystem);
-        private string TempPackRoot=> Path.Combine(this.PackRoot, TempPackFolder);
+        private string TempPackRoot => Path.Combine(this.PackRoot, TempPackFolder);
 
         [TestCase]
-        public void RemoveLooseObjectsInPackFiles()
+        [Order(1)]
```
jeschu1: Does order matter when you do EnlistmentPerTestCase?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/LooseObjectStepTests.cs
```diff
@@ -62,13 +53,13 @@ public void PutLooseObjectsInPackFiles()
...
-            this.GetLooseObjectFiles().Count.ShouldBeAtLeast(1);
+            int looseObjectCount = this.GetLooseObjectFiles().Count();
+            looseObjectCount.ShouldBeAtLeast(1);
 
             // This step should put the loose objects into a packfile
             this.Enlistment.LooseObjectStep();
 
-            this.GetLooseObjectFiles().Count.ShouldEqual(loooseObjectCount);
+            this.GetLooseObjectFiles().Count.ShouldEqual(looseObjectCount);
             this.CountPackFiles().ShouldEqual(1);
```
jeschu1: Should we verify that previuosly there were 0 packfiles?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/LooseObjectStepTests.cs
```diff
@@ -79,16 +70,74 @@ public void PutLooseObjectsInPackFiles()
...
+            // Delete/Move any starting loose objects and packfiles
             this.DeleteFiles(this.GetLooseObjectFiles());
+            this.MovePackFilesToTemp();
             this.GetLooseObjectFiles().Count.ShouldEqual(0);
-            int startingPackFileCount = this.CountPackFiles();
+            this.CountPackFiles().ShouldEqual(0);
+
+            // Copy and expand one pack
+            this.ExpandOneTempPack(copyPackBackToPackDirectory: true);
+            Assert.AreNotEqual(0, this.GetLooseObjectFiles().Count);
```
jeschu1: I don't think you need the above Assert, given the below ShouldBeAtLeast(1)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/LooseObjectStepTests.cs
```diff
@@ -79,16 +70,74 @@ public void PutLooseObjectsInPackFiles()
...
-            this.CountPackFiles().ShouldEqual(startingPackFileCount);
+            this.CountPackFiles().ShouldEqual(1);
+        }
+
+        [TestCase]
+        [Order(4)]
+        public void CorruptLooseObjectIsDeleted()
+        {
+            // Delete/Move any starting loose objects and packfiles
+            this.DeleteFiles(this.GetLooseObjectFiles());
```
jeschu1: These 4 functions look the same as the ones in `RemoveLooseObjectsInPackFiles`.  Consider a common function?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/LooseObjectStepTests.cs
derrickstolee: We do, it's just hidden in the diff collapse.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/LooseObjectStepTests.cs
```diff
@@ -25,50 +25,42 @@ public LooseObjectStepTests()
...
-            this.DeleteFiles(this.GetLooseObjectFiles());
-            this.MovePackFilesToTemp();
-            this.GetLooseObjectFiles().Count.ShouldEqual(0);
-            this.CountPackFiles().ShouldEqual(0);
+            this.ClearAllObjects();
 
             // Copy and expand one pack
             this.ExpandOneTempPack(copyPackBackToPackDirectory: true);
-            Assert.AreNotEqual(0, this.GetLooseObjectFiles().Count);
+            this.GetLooseObjectFiles().Count.ShouldEqual(0);
```
derrickstolee: This is wrong. Test failed because of it.

PR #1097, opened 2019-05-03
Enable building with Visual Studio 2019

jamill:
Update the Windows build script to be able to build with Visual
Studio 2019 tooling and update Readme with details. Includes the
following changes:

  * Tweak logic to find msbuild for different Visual Studio versions

  * Remove checks for unnecessary packages

  * Add checks for .Net Core 2.1 development packages

  * Tweak check for Windows 10 SDK (10.0.10240.0)

Note: I have not tested this yet without VS 2017 installed.

Related PR: #1093

### DIFF
--- a/Scripts/BuildGVFSForWindows.bat
+++ b/Scripts/BuildGVFSForWindows.bat
@@ -1,6 +1,6 @@
 @ECHO OFF
 SETLOCAL
-
+setlocal enabledelayedexpansion
 CALL %~dp0\InitializeEnvironment.bat || EXIT /b 10
 
 IF "%1"=="" (SET "Configuration=Debug") ELSE (SET "Configuration=%1")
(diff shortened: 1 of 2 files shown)


Scripts/BuildGVFSForWindows.bat
```diff
@@ -15,23 +15,32 @@ IF NOT EXIST %nuget% (
...
   set VsInstallDir=%%i
 )
 
+:: Assumes default installation location for Windows 10 SDKs
+IF NOT EXIST "c:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0" (
+  echo ERROR: Could not find Windows 10 SDK Version 10240
+  exit /b 1
+)
+
+for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -products * -requires Microsoft.Component.MSBuild Microsoft.VisualStudio.Workload.ManagedDesktop Microsoft.VisualStudio.Workload.NativeDesktop Microsoft.VisualStudio.Workload.NetCoreTools Microsoft.Net.Core.Component.SDK.2.1 -find MSBuild\**\Bin\amd64\MSBuild.exe`) do (
```
jamill: I don't like repeating all the requirements here, but maybe there is a better solution...

jrbriggs: IIRC the only thing we use VsInstallDir for is as a root to find msbuild. Can't we just remove the first vswhere invocation and change line 38 to test msbuild and give the same error message? It's still accurate.

Scripts/BuildGVFSForWindows.bat
```diff
@@ -15,23 +15,32 @@ IF NOT EXIST %nuget% (
...
+:: Use vswhere to find the latest VS installation with the msbuild component.
 :: See https://github.com/Microsoft/vswhere/wiki/Find-MSBuild
-for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -version "[15.0,16.0)" -products * -requires Microsoft.Component.MSBuild Microsoft.VisualStudio.Workload.ManagedDesktop Microsoft.VisualStudio.Workload.NativeDesktop Microsoft.VisualStudio.Workload.NetCoreTools Microsoft.Component.NetFX.Core.Runtime Microsoft.VisualStudio.Component.Windows10SDK.10240 -property installationPath`) do (
+for /f "usebackq tokens=*" %%i in (`%vswhere% -all -prerelease -latest -products * -requires Microsoft.Component.MSBuild Microsoft.VisualStudio.Workload.ManagedDesktop Microsoft.VisualStudio.Workload.NativeDesktop Microsoft.VisualStudio.Workload.NetCoreTools Microsoft.Net.Core.Component.SDK.2.1 -property installationPath`) do (
   set VsInstallDir=%%i
 )
 
+:: Assumes default installation location for Windows 10 SDKs
+IF NOT EXIST "c:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0" (
+  echo ERROR: Could not find Windows 10 SDK Version 10240
```
jrbriggs: Can we point the user to the readme like we do in the "can't find VS" error message?

PR #1098, opened 2019-05-03
Update PrjFSKextLogDaemon to forward messages to GitServiceDaemon

### LINKED ISSUE #648: Mac: Generate PrjFSKextVersion as part of the build

### LINKED ISSUE #927: Mac: Update PrjFSKextLogDaemon to communicate with the GitServiceDaemon
PrjFSKextLogDaemon listens to messages from the kext and writes them to the system log.  It should also connect to GitServiceDaemon to allow the kext to add messages to the telemetry stream.

wilbaker:
Resolves #927 
Resolves #648 

**Summary of Changes**
- Updated PrjFSKextLogDaemon to connect to GitServiceDaemon's socket (if present) and send it vnode cache health messages over its socket.  Currently these messages are being sent over the "vfs" socket.
- PrjFSKextLogDaemon will attempt to connect to the GitServiceDaemon at startup and every 30 minutes (if not already connected).
- Made a few small fixes/tweaks to `common.posix.cpp` based on issues I found while working on the PrjFSKextLogDaemon changes.

**TODO**
- [X] Decide if `s_messageListenerSocketPath` should be hardcoded or read from a config file.  (Decision: It will be hardcoded for now)
- [X] Send messages from the kext to the GitServiceDaemon (currently only vnode cache health data is being sent)
- [X] Decide if we want PrjFSKextLogDaemon to use the `Microsoft.Git.GVFS` provider name in the messages it sends to GitServiceDaemon
- [X] Send the correct PrjFS version number to the message listener
- [X] Add unit tests for JSON writing code
- [X] Remove duplication when logging to os_log and message listener
- [X] Update release build definition to ~~use kext build script~~ generate PrjFSVersion.h before building the kext

**Remaining Work (follow up PRs)**
- Update the kext to provide the VFS4G mount process with an ID that can be used to associate kext messages that are repo specific with their associated mount process. (#1180)
- Generate an xcconfig file with the PrjFS version, and use it to set the correct version in the kext's plist file. (#1181)
- Ensure the PrjFS/VFS4G version is properly stamped on the native binaries (#1181)
- Clean up how the shared user-mode code is structured (#1182)


ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -221,12 +239,141 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+    if (INVALID_SOCKET_FD == s_messageListenerSocket)
+    {
+        return;
+    }
+
+    // TODO: Properly version PrjFSKextLogDaemon and other Mac binaries
+    string fullMessage =
+    "{\""
+        "version\":\"0.6.XXX.X\","
+        "\"providerName\":\"Microsoft.Git.GVFS\","
```
wilbaker: @jeffhostetler @jrbriggs, let me know your thoughts on sending these messages over using the  "Microsoft.Git.GVFS" provider name.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -5,17 +5,29 @@
...
+using std::to_string;
 
 static const char PrjFSKextLogDaemon_OSLogSubsystem[] = "org.vfsforgit.prjfs.PrjFSKextLogDaemon";
+static const int INVALID_SOCKET_FD = -1;
 
 static os_log_t s_daemonLogger, s_kextLogger;
 static IONotificationPortRef s_notificationPort;
 
+static int s_messageListenerSocket = INVALID_SOCKET_FD;
+static string s_messageListenerSocketPath = "/usr/local/GitService/pipe/vfs-c780ac06-135a-4e9e-ab6c-d41e2d265baa";
```
wilbaker: Should this be in a config, and should that change be part of this PR?

It would be a bit of a hack but we also have the option of reading this from `GIT_TR2_EVENT`.

/cc: @jrbriggs @jeffhostetler for their thoughts

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeffhostetler: The provider name is an ETW holdover.  I only needed it in the protocol between the mount process and GitService to let me route it to the right GUID in ETW world.  For AppIns, I secretly abbreviate it to “VFS” and combine it with the event name.  That’s why you’re seeing justo events with “vfs.heartbeat” and etc.

It really doesn’t matter in the AI world because I can send everything to the same AI resource id/group (and without having to worry about registering a new (providerName, eventName) pair in some DB somewhere).

Once we drop ETW completely, we can revisit abbreviating them in the mount and kext processes.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeffhostetler: Just leave it for now or #define it somewhere.  Or see whe Matthew is doing on the mount side.

Don’t use the GIT_TR2_EVENT. It points to a different pipe.  And I’m deprecating it in favor of a system config setting that’ll be easier for the Mac installers to set.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -221,12 +239,141 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+        return;
+    }
+    
+    struct sockaddr_un socket_address;
+    memset(&socket_address, 0, sizeof(struct sockaddr_un));
+    
+    socket_address.sun_family = AF_UNIX;
+    size_t resultLength = strlcpy(socket_address.sun_path, s_messageListenerSocketPath.c_str(), sizeof(socket_address.sun_path));
+    
+    if (resultLength >= sizeof(socket_address.sun_path))
```
jeffhostetler: Would it be better to check for buffer space before doing the truncated copy?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: > see whe Matthew is doing on the mount side.

It looks like on the mount side we're shelling out to git and looking for the `gvfs.telemetry-pipe` config value (in the global or system git config).

> in favor of a system config setting that’ll be easier for the Mac installers to set.

This makes me wonder if all three applications (Git, VFS4G, PrjFSKextLogDaemon) should use the same config setting or each have their own.  In the case of PrjFSKextLogDaemon there no reason the config setting would have to be in a git config, it could be in a local PrjFSKextLogDaemon specific config file as well.

I guess it partially comes down to if we expect PrjFSKextLogDaemon to be used on machines without Git/VFS4G.  In the short to medium term I think the answer is that we don't expect \ scenarios like that (unlike Windows, ProjFS on Mac is not currently a standalone product).

/cc: @jrbriggs @mjcheetham

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -221,12 +239,141 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+        return;
+    }
+
+    // TODO: Properly version PrjFSKextLogDaemon and other Mac binaries
+    string fullMessage =
+    "{\""
+        "version\":\"0.6.XXX.X\","
+        "\"providerName\":\"Microsoft.Git.GVFS\","
+        "\"eventName\":\"PrjFSKextLogDaemon\","
+        "\"eventLevel\":2,"
```
jeffhostetler: You might check, but IIRC I got ride of the event level and opcode fields.

jeffhostetler: (On my iPad right now and can’t check...)

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -221,12 +239,141 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+    if (INVALID_SOCKET_FD != s_messageListenerSocket)
+    {
+        close(s_messageListenerSocket);
+        s_messageListenerSocket = INVALID_SOCKET_FD;
+    }
+}
+
+static void WriteJsonToMessageListener(const string& jsonMessage)
+{
+    if (INVALID_SOCKET_FD == s_messageListenerSocket)
```
wilbaker: Need to think about thread safety here, `WriteJsonToMessageListener` and `CreatePipeToMessageListener` could be called on separate threads (and there could be multiple calls to `WriteJsonToMessageListener` at the same time on different threads).

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -221,12 +239,141 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+            "\"CacheCapacity\":\"" + to_string(healthData.cacheCapacity) + "\","
+            "\"CacheEntries\":\"" + to_string(healthData.cacheEntries) + "\","
+            "\"InvalidationCount\":\"" + to_string(healthData.invalidateEntireCacheCount) + "\","
+            "\"CacheLookups\":\"" + to_string(healthData.totalCacheLookups) + "\","
+            "\"LookupCollisions\":\"" + to_string(healthData.totalLookupCollisions) + "\","
+            "\"FindRootHits\":\"" + to_string(healthData.totalFindRootForVnodeHits) + "\","
+            "\"FindRootMisses\":\"" + to_string(healthData.totalFindRootForVnodeMisses) + "\","
+            "\"RefreshRoot\":\"" + to_string(healthData.totalRefreshRootForVnode) + "\","
+            "\"InvalidateRoot\":\"" + to_string(healthData.totalInvalidateVnodeRoot) + "\","
+        "}");
```
jeffhostetler: I trust all of the above values are integers, so we don’t need to worry about escape/quoting them.

jeffhostetler: If they are integers, you can omit the dquotes as in “key”:123

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeffhostetler: Alternatively, you could have a kext provider name, but I’d have to add another translation entry to the service.  No big deal either way.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -221,12 +239,141 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
             healthData.totalFindRootForVnodeMisses,
             healthData.totalRefreshRootForVnode,
             healthData.totalInvalidateVnodeRoot);
+        
+        WriteJsonToMessageListener(
```
jeffhostetler: It’s hard to see the usage (here on my iPad) but I wonder if you want to pass the event name rather than have the message field be the determinant.  That is, pass eventName as “kext.health” for example. This would arrive to kusto as “vfs.kext.health”.  Then kusto queries would see it as a top-level field.

But I’m not sure how many message types you have. 

I thought I saw an error message too.  Maybe have it be eventName=“kext.error” and let it have a “message” and “ret” fields.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,6 +221,7 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
     IOReturn ret = IOConnectCallStructMethod(connection, LogSelector_FetchVnodeCacheHealth, nullptr, 0, &healthData, &out_size);
     if (ret == kIOReturnUnsupported)
     {
+        WriteJsonToMessageListener("{\"ErrorMessage\":\"IOConnectCallStructMethod failed for LogSelector_FetchVnodeCacheHealth, kIOReturnUnsupported \"}");
```
jeffhostetler: Did you want this to be a 2 field message? Later you have an error message with “ErrorMessage” and “ret” fields.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeffhostetler: I’m constrained on git.exe because of core Git on where I can get the value, but the kext and mount processes could share or not share config data as you want.  And yeah, requiring the mount daemon spawn a child process to get a config value seems expensive.

The git-guid and vfs-guid pipes are different.  The former summarizes all events for the client whereas the latter is just a proxy server. And the have a different protocol, so we can’t blend them. So I need to keep git.exe setting separate anyway.

jeffhostetler:
Hope this helps.  Ping me here if you have any questions.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: @wilbaker approach makes sense to me!  Yes probably some locking around connections and message sending.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: > Probably better to have a new provider entry, even if they route to the same AI endpoint.
The usermode and kernel mode pipe message schemas can diverge overtime then.

I think this sounds like something we could do in a follow up work item/PR?  @jeffhostetler would you agree?

jeffhostetler:
Yes this could wait for a later PR.  At that time we can ask if you need anything different in terms of how/when I proxy the events.
Such as monitoring the kext’s health or something that would differ from the plain proxying I’m doing for vfs events.

Jeff



Jeff

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: >I think this sounds like something we could do in a follow up work item/PR? @jeffhostetler would you agree?

The downside to delaying this work is that different versions of PrjFSKextLogDaemon will write the data in different formats, making it more difficult to consume.  Maybe we should get all of this sorted out now?

\+ @jrbriggs for his thoughts

Note to self:  Also need to verify that PrjFSKextLogDaemon is being versioned properly.

jeffhostetler: I’ll be back in the office on Monday and we can talk about stuff then if you want.
We can address the dedicated named pipe/socket, custom proxy, the format 
of the messages, and the changes to the provider and event name that we’ve 
talked about.  (On my iPad here and not sure if I’ve left anything out of this list or not.)

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: From what I can tell these are gone, but when you're back in the office could you confirm I read the code correctly? 😄

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: I just discussed this offline with @jrbriggs.  I won't hold this PR up on the provider name decision, and @jeffhostetler when you're back in we can decide if we want to update it before the next release.

We were also discussing that leaving the provider name as-is with "vfs.kext.*" events is probably OK too.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Per offline discussion with @jrbriggs we'll leave `s_messageListenerSocketPath` hardcoded for now.

ProjFS.Mac/PrjFSKextLogDaemon/JsonWriter.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <string>
+
+class JsonWriter
```
wilbaker: Add unit tests for this class

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
```diff
@@ -641,6 +652,84 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		26786AED228CD0F800F53311 /* ShellScript */ = {
```
wilbaker: @pmj @nickgra I didn't like updating each target to have call `GeneratePrjFSVersionHeader.sh` if the header is missing, but I couldn't find a better way of making sure that building any of the targets in Xcode would succeed if the version header was missing.

Any suggestions?

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -13,6 +18,9 @@ COVERAGEDIR=$ROOTDIR/BuildOutput/ProjFS.Mac/Coverage
 
 PROJFS=$SRCDIR/ProjFS.Mac
 
+echo "Generating PrjFSVersion.h as $VERSION..."
+$SCRIPTDIR/GeneratePrjFSVersionHeader.sh $VERSION || exit 1
```
wilbaker: @jrbriggs FYI I made a change here so that `ProjFS.Mac/Scripts/Build.sh` will generate a header file that contains the VFS4G/PrjFS version.

I also updated `PrjFSKextLogDaemon` and the kext/user-mode to use this version string rather than a hardcoded version.

Let me know if you see any issues with this approach.  I will be filing a follow up issue to also generate an xcconfig file so that the correct version can be stamped in the plist file.

jrbriggs: @wilbaker This approach makes sense to me, thanks!

ProjFS.Mac/PrjFSLib/Json/JsonWriter.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <string>
+
+class JsonWriter
```
wilbaker: @pmj @jeschu1 let me know your thoughts on where this class lives, and how I've done the unit tests.

With this PR things are setup as follows:

- There is a new `PrjFSLibTests` that tests the `PrjFSLib` dylib
- `JsonWriter.cpp`, though not required by `PrjFSLib`, is built as part of `PrjFSLib` (let me know what you think about this).
- I saw we're doing some weird things as far as directly compiling `.cpp` files from `PrjFSLib` in other projects (for example `PrjFSKextLogDaemon` builds `PrjFSUser.cpp` and now `JsonWriter.cpp` out of the `PrjFSLib` folder).  I did not change this as part of this PR.

Right now if there are no strong objections I'm leaning toward going with the current approach, and potentially having a follow up PR that cleans up `PrjFSLib` and how its files and binaries are being consumed.

Please let me know what you think and if you have any suggestions for cleaning this up.

ProjFS.Mac/PrjFSLibTests/JsonWriterTests.mm
```diff
@@ -0,0 +1,209 @@
...
+#include <utility>
+#include <vector>
+#import <XCTest/XCTest.h>
+
+using std::make_pair;
+using std::pair;
+using std::string;
+using std::vector;
+
+@interface JsonWriterTests : XCTestCase
```
wilbaker: @jeschu1 - FYI these tests get us 100% coverage of `JsonWriter`

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
pmj: You could create an otherwise empty target (File -> New -> Target… -> "Aggregate") and put your shell script in there. Then mark this new target as a dependency for all other relevant targets.

pmj: Additionally, I'd list the generated header file under output paths. I'm not entirely clear on where the version number is coming from, but if it's at all related to the contents of some file(s), then such files should be listed under input files so the shell script is only run when those files change.

ProjFS.Mac/PrjFSLib/Json/JsonWriter.hpp
pmj: > * I saw we're doing some weird things as far as directly compiling `.cpp` files from `PrjFSLib` in other projects (for example `PrjFSKextLogDaemon` builds `PrjFSUser.cpp` and now `JsonWriter.cpp` out of the `PrjFSLib` folder).  I did not change this as part of this PR.

Yeah, I think if we're going to have multiple such shared utility compilation units, we should probably move them to their own directory, where the various targets can include them in their builds or not as needed. Then for unit testing them we can just have a test target that builds all of (and only) the shared compilation units.

ProjFS.Mac/PrjFSKextLogDaemon/JsonWriter.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+    if ("{" != this->jsonBuffer)
+    {
+        this->jsonBuffer += ",";
+    }
+}
+
+void JsonWriter::AddKey(const string& key)
+{
+    this->jsonBuffer += "\"" + key + "\":";
```
pmj: Shouldn't this use the same escaping rules as AddStringValue?

pmj:
(FWIW, I'll try to do a comprehensive review of this either Tuesday or Wednesday - so far I've just picked out a few things in response to William's own comments.)

ProjFS.Mac/PrjFSKextLogDaemon/JsonWriter.cpp
wilbaker: Great suggestion, thanks!

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
wilbaker: > I'm not entirely clear on where the version number is coming from, but if it's at all related to the contents of some file(s), then such files should be listed under input files so the shell script is only run when those files change.

The version number comes from the Azure Pipelines build.  When it calls the build script it passes in the version:

```
Scripts/Mac/BuildGVFSForMac.sh $(BuildConfiguration) $(MacMajorAndMinorVersion).$(GVFSRevision)
```

And when a developer is building VFS4G on their machine by default we want to use "0.2.173.2" as the version.

I'm glad you brought this up because I noticed the "Mac XCode build kext" step of the release build is *not* using the build script.  It's calling XCode directly, and so that will need to be updated. 

@jrbriggs @nickgra do you know why that release build definition is not using `ProjFS.Mac/Scripts/Build.sh` to build the kext?

ProjFS.Mac/PrjFSLib/Json/JsonWriter.hpp
wilbaker: Filed #1182 for cleaning this up

ProjFS.Mac/PrjFS.xcodeproj/project.pbxproj
wilbaker: > Additionally, I'd list the generated header file under output paths. I'm not entirely clear on where the version number is coming from, but if it's at all related to the contents of some file(s), then such files should be listed under input files so the shell script is only run when those files change.

I added `${BUILD_ROOT}/../../PrjFSVersion.h` as an input path to the "Run Script" step (in a new aggregate target).

Listing the same path as both input and output caused some circular dependency issues.

The script in XCode only needs to be run if `${BUILD_ROOT}/../../PrjFSVersion.h` is missing, and so picking "input" over "output" seemed like the better choice, but let me know if you disagree.

wilbaker:
Next push will squat commits and address conflicts.

Old state of the branch can be found here:

https://github.com/wilbaker/VFSForGit/tree/kext_daemon_telemetry_preMay20rebase

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -1,21 +1,52 @@
...
 static const char PrjFSKextLogDaemon_OSLogSubsystem[] = "org.vfsforgit.prjfs.PrjFSKextLogDaemon";
+static const int INVALID_SOCKET_FD = -1;
 
 static os_log_t s_daemonLogger, s_kextLogger;
 static IONotificationPortRef s_notificationPort;
 
-static void StartLoggingKextMessages(io_connect_t connection, io_service_t service, os_log_t daemonLogger, os_log_t kextLogger);
+static int s_messageListenerSocket = INVALID_SOCKET_FD;
+static const string MessageListenerSocketPath = "/usr/local/GitService/pipe/vfs-c780ac06-135a-4e9e-ab6c-d41e2d265baa";
+static const string InfoMessageEventName = "info";
```
jeschu1: Is there a nice way to group the Info/Error/Heath together in C++?

It looks like from C++ 17 the below is valid:
struct EventName
{
    inline static const string Info= "info";
    inline static const string Error= "error";
    etc...
}

What we have certainly isn't a killer but it would be great if a more specific grouping is available.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -145,7 +145,11 @@ void DisableCRLFTranslationOnStdPipes()
 
 bool WriteToPipe(PIPE_HANDLE pipe, const char* message, size_t messageLength, /* out */ size_t* bytesWritten, /* out */ int* error)
 {
-    *bytesWritten = write(pipe, message, messageLength);
+    do
+    {
+        *bytesWritten = write(pipe, message, messageLength);
+    } while (*bytesWritten == -1 && errno == EINTR);
```
jeschu1: Curious about this change?  Is this to block while the pipe is down?  Could this stop a process from completing?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,29 +256,169 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+
+    JsonWriter messageWriter;
+    messageWriter.Add(MessageKey, message);
+    WriteJsonToMessageListener(ErrorMessageEventName, messageWriter);
+}
+
+static void LogKextMessage(os_log_type_t messageLogType, const char* message, int messageLength)
+{
+    os_log_with_type(s_kextLogger, messageLogType, "%{public}.*s", messageLength, message);
+    switch (messageLogType)
```
jeschu1: nit: would this make more sense as an if statement?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,29 +256,169 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+    }
+    
+    struct sockaddr_un socket_address;
+    memset(&socket_address, 0, sizeof(struct sockaddr_un));
+    
+    socket_address.sun_family = AF_UNIX;
+    size_t pathLength = MessageListenerSocketPath.length();
+    if (pathLength + 1 >= sizeof(socket_address.sun_path))
+    {
+        os_log(s_daemonLogger, "CreatePipeToMessageListener: Failed to copy socket path, insufficient buffer. Buffer size: %lu", sizeof(socket_address.sun_path));
```
jeffhostetler: This may be a nit, but some places in this commit you have %lu and some places have %llu.  Perhaps do a pass to double check these.  Thanks!

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,29 +256,169 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+        goto ClosePipeAndCleanup;
+    }
+    
+    strlcpy(socket_address.sun_path, MessageListenerSocketPath.c_str(), sizeof(socket_address.sun_path));
+    
+    if(0 == connect(s_messageListenerSocket, (struct sockaddr *) &socket_address, sizeof(struct sockaddr_un)))
+    {
+        os_log(s_daemonLogger, "CreatePipeToMessageListener: Connected to message listener");
```
jeffhostetler: another nit.  maybe include the name of the socket path??

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,29 +256,169 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+        return;
+    }
+    
+    error = errno;
+    os_log(s_daemonLogger, "CreatePipeToMessageListener: Failed to connect socket, error: %d", error);
```
jeffhostetler: and also here??

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,29 +256,169 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+    
+    string fullMessage = fullMessageWriter.ToString() + "\n";
+    size_t bytesWritten;
+    do
+    {
+        bytesWritten = write(s_messageListenerSocket, fullMessage.c_str(), fullMessage.length());
+    } while (bytesWritten == -1 && errno == EINTR);
+    
+    int error = errno;
+    if (bytesWritten != fullMessage.length())
```
jeffhostetler: The mac version of GitService runs in byte/stream mode and not in dgram/message mode, so we can do the usual write-loop logic for short writes, rather than aborting if you want.

ProjFS.Mac/PrjFSLibTests/JsonWriterTests.mm
jeschu1: @wilbaker big hi-fives on the unit tests.  You definitely cover some great cases there.

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -81,6 +89,38 @@ while read line; do
 	 [[ $line != *"RetainIOCount"* ]] && 
 	 [[ $line != *"ProviderMessaging_"* ]] && 
 	 [[ $line != *"RWLock_DropExclusiveToShared"* ]] && 
+	 [[ $line != *"PrjFS_"* ]] &&                                     #SHOULD ADD COVERAGE
```
jeschu1: There are a lot of exclusions now.  Can we break it up a bit with a comment
##### PrjjFsLib Excusions ######
Same for PrjFsKext

ProjFS.Mac/PrjFSLibTests/JsonWriterTests.mm
```diff
@@ -0,0 +1,225 @@
...
+- (void) testAddInt32 {
+    string expectedResult = "{\"testKey\":32}";
+    JsonWriter writer;
+    writer.Add("testKey", 32);
+    string jsonResult = writer.ToString();
+    XCTAssertTrue(
+        jsonResult == expectedResult,
+        "%s",
+        ("Expected result: " + expectedResult + " Result: " + jsonResult).c_str());
+}
```
jeffhostetler: Should we add a test case with a negative uint32 values ?

ProjFS.Mac/PrjFSLibTests/JsonWriterTests.mm
```diff
@@ -0,0 +1,225 @@
...
+    writer.Add("testKeyLine1\r\nTestKeyList2", "testValue");
+    writer.Add("\f\t key", 32);
+    string jsonResult = writer.ToString();
+    XCTAssertTrue(
+        jsonResult == expectedResult,
+        "%s",
+        ("Expected result: " + expectedResult + " Result: " + jsonResult).c_str());
+}
+
+- (void) testNestedJson {
```
jeffhostetler: maybe add tests where the nested sub-json structure is empty.

jeffhostetler: {"a":123,"b":{},"c":"xyz"} for example.

jeffhostetler: I suggest this because the "add comma if needed" logic is a bit tricky around empty sub objects.

jeffhostetler: Looks like you're not building multiple levels in the same structure, but rather using multiple variable instances, so this might not be a problem you.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Closing this thread out, based on our offline discussion we'll leave the provider name as "vfs.kext.*"

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Sure we can group these

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
wilbaker: As best I could tell, we weren't properly handling `write` failures due to interrupts:

```
     [EINTR]            A signal interrupts the write before it could be completed.
```

See also: https://www.gnu.org/software/libc/manual/html_node/Interrupted-Primitives.html#Interrupted-Primitives

wilbaker: As @jeffhostetler mentioned in a different comment:

> The mac version of GitService runs in byte/stream mode and not in dgram/message mode, so we can do the usual write-loop logic for short writes, rather than aborting if you want.

We are using stream mode here as well and shouldn't be aborting on short writes here either.

ProjFS.Mac/PrjFSLibTests/JsonWriterTests.mm
wilbaker: I will still add a test for this, thanks!

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: > You might check, but IIRC I got ride of the event level and opcode fields.

@jeffhostetler if my manual testing looks good I will assume that I read the service daemon code correctly and don't need to set these fields.  If it ends up that I do still need to set these please let me know 😃

jeffhostetler: yeah, the event level and opcode are gone.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Thanks! I'll switch the `size_t` format strings to use `%zu`

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: @jeschu1 let me know what you think of the `enum` I added here to get some grouping of values.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -36,7 +36,7 @@ PATH_STRING GetGVFSPipeName(const char *appName)
     if (enlistmentRootLength + 2 > sizeof(enlistmentRoot))
     {
         die(ReturnCode::PipeConnectError,
-            "Could not copy finalRootPath: %ls, insufficient buffer. enlistmentRootLength: %d, sizeof(enlistmentRoot): %d\n",
+            "Could not copy finalRootPath: %ls, insufficient buffer. enlistmentRootLength: %zu, sizeof(enlistmentRoot): %zu\n",
```
wilbaker: `%ls` should be `%s` here

wilbaker:
@pmj @jeschu1 please let me know if you have any feedback on these latest changes.  I'm aiming to merge these changes in by EOD tomorrow (May 23rd).

ProjFS.Mac/Scripts/Build.sh
```diff
@@ -5,6 +5,11 @@ if [ -z $CONFIGURATION ]; then
   CONFIGURATION=Debug
 fi
 
+VERSION=$2
+if [ -z $VERSION ]; then
+  VERSION="0.2.173.2"
```
pmj: Do we expect to update this manually on occasion?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -204,29 +265,222 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
...
+        case MessageType::Health:
+            return "health";
+    }
+    
+    return "invalid";
+}
+
+static void WriteJsonToMessageListener(MessageType messageType, const JsonWriter& jsonMessage)
+{
+    lock_guard<mutex> lock(s_messageListenerMutex);
```
pmj: Putting the `write()` inside a held lock could cause everything to get backed up - if reads stall on the other side of the socket for some reason (I/O…) that will fill up the socket's internal buffer, which will block the `write()` here, which means the mutex sits in the held state.

I don't think there's a great catch-all solution to this, just something to be aware of.

ProjFS.Mac/Scripts/Build.sh
wilbaker: "0.2.173.2" is a special version number that was picked years ago that means that VFS4G was built on a developers machine.  There are no plans to ever update it.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Yep that is definitely a risk, however, as best I could tell a backup here should not cause a backup in the kext itself (although messages might get dropped).

If we find that we do encounter problems with the reader getting stuck one option is moving all writes to a new background thread (which will help ensure we don't block reading messages from the kext and writing them to the macOS log), but I will leave that for a follow up PR if we end up needing it.

GVFS/GVFS.NativeHooks.Common/common.posix.cpp
```diff
@@ -36,7 +36,7 @@ PATH_STRING GetGVFSPipeName(const char *appName)
     if (enlistmentRootLength + 2 > sizeof(enlistmentRoot))
     {
         die(ReturnCode::PipeConnectError,
-            "Could not copy finalRootPath: %ls, insufficient buffer. enlistmentRootLength: %d, sizeof(enlistmentRoot): %d\n",
+            "Could not copy finalRootPath: %s, insufficient buffer. enlistmentRootLength: %zu, sizeof(enlistmentRoot): %zu\n",
```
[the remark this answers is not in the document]
wilbaker: FYI - Getting this to work correctly on VS 2017 required setting the `/Zc:__cplusplus` option.  More details here:

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

PR #1103, opened 2019-05-06
Enable Rename tests on Mac and add additional tests for coverage

jeschu1:
We want to test the specific cases of
- rename within a virtualization root
- file moved outside a virtualization root to in
- file moved inside a virtualization root to out

The events sent by the OS are
- Predelete on the From location
- Rename with a from and to location

Since we receive the predelete, no work is necessary by the rename for the 'from' location.
**With this is mind we do not want to populate the 'from' portion for renames in the kext.  We do need to populate it for hard links.**

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Categories.cs
+++ b/GVFS/GVFS.FunctionalTests/Categories.cs
@@ -15,11 +15,8 @@
             // machines but not on the build agents
             public const string FailsOnBuildAgent = "FailsOnBuildAgent";
 
-            // Tests that require #356 (old paths to be delivered with rename notifications)
-            public const string NeedsRenameOldPath = "NeedsRenameOldPath";
-
-            // Git related tests that are not yet passing on Mac
-            public const string M3 = "M3_AllGitCommands";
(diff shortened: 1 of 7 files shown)


jeschu1:
@pmj FYI as this affects #985
More details in the description, but the crux is since we get a predelete before the rename, there isn't work for the rename to do on the `from` path.   

We *do not* want to send both from/to for rename scenarios:
Reason 1: Let's avoid unnecessary extra messages
Reason 2: It actually causes messages to unnecessarily be put into ModifiedFiles.dat
- Predelete will remove a message from newlyCreatedFileAndFolderPaths
- Rename with from would then see newlyCreatedFileAndFolderPaths does not have the file and add it unnessarily to modifiedFiles.dat [here](https://github.com/Microsoft/VFSForGit/blob/d554f3c83a7b18d8f37865b89c60cfaeea7eb5df/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs#L588).

We do still need both from/to for hardlinks.  Sorry I just found this now!

PR #1104, opened 2019-05-06
Update libgit2 native binaries and include .so for Linux

chrisd8088:
Upgrade the LibGit2Sharp.NativeBinaries NuGet package from v1.0.165 (containing LibGit2Sharp v0.23.1) to [v2.0.278](https://www.nuget.org/packages/LibGit2Sharp.NativeBinaries/2.0.278), which contains the latest [LibGit2Sharp release v0.26](https://github.com/libgit2/libgit2sharp/releases/tag/v0.26).

LibGit2Sharp version 0.26 eliminates most external library dependencies other than OpenSSL.  In particular, by removing the dependency on libcurl, there is no longer a link-time requirement for a version of libcurl with deprecated `CURL_OPENSSL_3` symbols and SSLv3 support.

This allows us to include the compiled `libgit2-*.so` from the LibGit2Sharp.NativeBinaries NuGet package when building on Linux; otherwise, builds on Linux are unsuccessful or else require old and insecure libcurl installations, so this is a prerequisite for a Linux platform build (along with PRs #1119 and #1123).

/cc @jrbriggs, @kivikakk

### DIFF
--- a/GVFS/FastFetch/FastFetch.csproj
+++ b/GVFS/FastFetch/FastFetch.csproj
@@ -47,7 +47,7 @@
   
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.1.1-beta" />
-    <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="1.0.165" />
+    <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="2.0.278" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.0.2">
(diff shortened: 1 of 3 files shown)


wilbaker:
>I assume that if the Mac functional tests pass, then this is an acceptable update to the libraries.

That's my assumption as well.

chrisd8088:
Thanks, @derrickstolee and @wilbaker!  Will merge shortly -- and will include the full PR description in the merge commit this time!  :-)

PR #1107, opened 2019-05-06
Refactor for easier conversion to sqlite

kewillford:
This is only a refactoring to prepare for using SQLite and shouldn't contain any functional changes.  Started putting the code in a Database directory of GVFS.Common

It creates a couple of interfaces.  One for the placeholder data and one for the placeholder database.

Renamed some of the methods to be a little more generic like removing `AndFlush` and `AndPrepToWriteAllEntries` since SQLite will not have these.
  
Also replaced a `Dictionary<string, PlaceholderListDatabase.PlaceholderData>` with a `HashSet<string>` since it was never using the value of the dictionary.

### DIFF
--- a/GVFS/GVFS.Common/Database/IPlaceholderData.cs
+++ b/GVFS/GVFS.Common/Database/IPlaceholderData.cs
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace GVFS.Common
+﻿namespace GVFS.Common.Database
 {
     public interface IPlaceholderData
     {
--- a/GVFS/GVFS.Common/PlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/PlaceholderListDatabase.cs
@@ -1,4 +1,5 @@
-﻿using GVFS.Common.FileSystem;
+﻿using GVFS.Common.Database;
+using GVFS.Common.FileSystem;
 using GVFS.Common.Tracing;
 using System;
 using System.Collections.Generic;
@@ -6,7 +7,7 @@ using System.IO;
 
 namespace GVFS.Common
 {
-    public class PlaceholderListDatabase : FileBasedCollection
+    public class PlaceholderListDatabase : FileBasedCollection, IPlaceholderDatabase
     {
         // Special folder values must:
         // - Be 40 characters long
@@ -49,7 +50,7 @@ namespace GVFS.Common
         /// The EstimatedCount is "estimated" because it's simply (# adds - # deletes).  There is nothing to prevent
         /// multiple adds or deletes of the same path from being double counted
         /// </summary>
-        public int EstimatedCount { get; private set; }
+        public int Count { get; private set; }
 
         public static bool TryCreate(ITracer tracer, string dataFilePath, PhysicalFileSystem fileSystem, out PlaceholderListDatabase output, out string error)
         {
@@ -59,7 +60,7 @@ namespace GVFS.Common
             if (!temp.TryLoadFromDisk<string, string>(
                 temp.TryParseAddLine,
                 temp.TryParseRemoveLine,
-                (key, value) => temp.EstimatedCount++,
+                (key, value) => temp.Count++,
                 out error))
             {
                 temp = null;
@@ -72,22 +73,27 @@ namespace GVFS.Common
             return true;
         }
 
-        public void AddAndFlushFile(string path, string sha)
+        public void AddFile(string path, string sha)
         {
             this.AddAndFlush(path, sha);
         }
 
-        public void AddAndFlushFolder(string path, bool isExpanded)
+        public void AddPartialFolder(string path)
         {
-            this.AddAndFlush(path, isExpanded ? ExpandedFolderValue : PartialFolderValue);
+            this.AddAndFlush(path, PartialFolderValue);
         }
 
-        public void AddAndFlushPossibleTombstoneFolder(string path)
+        public void AddExpandedFolder(string path)
+        {
+            this.AddAndFlush(path, ExpandedFolderValue);
+        }
+
+        public void AddPossibleTombstoneFolder(string path)
         {
             this.AddAndFlush(path, PossibleTombstoneFolderValue);
         }
 
-        public void RemoveAndFlush(string path)
+        public void Remove(string path)
         {
             try
             {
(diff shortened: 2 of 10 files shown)


GVFS/GVFS.Common/Database/IPlaceholderData.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace GVFS.Common.Database
+{
+    public interface IPlaceholderData
+    {
+        string Path { get; }
+        string Sha { get; }
+        bool IsFolder { get; }
```
kewillford: Rename to `IsPartialFolder`?

derrickstolee: Are all three of these `bool`s independent? Should they be an `enum` instead?

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -49,7 +50,7 @@ private PlaceholderListDatabase(ITracer tracer, PhysicalFileSystem fileSystem, s
         /// The EstimatedCount is "estimated" because it's simply (# adds - # deletes).  There is nothing to prevent
         /// multiple adds or deletes of the same path from being double counted
         /// </summary>
-        public int EstimatedCount { get; private set; }
+        public int Count { get; private set; }
```
wilbaker: When we move to SQLite will this count be more accurate?

GVFS/GVFS.Common/Database/IPlaceholderData.cs
kewillford: Looking at this closer and `IsFolder` should stay because it really is if the placeholder is any type of folder and not just a partial folder.
```
public bool IsFolder
{
	get
	{
		return IsShaAFolder(this.Sha);
	}
}


public static bool IsShaAFolder(string shaValue)
{
	return shaValue.Equals(PartialFolderValue, StringComparison.Ordinal) ||
		shaValue.Equals(ExpandedFolderValue, StringComparison.Ordinal) ||
		shaValue.Equals(PossibleTombstoneFolderValue, StringComparison.Ordinal);
}
```

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -210,12 +216,12 @@ public void GetAllEntriesAndPrepToWriteAllEntries(out List<PlaceholderData> file
...
         }
 
-        public Dictionary<string, PlaceholderListDatabase.PlaceholderData> GetAllFileEntries()
+        public HashSet<string> GetAllFilePaths()
         {
             try
             {
-                Dictionary<string, PlaceholderListDatabase.PlaceholderData> filePlaceholdersFromDiskByPath =
-                    new Dictionary<string, PlaceholderListDatabase.PlaceholderData>(Math.Max(1, this.EstimatedCount), StringComparer.Ordinal);
+                HashSet<string> filePlaceholdersFromDiskByPath =
```
wilbaker: Minor: I think the "ByPath" part of the `filePlaceholdersFromDiskByPath` name was because this was a dictionary.  We could probably rename this to something simpler like `filePlaceholderPaths`

GVFS/GVFS.Common/PlaceholderListDatabase.cs
kewillford: Yes because it will be querying the database with 
```
SELECT COUNT(path) FROM Placeholders
```

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -165,12 +171,12 @@ public List<PlaceholderData> GetAllEntriesAndPrepToWriteAllEntries()
         ///     - If WriteAllEntriesAndFlush is *not* called entries that were added to the PlaceholderListDatabase after
         ///       calling GetAllEntriesAndPrepToWriteAllEntries will be lost
```
wilbaker: Update method names in the comments (I see `GetAllEntriesAndPrepToWriteAllEntries` here, I have not checked if the other renamed methods were mentioned in comments)

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -120,11 +126,11 @@ public void RemoveAndFlush(string path)
         ///     - If WriteAllEntriesAndFlush is *not* called entries that were added to the PlaceholderListDatabase after
         ///       calling GetAllEntriesAndPrepToWriteAllEntries will be lost
         /// </remarks>
-        public List<PlaceholderData> GetAllEntriesAndPrepToWriteAllEntries()
+        public List<IPlaceholderData> GetAllEntriesAndPrepToWriteAllEntries()
```
wilbaker: Should `GetAllEntriesAndPrepToWriteAllEntries()` be renamed as well (like  `GetAllEntriesAndPrepToWriteAllEntries(out List<PlaceholderData> filePlaceholders, out List<PlaceholderData> folderPlaceholders)` was)?

kewillford: That one I left because it is only a part of the `PlaceholderListDatabase` and not the interface because SQLite didn't necessarily _need_ it since there was already a `GetAllEntries` and  it is only used in tests and in the DiskLayoutUpgrades.  But if you feel differently that it should be part of the interface and renamed let me know.

wilbaker: > But if you feel differently that it should be part of the interface

I agree there is no need to add it to the interface (and it would be better not to add it).

> and renamed let me know.

I do think it would be nice to keep the name of this method and the old `GetAllEntriesAndPrepToWriteAllEntries` method in sync as the idea is they are supposed to do the exact same thing (other than returning the data in one collection or two).

I suppose the other option is we delete this method entirely and have everyone use the version that returns separate file and folder placeholders (and the caller can combine them if needed).  Thoughts on that approach?  

Based on a quick glance over how the upgraders are using the placeholder list it doesn't look like it would be too bad to switch them to the method in the interface.

One other question (that is slightly outside the scope of this PR).  Will this:

```
///     - If WriteAllEntriesAndFlush is *not* called entries that were added to the PlaceholderListDatabase after
///       calling GetAllEntriesAndPrepToWriteAllEntries will be lost	        ///       calling GetAllEntriesAndPrepToWriteAllEntries will be lost
```

Still be an issue after moving to SQLite?

kewillford: For this PR I will rename it to keep it in sync and look at deleting in another PR just to reduce the risk and functional changes.

The entries will not be "lost" because they will be added to the database but they will not be in the list of entries returned by `GetAllEntries` so I will need to check those code paths to make sure we don't need those added entries where we called `GetAllEntries`.

PR #1102, opened 2019-05-06
Rename LocalStorageRoot

wilbaker:
Follow up to #1074 (see comment here https://github.com/Microsoft/VFSForGit/pull/1074#issuecomment-489677734)

This PR makes a few tweaks to the changes in #1074:

- Renamed `Enlistment.LocalStorageRoot` to `WorkingDirectoryBackingRoot`
- Added comments to `Enlistment.WorkingDirectoryBackingRoot`
- Renamed `GVFSPlatform.LocalStorageRoot` to `WorkingDirectoryBackingRootName` to make it clearer that this property is not a path.

Please let me know if you can think of a better name than `WorkingDirectoryBackingRoot`, the goal of this PR is to replace `LocalStorageRoot` with a more descriptive name.

PR #1119, opened 2019-05-08
Rename WorkingDirectoryBackingRootPath constants

chrisd8088:
This PR primarily is a followup to #1102 to accommodate the fact that on Linux, the relative location of the working directory backing store will be a short path (i.e., `.gvfs/lower`) and not a single filename like `src`.  Hence we rename the `WorkingDirectoryBackingRootName` constants to `WorkingDirectoryBackingRootPath`, akin to how other short relative path constants are constructed in [`GVFS.Common.GVFSConstants.DotGVFS`](https://github.com/microsoft/VFSForGit/blob/7cc1786fcb2ca5d9be9ce7a40158eec8dacf1f4d/GVFS/GVFS.Common/GVFSConstants.cs#L107).

We also fix up the internal name of the `workingDirectoryBackingRoot` argument to the `GVFS.Common.Enlistment` class; it was possibly missed in #1102 because of the lowercase `l` in `localStorageRoot`.

And we add a `TODO` comment regarding the `git.cmd` script, which is presumably unused on non-Windows platforms and would be ignored even if renamed `git` as it's unlikely to be in the user's `PATH`.

These housekeeping changes are all in preparation, along with ~#1104~ (DONE), #1123, and #1128 (still WIP) for #1125, which adds the `GVFS.Platform.Linux` classes and related build tooling.

/cc @jrbriggs, @kivikakk

### DIFF
--- a/GVFS/GVFS/CommandLine/CloneVerb.cs
+++ b/GVFS/GVFS/CommandLine/CloneVerb.cs
@@ -662,6 +662,7 @@ namespace GVFS.CommandLine
             return new Result(true);
         }
 
+        // TODO(Linux), TODO(Mac): either adjust to "git" or remove entirely
         private void CreateGitScript(GVFSEnlistment enlistment)
         {
             FileInfo gitCmd = new FileInfo(Path.Combine(enlistment.EnlistmentRoot, "git.cmd"));
--- a/GVFS/GVFS.Common/Enlistment.cs
+++ b/GVFS/GVFS.Common/Enlistment.cs
@@ -10,7 +10,7 @@ namespace GVFS.Common
         protected Enlistment(
             string enlistmentRoot,
             string workingDirectoryRoot,
-            string localStorageRoot,
+            string workingDirectoryBackingRoot,
             string repoUrl,
             string gitBinPath,
             string gvfsHooksRoot,
@@ -24,7 +24,7 @@ namespace GVFS.Common
 
             this.EnlistmentRoot = enlistmentRoot;
             this.WorkingDirectoryRoot = workingDirectoryRoot;
-            this.WorkingDirectoryBackingRoot = localStorageRoot;
+            this.WorkingDirectoryBackingRoot = workingDirectoryBackingRoot;
             this.DotGitRoot = Path.Combine(this.WorkingDirectoryBackingRoot, GVFSConstants.DotGit.Root);
             this.GitBinPath = gitBinPath;
             this.GVFSHooksRoot = gvfsHooksRoot;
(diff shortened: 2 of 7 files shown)


GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -662,6 +662,7 @@ private void CheckNotInsideExistingRepo(string normalizedEnlistmentRootPath)
             return new Result(true);
         }
 
+        // TODO(Linux), TODO(Mac): either adjust to "git" or remove entirely
```
derrickstolee: Good call! An equivalent shell script could be:

```
#!/bin/sh

echo This repo was cloned using VFS for Git, and the git repo is in the 'src' directory
echo Switching you to the 'src' directory and rerunning your git command

cd src
git "$@"
```

derrickstolee: Of course, there are some special characters in the `git.cmd` that probably do color stuff, but I'm too lazy to figure out how to do that in POSIX.

chrisd8088: Well, and the other consideration is that where Windows will (I believe; I'm no expert) search the [current directory first](https://docs.microsoft.com/en-us/previous-versions//cc723564(v=technet.10)#command-search-sequence) for an executable, on Unix systems, unless someone happens to have put `.` in their `PATH` (which is generally frowned upon), that's unlikely.

So I'm not sure if there's really any value to this kind of "redirector" script on Mac or Linux, even if we named it `git`.  Plus I feel like it would be confusing for folks who know there needs to be a custom `git` binary to work with VFSForGit if we *also* have a script named the same thing, but which isn't actually that binary.

Anyway, I just wanted to flag it because it was something I bumped into along the road to `GVFS.Platform.Linux`.  :-)

derrickstolee: Good point! Then, it's worth leaving as a TODO until we have a concrete decision about what to do with this step. Apparently, it was an important feature for users confused about where to go after a `gvfs clone` step.

chrisd8088: I could also open an issue on this to track it for future work, if that's useful.

wilbaker: >I could also open an issue on this to track it for future work, if that's useful.

I think it's worthwhile to have an issue tracking this, could you file one?

chrisd8088: Done as #1138 -- thanks!

chrisd8088:
Thanks very much @wilbaker!

PR #1123, opened 2019-05-08
[POSIX, Mac] Refactor BSD-specific values out of POSIXFileSystem

chrisd8088:
Along with ~#1104~ (DONE), ~#1119~ (DONE), and #1128, this PR is a prerequisite to the addition of a `GVFS.Platform.Linux` set of classes in #1125.

In particular, we convert `POSIXFileSystem` into an abstract class and move the majority of its implementation into `MacFileSystem`, due to the fact that many of the syscall function signatures and flag values, plus the layout of the `struct stat` fields, differ between BSD/Darwin and Linux.  So we need to separate these out from the POSIX-level commonalities between the two.

The visibility of the members of the `NativeFileReader` subclass is also adjusted, and we remove some of its unused constants such as `Create` (for the `O_CREAT` flag, which isn't used here).

And we correct the return value of `getuid(2)` in `POSIXPlatform` to a 32-bit `uint`.  Note that on both Linux and BSD/Darwin, the `uid_t` `typedef` can be checked with the following:
```
gcc -E -xc -include 'sys/types.h' - </dev/null | grep [^u]uid_t | grep ^typedef
```

/cc @jrbriggs

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacFileSystem.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystem.cs
@@ -1,0 +1,157 @@
+﻿using GVFS.Common;
+using GVFS.Platform.POSIX;
+using System;
+using System.Runtime.InteropServices;
+
+namespace GVFS.Platform.Mac
+{
+    public class MacFileSystem : POSIXFileSystem
+    {
+        public override void ChangeMode(string path, ushort mode)
+        {
+            Chmod(path, mode);
+        }
+
+        public override bool HydrateFile(string fileName, byte[] buffer)
+        {
+            return NativeFileReader.TryReadFirstByteOfFile(fileName, buffer);
+        }
+
+        public override bool IsExecutable(string fileName)
+        {
+            NativeStat.StatBuffer statBuffer = this.StatFile(fileName);
+            return NativeStat.IsExecutable(statBuffer.Mode);
+        }
+
+        public override bool IsSocket(string fileName)
+        {
+            NativeStat.StatBuffer statBuffer = this.StatFile(fileName);
+            return NativeStat.IsSock(statBuffer.Mode);
+        }
+
+        [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
+        private static extern int Chmod(string pathname, ushort mode);
+
+        private NativeStat.StatBuffer StatFile(string fileName)
+        {
+            if (NativeStat.Stat(fileName, out NativeStat.StatBuffer statBuffer) != 0)
+            {
+                NativeMethods.ThrowLastWin32Exception($"Failed to stat {fileName}");
+            }
+
+            return statBuffer;
+        }
+
+        private static class NativeStat
+        {
+            // #define  S_IFMT      0170000     /* [XSI] type of file mask */
+            private static readonly ushort IFMT = Convert.ToUInt16("170000", 8);
+
+            // #define  S_IFSOCK    0140000     /* [XSI] socket */
+            private static readonly ushort IFSOCK = Convert.ToUInt16("0140000", 8);
+
+            // #define S_IXUSR     0000100     /* [XSI] X for owner */
+            private static readonly ushort IXUSR = Convert.ToUInt16("100", 8);
+
+            // #define S_IXGRP     0000010     /* [XSI] X for group */
+            private static readonly ushort IXGRP = Convert.ToUInt16("10", 8);
+
+            // #define S_IXOTH     0000001     /* [XSI] X for other */
+            private static readonly ushort IXOTH = Convert.ToUInt16("1", 8);
+
+            public static bool IsSock(ushort mode)
+            {
+                // #define  S_ISSOCK(m) (((m) & S_IFMT) == S_IFSOCK)    /* socket */
+                return (mode & IFMT) == IFSOCK;
+            }
+
+            public static bool IsExecutable(ushort mode)
+            {
+                return (mode & (IXUSR | IXGRP | IXOTH)) != 0;
+            }
+
+            [DllImport("libc", EntryPoint = "stat$INODE64", SetLastError = true)]
+            public static extern int Stat(string path, [Out] out StatBuffer statBuffer);
+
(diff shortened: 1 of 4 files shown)


chrisd8088:
Thanks @kivikakk -- and I think I managed to retain you as the primary author, since you did the work here!

GVFS/GVFS.Platform.POSIX/POSIXFileSystem.cs
```diff
@@ -65,138 +51,9 @@ public bool TryCreateDirectoryWithAdminAndUserModifyPermissions(string directory
...
-
-                [MarshalAs(UnmanagedType.ByValArray, SizeConst = 2)]
-                public long[] QSpare;     /* RESERVED: DO NOT USE! */
-            }
-        }
-
-        private class NativeFileReader
-        {
-            public const int ReadOnly = 0x0000;
-            public const int WriteOnly = 0x0001;
```
derrickstolee: I noticed that `WriteOnly` and `Create` were dropped, but they were not used. @wilbaker @nickgra do you know why they were there at all?

chrisd8088: As a bit of context, these are mappings of the libc API flags `O_WRONLY` and `O_CREAT`, as defined in, e.g., FreeBSD's [`fcntl.h`](https://github.com/freebsd/freebsd/blob/def45a363e51d353f063c8876ad00ddac259c019/sys/sys/fcntl.h#L76).

It looks to me as though there was some [refactoring of native filesystem methods](https://github.com/microsoft/VFSForGit/commit/dc13c08b942e908b481ccdcf1bfe3b3736a4bfe4#diff-a279408f959e444a7a8665ecae6e4dc7) back in December which resulted in the creation of [`FastFetch.NativeUnixMethods`](https://github.com/microsoft/VFSForGit/blob/29fd6a1746fabedea35ac920ef0a2f844aeacb3e/GVFS/FastFetch/NativeUnixMethods.cs#L9) as well as `GVFS.Platform.Mac.MacFileSystem`, which was then [migrated to `POSIXFileSystem`](https://github.com/microsoft/VFSForGit/commit/262772e61be3eb0513749fa49a25ed3186643f0d#diff-4d1bba707d97f523c1c9d2f1bc15dcd8) by @kivikakk recently so we could share the common POSIX bits with Linux.

My hunch would be that because some of these flags are used in the [`FastFetch.NativeUnixMethods`](https://github.com/microsoft/VFSForGit/blob/29fd6a1746fabedea35ac920ef0a2f844aeacb3e/GVFS/FastFetch/NativeUnixMethods.cs#L35) code, they were included here by accident.

It's a good question, because it points out that if we want to use `FastFetch` on Linux, we'll need to refactor `FastFetch.NativeUnixMethods`, which is currently going to work on BSD/Darwin only as these flags like `O_CREAT` have different values on Linux, and the fields of `struct stat` are also very different.  I've created [an issue](github/VFSForGit#17) in our backlog for that!

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -0,0 +1,157 @@
...
+using System;
+using System.Runtime.InteropServices;
+
+namespace GVFS.Platform.Mac
+{
+    public class MacFileSystem : POSIXFileSystem
+    {
+        public override void ChangeMode(string path, ushort mode)
+        {
+            Chmod(path, mode);
```
wilbaker: Why does this not work on Linux?  Is `mode_t` something other than `ushort`?

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -0,0 +1,157 @@
...
+                public int LSpare;         /* RESERVED: DO NOT USE! */
+
+                [MarshalAs(UnmanagedType.ByValArray, SizeConst = 2)]
+                public long[] QSpare;     /* RESERVED: DO NOT USE! */
+            }
+        }
+
+        private static class NativeFileReader
+        {
+            private const int ReadOnly = 0x0000;
```
wilbaker: Not your change, but would you mind putting a comment here:

```
// #define	O_RDONLY	0x0000		/* open for reading only */
private const int ReadOnly = 0x0000;
```

(This matches the style of comments in `NativeStat` and makes it easier to track down where these values come from)

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
chrisd8088: Yes, it's a `uint` -- so the `Chmod` DLL import wrapper is [different on Linux](https://github.com/github/VFSForGit/blob/6e457bc631f43d69681cb7693df0e36bdec3f7fb/GVFS/GVFS.Platform.Linux/LinuxFileSystem.cs#L33), and hence `ChangeMode()` ends up having to be in the OS-specific class rather than the base POSIX class, even though it looks [the same](https://github.com/github/VFSForGit/blob/6e457bc631f43d69681cb7693df0e36bdec3f7fb/GVFS/GVFS.Platform.Linux/LinuxFileSystem.cs#L12).  You can see these Linux variants in #1125.

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
chrisd8088: Sure thing!

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
chrisd8088: On Linux, it's a 32-bit `int` (assuming we're using fairly generic hardware and compiler, of course):
```
$ gcc -E -xc -include 'sys/types.h' - </dev/null | grep mode_t
typedef unsigned int __mode_t;
typedef __mode_t mode_t;
```

On Mac:
```
gcc -E -xc -include 'sys/types.h' - </dev/null | grep mode_t | grep ^typedef
typedef __uint16_t __darwin_mode_t;
typedef __darwin_mode_t mode_t;
```

PR #1125, opened 2019-05-08
[Linux] Add GVFS platform and build configuration for Linux

chrisd8088:
We add the core `GVFS.Platform.Linux` classes, derived from their `GVFS.Platform.Mac` equivalents but with appropriate changes for syscall argument signatures and flag values, type definitions, and structure fields (e.g., `mode_t` is a `uint`, and the layout of `struct stat` is quite different).

In order to avoid conflicting with the GNOME Virtual file system (GVfs), we use `.vfsforgit` as the name of our per-repository private directories, instead of `.gvfs`, as is used on Windows and Mac.

The build tooling is largely a duplicate of the corresponding Mac configuration, with some changes to ensure we build successfully on Linux (e.g., testing for `!= Windows` is no longer sufficient to determine that the platform is OS X, so we add distinct `IsLinux` and `IsOSX` build conditions to the `.cspro`j file for `GVFS.Upgrader`, among other subprojects).

The `GVFS.sln` configuration changes should ensure that the `ProjFS.Linux.Managed` and `GVFS.Platform.Linux` subprojects are built on all platforms (and their Mac and Windows equivalents
are built on Linux), and that the `GVFS.Service.Mac` classes are built on Linux, so as to allow the `GVFS.UnitTests` subproject to build and run successfully on all three platforms.

We also build the `GVFS.Hooks.Linux` and `GVFS.Mount.Linux` subprojects on both Linux and Windows, just as their Mac counterparts are built on both Mac and Windows.

The `GVFS.Service` binary is, for the time being, only a stub on Linux, as we have yet to determine how (or if) we should handle auto-mounting repositories.

Note that for now the `BuildGVFSForLinux.sh` script does not perform any installation or packaging.

* Resolves https://github.com/github/VFSForGit/issues/9.
* Resolves github/VFSForGit#15.
* Resolves https://github.com/github/VFSForGit/issues/20 and https://github.com/github/VFSForGit/issues/27.

/cc @jrbriggs, @wilbaker


Scripts/Linux/DownloadGVFSGit.sh
```diff
@@ -0,0 +1,7 @@
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
+
+BUILDDIR=$VFS_OUTPUTDIR/GVFS.Build
+GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
+cp $VFS_SRCDIR/nuget.config $BUILDDIR
+dotnet new classlib -n Restore.GitInstaller -o $BUILDDIR --force
+dotnet add $BUILDDIR/Restore.GitInstaller.csproj package --package-directory $VFS_PACKAGESDIR GitForMac.GVFS.Installer --version $GITVERSION
```
kivikakk: This will download the nupkg containing a dmg; we'll need to adjust this and/or make our own GitForLinux installer with the gvfs patches?

chrisd8088: Yes ... the required effort here (for Linux) remains murky to me.  The existing CI jobs already build a Linux version of the patched GitForWindows (from microsoft/git), which is nice.  But we'll need to review how we want to package, both for CI and for customers, VFSForGit and the patched GitForWindows.

GVFS/GVFS.PlatformLoader/PlatformLoader.Linux.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public static FileSystemVirtualizer CreateFileSystemVirtualizer(GVFSContext context, GVFSGitObjects gitObjects)
+        {
+            return new LinuxFileSystemVirtualizer(context, gitObjects);
+        }
+
+        public static void Initialize()
+        {
+            GVFSPlatform.Register(new LinuxPlatform());
+            return;
```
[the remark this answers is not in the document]
chrisd8088: This is just a copy-and-paste from Mac, and the Windows one has it too, so I'll pull it out of all of them.  Thanks!

kivikakk:
@chrisd8088 fyi, I've added two items to this PR:

* e7f64cf is just #1191 -- once that PR is merged, our next rebase will drop the commit from here.
* 30cd116 resolves a conflict that our latest rebase produces when building MirrorProvider. We don't build MirrorProvider in any CI which is why it doesn't show up, but libprojfs's CI _does_ build it (!) and turned the error up 😸

GVFS/GVFS.FunctionalTests/FileSystemRunners/SystemIORunner.cs
```diff
@@ -214,20 +214,27 @@ public override void ChangeMode(string path, ushort mode)
...
         }
 
         [DllImport("kernel32", SetLastError = true)]
         private static extern bool MoveFileEx(string existingFileName, string newFileName, int flags);
 
         [DllImport("libc", EntryPoint = "link", SetLastError = true)]
-        private static extern int MacCreateHardLink(string oldPath, string newPath);
+        private static extern int POSIXCreateHardLink(string oldPath, string newPath);
+
+        [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
```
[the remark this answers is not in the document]
chrisd8088: It's the next line: `uint mode` on Linux, but `ushort mode` on BSD/Darwin.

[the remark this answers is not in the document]
derrickstolee: Thanks! I must have misread that.

chrisd8088:
@wilbaker, would you be able to take a look at this for us?  We are, mostly, adding files to flesh out the Linux side of the GVFS provider.

I should note that I have some WIP which will reduce code duplication between `ProjFS.Linux` and `ProjFS.Mac`, and between `LinuxFileSystemVirtualizer` and `MacFileSystemVirtualizer`, but we'd like to get this PR landed into `features/linuxprototype` before it gets to be too large.

chrisd8088:
@wilbaker -- I realize there's a lot of fresh code here, since we're adding a whole new platform to the GVFS provider.  To that end we've tried to keep the sequence of commits as clean and logical as possible (hence all the rebasing), so you might find it easier to review by looking at the commit sequence, at least in part.

There are places where we have some code duplication between Mac and Linux, and that'll be the focus of another PR based on [`posix-projfs-virt-instance`](https://github.com/github/VFSForGit/tree/posix-projfs-virt-instance), which is still WIP.  But our hope is that this PR is relatively coherent now as just adding what's required to get the GVFS provider to build and run on Linux.

GVFS/GVFS.Platform.Linux/GVFS.Platform.Linux.csproj
```diff
@@ -0,0 +1,40 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\GVFS.Build\GVFS.cs.props" />
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp2.1;netstandard2.0</TargetFrameworks>
+    <Platforms>x64</Platforms>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
wilbaker: I didn't see any code marked as `unsafe` in `GVFS.Platform.Linux`, can `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` be removed?

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -0,0 +1,719 @@
...
+            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
+            this.virtualizationInstance.OnFilePreConvertToFull = this.NotifyFilePreConvertToFull;
+
+            uint threadCount = (uint)Environment.ProcessorCount * 2;
+
+            Result result = this.virtualizationInstance.StartVirtualizationInstance(
+                this.Context.Enlistment.WorkingDirectoryBackingRoot,
+                this.Context.Enlistment.WorkingDirectoryRoot,
+                threadCount,
+                this.IsUninitializedMount(this.Context.Enlistment.WorkingDirectoryBackingRoot));
```
wilbaker: Full disclosure: I'm still very new to the Linux architecture.

If `IsUninitializedMount` is false, should `StartVirtualizationInstance` be called at all?

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -0,0 +1,719 @@
...
+                relativeNewLinkPath: relativeNewLinkPath);
+        }
+
+        private Result OnEnumerateDirectory(
+            ulong commandId,
+            string relativePath,
+            int triggeringProcessId,
+            string triggeringProcessName)
+        {
+            relativePath = ConvertDotPath(relativePath);
```
wilbaker: Rather than making VFS4G call `ConvertDotPath` can your filesystem layer perform this conversion before calling `OnEnumerateDirectory`?

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -0,0 +1,719 @@
...
+        }
+
+        private Result OnPreDelete(string relativePath, bool isDirectory)
+        {
+            try
+            {
+                bool pathInsideDotGit = Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath);
+                if (pathInsideDotGit)
+                {
+                    if (relativePath.Equals(GVFSConstants.DotGit.Index, StringComparison.OrdinalIgnoreCase))
```
wilbaker: This is a much larger question/discussion than this specific line (and no need to hold this PR up on the discussion).

Does VFS4G on Linux need to work with case-sensitive file systems?  If so, we'l need to find a way to make the comparison type part of the platform rather than having hard-coded `OrdinalIgnoreCase` (there will probably need to be a few other changes as well, but that's the big one that jumps out at me).

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -0,0 +1,719 @@
...
+            this.OnFileRenamed(
+                relativeSourcePath: string.Empty,
+                relativeDestinationPath: relativeDestinationPath,
+                isDirectory: isDirectory);
+        }
+
+        private void OnHardLinkCreated(string relativeNewLinkPath)
+        {
+            this.OnHardLinkCreated(
+                relativeExistingFilePath: string.Empty,
```
wilbaker: FYI - This is changing on Mac (see #1184) and we will start delivering the second path to VFS4G.  You'll most likely want the same on Linux.

GVFS/GVFS.Hooks/GVFS.Hooks.Linux.csproj
```diff
@@ -0,0 +1,113 @@
...
+      <Link>Common\Tracing\EventOpcode.cs</Link>
+    </Compile>
+    <Compile Include="..\GVFS.Common\Tracing\ITracer.cs">
+      <Link>Common\Tracing\ITracer.cs</Link>
+    </Compile>
+    <Compile Include="..\GVFS.Common\Tracing\Keywords.cs">
+      <Link>Common\Tracing\Keywords.cs</Link>
+    </Compile>
+    <Compile Include="..\GVFS.Platform.Linux\LinuxPlatform.Shared.cs">
+      <Link>"LinuxPlatform.Shared.cs"</Link>
```
wilbaker: For consistency with Mac and Windows, link this in as "Linux\LinuxPlatform.Shared.cs"

GVFS.sln
```diff
@@ -201,243 +217,385 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "GVFS.PostIndexChangedHook.W
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug.Linux|x64 = Debug.Linux|x64
```
wilbaker: The good news is I was able to build your branch on Mac with `Scripts/Mac/BuildGVFSForMac.sh`.  The bad news is that something is configured in a way that VS for Mac doesn't like, but that can get sorted out before this is merged to master.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -34,7 +39,19 @@
...
+        <RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>
+      </PropertyGroup>
+      <ItemGroup>
+        <ProjectReference Include="..\GVFS.Platform.Linux\GVFS.Platform.Linux.csproj" />
+        <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Linux.cs">
+          <Link>PlatformLoader.Linux.cs</Link>
+        </Compile>
+      </ItemGroup>
+    </When>
+    <When Condition="'$(IsOSX)' == 'true'">
```
wilbaker: This `When` is also broken in VS for Mac and needs to be switched back to `Otherwise`

Scripts/Linux/BuildGVFSForLinux.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+. "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
+
+CONFIGURATION=$1
```
wilbaker: Not needed for this PR but at some point you'll probably want this script to take a version number like the Mac build script does so that the version number of the build can be passed through.

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: Good question -- in brief, yes, because the issue here is that we need the provider to inform us whether the top-level directory of a mount (or a remount of an unmounted working directory) should have its empty placeholder flag set or cleared.  The underlying "backing root" (e.g., `.vfsforgit/lower`) is created by the clone step (`gvfs clone`), but it has no placeholder flag set.

A user might mount (`gvfs mount`), visit the working directory and cause a hydration, and unmount(`gvfs unmount`), and then remount later, and we do want to start the virtualization again on the second `gvfs mount`.

Your question does make me wonder if we should choose a more descriptive name for the `IsUninitializedMount()` check ... @kivikakk, any thoughts here?  (Naming is never my strong suit! :-)

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: Oooo ... good point.  Yes, that's going to be an issue; we don't want to match on `.GIT` or `.Git`, etc.  I think we'll have to take that off into another PR, for sure.

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: Yup ... we've been tracking that work, and fortunately the lower-level code already supplies both paths.

GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -84,6 +88,15 @@ public static void Main(string[] args)
...
 
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                excludeCategories.Add(Categories.WindowsOnly);
+            }
+            else
+            {
+                excludeCategories.Add(Categories.LinuxOnly);
+            }
+
```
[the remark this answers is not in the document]
kivikakk: * We already exclude `MacOnly` because of that same following block, which adds `MacOnly` to the exclusion list if the platform is not `OSPlatform.OSX`.
* I'm unsure on what you're asking with the second question here. Incorporate what new exclusions where?

GVFS/GVFS.FunctionalTests/FileSystemRunners/FileSystemRunner.cs
```diff
@@ -16,6 +16,13 @@ public abstract class FileSystemRunner
                 new object[] { new BashRunner() },
             };
 
+        public static object[] AllLinuxRunners { get; } =
```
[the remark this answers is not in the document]
kivikakk: Sure!

GVFS/GVFS.Platform.Linux/GVFS.Platform.Linux.csproj
kivikakk: Yes, removed!

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
kivikakk: I don't have any strong opinions on the name — `MountNotYetHydrated`? Something that conveys a little more specifically what the test is for would be ideal.

GVFS/GVFS.Hooks/GVFS.Hooks.Linux.csproj
kivikakk: Thanks, fixed!

GVFS/FastFetch/FastFetch.csproj
```diff
@@ -33,16 +38,30 @@
...
+      <ItemGroup>
+        <ProjectReference Include="..\GVFS.Platform.Linux\GVFS.Platform.Linux.csproj" />
+        <Compile Include="..\GVFS.PlatformLoader\PlatformLoader.Linux.cs">
+          <Link>PlatformLoader.Linux.cs</Link>
+        </Compile>
+        <Compile Remove="NativeMethods.Mac.cs" />
+        <Compile Remove="NativeMethods.Windows.cs" />
+      </ItemGroup>
+    </When>
+    <When Condition="'$(IsOSX)' == 'true'">
```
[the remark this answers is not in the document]
kivikakk: Got it. The amount of trickery we're performing in the project files themselves does seem a bit beyond VS at times.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
kivikakk: :+1:

GVFS.sln
kivikakk: Hm, can you describe it at all? I just tried a reasonably fresh build with VS for Mac of this tree and everything worked OK.

I trashed `BuildOutput`, `packages`, `Publish`, ran enough of `BuildGVFSForMac.sh` to generate the constants and assembly version files, then built the solution. Everything looks fine.

---

I will note that I encounter issues regularly if I try to build with VS and then build script (or vice versa) without cleaning everything in between — VS gets confused, red-underlines everything, says `System` is undefined, etc. etc. Just to see if I could trigger it in this case, after building with VS per above, I ran the full `BuildGVFSForMac.sh`, and then again in VS, and this _one_ time it actually works.

(Well, actually, I get the incompatible SDK error thing on two projects, but that's not what I'm talking about and is already known!)

Scripts/Linux/BuildGVFSForLinux.sh
kivikakk: Done in 358254ec.

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
kivikakk: 👍

GVFS/GVFS.FunctionalTests/Program.cs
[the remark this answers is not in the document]
kivikakk: > * I'm unsure on what you're asking with the second question here. Incorporate what new exclusions where?

(To be clear, not trying to be obtuse! Just not parsing the question well.)

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: Captured as github/VFSForGit#29.

GVFS/GVFS.FunctionalTests/Program.cs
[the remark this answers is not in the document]
chrisd8088: My reading of @wilbaker's suggestion here is that because many of these functional tests have not yet been ported to the Mac (and are therefore Windows-only not by design, but only because of lack of time), we will want to, for now, also exclude these on Linux.  I'll make that change shortly!

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
chrisd8088: I'm going to go with `IsDirectoryEmpty()` and `IsWorkingDirectoryEmpty()` for `MirrorProvider` and `GVFS` respectively -- I think that captures (a) the difference between them, (b) the fact that we're testing whether the directory is "empty" (from the provider's standpoint) and needs to be flagged for future hydration, and (c) that we're really only testing a directory's contents.  The fact that it's a mount point is incidental to the tests.  I hope that works for everyone, and if not, we can change again!

GVFS/GVFS.FunctionalTests/Program.cs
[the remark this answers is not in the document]
chrisd8088: I think I'm going to leave this as-is for now, because we haven't yet run the functional tests on Linux (see github/VFSForGit#16) and so we may find that we need a different, likely larger, set of `LinuxTODO` tests.  Until we investigate, it's a little hard to know which ones from the `MacTODO` category will really be `POSIX-TODO` and apply to Linux as well, and which are Mac-specific.

But as I'm 100% certain we'll be revisiting these tests and categoroes very soon, I think I'll go ahead and merge with this as it stands, as a placeholder (projecting filesystem pun intended! :-)

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
wilbaker: @chrisd8088 FYI you'll also need to adjust how the placeholder database is created (see #1140) as currently the table is always being created with a case-insensitive path/key and that might need to be OS dependent as well.

wilbaker: (assuming you want to allow repos with file names that differ only in case)

GVFS/GVFS.FunctionalTests/Program.cs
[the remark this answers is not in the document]
wilbaker: @kivikakk @chrisd8088 apologies for the confusion.

The idea in the second point was to combine the new if/else with the existing one below so that it would look something like:

```
if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
{
    excludeCategories.Add(... stuff excluded on Mac ...);
}
else if (RuntimeInformation.IsOSPlatform(Linux))
{
    excludeCategories.Add(... stuff excluded on Linux ...);
}
else
{
   excludeCategories.Add(... stuff excluded on Windows ...);
}
```

It can get tricky to keep track of what we're excluding or including, and it might be easier to keep track if we set them all in one spot.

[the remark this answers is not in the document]
wilbaker: To clarify, when I wrote `excludeCategories.Add(... stuff excluded on Linux ...);` I meant call  `excludeCategories.Add` as many times as needed in each block to exclude what should be excluded on each platform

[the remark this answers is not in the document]
kivikakk: Ah, makes perfect sense. I'll do a PR for this -- thanks!

GVFS.sln
wilbaker: Hi, these are the errors I'm seeing:

> Error: NuGet packages need to be restored before building. NuGet MSBuild targets are missing and are needed for building. The NuGet MSBuild targets are generated when the NuGet packages are restored. (GVFS.Common)
>
> Error: NuGet packages need to be restored before building. NuGet MSBuild targets are missing and are needed for building. The NuGet MSBuild targets are generated when the NuGet packages are restored. (GVFS.Hooks.Mac)
>
> Error: NuGet packages need to be restored before building. NuGet MSBuild targets are missing and are needed for building. The NuGet MSBuild targets are generated when the NuGet packages are restored. (GVFS.Tests)

I tried deleting BuildOutput, packages, and Publish but it didn't help.

kivikakk: Trying to make my steps reproducible:
* Start with VS closed.
* Deleted `BuildOutput`, `packages`, `Publish`. The clone is at `~/microsoft/VFSForGit/src`, and the `VFSForGit` directory is empty except for the clone at `src`.
* Performed `git clean -fdx` in the clone.
* Opened `GVFS.sln` in Visual Studio for Mac (Community, 8.0.4 build 0).
* It automatically restored packages when opening the solution. (I'll omit the entire Package Console log, but the last line reads `Packages successfully restored.`)
* Tried to build the solution. Failed because I have `Debug.Windows` selected and a bunch of "This project type is not supported by Visual Studio Community 2019 for Mac." messages appeared.
* Switched to `Debug.Mac`. Failed because `GVFSConstants.GitVersion.cs` is not found.
* Made [these changes](https://gist.github.com/kivikakk/bbc1a0ca7d43ead9962eca3a0ebb8ba5) to `BuildGVFSForMac.sh` and ran the script to create the two generated `.cs` files.
* Built the solution again.
* ```
  ========== Build: 16 succeeded, 0 failed, 2 up-to-date, 7 skipped ==========

  Build successful.
  ```

It's possible the `git clean` step was necessary? I'm not sure.

wilbaker: Ahh thanks, I did forget to run `git clean`, let me give that a try.

PR #1128, opened 2019-05-09
[Linux] Implement update methods on Linux for placeholder files/dirs

chrisd8088:
For the Linux implementation of the equivalent of the [`PrjFSLib.Mac.DeleteFile()`](https://github.com/microsoft/VFSForGit/blob/29fd6a1746fabedea35ac920ef0a2f844aeacb3e/ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs#L72) method (and [PrjFS_DeleteFile()](https://github.com/microsoft/VFSForGit/blob/29fd6a1746fabedea35ac920ef0a2f844aeacb3e/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L527) C++ function), we rely on the [`projfs_get_attrs()`](https://github.com/github/libprojfs/blob/9f36ea1b4a3a38d97fd865049fcf11fd9ffe7435/include/projfs.h#L171) function in our libprojfs library, which allows us to read the `user.projection.empty` xattr.  If this xattr exists, the file or directory is not fully local, and may be deleted; otherwise, we return an error.

Our implementation of these functions follows the logic in the Mac `PrjFSLib.cpp` code, including the [small races](https://github.com/microsoft/VFSForGit/blob/master/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L546) in the `DeleteFile()` method between performing a `stat(2)` to check the nature of the file/dir, then a `getxattr(2)` to look for our projection flag xattr, and finally an `unlink(2)` or `rmdir(2)`.  (Note that unlike the Mac C++ implementation, here these low-level syscalls are actually performed in either libprojfs or in the C# standard library; they are just wrapped inside our calls to `ProjFS.GetProjState()`, `File.Delete()`, etc.)

Although we could try for a "more atomic" implementation by moving the code into libprojfs and acquiring a `flock(2)`, after consultation with @wilbaker this seems excessive, given that Git itself performs stat-followed-by-delete when trying to avoid deleting locally modified files, which is all this code is doing too.

Separately, we ensure Linux platform CI builds succeed by ignoring the `Profiling(Release)` builds (as done for [Mac as well](https://github.com/microsoft/VFSForGit/blob/29fd6a1746fabedea35ac920ef0a2f844aeacb3e/ProjFS.Mac/Scripts/Build.sh#L93)) and searching the usual system locations for our libprojfs headers and library.

Along with ~#1104~, ~#1119~, and ~#1123~ (ALL DONE), this PR is a prerequisite for the `GVFS.Platform.Linux` implementation in #1125.

Will resolve github/VFSForGit#13.

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -100,62 +100,47 @@ namespace PrjFSLib.Linux
             out UpdateFailureCause failureCause)
         {
             failureCause = UpdateFailureCause.NoFailure;
-            if (relativePath == string.Empty)
+            if (string.IsNullOrEmpty(relativePath))
             {
-                // mount point directory can not be deleted, so ignore
-                return Result.Success;
+                /* Our mount point directory can not be deleted; we would
+                 * receive an EBUSY error.  Therefore we just return
+                 * EDirectoryNotEmpty because that error is silently handled
+                 * by our caller in GitIndexProjection, and this is the
+                 * expected behavior (corresponding to the Mac implementation).
+                 */
+                return Result.EDirectoryNotEmpty;
             }
 
             string fullPath = Path.Combine(this.virtualizationRoot, relativePath);
-            try
+            bool isDirectory = Directory.Exists(fullPath);
+            Result result = Result.Success;
+            if (!isDirectory)
             {
-                if (Directory.Exists(fullPath))
-                {
-                    Directory.Delete(fullPath);
-                }
-                else
-                {
-                    // TODO(Linux): try to handle races with hydration?
-                    ProjectionState state;
-                    Result result = this.projfs.GetProjState(relativePath, out state);
+                // TODO(Linux): try to handle races with hydration?
+                ProjectionState state;
+                result = this.projfs.GetProjState(relativePath, out state);
 
-                    if (result == Result.EPathNotFound)
-                    {
-                        return Result.Success;
-                    }
-                    else if (result == Result.EAccessDenied || (result == Result.Success && state == ProjectionState.Full))
-                    {
-                        /* TODO(Linux): return EAccessDenied unless EPERM
-                         * (EPERM is returned when path is not a file or dir,
-                         *  which we want to treat as a full file)
-                         */
-                        failureCause = UpdateFailureCause.DirtyData;
-                        return Result.EVirtualizationInvalidOperation;
-                    }
-                    else if (result != Result.Success)
-                    {
-                        return result;
-                    }
-
-                    File.Delete(fullPath);
+                // treat unknown state (i.e., a special file) as a full file
+                if ((result == Result.Success && state == ProjectionState.Full) ||
+                    (result == Result.Invalid && state == ProjectionState.Unknown))
+                {
+                    failureCause = UpdateFailureCause.DirtyData;
+                    return Result.EVirtualizationInvalidOperation;
                 }
             }
-            catch (IOException ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
+
+            if (result == Result.Success)
+            {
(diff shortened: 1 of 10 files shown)


chrisd8088:
Hi @kivikakk and @derrickstolee -- would you mind taking a second look at this PR, now that I've ended up making some moderately significant changes (in 18a1486)?  In summary:

In particular, from studying the calling context for #1160, I realized we really should be returning `EDirectoryNotEmpty` when a directory removal fails for that reason, but AFAIK there's no way to pick out that specific error condition from the generic `IOException` [returned by `Directory.Delete()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.delete).  Hence the switch to using the libc syscalls instead.

For the same reason, we also return `EDirectoryNotEmpty` when asked to remove the repo's root (which we can reasonably assume to never be empty, since it should have `.git` in it); this is a belt-and-suspenders check so that even if #1160 is merged in some form, any other callers won't trigger the same issue.

My apologies for the extra go-round of another review cycle.

kivikakk:
> but AFAIK there's no way to pick out that specific error condition from the generic `IOException` [returned by `Directory.Delete()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.delete).

The `IOException`'s `Message` property will read `"Directory not empty"`, but I don't know how stable that text can be expected to be. The `HResult` corresponds to the errno, and indeed, on Linux trying to perform `Directory.Delete()` on a non-empty directory yields an `IOException` with a `HResult` of 39 (`ENOTEMPTY`).

All looks good otherwise, though I'd slightly prefer using the framework methods over libc where possible.

chrisd8088:
Me too, @kivikakk -- that is, I had originally written this using `Directory.Delete()` and `File.Delete()` and was happy to avoid any `DllImport`s.  I could go back if there's a stable way to convert `IOException` to a Linux `errno` -- thanks for the tip about the `HResult` property.  Some sources suggest bitwise masking that; there's also this dotnet/corefx#34221 thread about making `IOException` more fine-grained.

I did find one example in the existing GVFS codebase of non-Windows code [testing an exception's `HResult`](https://github.com/microsoft/VFSForGit/blob/a6622fb354ea577171ffb5e48604836b6b5dfe46/GVFS/GVFS.Common/Git/GitProcess.cs#L49).

I'm honestly not sure if it would make this code more or less readable/maintainable to go back to `Directory.Delete()` (and `File.Delete()`) -- I'm zonked tonight, but let me try it that way tomorrow and if it looks better, I'll add a second commit to this PR.

chrisd8088:
Hi @kivikakk and @derrickstolee -- I think third time might be a charm on this PR.  Would you mind doing a once-over and re-approving if it looks OK?  Thanks so much and apologies for all the churn.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -235,6 +238,50 @@ public virtual void StopVirtualizationInstance()
...
+            }
+            catch (IOException ex) when (ex is DirectoryNotFoundException)
+            {
+                return Result.EPathNotFound;
+            }
+            catch (IOException ex) when (ex is FileNotFoundException)
+            {
+                return Result.EFileNotFound;
+            }
+            catch (IOException ex)
```
kivikakk: For consistency with above you could consider `catch (IOException ex) when (ex.HResult == Errno.Constants.ENOTEMPTY)`.

chrisd8088: Good idea!  Done!

chrisd8088:
For reference, @derrickstolee, here's what's changed:
https://github.com/github/VFSForGit/compare/ac974c7..a50f57a

chrisd8088:
There's no delay to apologize for, @derrickstolee -- I only finished this up last night!  Thanks for the fast review!

PR #1130, opened 2019-05-09
Mac: Don't Fill Placeholders & Properly Remove Placeholder Flag

### LINKED ISSUE #1133: Mac: Determine if placeholders need to be created to file size
Since we hydrate on KAUTH_VNODE_READ_ATTRIBUTES and stat(), calls  KAUTH_VNODE_READ_ATTRIBUTES | KAUTH_VNODE_READ_SECURITY, this seems unnecessary.  

Credit to @pmj for finding the reference https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/vfs_vnops.c#L1446

jeschu1:
1. Don't fill placeholders with data to make size
  - Commands that create placeholders without hydration will receive a performance boost
  - In perf tests I saw a 4% overall drop for a million placeholders (123 second down to 118)

2. Properly remove placeholder flag when file is covered to full (KAUTH_VNODE_WRITE_DATA / KAUTH_VNODE_APPEND_DATA).
  - This will avoid calling PreConvertToFull again when the same Vnode Events are called.

resolves #1133

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -1060,7 +1060,7 @@ static PrjFS_Result HandleFileNotification(
         notificationType,
         nullptr /* destinationRelativePath */);
     
-    if (result == 0 && placeholderFile && PrjFS_NotificationType_PreConvertToFull == notificationType)
+    if (result == PrjFS_Result_Success && placeholderFile && PrjFS_NotificationType_PreConvertToFull == notificationType)
     {
         errno_t result = RemoveXAttrWithoutFollowingLinks(absolutePath, PrjFSFileXAttrName);
         if (0 != result)
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -389,13 +389,6 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
         goto CleanupAndFail;
     }
     
-    // Expand the file to the desired size
-    if (ftruncate(fileno(file), fileSize))
-    {
-        result = PrjFS_Result_EIOError;
-        goto CleanupAndFail;
-    }
-    
     fclose(file);
     file = nullptr;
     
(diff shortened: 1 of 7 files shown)


wilbaker:
Nit: Please tag the PR with the appropriate labels

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -389,13 +389,6 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
...
     }
     
-    // Expand the file to the desired size
-    if (ftruncate(fileno(file), fileSize))
-    {
-        result = PrjFS_Result_EIOError;
-        goto CleanupAndFail;
-    }
-    
     fclose(file);
```
wilbaker: Did you validate that [`stat` and `fstat`](https://linux.die.net/man/2/fstat) will see the correct size with this change?

pmj: > The change looks good overall, but I left a question on regarding testing out `stat` and `fstat`.
> 
> It would be good to hear @pmj's thoughts on this change as well.

I don't really seed `fstat()` being an issue - having a file descriptor implies the file has gone through `open()` so it'll be hydrated.

I'm personally not worried about `stat()` either as I've satisfied myself that the kernel’s implementation of it calls into the vnode listener with a hydrating action.

Nevertheless, it'd perhaps be worth adding a functional test for `stat()` - I don't know if there's a direct .NET wrapper for it we can easily access from the functional test runner, or if the functional tests are set up for easily adding some native code.[*] Not a dealbreaker for me if we don't, though - I suspect plenty of functional tests currently implicitly test this by using `stat()` as the backend to whatever high-level .NET stdlib functions they use.

[*] If it's currently not straightforward to use native code in functional tests, I wonder if that's something worth doing in itself, especially as macOS has some pretty odd syscalls such as [`exchangedata(2)`](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/exchangedata.2.html) that we may eventually want to test.

wilbaker: > Nevertheless, it'd perhaps be worth adding a functional test for stat()

I agree.  We already have some C# code in the product that calls `stat()` and so it should be straightforward to do it in the functional tests as well.

jeschu1:
> Unless it's super easy to do, I'm not convinced the stat() functional test is worth it.
@wilbaker and I decided the 'stat' command line call would do the trick.  That's easy to add and avoids the need to import the Native method/struct.

> PR looks good to go to me. There's one minor detail I noticed - "Importments" in the title should probably be "Improvements" - and the title is perhaps a little too generic for my liking overall. (Given that it'll presumably be used as the merge commit title.)
Great Feedback I'll update the title name!

PR #1131, opened 2019-05-09
PostFetchStep: delete and rewrite commit-graph on write error

derrickstolee:
We are no longer using 'verify' to check that our commit-graph
is valid, but we can still discover some errors during the 'write'
command. We currently never recover from this problem.

If we inspect the response from the 'write' command, then we can
delete and rewrite on failure. This will resolve some issues that
users are seeing right now.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
@@ -46,7 +46,12 @@ namespace GVFS.Common.Maintenance
                 string commitGraphLockPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", CommitGraphLock);
                 this.Context.FileSystem.TryDeleteFile(commitGraphLockPath);
 
-                this.RunGitCommand((process) => process.WriteCommitGraph(this.Context.Enlistment.GitObjectsRoot, this.packIndexes), nameof(GitProcess.WriteCommitGraph));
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Common/Maintenance/PostFetchStep.cs
```diff
@@ -46,7 +46,12 @@ protected override void PerformMaintenance()
                 string commitGraphLockPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", CommitGraphLock);
                 this.Context.FileSystem.TryDeleteFile(commitGraphLockPath);
 
-                this.RunGitCommand((process) => process.WriteCommitGraph(this.Context.Enlistment.GitObjectsRoot, this.packIndexes), nameof(GitProcess.WriteCommitGraph));
+                GitProcess.Result writeResult = this.RunGitCommand((process) => process.WriteCommitGraph(this.Context.Enlistment.GitObjectsRoot, this.packIndexes), nameof(GitProcess.WriteCommitGraph));
+
+                if (writeResult.ExitCodeIsFailure)
+                {
+                    this.LogErrorAndRewriteCommitGraph(activity, this.packIndexes);
```
alameenshah: This logs error and attempts to write commit graph again, right? - wondering why it would succeed the second time.

derrickstolee: It deletes the commit-graph first.

alameenshah: oh ok...

PR #1132, opened 2019-05-09
netcore build: fix runtime patch selection issues

jamill:
Issue:

When switching between command line and IDE builds, the build will often fail
with an error similar to:

  NETSDK1061: The project was restored using Microsoft.NETCore.App version
  2.1.9, but with current settings, version 2.1.0 would be used instead. To
  resolve this issue, make sure the same settings are used for restore and for
  subsequent operations such as build or publish. Typically this issue can occur
  if the RuntimeIdentifier property is set during build or publish but not during
  restore. For more information, see
  https://aka.ms/dotnet-runtime-patch-selection.

My understanding is that this can happen when the runtime used by the restore
operation does not match the runtime expected by the publish operation. As the
publish operation might not call restore itself, it will generate this error
when the runtime does not match the expected version.

To fix this, the BuildGVFSForMac.sh script will not include the runtime
identifier when calling restore. The projects that generate self-contained
applications will also set the `TargetLatestRuntimePatch` property to target
the latest runtime.

### DIFF
--- a/GVFS/GVFS.Platform.Mac/GVFS.Platform.Mac.csproj
+++ b/GVFS/GVFS.Platform.Mac/GVFS.Platform.Mac.csproj
@@ -4,6 +4,7 @@
     <TargetFrameworks>netcoreapp2.1;netstandard2.0</TargetFrameworks>
     <Platforms>x64</Platforms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
--- a/GVFS/GVFS.Platform.POSIX/GVFS.Platform.POSIX.csproj
+++ b/GVFS/GVFS.Platform.POSIX/GVFS.Platform.POSIX.csproj
@@ -4,6 +4,7 @@
     <TargetFrameworks>netcoreapp2.1;netstandard2.0</TargetFrameworks>
     <Platforms>x64</Platforms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
--- a/GVFS/GVFS.Service/GVFS.Service.Mac.csproj
+++ b/GVFS/GVFS.Service/GVFS.Service.Mac.csproj
@@ -9,6 +9,7 @@
     <Platforms>x64</Platforms>
     <RunTimeIdentifiers>osx-x64</RunTimeIdentifiers>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
   
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
--- a/ProjFS.Mac/PrjFSLib.Mac.Managed/PrjFSLib.Mac.Managed.csproj
+++ b/ProjFS.Mac/PrjFSLib.Mac.Managed/PrjFSLib.Mac.Managed.csproj
@@ -4,6 +4,7 @@
     <Platforms>x64</Platforms>
     <Configurations>Debug;Release</Configurations>
     <CodeAnalysisRuleSet>..\..\GVFS\GVFS.Build\GVFS.ruleset</CodeAnalysisRuleSet>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
   
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">


derrickstolee:
@jamill I had this same problem on my home machine after playing with [Stryker](https://github.com/stryker-mutator/stryker-net).  I'll give this topic a try later to see if it fixes the issue.

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -50,7 +50,7 @@ $VFS_SCRIPTDIR/GenerateCommonAssemblyVersion.sh $VERSION || exit 1
 # overriding build targets. The duplicate import generates warning MSB4011 that is ignored
 # by this switch.
 echo 'Restoring packages...'
-dotnet restore $VFS_SRCDIR/GVFS.sln /p:Configuration=$CONFIGURATION.Mac --packages $VFS_PACKAGESDIR /warnasmessage:MSB4011 || exit 1
+dotnet restore $VFS_SRCDIR/GVFS.sln --runtime osx-x64 /p:Configuration=$CONFIGURATION.Mac --packages $VFS_PACKAGESDIR /warnasmessage:MSB4011 || exit 1
```
wilbaker: >My understanding is that this can happen when the runtime used by the restore
operation does not match the runtime expected by the publish operation. As the
publish operation might not call restore itself, it will generate this error
when the runtime does not match the expected version.

The change looks simple enough, but I am having trouble understanding how passing in `--runtime osx-x64` makes a difference.  Does this mean that previously some step in restore or publish was not using osx-x64?

GVFS/GVFS.Platform.Mac/GVFS.Platform.Mac.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>netcoreapp2.1;netstandard2.0</TargetFrameworks>
     <Platforms>x64</Platforms>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
```
wilbaker: What runtime were we targeting before this change?

Or (based on https://aka.ms/dotnet-runtime-patch-selection) were we always publishing with the latest but building with a version that could potentially be different?

jamill: I think the problem is that the IDE tooling did not always specify a run-time version during the publish step, while our build scripts do. I *suspect* this lack of an explicit runtime identifier in the IDE builds for publish is what is causing confusion... without the `TargetLatestRuntimePath` property, the IDE generates:

1) A non-self contained version that targets systems with a minimum Framework  version of 2.1
2) A self contained exe with a v 2.1.9 version of the runtime

I think the error is in case this discrepancy is not OK (i.e. you have a hard dependency on the 2.1.9 version of the runtime).

> were we always publishing with the latest but building with a version that could potentially be different?

No - I think this error is telling us that there would be a discrepancy. The build scripts include the runtime version in the publish step, so I think it could use that information when running the publish command, and not have ambiguity between the version targeted by a self contained exe versus a version that is meant to be run in environments where the system provides the Framework libs.

Scripts/Mac/BuildGVFSForMac.sh
jamill: I think this means previously restore was not explicity using osx-64 from the botnet command, and instead would have picked it up from the project files. From the documentation:

> This is used to restore packages for runtimes not explicitly listed in the <RuntimeIdentifiers> tag in the .csproj file.

I am not sure if this change is absolutely necessary, as many (all?) of our projects specify the runtime identifier as well, but it seemed reasonable, as this is the only runtime we are targeting, and the other commands include it.

GVFS/GVFS.Platform.Mac/GVFS.Platform.Mac.csproj
jamill: > What runtime were we targeting before this change?

I think our build script would always generate self contained executables that targeted the latest patch version on the build machine.

wilbaker: >I think the problem is that the IDE tooling did not always specify a run-time version during the publish step, while our build scripts do.

For my own understanding, does "run-time version" mean something like `--runtime osx-x64`?

> A self contained exe with a v 2.1.9 version of the runtime

Do you know where it was picking up 2.1.9 from (and how did you figure this out)?

wilbaker: I just saw we were already including `<TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>` in most of the projects already (and that's probably reason enough to do it here).

Based on the link I think this sounds like us?

> Prefer the self-contained behavior. If the application is self-contained, this is the default behavior. If the application is framework-dependent, and requires the latest patch installed, set TargetLatestRuntimePatch to true in the project file.

"requires the latest patch installed" is the one part I'm not sure about.  Do we require that the latest patch is installed (it sounds like a good thing even if we don't have a strict dependency on it?).

This PR looks good, but I still don't have a good understanding of exactly what we're specifying when building and publishing.

CC: @nickgra as well, I saw he did some research into this in an earlier PR (https://github.com/microsoft/VFSForGit/pull/496#discussion_r232555856).

jamill: > For my own understanding, does "run-time version" mean something like --runtime osx-x64?

I think when you build a self contained executable (I.e. You specify a runtime id in the former command, or the project file has a runtime specified), it chooses the latest runtime available by default.

The specific versions here came from the error messages I encountered

jamill: > "requires the latest patch installed" is the one part I'm not sure about

I think this refers to "Framework Dependent Deployments" (FDD), or deployments where we would expect the target system to provide the framework. With this property set, "Self Contained Deployments" (which is what I think actually care about) include 2.1.9, and a FDD would require  the target system to have at 2.1.9 installed. Without this property, I think the publish command is complaining that FDD have a minimum target version of 2.1, but was built against the 2.1.9 runtime.

Even if we don't care about FDD, I suspect the publish commands issued by IDE builds still generates a FDD package (whereas our command line builds generally provide the '--runtime' parameter in the dotnet buil and publish commands) and this is causing ambiguity.

Scripts/Mac/BuildGVFSForMac.sh
jamill: Is this command even necessary at all still? With NETCore 2.0, the restore should be run implicitly by other dotnet commands...

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore?tabs=netcore2x

mjcheetham: Correct.. however there are can be subtle differences when it comes to differing package reference styles, and if you're mixing old and new MSBuild project formats. Since this is for the _Mac_ build only (which should be .NET Core only, and MSBuild 15 only) it should be ok to drop the explicit `restore`.

Although it should still be tested!

GVFS/GVFS.Platform.Mac/GVFS.Platform.Mac.csproj
mjcheetham: > Even if we don't care about FDD, I suspect the publish commands issued by IDE builds still generates a FDD package (whereas our command line builds generally provide the '--runtime' parameter in the dotnet buil and publish commands) and this is causing ambiguity.

I believe even command line builds create the FDD assemblies first, then do the self-contained ones after? At least that's what I've seen.

PR #1137, opened 2019-05-09
Upgrader: do not include Micrsoft.Diagnostic.Tracing.EventSource

jamill:
The upgrade verb attempts to copy
Microsoft.Diagnostic.Tracing.EventSource when copying the
ProductUpgrader to the temporary directory to run. However, this dll is
no longer needed or installed during installation.

Older installations will have copied this file into the install
directory will still have a copy for the upgrade verb to copy over.
However, clean installs, or if you uninstall VFS4G and re-install, will
not have this dll and upgrade will fail.

This is a quick fix - we will also see how to make this less fragile.
One option might be to install "upgrader app" into its own directory
under the VFS For Git installation directory, and then just copy the
directory to the temp location, instead of providing a list of files
that need to be kept in sync.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -36,7 +36,6 @@ namespace GVFS.Common
                 UpgraderToolConfigFile,
                 "GVFS.Common.dll",
                 "GVFS.Platform.Windows.dll",
-                "Microsoft.Diagnostics.Tracing.EventSource.dll",
                 "netstandard.dll",
                 "System.Net.Http.dll",
                 "Newtonsoft.Json.dll",
--- a/GVFS/GVFS.Mount/GVFS.Mount.Windows.csproj
+++ b/GVFS/GVFS.Mount/GVFS.Mount.Windows.csproj
@@ -41,10 +41,6 @@
     <Reference Include="Microsoft.Data.Sqlite, Version=2.2.4.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, processorArchitecture=MSIL">
       <HintPath>..\..\..\packages\Microsoft.Data.Sqlite.Core.2.2.4\lib\netstandard2.0\Microsoft.Data.Sqlite.dll</HintPath>
     </Reference>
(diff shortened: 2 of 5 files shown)


GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -36,7 +36,6 @@ public abstract class ProductUpgrader : IDisposable
                 UpgraderToolConfigFile,
                 "GVFS.Common.dll",
                 "GVFS.Platform.Windows.dll",
-                "Microsoft.Diagnostics.Tracing.EventSource.dll",
```
wilbaker: Should we also remove any references to `Microsoft.Diagnostics.Tracing` in projects\config files?

wilbaker:
>Older installations will have copied this file into the install directory will still have a copy for the upgrade verb to copy over.

Does the upgrader only need to worry about files that were included with its own installation?

I couldn't tell from your comment if there is an issue with older installations having this file around.

GVFS/GVFS.Common/ProductUpgrader.cs
jamill: Yes - we should. I will go and remove the other ones I find.

jamill:
> Does the upgrader only need to worry about files that were included with its own installation?
> 
> I couldn't tell from your comment if there is an issue with older installations having this file around.

The upgrader needs to copy over its dependencies. In this case, the file is no longer a dependency (so we do not need to copy it over), but it is still in the list of files that the upgrade verb tries to copy over. The reason why we didn't notice this before (and lessens the impact), is that users who already have this file in the installation directory (via an older installation) will still be able to upgrade, as this file is apparently not removed when installing a new version over an old version...

I will update the commit message to include the commit where this file was removed from the installer (cbe5787db2df7d9ee7724dc069836a0e5ddd2d8f)

jrbriggs:
Looks good, thanks @jamill. Please port to release branch.

jamill:
I added a new commit to remove references to the `Microsoft.Diagnostic.Tracing.EventSource` package as well.

I will only bring the first commit over when I port the changes to release branches.

/cc @wilbaker, @derrickstolee, @mjcheetham

PR #1140, opened 2019-05-09
Sqlite placeholders

kewillford:
I think this is far enough along to start taking a look at.  There is a lot of new code to handle the sqlite database but the other changes should be minimal.

I would like to get some feedback on how I am setting up the class for the connection and the class for the Placeholders table.  Something to keep in mind is if we are going to add ModifiedPaths to this same database, does this make it easy or is there a better way?

Currently it is structured with a `GVFSDatabase` class that initializes the database and keeps a pool of connections for consumers to use.  The connection is wrapped in a class that implements `IDbConnection` so that the `Dispose` method can return the underlying connection to the pool. Like the following:
```
using (IDbConnection pooled = this.database.GetPooledConnection())
using (SqliteCommand command = pooled.CreateCommand())
{
```

Another question is also how we show handle exceptions.  The `FileBasedCollection` would have try catches and throw a custom `FileBasedCollectionException`.  Should we be doing something similar for SQLite or just let exceptions bubble up?

Changed the `Count` property to be a method because it is now running a SQLite query and will take longer than returning a value and it being a method makes that more obvious.

Removed some constructors that caused me some confusion when I was adding the `IPlaceholderDatabase` as a parameter but I'm open to adding them back in if they make sense.

The most critical part to review would be the changes in `GitIndexProjection`.

TODO:
- [x] Add Upgrader
- [x] Error Handling
- [x] See what other possible tests I can add
- [x] Repair
- [x] Validate dehydrate works as expected
- [x] Validate diagnose works as expected
- [x] Add custom exception for sqlite database errors


GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,135 @@
...
+using GVFS.Common.Tracing;
+using Microsoft.Data.Sqlite;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GVFS.Common.Database
+{
+    public class GVFSDatabase : IDisposable
```
jamill: Consider including a class comment giving a brief description of what a `GVFSDatabase` class represents.

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,135 @@
...
+            return connection;
+        }
+
+        private void ReturnToPool(SqliteConnection connection)
+        {
+            if (this.disposed)
+            {
+                connection?.Dispose();
+            }
+            else if (!this.connectionPool.TryAdd(connection))
```
jamill: Can this throw? For example, if the pool has been marked complete with regards to additions?

https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1.tryadd?view=netframework-4.8

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -73,6 +69,15 @@ public static bool TryCreate(ITracer tracer, string dataFilePath, PhysicalFileSy
             return true;
         }
 
+        /// <summary>
+        /// The Count is "estimated" because it's simply (# adds - # deletes).  There is nothing to prevent
+        /// multiple adds or deletes of the same path from being double counted
+        /// </summary>
+        public int Count()
```
jamill: Maybe `EstimatedCount`?

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            return new GVFSConnection(this, connection);
+        }
+
+        private void Initialize()
+        {
+            using (IPooledConnection pooled = this.GetPooledConnection())
+            using (SqliteCommand command = pooled.Connection.CreateCommand())
+            {
+                command.CommandText = $"PRAGMA journal_mode=WAL;";
```
kewillford: Remove all the unnecessary `$` from strings

GVFS/GVFS.Common/Database/GVFSDatabase.cs
kewillford: I set the `disposed` to `true` before calling `CompleteAdding` but I see there is still a race if this code checked `disposed` and then the other code ran that sets the `disposed` and then calls `CompleteAdding` before this code runs the `TryAdd`.   I will have to handle this.  Thanks!

GVFS/GVFS.Common/PlaceholderListDatabase.cs
kewillford: This is part of the interface so it is just `Count()` because with SQLite it is not an estimate since it is running a SQL query to get the count.  This class will basically be deprecated with these changes and shouldn't be used except for in the disk layout upgrade.  I have even thought about renaming it to `DeprecatedPlaceholderListDatabase` to make that obvious.  What do you think?

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,135 @@
...
+            SqliteConnection connection = new SqliteConnection(this.sqliteConnectionString);
+            connection.Open();
+            return connection;
+        }
+
+        private void ReturnToPool(SqliteConnection connection)
+        {
+            if (this.disposed)
+            {
+                connection?.Dispose();
```
wilbaker: Does you use of `?.` imply that `connection` could be `null`?

If the connection could be `null` should we check if it's `null` before calling `this.connectionPool.TryAdd`?

GVFS/GVFS.Common/Database/IPlaceholderDatabase.cs
```diff
@@ -4,8 +4,7 @@ namespace GVFS.Common.Database
 {
     public interface IPlaceholderDatabase
     {
-        int Count { get; }
-
+        int Count();
```
wilbaker: Minor: Why was this changed from a property to a method?

If it stays a method could you update the name to include a verb (e.g. `GetCount`, `RetrieveCount`)

GVFS/GVFS.Common/GVFSConstants.cs
```diff
@@ -115,6 +115,7 @@ public static class Databases
                 public static readonly string PlaceholderList = Path.Combine(Name, "PlaceholderList.dat");
                 public static readonly string ModifiedPaths = Path.Combine(Name, "ModifiedPaths.dat");
                 public static readonly string RepoMetadata = Path.Combine(Name, "RepoMetadata.dat");
+                public static readonly string GVFSDatabase = Path.Combine(Name, "gvfs.sqlite");
```
wilbaker: I don't think we've had much guidance on this, but do you think the filename should use the new name (e.g. "VFSForGit.sqlite")?

I think it's okay to keep the code using the GVFS name for now.

cc: @jrbriggs for his thoughts

GVFS/GVFS.Common/Database/SqliteDatabase.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            return false;
+        }
+
+        public static string CreateConnectionString(string databasePath)
+        {
+            // Share-Cache mode allows multiple connections from the same process to share the same data cache
+            // http://www.sqlite.org/sharedcache.html
+            return $"data source={databasePath};Cache=Shared";
```
wilbaker: Have you tried running the new database without shared-cache mode?  The data-access pattern is different (than BlobSizes) and I'm not sure we can assume that shared-cache will be better.

GVFS/GVFS.Common/Database/Placeholders.cs
```diff
@@ -0,0 +1,199 @@
...
+            if (placeholder.Sha == null)
+            {
+                command.Parameters.Add("@sha", SqliteType.Text).Value = DBNull.Value;
+            }
+            else
+            {
+                command.Parameters.Add("@sha", SqliteType.Text).Value = placeholder.Sha;
+            }
+
+            command.ExecuteNonQuery();
```
wilbaker: Have you tried measuring the performance impact of using prepared statements, and if so, what did you find?

wilbaker:
> Another question is also how we show handle exceptions. The FileBasedCollection would have try catches and throw a custom FileBasedCollectionException. Should we be doing something similar for SQLite or just let exceptions bubble up?

FWIW we went with having a custom exception type for the BlobSizes database (`BlobSizesException`) to make it easier for callers to know what exception type(s) to expect.  The database code was able to map all possible exceptions to a single well know type.

GVFS/GVFS.Common/Database/Placeholders.cs
```diff
@@ -0,0 +1,199 @@
...
+        private GVFSDatabase database;
+
+        public Placeholders(GVFSDatabase database)
+        {
+            this.database = database;
+        }
+
+        public static void CreateTable(SqliteCommand command)
+        {
+            command.CommandText = @"CREATE TABLE IF NOT EXISTS [Placeholders] (path TEXT PRIMARY KEY, pathType TINYINT NOT NULL, sha char(40) ) WITHOUT ROWID;";
```
wilbaker: Style comment:  It looked a bit weird to me to have `CreateTable` accept a `SqliteCommand`, I would have expect `CreateTable` to create the `SqliteCommand` that it needs.

What would you think of having the method take a `SqliteConnection` instead?

GVFS/GVFS.Common/Database/Placeholders.cs
```diff
@@ -0,0 +1,199 @@
...
+    public class Placeholders : IPlaceholderDatabase
+    {
+        private GVFSDatabase database;
+
+        public Placeholders(GVFSDatabase database)
+        {
+            this.database = database;
+        }
+
+        public static void CreateTable(SqliteCommand command)
```
wilbaker: Related to comment on having `CreateTable` take a `SqliteCommand`, I'm having a bit of trouble understanding the relationship between `Placeholders` and `GVFSDatabase`.

- Why is `CreateTable` `static`?  If `Placeholders` has a `GVFSDatabase` member, could it just request a connection from the pool (like `Count` does)?
- If `Placeholders` is tied this closely to `GVFSDatabase`, should it be a nested class of `GVFSDatabase`?
- What are the pros/cons of making `Placeholders` is own class vs accessing placeholder data via `GVFSDatabase`?

GVFS/GVFS.Common/GVFSConstants.cs
jrbriggs: Given that this is an new file and has no compat concerns I think we switch to the new name to avoid building debt.

kewillford: So is this only for the name of this file or is it for all uses of `GVFS`?  Like should this variable be `VFSForGitDatabase` instead of `GVFSDatabase` and the class `GVFSDatabase`?

GVFS/GVFS.Common/Database/IPlaceholderDatabase.cs
kewillford: >Why was this changed from a property to a method?

Because it is now running a SQL query and not simply returning a stored value.  Making it a method makes it more obvious that it is running some code and could take longer that retrieving a value.  This is why `IEnumerable` has a `Count()` method and that is why I named it `Count` to follow that precedent but I can rename it.

GVFS/GVFS.Common/Database/Placeholders.cs
kewillford: So a little information on how I arrived here.  I originally had this code in the `GVFSDatabase` class because it is the class that is initializing the database and creating all the tables for the database but when creating the `Placeholders` class, I felt this needed to be a part of this class since this is the class that will be using the fields of the table and for easy reference to what the fields of the table would be I put them here.

>Why is CreateTable static? If Placeholders has a GVFSDatabase member, could it just request a connection from the pool (like Count does)?

Because `Placeholders` needs a `GVFSDatabase` but the `GVFSDatabase` needs to create the tables of the database in it's constructor there is kind of a chicken/egg problem here if it is not `static`.  I supposed we could have the `Placeholders` try to create the table in its constructor but I felt that was the responsibility of the `GVFSDatabase` class to make sure the database is initialized and ready to use.

>If Placeholders is tied this closely to GVFSDatabase, should it be a nested class of GVFSDatabase?

To me this is a separate issue that I can see either way.  I am making some changes so that I can unit test this class that will decouple this class from the `GVFSDatabase` class but it could still be nested because it is part of the `GVFSDatabase` class in a way.  I would still want to keep the ability to unit test it though so it would need to be `public` which would be an argument for not have it be nested.

>What are the pros/cons of making Placeholders is own class vs accessing placeholder data via GVFSDatabase?

The biggest con for putting everything in the `GVFSDatabase` is that it would grow and become harder and harder to maintain if we add modified paths, repo metadata, other tables, etc.  I guess the pro would be all the database stuff would be in one class?  I'm one for smaller classes that have specific responsibilities vs one that can do everything.

GVFS/GVFS.Common/Database/Placeholders.cs
kewillford: I did on the `Contains` method that I had and it did help.  The issue here would be that you need to use the same connection that the prepare was ran against.  This would require the class to keep connections around and do locking for using the commands that are prepared which would really add to the complexity so I will look into that after the PR is smoothed out.

GVFS/GVFS.Common/GVFSConstants.cs
wilbaker: > So is this only for the name of this file or is it for all uses of GVFS? 

I think the important part is the name of the file.  

 I don't think there'd be any objections to using `VFSForGitDatabase` but I also don't think you have to update the names in code right now either ("GVFS" is still used just about everywhere ).

GVFS/GVFS.Common/Database/Placeholders.cs
wilbaker: Thanks for all the details here.  What would you think of renaming `Placeholders` to `PlaceholdersTable`?

I think if the name of the class reflects that this class is specifically for interacting with a table in the database it might add some clarity to how/why responsibilities are split between these classes.

wilbaker: Actually, maybe put that suggestion on hold.  

I think there might be a few classes/interfaces to consider when it comes to naming `Placeholders`:

- `IPlaceholderDatabase` 
- `Placeholders`
- `GVFSDatabase`

If `IPlaceholderDatabase` doesn't have any "database" specific methods, maybe something like `IPlaceholderCollection` would be a more descriptive name?

And now I am torn on if the `Placeholders` name should have `Table` in it.  Do you see this class representing a table in the `GVFSDatabase`?

kewillford: >If IPlaceholderDatabase doesn't have any "database" specific methods, maybe something like IPlaceholderCollection would be a more descriptive name?

Yeah that sounds like a good change.

>Do you see this class representing a table in the GVFSDatabase?

Right now it is but it might not be the only one in the future.  For example if we add the modified paths to the database as another table and we have a method on a ModifiedPaths Table class that is Add but that method will also delete from the Placeholder Table then things are not as straight forward.  At that point though we may need another class that is more like the business logic that adds to the modified paths table and deletes from the placeholder table using a transaction.  So I think for this PR we can rename it to have `Table` in it.

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTests.cs
```diff
@@ -0,0 +1,428 @@
...
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+
+namespace GVFS.UnitTests.Common.Database
+{
+    [TestFixture]
+    public class PlaceholdersTests
```
kewillford: This class goes a little mock crazy so I'm not sure the value this brings vs the maintenance cost of all the mocking it is doing.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
```diff
@@ -0,0 +1,202 @@
...
+        public void AddExpandedFolder(string path)
+        {
+            using (IPooledConnection pooled = this.connectionPool.GetConnection())
+            using (IDbCommand command = pooled.Connection.CreateCommand())
+            {
+                Insert(command, new PlaceholderData() { Path = path, PathType = PlaceholderData.PlaceholderType.ExpandedFolder });
+            }
+        }
+
+        public void AddPossibleTombstoneFolder(string path)
```
mjcheetham: With all these repeated `Add[Type]Folder(string path)` methods, with nearly identical bodies, perhaps a single method `private void AddFolder(string path, PlaceholderData.PlaceholderType type)` would be cleaner, and have these `public` methods forward to the `private` one.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                Delete(command, path);
+            }
+        }
+
+        private static void Insert(IDbCommand command, PlaceholderData placeholder)
+        {
+            command.CommandText = "INSERT OR REPLACE INTO Placeholders (path, pathType, sha) VALUES (@path, @pathType, @sha);";
+            command.AddParameter("@path", DbType.String, placeholder.Path);
+            command.AddParameter("@pathType", DbType.Int32, (int)placeholder.PathType);
```
mjcheetham: `PlaceholderData.PathType` is an enumeration that is backed by a `byte`, but here we're casting it to `int`. Generally advice for enumerations is _do not specify a backing type_ unless there's a good reason.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
```diff
@@ -0,0 +1,202 @@
...
+
+            if (placeholder.Sha == null)
```
mjcheetham: Could also do...

```csharp
command.AddParameter("@sha", DbType.String, placeholder.Sha ?? DBNull.Value);
```

mjcheetham: Also do we ever expect `placeholder.Sha` to be an empty string?

GVFS/GVFS.Common/Database/SqliteDatabase.cs
```diff
@@ -0,0 +1,74 @@
...
+                try
+                {
+                    string sqliteConnectionString = CreateConnectionString(databasePath);
+                    using (SqliteConnection integrityConnection = new SqliteConnection(sqliteConnectionString))
+                    {
+                        integrityConnection.Open();
+
+                        using (SqliteCommand pragmaCommand = integrityConnection.CreateCommand())
+                        {
+                            pragmaCommand.CommandText = $"PRAGMA integrity_check;";
```
mjcheetham: nit:
```suggestion
                            pragmaCommand.CommandText = "PRAGMA integrity_check;";
```

GVFS/GVFS.Common/PlaceholderListDatabase.cs
mjcheetham: Maybe add the `[Obsolete]` attribute? Although that would introduce a warning at each use-site (which could be `#pramga disabled` in the layout upgrade code..).

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,142 @@
+﻿using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using Microsoft.Data.Sqlite;
```
derrickstolee: `using Microsoft.Data.Sqlite;` is unused.

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,142 @@
...
+    /// Handles setting up the database for storing data used by GVFS and
+    /// managing the connections to the database
+    /// </summary>
+    public class GVFSDatabase : IGVFSConnectionPool, IDisposable
+    {
+        private const int InitialPooledConnections = 5;
+        private const int MillisecondsWaitingToGetConnection = 50;
+
+        private bool disposed = false;
+        private ITracer tracer;
```
derrickstolee: Nothing is reading from the `tracer`.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+        }
+
+        public HashSet<string> GetAllFilePaths()
+        {
+            using (IPooledConnection pooled = this.connectionPool.GetConnection())
+            using (IDbCommand command = pooled.Connection.CreateCommand())
+            {
+                HashSet<string> fileEntries = new HashSet<string>();
+                command.CommandText = "SELECT path FROM Placeholders WHERE pathType = 0;";
```
derrickstolee: Should this `WHERE pathType = 0;` use an enum value instead of a literal?

GVFS/GVFS/RepairJobs/PlaceholderDatabaseRepairJob.cs
```diff
@@ -13,7 +14,7 @@ public class PlaceholderDatabaseRepairJob : RepairJob
         public PlaceholderDatabaseRepairJob(ITracer tracer, TextWriter output, GVFSEnlistment enlistment)
             : base(tracer, output, enlistment)
         {
-            this.databasePath = Path.Combine(this.Enlistment.DotGVFSRoot, GVFSConstants.DotGVFS.Databases.PlaceholderList);
+            this.databasePath = Path.Combine(this.Enlistment.DotGVFSRoot, GVFSConstants.DotGVFS.Databases.VFSForGit);
```
derrickstolee: This variable is no longer read, because you create `blobsizesDatabasePath` down below. Drop one?

GVFS/GVFS/RepairJobs/PlaceholderDatabaseRepairJob.cs
```diff
@@ -24,15 +25,10 @@ public override string Name
...
         {
             string error;
-            PlaceholderListDatabase placeholders;
-            if (!PlaceholderListDatabase.TryCreate(
-                this.Tracer,
-                this.databasePath,
-                new PhysicalFileSystem(),
-                out placeholders,
-                out error))
+            string blobsizesDatabasePath = Path.Combine(this.Enlistment.DotGVFSRoot, GVFSConstants.DotGVFS.Databases.VFSForGit);
```
derrickstolee: Consider just using `this.databasePath` as this database now has more than just sizes, right?

derrickstolee: I see from `BlobSizeDatabaseRepairJob` that the blob size database has a different path, so they are not tables in a combined database. Then definitely drop this name.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
kewillford: I will have to measure the memory and see.  We load all the placeholders into memory when we update them and if the user has most of the placeholders created ~3 million times 3 bytes is ~9MB that we save.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
derrickstolee: nit: I would have made this be `public int Count => this.count;` so the external readers do not need to change to a method. PLUS: whenever I see `.Count()` I immediately think it is a reference to the LINQ method that should never be used.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
kewillford: Because `DBNull.Value` and `placeholder.Sha` are not the same types you cannot use the `??` operator.

>do we ever expect placeholder.Sha to be an empty string?

No, it should either be `null` or a 40 character string.  I will add some validation for this.

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTableTests.cs
```diff
@@ -0,0 +1,435 @@
...
+        private const byte PathTypeExpandedFolder = 2;
+        private const byte PathTypePossibleTombstoneFolder = 3;
+        private const string DefaultSha = "SHA1";
+
+        [TestCase]
+        public void ConstructorTest()
+        {
+            Mock<IGVFSConnectionPool> mockConnectionPool = new Mock<IGVFSConnectionPool>(MockBehavior.Strict);
+            PlaceholdersTable placeholders = new PlaceholdersTable(mockConnectionPool.Object);
+            mockConnectionPool.VerifyAll();
```
derrickstolee: What is this testing? That `new PlaceholdersTable()` doesn't call any methods?

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTableTests.cs
```diff
@@ -0,0 +1,435 @@
...
+        }
+
+        [TestCase]
+        public void CountTest()
+        {
+            this.TestPlaceholders(
+                (placeholders, mockCommand) =>
+                {
+                    mockCommand.SetupSet(x => x.CommandText = "SELECT count(path) FROM Placeholders;");
+                    mockCommand.Setup(x => x.ExecuteScalar()).Returns(0);
```
derrickstolee: If you are making a custom setup here, could we use a value other than zero here? Could we have it return `123`, and verify that result?

derrickstolee: It just seems like there are lots of ways we could _accidentally_ be right with a zero result.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
```diff
@@ -0,0 +1,202 @@
...
+    public class PlaceholdersTable : IPlaceholderCollection
+    {
+        private IGVFSConnectionPool connectionPool;
+
+        public PlaceholdersTable(IGVFSConnectionPool connectionPool)
+        {
+            this.connectionPool = connectionPool;
+        }
+
+        public static void CreateTable(IDbCommand command)
```
wilbaker: I saw that this TODO was marked as complete:

> Validate diagnose works as expected

How does diagnose work with this change?  How will we be able to see the placeholder list entries?

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTableTests.cs
```diff
@@ -0,0 +1,435 @@
...
+        }
+
+        [TestCase]
+        public void GetAllFilePathsWithNoResults()
+        {
+            this.TestPlaceholdersWithReader(
+               (placeholders, mockCommand, mockReader) =>
+               {
+                   mockReader.Setup(x => x.Read()).Returns(false);
+                   mockCommand.SetupSet(x => x.CommandText = "SELECT path FROM Placeholders WHERE pathType = 0;");
```
derrickstolee: I like that the unit tests here use the literal "0". If you use the enum value in the implementation of `GetAllFilePaths()`, this test would detect a change in the specific enum values.

(We shouldn't change the enum values, but this gives us protection against it.)

GVFS/GVFS.Common/PlaceholderListDatabase.cs
kewillford: `Count()` is a part of the interface and I made it a method because I want users to think before using it because in the PlaceholderTable class it is making a query against the database and I don't want someone thinking it is just an in memory field that will be returned but is doing some work.

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTableTests.cs
```diff
@@ -0,0 +1,435 @@
...
+                       return readCalls == 1;
+                   });
+
+                   mockReader.Setup(x => x.GetString(0)).Returns(DefaultPath);
+                   mockCommand.SetupSet(x => x.CommandText = "SELECT path FROM Placeholders WHERE pathType = 0;");
+
+                   HashSet<string> filePaths = placeholders.GetAllFilePaths();
+                   filePaths.ShouldNotBeNull();
+                   filePaths.Count.ShouldEqual(1);
+                   filePaths.ShouldContain(x => x == DefaultPath);
```
derrickstolee: Should these `filePaths.ShouldContain(x => x == <thing>)` instead be `filePaths.Contains(<thing>).ShouldBeTrue()`? The `ShouldContain()` enumerates the set (it only knows its an `IEnumerable<T>`).

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTableTests.cs
```diff
@@ -0,0 +1,435 @@
...
+
+        private void PlaceholderDataShouldMatch(IPlaceholderData expected, IPlaceholderData actual)
+        {
+            actual.Path.ShouldEqual(expected.Path);
+            actual.IsFolder.ShouldEqual(expected.IsFolder);
+            actual.IsExpandedFolder.ShouldEqual(expected.IsExpandedFolder);
+            actual.IsPossibleTombstoneFolder.ShouldEqual(expected.IsPossibleTombstoneFolder);
+            actual.Sha.ShouldEqual(expected.Sha);
+        }
+    }
```
derrickstolee: I just want to say that these unit tests are fantastic!

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
kewillford: All the files in the `.gvfs\databases` directory will be copied to the diagnose file.  This will include the VFSForGit.sqlite file.  I have been using [DB Browser for SQLite](https://sqlitebrowser.org/) to look at this file and see the data.  It seems easy enough to use and browse/query the data.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -600,12 +592,6 @@ protected virtual void Dispose(bool disposing)
                     this.indexParsingThread.Dispose();
                     this.indexParsingThread = null;
                 }
-
```
derrickstolee: Are we no longer responsible for disposing?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1152,11 +1116,14 @@ private void UpdatePlaceholders()
...
-                    HashSet<string> folderPlaceholders =
-                        GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories ?
-                        new HashSet<string>(placeholderFoldersListCopy.Select(x => x.Path), StringComparer.OrdinalIgnoreCase) :
-                        null;
+                    // A hash of the placeholders is only required if the platform expands directories
+                    HashSet<string> existingPlaceholders = null;
+                    if (GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories)
+                    {
+                        IEnumerable<string> allPlaceholders = placeholderFoldersListCopy.Select(x => x.Path)
+                            .Union(this.placeholderDatabase.GetAllFilePaths());
```
derrickstolee: nit: Please line up the `.Union()` with the `.Select()` above. This can also be done by putting `.Select()` on its own line.

GVFS/GVFS.Common/Database/SqliteDatabase.cs
kewillford: I did a perf run without the shared cache and everything was slower, reading files, enumerating, as well as git checkouts so keeping this on is definitely better.

GVFS/GVFS.Common/Database/GVFSDatabase.cs
kewillford: In looking a the code closer I doesn't appear that this could ever be `null` so I am removing the `?`

GVFS/GVFS.Common/PlaceholderListDatabase.cs
derrickstolee: > `Count()` is a part of the interface

I missed that. Thanks!

GVFS/GVFS.Common/Database/Placeholders.cs
kewillford: I also ran some testing using a simple locking strategy with one prepared connection and command and things were slowed down because of the lock.  The next experiment will be to have a pool of prepared connections and commands to remove the need for the lock.

GVFS/GVFS.UnitTests/Common/Database/PlaceholdersTableTests.cs
kewillford: Yeah basically that the constructor isn't doing any work that needs to be tested so that if some functionality is added to it we will have a test that will need to be updated.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: Nope. This class now takes the interface in the constructor and that interface isn't disposable and the class `PlaceholderTable` isn't disposable either since it just uses the connections that the `GVFSDatabase` manages.  The `GVFSDatabase` class is the only one that needs to be disposed in `InProcessMount.UnmountAndStopWorkingDirectoryCallbacks`

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        private void Initialize()
+        {
+            IGVFSConnectionPool connectionPool = this;
+            using (IDbConnection connection = connectionPool.GetConnection())
+            using (IDbCommand command = connection.CreateCommand())
+            {
+                command.CommandText = "PRAGMA journal_mode=WAL;";
+                command.ExecuteNonQuery();
```
jamill: Do we need to handle the `InvalidOperationException` these commands can throw?

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                return;
+            }
+
+            this.disposed = true;
+            this.connectionPool.CompleteAdding();
+            while (!this.connectionPool.IsCompleted && this.connectionPool.TryTake(out IDbConnection connection))
+            {
+                connection.Dispose();
+            }
```
jamill: Do we need to dispose the `connectionPool`? The documentation on [msdn](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1?view=netframework-4.8) seems to indicate that we should.

GVFS/GVFS.Common/Database/PlaceholdersTable.cs
mjcheetham: `PlaceholderData` is a reference type and is being heap allocated. The `PlaceholderType PathType` member is a property, which will auto-create a private backing field.

> [..] is ~9MB that we save.

I'm not sure I know enough about how the CLR allocates to say if this would be true. Are instance fields packed or aligned to some n-byte boundary? In which order (does it matter)?

```csharp
class PlaceholderData
{
    private byte pathType;
    private int howAmIAligned; // 1 byte in or 4? 8?

    public byte PathType { get => pathType; set => pathType = value; }
}
```

kewillford:
@jamill for the exception, I was thinking of having a `GVFSDatabaseException` and simply catch and rethrow.  This is what was being done in the `PlaceholderListDatabase` so that if needed the code could catch that one exception instead of every possible exception that might happen while interacting with the GVFSDatabase.  What are your thoughts?

jamill:
> What are your thoughts?

That seems reasonable. We also had discussed having a 'GVFSException' that all other application exceptions could inherent from, so there would be a single base exception type for any exception we throw, and a single except type for us to catch. In that case, 'GVFSDatabaseException' could subclass 'GVFSException'.

I don't think any of this precludes just adding your exception type now, and I don't think we would make use of that distinction.

derrickstolee:
@kewillford What's the status here. Are you still doing investigations? I think this is good to go, and it's probably best to let it simmer in CI before a release.

Cc: @jrbriggs, @jamill in case they know the reason.

kewillford:
@derrickstolee, @wilbaker was going to take another look today.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace GVFS.Common
 {
-    public class PlaceholderListDatabase : FileBasedCollection, IPlaceholderDatabase
+    public class PlaceholderListDatabase : FileBasedCollection, IPlaceholderCollection
```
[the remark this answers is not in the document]
kewillford: Yes, it is only for the upgraders now and I thought about renaming it to `DeprecatedPlaceholderListDatabase` but `Legacy` should work as well.  I will only rename it for now and add an item to the back log to clean up all the methods and unit tests for this in a future PR if that works for you.

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -550,7 +552,21 @@ private void MountAndStartWorkingDirectoryCallbacks(CacheServerInfo cache)
                 this.tracer.RelatedInfo("Git status cache is not enabled");
             }
 
-            this.fileSystemCallbacks = this.CreateOrReportAndExit(() => new FileSystemCallbacks(this.context, this.gitObjects, RepoMetadata.Instance, virtualizer, gitStatusCache), "Failed to create src folder callback listener");
+            this.gvfsDatabase = this.CreateOrReportAndExit(() => new GVFSDatabase(this.context.FileSystem, this.context.Enlistment.EnlistmentRoot, new SqliteDatabase()), "Failed to create database connection");
```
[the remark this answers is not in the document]
kewillford: In my testing on the perf machine which has 36 cores it was only using 4 connections so I'm not sure we should open and keep around a bunch of connections that we will never use.  A new connection will get created and added back into the pool if there aren't any available to use so I'm not sure creating a bunch at the beginning has much benefit.  If we only set this to 1 but the projection change needs more it will expand the pool anyways so I found 5 to be a good number.  Let me know if you feel differently.

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        private void Insert(PlaceholderData placeholder)
+        {
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    command.CommandText = "INSERT OR REPLACE INTO Placeholder (path, pathType, sha) VALUES (@path, @pathType, @sha);";
+                    command.AddParameter("@path", DbType.String, placeholder.Path);
```
wilbaker: From what I've see in SO posts I *think* SQLite is case-sensitive.  Do we need to worry about duplicate entries being added, or is there something about how VFS4G uses the database that prevents that?

GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
```diff
@@ -181,8 +184,314 @@
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="..\GVFS.UnitTests\**\*.cs">
-      <Link>NetCore\%(RecursiveDir)\%(Filename)%(Extension)</Link>
```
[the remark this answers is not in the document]
kewillford: >Why did this recursive entry have to get broken out to all of the individual files?

VS goodness? :)  I will revert these changes.

GVFS/GVFS.Mount/InProcessMount.cs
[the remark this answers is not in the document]
wilbaker: > In my testing on the perf machine which has 36 cores it was only using 4 connections so I'm not sure we should open and keep around a bunch of connections that we will never use.

In those tests how many threads were creating files in parallel?

> A new connection will get created and added back into the pool if there aren't any available to use so I'm not sure creating a bunch at the beginning has much benefit.

That's a good point, it would be a one time hit as new connections are added and it looks like from that point on the connections stay around?  If that's the case and you've seen good perf with 5 then I agree there's no need to change this.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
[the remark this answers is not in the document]
wilbaker: > I will only rename it for now and add an item to the back log to clean up all the methods and unit tests for this in a future PR if that works for you.

That sounds good, thanks!

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1155,11 +1119,15 @@ private void UpdatePlaceholders()
                 int deleteFolderPlaceholderAttempted = 0;
                 using (BlobSizes.BlobSizesConnection blobSizesConnection = this.blobSizes.CreateConnection())
                 {
-                    // A hash of the folder placeholders is only required if the platform expands directories
-                    HashSet<string> folderPlaceholders =
-                        GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories ?
-                        new HashSet<string>(placeholderFoldersListCopy.Select(x => x.Path), StringComparer.OrdinalIgnoreCase) :
-                        null;
+                    // A hash of the placeholders is only required if the platform expands directories
```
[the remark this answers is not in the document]
kewillford: Will add a comment.

GVFS/GVFS.UnitTests.Windows/GVFS.UnitTests.Windows.csproj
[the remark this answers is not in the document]
wilbaker: I think this might happen if you try to add a file that lives in `GVFS.UnitTests` into `GVFS.UnitTests.Windows.csproj ` directly rather than adding it to `GVFS.UnitTests` and then reloading the `GVFS.UnitTests.Windows` project.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1155,11 +1119,15 @@ private void UpdatePlaceholders()
...
-                        GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories ?
-                        new HashSet<string>(placeholderFoldersListCopy.Select(x => x.Path), StringComparer.OrdinalIgnoreCase) :
-                        null;
+                    // A hash of the placeholders is only required if the platform expands directories
+                    HashSet<string> existingPlaceholders = null;
+                    if (GVFSPlatform.Instance.KernelDriver.EnumerationExpandsDirectories)
+                    {
+                        IEnumerable<string> allPlaceholders = placeholderFoldersListCopy
+                            .Select(x => x.Path)
+                            .Union(this.placeholderDatabase.GetAllFilePaths());
```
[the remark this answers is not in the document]
kewillford: >Do we have to get all placeholders this way for correctness?

No

>Or could we get get all of the entries out of the database?

We could but we already got all the folder placeholders and they have not changed since only the file placeholders were processed up to this point and if there are are a significant number of folder placeholders it will be slower to get them all from the database again.  I will add some comments here to help explain this.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/DiskLayoutUpgradeTests.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.IO;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
+{
+    public abstract class DiskLayoutUpgradeTests : TestsWithEnlistmentPerTestCase
+    {
+        protected const string PlaceholderListDatabaseDelimiter = "\0";
+
+        protected static readonly string PlaceholderListDatabaseContent = $@"A .gitignore{PlaceholderListDatabaseDelimiter}E9630E4CF715315FC90D4AEC98E16A7398F8BF64
+A Readme.md{PlaceholderListDatabaseDelimiter}583F1A56DB7CC884D54534C5D9C56B93A1E00A2B
```
wilbaker: Use `GVFSHelpers.PlaceholderFieldDelimiter` and remove `PlaceholderListDatabaseDelimiter`?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1658,6 +1605,7 @@ private bool RemoveFolderPlaceholderIfEmpty(IPlaceholderData placeholder)
                     break;
 
                 case FSResult.FileOrPathNotFound:
+                    this.placeholderDatabase.Remove(placeholder.Path);
```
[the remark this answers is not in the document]
kewillford: I had changed this at one point but must have reverted.  Let me make the change again and make sure all the tests are still passing.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/DiskLayoutUpgradeTests.cs
```diff
@@ -0,0 +1,133 @@
...
+            placeholderLines.ShouldContain(x => x == this.PartialFolderPlaceholderString("Scripts"));
+            placeholderLines.ShouldContain(x => x == this.PartialFolderPlaceholderString("GVFS"));
+            placeholderLines.ShouldContain(x => x == this.PartialFolderPlaceholderString("GVFS\\GVFS.Common"));
+            placeholderLines.ShouldContain(x => x == this.PartialFolderPlaceholderString("GVFS\\GVFS.Common\\Git"));
+            placeholderLines.ShouldContain(x => x == this.PartialFolderPlaceholderString("GVFS\\GVFS.Tests"));
+        }
+
+        protected void WriteOldPlaceholderListDatabase()
+        {
+            this.fileSystem.WriteAllText(Path.Combine(this.Enlistment.DotGVFSRoot, GVFSHelpers.PlaceholderListFile), PlaceholderListDatabaseContent);
```
wilbaker: On Mac the placeholder list would have had `/` path separators, but when the `MountUpgradesPlaceholderListDatabaseToSQLite` Mac function test calls this method a file will be written with `\` separators.

I think that we might need to run `GVFSHelpers.PlaceholderListFile` through a `Replace` that swaps the separators.

I don't think the other methods that use paths are called on Mac, but it would be worth confirming.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/MacDiskLayoutUpgradeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            this.Enlistment.MountGVFS();
+            this.Enlistment.UnmountGVFS();
+
+            // Validate the placeholders are in the SQLite placeholder database now
+            string placeholderDatabasePath = Path.Combine(this.Enlistment.DotGVFSRoot, TestConstants.Databases.VFSForGit);
+            placeholderDatabasePath.ShouldBeAFile(this.fileSystem);
+            string[] lines = GVFSHelpers.GetAllSQLitePlaceholdersAsString(placeholderDatabasePath).Split(new char[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+            lines.Length.ShouldEqual(10);
+            this.PlaceholderDatabaseShouldIncludeCommonLines(lines);
+            lines.ShouldContain(x => x == this.PartialFolderPlaceholderString("GVFS\\GVFS.Tests\\Properties"));
```
wilbaker: Similar question here, is it expected that we'll have `\` on Mac?

Does that cause any issues when looking up paths to update placeholders during projection changes?

GVFS/GVFS.Mount/InProcessMount.cs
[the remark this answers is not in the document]
kewillford: >In those tests how many threads were creating files in parallel?

Creating temp files runs 
`int numThreads = Math.Max(8, Environment.ProcessorCount);`

And the recursive read does
```
string filereadArgs = string.Format(
	"{0} --threads={1}",
	Path.Combine(this.WorkingDirectory, path),
	Environment.ProcessorCount * 1.5);
```

GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
```diff
@@ -69,11 +74,15 @@ public void BackgroundOperationCountMatchesBackgroundFileSystemTaskRunner()
...
             }
+
+            mockPlaceholderDb.VerifyAll();
         }
 
         [TestCase]
         public void GetMetadataForHeartBeatDoesNotChangeEventLevelWhenNoPlaceholderHaveBeenCreated()
         {
+            Mock<IPlaceholderCollection> mockPlaceholderDb = new Mock<IPlaceholderCollection>(MockBehavior.Strict);
+            mockPlaceholderDb.Setup(x => x.GetCount()).Returns(0);
```
wilbaker: In the other tests you set `GetCount` to return 1 rather than 0.  Was there any significance to using 0 here instead of 1?

GVFS/GVFS.UnitTests/Common/Database/GVFSDatabaseTests.cs
```diff
@@ -0,0 +1,131 @@
...
+    public class GVFSDatabaseTests
+    {
+        [TestCase]
+        public void ConstructorTest()
+        {
+            this.TestGVFSDatabase(null);
+        }
+
+        [TestCase]
+        public void ConstructorThrowsGVFSDatabaseException()
```
wilbaker: Also tag this test with `[Category(CategoryConstants.ExceptionExpected)]`?

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            if (!itemWasAdded)
+            {
+                connection.Dispose();
+            }
+        }
+
+        private void Initialize()
+        {
+            IGVFSConnectionPool connectionPool = this;
```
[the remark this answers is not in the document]
kewillford: Because the interface implementation is explicit
`IDbConnection IGVFSConnectionPool.GetConnection()`
You have to use the interface to call the method.

I guess we could have a private method `InternalGetConnection` or `GetConnectionImplementation`? and have the explicit interface call that method.  I can do that if it would make more sense.

GVFS/GVFS.UnitTests/Common/Database/PlaceholderTableTests.cs
```diff
@@ -0,0 +1,668 @@
...
+            mockConnection.Setup(x => x.CreateCommand()).Returns(mockCommand.Object);
+
+            PlaceholderTable.CreateTable(mockConnection.Object);
+            mockCommand.VerifyAll();
+            mockConnection.VerifyAll();
+        }
+
+        [TestCase]
+        [Category(CategoryConstants.ExceptionExpected)]
+        public void CreateTableThrowsExceptionNotWrappedInGVFSDatabaseException()
```
wilbaker: Oh it looks like it is expected the Create doesn't throw `GVFSDatabaseException`

GVFS/GVFS.Common/Database/PlaceholderTable.cs
wilbaker: Another concern, calls to `Remove` using the current case of the file might fail to find existing entries that only differ by case.

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -0,0 +1,223 @@
...
+            this.connectionPool = connectionPool;
+        }
+
+        public static void CreateTable(IDbConnection connection)
+        {
+            using (IDbCommand command = connection.CreateCommand())
+            {
+                command.CommandText = "CREATE TABLE IF NOT EXISTS [Placeholder] (path TEXT PRIMARY KEY, pathType TINYINT NOT NULL, sha char(40) ) WITHOUT ROWID;";
+                command.ExecuteNonQuery();
+            }
```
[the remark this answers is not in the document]
wilbaker: I see now based on the unit tests this is expected ... maybe this is OK? What was the reason for not using `GVFSDatabaseException` here?

GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
kewillford: To match the check below
`metadata.ShouldContain("PlaceholderCount", 0);`

GVFS/GVFS.Common/Database/PlaceholderTable.cs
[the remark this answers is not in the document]
kewillford: Didn't want the exception to be double wrapped because this is called from the `Initialize` of `GVFSDatabase` and that is already wrapped in a `try/catch` in the constructor.  The catch and throw of the `GVFSDatabaseException` here would be caught again and wrapped in another `GVFSDatabaseException` in the constructor of `GVFSDatabase`

GVFS/GVFS.Common/Database/GVFSDatabase.cs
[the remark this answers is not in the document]
wilbaker: > I can do that if it would make more sense.

I think the current code is good, I missed that the implementation was explicit, thanks!

GVFS/GVFS.Common/Database/PlaceholderTable.cs
wilbaker: As discussed in a different comment, path in the table is now case insensitive, marking this as resolved.

jrbriggs: @wilbaker @kewillford does it need to be case-sensitive on OSes where the underlying filesystem is case-sensitive?

kewillford: It will but we have so many collections that are `OrdinalIgnoreCase` that there would need to be a bunch of code changes to make VFSForGit case sensitive on those file systems.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/DiskLayoutUpgradeTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            string target = Path.Combine(this.Enlistment.EnlistmentRoot, "GVFS", "GVFS", "GVFS.UnitTests");
+            string link = Path.Combine(this.Enlistment.RepoRoot, "UnitTests");
+            ProcessHelper.Run("CMD.exe", "/C mklink /J " + link + " " + target);
+            target = Path.Combine(this.Enlistment.EnlistmentRoot, "GVFS", "GVFS", "GVFS.Installer");
+            link = Path.Combine(this.Enlistment.RepoRoot, "Installer");
+            ProcessHelper.Run("CMD.exe", "/C mklink /D " + link + " " + target);
+        }
+
+        protected void PlaceholderDatabaseShouldIncludeCommonLinesForUpgradeTestIO(string[] placeholderLines)
```
wilbaker: Do you think the `protected` methods that are only called in `WindowsDiskLayoutUpgradeTests` should be moved to `WindowsDiskLayoutUpgradeTests`?

Since they're using `\` I don't think they would ever apply to other platforms.

(No need to block the PR on this)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/DiskLayoutUpgradeTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        protected string FilePlaceholderString(params string[] pathParts)
+        {
+            return $"{Path.Combine(pathParts)}{GVFSHelpers.PlaceholderFieldDelimiter}0{GVFSHelpers.PlaceholderFieldDelimiter}";
+        }
+
+        protected string PartialFolderPlaceholderString(params string[] pathParts)
+        {
+            return $"{Path.Combine(pathParts)}{GVFSHelpers.PlaceholderFieldDelimiter}1{GVFSHelpers.PlaceholderFieldDelimiter}";
```
wilbaker: Minor: A named constant here for `1` would make it easier to understand why it's part of the string

turbonaitis:
I'm a bit late to the party, but do we have some performance numbers here? How was the old placeholder DB performing and by how much the SQLite solution improves it?

kewillford:
@turbonaitis here are some number with ~1 million placeholders
#### Old
Load and get all placeholders: `41.253 s`
Write and update placeholders: `21.940 s`

#### SQLite
Load and get all placeholders: `2.053 s`
Write and update placeholders: `23.940 s`

turbonaitis:
These look great, thanks for sharing!

PR #1143, opened 2019-05-10
Add .vsconfig file to declare required VS components

mjcheetham:
Declare the following VS components as required for this solution to build:

 - Microsoft.Component.MSBuild
 - Microsoft.VisualStudio.Workload.NativeDesktop
 - Microsoft.VisualStudio.Workload.ManagedDesktop
 - Microsoft.VisualStudio.Workload.NetCoreTools
 - Microsoft.Net.Core.Component.SDK.2.1
 - Microsoft.VisualStudio.Component.VC.v141.x86.x64
 - Microsoft.Net.Component.4.6.1.TargetingPack
 - Microsoft.Net.Component.4.6.1.SDK

Note that the Windows 10 SDK 10240 is also required, but this component is no longer shipped with VS2019.

https://devblogs.microsoft.com/setup/configure-visual-studio-across-your-organization-with-vsconfig/

### DIFF
--- a/.vsconfig
+++ b/.vsconfig
@@ -1,0 +1,13 @@
+{
+  "version": "1.0",
+  "components": [
+    "Microsoft.Component.MSBuild",
+    "Microsoft.VisualStudio.Workload.NativeDesktop"
+    "Microsoft.VisualStudio.Workload.ManagedDesktop",
+    "Microsoft.VisualStudio.Workload.NetCoreTools",
+    "Microsoft.Net.Core.Component.SDK.2.1",
+    "Microsoft.VisualStudio.Component.VC.v141.x86.x64",
+    "Microsoft.Net.Component.4.6.1.TargetingPack",
+    "Microsoft.Net.Component.4.6.1.SDK",
+  ]
+}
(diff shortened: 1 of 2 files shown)


.vsconfig
```diff
@@ -0,0 +1,13 @@
+{
+  "version": "1.0",
+  "components": [
```
derrickstolee: Isn't there a Windows SDK we require?

mjcheetham: [RTF~M~ PR Description](https://github.com/microsoft/VFSForGit/pull/1143#issue-277754551) 😝 

> Note that the Windows 10 SDK 10240 is also required, but this component is no longer shipped with VS2019.

vsconfig only works in VS2019+, and the SDK we require 10240 is no longer included in the VS installation.

jamill:
Thanks!

mjcheetham:
> I think this will be helpful. One thought - I can see this file being easy to miss when we update our dependencies, is there anything we can do to help whoever updates dependencies to also update this file? For instance, maybe mention in the README that VS 2019 will prompt you to install missing dependencies? I suspect we will remember to update the readme when we update dependencies, and maybe seeing a sentence about it in there will remind people to update this as well.

Updated Readme.

.vsconfig
derrickstolee: sorry I missed that. the v141  tools are not included (by default) but can be downloaded this way. Can the windows SDK stuff download, too?

mjcheetham: No. The v141 VC toolchain is available in VS2019 (just not selected by default).
image

The only available Windows SDKs in VS2019's installer are:
image

DevDiv have clearly decided to only support SDKs for currently supported Windows 10 versions. We must still support RS1 (apparently).

PR #1114, opened 2019-05-07
[Not to merge] User notification mac

alameenshah:
#### High level description

This PR adds the ability for `VFSForGit` to display native desktop notifications on the Mac. A new `launchd` `User Agent` app has been added - `VFSForGit.app`. The `User Agent` gets launched automatically during user login. It creates a UNIX domain socket and listens for incoming requests for notification display. Any client `VFSForGit` process that needs to notify the user (example when an upgrade is available, or when a repository mount failed), opens a communication pipe with the `User Agent` and sends it the text of the notification message. `User Agent` then uses Mac native `NSUserNotification` API to display the actual notification alert window to User.

The following sequence diagram demonstrates the flow of control as `GVFS.Service` tries to notify the user of results of auto-mount action.
UserNotification

#### `User Agent` app UI elements
The `User Agent` displays a `NSStatusItem` which is equivalent of the `tray` icon in `Windows`. Clicking on the `NSStatusItem` would show a drop down menu, with currently mounted repositories and an `About VFSForGit` link. The `About VFSForGit` link opens an about box which should currently installed version numbers for `VFSForGit` and `Git`. See the screenshots below

| UI Element  | Screenshot |
| ------------- | ------------- |
| Tray icon  | Screen Shot 2019-05-07 at 4 03 39 PM|
| Drop down menu  | Screen Shot 2019-05-07 at 4 04 14 PM |
| About Window  | Screen Shot 2019-05-07 at 4 04 25 PM |
| Sample Notification  | Screen Shot 2019-05-07 at 4 12 08 PM |

#### Files changed
- installer scripts (`CreateMacInstaller.sh`, `postinstall` & `preinstall`) to install `User Agent` as a `launchd` agent
- `BuildGVFSForMac.sh` to build the new `User Agent`
- `GVFS.Service/Handlers/NotificationHandler.Mac.cs` to connect to `User Agent` and send the notification request.
- UT cases are updated to use Mock NotificationHandler.

#### Open questions
- What should be the name of the `User Agent`. Right now it is named `VFSForGit.app`
- Directory structure of `User Agent` project. Right now I've kept it inside `GVFS/UserNotification`. It has a `Mac` subdirectory assuming we'll create a `Windows` subdirectory for the `Windows` `Service` in future.
- Contents of the `User Agent` UI. Right now it shows mounted repositories and link to about box.
- The notifications displayed are informational. User cannot trigger any action by clicking on it.
- Install location for `VFSForGit.app`.

#### Upcoming PRs
[...]

PR #1146, opened 2019-05-10
Platform native notifications - PR_1

alameenshah:
#### Boiler plate code for native Xcode app

This is the first in a series of upcoming changes to add platform
native notifications. This commit adds a native Mac app that shows
a status menu in the macOS menu bar. It contains the following
changes
 - Boilerplate code generated by Xcode for native apps
 - VFSForGit Icon asset files
 - UI .xib files
 - Placeholder source code files for upcoming changes.
 - Updated buildscript to build the app

The app will get installed, but will not get automatically launched.
It displays hard coded version strings in its About box and also
does not perform any useful function.

#### Reference
For a better understanding of the bigger picture refer [this PR](https://github.com/microsoft/VFSForGit/pull/1114).


wilbaker:
> The app will get installed, but will not get automatically launched.

When you say "will not get automatically launched" does that mean customers will not see the app unless they run it specifically themselves?

GVFS/GVFS.Notifications/VFSForGit.Mac/org.vfsforgit.usernotification.plist
```diff
@@ -0,0 +1,36 @@
...
+            that are set by launchd.
+            -->
+            <key>PATH</key>
+            <string>/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+        </dict>
+        <key>Label</key>
+        <string>org.vfsforgit.usernotification</string>
+        <key>ProgramArguments</key>
+        <array>
+            <string>/usr/local/vfsforgit/VFSForGit.app/Contents/MacOS/VFSForGit</string>
```
wilbaker: During the last offline discussion there was a question as to wether or not we'd have to use the name `VFSForGit`

- Are we required to use this name, or can it be overridden in a plist file?
- If our App name will always be used, should we include spaces in the name like `Microsoft Outlook.app`  `Visual Studio.app` do?

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -56,6 +56,9 @@ dotnet build $VFS_SRCDIR/GVFS.sln --runtime osx-x64 --framework netcoreapp2.1 --
 NATIVEDIR=$VFS_SRCDIR/GVFS/GVFS.Native.Mac
 xcodebuild -configuration $CONFIGURATION -workspace $NATIVEDIR/GVFS.Native.Mac.xcworkspace build -scheme GVFS.Native.Mac -derivedDataPath $VFS_OUTPUTDIR/GVFS.Native.Mac || exit 1
 
+USERNOTIFICATIONDIR=$VFS_SRCDIR/GVFS/GVFS.Notifications/VFSForGit.Mac
+xcodebuild -configuration $CONFIGURATION -project $USERNOTIFICATIONDIR/VFSForGit.xcodeproj build -scheme VFSForGit -derivedDataPath $VFS_OUTPUTDIR/GVFS.Notifications/VFSForGit.Mac || exit 1
```
wilbaker: Should this project be built as part of `GVFS.Native.Mac.xcworkspace`?

`GVFS.Native.Mac.xcworkspace` currently includes all of the native Mac VFS4G projects:

```
GVFS/GVFS.PostIndexChangedHook/GVFS.PostIndexChangedHook.Mac.xcodeproj/project.pbxproj
GVFS/GVFS.ReadObjectHook/GVFS.ReadObjectHook.Mac.xcodeproj/project.pbxproj
GVFS/GVFS.VirtualFileSystemHook/GVFS.VirtualFileSystemHook.Mac.xcodeproj/project.pbxproj
```
cc: @nickgra @jrbriggs @jeschu1 @jamill for their thoughts as well

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/GVFSTask.m
```diff
@@ -0,0 +1,26 @@
...
+
+@end
+
+// Placeholder class to get VFSForGit info that
+// gets displayed in the Status Item.
+@implementation GVFSTask
+
+- (NSString *) VFSForGitVersion
+{
+    // TODO : Get actual VFSForGit version
```
wilbaker: I recommend either addressing these TODOs in this PR or returning an empty string.  

I don't think we want to ship an application that will report its version incorrectly (even if we don't load it by default)

alameenshah:
> > The app will get installed, but will not get automatically launched.
> 
> When you say "will not get automatically launched" does that mean customers will not see the app unless they run it specifically themselves?

Yes - that is the behavior in this PR, because it offers no value. Once I get the Versions show, I'll enable auto launch.

GVFS/GVFS.Notifications/VFSForGit.Mac/org.vfsforgit.usernotification.plist
alameenshah: Am not able to override the name in info.plist file. Added spacing in the name - new name is `VFS For Git.app`.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/GVFSTask.m
alameenshah: Agree. Updated to return empty strings for version.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/AboutWindowController.m
```diff
@@ -0,0 +1,20 @@
+		#import "GVFSTask.h"
```
nickgra: ws

Scripts/Mac/BuildGVFSForMac.sh
nickgra: We got rid of the original `PrjFS.xcworkspace` we had for the kext code to move to a single .xcodeproj to work around some clunkiness around how workspaces work. I think `GVFS.Native.Mac.xcworkspace` was set up the in the way that it is because we were just copying what we did for the `PrjFS.xcworkspace`, I can't think of a good reason for it to be structured as a workspace.

I think we should convert the existing `GVFS.Native.Mac` workspace to a project (and then include this project as well), although that can be in a separate follow up. 

Adding @pmj as well since had opinions on this in the past.

nickgra:
> When you say "will not get automatically launched" does that mean customers will not see the app unless they run it specifically themselves?

What I believe this PR does is that it will 'install' the application by copying it to `/usr/local/vfsforgit`, but it will not configure it to be launched automatically via `launchd`. In general, most people don't go looking for .app files outside of `/Applications` so I think this is a fine interim state for us to be in.

Scripts/Mac/BuildGVFSForMac.sh
pmj: Xcode Workspaces are very loose aggregations of Projects. The only thing you really can do with them is create super-project build Schemes. One of the main downsides is that they don't allow cross-references between projects - e.g. you can't have a target in one project be a build dependency for a target in another project if the only connection between those projects is that they're contained in the same workspace.

You *can* however nest projects - in this case, the outer project can depend on targets from the inner project. This isn't exactly the same as having the targets in the same project of course - project-wide properties aren't propagated to sub-project targets, and the sub-project's targets can't depend on the super-project's targets.

Without looking at the merits of the different approaches for this specific case, my order of preference is normally:

 1. Add new "projects" (apps, tools, libraries, etc.) as *targets* to an existing Xcode Project.
 2. If the new thing really doesn't fit with an existing Xcode Project but uses something from an existing Xcode Project, then nest the existing Project inside a new Project containing the new target. The new target can then e.g. link against libraries built by the existing Project.
 3. If the new thing is only extremely loosely related to the existing Xcode Project, but you want to be able to build everything in one go from within Xcode, create a new, unrelated Project for the new thing, then nest both it and the existing Project inside a newly created umbrella project. The umbrella project contains an aggregate build target which simply lists targets from both sub-projects as dependencies.

I'm not aware of any scenario where you'd prefer a workspace over any of the above.

pmj:
While I'm here…

> @nickgra : What I believe this PR does is that it will 'install' the application by copying it to `/usr/local/vfsforgit`, but it will not configure it to be launched automatically via `launchd`. In general, most people don't go looking for .app files outside of `/Applications` so I think this is a fine interim state for us to be in.

Putting `.app` bundles below `/usr/` isn't really the *done thing*. If you want an app but not below `/Applications/`, the place to put it is normally below `/Library/Application Support/YourProduct/`. I certainly don't recommend putting an .app in `/usr/local/vfsforgit`.

Also, if by `launchd` you mean `LaunchAgents` or `LaunchDaemons` - I feel I should point out you don't launch *apps* via `LaunchAgents`; agents don't have a UI (unless they're pre-login agents, which let you add a UI to the login screen), they're basically daemons that run in the context of a specific user. Automatically launched *apps* (with UI) should be installed as *Login Items*, see documentation here: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLoginItems.html

nickgra:
> While I'm here…
> 
> > @nickgra : What I believe this PR does is that it will 'install' the application by copying it to `/usr/local/vfsforgit`, but it will not configure it to be launched automatically via `launchd`. In general, most people don't go looking for .app files outside of `/Applications` so I think this is a fine interim state for us to be in.
> 
> Putting `.app` bundles below `/usr/` isn't really the _done thing_. If you want to install an app but not below `/Applications/`, the place to put it is normally below `/Library/Application Support/YourProduct/`. I certainly don't recommend putting an .app in `/usr/local/vfsforgit`.
> 
> Also, if by `launchd` you mean `LaunchAgents` or `LaunchDaemons` - I feel I should point out you don't launch _apps_ via `LaunchAgents`; agents don't have a UI (unless they're pre-login agents, which let you add a UI to the login screen), they're basically daemons that run in the context of a specific user. Automatically launched _apps_ (with UI) should be installed as _Login Items_, see documentation here: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLoginItems.html

Thanks for calling that out @pmj! I think we should move the .app file to `/Library/Application Support/VFSForGit` as you propose. 

I feel like less strongly about Login Items vs launchd plists, @alameenshah can you look into programmatic ways we could add the .app to Login Items?

Scripts/Mac/BuildGVFSForMac.sh
nickgra: Thanks for the detailed answer Phil, I agree with your assessment that we could convert the workspace into a project. I've opened #1167 to track the conversion.

@alameenshah I think we can leave this as is for now and rework it as part of 1167.

nickgra:
I was looking at one of the build agents (GVFS-MACPRO-1) and I saw that the GUI app was installed and running in the top menu bar. Perhaps something was missed in the cleanup/uninstall steps?

Before we merge this, can you try running the functional tests and remote into the machine, validate that the icon appears, then validate that it gets uninstalled correctly at the end of the job?

alameenshah:
> > While I'm here…
> > > @nickgra : What I believe this PR does is that it will 'install' the application by copying it to `/usr/local/vfsforgit`, but it will not configure it to be launched automatically via `launchd`. In general, most people don't go looking for .app files outside of `/Applications` so I think this is a fine interim state for us to be in.
> > 
> > 
> > Putting `.app` bundles below `/usr/` isn't really the _done thing_. If you want to install an app but not below `/Applications/`, the place to put it is normally below `/Library/Application Support/YourProduct/`. I certainly don't recommend putting an .app in `/usr/local/vfsforgit`.
> > Also, if by `launchd` you mean `LaunchAgents` or `LaunchDaemons` - I feel I should point out you don't launch _apps_ via `LaunchAgents`; agents don't have a UI (unless they're pre-login agents, which let you add a UI to the login screen), they're basically daemons that run in the context of a specific user. Automatically launched _apps_ (with UI) should be installed as _Login Items_, see documentation here: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLoginItems.html
> 
> Thanks for calling that out @pmj! I think we should move the .app file to `/Library/Application Support/VFSForGit` as you propose.
> 
> I feel like less strongly about Login Items vs launchd plists, @alameenshah can you look into programmatic ways we could add the .app to Login Items?

1. Updated the install location to `/Library/Application Support/VFS For Git` directory.
2. Login items vs launchd agent - A few issues i see with Login Items - user can disable it/quit it, offers no ability to relaunch on crash, would be harder for VFSForGit installer to launch the app automatically in each logged-in user context (launchd offers bootstrap). Examples of other Microsoft `apps` that are Launchd Agents - Office updater `Microsoft AutoUpdate.app`, OneDrive updater - `StandaloneUpdater.app`.

nickgra:
> > > While I'm here…
> > > > @nickgra : What I believe this PR does is that it will 'install' the application by copying it to `/usr/local/vfsforgit`, but it will not configure it to be launched automatically via `launchd`. In general, most people don't go looking for .app files outside of `/Applications` so I think this is a fine interim state for us to be in.
> > > 
> > > 
> > > Putting `.app` bundles below `/usr/` isn't really the _done thing_. If you want to install an app but not below `/Applications/`, the place to put it is normally below `/Library/Application Support/YourProduct/`. I certainly don't recommend putting an .app in `/usr/local/vfsforgit`.
> > > Also, if by `launchd` you mean `LaunchAgents` or `LaunchDaemons` - I feel I should point out you don't launch _apps_ via `LaunchAgents`; agents don't have a UI (unless they're pre-login agents, which let you add a UI to the login screen), they're basically daemons that run in the context of a specific user. Automatically launched _apps_ (with UI) should be installed as _Login Items_, see documentation here: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLoginItems.html
> > 
> > 
> > Thanks for calling that out @pmj! I think we should move the .app file to `/Library/Application Support/VFSForGit` as you propose.
> > I feel like less strongly about Login Items vs launchd plists, @alameenshah can you look into programmatic ways we could add the .app to Login Items?
> 
> 1. Updated the install location to `/Library/Application Support/VFS For Git` directory.
> 2. Login items vs launchd agent - A few issues i see with Login Items - user can disable it/quit it, offers no ability to auto-relaunch on crash, would be harder for VFSForGit installer to launch the app automatically in each logged-in user context (launchd offers bootstrap). Examples of other Microsoft `apps` that are Launchd Agents - Office updater `Microsoft AutoUpdate.app`, OneDrive updater - `StandaloneUpdater.app`.

Sounds good, thanks!

alameenshah:
> Approved, assuming that we validate the behavior/installation/uninstallation of the .app on the build agents.

Thank you Nick. Updated the un-installer to remove `VFS For Git.app`. Also the app in the build agent got launched during the Function Testing of [the POC PR](https://github.com/microsoft/VFSForGit/pull/1114). Have cleanup the build agent.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
```diff
@@ -0,0 +1,59 @@
...
+        <customObject id="-3" userLabel="Application" customClass="NSObject"/>
+        <customObject id="Voe-Tx-rLC" customClass="AppDelegate">
+            <connections>
+                <outlet property="Window" destination="QvC-M9-y7g" id="ykX-6N-Ma5"/>
+            </connections>
+        </customObject>
+        <customObject id="YLy-65-1bz" customClass="NSFontManager"/>
+        <menu title="Main Menu" systemMenu="main" id="AYu-sK-qS6">
+            <items>
+                <menuItem title="VFSForGit" id="1Xt-HY-uBw">
```
mjcheetham: Should we not have spaces in our name here? Is `title="VFSForGit"` a user visible string?

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
```diff
@@ -0,0 +1,59 @@
...
+                    <menu key="submenu" title="VFSForGit" systemMenu="apple" id="uQy-DD-JDr">
+                        <items>
+                            <menuItem title="About VFSForGit" id="5kV-Vb-QxS">
+                                <modifierMask key="keyEquivalentModifierMask"/>
+                                <connections>
+                                    <action selector="orderFrontStandardAboutPanel:" target="-1" id="Exp-CZ-Vem"/>
+                                </connections>
+                            </menuItem>
+                            <menuItem isSeparatorItem="YES" id="VOq-y0-SEH"/>
+                            <menuItem title="Preferences…" keyEquivalent="," id="BOF-NM-1cW"/>
```
mjcheetham: ❤️  for the use of a real ellipsis character (…) ! 😁

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
```diff
@@ -0,0 +1,59 @@
...
+                            <menuItem isSeparatorItem="YES" id="wFC-TO-SCJ"/>
+                            <menuItem title="Quit VFSForGit" keyEquivalent="q" id="4sb-4s-VLi">
```
mjcheetham: Do we need to keep saying "VFSForGit" here in the menu? Could we not just say "Quit" (or at least add spaces to the text please)?

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
```diff
@@ -0,0 +1,59 @@
...
+                                    <action selector="terminate:" target="-1" id="Te7-pn-YzF"/>
+                                </connections>
+                            </menuItem>
+                        </items>
+                    </menu>
+                </menuItem>
+            </items>
+            <point key="canvasLocation" x="-66" y="25"/>
+        </menu>
+        <window title="VFSForGit" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" restorable="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" animationBehavior="default" id="QvC-M9-y7g">
```
mjcheetham: Not sure if these titles are read by VoiceOver (narrator)? Spaces in the name again if they are :)

GVFS/GVFS.Notifications/VFSForGit.Mac/org.vfsforgit.usernotification.plist
```diff
@@ -0,0 +1,36 @@
...
+        <dict>
+            <!--
+            Although VFSForGit.app gets launched in the GUI context of actively logged-in user, it
+            does not seem to inherit any user-specific environment variables (like user customized
+            $PATH specified in shell ~/.profile). VFSForGit needs /usr/local/bin to be in its 
+            search PATH environment, so specifying it here, along with rest of the default paths 
+            that are set by launchd.
+            -->
+            <key>PATH</key>
+            <string>/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
```
mjcheetham: Why are we needing the user's environment with /usr/local/bin in the $PATH for the GUI app? We probably shouldn't be relying on their PATH for lauching things? Why couldn't we just use the absolute paths to executables that we need (e.g., /usr/local/bin/vfsforgit ... /gvfs )

mjcheetham:
Could we see a screenshot of the menu based on the code in this PR? Thanks!

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
alameenshah: Updated

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
alameenshah: Added spaces. Usually apps show "Quit <app_name".

mjcheetham: Sure thing! Just checking. Better to be consistent with others :)

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/Base.lproj/MainMenu.xib
alameenshah: added spaces.

GVFS/GVFS.Notifications/VFSForGit.Mac/org.vfsforgit.usernotification.plist
alameenshah: In our processes we can use the absolute path names, no problem. But other tools like `git` (which we will use in the upcoming PR to get git version string) seems to need `PATH` environment setting. 

I will double check this (and remove this if not needed) in the upcoming PR when I start using `git` and also make the app an actual launchd agent.

alameenshah:
> Could we see a screenshot of the menu based on the code in this PR? Thanks!

Sure...
Screen Shot 2019-05-17 at 10 38 29 AM

mjcheetham:
> > Could we see a screenshot of the menu based on the code in this PR? Thanks!
> 
> Sure...
> Screen Shot 2019-05-17 at 10 38 29 AM

I'm glad that we've got icons that look great on retina screens :+1:

PR #1148, opened 2019-05-10
Fix handling of >>

jeschu1:
">>" will not call PreConvertToFull
If we get a modified event for a file that is a placeholder, add it to modifiedPaths.dat

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
@@ -345,10 +345,13 @@ BOOL APIENTRY DllMain( HMODULE hModule,
         }
 
         [TestCase, Order(9)]
-        public void WriteToHydratedFileAfterRemount()
+        public void AppendToHydratedFileAfterRemount()
         {
-            string virtualFilePath = this.Enlistment.GetVirtualPathTo("Test_EPF_WorkingDirectoryTests", "WriteToHydratedFileAfterRemount.cpp");
+            string fileToAppendEntry = "Test_EPF_WorkingDirectoryTests/WriteToHydratedFileAfterRemount.cpp";
+            string virtualFilePath = this.Enlistment.GetVirtualPathTo(fileToAppendEntry);
             string fileContents = virtualFilePath.ShouldBeAFile(this.fileSystem).WithContents();
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileToAppendEntry);
 
             // Remount
             this.Enlistment.UnmountGVFS();
@@ -356,6 +359,8 @@ BOOL APIENTRY DllMain( HMODULE hModule,
 
             string appendedText = "Text to append";
             this.fileSystem.AppendAllText(virtualFilePath, appendedText);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileToAppendEntry);
             virtualFilePath.ShouldBeAFile(this.fileSystem).WithContents(fileContents + appendedText);
         }
 
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -436,13 +436,20 @@ namespace GVFS.Platform.Mac
             this.Context.Tracer.RelatedError($"{nameof(MacFileSystemVirtualizer)}::{nameof(this.OnLogError)}: {errorMessage}");
         }
 
-        private void OnFileModified(string relativePath)
+        private void OnFileModified(string relativePath, bool isPlaceholderFile)
         {
             try
             {
                 if (Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath))
                 {
                     this.OnDotGitFileOrFolderChanged(relativePath);
+                }
+                else if (isPlaceholderFile == true)
+                {
+                    // The '>>' bash command does not send a KAUTH_VNODE_WRITE data event
+                    // To handle this, PrjFSLib checks if a file is a placeholder in the Modifed event and converts the file to
+                    // full after delivering this callback
+                    this.NotifyFilePreConvertToFull(relativePath);
                 }
             }
             catch (Exception e)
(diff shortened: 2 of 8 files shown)


GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -436,14 +436,18 @@ private void OnLogError(string errorMessage)
...
         {
             try
             {
                 if (Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath))
                 {
                     this.OnDotGitFileOrFolderChanged(relativePath);
                 }
+                else if (firstWrite == true)
+                {
+                    this.OnFilePreConvertToFull(relativePath);
```
jeschu1: @wilbaker quick question (no need to review the whole PR yet)

Is this sort of action of calling another callback from a callback, frowned upon?  I would think we'd want to do whatever logic OnFilePreConvertToFull did.  Alternatively I can easily add a new FileSystemTask.OperationType.OnModifiedFile and not do it this way.

wilbaker: I think in the case it's OK because the native library is about to convert the file to full (i.e. remove its placeholder attributes).

Could you add a comment here though explaining why we've found the current PreConvertToFull callback doesn't get us the coverage we need?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1057,6 +1057,7 @@ static PrjFS_Result HandleFileNotification(
         request->pid,
         request->procname,
         isDirectory,
+        placeholderFile,
         notificationType,
         nullptr /* destinationRelativePath */);
```
wilbaker: The code below (for PrjFS_NotificationType_PreConvertToFull) should also happen for file modified notifications when `placeholderFile` is true.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
jeschu1: Just a note to myself that this function needs to be updated to return a bool now.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
```diff
@@ -345,23 +345,28 @@ public void HydratingNestedFileUsesNameCaseFromRepo()
         }
 
         [TestCase, Order(9)]
-        public void WriteToHydratedFileAfterRemount()
+        public void AppendToHydratedFileAfterRemount()
```
jeschu1: I decided to use an existing test to verify the ">>".
I can also create a new one.
My thought process was, this does the work already and let's not increase the time of the functional tests.  Let me know if you the tradeoffs aren't fair here.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1057,10 +1057,13 @@ static PrjFS_Result HandleFileNotification(
...
     
-    if (result == 0 && placeholderFile && PrjFS_NotificationType_PreConvertToFull == notificationType)
+    if ((result == PrjFS_Result_Success) &&
```
jeschu1: @wilbaker, I'm not sure how the result==0 would have worked.  Thoughts?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
wilbaker: I think expanding the current test looks good.  To confirm, this test fails without your fix?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: @jeschu1 agreed, this looks like it was a bug.  Thanks for catching this!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
kewillford: I think this is good for this test but I would also like a test in `StatusTests` so that we are covering this case with `git status` as well.

jeschu1: Added a status test for good measure

ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs
```diff
@@ -193,8 +194,7 @@ public virtual Result StopVirtualizationInstance()
                     return this.OnPreDelete(relativePath, isDirectory);
 
                 case NotificationType.FileModified:
-                    this.OnFileModified(relativePath);
-                    return Result.Success;
+                    return this.OnFileModified(relativePath, isPlaceholderFile);
```
wilbaker: `OnFileModified` should stay as `void` return type.  Unlike `OnPreDelete` or `OnFilePreConvertToFull` this callback is informational.  If the provider returns a failure code we can't undo the modification, and the native library should still convert the file to full (i.e. remove the placeholder attributes).

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -436,19 +436,27 @@ private void OnLogError(string errorMessage)
...
         {
             try
             {
                 if (Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath))
                 {
                     this.OnDotGitFileOrFolderChanged(relativePath);
                 }
+                else if (isPlaceholderFile == true)
+                {
+                    // >> does not send a KAUTH_VNODE_WRITE data event
```
wilbaker: Super minor, when I saw ">>" here I thought it was just comment style, maybe:

"The '>>' bash command" instead of ">>"?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -436,19 +436,27 @@ private void OnLogError(string errorMessage)
...
+                    // To handle this we need to detect placeholders on Modifed events and perform the convert there
```
wilbaker: Maybe add one or to more details here, something like:

> // To handle this, PrjFSLib checks if a file is a placeholder in the Modified event and converts the file to full after delivering this callback

PR #1151, opened 2019-05-13
Updated Service response to EnablePrjFS request

### LINKED ISSUE #1053: gvfs service --mount-all fails on Mac
`gvfs service --unmount-all` succeeds, but `--mount-all` fails with PrjFlt load error message.

```
$ gvfs service --unmount-all

Unmounting repo at /Users/ameen/Work/Microsoft/VFSForGit/TestRepos/GVFS_9
Unmounting...Succeeded

Unmounting repo at /Users/ameen/Work/Microsoft/VFSForGit/TestRepos/GVFS_10
Unmounting...Succeeded
```
```
$ gvfs service --mount-all
Failed to enable PrjFlt: GVFS.Service responded with unexpected message: GVFS.Common.NamedPipes.NamedPipeMessages+EnableAndAttachProjFSRequest+Response
```

alameenshah:
Service (`RequestHandler.cs`) on the Mac, was sending responses to EnableProjFS requests in an incorrect format. This was causing `gvfs service --mount-all` to fail with “un-expected response message” error. Updated RequestHandler to first create a Message object from EnableProjFS Response and use it in its reply to client. This helps create a client parsable response with expected header and body.

Service response before the change -
`GVFS.Common.NamedPipes.NamedPipeMessages+EnableAndAttachProjFSRequest+Response`

Service response after the change -
`TRequestResponse|{\"State\":1,\"ErrorMessage\":null}”}`

Fixes #1053

### DIFF
--- a/GVFS/GVFS.UnitTests/Service/Mac/MacServiceTests.cs
+++ b/GVFS/GVFS.UnitTests/Service/Mac/MacServiceTests.cs
@@ -1,7 +1,9 @@
 ﻿using GVFS.Common;
 using GVFS.Common.FileSystem;
+using GVFS.Common.NamedPipes;
 using GVFS.Platform.Mac;
 using GVFS.Service;
+using GVFS.Service.Handlers;
 using GVFS.UnitTests.Mock.Common;
 using GVFS.UnitTests.Mock.FileSystem;
 using Moq;
@@ -47,6 +49,26 @@ namespace GVFS.UnitTests.Service.Mac
             service.Run();
 
             repoRegistry.VerifyAll();
+        }
+
+        [TestCase]
+        public void ServiceHandlesEnablePrjfsRequest()
+        {
+            string expectedServiceResponse = "TRequestResponse|{\"State\":1,\"ErrorMessage\":null}";
+            Mock<NamedPipeServer.Connection> connectionMock = new Mock<NamedPipeServer.Connection>(
+                MockBehavior.Strict,
+                null, // serverStream
+                this.tracer, // tracer
+                null); // isStopping
+            connectionMock.Setup(mp => mp.TrySendResponse(expectedServiceResponse)).Returns(true);
+
+            NamedPipeMessages.EnableAndAttachProjFSRequest request = new NamedPipeMessages.EnableAndAttachProjFSRequest();
+            request.EnlistmentRoot = string.Empty;
+
+            RequestHandler serviceRequestHandler = new RequestHandler(this.tracer, etwArea: string.Empty, repoRegistry: null);
+            serviceRequestHandler.HandleRequest(this.tracer, request.ToMessage().ToString(), connectionMock.Object);
+
+            connectionMock.VerifyAll();
         }
 
         [TestCase]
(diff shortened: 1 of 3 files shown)


jamill:
Thanks! The change and explanation look straight forward enough. Is there any tests we could add to cover this behavior?

jeschu1:
We have several service functional tests turned off, does it make sense to turn any on with this PR https://github.com/microsoft/VFSForGit/blob/master/GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/ServiceVerbTests.cs

GVFS/GVFS.UnitTests/Service/Mac/MacServiceTests.cs
```diff
@@ -49,6 +51,22 @@ public void ServiceStartTriggersAutoMountForCurrentUser()
             repoRegistry.VerifyAll();
         }
 
+        [TestCase]
+        public void ServiceHandlesEnablePrjfsRequest()
+        {
+            string expectedServiceResponse = "TRequestResponse|{\"State\":1,\"ErrorMessage\":null}";
```
alameenshah: Let me know if this string looks too strict.

alameenshah:
> Thanks! The change and explanation look straight forward enough. Is there any tests we could add to cover this behavior?

Ideally we should be having UT for all the Service Request Handler classes. Created a [ticket](https://github.com/microsoft/VFSForGit/issues/1153) for that.

For this specific PR, I've added a new test in `MacServiceTests` (which has other generic UT for `.netcore` target classes) to include a new test for EnablePrjFS handler response.

GVFS/GVFS.UnitTests/Service/Mac/MacServiceTests.cs
```diff
@@ -49,6 +51,22 @@ public void ServiceStartTriggersAutoMountForCurrentUser()
             repoRegistry.VerifyAll();
         }
 
+        [TestCase]
+        public void ServiceHandlesEnablePrjfsRequest()
+        {
+            string expectedServiceResponse = "TRequestResponse|{\"State\":1,\"ErrorMessage\":null}";
+            Mock<NamedPipeServer.Connection> connectionMock = new Mock<NamedPipeServer.Connection>(MockBehavior.Strict, null, this.tracer, null);
```
kewillford: What are the 2 `null`s you are passing here?  Will you use named parameters for them?

wilbaker:
If we haven't created a `milestones/M152` branch yet, should this PR be marked as M152?

cc: @jrbriggs

GVFS/GVFS.UnitTests/Service/Mac/MacServiceTests.cs
alameenshah: As discussed the `Mock` constructors don't allow this. Added comments instead of parameter labels.

```
public Mock();
public Mock(MockBehavior behavior);
public Mock(MockBehavior behavior, params object[] args);
```

wilbaker:
@alameenshah what was the verdict on this comment?

> We have several service functional tests turned off, does it make sense to turn any on with this PR https://github.com/microsoft/VFSForGit/blob/master/GVFS/GVFS.FunctionalTests/Tests/MultiEnlistmentTests/ServiceVerbTests.cs

cc: @jeschu1

PR #1154, opened 2019-05-14
ProjectedFSLib: do not copy into GVFS.Platform.Windows output dir

jamill:
Currently, ProjectedFSLib.dll is expected to be copied into the GVFS.Platform.Windows
output directory during a build, and from there is copied into the installer. Due to
how this is included in the project, it is not consistently copied into
the GVFS.Platform.Windows output directory, and this can cause a build
failure simiar to:

```
Error on line in C:\Sources\Vfs4Git\src\GVFS\GVFS.Installer.Windows\Setup.iss: Source file "C:\Sources\Vfs4Git\src\..\BuildOutput\GVFS.Windows\bin\x64\Debug\ProjectedFSLib.dll" does not exist.
```

To fix this, we could fix the project setup to have the file copied over
consistently. However, this file is not needed by the
GVFS.Platform.Windows project. Post RS4, ProjectedFSLib.dll should be
provided by the operating system and not copied over the product.
Previously, the product had to bring over its own copy of
ProjectedFSLib.dll, as it was not providided by the OS. By now, we
expect developers to have moved to a version of Windows that includes
this the ProjectedFSLib.dll component in box.

This also means we can remove logic for deleting this file as part of a
post build step on systems where this file should not be provided by the
product.

### DIFF
--- a/GVFS/GVFS.FunctionalTests.Windows/GVFS.FunctionalTests.Windows.csproj
+++ b/GVFS/GVFS.FunctionalTests.Windows/GVFS.FunctionalTests.Windows.csproj
@@ -189,10 +189,6 @@
       mkdir  $(TargetDir)\netcoreapp2.1
       xcopy /Y $(BuildOutputDir)\FastFetch\bin\$(Platform)\$(Configuration)\netcoreapp2.1\* $(TargetDir)\netcoreapp2.1
       xcopy /Y $(BuildOutputDir)\GVFS.FunctionalTests.LockHolder\bin\$(Platform)\$(Configuration)\netcoreapp2.1\* $(TargetDir)\netcoreapp2.1
-
-      REM If there is an inbox version of the ProjFS dll, we must use that one to guarantee compat with the sys, so delete whatever is in our TargetDir
-      if exist c:\Windows\System32\ProjectedFSLib.dll del $(TargetDir)\ProjectedFSLib.dll
-
 </PostBuildEvent>
   </PropertyGroup>
   <PropertyGroup>
--- a/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
+++ b/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
@@ -48,9 +48,6 @@
 
   <Target Name="CopyWindowsTestDependencies" AfterTargets="PostBuildEvent" Condition="'$(OS)' == 'Windows_NT'">
     <Copy SourceFiles="@(WindowsBuildOutputs)" DestinationFolder="$(TargetDir)\%(RecursiveDir)" SkipUnchangedFiles="true" OverwriteReadOnlyFiles="true" />
-
-    <!-- If there is an inbox version of the ProjFS dll, we must use that one to guarantee compat with the sys, so delete whatever is in our TargetDir -->
-    <Delete Files="$(TargetDir)\ProjectedFSLib.dll" Condition="Exists('c:\Windows\System32\ProjectedFSLib.dll')" />
   </Target>
   
 </Project>
--- a/GVFS/GVFS.Installer.Windows/Setup.iss
+++ b/GVFS/GVFS.Installer.Windows/Setup.iss
@@ -4,6 +4,7 @@
 ; General documentation on how to use InnoSetup scripts: http://www.jrsoftware.org/ishelp/index.php
 
 #define PrjFltDir PackagesDir + "\" + ProjFSNativePackage + "\filter" 
+#define ProjFSNativeLibDir PackagesDir + "\" + ProjFSNativePackage + "\lib" 
 #define VCRuntimeDir PackagesDir + "\GVFS.VCRuntime.0.2.0-build\lib\x64"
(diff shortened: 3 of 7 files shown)


GVFS/GVFS.Platform.Windows/packages.config
```diff
@@ -4,6 +4,7 @@
   <package id="Microsoft.Database.Collections.Generic" version="1.9.4" targetFramework="net461" />
   <package id="Microsoft.Database.Isam" version="1.9.4" targetFramework="net461" />
   <package id="Microsoft.Windows.ProjFS" version="1.0.19108.2" targetFramework="net461" />
+  <package id="GVFS.ProjFS" version="2019.411.1" targetFramework="net461" />
```
jamill: @wilbaker - I would be curious if there is a reason to not include this package as a dependency of `GVFS.Platform.Windows`...

mjcheetham:
> I think an alternative change would be to change this to a "Content" item.

I would vote for this option as it's more 'correct' IMO. This file does not play part in the build process but is a file that is required in the final product.

jamill:
> I would vote for this option as it's more 'correct' IMO.

I am not against this - the main reason I didn't choose this was because of the precedence with the libgit2 native binaries. If we go with this approach, I guess we should update the libgit2 native binary props to be consistent.

jrbriggs:
Changing to `Content` means the item will show up in the solution tree. I'm not opposed to that, just FYI.

mjcheetham:
> Changing to `Content` means the item will show up in the solution tree. I'm not opposed to that, just FYI.

Probably a good thing. Make your weak/runtime dependency on this file more visible?

GVFS/GVFS.Platform.Windows/packages.config
wilbaker: > I would be curious if there is a reason to not include this package as a dependency of GVFS.Platform.Windows

It's a bit weird but strictly speaking `GVFS.Platform.Windows` only has a dependency on the `Microsoft.Windows.ProjFS` package.

The only thing that *really* needs the `GVFS.ProjFS` package is the installer because the package is only needed if ProjFS is not available as part of the operating system.

Was there a reason you had to add this?

wilbaker: A bit more history here, `GVFS.Platform.Windows` did have a reference to `GVFS.ProjFS` when the managed API and native libraries were in the same package.  When they were split into two packages we removed the reference to `GVFS.ProjFS`.

Unless this was required to fix the build I recommend taking the "GVFS.ProjFS" reference back out, as `GVFS.Platform.Windows` should have to care about the native library.

jamill: > The only thing that really needs the GVFS.ProjFS package is the installer because the package is only needed if ProjFS is not available as part of the operating system.

What seems awkward is that the the `ProjectedFSLib.dll` is copied into the `GVFS.Platform.Windows` project's output directory, and that the installer copies it out of that directory.

When would we build `GVFS.Platform.Windows` and not want to copy `ProjectedFSLib.dll` into the output directory. Is there a scenario where this actually happens today?

jamill:
Looks like our build machines are pre RS4

> OS Version: Microsoft Windows NT 10.0.17134

RS4 is 1803

/cc @wilbaker @kewillford

jamill:
The actual failure is:

> 1) Invalid : C:/agent/_work/11/BuildOutput/GVFS.UnitTests.Windows/bin/x64/debug/GVFS.UnitTests.Windows.exe
Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
  ----> Could not load file or assembly 'ProjectedFSLib.Managed.dll' or one of its dependencies. The specified module could not be found.
  ----> Could not load file or assembly 'ProjectedFSLib.Managed.dll' or one of its dependencies. The specified module could not be found.


I am still investigating to see if this could be caused by running my changes on a Pre-RS4 machine...

wilbaker:
> Looks like our build machines are pre RS4 - from the failing tests:

Our build machine pools is RS4+, but it appears that the Hosted Agents pool is not (or it does not have the ProjFS optional feature enabled).

As we discussed offline, it's probably easiest to copy the native DLL to the `GVFS.UnitTests.Windows` output folder.

jamill:
> As we discussed offline, it's probably easiest to copy the native DLL to the GVFS.UnitTests.Windows output folder.

My latest commit does this. I used @kewillford's code to conditionally delete the ProjectedFSLib.dll if the dll exists in the system folder already... I am still not sure this is last step is necessary, as we are not actually mounting repositories in the unit test code (I don't think...).

If we don't like this approach, or if it has problems, it might just be easier to go with the original approach (but not explicitly include the `ProjectedFSLib.dll` package dependency in `GVFS.Platform.Windows`)

wilbaker:
> I used @kewillford's code to conditionally delete the ProjectedFSLib.dll if the dll exists in the system folder already... I am still not sure this is last step is necessary, as we are not actually mounting repositories in the unit test code (I don't think...).

I don't think this step should be necessary.  Assuming the tests pass with your latest changes I think it'd be worth taking the conditional delete out and see if everything still passes.

PR #1155, opened 2019-05-14
GVFSService: platforms should use appropriate pipe name

jamill:
On POSIX platforms the pipe name should be a rooted path so that we have
full control over the pipe path. If a relative path is used, then the
pipe is created in a temporary dictory, which can be a different
location depending on the user's context (i.e. running as root vs normal
user). This can result in an process running as root not being able to
communicate with the GVFS Service.

This change enables the different platforms to create a pipe path that
is suitable for the platform, and encapsulates the logic for creating
the pipe name for client / server in a single location.

### DIFF
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -117,6 +117,13 @@ namespace GVFS.Platform.POSIX
             return POSIXPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
         }
 
+        public override string GetGVFSServiceNamedPipeName(string serviceName)
+        {
+            // Pipes are stored as files on POSIX, use a rooted pipe name
+            // in the same location as the service to keep full control of the location of the file
+            return this.GetDataRootForGVFSComponent(serviceName) + ".pipe";
+        }
+
         public override bool IsConsoleOutputRedirectedToFile()
         {
             return POSIXPlatform.IsConsoleOutputRedirectedToFileImplementation();
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -199,6 +199,11 @@ namespace GVFS.Platform.Windows
             return WindowsPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
         }
 
+        public override string GetGVFSServiceNamedPipeName(string serviceName)
+        {
+            return serviceName + ".pipe";
+        }
+
         public override void ConfigureVisualStudio(string gitBinPath, ITracer tracer)
         {
             try
--- a/GVFS/GVFS.Service/GVFSService.Mac.cs
+++ b/GVFS/GVFS.Service/GVFSService.Mac.cs
@@ -42,7 +42,7 @@ namespace GVFS.Service
 
                 if (!string.IsNullOrEmpty(this.serviceName))
                 {
-                    string pipeName = this.serviceName + ".Pipe";
+                    string pipeName = GVFSPlatform.Instance.GetGVFSServiceNamedPipeName(this.serviceName);
                     this.tracer.RelatedInfo("Starting pipe server with name: " + pipeName);
 
                     using (NamedPipeServer pipeServer = NamedPipeServer.StartNewServer(
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -86,7 +86,7 @@ namespace GVFS.CommandLine
         {
             get
             {
-                return this.ServiceName + ".Pipe";
+                return GVFSPlatform.Instance.GetGVFSServiceNamedPipeName(this.ServiceName);
             }
         }
 
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -40,6 +40,7 @@ namespace GVFS.Common
         public abstract bool IsProcessActive(int processId);
         public abstract void IsServiceInstalledAndRunning(string name, out bool installed, out bool running);
         public abstract string GetNamedPipeName(string enlistmentRoot);
+        public abstract string GetGVFSServiceNamedPipeName(string serviceName);
         public abstract NamedPipeServerStream CreatePipeByName(string pipeName);
 
         public abstract string GetOSVersionInformation();
--- a/GVFS/GVFS.Service/GVFSService.Windows.cs
+++ b/GVFS/GVFS.Service/GVFSService.Windows.cs
@@ -52,7 +52,7 @@ namespace GVFS.Service
                 this.repoRegistry.Upgrade();
                 this.requestHandler = new WindowsRequestHandler(this.tracer, EtwArea, this.repoRegistry);
 
-                string pipeName = this.serviceName + ".Pipe";
+                string pipeName = GVFSPlatform.Instance.GetGVFSServiceNamedPipeName(this.serviceName);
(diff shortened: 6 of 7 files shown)


GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -199,6 +199,11 @@ public override string GetNamedPipeName(string enlistmentRoot)
             return WindowsPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
         }
 
+        public override string GetGVFSServiceNamedPipeName()
+        {
+            return GVFSConstants.Service.ServiceName + ".pipe";
```
alameenshah: Question - On `POSIXPlatform.cs` we are dynamically getting the root directory path for `Service` and constructing the pipe name by appending `.pipe` string to it. But here on Windows, we are using a fixed `<service_name>.pipe` string. Why is this different between platforms - where does the pipe file get created on Windows.

mjcheetham: > where does the pipe file get created on Windows.

On Windows pipes are not files, but exist in a special namespace for Pipes: `\.\pipe\<NAME>`

mjcheetham: https://docs.microsoft.com/en-us/windows/desktop/ipc/pipe-names

GVFS/GVFS.Service/GVFSService.Mac.cs
```diff
@@ -20,18 +20,18 @@ public class GVFSService
...
         private RequestHandler requestHandler;
 
-        public GVFSService(
-            ITracer tracer,
-            string serviceName,
-            IRepoRegistry repoRegistry)
+        public GVFSService(
+            ITracer tracer,
+            string serviceName,
+            IRepoRegistry repoRegistry)
```
alameenshah: Are these changes caused by File Encoding differences between Windows and Mac? This has become a common issue while editing cross-plat files.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
mjcheetham: Since in *NIX pipes exist in the unified file system namespace they need to go somewhere under `/`.
A pipe with a simple name gets created directly under `Path.GetTempPath()` with `"CoreFXPipe_"` prefix.

See [here](https://github.com/dotnet/corefx/blob/e12bfba58e3581dc475e804688bef1c66d70226c/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs#L49-L52) for more information.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -85,8 +85,8 @@ public string InternalParameters
         public string ServicePipeName
         {
             get
-            {
-                return this.ServiceName + ".Pipe";
+            {
+                return GVFSPlatform.Instance.GetGVFSServiceNamedPipeName();
```
wilbaker: I suspect the change from `this.ServiceName + ".Pipe"` to `GVFSPlatform.Instance.GetGVFSServiceNamedPipeName` will break some things on Windows because during the functionals tests the test service's pipe is named `Test.GVFS.Service`.

Perhaps `GetGVFSServiceNamedPipeName` should take a service name as a parameter?

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -117,6 +117,13 @@ public override string GetNamedPipeName(string enlistmentRoot)
             return POSIXPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
         }
 
+        public override string GetGVFSServiceNamedPipeName()
+        {
+            // Pipes are stored as files on POSIX, use a rooted pipe name
+            // in the same location as the service to keep full control of the location of the file
+            return this.GetDataRootForGVFSComponent(GVFSConstants.Service.ServiceName) + ".pipe";
```
wilbaker: Not your change, and looking for your opinion.  

I've had some trouble remembering what `GetDataRootForGVFSComponent` returns (and if it's a global path).

Do you think this should be renamed (perhaps to something like `GetGlobalDataRootForGVFSComponent`)?

It looks like in the git world "global" means per-user, and so that would seem to be consistent.

wilbaker: > It looks like in the git world "global" means per-user, and so that would seem to be consistent.

I guess except on Windows `GetDataRootForGVFSComponent` returns something system wide ... maybe there is a better name.

GVFS/GVFS/CommandLine/GVFSVerb.cs
jamill: I missed the test service! thanks. I will update to take in the service name.

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
jamill: I wonder if `GetDataDirectoryForService`, `GetDataDirectoryForFeature` would be easier... but I am not sure... Also - what is global on Windows is not currently global on macOS...

GVFS/GVFS.Service/GVFSService.Mac.cs
jamill: yes... I have wondered if we could help solve this with a setting in editorconfig. I think the expectation is that all lines are crlf terminated, but that gets mixed up...

chrisd8088:
This is not immediately relevant to this PR, but the changes here reminded me to note that the current code which generates named piped names on POSIX platforms (Mac, Linux) does not constrain them to the limits [imposed on domain sockets](https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars) of about 100 characters.

One can generate a stack trace like the following by simply mixing up the order of arguments on the `gvfs clone` command line:

```
Unhandled Exception: System.ArgumentOutOfRangeException: The path '.../.gvfs/GVFS_NetCorePipe' is of an invalid length for use with domain sockets on this platform.  The length must be between 1 and 108 characters, inclusive.
...
   at System.IO.Pipes.NamedPipeClientStream.Connect(Int32 timeout)
   at GVFS.Common.NamedPipes.NamedPipeClient.Connect(Int32 timeoutMilliseconds) in .../VFSForGit/GVFS/GVFS.Common/NamedPipes/NamedPipeClient.cs:line 29
   at GVFS.CommandLine.GVFSVerb.IsExistingPipeListening(String enlistmentRoot) in .../VFSForGit/GVFS/GVFS/CommandLine/GVFSVerb.cs:line 360
   at GVFS.CommandLine.CloneVerb.CheckNotInsideExistingRepo(String normalizedEnlistmentRootPath)
...
```

I could put this into an issue if it's not something you'd like to address in this PR, but I thought I'd mention it here because creating pipe names using absolute paths may also sometimes trigger this exception, depending on the path.

jamill:
> I could put this into an issue if it's not something you'd like to address in this PR, but I thought I'd mention it here because creating pipe names using absolute paths may also sometimes trigger this exception, depending on the path.

Thanks for pointing this out! If I understand the issue correctly, I think we are most susceptible to this in the pipe for the mount process, which depends on the enlistment path. I think the paths for the GVFS Service's named pipes are more constrained / controlled, as they do not depend on a mount's directory.

Let's file a separate issue for this - would you mind filing it? If not, I can go ahead and file it to make sure we track it.

GVFS/GVFS.Service/GVFSService.Mac.cs
nickgra: Setting up the `.editorconfig` to enforce line endings and normalizing the files that are currently mixed has been on my backlog of things I wanted to do for a long time. I think the difficult part there is validating that it behaves as expected in all the IDEs we use.

PR #1156, opened 2019-05-14
Clone and mount in unattended mode for loose object functional tests

derrickstolee:
The functional tests for loose objects seem to be a bit flaky. One reason may be that some background operations are interfering at some points. Clone in unattended mode to ensure no background operations are happening during our test runs.

I'll run the functional tests a few times with this to try and catch the flakiness. If this doesn't work, then we will need to relax the conditions on the tests.


derrickstolee:
This didn't work. I'll create a new PR after I change the tests.

PR #1157, opened 2019-05-14
GitProcess: Enable lower ProcessPriorityClass

derrickstolee:
In an effort to have fewer complaints for running things in the background, give background Git commands a lower process priority.

See [ProcessPriorityClass enum documentation](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processpriorityclass?view=netframework-4.8) for details of how this should work. By specifying "BelowNormal", we are allowing the process to run while a user is at the machine, but the process should not keep them from doing their work. The only lower priority is `Idle` which implies that the user is not using the machine, such as with a screensaver. I'm not sure if we want to go that low.

/cc @jrbriggs as he had an idea of what to do here, and this is the only kind of priority I could find.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -83,6 +83,8 @@ namespace GVFS.Common.Git
             }
         }
 
+        public bool ForBackground { get; set; }
+
         public static Result Init(Enlistment enlistment)
         {
             return new GitProcess(enlistment).InvokeGitOutsideEnlistment("init \"" + enlistment.WorkingDirectoryBackingRoot + "\"");
@@ -688,6 +690,7 @@ namespace GVFS.Common.Git
 
             Process executingProcess = new Process();
             executingProcess.StartInfo = processInfo;
+
             return executingProcess;
         }
 
@@ -748,6 +751,11 @@ namespace GVFS.Common.Git
                             }
 
                             this.executingProcess.Start();
+
+                            if (this.ForBackground)
+                            {
+                                this.executingProcess.PriorityClass = ProcessPriorityClass.BelowNormal;
+                            }
                         }
 
                         writeStdIn?.Invoke(this.executingProcess.StandardInput);
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -83,7 +83,7 @@ namespace GVFS.Common.Git
             }
         }
 
-        public bool ForBackground { get; set; }
+        public bool LowerPriority { get; set; }
 
         public static Result Init(Enlistment enlistment)
         {
@@ -752,7 +752,7 @@ namespace GVFS.Common.Git
 
                             this.executingProcess.Start();
 
-                            if (this.ForBackground)
+                            if (this.LowerPriority)
                             {
                                 this.executingProcess.PriorityClass = ProcessPriorityClass.BelowNormal;
                             }
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -237,7 +237,7 @@ namespace GVFS.Common.Maintenance
                 }
 
                 this.MaintenanceGitProcess = this.Context.Enlistment.CreateGitProcess();
-                this.MaintenanceGitProcess.ForBackground = true;
+                this.MaintenanceGitProcess.LowerPriority = true;
             }
 
             try
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -237,6 +237,7 @@ namespace GVFS.Common.Maintenance
                 }
 
                 this.MaintenanceGitProcess = this.Context.Enlistment.CreateGitProcess();
+                this.MaintenanceGitProcess.ForBackground = true;
             }
 
             try


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -688,6 +690,12 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
 
             Process executingProcess = new Process();
             executingProcess.StartInfo = processInfo;
+
+            if (this.ForBackground)
+            {
+                executingProcess.PriorityClass = ProcessPriorityClass.BelowNormal;
```
wilbaker: I think that this can only be set after the process has started (and I don't think that's happened yet at this point).

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.priorityclass?view=netframework-4.6.1

>Exceptions
Win32Exception
Process priority information could not be set or retrieved from the associated process resource.
>
>-or-
>
>The process identifier or process handle is zero. (The process has not been started.)

wilbaker:
FYI - Some details on how this works on Mac and Linux from what I could find in the .NET Core source.

[The `PriorityClass` setter](https://github.com/dotnet/corefx/blob/6ca16758a5d454c1f1b04975bf55f259dd71fc49/src/System.Diagnostics.Process/src/System/Diagnostics/Process.cs#L477) calls the [`PriorityClassCore` setter](https://github.com/dotnet/corefx/blob/007571b90c3cc555cd30f3b467b87c22f9f665e0/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Unix.cs#L271), which has the following comment/values:

```
// This mapping is relatively arbitrary.  0 is normal based on the man page,
// and the other values above and below are simply distributed evenly.

int pri = 0; // Normal
switch (value)
{
    case ProcessPriorityClass.RealTime: pri = -19; break;
    case ProcessPriorityClass.High: pri = -11; break;
    case ProcessPriorityClass.AboveNormal: pri = -6; break;
    case ProcessPriorityClass.BelowNormal: pri = 10; break;
    case ProcessPriorityClass.Idle: pri = 19; break;
    default:
        Debug.Assert(value == ProcessPriorityClass.Normal, "Input should have been validated by caller");
        break;
}

int result = Interop.Sys.SetPriority(Interop.Sys.PriorityWhich.PRIO_PROCESS, _processId, pri);
```

And `Interop.Sys.SetPriority` ends up calling [`setpriority`](https://github.com/dotnet/corefx/blob/b10e8d67b260e26f2e47750cf96669e6f48e774d/src/Native/Unix/System.Native/pal_process.c#L738)

GVFS/GVFS.Common/Git/GitProcess.cs
derrickstolee: And fittingly, the functional tests around the loose objects and packfile maintenance steps failed. Thanks!

derrickstolee:
If anyone has ideas of how to test this, I'm open to suggestions.

For my manual test, I loaded Task Manager, went to the "Details" page, and added the "Base priority" column. Then, I ran the `PackfileMaintenanceStep` using the `gvfs dehydrate` command. I watched `git.exe` pop up with "Below Normal" priority.

jeschu1:
> my manual test, I loaded Task Manager, went to the "Details" page, and added the "Base priority" column. Then, I ran the PackfileMaintenanceStep using the gvfs dehydrate command. I watched git.exe pop up with "Below Normal" priority.

Do you have an idea of how much longer the job takes with "Below Normal" priority?  It would be a good to test running some meaty things on the machine and try it with/without the priority change.

I can't think of a great automated test.  I imagine we could run a background job on a functional test and verify it was set up with "Below Normal" priority, but I don't think that would be worth it.  I'd be more interested in knowing the impact of changing the priority.

derrickstolee:
> Do you have an idea of how much longer the job takes with "Below Normal" priority? It would be a good to test running some meaty things on the machine and try it with/without the priority change.

The tricky part here is that if the computer is not in use, then it shouldn't take any longer, right? But if someone is running a build, then the step will now take a lot longer than before. It won't use any resources while it is being skipped in favor of other processes, so these longer times shouldn't have user-facing effects.

jeschu1:
> The tricky part here is that if the computer is not in use, then it shouldn't take any longer, right? But if someone is running a build, then the step will now take a lot longer than before. It won't use any resources while it is being skipped in favor of other processes, so these longer times shouldn't have user-facing effects.

I agree.  In the past we've been concerned about long running background jobs.  Should we continue to be concerned with this or accept we are going to get some slow runs by design (and they aren't hurting anything because of the priority drop)?

I think its worth doing the simple test you describe locally where you run a build while maintence is running as a data point.

jrbriggs:
>  Should we continue to be concerned with this or accept we are going to get some slow runs by design (and they aren't hurting anything because of the priority drop)?

Things running slower in the background is not only acceptable, but desirable. There's no rush to fix something that's been a problem for weeks or months. We can do it a few seconds slower and not irritate users.

jamill:
:+1: I like marking the background tasks to run at a lower priority. At least it conveys our intention of how we want the process to run.

jeschu1:
I'm only advocating we test it once manually, during a build (or another busy time) before shipping - and report the results as a benchmark.  I definitely think we should take the change.

derrickstolee:
I have run this locally, and watched the processes start with lower priority. Measuring the end-to-end time change is unlikely to have conclusive evidence, because (1) these commands already have high variance between runs, and (2) it will change depending on the load we are putting on the machine.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -83,6 +83,8 @@ public GitProcess(string gitBinPath, string workingDirectoryRoot, string gvfsHoo
             }
         }
 
+        public bool ForBackground { get; set; }
```
jamill: I would consider including a comment about what this means. Background can be a general without context... e.g. that this process will be run with lower priority to not impact other uses of the machine..

derrickstolee: I'll just rename it to `LowerPriority` so the name is self-documenting.

PR #1159, opened 2019-05-14
MacPlatform: enable GVFS config verb

jamill:


GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -7,19 +7,24 @@
...
     {
-        public POSIXPlatform() : base(
+        protected POSIXPlatform() : this(
             underConstruction: new UnderConstructionFlags(
                 supportsGVFSUpgrade: false,
                 supportsGVFSConfig: false))
         {
         }
 
+        protected POSIXPlatform(UnderConstructionFlags underConstruction)
```
nickgra: It's not apparent to me from this PR or the description, why does `POSIXPlatform` not support config?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -8,10 +8,13 @@ namespace GVFS.Platform.Mac
 {
     public partial class MacPlatform : POSIXPlatform
     {
-        public MacPlatform()
+        public MacPlatform() : base (
+             underConstruction: new UnderConstructionFlags(
+                supportsGVFSUpgrade: false,
+                supportsGVFSConfig: true))
```
wilbaker: With this change can the `ConfigVerbTests` functional tests be turned on for Mac?

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
jamill: I implemented this at the POSIX layer, but I hadn't tested it on a Linux platform. I thought this was a good middle ground between implement at the common layer, but enable at the layer that is tested.

Also, the overarching change will include upgrader related config settings, which also will not be tested on Linux at initially.

PR #1160, opened 2019-05-15
Add repo root dir to retention list before projection updates

chrisd8088:
The `GitIndexProjection.UpdatePlaceholders()` method, after processing the list of placeholder files, iterates over the list of placeholder folders and tries to remove any which are actually files, or not in the projection, or a possible tombstone.

The last placeholder folder in the ordered list appears to always correspond to the root of the repository; this placeholder has an empty path string, and `IsPathProjected()` returns `false` for both the `isProjected` and `isFolder` flags for this entry.

Because both flags are `false`, `RemoveFolderPlaceholderIfEmpty()` is called, which then results in a call to `DeleteFile()` in the lower-level `ProjFS` layer.  This always fails, because the root folder is never empty (it contains, among other things, the `.git` folder), but since `RemoveFolderPlaceholderIfEmpty()` handles the `FSResult.DirectoryNotEmpty` error code, no exception is propagated.

We can avoid these unnecessary attempts to remove the repository's root folder on `git checkout` operations by adding the empty path to the `folderPlaceholdersToKeep` list.

This behaviour was noted while testing #1128, which implements the underlying `DeleteFile()` method on Linux, where the use of a true filesystem mount point for the root of the VFSForGit repository means that an attempt to `rmdir(2)` the repository's root directory does not return `ENOTEMPTY` but rather `EBUSY` (`Device or resource busy`).  This is not silently handled by `RemoveFolderPlaceholderIfEmpty()` but instead logs an error.

/cc @jrbriggs, @kivikakk

### DIFF
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1111,7 +1111,10 @@ namespace GVFS.Virtualization.Projection
                 numThreads = Math.Min(numThreads, placeholderFilesListCopy.Count / minItemsPerThread);
                 numThreads = Math.Max(numThreads, 1);
 
+                // folderPlaceholdersToKeep always contains the empty path so as to avoid unnecessary attempts
+                // to remove the repository's root folder.
                 ConcurrentHashSet<string> folderPlaceholdersToKeep = new ConcurrentHashSet<string>();
+                folderPlaceholdersToKeep.Add(string.Empty);
 
                 // updatedPlaceholderDictionary and updatedPlaceholderBag are mutually exclusive.
                 //  - On platforms that expand on enumeration: updatedPlaceholderDictionary is used (required for ReExpandFolder)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1175,9 +1175,11 @@ private void UpdatePlaceholders()
                             // 1. not in the projection anymore
                             // 2. in the projection but is not a folder in the projection
                             // 3. Folder placeholder is a possible tombstone
-                            if (!isProjected ||
+                            // and if the folder is not the root folder of the repository
+                            if ((!isProjected ||
                                 !isFolder ||
-                                folderPlaceholder.IsPossibleTombstoneFolder)
+                                folderPlaceholder.IsPossibleTombstoneFolder) &&
+                                folderPlaceholder.Path != string.Empty)
```
derrickstolee: We prefer `!string.IsNullOrEmpty(folderPlaceholder.Path)` for comparing against an empty string.

kewillford: How about we add the string.Empty to the `folderPlaceholdersToKeep` when we initialize it since we will always be keeping that folder?

chrisd8088: Thanks for the tip -- while I didn't use it here because of @kewillford's suggestion below, I did end up using it in #1128!

chrisd8088: I thought about that too, but was concerned it might "match against everything" and short-circuit all the path comparisons.  However, a little testing and a more careful read of the code shows this is not at all the case, so I rebased with your suggestion implemented in 46cca89.  Thanks!

chrisd8088:
@wilbaker -- reworked using @kewillford's much tidier suggestion!

PR #1162, opened 2019-05-15
Relax LooseObjectStepTests from exact object counts

derrickstolee:
We are having some flaky tests because _somtimes_ we get one too many or one too few loose objects in our `LooseObjectStepTests`. The only reason this can happen (as I see it) is that we are downloading a loose object due to the virtualization layer, and that file exists in a pack somehow. This causes us to delete or add an object during the `LooseObjectsStep` and have off-by-one errors. Instead of seeing 160 objects (for example) we see 159 or 161.

Relax the test conditions to verify that we did not delete all of the loose objects during the steps instead of requiring the count is exact. (We already relaxed this in one direction in #1116.)


derrickstolee:
Marking as ready for review since I had five functional test runs succeed. Happy to run more, if we want to.

PR #1168, opened 2019-05-15
Functional Test Fix

jeschu1:
This should fail

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
@@ -1,4 +1,5 @@
-﻿using GVFS.FunctionalTests.Tools;
+﻿using GVFS.FunctionalTests.FileSystemRunners;
+using GVFS.FunctionalTests.Tools;
 using GVFS.Tests.Should;
 using NUnit.Framework;
 using System;
@@ -136,7 +137,13 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         [TestCase]
         public void AppendFile()
         {
-            this.AppendAllText("More Data", @"Readme.md");
+            BashRunner bash = new BashRunner();
+            string filePath = @"Readme.md";
+            string content = "Apended Data";
+            string virtualFile = Path.Combine(this.Enlistment.RepoRoot, filePath);
+            string controlFile = Path.Combine(this.ControlGitRepo.RootPath, filePath);
+            bash.AppendAllText(virtualFile, content);
+            bash.AppendAllText(controlFile, content);
 
             this.ValidateGitCommand("status");
         }
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
@@ -136,7 +136,7 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         [TestCase]
         public void AppendFile()
         {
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
```diff
@@ -131,12 +131,12 @@ public void ModifyingHeadRefInvalidatesCache()
...
             this.ValidateGitCommand("status");
-        }
-
+        }
+
         [TestCase]
         public void AppendFile()
         {
-            this.AppendAllText(@"Readme.md", "More Data");
+            this.AppendAllText("More Data", @"Readme.md");
```
derrickstolee: This worked before because "More Data" wasn't a file, so appending was the same as a first write?

jeschu1: Parameters were the wrong way round.  

I wanted to verify fixing them made them, made the test fail (cause append is actually broken in cases where the file wasn't hydrated).  I also wanted to test this on a clean machine.

So by virtue of this passing either:
1. The test hydrates the file
2. There is a race condition
3. something else is wrong

I've pulled the branch to dig deeper

GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
```diff
@@ -136,7 +137,13 @@ public void ModifyingHeadRefInvalidatesCache()
         [TestCase]
         public void AppendFile()
         {
-            this.AppendAllText("More Data", @"Readme.md");
+            BashRunner bash = new BashRunner();
```
wilbaker: Minor: Could add some comments here describing why we're using Bash specifically?  It might also be worth updating the test name.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
```diff
@@ -136,7 +137,13 @@ public void ModifyingHeadRefInvalidatesCache()
...
         public void AppendFile()
         {
-            this.AppendAllText("More Data", @"Readme.md");
+            BashRunner bash = new BashRunner();
+            string filePath = @"Readme.md";
+            string content = "Apended Data";
+            string virtualFile = Path.Combine(this.Enlistment.RepoRoot, filePath);
+            string controlFile = Path.Combine(this.ControlGitRepo.RootPath, filePath);
+            bash.AppendAllText(virtualFile, content);
+            bash.AppendAllText(controlFile, content);
```
wilbaker: Do we have a test that validates that the contents of the file after using `>>` is correct?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/StatusTests.cs
derrickstolee: The test is not specific enough in what it is checking, then.

All the `ValidateGitCommand()` method does is check that we get the same output in vanilla Git as we do in VFS-enabled Git. But in the old code, we just saw a new untracked "More Data" file.

We should be specific about what we are expecting to see in the output, and what content should be in the file. These questions should help design a test that prevents this bug in the future:

1. What is the content in Readme.md?
2. What _specifically_ should `git status` say about Readme.md?

jeschu1: @derrickstolee  agreed!  This would a quick draft test to validate failure on the build server.  As @wilbaker  pointed out we need to validate the file contents and add some comments around what we're trying to achieve.  I'm closing this PR, but the actual one we're taking will have those details.

PR #1169, opened 2019-05-15
Fix Append Files

jeschu1:
- We should call MessageType_KtoU_NotifyFilePreConvertToFull on KAUTH_VNODE_APPEND_DATA
- Revert Previous method of relying on FILEOP operations

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -366,7 +366,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_READ_DATA |
                 KAUTH_VNODE_WRITE_DATA |
                 KAUTH_VNODE_EXECUTE |
-                KAUTH_VNODE_DELETE)) // Hydrate on delete to ensure files are hydrated before rename operations
+                KAUTH_VNODE_DELETE | // Hydrate on delete to ensure files are hydrated before rename operations
+                KAUTH_VNODE_APPEND_DATA))
         {
             if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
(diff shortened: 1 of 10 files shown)


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -366,7 +366,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_READ_DATA |
                 KAUTH_VNODE_WRITE_DATA |
                 KAUTH_VNODE_EXECUTE |
-                KAUTH_VNODE_DELETE)) // Hydrate on delete to ensure files are hydrated before rename operations
+                KAUTH_VNODE_DELETE |
+                KAUTH_VNODE_APPEND_DATA)) // Hydrate on delete to ensure files are hydrated before rename operations
```
wilbaker: Nit: This comment should be on the same line as `KAUTH_VNODE_DELETE`

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -116,6 +116,8 @@ - (void) testEmptyFileHydrates {
         KAUTH_VNODE_WRITE_DATA,
         KAUTH_VNODE_EXECUTE,
         KAUTH_VNODE_DELETE,
+        KAUTH_VNODE_APPEND_DATA,
+        KAUTH_VNODE_ADD_SUBDIRECTORY,
```
wilbaker: We shouldn't receive `KAUTH_VNODE_ADD_SUBDIRECTORY` for file vnodes.  Should we only be adding `KAUTH_VNODE_APPEND_DATA` here?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -191,6 +195,8 @@ - (void) testNonEmptyFileDoesNotHydrate {
...
```
wilbaker: Similar comment here, it looks a bit strange to have `KAUTH_VNODE_ADD_SUBDIRECTORY` for a file test

wilbaker: (this comment applies to all of the file specific tests)

wilbaker:
Other than some suggestions for the unit tests this is looking good!

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
jeschu1: I'll fix that up, I had it in only because the code is the same, but you're correct this is a case that shouldn't happen

PR #1177, opened 2019-05-20
refactor ReadLink into MirrorProvider.POSIX

kivikakk:
Refactor our uses of libc's `readlink` in the MirrorProvider for Mac and Linux.

From @chrisd8088's suggestion at https://github.com/github/VFSForGit/pull/21#discussion_r284493927.

/cc @derrickstolee @wilbaker

### DIFF
--- a/MirrorProvider/MirrorProvider.POSIX/POSIXNative.cs
+++ b/MirrorProvider/MirrorProvider.POSIX/POSIXNative.cs
@@ -1,0 +1,28 @@
+﻿using System.Runtime.InteropServices;
+using System.Text;
+
+namespace MirrorProvider.POSIX
+{
+    public static class POSIXNative
+    {
+        public static string ReadLink(string path)
+        {
+            const ulong BufSize = 4096;
+            byte[] targetBuffer = new byte[BufSize];
+            long bytesRead = __ReadLink(path, targetBuffer, BufSize);
+            if (bytesRead < 0)
+            {
+                return null;
+            }
+
+            targetBuffer[bytesRead] = 0;
+            return Encoding.UTF8.GetString(targetBuffer);
+        }
+
+        [DllImport("libc", EntryPoint = "readlink", SetLastError = true)]
+        public static extern long __ReadLink(
+            string path,
+            byte[] buf,
+            ulong bufsize);
+    }
+}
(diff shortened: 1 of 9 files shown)


chrisd8088:
This looks great to me, @kivikakk ... thank you!

MirrorProvider/MirrorProvider.POSIX/POSIXNative.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            const ulong BufSize = 4096;
+            byte[] targetBuffer = new byte[BufSize];
+            long bytesRead = __ReadLink(path, targetBuffer, BufSize);
+            if (bytesRead < 0)
+            {
+                return null;
+            }
+
+            targetBuffer[bytesRead] = 0;
```
derrickstolee: What happens if we read exactly `BufSize` bytes? The [`man` page](http://man7.org/linux/man-pages/man2/readlink.2.html) appears to imply it will silently truncate, but can entirely fill the available bytes:

```
      readlink() places the contents of the symbolic link pathname in the
       buffer buf, which has size bufsiz.  readlink() does not append a null
       byte to buf.  It will (silently) truncate the contents (to a length
       of bufsiz characters), in case the buffer is too small to hold all of
       the contents.
```

Perhaps `byte[] targetBuffer = new byte[BufSize + 1];`?

derrickstolee: Or, should we throw an exception if we suspect the link was truncated (`bytesRead == BufSize`)?

kivikakk: Indeed, I had shifted this code unchanged during this refactor, but I think you're right, and that this has the potential to crash, as it could return `BufSize` if the link target was exactly 4096 bytes.

I'm inclined to think throwing is cleaner, although right now the code expects a `null` for failure to `readlink` -- callers know how to deal with that.

We can use the three-argument variant of `Encoding.GetString` to avoid having to write the nul byte into the array ourselves, but we should still return null on `bytesRead == BufSize`, as it's ambiguous as to whether it was actually a 4096 byte target, or whether it was larger and truncated.

Changes made in 262f7c4c.

kivikakk:
Thanks for the review! I've made changes accordingly.

derrickstolee:
@kivikakk Sorry about the test noise. The test is fixed in master with #1162. How often do you update this prototype branch with changes from master?

chrisd8088:
@derrickstolee -- I've been rebasing from master relatively often of late, but good to know there's a fix for one of the flaky CI tests; I'll pull that in soon.

PR #1184, opened 2019-05-21
Mac: Support Hardlinks

### LINKED ISSUE #337: Mac: Add both source and target path to ModifiedPaths.dat when hardlinks are created

jeschu1:
Add handling for RelativeFromPath to support Hardlinks 
- add both from / to paths to ModifiedFiles.dat

resolves #337

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
@@ -391,6 +391,59 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
         }
 
+        [TestCase, Order(18)]
+        public void HardlinkFromOutsideRepoToInside()
+        {
+            string fileName = "OutsideRepoFileForHardlink.txt";
+            string fileOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileName);
+            this.fileSystem.WriteAllText(fileOutsideRepo, "Contents for the new file");
+            fileOutsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
+
+            string fileHardlinkedInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
+            this.fileSystem.CreateHardLink(fileHardlinkedInsideRepo, fileOutsideRepo);
+            fileHardlinkedInsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+        }
+
+        [TestCase, Order(19)]
+        public void HardlinkFromInsideRepoToInside()
+        {
+            string fileName = "InsideRepoToInsideRepoFileForHardlink.txt";
+            string fileInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
+            this.fileSystem.WriteAllText(fileInsideRepo, "Contents for the new file");
+            fileInsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+
+            string fileHardLinkedOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileName);
+            this.fileSystem.CreateHardLink(fileHardLinkedOutsideRepo, fileInsideRepo);
+            fileHardLinkedOutsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+        }
+
+        [TestCase, Order(20)]
+        public void HardlinkInsideRepo()
+        {
+            string fileName = "InsideRepoFileForHardlink.txt";
+            string fileInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
+            this.fileSystem.WriteAllText(fileInsideRepo, "Contents for the new file");
+            fileInsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+
+            string fileNameLink = "HardLinkInsideRepoLink.txt";
+            string fileLinkInsideRepo = this.Enlistment.GetVirtualPathTo(fileNameLink);
+            this.fileSystem.CreateHardLink(fileLinkInsideRepo, fileInsideRepo);
+            fileLinkInsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
+        }
+
         [DllImport("GVFS.NativeTests.dll", CharSet = CharSet.Unicode)]
         private static extern bool SupersedeFile(string path, [MarshalAs(UnmanagedType.LPStr)]string newContent);
 
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
@@ -394,54 +394,76 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         [TestCase, Order(18)]
         public void HardlinkFromOutsideRepoToInside()
         {
-            string fileName = "OutsideRepoFileForHardlink.txt";
+            string fileName = "OutsideRepoToInside_FileForHardlink.txt";
             string fileOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileName);
             this.fileSystem.WriteAllText(fileOutsideRepo, "Contents for the new file");
             fileOutsideRepo.ShouldBeAFile(this.fileSystem);
             this.Enlistment.WaitForBackgroundOperations();
             GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
 
-            string fileHardlinkedInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
-            this.fileSystem.CreateHardLink(fileHardlinkedInsideRepo, fileOutsideRepo);
-            fileHardlinkedInsideRepo.ShouldBeAFile(this.fileSystem);
+            string fileNameLink = "OutsideRepoToInside_RepoLink.txt";
+            string fileLinkInsideRepo = this.Enlistment.GetVirtualPathTo(fileNameLink);
+            this.fileSystem.CreateHardLink(fileLinkInsideRepo, fileOutsideRepo);
             this.Enlistment.WaitForBackgroundOperations();
-            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
+            fileNameLink.ShouldBeAFile(this.fileSystem);
         }
 
         [TestCase, Order(19)]
-        public void HardlinkFromInsideRepoToInside()
+        public void HardlinkFromInsideRepoToOutside()
         {
-            string fileName = "InsideRepoToInsideRepoFileForHardlink.txt";
+            string fileName = "InsideRepoToOutside_FileForHardlink.txt";
             string fileInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
             this.fileSystem.WriteAllText(fileInsideRepo, "Contents for the new file");
             fileInsideRepo.ShouldBeAFile(this.fileSystem);
             this.Enlistment.WaitForBackgroundOperations();
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
 
-            string fileHardLinkedOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileName);
-            this.fileSystem.CreateHardLink(fileHardLinkedOutsideRepo, fileInsideRepo);
-            fileHardLinkedOutsideRepo.ShouldBeAFile(this.fileSystem);
+            string fileNameLink = "InsideRepoToOutside_RepoLink.txt";
+            string fileLinkOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileNameLink);
+            this.fileSystem.CreateHardLink(fileLinkOutsideRepo, fileInsideRepo);
+            fileLinkOutsideRepo.ShouldBeAFile(this.fileSystem);
             this.Enlistment.WaitForBackgroundOperations();
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileNameLink);
         }
 
         [TestCase, Order(20)]
         public void HardlinkInsideRepo()
         {
-            string fileName = "InsideRepoFileForHardlink.txt";
+            string fileName = "InsideRepo_FileForHardlink.txt";
             string fileInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
             this.fileSystem.WriteAllText(fileInsideRepo, "Contents for the new file");
             fileInsideRepo.ShouldBeAFile(this.fileSystem);
             this.Enlistment.WaitForBackgroundOperations();
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
 
-            string fileNameLink = "HardLinkInsideRepoLink.txt";
+            string fileNameLink = "InsideRepo_RepoLink.txt";
             string fileLinkInsideRepo = this.Enlistment.GetVirtualPathTo(fileNameLink);
             this.fileSystem.CreateHardLink(fileLinkInsideRepo, fileInsideRepo);
-            fileLinkInsideRepo.ShouldBeAFile(this.fileSystem);
             this.Enlistment.WaitForBackgroundOperations();
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
+            fileLinkInsideRepo.ShouldBeAFile(this.fileSystem);
+        }
+
+        [TestCase, Order(21)]
+        public void HardlinkExistingFileInRepo()
+        {
+            string fileName = "GVFS/GVFS.Mount/Program.cs";
+            string fileNameLink = "HardLinkToReadme";
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileNameLink);
+
+            string fileInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
+            string fileLinkInsideRepo = this.Enlistment.GetVirtualPathTo(fileNameLink);
+            this.fileSystem.CreateHardLink(fileLinkInsideRepo, fileInsideRepo);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
+            fileInsideRepo.ShouldBeAFile(this.fileSystem);
+            fileLinkInsideRepo.ShouldBeAFile(this.fileSystem);
         }
 
         [DllImport("GVFS.NativeTests.dll", CharSet = CharSet.Unicode)]
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
@@ -407,18 +407,15 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.Enlistment.WaitForBackgroundOperations();
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
             GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
-            fileNameLink.ShouldBeAFile(this.fileSystem);
+            fileLinkInsideRepo.ShouldBeAFile(this.fileSystem);
         }
 
         [TestCase, Order(19)]
         public void HardlinkFromInsideRepoToOutside()
         {
-            string fileName = "InsideRepoToOutside_FileForHardlink.txt";
+            string fileName = "Readme.md";
             string fileInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
-            this.fileSystem.WriteAllText(fileInsideRepo, "Contents for the new file");
-            fileInsideRepo.ShouldBeAFile(this.fileSystem);
-            this.Enlistment.WaitForBackgroundOperations();
-            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
 
             string fileNameLink = "InsideRepoToOutside_RepoLink.txt";
             string fileLinkOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileNameLink);
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -218,7 +218,7 @@ namespace GVFS.Platform.Mac
             this.virtualizationInstance.OnPreDelete = this.OnPreDelete;
             this.virtualizationInstance.OnNewFileCreated = this.OnNewFileCreated;
             this.virtualizationInstance.OnFileRenamed = this.OnFileRenamed;
-            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
+            this.virtualizationInstance.OnHardLinkCreated = this.NotifyHardLinkCreated;
             this.virtualizationInstance.OnFilePreConvertToFull = this.NotifyFilePreConvertToFull;
 
             uint threadCount = (uint)Environment.ProcessorCount * 2;
@@ -551,10 +551,10 @@ namespace GVFS.Platform.Mac
                 isDirectory: isDirectory);
         }
 
-        private void OnHardLinkCreated(string relativeNewLinkPath)
+        private void NotifyHardLinkCreated(string relativeNewLinkPath, string relativeFromPath)
         {
             this.OnHardLinkCreated(
-                relativeExistingFilePath: string.Empty,
+                relativeExistingFilePath: relativeFromPath,
                 relativeNewLinkPath: relativeNewLinkPath);
         }
 
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -218,7 +218,7 @@ namespace GVFS.Platform.Mac
             this.virtualizationInstance.OnPreDelete = this.OnPreDelete;
             this.virtualizationInstance.OnNewFileCreated = this.OnNewFileCreated;
             this.virtualizationInstance.OnFileRenamed = this.OnFileRenamed;
-            this.virtualizationInstance.OnHardLinkCreated = this.NotifyHardLinkCreated;
+            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
             this.virtualizationInstance.OnFilePreConvertToFull = this.NotifyFilePreConvertToFull;
 
             uint threadCount = (uint)Environment.ProcessorCount * 2;
@@ -549,13 +549,6 @@ namespace GVFS.Platform.Mac
                 relativeSourcePath: string.Empty,
                 relativeDestinationPath: relativeDestinationPath,
                 isDirectory: isDirectory);
-        }
-
-        private void NotifyHardLinkCreated(string relativeNewLinkPath, string relativeFromPath)
-        {
-            this.OnHardLinkCreated(
-                relativeExistingFilePath: relativeFromPath,
-                relativeNewLinkPath: relativeNewLinkPath);
         }
 
         private Result OnEnumerateDirectory(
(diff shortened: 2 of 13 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1078,14 +1099,20 @@ static PrjFS_Result HandleFileNotification(
     const MessageHeader* request,
     const char* relativePath,
     const char* absolutePath,
+    const char* relativeFromPath,
     bool isDirectory,
     PrjFS_NotificationType notificationType)
 {
+    // Don't send null to providers
+    relativePath = relativePath != nullptr ? relativePath : "";
```
jeschu1: @wilbaker what are your thoughts on having the check here.  Option 2 would be send "" instead of nullptr in the TrySendMessage many levels up.  Both approachs have pros/cons.  Since we were already doing the nullptr I stuck with it for now.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -269,17 +269,24 @@ protected void OnHardLinkCreated(string relativeExistingFilePath, string relativ
...
-                    if (pathInDotGit)
+                    if (newLinkPathInDotGit)
                     {
                         this.OnDotGitFileOrFolderChanged(relativeNewLinkPath);
                     }
-                    else
+
+                    if (existingPathInDotGit)
+                    {
+                        this.OnDotGitFileOrFolderChanged(relativeExistingFilePath);
```
wilbaker: I can't think of a scenario where `OnDotGitFileOrFolderChanged` would need to be called for `relativeExistingFilePath`, did you have one in mind?

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -269,17 +269,24 @@ protected void OnHardLinkCreated(string relativeExistingFilePath, string relativ
...
+                    }
+
+                    if (!(newLinkPathInDotGit && existingPathInDotGit))
                     {
-                        this.FileSystemCallbacks.OnFileHardLinkCreated(relativeNewLinkPath);
+                        this.FileSystemCallbacks.OnFileHardLinkCreated(relativeNewLinkPath, relativeExistingFilePath);
```
wilbaker: If there are hardlinks like this:

- Outside Root -> Inside .git
- Inside .git -> Outside root

Will `OnFileHardLinkCreated` be called?  I'm wondering if we need an `IsNullOrEmpty` check before making this call.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
```diff
@@ -353,44 +353,97 @@ public void SupersededFileAddedToModifiedPathsAndSkipWorktreeBitCleared()
...
+
+            string fileHardlinkedInsideRepo = this.Enlistment.GetVirtualPathTo(fileName);
+            this.fileSystem.CreateHardLink(fileHardlinkedInsideRepo, fileOutsideRepo);
+            fileHardlinkedInsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+        }
+
+        [TestCase, Order(19)]
+        public void HardlinkFromInsideRepoToInside()
```
wilbaker: This test name says `FromInsideRepoToInside` but it looks like one of the files being linked is outside of the repo.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
```diff
@@ -353,44 +353,97 @@ public void SupersededFileAddedToModifiedPathsAndSkipWorktreeBitCleared()
...
+
+            string fileHardLinkedOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileName);
+            this.fileSystem.CreateHardLink(fileHardLinkedOutsideRepo, fileInsideRepo);
+            fileHardLinkedOutsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+        }
+
+        [TestCase, Order(20)]
+        public void HardlinkInsideRepo()
```
wilbaker: Also add a test where the target of the hardlink is a file being projected by VFS4G that is not already in the modified paths?

At the end of the test we're validating:

```
            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
```

But `fileName` would already be in the modified paths from when it was created.

ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs
```diff
@@ -205,7 +206,7 @@ public virtual Result StopVirtualizationInstance()
                     return Result.Success;
 
                 case NotificationType.HardLinkCreated:
-                    this.OnHardLinkCreated(relativePath);
+                    this.OnHardLinkCreated(relativePath, relativeFromPath);
```
wilbaker: What would you think of doing the `null` check here?

We'd still have to handle the `null` when logging in the native code, but by doing the check here we could restrict it to `HardLinkCreated`.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I'll throw an option 3 out there as well, making the check in the hardlink case of `OnNotifyOperation`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -679,7 +681,13 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     // whereas messages originating in the kext's vnode handler will only fill
     // the fsid/inode, so we need to look up the path below.
     char pathBuffer[PrjFSMaxPath];
-    if (request.paths[MessagePath_Target] == nullptr)
+    if (request.paths[MessagePath_Target] == nullptr &&
+        request.paths[MessagePath_From] != nullptr)
+    {
+       // MessagePath_Target may be null, when MessagePath_From is populated
+       // In these instances don't try to popute absolute/relative path
```
wilbaker: Typo: `popute` -> `populate`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -679,7 +681,13 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
     // whereas messages originating in the kext's vnode handler will only fill
     // the fsid/inode, so we need to look up the path below.
```
wilbaker: This comment is now a bit outdated:

```
// We expect a non-null request.path for messages sent from the FILEOP handler,
// whereas messages originating in the kext's vnode handler will only fill
// the fsid/inode, so we need to look up the path below.
```

Could you update to include all of the possible scenarios?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1042,29 +1057,35 @@ static PrjFS_Result HandleNewFileInRootNotification(
...
+    const char* relativeFromPath,
     bool isDirectory,
     PrjFS_NotificationType notificationType)
 {
 #ifdef DEBUG
     cout
         << "HandleNewFileInRootNotification: "
-        << absolutePath
-        << " (root-relative: " << relativePath << ")"
+        << (absolutePath == nullptr ? "[NULL]" : absolutePath)
```
wilbaker: Under what conditions would `absolutePath ` be `nullptr` in `HandleNewFileInRootNotification`?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1042,29 +1057,35 @@ static PrjFS_Result HandleNewFileInRootNotification(
...
         << " Pid: " << request->pid
         << " notificationType: " << NotificationTypeToString(notificationType)
         << " isDirectory: " << isDirectory << endl;
 #endif
 
-    // Whenever a new file shows up in the root, we need to check if its ancestor
-    // directories are flagged as in root.  If they are not, flag them as in root and
-    // notify the provider
-    FindNewFoldersInRootAndNotifyProvider(request, relativePath);
+    if (relativePath != nullptr)
```
wilbaker: Same comment here, in `HandleNewFileInRootNotification` I would expect `relativePath` and `absolutePath` to always be non-null, and `relativeFromPath` to always be `nullptr`.

Assuming that's the case I don't think we should add code to handle `nullptr` when it's not expected.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1078,14 +1099,20 @@ static PrjFS_Result HandleFileNotification(
...
+    // Don't send null to providers
+    relativePath = relativePath != nullptr ? relativePath : "";
+    relativeFromPath = relativeFromPath != nullptr ? relativeFromPath : "";
+    
 #ifdef DEBUG
     cout
         << "PrjFSLib.HandleFileNotification: "
-        << absolutePath
+        << (absolutePath == nullptr ? "[NULL]" : absolutePath)
         << " (root-relative: " << relativePath << ")"
```
wilbaker: If we're checking that `absolutePath` is `nullptr`, do we need a similar check for `relativePath`?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1098,6 +1125,7 @@ static PrjFS_Result HandleFileNotification(
     PrjFS_Result result = s_callbacks.NotifyOperation(
         0 /* commandId */,
         relativePath,
```
wilbaker: What is preventing us from sending a null `relativePath` path to the provider?  What will `relativePath` be when a hardlink is created outside of the repo to a file that's inside the repo? 

(As discussed in a separate comment these checks could also be moved into `OnNotifyOperation`)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -781,14 +793,16 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
             }
             cout << endl;
 #endif
+            bool isDirectory = requestHeader->messageType == MessageType_KtoU_NotifyDirectoryRenamed;
```
wilbaker: There is no need to check `requestHeader->messageType == MessageType_KtoU_NotifyDirectoryRenamed` because we're in the `MessageType_KtoU_NotifyFileHardLinkCreated` case.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -758,19 +766,23 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
-            cout << "PrjFSLib.HandleKernelRequest: hard-linked " << (fromPath ?: "[out-of-root path]") << " -> " << (absolutePath ?: "[out-of-root path]") << " (absolute), ";
+            const char* absoluteFromPath = request.paths[MessagePath_From];
+            const char* relativeFromPath = nullptr;
+            if (absoluteFromPath != nullptr)
+            {
+                relativeFromPath = GetRelativePath(absoluteFromPath, s_virtualizationRootFullPath.c_str());
+            }
+            
+#ifdef DEBUG
+            cout << "PrjFSLib.HandleKernelRequest: " << (requestHeader->messageType == MessageType_KtoU_NotifyFileHardLinkCreated ? "hard-linked " : "renamed ");
```
wilbaker: There is no need to check `requestHeader->messageType == MessageType_KtoU_NotifyFileHardLinkCreated` because we're in the `MessageType_KtoU_NotifyFileHardLinkCreated` case of the switch statement.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -781,14 +793,16 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
             }
             cout << endl;
 #endif
+            bool isDirectory = requestHeader->messageType == MessageType_KtoU_NotifyDirectoryRenamed;
             
-            if (relativePath != nullptr)
+            if (relativePath != nullptr || relativeFromPath != nullptr)
             {
                 result = HandleNewFileInRootNotification(
```
wilbaker: We should only be calling `HandleNewFileInRootNotification` if `relativePath` is non-null, otherwise there's not a new file in the root.  If `relativePath` is null we should call `HandleFileNotification` instead to avoid the walk up the tree that checks for new directories.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I see now what's going on.  See comments above when we call `HandleNewFileInRootNotification`.  We should only call `HandleNewFileInRootNotification` when the hardlink results in a new file inside the root.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I see now how this can happen, see comments above where `HandleNewFileInRootNotification` is called.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -218,7 +218,7 @@ protected override bool TryStart(out string error)
             this.virtualizationInstance.OnPreDelete = this.OnPreDelete;
             this.virtualizationInstance.OnNewFileCreated = this.OnNewFileCreated;
             this.virtualizationInstance.OnFileRenamed = this.OnFileRenamed;
-            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
+            this.virtualizationInstance.OnHardLinkCreated = this.NotifyHardLinkCreated;
```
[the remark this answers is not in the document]
jeschu1: Otherwise we clash with the base class's OnHardLinkCreated.  
This seems to be the way NotifyFilePreConvertToFull got around the same problem.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
jeschu1: All of the git folder cases are a bit strange to me, but maybe you can enlighten me here :-)
I'll followup offline

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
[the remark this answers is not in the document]
mjcheetham: ..but since `NotifyHardLinkCreated` just calls through to `this.OnHardLinkCreated` why does `NotifyHardLinkCreated` need to exist anyway? Couldn't we just do..

```suggestion
            this.virtualizationInstance.OnHardLinkCreated = this.OnHardLinkCreated;
```

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
jeschu1: Summary of the offline coversation (as far as I understand)

The .git folder tracking is mainly for timestamps / seeing the if the git repo changed.  We care about new links here (an update could be done via a hardlink / file replacement).  We shouldn't track the existing file for a hardlink.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
jeschu1: The logic will be updated to what was there previously

ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs
jeschu1: Thinking about this more, the other calls expect to not have null (and don't check for it).  I think we should ensure the provider doesn't send null across to be consistent.  

If we think we should check for null here, then we should really update all the calls for the possibility of null.

What are your thoughts regarding?

wilbaker: Ahh good point.

I think I'm leaning toward making sure the kext just doesn't send null for `relativeFromPath` (for 2-path messages).  What do you think?

jeschu1: That's what I got with the latest updates.  Hopefully it's clear.  

For 1 path messages the From can be null, but we shouldn't be using it.  I've been debating if we should send string.empty for non-two path messages, but since those shouldn't use them I'm thinking null for now.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: Updated so only the second path can be null

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: I've updated to send "" for relative paths in this case.  If GetRelativePath gets "", it will return "".

jeschu1:
@wilbaker I've gone with the following approach:

relativePath will be "" (not nullptr) when outside the Virtualization Root.  This is because nullptr has a special meaning for us, as it may look up via fsid.

fromRelative path *can* be nullptr.  This is because it is non-relevant to most message types.  We also don't have any special lookup handling for it like relativePath.  It will be "" in hardlink messages to signify outside the virtualization root.

Let me know what you think and thanks for all the great feedback!!!!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
```diff
@@ -353,42 +353,117 @@ public void SupersededFileAddedToModifiedPathsAndSkipWorktreeBitCleared()
...
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
+
+            string fileNameLink = "OutsideRepoToInside_RepoLink.txt";
+            string fileLinkInsideRepo = this.Enlistment.GetVirtualPathTo(fileNameLink);
+            this.fileSystem.CreateHardLink(fileLinkInsideRepo, fileOutsideRepo);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileNameLink);
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, fileName);
+            fileNameLink.ShouldBeAFile(this.fileSystem);
```
wilbaker: It looks like there is a copy/paste error on this line that is causing your functional test failure:

`fileNameLink.ShouldBeAFile` should be `fileLinkInsideRepo.ShouldBeAFile`

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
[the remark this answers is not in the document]
wilbaker: > ..but since NotifyHardLinkCreated just calls through to this.OnHardLinkCreated why does NotifyHardLinkCreated need to exist anyway? Couldn't we just do..

That's a good point.  I think the parameters here would need to be re-ordered to match `OnHardLinkCreated`:

`NotifyHardLinkCreated(string relativeNewLinkPath, string relativeFromPath)` - > `NotifyHardLinkCreated(string relativeFromPath, string relativeNewLinkPath)`

But that seems like a good change anyway as it would make the API more consistent with Windows:

```
        private void NotifyHardlinkCreated(
            string relativeExistingFilePath,
            string relativeNewLinkPath,
            uint triggeringProcessId,
            string triggeringProcessImageFileName)
        {
            this.OnHardLinkCreated(relativeExistingFilePath, relativeNewLinkPath);
        }
```

> This seems to be the way NotifyFilePreConvertToFull got around the same problem.

If we tweaked the Mac's managed API for `NotifyFilePreConvertToFull` to return `true`/`false` we could use @mjcheetham's suggestion there too, but I'd also be OK leaving that as-is as the code is a bit unrelated to this PR.

GVFS/GVFS.UnitTests/Virtualization/FileSystemCallbacksTests.cs
```diff
@@ -257,7 +257,7 @@ public void FileAndFolderCallbacksScheduleBackgroundTasks()
 
                 this.CallbackSchedulesBackgroundTask(
                     backgroundTaskRunner,
-                    (path) => fileSystemCallbacks.OnFileHardLinkCreated(path),
+                    (path) => fileSystemCallbacks.OnFileHardLinkCreated(path, string.Empty),
```
wilbaker: Update this call to the version of `CallbackSchedulesBackgroundTask` that expects and action that accepts two paths?

```
        private void CallbackSchedulesBackgroundTask(
            MockBackgroundFileSystemTaskRunner backgroundTaskRunner,
            Action<string, string> callback,
            string oldPath,
            string newPath,
            FileSystemTask.OperationType operationType)
        {
```

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
```diff
@@ -353,42 +353,117 @@ public void SupersededFileAddedToModifiedPathsAndSkipWorktreeBitCleared()
...
+            fileInsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
+
+            string fileNameLink = "InsideRepoToOutside_RepoLink.txt";
+            string fileLinkOutsideRepo = Path.Combine(this.Enlistment.EnlistmentRoot, fileNameLink);
+            this.fileSystem.CreateHardLink(fileLinkOutsideRepo, fileInsideRepo);
+            fileLinkOutsideRepo.ShouldBeAFile(this.fileSystem);
+            this.Enlistment.WaitForBackgroundOperations();
+            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
```
wilbaker: Similar comment here (to the earlier comment I left in `HardlinkInsideRepo`).

This test is not validating that the creation of the hardlink resulted in `fileName` being added to modified paths as it was already added above when the file was written:

```
            this.fileSystem.WriteAllText(fileInsideRepo, "Contents for the new file");
            fileInsideRepo.ShouldBeAFile(this.fileSystem);
            this.Enlistment.WaitForBackgroundOperations();
            GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.fileSystem, fileName);
```

Could you update this test to do something like `HardlinkExistingFileInRepo` where the file being linked to is not not currently in modified paths?

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -269,18 +269,15 @@ protected void OnHardLinkCreated(string relativeExistingFilePath, string relativ
...
-                    {
-                        this.FileSystemCallbacks.OnFileHardLinkCreated(relativeNewLinkPath);
-                    }
+                if (newLinkPathInDotGit)
+                {
+                    this.OnDotGitFileOrFolderChanged(relativeNewLinkPath);
+                }
+                else
+                {
+                     this.FileSystemCallbacks.OnFileHardLinkCreated(relativeNewLinkPath, relativeExistingFilePath);
```
wilbaker: It looks like with this change `this.FileSystemCallbacks.OnFileHardLinkCreated` would still be called in this scenario (when it should not be):

- `relativeNewLinkPath == string.Empty`
- `relativeExistingFilePath` -> Path inside the .git folder

And `FileSystemCallbacks.OnFileHardLinkCreated` is not getting called in this scenario (when it should be):

- `relativeNewLinkPath` -> Path in the .git folder
- `relativeExistingFilePath` -> Path inside the working tree

I think you should be able to write a functional test that catches the second scenario (I suspect the first scenario is only a perf issue)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -603,7 +603,7 @@ KEXT_STATIC int HandleFileOpOperation(
                         currentVnode,
                         vnodeFsidInode,
                         newPath,
-                        (messageFromProvider && targetRoot == fromRoot) ? fromPath : nullptr, // Only send "from" path if in the same root.
+                        (messageFromProvider && targetRoot == fromRoot) ? fromPath : "", // Send "", to specify there is a fromPath but it is not relavent
```
wilbaker: Nit:  Suggestion on rewording the comment:

`// Send "" to specify that the fromPath is not in the same root`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -716,8 +718,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
                 << dec << request.messageHeader->fsidInode.inode
                 << " -> '"
                 << pathBuffer
-                << "' -> relative path '"
-                << (relativePath != nullptr ? relativePath : "[NULL]")
+                << "' -> relative path '" << relativePath
```
wilbaker: Why was the `nullptr` check removed here?  I noticed we're still checking on line 775.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: Unresolving this conversation, I still see this check:

`requestHeader->messageType == MessageType_KtoU_NotifyFileHardLinkCreated`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -758,37 +759,42 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
+            }
+            else
             {
                 result = HandleNewFileInRootNotification(
                     requestHeader,
                     relativePath,
                     absolutePath,
-                    false, // isDirectory; TODO: may yet be a directory
+                    relativeFromPath,
+                    false,
```
wilbaker: Could you add this comment back:

`// isDirectory; TODO: may yet be a directory`

Here and above in the call to `HandleFileNotification`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GitFilesTests.cs
jeschu1: I agree that is a better test on it!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: relativeFromPath can be null, while relativePath can not.

Two reasons:
- relativePath as nullptr can mean we need to look up path.  We don't want to confuse ourselves with it having a double meaning.  It is always sent as "" when for paths outside the virtualization root.
- realtiveFromPath is not relevant for all messages, hence the allowance of null.  For messages it is relevant for it is sent as "".

That said, on 775 fromRelativePath should never be nullptr since we are dealing with a hardlink.  It should be "".  I'll update that line good catch!

wilbaker: > relativeFromPath can be null, while relativePath can not.

`relativePath` is the value returned by `GetRelativePath`, and it looks like under certain error conditions `GetRelativePath` can return return `nullptr`.

It looks like we should have been doing a better job of handing that before (e.g. logging an error and returning).  Can you add a check for that scenario (or if it's possible `GetRelativePath` will never return `nullptr` on this code path that would be good to know as well).

jeschu1: As discussed offline if we can't find a Relative path, we could end up in a bad state.

Code is being updated to cleanup in return if relativepath can't be found.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
jeschu1: Let me know what you think now.  I can add a specific functional test for relativeNewLinkPath from .git, and relativeExistingFile path from the working tree, but I'm a bit cautious of adding too many functional tests (already added a bunch).  I would expect this to be a rare case as well.  Let me know what you think.  I can add the test if we really think we need it.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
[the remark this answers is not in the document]
jeschu1: I think in this case less code is good so I've updated with @mjcheetham 's suggestion to call `this.OnHardLinkCreated` .  Thank you @wilbaker for calling out that the params are switched, i'm sure i would have fallen for that!

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
wilbaker: With the current check this scenario will not be handled correctly:

`relativeNewLinkPath` -> .git\\git_file
`relativeExistingFilePath` -> ""

Because it will result in:

`if (!(false && false))`

And `OnFileHardLinkCreated` will be called when neither path is in the working tree.

What would you think of making this check a bit lengthier, but perhaps easier to understand:

```
// call OnFileHardLinkCreated if either relativeNewLinkPath or relativeExistingFilePath are in the working tree
if ( (!newLinkPathInDotGit && !string.IsNullOrEmpty(relativeNewLinkPath) ||
     (!relativePathInDotGit && !string.IsNullOrEmpty(relativeExistingFilePath) )
```

Or even the *slightly* less efficient:

```
bool newLinkInWorkingDir = !newLinkPathInDotGit && !string.IsNullOrEmpty(relativeNewLinkPath);
bool existingFileInWorkingDir = !relativePathInDotGit && !string.IsNullOrEmpty(relativeExistingFilePath);
if ( newLinkInWorkingDir || existingFileInWorkingDir)
```

I think it would be find to always compute `newLinkInWorkingDir` and `existingFileInWorkingDir` to make the condition easier to read.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
```diff
@@ -269,18 +269,17 @@ protected void OnHardLinkCreated(string relativeExistingFilePath, string relativ
         {
             try
             {
-                if (!string.IsNullOrEmpty(relativeNewLinkPath))
+                bool newLinkPathInDotGit = FileSystemCallbacks.IsPathInsideDotGit(relativeNewLinkPath);
+                bool relativePathInDotGit = FileSystemCallbacks.IsPathInsideDotGit(relativeExistingFilePath);
```
wilbaker: Rename `relativePathInDotGit` to something like `existingFilePathInDotGit` to make it clearer which path we're referring to?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -758,36 +767,45 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
...
             
-            if (relativePath != nullptr)
+            if (strcmp(relativePath, "") == 0)
+            {
+                result = HandleFileNotification(
+                    requestHeader,
+                    relativePath,
+                    absolutePath,
+                    relativeFromPath,
+                    false, // isDirectory; TODO: may yet be a directory
```
wilbaker: Not your change but I think we can remove this `TODO`.  It was added when `MessageType_KtoU_NotifyFileHardLinkCreated` could be called for directories, and it no longer will be.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: Since we simplied path to send "" when non-relavent, this comment is valid again.

GVFS/GVFS.Virtualization/FileSystem/FileSystemVirtualizer.cs
jeschu1: I think the code is actually works the way it stands.  

The second scenario was indeed be false, but 
```
if (newLinkPathInDotGit)
{
   this.OnDotGitFileOrFolderChanged(relativeNewLinkPath);
}
```
Would have been called previously making the status cache invalid.  
Calling OnFileHardLinkCreated does nothing for the scenario you called out as far as I can see since the path is a in a git folder.  Please let me know if I'm incorrect. :-)

That said the way the code stands now is obviously confusing so i'm updating it an a separate commit to be more similar with how FileRename is handled.  Since that one does the "IsNullOrEmpty" check in the FileSystemCallback I think it's fine to do the same here.  I"ll wait until your back to give a thumbs up and confirm this new change isn't confusing.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: Since you can do a hardlink for a directory I think it's "OK" to leave.

It may not be a current requirement, but maybe in the future someday?

PR #1189, opened 2019-05-22
Improve Linux event-handling robustness and allow provider-generated events

chrisd8088:
When the provider process generates a permission request event (e.g., by deleting a file in order to conform the working directory during a `git checkout`), this event is then delivered back to the provider itself for validation.

We want to always allow those provider-generated permission-request events to succeed, so we return `PROJFS_ALLOW` for them, which we were previously failing to do.

Separately, we may fail to read the `/proc/{pid}/cmdline` file on Linux for a variety of reasons, not solely that it does not exist.

In all cases, we should catch the exception and simply return an empty string; otherwise, the unhandled exception will unexpectedly halt the provider process.

Thirdly, we've been sending the incorrect path on `OnFileRenamed` events; it is [expected](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L543) to be the destination (a.k.a. target) path.

And lastly, we can remove a chunk of unnecessary overhead and code when handling non-projection events, because our `OnNotifyOperation()` method (which was derived from the [Mac version](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/ProjFS.Mac/PrjFSLib.Mac.Managed/VirtualizationInstance.cs#L180)), doesn't need most of its arguments.  Neither does the Mac one, but its signature has to match [the one implemented](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp#L1091) in its corresponding C++ library.

/cc @kivikakk

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -411,11 +411,7 @@ namespace PrjFSLib.Linux
                 if (nt == NotificationType.FileRenamed || nt == NotificationType.HardLinkCreated)
                 {
                     currentRelativePath = PtrToStringUTF8(ev.TargetPath);
-
-                    if (nt == NotificationType.HardLinkCreated)
-                    {
-                        relativePath = currentRelativePath;
-                    }
+                    relativePath = currentRelativePath;
                 }
 
                 result = this.projfs.GetProjAttrs(
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -397,41 +397,22 @@ namespace PrjFSLib.Linux
             }
 
             bool isDirectory = (ev.Mask & ProjFS.Constants.PROJFS_ONDIR) != 0;
-            string triggeringProcessName = GetProcCmdline(ev.Pid);
-            byte[] providerId = new byte[PlaceholderIdLength];
-            byte[] contentId = new byte[PlaceholderIdLength];
-            string relativePath = PtrToStringUTF8(ev.Path);
-            Result result = Result.Success;
+            string relativePath;
 
-            if (!isDirectory)
+            if (nt == NotificationType.FileRenamed ||
+                nt == NotificationType.HardLinkCreated)
             {
-                string currentRelativePath = relativePath;
-
-                // TODO(Linux): can other intermediate file ops race us here?
-                if (nt == NotificationType.FileRenamed || nt == NotificationType.HardLinkCreated)
-                {
-                    currentRelativePath = PtrToStringUTF8(ev.TargetPath);
-                    relativePath = currentRelativePath;
-                }
-
-                result = this.projfs.GetProjAttrs(
-                    currentRelativePath,
-                    providerId,
-                    contentId);
+                relativePath = PtrToStringUTF8(ev.TargetPath);
+            }
+            else
+            {
+                relativePath = PtrToStringUTF8(ev.Path);
             }
 
-            if (result == Result.Success)
-            {
-                result = this.OnNotifyOperation(
-                    commandId: 0,
-                    relativePath: relativePath,
-                    providerId: providerId,
-                    contentId: contentId,
-                    triggeringProcessId: ev.Pid,
-                    triggeringProcessName: triggeringProcessName,
-                    isDirectory: isDirectory,
-                    notificationType: nt);
-            }
+            Result result = this.OnNotifyOperation(
+                relativePath: relativePath,
+                isDirectory: isDirectory,
+                notificationType: nt);
 
             int ret = -result.ToErrno();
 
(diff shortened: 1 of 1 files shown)


ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -361,7 +361,7 @@ private int HandleNonProjEvent(ref ProjFS.Event ev, bool perm)
             // ignore events triggered by own process to prevent deadlocks
             if (this.IsProviderEvent(ev))
             {
-                return 0;
+                return perm ? (int)ProjFS.Constants.PROJFS_ALLOW : 0;
```
kivikakk: 👍

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -291,7 +291,7 @@ private static string GetProcCmdline(int pid)
                     return parts.Length > 0 ? parts[0] : string.Empty;
                 }
             }
-            catch (IOException ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
+            catch
```
kivikakk: This has me feeling a bit on edge. What other kinds of exceptions are there?

Perhaps we should continue to `return string.Empty;` in all cases, but log the error if the exception isn't an expected `IOException` per the current logic?

chrisd8088: Unfortunately I don't think there's a logging infra this deep in the stack ... the GVFS provider has its `Tracer` framework, but here we don't have access to that, AFAIK.

The errors I've seen tend be due to races and occur when recovering from a sleep state, but all that's in the `gvfs_mount_process....log` is `Unhandled Exception`, I believe.  I've been trying to reproduce them for a few days, but they are something of a heisenbug.

In the end, though, I'm not sure it matters.  The code here has to be robust against all exceptions, regardless of their cause, and the retrieval of the process's command line is -- as far as I can tell -- purely informational.  It's used in the higher-level code in its [logging](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L356) in `OnGetFileStream()` callbacks, and in its [telemetry](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs#L461) [statistics](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs#L297) collected by [`CreatePlaceholders()`](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L644) in its `OnEnumerateDirectory()` callbacks.

Your question did prompt me to notice that there's no need to call this on the non-projection notifications, because the result is unused!  So we can actually pull out a big chunk of code here!  See 97189a1 for a nice cleanup -- thank you for the prompt!!

kivikakk: > The code here has to be robust against all exceptions, regardless of their cause, and the retrieval of the process's command line is — as far as I can tell — purely informational. It's used in the higher-level code only in its [logging](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L356) in `OnGetFileStream()` callbacks, and in the [telemetry](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs#L461) [statistics](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs#L297) collected by [`CreatePlaceholders()`](https://github.com/microsoft/VFSForGit/blob/ceaaa10d19aa587e5eb2cac63dfce74d70f43749/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L644) in its `OnEnumerateDirectory()` callbacks

:+1::+1: Makes sense!

> Your question did prompt me to notice that there's no need to call this on the non-projection notifications, because the result is unused! So we can actually pull out a big chunk of code here! See [d902623](https://github.com/microsoft/VFSForGit/commit/d9026232d0cdd705f508a4780ed2914f00a0c3bf) for a nice cleanup -- thank you for the prompt!!

Super good result!

image

chrisd8088:
I think this one is ready for a review, @derrickstolee -- thanks very much, and have a good weekend!  (And thanks for fixing those CI issues! 👍)

PR #1190, opened 2019-05-22
Add notarize shell script for notarizing the kext with Apple

kewillford:
This is a script taken from [BuildXL](https://github.com/microsoft/BuildXL) that will take a zip file and notarize the contained files with Apple using their notarization service.  This will be used to send the kext and get it notarized so that it will work on [macOS 10.14.5 and above](https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution)


nickgra:
Should we leave a note in the script that we sourced the script from the BuildXL repo and leave a link to it in their repo?

Issue #526, opened 2018-11-16
Mac: Use the WILL_RENAME file op event on Mojave to determine if PreDelete is part of a rename


wilbaker:
In Mojave Apple added a new WILL_RENAME file op event that can be used to determine if DELETE vnode operations are part of a rename.

VFS4G should use this file op event to determine if it needs to hydrate a file and either expand/hydrate directories (or block them from being deleted)

wilbaker:
Another option here is that WILL_RENAME could be used to determine if we should prevent the rename of directories to match the behavior on Windows

wilbaker:
@pmj before starting this work please confirm with @jrbriggs that we have confirmation that we can set Mojave as the minimum supported macOS version.

/cc: @nickgra

pmj:
It looks like the `KAUTH_FILEOP_WILL_RENAME` -> `KAUTH_VNODE_DELETE` pairing is reliable (i.e. we *always* get a `KAUTH_VNODE_DELETE` after a `KAUTH_FILEOP_WILL_RENAME`), and intentionally so:

https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/vfs/vfs_subr.c#L6255

I've not just taken the comment at face value but looked at the code for `vnode_authorize()` and it appears each registered vnode listener kext does indeed get a chance to see the event even if an earlier kext's listener has already denied the operation.

This is good news, as it simplifies the implementation somewhat.

pmj:
With #1193 merged, there's nothing left to do here as far as I'm aware.

PR #1193, opened 2019-05-23
Mac ProjFS: Only hydrate/expand when deleting due to rename

pmj:
This change utilises the KAUTH_FILEOP_WILL_RENAME event which was introduced with Mojave/10.14 to determine if a KAUTH_VNODE_DELETE authorisation check is due to a rename() operation. In this case, the file or directory genuinely needs hydrating, otherwise hydration can be skipped.

This change also adds unit tests for the new rename operation tracking code, and adapts existing unit tests so hydration is not expected when a KAUTH_VNODE_DELETE check occurs outside of a rename() operation.

This implements #526.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -248,6 +248,14 @@ KEXT_STATIC int HandleVnodeOperation(
     bool isDeleteAction = false;
     bool isDirectory = false;
 
+    {
+        PerfSample considerVnodeSample(&perfTracer, PrjFSPerfCounter_VnodeOp_BasicVnodeChecks);
+        if (!VnodeIsEligibleForEventHandling(currentVnode))
+        {
+            goto CleanupAndReturn;
+        }
+    }
+
     if (!ShouldHandleVnodeOpEvent(
             &perfTracer,
             context,
@@ -818,15 +826,6 @@ KEXT_STATIC bool ShouldHandleVnodeOpEvent(
             //    request in progress is advisory, rather than authoritative.  Listeners
             //    performing consequential work (i.e. not strictly checking authorisation)
             //    may test this flag to avoid performing unnecessary work."
-            *kauthResult = KAUTH_RESULT_DEFER;
-            return false;
-        }
-    }
-    
-    {
-        PerfSample considerVnodeSample(perfTracer, PrjFSPerfCounter_VnodeOp_ShouldHandle_BasicVnodeChecks);
-        if (!VnodeIsEligibleForEventHandling(vnode))
-        {
             *kauthResult = KAUTH_RESULT_DEFER;
             return false;
         }
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -2,7 +2,10 @@
 #include <sys/kauth.h>
 #include <sys/proc.h>
 #include <kern/assert.h>
+#include <libkern/version.h>
+#include <kern/thread.h>
 
+#include "KauthHandlerPrivate.hpp"
 #include "public/PrjFSCommon.h"
 #include "public/PrjFSPerfCounter.h"
 #include "public/PrjFSXattrs.h"
@@ -19,15 +22,28 @@
 #include "ProviderMessaging.hpp"
 #include "VnodeCache.hpp"
 #include "Memory.hpp"
+#include "ArrayUtilities.hpp"
 
 #ifdef KEXT_UNIT_TESTING
 #include "KauthHandlerTestable.hpp"
 #endif
 
+template <typename T, typename MIN_T, typename MAX_T>
+    auto clamp(const T& value, const MIN_T& min, const MAX_T& max)
+{
+    return value < min ? min : (value > max ? max : value);
+}
+
 enum ProviderCallbackPolicy
 {
     CallbackPolicy_AllowAny,
     CallbackPolicy_UserInitiatedOnly,
+};
+
+struct PendingRenameOperation
+{
+    vnode_t vnode;
+    thread_t thread;
 };
 
 // Function prototypes
@@ -101,11 +117,19 @@ KEXT_STATIC bool ShouldHandleFileOpEvent(
     VirtualizationRootHandle* root,
     int* pid);
 
+KEXT_STATIC bool InitPendingRenames();
+KEXT_STATIC void CleanupPendingRenames();
+KEXT_STATIC void ResizePendingRenames(uint32_t newMaxPendingRenames);
+
 // State
 static kauth_listener_t s_vnodeListener = nullptr;
 static kauth_listener_t s_fileopListener = nullptr;
 
 static atomic_int s_numActiveKauthEvents;
+static SpinLock s_renameLock;
+static PendingRenameOperation* s_pendingRenames = nullptr;
+KEXT_STATIC uint32_t s_pendingRenameCount = 0;
+KEXT_STATIC uint32_t s_maxPendingRenames = 0;
 
 // Public functions
 kern_return_t KauthHandler_Init()
@@ -130,6 +154,11 @@ kern_return_t KauthHandler_Init()
         goto CleanupAndFail;
     }
     
+    if (!InitPendingRenames())
+    {
+        goto CleanupAndFail;
+    }
+
     s_vnodeListener = kauth_listen_scope(KAUTH_SCOPE_VNODE, HandleVnodeOperation, nullptr);
     if (nullptr == s_vnodeListener)
     {
@@ -178,6 +207,8 @@ kern_return_t KauthHandler_Cleanup()
     
     WaitForListenerCompletion();
 
+    CleanupPendingRenames();
+    
     if (VnodeCache_Cleanup())
     {
         result = KERN_FAILURE;
@@ -216,6 +247,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
     }
 }
 
+KEXT_STATIC bool InitPendingRenames()
+{
+    if (version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave) // Only need to track renames on Mojave and newer
+    {
+        s_renameLock = SpinLock_Alloc();
+        s_maxPendingRenames = 8; // Arbitrary choice, should be maximum number of expected concurrent threads performing renames, but array will resize on demand
+        s_pendingRenameCount = 0;
+        s_pendingRenames = Memory_AllocArray<PendingRenameOperation>(s_maxPendingRenames);
+        if (!SpinLock_IsValid(s_renameLock) || s_pendingRenames == nullptr)
+        {
+            return false;
+        }
+
+        Array_DefaultInit(s_pendingRenames, s_maxPendingRenames);
+    }
+    
+    return true;
+}
+
+KEXT_STATIC void CleanupPendingRenames()
+{
+    if (version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave)
+    {
+        if (SpinLock_IsValid(s_renameLock))
+        {
+            SpinLock_FreeMemory(&s_renameLock);
+        }
+
+        if (s_pendingRenames != nullptr)
+        {
+            assert(s_pendingRenameCount == 0);
+            Memory_FreeArray(s_pendingRenames, s_maxPendingRenames);
+            s_pendingRenames = nullptr;
+            s_maxPendingRenames = 0;
+        }
+    }
+}
+
+KEXT_STATIC void ResizePendingRenames(uint32_t newMaxPendingRenames)
+{
+    PendingRenameOperation* newArray = Memory_AllocArray<PendingRenameOperation>(newMaxPendingRenames);
+    assert(newArray != nullptr);
+    PendingRenameOperation* arrayToFree = nullptr;
+    uint32_t arrayToFreeLength = 0;
+    
+    SpinLock_Acquire(s_renameLock);
+    {
+        if (newMaxPendingRenames > s_maxPendingRenames)
+        {
+            Array_CopyElements(newArray, s_pendingRenames, s_maxPendingRenames);
+            Array_DefaultInit(newArray + s_maxPendingRenames, newMaxPendingRenames - s_maxPendingRenames);
+            
+            arrayToFree = s_pendingRenames;
+            arrayToFreeLength = s_maxPendingRenames;
+            
+            s_pendingRenames = newArray;
+            s_maxPendingRenames = newMaxPendingRenames;
+        }
+        else
+        {
+            arrayToFree = newArray;
+            arrayToFreeLength = newMaxPendingRenames;
+        }
+    }
+    SpinLock_Release(s_renameLock);
+    
+    if (arrayToFree != nullptr)
+    {
+        assert(arrayToFreeLength > 0);
+        Memory_FreeArray(arrayToFree, arrayToFreeLength);
+    }
+}
+
+KEXT_STATIC void RecordPendingRenameOperation(vnode_t vnode)
+{
+    assertf(version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave, "This function should only be called from the KAUTH_FILEOP_WILL_RENAME handler, which is only supported by Darwin 18 (macOS 10.14 Mojave) and newer (version_major = %u)", version_major);
+    thread_t myThread = current_thread();
+    
+    bool resizeTable;
+    do
+    {
+        resizeTable = false;
+        uint32_t resizeTableLength = 0;
+        
+        SpinLock_Acquire(s_renameLock);
+        {
+            if (s_pendingRenameCount < s_maxPendingRenames)
+            {
+                s_pendingRenames[s_pendingRenameCount].thread = myThread;
+                s_pendingRenames[s_pendingRenameCount].vnode = vnode;
+                ++s_pendingRenameCount;
+            }
+            else
+            {
+                for (uint32_t i = 0; i < s_pendingRenameCount; ++i)
+                {
+                    assert(s_pendingRenames[i].thread != myThread);
+                }
+
+                resizeTable = true;
+                
+                resizeTableLength = static_cast<uint32_t>(clamp(s_maxPendingRenames * UINT64_C(2), 1u, UINT32_MAX));
+                assert(resizeTableLength > s_maxPendingRenames);
+            }
+        }
+        SpinLock_Release(s_renameLock);
+        
+        if (resizeTable)
+        {
+            if (resizeTableLength > 16)
+            {
+                KextLog_Error("Warning: RecordPendingRenameOperation is causing pending rename array resize to %u items.", resizeTableLength);
+            }
+            ResizePendingRenames(resizeTableLength);
+        }
+    } while (resizeTable);
+}
+
+KEXT_STATIC bool DeleteOpIsForRename(vnode_t vnode)
+{
+    if (version_major < PrjFSDarwinMajorVersion::MacOS10_14_Mojave)
+    {
+        // High Sierra and earlier do not support WILL_RENAME notification, so we have to assume any delete may be caused by a rename
+        assert(s_pendingRenameCount == 0);
+        assert(s_maxPendingRenames == 0);
+        return true;
+    }
+    
+    bool isRename = false;
+    
+    thread_t myThread = current_thread();
+    
+    SpinLock_Acquire(s_renameLock);
+    {
+        for (uint32_t i = 0; i < s_pendingRenameCount; ++i)
+        {
+            if (s_pendingRenames[i].thread == myThread)
+            {
+                isRename = true;
+                assert(s_pendingRenames[i].vnode == vnode);
+                --s_pendingRenameCount;
+                if (i != s_pendingRenameCount)
+                {
+                    s_pendingRenames[i] = s_pendingRenames[s_pendingRenameCount];
+                }
+                
+                s_pendingRenames[s_pendingRenameCount] = PendingRenameOperation{};
+                break;
+            }
+        }
+    }
+    SpinLock_Release(s_renameLock);
+
+    return isRename;
+}
+
 // Private functions
 KEXT_STATIC int HandleVnodeOperation(
     kauth_cred_t    credential,
@@ -247,6 +434,7 @@ KEXT_STATIC int HandleVnodeOperation(
     char procname[MAXCOMLEN + 1] = "";
     bool isDeleteAction = false;
     bool isDirectory = false;
+    bool isRename = false;
 
     {
         PerfSample considerVnodeSample(&perfTracer, PrjFSPerfCounter_VnodeOp_BasicVnodeChecks);
@@ -254,6 +442,15 @@ KEXT_STATIC int HandleVnodeOperation(
         {
             goto CleanupAndReturn;
         }
+    }
+
+    isDeleteAction = ActionBitIsSet(action, KAUTH_VNODE_DELETE);
+    if (isDeleteAction)
+    {
+        // This removes a matching entry from the array, so must run under the same
+        // conditions as the original RecordPendingRenameOperation call - hence early
+        // in the callback.
+        isRename = DeleteOpIsForRename(currentVnode);
     }
 
     if (!ShouldHandleVnodeOpEvent(
@@ -270,22 +467,21 @@ KEXT_STATIC int HandleVnodeOperation(
         goto CleanupAndReturn;
     }
     
-    isDeleteAction = ActionBitIsSet(action, KAUTH_VNODE_DELETE);
     isDirectory = vnode_isdir(currentVnode);
     
     if (isDirectory)
     {
-        if (ActionBitIsSet(
+        if (isRename ||
+            ActionBitIsSet(
                 action,
                 KAUTH_VNODE_LIST_DIRECTORY |
                 KAUTH_VNODE_SEARCH |
                 KAUTH_VNODE_READ_SECURITY |
                 KAUTH_VNODE_READ_ATTRIBUTES |
-                KAUTH_VNODE_READ_EXTATTRIBUTES |
-                KAUTH_VNODE_DELETE))
+                KAUTH_VNODE_READ_EXTATTRIBUTES))
         {
-            // Recursively expand directory on delete to ensure child placeholders are created before rename operations
-            if (isDeleteAction)
+            // Recursively expand directory on rename as user will expect the moved directory to have the same contents as in its original location
+            if (isRename)
             {
                 // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
                 if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
@@ -339,7 +535,8 @@ KEXT_STATIC int HandleVnodeOperation(
     }
     else
     {
-        if (ActionBitIsSet(
+        if (isRename || // Hydrate before a file is moved as the user will not expect an empty file at the new location
+            ActionBitIsSet(
                 action,
                 KAUTH_VNODE_READ_ATTRIBUTES |
                 KAUTH_VNODE_WRITE_ATTRIBUTES |
@@ -348,7 +545,6 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_READ_DATA |
                 KAUTH_VNODE_WRITE_DATA |
                 KAUTH_VNODE_EXECUTE |
-                KAUTH_VNODE_DELETE | // Hydrate on delete to ensure files are hydrated before rename operations
                 KAUTH_VNODE_APPEND_DATA))
         {
             if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
@@ -763,6 +959,15 @@ KEXT_STATIC int HandleFileOpOperation(
                 &kauthError))
         {
             goto CleanupAndReturn;
+        }
+    }
+    else if (KAUTH_FILEOP_WILL_RENAME == action)
+    {
+        currentVnode = reinterpret_cast<vnode_t>(arg0);
+        if (VnodeIsEligibleForEventHandling(currentVnode))
+        {
+            // Records thread/vnode as rename() operation in progress, enabling optimisation in subsequent DELETE vnode listener handler
+            RecordPendingRenameOperation(currentVnode);
         }
     }
     
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -623,7 +623,7 @@ KEXT_STATIC int HandleVnodeOperation(
                 root,
                 isDirectory ?
                 MessageType_KtoU_NotifyDirectoryPreDelete :
-                MessageType_KtoU_NotifyFilePreDelete,
+                isRename ? MessageType_KtoU_NotifyFilePreDeleteFromRename : MessageType_KtoU_NotifyFilePreDelete,
                 currentVnode,
                 vnodeFsidInode,
                 nullptr, // path not needed, use fsid/inode
--- a/ProjFS.Mac/PrjFSKext/Locks.cpp
+++ b/ProjFS.Mac/PrjFSKext/Locks.cpp
@@ -180,3 +180,30 @@ void RWLock_DropExclusiveToShared(RWLock& rwLock)
 #endif
 }
 
+
+SpinLock SpinLock_Alloc()
+{
+    return SpinLock { lck_spin_alloc_init(s_lockGroup, LCK_ATTR_NULL) };
+}
+
+void SpinLock_FreeMemory(SpinLock* lock)
+{
+    assert(lock->p != nullptr);
+    lck_spin_free(lock->p, s_lockGroup);
+    lock->p = nullptr;
+}
+
+bool SpinLock_IsValid(SpinLock lock)
+{
+    return lock.p != nullptr;
+}
+
+void SpinLock_Acquire(SpinLock lock)
+{
+    lck_spin_lock(lock.p);
+}
+
+void SpinLock_Release(SpinLock lock)
+{
+    lck_spin_unlock(lock.p);
+}
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -853,5 +853,21 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 0);
 }
 
+- (void) testIneligibleFilesystemType
+{
+    shared_ptr<mount> testMountNone = mount::Create("msdos", fsid_t{}, 0);
+    shared_ptr<vnode> testVnodeNone = vnode::Create(testMountNone, "/Volumes/USBSTICK");
+    XCTAssertEqual(
+        KAUTH_RESULT_DEFER,
+        HandleVnodeOperation(
+            nullptr,
+            nullptr,
+            KAUTH_VNODE_ADD_FILE,
+            reinterpret_cast<uintptr_t>(context),
+            reinterpret_cast<uintptr_t>(testVnodeNone.get()),
+            0,
+            0));
+    XCTAssertFalse(MockCalls::DidCallFunction(ProviderMessaging_TrySendRequestAndWaitForResponse));
+}
 
 @end
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -430,7 +430,7 @@ static void TestForAllSupportedDarwinVersions(void(^testBlock)(void))
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
             _,
-            MessageType_KtoU_NotifyFilePreDelete,
+            MessageType_KtoU_NotifyFilePreDeleteFromRename,
             testFileVnode.get(),
             _,
             _,
@@ -550,7 +550,7 @@ static void TestForAllSupportedDarwinVersions(void(^testBlock)(void))
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
            _,
-            MessageType_KtoU_NotifyFilePreDelete,
+            MessageType_KtoU_NotifyFilePreDeleteFromRename,
             testFileVnode.get(),
             _,
             _,
@@ -574,7 +574,7 @@ static void TestForAllSupportedDarwinVersions(void(^testBlock)(void))
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
            _,
-            MessageType_KtoU_NotifyFilePreDelete,
+            MessageType_KtoU_NotifyFilePreDeleteFromRename,
             otherTestFileVnode.get(),
             _,
             _,
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -21,7 +21,11 @@
 using std::make_tuple;
 using std::shared_ptr;
 using std::vector;
+using std::string;
 using KextMock::_;
+
+// Darwin version of running kernel
+int version_major = PrjFSDarwinMajorVersion::MacOS10_14_Mojave;
 
 class PrjFSProviderUserClient
 {
@@ -35,15 +39,31 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     vnode->xattrs.insert(make_pair(PrjFSFileXAttrName, rootXattrData));
 }
 
+static void TestForDarwinVersionRange(int versionMin, int versionMax, void(^testBlock)(void))
+{
+    const int savedVersion = version_major;
+    for (int version = versionMin; version <= versionMax; ++version)
+    {
+        version_major = version;
+        testBlock();
+    }
+    version_major = savedVersion;
+}
+
+static void TestForAllSupportedDarwinVersions(void(^testBlock)(void))
+{
+    TestForDarwinVersionRange(PrjFSDarwinMajorVersion::MacOS10_13_HighSierra, PrjFSDarwinMajorVersion::MacOS10_15_Catalina, testBlock);
+}
+
 @interface HandleVnodeOperationTests : PFSKextTestCase
 @end
 
 @implementation HandleVnodeOperationTests
 {
     vfs_context_t context;
-    const char* repoPath;
-    const char* filePath;
-    const char* dirPath;
+    string repoPath;
+    string filePath;
+    string dirPath;
     VirtualizationRootHandle dummyRepoHandle;
     PrjFSProviderUserClient dummyClient;
     pid_t dummyClientPid;
@@ -75,7 +95,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     testDirVnode = testMount->CreateVnodeTree(dirPath, VDIR);
 
     // Register provider for the repository path (Simulate a mount)
-    VirtualizationRootResult result = VirtualizationRoot_RegisterProviderForPath(&dummyClient, dummyClientPid, repoPath);
+    VirtualizationRootResult result = VirtualizationRoot_RegisterProviderForPath(&dummyClient, dummyClientPid, repoPath.c_str());
     XCTAssertEqual(result.error, 0);
     self->dummyRepoHandle = result.root;
 
@@ -102,6 +122,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     cacheWrapper.FreeCache();
     
     VirtualizationRoots_Cleanup();
+    CleanupPendingRenames();
     vfs_context_rele(context);
     MockVnodes_CheckAndClear();
     MockCalls::Clear();
@@ -125,8 +146,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     testFileVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
     SetPrjFSFileXattrData(testFileVnode);
     
-    const int actionCount = 9;
-    kauth_action_t actions[actionCount] =
+    kauth_action_t actions[] =
     {
         KAUTH_VNODE_READ_ATTRIBUTES,
         KAUTH_VNODE_WRITE_ATTRIBUTES,
@@ -135,10 +155,10 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
         KAUTH_VNODE_READ_DATA,
         KAUTH_VNODE_WRITE_DATA,
         KAUTH_VNODE_EXECUTE,
-        KAUTH_VNODE_DELETE,
         KAUTH_VNODE_APPEND_DATA,
     };
-    
+    const int actionCount = std::extent<decltype(actions)>::value;
+
     for (int i = 0; i < actionCount; i++)
     {
         XCTAssertTrue(HandleVnodeOperation(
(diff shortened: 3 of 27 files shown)


pmj:
**Review Notes:**

You'll notice that the first commit in this series, which actually implements the functionality, passes all tests, while the second commit does not - this is the one that kills High Sierra compatibility, and it looks like the Mac Functional tests are still running on a HS machine. I've run the functional tests locally on a Mojave machine too, and it's all working fine.

Personally, as I've said, I'm happy to maintain HS compatibility for the moment - it's quite isolated in the kext, and I don't foresee anything starting to depend on this optimisation in the near future.

I also don't know what the status of the "large build" is with regard to Mojave, perhaps @nickgra can comment on that and what we should do with regard to that in this PR?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +255,102 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
 
+static void ResizePendingRenames(uint32_t newMaxPendingRenames)
+{
+    PendingRenameOperation* newArray = Memory_AllocArray<PendingRenameOperation>(newMaxPendingRenames);
+    PendingRenameOperation* arrayToFree = nullptr;
+    uint32_t arrayToFreeLength = 0;
+    
+    SpinLock_Acquire(s_renameLock);
+    {
+        if (newMaxPendingRenames >= s_maxPendingRenames)
```
jeschu1: Should this just be '>'?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +255,102 @@ KEXT_STATIC void UseMainForkIfNamedStream(
     }
 }
 
+static void ResizePendingRenames(uint32_t newMaxPendingRenames)
+{
+    PendingRenameOperation* newArray = Memory_AllocArray<PendingRenameOperation>(newMaxPendingRenames);
```
jeschu1: Does this memory get freed if newMaxPendingRenames < s_maxPendingRenames

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +255,102 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+                if (s_pendingRenames[i].thread == nullptr)
+                {
+                    s_pendingRenames[i].thread = myThread;
+                    s_pendingRenames[i].vnode = vnode;
+                    inserted = true;
+                    break;
+                }
+                else
+                {
+                    assert(s_pendingRenames[i].thread != myThread);
```
jeschu1: Assert cause Kernel panics right? 
Could this happen:
  1. Rename A on Thread 1
  2. Rename B on Thread 1
  3. PreDelete A on Thread 1
Want to verify how these events come in and are processed.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +255,102 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+            {
+                resizeTable = true;
+                resizeTableLength = MAX(s_maxPendingRenames * 2u, 1u);
+            }
+        }
+        SpinLock_Release(s_renameLock);
+        
+        if (resizeTable)
+        {
+            ResizePendingRenames(resizeTableLength);
```
jeschu1: Since we are not under lock here, I think 2 threads could have gone thru the previous loop.  Then they would both try to do the resize.  Does that sound correct?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -129,7 +143,19 @@ kern_return_t KauthHandler_Init()
...
+
+    s_renameLock = SpinLock_Alloc();
+    s_maxPendingRenames = 8; // Arbitrary choice, should be maximum number of expected concurrent threads performing renames, but array will resize on demand
+    s_pendingRenames = Memory_AllocArray<PendingRenameOperation>(s_maxPendingRenames);
+    if (!SpinLock_IsValid(s_renameLock) || s_pendingRenames == nullptr)
+    {
+        goto CleanupAndFail;
+    }
+
+    Array_DefaultInit(s_pendingRenames, s_maxPendingRenames);
```
jeschu1: Curious why we fill the array with objects rather than allowing empty slots to be null.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -129,7 +143,19 @@ kern_return_t KauthHandler_Init()
     {
         goto CleanupAndFail;
     }
-    
+
+
+    s_renameLock = SpinLock_Alloc();
+    s_maxPendingRenames = 8; // Arbitrary choice, should be maximum number of expected concurrent threads performing renames, but array will resize on demand
```
jeschu1: I could see some applications doing a bunch of renames (this is how a lot of editors seem to save).   Do we know how many threads the Kernel runs with?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -247,6 +382,14 @@ KEXT_STATIC int HandleVnodeOperation(
     char procname[MAXCOMLEN + 1] = "";
     bool isDeleteAction = false;
     bool isDirectory = false;
+    bool isRename = true;
+
+    isDeleteAction = ActionBitIsSet(action, KAUTH_VNODE_DELETE);
+    if (isDeleteAction && VnodeIsEligibleForEventHandling(currentVnode))
+    {
+        isRename = DeleteOpIsForRename(currentVnode);
+        KextLog_FileInfo(currentVnode, "HandleVnodeOperation thread = %p, vnode = %p: KAUTH_VNODE_DELETE: isRename = %s", current_thread(), currentVnode, isRename ? "YES" : "NO");
```
jeschu1: Debugging info?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -756,6 +898,18 @@ KEXT_STATIC int HandleFileOpOperation(
...
+    else if (KAUTH_FILEOP_WILL_RENAME == action)
+    {
+        currentVnode = reinterpret_cast<vnode_t>(arg0);
+        if (VnodeIsEligibleForEventHandling(currentVnode))
+        {
+            const char* fromPath = reinterpret_cast<const char*>(arg1);
+            const char* toPath = reinterpret_cast<const char*>(arg2);
+            KextLog_Info("HandleFileOpOperation: KAUTH_FILEOP_WILL_RENAME vnode %p '%s' -> '%s'", currentVnode, fromPath, toPath);
+
+            RecordPendingRenameOperation(currentVnode);
```
jeschu1: Maybe a comment here, 

// We store rename operations, so we can identify if the following PreDelete is part of a rename

jeschu1:
Just an FYI, I think our target users are not quite on Mojave although that will be a requirement for go-live.  @nickgra for reference

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: There is a kernel thread for every thread in every process, plus it can create its own. So there are usually thousands. Very few will be between the kauth fileop and vnode authorisation calls in the rename code at any time though, and if there are more than 8, well, that's why allow resizing.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: In C++, initialising a variable of type `T` with `T()` or `T{}` means it is "value-initialised", which ends up being zero-initialisation for primitive (member) types (otherwise, primitive members are zero-initialised, then the default constructor is called). So in this case `PendingRenameOperation()` means constructing a `PendingRenameOperation` which has both its fields (which are pointers, so primitive types) zero-initialised. (i.e. `.vnode` and `.thread` are both `nullptr`, which is what we want.)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes, it probably should. We should probably also not be reinventing array resizing so many times… (As I've mentioned in other PRs, my preferred option would be a resizeable array template.)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: This won't happen because a call with `KAUTH_FILEOP_WILL_RENAME` is always immediately followed by a `KAUTH_VNODE_DELETE`.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes, that's something we just have to deal with.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: > Does this memory get freed if newMaxPendingRenames < s_maxPendingRenames

Looks like I messed that up, thanks for catching it! This code went through a number of iterations, and it looks like I lost sight of that edge case.

> What prevents newMaxPendingRenames from becoming arbitrary large?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yep, looks like that one made it past my removal pass.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +255,102 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+
+static bool DeleteOpIsForRename(vnode_t vnode)
+{
+    bool isRename = false;
+    
+    thread_t myThread = current_thread();
+    
+    SpinLock_Acquire(s_renameLock);
+    {
+        for (uint32_t i = 0; i < s_maxPendingRenames; ++i)
```
pmj: Heads up: I will change this logic slightly so that we don't have to scan the whole array every time. (which is mostly empty most of the time)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -101,11 +110,16 @@ KEXT_STATIC bool ShouldHandleFileOpEvent(
...
+static void ResizePendingRenames(uint32_t newMaxPendingRenames);
+
 // State
 static kauth_listener_t s_vnodeListener = nullptr;
 static kauth_listener_t s_fileopListener = nullptr;
 
 static atomic_int s_numActiveKauthEvents;
+static SpinLock s_renameLock;
+static PendingRenameOperation* s_pendingRenames = nullptr;
+static uint32_t s_maxPendingRenames = 0;
```
jeschu1: How about s_pendingRenamesSize versus s_maxPendingRenames?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Is the issue that we can't resize under lock?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I'm not saying it's the best naming convention, but we've [used this naming pattern elsewhere](https://github.com/microsoft/VFSForGit/blob/8c61c1ad051c97545a2dd2b6b0e22fca9b3b7fa1/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp#L26).

The other thing about "size" is that it's potentially ambiguous. Is it a number of array items? Size in bytes?

jeschu1: Fair point, there is precedence.  I know what it means now.

pmj:
> Just an FYI, I think our target users are not quite on Mojave although that will be a requirement for go-live. @nickgra for reference

That's very much the impression I'm getting. When we talked about this change out-of-band before I embarked on it, everyone seemed very much against having any backwards compatibility code in it, which is why I'm vocally calling attention to the issue. I don't think we can realistically integrate this change if we don't have explicit HS support, at least initially.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes, blocking memory allocations are iffy while holding a lock as you risk deadlocking. (There's an in-depth explanation somewhere in #500.) During a spinlock they're a complete no-go.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: This should be fixed now.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: > > Does this memory get freed if newMaxPendingRenames < s_maxPendingRenames
> 
> Looks like I messed that up, thanks for catching it! This code went through a number of iterations, and it looks like I lost sight of that edge case.

This is now fixed.

pmj:
This latest version, rebased on master, fixes the first set of review feedback as well as a bunch of bugs that were surfaced by the now-working `assert()` implementation (#1200). I've also reworked the array data structure so that empty entries are not repeatedly searched.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,158 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+        if (resizeTable)
+        {
+            ResizePendingRenames(resizeTableLength);
+        }
+    } while (resizeTable);
+}
+
+static bool DeleteOpIsForRename(vnode_t vnode)
+{
+    if (version_major < 18)
```
jeschu1: Am I right in assuming your updates the code is backwards compatable?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,158 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+    SpinLock_Acquire(s_renameLock);
+    {
+        for (uint32_t i = 0; i < s_pendingRenameCount; ++i)
+        {
+            if (s_pendingRenames[i].thread == myThread)
+            {
+                isRename = true;
+                assert(s_pendingRenames[i].vnode == vnode);
+                --s_pendingRenameCount;
+                if (i != s_pendingRenameCount)
```
jeschu1: The idea here is that you lower pendingRenameCount by one, and swap the new free index with what was at the end.

Code looks correct, but it's a lot to take in.  Some comments here would be helpful.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -247,6 +424,14 @@ KEXT_STATIC int HandleVnodeOperation(
     char procname[MAXCOMLEN + 1] = "";
     bool isDeleteAction = false;
     bool isDirectory = false;
+    bool isRename = true;
+
+    isDeleteAction = ActionBitIsSet(action, KAUTH_VNODE_DELETE);
+    if (isDeleteAction && VnodeIsEligibleForEventHandling(currentVnode))
```
jeschu1: Can we do this check after the ShouldHandleVnodeOpEvent?

Then I don't think we'd need a VnodeIsEligibleForEventHandling check.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -302,8 +486,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_SEARCH |
                 KAUTH_VNODE_READ_SECURITY |
                 KAUTH_VNODE_READ_ATTRIBUTES |
-                KAUTH_VNODE_READ_EXTATTRIBUTES |
-                KAUTH_VNODE_DELETE))
+                KAUTH_VNODE_READ_EXTATTRIBUTES)
+            || (isDeleteAction && isRename))
         {
```
jeschu1: nit:

``             KAUTH_VNODE_READ_EXTATTRIBUTES)  || 
               (isDeleteAction && isRename)) ``

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -369,8 +553,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_READ_DATA |
                 KAUTH_VNODE_WRITE_DATA |
                 KAUTH_VNODE_EXECUTE |
-                KAUTH_VNODE_DELETE | // Hydrate on delete to ensure files are hydrated before rename operations
-                KAUTH_VNODE_APPEND_DATA))
+                KAUTH_VNODE_APPEND_DATA)
+            || (isDeleteAction && isRename)) // Hydrate on delete to ensure files are hydrated before rename operations
```
jeschu1: nit: Same as above "||" should go after the KAUTH_VNODE_APPEND_DATA

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -756,6 +940,18 @@ KEXT_STATIC int HandleFileOpOperation(
...
         }
     }
+    else if (KAUTH_FILEOP_WILL_RENAME == action)
+    {
+        currentVnode = reinterpret_cast<vnode_t>(arg0);
+        if (VnodeIsEligibleForEventHandling(currentVnode))
+        {
+            const char* fromPath = reinterpret_cast<const char*>(arg1);
+            const char* toPath = reinterpret_cast<const char*>(arg2);
+            KextLog_Info("HandleFileOpOperation: KAUTH_FILEOP_WILL_RENAME vnode %p '%s' -> '%s'", currentVnode, fromPath, toPath);
```
jeschu1: nit: placeholder to make sure we remove the debugging info above.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,158 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+    if (arrayToFree != nullptr)
+    {
+        assert(arrayToFreeLength > 0);
+        Memory_FreeArray(arrayToFree, arrayToFreeLength);
+    }
+}
+
+static void RecordPendingRenameOperation(vnode_t vnode)
+{
+    assertf(version_major >= 18, "KAUTH_FILEOP_WILL_RENAME is only supported by Darwin 18 and newer (version_major = %u)", version_major);
```
jeschu1: It looks like we added some backwards compat in this PR, but due to this line I think we panic if we aren't on 18+.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -642,4 +642,13 @@ - (void)testShouldHandleFileOpEvent_VnodeCacheUpdated {
     }
 }
 
+- (void)testResizePendingRenamesRace
+{
+    // No assertions needed here, the assertions in the kext code and mock environment suffice
+    InitPendingRenames();
+    ResizePendingRenames(16);
+    ResizePendingRenames(16);
```
jeschu1: Would it be possible to test that s_pendingRenames is indeed 16, as is the actual array size?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -166,6 +169,105 @@ - (void) testEmptyFileHydrates {
...
+            testFileVnode.get(),
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            nullptr));
+    MockCalls::Clear();
+
```
jeschu1: nit: I think removing the blank line will help clarity

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -166,6 +169,105 @@ - (void) testEmptyFileHydrates {
     }
 }
 
+- (void) testEmptyFileDeleteNonHydrationOptimisation {
```
jeschu1: nit: testFileDeleteHydratesOnlyWhenNecessary ?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: I'd really love a unit test here.  One that explicity swaps a vnode not at the end.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: We cannot move this because `DeleteOpIsForRename` clears the entry from the array, so if we didn’t call it the entries would start piling up.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: This gives me another idea.

My main concern with our current approach is that something may go wrong and we suddently have a very large data object for holding renames.  This has a cost for sure.

Thoughts on making the data size making the data size X amount big at the start (maybe 2,000 ).  If we go over that roof we log an error and revert to always hydrating like we would on High Sierra.  I think your memory allocation outside of the lock is sound, but having a top roof would simpler and avoid my fear of a large object size.

I totally realize I may be being too cautious here.  Thoughts?  @wilbaker @pmj

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: There’s a version check before calling this function. The first iteration of this PR was also backwards compatible but had a second commit which removed that compatibility. That broke the functional tests however, and I’ve not pushed it on the rebased version for now.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Is the concern `ShouldHandleVnodeOpEvent` may fail for another reason and we'd miss calling `DeleteOpIsForRename`

If having the check here is a must I'd suggest doing an explicit `VnodeIsEligibleForEventHandling` check for all nodes and removing it from `ShouldHandleVnodeOpEvent`.  I think that may be easier to understand.  

Some comments would be great too :-)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Since this only gets called from` KAUTH_FILEOP_WILL_RENAME` you should be on 18+ if you get here.
Maybe an extra comment: `// This function should be called from KAUTH_FILEOP_WILL_RENAME`

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: > Is the concern `ShouldHandleVnodeOpEvent` may fail for another reason and we'd miss calling `DeleteOpIsForRename`

Precisely.

> If having the check here is a must I'd suggest doing an explicit `VnodeIsEligibleForEventHandling` check for all nodes and removing it from `ShouldHandleVnodeOpEvent`. I think that may be easier to understand.

That is absolutely a change I was planning to make, but as part of #900. As it's fairly standalone I can apply it here though.

> Some comments would be great too :-)

OK, will do.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I assumed that was implicit in the assert message, but I can make that a bit clearer.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: Yes, can do. Really, we should be doing this all in a generic resizeable array data structure, so we can test that in isolation and don't have to go through all the contortions of exposing the guts of our specific data structure to the test framework.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I think that level of caution (1) is excessive and also (2) doesn't match the level we've applied elsewhere.

1. The pathological case you're worried about is that of every single thread in the entire operating system renaming a file at the exact same time, and all of them being preempted between the `KAUTH_FILEOP_WILL_RENAME` and `KAUTH_VNODE_DELETE` kauth callbacks. The maximum number of processes in the system is given by the `kern.maxproc` sysctl, which seems to scale with RAM - I just checked the value on 2 systems, and it's 4256 on a Mac with 32GiB, and 2128 on one with 16GiB. Each process may spawn up to `kern.num_taskthreads` threads, which also seems to scale with RAM, with limits of 8192 and 4096 for those two Macs. So on that beefier Mac, we have an absolute upper limit of `4256 * 8192 = 34865152` array entries in use. As we're growing the array by powers of 2 and each element is 16 bytes, that means the array would be `64 * 1024 * 1024 * 16 = 1GiB`. While very large, it won't take down the system. Plus, I don't think even a very determined attacker would be able to deliberately cause that worst-case preemption scenario without gaining kernel access themselves - and once they have kernel access, they don't need to exploit this kext anymore anyway. We certainly won't get anywhere near those values by accident. My fairly busy, long-running Mac currently has about 4500 threads according to Activity Monitor, so if all of them ended up in that pathological rename-preemption case, the array would only be 8192 items or 128KiB long, which is pretty unremarkable these days.
2. By far the biggest risk our kext poses to overall system stability is the fact that our kauth handlers block pending a response from user space providers. The adverse interaction this has had with AV software (#354) in the past is one example of that. (I suspect the occasional, undiagnosed reports we get of slowdowns/hangs are related to this too.)

Finally, I don't particularly like "high water mark" solutions as that code tends to be very badly tested in the real world, even if we have unit tests for it. So it's not uncommon that the very code that's supposed to prevent Bad Things when resource utilisation goes beyond a certain level actually causes other Bad Things because it only runs on one in a million cases and those bugs are usually not even reported.

jeschu1: @pmj those are great points

I think we need to pick a number to, at a minimum, log that we got that high.  If we don't expect to hit say 2,000 (or X) we should log that.  If we log as a error it will get swept into telemetry and we can examine later.  To be clear I doubt a user would ever hit a high number and the code looks correct to me.  

In the unlikely event that this does happen it would be a signal that either 1) something is wrong or 2) we should consider a different data structure for holding rename events.

This is the only point of contention left for me on the PR.  Everything else looks good :-)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: > That is absolutely a change I was planning to make, but as part of #900. As it's fairly standalone I can apply it here though.

I think it would be great to apply here if you don't mind.  It makes me it easier for me at least to read :-)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Thanks @pmj.  I think a comment saying
// This function i called from `KAUTH_FILEOP_WILL_RENAME` a 18+ feature would be helpful

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yeah, we can definitely log a high value. In principle we can also shrink the array back down, but I doubt we'll even fill the default size of 8, so let's log if we grow beyond 16, say?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've updated the assert message to be more descriptive, no point splitting the information between the assert and a comment.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've now done that, as a separate commit in the series as it required an additional test to maintain 100% coverage and some updates to perf sampling which are completely unrelated to #526.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've now added [`testPendingRenamesOutOfOrderInsertAndRemoval`](https://github.com/microsoft/VFSForGit/pull/1193/commits/753ff3236def50cc95bbfec98480b1f50f2c35d5#diff-3da2dbea50d618b6315c407d65c25658R677) which covers this situation.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: The error is now [logged in `RecordPendingRenameOperation`](https://github.com/microsoft/VFSForGit/pull/1193/commits/753ff3236def50cc95bbfec98480b1f50f2c35d5#diff-ab3be22ed1fbe1779471dbe568982eb0R355) and this behaviour is tested in [`testResizePendingRenamesLogsLargeArrayError`](https://github.com/microsoft/VFSForGit/pull/1193/commits/753ff3236def50cc95bbfec98480b1f50f2c35d5#diff-3da2dbea50d618b6315c407d65c25658R657).

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've riffed on the wording, [but we now have a comment in this location](https://github.com/microsoft/VFSForGit/pull/1193/commits/753ff3236def50cc95bbfec98480b1f50f2c35d5#diff-ab3be22ed1fbe1779471dbe568982eb0R962).

pmj:
I've pushed a major update to this, incorporating all the feedback so far. Note that there are now 2 commits, the first a tiny step towards resolving #900, and the second actually fixing #526. I can push a third commit, which was part of the very first pass and removed OS backwards compatibility. However, as the functional tests are all run on 10.13, this fails, so I'm not sure there's much point. I can maintain a 10.14+ branch once which I'll periodically rebase on master once this PR has been merged if you like.

pmj:
The latest push rebases on current master, resolving conflicts.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+            {
+                for (uint32_t i = 0; i < s_pendingRenameCount; ++i)
+                {
+                    assert(s_pendingRenames[i].thread != myThread);
+                }
+
+                resizeTable = true;
+                
+                resizeTableLength = MAX(1u, MIN(UINT32_MAX, s_maxPendingRenames * UINT64_C(2)));
+                assert(resizeTableLength > s_maxPendingRenames);
```
jeschu1: To confirm, whats the largest this is allowed to grow to?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: I'm really happy with the logging there.

My thought process here is, we fall back to the old behavior for high-sierra and will never have this optimization there.  So if we fall back on later versions once we hit an unexpected size, not the end of the world.  It would also let us remove a good chunk of code.  Anyways food for thought!

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -344,4 +346,55 @@ - (void)testUseMainForkIfNamedStream {
     vnode_put(testVnode);
 }
 
+- (void)testResizePendingRenamesRace
+{
+    InitPendingRenames();
+    XCTAssertLessThan(s_maxPendingRenames, 16);
```
jeschu1: Very small nit:  If we know what the value should be lets assertEqual

jeschu1:
This PR is looking very good to me!  I think backwards compat is the right answer here and it isn't too much trouble.  I'm still not sure on growing / fixed length for the array, but thats a small implemenation detail in the overall PR.

@wilbaker can you give this a pass too

wilbaker:
> @wilbaker can you give this a pass too

Sure I'll take a look today, @pmj apologies for the delay!

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -711,6 +1031,18 @@ - (void) testDeleteDirectoryWhenFirstRequestFails {
 
 - (void) testDeleteDirectoryWhenSecondRequestFails {
```
wilbaker: Naming suggestion: `testDeleteDirectoryWhenSecondRequestFails` -> `testRenameDirectoryWhenSecondRequestFails` or `testDeleteDirectoryForRenameWhenSecondRequestFails`

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -853,5 +1185,21 @@ - (void) testWriteDataToForkedPath {
     XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 0);
 }
 
+- (void) testIneligibleFilesystemType
```
wilbaker: Were we just missing coverage for this case before?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -853,5 +1185,21 @@ - (void) testWriteDataToForkedPath {
...
+{
+    shared_ptr<mount> testMountNone = mount::Create("msdos", fsid_t{}, 0);
+    shared_ptr<vnode> testVnodeNone = vnode::Create(testMountNone, "/Volumes/USBSTICK");
+    XCTAssertEqual(
+        KAUTH_RESULT_DEFER,
+        HandleVnodeOperation(
```
wilbaker: Should we also try calling `HandleFileOpOperation` for WILL_RENAME and validate that `s_pendingRenameCount` does not increase?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: I guess as we're directly testing the data structure itself here we can do that. Otherwise I'd say we don't really care about the exact growth algorithm.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: `MAX(1u, MIN(UINT32_MAX, s_maxPendingRenames * UINT64_C(2)))` should clamp to 1..UINT32_MAX inclusive. I should probably wrap that in a `clamp()` template function for better readability. Not 100% sure on the most intuitive parameter order for that - `clamp(min, value, max)` or `clamp(value, min, max)`?

jrbriggs: For consistency with the standard library, the latter: `clamp(val, min, max)`

ProjFS.Mac/PrjFSKextTests/TestLocks.cpp
```diff
@@ -39,3 +39,25 @@ void RWLock_FreeMemory(RWLock* rwLock)
...
     rwLock->p = nullptr;
 }
+
+void SpinLock_Acquire(SpinLock lock)
+{
+}
+
+void SpinLock_Release(SpinLock lock)
+{
+}
```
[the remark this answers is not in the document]
pmj: We currently don't have any such validation in the `RWLock` mocks either. I'd definitely like to implement that at some stage, ideally with some APIs for explicitly simulating race conditions. Feel free to create an issue for this and assign it to me, but I think the scope of this PR is already large enough.

ProjFS.Mac/PrjFSKext/Locks.cpp
```diff
@@ -180,3 +180,30 @@ void RWLock_DropExclusiveToShared(RWLock& rwLock)
 #endif
 }
 
+
+SpinLock SpinLock_Alloc()
+{
+    return SpinLock { lck_spin_alloc_init(s_lockGroup, LCK_ATTR_NULL) };
+}
+
+void SpinLock_FreeMemory(SpinLock* lock)
```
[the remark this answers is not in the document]
pmj: I was following the precedent set by

```
Mutex Mutex_Alloc();
void Mutex_FreeMemory(Mutex* mutex);
```

and

```
RWLock RWLock_Alloc();
void RWLock_FreeMemory(RWLock* rwLock);
```

One advantage of the pointer (or reference) based API is that the internal pointer reliably gets zeroed out, which is what the `*_IsValid()` functions test for.

For ever so slightly more ergonomic use we could replace the pointer with a reference (`SpinLock& lock`), but I think the safety aspect of zeroing out the internal pointer is sound. If we were going to change our lock wrappers fundamentally, my suggestion would be to make them properly encapsulated C++ classes with safe constructors and destructors.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
     }
 }
 
+KEXT_STATIC bool InitPendingRenames()
+{
+    if (version_major >= 18) // Only need to track renames on Mojave and newer
```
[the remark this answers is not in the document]
pmj: Good idea. I'll push that in the next iteration.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+            Memory_FreeArray(s_pendingRenames, s_maxPendingRenames);
+            s_pendingRenames = nullptr;
+            s_maxPendingRenames = 0;
+        }
+    }
+}
+
+KEXT_STATIC void ResizePendingRenames(uint32_t newMaxPendingRenames)
+{
+    PendingRenameOperation* newArray = Memory_AllocArray<PendingRenameOperation>(newMaxPendingRenames);
```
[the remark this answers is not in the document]
pmj: I don't see the need for using a `_noblock` fast path here because this code is going to be called so rarely anyway. And as we call this only when no lock is held, the blocking version is safe.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+        }
+        else
+        {
+            arrayToFree = newArray;
+            arrayToFreeLength = newMaxPendingRenames;
+        }
+    }
+    SpinLock_Release(s_renameLock);
+    
+    if (arrayToFree != nullptr)
```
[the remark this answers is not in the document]
pmj: If `s_maxPendingRenames` is initially `nullptr´. This won't happen in the production kext, but allowing it simplifies unit tests.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +241,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+            {
+                isRename = true;
+                assert(s_pendingRenames[i].vnode == vnode);
+                --s_pendingRenameCount;
+                if (i != s_pendingRenameCount)
+                {
+                    s_pendingRenames[i] = s_pendingRenames[s_pendingRenameCount];
+                }
+                
+                s_pendingRenames[s_pendingRenameCount] = PendingRenameOperation{};
```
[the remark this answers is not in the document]
pmj: No, `current_thread()` never returns `nullptr`. Additionally, because we only check the first `s_pendingRenameCount` array elements, we will never test against a default-initialised entry.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -681,6 +992,15 @@ - (void) testDeleteDirectoryWithDisappearingVirtualizationRoot {
             ActiveProvider_Disconnect(self->dummyRepoHandle, &self->dummyClient);
         });
     testDirVnode->attrValues.va_flags = FileFlags_IsInVirtualizationRoot;
```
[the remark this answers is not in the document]
pmj: > Did the `KAUTH_FILEOP_WILL_RENAME` file op have to be added for this test to work?

Looks like it did not - I'm not sure why I added it. I'll remove it again.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: Commit c8b4573 moves the eligibility check outside the `ShouldHandle…` function, so there's a new code path in `HandleVnodeOperation`, and this path isn't covered by existing tests, no.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: Good idea; I'll add that to `HandleFileOpOperationTests(.mm)` though as it's not directly relevant to this test.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Thanks for bringing that to my attention, @jrbriggs - I'd missed that the standard library had gained such a function in C++17. Not that we can use it in kext code anyway, so I'll add our own version.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -364,8 +478,166 @@ - (void) testDeleteFile {
...
+        reinterpret_cast<uintptr_t>(context),
+        reinterpret_cast<uintptr_t>(otherTestFileVnode.get()),
+        0,
+        0) == KAUTH_RESULT_DEFER);
+
+    XCTAssertTrue(MockCalls::DidCallFunctionsInOrder(
+        ProviderMessaging_TrySendRequestAndWaitForResponse,
+        make_tuple(
+            _,
+            MessageType_KtoU_NotifyFilePreDelete,
```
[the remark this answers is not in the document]
pmj: Yes, I didn't have time to update this PR after rebasing on post-#1215 last night, I've fixed it in the next iteration.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: > My thought process here is, we fall back to the old behavior for high-sierra and will never have this optimization there.

We *are* scheduled to remove the High Sierra code path in a few weeks' time.

> It would also let us remove a good chunk of code.

I guess I'm not convinced we should be running scared every time we need to resize a data structure.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -273,8 +471,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_SEARCH |
                 KAUTH_VNODE_READ_SECURITY |
                 KAUTH_VNODE_READ_ATTRIBUTES |
-                KAUTH_VNODE_READ_EXTATTRIBUTES |
-                KAUTH_VNODE_DELETE))
+                KAUTH_VNODE_READ_EXTATTRIBUTES) ||
+            (isDeleteAction && isRename))
```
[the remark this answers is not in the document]
pmj: Ah indeed, good idea. I think I defaulted it to `true` originally because the version check was somewhere else than it now is, but with the current code, a default of `false` does indeed make it simpler.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -273,8 +471,8 @@ KEXT_STATIC int HandleVnodeOperation(
...
         {
             // Recursively expand directory on delete to ensure child placeholders are created before rename operations
             if (isDeleteAction)
```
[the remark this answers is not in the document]
pmj: I've just been struck by a thought: have we even been handling this case correctly? (mostly unrelated to this change) If we move a directory out of a repo, and it's recursively enumerated, won't it just end up full of empty files? I didn't think recursive enumeration hydrated files.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -340,8 +538,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 KAUTH_VNODE_READ_DATA |
                 KAUTH_VNODE_WRITE_DATA |
                 KAUTH_VNODE_EXECUTE |
-                KAUTH_VNODE_DELETE | // Hydrate on delete to ensure files are hydrated before rename operations
-                KAUTH_VNODE_APPEND_DATA))
+                KAUTH_VNODE_APPEND_DATA) ||
+            (isDeleteAction && isRename)) // Hydrate on delete to ensure files are hydrated before rename operations
```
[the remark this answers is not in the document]
pmj: The comment then becomes somewhat tautological, so I've updated it to be about the ultimate reason why we're doing this, user expectation.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -166,6 +169,105 @@ - (void) testEmptyFileHydrates {
...
+            _,
+            _,
+            _,
+            nullptr));
+    MockCalls::Clear();
+    CleanupPendingRenames();
+    version_major = 18;
+}
+
+- (void) testRenamingEmptyFileHydrates
```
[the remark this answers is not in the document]
pmj: Good point, I've removed `testRenamingEmptyFileHydrates` as `testHydrationOnDeleteWhenRenamingFile` has a counterpart in `testDeleteFileNonRenamed`.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -403,6 +675,45 @@ - (void) testDeleteDir {
     XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 2);
 }
 
+- (void) testDeleteDirNonRenamed
```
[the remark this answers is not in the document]
pmj: To make this type of test easier and less repetitive to write, I've added helper functions `TestForDarwinVersionRange()` and `TestForAllSupportedDarwinVersions()` and used the latter in this test and updated `testFileDeleteHydratesOnlyWhenNecessary` with it too.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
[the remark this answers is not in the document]
pmj: Ugh, looks like the rebase might have moved the change to a different test. The coverage provided by `testDeleteDirectoryWithDisappearingVirtualizationRoot`, so I've re-added it there.

ProjFS.Mac/PrjFSKext/Locks.cpp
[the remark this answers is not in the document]
wilbaker: > One advantage of the pointer (or reference) based API is that the internal pointer reliably gets zeroed out, which is what the …_IsValid() functions test for.

Ahh good point, I missed that thanks!

ProjFS.Mac/PrjFSKextTests/TestLocks.cpp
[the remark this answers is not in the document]
wilbaker: Sounds good, I agree no need to make the changes in this PR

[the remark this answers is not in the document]
wilbaker: I filed #1254 for this, but I don't think it's anything we need to pick up soon

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
wilbaker: > If we move a directory out of a repo, and it's recursively enumerated, won't it just end up full of empty files? I didn't think recursive enumeration hydrated files.

Yep this has been broken and currently you can only rename directories within a repo.  We had #515 open to fix it, but then we re-purposed #515 to say that we should block renames of projected folders.

Rather than fixing the scenario you've called out by implementing recursive hydration the plan is to set the minimum version to Mojave as soon as we can and then block renames of projected folders.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: > I guess I'm not convinced we should be running scared every time we need to resize a data structure.

No disagreement there, there are definitely times to take on complexity.  To me this is similar to Vnode cache where we decided to fall back to a slower path rather than worry about growing memory.  That said the code looks correct, and we have good alerts in place so if we need to adjust in the future we can.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -216,6 +247,162 @@ KEXT_STATIC void UseMainForkIfNamedStream(
...
+                assert(resizeTableLength > s_maxPendingRenames);
+            }
+        }
+        SpinLock_Release(s_renameLock);
+        
+        if (resizeTable)
+        {
+            if (resizeTableLength > 16)
+            {
+                KextLog_Error("Warning: RecordPendingRenameOperation is causing pending rename array resize to %u items.", resizeTableLength);
```
jeschu1: If we have a process doing a large amount of renames could this hammer telemetry?  I'm wondering if ((resizeTableLength % 16) == 0) may be more appropriate?

pmj: > If we have a process doing a large amount of renames could this hammer telemetry?

No, we only log on *resize*, and there's a `O(log2(N))` number of those for some maximum number of pending resizes. So even for say 1000000 pending resizes we'd get about 16 log messages.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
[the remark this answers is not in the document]
pmj: Ahh, right, that was the reason for wanting to block them. I can integrate that blocking in my work on #182 for Mojave+ only.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: Good point, this looks good then!

pmj:
@jeschu1 mentioned that there would be a conflict between this change and #1215 regarding metadata removal, but it's evidently not triggering any functional test failures. After trying to think it through, I have to admit I still can't quite work out what the problem is, and I've run the functional tests locally on Mojave, and they pass, so they're no help in pinpointing the problem. (the azp pipeline seems to have a codesigning issue)

I've archived the commit-by-commit review feedback fixes in this branch: https://github.com/pmj/VFSForGit/tree/526-only-hydrate-on-rename-delete-archived and I'm about to force-push a new version of this PR branch that's squashed and rebased on latest master. If the metadata issue turns out to be a non-issue, I'm happy for it to be merged as-is, otherwise @jeschu1 could you please implement that fix?

pmj:
Sorry for the late reply @jeschu1 - I did take a quick look at your commits and they all look good to me. Thanks for sorting that out!

PR #1194, opened 2019-05-23
Mac: Refactor testShouldHandleFileOpEvent test:

wilbaker:
Resolves #1187

  - Split testShouldHandleFileOpEvent into separate test cases with more descriptive names
  - Fix an issue where the `testShouldHandleFileOpEvent_VnodeCacheUpdated` tests were sometimes failing due to vnode cache collisions from other tests
  - Updated the provider initiated I/O and provider offline scenarios to use file vnodes rather than the repo root vnodes

PR #1199, opened 2019-05-24
Move ShouldHandleFileOpEvent unit tests into their own class

### LINKED ISSUE #1195: Move ShouldHandleFileOpEvent tests into their own class
Update the kext unit tests based on the comments left by @pmj in #1194 

In addition to moving `ShouldHandleFileOpEvent` tests, also:

> I realise the old code did this too, but the #defined name, PrjFSProviderUserClient is generally preferable in code - that's the whole point of the macro, if the reverse DNS prefix changes, you don't have to update all of the code.
>
>Might be worth doing a search for the use of the explicitly-prefixed name, I notice quite a few instances of it being used in this file.

wilbaker:
Additionally:
- Reorder PrjFSKextTests files so they're in alphabetical order
- Replace the usage of org_vfsforgit_PrjFSProviderUserClient with PrjFSProviderUserClient

Follow up work for comments left on #1194 after it was merged.

Resolves #1195


ProjFS.Mac/PrjFSKextTests/ShouldHandleFileOpTests.mm
```diff
@@ -0,0 +1,298 @@
...
+#include "MockProc.hpp"
+#include "MockVnodeAndMount.hpp"
+#include "VnodeCacheEntriesWrapper.hpp"
+
+#import <XCTest/XCTest.h>
+
+using std::shared_ptr;
+
+// Dummy implementation
+class org_vfsforgit_PrjFSProviderUserClient
```
pmj: This is still using the reverse-DNS prefixed name.

ProjFS.Mac/PrjFSKextTests/ShouldHandleFileOpTests.mm
```diff
@@ -0,0 +1,298 @@
...
+            &self->perfTracer,
+            self->context,
+            self->repoRootVnode.get(),
+            nullptr, // path
+            KAUTH_FILEOP_RENAME,
+            true, // isDirectory,
+            &testRootHandle,
+            &pid));
+
+    PrjFSProviderUserClient userClient;
```
pmj: There's probably no harm moving this to an ivar - various other tests also use one of these.

ProjFS.Mac/PrjFSKextTests/ShouldHandleFileOpTests.mm
```diff
@@ -0,0 +1,298 @@
...
+        ShouldHandleFileOpEvent(
+            &self->perfTracer,
+            self->context,
+            self->testVnodeFile.get(),
+            nullptr, // path
+            KAUTH_FILEOP_OPEN,
+            false, // isDirectory,
+            &rootHandle,
+            &pid));
+    XCTAssertTrue(rootHandle == testRootHandle);
```
pmj: While we're poking around in here, would it be worth changing these assertions to `XCTAssertEqual`?

ProjFS.Mac/PrjFSKextTests/ShouldHandleFileOpTests.mm
```diff
@@ -0,0 +1,298 @@
...
+        0,
+        self->repoRootVnode.get(),
+        self->repoRootVnode->GetVid(),
+        FsidInode{ self->repoRootVnode->GetMountPoint()->GetFsid(), self->repoRootVnode->GetInode() },
+        self->repoPath.c_str());
+    XCTAssertTrue(VirtualizationRoot_IsValidRootHandle(testRootHandle));
+
+    int pid;
+    // Fail when the provider is not online
+    s_virtualizationRoots[0].providerUserClient = nullptr;
```
pmj: This should really be an `ActiveProvider_Disconnect()` call. (In which case there needs to be a `vnode_get(self->repoRootVnode.get())` after the insertion above to balance the `vnode_put` in Disconnect; this is a bit ugly and I believe I fixed the insertion to perform the `_get` in #500. I think I'll extract the simpler bug fixes from that PR and resubmit them as a standalone one.)

wilbaker: I see that you made this change in #1200.  

I will leave this as-is here and wait to merge this PR in until after you've merged #1200.

wilbaker:
Waiting on #1200 to be merged before merging this PR.

ProjFS.Mac/PrjFSKextTests/ShouldHandleFileOpTests.mm
wilbaker: Actually I see that this is a slightly different suggestion, fixing now

PR #1200, opened 2019-05-24
Mac kext testing: assert() integration

pmj:
This patch series enables `assert()` and `assertf()` in the "testable" kext build. This plus the supporting infrastructure are implemented in the first commit:

 * Enables assertions in the testable kext build by `#define`ing `MACH_ASSERT=1`.
 * Implements the back-end Assert() and panic() functions used by the assert() and assertf() macros, respectively. These back-end functions call into XCTest's test failure reporting mechanism, so failed kext assertions in unit tests are now reported as regular unit test failures.
 * To facilitate this, the current test case needs to be known outside the test's Objective-C class, so we track that in thread-local storage by subclassing XCTestCase and deriving all tests from the new class, and perform current test registration/deregistration during setUp/tearDown.
 * Some unit tests deliberately cause failed assertions, so a mechanism for ignoring a specific number of them in a test is provided.
 * Not having full coverage in panic/Assert/etc. is a good thing, so those have been added to the 100% test coverage exceptions

The remaining commits fix unit test breakage that is caused by this - mostly cases of `VirtualizationRoots_Cleanup()` calls in a state where mock providers haven't yet disconnected. There's also a test which deliberately causes a failed assertion, which now needs to explicitly declare this fact.

**Reviewer notes:**

 * Commit-by-commit review probably will make a lot more sense than viewing all changes in one go.
 * The lack of support for an `assert()` with teeth in the testable kext has been annoying me for a while, so although it's not explicitly listed in #900, I've gone ahead and done it now as I ran into a situation where it would have helped while working on #526/#1193.

### DIFF
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -85,7 +85,6 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     ProvidermessageMock_ResetResultCount();
     ProviderMessageMock_SetDefaultRequestResult(true);
     ProviderMessageMock_SetSecondRequestResult(true);
-    ProviderMessageMock_SetCleanupRootsAfterRequest(false);
 }
 
 - (void) tearDown
@@ -659,7 +658,11 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
 }
 
 - (void) testDeleteDirectoryWithDisappearingVirtualizationRoot {
-    ProviderMessageMock_SetCleanupRootsAfterRequest(true);
+    ProviderMessageMock_SetRequestSideEffect(
+        [&]()
+        {
+            ActiveProvider_Disconnect(self->dummyRepoHandle, &self->dummyClient);
+        });
     testDirVnode->attrValues.va_flags = FileFlags_IsInVirtualizationRoot;
     XCTAssertTrue(HandleVnodeOperation(
         nullptr,
@@ -750,7 +753,11 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
 }
 
 - (void) testWriteWithDisappearingVirtualizationRoot {
-    ProviderMessageMock_SetCleanupRootsAfterRequest(true);
+    ProviderMessageMock_SetRequestSideEffect(
+        [&]()
+        {
+            ActiveProvider_Disconnect(self->dummyRepoHandle, &self->dummyClient);
+        });
     testFileVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
     XCTAssertTrue(HandleVnodeOperation(
         nullptr,
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -9,7 +9,7 @@
 #include "../PrjFSKext/ProviderMessaging.hpp"
 #include "../PrjFSKext/public/PrjFSXattrs.h"
 #include "../PrjFSKext/kernel-header-wrappers/kauth.h"
-#import <XCTest/XCTest.h>
+#import "KextAssertIntegration.h"
 #import <sys/stat.h>
 #include "KextMockUtilities.hpp"
 #include "MockVnodeAndMount.hpp"
@@ -35,7 +35,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     vnode->xattrs.insert(make_pair(PrjFSFileXAttrName, rootXattrData));
 }
 
-@interface HandleVnodeOperationTests : XCTestCase
+@interface HandleVnodeOperationTests : PFSKextTestCase
 @end
 
 @implementation HandleVnodeOperationTests
@@ -55,6 +55,8 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
 
 - (void) setUp
 {
+    [super setUp];
+    
     kern_return_t initResult = VirtualizationRoots_Init();
     XCTAssertEqual(initResult, KERN_SUCCESS);
     context = vfs_context_create(NULL);
@@ -99,6 +101,8 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     MockVnodes_CheckAndClear();
     MockCalls::Clear();
     MockProcess_Reset();
+
+    [super tearDown];
 }
 
 - (void) testEmptyFileHydrates {
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -44,6 +44,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     const char* repoPath;
     const char* filePath;
     const char* dirPath;
+    VirtualizationRootHandle dummyRepoHandle;
     PrjFSProviderUserClient dummyClient;
     pid_t dummyClientPid;
     shared_ptr<mount> testMount;
@@ -76,7 +77,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
     // Register provider for the repository path (Simulate a mount)
     VirtualizationRootResult result = VirtualizationRoot_RegisterProviderForPath(&dummyClient, dummyClientPid, repoPath);
     XCTAssertEqual(result.error, 0);
-    vnode_put(s_virtualizationRoots[result.root].rootVNode);
+    self->dummyRepoHandle = result.root;
 
     MockProcess_AddContext(context, 501 /*pid*/);
     MockProcess_SetSelfPid(501);
@@ -89,6 +90,11 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
 
 - (void) tearDown
 {
+    if (VirtualizationRoot_GetActiveProvider(self->dummyRepoHandle).isOnline)
+    {
+        ActiveProvider_Disconnect(self->dummyRepoHandle, &self->dummyClient);
+    }
+
     testMount.reset();
     repoRootVnode.reset();
     testFileVnode.reset();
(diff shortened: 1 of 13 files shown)


pmj:
Latest push merely rebases on master.

ProjFS.Mac/PrjFSKextTests/VnodeCacheTests.mm
```diff
@@ -594,6 +598,7 @@ - (void)testFindVnodeRootFromDiskAndUpdateCache_InvalidUpdateCacheBehaviorValue
     // In production FindVnodeRootFromDiskAndUpdateCache would panic when it sees an
     // invalid UpdateCacheBehavior, but in user-mode the assertf if a no-op
     VirtualizationRootHandle rootHandle;
+    [self setExpectedFailedKextAssertionCount:1];
```
jeschu1: So this means we will hit exactly one assert in the called test code below?

To confirm asserts in test cases don't cause failures?

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return;
+    }
+
+    NSString* expressionString = [NSString stringWithCString:expression encoding:NSUTF8StringEncoding];
+    _XCTFailureHandler(test, YES, file, line, expressionString, @"");
+}
+
+// catches kext assertf() - currently no source file & line number support
+void panic(const char* format, ...)
```
jeschu1: Do we have tests that call panics?

ProjFS.Mac/PrjFSKextTests/VnodeCacheTests.mm
pmj: Failed assertions in kext code when run in a test *will* now cause test failures, unless explicitly declared as expected as above.

If in the course of running this particular test case, 0 assertions failed in kext code, the test will fail. Likewise if 2 assertions fail.

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
pmj: The test `testFindVnodeRootFromDiskAndUpdateCache_InvalidUpdateCacheBehaviorValue` where you commented on the `[self setExpectedFailedKextAssertionCount:1];` call actually fails an `assertf()` in the kext, and this causes `panic()` to be called. (See `kern/assert.h` for the definitions of `assert()` and `assertf()` macros in the kernel.)

jeschu1:
This also contains some nice cleanups for unit tests.  
While i am not an objective C expert i can definitely work with this.  

@wilbaker for his thoughts too.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -99,6 +106,19 @@ - (void) tearDown
     MockVnodes_CheckAndClear();
     MockCalls::Clear();
     MockProcess_Reset();
+
+    [super tearDown];
+}
+
+- (void) removeAllVirtualizationRoots
```
wilbaker: Outside of coding style, is there any reason this helper function is written in Objective C rather than C++?  (In other words, is this required for asserts to work as expected?)

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -389,6 +391,12 @@ - (void)testShouldHandleFileOpEvent_RootHandle {
             true, // isDirectory,
             &testRootHandle,
             &pid));
+
+    if (VirtualizationRoot_IsValidRootHandle(testRootHandle))
```
wilbaker: Why do we need to check `VirtualizationRoot_IsValidRootHandle`?

Couldn't  we always know at this point whether `testRootHandle` will be valid?  Or is this in case something above in the test fails and we don't want to poison things for the rest of the test?

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
```diff
@@ -0,0 +1,90 @@
+#import "KextAssertIntegration.h"
+#include <pthread.h>
+
+extern void Assert(const char* file, unsigned line, const char* expression);
+
+static pthread_key_t s_currentTestCaseKey;
+
+static PFSKextTestCase* CurrentTestCase()
+{
+    return (__bridge PFSKextTestCase*)pthread_getspecific(s_currentTestCaseKey);
```
wilbaker: What does `__bridge` do?

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
```diff
@@ -0,0 +1,90 @@
...
+    }
+    else
+    {
+        return false;
+    }
+}
+
+@end
+
+// catches kext assert()
```
wilbaker: Does `Assert` get wired up to catch `assert()`?

ProjFS.Mac/PrjFSKextTests/VnodeCacheTests.mm
pmj: No, failed `assert()`s previously and now cause a process termination. (via `abort()` I think) Not ideal, and something I'll fix up at some point - probably by defining our own assertion macro that can be used outside the test class. (You can't use the `XCTAssert*` macros outside the test itself because their expansion references `self`.)

ProjFS.Mac/PrjFSKextTests/MemoryTests.mm
```diff
@@ -1,7 +1,7 @@
-#import <XCTest/XCTest.h>
+#import "KextAssertIntegration.h"
 #include "../PrjFSKext/Memory.hpp"
 
-@interface MemoryTests : XCTestCase
+@interface MemoryTests : PFSKextTestCase
```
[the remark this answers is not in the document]
pmj: No, although we need to remember to insert the call to `[super setUp]` if we do end up needing to override `setUp` in these classes.

Objective-C methods behave like C++ `virtual` methods, so if the subclass doesn't define its own overriding implementation, the `super` implementation is used.

ProjFS.Mac/PrjFSKextTests/ProviderMessagingMock.cpp
```diff
@@ -9,10 +9,14 @@
...
 #include "ProviderMessagingMock.hpp"
+#include <functional>
+
+using std::function;
+using std::move;
 
 bool static s_defaultRequestResult = true;
 bool static s_secondRequestResult = true;
-bool static s_cleanupRootsAfterRequest = false;
+static function<void()> s_trySendRequestSideEffect;
```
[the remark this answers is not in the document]
pmj: I haven't because `std::function` is a class and provides a default constructor which is appropriate to be used in this case. We could explicitly invoke it, but that syntax is rather wordy:

```
static function<void()> s_trySendRequestSideEffect = function<void()>();
```

I'm not convinced that's an improvement.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: It references ivars in the test class (I've made this explicit via `self->`, which is technically not required) so we'd need to pass pointers/references to those ivars as arguments if we implemented this as a free function. I didn't see an advantage in doing that. The decision has nothing to do with asserts though.

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
pmj: > Or is this in case something above in the test fails and we don't want to poison things for the rest of the test?

Exactly this.

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
pmj: It's used for casting between pointers to Objective-C objects and regular pointers. (typically `void*` as here, or Core Foundation types for "toll-free bridged" classes, such as `NSString` <-> `CFStringRef`.)

The reason is that Objective-C pointers use ARC, automatic reference counting (unless this is disabled via a compiler flag). This means that `PFSKextTestCase*` is effectively a "smart" pointer comparable to C++'s `std::shared_ptr<>`. When casting to an opaque type such as `void*`, it needs to be made clear to the compiler whether ownership of the object is transferred, whether an additional reference should be taken, or whether the opaque pointer is effectively a weak pointer (as it is here).

A more detailed explanation of ARC and bridged casts [can be found in the llvm/clang documentation](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#bridged-casts).

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
pmj: [Yes, in the kernel, `assert()` is implemented as a macro that expands to calling the `Assert()` function.](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/osfmk/kern/assert.h#L91) The macro is needed to make the filename, line number, and textual representation of the expression that unexpectedly evaluated to false appear in the error message, generating the actual error message and panic are implemented in `Assert()`. Implementing our own version of that function means we can do what we want with that information.

pmj: [Yes, in the kernel, `assert()` is implemented as a macro that expands to calling the `Assert()` function.](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/osfmk/kern/assert.h#L91) The macro is needed to make the filename, line number, and textual representation of the expression that unexpectedly evaluated to false appear in the error message, generating the actual error message and panic are implemented in `Assert()`. Implementing our own version of that function means we can do what we want with that information.

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
pmj: It's used for casting between pointers to Objective-C objects and regular pointers. (typically `void*` as here, or Core Foundation types for "toll-free bridged" classes, such as `NSString` <-> `CFStringRef`.)

The reason is that Objective-C pointers use ARC, automatic reference counting (unless this is disabled via a compiler flag). This means that `PFSKextTestCase*` is effectively a "smart" pointer comparable to C++'s `std::shared_ptr<>`. When casting to an opaque type such as `void*`, it needs to be made clear to the compiler whether ownership of the object is transferred, whether an additional reference should be taken, or whether the opaque pointer is effectively a weak pointer (as it is here).

A more detailed explanation of ARC and bridged casts [can be found in the llvm/clang documentation](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#bridged-casts).

pmj:
Thanks for the review everyone. Looks like it's not likely there will be any objections here so I'll merge, but feel free to keep asking questions if my answers so far are insufficient.

ProjFS.Mac/PrjFSKextTests/KextAssertIntegration.m
jeschu1: I was thinking this exact question, thanks for asking it @wilbaker!

PR #1202, opened 2019-05-24
Platform native notifications PR_3

alameenshah:
#### Enable display of Automount notifications

This is the third incremental PR to enable platform native notifications on the Mac. In this PR, `GVFS.Service` sends auto-mount notifications to `VFS For Git.app` for display to user. `Service` displays two separate notifications - 1. a generic message when `GVFS.Service` begins on auto-mounting registered repositories and 2. for each registered repository with repo specific auto-mount success/failure messaging.

#### Details
- New VFSForGitNotification class. It supports 3 different notifications - AutomountStart, MountSuccess, MountFailure. It will not display any other notification.
- VFSForGitNotification defines the title & message body for each of the supported notification. It is not possible for `GVFS.Service` or another client to display random text.
- `MessageListener` class creates a socket and waits for notification requests from client.
- MessageParser class to parse the incoming JSON formatted message text.
- New UT for VFSForGitNotification & MessageParser classes.
- `GVFS.Service` does not wait for a reply from `VFS For Git.app` for the notification requests that it sends.
- `VFS For Git.app` get automatically launched after installation.
- `VFS For Git.app` has the same app version as `gvfs version`.

#### Message processing sequence
`AppDelegate` class creates and installs an instance of `MessageListener`. `MessageListener` creates a socket and waits for incoming messages. The following sequence diagram shows the flow of a message that arrives in `MessageListener`.
UserNotification-2

#### Testing notes
- Installer automatically launches `VFS For Git.app` after successful install. If there were repositories that were registered for auto-mount before upgrade install, then auto-mount notifications for each of those repos would get displayed in the native Notification Center as `GVFS.Service` auto mounts them.
- Tried sending a few hundred notifications. NotificationCenter collated notifications beyond 10 and created the following single notification (separate ones were still available in a long scroll view inside the NotificationCenter) - 
Screen Shot 2019-06-12 at 3 00 08 PM

- Tried sending huge (500MB) garbage data from client - memory usage of `VFS For Git.app` spiked, JSON Parsing failed, the app logged parse error and its memory footprint went down to normal.
- Tried multiple clients sending messages simultaneously. No issues observed.

#### Reference
For a better understanding of the bigger picture refer [this PR](https://github.com/microsoft/VFSForGit/pull/1114).


GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/MessageListener.h
```diff
@@ -0,0 +1,15 @@
+#import <Foundation/Foundation.h>
+
+typedef void (^NewMessageCallback) (NSDictionary *_Nonnull messageInfo);
```
HiKumar-MS: Recent versions of Xcode insert the following markers in header files, to default to _Nonnull for all parameters. This class looks like would benefit from that instead of explicitly adding _Nonnull decorators everywhere.
NS_ASSUME_NONNULL_BEGIN
…your code...
NS_ASSUME_NONNULL_END

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/MessageListener.h
```diff
@@ -0,0 +1,15 @@
...
+
+@interface MessageListener : NSObject
```
HiKumar-MS: Use project specific 3-letter prefix for class names

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/MessageParser.h
```diff
@@ -0,0 +1,9 @@
+#import <Foundation/Foundation.h>
+
+@interface MessageParser : NSObject
+
++ (BOOL) tryParseMessage:(NSDictionary *_Nonnull *_Nonnull) message
```
HiKumar-MS: Output/optional parameters are usually at the end.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/MessageParser.m
```diff
@@ -0,0 +1,46 @@
...
+        return NO;
+    }
+    
+    if ([messageStr hasPrefix:(NSString *) NotificationPrefix])
+    {
+        messageStr = [messageStr substringFromIndex:[NotificationPrefix length]];
+    }
+    
+    messageStr =
+    [messageStr stringByTrimmingCharactersInSet:[NSCharacterSet controlCharacterSet]];
```
HiKumar-MS: Formatting: assignment should be on the same line

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.h
```diff
@@ -0,0 +1,22 @@
...
+
+@interface VFSForGitNotification : NSObject
+
+@property (assign, readonly) Identifier identifier;
+@property (copy, readonly, nonnull) NSString *title;
+@property (copy, readonly, nonnull) NSString *message;
+
++ (BOOL) tryCreateNotification:(VFSForGitNotification *_Nonnull *_Nonnull) notification
+                          from:(NSDictionary *_Nonnull) jsonMessage
+                         error:(NSString *_Nonnull *_Nonnull) error;
```
HiKumar-MS: Add _autoreleasing decorator to output error parameter type

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -0,0 +1,142 @@
+#import "VFSForGitNotification.h"
+
+NSString const * IdentifierKey = @"Id";
+NSString const * EnlistmentKey = @"Enlistment";
+NSString const * EnlistmentCountKey = @"EnlistmentCount";
+NSString const * TitleKey = @"Title";
+NSString const * MessageKey = @"Message";
+
+NSString const * AutomountTitle = @"GVFS AutoMount";
```
HiKumar-MS: "NSString * const foo = @"bar";
You won't need to const_cast it to NSString*.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -0,0 +1,142 @@
...
+
+@end
+
+@implementation VFSForGitNotification
+
++ (BOOL) tryCreateNotification:(VFSForGitNotification *_Nonnull *_Nonnull) notification
+                          from:(NSDictionary *_Nonnull) jsonMessage
+                         error:(NSString *_Nonnull *_Nonnull) error
+{
+    NSParameterAssert(notification);
```
HiKumar-MS: I haven't used NSParameterAssert() but it seems like this will throw an exception and trigger a termination of the app. Are you sure all the different instances you have added so far, are all crash-worthy and non-recoverable conditions? In a lot of cases, nil parameters may be ignorable, like in the case of output error parameters if the caller doesn't care about the error then it should be fine to fail without the output error.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -0,0 +1,142 @@
...
+                                         error:(NSString *_Nonnull *_Nonnull) error
+{
+    return self =
+    [self initAsMountWithMessage:messageDict
+                           title:(NSString *) AutomountTitle
+                   messageFormat:(NSString *) AutomountFailureMessageFormat
+                           error:error];
+}
+
+- (instancetype) initAsMountWithMessage:(NSDictionary *_Nonnull) messageDict
```
HiKumar-MS: Formatting: in ObjC usually we don't put a space between parameter type and parameter name:
initAsMountWithMessage:(NSDictionary *_Nonnull)messageDict

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -0,0 +1,142 @@
...
+                           title:(NSString *) AutomountTitle
+                   messageFormat:(NSString *) AutomountSuccessMessageFormat
+                           error:error];
+}
+
+- (instancetype) initAsMountFailureWithMessage:(NSDictionary *_Nonnull) messageDict
+                                         error:(NSString *_Nonnull *_Nonnull) error
+{
+    return self =
+    [self initAsMountWithMessage:messageDict
```
HiKumar-MS: Formatting: assignment should be on same line.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -0,0 +1,142 @@
...
+                           title:(NSString *) AutomountTitle
+                   messageFormat:(NSString *) AutomountFailureMessageFormat
+                           error:error];
+}
+
+- (instancetype) initAsMountWithMessage:(NSDictionary *_Nonnull) messageDict
+                                  title:(NSString *_Nonnull) title
+                          messageFormat:(NSString *_Nonnull) messageFormat
+                                  error:(NSString *_Nonnull *_Nonnull) error
```
HiKumar-MS: Doesn't "*_Nonnull *_Nonnull" imply error!=nil && (*error)!=nil? That doesn't look right for an output parameter

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/MessageListener.h
alameenshah: Updated to use "VFS".

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: Yep... Thank you!

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: Makes sense. Have removed the hard requirement for output error args. There are 4 other places, where `NSParameterAssert` is used. These are there to catch programmer errors (like for example `VFSMessageParser::tryParseData` invoked with `nil` arg for capturing the output  `parsedMessage`). I think it would be useful to rather crash and collect the logs, if it happens.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: It does look confusing. Couple references - 
1. [Apple documentation](https://developer.apple.com/swift/blog/?id=25). See the line (although it does not exactly match our case) ```More complex pointer types like id * must be explicitly annotated. For example, to specify a non-nullable pointer to a nullable object reference, use _Nullable id * _Nonnull.```
2. Did search in Apple frameworks, could find a few sample code
```
MPSKeyedArchiver.h

+ (nullable id)unarchiveTopLevelObjectWithData:(NSData *__nonnull)data
                                         error:(NSError * __nullable * __nullable)error NS_UNAVAILABLE;
```

```
MTKModel.h

- (nullable instancetype)initWithMesh:(nonnull MDLMesh *)mesh
                               device:(nonnull id<MTLDevice>)device
                                error:(NSError * __nullable * __nullable)error;
```

alameenshah: Let me know if there is a more readable alternative...

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: Also - I've retained the ```_Nonnull``` specifier for errors (even though there is no `NSParameterAssert` for them.) It helps in showing a compiler warning when passing `nil` for errors.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.h
```diff
@@ -13,10 +15,12 @@ typedef NS_ENUM(NSInteger, Identifier)
...
 @property (assign, readonly) Identifier identifier;
-@property (copy, readonly, nonnull) NSString *title;
-@property (copy, readonly, nonnull) NSString *message;
+@property (copy, readonly) NSString *title;
+@property (copy, readonly) NSString *message;
 
-+ (BOOL) tryCreateNotification:(VFSForGitNotification *_Nonnull *_Nonnull) notification
-                          from:(NSDictionary *_Nonnull) jsonMessage
-                         error:(NSString *_Nonnull *_Nonnull) error;
++ (BOOL)tryValidateMessage:(NSDictionary *_Nonnull)jsonMessage
```
HiKumar-MS: With NS_ASSUME_NONNULL_BEGIN, now you shouldn't need _Nonnull decorators in the method parameters

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -1,90 +1,97 @@
...
         default:
         {
             *notification = nil;
-            *error = @"Unrecognised message id";
+            initError = @"Unrecognised message id";
             break;
         }
     }
     
+    if (error)
```
HiKumar-MS: Avoid using pointers as bools, compare to nil instead.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -1,90 +1,97 @@
...
 @end
 
 @implementation VFSForGitNotification
 
-+ (BOOL) tryCreateNotification:(VFSForGitNotification *_Nonnull *_Nonnull) notification
-                          from:(NSDictionary *_Nonnull) jsonMessage
-                         error:(NSString *_Nonnull *_Nonnull) error
++ (BOOL)tryValidateMessage:(NSDictionary *_Nonnull)jsonMessage
+         buildNotification:(VFSForGitNotification *_Nonnull *_Nonnull)notification
+                     error:(NSString *_Nonnull __autoreleasing *_Nonnull)error
```
HiKumar-MS: Usually an "error" parameter is of NSError type, not NSString. In objc, NSError objects are very common and preferred over throwing exceptions generally.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -1,90 +1,97 @@
...
-                                                     error:(NSString *_Nonnull *_Nonnull) error
+- (instancetype _Nullable)initAsAutomountStartWithMessage:(NSDictionary *_Nonnull)messageDict
+                                                    error:(NSString *_Nonnull *_Nonnull)error
 {
     if (self = [super init])
     {
         id repoCount = messageDict[EnlistmentCountKey];
         if (repoCount && [repoCount isKindOfClass:[NSNumber class]])
         {
             _title = [(NSString *) AutomountTitle copy];
```
HiKumar-MS: Const cast not necessary

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSMessageListener.m
```diff
@@ -0,0 +1,133 @@
+#import <netinet/in.h>
+#import <sys/socket.h>
+#import <sys/un.h>
+#import "VFSMessageListener.h"
+#import "VFSMessageParser.h"
+
+NSString const * NotificationServerPipeName = @"vfsforgit.notification";
```
HiKumar-MS: NSString * const and remove const cast below

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
HiKumar-MS: Those are all "nullable" decorations, you are using the opposite "nonnull" decoration. Both error and *error can be null/nil, so you would use __nullable decoration for that.

alameenshah: makes sense. Updated to use the default nullability for `NSError`s.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.h
alameenshah: Removed it for object references. As discussed - still need it for double pointers.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: Makes sense to keep it consistent, updated all errors to NSError type.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: Update - changed the types of all errors to NSError and also using the default nullability specifiers for NSError type.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSMessageParser.m
```diff
@@ -4,20 +4,23 @@
...
                                              encoding:NSUTF8StringEncoding]))
     {
-        if (error)
+        if (error != nil)
         {
-            *error = [NSString stringWithFormat:@"%@: error reading data",
-                      NSStringFromSelector(_cmd)];
+            NSString *info = [NSString stringWithFormat:@"%@: ERROR: error reading data.",
+                              NSStringFromSelector(_cmd)];
+            *error = [NSError errorWithDomain:@"VFS For Git"
```
HiKumar-MS: Move the domain string out to a single extern constant (declaration in some common header, single initialization in a .m file). Also see naming convention for custom error domains: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorObjectsDomains/ErrorObjectsDomains.html
_You can create your own error domains and error codes for use in your own frameworks, or even in your own applications. It is recommended that the string constant for the domain be of the form com.company.framework_or_app.ErrorDomain_.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
```diff
@@ -0,0 +1,183 @@
...
+    NSParameterAssert(notification);
+    NSParameterAssert(jsonMessage);
+    
+    id identifier = jsonMessage[IdentifierKey];
+    if (![identifier isKindOfClass:[NSNumber class]])
+    {
+        if (error != nil)
+        {
+            *error = [NSError errorWithDomain:@"VFS For Git"
+                                         code:-1
```
HiKumar-MS: Suggestion: Ideally each error code should be unique to identify the failure. Helps when you want the caller to handle particular error conditions (by comparing error code numbers/enums).

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSMessageListener.m
```diff
@@ -0,0 +1,133 @@
...
+                                               object:nil];
+    
+    self.connectionHandle = [[NSFileHandle alloc] initWithFileDescriptor:socketDescriptor
+                                                          closeOnDealloc:YES];
+    [self.connectionHandle acceptConnectionInBackgroundAndNotify];
+    
+    return YES;
+}
+
+- (void)stopListening
```
HiKumar-MS: Add a -dealloc method to ensure the cleanup happens correctly on termination or whenever the object is released. Otherwise you are relying on the caller to call this appropriately. It may be fine right now because stopListening is called on application terminate, but could change with code refactoring etc.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSMessageParser.m
alameenshah: Moved Notification Agent specific errors to a separate file - `VFSNotificationErrors.h`. `VFSNotificationErrors.m` has only a single line with definition of the Domain - `VFSForGitNotificationErrorDomain`.

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSForGitNotification.m
alameenshah: Added Notification specific error codes in `VFSNotificationErrors.h`

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/IPC/VFSMessageListener.m
alameenshah: yep... added.

Couple notes
1. `[super dealloc]` is not added because of `ARC`.
2. [Apple docs](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1413994-removeobserver?language=objc) suggest it is not required to un-register notification observers from inside a `dealloc` method.

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -118,6 +118,9 @@ function CopyBinariesToInstall()
     copyNotificationApp="cp -Rf \"${VFS_OUTPUTDIR}/GVFS.Notifications/VFSForGit.Mac/Build/Products/$CONFIGURATION/VFS For Git.app\" \"${STAGINGDIR}/${LIBRARYAPPSUPPORTDESTINATION}/.\""
     eval $copyNotificationApp || exit 1
     
+    copyNotificationPlist="cp -Rf \"${SOURCEDIRECTORY}/../GVFS.Notifications/VFSForGit.Mac/org.vfsforgit.usernotification.plist\" \"${STAGINGDIR}/${AGENTPLISTDESTINATION}/.\""
```
alameenshah: installs `VFS For Git.app` as launch agent

GVFS/GVFS.Installer.Mac/scripts/preinstall
```diff
@@ -10,20 +10,26 @@ fi
...
+        fi
+    done
+done
 
 KEXTBUNDLEID="org.vfsforgit.PrjFSKext"
 isKextLoadedCmd="/usr/sbin/kextstat -l -b $KEXTBUNDLEID | wc -l"
 isKextLoaded=$(eval $isKextLoadedCmd)
 if [ "$isKextLoaded" -gt 0 ]; then
-	unloadCmd="/sbin/kextunload -b $KEXTBUNDLEID"
+    unloadCmd="/sbin/kextunload -b $KEXTBUNDLEID"
```
alameenshah: replaced a tab with spaces

GVFS/GVFS.Installer.Mac/scripts/preinstall
```diff
@@ -32,7 +38,7 @@ LEGACYKEXTBUNDLEID="io.gvfs.PrjFSKext"
 isKextLoadedCmd="/usr/sbin/kextstat -l -b $LEGACYKEXTBUNDLEID | wc -l"
 isKextLoaded=$(eval $isKextLoadedCmd)
 if [ "$isKextLoaded" -gt 0 ]; then
-	unloadCmd="/sbin/kextunload -b $LEGACYKEXTBUNDLEID"
+    unloadCmd="/sbin/kextunload -b $LEGACYKEXTBUNDLEID"
```
alameenshah: replaced a tab with spaces

GVFS/GVFS.Service/RepoRegistry.cs
```diff
@@ -178,19 +181,27 @@ public void AutoMountRepos(string userId, int sessionId)
                     return;
                 }
 
-                this.SendNotification(sessionId, "GVFS AutoMount", "Attempting to mount {0} GVFS repo(s)", activeRepos.Count);
+                this.SendNotification(
+                    sessionId,
+                    NamedPipeMessages.Notification.Request.Identifier.AutomountStart,
```
jrbriggs: How should we thinking about interacting with users on macOS via notifications? Do they expect these to be fired only for special situations, or do they expect chatty notifications? I worry that notifying that we're about to mount  X repos and then firing X success messages is too much. We could
1. Only fire failure to mount
2. Continue firing start, but only fire one notification summarizing what mounted successfully.

Thoughts?

jrbriggs: The more I think about it, the more I like option 2. It's also a nice clue that you can start working. Thoughts?

alameenshah: yes, that makes sense. Let me take a look at the options to minimize auto-mount notifications count...

GVFS/GVFS.Notifications/VFSForGit.Mac/StatusMenuItem/UI/VFSAppDelegate.m
```diff
@@ -1,30 +1,41 @@
...
+    self.messageListener = [[VFSMessageListener alloc]
+        initWithSocket:NSTemporaryDirectory()
+        callback:^(NSDictionary *messageInfo)
+        {
+            [self displayNotification:messageInfo];
+        }];
+    
+    [self.messageListener startListening];
     
     VFSProductInfoFetcher *productInfoFetcher =
```
nickgra: nit: what's going on with the whitespace here? Can we try getting Xcode to make the whitespace conform to the usual standard?

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -59,6 +59,10 @@ xcodebuild -configuration $CONFIGURATION -workspace $NATIVEDIR/GVFS.Native.Mac.x
 USERNOTIFICATIONDIR=$VFS_SRCDIR/GVFS/GVFS.Notifications/VFSForGit.Mac
 USERNOTIFICATIONPROJECT="$USERNOTIFICATIONDIR/VFSForGit.xcodeproj"
 USERNOTIFICATIONSCHEME="VFS For Git"
+updateAppVersionCmd="(cd \"$USERNOTIFICATIONDIR\" && /usr/bin/xcrun agvtool new-marketing-version \"$VERSION\")"
+echo $updateAppVersionCmd
```
nickgra: For debugging?

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -59,6 +59,10 @@ xcodebuild -configuration $CONFIGURATION -workspace $NATIVEDIR/GVFS.Native.Mac.x
 USERNOTIFICATIONDIR=$VFS_SRCDIR/GVFS/GVFS.Notifications/VFSForGit.Mac
 USERNOTIFICATIONPROJECT="$USERNOTIFICATIONDIR/VFSForGit.xcodeproj"
 USERNOTIFICATIONSCHEME="VFS For Git"
+updateAppVersionCmd="(cd \"$USERNOTIFICATIONDIR\" && /usr/bin/xcrun agvtool new-marketing-version \"$VERSION\")"
```
nickgra: How does a `new-marketing-version` interact with the Info.plist? Should we move to using this to set the version of everything else that is native (the kext, log daemons, etc)?

Scripts/Mac/BuildGVFSForMac.sh
alameenshah: ya. just to make sure the project directory and version are specified correctly.

Scripts/Mac/BuildGVFSForMac.sh
alameenshah: `new-marketing-version` updates `CFBundleShortVersionString` in the `Info.plist` file. 

>>Should we move to using this to set the version of everything else that is native (the kext, log daemons, etc)?

I can take a look at how this is done on other native code and we can discuss offline.

GVFS/GVFS.Service/RepoRegistry.cs
alameenshah: As discussed - notifying only on auto-mount failures.

PR #1203, opened 2019-05-24
[TESTING] Comment out code for replacing inbox ProjFS in functional tests

wilbaker:


GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -30,8 +30,10 @@ public static void Main(string[] args)
 
             if (runner.HasCustomArg("--replace-inbox-projfs"))
             {
-                Console.WriteLine("Tests will replace inbox ProjFS");
-                GVFSTestConfig.ReplaceInboxProjFS = true;
+                Console.WriteLine("Pausing to allow for manual adjustments of the GVFS installation");
+                System.Threading.Thread.Sleep(1000 * 60 * 5);
```
wilbaker: This sleep was only needed because this branch doesn't have the Managed API fix and I had to manually install it.  When the fix is ready, this sleep can be removed.

PR #1207, opened 2019-05-28
Mac kext: Fix for potential vnode_put() race in VirtualizationRoot_RegisterProviderForPath

pmj:
Previously, VirtualizationRoot_RegisterProviderForPath() called vnode_mount() on a vnode that was not guaranteed to have a strong reference on it after the lock was dropped. This patch changes InsertVirtualizationRoot_Locked so it calls vnode_get() on the vnode in the case of an active provider rather than leaving the outer function to deal with it, and VirtualizationRoot_RegisterProviderForPath() now always retains the iocount until *after* disabling the auth cache, subsequently always dropping it with `vnode_put()`.

Due to the change, InsertVirtualizationRoot_Locked() now also pairs nicely with ActiveProvider_Disconnect(), which tidies up some unit tests a little.

**Notes:**

 * This change builds on #1200, this PR is only about the final commit.
 * This fix first appeared in #500 but as that still has not been merged and #1200 resurfaced the issue, I've submitted it in isolation.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -389,6 +389,14 @@ KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProvid
     {
         assert(rootIndex < s_maxVirtualizationRoots);
         assert(!s_virtualizationRoots[rootIndex].inUse);
+        
+        // Retain a strong reference to the vnode if we have an active provider on it
+        if (userClient != nullptr)
+        {
+            errno_t error = vnode_get(vnode);
+            assertf(error == 0, "The calling code should already hold an iocount on vnode, so vnode_get should never fail. error = %d", error);
+        }
+        
         VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
         
         root->providerUserClient = userClient;
@@ -478,7 +486,10 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
                             strlcpy(root.path, virtualizationRootCanonicalPath, sizeof(root.path));
                             KextLog_File(virtualizationRootVNode, "VirtualizationRoot_RegisterProviderForPath: registered provider (PID %d, IOUC %p) for virtualization root %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u.",
                                 clientPID, KextLog_Unslide(userClient), rootIndex, root.path, root.rootFsid.val[0], root.rootFsid.val[1], root.rootInode, KextLog_Unslide(virtualizationRootVNode), rootVid);
-                            virtualizationRootVNode = NULLVP; // transfer ownership
+
+                            // Acquire strong reference while provider is connected
+                            err = vnode_get(virtualizationRootVNode);
+                            assert(err == 0); // we already hold 1 strong reference from vnode_lookup so this should always succeed
                         }
                     }
                     else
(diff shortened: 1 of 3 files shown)


ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -510,17 +518,16 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
         }
     }
     
-    if (NULLVP != virtualizationRootVNode)
+    if (VirtualizationRoot_IsValidRootHandle(rootIndex))
     {
-        vnode_put(virtualizationRootVNode);
+        vfs_setauthcache_ttl(vnode_mount(virtualizationRootVNode), 0);
```
pmj: Ideally, I'd like to verify this in tests by requiring `vnode_mount()` to only be called on vnodes with an active refcount, but this is currently not practical as a vast number of unit tests benignly violate this elsewhere.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -510,17 +518,16 @@ VirtualizationRootResult VirtualizationRoot_RegisterProviderForPath(PrjFSProvide
...
     }
-    
-    if (rootIndex >= 0)
+
+    if (NULLVP != virtualizationRootVNode)
     {
-        VirtualizationRoot* root = &s_virtualizationRoots[rootIndex];
-        vfs_setauthcache_ttl(vnode_mount(root->rootVNode), 0);
+        vnode_put(virtualizationRootVNode);
```
jeschu1: Newbie to this area of code, I see how we have a "get" on both paths above. 

How did this work previously?  What woudl the "put" be pairing with?

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -394,7 +394,6 @@ - (void)testShouldHandleFileOpEvent_RootHandle {
 
     if (VirtualizationRoot_IsValidRootHandle(testRootHandle))
     {
-        vnode_get(repoRootVnode.get()); // InsertVirtualizationRoot_Locked doesn't _get directly, but ActiveProvider_Disconnect does _put
         ActiveProvider_Disconnect(testRootHandle, &userClient);
```
jeschu1: The changes in this file make me very happy :-)

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Previously, we'd skip both a put and a get by setting `virtualizationRootVNode = NULLVP;` on line 491, which has now been deleted. This particular `vnode_put()` here now always pairs with the `vnode_lookup()` at the top of the function - `vnode_lookup` returns a vnode with an iocount if it succeeds. Previously we'd normally pair that with a `vnode_put` when the provider disconnects, but there's a race condition where `vfs_setauthcache_ttl(vnode_mount(root->rootVNode), 0)` could hypothetically be called on a vnode without a refcount if the provider immediately disconnected. This way is also somewhat clearer in my opinion - and it makes the next stage, fixing the `s_virtualizationRoots` resize, more straightforward.

pmj:
No changes in the latest push, just rebased on latest master to keep the branch fresh while awaiting further review.

pmj:
I've rebased this branch on master and fixed the conflicts, no changes otherwise.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -389,6 +389,14 @@ KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProvid
     {
         assert(rootIndex < s_maxVirtualizationRoots);
         assert(!s_virtualizationRoots[rootIndex].inUse);
+        
+        // Retain a strong reference to the vnode if we have an active provider on it
```
wilbaker: Is the main reason for doing this here (rather than in `VirtualizationRoot_RegisterProviderForPath`) that it makes the unit tests cleaner?

pmj: Having `InsertVirtualizationRoot_Locked` pair nicely with `ActiveProvider_Disconnect()` also is a necessary requirement for the `s_virtualizationRoots` array resize tidyup that was part of #500 together with this fix. I'm going to pull that cleanup out into a separate PR as well once the higher priority stuff has been dealt with.

Issue #182, opened 2018-08-16
Mac: Ensure that the kext blocks file I\O that that is not allowed when VFSForGit is unmounted


wilbaker:
We *used* to have this on Mac, but it had to be removed in #191 as it was preventing:

- VFS4G from modifying the .git/hooks a as part of mounting
- `gvfs repair` from being able to fix up files in the .git folder

As the feature is re-enabled we'll need to consider how we can still support the scenarios above.  Perhaps we can support a new xattr that VFS4G sets on folders letting the kext know that offline modifications are allowed?  Or maybe there is some way that we can ACL/mark the VFS4G processes so that the kext knows that they should be allowed to make offline modifications?

pmj:
> Or maybe there is some way that we can ACL/mark the VFS4G processes so that the kext knows that they should be allowed to make offline modifications?

Can we have these processes, or at least a parent process of these processes register with the kext? From the kext's point of view, this is going to be the easiest method: any process wanting to modify offline roots needs to open a connection to the kext. We can either create a new `IOUserClient` subclass for this, or possibly roll it into the "provider" client subclass - for example we could start allowing offline modifications from provider `IOServiceOpen()` onwards, before registering as a provider with a path. If the process doesn't actually need to become a provider, it just stays in this state until it's done (`IOServiceClose()` or process exits).

If it is helpful, we can also allow entire process hierarchies to modify offline roots, in a similar way to how we check if a process was spawned by a user by walking up the parent process chain.

wilbaker:
> Can we have these processes, or at least a parent process of these processes register with the kext? 

We could have these processes register with the kext.  In both the cases called out in the description:

- VFS4G from modifying the .git/hooks a as part of mounting
- `gvfs repair` from being able to fix up files in the .git folder

We own the code and I don't think it would be a big deal for the process to make a call to register itself.

>  If the process doesn't actually need to become a provider, it just stays in this state until it's done (IOServiceClose() or process exits).

There are no plans to have these processes become providers.  They might launch the provider process as part of doing their work, but we could have them unregister with the kext before doing so (in case that simplifies anything).

CC: @kewillford as FYI and to double check my statements on provider/process lifetimes

wilbaker:
> If it is helpful, we can also allow entire process hierarchies to modify offline roots, in a similar way to how we check if a process was spawned by a user by walking up the parent process chain.

We *might* need these, I'll have to check if `gvfs repair` launches any child processes.

[...]

PR #1208, opened 2019-05-28
Mac ProjFS: Deny I/O on offline roots, with exceptions, part 1

pmj:
This is the first stage of implementing #182 - the kext now blocks write I/O to empty or placeholder files on offline roots.

I/O needs to be blocked when:

1. A message failed to deliver to the provider.
2. A process attempts to write to an empty file in an offline root (this file would subsequently be overwritten by a hydration event)
3. A process attempts to write to a placeholder (hydrated unmodified) file when the provider is offline. The file would not show up in git status and could be overwritten by a subsequent checkout/rebase/merge.
4. A process attempts to rename a file in an offline root.
5. A process attempts to read/execute an empty file in an offline root. The result of the read would be bad data, so failing with denied authorisation is preferable to letting the bad data propagate.
6. A process attempts to create files or directories in an offline root.

This change implements cases 1-4. Cases 5-6 will be covered in a future change.

At first glance this seems like it would be a very simple change, but in practice, some processes must be allowed unfettered access to files in offline roots. So a large part of this patch set is dedicated to implementing a system for allowing exceptions - there is a new type of IOKit user client for processes which need unrestricted access to offline roots, and this needed to be implemented all the way through from kext, via native lib, to (managed) VFS4G code proper.

Next, the vnode handler, which is in the actual business of denying or allowing file access, has been expanded to allow returning different results based on whether the root's provider is offline.

Finally, "deny" results are actually returned for a number of different cases - so far, this only includes writes. Even this caused some test failures, so the test process itself needs to register as an exception with unrestricted offline root access. This code was kindly provided by @jeschu1.

**Review Notes:**

This is a pretty big change set so I recommend reviewing commit-by-commit:

 1. **Minor macro tweak** - tiny bug fix to the `KextLog_VnodeOp` macro for an issue I ran into while working on this. The `vnodeType` argument can itself be the result of an expression, and so must be surrounded by parentheses to avoid operator precedence problems.
 2. **Adds system for registering processes for offline root I/O** - This implements the kext-side `IOUserClient` subclass for processes which need access to offline roots, code for maintaining the active list of these exceptions, plus some seeds of actual exception logic in `TryGetVirtualizationRoot`. (At this stage no exception is actually needed because no access is denied yet.) There are also unit tests for covering much of this new code.
 3. **User-space support for offline I/O process registration.** - This adds the ability to register and deregister as an offline root exception process to the native PrjFS library, along with calls to the new functions from the managed side from the `Main()` method of the `gvfs` tool.
 4. **Preparation for denying access to offline roots.** - This change identifies the various sites in the vnode handler where access will need to be denied, and prepares the code such that this behaviour can easily be enabled in a subsequent change. Doing so adds some new code paths, so new tests are provided to cover those paths.
 5. **Register/Unregister for offline I/O in functional tests** - This commit by @jeschu1 registers the functional test runner as a process which is allowed to perform I/O on offline roots. This is a requirement for the tests to pass once the subsequent commits are applied.
 6. **Denies access when provider does not respond** - If the provider process does not respond to a message from the vnode handler, we deny the operation, whatever it was. This is perhaps not ideal; a cleaner way to implement this would be to behave as if the provider was offline. That looks like it'll involve some refactoring of the vnode/fileop handlers, which I'll leave for #900.
 7. **Denies write access to empty files in offline root** - What it says on the tin. Writing to empty files makes no sense, when the provider comes online, it'll think the file is still empty and hydrate it, overwriting whatever the user stored there. Adds a test that verifies that such accesses are indeed denied by the vnode handler. 
 8. **Denies write access to placeholder files in offline root** - Placeholder (hydrated, non-full) files in an offline root must not be writable, or the provider will miss the PreConvertToFull/MileModified events and incorrectly track it as unmodified. This is prevented by this change; tests are updated to reflect the new expected behaviour.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
@@ -4,6 +4,7 @@ using NUnit.Framework;
 using System;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 
 namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
@@ -12,6 +13,26 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
     [Category(Categories.ExtraCoverage)]
     public class RepairTests : TestsWithEnlistmentPerTestCase
     {
+        private const string PrjFSLibPath = "libPrjFSLib.dylib";
+
+        [OneTimeSetUp]
+        public void TurnOfflineIOOn()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                 RegisterForOfflineIO();
+            }
+        }
+
+        [OneTimeTearDown]
+        public void TurnOfflineIOOff()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                UnregisterForOfflineIO();
+            }
+        }
+
         [TestCase]
         public void NoFixesNeeded()
         {
@@ -153,6 +174,12 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerTestCase
             result.ExitCode.ShouldEqual(0, result.Errors);
             this.Enlistment.MountGVFS();
         }
+
+        [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_RegisterForOfflineIO")]
+        private static extern uint RegisterForOfflineIO();
+
+        [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_UnregisterForOfflineIO")]
+        private static extern uint UnregisterForOfflineIO();
 
         private void CreateCorruptIndexAndRename(string indexPath, Action<FileStream, FileStream> corruptionAction)
         {
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -133,6 +133,7 @@ static SpinLock s_renameLock;
 static PendingRenameOperation* s_pendingRenames = nullptr;
 KEXT_STATIC uint32_t s_pendingRenameCount = 0;
 KEXT_STATIC uint32_t s_maxPendingRenames = 0;
+static bool s_osSupportsRenameDetection = false;
 
 // Public functions
 kern_return_t KauthHandler_Init()
@@ -252,7 +253,9 @@ KEXT_STATIC void UseMainForkIfNamedStream(
 
 KEXT_STATIC bool InitPendingRenames()
 {
-    if (version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave) // Only need to track renames on Mojave and newer
+    // Only need to/can track renames on Mojave and newer
+    s_osSupportsRenameDetection = (version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave);
+    if (s_osSupportsRenameDetection)
     {
         s_renameLock = SpinLock_Alloc();
         s_maxPendingRenames = 8; // Arbitrary choice, should be maximum number of expected concurrent threads performing renames, but array will resize on demand
@@ -271,7 +274,7 @@ KEXT_STATIC bool InitPendingRenames()
 
 KEXT_STATIC void CleanupPendingRenames()
 {
-    if (version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave)
+    if (s_osSupportsRenameDetection)
     {
         if (SpinLock_IsValid(s_renameLock))
         {
@@ -325,7 +328,7 @@ KEXT_STATIC void ResizePendingRenames(uint32_t newMaxPendingRenames)
 
 KEXT_STATIC void RecordPendingRenameOperation(vnode_t vnode)
 {
-    assertf(version_major >= PrjFSDarwinMajorVersion::MacOS10_14_Mojave, "This function should only be called from the KAUTH_FILEOP_WILL_RENAME handler, which is only supported by Darwin 18 (macOS 10.14 Mojave) and newer (version_major = %u)", version_major);
+    assertf(s_osSupportsRenameDetection, "This function should only be called from the KAUTH_FILEOP_WILL_RENAME handler, which is only supported by Darwin 18 (macOS 10.14 Mojave) and newer (version_major = %u)", version_major);
     thread_t myThread = current_thread();
     
     bool resizeTable;
@@ -370,7 +373,7 @@ KEXT_STATIC void RecordPendingRenameOperation(vnode_t vnode)
 
 KEXT_STATIC bool DeleteOpIsForRename(vnode_t vnode)
 {
-    if (version_major < PrjFSDarwinMajorVersion::MacOS10_14_Mojave)
+    if (!s_osSupportsRenameDetection)
     {
         // High Sierra and earlier do not support WILL_RENAME notification, so we have to assume any delete may be caused by a rename
         assert(s_pendingRenameCount == 0);
@@ -493,7 +496,10 @@ KEXT_STATIC int HandleVnodeOperation(
                         pid,
                         // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
                         CallbackPolicy_UserInitiatedOnly,
-                        false, // allow deleting offline directories even if not expanded
+                        // We want to block directory renames when provider is offline, but we can only do this on
+                        // newer OS versions as any delete operation could be a rename on High Sierra and older,
+                        // so on those versions, isRename is true for all delete events.
+                        s_osSupportsRenameDetection,
                         &root,
                         &vnodeFsidInode,
                         &kauthResult,
@@ -706,7 +712,9 @@ KEXT_STATIC int HandleVnodeOperation(
                 pid,
                 // Allow any user to delete individual files, as this generally doesn't cause nested kauth callbacks.
                 CallbackPolicy_AllowAny,
-                false, // allow deletes even if provider offline
+                // Allow deletes even if provider offline, except renames. Allow all deletes
+                // on OS versions where we can't distinguish renames & other deletes.
+                isRename && s_osSupportsRenameDetection,
                 &root,
                 &vnodeFsidInode,
                 &kauthResult,
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -669,8 +669,8 @@ KEXT_STATIC int HandleVnodeOperation(
                 if (providerStatus == Provider_IsOffline)
                 {
                     assert(!shouldMessageProvider);
-                    // TODO(Mac): Write access to placeholder files in offline roots should be denied (#182)
-                    //kauthResult = KAUTH_RESULT_DENY;
+                    // Deny write access to placeholders in an offline root
+                    kauthResult = KAUTH_RESULT_DENY;
                     goto CleanupAndReturn;
                 }
                 
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -572,6 +572,12 @@ KEXT_STATIC int HandleVnodeOperation(
         {
             if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
+                bool isWriteOperation = ActionBitIsSet(
+                    action,
+                    KAUTH_VNODE_WRITE_ATTRIBUTES |
+                    KAUTH_VNODE_WRITE_EXTATTRIBUTES |
+                    KAUTH_VNODE_WRITE_DATA |
+                    KAUTH_VNODE_APPEND_DATA);
                 if (!TryGetVirtualizationRoot(
                         &perfTracer,
                         context,
@@ -579,8 +585,8 @@ KEXT_STATIC int HandleVnodeOperation(
                         pid,
                         // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
                         CallbackPolicy_UserInitiatedOnly,
-                        // TODO(Mac): Access to empty files in an offline root should be denied except for deletions (#182)
-                        false, // denyIfOffline
+                        // Prevent write access to empty files in offline roots. For now allow reads, and always allow the user to delete files.
+                        isWriteOperation,
                         &root,
                         &vnodeFsidInode,
                         &kauthResult,
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -28,12 +28,6 @@
 #include "KauthHandlerTestable.hpp"
 #endif
 
-template <typename T, typename MIN_T, typename MAX_T>
-    auto clamp(const T& value, const MIN_T& min, const MAX_T& max)
-{
-    return value < min ? min : (value > max ? max : value);
-}
-
 enum ProviderCallbackPolicy
 {
     CallbackPolicy_AllowAny,
@@ -1132,6 +1126,10 @@ static bool TryGetVirtualizationRoot(
         perfTracer->IncrementCount(PrjFSPerfCounter_VnodeOp_GetVirtualizationRoot_ProviderOffline);
         
         *kauthResult = KAUTH_RESULT_DEFER;
+        if (!VirtualizationRoots_ProcessMayAccessOfflineRoots(pidMakingRequest))
+        {
+        }
+        
         return false;
     }
     
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -34,6 +34,13 @@ enum ProviderCallbackPolicy
     CallbackPolicy_UserInitiatedOnly,
 };
 
+enum ProviderStatus
+{
+    Provider_StatusUnknown,
+    Provider_IsOffline,
+    Provider_IsOnline,
+};
+
 struct PendingRenameOperation
 {
     vnode_t vnode;
@@ -91,12 +98,14 @@ static bool TryGetVirtualizationRoot(
     const vnode_t vnode,
     pid_t pidMakingRequest,
     ProviderCallbackPolicy callbackPolicy,
-    
+    bool denyIfOffline,
+
     // Out params:
     VirtualizationRootHandle* root,
     FsidInode* vnodeFsidInode,
     int* kauthResult,
-    int* kauthError);
+    int* kauthError,
+    ProviderStatus* _Nullable providerStatus = nullptr);
 
 KEXT_STATIC bool ShouldHandleFileOpEvent(
     // In params:
@@ -477,8 +486,18 @@ KEXT_STATIC int HandleVnodeOperation(
             // Recursively expand directory on rename as user will expect the moved directory to have the same contents as in its original location
             if (isRename)
             {
-                // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                if (!TryGetVirtualizationRoot(
+                        &perfTracer,
+                        context,
+                        currentVnode,
+                        pid,
+                        // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
+                        CallbackPolicy_UserInitiatedOnly,
+                        false, // allow deleting offline directories even if not expanded
+                        &root,
+                        &vnodeFsidInode,
+                        &kauthResult,
+                        kauthError))
                 {
                     goto CleanupAndReturn;
                 }
@@ -502,8 +521,18 @@ KEXT_STATIC int HandleVnodeOperation(
             }
             else if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
-                // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                if (!TryGetVirtualizationRoot(
+                        &perfTracer,
+                        context,
+                        currentVnode,
+                        pid,
+                        // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
+                        CallbackPolicy_UserInitiatedOnly,
+                        false, // allow reading offline directories even if not expanded
+                        &root,
+                        &vnodeFsidInode,
+                        &kauthResult,
+                        kauthError))
                 {
                     goto CleanupAndReturn;
                 }
@@ -543,8 +572,19 @@ KEXT_STATIC int HandleVnodeOperation(
         {
             if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
-                // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                if (!TryGetVirtualizationRoot(
+                        &perfTracer,
+                        context,
+                        currentVnode,
+                        pid,
+                        // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
+                        CallbackPolicy_UserInitiatedOnly,
+                        // TODO(Mac): Access to empty files in an offline root should be denied except for deletions (#182)
+                        false, // denyIfOffline
+                        &root,
+                        &vnodeFsidInode,
+                        &kauthResult,
+                        kauthError))
                 {
                     goto CleanupAndReturn;
                 }
@@ -569,19 +609,68 @@ KEXT_STATIC int HandleVnodeOperation(
             
             if (ActionBitIsSet(action, KAUTH_VNODE_WRITE_DATA | KAUTH_VNODE_APPEND_DATA))
             {
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                /* At this stage we know the file is NOT empty, but it could be
+                 * in either placeholder (hydrated) or full (modified/added to
+                 * watchlist) state.
+                 *
+                 * If it's a placeholder, we need to notify the provider before
+                 * allowing modifications so that it is tracked as a full file
+                 * (and the placeholder xattr is removed).
+                 * If the provider is offline in this situation, we want to
+                 * prevent writes (by most processes) to the file so that the
+                 * modifications don't go undetected.
+                 *
+                 * If it's already a full file on the other hand, nothing needs
+                 * to be done, so we can allow writes even when the provider is
+                 * offline.
+                 */
+                ProviderStatus providerStatus;
+                bool shouldMessageProvider = TryGetVirtualizationRoot(
+                    &perfTracer,
+                    context,
+                    currentVnode,
+                    pid,
+                    CallbackPolicy_UserInitiatedOnly,
+                    // At this stage, we don't yet know if the file is still a placeholder, so don't deny yet even if offline
+                    false, // denyIfOffline
+                    &root,
+                    &vnodeFsidInode,
+                    &kauthResult,
+                    kauthError,
+                    &providerStatus);
+                
+                if (!shouldMessageProvider)
                 {
-                    goto CleanupAndReturn;
+                    // If we don't need to message the provider, that's normally
+                    // the end of it, but in the case where the provider is
+                    // offline we need to do further checks.
+                    if (providerStatus != Provider_IsOffline)
+                    {
+                        goto CleanupAndReturn;
+                    }
                 }
                 
                 PrjFSFileXAttrData rootXattr = {};
                 SizeOrError xattrResult = Vnode_ReadXattr(currentVnode, PrjFSFileXAttrName, &rootXattr, sizeof(rootXattr));
                 if (xattrResult.error == ENOATTR)
                 {
-                    // Only notify if the file is still a placeholder
+                    // If the file does not have the attribute, this means it's
+                    // already in the "full" state, so no need to send a provider
+                    // message or block the I/O if the provider is offline
                     goto CleanupAndReturn;
                 }
                 
+                if (providerStatus == Provider_IsOffline)
+                {
+                    assert(!shouldMessageProvider);
+                    // TODO(Mac): Write access to placeholder files in offline roots should be denied (#182)
+                    //kauthResult = KAUTH_RESULT_DENY;
+                    goto CleanupAndReturn;
+                }
+                
+                assert(shouldMessageProvider);
+                assert(providerStatus == Provider_IsOnline);
+
                 PerfSample preConvertToFullSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreConvertToFull);
                 
                 if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
@@ -604,8 +693,18 @@ KEXT_STATIC int HandleVnodeOperation(
     
     if (isDeleteAction)
     {
-        // Allow any user to delete individual files, as this generally doesn't cause nested kauth callbacks.
-        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_AllowAny, &root, &vnodeFsidInode, &kauthResult, kauthError))
+        if (!TryGetVirtualizationRoot(
+                &perfTracer,
+                context,
+                currentVnode,
+                pid,
+                // Allow any user to delete individual files, as this generally doesn't cause nested kauth callbacks.
+                CallbackPolicy_AllowAny,
+                false, // allow deletes even if provider offline
+                &root,
+                &vnodeFsidInode,
+                &kauthResult,
+                kauthError))
         {
             goto CleanupAndReturn;
         }
@@ -1083,14 +1182,21 @@ static bool TryGetVirtualizationRoot(
     const vnode_t vnode,
     pid_t pidMakingRequest,
     ProviderCallbackPolicy callbackPolicy,
+    bool denyIfOffline,
 
     // Out params:
     VirtualizationRootHandle* root,
     FsidInode* vnodeFsidInode,
     int* kauthResult,
-    int* kauthError)
+    int* kauthError,
+    ProviderStatus* _Nullable providerStatus)
 {
     PerfSample findRootSample(perfTracer, PrjFSPerfCounter_VnodeOp_GetVirtualizationRoot);
+    
+    if (providerStatus != nullptr)
+    {
+        *providerStatus = Provider_StatusUnknown;
+    }
     
     *root = VnodeCache_FindRootForVnode(
         perfTracer,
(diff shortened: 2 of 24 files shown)


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -484,7 +493,9 @@ KEXT_STATIC int HandleVnodeOperation(
             if (isRename)
             {
                 // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly,
+                    false, // allow deleting offline directories even if not expanded
```
jeschu1: @wilbaker to double check do we want to allow deleting offline directories?

pmj: I'm not @wilbaker, but my understanding was that we want to support the case where a root is not mountable (e.g. access to server repo is no longer available, no network connection, etc.) but the user wants to delete the whole thing simply by performing an `rm -rf` on it.

jeschu1: :-) That makes sense as a use case.  @pmj was there a specific set of use cases allowed given as a guide?  I looked in the attached issue but may have missed something.

pmj: I don't think so. I just remember that case being mentioned at some point. Possibly as far back as our original kext prototype. In fact, it might be documented in the source code there?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -549,8 +562,13 @@ KEXT_STATIC int HandleVnodeOperation(
         {
             if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
-                // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                bool isWriteOperation = ActionBitIsSet(action, KAUTH_VNODE_WRITE_ATTRIBUTES | KAUTH_VNODE_WRITE_EXTATTRIBUTES | KAUTH_VNODE_WRITE_DATA | KAUTH_VNODE_APPEND_DATA);
```
jeschu1: nit: this line is a bit long.  Can we put the parameters for ActionBitIsSet on a separate line?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -584,10 +616,20 @@ KEXT_STATIC int HandleVnodeOperation(
                 SizeOrError xattrResult = Vnode_ReadXattr(currentVnode, PrjFSFileXAttrName, &rootXattr, sizeof(rootXattr));
                 if (xattrResult.error == ENOATTR)
                 {
-                    // Only notify if the file is still a placeholder
+                    // Only notify or block writes if the file is still a placeholder
                     goto CleanupAndReturn;
                 }
                 
+                if (providerStatus == Provider_IsOffline)
```
jeschu1: I'm finding the logic in this block a bit hard to follow.  Perhaps some comments would help?

If shouldMessageProvider is false, but providerStatus is Online or Unknown and kauthResult != deny we continue
Then is providerStatus is Offline, we assert if shouldMessageProvider?  

Why does this not follow the normal if (!shouldProvider) goto CleanupAndReturn flow?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -549,8 +562,13 @@ KEXT_STATIC int HandleVnodeOperation(
...
             {
-                // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                bool isWriteOperation = ActionBitIsSet(action, KAUTH_VNODE_WRITE_ATTRIBUTES | KAUTH_VNODE_WRITE_EXTATTRIBUTES | KAUTH_VNODE_WRITE_DATA | KAUTH_VNODE_APPEND_DATA);
+                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid,
+                    // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
+                    CallbackPolicy_UserInitiatedOnly,
+                    // Prevent write access to empty files in offline roots. For now allow reads, and always allow the user to delete files.
+                    isWriteOperation,
+                    &root, &vnodeFsidInode, &kauthResult, kauthError))
```
jeschu1: nit: Can we have all the params on their own line for readability?

ProjFS.Mac/PrjFSKext/KextLog.hpp
```diff
@@ -110,7 +110,7 @@ template <typename... args>
 
 #define KextLog_VnodeOp(vnode, vnodeType, procname, action, message) \
     do { \
-        if (VDIR == vnodeType) \
+        if (VDIR == (vnodeType)) \
```
jeschu1: Are these ()'s needed?

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
```diff
@@ -180,7 +180,7 @@ bool ProviderMessaging_TrySendRequestAndWaitForResponse(
         if (0 != sendError)
         {
             // TODO: appropriately handle unresponsive providers
-            *kauthResult = KAUTH_RESULT_DEFER;
+            *kauthResult = KAUTH_RESULT_DENY;
```
jeschu1: This change seems unrelated to the PR, but I think makes sense.  Can you confirm?
If we can't send a message fail the request.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -653,3 +667,112 @@ VirtualizationRootHandle ActiveProvider_FindForPath(const char* _Nonnull path)
     
     return matchingHandle;
 }
+
+bool VirtualizationRoots_ProcessMayAccessOfflineRoots(pid_t pid)
```
jeschu1: I want to make sure I understand this flow.  Is the check here to verify the request started when we were in offline mode?  How does a pid get into this array?  What are we trying to prevent here?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -653,3 +667,112 @@ VirtualizationRootHandle ActiveProvider_FindForPath(const char* _Nonnull path)
...
+    return result;
+}
+
+bool VirtualizationRoots_AddOfflineIOProcess(pid_t pid)
+{
+    bool success = false;
+    bool retry;
+    pid_t* resizeMemory = nullptr;
+    uint32_t resizeMemoryArrayLength = 0;
+
```
jeschu1: Same question I always ask, do we need the complexity of a resizable array here or do we want to say we can handle at most 1000 offline processes. I imagine I know the answer :-).

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -653,3 +667,112 @@ VirtualizationRootHandle ActiveProvider_FindForPath(const char* _Nonnull path)
...
+{
+    bool success = false;
+    bool retry;
+    pid_t* resizeMemory = nullptr;
+    uint32_t resizeMemoryArrayLength = 0;
+    
+    do
+    {
+        retry = false;
+        RWLock_AcquireExclusive(s_virtualizationRootsLock);
```
jeschu1: It would be easier for me to read if we did

if (s_offlineIOPIDCount >= s_maxOfflineIOPIDs)
{
   - resize logic
}

Add offlineRoot

Not sure why this needs to be in a do/while

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: @pmj I think it would be useful to document the list of offline operations we are allowing in the PR description.  
I can see why delete is allowed.  However, if someone unmounts, deletes a single file, and then remounts we'll have incorrect status.  This may be a known possibility with the Windows implemenation.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Walking through the logic top to bottom:

* If the operation would cause no message to be sent to a provider, we're done with the logic, because there's nothing for an offline provider to miss.
* If `TryGetVirtualizationRoot` decided internally that the operation should be blocked, there's also not really anything else to do. I guess the `providerStatus != Provider_IsOffline` check looks redundant in this case, but I'll double-check that before I remove it.

Next, the attribute check:

* If the file is *not* a placeholder (it's a "full" file), we:
 - don't need to `NotifyFilePreConvertToFull`
 - we can allow writes even if the provider is offline because it does no harm (there's no message an offline provider would miss)
* If it *is* a placeholder, we need to send a message if the provider is online, or block the write if it's offline.

I agree a few more comments would help here.

ProjFS.Mac/PrjFSKext/KextLog.hpp
pmj: Yes, because it's a macro. When used, the arguments are simply expanded as text, which can have unintended operator precedence consequences. For example, if we have:

    KextLog_VnodeOp(vn, isDir ? VDIR : VREG, ...);

The `if` statement expands to:

   if (VDIR == isDir ? VDIR : VREG)

Which is equivalent to:

   if ((VDIR == isDir) ? VDIR : VREG)

or

   if (true)

because both `VDIR` and `VREG` are non-zero. This is basically never what the user of the macro wanted!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -584,10 +616,20 @@ KEXT_STATIC int HandleVnodeOperation(
...
-                    // Only notify if the file is still a placeholder
+                    // Only notify or block writes if the file is still a placeholder
                     goto CleanupAndReturn;
                 }
                 
+                if (providerStatus == Provider_IsOffline)
+                {
+                    assert(!shouldMessageProvider);
+                    // Deny write access to placeholders in an offline root
+                    kauthResult = KAUTH_RESULT_DENY;
```
jeschu1: If we got to this line we don't have the PrjFSFileXAttrName attribute and are a full file correct?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: This is the check for whether this is a process that has registered for requiring access to offline roots. `s_offlineIOPIDs` holds the PIDs for all processes that have opted-in to this.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: If we allow registration to fail because the array is full, that failure will propagate to user space, so certain `gvfs` commands could start failing. I'm not sure that's worth a slight reduction in complexity.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: This is the usual thing where we need to avoid allocating memory while holding the lock, and when we drop the lock to allocate, we introduce the possibility of a race condition, hence the need for a retry. So the loop needs to be there. It doesn't have to be structured in exactly this way, but the other option takes the following form:

    Lock();
    while (needsResizing?)
    {
      // [work out new array size]
      Unlock();
      // [alloc memory for new array size]
      Lock();
      // [try to resize array with memory alloc'd during dropped lock]
    }
    // insert item
    Unlock();

So you end up with those unbalanced lock/unlock calls which definitely violate our coding guidelines; I don't think there's much difference readability wise, but I can flesh out the alternate version for a full comparison if you like.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: No, we *do* have the attribute (as `xattrResult.error != ENOATTR`) and therefore have a placeholder file, so with the provider offline, we want to deny modifications.

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
pmj: It's not really unrelated. Prior to this PR, we've been allowing operations to go ahead under all circumstances. In this PR, we want to deny operations if allowing them to go ahead would mean that the provider misses messages. So the goal is to deny access regardless of the *reason* for the missed message - whether this is an offline provider or some other message delivery problem.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
jeschu1: Got it so basically any pids that have triggered "PrjFS_RegisterForOfflineIO" would be allowed?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
jeschu1: The block you have above is super readable to me.  Especially if you put that block in comments.

Something like:

```
Lock()
bool needsResizing = s_offlineIOPIDCount >= s_maxOfflineIOPIDs;
while (needsResizing)
{
   // We may need to retry since we attempt to allocate memory outside of the lock
  // [work out new array size]
  Unlock();
  // [alloc memory for new array size]
  Lock();
  // [try to resize array with memory alloc'd during dropped lock]
  // If we successful resized, needsResizing = false
}
// insert item
Unlock();

```
Perhaps some of the variables are also confusing me a bit.  resizeMemory sounds like a bool, but it's the resizeMemory.

This is a someone foreign paradigm to most folks on my team, but I'm sure you have extensive experience with it.  Any changes we do to make it ridiculously easy to understand will be appreciated.

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
jeschu1: "or some other message delivery problem" this bit could have been existing.

I think it's a fine change and makes sense in this PR.

jeschu1:
On Windows it appears I can write to any hydrated file (no need to register the pid i.e. PrjFS_RegisterForOfflineIO).

Curious why we didnt' do the same as Windows here for Mac.

pmj:
> On Windows it appears I can write to any hydrated file (no need to register the pid i.e. PrjFS_RegisterForOfflineIO).

Hmm. I'm assuming this means any thusly modified file doesn't show up in `git status` when you do remount the root?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Correct, that's the idea.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Looking at this again, I was right about the `providerStatus != Provider_IsOffline` check.

`shouldMessageProvider` can be `false` if:

 1. The file is not in a virtualisation root.
 2. The provider is offline
 3. It is the provider itself that is causing the I/O.
 4. The I/O is caused by a non-user-initiated process.

Cases 1, 3, & 4 require no further processing.

In the offline provider case, we still need to decide if we need to block the access in case it would lead to a missed message (it would cause a placeholder -> full transition).

I'm restructuring this code slightly, and adding a lot more comments.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Done!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Done, for all calls to this function.

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
pmj: I've clarified with a comment, but yeah, I think we should leave this in.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: I've reworked the resize in a3618a4 - I'll leave this as a separate commit for now so the 2 variants can be compared side-by-side. I agree the new version is clearer.

pmj:
I've just rebased this PR on latest master and incorporated the review feedback so far. I've not made any changes outside review feedback.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've commented on the original issue #182 to solicit feedback on the list, and I've added the current list to the PR's description.

jeschu1:
> Hmm. I'm assuming this means any thusly modified file doesn't show up in git status when you do remount the root?

@pmj that is correct.

I had an offline conversation with @wilbaker  and the net net is that there are differences between the mac & windows implementation.  On windows you only modified hydrated files if the directory is marked as full.  We don't have that concept on Mac.  Therefore the way you've done this PR with registering for offline writes is the most ideal approach we could think of.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
jeschu1: I really like the new verison.  Thank you for taking the time to simplify.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -575,19 +621,68 @@ KEXT_STATIC int HandleVnodeOperation(
...
+                 * If the provider is offline in this situation, we want to
+                 * prevent writes (by most processes) to the file so that the
+                 * modifications don't go undetected.
+                 *
+                 * If it's already a full file on the other hand, nothing needs
+                 * to be done, so we can allow writes even when the provider is
+                 * offline.
+                 */
+                ProviderStatus providerStatus;
+                bool shouldMessageProvider = TryGetVirtualizationRoot(
```
jeschu1: nit: do we need a seperate varaible for shouldMessageProvider?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -575,19 +621,68 @@ KEXT_STATIC int HandleVnodeOperation(
...
+                    // already in the "full" state, so no need to send a provider
+                    // message or block the I/O if the provider is offline
                     goto CleanupAndReturn;
                 }
                 
+                if (providerStatus == Provider_IsOffline)
+                {
+                    assert(!shouldMessageProvider);
+                    // Deny write access to placeholders in an offline root
+                    kauthResult = KAUTH_RESULT_DENY;
```
jeschu1: I *think* this is OK but it would be worth a second pair of eyes.  
At this point we are ignoring any checks from TryGetVirtualizationRoot, and making the decision to deny solely on if we have a placeholder and we're offline.  That makes snese to me.

pmj: Note that `TryGetVirtualizationRoot` will *never* return `true` in the case of an offline provider, so there's not really anything to ignore.

jeschu1: @pmj what I meant is TryGetVirtualizationRoot could return `false` for a reason other than offline provider.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: > Isn't there also the case where we can have two threads do
> 
> resizeArrayLength = MIN(MAX(1u, s_maxOfflineIOPIDs * 2u), UINT32_MAX);
> And allocate a block of memory for that size?
> 
> If so I think everything works we just copy twice. Perhaps a comments explaining that to make it simple for the reader.

No, we shouldn't copy twice as the race will be detected by the `resizeArrayLength > s_maxOfflineIOPIDs` check, in which case one thread's alloc'd memory is freed again without using it. At least that's what the code is *supposed* to do, please check that also matches your interpretation. :-)

pmj:
> Isn't there also the case where we can have two threads do
> 
> resizeArrayLength = MIN(MAX(1u, s_maxOfflineIOPIDs * 2u), UINT32_MAX);
> And allocate a block of memory for that size?
> 
> If so I think everything works we just copy twice. Perhaps a comments explaining that to make it simple for the reader.

No, we shouldn't copy twice as the race will be detected by the `resizeArrayLength > s_maxOfflineIOPIDs` check, in which case one thread's alloc'd memory is freed again without using it. At least that's what the code is supposed to do, please check that also matches your interpretation. :-)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes, and that's already caught earlier on, on lines 656-662. (Of course if by "ignoring any checks" you mean we've already dealt with them then we're just vigorously agreeing with one another.)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I found it more readable this way, given the 12-line `TryGetVirtualizationRoot` function call.

jeschu1:
> o, we shouldn't copy twice as the race will be detected by the resizeArrayLength > s_maxOfflineIOPIDs check, in which case one thread's alloc'd memory is freed again without using it. At least that's what the code is supposed to do, please check that also matches your interpretation. :-)

:-) What I was thinking is

Thread 1 enters
   uint32_t resizeArrayLength = MIN(MAX(1u, s_maxOfflineIOPIDs * 2u), UINT32_MAX); 
Then release the lock and calls 
   pid_t* resizedArrayMemory = Memory_AllocArray<pid_t>(resizeArrayLength); 
It then waits for the lock

Thread 2 enters gets the lock (while thread 1 is waiting above)
   uint32_t resizeArrayLength = MIN(MAX(1u, s_maxOfflineIOPIDs * 2u), UINT32_MAX); 
Then calls 
   pid_t* resizedArrayMemory = Memory_AllocArray<pid_t>(resizeArrayLength); 

In that case they'd have the same array length I think.  Let me know if I missed something.  Threading is fun cause it's like little puzzles :-).

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -575,19 +621,68 @@ KEXT_STATIC int HandleVnodeOperation(
...
+                    goto CleanupAndReturn;
+                }
+                
+                assert(shouldMessageProvider);
```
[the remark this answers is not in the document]
pmj: I've left them in as a sanity check of the expected invariants - if they ever fail, `TryGetVirtualizationRoot()` has been modified to violate the assumption we've baked into this code that if the provider is offline, `TryGetVirtualizationRoot()` always returns false.

Hopefully with the #900 refactorings, we can simplify this entire code and reduce the level of defensive coding…

pmj:
> In that case they'd have the same array length I think. Let me know if I missed something. Threading is fun cause it's like little puzzles :-).

They will indeed have allocated the same amount of memory. Only one of them "wins" the `if (resizeArrayLength > s_maxOfflineIOPIDs)` check, however, and resizes the array. Once the array is resized, `s_maxOfflineIOPIDs` will be equal to `resizeArrayLength` by the time the second thread gets to that check, so it'll take the `else` branch which just frees the memory again.

It might make more sense if you read that `if` statement as "will this memory we've just allocated actually grow the array?"

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: I get that.  All of them are 12-lines though.  Since the other calls don't assign a variable and do an assert check later on I'd prefer to keep it consistent.  Otherwise me in a year reading this code goes, why did we do this one differently than the others?   

If you feel strongly about it can you add a comment as to why it breaks the convention the other ones have?  It certainly isn't wrong, but it will be a helpful refresher on why it's different.

pmj: I don't like the other call sites either, but the logic that follows them is much less complex so I don't feel the meaning of everything needs spelling out so much. I don't mind removing the variable though. (and the `assert()`s that check it later on)

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: I think we are :-)

jeschu1:
> They will indeed have allocated the same amount of memory. Only one of them "wins" the if (resizeArrayLength > s_maxOfflineIOPIDs) check, however, and resizes the array. Once the array is resized, s_maxOfflineIOPIDs will be equal to resizeArrayLength by the time the second thread gets to that check, so it'll take the else branch which just frees the memory again.

Ah yes, in the case I described s_maxOfflineIOPIDs would have been altered so the case holds.  THanks for pointing that out, yes looks like all cases are covered!

wilbaker:
> @wilbaker can you please also review?

Sure taking a look now

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -484,7 +493,17 @@ KEXT_STATIC int HandleVnodeOperation(
...
             {
                 // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
-                if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))
+                if (!TryGetVirtualizationRoot(
+                        &perfTracer,
+                        context,
+                        currentVnode,
+                        pid,
+                        CallbackPolicy_UserInitiatedOnly,
+                        false, // allow deleting offline directories even if not expanded
```
wilbaker: > `// allow deleting offline directories even if not expanded`

This is inside of `if (isRename)`, and we should not allow renames when the provider is offline.

Should the comment read something like this?:

```
// Until Mojave is set as our minimum version, we must allow offline 
// access when 'isRename' is true as on High Sierra 'isRename' is true for deletes (which should be allowed)
```

Additionally/alternatively should we check `version_major` and only pass in `false` when we're pre-Mojave?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -653,3 +668,103 @@ VirtualizationRootHandle ActiveProvider_FindForPath(const char* _Nonnull path)
...
+{
+    bool success = false;
+    
+    // Note: lock may temporarily be dropped within
+    RWLock_AcquireExclusive(s_virtualizationRootsLock);
+    {
+        while (s_offlineIOPIDCount >= s_maxOfflineIOPIDs && s_maxOfflineIOPIDs < UINT32_MAX)
+        {
+            // Array must be resized
+            uint32_t resizeArrayLength = MIN(MAX(1u, s_maxOfflineIOPIDs * 2u), UINT32_MAX);
```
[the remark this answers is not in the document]
wilbaker: Related: I agree with what @jeschu1 mentioned in an earlier comment that we're not gaining much from the user experience side with a resizable array (and the resizable array does make things more complicated).  

Users won't have more than a handful of `gvfs` commands running at once and so if we had a fixed length array with something like 64 entries I don't think we'd ever exceed it and it would be OK to fail the offline request if we did.

cc: @jrbriggs for his thoughts as well

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -881,6 +882,67 @@ -(void) testWriteFileHydrated {
...
+            _,
+            _,
+            _,
+            _,
+            _,
+            _,
+            nullptr));
+}
+
+-(void) testWriteFlaggedNonRepoFile
```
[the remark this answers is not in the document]
pmj: Correct. Without a test for this, we'd lack coverage for the following code path in `HandleVnodeOperation()`:

```
    if (isDirectory)
    {
        […]
    }
    else
    {
        if (isRename || […])
        {
            […]
            if (ActionBitIsSet(action, KAUTH_VNODE_WRITE_DATA | KAUTH_VNODE_APPEND_DATA))
            {
                […]
                if (!shouldMessageProvider)
                {
                    if (providerStatus != Provider_IsOffline)
                    {
                        goto CleanupAndReturn; // <------ HERE
                    }
                }
```

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -1129,16 +1191,21 @@ - (void) testWriteWithDisappearingVirtualizationRoot {
         [&]()
         {
             ActiveProvider_Disconnect(self->dummyRepoHandle, &self->dummyClient);
+            // mark as hydrated
+            testFileVnode->attrValues.va_flags &= ~FileFlags_IsEmpty;
```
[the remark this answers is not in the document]
pmj: I guess our code doesn't test that the empty flag was removed, only whether the xattr is there, so it's not required for passing the test to remove the flag, but I'd be uneasy with the half-arsed approach.

I'm going to re-check what exactly we're using this test for though, I no longer remember after a month so that's probably a good indication that we need some comments in there.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -1196,4 +1271,108 @@ - (void) testIneligibleFilesystemType
     XCTAssertFalse(MockCalls::DidCallFunction(ProviderMessaging_TrySendRequestAndWaitForResponse));
 }
 
+- (void) testOfflineRootDeniesAccessToEmptyFile
```
wilbaker: Would it be possible to add a test to validate the offline rename scenarios on pre/post Mojave?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: > If we allow registration to fail because the array is full, that failure will propagate to user space, so certain `gvfs` commands could start failing. I'm not sure that's worth a slight reduction in complexity.

On the flip side, if we introduce a bug in how our offline processes register/unregister it could lead to the array growing to a massive size (`UINT32_MAX / sizeof(s_offlineIOPIDs[0])`).

Users should never be running more than a handful of `gvfs commands at once` (and the functional tests are limited by the number of cores on the machine).

My preference is to pick a reasonably sized fixed-length array (e.g. 128) and that way benefit by:

- Not needing to worry about resizing the array
- Easier to detect bugs in the register/unregister flow

@jrbriggs, @jeschu1, do you have any concerns limiting ourselves in this way?

jeschu1: I believe the downside of a fixed array would be that if you have over 128 offline unique pids trying to write to a unmounted repo we deny the request for those over 128.

 I don't think we should ever hit 128 and if we do my vote would be to fail, since it could be a sign of something else gone wrong.

@pmj please let me know if my understanding is incorrect

wilbaker: > I don't think we should ever hit 128 and if we do my vote would be to fail, since it could be a sign of something else gone wrong.

I agree, if we find ourselves in this situation something has definitely gone wrong.

jrbriggs: I like the idea of just fixing the size at 128.

ProjFS.Mac/PrjFSKext/ProviderMessaging.cpp
```diff
@@ -180,7 +180,9 @@ bool ProviderMessaging_TrySendRequestAndWaitForResponse(
         if (0 != sendError)
         {
             // TODO: appropriately handle unresponsive providers
-            *kauthResult = KAUTH_RESULT_DEFER;
+
```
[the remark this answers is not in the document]
pmj: So the case

    if (0 != sendError)

actually doesn't necessarily cover (my understanding of) "unresponsive" providers. "Unresponsive" to me indicates we manage to deliver (some) messages to the provider  successfully, but the provider does not reply.

So while the comment is misplaced here, I think we should move it into the

            while (!message.receivedResult &&
                   !s_isShuttingDown)

loop where we are indeed not doing anything about unresponsive providers.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
pmj: Good idea, I've added a new test `testOfflineRootDeniesRename` as there's a bit more to do in that case and extending `testOfflineRootDeniesAccessToEmptyFile` would be messy.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -481,10 +485,12 @@ - (void) testDeleteFileNonRenamed
         _,
         _,
         nullptr));
+    CleanupPendingRenames();
```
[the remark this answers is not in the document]
pmj: The (new with this PR) `s_osSupportsRenameDetection` variable could end up being in an inconsistent state for these tests otherwise. I was on the fence whether to set `s_osSupportsRenameDetection` when setting `version_major` instead, but relying on uninitialised pending rename array state in the tests isn't great anyway. We specifically *don't* want to call `InitPendingRenames();`/`CleanupPendingRenames();` from `setUp`/`tearDown` because their version-specific so they need to be called inside the block passed to `TestForAllSupportedDarwinVersions()`, not before/after the whole test method.

It's not a super nice solution, I'll grant you that. In an ideal world we'd probably do some fancy stuff to run the whole test suite for each supported `major_version` instead of doing that on a test-by-test basis, but I don't have an immediate suggestion on how to do that with XCTest.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerTestCase/RepairTests.cs
```diff
@@ -154,6 +175,12 @@ public void FixesCorruptGitConfig()
             this.Enlistment.MountGVFS();
         }
 
+        [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_RegisterForOfflineIO")]
+        private static extern uint RegisterForOfflineIO();
+
+        [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_DeregisterForOfflineIO")]
+        private static extern uint DeregisterForOfflineIO();
```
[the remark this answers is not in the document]
pmj: All these instances should now be renamed.

ProjFS.Mac/PrjFSKext/PrjFSService.cpp
```diff
@@ -124,6 +125,16 @@ IOReturn PrjFSService::newUserClient(
             }
         }
         break;
+    case UserClientType_OfflineIO:
+        {
+            PrjFSOfflineIOUserClient* provider_client = new PrjFSOfflineIOUserClient();
```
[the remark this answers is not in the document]
pmj: Done.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -1196,4 +1263,108 @@ - (void) testIneligibleFilesystemType
...
+}
+
+- (void) testOfflineRootAllowsRegisteredProcessAccessToEmptyFile
+{
+    testFileVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
+    SetPrjFSFileXattrData(testFileVnode);
+
+    ActiveProvider_Disconnect(self->dummyRepoHandle, &self->dummyClient);
+
+    XCTAssertTrue(VirtualizationRoots_AddOfflineIOProcess(501));
```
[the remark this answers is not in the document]
pmj: This is now called `RunningMockProcessPID`.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -509,7 +527,17 @@ KEXT_STATIC int HandleVnodeOperation(
             else if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
                 // Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
```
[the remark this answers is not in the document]
pmj: Done.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -549,8 +577,25 @@ KEXT_STATIC int HandleVnodeOperation(
...
+                    KAUTH_VNODE_APPEND_DATA);
+                if (!TryGetVirtualizationRoot(
+                        &perfTracer,
+                        context,
+                        currentVnode,
+                        pid,
+                        // Prevent system services from hydrating files as this tends to cause deadlocks with the kauth listeners for Antivirus software
+                        CallbackPolicy_UserInitiatedOnly,
+                        // Prevent write access to empty files in offline roots. For now allow reads, and always allow the user to delete files.
+                        isWriteOperation,
```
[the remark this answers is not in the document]
pmj: Good catch, I've added that now and also added test coverage for that case.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: I've sorted that out now, with a version-dependent flow.

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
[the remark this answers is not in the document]
pmj: Yep, that sounds spot on. I've changed the xattr logic and also added some comments explaining what's being tested and how.

pmj:
I think I've now addressed all the feedback, please re-check!

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -653,3 +661,71 @@ VirtualizationRootHandle ActiveProvider_FindForPath(const char* _Nonnull path)
...
+    RWLock_ReleaseShared(s_virtualizationRootsLock);
+
+    return result;
+}
+
+bool VirtualizationRoots_AddOfflineIOProcess(pid_t pid)
+{
+    bool success = false;
+    
+    // Note: lock may temporarily be dropped within
```
wilbaker: This comment is no longer accurate and can be removed

PR #1212, opened 2019-05-28
Fix Mac installer issues

### LINKED ISSUE #1210: macOS installer: automount fails
The macOS installer fails to re-mount repositories. The following error is contained in the service logs:

```"ErrorMessage":"LaunchProcess ERROR: Could not launch /bin/launchctl"}```

### PR DESCRIPTION
**Pull request by alameenshah**
#### Fix couple installer issues
1. Mac installer relies on GVFS.Service to do the auto-mount of registered repos. When GVFS.Service gets launched (post installation) it (Service) calls gvfs mount in the background (without user interaction). In this case if the kext is not loaded, then mount will fail. Updating the installer to load kext would help in scenario. Also updated postflight script to log output of kextstat command that is used to determine if macOS autoloaded kext or not. A non empty output means kext got autoloaded or there was an error running kextstat.

2. Don't install copy of "VFS For Git.app" in /usr/loca/vfsforgit directory.

Fixes #1210

### DIFF
--- a/GVFS/GVFS.Installer.Mac/scripts/postinstall
+++ b/GVFS/GVFS.Installer.Mac/scripts/postinstall
@@ -1,4 +1,29 @@
 #!/bin/bash
+
+# Load PrjFSKext if it is not loaded already
+# PrjFSKext is an IOKit kext and should get automatically loaded 
+# by macOS after install. But the system does not seem to auto-load it
+# consistently after every install. The code below checks if it has
+# been auto-loaded. If not, it will attempt to load it.
+# The BundleID of the Kext(ProjFS.Mac/PrjFS.xcodeproj) is defined in
+# ProjFS.Mac/PrjFS.xcodeproj project.
+KEXTBUNDLEID="org.vfsforgit.PrjFSKext"
+KEXTPATH="/Library/Extensions/PrjFSKext.kext"
+kextstatCmd="/usr/sbin/kextstat -l -b $KEXTBUNDLEID"
+kextstatOutput=$(eval $kextstatCmd)
+if [[ -z "${kextstatOutput// }" ]]; then
+# load the kext using kextload command. Installer is run as the
+# admin user, so we already have required privileges. In case load
+# still fails, then exit 1. This will cause the installer to display
+# an installation failed error message.
+    loadCmd="/sbin/kextload \"$KEXTPATH\""
+    echo $loadCmd
+    eval $loadCmd || exit 1
+else
+    echo "$kextstatCmd returned non-zero output. This might possibly indicate an error."
+    echo "$kextstatOutput"
+fi
+
 loadCmd="sudo launchctl load -w /Library/LaunchDaemons/org.vfsforgit.prjfs.PrjFSKextLogDaemon.plist"
 echo "Loading PrjFSKextLogDaemon: '$loadCmd'..."
 eval $loadCmd || exit 1

### CODE REVIEW

File: GVFS/GVFS.Installer.Mac/scripts/postinstall [hunk 1]
```diff
@@ -1,4 +1,15 @@
 #!/bin/bash
+
+KEXTBUNDLEID="org.vfsforgit.PrjFSKext"
+KEXTPATH="/Library/Extensions/PrjFSKext.kext"
+isKextLoadedCmd="/usr/sbin/kextstat -l -b $KEXTBUNDLEID | wc -l"
+isKextLoaded=$(eval $isKextLoadedCmd)
```
jamill: Might be good to include a comment about when we might expect the kext to be loaded (or not).

GVFS/GVFS.Installer.Mac/scripts/postinstall
```diff
@@ -1,4 +1,15 @@
 #!/bin/bash
+
+KEXTBUNDLEID="org.vfsforgit.PrjFSKext"
```
jamill: Maybe include a comment about where this ID comes from?

GVFS/GVFS.Installer.Mac/scripts/postinstall
```diff
@@ -1,4 +1,15 @@
...
+KEXTPATH="/Library/Extensions/PrjFSKext.kext"
+isKextLoadedCmd="/usr/sbin/kextstat -l -b $KEXTBUNDLEID | wc -l"
+isKextLoaded=$(eval $isKextLoadedCmd)
+if [ "$isKextLoaded" -eq 0 ]; then
```
jamill: Are there any error conditions we might have to account for here?

GVFS/GVFS.Installer.Mac/scripts/postinstall
```diff
@@ -1,4 +1,15 @@
...
+    loadCmd="/sbin/kextload \"$KEXTPATH\""
+    echo $loadCmd
```
jamill: In other places, it looks like we include a little more context around the operation, rather than just the command. Should we do that here as well?

GVFS/GVFS.Installer.Mac/scripts/postinstall
alameenshah: Added a short description in the beginning.

GVFS/GVFS.Installer.Mac/scripts/postinstall
alameenshah: Added a short comment with info on what we are trying to do.

GVFS/GVFS.Installer.Mac/scripts/postinstall
alameenshah: `kextstat` can return console output which can indicate couple things
- macOS successfully auto-loaded our kext after install. 
- there was a `kextstat` error.

Updated `postflight` to log`kextstat` output, in case it was a `kextstat` error.

GVFS/GVFS.Installer.Mac/scripts/postinstall
```diff
@@ -1,4 +1,29 @@
...
+if [[ -z "${kextstatOutput// }" ]]; then
+# load the kext using kextload command. Installer is run as the
+# admin user, so we already have required privileges. In case load
+# still fails, then exit 1. This will cause the installer to display
+# an installation failed error message.
+    loadCmd="/sbin/kextload \"$KEXTPATH\""
+    echo $loadCmd
+    eval $loadCmd || exit 1
+else
+    echo "$kextstatCmd returned non-zero output. This might possibly indicate an error."
```
jrbriggs: When wouldn't this be an error? What action do we expect a user to take here? This would be a very frustrating error message to see because I would walk away feeling helpless.

GVFS/GVFS.Installer.Mac/scripts/postinstall
```diff
@@ -1,4 +1,29 @@
 #!/bin/bash
+
+# Load PrjFSKext if it is not loaded already
+# PrjFSKext is an IOKit kext and should get automatically loaded 
+# by macOS after install. But the system does not seem to auto-load it
+# consistently after every install. The code below checks if it has
+# gotten auto-loaded. If not load it.
```
jrbriggs:
```suggestion
# been auto-loaded. If not, it will attempt to load it.
```

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -67,7 +67,6 @@ echo 'Copying native binaries to Publish directory...'
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.ReadObjectHook $VFS_PUBLISHDIR || exit 1
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.VirtualFileSystemHook $VFS_PUBLISHDIR || exit 1
 cp $VFS_OUTPUTDIR/GVFS.Native.Mac/Build/Products/$CONFIGURATION/GVFS.PostIndexChangedHook $VFS_PUBLISHDIR || exit 1
-cp -Rf $VFS_OUTPUTDIR/GVFS.Notifications/VFSForGit.Mac/Build/Products/$CONFIGURATION/VFS\ For\ Git.app $VFS_PUBLISHDIR || exit 1
```
jrbriggs: Why was this line removed?

GVFS/GVFS.Installer.Mac/scripts/postinstall
alameenshah: >> When wouldn't this be an error?

Two scenarios here -
1. If macOS successfully auto-loaded the kext before this script(`postinstall`), then this will not be an error. This would be the best case scenario. User need to take no action here. The output of `echo` goes to the installer log file only, NO install error will get displayed to the user. The installer log would be useful to us developers (in case of issues like - installation succeeded but still `gvfs mount` command prompts for `kextload`).
2. `kextstat` itself fails with error output. I haven't seen this happen, but `man kextstat` does say kextstat can possibly exit with error. In this case, again, user will not see any error in installer UI. `gvfs mount` will prompt user to manually do a `kextload`. `GVFS.Service` auto-mount will fail, with error messaging in `Service` logs.

>> What action do we expect a user to take here? 

None. The two lines of logs are there to help us developers debug kext load issue (in the 2nd case above) during install. 

>> This would be a very frustrating error message to see because I would walk away feeling helpless.

Agree. Installer displays no error. The two lines go to installer log only.

Scripts/Mac/BuildGVFSForMac.sh
alameenshah: We were installing `VFS For Git.app` redundantly in two different directories - 1. `/Library/Application Support/VFS For Git` and 2. `/usr/local/vfsforgit`. This is not required and seems to cause weird installer issues during testing. I've already removed this in pending [PR](https://github.com/microsoft/VFSForGit/pull/1202). Just wanted to see if it helps @jamill with the weird `VFS For Git.app` permissions issue that he is seeing.

PR #1215, opened 2019-05-30
Remove Placeholder after calling PreDelete

### LINKED ISSUE #951: Mac: File renames should convert files to full
See related #515 for folder rename scenario, and #526 for related scenario to detect if PreDelete is for a rename.

When processing PreDelete vnode operations, VFS4G is currently hydrating files (because they might be part of a rename that will move a file outside of the virtualization root).

In addition to hydrating the files, VFS4G should also remove the placeholder xattr (and possibly flag(s)?) so that the file is treated as a full file.

jeschu1:
- PreDelete will perform a convert to Full
- This requires we hydrate the file first, hence the call order changes in KauthHandler

resolves #951

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -265,35 +265,6 @@ KEXT_STATIC int HandleVnodeOperation(
     isDeleteAction = ActionBitIsSet(action, KAUTH_VNODE_DELETE);
     isDirectory = vnode_isdir(currentVnode);
     
-    if (isDeleteAction)
-    {
-        // Allow any user to delete individual files, as this generally doesn't cause nested kauth callbacks.
-        ProviderCallbackPolicy callbackPolicy = isDirectory ? CallbackPolicy_UserInitiatedOnly : CallbackPolicy_AllowAny;
-        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, callbackPolicy, &root, &vnodeFsidInode, &kauthResult, kauthError))
-        {
-            goto CleanupAndReturn;
-        }
-        
-        PerfSample preDeleteSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreDelete);
-        
-        if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
-                root,
-                isDirectory ?
-                    MessageType_KtoU_NotifyDirectoryPreDelete :
-                    MessageType_KtoU_NotifyFilePreDelete,
-                currentVnode,
-                vnodeFsidInode,
-                nullptr, // path not needed, use fsid/inode
-                nullptr, // source path N/A
-                pid,
-                procname,
-                &kauthResult,
-                kauthError))
-        {
-            goto CleanupAndReturn;
-        }
-    }
-    
     if (isDirectory)
     {
         if (ActionBitIsSet(
@@ -432,6 +403,36 @@ KEXT_STATIC int HandleVnodeOperation(
             }
         }
     }
+    
+    if (isDeleteAction)
+    {
+        // Allow any user to delete individual files, as this generally doesn't cause nested kauth callbacks.
+        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_AllowAny, &root, &vnodeFsidInode, &kauthResult, kauthError))
+        {
+            goto CleanupAndReturn;
+        }
+                
+        PerfSample preDeleteSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreDelete);
+        
+        // Predeletes must be sent after hydration since they may convert the file to full
+        if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                root,
+                isDirectory ?
+                MessageType_KtoU_NotifyDirectoryPreDelete :
+                MessageType_KtoU_NotifyFilePreDelete,
+                currentVnode,
+                vnodeFsidInode,
+                nullptr, // path not needed, use fsid/inode
+                nullptr, // source path N/A
+                pid,
+                procname,
+                &kauthResult,
+                kauthError))
+        {
+            goto CleanupAndReturn;
+        }
+    }
+
     
 CleanupAndReturn:
     if (putVnodeWhenDone)
--- a/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
@@ -339,7 +339,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
             _,
-            MessageType_KtoU_NotifyFilePreDelete,
+            MessageType_KtoU_HydrateFile,
             testFileVnode.get(),
             _,
             _,
@@ -350,8 +350,8 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
             nullptr),
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
-           _,
-            MessageType_KtoU_HydrateFile,
+            _,
+            MessageType_KtoU_NotifyFilePreDelete,
             testFileVnode.get(),
             _,
             _,
@@ -378,7 +378,7 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
             _,
-            MessageType_KtoU_NotifyDirectoryPreDelete,
+            MessageType_KtoU_RecursivelyEnumerateDirectory,
             testDirVnode.get(),
             _,
             _,
@@ -389,8 +389,8 @@ static void SetPrjFSFileXattrData(const shared_ptr<vnode>& vnode)
             nullptr),
         ProviderMessaging_TrySendRequestAndWaitForResponse,
         make_tuple(
-           _,
-            MessageType_KtoU_RecursivelyEnumerateDirectory,
+            _,
+            MessageType_KtoU_NotifyDirectoryPreDelete,
             testDirVnode.get(),
             _,
             _,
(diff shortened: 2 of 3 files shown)


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -433,6 +404,36 @@ KEXT_STATIC int HandleVnodeOperation(
         }
     }
     
+    if (isDeleteAction)
+    {
+        // Allow any user to delete individual files, as this generally doesn't cause nested kauth callbacks.
+        ProviderCallbackPolicy callbackPolicy = isDirectory ? CallbackPolicy_UserInitiatedOnly : CallbackPolicy_AllowAny;
+        if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, callbackPolicy, &root, &vnodeFsidInode, &kauthResult, kauthError))
```
wilbaker: It looks like we'll have already looked up the root by now for deletes, can we remove this call to `TryGetVirtualizationRoot`?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -433,6 +404,36 @@ KEXT_STATIC int HandleVnodeOperation(
...
+        {
+            goto CleanupAndReturn;
+        }
+                
+        PerfSample preDeleteSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreDelete);
```
wilbaker: Could you add a comment explaining why the PreDeletes are sent after the hydrate messages?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1099,7 +1099,9 @@ static PrjFS_Result HandleFileNotification(
         notificationType,
         nullptr /* destinationRelativePath */);
     
-    if (result == PrjFS_Result_Success && placeholderFile && PrjFS_NotificationType_PreConvertToFull == notificationType)
+    if (result == PrjFS_Result_Success &&
+        placeholderFile &&
+        (PrjFS_NotificationType_PreConvertToFull == notificationType || PrjFS_NotificationType_PreDelete == notificationType))
```
wilbaker: Maybe add a comment here saying that we convert to full for PreDelete because until we have WILL_RENAME support the PreDelete could be for a rename?

ProjFS.Mac/PrjFSKextTests/HandleOperationTests.mm
```diff
@@ -669,6 +669,20 @@ - (void) testDeleteDirectoryWithDisappearingVirtualizationRoot {
     XCTAssertTrue(MockCalls::CallCount(ProviderMessaging_TrySendRequestAndWaitForResponse) == 1);
 }
 
+- (void) testDeleteDirectoryWhenFirstRequestFails {
```
wilbaker: Could you add some comments to make it a bit clearer what's being validated?

I think we're validating that when processing `KAUTH_VNODE_DELETE` the kext will not send additional messages if the first `TrySendRequestAndWaitForResponse` fails, but a comment here would be good.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
jeschu1: I can, the old way also did 2 looksup, but it's fair to say that was never a hard requirement.

wilbaker: Yeah I noticed that too.  I could not think of a reason that was required though, maybe it was just an oversight?

jeschu1: I *think* and it would be great to get confirmation on this, since the virtualization roots aren't on lock they could change between calls.  That said nothing is under lock so it really doesn't guarantee you anything just that you checked again a few lines down.

I think it was probably just a code copy oversight.  It's also easier now to tell we have the virtualization root based on where the code is now.

jeschu1: There actually is a notable change worth calling out.  For the Predelete we were looking up the root with "CallbackPolicy_AllowAny". 

However the following check for Hydrate would have been called with "CallbackPolicy_UserInitiatedOnly"

@pmj can you elaborate on the desired behavior here?  For Deletes I would expect both the Hydrate and PreDelete to be processed (or neither).

pmj: Regarding calling `TryGetVirtualizationRoot` multiple times - sorting this out is *very* high on my list of things to do as part of #900, I'm hoping we can do that next after #182 and #526 are dealt with.

Regarding predelete vs hydrate policy: for directories it looks like it's already consistent. For files you're right, it doesn't really make sense to have a discrepancy in policy here. Note that from #1193 on we have a distinction between delete-because-of-rename and other delete operations, and we won't be hydrating in the non-rename case anyway. For now, perhaps change everything to `CallbackPolicy_UserInitiatedOnly` and put a TODO: comment anywhere we might want to have an `AllowAny` policy in future.

To be honest I think the whole function needs a complete overhaul as I've said in #900. My current favoured approach would be not to interleave decision logic with actually doing anything destructive, i.e. sending the messages. So collect the messages we want to send, and once we've made it all the way through to decide what we want to do, actually go and send them. Separating those 2 aspects will also make it easier to only gather necessary information (such as the virtualisation root handle) once rather than keep going back for the same thing over and over.

jeschu1: Thanks @pmj, that does make sense!

Do we think we'll use `CallbackPolicy_AllowOnly` in the future?  It looks like it was only on deletes.  Should system processes be deleting files?

pmj: > Do we think we'll use `CallbackPolicy_AllowOnly` in the future? It looks like it was only on deletes. Should system processes be deleting files?

In some cases, user-initiated file operations in Finder will be farmed out to a root-owned helper tool. This has `launchd` as its parent process, so `CallbackPolicy_UserInitiatedOnly` would block it, for example.

wilbaker: **TLDR:** I was wrong, we will not always have looked up the root by now.  When deleting a file we only look up the root if we're hydrating the file first.

----

**Details**

Taking a closer look, I'm not sure the old code was properly handling `UserInitiatedOnly` vs `AllowAny` in the first place for *directories*.

The old logic did not bail out after sending the PreDelete unless either:
- `TryGetVirtualizationRoot` failed
- `ProviderMessaging_TrySendRequestAndWaitForResponse` returned false

As best I can tell, this means that if a system process was trying to delete a *directory*, the following would occur:
1. PreDelete notification is sent to the provider
2. The root lookup here fails due to the policy restriction:

```
// Prevent system services from expanding directories as part of enumeration as this tends to cause deadlocks with the kauth listeners for Antivirus software
if (!TryGetVirtualizationRoot(&perfTracer, context, currentVnode, pid, CallbackPolicy_UserInitiatedOnly, &root, &vnodeFsidInode, &kauthResult, kauthError))	                
{
    goto CleanupAndReturn;
}
```

End result: The delete is denied, and so there was no point in sending a PreDelete message in the first place.

I think that leaves us with the file scenario, it which case I was wrong.   We might have to look up the root again because we would have only looked up the root if the file was empty and needed to be hydrated.

So I think we could simplify this code to always use `CallbackPolicy_AllowAny` (we would have bailed out early if something required ` CallbackPolicy_UserInitiatedOnly` and that lookup failed).  Additionally (and ideally) we would *only* make this call if we didn't already look up the root somewhere above.

@pmj @jeschu1 does this sound right to you?

jeschu1: **Looking at the old code we did:**
`        ProviderCallbackPolicy callbackPolicy = isDirectory ? CallbackPolicy_UserInitiatedOnly : CallbackPolicy_AllowAny;`

**Previous Behavior was**
System initiated directory `DELETE` sends no messages
System initiated file  `DELETE` sends `MessageType_KtoU_NotifyFilePreDelete` message but not  `MessageType_KtoU_HydrateFile`.

**The way the code is structured now**
System initiated directory `DELETE` sends no messages
System initiated file  `DELETE` on a hydrated file, we will always send `MessageType_KtoU_NotifyFilePreDelete`.
System initiated file  `DELETE` on an unhydrated file, sends no messages

Since we now remove the placeholder tag on `MessageType_KtoU_NotifyFilePreDelete` the file must be hydrated before calling it.

@pmj is the desired functionality that we allow  `DELETE` to process regardless of who initiated them?  If so I can change make those changes accordingly so we'll both hydrate and call predelete.  Would `DELETE` be the only command we'd want to ignore owner for at this time?

pmj: The problem with system processes is primarily hydration/expansion, because that causes file writes by the provider which in turn causes callbacks in e.g. an AV kext's kauth handler, which causes activity in the AV daemon, etc.

So I think we basically want `CallbackPolicy_AllowAny` for any code path that doesn't cause hydration/expansion.

jeschu1: I think the simplest approach for now is to do the `CallbackPolicy_AllowAny` as suggested.  Thanks @pmj and @wilbaker!  If we reach this point in code the file must be hydrated so it's safe to make this change.

@pmj a new message type is definitely an option, but considering we're reworking renames I wouldn't do the change in this PR as it's a bit more invasive.

I'd like to discuss callbackPolicy more offline, but definitely not a blocker or change for this PR.

jeschu1:
> The only comment I have is that it might be good to have a functional test for this? Or perhaps we already do but it's ignored on Mac so far?

My Rename PR has a bunch of tests that will hit this code path.  We also have good Delete tests here.  The one bit that hard to functional test is running as a different user (at least on windows).  Generally that becomes tricky for security reasons.

pmj:
> My Rename PR has a bunch of tests that will hit this code path. We also have good Delete tests here. The one bit that hard to functional test is running as a different user (at least on windows). Generally that becomes tricky for security reasons.

Yeah, I'm not suggesting testing *all* code paths via functional tests here, just making sure renaming an empty file now works correctly. (presumably it previously did not - otherwise what's this PR for?)

jeschu1:
> Yeah, I'm not suggesting testing all code paths via functional tests here, just making sure renaming an empty file now works correctly. (presumably it previously did not - otherwise what's this PR for?)

Great question!  The functionality worked before, the issue was we may try to hydrate twice (once for predelete and once for any other hydrating action).  Since we remove the flag now on predelete, it will only hydrate once.  It's a performance optimization.

wilbaker:
@pmj @jeschu1

> The functionality worked before, the issue was we may try to hydrate twice (once for predelete and once for any other hydrating action). Since we remove the flag now on predelete, it will only hydrate once. It's a performance optimization.

I don't think we would have hydrated twice before as the `FileFlags_IsEmpty` flag would have been cleared after the first hydration.  

This PR gets us the first half of what's required for #606, and helps tighten up correctness a bit in rename scenarios.  The update placeholder helper functions should fail on a file that's been renamed, and we get that "failure after rename" behavior by removing the placeholder xattr.

pmj:
Thanks for the explanations everyone. Looks like I’d misunderstood the purpose of the changes.

PR #1223, opened 2019-05-30
Design Document for the Analytics Tool

Halterer:
Here is the draft of the design document for the analytics tool project. Hopefully everyone can give me some feed back and I'll continue to update it throughout the day based on feedback. If you view it in the markdown view rather than source mode you can also see the diagram at the bottom.


Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+This document details the design of and plan for the **VFS4G analytics tool**. The tool aims to help users identify reasons why their enlistment may be slowing down and the `gvfs` commands running slowly. By collecting and reporting statistics, the power to uncover and hopefully fix problems as they arrive moves to the user, empowering them to use the software to its fullest potential.
+
+## Problem
+
+When users unintentionally hydrate too many files within their enlistment, commands like checkout and commit increase their running time by several orders of magnitude. This causes unacceptable losses in performance and greatly hinders users as they go about their daily development cycle. The primary metric of the overhydration is the number of placeholder files which have been pulled from the remote into the local system. Slowdowns may also stem from having many modified files or directories which increases the amount of work that git does under the hood. Additionally, if a local enlistment has been in use for a long time, the number of loose object may eventually reach the order of several thousand in which case that also contributes to the loss in performance.
```
wilbaker: > When users unintentionally hydrate too many files within their enlistment, commands like checkout and commit increase their running time by several orders of magnitude.

A small clarification on this point.  The performance issues that we've seen from users are usually caused by one (or both) of the following:

- Having too many placeholders on disk
- Having too many modified/full files (i.e. the files in "modified paths")

Hydrating files (i.e. populating empty placeholders with their contents) doesn't impact the steps VFS4G will need to take when processing a command like `git checkout`:

1. Check if the placeholder file that's on disk needs to be updated/deleted
2. If the placeholder needs to be updated, update it.  The helper function that VFS4G calls does this by:
    1. Deleting the old placeholder
    2. Writing a new placeholder
3. Store the updated information about the placeholder in the placeholder list

You might hear people referring to the problem of "over hydration" but from a technical standpoint it's not the hydrating specifically that's the problem.  It's having too many files/folders in the working directory.

Halterer: That makes sense. I rewrote for clarity of these points both for others but also to keep myself in track:

"Over the lifetime of an enlistment, sometimes a user's working directory fills up with too many entries. Generally this means having too many placeholders on their disk (have been read and pulled from remote) or too many files which are fully on disk (have been modified). This causes unacceptable losses in performance and greatly hinders users as they go about their daily development cycle. Additionally, if a local enlistment has been in use for a long time, the number of loose object may eventually reach the order of several thousand in which case that also contributes to the loss in performance."

wilbaker: The re-write looks good.  I think you could simplify this:

> too many files which are fully on disk (have been modified)

To:
> too many modified files

Unfortunately some of the terminology used in VFS4G and ProjFS is not super straightforward:
- "Full files" -> Regular/plain NTFS files 
- "Hydrated placeholder" or "placeholder with contents" -> File that has all of its contents on disk, but still has ProjFS placeholder metadata (i.e. reparse data) attached to it.

And so both placeholders and modified files can be "fully on disk" as far as their contents are concerned.

Halterer: Got it! Good catch, I didn't know that but I'm glad that I do now. I'm pretty certain I made several more instances of that mistake so I'll go through and clarify there too

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Provided Statistics
+
+### Safe to access
+
+- **Modified paths** - The number of files which have been edited and have been handed over to git
+- **Git index** - The differences between modified paths and the index tracked by git
+- **Placeholders** - A total number of the placeholders being tracked in VFS4G and a list of the placeholders in the system (Note that this contains some upper level directories and thus not all of their children are included in the count without enumeration)
```
wilbaker: > Note that this contains some upper level directories and thus not all of their children are included in the count without enumeration

Two small things to be aware of:

- The placeholders collection will always contain *all* placeholders, even those files whose parent folders are in the placeholder list.

- The "Modified paths" can also contain top level directories, and it has the behavior you mentioned here.  If a directory is in modified paths its children will not be included.

Halterer: I updated that for correctness, thank you!

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Problem
+
+When users unintentionally hydrate too many files within their enlistment, commands like checkout and commit increase their running time by several orders of magnitude. This causes unacceptable losses in performance and greatly hinders users as they go about their daily development cycle. The primary metric of the overhydration is the number of placeholder files which have been pulled from the remote into the local system. Slowdowns may also stem from having many modified files or directories which increases the amount of work that git does under the hood. Additionally, if a local enlistment has been in use for a long time, the number of loose object may eventually reach the order of several thousand in which case that also contributes to the loss in performance.
+
+## Provided Statistics
+
+### Safe to access
+
+- **Modified paths** - The number of files which have been edited and have been handed over to git
```
kewillford: Like William said this file is a list of file and folder paths that git will pay attention to and a folder path could cover many files.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Problem
+
+When users unintentionally hydrate too many files within their enlistment, commands like checkout and commit increase their running time by several orders of magnitude. This causes unacceptable losses in performance and greatly hinders users as they go about their daily development cycle. The primary metric of the overhydration is the number of placeholder files which have been pulled from the remote into the local system. Slowdowns may also stem from having many modified files or directories which increases the amount of work that git does under the hood. Additionally, if a local enlistment has been in use for a long time, the number of loose object may eventually reach the order of several thousand in which case that also contributes to the loss in performance.
+
+## Provided Statistics
+
+### Safe to access
```
kewillford: What does this mean?  Does `access` mean only to read the information from?

Also is this a list of the statistics that you are going to be tracking or the files that will be read?

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **Modified paths** - The number of files which have been edited and have been handed over to git
+- **Git index** - The differences between modified paths and the index tracked by git
```
kewillford: Not sure what this means or what you will be using the index for.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- **Placeholders** - A total number of the placeholders being tracked in VFS4G and a list of the placeholders in the system (Note that this contains some upper level directories and thus not all of their children are included in the count without enumeration)
+- **Troublesome programs** - Similar to what is already being logged in the heartbeat, look at which services or programs are responsible for hydrating large amounts of files in the enlistment
+
+### Potentially detrimental to access on a mounted enlistment
+
+- **Placeholder details** - Crawling the file system to enumerate directories of placeholders may unintentionally further hydrate files (IE downloading file sizes), but doing so would allow for greater details such as the percent hydration of folders, the total number of hydrated files, the type of files which are most hydrated, etc.
+- **File system** - Provide enumerated details for files in the placeholder list or the modified path list
```
kewillford: Anything in the modified paths list should be okay to get details on since they has been written to and there will not be any events happening inside of VFSForGit on them.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Background Information
+
+- **Modified paths** is the list of files which VFS4G has completely pulled from the remote and whose contents are on disk in their entirety. These files are handed to git via a hook when any git command requiring the list of changed files is called
```
kewillford: This is a list of file and folder **paths** and they aren't necessarily pulled from the remote.  If I create a new file, it will be in the modified paths but will not have come from any remote.  

>These files are handed to git

These paths are handed to git

>when any git command requiring the list of changed files is called

when a git command reads its index to populate its list of files that it is tracking.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- **Placeholders** are files which have been opened by the local system with a read tag but without a write tag. They are objects tracked by VFS4G since git should not have to internally scan them since it is guaranteed that they have not been modified and accordingly do not have to be rehashed for differences. Their contents are on disk, but control over them has not been fully ceded over to the system. They decrease performance because VFS4G has to track all of them individually, and this can cause certain `gvfs` commands to be slow
```
kewillford: >control over them has not been fully ceded over to the system.

over to git?

>this can cause certain `gvfs` commands to be slow

`git` commands that need to update the projection

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- **File system** - Provide enumerated details for files in the placeholder list or the modified path list
+
+## Background Information
+
+- **Modified paths** is the list of files which VFS4G has completely pulled from the remote and whose contents are on disk in their entirety. These files are handed to git via a hook when any git command requiring the list of changed files is called
+- **Placeholders** are files which have been opened by the local system with a read tag but without a write tag. They are objects tracked by VFS4G since git should not have to internally scan them since it is guaranteed that they have not been modified and accordingly do not have to be rehashed for differences. Their contents are on disk, but control over them has not been entirely ceded over to the system. They decrease performance because VFS4G has to track all of them individually, and this can cause certain `gvfs` commands to be slow
```
kewillford: >control over them has not been entirely ceded over to the system.

What "system"?  git?

>certain `gvfs` commands to be slow

certain `git` command that need to update the projection to be slow

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+### Safe to access
+
+- **Modified paths** - The number of files which have been edited and have been handed over to git
+- **Git index** - The differences between modified paths and the index tracked by git
+- **Placeholders** - A total number of the placeholders being tracked in VFS4G and a list of the placeholders in the system (Note that this contains some upper level directories and thus not all of their children are included in the count without enumeration)
+- **Troublesome programs** - Similar to what is already being logged in the heartbeat, look at which services or programs are responsible for hydrating large amounts of files in the enlistment
+
+### Potentially detrimental to access on a mounted enlistment
+
+- **Placeholder details** - Crawling the file system to enumerate directories of placeholders may unintentionally further hydrate files (IE downloading file sizes), but doing so would allow for greater details such as the percent hydration of folders, the total number of hydrated files, the type of files which are most hydrated, etc.
```
derrickstolee: You should be able to avoid hydrating the filesystem if you use `git ls-tree` to discover the folder layout instead of using the filesystem.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- **File system** - Provide enumerated details for files in the placeholder list or the modified path list
```
derrickstolee: I'm not sure what "enumerated details" means. Could you give a concrete example?

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+
+- **Modified paths** is the list of files which VFS4G has completely pulled from the remote and whose contents are on disk in their entirety. These files are handed to git via a hook when any git command requiring the list of changed files is called
+- **Placeholders** are files which have been opened by the local system with a read tag but without a write tag. They are objects tracked by VFS4G since git should not have to internally scan them since it is guaranteed that they have not been modified and accordingly do not have to be rehashed for differences. Their contents are on disk, but control over them has not been entirely ceded over to the system. They decrease performance because VFS4G has to track all of them individually, and this can cause certain `gvfs` commands to be slow
+- The **Git index** is gits internal list of files which have been modified (tracked or untracked) which is what is used for commit commands to keep score of everything going on inside of the enlistment. It is useful to this project by means of comparing it to the list of **hydrated files** to deduce if those files should truly be hydrated or if it may have occurred incidentally.
```
kewillford: `gits` -> `git's`

The index is only the files that are being tracked by git.  git untracked files are not something that are in the index, so I would have this say something like
```
is git's internal list of files which it is tracking for creating  the next `commit`.  
```

I'm also not sure what you are going to use the index for and what comparisons you will be doing.  You may want to talk a little about the skip-worktree bit and how that will be used in this comparison since those are the files that should not be in the modified paths list but could be a placeholder but might not be in either list.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- **Troublesome programs** - Similar to what is already being logged in the heartbeat, look at which services or programs are responsible for hydrating large amounts of files in the enlistment
+
+### Potentially detrimental to access on a mounted enlistment
+
+- **Placeholder details** - Crawling the file system to enumerate directories of placeholders may unintentionally further hydrate files (IE downloading file sizes), but doing so would allow for greater details such as the percent hydration of folders, the total number of hydrated files, the type of files which are most hydrated, etc.
+- **File system** - Provide enumerated details for files in the placeholder list or the modified path list
+
+## Background Information
+
+- **Modified paths** is the list of files which VFS4G has completely pulled from the remote and whose contents are on disk in their entirety. These files are handed to git via a hook when any git command requiring the list of changed files is called
```
wilbaker: A few more details on modified paths.

Modified paths contains:
- Files and directories that have been deleted
- Newly created files/directories (created by the user), including the destination path when a file is renamed.
- File that have been modified by the user (more specifically, opened for write)

Fully hydrated placeholders are not included in modified paths, even though their contents are entirely on disk.

Documents/AnalyticsToolDesignDocument.md
Halterer: I meant along the lines of anything that requires crawling the disk for information about different files. One of the things I've aimed to learn with this review is the different statistics that people think are critical to have in an analytics tool so I was using this as more of an umbrella term for anything like that. If there is no real reason to do this, especially given your above comment, then I can remove it. It sounds like we can get just about everything we need from `git ls-tree -r HEAD` so then this whole section about detrimental/risky access can be nixed

Documents/AnalyticsToolDesignDocument.md
Halterer: I thought fully hydrated placeholders were what files became once they were opened with a write handle. So then that term is for still just placeholders that have some data associated with them from ProjFS?

Documents/AnalyticsToolDesignDocument.md
Halterer: Alright I removed that section completely and added in the recommended command to the modified paths entry where crawling the disk is potentially useful

Documents/AnalyticsToolDesignDocument.md
Halterer: This is a list of the statistics I plan to extract from the enlistment and provide to the user as a whole, and previously it included some of the assets I was going to need to get them as far as I could tell. It has since been updated, after I finish reading through the recent feedback I will recommit the updated document. I have since removed the "safe to access" tag which I had initially meant to contain the statistics where if we accessed them there was no risk of accidentally hydrating files on disk farther than already done (like we talked about last night with downloading file sizes).

Documents/AnalyticsToolDesignDocument.md
Halterer: I added the following to the end of the git index bullet point, it might clear things up.

"This might be useful in determining what differences the local repository has to predict the complexity and potentially slowdowns of an upcoming commit."

It is also possible however that this is information which could easily be accessed elsewhere or maybe even that everything within the modified paths will always align perfectly with the index (which is suspect might be the case)

Documents/AnalyticsToolDesignDocument.md
Halterer: This has since been removed and updated 😄

Documents/AnalyticsToolDesignDocument.md
Halterer: Thank you for the clarification, I missed that. Updated to correctly describe (hopefully) the system 👍

Documents/AnalyticsToolDesignDocument.md
wilbaker: > I thought fully hydrated placeholders were what files became once they were opened with a write handle.

Fully hydrated placeholders are what empty placeholder files become when they're read.

The exact behavior of files is slightly different on Window and Mac:

**Windows**

- Virtual File (*File owned/updated by VFS4G*): File does not exist on disk but is injected into enumeration results by VFS4G.
- Empty Placeholder (*File owned/updated by VFS4G*): File has been opened without write access.  No data has been read from the file.
- Hydrated Placeholder (*File owned/updated by VFS4G*): At least one byte has been read from the file.  If an empty placeholder file is opened for write it will be hydrated as well, but it will immediately transition to the "full file" state.
- Full File (*File owned/updated by Git*): A regular/plain NTFS file.  There is no special ProjFS data on the file.  Newly created files are "full files" as well as placeholder files that were opened for write.
- Tombstone (*File owned/updated by Git*): A special hidden ProjFS file.  Tombstone files suppress virtual files from being displayed during enumeration and they are created when a user deletes a file that VFS4G is currently projecting.

**Mac**
- Empty Placeholder (*File owned/updated by VFS4G*): Created when expanding a directory (Mac specific).  Files remain empty until something attempts to read/write data or attributes.
- Hydrated Placeholder (*File owned/updated by VFS4G*): At least one byte has been read from the file *or* the attributes of the file have been read.  If an empty placeholder file is opened for write it will be hydrated as well, but it will immediately transition to the "full file" state.
- Full File (*File owned/updated by Git*): A regular/plain file.  There is no special ProjFS data on the file.  Newly created files are "full files" as well as placeholder files that were opened for write.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
+# VFS For Git: Analytics Tooling Design Document
+
+This document details the design of and plan for the **VFS4G analytics tool**. The tool aims to help users identify reasons why their enlistment may be slowing down and the `gvfs` commands running slowly. By collecting and reporting statistics, the power to uncover and hopefully fix problems as they arrive moves to the user, empowering them to use the software to its fullest potential.
```
wilbaker: > The tool aims to help users identify reasons why their enlistment may be slowing down and the `gvfs` commands running slowly.

- To confirm, do you mean `git` or `gvfs` commands here?
- Will the tool help identify which commands are running slowly, or why they might be running slowly?

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- The **file system** must be carefully considered in the case of working with a mounted repository. VFS4G intercepts file I/O calls at the kernel level to inject its own behavior (the projected file system and if needed a network call to pull down the file which the system asked for in the correct format). Because of this, enumerating directories can cause additional data to be pulled down from the remote by hydrating extra files or further hydrating those that already exist on disk. Accordingly, care must be taken when interacting with the file system lest the tool designed to help users increase enlistment heath end up causing more problems.
+
+## Components
+
+- **API** - The server component of the client-server relationship. Responsible for the logic and computation of statistical data, as well as the assembly and sending of a **json statistics object** over the named pipe to the client. It sits within the `InProgressMount` process and interacts with the VFS4G internals
```
derrickstolee: This **API** idea still worries me. Putting a new API in the named pipe layer still seems unnecessary (until we are forced to do it because we find a reason we can't access a resource from multiple processes). Instead, I'd start with the verb, and what work the verb needs to do in order to output a simple text-based message to a user summarizing what you found. Something like:

```
> gvfs statistics
Paths tracked by Git:   3,123,456 (100%)
Placeholders:      123,456 (4%)
Modified Paths:      1,234 (<1%)

Repo status: Healthy (no recommended action)

> (do stuff)

> gvfs statistics
Paths tracked by Git:   3,123,456 (100%)
Placeholders:      3,523,456 (111%)
Modified Paths:      500,000 (24%)

Repo status: Over-hydrated
Recommended Action: Run "gvfs dehydrate --confirm --no-status"
Take action? (y/n): y

Starting dehydration. All of your existing files will be backed up in C:\_git\os\dehydrate_backup\20190530_151645
WARNING: If you abort the dehydrate after this point, the repo may become corrupt

Unmounting...Succeeded
Authenticating...Succeeded
Backing up your files...Succeeded
Downloading git objects...Succeeded
Recreating git index...Succeeded
Mounting...Succeeded

Paths tracked by Git:   3,123,456 (100%)
Placeholders:      0 (0%)
Modified Paths:   0 (0%)
>
```
(Btw, having something like the input/output example above would be a good concrete example of something you could put in this document.)

You can take a look at `RepairVerb` for the concept of "repair jobs" which are different things that verb can do and report to the verb for details. Likely, you will want these different analyses to fit that sort of plug-in model.

Finally, you can still specify a JSON API layer, but as output from the _verb_, so your "additional visualization tools" could call the verb directly with a parameter (say, `--json`) and interpret the output nicely.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,48 @@
...
+- **Verb** - The `gvfs statistics` verb will be the component that handles the client side interaction with the API. When run, it will collect and store the statistics from the enlistment to either be displayed immediately or collected for trend purposes to then provide understanding in the future. This is also the point where data can be sent to telemetry for our internal analysis
+- **CLI Tool** - This tool will be built into the above verb and used to visualize and understand the data. It will simply output the analytics data to the command line and will also be used during the design period to test the API and verb
```
derrickstolee: I'm confused about the distinction between the **Verb** and the **CLI Tool**. A verb is specifically what executes when we run `gvfs <verbname>` in the command-line.

Documents/AnalyticsToolDesignDocument.md
Halterer: It's possible that the git index might not be a useful item in this tool. I was thinking along the lines of it containing the significant files which contribute to the bulk of work being done on any given git command (if my understanding serves me correctly) that it would contain useful information, especially when coupled with the other lists of files in the `.gvfs` directory. Do you think that the index won't end up being beneficial to the project?

Documents/AnalyticsToolDesignDocument.md
Halterer: I do mean git, thanks for catching that. I think tracking command run times might be a useful metric to capture so I'll go ahead and add it to the list of statistics planned for reporting

Documents/AnalyticsToolDesignDocument.md
Halterer: I'm envisioning the verb having overall functionality to do other things, I'm not sure if something like the verb is actually used when running on a timer to collect data (perhaps that is a different type of system in place like you had mentioned some yesterday). I envision the CLI tool (looking back, that's an ambiguous name) being an option for the verb which attempts to present the information to the user in a somewhat beautified manner. Perhaps the two are really the same thing and the default action that occurs when you run `gvfs statistics` is something like you showed above or the mock up I created and will push shortly. Then those two are really just one and the same which makes sense

Documents/AnalyticsToolDesignDocument.md
kewillford: I think it could be useful in telling how many files git is tracking an of those how many git is controlling/keeping up to date.  See Stolee's example below.  You could have:
```
Paths tracked by Git:   3,123,456 (100%)
Paths git is updating:   1 (0%)
```

Documents/AnalyticsToolDesignDocument.md
Halterer: Ironically as you were suggesting a console mock up Jameson was at my desk saying the same thing. I banged out an example locally but I will revisit it now with some of the ideas you had in yours as well as a loose example of what I will be building towards. As for the API, it does make sense to give a `--json` parameter as another method of passing data from the overall system. I'll take a look at the repair verb and consider that. I will continue working on plans for the system as a whole and investigate which statistics we can access readily from the verb vs. which ones may not be as accessible and might require the API.

kewillford: I like this!  I was thinking something similar but by folder so
```
DirectoryA - X file placeholders Y folder placeholder Z git files
DirectoryB - X1 file placeholders Y1 folder placeholder Z1 git files
etc...
```
And then you could run `gvfs statistics DirectoryA` to get the stats on that directory

Documents/AnalyticsToolDesignDocument.md
derrickstolee: If you are thinking instead about the object model, then it would be something like:

* `StatisticsVerb`
  - Lives in `GVFS/GVFS/CommandLine`
  - Handles all parameter handling and user-facing UI decisions
* `StatisticsWorker` (or something)
  - Lives in `GVFS/GVFS.Common/Statistics`
  - Runs multiple `StatisticsJob` instances, say `PlaceholderStatisticsJob` and `ModifiedFilesStatisticsJob`.

Then, you are talking about the difference between the command-line interface (verb) and the business logic (worker). The nice thing about the split here is that you can have the mount process also run the statistics on a regular basis and output the results a different way (telemetry, platform native notifications).

Documents/AnalyticsToolDesignDocument.md
Halterer: Ah okay I see now. I wasn't sure if a "fully hydrated placeholder" was a synonym for "full file" or for "hydrated placeholder" but now I see that it is the latter.

Documents/AnalyticsToolDesignDocument.md
derrickstolee: Definitely put Kevin's idea in the document, it's really good. Save it for a follow-up task after your initial prototype is complete.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,108 @@
...
+```terminal
+$ gvfs statistics
+
+Repository statistics
+Total paths tracked by git:        3,123,456 |  100%
+Total number of placeholders:        123,456 |    4%
+Total number of modififed paths:       1,234 |   <1%
+
+Total hydration percentage:                       4%
+Most hydrated directories:
```
derrickstolee: Oh, this summary of the few most-hydrated directories is so much better than the full list of directories I had in my head. Awesome.

Make a note for later: when do we stop this list? Perhaps include directories whose hydration percentage is at least the total hydration percentage? That way, we can easily see if they have 25% of their root directories 100% hydrated, or have 100% of their root directories 25% hydrated.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,108 @@
...
+  98% | src/main    | Primarily hydrated by "Sublime Text Editor"
+  79% | src/util    | Primarily hydrated by "Visual Studio 2019"
+  58% | assets/img  | Primarily hydrated by "cmd.exe"
+
+Repository status:  Healthy
+Recommended action: None
+
+To view git command performance run with "--perf"
```
derrickstolee: We will want to think about progress indicators if we are going a `git` perf benchmark. Love the idea!

kewillford: I wonder where this information will be stored because VFSForGit does not have the complete picture since it only knows about commands that take the lock and might not be a total picture of the time.  We should see if trace2 or something @jeffhostetler has done can help give us this information.

derrickstolee: Oh, I was thinking we'd have some set of benchmark commands that we run as part of the `statistics` verb. The progress indicator would say "Running `git log --graph -n 10`" or something as we run each command. It's important to find a set of read-only commands to test.

derrickstolee: Having a user self-serve certain Git statistics from their trace2 logs could be interesting, especially for external customers.

Documents/AnalyticsToolDesignDocument.md
wilbaker: > This API idea still worries me. Putting a new API in the named pipe layer still seems unnecessary (until we are forced to do it because we find a reason we can't access a resource from multiple processes). Instead, I'd start with the verb, and what work the verb needs to do in order to output a simple text-based message to a user summarizing what you found.

I agree with @derrickstolee here.  @kewillford, do you know if the SQLite DB for placeholders will allow a second process to connect/read while VFS4G has open connection(s)?

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,108 @@
...
+- **Utility tools for dehydration** - *Stretch goal* | After the analytics are collected, these components will provide utilities to help the user get their enlistment back to a healthier state. The ideal goal is to have VFS4G suggest uses to the user (EG: `This directory was hydrated by Sublime text editor. You can dehydrate it by running "gvfs dehydrate --confirm -d /path/to/repo"`)
+- **Additional visualization tools** - *Stretch goal* | With a comprehensive API in place, the possibility opens up for different visualization tools to be created that consume the json data and display it to the user in different ways, potentially even eventually interfacing with the utility tools for an interactive GUI
+
```
wilbaker: One more stretch goal to consider adding to the bottom of your list:

*Raising notifications to the user* - Is there someway we could alert the user that there's a potential issue and encourage them to run `gvfs statistics`?  Assuming the verb is non-intrusive one option might be having `GVFS.Service` run the command shortly after automatically mounting repos and presenting the user with a notification if hydration levels are over a certain threshold.

Documents/AnalyticsToolDesignDocument.md
kewillford: Yes SQLite allows for multiple connections from different processes.

Documents/AnalyticsToolDesignDocument.md
Halterer: Thanks a good idea! I hadn't thought of that, I'll note that down in the document as the default behavior. This morning I'm working on adding the other features and parameters that me and Jameson briefly discusses and I think will be cool. One of them would be a `--hydration` parameter which defaults to the root and just shows more information about the hydration of different directories, or could be pointed at a cone of the tree and then would go more in depth from there.

Documents/AnalyticsToolDesignDocument.md
Halterer: That's a really go idea, I definitely like that a lot. I'm still trying to hash out what percentage exactly indicates an overhydrated directory. I'm guessing either somebody has a pretty good guess or we can use some data to determine that

Documents/AnalyticsToolDesignDocument.md
jamill: > Having a user self-serve certain Git statistics from their trace2 logs could be interesting, especially for external customers.

One idea that has been discussed - the Git service daemon has the ability to dispatch data to multiple subscribers. I think it could dump trace2 logs (or some version of them) to disk, as well as whatever else it does. This would not address external scenarios (maybe for external use scenarios git can be configured to write out the trace2 logs, instead of going through a broker service).

Documents/AnalyticsToolDesignDocument.md
Halterer: Yeah splitting them up that way makes sense to me. I need to think some about how I'll refactor the document but I definitely think this fits more closely with the model I had in my head than the one I had on paper.

jeffhostetler:
This looks very nice.  I don't have anything to add at this point, but just wanted to say that I think this will be a very nice addition to the product.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,207 @@
+# VFS For Git: Analytics Tooling Design Document
+
+This document details the design of and plan for the **VFS4G analytics tool**. The tool aims to help users identify reasons why their enlistment may be slowing down and the `git` commands running slowly on a VFS4G mounted repository. By collecting and reporting statistics, the power to uncover and hopefully fix problems as they arrive moves to the user, empowering them to use the software to its fullest potential.
```
nickgra: High level question: 
Do we use the word 'enlistment' to describe a `gvfs clone`? In my mind, enlistment evokes the world/tooling that our internal customers used before VFS4G. 

I have been trying to refer to them as 'clones', but I've been sloppy about it for sure :)

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,207 @@
...
+Total paths tracked by git:        3,123,456 |  100%
+Total number of placeholders:        123,456 |    4%
+Total number of modififed paths:       1,234 |   <1%
+
+Total hydration percentage:                       4%
+Most hydrated directories:
+  98% | src/main    | Primarily hydrated by "Sublime Text Editor"
+  79% | src/util    | Primarily hydrated by "Visual Studio 2019"
+  58% | assets/img  | Primarily hydrated by "Photoshop"
+
```
nickgra: Could/should we extend this to include folder placeholders as well as file placeholders?

Documents/AnalyticsToolDesignDocument.md
Halterer: If I'm going to be honest its definitely just something that I totally picked up from listening to other people. I think Jameson has been using it to refer to an instance of a repository and as I've been most closely working with him that's the meaning I've taken to it as well. Is there a concrete meaning for the word that has been decided? I've already considered and I think Jameson agrees with potentially adding a glossary to define a few things. Perhaps "hydration" may also belong there, though I did define it as best I could in the document. I'd love to hear your feedback about that!

Documents/AnalyticsToolDesignDocument.md
Halterer: I think I'm going to eventually add a verbose option to the tool which will give those outputs as well. Just that in the base case I'm assuming that this is being used by someone a little higher level who may not even fully understand what a placeholder or modified path is so I was trying to present the simplest statistics first. I figured that the extra options for the tool might be a little more detailed than the scope of this document.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,207 @@
...
+Repository status:  Healthy
+Recommended action: None
+```
+
+## Background Information
+
+- **Modified paths** is the list of files which have been modified and thus correctly reflect everything on the disk. The paths here point to directories and files which have either been modified, created, or deleted. These files are handed to git when a git command reads its index in order to populate its list of files being tracked
```
jeschu1: **Modified paths** is the list of files which *may* have been modified

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,207 @@
...
+
+### `gvfs statistics --hydration` (With no path provided it runs on the root)
+
+```terminal
+$ gvfs statistics --hydration
+
+Total hydration level: 145%
+
+Programs responsible
+Sublime Text Editor |  78% | 143,566 files | 80,435 placeholders | 63,131 modified paths
```
jeschu1: How will we capture which app created a placholder / modified path?

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,207 @@
...
+- There are several **files states** through which a file in the enlistment can transition. Depending on where files are in this life cycle, performance can be better or worse. Files begin as ***virtual files*** which do not exist on disk but are projected into the file system. They transition to ***empty placeholders*** when the file is opened without a write handle but data has yet to be read. Once data has been read from the file, it becomes a ***hydrated placeholder*** which is managed by VFS4G but has its entirety on the disk. If a file is opened with a write handle, it becomes a ***full file*** with control ceded to git. Full files are tracked in the modified paths list. Additionally, if a file is created it is a full file. The final file type is a ***tombstone***, a special hidden ProjFS file denoting that a file was deleted and thus it should not be projected into the system any longer.
+
+![File Life Cycle](FileLifeCycle.png)
```
wilbaker: I think the diagram here is really nice.  

A few minor suggestions on the state transitions:

- Virtual Files always transition to Empty Placeholder first.  They will not transition directly to Tombstone or Full File (those two arrows can be removed).
- It might be worth mentioning in the diagram that "Virtual File" and "Tombstone" are Windows specific
- You could consider changing the "Data is read" text to something like "Data is required" as this transition will happen for reads or for writes that require the current file contents (e.g. appending to a file).

derrickstolee:
@jamill: is it time to close this?

PR #1224, opened 2019-05-30
ProductUpgrader: platform specific upgrade directories

jamill:
Different platforms have different file system layouts / capabilities. The existing upgrade logic is Windows centric and does not match the behaviors of other systems.

### Upgrade components that interact with the fileystem:
---

Component | description | data scope | Windows | Mac
----------- | ------------ | ----------- | --------- | ---- |
Upgrader Application | Temporary copy of upgrader application upgrade main application | Protected | ProgramData\GVFS\GVFS.Upgrade\Tools |  /usr/local/vfsforgit_upgrader/Tools
gvfs.config | config file for upgrade related settings | Protected | ProgramData\GVFS\gvfs.config | /usr/local/vfsforgit/gvfs.config
Upgrade Package Download | Location to download upgrade package | Protected | ProgramData\GVFS\GVFS.Upgrade\Download | /usr/local/vfsforgit_upgrader/Download
Upgrader Package Unpack | Location to unpack the downloaded update package | Protected | ProgramData\GVFS\GVFS.Upgrade\Download\InstallerTemp |  /usr/local/vfsforgit_upgrader/Download/InstallerTemp
Upgrader Logs | Directory to write upgrade logs to | Writable by user account | ProgramData\GVFS\GVFS.Upgrade\UpgraderLogs | ~/Library/Application Support/GVFS/GVFS.Upgrade/UpgraderLogs
Upgrade Available File | File indicating whether an upgrade is available | Written by GVFS.Service process<br><br>Windows: protected<br><br>Mac: writable by user account | ProgramData\GVFS\GVFS.Upgrade\Tools\HighestAvailableVersion |  ~/Library/Application Support/GVFS/GVFS.Upgrade/HighestAvailableVersion

### Data Scopes
---
This is not generally applicable across platforms, but is a model for how things are set up on Mac.

Scope | Usage | Windows | Mac
------ | ------- | -------- | -----
Non-protected Data | Unprivileged data that user accounts can access.  | Logs directory |  Logs and HighestAvailableVersion file. <br><br> ~/Library/Application&nbsp;Support/GVFS/GVFS.Upgrade
Protected Data | Requires root / elevation to mutate | ProgramData\GVFS\GVFS.Upgrade | /usr/local/vfsforgit_upgrader

### Motivation

On Windows, the upgrade related directories are located under the "service" level directory structure, and is system scoped and requires elevated permissions to modify (except for specific exemptions). The base directory is `ProgramData\GVFS\GVFS.Upgrade`. We have to tweak the permissions on the UpgraderLogs directory in order to allow non-elevated access, as non-elevated processes need to be able to write logs here.

On Mac, the root of the "service" data directories is a user scoped, and not system scoped. As the upgrade flow uses ACLs to protect the data, we need to find another directory.

This is a proposed change to enable VFS for Git to account for the platform differences.

### Mac File System Hierarchy
There are several candidates for where we can place "protected" data, including /usr/local, or /var. The proposed location for the Upgrader app and related data is: `/usr/local/vfsforgit_upgrader`, but I could see putting the data (and / or) application in `/var` as well.


derrickstolee:
@jamill Thanks for the detailed PR description! It is truly enlightening all the things that are going on here and how we need to change things for Mac.

GVFS/GVFS.Common/MacProductUpgraderInfo.cs
```diff
@@ -0,0 +1,27 @@
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common
+{
+    public class MacProductUpgraderInfo : ProductUpgraderInfoImpl
```
kewillford: Again in this PR, I don't think we should have platform specific code in `GVFS.Common`.

jamill: Yes - they currently share the same pattern. At a higher level, I think the question is do we want upgrader specific code in the platform, or the upgrader project to have knowledge of the platforms.

How would you have the upgrade code load the correct instance type (based on platform)?

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -16,24 +16,39 @@ public ProductUpgraderInfo(ITracer tracer, PhysicalFileSystem fileSystem)
             this.fileSystem = fileSystem;
         }
 
+        public static ProductUpgraderInfoImpl GetProductUpgraderInfoImpl()
```
kewillford: This just doesn't feel right to me.  Especially when you are calling this method for every call to get a directory?  Can we have an interface and create the instance that implements it only once and pass around the interface where we need it and get rid of all the statics?

GVFS/GVFS.Common/MacProductUpgraderInfo.cs
jamill: I am going to look at moving this to the platform specific projects

GVFS/GVFS.Common/ProductUpgraderInfo.cs
jamill: I like the final goal here - I can at least avoid creating a new instance for each call.

GVFS/GVFS.Common/ProductUpgraderInfoImpl.cs
```diff
@@ -0,0 +1,26 @@
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common
+{
+    public abstract class ProductUpgraderInfoImpl
```
jamill: I think the eventual goal here is to combine this and the `ProductUpgraderInfo` classes (or maybe this should just be a subclass of that type)...

nickgra:
I reviewed the description/design and it looks good to me!

Your model works for protecting the `vfsforgit_upgrader` folder if the user hasn't messed with the permissions on the `/usr/local` folder, but if the mode was modified to allow anyone to write to the folder, we could be vulnerable to the attack where someone could replace the `vfsforgit_upgrader` folder with their own content and have VFS4G execute it as admin. I don't know enough about ACLs on Windows, but we might be vulnerable in the same way there. 
(I don't have any immediate ideas on how we could fix this off the top of my head, besides resetting the permissions on `/usr/local` before we lay down the upgrader folder. I'll keep thinking on it though.)

Also, there are some places in your PR description where you have `/usr/local.vfsforgit_upgrader` instead of `/usr/local/vfsforgit_upgrader`.

alameenshah:
One more thought about using `/usr/local.vfsforgit_upgrader/Download/InstallerTemp` to store temp files. There is this directory `/private/var/tmp` which is special directory with a [sticky bit](https://en.wikipedia.org/wiki/Sticky_bit) set. We wouldn't need to set custom ACLs for protection.

jamill:
Thanks for the feedback!

@alameenshah -

> Is Protected Data directory really needed on a Mac?

Good question. As you pointed out, locked files are not an issue on Mac, but I was not sure if there could be other problems. For instance, could an operation cause an assembly load *after* we have overwritten the original application directory, and result in an "unexpected" assembly being found?

> Can we utilize the fact that our installers are code signed using Microsoft certificate, to avoid the need for additional protection?

Another good point - this will help with the installer(s) themselves, but there could be scripts or other artifacts included with the installer package.

@nickgra -

> but if the mode was modified to allow anyone to write to the folder, we could be vulnerable to the attack

Good point. We had an assumption that the user would not tweak this part of their system. We could include an additional check that the system directories are ACL'ed as we expect, and handle the cases where it isn't (maybe refuse to install?)

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -49,7 +49,7 @@ public class GitHubUpgrader : ProductUpgrader
         {
             this.Config = upgraderConfig;
 
-            string upgradesDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
+            string upgradesDirectoryPath = ProductUpgraderInfo.GetUpgradeBasicDataDirectory();
```
jamill: This should point to the `ProtectedDataDirectory`

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -21,19 +21,29 @@ public static string CurrentGVFSVersion()
...
+
+        public static string GetUpgradeProtectedDataDirectory()
+        {
+            return GVFSPlatform.Instance.GetUpgradeProtectedDataDirectory();
+        }
+
+        public static string GetUpgradeBasicDataDirectory()
+        {
+            return GVFSPlatform.Instance.GetUpgradeNonProtectedDataDirectory();
         }
```
kewillford: Can we use the same name and terminology throughout the code.  Here `Protected` matches but `Basic == NonProtected`?  I was also thinking that the terms should be `Restricted` and `User` especially since `Basic` is a little to generic and doesn't really describe it.

GVFS/GVFS.Platform.Mac/MacProductUpgraderPlatformStrategy.cs
```diff
@@ -21,8 +21,7 @@ public override bool TryPrepareLogDirectory(out string error)
 
         public override bool TryPrepareApplicationDirectory(out string error)
         {
-            string rootDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
-            string toolsDirectoryPath = Path.Combine(rootDirectoryPath, ProductUpgrader.ToolsDirectory);
+            string toolsDirectoryPath = ProductUpgraderInfo.GetUpgradeApplicationDirectory();
```
kewillford: Why do we call it `toolsDirectoryPath` but we are getting the `ApplicationDirectory`?  Do `tools == Application` or are the tools just located in the Application directory?

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -131,21 +131,21 @@ protected override void Execute(GVFSEnlistment enlistment)
                         {
                             // upgrader
                             this.CopyAllFiles(
-                                ProductUpgraderInfo.GetUpgradesDirectoryPath(),
```
jamill: This code will need to be adjusted for new log directory paths.

GVFS/GVFS.Common/GitHubUpgrader.cs
```diff
@@ -48,9 +48,6 @@ public class GitHubUpgrader : ProductUpgrader
             : base(currentVersion, tracer, dryRun, noVerify, fileSystem)
         {
             this.Config = upgraderConfig;
-
-            string upgradesDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
```
jamill: TODO: follow up on why we were creating this directory, and see if we still need to create a directory here.

GVFS/GVFS.Common/ProductUpgraderInfo.cs
jamill: As the concept of base `User` writable directory does not exist across platforms at this time, I removed `Basic` for now. I still have `Protected`, but I like `Restricted` as well.

GVFS/GVFS.Platform.Mac/MacProductUpgraderPlatformStrategy.cs
jamill: `Tools` was how it was referred to previously. I think `UpgradeApplication` is a bit more descriptive now - I will go through and update the terminology to be more consistent.

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -70,6 +72,27 @@ public static void Register(GVFSPlatform platform)
         public abstract bool IsElevated();
         public abstract string GetCurrentUser();
         public abstract string GetUserIdFromLoginSessionId(int sessionId, ITracer tracer);
+
+        /// <summary>
+        /// Get the directory for upgrades that is permissioned to
+        /// require elevated privledes to modify. This can be used for
```
wilbaker: typo: `privledes` -> `privileges`

GVFS/GVFS.Common/GitHubUpgrader.cs
wilbaker: I'm not sure if something else would have created the directory by the time `RecordHighestAvailableVersion` runs, but it looks like if the directory did not exist for `RecordHighestAvailableVersion` the call to `WriteAllText` in `RecordHighestAvailableVersion` would throw a `DirectoryNotFoundException`.

There might be other methods that `GetUpgradesDirectoryPath` that similarly assume the directory already exists.

GVFS/GVFS.Common/ProductUpgraderInfo.Shared.cs
```diff
@@ -9,15 +9,20 @@ public partial class ProductUpgraderInfo
...
 
-        public static bool IsLocalUpgradeAvailable(ITracer tracer, string gvfsDataRoot)
+        public static bool IsLocalUpgradeAvailable(ITracer tracer, string highestAvailableVersionDirectory)
         {
             try
             {
-                string upgradesDirectory = Path.Combine(gvfsDataRoot, UpgradeDirectoryName);
-
-                return File.Exists(GetHighestAvailableVersionFilePath(upgradesDirectory));
+                return File.Exists(GetHighestAvailableVersionFilePath(highestAvailableVersionDirectory));
```
wilbaker: Also update the parameter name for `GetHighestAvailableVersionFilePath'?

```
GetHighestAvailableVersionFilePath(string upgradesDirectory)
```

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -21,19 +21,35 @@ public static string CurrentGVFSVersion()
...
 
         public static string GetAssetDownloadsPath()
         {
-            return Path.Combine(GetUpgradesDirectoryPath(), DownloadDirectory);
+            return Path.Combine(
+                GVFSPlatform.Instance.GetUpgradeProtectedDataDirectory(),
+                ProductUpgraderInfo.DownloadDirectory);
+        }
+
+        public static string GetHighestAvailableVersionDirectory()
```
wilbaker: Can/should this method be removed and replaced with calls to `GVFSPlatform.Instance.GetUpgradeHighestAvailableVersionDirectory()` directly?

GVFS/GVFS.Common/ProductUpgraderInfo.cs
```diff
@@ -21,19 +21,35 @@ public static string CurrentGVFSVersion()
...
-        {
-            return GVFSPlatform.Instance.GetDataRootForGVFSComponent(UpgradeDirectoryName);
+        public static string GetUpgradeApplicationDirectory()
+        {
+            return Path.Combine(
+                GVFSPlatform.Instance.GetUpgradeProtectedDataDirectory(),
+                ProductUpgraderInfo.ApplicationDirectory);
+        }
+
+        public static string GetParentLogDirectoryPath()
```
wilbaker: Can/should this method be removed and replaced with calls to `GVFSPlatform.Instance.GetUpgradeLogDirectoryParentDirectory()` directly?

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -24,6 +26,14 @@ public MacPlatform()
         public override GVFSPlatformConstants Constants { get; } = new MacPlatformConstants();
         public override IPlatformFileSystem FileSystem { get; } = new MacFileSystem();
 
+        public override string GVFSConfigPath
+        {
+            get
+            {
+                return Path.Combine(this.Constants.GVFSBinDirectoryPath, LocalGVFSConfig.FileName);
```
wilbaker: Note: I am not an expert on how file permissions work on Mac, so there might be no issues here!

Does `sudo` care at all about what group a user is in?  I'm wondering if two users are modifying `/usr/local/vfsforgit/gvfs.config` is there any risk that one of them will block the other from being able to make further changes (using `sudo`)?

Similarly, is there any chance that one user's changes might prevent another users from being able to read the file (without using `sudo`)?

jamill: Great question(s)!

Root is its own account, so when running a command via sudo, `root` will be the owner of this file. The mode bits on the file will be such that `root` is the only account that can modify the file, while it will be readable by everyone. If two different users attempt to modify this file via a `sudo` command, it will be the `root` account that is making changes.

/cc @jeffhostetler to check my understanding :)

PR #1225, opened 2019-05-30
Upgrader - platform specific directory creation

jamill:
The upgrade logic between the different platforms has some high level differences that need to be accounted for:

1) Creation of directories with the correct ACLs
    - On Mac, the directories will be created with the expected ACLs, and we will not need to re-acl the directories.

There are other ways to handle the platform differences, but including:

  - performing the re-ACLing of directories on non-windows platforms. We could re-use more of the top level logic, but then we would be doing extra work with ACLs that is not needed.

  - Do no re-ACL on non-windows platforms. I did not want to cause confusion with the methods not doing what they advertise (especially around ACLs), in case someone misses this subtlety in the future. This would be hard from a maintainability standpoint...

2) Whether the Upgrader is responsible for remounting or not
    - On Windows, the upgrader itself is responsible for re-mounting after upgrade, while on Mac, the GVFS.Service will auto remount when the service starts up.

There is still some further work I need to do here to get this into shape, but I wanted to put this up for initial feedback on the general direction.

### DIFF
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -73,6 +73,10 @@ namespace GVFS.Common
             ITracer tracer,
             string lockPath);
 
+        public abstract ProductUpgraderPlatformStrategy CreateProductUpgraderPlatformInteractions(
+            PhysicalFileSystem fileSystem,
+            ITracer tracer);
+
         public bool TryGetNormalizedPathRoot(string path, out string pathRoot, out string errorMessage)
         {
             pathRoot = null;
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -138,16 +138,19 @@ namespace GVFS.Common
             public UnderConstructionFlags(
                 bool supportsGVFSUpgrade = true,
                 bool supportsGVFSConfig = true,
-                bool requiresDeprecatedGitHooksLoader = false)
+                bool requiresDeprecatedGitHooksLoader = false,
+                bool supportsNuGetEncryption = true)
             {
                 this.SupportsGVFSUpgrade = supportsGVFSUpgrade;
                 this.SupportsGVFSConfig = supportsGVFSConfig;
                 this.RequiresDeprecatedGitHooksLoader = requiresDeprecatedGitHooksLoader;
+                this.SupportsNuGetEncryption = supportsNuGetEncryption;
             }
 
             public bool SupportsGVFSUpgrade { get; }
             public bool SupportsGVFSConfig { get; }
             public bool RequiresDeprecatedGitHooksLoader { get; }
+            public bool SupportsNuGetEncryption { get; }
         }
     }
 }
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -27,21 +27,40 @@ namespace GVFS.Common
 
         protected bool noVerify;
         protected bool dryRun;
+        protected ProductUpgraderPlatformStrategy productUpgraderPlatformStrategy;
 
         private static readonly string UpgraderToolName = GVFSPlatform.Instance.Constants.GVFSUpgraderExecutableName;
 
-        public ProductUpgrader(
+        protected ProductUpgrader(
             string currentVersion,
             ITracer tracer,
             bool dryRun,
             bool noVerify,
             PhysicalFileSystem fileSystem)
+            : this(
+                  currentVersion,
+                  tracer,
+                  dryRun,
+                  noVerify,
+                  fileSystem,
+                  GVFSPlatform.Instance.CreateProductUpgraderPlatformInteractions(fileSystem, tracer))
+        {
+        }
+
+        protected ProductUpgrader(
+            string currentVersion,
+            ITracer tracer,
+            bool dryRun,
+            bool noVerify,
+            PhysicalFileSystem fileSystem,
+            ProductUpgraderPlatformStrategy productUpgraderPlatformStrategy)
         {
             this.installedVersion = new Version(currentVersion);
             this.dryRun = dryRun;
             this.noVerify = noVerify;
             this.tracer = tracer;
             this.fileSystem = fileSystem;
+            this.productUpgraderPlatformStrategy = productUpgraderPlatformStrategy;
         }
 
         /// <summary>
@@ -154,20 +173,7 @@ namespace GVFS.Common
             string rootDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
             string toolsDirectoryPath = Path.Combine(rootDirectoryPath, ToolsDirectory);
 
-            Exception deleteDirectoryException;
-            if (this.fileSystem.DirectoryExists(toolsDirectoryPath) &&
-                !this.fileSystem.TryDeleteDirectory(toolsDirectoryPath, out deleteDirectoryException))
-            {
-                upgraderToolPath = null;
-                error = $"Failed to delete {toolsDirectoryPath} - {deleteDirectoryException.Message}";
-                this.TraceException(deleteDirectoryException, nameof(this.TrySetupToolsDirectory), $"Error deleting {toolsDirectoryPath}.");
-                return false;
-            }
-
-            if (!this.fileSystem.TryCreateOrUpdateDirectoryToAdminModifyPermissions(
-                    this.tracer,
-                    toolsDirectoryPath,
-                    out error))
+            if (!this.productUpgraderPlatformStrategy.TryPrepareApplicationDirectory(out error))
             {
                 upgraderToolPath = null;
                 return false;
@@ -227,10 +233,7 @@ namespace GVFS.Common
 
         protected virtual bool TryCreateAndConfigureDownloadDirectory(ITracer tracer, out string error)
         {
-            return this.fileSystem.TryCreateOrUpdateDirectoryToAdminModifyPermissions(
-                tracer,
-                ProductUpgraderInfo.GetAssetDownloadsPath(),
-                out error);
+            return this.productUpgraderPlatformStrategy.TryPrepareDownloadDirectory(out error);
         }
 
         protected virtual void RunInstaller(string path, string args, out int exitCode, out string error)
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -20,6 +20,7 @@ namespace GVFS.Common
 
     public abstract class ProductUpgrader : IDisposable
     {
+        public const string ToolsDirectory = "Tools";
         protected readonly Version installedVersion;
         protected readonly ITracer tracer;
         protected readonly PhysicalFileSystem fileSystem;
@@ -27,7 +28,6 @@ namespace GVFS.Common
         protected bool noVerify;
         protected bool dryRun;
 
-        private const string ToolsDirectory = "Tools";
         private static readonly string UpgraderToolName = GVFSPlatform.Instance.Constants.GVFSUpgraderExecutableName;
 
         public ProductUpgrader(
--- a/GVFS/GVFS.Common/ProductUpgraderPlatformStrategy.cs
+++ b/GVFS/GVFS.Common/ProductUpgraderPlatformStrategy.cs
@@ -1,0 +1,38 @@
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common
+{
+    public abstract class ProductUpgraderPlatformStrategy
+    {
+        public ProductUpgraderPlatformStrategy(PhysicalFileSystem fileSystem, ITracer tracer)
+        {
+            this.FileSystem = fileSystem;
+            this.Tracer = tracer;
+        }
+
+        protected PhysicalFileSystem FileSystem { get; }
+        protected ITracer Tracer { get; }
+
+        public abstract bool TryPrepareLogDirectory(out string error);
+
+        public abstract bool TryPrepareApplicationDirectory(out string error);
+
+        public abstract bool TryPrepareDownloadDirectory(out string error);
+
+        protected void TraceException(Exception exception, string method, string message)
+        {
+            this.TraceException(this.Tracer, exception, method, message);
+        }
+
+        protected void TraceException(ITracer tracer, Exception exception, string method, string message)
+        {
+            EventMetadata metadata = new EventMetadata();
+            metadata.Add("Method", method);
+            metadata.Add("Exception", exception.ToString());
+            tracer.RelatedError(metadata, message);
+        }
+    }
+}
--- a/GVFS/GVFS.Platform.Mac/MacProductUpgraderPlatformStrategy.cs
+++ b/GVFS/GVFS.Platform.Mac/MacProductUpgraderPlatformStrategy.cs
@@ -1,0 +1,63 @@
+using GVFS.Common;
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Platform.Mac
+{
+    public class MacProductUpgraderPlatformStrategy : ProductUpgraderPlatformStrategy
+    {
+        public MacProductUpgraderPlatformStrategy(PhysicalFileSystem fileSystem, ITracer tracer)
+        : base(fileSystem, tracer)
+        {
+        }
+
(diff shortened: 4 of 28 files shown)


GVFS/GVFS.Common/ProductUpgraderPlatformStrategy.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public abstract class ProductUpgraderPlatformStrategy
+    {
+        public ProductUpgraderPlatformStrategy(PhysicalFileSystem fileSystem, ITracer tracer)
+        {
+            this.FileSystem = fileSystem;
+            this.Tracer = tracer;
+        }
+
+        protected PhysicalFileSystem FileSystem { get; private set; }
```
derrickstolee: nit: drop the `private set;` and it will become read-only (only assignable in the constructor).

GVFS/GVFS.Common/ProductUpgraderPlatformStrategyFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public static ProductUpgraderPlatformStrategy Create(
+            PhysicalFileSystem fileSystem,
+            ITracer tracer)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return new MacProductUpgraderPlatformStrategy(fileSystem, tracer);
+            }
+            else
```
derrickstolee: Perhaps be specific here and have the `else` condition throw a `NotSupportedException` or something? That way the Linux platform can slot in easilyi.

GVFS/GVFS.Common/WindowsProductUpgraderPlatformStrategy.cs
```diff
@@ -0,0 +1,75 @@
...
+            string rootDirectoryPath = ProductUpgraderInfo.GetUpgradesDirectoryPath();
+            string toolsDirectoryPath = Path.Combine(rootDirectoryPath, ProductUpgrader.ToolsDirectory);
+
+            Exception deleteDirectoryException;
+            if (this.FileSystem.DirectoryExists(toolsDirectoryPath) &&
+                !this.FileSystem.TryDeleteDirectory(toolsDirectoryPath, out deleteDirectoryException))
+            {
+                error = $"Failed to delete {toolsDirectoryPath} - {deleteDirectoryException.Message}";
+
+                /// this.TraceException(deleteDirectoryException, nameof(this.TrySetupToolsDirectory), $"Error deleting {toolsDirectoryPath}.");
```
derrickstolee: Is this intended to be commented-out?

GVFS/GVFS.Upgrader/MacUpgradeOrchestrator.cs
```diff
@@ -0,0 +1,22 @@
+using CommandLine;
```
derrickstolee: nit: many of these `using`s seem unused.

GVFS/GVFS.Upgrader/Program.cs
```diff
@@ -9,8 +9,18 @@ public static void Main(string[] args)
         {
             GVFSPlatformLoader.Initialize();
 
-            Parser.Default.ParseArguments<UpgradeOrchestrator>(args)
+            UpgradeOrchestrator upgraderType = UpgradeOrchestratorFactory.Create();
+
+            if (upgraderType is MacUpgradeOrchestrator)
```
derrickstolee: This whole chunk is confusing to me.

1. We should use our base type to avoid this `is` convention.
2. We create `upgraderType` just to check the `is`, and then we don't even use it.

It would be better if the `UpgradeOrchestrator` had some `ExecuteOnArgs(string [] args)` method that did this internal logic according to its own type.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
```diff
@@ -152,6 +152,35 @@ public void Execute()
             Environment.ExitCode = (int)this.ExitCode;
         }
 
+        protected virtual bool TryMountRepositories(out string consoleError)
```
derrickstolee: Should this be `abstract` in `UpgradeOrchestrator` and the implementation actually live in a `WindowsUpgradeOrchestrator`?

GVFS/GVFS.Upgrader/Program.cs
derrickstolee: I see now that there is no `WindowsUpgradeOrchestrator`, which would probably be helpful here.

GVFS/GVFS.Upgrader/UpgradeOrchestratorFactory.cs
```diff
@@ -0,0 +1,14 @@
+namespace GVFS.Upgrader
+{
+    public static class UpgradeOrchestratorFactory
+    {
+        public static UpgradeOrchestrator Create()
+        {
+#if MACOS_BUILD
+            return new MacUpgradeOrchestrator();
+#else
```
derrickstolee: `#elseif WINDOWS_BUILD` to avoid issues with Linux platform later.

GVFS/GVFS.Common/MacProductUpgraderPlatformStrategy.cs
```diff
@@ -0,0 +1,62 @@
+using GVFS.Common.FileSystem;
+using GVFS.Common.Tracing;
+using System;
+using System.IO;
+
+namespace GVFS.Common
+{
+    public class MacProductUpgraderPlatformStrategy : ProductUpgraderPlatformStrategy
```
kewillford: Still looking but I'm not sure what `PlatformStrategy` means.  

I also don't think we should have platform specific code going into `GVFS.Common`.  Looking at this and it seems these should be a part of `IPlatformFileSystem` with a `TryPrepareDirectory(string path, out string error)`.  We might even be able to remove the 2 permission method from that interface since they appear only to be for Windows.
```
        bool TryCreateDirectoryWithAdminAndUserModifyPermissions(string directoryPath, out string error);
        bool TryCreateOrUpdateDirectoryToAdminModifyPermissions(ITracer tracer, string directoryPath, out string error);
```

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
```diff
@@ -26,6 +26,7 @@
     <When Condition="'$(OS)' == 'Windows_NT'">
       <PropertyGroup>
         <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
+	<DefineConstants>$(DefineConstants);WINDOWS_BUILD</DefineConstants>
```
kewillford: Looks like this project is already referencing the Windows or the Mac Platform projects so why can't we have the platform specific code live in those projects instead of adding another way to create platform specific code?

GVFS/GVFS.Common/MacProductUpgraderPlatformStrategy.cs
jamill: > it seems these should be a part of IPlatformFileSystem with a TryPrepareDirectory(string path, out string error)

I like this idea - but the problem I ran into is that that action depends on the directory (which the upgrader has the most context on).

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
jamill: That is an option - but then we are putting upgrader specific code in the platform project...

jamill: I will look at moving this into the platform projects

jamill: I will look at moving this into the platform projects

jamill: I will look at moving this into platform specific projects

GVFS/GVFS.Common/ProductUpgraderPlatformStrategyFactory.cs
jamill: I liked this suggestion - but moved the "Creation" logic into the Platform classes, so that should help with the Linux platform.

GVFS/GVFS.Common/WindowsProductUpgraderPlatformStrategy.cs
jamill: I was lazy for my initial round of changes. The base class did not have the `TraceException` functionality. I added this to the base class, and uncommented this (and similar) lines.

GVFS/GVFS.Upgrader/GVFS.Upgrader.csproj
jamill: After thinking about this a bit more - I moved the platform specific code out of the Common project, but I am still am not sure about the `GVFS.Upgrader` project. This project is almost a separate application, and I didn't want to bring high level "other application" specific logic into the Platform classes.

I am open to thoughts on this, but I wasn't going to convert this just yet.

GVFS/GVFS.Common/MacProductUpgraderPlatformStrategy.cs
jamill: If there is an appropriate abstraction here, I am open to adding these directly to the `IPlatformFileSystem` class, but I haven't figured out a non-upgrader specific abstraction here... I will continue to think about this.

jamill:
Thanks @derrickstolee for the review!

> This means having an abstract base class with an implementation for each platform.

I think this makes sense as a goal. The problem I ran into is that we have a bunch of existing tests that exercise the current concrete implementation. If we make the `UpgradeOrchestrator` class abstract, then these tests will need reaction work... Either a "Testable" `UpgradeOrchestrator` class to test the common logic (which is most of the class), moving the tests to follow the new concrete implementations, etc... I like the suggestion - so let me investigate it a bit further.

> It would be nice to do some research into the Linux platform a bit here so we can figure out what can be combined into a POSIX platform class 

Good idea! I think a lot (or at least some) of the logic that is in the Mac specific class can actually go in the POSIX layer instead (as you noted).

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -73,6 +73,10 @@ public static void Register(GVFSPlatform platform)
             ITracer tracer,
             string lockPath);
 
+        public abstract ProductUpgraderPlatformStrategy CreateProductUpgraderPlatformInteractions(
```
jamill: I think the `ProductUpgraderPlatformStrategy` name is overly verbose and at the same time not clear. I don't think the `Interactions` suffix is any better... Still thinking for a better name.

GVFS/GVFS.Upgrader/UpgradeOrchestrator.cs
jamill: Yes - ideally this should be the case. For the current state of this, this interacts with how the tests are architected. I am still looking into what it would take to move to this structure (in which case most of the logic would remain the same, but the implementation would slide down into a Windows specific type). If it introduces a lot of churn, I was going to defer this (and not the followup work item)

GVFS/GVFS.Upgrader/Program.cs
jamill: I was able to clean this up a bit. I now have a stable `UpgradeOptions` class that then gets passed to a specific `UpgradeOrchestrator` implementation.

GVFS/GVFS.Upgrader/WindowsUpgradeOrchestrator.cs
```diff
@@ -0,0 +1,55 @@
...
+            : base(upgrader, tracer, fileSystem, preRunChecker, input, output)
+        {
+        }
+
+        public WindowsUpgradeOrchestrator(UpgradeOptions options)
+            : base(options)
+        {
+        }
+
+        protected override bool TryMountRepositories(out string consoleError)
```
derrickstolee: Should this be in the base (as `virtual`) and then the Mac version can leave it unimplemented until it is supported?

GVFS/GVFS.Upgrader/UpgradeOptions.cs
```diff
@@ -0,0 +1,22 @@
+using CommandLine;
+
+namespace GVFS.Upgrader
+{
+    [Verb("UpgradeOrchestrator", HelpText = "Upgrade VFS for Git.")]
```
derrickstolee: This `Verb` annotation seems out of place.

jamill: This usage is different than how we generally structure this in our product, but it does match the intended usage:

https://github.com/commandlineparser/commandline#for-verbs

I think we could ask whether we need this to be a verb, as this is being used to process command line arguments for a program that doesn't actually have verbs. I will note this as a potential cleanup.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -154,20 +173,7 @@ public virtual bool TrySetupToolsDirectory(out string upgraderToolPath, out stri
...
-                return false;
-            }
-
-            if (!this.fileSystem.TryCreateOrUpdateDirectoryToAdminModifyPermissions(
-                    this.tracer,
-                    toolsDirectoryPath,
-                    out error))
+            if (!this.productUpgraderPlatformStrategy.TryPrepareApplicationDirectory(out error))
             {
                 upgraderToolPath = null;
```
alameenshah: Not directly related to this PR. Can you take a look at lines #182 through #186. `TrySetupToolsDirectory` is trying to copy everything inside `C:\Program Files\GVFS` to "C:\ProgramData\GVFS\GVFS.Upgrade\Tools". The whole set of `GVFS` assemblies get copied. This is not the intent (we only need GVFS.Upgrade.exe and its dependencies), right?

GVFS/GVFS.Common/ProductUpgraderPlatformStrategy.cs
```diff
@@ -0,0 +1,38 @@
...
+        protected PhysicalFileSystem FileSystem { get; }
+        protected ITracer Tracer { get; }
+
+        public abstract bool TryPrepareLogDirectory(out string error);
+
+        public abstract bool TryPrepareApplicationDirectory(out string error);
+
+        public abstract bool TryPrepareDownloadDirectory(out string error);
+
+        protected void TraceException(Exception exception, string method, string message)
```
alameenshah: This is a useful utility method, that is not related to `ProductUpgraderPlatformStrategy`. Thoughts on moving this to `JsonTracer` as a new `RelatedException` method?

GVFS/GVFS.Platform.Mac/MacProductUpgraderPlatformStrategy.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (this.FileSystem.DirectoryExists(toolsDirectoryPath) &&
+                !this.FileSystem.TryDeleteDirectory(toolsDirectoryPath, out deleteDirectoryException))
+            {
+                error = $"Failed to delete {toolsDirectoryPath} - {deleteDirectoryException.Message}";
+
+                this.TraceException(deleteDirectoryException, nameof(this.TryPrepareApplicationDirectory), $"Error deleting {toolsDirectoryPath}.");
+                return false;
+            }
+
+            this.FileSystem.CreateDirectory(toolsDirectoryPath);
```
alameenshah: Upgrade verb runs elevated on the Mac too, right? Would this end up creating a root:admin owned directory inside user's `$HOME/Application\ Support/GVFS`? (root owned directory inside user home will be unusual.)

GVFS/GVFS.Common/ProductUpgraderPlatformStrategy.cs
jamill: Yeah - good point. We duplicate this utility method is several places, and we should probably centralize it. I will include it as a potential cleanup (and standardize this everywhere).

GVFS/GVFS.Platform.Mac/MacProductUpgraderPlatformStrategy.cs
jamill: > Would this end up creating a root:admin owned directory inside user's $HOME/Application\ Support/GVFS?

Good question - the "application" directory will be created under /usr/local/vfsforgit_upgrader, which is root owned.

GVFS/GVFS.Common/ProductUpgrader.cs
jamill: > The whole set of GVFS assemblies get copied. This is not the intent (we only need GVFS.Upgrade.exe and its dependencies), right?

The problem was that it was fragile to keep the list of dependencies straight (across all platforms). We now copy over the entire "application", including some extra files, but I don't think there should be an issue with copying over these extra file... If there are concerns with copying the extra files, please let me know :)

alameenshah: I see...

PR #1129, opened 2019-05-09
Upgrader for macOS MVP

jamill:
This issue is to track the work for building the upgrade functionality for macOS. It will also contain a proof of concept for the end-to-end functionality. As the PRs for individual tasks are ready, I will link them here.

### Tasks
- [x] #1059  GVFS.Upgrader project builds as a netcore project. 
- [x] #1155 macOS: rooted pipe names for GVFS Service
- [x] #1151 macOS: get mount-all command working
- [x] #1159 macOS: GVFS.Config
- [x] #1159 macOS: IsElevated implementation
- [x] #1228 upgrade verb: copy entire application directory
- [x] #1239 netcore: NuGet credentials tweaks
- [x] #1234  macOS: IsServiceIstalledAndRunning implementation
- [x] #1234 macOS: Program locator adjustments
- [x] #1234 macOS: InstallerPreRunChecker adjustments
- [x] #1225 upgrader: handle platform differences for high level upgrade orchestration
- [x] #1230 upgrader: enable running a command provided by host environment 
- [x] #1224 upgrader: handle different platform file system hierarchies
- [x] #1284 POSIX: support "inline" upgrade
- [x] #1284 macOS: Kext supports upgrade

### Deferred / Followup Tasks
- Automated tests for `CopyDirectoryRecursive`
- netcore: NuGet package verification

PR #1228, opened 2019-05-31
Upgrader: copy entire application directory

jamill:
When upgrading VFS for Git via the built-in upgrade logic, the upgrade verb will copy the upgrader application to a temporary directory, and run the upgrade app from the temporary directory. The main reason for this is so that the upgrade application can update the original application without open file handles getting in the way (and preventing the deletion of the application that is being updated).

The upgrade verb currently has a hard coded list of dependencies (files) required for the upgrader application to run. This can be fragile as we need to keep this list in sync with the actual dependencies, across multiple platforms. This can easily break if the dependencies change, and we update the list of files included with the installer, but not with the upgrader application. A specific occurrence of this happened in #1144.

Instead, just copy over the whole VFS for Git application directory when creating a temporary copy of the upgrader application to run. We might copy over a few additional files, but this should not cause any noticeable downsides.

#### TODO:
- ~~[ ] Tests~~ copied to backlog in #1129

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
@@ -39,6 +39,26 @@ namespace GVFS.Common.FileSystem
             directory.Delete();
         }
 
+        public virtual void CopyDirectoryRecursive(string srcDirectoryPath, string dstDirectoryPath)
+        {
+            DirectoryInfo srcDirectory = new DirectoryInfo(srcDirectoryPath);
+
+            if (!this.DirectoryExists(dstDirectoryPath))
+            {
+                this.CreateDirectory(dstDirectoryPath);
+            }
+
+            foreach (FileInfo file in srcDirectory.EnumerateFiles())
+            {
+                this.CopyFile(file.FullName, Path.Combine(dstDirectoryPath, file.Name), overwrite: true);
+            }
+
+            foreach (DirectoryInfo subDirectory in srcDirectory.EnumerateDirectories())
+            {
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -39,6 +39,23 @@ public virtual void DeleteDirectory(string path, bool recursive = true)
             directory.Delete();
         }
 
+        public virtual void CopyDirectory(string source, string destination)
```
jrbriggs: naming nit: `CopyDirectoryRecursive` ?

jamill:
> I see your "Tests" TODO. I'm assuming you just want a unit test on the recursive copy, right?

I have been thinking about how to test this. My initial idea was to do a unit test to verify that a source directory (with some structure) was copied to a destination directory as expected, but then I would be testing the `PhysicalFileSystem` with a `MockPhysicalFileSystem`. Another thought was to just make a functional test for this... 

> Is there any reason the CopyDirectoryRecursive() method shouldn't be an extension method on a FileSystem object? That would allow calling it on a mock object during the test.

That might help with how to test this. I don't think there is any reason why it couldn't be an extension method. We don't have any other extension methods for this class (yet), so this would be the first. It seems similar in function to existing methods such as `DeleteDirectory`, so I wanted to keep them close.

> I'm assuming our functional tests already cover what we can in the upgrade scenario.

I think we are a bit thin here - we have discussed improving this, and I added a GitHub issue to track this work (in case we don't have an existing issue).

jamill:
I am going to proceed with this change as is. I included the test topic as a deferred / follow up item in the top level issue tracking the macOS upgrader (#1129)

PR #1230, opened 2019-06-01
 Upgrader: add command install action

jamill:
When the upgrader runs install actions defined in the installer manifest, it previously only supported starting an application or document contained in the installer package. The installer author might
want to run instead a command that is provided by the host operating
system / environment. For example, on macOS the installer author might want to run bash
or the "installer" binary directly as one of the install actions defined in the installer payload.

As any command can be included as an installer action, there is not any additional verification we can make about the command, except for validating the integrity of the containing NuGet package, and enforcing that the package contents have not been tampered with before we run the command.

This does not currently include a mechanism to define an environment variable.

### DIFF
--- a/GVFS/GVFS.Common/NuGetUpgrade/InstallActionInfo.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/InstallActionInfo.cs
@@ -7,13 +7,15 @@ namespace GVFS.Common.NuGetUpgrade
         /// </summary>
         public const string ManifestEntryInstallationIdToken = "installation_id";
         public const string ManifestEntryLogDirectoryToken = "log_directory";
+        public const string ManifestEntryInstallerBaseDirectoryToken = "installer_base_path";
 
-        public InstallActionInfo(string name, string version, string args, string installerRelativePath)
+        public InstallActionInfo(string name, string version, string args, string installerRelativePath, string command)
         {
             this.Name = name;
             this.Version = version;
             this.Args = args;
             this.InstallerRelativePath = installerRelativePath;
+            this.Command = command;
         }
 
         /// <summary>
@@ -27,8 +29,8 @@ namespace GVFS.Common.NuGetUpgrade
         public string Name { get; }
 
         /// <summary>
-        /// The path to the installer, relative to the
-        /// content directory of the NuGet package.
+        /// The path to the application or document to start. The path
+        /// is relative to the content directory of the NuGet package.
         /// </summary>
         public string InstallerRelativePath { get; }
 
@@ -36,5 +38,12 @@ namespace GVFS.Common.NuGetUpgrade
         /// The version of the component that this entry installs
         /// </summary>
         public string Version { get; }
+
+        /// <summary>
+        /// The command to run. If this is present, the command is run
+        /// directly (with the processed args), and the
+        /// InstallerRelativePath property is ignored.
+        /// </summary>
+        public string Command { get; }
     }
 }
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
@@ -488,7 +488,6 @@ namespace GVFS.Common.NuGetUpgrade
 
         private bool TryGetPersonalAccessToken(string credentialUrl, ITracer tracer, out string token, out string error)
         {
-            error = null;
             return this.credentialStore.TryGetCredential(this.tracer, credentialUrl, out string username, out token, out error);
         }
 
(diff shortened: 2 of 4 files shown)


GVFS/GVFS.Common/NuGetUpgrade/InstallActionInfo.cs
```diff
@@ -27,14 +29,21 @@ public InstallActionInfo(string name, string version, string args, string instal
...
         /// </summary>
         public string InstallerRelativePath { get; }
 
         /// <summary>
         /// The version of the component that this entry installs
         /// </summary>
         public string Version { get; }
+
+        /// <summary>
+        /// The command to run. If this is present, the command is run
```
derrickstolee: "If this is present" probably means that you [want the `string` to be nullable (`string?`)](https://blog.cdemi.io/whats-coming-in-c-8-0-nullable-reference-types/).

mjcheetham: We're not using C# 8 in VFS for Git yet. Given how hard it is to get any new syntax into the project, adopting something like non-nullable reference types is not going to happen sadly.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -150,10 +150,12 @@ public class NuGetUpgrader : ProductUpgrader
...
         /// <param name="installationId">A unique installer ID to replace the installer_id token with.</param>
         /// <returns>The argument string with tokens replaced.</returns>
-        public static string ReplaceArgTokens(string src, string installationId, string logsDirectory)
+        public static string ReplaceArgTokens(string src, string installationId, string logsDirectory, string installerBaseDirectory)
         {
-            string dst = src.Replace(NuGetUpgrader.replacementToken(InstallActionInfo.ManifestEntryLogDirectoryToken), logsDirectory)
-                .Replace(NuGetUpgrader.replacementToken(InstallActionInfo.ManifestEntryInstallationIdToken), installationId);
+            string dst = src
+                .Replace(NuGetUpgrader.replacementToken(InstallActionInfo.ManifestEntryLogDirectoryToken), logsDirectory)
+                .Replace(NuGetUpgrader.replacementToken(InstallActionInfo.ManifestEntryInstallationIdToken), installationId)
```
mjcheetham: Not your change, but why is `NuGetUpgader.replacementToken` in camelCase and not PascalCase?

jamill: I was the one who originally wrote this - it should be PascalCased. I will fix this up in a separate commit (but just include it with this PR)

alameenshah:
>> When running install actions, we previously only supported starting an
application or document contained in the installer package. You might
want to run instead a command that is provided by the host operating
system / environment. For example, on macOS you might want to run bash
or the "installer" binary directly.

Could you please update the description with a few additional details.
1. Can regular mac end user provide a command for `gvfs upgrade` to run? Or is this command included in the installer payload?
2. Similar to the downloaded upgrade package validation, is there any sort of validation that we can do on the command?
3. How does end user specify an environment setting? Is this the bash environment variables or `gvfs config`?

GVFS/GVFS.Common/NuGetUpgrade/InstallActionInfo.cs
jamill: I like the idea of the type being more clear about how it is intended to be used, but in this case, I think if we were to switch to use nullable strings, we should do that in more than just this one place (at least this entire file, and maybe a larger scope of the project to be consistent).

jamill:
@alameenshah Thanks for the review! I updated the description to address the (good) questions you brought up.

GVFS/GVFS.Common/NuGetUpgrade/NuGetUpgrader.cs
```diff
@@ -383,7 +386,15 @@ public override bool TryRunInstaller(InstallActionWrapper installActionWrapper,
                             {
                                 if (!this.dryRun)
                                 {
-                                    this.RunInstaller(installerPath, processedArgs, out installerExitCode, out localError);
+                                    if (!string.IsNullOrEmpty(entry.Command))
+                                    {
+                                        this.RunInstaller(entry.Command, processedArgs, out installerExitCode, out localError);
```
kewillford: Who controls the command that is ran here?  Are there any security concerns?

kewillford: Should we scope this down or allow any command?

jamill: The command comes from the "installer payload". Inside the "installer payload" is a manifest that describes the commands that should be run during upgrade / installation. 

> Are there any security concerns?

This is a good item to keep in mind - this is a quite critical component that we need to protect, and one that we have put a good deal of thought / effort into. If you have any suggestions on how we can minimize concerns here, I would love to follow up on them. We protect the components that control where we download packages from, we verify the integrity of the NuGet package that is downloaded (by default - you can currently opt-out of this verification, and net core does not currently support verification yet), and we download / unpack the package into locations that are only writable by root.

> Should we scope this down or allow any command

Interesting idea... It is quite broad and flexible right now (by design).

I will point out, that this change does not broaden the attack surface, as we already had the ability to run scripts in the upgrader package (by design).

PR #1084, opened 2019-04-30
Support alternate names for the .gvfs folder

kivikakk:
On Linux, the existence of a GNOME Virtual file system `.gvfs` directory precludes the use of VFS for Git entirely. For both this reason, and to fully disambiguate the VFSForGit project from GVfs, it would be preferable to use a different folder name for VFSForGit's metadata, such as `.vfsforgit`.

This PR introduces support for modifying the name of `.gvfs` folder on a per-platform basis. `GVFSPlatform.Instance.Constants.DotGVFSRoot` now holds the appropriate name of that folder.

Fixes #987. 

/cc @chrisd8088 @jrbriggs @wilbaker @nickgra

PR #1232, opened 2019-06-03
Perform case-sensitive path comparisons on Linux

kivikakk:
On Windows and Mac, path comparisons are case-insensitive, as the filesystems are case-insensitive -- `.git` and `.GIT` are the same path name, and you can't have two different files or directories with names that only differ in case.

On Linux, filesystems (usually) are case-sensitive -- it is possible to have `.git`, `.GIT` and `.Git` entries in the one directory, all referring to different things. Accordingly, we need to do all path comparisons in a case-sensitive fashion.

This change should be a no-op for Windows and Mac, inasmuch as no different values will result after the change. This change does mean `GVFSPlatform.Instance` is being relied on in more places, and it's difficult to audit if a given piece of code could be called without `GVFSPlatform.Instance` already being initialised. (This bit me when working on #1084, where I tried to use `GVFSPlatform.Instance` in `GVFS.FunctionalTests.LockHolder`, which silently died with NRE because it wasn't initialised.)

Nonetheless, unit and (Windows) functional tests pass, so it might be fine.

Fixes github#29.
/cc @chrisd8088

### DIFF
--- a/GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
+++ b/GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
@@ -2,6 +2,7 @@
 using GVFS.Common.Prefetch.Git;
 using GVFS.Tests;
 using GVFS.Tests.Should;
+using GVFS.UnitTests.Category;
 using GVFS.UnitTests.Mock.Common;
 using GVFS.UnitTests.Mock.Git;
 using NUnit.Framework;
@@ -102,6 +103,7 @@ namespace GVFS.UnitTests.Prefetch
         // Delete a folder with two sub folders each with a single file
         // Readd it with a different casing and same contents
         [TestCase]
+        [Category(CategoryConstants.CaseInsensitiveFileSystemOnly)]
         public void ParsesCaseChangesAsAdds()
         {
             MockTracer tracer = new MockTracer();
--- a/GVFS/FastFetch/WorkingTree.cs
+++ b/GVFS/FastFetch/WorkingTree.cs
@@ -26,7 +26,7 @@ namespace FastFetch
 
             Parallel.ForEach(
                 dir.EnumerateDirectories().Where(subdir =>
-                    (!subdir.Name.Equals(GVFSConstants.DotGit.Root, StringComparison.OrdinalIgnoreCase) &&
+                    (!subdir.Name.Equals(GVFSConstants.DotGit.Root, GVFSPlatform.Instance.Constants.PathComparison) &&
                      !subdir.Attributes.HasFlag(FileAttributes.ReparsePoint))),
                 subdir => { ForAllDirectories(subdir, asyncParallelCallback); });
         }
--- a/GVFS/GVFS.Common/FileSystem/HooksInstaller.cs
+++ b/GVFS/GVFS.Common/FileSystem/HooksInstaller.cs
@@ -29,7 +29,7 @@ namespace GVFS.Common.FileSystem
         {
             IEnumerable<string> valuableHooksLines = defaultHooksLines.Where(line => !string.IsNullOrEmpty(line.Trim()));
 
-            if (valuableHooksLines.Contains(GVFSPlatform.Instance.Constants.GVFSHooksExecutableName, StringComparer.OrdinalIgnoreCase))
+            if (valuableHooksLines.Contains(GVFSPlatform.Instance.Constants.GVFSHooksExecutableName, GVFSPlatform.Instance.Constants.PathComparer))
             {
                 throw new HooksConfigurationException(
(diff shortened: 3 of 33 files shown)


kivikakk:
I thought I'd give it a try! Looks like maybe I don't have permissions?

derrickstolee:
> I thought I'd give it a try! Looks like maybe I don't have permissions?

It looks like the mac build worked for you. Sorry that you don't have access to the build results themselves. It's unfortunate that we always get those "No pipelines are associated..." messages.

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -104,6 +105,11 @@ public void CanParseBackwardsDiff()
         [TestCase]
         public void ParsesCaseChangesAsAdds()
         {
+            if (GVFSPlatform.Instance.Constants.CaseSensitiveFileSystem)
```
derrickstolee: I wonder if there is a way to do this with test categories using the `[]` annotations.

GVFS/GVFS.Common/PlaceholderListDatabase.cs
```diff
@@ -260,7 +260,7 @@ public void WriteAllEntriesAndFlush(IEnumerable<IPlaceholderData> updatedPlaceho
 
         private IEnumerable<string> GenerateDataLines(IEnumerable<IPlaceholderData> updatedPlaceholders)
```
wilbaker: FYI - On master the placeholders are now being stored in a SQLite database (see #1140).  When updating the Linux feature branch to pick up that change you'll need to make the table creation call aware of whether the platform is case sensitive (currently the table is created with a case-insensitive path/key).

wilbaker:
FYI - If you need to support repos that have files whose names only differ by case, you'll also need to update `GitIndexProjection.SortedFolderEntries`.  

It uses `LazyUTF8String.CaseInsensitiveCompare` and it will need to use a new `CaseSensitiveCompare` if you need to support repos that allow names that only differ by case.

Please note that this is an **extremely** hot path, and has been tuned to eliminate as many instructions as possible (it's hit for every entry when parsing the index) and so we'll have to be careful with how this code is updated.

Related, there is a project `GVFS.PerfProfiling` that can be used to measure the perf of VFS4G's index loading/parsing code which I've found to be helpful when making changes in this area.  Unfortunately it looks like it hasn't been converted to .NET core yet, and so right now it can only be run on Windows.

derrickstolee:
> FYI - If you need to support repos that have files whose names only differ by case, you'll also need to update `GitIndexProjection.SortedFolderEntries`.
> 
> It uses `LazyUTF8String.CaseInsensitiveCompare` and it will need to use a new `CaseSensitiveCompare` if you need to support repos that allow names that only differ by case.

Sounds like we will need a new set of functional tests that only run on Linux that check case-sensitive comparisons by explicitly adding files or folders that differ only by case, otherwise, we will never know if we've checked all the things we need to make this switch.

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
kivikakk: Time to find out!

kivikakk: be79df24 uses this approach -- thanks!

kivikakk:
For now we've decided to not support in-tree files whose name only differ by case -- this PR will be scoped only to the operations VFS for Git performs on the filesystem for its own functioning (like locating the `.git` directory, or matching `.keep` and `.idx` files, etc.) -- the repository content will remain case-insensitive.

kivikakk:
I think this should be good to go -- @derrickstolee @wilbaker fyi in case you'd like to re-review. I've lessened the scope of the change so nothing that impacts tracked content should be changed now.

chrisd8088:
Purely as a note for future work (not for this PR, unless someone feels it's important), I have a suspicion there are a [few](https://github.com/microsoft/VFSForGit/blob/c7df435fb69761a8d1419bf86b653dee66b930da/GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs#L155) [places](https://github.com/microsoft/VFSForGit/blob/c7df435fb69761a8d1419bf86b653dee66b930da/GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs#L168) in the functional test suite where we may want to have a similar OS-specific path comparator instead of `StringComparison.OrdinalIgnoreCase()`.

kivikakk:
The paths referred to in those instances are in-tree paths (tracked content), and so are not in the scope of the change proposed in this PR. We'll only change those if we proceed to make VFSForGit support case-sensitivity in the repository contents. Right now we're just behaving appropriately with respect to the backing store.

PR #1234, opened 2019-06-03
Upgrader: implement cross platform functionality

jamill:
Includes the following changes to enable cross platform functionality required by the upgrade functionality:

 - Platform specific blocking process lists
 - Platform specific ability to locate programs
 - Implement IsServiceInstalledAndRunning for Mac platform

To see how these changes fit in with the larger Mac Upgrader work, please look at #1129

### DIFF
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -106,6 +106,8 @@ namespace GVFS.Common
 
             public abstract string GVFSExecutableName { get; }
 
+            public abstract string ProgramLocaterCommand { get; }
+
             public string GVFSHooksExecutableName
             {
                 get { return "GVFS.Hooks" + this.ExecutableExtension; }
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -108,6 +108,18 @@ namespace GVFS.Common
 
             public abstract string ProgramLocaterCommand { get; }
 
+            /// <summary>
+            /// Different platforms can have different requirements
+            /// around which processes can block upgrade. For example,
+            /// on Windows, we will block upgrade if any GVFS commands
+            /// are running, but on POSIX platforms, we relax this
+            /// constraint to allow upgrade to run while the upgrade
+            /// command is running. Another example is that
+            /// Non-windows platforms do not block upgrade when bash
+            /// is running.
+            /// </summary>
+            public abstract HashSet<string> UpgradeBlockingProcesses { get; }
+
             public string GVFSHooksExecutableName
             {
                 get { return "GVFS.Hooks" + this.ExecutableExtension; }
--- a/GVFS/GVFS.Common/ProcessHelper.cs
+++ b/GVFS/GVFS.Common/ProcessHelper.cs
@@ -62,9 +62,9 @@ namespace GVFS.Common
             return currentVersion.Major == 0;
         }
 
-        public static string WhereDirectory(string processName)
+        public static string GetProgramLocation(string programLocaterCommand, string processName)
         {
-            ProcessResult result = ProcessHelper.Run("where", processName);
+            ProcessResult result = ProcessHelper.Run(programLocaterCommand, processName);
             if (result.ExitCode != 0)
             {
                 return null;
--- a/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
+++ b/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
@@ -34,7 +34,7 @@
         
         function GetBlockingProcesses()
         {
-            var watchList = [ "gvfs", "GVFS.Mount", "git", "gitk", "Wish" ];
+            var watchList = [ "GVFS.Mount", "git", "gitk", "Wish" ];
             var blockingProcesses = new Array();
             for (var process of watchList)
             {
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -207,7 +207,7 @@ namespace GVFS.Platform.POSIX
 
             public override HashSet<string> UpgradeBlockingProcesses
             {
-                get { return new HashSet<string> { "GVFS.Mount", "git", "wish" }; }
+                get { return new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "GVFS.Mount", "git", "wish" }; }
             }
         }
     }
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -199,6 +199,12 @@ namespace GVFS.Platform.POSIX
             {
                 get { return "gvfs"; }
             }
+
+            public override string ProgramLocaterCommand
+            {
+                get { return "which"; }
+            }
+
         }
     }
 }
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -205,6 +205,10 @@ namespace GVFS.Platform.POSIX
                 get { return "which"; }
             }
 
+            public override HashSet<string> UpgradeBlockingProcesses
+            {
+                get { return new HashSet<string> { "GVFS.Mount", "git", "wish" }; }
+            }
         }
     }
 }
--- a/GVFS/GVFS.Common/IProcessRunner.cs
+++ b/GVFS/GVFS.Common/IProcessRunner.cs
@@ -1,0 +1,12 @@
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Interface around process helper methods. This is to enable
+    /// testing of components that interact with the ProcessHelper
+    /// static class.
+    /// </summary>
+    public interface IProcessRunner
+    {
+        ProcessResult Run(string programName, string args, bool redirectOutput);
+    }
+}
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -12,8 +12,6 @@ namespace GVFS.Upgrader
 {
     public class InstallerPreRunChecker
     {
-        private static readonly HashSet<string> BlockingProcessSet = new HashSet<string> { "GVFS", "GVFS.Mount", "git", "ssh-agent", "bash", "wish", "git-bash" };
-
         private ITracer tracer;
 
         public InstallerPreRunChecker(ITracer tracer, string commandToRerun)
@@ -139,7 +137,7 @@ namespace GVFS.Upgrader
 
             foreach (Process process in allProcesses)
             {
-                if (process.Id == currentProcessId || !BlockingProcessSet.Contains(process.ProcessName))
+                if (process.Id == currentProcessId || !GVFSPlatform.Instance.Constants.UpgradeBlockingProcesses.Contains(process.ProcessName))
                 {
                     continue;
                 }
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -153,7 +153,7 @@ namespace GVFS.Upgrader
 
         protected virtual bool TryRunGVFSWithArgs(string args, out string consoleError)
         {
-            string gvfsDirectory = ProcessHelper.WhereDirectory(GVFSPlatform.Instance.Constants.GVFSExecutableName);
+            string gvfsDirectory = ProcessHelper.GetProgramLocation(GVFSPlatform.Instance.Constants.ProgramLocaterCommand, GVFSPlatform.Instance.Constants.GVFSExecutableName);
             if (!string.IsNullOrEmpty(gvfsDirectory))
             {
                 string gvfsPath = Path.Combine(gvfsDirectory, GVFSPlatform.Instance.Constants.GVFSExecutableName);
--- a/GVFS/GVFS.Common/ProcessRunnerImpl.cs
+++ b/GVFS/GVFS.Common/ProcessRunnerImpl.cs
@@ -1,0 +1,16 @@
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Default product implementation of IProcessRunner
+    /// interface. Delegates calls to static ProcessHelper class. This
+    /// class can be used to enable testing of components that call
+    /// into the ProcessHelper functionality.
+    /// </summary>
+    public class ProcessRunnerImpl : IProcessRunner
+    {
+        public ProcessResult Run(string programName, string args, bool redirectOutput)
+        {
+            return ProcessHelper.Run(programName, args, redirectOutput);
+        }
+    }
+}
--- a/GVFS/GVFS.PerfProfiling/ProfilingEnvironment.cs
+++ b/GVFS/GVFS.PerfProfiling/ProfilingEnvironment.cs
@@ -26,7 +26,7 @@ namespace GVFS.PerfProfiling
         private GVFSEnlistment CreateEnlistment(string enlistmentRootPath)
         {
             string gitBinPath = GVFSPlatform.Instance.GitInstallation.GetInstalledGitBinPath();
-            string hooksPath = ProcessHelper.WhereDirectory(GVFSPlatform.Instance.Constants.GVFSHooksExecutableName);
+            string hooksPath = ProcessHelper.GetProgramLocation(GVFSPlatform.Instance.Constants.ProgramLocaterCommand, GVFSPlatform.Instance.Constants.GVFSHooksExecutableName);
 
             return GVFSEnlistment.CreateFromDirectory(enlistmentRootPath, gitBinPath, hooksPath, authentication: null);
         }
--- a/GVFS/GVFS.Platform.Mac/MacDaemonController.cs
+++ b/GVFS/GVFS.Platform.Mac/MacDaemonController.cs
@@ -1,0 +1,74 @@
+using GVFS.Common;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Platform.Mac
+{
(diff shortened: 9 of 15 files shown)


GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -88,6 +88,8 @@ public bool TryGetNormalizedPathRoot(string path, out string pathRoot, out strin
             return true;
         }
 
+        public abstract string GetProgramLocater();
```
jamill: Feedback from #1129 - make this a property and move into `GVFSPlatformConstants`

GVFS/GVFS.Common/ProcessHelperImpl.cs
```diff
@@ -0,0 +1,16 @@
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Default product implementation of IProcessHelper
+    /// interface. Delegates calls to static ProcessHelper class. This
+    /// class can be used to enable testing of components that call
+    /// into the ProcessHelper functionality.
+    /// </summary>
+    public class ProcessHelperImpl : IProcessHelper
```
mjcheetham: How much effort is it to also remove the `static class ProcessHelper` fully?

GVFS/GVFS.Platform.Mac/MacServiceProcess.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Platform.Mac
+{
+    /// <summary>
+    /// Class to query the configured services on macOS
+    /// </summary>
+    public class MacServiceProcess
```
mjcheetham: I would name this something like `MacDaemonManager` because:

1. "service" is the Windows term.. the *NIX term is "daemon".
2. `MacServiceProcess` to me says this class models the Mac daemon process, but it actually wraps interactions with `launchctl`

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -176,6 +176,11 @@ public override bool IsGitStatusCacheSupported()
             return false;
         }
 
+        public override string GetProgramLocater()
```
mjcheetham: Feels like the `ProgramLocator` should instead be a type, rather than a string name of a program to execute.

```csharp
interface IExecutableLocator
{
    /// <summary>
    /// Get the full path to the given executable, or return `null` if not found.
    /// </summary>
    string Locate(string executableName);

// or if we want to use the Try* pattern...

    /// <summary>
    /// Try and get the full path to the given executable
    /// </summary>
    bool TryLocate(string executableName, out string fullPath);
}

class (Posix|Windows)ExecutableLocator : IExecutableLocator {...}
```

The implementation on Windows might shell out to `where` and on POSIX platforms to `which`, and might also augment the search paths, or call `which -a` to get the full list, and prefer a particular version over another.

GVFS/GVFS.Common/ProcessHelperImpl.cs
jamill: We have ~100 references to `ProcessHelper`, not impossible, but still a bit of a change.

GVFS/GVFS.Platform.Mac/MacServiceProcess.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public MacServiceProcess(IProcessHelper processHelper)
+        {
+            this.processHelper = processHelper;
+        }
+
+        public bool TryGetServices(out List<ServiceInfo> services, out string error)
+        {
+            ProcessResult result = this.processHelper.Run(LaunchCtlPath, LaunchCtlArg, true);
```
alameenshah: Does this return expected results when `gvfs upgrade --confirm` is run elevated? `GVFS.Service` on the Mac is run per user, so running as root might return a list with no `GVFS.Service` in it (unless root user has logged-in as well through login window.)

jamill: hmm... that might be the case... I will investigate this more - do you have suggestions for an alternative here?

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -106,6 +106,10 @@ public abstract class GVFSPlatformConstants
 
             public abstract string GVFSExecutableName { get; }
 
+            public abstract string ProgramLocaterCommand { get; }
```
jamill: This should probably be `Locator` to be consistent

GVFS/GVFS.Common/IProcessHelper.cs
```diff
@@ -0,0 +1,12 @@
+namespace GVFS.Common
+{
+    /// <summary>
+    /// Interface around process helper methods. This is to enable
+    /// testing of components that interact with the ProcessHelper
+    /// static class.
+    /// </summary>
+    public interface IProcessHelper
```
kewillford: I would name this `IProcessRunner` or `IProcessLauncher`

GVFS/GVFS.Common/ProcessHelper.cs
```diff
@@ -62,9 +62,9 @@ public static bool IsDevelopmentVersion()
             return currentVersion.Major == 0;
         }
 
-        public static string WhereDirectory(string processName)
+        public static string WhereDirectory(string programLocator, string processName)
```
kewillford: Can we name this something like `GetProgramLocation`?

GVFS/GVFS.Common/ProcessHelperImpl.cs
kewillford: I would name this `ProcessRunner` or `ProcessLauncher` and if we add more to it rename it but since all it is doing is one thing right now we can use a specific name for the class.

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Platform.Mac
+{
+    /// <summary>
+    /// Class to query the configured services on macOS
+    /// </summary>
+    public class MacDaemonManager
```
kewillford: `Manager` doesn't seem like the right term here since it is only for getting a list of the daemons.  Also don't think we need `Mac` in the name since it could possibly be used for Linux?  What about `DaemonList`?

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        private const string LaunchCtlPath = @"/bin/launchctl";
+        private const string LaunchCtlArg = @"list";
+
+        private IProcessHelper processHelper;
+
+        public MacDaemonManager()
+        : this(new ProcessHelperImpl())
```
kewillford: Can we get by with deleting this constructor and have the caller create the dependency and pass it in?

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        private bool TryParseOutput(string output, out List<DaemonInfo> daemonInfos, out string error)
+        {
+            daemonInfos = new List<DaemonInfo>();
+
+            // 1st line is the header, skip it
+            foreach (string line in output.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries).Skip(1))
+            {
+                // The expected output is a list of tab delimited entried:
+                // PID\tSTATUS\tLABEL
```
kewillford: How likely is the output of this to change?

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
```diff
@@ -0,0 +1,79 @@
...
+
+                if (tokens.Length != 3)
+                {
+                    daemonInfos = null;
+                    error = $"Unexpected number of tokens in line: {line}";
+                    return false;
+                }
+
+                string label = tokens[2];
+                bool isRunning = int.TryParse(tokens[0], out _);
```
kewillford: Do we ever need to check the status for things like stopping?

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -199,6 +199,16 @@ public override string GVFSExecutableName
...
             }
+
+            public override string ProgramLocaterCommand
+            {
+                get { return "which"; }
+            }
+
+            public override HashSet<string> UpgradeBlockingProcesses
+            {
+                get { return new HashSet<string> { "GVFS.Mount", "git", "wish" }; }
```
kewillford: Why is this a different list of processes?  Before it was shared correct and now they are separate and different than the list before this change.  Does it hurt anything to have the same list?  Also I was wondering if these `HashSet`s should be case insensitive?

GVFS/GVFS.UnitTests/Platform.Mac/MacDaemonManagerTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace GVFS.UnitTests.Platform.Mac
+{
+    [TestFixture]
+    public class MacServiceProcessTests
```
kewillford: Can we add any tests for `IsServiceInstalledAndRunning`?

GVFS/GVFS.UnitTests/Platform.Mac/MacDaemonManagerTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        [TestCase]
+        public void CanGetServices()
```
kewillford: Do we need any failure tests like a `ProcessResult` that is an error?

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
```diff
@@ -34,7 +34,7 @@
         
         function GetBlockingProcesses()
         {
-            var watchList = [ "gvfs", "GVFS.Mount", "git", "gitk", "Wish" ];
```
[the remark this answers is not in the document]
jamill: On the Mac, we can run the installer while the ‘GVFS upgrade’ verb is running - as do not have the restriction we have on windows where we cannot delete the application while it is running. This allows us to run upgrade in the same window as the upgrade verb, and not in a separate window.

This means we need to allow running the installer while the GVFS upgrade command is running. We can come back and tighten up this logic if needed (maybe check parent processes?).

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
jamill: I think this is stable. I thought about following up with a functional test...

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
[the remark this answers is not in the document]
kewillford: What about if I run some gvfs command that could take a while like `gvfs dehydrate` and run `gvfs upgrade` in another prompt?  Will it be okay?  Will the mount get ripped out from under the dehydrate?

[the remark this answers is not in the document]
jamill: Good points - 

`GVFS.Mount` is still a blocking process - so if there is a mount process running, we will not run the installer.

More generally, I will note this in the list of things to improve on. For instance, maybe:

1. The GVFS upgrade verb can check for other running GVFS commands
2. The upgrade process can check to only not block for `GVFS upgrade` verbs (if we can get the command line args for the command),
3. The upgrade process can check and not run if a GVFS command that is not in a parent process is running (also - I need to investigate if this is possible).

I think for the initial version, it is OK to not block for running GVFS commands (and provide instructions for initial users to not run upgrade while running other GVFS commands).

> What about if I run some gvfs command that could take a while like gvfs dehydrate and run gvfs upgrade in another prompt?

I wonder if it is ok to run `GVFS unmount` while the repository is being dehydrated? If not, should the `gvfs unmount` command not run if it is not safe?

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
jamill: `MacDaemonLister`?

> Also don't think we need Mac in the name

This implementation is specific to macOS (I think `launchctl` is more of a macOS artifact than Linux).

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
jamill: > I was wondering if these HashSets should be case insensitive?

probably! I will update this.

I will add more comments around the motivation for having different lists. The commit message includes this, but I should include it in the PR description / code comments.

GVFS/GVFS.Common/GVFSPlatform.cs
jamill: Made everything `Locater` instead

GVFS/GVFS.Common/IProcessHelper.cs
jamill: Renamed this to `IProcessRunner`

GVFS/GVFS.Platform.Mac/MacDaemonManager.cs
jamill: I don't think we need this at the moment...

GVFS/GVFS.UnitTests/Platform.Mac/MacDaemonManagerTests.cs
jamill: Noted for followup.

GVFS/GVFS.UnitTests/Platform.Mac/MacDaemonManagerTests.cs
jamill: Noted for followup

GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
jamill: Noted for Followup

GVFS/GVFS.Common/ProcessHelperImpl.cs
jamill: Noted for followup

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -106,6 +106,20 @@ public abstract class GVFSPlatformConstants
...
+
+            /// <summary>
+            /// Different platforms can have different requirements
+            /// around which processes can block upgrade. For example,
+            /// on Windows, we will block upgrade if any GVFS commands
+            /// are running, but on POSIX platforms, we relaxe this
```
mjcheetham: typo:

```suggestion
            /// are running, but on POSIX platforms, we relax this
```

GVFS/GVFS.Common/ProcessHelper.cs
```diff
@@ -62,9 +62,9 @@ public static bool IsDevelopmentVersion()
             return currentVersion.Major == 0;
         }
 
-        public static string WhereDirectory(string processName)
+        public static string GetProgramLocation(string programLocaterCommand, string processName)
```
mjcheetham: Since we're not going with the `ILocator` component idea (just yet), we might just want to replace `string programLocatorCommand` with `GVFSPlatform.Instance.Constants.ProgramLocatorCommand` since this method is parsing output which is specific to the locator command, and also in practice this method is _always_ going to be called with `GVFS.Platform.Instance.Constants.ProgramLocatorCommand`.

GVFS/GVFS.Platform.Mac/MacDaemonLister.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace GVFS.Platform.Mac
+{
+    /// <summary>
+    /// Class to query the configured services on macOS
+    /// </summary>
+    public class MacDaemonLister
```
mjcheetham: I'm still not liking this name.. "Lister". What is the aversion to "Manager" or "Controller"?
At the moment all we're implementing is 'list', but why not 'stop' or 'restart'? You wouldn't expect a "MacDaemonStopper" class.

GVFS/GVFS.Common/ProcessHelper.cs
jamill: The problem is that this `ProcessHelper` class is included in the Hooks projects, which do not have access to `GVFSPlatform` - so it has to be passed in from callers who do (always) have access to `GVFSPlatform`

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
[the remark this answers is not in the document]
kewillford: >I wonder if it is ok to run GVFS unmount while the repository is being dehydrated? If not, should the gvfs unmount command not run if it is not safe?

There is some protection with the lock and not running if an unmount is in process or another command has the lock but there could be some gaps in that logic.  Not sure if we have much testing around this for each of the commands.  But upgrade is kind of outside that but maybe the upgrade will fail when it tries to unmount the repos and something is already running that has the lock?

GVFS/GVFS.Platform.Mac/MacDaemonLister.cs
jamill: Yeah... I don't really have an aversion to any of these names...

I wanted this to be semi-similar to [ServiceController](https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontroller?view=netframework-4.8) in spirit, which seems to point towards `MacDaemonController` or `DaemonController` if it were targeting at POSIX.

I don't think we have any immediate need to flesh it out the functionality at this point... I am fine with either its current name for current functionality, or with a more aspirational name for if we ever expand it.

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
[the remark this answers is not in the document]
jamill: I will note this for follow up (I will update the MVP issue with a list of deferred items, and might copy that into a new issue for post MVP items).

PR #1239, opened 2019-06-05
NuGetFeed: only encrypt in memory credential  on supported platforms

jamill:
The NuGet Client API has flags to control whether a password is stored
in clear text or not. This flag controls:

  1) Whether the password is stored in clear text when persisted to a
  config file on disk.

  2) Whether the password is stored encrypted in memory

Encrypting the password is only supported on Windows (and Mono)
platforms, and not on netcore platfroms.

As VFS for Git does not actually update any configuration files, the
flag only controls how the password is stored in memory. For netcore
platforms, do not set this flag. As VFS for Git does not persist this
data to disk, the tradeoff is whether the process is encrypting the
password in memory or not. VFS for Git itself is working with the
plain text password, so this is not broading the risk in this aspect.

### DIFF
--- a/GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
+++ b/GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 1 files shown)


jamill:
I am open to feedback on this, including if I am going about this in the completely wrong way. One option is to just make the web requests ourselves and not use the NuGet client APIs.

nickgra:
What kind of 'password' do we have in memory? Is this going to be the PAT that the GCM stores? Or are we going to have a flow where this could be an authorization token that is generated with a PAT (or via a GUI pop up dialog box like ADAL can create)? I would hope this isn't someone's actual AD credential :)

jamill:
> What kind of 'password' do we have in memory? Is this going to be the PAT that the GCM stores? Or are we going to have a flow where this could be an authorization token that is generated with a PAT (or via a GUI pop up dialog box like ADAL can create)? I would hope this isn't someone's actual AD credential :)

This is going to be whatever credential the GCM returns. Currently, for the GCM that ships with VFS for Git, this is a PAT.

Issue #942, opened 2019-03-18
Track what processes create folder placeholders, and log how many folders were removed during projection changes


wilbaker:
Currently VFS4G tracks which processes have been creating file placeholders (and records the top 10 in heartbeats), but it does not track what processes are creating folder placeholders.

VFS4G should start tracking both.

nickgra:
Is this a dupe of #490? @wilbaker

wilbaker:
Yep good call @nickgra.  I will close the older issue as this issue has a few more details.

nickgra:
Picking this one up per my offline conversation with @jeschu1 and @wilbaker (I have a hacky version that will be cleaned up before PR).

nickgra:
> Additionally, VFS4G should record how many empty directories it was able to delete when it updates placeholders as part of a projection change.

I have split this out into a separate issue. #1255

PR #1241, opened 2019-06-06
Update heartbeats to contain folder placeholder and file hydration metadata

### LINKED ISSUE #1172: Track what processes hydrate files and log them in heartbeats, like we do for placeholders
As a companion to #942, we should consider logging which processes are hydrating files. 
It's possible that customers could have an experience where a process/service is hydrating files and is impacting their experience, but if we had this data in the heartbeat, we could diagnose and identify this issue.

nickgra:
Fixes #1172 #942 

Update our heartbeats to contain metadata on what processes are creating folder placeholders and how many were created since the last heartbeat. 

While we're here, let's do the same thing for file hydration to get ahead of any potential issues in the future.

I've restructured the format a bit to make it more readable with the new fields:
```
[2019-06-05 18:03:27 -07:00] Heartbeat {"FilePlaceholderCreation":{"ProcessName1":"\\Device\\HarddiskVolume4\\Users\\nickgra\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe","ProcessCount1":23},"FolderPlaceholderCreation":{"ProcessName1":"\\Device\\HarddiskVolume4\\Users\\nickgra\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe","ProcessCount1":1371},"FilePlaceholdersHydrated":{"ProcessName1":"\\Device\\HarddiskVolume4\\Users\\nickgra\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe","ProcessCount1":4},"ModifiedPathsCount":1,"FilePlaceholderCount":23,"FolderPlaceholderCount":1371,"GitStatusCache.StatusScanCount":1,"GitStatusCache.CacheReadyCount":1,"EnlistmentId":"349f2ee1c40e4c79ab7482ad5966aec8","PhysicalDiskInfo":{"DriveLetter":"E","VolumeDriveType":"Fixed","VolumeFileSystem":"NTFS","VolumeFileSystemLabel":"Data","VolumeSize":"1024073920512","VolumeSizeRemaining":"707978559488"},"Version":"0.2.173.2","MinutesUptime":0,"MinutesSinceLast":0}
```

> [2019 - 06 - 05 18: 03: 27 - 07: 00] Heartbeat {
    "FilePlaceholderCreation": {
        "ProcessName1": "\\Device\\HarddiskVolume4\\Users\\nickgra\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe",
        "ProcessCount1": 23
    },
    "FolderPlaceholderCreation": {
        "ProcessName1": "\\Device\\HarddiskVolume4\\Users\\nickgra\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe",
        "ProcessCount1": 1371
    },
    "FilePlaceholdersHydrated": {
        "ProcessName1": "\\Device\\HarddiskVolume4\\Users\\nickgra\\AppData\\Local\\Programs\\Microsoft VS Code\\Code.exe",
        "ProcessCount1": 4
    },
    "ModifiedPathsCount": 1,
    "FilePlaceholderCount": 23,
    "FolderPlaceholderCount": 1371,
    "GitStatusCache.StatusScanCount": 1,
    "GitStatusCache.CacheReadyCount": 1,
    "EnlistmentId": "349f2ee1c40e4c79ab7482ad5966aec8",
    "PhysicalDiskInfo": {
        "DriveLetter": "E",
        "VolumeDriveType": "Fixed",
        "VolumeFileSystem": "NTFS",
        "VolumeFileSystemLabel": "Data",
        "VolumeSize": "1024073920512",
        "VolumeSizeRemaining": "707978559488"
    },
    "Version": "0.2.173.2",
    "MinutesUptime": 0,
    "MinutesSinceLast": 0
}

To-Do:
- [x] Performance analysis (I think this impacts hot paths)
- [x] Validate behavior on macOS


GVFS/GVFS.Common/HeartbeatThread.cs
```diff
@@ -6,7 +6,7 @@ namespace GVFS.Common
 {
     public class HeartbeatThread
     {
-        private static readonly TimeSpan HeartBeatWaitTime = TimeSpan.FromMinutes(60);
+        private static readonly TimeSpan HeartBeatWaitTime = TimeSpan.FromSeconds(30);
```
nickgra: To be reverted after testing.

nickgra:
Some of these files were normalized to CRLF since they had mixed line endings. I can revert those before submitting this, but perhaps this is a good sign that I should pick up the work to configure our editorconfig and normalize the mess that VSMac made :)

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -175,6 +175,57 @@ public void Remove(string path)
...
         }
 
+        public int GetFilePlaceholdersCount()
+        {
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    command.CommandText = $"SELECT count(path) FROM Placeholder WHERE pathType = {(int)PlaceholderData.PlaceholderType.File};";
```
jeschu1: I think what you have is probably the way to go.

There could be another option maybe to do:
    SELECT count(path) FROM Placeholder GROUP BY pathType
That should negate the need for 3 separate queries.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -522,16 +522,16 @@ public FileSystemTaskResult AddMissingModifiedFiles()
                     using (ITracer activity = this.context.Tracer.StartActivity(
                         nameof(this.indexParser.AddMissingModifiedFilesAndRemoveThemFromPlaceholderList),
                         EventLevel.Informational))
-                    {
```
jeschu1: I'm definitely on team fix line endings!

Since there are no actual changes in this file though, I would remove it from the commit.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -259,30 +263,20 @@ public bool IsReadyForExternalAcquireLockRequests(NamedPipeMessages.LockData req
...
-
-                    metadata.Add("ProcessName" + count, processCount.Key);
-                    metadata.Add("ProcessCount" + count, processCount.Value.Count);
-                }
-
-                eventLevel = EventLevel.Informational;
-            }
+            metadata.Add(
+                "FilePlaceholderCreation",
+                this.GetProcessInteractionData(ref this.filePlaceHolderCreationCount, ref eventLevel));
```
jeschu1: Why do these need to passed by reference?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -259,30 +263,20 @@ public bool IsReadyForExternalAcquireLockRequests(NamedPipeMessages.LockData req
...
+                this.GetProcessInteractionData(ref this.folderPlaceHolderCreationCount, ref eventLevel));
+            metadata.Add(
+                "FilePlaceholdersHydrated",
+                this.GetProcessInteractionData(ref this.fileHydrationCount, ref eventLevel));
 
             metadata.Add("ModifiedPathsCount", this.modifiedPaths.Count);
-            metadata.Add("PlaceholderCount", this.placeholderDatabase.GetCount());
+            metadata.Add("FilePlaceholderCount", this.placeholderDatabase.GetFilePlaceholdersCount());
+            metadata.Add("FolderPlaceholderCount", this.placeholderDatabase.GetFolderPlaceholdersCount());
+
```
jeschu1: GetExpandedFolderCount? too?

jeschu1: I don't see where GetExpandedFolderCount is used, if we don't need it maybe drop it?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -506,6 +513,35 @@ private static bool CheckConditionWithRetry(Func<bool> predicate, int retries, i
...
+                    if (count > 10)
+                    {
+                        break;
+                    }
+
+                    metadata.Add("ProcessName" + count, processCount.Key);
+                    metadata.Add("ProcessCount" + count, processCount.Value.Count);
+                }
+
+                eventLevel = EventLevel.Informational;
```
jeschu1: Why do we need to change the event level?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
mjcheetham: It appears the `GetProcessInteractionData` method is setting the concurrent dictionary reference to a new instance before iterating on it, effectively resetting the dictionary each time.

Why are we doing this on the `EventMetadata.Add`?
If you do want to clear the various dictionaries I would make it explicit and obvious by calling `.Clear` on each dictionary here in _this_ method, not hidden inside the `GetProcessInteractionData` method.

Methods named "get" should probably not mutate things.

jeschu1: @mjcheetham .Clear suggestion would also negate the need to pass by ref.

mjcheetham:
```csharp
metadata.Add(
    "FilePlaceholderCreation",
    this.GetProcessInteractionData(this.filePlaceHolderCreationCount));
this.filePlaceHolderCreationCount.Clear();

// .. repeated for the other dictionaries ..

eventLevel = EventLevel.Informational;
```

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -506,6 +513,35 @@ private static bool CheckConditionWithRetry(Func<bool> predicate, int retries, i
...
+            EventMetadata metadata = new EventMetadata();
+
+            if (collectedData.Count > 0)
+            {
+                ConcurrentDictionary<string, PlaceHolderCreateCounter> localData = collectedData;
+                collectedData = new ConcurrentDictionary<string, PlaceHolderCreateCounter>(StringComparer.OrdinalIgnoreCase);
+
+                int count = 0;
+                foreach (KeyValuePair<string, PlaceHolderCreateCounter> processCount in
+                    localData.OrderByDescending((KeyValuePair<string, PlaceHolderCreateCounter> kvp) => kvp.Value.Count))
```
jeschu1: foreach (KeyValuePair<string, PlaceHolderCreateCounter> processCount in 
                    localData.OrderByDescending((KeyValuePair<string, PlaceHolderCreateCounter> kvp) => kvp.Value.Count).Take(10))

Then you don't need the count variable.

jeschu1:
@nickgra Fantastic job on this so far!   I left so pretty minor comments.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
wilbaker: > Methods named "get" should probably not mutate things.

Agreed, this method as a whole could be better named as well (e.g. `GetAndResetHeartBeatMetadata`)

> If you do want to clear the various dictionaries I would make it explicit and obvious by calling `.Clear`

Some background info on this discussion:

The idea of using a swap here rather than `Clear` was that so if threads were adding new entries to `placeHolderCreationCount` while this method was processing the collection we'd reduce our chances of dropping anything new that was added (before calling `Clear`), and this method could take its time in processing the data.

I just tool a look and saw that `placeHolderCreationCount` is not currently marked as `volatile` though, and so even with the current code we could drop some new entries.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
nickgra: Good catch, I thought this would be interesting data to have, but upon further thought, I am not going to use it.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
nickgra: Thanks for all the feedback here. I renamed some methods here and reworked how we reason about the EventLevel (which controls if we actually log this message to the mount log file).

William called out why we have the current 'swap' approach for the dictionaries and that there is still a race condition in our current approach. While the Clear approach is more explicit in our intent, I believe it would widen the window of the race condition, so I'd like to keep the swap approach as is for now.

I renamed methods here to try and make things more explicit. What does everyone think now? If it's still confusing, I can switch to `Clear`.

wilbaker: I think the way you've cleaned up the code looks good, the new method names make it easier to follow what's going on.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -506,6 +518,34 @@ private static bool CheckConditionWithRetry(Func<bool> predicate, int retries, i
             return result;
         }
 
+        private EventMetadata GetProcessInteractionData(ConcurrentDictionary<string, PlaceHolderCreateCounter> collectedData, ref bool logToFile)
+        {
+            EventMetadata metadata = new EventMetadata();
+
+            if (collectedData.Count > 0)
+            {
+                int count = 0;
```
jeschu1: I don't think you need the count variable at all.  Whippee less code! :-)

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -506,6 +518,34 @@ private static bool CheckConditionWithRetry(Func<bool> predicate, int retries, i
...
+
+                logToFile = true;
+            }
+
+            return metadata;
+        }
+
+        private ConcurrentDictionary<string, PlaceHolderCreateCounter> GetAndResetProcessCountMetadata(ref ConcurrentDictionary<string, PlaceHolderCreateCounter> collectedData)
+        {
+            ConcurrentDictionary<string, PlaceHolderCreateCounter> localData = collectedData;
```
jeschu1: I understand *why* we are doing this, but I think it is confusing.

Maybe a helpful comment at the top like:
// Captures the current state of dictionary, and resets it
// This approach is optimal for our use case to preserve all entries while avoiding additional locking

Because what I think you want in a perfect world is to lock the 3 dictionaries why you examine the data.  But I can see that being a pain and more code verbosity/complexity.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -256,36 +260,31 @@ public bool IsReadyForExternalAcquireLockRequests(NamedPipeMessages.LockData req
...
-                        break;
-                    }
-
-                    metadata.Add("ProcessName" + count, processCount.Key);
-                    metadata.Add("ProcessCount" + count, processCount.Value.Count);
-                }
 
-                eventLevel = EventLevel.Informational;
-            }
+            ConcurrentDictionary<string, PlaceHolderCreateCounter> storedProcessData = this.GetAndResetProcessCountMetadata(ref this.filePlaceHolderCreationCount);
```
jeschu1: nit: I find the use of storedProcessData a bit hard to follow.  Can we avoid this variable and just do

`this.GetProcessInteractionData(this.GetAndResetProcessCountMetadata(ref this.filePlaceHolderCreationCount), ref logToFile));`

Same comment for other two instances

jeschu1:
@nickgra , I left some minor final feedback.  Nice job on this, you can steal my tickets anytime! :-D

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
nickgra: We append the count onto `ProcessName` to sort processes by the ones that are the most active to least active. Now that the EventMetadata is nested, maybe it's not needed?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -256,36 +260,28 @@ public bool IsReadyForExternalAcquireLockRequests(NamedPipeMessages.LockData req
...
-
-                    metadata.Add("ProcessName" + count, processCount.Key);
-                    metadata.Add("ProcessCount" + count, processCount.Value.Count);
-                }
 
-                eventLevel = EventLevel.Informational;
-            }
+            metadata.Add(
+                "FilePlaceholderCreation",
+                this.GetProcessInteractionData(this.GetAndResetProcessCountMetadata(ref this.filePlaceHolderCreationCount), ref logToFile));
```
jeschu1: Another thought here.

Do we really need to clear the list between each heartbeat?  Would it be more beneficial to keep a running total of largest hydrators across the mount?  You can get all the same information with either approch but a running counter would be simplier.  Both approaches would have pros/cons.   @wilbaker

wilbaker: The historical reasons for clearing each heartbeat:

- We only want to write the heartbeat to the log file when something has changed, and starting from scratch for each heartbeat makes it easier to know if anything has changed in subsequent heartbeats.

- When helping out users you'll be able to immediately tell which processes created placeholders in each time slice (without having to search for older heartbeats and calculate the new values).

- Our main customer was processing this data as well, and resetting it each heartbeat was helpful for the diagnostics/reporting.

My personal preference is to continue to reset each heartbeat, because I've found it easier to reason about what happened when I look at the log files.

PR #1242, opened 2019-06-06
Mac: Add FileSystemType

### LINKED ISSUE #1134: Mac: Record filesystem type (APFS / HFS+) in telemetry stream
Run performance testing using both, and note the differences

jeschu1:
Adds Disk information to heartbeats for Mac.

resolves #1134

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacPlatform.cs
+++ b/GVFS/GVFS.Platform.Mac/MacPlatform.cs
@@ -2,7 +2,11 @@
 using GVFS.Common.FileSystem;
 using GVFS.Common.Tracing;
 using GVFS.Platform.POSIX;
+using System.Collections.Generic;
 using System.IO;
+using System.Xml;
+using System.Xml.Linq;
+using System.Xml.XPath;
 
 namespace GVFS.Platform.Mac
 {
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -32,36 +36,62 @@ public override string GetDataRootForGVFS()
...
         {
             return new MacFileBasedLock(fileSystem, tracer, lockPath);
-        }
-
+        }
+
+        public override Dictionary<string, string> GetPhysicalDiskInfo(string path, bool sizeStatsOnly)
+        {
+            // DiskUtil will return disk statistics in xml format
+            ProcessResult processResult = ProcessHelper.Run("diskutil", "info -plist /", true);
```
jeschu1: Note this opens the door for us also recording things like volume size, size used, remaining, etc..

@mjcheetham of all the fancy ways to get the info from diskutil, I think parsing the xml seems like the cleanest.

mjcheetham: The XDocument streaming APIs are very fast and reliable from my experience.
(so long as you take precautions around schema validation etc)

jeschu1:
> Looks good to me, except.. all those whitespace changes? I cannot tell in GitHub's UI.. are they removing or adding trailing whitespace? If it's removing 😄  otherwise 😢

Thanks @mjcheetham !  The diff is line endings.  A lot of the .cs mac files seemed to have a mixture.  Saving in VS Code makes them all ^M style endings which last I checked was what we want.  So this fixes them.  I can (with some pain) not change those line endings but it will be pain again the next time the file gets updated.

PR #1246, opened 2019-06-06
update Mac build steps to install Java first

seanknox:
Moved the Java install step before the git-credential-manager install step.

The Mac build instructions currently instruct the user to install `git-credential-manager` via Homebrew first and then install Java, but doing that fails:

```
git-credential-manager: Java 1.6+ is required to install this formula.
Install AdoptOpenJDK with Homebrew Cask:
  brew cask install adoptopenjdk
```


jrbriggs:
We're actively trying to replace that GCM with a new one, which is already available to internal customers, and will be available to external customers within the sprint. To avoid further confusing the issue, I'm not going to take this PR.

PR #1248, opened 2019-06-06
LooseObjectsStep: report full object cache size

derrickstolee:
As the PackfileMaintenanceStep is now running in production, I noticed a user's repo was growing significantly between runs. It turns out this user had many loose objects and the packs created by the LooseObjectsStep were large.

Add extra object size checks and pack-size checks to the LooseObjectsStep so we can report the total effect on the object cache's contents.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -41,9 +41,10 @@ namespace GVFS.Common.Maintenance
         protected override string LastRunTimeFilePath => Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", LooseObjectsLastRunFileName);
         protected override TimeSpan TimeBetweenRuns => TimeSpan.FromDays(1);
 
-        public int CountLooseObjects()
+        public void CountLooseObjects(out int count, out long size)
         {
-            int count = 0;
+            count = 0;
+            size = 0;
 
             foreach (string directoryPath in this.Context.FileSystem.EnumerateDirectories(this.Context.Enlistment.GitObjectsRoot))
             {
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.UnitTests/Maintenance/LooseObjectStepTests.cs
```diff
@@ -143,9 +143,10 @@ public void LooseObjectsCount()
             this.TestSetup(DateTime.UtcNow.AddDays(-7));
 
             LooseObjectsStep step = new LooseObjectsStep(this.context, requireCacheLock: false, forceRun: false);
-            int count = step.CountLooseObjects();
+            step.CountLooseObjects(out int count, out long size);
 
             count.ShouldEqual(3);
+            size.ShouldEqual(11);
```
jeschu1: I think a small comment on how we got to 3 / 11, would be helpful for readability

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -211,16 +213,27 @@ protected override void PerformMaintenance()
                         }
                     }
 
-                    int beforeLooseObjectsCount = this.CountLooseObjects();
+                    this.CountLooseObjects(out int beforeLooseObjectsCount, out long beforeLooseObjectsSize);
+                    this.GetPackFilesInfo(out int beforePackCount, out long beforePackSize, out bool _);
                     GitProcess.Result gitResult = this.RunGitCommand((process) => process.PrunePacked(this.Context.Enlistment.GitObjectsRoot), nameof(GitProcess.PrunePacked));
```
jamill: I don't think this `gitResult` is being used anywhere - is there any decisions we should make on the result (success or failure) of this operation or record the result to telemetry?

PR #1250, opened 2019-06-07
Mac: Switch on the UTF8 tests

nickgra:
When we were standing up the Functional Tests on macOS, we saw failures in tests that involved UTF-8 strings on the Azure Pipelines hosted macOS agents which we could never repro locally. Now that they've moved to a newer image, these tests now pass on both our local dev machines, our dedicated functional test machines and the Azure Pipelines hosted machines. 

Let's enable the tests.


jeschu1:
Looks like this works, any reason why it's still in draft?

PR #1257, opened 2019-06-07
Mac: Bash Functional Test Hang

### LINKED ISSUE #554: Mac: Investigate why BashRunner.DeleteDirectoryWithUnlimitedRetries is hanging only in GVFSFunctionalTestEnlistment
In an earlier PR, @wilbaker discovered this hang when standing up GVFS.Hooks on MacOS. Upon doing some code cleanup, I tried folding all uses of this code together. This worked fine for the uses in SharedCacheTests and FastFetchTests, but not for GVFSEnlistmentTests.

jeschu1:
Add '''s to directory being accessed

Previously we were trying to delete a directory with a space indefinitely.  By adding 's around the directory it is deleted successfully.  While we're in here ensure all paths are in 's.

Resolves #554


wilbaker:
Please set the appropriate labels and milestone before merging

PR #1260, opened 2019-06-10
Test Rename File Attribute (For Testing Only)

jeschu1:
New MessageType for DeleteFromRename

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -753,6 +753,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize)
             
         case MessageType_KtoU_NotifyFileModified:
         case MessageType_KtoU_NotifyFilePreDelete:
+        case MessageType_KtoU_NotifyFilePreDeleteFromRename:
         case MessageType_KtoU_NotifyDirectoryPreDelete:
         case MessageType_KtoU_NotifyFilePreConvertToFull:
         {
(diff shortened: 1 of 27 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1125,7 +1125,7 @@ static PrjFS_Result HandleFileNotification(
     // We have to convert to full for PreDelete until we have WILL_RENAME support.  The PreDelete could be for a rename.
```
wilbaker: This comment should be updated to explain why we convert to full for `PreDeleteFromRename` (or it should be removed)

jeschu1:
@wilbaker can you have a look at the last commit, before I cherrypick the 2 commits over to @pmj 's PR and merge?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1122,10 +1123,10 @@ static PrjFS_Result HandleFileNotification(
         notificationType,
         nullptr /* destinationRelativePath */);
     
-    // We have to convert to full for PreDelete until we have WILL_RENAME support.  The PreDelete could be for a rename.
+    // We have to convert to full for PrjFS_NotificationType_PreDeleteFromRename in order to move the file
```
wilbaker: Minor:  This comment isn't quite accurate, as the file could be moved even if we don't convert it to full.

What if we went with something like:

```
// Remove placeholder xattrs for renames (PreDeleteFromRename) because:
//   - Renames are treated as "delete old path" + "create new path", and new files should not be marked as placeholders
//   - The target of the rename might be outside of the root, and we should not allow placeholder files to 
//     live outside of virtualization roots (as the kext won't know what provider they below to)

PR #1263, opened 2019-06-10
Log Folder Placeholders Removed

### LINKED ISSUE #1255: Record how many empty directories are deleted during update placeholders
See #942 for more context

VFS4G should record how many empty directories it was able to delete when it updates placeholders as part of a projection change.

jeschu1:
resolves #1255

### DIFF
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1117,6 +1117,8 @@ namespace GVFS.Virtualization.Projection
                 this.blobSizes.Flush();
 
                 int deleteFolderPlaceholderAttempted = 0;
+                int folderPlaceholdersDeleted = 0;
+                int folderPlaceholdersPathNotFound = 0;
                 using (BlobSizes.BlobSizesConnection blobSizesConnection = this.blobSizes.CreateConnection())
                 {
                     // A hash of the placeholders is only required if the platform expands directories
@@ -1157,7 +1159,18 @@ namespace GVFS.Virtualization.Projection
                                 !isFolder ||
                                 folderPlaceholder.IsPossibleTombstoneFolder)
                             {
-                                keepFolder = !this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder);
+                                FSResult result = this.RemoveFolderPlaceholderIfEmpty(folderPlaceholder);
+                                if (result == FSResult.Ok)
+                                {
+                                    ++folderPlaceholdersDeleted;
+                                    keepFolder = false;
+                                }
+                                else if (result == FSResult.FileOrPathNotFound)
+                                {
+                                    ++folderPlaceholdersPathNotFound;
+                                    keepFolder = false;
+                                }
+
                                 ++deleteFolderPlaceholderAttempted;
                             }
 
@@ -1200,7 +1213,9 @@ namespace GVFS.Virtualization.Projection
                     millisecondsUpdatingFilePlaceholders,
                     millisecondsUpdatingFolderPlaceholders,
                     millisecondsWriteAndFlush,
-                    deleteFolderPlaceholderAttempted);
+                    deleteFolderPlaceholderAttempted,
+                    folderPlaceholdersDeleted,
+                    folderPlaceholdersPathNotFound);
             }
         }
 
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1165,6 +1166,10 @@ private void UpdatePlaceholders()
                             {
                                 this.AddParentFoldersToListToKeep(folderPlaceholder.Path, folderPlaceholdersToKeep);
                             }
+                            else
+                            {
+                                ++deleteFolderPlaceHolder;
```
kewillford: I would move this down after line 1182 since that is where we are actually deleting the placeholder from the database.  Also if we want as much information as possible I would log the `deleteFolderPlaceholderAttempted` and have a counter for the ReExpanded folders so we know how many of those as well?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1117,6 +1117,7 @@ private void UpdatePlaceholders()
                 this.blobSizes.Flush();
 
                 int deleteFolderPlaceholderAttempted = 0;
+                int deleteFolderPlaceHolder = 0;
```
wilbaker: Nit: Could you rename `deleteFolderPlaceHolder` to be slightly more descriptive?  A few ideas:

- `folderPlaceholdersDeleted`
- `deleteFolderPlaceholderSucceeded`

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1180,10 +1181,13 @@ private void UpdatePlaceholders()
                         {
                             existingPlaceholders?.Remove(folderPlaceholder.Path);
                             this.placeholderDatabase.Remove(folderPlaceholder.Path);
+                            ++deleteFolderPlaceHolder;
```
wilbaker: Currently this is counting both folder placeholders that were deleted by VFS4G, and those that did not exist (but VFS4G was trying to delete).  Do you think these counts should be combined?

cc: @kewillford for his thoughts as well

kewillford: @jeschu1 This depends on what we are trying to get out of this logging.  Are we trying to determine what is causing something to be slow by having a count on the folder placeholders?  Or is it so we know an can report on the number of empty folders that the user had?  There are multiple counts here that could be useful. 
1. The number of folder placeholder that we attempt to delete.  This was the cause of performance issues before we added the check for if it was in the projection.
2. The number of folder placeholders that actually get deleted.  I suspect in most cases now this will be equal to the attempts but would be interesting to know when and how often the are not.
3. The number of deletes that are performed on the database.  Again this would be interesting to see how this number compares with the others.

So I guess it all depends on what we will be using these number for in the logging.

wilbaker: To add some additional context here, I originally filed the issue because I found it difficult to figure where time was being spent during projection changes (and what VFS4G was doing).

I think the more details we can include the better, it will help with understanding what VFS4G was doing during a projection change.

jeschu1: I'm wondering if a better place for this data would be [here](https://github.com/microsoft/VFSForGit/blob/c6ab22dde56d91591b02026860ccb3e9990d6560/GVFS/GVFS.Common/GVFSLock.cs#L332).  @kewillford i'll get in touch regarding.

wilbaker: > I'm wondering if a better place for this data would be here

Is that data currently written to the log file as well?

jeschu1: > Is that data currently written to the log file as well?

It looks like it should be.  If it isn't I'll keep this as standalone telemetry.

> RemoveFolderPlaceholderIfEmpty

Do you think we should pull this apart to distinguish when this.fileSystemVirtualizer.DeleteFile gets OK versus FileOrPathNotFound.  @kewillford didnt' have strong feelings here so let me know what you think @wilbaker

wilbaker: @jeschu1 I think it would be worth distinguishing them.  Could we update `RemoveFolderPlaceholderIfEmpty` to return an enum value, and then use that to decide which count to increment?

jeschu1: @wilbaker @kewillford please give this a fresh look

@wilbaker I confirmed the added values go to the log

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1431,22 +1446,18 @@ private string GetNewProjectedShaForPlaceholder(string path)
...
-        /// <c>true</c>If the folder placeholder was deleted
-        /// <c>false</c>If RemoveFolderPlaceholderIfEmpty failed attempting to remove the folder placeholder
+        /// <returns> Result of trying to delete the PlaceHolder
         /// </returns>
-        private bool RemoveFolderPlaceholderIfEmpty(IPlaceholderData placeholder)
+        private FSResult RemoveFolderPlaceholderIfEmpty(IPlaceholderData placeholder)
         {
             UpdateFailureReason failureReason = UpdateFailureReason.NoFailure;
             FileSystemResult result = this.fileSystemVirtualizer.DeleteFile(placeholder.Path, FolderPlaceholderDeleteFlags, out failureReason);
             switch (result.Result)
```
kewillford: This could be changed into an `if` statement to log and always return the `result.Result` but would have a bigger diff so I think this is good for now.

jeschu1:
Thanks for the quick review and help @kewillford , @wilbaker do you want to give this a quick look before I merge?

Issue #1205, opened 2019-05-27
Mac: Frequent hangs/stalls during functional tests


pmj:
Mac functional test runs have occasionally been hitting the timeout of 180 minutes over the last few weeks, but either I've been extremely unlucky or they've become more frequent recently.

I've re-queued PR #1200 a few times now, each time hitting either #1204 or the timeout problem. The changes in the PR itself only affect kext unit tests which complete without issue according to the logs.

Recent example:  https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9557200&view=results (retained)

Previous runs seem to have disappeared from the log already, I have queued some new builds now and will retain and add them here if they time out.

As the functional tests [appear to still be running on macOS 10.13 (High Sierra)](https://github.com/microsoft/VFSForGit/pull/1193#issuecomment-495285780) could this be the kext-side IOSharedDataQueue issue yet again? (#161) If so, can we either upgrade the test runner machines or implement the kext-side workaround as well?

Note: #1165 may help with diagnosing this.

pmj:
Another one:

https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9571613&view=results

pmj:
https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9581752&view=results

pmj:
https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9583367&view=results

pmj:
https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=9609994&view=results

wilbaker:
@pmj if you notice this happen again when I'm online could you send me a message?  I'd like to to hop onto one of the machines while a hang is in progress.

wilbaker:
Closed by #1273

wilbaker:
(Fixed in master in #1276)

PR #1267, opened 2019-06-11
Add logging to clone/mount paths

wilbaker:
Part of the work required for #1205

- Add additional logging to help diagnose hangs during clone/mount

### DIFF
--- a/GVFS/GVFS.Platform.POSIX/ProcessLauncher.cs
+++ b/GVFS/GVFS.Platform.POSIX/ProcessLauncher.cs
@@ -29,6 +29,8 @@ namespace GVFS.Platform.POSIX
                     NetCoreMethods.AllocNullTerminatedArray(argv, ref argvPtr);
                     NetCoreMethods.AllocNullTerminatedArray(envp, ref envpPtr);
 
+                    tracer.RelatedInfo($"{nameof(StartBackgroundProcess)}: Forking process and starting {programName}");
+
                     // Fork the child process
                     int processId = Fork();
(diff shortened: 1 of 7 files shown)


GVFS/GVFS.Platform.POSIX/ProcessLauncher.cs
```diff
@@ -54,40 +56,29 @@ public static void StartBackgroundProcess(ITracer tracer, string programName, st
...
+            // daemons.
+            //
+            // If nochdir is zero, daemon() changes the process's current working
+            // directory to the root directory("/"); otherwise, the current working
+            // directory is left unchanged.
+            //
+            // If noclose is zero, daemon() redirects standard input, standard
+            // output and standard error to /dev/ null; otherwise, no changes are
+            // made to these file descriptors.
+            if (Daemon(nochdir: 1, noclose: 0) != 0)
```
jeschu1: Using Daemon is just a good general cleanup aside from the extra logging correct?  Was there any other reason for this?

wilbaker: It was just to move us one step closer to standard methods/functions and away from our own custom solution.

However, this is not strictly required for the logging and so I might just take that change out for now and focus on getting us onto `Process.Start`

wilbaker: @jeschu1 I decided to back this change out for now and just go with the extra logging.  I've leave the launching code as-is until we have a better fix for the hang.

PR #1268, opened 2019-06-11
NuGetUpgrader: fix issue with upgrader using interop services

jamill:
Instead of querying interop services at runtime (with a dll that is not
included with the windows install), determine this at compile time using the
under construction flags.

(cherry picked from commit fe04e7daf4705e2795157ccc11956d422d35c31b)

### DIFF
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -138,16 +138,19 @@ namespace GVFS.Common
             public UnderConstructionFlags(
                 bool supportsGVFSUpgrade = true,
                 bool supportsGVFSConfig = true,
-                bool requiresDeprecatedGitHooksLoader = false)
+                bool requiresDeprecatedGitHooksLoader = false,
+                bool supportsNuGetEncryption = true)
             {
                 this.SupportsGVFSUpgrade = supportsGVFSUpgrade;
                 this.SupportsGVFSConfig = supportsGVFSConfig;
                 this.RequiresDeprecatedGitHooksLoader = requiresDeprecatedGitHooksLoader;
+                this.SupportsNuGetEncryption = supportsNuGetEncryption;
             }
 
             public bool SupportsGVFSUpgrade { get; }
             public bool SupportsGVFSConfig { get; }
             public bool RequiresDeprecatedGitHooksLoader { get; }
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Common/NuGetUpgrade/NuGetFeed.cs
```diff
@@ -176,7 +177,7 @@ private void SetSourceRepository()
             this.sourceRepository = Repository.Factory.GetCoreV3(this.feedUrl);
             if (!string.IsNullOrEmpty(this.personalAccessToken))
             {
-                this.sourceRepository.PackageSource.Credentials = BuildCredentialsFromPAT(this.personalAccessToken);
+                this.sourceRepository.PackageSource.Credentials = BuildCredentialsFromPAT(this.personalAccessToken, !this.platformSupportsEncryption);
```
wilbaker: What's the advantage of having the `platformSupportsEncryption` member variable vs. using `GVFSPlatform.Instance.UnderConstruction.SupportsNuGetEncryption` here?

jamill: It could have gone there as well, but I wanted to keep the GVFSPlatform out of NuGetFeed (and not introduce static references in that component)

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -138,16 +138,19 @@ public class UnderConstructionFlags
...
                 this.SupportsGVFSUpgrade = supportsGVFSUpgrade;
                 this.SupportsGVFSConfig = supportsGVFSConfig;
                 this.RequiresDeprecatedGitHooksLoader = requiresDeprecatedGitHooksLoader;
+                this.SupportsNuGetEncryption = supportsNuGetEncryption;
             }
 
             public bool SupportsGVFSUpgrade { get; }
             public bool SupportsGVFSConfig { get; }
             public bool RequiresDeprecatedGitHooksLoader { get; }
+            public bool SupportsNuGetEncryption { get; }
```
wilbaker: As discussed offline, `UnderConstructionFlags` has been for flags where we have remaining work on our side for some platforms.  If we're waiting on NuGet to support encryption the flag/property could be moved out of `UnderConstructionFlags`.

Issue #1262, opened 2019-06-10
[TODO] Platform native notifications


alameenshah:
1. Add logging to file. Make the log file available to gvfs diagnose.
2. Simplify VFSProcessRunner:tryRunExecutable. https://github.com/microsoft/VFSForGit/pull/1176#discussion_r290927588

PR #1270, opened 2019-06-11
VFSForGit installer non-relocatable

alameenshah:
- Added component-plist file for "VFS For Git" package. This is an xml file that contains specifications for how installer should handle installing bundles in VFSForGit package. 
 It is created using the command - 
 `pkgbuild --analyze --root ../BuildOutput/GVFS.Installer.Mac/bin/x64/Debug/netcoreapp2.1/osx-x64/Staging vfsforgit_components.plist`.
 The file has all default flags except `BundleIsRelocatable` which is set to false, for "VFS For Git.app". At install time, this flag stops the installer from searching the target disc for a copy of the app (on developer machines it often times picks the copy from BuildOutput directory), which it then upgrades. Now the installer will always fresh-install (or upgrade) the specific path - `/Library/Application\ Support/VFS For Git/VFS For Git.app` only. 
- Also updated CreateInstaller script to use the component-plist file.

### DIFF
--- a/GVFS/GVFS.Installer.Mac/vfsforgit_components.plist
+++ b/GVFS/GVFS.Installer.Mac/vfsforgit_components.plist
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<array>
+	<dict>
+		<key>BundleHasStrictIdentifier</key>
+		<true/>
+		<key>BundleIsRelocatable</key>
+		<false/>
+		<key>BundleIsVersionChecked</key>
+		<true/>
+		<key>BundleOverwriteAction</key>
+		<string>upgrade</string>
+		<key>RootRelativeBundlePath</key>
+		<string>Library/Application Support/VFS For Git/VFS For Git.app</string>
+	</dict>
+	<dict>
+		<key>BundleIsVersionChecked</key>
+		<true/>
+		<key>BundleOverwriteAction</key>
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Installer.Mac/vfsforgit_components.plist
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<array>
+	<dict>
+		<key>BundleHasStrictIdentifier</key>
+		<true/>
+		<key>BundleIsRelocatable</key>
+		<false/>
+		<key>BundleIsVersionChecked</key>
```
wilbaker: How do we set the version of the bundle?

nickgra: This might be the key to understanding why we suddenly got hit with this issue, we may not be versioning it correctly today. Either way, I believe this is still a correct change to make.

jamill: > This might be the key to understanding why we suddenly got hit with this issue

We started seeing this shortly after we included the notification app in the installer :)

alameenshah: >> How do we set the version of the bundle?

Right now we are not updating the Bundle version (it is set to 1.0). I will take a look at the options (scripts used to set the ProjFS kext bundle version & also Apple's versioning tool `agvtool`) and incorporate version updating into the build script. Updated the [TODO](https://github.com/microsoft/VFSForGit/issues/1262) issue.

PR #1271, opened 2019-06-12
[WIP] Incremental commit graph (PR Build of Git)

derrickstolee:
The incremental commit-graph format is designed to speed up commit-graph writes, and especially to allow fast verification of the new data. At least, most of the time.

By only writing the new data to a small "tip" commit-graph file, we can write much less data and only verify the new tip file.

Current settings include `--size-multiple=4`: This guarantees each level of the chain is half the size of the level below. This means the chain length will be limited by log_4(N) where N is the number of commits. This means ~6 levels is the maximum theoretical chain length with 4,000,000 commits. We are not currently using the `--max-commits` setting.

With these settings, the `git commit-graph verify` command will be much faster on most new tip files. For example, a tip graph with 3,300 commits uses 183 KB of data and took 0.7 seconds to verify.

TODOs:

* [ ] Check functional tests.
* [x] Reconsider the `--size-multiple=4` which reduces the chain length at the cost of more "complete" writes. May not be worth it.
* [ ] Run perf tests on the performance machine (only when we know the format is final, to avoid problems).
* [ ] Add a directory listing of the `info/commit-graphs` dir to `gvfs diagnose`.
* [ ] Add the contents of `info/commit-graphs/commit-graph-chain` to `gvfs diagnose`.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbWithoutSharedCacheTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbWithoutSharedCacheTests.cs
@@ -289,7 +289,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             // Multi-pack-index write happens even if the prefetch downloads nothing, while
             // the commit-graph write happens only when the prefetch downloads at least one pack
 
-            string graphPath = Path.Combine(this.Enlistment.GetObjectRoot(this.fileSystem), "info", "commit-graph");
+            string graphPath = Path.Combine(this.Enlistment.GetObjectRoot(this.fileSystem), "info", "commit-graphs", "commit-graph-chain");
             string graphLockPath = graphPath + ".lock";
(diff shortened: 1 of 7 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbWithoutSharedCacheTests.cs
```diff
@@ -289,7 +289,7 @@ public void PrefetchCleansUpOphanedLockFiles()
             // Multi-pack-index write happens even if the prefetch downloads nothing, while
             // the commit-graph write happens only when the prefetch downloads at least one pack
 
-            string graphPath = Path.Combine(this.Enlistment.GetObjectRoot(this.fileSystem), "info", "commit-graph");
+            string graphPath = Path.Combine(this.Enlistment.GetObjectRoot(this.fileSystem), "info", "commit-graphs", "commit-graph-chain");
```
derrickstolee: We will need something other than `git commit-graph read` to ensure the graph is build. Probably a verify check.

PR #1278, opened 2019-06-13
Fix performance regression in LibGit2RepoInvoker

### LINKED ISSUE #1277: LibGit2 repo startup time affects read-object hook and hydration
The LibGit2Repo pool used to create a bunch of libgit2 repos and leave them open for the entire mount process. This needed to change for the packfile maintenance step to successfully delete packfiles. However, the way we dispose of the libgit2 repos on a timer affects performance in a bad way: it can take on the order of seconds to recreate one. 

Instead of disposing on a timer, instead we should dispose when the packfile maintenance starts and recreate the repo after it is done. If a request comes in during the maintenance, then we should honor it and create a repo. This may cause a stale .idx file, but we log those and clean them up the next time. Further, this isolates the slow startup to at most once a day.

derrickstolee:
Today, we noticed a performance regression from 1.0.19074.2 to 1.0.19130.1 (M150).

In that release, we changed the way we allocate and refresh LibGit2Repo objects. In order to allow packfile deletions, we need to drop the handles that LibGit2 keeps open during the duration of the LibGit2Repo. The change was to share one repo object and to dispose of it after 15 minutes of inactivity. However, that doesn't take into account the startup time required to initialize the data. This can be a lot worse if the user has thousands of pack-files.

To fix this, change the model. With this change, the PackfileMaintenanceStep will request that the LibGit2Repos are closed before performing the expire command, and then will reinitialize the LibGit2 repo right after. This will keep the "warmup" time as part of the initial mount or as the background operation.

If the read-object hook or a file hydration comes in while the expire command is running, a LibGit2 repo will be initialized and it _may_ cause some .idx files to stick around until the next run. This was already the case, but then the warmup would happen during those operations. This will happen at most once a day.

Resolves #1277

### DIFF
--- a/GVFS/GVFS.Common/Git/GitRepo.cs
+++ b/GVFS/GVFS.Common/Git/GitRepo.cs
@@ -44,12 +44,20 @@ namespace GVFS.Common.Git
             Unknown,
         }
 
-        public bool HasActiveLibGit2Repo => this.libgit2RepoInvoker?.IsActive == true;
-
         public GVFSLock GVFSLock
         {
             get;
             private set;
+        }
+
+        public void CloseActiveRepo()
+        {
+            this.libgit2RepoInvoker?.DisposeSharedRepo();
+        }
+
+        public void OpenRepo()
+        {
+            this.libgit2RepoInvoker?.InitializedSharedRepo();
         }
 
         public bool TryGetIsBlob(string sha, out bool isBlob)
--- a/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
@@ -6,39 +6,21 @@ namespace GVFS.Common.Git
 {
     public class LibGit2RepoInvoker : IDisposable
     {
-        private static readonly TimeSpan DefaultRepositoryDisposalPeriod = TimeSpan.FromMinutes(15);
-
         private readonly Func<LibGit2Repo> createRepo;
         private readonly ITracer tracer;
         private readonly object sharedRepoLock = new object();
-        private readonly TimeSpan sharedRepositoryDisposalPeriod;
         private volatile bool disposing;
         private volatile int activeCallers;
         private LibGit2Repo sharedRepo;
         private Timer sharedRepoDisposalTimer;
 
-        public LibGit2RepoInvoker(ITracer tracer, Func<LibGit2Repo> createRepo, TimeSpan? disposalPeriod = null)
+        public LibGit2RepoInvoker(ITracer tracer, Func<LibGit2Repo> createRepo)
         {
             this.tracer = tracer;
             this.createRepo = createRepo;
 
-            if (!disposalPeriod.HasValue || disposalPeriod.Value <= TimeSpan.Zero)
-            {
-                this.sharedRepositoryDisposalPeriod = DefaultRepositoryDisposalPeriod;
-            }
-            else
-            {
-                this.sharedRepositoryDisposalPeriod = disposalPeriod.Value;
-            }
-
-            this.sharedRepoDisposalTimer = new Timer(
-                (state) => this.DisposeSharedRepo(),
-                state: null,
-                dueTime: this.sharedRepositoryDisposalPeriod,
-                period: this.sharedRepositoryDisposalPeriod);
+            this.sharedRepo = this.createRepo();
         }
-
-        public bool IsActive => this.sharedRepo != null;
 
         public void Dispose()
         {
@@ -81,27 +63,7 @@ namespace GVFS.Common.Git
             }
         }
 
-        private LibGit2Repo GetSharedRepo()
-        {
-            lock (this.sharedRepoLock)
-            {
-                if (this.disposing)
-                {
-                    return null;
-                }
-
-                this.sharedRepoDisposalTimer?.Change(this.sharedRepositoryDisposalPeriod, this.sharedRepositoryDisposalPeriod);
-
-                if (this.sharedRepo == null)
-                {
-                    this.sharedRepo = this.createRepo();
-                }
-
-                return this.sharedRepo;
-            }
-        }
-
-        private void DisposeSharedRepo()
+        public void DisposeSharedRepo()
         {
             lock (this.sharedRepoLock)
             {
@@ -112,6 +74,29 @@ namespace GVFS.Common.Git
 
                 this.sharedRepo?.Dispose();
                 this.sharedRepo = null;
+            }
+        }
+
+        public void InitializedSharedRepo()
+        {
+            this.GetSharedRepo();
+        }
+
+        private LibGit2Repo GetSharedRepo()
+        {
+            lock (this.sharedRepoLock)
+            {
+                if (this.disposing)
+                {
+                    return null;
+                }
+
+                if (this.sharedRepo == null)
+                {
+                    this.sharedRepo = this.createRepo();
+                }
+
+                return this.sharedRepo;
             }
         }
     }
--- a/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
@@ -19,7 +19,7 @@ namespace GVFS.Common.Git
             this.tracer = tracer;
             this.createRepo = createRepo;
 
-            this.sharedRepo = this.createRepo();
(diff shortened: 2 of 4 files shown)


GVFS/GVFS.Common/Git/GitRepo.cs
```diff
@@ -44,14 +44,22 @@ private enum LooseBlobState
...
-
         public GVFSLock GVFSLock
         {
             get;
             private set;
         }
 
+        public void CloseActiveRepo()
+        {
+            this.libgit2RepoInvoker?.DisposeSharedRepo();
```
jamill: IIUC - this can be called after the `GitRepo` has been disposed, and that is why we are using the `?.` operator here? Assuming my interpretation is correct, it is a bit subtle as to why we we use this pattern here, but not elsewhere in the file - maybe a small comment to let future readers know why this is.

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
```diff
@@ -25,32 +23,65 @@ public void Setup()
...
-            this.numConstructors.ShouldEqual(0);
+            this.numConstructors.ShouldEqual(1);
         }
 
         [TestCase]
-        public void CreatesRepoOnTryInvoke()
+        public void CreatesOnRequestAfterClosed()
         {
-            this.numConstructors.ShouldEqual(0);
+            this.numConstructors.ShouldEqual(1);
```
jamill: Should we assert the initial number of disposals (equals 0), as we do with the initial number of constructions?

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
```diff
@@ -25,32 +23,65 @@ public void Setup()
...
 
             this.invoker.TryInvoke(repo => { return true; }, out bool result);
-            result.ShouldEqual(true);
+
+            this.numDisposals.ShouldEqual(1);
+            this.numConstructors.ShouldEqual(2);
+        }
+
+        [TestCase]
+        public void CreatesOnInvokeAfterClosed()
```
jamill: The test case names make it clear what they are testing - thanks!

jamill:
I think the main issue was the potential null ref exception - otherwise looks good Thanks!

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -150,7 +138,23 @@ protected override void PerformMaintenance()
...
+                    // running CloseActiveRepos() here, we ensure that we do not run twice with the same
+                    // LibGit2Repo active across two calls. A "new" repo should not hold handles to .idx files
+                    // that do not have corresponding .pack files, so we will clean them up in CleanStaleIdxFiles().
+                    this.Context.Repository.CloseActiveRepo();
+
+                    GitProcess.Result expireResult = this.RunGitCommand((process) => process.MultiPackIndexExpire(this.Context.Enlistment.GitObjectsRoot), nameof(GitProcess.MultiPackIndexExpire));
+                }
+                finally
+                {
+                    this.Context.Repository.OpenRepo();
```
wilbaker: Moving review to this PR (from #1279) as I see some other comments have been posted on this one.

Should this `finally` block run even if the exception thrown is a `StoppingException`?  I wasn't sure that we need to hit:

```
            // Run a test on the shared repo to ensure the object store
            // is loaded, as that is what takes a long time with many packs.
            this.GetSharedRepo().ObjectExists(GVFSConstants.AllZeroSha);
```

during shutdown.

(Note: I'm not sure if there are any other exception types that might be thrown from `RunGitCommand`)

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
```diff
@@ -25,32 +23,65 @@ public void Setup()
...
-        public void DoesNotCreateRepoOnConstruction()
+        public void DoesCreateRepoOnConstruction()
         {
-            this.numConstructors.ShouldEqual(0);
+            this.numConstructors.ShouldEqual(1);
         }
 
         [TestCase]
-        public void CreatesRepoOnTryInvoke()
+        public void CreatesOnRequestAfterClosed()
```
wilbaker: I'm not sure my understanding of what this test is validating is correct.  Is it that `InitializedSharedRepo` creates a LibGit2 after `DisposeSharedRepo` has been called?

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
```diff
@@ -25,32 +23,65 @@ public void Setup()
...
+
+            this.numDisposals.ShouldEqual(1);
+            this.numConstructors.ShouldEqual(1);
+
+            this.invoker.TryInvoke(repo => { return true; }, out bool result);
+
+            this.numDisposals.ShouldEqual(1);
+            this.numConstructors.ShouldEqual(2);
+
+            this.invoker.InitializedSharedRepo();
```
wilbaker: Maybe add a comment here that `InitializedSharedRepo` should not construct another libgit2 if after one was created for TryInvoke?

GVFS/GVFS.Common/Git/GitRepo.cs
derrickstolee: The unit tests for PackfileMaintenanceStep use the constructor that leaves this `null`. That's the only reason.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
derrickstolee: I never _expect_ other exceptions, but this is a performance optimization, so skipping the `OpenRepo()` call doesn't break us.

derrickstolee: So I'll remove the `try`/`finally`.

GVFS/GVFS.UnitTests/Common/LibGit2RepoInvokerTests.cs
derrickstolee: `Initialize[d]SharedRepo` had an extra 'd' which was a typo and may have added to the confusion.

I will rename this test `CreatedByInitializeAfterClosed`.

GVFS/GVFS.Common/Git/LibGit2RepoInvoker.cs
```diff
@@ -6,40 +6,22 @@ namespace GVFS.Common.Git
...
-        private static readonly TimeSpan DefaultRepositoryDisposalPeriod = TimeSpan.FromMinutes(15);
-
         private readonly Func<LibGit2Repo> createRepo;
         private readonly ITracer tracer;
         private readonly object sharedRepoLock = new object();
-        private readonly TimeSpan sharedRepositoryDisposalPeriod;
         private volatile bool disposing;
         private volatile int activeCallers;
         private LibGit2Repo sharedRepo;
         private Timer sharedRepoDisposalTimer;
```
kewillford: This `Timer` can be deleted now?

derrickstolee:
@kewillford I liked your questions so much I put them in the PR description. Also, you led me to double-check the numbers and found that I needed another commit.

jeschu1:
@derrickstolee great summary and description of the problem / fix!

I notice we have several functional test failures.  It looks like they *may* be related to left over .idx files.  
Ex:
Prefetching commits and trees failed: Unable to delete C:\Repos\GVFSFunctionalTests\enlistment\3ae33ecc316b48eeb393\.gvfs\.gvfsCache\735ad590dc55429aa9764a55743ed1df\gitObjects\pack\prefetch-1560524212-55e96e5fe5a54c988154a4f86b32c997.idx

Could this be because we aren't disposing (i.e. the idx is still in use)?

jamill:
Maybe prefetch should shutdown the libgit2 repo while it is running (and restart it when it finishes...)?

derrickstolee:
I’ll be able to get to the prefetch thing on Monday. I think this has always been a potential problem, but the change in this PR is that we preload the libgit2 data on mount time, creating handles.

derrickstolee:
All tests pass with the latest updates (timing the mount/unmount with the places where we want to delete packs during prefetch).

jamill:
Thanks! Latest changes look good to me

PR #1280, opened 2019-06-16
Mac kext: Drop lock for allocating virtualization root array memory

pmj:
This is an old fix from #500 which I've rebased/updated, isolated, and simplified a little.

Performing a blocking memory allocation while holding a lock in a vnode listener callback is not necessarily safe. (There is potential for a deadlock in low-memory situations.) When growing the virtualization root array, this was exactly what the existing code was doing. This change restructures the virtualization root insertion logic so that the lock can be dropped for allocating the larger memory block. Dropping the lock means it must be re-checked whether an insertion is still necessary, which was not readily possible with the old code structure.

A bunch of existing unit tests were calling `InsertVirtualizationRoot_Locked()` which no longer exists, these have been updated to call either `FindOrInsertVirtualizationRoot_LockedMayUnlock` or `VirtualizationRoot_RegisterProviderForPath` as appropriate.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
+++ b/ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
@@ -15,6 +15,7 @@
 #include "kernel-header-wrappers/stdatomic.h"
 #include "VnodeUtilities.hpp"
 #include "PerformanceTracing.hpp"
+#include "ArrayUtilities.hpp"
 
 #ifdef KEXT_UNIT_TESTING
 #include "VirtualizationRootsTestable.hpp"
@@ -24,6 +25,7 @@ static RWLock s_virtualizationRootsLock = {};
 
 // Current length of the s_virtualizationRoots array
 KEXT_STATIC uint16_t s_maxVirtualizationRoots = 0;
+static const uint16_t MaxVirtualizationRootsLimit = INT16_MAX + 1u;
 KEXT_STATIC VirtualizationRoot* s_virtualizationRoots = nullptr;
 
 // Looks up the vnode/vid and fsid/inode pairs among the known roots
@@ -32,13 +34,15 @@ static VirtualizationRootHandle FindRootAtVnode_Locked(vnode_t vnode, uint32_t v
 static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHandle, vnode_t vnode, uint32_t vid, FsidInode fileId);
 static bool FsidsAreEqual(fsid_t a, fsid_t b);
 KEXT_STATIC bool PathInsideDirectory(const char* directoryPath, const char* path);
+static void GrowVirtualizationRootArrayWithMemory_Locked(VirtualizationRoot* newMemory, uint16_t newLength);
 
 // Looks up the vnode and fsid/inode pair among the known roots, and if not found,
 // detects if there is a hitherto-unknown root at vnode by checking attributes.
 static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t vnode, const FsidInode& vnodeFsidInode);
 
 static VirtualizationRootHandle FindUnusedIndex_Locked();
-KEXT_STATIC VirtualizationRootHandle InsertVirtualizationRoot_Locked(PrjFSProviderUserClient* userClient, pid_t clientPID, vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path);
+
+KEXT_STATIC VirtualizationRootHandle FindOrInsertVirtualizationRoot_LockedMayUnlock(vnode_t vnode, uint32_t vid, FsidInode persistentIds, const char* path);
 
 ActiveProviderProperties VirtualizationRoot_GetActiveProvider(VirtualizationRootHandle rootHandle)
 {
@@ -83,7 +87,8 @@ kern_return_t VirtualizationRoots_Init()
         return KERN_FAILURE;
     }
     
-    s_maxVirtualizationRoots = 128;
+    // Start with a small size so the resizing logic is regularly tested
+    s_maxVirtualizationRoots = 4;
     s_virtualizationRoots = Memory_AllocArray<VirtualizationRoot>(s_maxVirtualizationRoots);
     if (nullptr == s_virtualizationRoots)
     {
@@ -227,18 +232,10 @@ static VirtualizationRootHandle FindOrDetectRootAtVnode(vnode_t _Nonnull vnode, 
  
             RWLock_AcquireExclusive(s_virtualizationRootsLock);
             {
-                // Vnode may already have been inserted as a root in the interim
-                rootIndex = FindRootAtVnode_Locked(vnode, vid, vnodeFsidInode);
+                rootIndex = FindOrInsertVirtualizationRoot_LockedMayUnlock(vnode, vid, vnodeFsidInode, path);
                 
-                if (RootHandle_None == rootIndex)
-                {
-                    // Insert new offline root
-                    rootIndex = InsertVirtualizationRoot_Locked(nullptr, 0, vnode, vid, vnodeFsidInode, path);
-                    
-                    // TODO: error handling
-                    assert(rootIndex >= 0);
-                }
-
+                // TODO: error handling
+                assert(rootIndex >= 0);
             }
             RWLock_ReleaseExclusive(s_virtualizationRootsLock);
         }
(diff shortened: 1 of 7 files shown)


ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -83,7 +86,8 @@ kern_return_t VirtualizationRoots_Init()
         return KERN_FAILURE;
     }
     
-    s_maxVirtualizationRoots = 128;
+    // Start with a small size so the resizing logic is regularly tested
+    s_maxVirtualizationRoots = 4;
```
wilbaker: If this is initialized to 4, why does the unit test need this code?

```
    // Start with 4 "full" array items, forcing resize on next insertion
    Memory_FreeArray(s_virtualizationRoots, s_maxVirtualizationRoots);
    s_maxVirtualizationRoots = 4;
    s_virtualizationRoots = Memory_AllocArray<VirtualizationRoot>(s_maxVirtualizationRoots);
    memset(s_virtualizationRoots, 0, s_maxVirtualizationRoots * sizeof(s_virtualizationRoots[0]));
```

Is it just to make sure the array has a size of 4 for the test?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -380,43 +350,66 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
...
+            if (newLength <= s_maxVirtualizationRoots)
+            {
+                KextLog_Error("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array for root at '%s' failed: out of valid indices, array has reached %u items\n", path, s_maxVirtualizationRoots);
+                return RootHandle_None;
+            }
+        
+            // Must drop lock to safely allocate memory with blocking alloc
+            RWLock_ReleaseExclusive(s_virtualizationRootsLock);
+            VirtualizationRoot* grownArray = Memory_AllocArray<VirtualizationRoot>(newLength);
+            KextLog_Info("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array to %u items for root at '%s' with non-blocking alloc failed, blocking alloc returned %p\n", newLength, path, grownArray);
```
wilbaker: > growing virtualization root array to %u items for root at '%s' with non-blocking alloc failed

Is this log message accurate? I can't see where we tried to grow the array with a non-blocking alloc first.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -380,43 +350,66 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
...
+            uint16_t newLength = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
+            if (newLength <= s_maxVirtualizationRoots)
+            {
+                KextLog_Error("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array for root at '%s' failed: out of valid indices, array has reached %u items\n", path, s_maxVirtualizationRoots);
+                return RootHandle_None;
+            }
+        
+            // Must drop lock to safely allocate memory with blocking alloc
+            RWLock_ReleaseExclusive(s_virtualizationRootsLock);
+            VirtualizationRoot* grownArray = Memory_AllocArray<VirtualizationRoot>(newLength);
```
wilbaker: Check if `Memory_AllocArray` returned null?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -380,43 +350,66 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
...
+            KextLog_Info("FindOrInsertVirtualizationRoot_LockedMayUnlock: growing virtualization root array to %u items for root at '%s' with non-blocking alloc failed, blocking alloc returned %p\n", newLength, path, grownArray);
+            RWLock_AcquireExclusive(s_virtualizationRootsLock);
+            
+            uint16_t newLengthAgain = MIN(s_maxVirtualizationRoots * 2u, INT16_MAX + 1u);
+            if (newLengthAgain != newLength || newLength == s_maxVirtualizationRoots)
```
wilbaker: Why do we need to check `newLength == s_maxVirtualizationRoots`?  

I'm having trouble finding a scenario where `newLengthAgain != newLength` would be false but `newLength == s_maxVirtualizationRoots` would be true.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -380,43 +350,66 @@ static void RefreshRootVnodeIfNecessary_Locked(VirtualizationRootHandle rootHand
...
+    root->rootVNode = virtualizationRootVNode;
+    root->rootVNodeVid = rootVid;
+    root->rootFsid = persistentIds.fsid;
+    root->rootInode = persistentIds.inode;
+    if (path != nullptr)
+    {
+        strlcpy(root->path, path, sizeof(root->path));
+    }
+
+    KextLog_File(virtualizationRootVNode, "InsertVirtualizationRoot_Locked: virtualization root inserted at index %d: (path: \"%s\", fsid: 0x%x:%x, inode: 0x%llx) directory vnode %p:%u.",
```
wilbaker: Nit: `"InsertVirtualizationRoot_Locked:` -> `"FindOrInsertVirtualizationRoot_LockedMayUnlock:`

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: * `newLengthAgain != newLength` is of course `true` if another thread has beaten us to the resize.
 * However, if the other thread resizing the array caused it to hit the maximum of `INT16_MAX + 1u` items, then `newLength` and `newLengthAgain` will both be equal to that value, as will `s_maxVirtualizationRoots`. In that case, there's not really any "resizing" happening.

So I think the code is correct, but the intention is perhaps not clear. I think I might replace  `INT16_MAX + 1u` with a symbolic constant and compare `s_maxVirtualizationRoots` to that as the second part of this condition.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: > Is it just to make sure the array has a size of 4 for the test?

Yes, in case any other tests left the array in some other state. I guess we could init the array during the test's `setUp` but I think that would move that precondition of the test too far apart from the initialisation code, so I'd say in that case we should move the test to a new test class that just contains the tests for the array resizing logic.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: For blocking allocs, this should pretty much never happen, and the caller currently asserts on this function succeeding, but I guess it's good to be explicit about it. I've added that check now, it'll be in the next push.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: Sorry, that's a reference to code I removed. No point keeping this message either.

pmj:
> Apologies for the delay in getting to this review.
> 
> Overall the changes look good, just a few minor comments and questions.

Thanks for the review! I've made adjustments according to your feedback in 941df90 - I've left it as an independent commit for now to make my updates obvious. I'll squash it into the main commit once everyone is happy.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
wilbaker: > However, if the other thread resizing the array caused it to hit the maximum of INT16_MAX + 1u items, then newLength and newLengthAgain will both be equal to that value, as will s_maxVirtualizationRoots. In that case, there's not really any "resizing" happening.

Ok that makes sense thanks!

> I think I might replace INT16_MAX + 1u with a symbolic constant and compare s_maxVirtualizationRoots to that as the second part of this condition.

I think that would make this check much easier to follow, could you make that change before merging?

pmj: > I think that would make this check much easier to follow, could you make that change before merging?

Thanks for reminding me - I've now tidied that up.

PR #1281, opened 2019-06-17
PrefetchStep: don't send post-fetch request if no new packs

derrickstolee:
Every time the post-fetch step runs, we write a new multi-pack-index. This is important to catch any leftover packs that were added by `gvfs prefetch --files` or by the loose objects step.

However, for a normal `git fetch` that calls `gvfs prefetch --commits`, there are likely no new packs. This previously triggered a post-fetch step anyway, and now it will not make that request if there are no new packs. This can save about one minute of CPU time per "trivial" prefetch.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
@@ -64,7 +64,7 @@ namespace GVFS.Common.Maintenance
                 this.UpdateKeepPacks();
             }
 
-            this.TrySchedulePostFetchJob(packIndexes);
+            this.SchedulePostFetchJob(packIndexes);
 
             return true;
         }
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.Common/Maintenance/PrefetchStep.cs
```diff
@@ -259,8 +259,13 @@ private bool TryGetMaxGoodPrefetchTimestamp(out long maxGoodTimestamp, out strin
             return true;
         }
 
-        private bool TrySchedulePostFetchJob(List<string> packIndexes)
+        private void TrySchedulePostFetchJob(List<string> packIndexes)
```
derrickstolee: This `bool` to `void` change is because the only caller does not use the return type.

kewillford: Should we change the method name as well then and remove the `Try`?

derrickstolee: An excellent point.

derrickstolee:
/azd run GitHub VFSForGit Mac Functional Tests

PR #1283, opened 2019-06-18
Incremental commit graph

derrickstolee:
The incremental commit-graph format is designed to speed up commit-graph writes, and especially to allow fast verification of the new data. At least, most of the time.

Uses the new feature in microsoft/git#156.

By only writing the new data to a small "tip" commit-graph file, we can write much less data and only verify the new tip file.

Current settings include `--size-multiple=4`: This guarantees each level of the chain is half the size of the level below. This means the chain length will be limited by log_4(N) where N is the number of commits. This means ~6 levels is the maximum theoretical chain length with 4,000,000 commits. We are not currently using the `--max-commits` setting.

With these settings, the `git commit-graph verify` command will be much faster on most new tip files. For example, a tip graph with 3,300 commits uses 183 KB of data and took 0.7 seconds to verify.

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20190626.2</GitPackageVersion>
+    <GitPackageVersion>2.20190701.1</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -529,7 +529,7 @@ namespace GVFS.Common.Git
         /// </summary>
         public Result WriteCommitGraph(string objectDir, List<string> packs)
         {
-            string command = "commit-graph write --stdin-packs --append --object-dir \"" + objectDir + "\"";
+            string command = "commit-graph write --stdin-packs --split --size-multiple=4 --object-dir \"" + objectDir + "\"";
             return this.InvokeGitInWorkingDirectoryRoot(
                 command,
                 useReadObjectHook: true,
@@ -547,7 +547,7 @@ namespace GVFS.Common.Git
 
         public Result VerifyCommitGraph(string objectDir)
         {
-            string command = "commit-graph verify --object-dir \"" + objectDir + "\"";
+            string command = "commit-graph verify --shallow --object-dir \"" + objectDir + "\"";
             return this.InvokeGitInWorkingDirectoryRoot(command, useReadObjectHook: true);
         }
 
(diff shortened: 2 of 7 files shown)


derrickstolee:
image

Here is my `gitObjects/info/commit-graphs` folder after a few iterations.

GVFS/GVFS.Build/GVFS.props
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20190610.5</GitPackageVersion>
+    <GitPackageVersion>2.20190621.1-pr</GitPackageVersion>
```
derrickstolee: Change to release build, when available.

derrickstolee:
Going to close and reopen to retrigger builds.

derrickstolee:
Here is an update on my commit-graphs list after a few weeks:

image

Note the modified dates: 6/18, 6/26, 6/27, 6/28, & 6/28. That gives an idea of how often these are merging down.

(the last update was on June 28th when I installed a different version of VFS for Git for testing.)

derrickstolee:
Perf tests checked out with no change outside of error.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -529,7 +529,7 @@ public Result PackObjects(string filenamePrefix, string gitObjectsDirectory, Act
         /// </summary>
         public Result WriteCommitGraph(string objectDir, List<string> packs)
         {
-            string command = "commit-graph write --stdin-packs --append --object-dir \"" + objectDir + "\"";
+            string command = "commit-graph write --stdin-packs --split --size-multiple=4 --object-dir \"" + objectDir + "\"";
```
kewillford: Do we need some tests the write the old and convert to incremental and the other way as well or I'm guessing there are git tests for that?

derrickstolee: `t/t5323-split-commit-graph.sh` has [a test that upgrades from the "flat" file to a chain](https://github.com/microsoft/git/blob/vfs-2.22.0/t/t5324-split-commit-graph.sh#L108-L120).

kewillford: Also is it git that is moving the file from the `info/commit-graph` location to the new `info/commit-graphs/X` with the `split` command?

derrickstolee: Yes. Git realizes that we are adding commits on top of an existing `info/commit-graph` file and moves it to `info/commit-graphs/graph-{hash}.graph` before completing the write of the next level (the data is already written to a temp file before the move) and updating the `info/commit-graphs/commit-graph-chain` file to point to it.

During this "upgrade" step, there is a brief moment where `info/commit-graph` does not exist and `info/commit-graphs/commit-graph-chain` does not exist yet. A process that looks for a commit-graph in that window may have slow performance, but not incorrect results.

derrickstolee:
Five performance runs all looked good. Merging.

PR #1287, opened 2019-06-20
use repository relative paths in DiffTreeResult.TargetPath

kivikakk:
We construct absolute paths at point of use -- some cases use the `Enlistment`'s `WorkingDirectoryRoot`, and others use the `WorkingDirectoryBackingRoot`.  (These are identical on Windows and Mac, but different on Linux.)

As a result of this change:

* `DiffTreeResult` only deals in <var title="N.B.! This is the opposite of the Sith.">relative, repository-internal paths</var>. These are not prefixed with the directory separator character.
* `HydrateFilesStage` takes the working directory root to trigger projections by the projected fs.
* `BlobPrefetcher` likewise deals in repository-internal paths. Note that `--folders /` now produces an empty string folder path, which **does** make it into the lists later used for determining whether to include a file. This is a bit awkward, but works in all the ways we want it to.
* `FastFetch.Index` no longer tries to convert between full paths and git relative paths.

This fixes all `GVFS.FunctionalTests.Tests.EnlistmentPerFixture.PrefetchVerbTests` tests on Linux.

### DIFF
--- a/GVFS/FastFetch/Index.cs
+++ b/GVFS/FastFetch/Index.cs
@@ -213,7 +213,7 @@ namespace FastFetch
                     addedOrEditedLocalFiles,
                     (localPath) =>
                     {
-                        string gitPath = FromDotnetFullPathToGitRelativePath(localPath, this.repoRoot);
+                        string gitPath = localPath.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
                         long offset;
                         if (this.indexEntryOffsets.TryGetValue(gitPath, out offset))
                         {
--- a/GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
+++ b/GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
@@ -125,7 +125,7 @@ namespace GVFS.Common.Prefetch.Git
                     GitProcess.Result result = this.git.DiffTree(
                         sourceTreeSha,
                         targetTreeSha,
-                        line => this.EnqueueOperationsFromDiffTreeLine(this.tracer, this.enlistment.WorkingDirectoryBackingRoot, line));
+                        line => this.EnqueueOperationsFromDiffTreeLine(this.tracer, line));
 
                     if (result.ExitCodeIsFailure)
                     {
@@ -143,11 +143,12 @@ namespace GVFS.Common.Prefetch.Git
                 metadata.Add("DirectoryOperationsCount", this.TotalDirectoryOperations);
                 metadata.Add("FileDeleteOperationsCount", this.TotalFileDeletes);
                 metadata.Add("RequiredBlobsCount", this.RequiredBlobs.Count);
+                metadata.Add("FileAddOperationsCount", this.FileAddOperations.Sum(kvp => kvp.Value.Count));
                 activity.Stop(metadata);
             }
         }
 
-        public void ParseDiffFile(string filename, string repoRoot)
+        public void ParseDiffFile(string filename)
         {
             using (ITracer activity = this.tracer.StartActivity("PerformDiff", EventLevel.Informational))
             {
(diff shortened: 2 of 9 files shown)


kivikakk:
More test failures sprouted up. I'll re-request review when done.

kivikakk:
Ready for review.

GVFS/FastFetch/Index.cs
```diff
@@ -213,7 +213,7 @@ private bool UpdateFileInformationFromDiskForFiles(MemoryMappedViewAccessor inde
                     addedOrEditedLocalFiles,
                     (localPath) =>
                     {
-                        string gitPath = FromDotnetFullPathToGitRelativePath(localPath, this.repoRoot);
+                        string gitPath = localPath.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator);
```
derrickstolee: This seems wrong for two, conflicting reasons:

1. If this was here, then `localPath` needs removing `this.repoRoot`. Is `localPath` already local and `this.repoRoot` empty?
2. There is another instance of `FromDotnetFullPathToGeitRelativePath()` in this file. Why only one changed?

I'm having trouble resolving the dissonance here, as the method context looks like `localPath` shouldn't contain `this.repoRoot`, in which case this would be a bug!

@kivikakk: you were in here, so how did you find that this needed changing? Can you resolve my confusion here?

kivikakk: 1. The purpose of this change is to exactly make `this.repoRoot` no longer part of these paths. (`this.repoRoot` is never empty.) The problem is that there are multiple repository roots depending on what you want -- the working tree, or the backing store. These are the same on Windows and Mac and therefore it was fine to use absolute paths in many places in the VFS4G codebase, because that absolute path was fit for use in whatever place you might want to, whether that's triggering hydration (working tree) or writing the hydrated file contents to disk (backing store).

   On Linux, these paths are different -- `src/` for the working tree, `.vfsforgit/lower/` for the backing store. Accordingly, we can't meaningfully pass around absolute paths because any given absolute path is only fit for one of those two purposes. I briefly considered continuing to use the absolute working tree path (`src/`) everywhere and replacing the prefix with the backing store path (`.vfsforgit/linux/`) where needed, but while a smaller change, is a huge hack that requires a lot of explanation, and is essentially technical debt that would come to bite us later, as it doesn't require a Windows or Mac programmer to think about which of the two paths are actually desired here.

   Accordingly, this change changes our internal diff objects to use git-relative paths everywhere, and construct appropriate absolute paths at time of use. `addedOrEditedLocalFiles` comes from `DiffHelper` and so has relative paths, which is why we no longer call `FromDotnetFullPathToGitRelativePath`.

2. I originally blindly changed both without looking at the contexts of each, and hit a new error which vexed me for a while. The remaining use is on `file.FullName` where `file` is a `FileInfo` from `FastFetch.WorkingTree.ForAllFiles`. These are absolute paths that come directly from the filesystem and are guaranteed to have a `this.repoRoot` prefix, so we must remove that prefix to correctly index into `this.indexEntryOffsets`.

derrickstolee: Thanks for the context!

kivikakk:
I'll wait until my Friday to merge this if anyone else wants to :eyes:!

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -143,19 +143,20 @@ public void PerformDiff(string sourceTreeSha, string targetTreeSha)
                 metadata.Add("DirectoryOperationsCount", this.TotalDirectoryOperations);
                 metadata.Add("FileDeleteOperationsCount", this.TotalFileDeletes);
                 metadata.Add("RequiredBlobsCount", this.RequiredBlobs.Count);
+                metadata.Add("FileAddOperationsCount", this.FileAddOperations.Sum(kvp => kvp.Value.Count));
```
chrisd8088: This is just a bit of extra logging metadata, it looks like.

kivikakk: 👍  Yep; it was useful for me to have, and fit the theme of the other items being logged.

PR #1288, opened 2019-06-20
[Linux] add pre-rename and post-delete Linux events

chrisd8088:
The `GVFS.Virtualization.FileSystemCallbacks` background task processing code is responsible for adding and removing paths to the GVFS modified-paths database, and in the case of rename
file operations, needs to remove the old path from the database.

In the Mac Darwin/BSD kauth design, a rename file operation is preceded by a `KAUTH_VNODE_DELETE` permission check, which the ProjFS.Mac kext reports as a `PreDelete` notification.  This in turn ensures that the `FileSystemCallbacks.ExecuteBackgroundOperation()` method will remove the source path from the modified-paths database.  However, the ProjFS.Linux
and libprojfs implementation did not include a parallel pre-delete event prior to rename file operations, leading to functional test failures on Linux.

Because there is no necessity to follow the BSD design, we can introduce true pre-rename (`PROJFS_MOVE_PERM`) permission requests in Linux, as well as post-delete events.  This brings the libprojfs/ProjFS.Linux design somewhat closer to the Windows projectedfslib/PrjFlt one.

In particular, this allows the GVFS `LinuxFileSystemVirtualizer` to enforce restrictions on renaming the `.git/index` file, akin to the ones enforced on Windows.

By reporting both the source and target (destination) paths in the post-rename events, we permit the `FileSystemCallbacks` task processing code to add and remove the paths to the database, without having to do so in either the pre-rename event or a BSD-like pre-delete event delivered prior to a rename.

We can also report both the existing and new paths to the `HardLinkCreated` event, although only the latter is ultimately used.

Note that the `PR - Linux - Build and Unit Test` CI job is expected to fail because we have a parallel change to make to libprojfs; see github/libprojfs#77.

/cc @kivikakk

### DIFF
--- a/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
+++ b/ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
@@ -30,14 +30,15 @@ namespace PrjFSLib.Linux
         public virtual GetFileStreamCallback OnGetFileStream { get; set; }
         public virtual LogErrorCallback OnLogError { get; set; }
 
-        public virtual NotifyFileModified OnFileModified { get; set; }
-        public virtual NotifyFilePreConvertToFullEvent OnFilePreConvertToFull { get; set; }
         public virtual NotifyPreDeleteEvent OnPreDelete { get; set; }
         public virtual NotifyPreRenameEvent OnPreRename { get; set; }
+        public virtual NotifyFilePreConvertToFullEvent OnFilePreConvertToFull { get; set; }
+
         public virtual NotifyNewFileCreatedEvent OnNewFileCreated { get; set; }
         public virtual NotifyFileDeletedEvent OnFileDeleted { get; set; }
         public virtual NotifyFileRenamedEvent OnFileRenamed { get; set; }
         public virtual NotifyHardLinkCreatedEvent OnHardLinkCreated { get; set; }
+        public virtual NotifyFileModified OnFileModified { get; set; }
 
         public virtual Result StartVirtualizationInstance(
             string storageRootFullPath,
@@ -466,13 +467,17 @@ namespace PrjFSLib.Linux
             {
                 nt = NotificationType.PreRename;
             }
-            else if ((ev.Mask & ProjFS.Constants.PROJFS_CLOSE_WRITE) != 0)
+            else if ((ev.Mask & ProjFS.Constants.PROJFS_OPEN_PERM) != 0)
             {
-                nt = NotificationType.FileModified;
+                nt = NotificationType.PreConvertToFull;
             }
             else if ((ev.Mask & ProjFS.Constants.PROJFS_CREATE) != 0 && !isLink)
             {
                 nt = NotificationType.NewFileCreated;
+            }
+            else if ((ev.Mask & ProjFS.Constants.PROJFS_DELETE) != 0)
+            {
+                nt = NotificationType.FileDeleted;
             }
             else if ((ev.Mask & ProjFS.Constants.PROJFS_MOVE) != 0)
             {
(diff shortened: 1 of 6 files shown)


chrisd8088:
This one (despite the known Linux CI test failure, which is due to a dependency not in libprojfs master yet) ready for review, @derrickstolee and @wilbaker -- once you're happy with this, I will commit github/libprojfs#77 to master, at which point the Linux CI test will pass, and it'll be all green.

chrisd8088:
Hi @wilbaker ... would you have a chance to look this over?  (And note that the Linux CI failure is expected, for now.)  I hope I've captured the sense of the Windows logic in this Linux code, but would appreciate a double-check on that!  Thanks very much, Chris.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -460,14 +463,22 @@ private int HandleNonProjEvent(ref ProjFS.Event ev, bool perm)
             {
                 nt = NotificationType.PreDelete;
             }
-            else if ((ev.Mask & ProjFS.Constants.PROJFS_CLOSE_WRITE) != 0)
+            else if ((ev.Mask & ProjFS.Constants.PROJFS_MOVE_PERM) != 0)
             {
-                nt = NotificationType.FileModified;
+                nt = NotificationType.PreRename;
+            }
+            else if ((ev.Mask & ProjFS.Constants.PROJFS_OPEN_PERM) != 0)
```
wilbaker: Unrelated to this PR: Out of curiosity what triggers the `PROJFS_OPEN_PERM` event being delivered?

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
```diff
@@ -540,24 +551,31 @@ private int HandlePermEvent(ref ProjFS.Event ev)
                 case NotificationType.PreDelete:
                     return this.OnPreDelete(relativePath, isDirectory);
 
-                case NotificationType.FileModified:
-                    this.OnFileModified(relativePath);
-                    return Result.Success;
+                case NotificationType.PreRename:
+                    return this.OnPreRename(relativePath, relativeDestinationPath, isDirectory);
```
wilbaker: What will `relativePath` and `relativeDestinationPath` be when the rename crosses the virtualization root boundary?

chrisd8088: Good question!  Fortunately, that can't happen, because `rename(2)` won't work across device boundaries, i.e., between mounted filesystems, and the whole repository/working directory is a separate mounted filesystem.

As it happens, I just rewrote some functional tests with this in mind -- see 6eb3597ece6cfb436a2478a5a50771fe5affbf69, which uses `BashRunner` instead of `SystemIORunner` on Linux when executing tests which move things in/out of the repository.  That's because command-line `mv` will helpfully fall back to a copy-and-delete strategy.  But an actual `rename()` file operation will error out before it gets to the point that we'd need to send any notifications.

ProjFS.Linux/PrjFSLib.Linux.Managed/VirtualizationInstance.cs
chrisd8088: The short version is that we deliver that event just before a file (not a directory) is opened for writing, and only proceed if the handler returns `PROJFS_ALLOW`.

The main call is in libprojfs in the [`open()` operation handler](https://github.com/github/libprojfs/blob/95cb9cafd1b23d7900b153ef8237f3c265c401af/lib/projfs.c#L985) as well as in the `rename()` or `truncate()` operation handlers also; all those calls to `project_file()` then [invoke](https://github.com/github/libprojfs/blob/95cb9cafd1b23d7900b153ef8237f3c265c401af/lib/projfs.c#L744) `project_locked_path()` which [sends the `PROJFS_OPEN_PERM` event](https://github.com/github/libprojfs/blob/95cb9cafd1b23d7900b153ef8237f3c265c401af/lib/projfs.c#L529).

wilbaker: Ok thanks!

PR #1293, opened 2019-06-21
Add include verb for allowing projection of only specific folders 

kewillford:
This is a feature to only include specified folders in the projection.  I'm not stuck with the naming that I have chosen so if you have better ones please let me know.  

You should be able to clone as normal and it will project everything until something is added to the included folder database via the `gvfs include` verb.

The `gvfs include` verb includes the following:
```
-a <folderlist> // add ; separated folders to the list to include
-r <folderlist> // remove ; separated folders to the list
-l // list the folders that are currently in the list
```
Examples:
```
gvfs include -l
gvfs include -a GVFS/GVFS;Scripts
gvfs include -r Scripts
```

You can use both the `-a` and `-r` in the same invocation.  The removes will be applied first and then the adds.

You must have a clean `git status` in order to add or remove folders so the the code can reproject without issue.

These changes currently don't do anything with the modified paths so if you create a directory or get a conflict that is outside the included folders then the paths will be added to the modified path list and continue to be in the working directory for git to keep up to date.

TODO:
- [x] Testing on Mac
- [x] Decide how to handle edge cases like creating a directory that is in the projection but not in the included list
- [x] Test conflict scenarios and make sure they work as expected
- [x] Performance test when no included folders
- [x] Performance test while there are included folders

### DIFF
--- a/GVFS/GVFS.Common/Database/IncludedFolderTable.cs
+++ b/GVFS/GVFS.Common/Database/IncludedFolderTable.cs
@@ -1,0 +1,89 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.IO;
+using System.Text;
+
+namespace GVFS.Common.Database
+{
+    public class IncludedFolderTable : IIncludedFolderCollection
+    {
+        private IGVFSConnectionPool connectionPool;
+
+        public IncludedFolderTable(IGVFSConnectionPool connectionPool)
+        {
+            this.connectionPool = connectionPool;
+        }
+
+        public static void CreateTable(IDbConnection connection)
+        {
+            using (IDbCommand command = connection.CreateCommand())
+            {
+                command.CommandText = "CREATE TABLE IF NOT EXISTS [IncludedFolder] (path TEXT PRIMARY KEY COLLATE NOCASE) WITHOUT ROWID;";
+                command.ExecuteNonQuery();
+            }
+        }
+
+        public void Add(string directory)
+        {
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    command.CommandText = "INSERT OR REPLACE INTO IncludedFolder (path) VALUES (@path);";
+                    command.AddParameter("@path", DbType.String, directory);
+                    command.ExecuteNonQuery();
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new GVFSDatabaseException($"{nameof(IncludedFolderTable)}.{nameof(this.Add)}({directory}) Exception: {ex.ToString()}", ex);
+            }
+        }
+
+        public List<string> GetAll()
+        {
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    List<string> directories = new List<string>();
+                    command.CommandText = $"SELECT path FROM IncludedFolder;";
+                    using (IDataReader reader = command.ExecuteReader())
+                    {
+                        while (reader.Read())
+                        {
+                            directories.Add(reader.GetString(0));
+                        }
+                    }
+
+                    return directories;
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new GVFSDatabaseException($"{nameof(PlaceholderTable)}.{nameof(this.GetAll)} Exception: {ex.ToString()}", ex);
+            }
+        }
+
+        public void Remove(string directory)
+        {
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    command.CommandText = "DELETE FROM IncludedFolder WHERE path = @path;";
+                    command.AddParameter("@path", DbType.String, directory);
+                    command.ExecuteNonQuery();
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new GVFSDatabaseException($"{nameof(IncludedFolderTable)}.{nameof(this.Remove)}({directory}) Exception: {ex.ToString()}", ex);
+            }
+        }
+    }
+}
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
@@ -136,6 +136,46 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             projectedFile.ShouldBeAFile(this.fileSystem);
         }
 
+        [TestCase, Order(6)]
+        public void ReadFileThenChangingIncludeFoldersShouldRemoveFileAndFolder()
+        {
+            string fileToRead = Path.Combine(this.Enlistment.RepoRoot, "Scripts", "RunFunctionalTests.bat");
+            this.fileSystem.ReadAllText(fileToRead);
+
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
+
+            string folderPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts");
+            folderPath.ShouldNotExistOnDisk(this.fileSystem);
+            fileToRead.ShouldNotExistOnDisk(this.fileSystem);
+        }
+
+        [TestCase, Order(7)]
+        public void ModifiedFileShouldNotAllowIncludedFolderChange()
+        {
+            string modifiedPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts", "RunFunctionalTests.bat");
+            this.fileSystem.WriteAllText(modifiedPath, "New Contents");
+
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            this.ValidateIncludedFolders(new string[0]);
+        }
+
+        [TestCase, Order(8)]
+        public void ModifiedFileAndCommitThenChangingIncludeFoldersShouldKeepFileAndFolder()
+        {
+            string modifiedPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts", "RunFunctionalTests.bat");
+            this.fileSystem.WriteAllText(modifiedPath, "New Contents");
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "add .");
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -m Test");
+
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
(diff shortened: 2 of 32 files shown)


GVFS/GVFS/CommandLine/IncludeVerb.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        private void CheckGitStatus(ITracer tracer, GVFSEnlistment enlistment)
+        {
+            GitProcess.Result statusResult = null;
+            if (!this.ShowStatusWhileRunning(
+                () =>
+                {
+                    GitProcess git = new GitProcess(enlistment);
+                    statusResult = git.Status(allowObjectDownloads: false, useStatusCache: false);
```
jrbriggs: I noticed that our status invocation doesn't override status.showuntrackedfiles, so there is a edge case here where a user has untracked file(s) but a naked `git status` won't show them, reporting working tree clean. That might lead to some wonkiness, especially if (going even more edge-casey here) the file has the same name as a directory that is being added to the projection.

Maybe we should either always force that config setting to true when we call status, if that's more correct for all usages of `Status()` in our codebase,  or we could plumb in an option to the `Status()` method.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -647,7 +654,10 @@ private void AddItemFromIndexEntry(GitIndexEntry indexEntry)
         {
             if (indexEntry.BuildingProjection_HasSameParentAsLastEntry)
             {
-                indexEntry.BuildingProjection_LastParent.AddChildFile(indexEntry.BuildingProjection_GetChildName(), indexEntry.Sha);
+                if (indexEntry.BuildingProjection_ShouldInclude || indexEntry.BuildingProjection_ShouldIncludeRecursive)
```
kewillford: Need to add `this.rootIncludedFolder.Children.Count == 0 ||` here.

GVFS/GVFS/CommandLine/IncludeVerb.cs
kewillford: Thanks I will test this more.  There is also the question of ignored files and how we would handle them but like you said these are really on the edge of doing some weird things but we should definitely try to handle them as best we can.  BTW the status in VFSForGit is returning:
```
On branch master

nothing to commit (use -u to show untracked files)
```
Which is weird because running `git status` from the command line gives you
```
C:\gvfs_repos\ft2\src>git status
On branch master

```

GVFS/GVFS/CommandLine/IncludeVerb.cs
```diff
@@ -0,0 +1,173 @@
...
+    public class IncludeVerb : GVFSVerb.ForExistingEnlistment
+    {
+        private const string IncludeVerbName = "include";
+
+        [Option(
+            'a',
+            "add",
+            Required = false,
+            Default = "",
+            HelpText = "A semicolon-delimited list of folders to include. Wildcards are not supported.")]
```
derrickstolee: We will want to allow `--stdin-add` or something so we don't run out of room in the command-line arguments when scoper runs. Can be delayed to a later PR.

derrickstolee:
Ok, I finally got the GitCommandTests working while watching the partial mode. Mostly, these check that the results of Git commands in a vanilla Git repo match those in the GVFS repo with some included folders. Additionally, it verifies the working directory contents match _except_ outside the specified cone. The goal here is just to ensure we are not changing any Git behavior within our cone. We will need to add tests that check how Git behaves when messing around outside the cone; that can be more reactionary to problems we see users hitting as they test the feature.

Please consider cherry-picking [this commit](https://github.com/derrickstolee/vfsforgit/commit/d0cd7ccfb3847e5407d85508fa4e9c4463a21644) which had [this successful build](https://dev.azure.com/gvfs/ci/_build/results?buildId=13523).

Note that I created the [PR - Windows - Functional Tests (Include Mode)](https://dev.azure.com/gvfs/ci/_build?definitionId=23&_a=summary) to run only the GitCommandTests with the IncludeMode option so we don't make the existing functional tests longer. This build can be turned on for `master` after this PR completes.

One thing the tests are missing is any behavior where Git does an interaction outside the included cone. We should consider setting up a merge conflict outside the cone and confirm the behavior is not terrible. We can create specific functional tests for those types of cases.

Finally, I'm testing some more to see if we can add a second-level included folder, such as "GVFS/GVFS". It is likely that some other sub-folders of the GVFS folder need to be added for the tests to work. I'll provide that change as a separate commit.

derrickstolee:
In [this commit](https://github.com/derrickstolee/VFSForGit/commit/c10e8ba2bf0ba53c6add2749f83f1928169c79b9), I converted the "GVFS" folder to "GVFS/GVFS" and "GVFS/GVFS.Common" and the [tests passed](https://dev.azure.com/gvfs/ci/_build/results?buildId=13525).

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            this.fileSystem.DeleteFile(fileToCreate);
+        }
+
+        [TestCase, Order(8)]
+        public void ModifiedFileShouldNotAllowIncludedFolderChange()
+        {
+            string modifiedPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts", "RunFunctionalTests.bat");
+            this.fileSystem.WriteAllText(modifiedPath, "New Contents");
+
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
```
jamill: Would we expect an error here (if gvfsProcess was not allowed to actually add an included folder)? Or maybe the actual return text of running this command does indicate there was an error.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            this.ValidateIncludedFolders(new string[0]);
+        }
+
+        [TestCase, Order(9)]
+        public void ModifiedFileAndCommitThenChangingIncludeFoldersShouldKeepFileAndFolder()
```
jamill: Seems like the high level behavior to validate is that folders containing modified files stay in the "include" list?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -408,10 +410,18 @@ public void OnFilePreDelete(string relativePath)
             this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnFilePreDelete(relativePath));
         }
 
-        public void OnFolderCreated(string relativePath)
+        public bool OnFolderCreated(string relativePath)
```
jamill: Can we include a comment on what the return value indicates?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -408,10 +410,18 @@ public void OnFilePreDelete(string relativePath)
             this.backgroundFileSystemTaskRunner.Enqueue(FileSystemTask.OnFilePreDelete(relativePath));
         }
 
-        public void OnFolderCreated(string relativePath)
+        public bool OnFolderCreated(string relativePath)
         {
+            GitIndexProjection.PathProjectionState pathProjectionState = this.GitIndexProjection.GetPathProjectionState(relativePath);
+            if (pathProjectionState == GitIndexProjection.PathProjectionState.Excluded)
+            {
+                this.GitIndexProjection.TryAddIncludedFolder(relativePath);
```
jamill: Do we care about success / failure here? When might this be ok to fail?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -524,7 +524,10 @@ private void OnNewFileCreated(string relativePath, bool isDirectory)
                         }
                         else
                         {
-                            this.FileSystemCallbacks.OnFolderCreated(relativePath);
+                            if (this.FileSystemCallbacks.OnFolderCreated(relativePath))
```
jamill: Should we include any comments on why we are calling `OnEnemerateDirectory` callback in response to file creation (when return value is true)?

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1003,31 +1003,9 @@ private HResult QueryFileNameHandler(string virtualPath)
                     if (isDirectory)
                     {
                         GitCommandLineParser gitCommand = new GitCommandLineParser(this.Context.Repository.GVFSLock.GetLockedGitCommand());
-                        if (gitCommand.IsValidGitCommand)
+                        if (gitCommand.IsValidGitCommand || this.FileSystemCallbacks.OnFolderCreated(virtualPath))
```
jamill: The new logic is subtle - mixing control logic with business functionality (i.e. including `OnFolderCreated(...)` inside the top control structure next to `gitCommand.IsValidGitCommand`). - I think it easier to read if we left the previous structure in place, even if not as compact. Just my thoughts - up to you :)

```
if (gitCommand.IsValidGitCommand)
{
    this.MarkDirectoryAsPlacholder(....)
}
else
{
    if (this.FileSystemCallbacks.OnFolderCreated(virtualPath))
    {
        this.MarkDirectoryAsPlacholder(....)
    }
}
```

GVFS/GVFS.Common/Database/IncludedFolderTable.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            this.connectionPool = connectionPool;
+        }
+
+        public static void CreateTable(IDbConnection connection)
+        {
+            using (IDbCommand command = connection.CreateCommand())
+            {
+                command.CommandText = "CREATE TABLE IF NOT EXISTS [IncludedFolder] (path TEXT PRIMARY KEY COLLATE NOCASE) WITHOUT ROWID;";
+                command.ExecuteNonQuery();
```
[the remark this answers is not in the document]
kewillford: Yes and it's try/catch is in the `GVFSDatabase` constructor.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
kewillford: Modified files get added to the ModifiedPaths list and right now the is only additive so once something is on that list it will stay there and git is in control of it regardless of what the include list is.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: Good point.  I will check the return code before returning `true`.  My one concern here would be that a user adds a folder that is excluded that would have millions of files and folders and if we fail adding it to the include it will now be in the modified paths and could cause performance issues for which they can't fix without a full dehydrate until we get the dehydrate folders finished.  But this shouldn't happen all that often so I am okay with this.

PR #1300, opened 2019-06-26
Update Git to include status deserialize fix

derrickstolee:
Includes status deserialziation bug fix from microsoft/git#157.

Further, update the dehydrate verb to specifically use `-uall` when checking for local changes.

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20190610.5</GitPackageVersion>
+    <GitPackageVersion>2.20190626.2</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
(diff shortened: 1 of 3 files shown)


kewillford:
We will need the changes in [this commit ](https://github.com/microsoft/VFSForGit/pull/1293/commits/1add9562d90d5a8f337e4c680f6a9de8fd51ee1f) so that if a user has the config for `status.showUntrackedFiles` set to `no`, they will be able to dehydrate when they have a clean working tree since a bare `git status` will now have `nothing to commit (use -u to show untracked files)` even when the working directory is clean and it is looking for `nothing to commit, working tree clean`.

derrickstolee:
> We will need the changes in [this commit ](https://github.com/microsoft/VFSForGit/pull/1293/commits/1add9562d90d5a8f337e4c680f6a9de8fd51ee1f) so that if a user has the config for `status.showUntrackedFiles` set to `no`, they will be able to dehydrate when they have a clean working tree since a bare `git status` will now have `nothing to commit (use -u to show untracked files)` even when the working directory is clean and it is looking for `nothing to commit, working tree clean`.

I did a cherry-pick, but dropped the change to `IncludeVerb` as I did that. Hopefully that's what you need.

PR #1301, opened 2019-06-26
simplify Git functional tests of case-sensitive file paths

chrisd8088:
The `FileShouldHaveCaseMatchingName()` and related `FolderShouldHaveCaseMatchingName()` methods used in the GitCommands functional tests currently take two arguments, first a file path, and then a file name which must match case-sensitively against the last component of the path.

On Linux, the use of at least one of these matching tests fails due to an incorrect usage: the caller,
`CaseOnlyRenameFileAndChangeBranches()`, is passing a file path which differs in case from the case-sensitive name.  This happens to work fine on Windows and Mac because the intended file is found regardless of case, but on Linux, it fails as the intended file can't be located.

In order to clarify the function of these matching test methods, we change them to take a single, case-sensitive file path, and then check that the file or folder found at that path has the matching case in its parent directory's listing.  This is effectively a no-op on Linux, but should retain the same functionality as before on other platforms.

/cc @kivikakk


chrisd8088:
Thanks, team!  ⚽️

Issue #1299, opened 2019-06-25
GVFSEnlistmentHealthTests refactoring and improvements


Halterer:
Increase name specificity: https://github.com/microsoft/VFSForGit/pull/1249#discussion_r297345713
Ensure behavior for directories with tied health scores (sorting alphanumerically) by checking to see their names: https://github.com/microsoft/VFSForGit/pull/1249#discussion_r297348302
Change the values being checked to be based on input data instead of hardcoded: https://github.com/microsoft/VFSForGit/pull/1249#discussion_r297365301
Refactor unit testing to use the [builder design](https://ardalis.com/improve-tests-with-the-builder-pattern-for-test-data): https://github.com/microsoft/VFSForGit/pull/1249#discussion_r297369251

PR #1302, opened 2019-06-27
Unit Test Refactoring

Halterer:
Implements the suggestions from issue https://github.com/microsoft/VFSForGit/issues/1299

 - A test case was renamed to better specify the case being tested.
 - Added a few more checks in test cases to test small missed behavior
 - Used the test data to determine expected results instead of hardcoded values to increase readability
 - Added comments in several test cases to increase the readability of checks
 - Refactored the test fixture to use the builder design pattern to create `GVFSEnlistmentPathData` objects in a much more expressive way
 - Slight change to `GVFSEnlistmentHealthCalculator` to allow arrays to be passed instead of only `List` objects

### DIFF
--- a/GVFS/GVFS.UnitTests/Common/GVFSEnlistmentHealthTests.cs
+++ b/GVFS/GVFS.UnitTests/Common/GVFSEnlistmentHealthTests.cs
@@ -1,7 +1,6 @@
 ﻿using GVFS.Common;
 using GVFS.Tests.Should;
 using NUnit.Framework;
-using System.Collections.Generic;
 
 namespace GVFS.UnitTests.Common
 {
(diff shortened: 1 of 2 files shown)


Halterer:
Not ready for review just yet, need to write a quick patch.

GVFS/GVFS.UnitTests/Common/GVFSEnlistmentHealthTests.cs
```diff
@@ -101,59 +105,68 @@ public void VariedDirectoryFormatting()
...
+                .AddPlaceholderFiles("A/1.js", $"A{this.sep}2.js", "/A/1.js", $"{this.sep}A{this.sep}1.js")
+                .AddModifiedPathFiles("B/1.js", $"B{this.sep}2.js", "/B/1.js", $"{this.sep}B{this.sep}1.js")
+                .Build();
 
-            this.enlistmentHealthData = this.GenerateStatistics("A/");
+            this.enlistmentHealthData = this.GenerateStatistics(pathData, "A/");
 
-            this.enlistmentHealthData.GitTrackedItemsCount.ShouldEqual(3);
-            this.enlistmentHealthData.PlaceholderCount.ShouldEqual(4);
+            this.enlistmentHealthData.GitTrackedItemsCount.ShouldEqual(pathData.GitFilePaths.Count / 2);
```
jamill: instead of `pathData.GitFilePaths.Count / 2`, could we do something like `expectedCountofFilesInDirectoryA`, or possibly do something like:

```
string [] filesPathsDirectoryA = new string [ {"A/1.js", "A/2.js", ....};
string [] dilesPathsDirectoryB = new string [ {"B/1.js", "/B/2.js", ....};

new PathDataBuild().AddGitFiles(filePathsDirectoryA).AddGitFiles(filePathsDirectoryB)....
this.enlistmentHealthData.GitTrackedItemsCount.ShouldEqual(filesPathsDirectoryA.Count);
```

Issue #1298, opened 2019-06-25
Small readability improvements for GVFSEnlistmentHealthCalculator


Halterer:
https://github.com/microsoft/VFSForGit/pull/1249#discussion_r297353860: Out of date comment
https://github.com/microsoft/VFSForGit/pull/1249#discussion_r297355484: Poorly worded precondition comment

PR #1303, opened 2019-06-28
Health Feature Pull Request

Halterer:
Open PR for the merge of `features/health` into `master` to track activity.

The first full version of `gvfs health` to be merged into master. The feature now has (windows only) functional tests, as well as unit tests. The primary health metrics the total number of hydrated files, and the verb can filter into a specified directory for more localized information and troubleshooting.

To-do list:

- [x] Refactor unit tests (https://github.com/microsoft/VFSForGit/issues/1299), addressed by PR https://github.com/microsoft/VFSForGit/pull/1302
- [x] Move the health feature files into a distinct folder within `GVFS.Common` and pull out inner classes to simplify naming, addressed by PR https://github.com/microsoft/VFSForGit/pull/1307
- [x] Create functional tests, addressed by PR https://github.com/microsoft/VFSForGit/pull/1313
- [x] Update and ensure comment quality (https://github.com/microsoft/VFSForGit/issues/1298), addressed by PR https://github.com/microsoft/VFSForGit/pull/1321
- [x] Update health metric, addressed by PR https://github.com/microsoft/VFSForGit/pull/1325


Halterer:
Removing to to-do for "Pull the builder pattern into `GVFS.Common` and use it in `GVFSEnlistmentHealthCalculator` as an unnecessary level of abstraction

Halterer:
In the hopes of getting this larger set of changes merged in before the M155 release, I'm going to punt on several of the refactoring work items. They remain as issues assigned to me to be circled back to.

Documents/AnalyticsToolDesignDocument.md
```diff
@@ -0,0 +1,207 @@
...
+This document details the design of and plan for the **VFS4G analytics tool**. The tool aims to help users identify reasons why their enlistment may be slowing down and the `git` commands running slowly on a VFS4G mounted repository. By collecting and reporting statistics, the power to uncover and hopefully fix problems as they arrive moves to the user, empowering them to use the software to its fullest potential.
+
+## Problem
+
+Over the lifetime of an enlistment, sometimes a user's working directory fills up with too many entries. Generally this means having too many placeholders on their disk (have been read and pulled from remote) or too many modified files. This causes unacceptable losses in performance and greatly hinders users as they go about their daily development cycle. Additionally, if a local enlistment has been in use for a long time, the number of loose object may eventually reach the order of several thousand in which case that also contributes to the loss in performance.
+
+## End Goal
+
+```terminal
+$ gvfs statistics
```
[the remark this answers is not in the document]
jamill: Thanks for pointing this out - Including an out-of-date document could add to confusion. I would remove this design document from this change. We can include the updated design doc in a future PR (if we want to include it with the repository at all - it does not look like we include many design docs in the actual repository).

jamill:
@kewillford  Thanks for trying this out!

> On the same repo where I opened all the files for write, I deleted a main directory and brought it back by running git reset --hard. After that I would have expected the hydration percentage to be 100% still but got this even though all the files in the index have the skipworktree bit off still.

It looks like the current algorithm only includes the actual files in PlaceHolders / Modified paths lists, and does not include the files with the skip worktree bit turned off. I think the `CalculateStatistics` method will need to be tweaked to account for this.

Halterer:
@kewillford Thanks for all of the great feedback! I've been reading over it and plan to implement several changes when I return to the office. Including a spinner or some kind of progress indicator is a really good idea since it can take a while to run. I will certainly add a message for when a file does not exist, that was a good catch and maybe we want a functional test for that? I will also remove the design document from the set of changes, I didn't mean for it to get in. I'll look into the wording as specified in your feedback #3, but yeah there it just refers to items, so folders are included. I will also fix the spacing issue with commas in the list.

Halterer:
> @kewillford Thanks for trying this out!
> 
> > On the same repo where I opened all the files for write, I deleted a main directory and brought it back by running git reset --hard. After that I would have expected the hydration percentage to be 100% still but got this even though all the files in the index have the skipworktree bit off still.
> 
> It looks like the current algorithm only includes the actual files in PlaceHolders / Modified paths lists, and does not include the files with the skip worktree bit turned off. I think the `CalculateStatistics` method will need to be tweaked to account for this.

Currently, the verb is supposed to handle the files with the skip worktree bit turned off. Its in the same function that handles the git index, `GetPathsFromGitIndex()`. The path data objects takes them in

jamill:
> Its in the same function that handles the git index, GetPathsFromGitIndex(). The path data objects takes them in

ah - one issue is with the [AddGitTrackingPaths](https://github.com/microsoft/VFSForGit/pull/1303/files#diff-46d4cb26b63a1016eb97d4204a61b8b5R27) method:

```
        public void AddGitTrackingPaths(IEnumerable<string> gitTrackingPaths)
        {
            this.ModifiedFilePaths.Union(gitTrackingPaths);
        }
```

The `Union` method does not add the items to the existing list - it generates a new result.

Halterer:
Pushed a patch which should fix the issues with the non skip worktree files not being correctly included (thanks @kewillford for the steps to reproduce that particular situation). Also fixed the formatting issue with commas in the list of sub directories. Going to work on the UX changes suggested above.

GVFS/GVFS.Common/HealthCalculator/EnlistmentPathData.cs
```diff
@@ -0,0 +1,56 @@
...
+            this.PlaceholderFilePaths = new List<string>();
+            this.ModifiedFolderPaths = new List<string>();
+            this.ModifiedFilePaths = new List<string>();
+        }
+
+        public void AddGitTrackingPaths(IEnumerable<string> gitTrackingPaths)
+        {
+            // Ensure that both lists of paths have been normalized before unioning to ensure no duplicity
+            List<string> normalizedGitTrackingPaths = gitTrackingPaths.ToList();
+            this.NormalizePaths(normalizedGitTrackingPaths);
```
jamill: Having multiple places where we normalize paths makes it hard to reason about when we are dealing with normalized vs un-normalized paths, and results in duplicate normalization calls for the same set of files.

Can we simplify the flow here? maybe by unioning the data after everything is normalized via the `NormalizePaths` method (maybe by stroring the paths with skip worktree bit set)?

GVFS/GVFS.Common/HealthCalculator/EnlistmentHealthCalculator.cs
```diff
@@ -0,0 +1,162 @@
...
+            }
+
+            gitTrackedItemsCount += this.CategorizePaths(this.enlistmentPathData.GitFolderPaths, gitTrackedItemsDirectoryTally, parentDirectory);
+            gitTrackedItemsCount += this.CategorizePaths(this.enlistmentPathData.GitFilePaths, gitTrackedItemsDirectoryTally, parentDirectory);
+            placeholderCount += this.CategorizePaths(this.enlistmentPathData.PlaceholderFolderPaths, hydratedFilesDirectoryTally, parentDirectory);
+            placeholderCount += this.CategorizePaths(this.enlistmentPathData.PlaceholderFilePaths, hydratedFilesDirectoryTally, parentDirectory);
+            modifiedPathsCount += this.CategorizePaths(this.enlistmentPathData.ModifiedFolderPaths, hydratedFilesDirectoryTally, parentDirectory);
+            modifiedPathsCount += this.CategorizePaths(this.enlistmentPathData.ModifiedFilePaths, hydratedFilesDirectoryTally, parentDirectory);
+
+            Dictionary<string, int> mostHydratedDirectories = new Dictionary<string, int>();
```
jamill: For consistency, we might want to explicitly include the string comparer. We are not looking up values in the dictionary at the moment, but it is better to specify the correct string comparer in case someone needs to use that functinality in the future.

jeffhostetler:
Hey, just had a chance to take a look at this and saw a few things that I thought I'd mention.

[1] The total "total files in head commit" seems to be the sum of the files and directories, rather than just the number of files.  Perhaps, print 2 line items here rather than the sum.

[2] Likewise, in the "files managed by *" lines, are these also sums of the files and directories?
If you have access to the data, perhaps, split these into sum-file and sum-dir fields.

[3] And I think the "Most hydrated..." lines are similarly sums of file and directory counts.
Maybe have separate columns for these.

I'd like to be able to compare your numbers with mine and it would help if they were
broken out by type.

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
```diff
@@ -92,6 +92,18 @@ public string Status(string trace = null)
...
 
+        public string Health(string directory = null)
+        {
+            if (string.IsNullOrEmpty(directory))
+            {
+                return this.CallGVFS("health " + this.enlistmentRoot);
+            }
+            else
+            {
+                return this.CallGVFS("health -d " + directory + " " + this.enlistmentRoot);
```
derrickstolee: Do any of the paths in these commands need their directories quoted? What if someone has this inside a directory with a space?

jamill:
This looks good to me - I can't wait for people to get to use this :) Thanks!

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
Halterer: From what I have seen - and perhaps more to the point, with the directories I chose to use when writing the functional tests, they all have underscores instead of spaces. It is possible that if someone were to edit the functional tests they could point it to a directory that had a space in which case it seems like they would need to be quoted. Is this an issue we could see potentially happening?

derrickstolee: The issue is that `this.enlistmentRoot` could have a space in it. If `directory` is always local to the test repo, then there will not be space there.

Halterer: Ah alright, that makes sense. I can add in the quotes around `this.enlistmentRoot` to ensure that isn't a problem. Does it make sense to add them around `directory` just in case? Otherwise, I can leave it as is.

derrickstolee:
@halterer please create an identical PR to merge `features/health` into `milestones/M155` so this can make the M155 cutoff. Thanks!

PR #1305, opened 2019-06-28
createMacInstaller: generate standalone git pkg installers

jamill:
When creating the mac installer, downstream consumers might want to
use the git for mac pkg installer. Copy this pkg installer to the
output directory.

### DIFF
--- a/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
+++ b/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
@@ -211,6 +211,10 @@ function CreateMetaDistribution()
     eval $copyGitPkgCmd
 
     UpdateDistributionFile "$GITPKGNAME" "$GITVERSIONSTRING"
+
+    copyGitPkgInstallerCmd="/bin/cp \"${GITINSTALLERPATH}\" \"${BUILDOUTPUTDIR}/\""
+    echo $copyGitPkgInstallerCmd
+    eval $copyGitPkgInstallerCmd || exit 1
     
     METAPACKAGENAME="$INSTALLERPACKAGENAME-Git.$GITVERSION.pkg"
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -211,6 +211,10 @@ function CreateMetaDistribution()
     eval $copyGitPkgCmd
 
     UpdateDistributionFile "$GITPKGNAME" "$GITVERSIONSTRING"
+
```
jamill: I will probably break this out into a more appropriate function (or maybe have a function that pulls out the pkg from the dmg, and then we can create the meta-installer separately).

I also was going to look at having the git for mac build publish the .pkg directly, so we did not need to extract it from the dmg

wilbaker:
> Copy this pkg installer to the output directory.

I am unfamiliar with how the installers are built.  What does copying the pkg installer to the output directory do?  Does something else come along and pick up things from that directory?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -211,6 +211,10 @@ function CreateMetaDistribution()
...
+    copyGitPkgInstallerCmd="/bin/cp \"${GITINSTALLERPATH}\" \"${BUILDOUTPUTDIR}/\""
```
kewillford: The copy above uses `-Rf` does this one need either of those options?

jamill:
>  Does something else come along and pick up things from that directory?

Yes - our build pipeline will publish everything in that directory as an artifact, so it will be available for other stages / pipelines to consume. This will allow downstream steps to consume this pkg.

jamill:
After further investigation, I don't think we actually need this change at this point. We already publish the necessary packages, but in a slightly different location.

PR #1307, opened 2019-06-28
Reorganize health assets

Halterer:
Renamed the following classes:
- `GVFSEnlistmentHealthCalculator` -> `EnlistmentHealthCalculator`
- `GVFSEnlistmentPathData` -> `EnlistmentPathData`
- `GVFSEnlistmentHealthData` -> `EnlistmentHealthData`

Extracted `EnlistmentPathData` and `EnlistmentHealthData` to be standalone
classes instead of inner classes of `EnlistmentHealthCalculator`.
Created a `GVFS.Common.HealthCalculator` directory to house the 3 health classes

### DIFF
--- a/GVFS/GVFS.Common/HealthCalculator/EnlistmentHealthCalculator.cs
+++ b/GVFS/GVFS.Common/HealthCalculator/EnlistmentHealthCalculator.cs
@@ -10,18 +10,18 @@ namespace GVFS.Common
     /// internally stores the calculated information. Compute or recompute via CalculateStatistics
     /// with an optional parameter to only look for paths which are under the specified directory
     /// </summary>
-    public class GVFSEnlistmentHealthCalculator
+    public partial class EnlistmentHealthCalculator
     {
         // In the context of this class, hydrated files are placeholders or modified paths
         // The total number of hydrated files is this.PlaceholderCount + this.ModifiedPathsCount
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.Common/HealthCalculator/EnlistmentHealthCalculator.cs
```diff
@@ -10,18 +10,18 @@ namespace GVFS.Common
     /// internally stores the calculated information. Compute or recompute via CalculateStatistics
     /// with an optional parameter to only look for paths which are under the specified directory
     /// </summary>
-    public class GVFSEnlistmentHealthCalculator
+    public partial class EnlistmentHealthCalculator
```
kewillford: Why is this class now `partial`?

Halterer: Oh whoops that was a line I missed when doing the refactoring. I tried to use the built in tools to extract inner classes, but it just created a new file and left them as members of `EnlistmentHealthCalculator` but in new files, hence why it was automatically modified to be a partial class. I missed that line when I went back through to change what had been automatically refactored.

Halterer: I'll very quickly push that fix up.

PR #1308, opened 2019-07-01
Updating README.md

turbonaitis:
Now, that local cache path defaults to `$HOME/.gvfsCache`, there's no need to specify --local-cache-path explicitly.


turbonaitis:
> Thanks for this!

It's nothing :) But I see that the funcitonal tests are failing and so I can't merge.

turbonaitis:
@derrickstolee I don't believe my change to README.md can trigger a functional test failure - could you retrigger them?

PR #1312, opened 2019-07-02
Use the new signing certificate.

jrbriggs:
The `Microsoft` certificate is a SHA1 certificate. Move to the hybrid certificate.


wilbaker:
How did you know you should make this change?

jrbriggs:
> How did you test this change or are you going to run the signing build to test it?

@kewillford  I'll run a real-sign build once we're into a release branch. I'm reluctant to generate a signed build off a topic branch.

>How did you know you should make this change?

@wilbaker Owners of real-signed builds were instructed to move away from the SHA1 certs, and there is a table provided for which cert name to move to. I moved from `Microsoft` to `Microsoft400` as per the table.

PR #1313, opened 2019-07-03
Health functional tests

Halterer:
Functional tests for the health verb are now included in the solution. There are five tests:

Freshly cloned repository
Placeholders
Modified paths
Placeholders turning into modified paths
Filtering by a target directory

There is an according bug fix to the way non skip-worktree files are parsed ensuring all folders are correctly included in the returned list.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
+++ b/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
@@ -32,6 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
     <ProjectReference Include="..\GVFS.Tests\GVFS.Tests.csproj" />
   </ItemGroup>
 
--- a/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
+++ b/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\GVFS.Build\GVFS.cs.props" />
   <PropertyGroup>
     <OutputType>Exe</OutputType>
--- a/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
+++ b/GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\GVFS.Build\GVFS.cs.props" />
   <PropertyGroup>
     <OutputType>Exe</OutputType>
@@ -32,7 +32,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
     <ProjectReference Include="..\GVFS.Tests\GVFS.Tests.csproj" />
   </ItemGroup>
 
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj
```diff
@@ -32,6 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\GVFS.Common\GVFS.Common.csproj" />
```
jamill: What is included from `GVFS.Common.csproj`? The functional tests historically have not taken a dependency on product code.

Halterer: I think I accidentally imported it once while I was trying to look at something, and even though I removed it from `HealthTests.cs` it looks like it ended up staying there. I looked at the `git diff` this time to try and make sure I didn't accidentally push anything which I meant to remove but apparently I missed this. After you're done reviewing I'll push a commit removing this dependency here

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/HealthTests.cs
```diff
@@ -0,0 +1,143 @@
...
+    public class HealthTests : TestsWithEnlistmentPerFixture
+    {
+        [TestCase, Order(0)]
+        public void AfterCloningRepoIsPerfectlyHealthy()
+        {
+            // .gitignore is always a placeholder on creation
+            // .gitconfig is always a modified path in functional tests since it is written at run time
+            string[] healthOutputLines = this.GetHealthOutputLines();
+
+            healthOutputLines[0].ShouldEqual("Health of directory: ");
```
jamill: It looks like there is a lot of common validation between each of the methods - can we put it into a helper method:

```
ValidatedHealthOutput(string directory, int totalFiles, int fastFiles, int slowFiles, List<ExpectedDirectoryInfo> expectedDirInfo, bool isHealthy)
```

We would have to handle calculating the "expected" percentage as well, or pass that in separately.

Halterer: I'm going to go ahead and punt this for now. The output format is expected to change, potentially even early next week so I don't think it makes sense to invest time in creating a method or class for it just yet until that becomes settled. When the output is more finalized, the functional tests will also be updated to reflect that

jamill: > The output format is expected to change

This is a reason to consolidate the validation logic into a single method. As is, if the output changes, you need to modify every test method. If you consolidate the validation into a single method, then you only need to update 1 location to react to the output changing.

Halterer: So then in the new function, also copy over all of the behavior for padding the lines correctly and formatting directory names as well?

kewillford: This is one of the reasons that I suggested to implement the `--json` option first might make the functional tests easier to write because we don't care about all the formatting of the output.  We only care that the directories, numbers, and percentages are correct.  Json would make it simple to verify.

jamill: I see - what I was thinking was that we would validate the general information in the output, rather than the specific formatting at this time.

At this point, my feeling is the most value of these tests would come from validating that we can run the command and get the high level data back that we expected. I would rather we validate the high level aspects (and be more resilient to changes in the structure of the presentation), rather then validate the _exact_ output, that we would need to update (in several places) with each change to the presentation of the data.

What do you think?

Halterer: One option is to delegate to functions like `EnsureTotalFileInfo(int expectedTotalFiles, decimal expectedTotalFilePercent)` from within a larger function like you said.

Halterer: The other thought is that when I update the output format to either json or whatever else, I will be nuking all functionality and nuking it in 5 places is not much different from one, especially if that one involves a lot of refactoring. I could see it going either way. I'm just trying to ensure all baseline functionality is in before next Friday so it will make it to the release that gets cut.

PR #1315, opened 2019-07-04
Fix GitProcess PATH

gonix:
- Use platform dependent PATH separator.
- Do not prepend `this.gitBinPath` which is path to binary (not a directory) and has no effect in PATH resolution anyway.
- Passthrough entire PATH to child process. Git might call 3rd party helpers (e.g. credentials helpers) that expect standard POSIX utilities to be available in PATH.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -675,11 +675,10 @@ namespace GVFS.Common.Git
 
             processInfo.EnvironmentVariables["GIT_TERMINAL_PROMPT"] = "0";
             processInfo.EnvironmentVariables["GCM_VALIDATE"] = "0";
-            processInfo.EnvironmentVariables["PATH"] =
-                string.Join(
-                    ";",
-                    this.gitBinPath,
-                    this.gvfsHooksRoot ?? string.Empty);
+            if (!string.IsNullOrWhiteSpace(this.gvfsHooksRoot))
+            {
+                processInfo.EnvironmentVariables["PATH"] = this.gvfsHooksRoot + Path.PathSeparator + processInfo.EnvironmentVariables["PATH"];
+            }
 
             if (gitObjectsDirectory != null)
             {


gonix:
It looks like `GitHub VFSForGit Mac Functional Tests` job is stuck (for 6 days). Is it possible to restart it?

gonix:
Looks like I don't have access to see why `GitHub VFSForGit Mac Functional Tests` failed.
Any pointers what went wrong?

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -675,11 +675,10 @@ public Process GetGitProcess(string command, string workingDirectory, string dot
...
             processInfo.EnvironmentVariables["GCM_VALIDATE"] = "0";
-            processInfo.EnvironmentVariables["PATH"] =
-                string.Join(
-                    ";",
-                    this.gitBinPath,
-                    this.gvfsHooksRoot ?? string.Empty);
+            if (!string.IsNullOrWhiteSpace(this.gvfsHooksRoot))
+            {
+                processInfo.EnvironmentVariables["PATH"] = this.gvfsHooksRoot + Path.PathSeparator + processInfo.EnvironmentVariables["PATH"];
+            }
```
kewillford: Looking back through the history of why this was added and it was when the code was using `git cat-file` to hydrate files which we no longer use.  It was causing an infinite loop when the `PATH` contained a path in the Repo.  I just tested that scenario and everything worked correctly with this code remove, so I'm thinking that we can completely delete this code.  I'm also hoping we can delete the `gvfsHooksRoot` member here as well as in the `Enlistment`.  I will give that a try and see.

kewillford: [Here are the changes](https://github.com/microsoft/VFSForGit/compare/master...kewillford:remove-hooks-path) that I made removing this code as well as all references to the `gvfsHooksRoot` and ran all tests against it as well as some manual testing and didn't hit any issues.  @wilbaker any reason you can think of that we can't remove this completely?

turbonaitis: How could we move forward with this change? If the changes mentioned above work better, shall @kewillford create a new pull request and we abandon this one?

kewillford: I will open a PR for the changes I made.  I still need to do a bit more testing with it to make sure we aren't going to cause any issues with the large repos and their tools with how they are using the `PATH`.

turbonaitis:
Thanks for the update! That, I guess, means, that we should abandon this
PR, as this code is going to be removed altogether?

gonix:
Abandoning this in favor of #1351.

PR #1317, opened 2019-07-08
Setup.iss: Update GitHub URL

derrickstolee:
I don't know where this actually shows up as part of the generated installer, but it can't hurt to update this correctly.


GVFS/GVFS.Installer.Windows/Setup.iss
```diff
@@ -22,7 +22,7 @@
 #define MyAppInstallerVersion GetFileVersion(GVFSDir + "\GVFS.exe")
 #define MyAppPublisher "Microsoft Corporation"
 #define MyAppPublisherURL "http://www.microsoft.com"
-#define MyAppURL "https://github.com/Microsoft/gvfs"
+#define MyAppURL "https://github.com/Microsoft/vfsforgit"
```
jrbriggs: One tiny nit, I think the canonical capitalization is now
```suggestion
#define MyAppURL "https://github.com/microsoft/VFSForGit"
```

PR #1318, opened 2019-07-09
Mac build scripts: consume git installer pkg directly

jamill:
Update VFS4G build scripts to consume the macOS pkg installer package directly

Previously, the GitForMac.GVFS.Installer NuGet package contained only
the git for mac installer as a .dmg file. This .dmg package contained
the Git for Mac installer .pkg.

The VFS4G installer takes a .pkg installer as input to the combined
VFS4G / Git for Mac installer. As the NuGet package only contained the
.dmg, the script to build the VFS4G combined installer had to mount
the .dmg and extract the .pkg file.

The GitForMac.GVFS.Installer package has been updated to include both
a .dmg and .pkg file, so the installer script can now consume the
installer .pkg directory, instead of mounting and extracting it from
the .dmg file.

This change also relaxes the regex pattern to correctly handle git versions with multiple hyphens. Previously, this would fail to capture the full version string for versions such as:

1.2.3-DRAFT-pr

### DIFF
--- a/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
+++ b/GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
@@ -179,49 +179,34 @@ function CreateVFSForGitDistribution()
 function CreateMetaDistribution()
 {
     GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
-    GITDMGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.dmg)" || exit 1
-    GITDMGNAME="${GITDMGPATH##*/}"
-    GITINSTALLERNAME="${GITDMGNAME%.dmg}"
-    GITVERSIONSTRING=`echo $GITINSTALLERNAME | cut -d"-" -f2`
-    GITPKGNAME="$GITINSTALLERNAME.pkg"
-    
+    GITINSTALLERPKGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.pkg)" || exit 1
+
+    GITPKGNAME="${GITINSTALLERPKGPATH##*/}"
+    GITINSTALLERPKGNAME="${GITPKGNAME%.pkg}"
+    GITVERSIONSTRING=`echo $GITINSTALLERPKGNAME | cut -d"-" -f2`
+
     if [[ -z "$GITVERSION" || -z "$GITVERSIONSTRING" ]]; then
         echo "Error creating metapackage: could not determine Git package version."
         exit 1
     fi
     
-    if [ ! -f "$GITDMGPATH" ]; then
-        echo "Error creating metapackage: could not find Git disk image."
-        exit 1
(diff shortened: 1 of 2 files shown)


wilbaker:
Could you update the description with more details on why these changes were needed?

Scripts/Mac/GetGitVersionNumber.sh
```diff
@@ -1,5 +1,5 @@
 . "$(dirname ${BASH_SOURCE[0]})/InitializeEnvironment.sh"
 
 GVFSPROPS=$VFS_SRCDIR/GVFS/GVFS.Build/GVFS.props
-GITVERSION="$(cat $GVFSPROPS | grep GitPackageVersion | grep -Eo '[0-9.]+(-\w+)?')"
-echo $GITVERSION
\ No newline at end of file
+GITVERSION="$(cat $GVFSPROPS | grep GitPackageVersion | grep -Eo '[0-9.]+(-\w+)*')"
```
wilbaker: Why was this change needed?

GVFS/GVFS.Installer.Mac/CreateMacInstaller.sh
```diff
@@ -179,49 +179,34 @@ function CreateVFSForGitDistribution()
 function CreateMetaDistribution()
 {
     GITVERSION="$($VFS_SCRIPTDIR/GetGitVersionNumber.sh)"
-    GITDMGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.dmg)" || exit 1
-    GITDMGNAME="${GITDMGPATH##*/}"
-    GITINSTALLERNAME="${GITDMGNAME%.dmg}"
-    GITVERSIONSTRING=`echo $GITINSTALLERNAME | cut -d"-" -f2`
-    GITPKGNAME="$GITINSTALLERNAME.pkg"
-    
+    GITINSTALLERPKGPATH="$(find $VFS_PACKAGESDIR/gitformac.gvfs.installer/$GITVERSION -type f -name *.pkg)" || exit 1
```
wilbaker: I've not been following much of the installer discussions, what are the pros of a .pkg (over a .dmg)?

jamill:
> Could you update the description with more details on why these changes were needed?

Thanks - the commits themselves contained descriptive descriptions, but I didn't update the PR description. I updated the PR description to be more meaningful.

Scripts/Mac/GetGitVersionNumber.sh
wilbaker: See the answer in the description now, thanks!

wilbaker:
> The GitForMac.GVFS.Installer package has been updated to include both
a .dmg and .pkg file, so the installer script can now consume the
installer .pkg directory, instead of mounting and extracting it from
the .dmg file.

Is there something/someone who is consuming the .dmgs?  I am not sure if making a .dmg available is a standard Apple thing.

jamill:
> Is there something/someone who is consuming the .dmgs? I am not sure if making a .dmg available is a standard Apple thing.

Great question. After we transition this build off of that requirement, maybe we can stop publishing the .dmg.

PR #1319, opened 2019-07-09
Don't change version number for dev builds

alameenshah:
Apple's `agvtool` changes version number contained in `Info.plist` files during build. `Info.plist` is part of `VFSForGit` sources and the version changes created by the `agvtool` end up creating noise in `git` command outputs.

This PR updates the build script to not trigger version update when running dev builds.

### DIFF
--- a/Scripts/Mac/BuildGVFSForMac.sh
+++ b/Scripts/Mac/BuildGVFSForMac.sh
@@ -7,9 +7,13 @@ if [ -z $CONFIGURATION ]; then
   CONFIGURATION=Debug
 fi
 
+runVersionUpdater="yes"
 VERSION=$2
 if [ -z $VERSION ]; then
   VERSION="0.2.173.2"
+  
+  # don't update version number for developer builds
+  runVersionUpdater="no"
 fi
 
 if [ ! -d $VFS_OUTPUTDIR ]; then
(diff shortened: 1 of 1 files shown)


Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -59,9 +59,12 @@ xcodebuild -configuration $CONFIGURATION -workspace $NATIVEDIR/GVFS.Native.Mac.x
 USERNOTIFICATIONDIR=$VFS_SRCDIR/GVFS/GVFS.Notifications/VFSForGit.Mac
 USERNOTIFICATIONPROJECT="$USERNOTIFICATIONDIR/VFSForGit.xcodeproj"
 USERNOTIFICATIONSCHEME="VFS For Git"
-updateAppVersionCmd="(cd \"$USERNOTIFICATIONDIR\" && /usr/bin/xcrun agvtool new-marketing-version \"$VERSION\")"
-echo $updateAppVersionCmd
-eval $updateAppVersionCmd
+# don't update version number for developer builds
+if [[ $VERSION != 0* ]]; then
```
jrbriggs: Our releases to mac early adopters are numbered 0.6.xxxxx.y. Can we make this test whether a version was explicitly provided to the script instead?

alameenshah: Updated to only run version updater tool when an explicit version is provided.

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -59,9 +61,12 @@ xcodebuild -configuration $CONFIGURATION -workspace $NATIVEDIR/GVFS.Native.Mac.x
 USERNOTIFICATIONDIR=$VFS_SRCDIR/GVFS/GVFS.Notifications/VFSForGit.Mac
 USERNOTIFICATIONPROJECT="$USERNOTIFICATIONDIR/VFSForGit.xcodeproj"
 USERNOTIFICATIONSCHEME="VFS For Git"
-updateAppVersionCmd="(cd \"$USERNOTIFICATIONDIR\" && /usr/bin/xcrun agvtool new-marketing-version \"$VERSION\")"
-echo $updateAppVersionCmd
-eval $updateAppVersionCmd
+# don't update version number for developer builds
```
jamill: Should this comment be up where we set this variable?

Scripts/Mac/BuildGVFSForMac.sh
```diff
@@ -59,9 +61,12 @@ xcodebuild -configuration $CONFIGURATION -workspace $NATIVEDIR/GVFS.Native.Mac.x
...
+if [ "$runVersionUpdater" == "yes" ]; then
+    updateAppVersionCmd="(cd \"$USERNOTIFICATIONDIR\" && /usr/bin/xcrun agvtool new-marketing-version \"$VERSION\")"
+    echo $updateAppVersionCmd
+    eval $updateAppVersionCmd
```
jamill: Do we need to check the return code of this command (and exit on failure?)

Scripts/Mac/BuildGVFSForMac.sh
alameenshah: that would be more appropriate, updated

Scripts/Mac/BuildGVFSForMac.sh
alameenshah: added an exit on error.

PR #1323, opened 2019-07-10
Dehydrate folders

kewillford:
### Allow the dehydration of specific folders
This change allows the dehydration of specific folders in the enlistment instead of having to dehydrate everything.  There is a lot happening that is from how dehydrate worked before, the main points now are:
1. Check status to make sure we are in a clean state.
2. Unmount
3. Delete folders passed in.  _Should this be a move to a backup location?_
4. Mount
5. Send message with list of folders to mount process for dehydration
    1. Remove modified paths for the folders
    2. Remove placeholders for the folders
    3. Write a new placeholder folder

#### Do we need to backup anything?
_My thought is we don't need to since we ran status and it came back clean no data should be lost. But do we need and could we have a rollback if something goes wrong?_
#### What do we do when there is an error for one of the folders?
_I suppose there are different actions depending on where this happens.  For example if deleting the folder fails we don't need to send it to the mount process for dehydration but still need to let the user know.  But if something happens in the dehydration in the mount process what should be done?_
### Todos
- [x] Functional tests
- [x] Error Handling
- [x] Clean up UI messages
- [x] Testing on Mac

### DIFF
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -28,11 +28,6 @@ namespace GVFS.Common.Database
             }
         }
 
-        public static string NormalizePath(string path)
-        {
-            return path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar).Trim().Trim(Path.DirectorySeparatorChar);
-        }
-
         public int GetCount()
         {
             try
@@ -159,7 +154,7 @@ namespace GVFS.Common.Database
         public List<IPlaceholderData> RemoveStartingWith(string path)
         {
             // Normalize the path to match what will be in the database
-            path = NormalizePath(path);
+            path = GVFSDatabase.NormalizePath(path);
 
             try
             {
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Data;
+using System.IO;
 
 namespace GVFS.Common.Database
 {
@@ -25,6 +26,11 @@ namespace GVFS.Common.Database
                 command.CommandText = $"CREATE TABLE IF NOT EXISTS [Placeholder] (path TEXT PRIMARY KEY{collateConstraint}, pathType TINYINT NOT NULL, sha char(40) ) WITHOUT ROWID;";
                 command.ExecuteNonQuery();
             }
+        }
+
+        public static string NormalizePath(string path)
+        {
+            return path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar).Trim().Trim(Path.DirectorySeparatorChar);
         }
 
         public int GetCount()
@@ -161,6 +167,9 @@ namespace GVFS.Common.Database
 
         public void RemoveStartingWith(string path)
         {
+            // Normalize the path to match what will be in the database
+            path = NormalizePath(path);
+
             try
             {
                 using (IDbConnection connection = this.connectionPool.GetConnection())
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -153,6 +153,8 @@ namespace GVFS.Common.Database
 
         public List<IPlaceholderData> RemoveAllEntriesForFolder(string path)
         {
+            const string fromWhereClause = "FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirectorySeparator;";
+
             // Normalize the path to match what will be in the database
             path = GVFSDatabase.NormalizePath(path);
 
@@ -162,12 +164,12 @@ namespace GVFS.Common.Database
                 using (IDbCommand command = connection.CreateCommand())
                 {
                     List<IPlaceholderData> removedPlaceholders = new List<IPlaceholderData>();
-                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirecotrySeparator;";
+                    command.CommandText = $"SELECT path, pathType, sha {fromWhereClause}";
                     command.AddParameter("@path", DbType.String, $"{path}");
-                    command.AddParameter("@pathWithDirecotrySeparator", DbType.String, $"{path + Path.DirectorySeparatorChar}%");
+                    command.AddParameter("@pathWithDirectorySeparator", DbType.String, $"{path + Path.DirectorySeparatorChar}%");
                     ReadPlaceholders(command, data => removedPlaceholders.Add(data));
 
-                    command.CommandText = "DELETE FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirecotrySeparator;";
+                    command.CommandText = $"DELETE {fromWhereClause}";
 
                     lock (this.writerLock)
                     {
@@ -179,7 +181,7 @@ namespace GVFS.Common.Database
             }
             catch (Exception ex)
             {
-                throw new GVFSDatabaseException($"{nameof(PlaceholderTable)}.{nameof(this.Remove)}({path}) Exception", ex);
+                throw new GVFSDatabaseException($"{nameof(PlaceholderTable)}.{nameof(this.RemoveAllEntriesForFolder)}({path}) Exception", ex);
             }
         }
 
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -54,36 +54,27 @@ namespace GVFS.Common.Database
         {
             try
             {
-                filePlaceholders = new List<IPlaceholderData>();
-                folderPlaceholders = new List<IPlaceholderData>();
+                List<IPlaceholderData> tempFilePlaceholders = new List<IPlaceholderData>();
+                List<IPlaceholderData> tempFolderPlaceholders = new List<IPlaceholderData>();
                 using (IDbConnection connection = this.connectionPool.GetConnection())
                 using (IDbCommand command = connection.CreateCommand())
                 {
                     command.CommandText = "SELECT path, pathType, sha FROM Placeholder;";
-                    using (IDataReader reader = command.ExecuteReader())
+                    ReadPlaceholders(command, data =>
                     {
-                        while (reader.Read())
+                        if (data.PathType == PlaceholderData.PlaceholderType.File)
                         {
-                            PlaceholderData data = new PlaceholderData();
-                            data.Path = reader.GetString(0);
-                            data.PathType = (PlaceholderData.PlaceholderType)reader.GetByte(1);
-
-                            if (!reader.IsDBNull(2))
-                            {
-                                data.Sha = reader.GetString(2);
-                            }
-
-                            if (data.PathType == PlaceholderData.PlaceholderType.File)
-                            {
-                                filePlaceholders.Add(data);
-                            }
-                            else
-                            {
-                                folderPlaceholders.Add(data);
-                            }
+                            tempFilePlaceholders.Add(data);
                         }
-                    }
+                        else
+                        {
+                            tempFolderPlaceholders.Add(data);
+                        }
+                    });
                 }
+
+                filePlaceholders = tempFilePlaceholders;
+                folderPlaceholders = tempFolderPlaceholders;
             }
             catch (Exception ex)
             {
@@ -165,7 +156,7 @@ namespace GVFS.Common.Database
             this.Insert(new PlaceholderData() { Path = path, PathType = PlaceholderData.PlaceholderType.PossibleTombstoneFolder });
         }
 
-        public void RemoveStartingWith(string path)
+        public List<IPlaceholderData> RemoveStartingWith(string path)
         {
             // Normalize the path to match what will be in the database
             path = NormalizePath(path);
@@ -175,9 +166,19 @@ namespace GVFS.Common.Database
                 using (IDbConnection connection = this.connectionPool.GetConnection())
                 using (IDbCommand command = connection.CreateCommand())
                 {
-                    command.CommandText = "DELETE FROM Placeholder WHERE path LIKE @path;";
+                    List<IPlaceholderData> removedPlaceholders = new List<IPlaceholderData>();
+                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path LIKE @path";
                     command.AddParameter("@path", DbType.String, $"{path}%");
-                    command.ExecuteNonQuery();
+                    ReadPlaceholders(command, data => removedPlaceholders.Add(data));
+
+                    command.CommandText = "DELETE FROM Placeholder WHERE path LIKE @path;";
+
+                    lock (this.writerLock)
+                    {
+                        command.ExecuteNonQuery();
+                    }
+
+                    return removedPlaceholders;
                 }
             }
             catch (Exception ex)
@@ -239,6 +240,26 @@ namespace GVFS.Common.Database
             catch (Exception ex)
             {
                 throw new GVFSDatabaseException($"{nameof(PlaceholderTable)}.{nameof(this.GetCount)} Exception", ex);
+            }
+        }
+
+        private static void ReadPlaceholders(IDbCommand command, Action<PlaceholderData> dataHandler)
+        {
+            using (IDataReader reader = command.ExecuteReader())
+            {
+                while (reader.Read())
+                {
+                    PlaceholderData data = new PlaceholderData();
+                    data.Path = reader.GetString(0);
+                    data.PathType = (PlaceholderData.PlaceholderType)reader.GetByte(1);
+
+                    if (!reader.IsDBNull(2))
+                    {
+                        data.Sha = reader.GetString(2);
+                    }
+
+                    dataHandler(data);
+                }
             }
         }
 
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -151,7 +151,7 @@ namespace GVFS.Common.Database
             this.Insert(new PlaceholderData() { Path = path, PathType = PlaceholderData.PlaceholderType.PossibleTombstoneFolder });
         }
 
-        public List<IPlaceholderData> RemoveStartingWith(string path)
+        public List<IPlaceholderData> RemoveAllEntriesForFolder(string path)
         {
             // Normalize the path to match what will be in the database
             path = GVFSDatabase.NormalizePath(path);
@@ -162,11 +162,12 @@ namespace GVFS.Common.Database
                 using (IDbCommand command = connection.CreateCommand())
                 {
                     List<IPlaceholderData> removedPlaceholders = new List<IPlaceholderData>();
-                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path LIKE @path";
-                    command.AddParameter("@path", DbType.String, $"{path}%");
+                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirecotrySeparator;";
+                    command.AddParameter("@path", DbType.String, $"{path}");
+                    command.AddParameter("@pathWithDirecotrySeparator", DbType.String, $"{path + Path.DirectorySeparatorChar}%");
                     ReadPlaceholders(command, data => removedPlaceholders.Add(data));
 
-                    command.CommandText = "DELETE FROM Placeholder WHERE path LIKE @path;";
+                    command.CommandText = "DELETE FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirecotrySeparator;";
 
                     lock (this.writerLock)
                     {
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -159,6 +159,24 @@ namespace GVFS.Common.Database
             this.Insert(new PlaceholderData() { Path = path, PathType = PlaceholderData.PlaceholderType.PossibleTombstoneFolder });
         }
 
+        public void RemoveStartingWith(string path)
+        {
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    command.CommandText = "DELETE FROM Placeholder WHERE path LIKE @path;";
+                    command.AddParameter("@path", DbType.String, $"{path}%");
+                    command.ExecuteNonQuery();
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new GVFSDatabaseException($"{nameof(PlaceholderTable)}.{nameof(this.Remove)}({path}) Exception", ex);
+            }
+        }
+
         public void Remove(string path)
         {
             try
--- a/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
@@ -273,6 +273,11 @@ namespace GVFS.Common
             }
         }
 
+        public void RemoveStartingWith(string path)
+        {
+            throw new NotImplementedException();
+        }
+
         private IEnumerable<string> GenerateDataLines(IEnumerable<IPlaceholderData> updatedPlaceholders)
         {
             HashSet<string> keys = new HashSet<string>(GVFSPlatform.Instance.Constants.PathComparer);
--- a/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
@@ -273,7 +273,7 @@ namespace GVFS.Common
             }
         }
 
-        List<IPlaceholderData> IPlaceholderCollection.RemoveStartingWith(string path)
+        List<IPlaceholderData> IPlaceholderCollection.RemoveAllEntriesForFolder(string path)
         {
             throw new NotImplementedException();
         }
--- a/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
@@ -278,6 +278,16 @@ namespace GVFS.Common
             throw new NotImplementedException();
         }
 
+        List<IPlaceholderData> IPlaceholderCollection.RemoveStartingWith(string path)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void AddPlaceholderData(IPlaceholderData data)
+        {
+            throw new NotImplementedException();
+        }
+
         private IEnumerable<string> GenerateDataLines(IEnumerable<IPlaceholderData> updatedPlaceholders)
         {
             HashSet<string> keys = new HashSet<string>(GVFSPlatform.Instance.Constants.PathComparer);
--- a/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
+++ b/GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
@@ -273,11 +273,6 @@ namespace GVFS.Common
             }
         }
 
-        public void RemoveStartingWith(string path)
-        {
-            throw new NotImplementedException();
-        }
-
         List<IPlaceholderData> IPlaceholderCollection.RemoveStartingWith(string path)
         {
             throw new NotImplementedException();
--- a/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
+++ b/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
@@ -114,6 +114,20 @@ namespace GVFS.Common
             return true;
         }
 
+        public void RemoveAllEntriesForFolder(string path)
+        {
+            string entry = this.NormalizeEntryString(path, isFolder: true);
+            foreach (string modifiedPath in this.modifiedPaths)
+            {
+                if (modifiedPath.StartsWith(entry, StringComparison.OrdinalIgnoreCase))
+                {
+                    this.modifiedPaths.TryRemove(modifiedPath);
+                }
+            }
+
+            this.WriteAllEntriesAndFlush();
+        }
+
         public bool TryRemove(string path, bool isFolder, out bool isRetryable)
         {
             isRetryable = true;
--- a/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
+++ b/GVFS/GVFS.Common/ModifiedPathsDatabase.cs
@@ -83,10 +83,10 @@ namespace GVFS.Common
             return this.modifiedPaths.Contains(entry);
         }
 
-        public bool ContainsParentFolder(string path)
+        public bool ContainsParentFolder(string path, out string parentFolder)
         {
             string entry = this.NormalizeEntryString(path, isFolder: false);
-            return this.ContainsParentDirectory(entry);
+            return this.ContainsParentDirectory(entry, out parentFolder);
         }
 
         public IEnumerable<string> GetAllModifiedPaths()
@@ -235,8 +235,13 @@ namespace GVFS.Common
 
         private bool ContainsParentDirectory(string modifiedPath)
         {
+            return this.ContainsParentDirectory(modifiedPath, out _);
+        }
+
+        private bool ContainsParentDirectory(string modifiedPath, out string parentFolder)
+        {
             string[] pathParts = modifiedPath.Split(new char[] { GVFSConstants.GitPathSeparator }, StringSplitOptions.RemoveEmptyEntries);
-            string parentFolder = string.Empty;
+            parentFolder = string.Empty;
             for (int i = 0; i < pathParts.Length - 1; i++)
             {
                 parentFolder += pathParts[i] + GVFSConstants.GitPathSeparatorString;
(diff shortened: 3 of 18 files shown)


GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -294,6 +294,18 @@ public EventMetadata GetAndResetHeartBeatMetadata(out bool logToFile)
             return metadata;
         }
 
+        public bool DehydrateFolder(string relativePath)
+        {
+            // Remove all modified paths that start with the path
+            this.modifiedPaths.RemoveAllEntriesForFolder(relativePath);
+
+            // Remove all placeholders that start with the path
+            this.placeholderDatabase.RemoveStartingWith(relativePath);
```
jeschu1: nit: Maybe call tihs RemoveAllEntriesForFolder as well?

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -157,6 +157,24 @@ public void AddPossibleTombstoneFolder(string path)
...
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    command.CommandText = "DELETE FROM Placeholder WHERE path LIKE @path;";
+                    command.AddParameter("@path", DbType.String, $"{path}%");
+                    command.ExecuteNonQuery();
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new GVFSDatabaseException($"{nameof(PlaceholderTable)}.{nameof(this.Remove)}({path}) Exception", ex);
```
jeschu1: This seems like the standard pattern, but what state are things left in the exception gets thrown.  How would we suggest the user recover from that?

jrbriggs:
>Do we need to backup anything?
_My thought is we don't need to since we ran status and it came back clean no data should be lost. But do we need and could we have a rollback if something goes wrong?_

I agree -- rather than trying to build the potential for recovering lost work, let's rather prevent losing work entirely. If we have a clean status I think that would be sufficient. The only edge case I think we could hit is when we have missed a path being added to modified paths under the dehydrate root. Should we, prior to dehydrating, add the entire folder to modified paths so Git grovels everything for a last-minute sanity check?

>What do we do when there is an error for one of the folders?
_I suppose there are different actions depending on where this happens. For example if deleting the folder fails we don't need to send it to the mount process for dehydration but still need to let the user know. But if something happens in the dehydration in the mount process what should be done?_
This is a bit tricky. We might want to take only one folder to dehydrate at a time, which might be easier to script. If we take multiple folders and fail one in the middle it might be harder for a consumer of the feature to know how to resume. The tradeoff would be spinning up multiple processes to dehydrate if there were multiple folders to process. Do you have an idea how much extra time that would incur?

GVFS/GVFS.Common/Database/PlaceholderTable.cs
kewillford: Usually this would result in VFS for Git crashing and having an error entry in the logs.  If VFS for Git cannot run queries against the placeholder database then there is something really wrong and we don't want things to get worse.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: That is not what happens from the perspective of the `placeholderDatabase`. It is using a `LIKE 'path%`.  You could pass a partial folder name and it would remove all the entries that start with that partial folder name.  It doesn't know about folders only paths.

kewillford:
>We might want to take only one folder to dehydrate at a time, which might be easier to script. If we take multiple folders and fail one in the middle it might be harder for a consumer of the feature to know how to resume. The tradeoff would be spinning up multiple processes to dehydrate if there were multiple folders to process. Do you have an idea how much extra time that would incur?

The big issue here would be that the dehydrate unmounts the repo to do some of the work then mounts and does some more work which if there were multiple processes spinning up cause havoc on any other dehydrates that might be in the process of doing work.

I will think more on failure scenarios and how we can better handle any failures.

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -158,6 +155,33 @@ public void AddPossibleTombstoneFolder(string path)
...
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    List<IPlaceholderData> removedPlaceholders = new List<IPlaceholderData>();
+                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path LIKE @path";
+                    command.AddParameter("@path", DbType.String, $"{path}%");
+                    ReadPlaceholders(command, data => removedPlaceholders.Add(data));
+
+                    command.CommandText = "DELETE FROM Placeholder WHERE path LIKE @path;";
+                    command.ExecuteNonQuery();
```
kewillford: Need to lock

GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
```diff
@@ -273,6 +273,21 @@ public void WriteAllEntriesAndFlush(IEnumerable<IPlaceholderData> updatedPlaceho
             }
         }
 
+        public void RemoveStartingWith(string path)
```
kewillford: Remove this method.  I don't think it is used anymore.

wilbaker:
>I agree -- rather than trying to build the potential for recovering lost work, let's rather prevent losing work entirely. If we have a clean status I think that would be sufficient. The only edge case I think we could hit is when we have missed a path being added to modified paths under the dehydrate root. Should we, prior to dehydrating, add the entire folder to modified paths so Git grovels everything for a last-minute sanity check?

One other possible edge case is files that are git ignored (and shouldn't have been ignored).  However, I think we can address that with the message that `dehydrate` is much like `git clean` in that ignored files will be removed.

GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
wilbaker: Looks like this method is still being used here:

```
        public bool DehydrateFolder(string relativePath)
        {
            List<IPlaceholderData> removedPlaceholders;
            List<string> removedModifiedPaths;
            bool successful = false;

            try
            {
                removedPlaceholders = this.placeholderDatabase.RemoveStartingWith(relativePath);
```

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                    foreach (string folder in folders)
+                    {
+                        string fullPath = Path.Combine(enlistment.WorkingDirectoryRoot, folder);
+                        string backupPath = Path.Combine(backupRoot, folder);
+                        if (this.fileSystem.DirectoryExists(fullPath))
+                        {
+                            if (!this.TryIO(tracer, () => this.fileSystem.CopyDirectoryRecursive(fullPath, backupPath), $"Backing up {folder} to {backupPath}", out ioError) ||
+                                !this.TryIO(tracer, () => this.fileSystem.DeleteDirectory(fullPath), $"Deleting {fullPath}", out ioError))
+                            {
+                                this.WriteMessage(tracer, $"Backup failed for {folder} and will not be dehydrated. {ioError}");
```
wilbaker: It looks like `ioError` will not contain the description of what IO action we were attempting.  Should `TryIO` be updated to include the description in `ioError`?  Or alternatively, should the copy and delete have their own code blocks so that the message we write here can say specifically which step failed?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                    }
+
+                    return true;
+                },
+                "Backing up folders"))
+            {
+                this.ReportErrorAndExit(tracer, "Dehydrate for folders failed.");
+            }
+
+            this.Mount(tracer);
```
wilbaker: If this call to `Mount` were to fail, what would a user do to repair their repo and try the dehydrate again?  

I'm wondering if the dehydrate verb should send over the dehydrate message even when the folder does not exist on disk.  Would doing so harm anything?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                foreach (string folder in response.SuccessfulFolders)
                 {
-                    this.Output.WriteLine();
-                    this.WriteMessage(tracer, "ERROR: Backup failed. We will attempt to mount, but you may need to reclone if that fails");
+                    this.WriteMessage(tracer, $"{folder} folder successfully dehydrated.");
+                }
 
+                foreach (string folder in response.FailedFolders)
+                {
+                    this.WriteMessage(tracer, $"{folder} folder failed to dehydrate. You may need to clean the working directory and retry the dehydrate.");
```
wilbaker: > may need to clean the working directory

What does "clean the working directory" mean in this context?  

Are there more explicit instructions we can provide (e.g. unmount your repo and delete the failed folder, run `git clean` for the folder, etc.)?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                else
+                {
+                    string[] folders = this.Folders.Split(new[] { ";" }, StringSplitOptions.RemoveEmptyEntries);
+
+                    if (folders.Length > 0)
                     {
-                        // Converting the src folder to partial must be the final step before mount
-                        this.PrepareSrcFolder(tracer, enlistment);
-                        this.Mount(tracer);
+                        if (cleanStatus)
```
wilbaker: The output for `CheckGitStatus` says:

> "If this takes too long, you can abort and run dehydrate with --no-status to skip this safety check."

But it looks like that is not true when a user specifies specific folders.  Omit that line or update the text to indicate that it can only be used in "full dehydrate" mode (assuming we always want to require it)?  (Same comment for that option's `HelpText`).

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -34,6 +36,13 @@ public class DehydrateVerb : GVFSVerb.ForExistingEnlistment
             HelpText = "Skip 'git status' before dehydrating")]
         public bool NoStatus { get; set; }
 
+        [Option(
+            "folders",
+            Default = "",
+            Required = false,
+            HelpText = "The folders to dehydrate separated by ;")]
+        public string Folders { get; set; }
```
wilbaker: Could you add the value of this new `Option` to the `EventMetadata` that's written in the start event (see line 63)?

```
                    new EventMetadata
                    {
                        { "Confirmed", this.Confirmed },
                        { "NoStatus", this.NoStatus },
                        { "NamedPipeName", enlistment.NamedPipeName },
                        { nameof(this.EnlistmentRootPathParameter), this.EnlistmentRootPathParameter },
                    });
```

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -114,12 +123,22 @@ protected override void Execute(GVFSEnlistment enlistment)
                     return;
```
wilbaker: It looks like the above `"WARNING: THIS IS AN EXPERIMENTAL FEATURE` message needs to be updated (or skipped) when using the `Folders` option.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -196,6 +196,12 @@ public override FileSystemResult WritePlaceholderDirectory(string relativePath)
             return new FileSystemResult(HResultToFSResult(result), unchecked((int)result));
         }
 
+        public override FileSystemResult DehydrateFolder(string relativePath)
+        {
+            // Don't need to do anything here because the parent will reproject the folder.
```
wilbaker: Do we need to ensure the parent folder is partial?  I am thinking about a scenario like this:

Repo:
```
src
  |
  -> Foo
    |
    -> bar
      |
      -> baz.txt
```

1. rmdir /s/q Foo
2. git checkout -- Foo
3. gvfs dehydrate --folders Foo\bar

When dehydrate is run "bar" will not be reprojected because Foo is a full folder.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -212,6 +212,20 @@ public override FileSystemResult WritePlaceholderDirectory(string relativePath)
...
 
+        public override FileSystemResult DehydrateFolder(string relativePath)
+        {
+            FileSystemResult result = new FileSystemResult(FSResult.Ok, 0);
+            GitIndexProjection.PathSparseState sparseState = this.FileSystemCallbacks.GitIndexProjection.GetFolderPathSparseState(relativePath);
+
+            if (sparseState == GitIndexProjection.PathSparseState.Included)
+            {
+                result = this.WritePlaceholderDirectory(relativePath);
+                this.FileSystemCallbacks.OnPlaceholderFolderCreated(relativePath, string.Empty);
```
wilbaker: Could you add a comment here explaining why we need to do this when the folder is Included?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+
+        private void DehydrateFolders(JsonTracer tracer, GVFSEnlistment enlistment, string backupRoot, string[] folders)
+        {
+            List<string> foldersToDehydrate = new List<string>();
+
+            if (!this.ShowStatusWhileRunning(
+                () =>
+                {
+                    string ioError;
+                    foreach (string folder in folders)
```
wilbaker: Do we need any checks in here for nested folders?

What happens if a user passes in `Foo\bar;Foo`?  Would the copy of `Foo\bar` first cause the copy of its parent to fail?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -296,6 +296,48 @@ public EventMetadata GetAndResetHeartBeatMetadata(out bool logToFile)
...
+            {
+                removedPlaceholders = this.placeholderDatabase.RemoveStartingWith(relativePath);
+                removedModifiedPaths = this.modifiedPaths.RemoveAllEntriesForFolder(relativePath);
+                FileSystemResult result = this.fileSystemVirtualizer.DehydrateFolder(relativePath);
+                successful = result.Result == FSResult.Ok;
+            }
+            catch (Exception ex)
+            {
+                EventMetadata metadata = this.CreateEventMetadata(relativePath, ex);
+                this.context.Tracer.RelatedError(metadata, $"{nameof(this.DehydrateFolder)} threw and exception");
```
wilbaker: Typo: `threw and` -> `threw an`

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -296,6 +296,48 @@ public EventMetadata GetAndResetHeartBeatMetadata(out bool logToFile)
...
+                    {
+                        this.placeholderDatabase.AddPlaceholderData(data);
+                    }
+                }
+
+                if (removedModifiedPaths != null)
+                {
+                    foreach (string modifiedPath in removedModifiedPaths)
+                    {
+                        this.modifiedPaths.TryAdd(modifiedPath, isFolder: modifiedPath.EndsWith(GVFSConstants.GitPathSeparatorString), isRetryable: out bool isRetryable);
```
wilbaker: Do we need to check the return value of `TryAdd`?  Should it always succeed?

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -26,6 +27,11 @@ public static void CreateTable(IDbConnection connection)
             }
         }
 
+        public static string NormalizePath(string path)
```
wilbaker: I think this method is only used in `RemoveStartingWith`, should it be private (or is there somewhere else we should call it from)?

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -158,6 +155,37 @@ public void AddPossibleTombstoneFolder(string path)
...
+            path = NormalizePath(path);
+
+            try
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    List<IPlaceholderData> removedPlaceholders = new List<IPlaceholderData>();
+                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path LIKE @path";
+                    command.AddParameter("@path", DbType.String, $"{path}%");
```
wilbaker: As best I can tell the normalized path for folders will not have a trailing directory separator (the final call on the path is `Trim(Path.DirectorySeparatorChar)`).

I think this means that if `RemoveStartingWith` were called for the folder `Foo`, this query would also return placeholders like `FooBar\baz.txt`

If there's not one already, a functional test for this scenario would be good.

GVFS/GVFS.Common/LegacyPlaceholderListDatabase.cs
kewillford: That is using the interface version below that returns the list.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
kewillford: So this was a little messy because when you delete `Foo` that adds an entry to the modified paths and if you dehydrate any children that causes issues because the skip worktree bit will not be set for any children so nothing will be part of the projection even if we mark the parent folder as partial.  So what I ended up doing was preventing dehydration of folders that have a parent folder entry in the modified paths.

We could do more work here to maybe get this to work like add all the children that aren't being dehydrated to the the modified paths so we can remove the child specified for dehydration but to me that seems overkill and could introduce more risk and bugs.

Let me know what you think.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
kewillford: Wrote a test for this case and it works without issue because after copying, the `Foo\bar` folder is deleted and will not be copied again when copying `Foo`.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
wilbaker: > So what I ended up doing was preventing dehydration of folders that have a parent folder entry in the modified paths.

- Is this only a problem if the parent is in ModifiedPaths.dat?  Or is there a similar issue dehydrating folders that have children that are full folders?

- If a user *really* wanted Foo\bar dehydrated (in the above example) could they do so by dehydrating Foo?  

I think it makes sense to start with the restriction you mentioned and see what feedback we get from users. Ideally we'd have some manual work-around we could provide if someone manages to get stuck by this (even if it's as drastic as dehydrating the while repo) and so I'm interested in hearing your thoughts there.

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: I'm not sure there is anything we can do if it is false.  It was already logged in the `TryAdd`.  We could retry a few times when `isRetryable` is true.  I guess the other option is to add it to the background queue and let it handle adding?  That seems a little hacky.  Let me know your thoughts.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
kewillford: >Is this only a problem if the parent is in ModifiedPaths.dat? Or is there a similar issue dehydrating folders that have children that are full folders?

This is only a problem when there is a parent in the ModifiedPaths.dat.  If there are children that are full folders all the modified paths and placeholders for them are removed because it uses a starts with when removing entries.

>If a user really wanted Foo\bar dehydrated (in the above example) could they do so by dehydrating Foo?

Yes and I have a message that lets the user know that is what they would need to do to dehydrate.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                    {
+                        string fullPath = Path.Combine(enlistment.WorkingDirectoryRoot, folder);
+                        string backupPath = Path.Combine(backupRoot, folder);
+                        if (this.fileSystem.DirectoryExists(fullPath))
```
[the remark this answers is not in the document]
kewillford: What were your thoughts here?  Do we need to treat them differently?  They will be in the backup location but we don't do anything with that unless maybe manually moving things back into the repo?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,31 +164,139 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                        {
+                            if (!this.TryIO(tracer, () => this.fileSystem.CopyDirectoryRecursive(fullPath, backupPath), $"Backing up {folder} to {backupPath}", out ioError) ||
```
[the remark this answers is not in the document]
kewillford: We would then need platform specific handling here which I would like to avoid if at all possible.  Aren't the empty placeholder files just files with specific reparse data and copying should work like with any other file?  Is there an issue with moving the file with the reparse data?  Will it cause issues with explorer or windows?  I will do some testing and see if I run into anything.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
kewillford: >If this call to Mount were to fail, what would a user do to repair their repo and try the dehydrate again?

I guess that would depend on why the mount failed.  Did you have some thoughts on what would cause this?

>I'm wondering if the dehydrate verb should send over the dehydrate message even when the folder does not exist on disk. Would doing so harm anything?

Thinking this through and I don't think it would harm anything.  It would remove modified paths and placeholders for the folder and if it was included create the placeholder on Mac but that shouldn't be the case because there was a clean `git status` so it shouldn't be in the index as included but not on disk.  I will try it out and see.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -109,17 +119,46 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+
+To actually execute the dehydrate, run 'gvfs dehydrate --confirm --folders <folder list>'
+from a parent of the folders list.
 ");
 
                     return;
                 }
 
-                this.CheckGitStatus(tracer, enlistment);
+                bool fullDehydrate = string.IsNullOrEmpty(this.Folders);
```
kewillford: Move this above line 104 to use there as well.

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -158,6 +150,38 @@ public void AddPossibleTombstoneFolder(string path)
             this.Insert(new PlaceholderData() { Path = path, PathType = PlaceholderData.PlaceholderType.PossibleTombstoneFolder });
         }
 
+        public List<IPlaceholderData> RemoveAllEntriesForFolder(string path)
```
kewillford: Need a unit test.

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -158,6 +150,38 @@ public void AddPossibleTombstoneFolder(string path)
...
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    List<IPlaceholderData> removedPlaceholders = new List<IPlaceholderData>();
+                    command.CommandText = "SELECT path, pathType, sha FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirecotrySeparator;";
+                    command.AddParameter("@path", DbType.String, $"{path}");
+                    command.AddParameter("@pathWithDirecotrySeparator", DbType.String, $"{path + Path.DirectorySeparatorChar}%");
+                    ReadPlaceholders(command, data => removedPlaceholders.Add(data));
+
+                    command.CommandText = "DELETE FROM Placeholder WHERE path = @path OR path LIKE @pathWithDirecotrySeparator;";
```
kewillford: Move the `FROM ...` into a shared string for both commands.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
kewillford: I think I understand what you were trying to point out that if the mount fails but we have already deleted the folders that were passed in, if the user runs dehydrate again we would still want to send the dehydrate message to clean up those folders which makes sense.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -221,12 +405,26 @@ private void CheckGitStatus(ITracer tracer, GVFSEnlistment enlistment)
...
                         this.WriteMessage(tracer, statusResult.Output);
                         this.WriteMessage(tracer, "git status reported that you have dirty files");
-                        this.WriteMessage(tracer, "Either commit your changes or run dehydrate with --no-status");
+                        if (fullDehydrate)
+                        {
+                            this.WriteMessage(tracer, "Either commit your changes or run dehydrate with --no-status");
+                        }
+                        else
+                        {
+                            this.WriteMessage(tracer, "Either commit your changes or run dehydrate with --no-status");
```
kewillford: Need to update this message.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -145,36 +184,178 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                foreach (string folder in response.SuccessfulFolders)
                 {
-                    this.Output.WriteLine();
-                    this.WriteMessage(tracer, "ERROR: Backup failed. We will attempt to mount, but you may need to reclone if that fails");
+                    this.WriteMessage(tracer, $"{folder} folder successfully dehydrated.");
+                }
 
+                foreach (string folder in response.FailedFolders)
+                {
+                    this.WriteMessage(tracer, $"{folder} folder failed to dehydrate. You may need to reset the working directory by deleting {folder} running `git reset --hard` and retry the dehydrate.");
```
kewillford: Add some commas to this text.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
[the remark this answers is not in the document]
kewillford: Okay I did some testing and added functional tests but here are a couple of points.

>I think the scenario I was wondering about is a bit different. What would/should happen in a scenario like this:
```
// Delete the projected directory "Bar".  A tombstone will be created for "Bar"
C:\Repos\Foo\src>rmdir /s/q Bar

// Dehydrate Bar, what should happen?
C:\Repos\Foo\src>gvfs dehydrate Bar
```

In this case the dehydrate would fail because there would not be a clean git status but you could commit and then dehydrate which is what your second example does but what I have found is that folder tombstones appear as junctions when unmounted.  This is the case whether or not there is another mount running.  So it doesn't get `{folder} did not exist to dehydrate.` but is able to delete the junction.  So I think we are good.

GVFS/GVFS.Common/ModifiedPathsDatabase.cs
```diff
@@ -209,9 +234,14 @@ private bool TryParseRemoveLine(string line, out string key, out string error)
         }
 
         private bool ContainsParentDirectory(string modifiedPath)
+        {
+            return this.ContainsParentDirectory(modifiedPath, out _);
+        }
+
+        private bool ContainsParentDirectory(string modifiedPath, out string parentFolder)
```
jamill: `ContainsParentDirectory` and `ContainsParentFolder` have quite similar signatures / names

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -143,14 +163,363 @@ public void DehydrateShouldFailOnWrongDiskLayoutVersion()
...
+            {
+                directoryToCreate.ShouldNotExistOnDisk(this.fileSystem);
+            }
+            else
+            {
+                directoryToCreate.ShouldBeADirectory(this.fileSystem);
+            }
+        }
+
+        private void PlaceholderShouldContain(params string[] paths)
```
jamill: nit: should plurality match between `PlaceholderShouldContain` and `PlaceholdersShouldNotContain`?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -143,14 +163,363 @@ public void DehydrateShouldFailOnWrongDiskLayoutVersion()
...
+        {
+            string[] placeholderLines = this.GetPlaceholderDatabaseLines();
+            foreach (string path in paths)
+            {
+                placeholderLines.ShouldContain(x => x.StartsWith(path + GVFSHelpers.PlaceholderFieldDelimiter));
```
jamill: Does capitalization of the paths being matched matter here (possibly dependent on the platform?)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -143,14 +163,363 @@ public void DehydrateShouldFailOnWrongDiskLayoutVersion()
...
+            }
+        }
+
+        private void PlaceholdersShouldNotContain(params string[] paths)
+        {
+            string[] placeholderLines = this.GetPlaceholderDatabaseLines();
+            foreach (string path in paths)
+            {
+                placeholderLines.ShouldNotContain(x => x.StartsWith(path + Path.DirectorySeparatorChar) || x.Equals(path, StringComparison.OrdinalIgnoreCase));
```
jamill: ah - here we are explicit with comparison for the `Equals` method - should be explicit for the `StartsWith` as well?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -31,9 +34,16 @@ public class DehydrateVerb : GVFSVerb.ForExistingEnlistment
             "no-status",
             Default = false,
             Required = false,
-            HelpText = "Skip 'git status' before dehydrating")]
+            HelpText = "Skip 'git status' before dehydrating only valid when not using folders option")]
```
jamill: Is this missing punctuation? `...before dehydrating. Only valid...`

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -296,6 +296,57 @@ public EventMetadata GetAndResetHeartBeatMetadata(out bool logToFile)
             return metadata;
         }
 
+        public bool DehydrateFolder(string relativePath)
```
jamill: This is close to a `Try...` patterned method. If we use a `Try` prefix, it would at least indicate to callers they need to check the result.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -198,6 +567,11 @@ private ProcessResult RunDehydrateProcess(bool confirm, bool noStatus)
                 dehydrateFlags += " --no-status ";
             }
 
+            if (foldersToDehydrate.Length > 0)
+            {
+                dehydrateFlags += $" --folders {string.Join(";", foldersToDehydrate)}";
```
jamill: Should path separator be a const somewhere (that both client / server ends of pipe can read)?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: The modified paths has a `TryAdd` so it shouldn't throw but the placeholder add could so added a try/catch around it and renamed the method to `TryDehydrateFolder`

GVFS/GVFS/CommandLine/DehydrateVerb.cs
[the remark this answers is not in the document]
cgallred: > Aren't the empty placeholder files just files with specific reparse data and copying should work like with any other file? Is there an issue with moving the file with the reparse data? Will it cause issues with explorer or windows?
> 
> I can't remember how empty placeholders will behave in this scenario (@cgallred might).
> 
> I will do some testing and see if I run into anything.
> 
> When you give this a test please be sure to include a scenario were there is a second mounted repo on the same volume while the dehydrate is being run.  My concern was that ProjFS might return an error like ERROR_FILE_SYSTEM_VIRTUALIZATION_UNAVAILABLE if you attempt to read the empty placeholder while the provider is not running.
> 
> We would then need platform specific handling here which I would like to avoid if at all possible.
> 
> That would definitely be a downside to the approach I suggested, however, if a large cone of the repo is being dehydrated the move might be significantly faster than a copy.  If the performance of the copy is not a problem (and there are no issues with copying empty placeholders) then I think we could stick with copy and keep the move option in mind if we find ourselves having to speed up this path.

Sorry for the delay, I was on vacation.  If the provider is running then copying will result in a content recall.  The file system doesn't have a "copy" operation.  Copy is an operation synthesized at the Win32 API level.  In the file system we see a copy as open-read-close for the copy source and open(possibly w/ create new)-write-close for the copy destination.  The file system does have a move/rename operation.  That should not result in a content recall.

PR #1328, opened 2019-07-12
Move Enlistment Directory For Functional tests to ~/GVFS.FT

jeschu1:
"/" is readonly on Catalina so this change is needed to create Enlistment directories

Unfortunately there are several line ending fixes in the commit as well.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Settings.cs
+++ b/GVFS/GVFS.FunctionalTests/Settings.cs
@@ -46,7 +46,9 @@ namespace GVFS.FunctionalTests.Properties
                 }
                 else
                 {
-                    string root = "/GVFS.FT";
+                    string root = Path.Combine(
+                        Environment.GetEnvironmentVariable("HOME"),
+                        "GVFS.FT");
                     EnlistmentRoot = Path.Combine(root, "test");
                     ControlGitRepoRoot = Path.Combine(root, "control");
                     FastFetchBaseRoot = Path.Combine(root, "FastFetch");
--- a/Scripts/Mac/RunFunctionalTests.sh
+++ b/Scripts/Mac/RunFunctionalTests.sh
@@ -7,9 +7,7 @@ if [ -z $CONFIGURATION ]; then
   CONFIGURATION=Debug
 fi
 
-sudo mkdir /GVFS.FT
-sudo chown $USER /GVFS.FT
-
+mkdir ~/GVFS.FT
 if [ "$2" != "--test-gvfs-on-path" ]; then
   echo "Calling LoadPrjFSKext.sh as --test-gvfs-on-path not set..."
   $VFS_SRCDIR/ProjFS.Mac/Scripts/LoadPrjFSKext.sh $CONFIGURATION


GVFS/GVFS.FunctionalTests/Settings.cs
```diff
@@ -30,23 +30,23 @@ public static void Initialize()
...
-                }
-                else
+                    FastFetchControl = Path.Combine(FastFetchBaseRoot, "control");
+                    PathToGVFSService = @"GVFS.Service.exe";
+                    BinaryFileNameExtension = ".exe";
+                }
+                else
                 {
-                    string root = "/GVFS.FT";
+                    string root = "~/GVFS.FT";
```
wilbaker: What does `Path.Combine` return when the path includes `~`?

I ask because when we updated the default `.gvfsCache` location to be in `~` we had to use `Environment.GetEnvironmentVariable("HOME")` because the expansion of `~` was bash specific.

https://github.com/microsoft/VFSForGit/blob/fa70093b0795f6dac167e8e361f2b3249e06d1de/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs#L219

Scripts/Mac/RunFunctionalTests.sh
```diff
@@ -7,9 +7,8 @@ if [ -z $CONFIGURATION ]; then
   CONFIGURATION=Debug
 fi
 
-sudo mkdir /GVFS.FT
-sudo chown $USER /GVFS.FT
-
+sudo mkdir ~/GVFS.FT
```
wilbaker: Does this still need `sudo`?

Scripts/Mac/RunFunctionalTests.sh
```diff
@@ -7,9 +7,8 @@ if [ -z $CONFIGURATION ]; then
...
+sudo chown $USER ~/GVFS.FT
```
wilbaker: Is this line still needed (now that we're creating the directory in HOME)?

GVFS/GVFS.FunctionalTests/Settings.cs
```diff
@@ -30,23 +30,23 @@ public static void Initialize()
...
-                    FastFetchControl = Path.Combine(FastFetchBaseRoot, "control");
-                    PathToGVFSService = @"GVFS.Service.exe";
-                    BinaryFileNameExtension = ".exe";
-                }
-                else
+                    FastFetchControl = Path.Combine(FastFetchBaseRoot, "control");
+                    PathToGVFSService = @"GVFS.Service.exe";
+                    BinaryFileNameExtension = ".exe";
+                }
+                else
```
pmj: There are a lot of whitespace (end-of-line?) changes here. Was the LF/CRLF situation here in a muddle before, and it's now homogeneous? Unfortunately, the GitHub diff doesn't show the nature of the newline characters.

Scripts/Mac/RunFunctionalTests.sh
pmj: Incidentally, the `-p` argument to `mkdir` might be appropriate here to avoid spamming an error when the directory already exists.

GVFS/GVFS.FunctionalTests/Settings.cs
jeschu1: @pmj +1 to wishing we could see line endings on github!  The file was saved with CRLF which should be fixing any issues.  I think at some point we should consider updating all line endings in the repo and failing builds based on them changing.  Of course fixing all the line ends also pollutes history so not an easy decision.

Scripts/Mac/RunFunctionalTests.sh
jeschu1: Course not :-)

Scripts/Mac/RunFunctionalTests.sh
jeschu1: @pmj I was also thinking "-p" might reduce noise.  This appears to be existing.  @wilbaker  any reason why this didn't have the "-p" arg previously?

wilbaker: > any reason why this didn't have the "-p" arg previously?

I suspect it was that we either didn't know about the option or didn't mind the noise.  

`-p` will also create intermediate directories if they're missing and so the absence of `-p` could be a sanity check that the intermediate directories that we expect to be present are in fact present.

jeschu1: `-p` will also create intermediate directories if they're missing and so the absence of -p could be a sanity check that the intermediate directories that we expect to be present are in fact present.

I could see that.  Since this is pre-existing and outside the scope of my changes I'll leave it be.

PR #1330, opened 2019-07-16
Add sparse verb for allowing projection of only specific folders

kewillford:
This is a feature to only include specified folders in the projection.  I'm not stuck with the naming that I have chosen so if you have better ones please let me know.  

You should be able to clone as normal and it will project everything until something is added to the sparse folder database via the `gvfs sparse` verb.

The `gvfs sparse` verb includes the following:
```
-a <folderlist> // add ; separated folders to the list to include
-r <folderlist> // remove ; separated folders to the list
-l // list the folders that are currently in the list
```
Examples:
```
gvfs sparse -l
gvfs sparse -a GVFS/GVFS;Scripts
gvfs sparse -r Scripts
```

You can use both the `-a` and `-r` in the same invocation.  The removes will be applied first and then the adds.

You must have a clean `git status` in order to add or remove folders so the the code can reproject without issue.

These changes currently don't do anything with the modified paths so if you create a directory or get a conflict that is outside the included folders then the paths will be added to the modified path list and continue to be in the working directory for git to keep up to date.


GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
```diff
@@ -166,7 +166,7 @@ private static void AddFiles(SortedFolderEntries entries, params string[] names)
             for (int i = 0; i < names.Length; i++)
             {
                 LazyUTF8String entryString = ConstructLazyUTF8String(names[i]);
-                entries.AddFile(entryString, new byte[20]);
+                entries.AddFile(entryString, new byte[20], isIncluded: true);
```
wilbaker: Are there any unit tests that should be added where `isIncluded` is false?  (e.g. validating that `isIncluded` is properly set by `AddFile`)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -14,6 +14,7 @@
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
```
wilbaker: Commenting here to record some of our offline conversation.  

It looks like with these changes users might see files start showing up outside of their included folders if they use `git reset`:

- `git reset --soft`: Will report files outside of the included folders
- `git reset --mixed`: This could result in files outside of the include folders being added to ModifiedPaths.dat

We probably don't need to do anything for this now, but it might be worth mentioned in the release notes and/or having a follow up issue to see if there's anything we can do to improve this experience.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.FolderEntryData.cs
```diff
@@ -9,6 +9,12 @@ internal abstract class FolderEntryData
         {
             public LazyUTF8String Name { get; protected set; }
             public abstract bool IsFolder { get; }
+            public bool IsIncluded { get; set; } = true;
```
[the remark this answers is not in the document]
kewillford: It has to be in this base class and not just in the `FileData` because the code also has to know if a folder is included or not.  Measuring the memory and for ~3 million entries in the index it increased the memory usage ~20 MB from ~710 MB to ~730 MB.  So doesn't seem like a big enough amount to change it.  What do you think?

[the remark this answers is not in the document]
wilbaker: > It has to be in this base class and not just in the FileData because the code also has to know if a folder is included or not

Sorry it looks like I left a half thought.  If we did decide to try to steal a bit for files we'd need to make `IsIncluded` abstract and add the bool to the folder data.

> Measuring the memory and for ~3 million entries in the index it increased the memory usage ~20 MB from ~710 MB to ~730 MB. So doesn't seem like a big enough amount to change it. What do you think?

I agree that doesn't sound bad.  I think we can stick with the current code, I'll mark this comment as resolved.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
```diff
@@ -67,6 +67,21 @@ public bool BuildingProjection_HasSameParentAsLastEntry
...
+            {
+                get; set;
+            }
+
+            public bool BuildingProjection_ShouldIncludeRecursive
+            {
+                get; set;
+            }
+
+            public IncludedFolderData BuildingProjection_LastEntryIncludedFolder
```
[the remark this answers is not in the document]
kewillford: Renamed this to `BuildingProjection_IncludedFolderToCheck` as well as added a comment.  Hopefully that name makes it more understandable.

GVFS/GVFS.Common/Database/GVFSDatabase.cs
```diff
@@ -118,6 +118,7 @@ private void Initialize()
                 }
 
                 PlaceholderTable.CreateTable(connection);
+                IncludedFolderTable.CreateTable(connection);
```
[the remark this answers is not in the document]
kewillford: I haven't seen any in the performance tests that I have ran.

GVFS/GVFS/CommandLine/IncludeVerb.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                            return;
+                        }
+
+                        // Make sure there is a clean git status before allowing inclusions to change
+                        this.CheckGitStatus(tracer, enlistment);
+
+                        if (haveFoldersToRemove)
+                        {
+                            foreach (string directoryPath in this.Remove.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries))
```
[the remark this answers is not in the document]
kewillford: I will add to the help text about being relative to the repo root and we can get feedback to see if we need to do more than that.

BTW `GVFS` and `GVFS\` should be considered the same because the code that creates the `IncludedFolderData` does this:
```
string[] folders = directoryPath.Split(new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar }, StringSplitOptions.RemoveEmptyEntries);
```

GVFS/GVFS/CommandLine/IncludeVerb.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        protected override void Execute(GVFSEnlistment enlistment)
+        {
+            using (JsonTracer tracer = new JsonTracer(GVFSConstants.GVFSEtwProviderName, "Include"))
+            {
+                try
+                {
+                    tracer.AddLogFileEventListener(
+                        GVFSEnlistment.GetNewGVFSLogFileName(enlistment.GVFSLogsRoot, GVFSConstants.LogFileTypes.Include),
+                        EventLevel.Informational,
```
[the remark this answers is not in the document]
kewillford: I'm questioning the value of even having the logs for this verb specifically or if we should just output to the console any error that happen?  Otherwise there could be a lot of include logs that will have one line or that status failed.  I guess there might be other errors that would be good to know about but I think for this verb it would be nice if we had just one file instead of a file for every invocation of the verb. 

Most of my logs only have this.
```
[2019-07-16 16:41:26 -06:00] Include (Stop) {"DurationMs":14044}
```

What are your thoughts?

GVFS/GVFS/CommandLine/IncludeVerb.cs
```diff
@@ -0,0 +1,182 @@
...
+            Default = "",
+            HelpText = "A semicolon-delimited list of folders to remove for being included. Wildcards are not supported.")]
+        public string Remove { get; set; }
+
+        [Option(
+            'l',
+            "list",
+            Required = false,
+            Default = false,
+            HelpText = "List of folders to for being included in projection.")]
```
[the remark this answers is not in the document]
kewillford: Yeah that was almost language from another planet :)

GVFS/GVFS.Common/Database/GVFSDatabase.cs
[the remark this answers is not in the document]
wilbaker: Ok thanks! I'll mark this as resolved

GVFS/GVFS/CommandLine/IncludeVerb.cs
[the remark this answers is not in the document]
wilbaker: > I'm questioning the value of even having the logs for this verb specifically or if we should just output to the console any error that happen?

> Otherwise there could be a lot of include logs that will have one line or that status failed.

> Most of my logs only have this.

Maybe we should only create a log file when users specify folders to add or remove?  I would guess we won't get much value from having logs when `gvfs include` was only used to list the current contents.  However, I could see it being useful to see the history of how folders were added/removed when investigating support issues.

If we're not already I think it would be helpful to log the arguments to the verb so we know what is being added or removed.

What do you think?

GVFS/GVFS/CommandLine/IncludeVerb.cs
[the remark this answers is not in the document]
wilbaker: > BTW GVFS and GVFS\ should be considered the same because the code that creates the IncludedFolderData does this:

Ok thanks I missed this!

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -524,7 +524,12 @@ private void OnNewFileCreated(string relativePath, bool isDirectory)
                         }
                         else
                         {
-                            this.FileSystemCallbacks.OnFolderCreated(relativePath);
+                            if (this.FileSystemCallbacks.OnFolderCreated(relativePath))
+                            {
+                                // Need to enumerate the directory to get and create placeholders
```
wilbaker: Could you update this comment to include what `OnFileCreated` returning true means? (I saw you mentioned that in the Windows callback code).

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -622,10 +627,18 @@ private Result CreatePlaceholders(string directoryRelativePath, IEnumerable<Proj
...
                     metadata.Add("fileInfo.Size", fileInfo.Size);
                     metadata.Add("fileInfo.IsFolder", fileInfo.IsFolder);
+                    metadata.Add(nameof(result), result.ToString());
                     metadata.Add(nameof(sha), sha);
                     this.Context.Tracer.RelatedError(metadata, $"{nameof(this.CreatePlaceholders)}: Write placeholder failed");
 
-                    return result;
+                    if (result == Result.EIOError)
+                    {
+                        this.FileSystemCallbacks.OnFileConvertedToFull(childRelativePath);
```
wilbaker: Minor: Could you add a comment here explaining why we're calling `OnFileConvertedToFull` and what "include mode" scenario we expect to hit this in?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+        [Category(Categories.MacOnly)]
+        public void CreateFolderThenFileShouldAddToIncludedListAndStartProjecting()
+        {
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
+
+            string newFolderPath = Path.Combine(this.Enlistment.RepoRoot, "GVFS", "GVFS.Common");
+            newFolderPath.ShouldNotExistOnDisk(this.fileSystem);
+            Directory.CreateDirectory(newFolderPath);
+            string newFilePath = Path.Combine(newFolderPath, "test.txt");
```
wilbaker: Could you also add a test that does this on Mac but adds a file that is already in the projection?

GVFS/GVFS.UnitTests/Mock/Virtualization/Projection/MockGitIndexProjection.cs
```diff
@@ -67,6 +67,11 @@ public override bool IsProjectionParseComplete()
             return this.ProjectionParseComplete;
         }
 
+        public override PathProjectionState GetPathProjectionState(string virtualPath)
+        {
+            return PathProjectionState.Included;
```
wilbaker: Similar comment to the comment in `SortedFolderEntriesTests.cs`.  Are there any unit tests that should be run with `GetPathProjectionState` returning something other than `Included`?

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -408,10 +410,28 @@ public void OnFilePreDelete(string relativePath)
...
+        /// <returns>
+        /// true when the folder is successfully added to the include list because it is in the projection but currently excluded.
+        /// false when the folder was not excluded or there was a failure adding to the included list.
+        /// </returns>
+        public bool OnFolderCreated(string relativePath)
         {
+            GitIndexProjection.PathProjectionState pathProjectionState = this.GitIndexProjection.GetPathProjectionState(relativePath);
+            if (pathProjectionState == GitIndexProjection.PathProjectionState.Excluded)
+            {
+                if (this.GitIndexProjection.TryAddIncludedFolder(relativePath))
```
wilbaker: Would you expect `TryAddIncludedFolder` to fail under normal circumstances?

I'm wondering if we should log a warning/error either here or in `TryAddIncludedFolder` if the add fails.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -414,6 +429,52 @@ public virtual void GetFileTypeAndMode(string filePath, out FileType fileType, o
...
+                this.GetChildNameAndParentKey(virtualPath, out string fileName, out string parentKey);
+                FolderEntryData data = this.GetProjectedFolderEntryData(
+                    blobSizesConnection: null,
+                    childName: fileName,
+                    parentKey: parentKey);
+
+                if (data != null && !data.IsIncluded)
+                {
+                    data.Include();
+                    this.includedFolderCollection.Add(virtualPath);
```
wilbaker: Will this throw if the entry already exists?

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1046,6 +1033,33 @@ private HResult QueryFileNameHandler(string virtualPath)
             }
         }
 
+        private void MarkDirectoryAsPlaceholder(
+            string virtualPath,
+            bool isDirectory,
```
wilbaker: You can remove `isDirectory`,  it's not being used

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1005,29 +1005,16 @@ private HResult QueryFileNameHandler(string virtualPath)
...
-                            else
-                            {
-                                EventMetadata metadata = this.CreateEventMetadata(virtualPath);
-                                metadata.Add("isDirectory", isDirectory);
-                                metadata.Add("triggeringProcessId", triggeringProcessId);
-                                metadata.Add("triggeringProcessImageFileName", triggeringProcessImageFileName);
-                                metadata.Add("HResult", hr.ToString());
-                                this.Context.Tracer.RelatedError(metadata, nameof(this.NotifyNewFileCreatedHandler) + "_" + nameof(this.virtualizationInstance.MarkDirectoryAsPlaceholder) + " error");
-                            }
+                            this.MarkDirectoryAsPlaceholder(virtualPath, isDirectory, triggeringProcessId, triggeringProcessImageFileName);
```
wilbaker: What is the expected behavior if git creates a directory that's currently excluded?

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1046,6 +1033,33 @@ private HResult QueryFileNameHandler(string virtualPath)
...
+            HResult hr = this.virtualizationInstance.MarkDirectoryAsPlaceholder(
+                directoryPath,
+                FolderContentId,
+                PlaceholderVersionId);
+
+            if (hr == HResult.Ok)
+            {
+                this.FileSystemCallbacks.OnPlaceholderFolderCreated(virtualPath, triggeringProcessImageFileName);
+            }
+            else
```
wilbaker: If this fails in the "OnFolderCreated returns true" scenario, should we add the folder to ModifiedPaths.dat?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+        public void CreatingFolderShouldAddToIncludedListAndStartProjecting()
+        {
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
+
+            string newFolderPath = Path.Combine(this.Enlistment.RepoRoot, "GVFS", "GVFS.Common");
+            newFolderPath.ShouldNotExistOnDisk(this.fileSystem);
+            Directory.CreateDirectory(newFolderPath);
+            newFolderPath.ShouldBeADirectory(this.fileSystem);
+            string[] fileSystemEntries = Directory.GetFileSystemEntries(newFolderPath);
```
jamill: Do we have options on how to handle the case where a user tries to create a directory that is outside the scope of included folders? Could we, for instance, block creation? I am not arguing that we should do that, but wondering what is possible. I think it is at least _unexpected_ if I create a directory, and then find that it is pre-populated.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -101,18 +106,21 @@ public partial class GitIndexProjection : IDisposable, IProfilerOnlyIndexProject
...
+            this.ClearProjectionCaches();
         }
 
-        // For Unit Testing
         protected GitIndexProjection()
         {
+            this.projectionReadWriteLock = new ReaderWriterLockSlim();
+            this.projectionParseComplete = new ManualResetEventSlim(initialState: false);
+            this.wakeUpIndexParsingThread = new AutoResetEvent(initialState: false);
+            this.rootFolderData.ResetData(new LazyUTF8String("<root>"));
```
wilbaker: Why did this line need to be added as part of these changes?

`this.rootFolderData.ResetData(new LazyUTF8String("<root>"));`

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
```diff
@@ -42,6 +49,27 @@ public bool TryMount(out string output)
...
+        {
+            return this.CallGVFS($"include {this.enlistmentRoot} -a {string.Join(";", folders)}");
+        }
+
+        public string RemoveIncludedFolders(params string[] folders)
+        {
+            return this.CallGVFS($" include {this.enlistmentRoot} -r {string.Join(";", folders)}");
+        }
+
+        public string[] IncludedFoldersList()
```
jamill: From the above method names, I kindof expected `ListIncludedFolders` :)

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
jamill: I think it is hard to understand from just reading the callsite what the return value indicates (none of the other callback methods have a return value).  As an alternative, maybe have an outparam that returns a bool, that way you can document how the value is to be interpreted in the method signature.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
kewillford: I don't think we want to block the user from doing something that they may need to do.  Currently it will start projecting what should be in the folder and I think that is what would be most helpful to the user because if we
1. Block creation then the user is blocked if that is what they really need to do and we have no way to inform the user that they need to run a `gvfs include -a <folder>` in order to have that folder.
2. Allow them to create folders without showing the projection.  The user could create files that already exist in the index, in which case it would show as modified with only the content from the newly created file and they would have to run a `git checkout -- <file>` to get the content from the commit.

Starting the projection when the folder is created will let the user know the folder is currently a part of the repo and they will be able to see all the files that already exist in that folder, which to me would be more helpful in knowing what happened as soon as possible instead of later when I run a `git status` or something and wonder why.

GVFS/GVFS/CommandLine/IncludeVerb.cs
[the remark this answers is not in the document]
kewillford: Sounds good thanks!

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
kewillford: I like that idea @jamill Thanks!

GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
kewillford: >Would you expect TryAddIncludedFolder to fail under normal circumstances?

No

>if we should log a warning/error either here or in TryAddIncludedFolder if the add fails.

Will do.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: >Will this throw if the entry already exists?

No because the SQLite command is 
```
INSERT OR REPLACE INTO IncludedFolder (path) VALUES (@path);
```
Which will not do anything if it already exists since we are only using the primary key.

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
kewillford: Changed to `GetIncludedFolders` since that seemed to fit better with current coding conventions.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
jamill: Thanks for the detailed response! I was mainly curious about what the possible options were. I think all options have the potential to cause confusion.

Is there any precedent from similar source control systems (if there is anything comparable...) that would point to what users would be familiar with?

I assume that creating new files outside the scope of "included" files will be a relatively (or at least not the common) scenario, and most likely indicates a deficit in the list of included directories? If I need to access files outside of this scope, I would expect to add the directory to the included list first.

If I run `mkdir`, it also might not be clear that there are new files because I happened to create a directory that collides with an non-included directory. There is no way to explain tothe user why the new entries are now present.

If the command was run as part of a script, it might hide the fact that the commands were running in an environment with insufficient include list entries.

Anyway - just my thoughts on what I *might* expect - I agree that blocking with an error is not a good experience either. Maybe just something for us to keep an eye out for as people use it.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: I thought that I made this change for unit tests that were getting a `NullReferenceException` but maybe I fixed that with some other change because it is no longer happening.

mjcheetham:
> I'm not stuck with the naming that I have chosen so if you have better ones please let me know.

It seems weird to have a verb called `include` and it also take options for removing paths from the projection. Why not name it `projection`?

```shell
gvfs projection --add src/a;src/b;src/c --remove test/
```

..or if the CLI lib we're using supports it..

```shell
gvfs projection --add src/a --add src/b --add src/c --remove test/
```

For listing it would be:

```shell
gvfs projection --list
```

..and we can extend this verb in the future for other projection-specific operations.

```shell
gvfs projection --refresh
gvfs projection --info
```

etc.

kewillford:
@mjcheetham thanks for the suggestion!  The one issue I would have with `projection` is the
```
gvfs projection --list
```
Should that list the folders that get stored during add?  All the items for the index that would be projected?   It might also be a overloaded term in the code base since we already have the `GitIndexProjection` and now we would have the `ProjectionFolders` which could become confusing with each other.

Other terms I think might work are:
```
limit
narrow // kind of fits with a narrow clone
constraint
bound
only-hydrate // this was John's original for the feature
folders // might be a little to generic
```

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,6 +892,34 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
                 }
 
                 parentFolder = parentFolder.ChildEntries.GetOrAddFolder(indexEntry.BuildingProjection_PathParts[pathIndex]);
+
+                if (this.rootIncludedFolder.Children.Count > 0)
+                {
+                    if (indexEntry.BuildingProjection_LastEntryIncludedFolder == null)
+                    {
+                        indexEntry.BuildingProjection_LastEntryIncludedFolder = this.rootIncludedFolder;
```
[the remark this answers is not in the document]
wilbaker: Discussed offline:  If the parent folder changes then `ClearLastParent` would have been called.

GVFS/GVFS.UnitTests/Platform.Mac/MacFileSystemVirtualizerTests.cs
```diff
@@ -90,6 +90,7 @@ public void UpdatePlaceholderIfNeeded()
             const string UpdatePlaceholderFileName = "testUpdatePlaceholder.txt";
```
wilbaker: Are there any unit tests that can be added for the changes to `MacFileSystemVirtualizer`?

GVFS/GVFS.UnitTests/Common/Database/GVFSDatabaseTests.cs
```diff
@@ -87,11 +87,25 @@ private void TestGVFSDatabase(Action<GVFSDatabase> testCode, bool throwException
...
+
+            mockCommand3.Setup(x => x.Dispose());
+
             List<Mock<IDbConnection>> mockConnections = new List<Mock<IDbConnection>>();
             Mock<IDbConnection> mockConnection = new Mock<IDbConnection>(MockBehavior.Strict);
             mockConnection.SetupSequence(x => x.CreateCommand())
                           .Returns(mockCommand.Object)
-                          .Returns(mockCommand2.Object);
+                          .Returns(mockCommand2.Object)
+                          .Returns(mockCommand3.Object);
```
wilbaker: Should there be any validation for `mockCommand3` like we have for the other mock commands below?

```
            mockCommand.Verify(x => x.Dispose(), Times.Once);
            mockCommand2.Verify(x => x.Dispose(), Times.Once);
            mockConnections.ForEach(connection => connection.Verify(x => x.Dispose(), Times.Once));	            
            
            mockCommand.VerifyAll();
            mockCommand2.VerifyAll();
```

(Full disclosure, I am not exactly sure how these `mockCommand`s work in this test)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+            // Remove all included folders should make all folders appear again
+            string[] directories = Directory.GetDirectories(this.Enlistment.RepoRoot);
+            directories.ShouldMatchInOrder(this.allRootDirectories);
+            this.ValidateIncludedFolders(new string[0]);
+        }
+
+        [TestCase, Order(1)]
+        public void BasicTestsAddingAndRemoving()
```
wilbaker: Update the name of this test?  I don't see anything being removed.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            string childPath = Path.Combine(this.mainIncludedFolder, "CommandLine");
+            this.gvfsProcess.AddIncludedFolders(childPath);
+            string[] directories = Directory.GetDirectories(Path.Combine(this.Enlistment.RepoRoot, this.mainIncludedFolder));
+            directories.Length.ShouldEqual(1);
+            directories[0].ShouldEqual(Path.Combine(this.Enlistment.RepoRoot, childPath));
+            this.ValidateIncludedFolders(childPath);
+
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            directories = Directory.GetDirectories(Path.Combine(this.Enlistment.RepoRoot, this.mainIncludedFolder));
+            directories.ShouldMatchInOrder(this.directoriesInMainFolder);
```
wilbaker: Also validate `directories.Length` is now > 1?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
+
+            string folderPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts");
+            folderPath.ShouldBeADirectory(this.fileSystem);
+            string[] fileSystemEntries = Directory.GetFileSystemEntries(folderPath);
+            fileSystemEntries.Length.ShouldEqual(2);
+            fileToCreate.ShouldBeAFile(this.fileSystem);
+        }
+
+        private void ValidateIncludedFolders(params string[] folders)
```
wilbaker: Naming suggestion: `ValidateIncludedFolders` -> `ValidateFoldersInIncludeList` or `FoldersShouldBeInIncludeList`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+
+            // Add subfolder of main folder and make sure it stays recursive
+            string subFolder = Path.Combine(this.mainIncludedFolder, "Properties");
+            this.gvfsProcess.AddIncludedFolders(subFolder);
+            string folder = this.Enlistment.GetVirtualPathTo(subFolder);
+            folder.ShouldBeADirectory(this.fileSystem);
+            this.ValidateIncludedFolders(this.mainIncludedFolder, subFolder);
+
+            folder = this.Enlistment.GetVirtualPathTo(this.mainIncludedFolder, "CommandLine");
+            folder.ShouldBeADirectory(this.fileSystem);
```
wilbaker: Also validate that `Path.Combine(this.mainIncludedFolder, "Properties")` is still present?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            string[] fileSystemEntries = Directory.GetFileSystemEntries(newFolderPath);
+            fileSystemEntries.Length.ShouldEqual(32);
+
+            string projectedFolder = Path.Combine(newFolderPath, "Git");
+            projectedFolder.ShouldBeADirectory(this.fileSystem);
+            fileSystemEntries = Directory.GetFileSystemEntries(projectedFolder);
+            fileSystemEntries.Length.ShouldEqual(13);
+
+            string projectedFile = Path.Combine(newFolderPath, "ReturnCode.cs");
+            projectedFile.ShouldBeAFile(this.fileSystem);
```
wilbaker: Call `ValidateIncludedFolders` again here and verify that `newFolderPath` is now in the list? (I am guessing that's what should happen?)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            string[] fileSystemEntries = Directory.GetFileSystemEntries(newFolderPath);
+            fileSystemEntries.Length.ShouldEqual(33);
+
+            string projectedFolder = Path.Combine(newFolderPath, "Git");
+            projectedFolder.ShouldBeADirectory(this.fileSystem);
+            fileSystemEntries = Directory.GetFileSystemEntries(projectedFolder);
+            fileSystemEntries.Length.ShouldEqual(13);
+
+            string projectedFile = Path.Combine(newFolderPath, "ReturnCode.cs");
+            projectedFile.ShouldBeAFile(this.fileSystem);
```
wilbaker: Same comment here: Call `ValidateIncludedFolders` again with `newFolderPath`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            fileToCreate.ShouldBeAFile(this.fileSystem);
+        }
+
+        private void ValidateIncludedFolders(params string[] folders)
+        {
+            HashSet<string> actualIncludedFolders = new HashSet<string>(this.gvfsProcess.IncludedFoldersList());
+            folders.Length.ShouldEqual(actualIncludedFolders.Count);
+            foreach (string expectedFolder in folders)
+            {
+                actualIncludedFolders.Contains(expectedFolder).ShouldBeTrue($"{expectedFolder} not found in actual folder list");
```
wilbaker: Should/can we also verify that there are no extra/unexpected entries in `actualIncludedFolders`?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+using GVFS.FunctionalTests.Tools;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.IO;
+
+namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
+{
+    [TestFixture]
+    public class IncludedFolderTests : TestsWithEnlistmentPerFixture
```
wilbaker: Could you add a test that files of non-recursive included parent folders are projected?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        [TestCase, Order(8)]
+        public void ModifiedFileShouldNotAllowIncludedFolderChange()
+        {
+            string modifiedPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts", "RunFunctionalTests.bat");
+            this.fileSystem.WriteAllText(modifiedPath, "New Contents");
+
+            string output = this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            output.ShouldContain("Include was aborted");
```
wilbaker: Also validate a non-zero return code (assuming that change is made to the verb)?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            this.fileSystem.WriteAllText(modifiedPath, "New Contents");
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "add .");
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -m Test");
+
+            this.gvfsProcess.AddIncludedFolders(this.mainIncludedFolder);
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
+
+            string folderPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts");
+            folderPath.ShouldBeADirectory(this.fileSystem);
+            modifiedPath.ShouldBeAFile(this.fileSystem);
```
wilbaker: Why were these kept? Is it because the file is now a full file?

What would happen if the file were deleted instead of modified? (it might be useful to add a test for that as well)

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
```diff
@@ -0,0 +1,262 @@
...
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "add .");
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -m Test");
+
+            this.gvfsProcess.RemoveIncludedFolders("Scripts");
+            this.ValidateIncludedFolders(this.mainIncludedFolder);
+
+            string folderPath = Path.Combine(this.Enlistment.RepoRoot, "Scripts");
+            folderPath.ShouldBeADirectory(this.fileSystem);
+            string[] fileSystemEntries = Directory.GetFileSystemEntries(folderPath);
+            fileSystemEntries.Length.ShouldEqual(2);
```
wilbaker: Is this 2 because "RunFunctionalTests.bat" from the previous test is still there? (If so could you add a comment?)

Also, would the behavior be the same if the new file were created in a folder that did not previously have a modified file in it?  (should that be its own test?)

mjcheetham:
> Should that list the folders that get stored during add? All the items for the index that would be projected?

To keep things simple, probably just a list of the explicit `add`s and `remove`s?? Like

```shell
/os/src $ gvfs projection --list

+ foo
+ bar
- bar/common
- bar/tests
+ misc
```

I guess there could also be another command (out of scope) that would behave like an `ls` and list all items in the current directory level and their projection status.

```shell
/os $ gvfs projection --list

Type  Name       Projection state
--------------------------------------------
dir   src

/os/src $ cd src
/os/src $ gvfs projection --list

Type   Name             Projection state
--------------------------------------------
dir    foo              Included
dir    bar              Partial
dir    misc
file   main.c
file   .gitattributes
file   Makefile
file   LICENSE

/os/src $ cd bar
/os/src/bar $ gvfs projection --list

Type   Name         Projection state
--------------------------------------------
dir    common       Excluded
dir    tests        Excluded
file   bar.c
```

> It might also be a overloaded term in the code base since we already have the `GitIndexProjection` and now we would have the `ProjectionFolders` which could become confusing with each other.

I would say we should prioritise the user experience, and pick the name that makes most sense. I doubt we'd get confused in the codebase.

kewillford:
@mjcheetham I understand where you are coming from and it makes sense but there are some things we would have to clarify so the user wouldn't be confused.  For example:
>To keep things simple, probably just a list of the explicit `add`s and `remove`s?? Like

Currently this is simply a SQLite database with the list of folders to include in the projection and it is not tracking adds and removes.  It looks like you are suggesting to track what is adds and removes so I could add `foo` and all the children would to included but if I removed `foo\bar` and `bar` was a child then it would be removed?  This could be doable but is added complexity for the user and for the code.  It would simplify the case where a user wanted to include all the child folders except for one but I don't know of a requirement for this so I would rather keep it simple for now.

Also in your example for `ls` the included/excluded state is separate from the file/folder's projection state.  There can be things included that are virtual/partial/full or excluded that are partial/full.  All of this information I'm not sure the user is concerned with.  For now the requirements were very simple and then we can add to it as we get feedback.

I do think we need to settle on what the verb name should be and `projection` might not fit currently if there is an expectation that it has to do with the file or folder's on disk projection state as opposed to whether or not it is projected but if we are going to continue to enhance the functionality of the verb to encompass that I can see it working.

cc: @jrbriggs

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
kewillford: Yes it is because files that get added to ModifiedPaths.dat will stay there and git is in charge of them even if they are outside the sparse scope.  Added a delete test as well in which case the file and folder are gone but if you checkout a commit with the file then it will appear when outside the scope.

GVFS/GVFS.UnitTests/Mock/Virtualization/Projection/MockGitIndexProjection.cs
kewillford: Look into this and it was a lot more work than I thought so I think I will add an issue.  The `MockGitIndexProjection` is kind of a half mock in that it overrides some methods but derives from `GitIndexProjection` but calls a constructor that doesn't setup anything so I was getting a bunch of `NullReferenceException`s and that led me down a rabbit hole.

kewillford: Added issue #1366

GVFS/GVFS.UnitTests/Platform.Mac/MacFileSystemVirtualizerTests.cs
kewillford: Added an issue to track these changes. #1365

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/IncludedFolderTests.cs
kewillford: Yeah I updated this to use a different path to remove that confusion.

GVFS/GVFS/CommandLine/IncludeVerb.cs
```diff
@@ -0,0 +1,182 @@
...
+                                includedFolderTable.Remove(directoryPath);
+                            }
+                        }
+
+                        if (haveFoldersToAdd)
+                        {
+                            foreach (string directoryPath in this.Add.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries))
+                            {
+                                includedFolderTable.Add(directoryPath);
+                            }
```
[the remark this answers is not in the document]
kewillford: I'm not sure that really matters.  What would the user action be from those messages?  Also it would be more queries to the database because we would need to check each path before taking any action.  The only thing I can think of this being useful for is multiple scripts that might be adding the same paths to go clean up the scripts.  I'm going to hold off on this for now until we get some real users feedback.  Let me know if you want me to create an issue for this enhancement.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,6 +892,34 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
                 }
 
                 parentFolder = parentFolder.ChildEntries.GetOrAddFolder(indexEntry.BuildingProjection_PathParts[pathIndex]);
+
+                if (this.rootIncludedFolder.Children.Count > 0)
+                {
+                    if (indexEntry.BuildingProjection_LastEntryIncludedFolder == null)
```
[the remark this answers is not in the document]
kewillford: These will be removed with the folder only IsIncluded PR.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,6 +892,34 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
...
+                if (this.rootIncludedFolder.Children.Count > 0)
+                {
+                    if (indexEntry.BuildingProjection_LastEntryIncludedFolder == null)
+                    {
+                        indexEntry.BuildingProjection_LastEntryIncludedFolder = this.rootIncludedFolder;
+                    }
+
+                    if (!indexEntry.BuildingProjection_ShouldIncludeRecursive)
+                    {
+                        string folderName = indexEntry.BuildingProjection_PathParts[pathIndex].GetString();
```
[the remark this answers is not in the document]
kewillford: Resolved with #1345

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,6 +892,34 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
...
+                        if (indexEntry.BuildingProjection_LastEntryIncludedFolder.Children.ContainsKey(folderName) &&
+                            indexEntry.BuildingProjection_LastEntryIncludedFolder.Children[folderName].Depth == pathIndex)
+                        {
+                            indexEntry.BuildingProjection_ShouldInclude = true;
```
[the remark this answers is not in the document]
kewillford: Resolved with #1345

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.GitIndexEntry.cs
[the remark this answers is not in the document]
kewillford: This will be removed with #1345

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
kewillford: The directory will be marked as a placeholder and start projecting and it should get removed once a index change happens if nothing adds to the modified paths and they are all placeholders still but if the skipworktree bit is turned off, paths will get added to the modified paths and the folder will not get removed.  I did some testing around the reset scenarios and this is what happened.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
kewillford: The risk there is that if that is a top level folder and that folder gets added to the ModifiedPaths.dat, it could show all the files for that top level folder as deleted in the next status and any checkout after that could be updating a lot of files and we don't currently have a way for the user to dehydrate that folder once it gets added.

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -0,0 +1,242 @@
...
+using System.Text;
+
+namespace GVFS.CommandLine
+{
+    [Verb(
+        SparseVerb.SparseVerbName,
+        HelpText = @"List, add, or remove from the list of folders that are included in VFS for Git's projection.
+Folders need to be relative to the repos root directory.")
+    ]
+    public class SparseVerb : GVFSVerb.ForExistingEnlistment
```
wilbaker: Looks like the comment was dropped when this class was renamed.  Should we print out a message about this being experimental until we get feedback on its behavior (similar to the messages we print for dehydrate/repair)?

kewillford: I'm on the fence for this because will we need to add the `--confirm` to the sparse verb as well then?

I should add some text here that explains when there are zero folders, everything will be projected.

wilbaker: I personally don't think we'd need a `--confirm`, just a message to dissuade anyone from taking a hard dependency on the current behavior.

jrbriggs: I think just prepending "EXPERIMENTAL: " to the help text will suffice for letting users know this entire command is subject to change.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.FolderData.cs
```diff
@@ -29,14 +31,22 @@ public void ResetData(LazyUTF8String name)
...
             }
 
-            public FileData AddChildFile(LazyUTF8String name, byte[] shaBytes)
+            public void Include()
             {
-                return this.ChildEntries.AddFile(name, shaBytes);
+                this.IsIncluded = true;
+                for (int i = 0; i < this.ChildEntries.Count; i++)
+                {
+                    FolderData childFolder = this.ChildEntries[i] as FolderData;
```
wilbaker: Minor/Nit: In GitIndexProject I saw cast to `FolderData` using this approach:

```
if (childEntry.IsFolder)
{
   FolderData folderData = (FolderData)childEntry;
```

Use the same approach here?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -873,68 +975,68 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
...
-                }
+                this.projectionFolderCache.TryAdd(folderPath, folderData);
             }
 
             return true;
         }
 
         /// <summary>
-        /// Finds the FolderEntryData for the path provided
+        /// Takes a path and get the FolderData object fo that path if it exist and is a folder
```
wilbaker: Typos: 
`get` -> `gets`
`object fo that` -> `object for that`
`if it exist` -> `if it exists`

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -873,68 +975,68 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
...
+                return true;
+            }
+            else
+            {
+                EventMetadata metadata = CreateEventMetadata();
+                metadata.Add("folderPath", folderPath);
+                metadata.Add(TracingConstants.MessageKey.InfoMessage, "Found file at path");
+                this.context.Tracer.RelatedEvent(
+                    EventLevel.Informational,
+                    $"{nameof(this.TryGetOrAddFolderDataFromCache)}_FileAtPath",
```
wilbaker: `TryGetOrAddFolderDataFromCache` -> `TryGetFolderDataFromTreeUsingPath`

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -1046,6 +1034,32 @@ private HResult QueryFileNameHandler(string virtualPath)
...
+            {
+                EventMetadata metadata = this.CreateEventMetadata(virtualPath);
+                metadata.Add("isDirectory", true);
+                metadata.Add("triggeringProcessId", triggeringProcessId);
+                metadata.Add("triggeringProcessImageFileName", triggeringProcessImageFileName);
+                metadata.Add("HResult", hr.ToString());
+                this.Context.Tracer.RelatedError(metadata, nameof(this.NotifyNewFileCreatedHandler) + "_" + nameof(this.virtualizationInstance.MarkDirectoryAsPlaceholder) + " error");
```
wilbaker: I think you could probably replace 

`nameof(this.NotifyNewFileCreatedHandler) + "_" + nameof(this.virtualizationInstance.MarkDirectoryAsPlaceholder)` 

with 

`nameof(this.MarkDirectoryAsPlaceholder)`

PR #1331, opened 2019-07-16
Extend GitRepoTests to allow include mode

kewillford:
The GitRepoTests can be initialized with a ValidateWorkingTree
mode that is currently a boolean. If enabled, we will add extra
checks to make sure the working tree of our virtual enlistment
matches the enlistment from a vanilla Git repo.

To extend this to the "include" mode, we need to first modify
this boolean into an enum and add it as an extra option.

Second, we need to be very careful about what we are comparing.
We will no longer virtualize outside of the included folders.
Be careful to include files in the root directory.

The functional tests can then be run with "--partial-mode" to
exercise only this option, which will isolate the test failures
to this setting AND keep the test suite from getting too slow.

Currently, the partial mode space is probably wider than
necessary and can be reduced in later iterations. This
does provide _some_ coverage.

Signed-off-by: Derrick Stolee <dstolee@microsoft.com>


GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs
```diff
@@ -251,12 +252,56 @@ public DirectoryInfo WithAttribute(FileAttributes attribute)
...
+            private static bool IsMatchedPath(FileSystemInfo info, string repoRoot, HashSet<string> prefixes)
+            {
+                if (prefixes == null)
+                {
+                    return true;
+                }
+
+                string localPath = info.FullName.Substring(repoRoot.Length + 1);
+
+                if (localPath.Equals(".git"))
```
wilbaker: Should this also be `OrdinalIgnoreCase`?

GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs
```diff
@@ -297,24 +343,27 @@ public DirectoryInfo WithAttribute(FileAttributes attribute)
...
                             {
                                 expectedMoved = expectedEnumerator.MoveNext();
+
                                 continue;
                             }
                         }
 
                         Assert.Fail($"File names don't match: expected: {expectedEnumerator.Current.FullName} actual: {actualEnumerator.Current.FullName}");
+                        expectedMoved = expectedEnumerator.MoveNext();
+                        continue;
```
wilbaker: Why was this change needed?  With the call to `Assert.Fail` on line 351 will we ever hit this code?

GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs
```diff
@@ -297,24 +343,27 @@ public DirectoryInfo WithAttribute(FileAttributes attribute)
...
                     }
                     else
                     {
-                        (actualEnumerator.Current.Attributes & FileAttributes.Directory).ShouldNotEqual(FileAttributes.Directory, $"expected path: {expectedEnumerator.Current.FullName} actual path: {actualEnumerator.Current.FullName}");
+                        (actualEnumerator.Current.Attributes & FileAttributes.Directory).ShouldNotEqual(FileAttributes.Directory, $"expected file path: {expectedEnumerator.Current.FullName} actual directory path: {actualEnumerator.Current.FullName}");
 
                         FileInfo expectedFileInfo = (expectedEnumerator.Current as FileInfo).ShouldNotBeNull();
                         FileInfo actualFileInfo = (actualEnumerator.Current as FileInfo).ShouldNotBeNull();
-                        actualFileInfo.Length.ShouldEqual(expectedFileInfo.Length);
+                        actualFileInfo.Length.ShouldEqual(expectedFileInfo.Length, $"File lengths do not agree for {expectedEnumerator.Current.FullName} vs {actualEnumerator.Current.FullName}");
```
wilbaker: Include the lengths in the message as well?  (I can't remember if the default message makes it clear which values are expected and actual.)

GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs
kewillford: Not sure why @derrickstolee added this.  Seems unnecessary but maybe debugging you can continue through the `Fail`?

derrickstolee: I think I had changed the assert to be something that concatenated the error (so I could see the full list of differences) but somehow forgot to revert this change.

derrickstolee: I think I had changed the assert to be something that concatenated the error (so I could see the full list of differences) but somehow forgot to revert this change.

GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -56,10 +63,18 @@ public static void Main(string[] args)
             }
             else
             {
+                Settings.ValidateWorkingTreeMode validateMode = Settings.ValidateWorkingTreeMode.Full;
+
+                if (runner.HasCustomArg("--partial-mode"))
```
derrickstolee: To be consistent, should we rename this '--include-mode"?

GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs
```diff
@@ -251,12 +252,54 @@ public DirectoryInfo WithAttribute(FileAttributes attribute)
...
+                {
+                    if (localPath.StartsWith(prefixDir, StringComparison.OrdinalIgnoreCase))
+                    {
+                        return true;
+                    }
+
+                    if (prefixDir.StartsWith(localPath, StringComparison.OrdinalIgnoreCase) &&
+                        Directory.Exists(info.FullName))
+                    {
+                        // For example: localPath = "GVFS" and prefix is "GVFS\\GVFS".
```
wilbaker: > prefix is "GVFS\\GVFS"

Do we have an prefixes in our test that are more than one folder deep?

I might have missed it, but as best I could tell everything in `IncludeModeFolders` is a top level folder.

kewillford: Sorry this was my fault.  I didn't pull it [this commit](https://github.com/derrickstolee/VFSForGit/commit/c10e8ba2bf0ba53c6add2749f83f1928169c79b9)

PR #1340, opened 2019-07-19
PrjFSKextLogDaemon: Don't log every time a dropped kext message is detected

### LINKED ISSUE #1326: Mac: PrjFSKextLogDaemon records too many "One or more kext log messages have been dropped" errors
**NOTE** So far we've only see this issue when developers make local changes that add a lot of logging to the kext.

When the kext logging daemon detects a dropped message:

https://github.com/microsoft/VFSForGit/blob/37816dc3f6ce854e8609ff65bc9432bd2d14779a/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp#L297

It will log an error "One or more kext log messages have been dropped".  This can cause a problem if the kext is emitting so much logging that messages are being continuously dropped, as the "message dropped" error will flood the logs.

The logging of this error should be scaled back.  For example, the logging daemon could track how many messages have been dropped and log the count as an error (if > 1) every time the kext health data is recorded (currently every 30 min).

wilbaker:
Resolves #1326 

When developers add a lot of extra logging to the kext, it can result in a lot of dropped log messages, and the daemon log logging each time this occurred.  This lead to a lot of extra log messages about messages being dropped.

To avoid generating so much logging, the behavior of the daemon has been updated:

- The first time (in each 30 minute interval) that the daemon detects a missing log message it will log an error
- Every 30 minutes the daemon will record the total number of missed messages in the last interval and reset the count.

**TODO**

- [X] Manually test changes with dropped messages 

Sample logging (with the daemon's logging interval shortened to 2 minutes):

image

### DIFF
--- a/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
+++ b/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
@@ -44,8 +44,8 @@ static void StartLoggingKextMessages(io_connect_t connection, io_service_t servi
 static void HandleSigterm(int sig, siginfo_t* info, void* uc);
 static void SetupExitSignalHandler();
 
-static dispatch_source_t StartKextHealthDataPolling(io_connect_t connection);
-static bool TryFetchAndLogKextHealthData(io_connect_t connection);
+static dispatch_source_t StartPeriodicLoggingTimer(io_connect_t connection);
+static void FetchAndLogKextHealthData(io_connect_t connection);
 
 static void ReportDroppedKextMessages();
 
@@ -203,7 +203,7 @@ static void StartLoggingKextMessages(io_connect_t connection, io_service_t prjfs
     });
     dispatch_resume(logDataQueue->dispatchSource);
 
-    dispatch_source_t timer = StartKextHealthDataPolling(connection);
+    dispatch_source_t timer = StartPeriodicLoggingTimer(connection);
 
     PrjFSService_WatchForServiceTermination(
         prjfsService,
@@ -244,7 +244,7 @@ static void SetupExitSignalHandler()
     sigaction(SIGTERM, &newAction, &oldAction);
 }
 
-static dispatch_source_t StartKextHealthDataPolling(io_connect_t connection)
+static dispatch_source_t StartPeriodicLoggingTimer(io_connect_t connection)
 {
     dispatch_source_t timer = dispatch_source_create(DISPATCH_SOURCE_TYPE_TIMER, 0, 0, dispatch_get_main_queue());
     dispatch_source_set_timer(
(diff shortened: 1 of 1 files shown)


jeschu1:
@wilbaker the premises here is great.  If we dropped one message, the probably of dropping more goes up and we don't want to log a ridiculous amount of times that we dropped a message.

That said this was exposed by a developer testing by entering lots of log messages.  I think we should agree a strategy for developers to use for logging in test.  We certainly wouldn't want any of the 'test' messages the user entered coming up in telemetry either.  

I know commenting out the WriteJson is one method, is there a simple env variable we can change possible?  If you aren't sure I can ask around here :-)

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -283,6 +287,18 @@ static bool TryFetchAndLogKextHealthData(io_connect_t connection)
     return true;
 }
 
+static void ReportDroppedKextMessages()
+{
+    uint32_t droppedMessageCount = atomic_exchange(&s_droppedMessageCount, 0U);
```
jeschu1: This makes sense, we zero out s_droppedMessageCount and assign the current value to droppedMessageCount.  The convention may not be obvious to a kext newbe so consider a comment.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -296,12 +312,11 @@ static void LogKextMessage(os_log_type_t messageLogType, uint32_t messageFlags,
 {
     if (messageFlags & LogMessageFlag_LogMessagesDropped)
     {
-        // One or more earlier messages have been dropped
-        os_log_error(s_kextLogger, "One or more kext log messages have been dropped");
-        
-        JsonWriter messageWriter;
-        messageWriter.Add(MessageKey, "One or more kext log messages have been dropped");
-        WriteJsonToMessageListener(MessageType::Error, messageWriter);
+        if (1 == ++s_droppedMessageCount)
```
jeschu1: I was debating if we needed a log on the first message and a total count every 30 minutes, but i think it's not too much noise and a good call.
If we only did every 30 minutes we may message a message if the Kext is unloaded in that block of time.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -196,11 +203,7 @@ static void StartLoggingKextMessages(io_connect_t connection, io_service_t prjfs
     });
     dispatch_resume(logDataQueue->dispatchSource);
 
-    dispatch_source_t timer = nullptr;
-    if (TryFetchAndLogKextHealthData(connection))
```
jeschu1: Is it worth saving the result of `TryFetchAndLogKextHealthData ` the first time it's called?  If so maybe we don't call it in the next interval to match the previous functionality?  

It looks like we no longer use the bool result of `TryFetchAndLogKextHealthData`

wilbaker:
> That said this was exposed by a developer testing by entering lots of log messages. I think we should agree a strategy for developers to use for logging in test. We certainly wouldn't want any of the 'test' messages the user entered coming up in telemetry either.
>
> I know commenting out the WriteJson is one method, is there a simple env variable we can change possible? If you aren't sure I can ask around here :-)

@jeschu1 the logging daemon will only report errors to telemetry, and so I think the easiest option would be log test messages at the INFO or DEFAULT levels:

```
// Log message levels, these correspond to their os_log counterparts
enum KextLog_Level : uint32_t
{
    KEXTLOG_ERROR = 0,   // For important failures, always gets logged and causes INFO messages to be committed too.
    KEXTLOG_DEFAULT = 2, // Run-of-the-mill messages that shouldn't be too frequent as they always are logged to disk.
    KEXTLOG_INFO = 1,    // Verbose logging that might help with diagnosing problems; these don't usually get logged to disk, except when an ERROR is loggod.
};
```

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Yeah I went back and forth on the same thing, but I decided I didn't want us to ever lose that a message was dropped.

As you pointed out if we only logged in the timer and things are shut down between it fires we would miss that a message was dropped.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: > If so maybe we don't call it in the next interval to match the previous functionality?

@jeschu1 I think the previous behavior was actually less than ideal, because we'd call `TryFetchAndLogKextHealthData` here and then immediately again when the timer in `StartKextHealthDataPolling` started.

The old idea had been that if we're unable to fetch the health data there is no need to start the timer, but I realized while working on this we also use the timer to re-connect to the kext if there are connection issues, and so I *think* it's OK to keep trying every 30 minutes if the first attempt fails.  Let me know if you can think of any problems with this new approach.

> It looks like we no longer use the bool result of TryFetchAndLogKextHealthData

Nice catch, I will clean this up and see what should be renamed.

PR #1343, opened 2019-07-22
Only setting GIT_TERMINAL_PROMPT=0 where we don't have the terminal

turbonaitis:
Currently GIT_TERMINAL_PROMPT is always set to 0, when git is invoked, which prevents things, like credential helpers, from prompting from user input in the terminal, even during `gvfs clone` and similar scenarios.

We need to differentiate between, when we have terminal and not.

This is an initial version, to start the conversation.

### DIFF
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -18,6 +18,8 @@ namespace GVFS.Common
         public const string WorkingDirectoryRootName = "src";
         public const string UnattendedEnvironmentVariable = "GVFS_UNATTENDED";
 
+        public const string InteractiveEnvironmentVariable = "GVFS_INTERACTIVE";
+
         public const string DefaultGVFSCacheFolderName = ".gvfsCache";
 
         public const string GitIsNotInstalledError = "Could not find git.exe.  Ensure that Git is installed.";
--- a/GVFS/GVFS.Common/GVFSEnlistment.Shared.cs
+++ b/GVFS/GVFS.Common/GVFSEnlistment.Shared.cs
@@ -6,28 +6,48 @@ namespace GVFS.Common
 {
     public partial class GVFSEnlistment
     {
-        private static bool? isUnAttendedByDefault;
-        public static void SetIsUnattendedByDefault(bool defaultValue)
+        private static bool? isInteractiveByDefault;
+        public static void SetIsInteractiveByDefault(bool defaultValue)
         {
-            if (isUnAttendedByDefault.HasValue)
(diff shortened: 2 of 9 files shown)


turbonaitis:
Looking at this change now, this is a bit of a hack. What we really need is a way to determine if we're running interactively (with the user present and having terminal access) and change our behavior accordingly (both in this case and in some others). 

I already see there is something like this in `GVFS_UNATTENDED` environment variable, but I'm sure how it's used. Also, things like `GVFS.Mount` and `GVFS.Service` are non-interactive by default, so probably shouldn't require setting an environment variable. 

Question: where should we put such global state of the app? Enlistment is not exactly that.

turbonaitis:
OK, so in the end I've introduced a new environment variable and a new check to determine if we're running VFS4G interactively. All tests are (finally) passing.

turbonaitis:
Any thoughts on this one?

turbonaitis:
@wilbaker any thoughts on this? This has been open for a while now and I'd love to move forward with it in some way. Happy to have a synchronous chat, if you want to better understand, where this is coming from :)

wilbaker:
@kewillford I couldn't find any comments about why this line was added:

```
processInfo.EnvironmentVariables["GIT_TERMINAL_PROMPT"] = "0";
```

In the original PR that added it.  Any ideas why this change was made in the first place?

turbonaitis:
> @kewillford I couldn't find any comments about why this line was added:
> 
> ```
> processInfo.EnvironmentVariables["GIT_TERMINAL_PROMPT"] = "0";
> ```
> 
> In the original PR that added it. Any ideas why this change was made in the first place?

[GIT_TERMINAL_PROMPT](https://git-scm.com/docs/git/2.3.0#git-emGITTERMINALPROMPTem), when set to 0, signals to git (and any sub processes), that there's no terminal access. This prevents git from for example asking for users password in the terminal. This is necessary, when trying to do `git credential fill` from the background proess (GVFS.Mount), as otherwise git would try to wait for password input from the user (and either would fail with an exception, or wait forever, not sure).

The reason for this change, is that for cases, when we are actually in the terminal (during all interactive gvfs commands - clone, mount etc), we can prompt for users password in the terminal.  We'd use this variable in our custom git credential helper, to improve user experience. Currently, we're always displaying a GUI dialog, that gets user credentials, but it often gets lost, or ignored by the users and they keep reporting, that "I ran gvfs mount - and it hangs there forever". If we set this environment variable appropriately for the situation, for user interactions from the terminal, our git credential helper could inform the user right there, that there's a need for credentials etc. 

Does that make sense?

derrickstolee:
Closing because stale.

PR #1345, opened 2019-07-22
Use IsIncluded only on folder entries

kewillford:
@wilbaker these are the changes to use only the `IsIncluded` on a folder.

### DIFF
--- a/GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
+++ b/GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
@@ -89,7 +89,7 @@ namespace GVFS.UnitTests.Virtualization.Git
         {
             SortedFolderEntries sfe = SetupDefaultEntries();
             LazyUTF8String name = ConstructLazyUTF8String("{{shouldbeattheend");
-            sfe.AddFolder(name);
+            sfe.AddFile(name, new byte[20]);
             sfe[defaultFiles.Length + defaultFolders.Length].Name.ShouldEqual(name, "Item added at incorrect index.");
         }
 
@@ -98,7 +98,7 @@ namespace GVFS.UnitTests.Virtualization.Git
         {
             SortedFolderEntries sfe = SetupDefaultEntries();
             LazyUTF8String name = ConstructLazyUTF8String("((shouldbeatthestart");
-            sfe.AddFolder(name);
+            sfe.AddFile(name, new byte[20]);
             sfe[0].Name.ShouldEqual(name, "Item added at incorrect index.");
         }
 
@@ -117,15 +117,45 @@ namespace GVFS.UnitTests.Virtualization.Git
         }
 
         [TestCase]
-        public void AddItemWithIsIncludedFalse()
+        public void AddFolderNoIncludeChildrenShouldBeIncluded()
         {
             SortedFolderEntries sfe = SetupDefaultEntries();
             LazyUTF8String name = ConstructLazyUTF8String("IsIncludedFalse");
-            sfe.AddFile(name, new byte[20], isIncluded: false);
-            sfe.TryGetValue(name, out FolderEntryData folderEntryData).ShouldBeTrue();
-            folderEntryData.ShouldNotBeNull();
-            folderEntryData.IsFolder.ShouldBeFalse();
-            folderEntryData.IsIncluded.ShouldBeFalse();
+            sfe.GetOrAddFolder(new[] { name }, partIndex: 0, parentIsIncluded: false, rootSparseFolderData: new SparseFolderData());
+            ValidateFolder(sfe, name, isIncludedValue: true);
+        }
+
+        [TestCase]
+        public void AddFolderWhereParentIncludedIsFalseAndIncluded()
+        {
+            SortedFolderEntries sfe = SetupDefaultEntries();
+            LazyUTF8String name = ConstructLazyUTF8String("Child");
+            SparseFolderData sparseFolderData = new SparseFolderData();
+            sparseFolderData.Children.Add("Child", new SparseFolderData());
+            sfe.GetOrAddFolder(new[] { name }, partIndex: 0, parentIsIncluded: false, rootSparseFolderData: sparseFolderData);
+            ValidateFolder(sfe, name, isIncludedValue: false);
+        }
+
+        [TestCase]
+        public void AddFolderWhereParentIncludedIsTrueAndChildIsIncluded()
+        {
+            SortedFolderEntries sfe = SetupDefaultEntries();
+            LazyUTF8String name = ConstructLazyUTF8String("Child");
+            SparseFolderData sparseFolderData = new SparseFolderData();
+            sparseFolderData.Children.Add("Child", new SparseFolderData());
+            sfe.GetOrAddFolder(new[] { name }, partIndex: 0, parentIsIncluded: true, rootSparseFolderData: sparseFolderData);
+            ValidateFolder(sfe, name, isIncludedValue: true);
+        }
+
+        [TestCase]
+        public void AddFolderWhereParentIncludedIsTrueAndChildIsNotIncluded()
+        {
+            SortedFolderEntries sfe = SetupDefaultEntries();
+            LazyUTF8String name = ConstructLazyUTF8String("ChildNotIncluded");
+            SparseFolderData sparseFolderData = new SparseFolderData();
+            sparseFolderData.Children.Add("Child", new SparseFolderData());
+            sfe.GetOrAddFolder(new[] { name }, partIndex: 0, parentIsIncluded: true, rootSparseFolderData: sparseFolderData);
+            ValidateFolder(sfe, name, isIncludedValue: false);
         }
 
         [TestCase]
@@ -178,10 +208,9 @@ namespace GVFS.UnitTests.Virtualization.Git
             for (int i = 0; i < names.Length; i++)
             {
                 LazyUTF8String entryString = ConstructLazyUTF8String(names[i]);
-                entries.AddFile(entryString, new byte[20], isIncluded: true);
+                entries.AddFile(entryString, new byte[20]);
                 entries.TryGetValue(entryString, out FolderEntryData folderEntryData).ShouldBeTrue();
                 folderEntryData.ShouldNotBeNull();
-                folderEntryData.IsIncluded.ShouldBeTrue();
                 folderEntryData.IsFolder.ShouldBeFalse();
             }
         }
@@ -191,12 +220,20 @@ namespace GVFS.UnitTests.Virtualization.Git
             for (int i = 0; i < names.Length; i++)
             {
                 LazyUTF8String entryString = ConstructLazyUTF8String(names[i]);
-                entries.AddFolder(entryString);
-                entries.TryGetValue(entryString, out FolderEntryData folderEntryData).ShouldBeTrue();
-                folderEntryData.ShouldNotBeNull();
-                folderEntryData.IsIncluded.ShouldBeTrue();
-                folderEntryData.IsFolder.ShouldBeTrue();
+                entries.GetOrAddFolder(new[] { entryString }, partIndex: 0, parentIsIncluded: true, rootSparseFolderData: new SparseFolderData());
+                ValidateFolder(entries, entryString, isIncludedValue: true);
             }
+        }
+
+        private static void ValidateFolder(SortedFolderEntries entries, LazyUTF8String entryToValidate, bool isIncludedValue)
+        {
+            entries.TryGetValue(entryToValidate, out FolderEntryData folderEntryData).ShouldBeTrue();
+            folderEntryData.ShouldNotBeNull();
+            folderEntryData.IsFolder.ShouldBeTrue();
+
+            FolderData folderData = folderEntryData as FolderData;
+            folderData.ShouldNotBeNull();
+            folderData.IsIncluded.ShouldEqual(isIncludedValue, "IsIncluded does not match expected value.");
         }
     }
 }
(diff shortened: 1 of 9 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -691,17 +675,18 @@ private static List<ProjectedFileInfo> ConvertToProjectedFileInfos(SortedFolderE
             {
                 childEntry = sortedFolderEntries[i];
 
-                if (childEntry.IsIncluded)
+                if (childEntry.IsFolder)
```
derrickstolee: Perhaps `if ((childEntry as FolderData)?.IsIncluded == true)`? Maybe that's too terse and the nested `if` is better.

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
```diff
@@ -98,7 +98,7 @@ public void AddItemAtTheBeginning()
         {
             SortedFolderEntries sfe = SetupDefaultEntries();
             LazyUTF8String name = ConstructLazyUTF8String("((shouldbeatthestart");
-            sfe.AddFolder(name);
+            sfe.AddFile(name, new byte[20]);
```
wilbaker: Why was this test (and `AddItemAtEnd`) changed to insert a file rather than a folder?

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
```diff
@@ -117,15 +117,45 @@ public void ValidateOrderOfDefaultEntries()
...
         {
             SortedFolderEntries sfe = SetupDefaultEntries();
             LazyUTF8String name = ConstructLazyUTF8String("IsIncludedFalse");
-            sfe.AddFile(name, new byte[20], isIncluded: false);
-            sfe.TryGetValue(name, out FolderEntryData folderEntryData).ShouldBeTrue();
-            folderEntryData.ShouldNotBeNull();
-            folderEntryData.IsFolder.ShouldBeFalse();
-            folderEntryData.IsIncluded.ShouldBeFalse();
+            sfe.GetOrAddFolder(new[] { name }, partIndex: 0, parentIsIncluded: false, rootSparseFolderData: new SparseFolderData());
+            ValidateFolder(sfe, name, isIncludedValue: true);
```
wilbaker: Is `isIncludedValue` true because `rootSparseFolderData` is empty?

I had some trouble parsing the name of this test, perhaps something like `AddFolderWhenSparseFolderDataIsEmtpy` or `FoldersShouldBeVisibleWhenSparseFolderDataIsEmpty`?

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
```diff
@@ -117,15 +117,45 @@ public void ValidateOrderOfDefaultEntries()
...
+
+        [TestCase]
+        public void AddFolderWhereParentIncludedIsTrueAndChildIsNotIncluded()
+        {
+            SortedFolderEntries sfe = SetupDefaultEntries();
+            LazyUTF8String name = ConstructLazyUTF8String("ChildNotIncluded");
+            SparseFolderData sparseFolderData = new SparseFolderData();
+            sparseFolderData.Children.Add("Child", new SparseFolderData());
+            sfe.GetOrAddFolder(new[] { name }, partIndex: 0, parentIsIncluded: true, rootSparseFolderData: sparseFolderData);
+            ValidateFolder(sfe, name, isIncludedValue: false);
```
wilbaker: Is there a test to handle the case where a parent folder is recursively included?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1008,51 +965,50 @@ private FolderData AddFileToTree(GitIndexEntry indexEntry)
...
 
-        /// <summary>
-        /// Finds the FolderEntryData for the path provided
-        /// </summary>
-        /// <param name="pathParts">Path to desired entry</param>
-        /// <param name="folderEntryData">Out: FolderEntryData for pathParts</param>
-        /// <returns>True if the path could be found in the tree, and false otherwise</returns>
-        /// <remarks>If the root folder is desired, the pathParts should be an empty array</remarks>
-        private bool TryGetFolderEntryDataFromTree(LazyUTF8String[] pathParts, out FolderEntryData folderEntryData)
+        private bool TryGetFolderDataFromTreeUsingPath(string folderPath, out FolderData folderData)
```
wilbaker: Could you add a comment for this method explaining what it does?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1062,14 +1018,13 @@ private bool TryGetFolderEntryDataFromTree(LazyUTF8String[] pathParts, out Folde
         /// childPathParts.
```
wilbaker: Minor:  Above comment is no longer accurate:

```
        /// Finds the FolderEntryData at the specified depth of the path provided.  If depth == pathParts.Length
        /// TryGetFolderEntryDataFromTree will find the FolderEntryData specified in pathParts.  If
        /// depth is less than pathParts.Length, then TryGetFolderEntryDataFromTree will return an ancestor folder of
        /// childPathParts.
```

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1501,7 +1456,7 @@ private string GetNewProjectedShaForPlaceholder(string path)
             HashSet<string> existingPlaceholders)
         {
             bool foundFolder = this.TryGetOrAddFolderDataFromCache(relativeFolderPath, out FolderData folderData);
-            if (!foundFolder || !folderData.IsIncluded)
+            if (!foundFolder)
```
wilbaker: Why was the `!folderData.IsIncluded` check removed here?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -485,7 +469,7 @@ public virtual bool IsPathProjected(string virtualPath, out string fileName, out
                 childName: fileName,
```
wilbaker: Maybe add a comment above the call to `GetProjectedFolderEntryData` that mentions `null` will be returned if a path's parent folder is not included.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: Out of curiosity, how would you expect the performance of `as` to compare to the virtual `IsFolder` call?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -428,45 +428,29 @@ public virtual void GetFileTypeAndMode(string filePath, out FileType fileType, o
...
             try
             {
-                this.GetChildNameAndParentKey(virtualPath, out string fileName, out string parentKey);
-                FolderEntryData data = this.GetProjectedFolderEntryData(
-                    blobSizesConnection: null,
-                    childName: fileName,
-                    parentKey: parentKey);
-
-                if (data != null && !data.IsIncluded)
+                if (this.TryGetFolderDataFromTreeUsingPath(virtualPath, out FolderData folderData) &&
```
wilbaker: It looks like with this change we'll be skipping the lookup in the cache that would happen in `GetProjectedFolderEntryData`.  Do we know at this point that this folder won't be in the cache?

(I am guessing that because we're in `TryAddSparseFolder` we'd know the folder is not in the cache?)

Adding a comment about out expectations of the cache would be helpful.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -428,45 +428,29 @@ public virtual void GetFileTypeAndMode(string filePath, out FileType fileType, o
...
-            if (data == null)
-            {
-                return PathSparseState.NotFound;
-            }
-            else if (data.IsIncluded)
-            {
-                return PathSparseState.Included;
-            }
-            else
+            if (this.TryGetFolderDataFromTreeUsingPath(virtualPath, out FolderData folderData))
```
wilbaker: Similar question as in `TryAddSparseFolder`, why was this changed from `GetProjectedFolderEntryData` to `TryGetFolderDataFromTreeUsingPath`?

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
kewillford: Simply because it is an easier call.  The `AddFolder` method was removed and now there is
```
public FolderData GetOrAddFolder(
    LazyUTF8String[] pathParts,
    int partIndex,
    bool parentIsIncluded,
    SparseFolderData rootSparseFolderData)
```

Which is more setup work to call and checking the order shouldn't matter whether it is a file or a folder

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
kewillford: >Is isIncludedValue true because rootSparseFolderData is empty?

Yes because if the root sparse folder data doesn't have any children then everything should be included.

I will update the name of the test.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.SortedFolderEntries.cs
```diff
@@ -88,27 +88,57 @@ public void Clear()
...
+                            }
+                        }
+                    }
+                    else
+                    {
+                        isIncluded = false;
+                    }
+                }
+
+                return this.InsertFolder(pathParts[partIndex], ~index, isIncluded: isIncluded);
```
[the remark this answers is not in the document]
kewillford: If the entry is not found the index is set the in the complement of the value where it should be inserted.  This is so there only needs to be one call to look something up instead of looking it up to discover it isn't found and looking up the index where it should be inserted when it isn't found.

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
wilbaker: Ok thanks!

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.SortedFolderEntries.cs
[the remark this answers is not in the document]
wilbaker: Thanks that makes sense

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: Running a simple test and `as` was in most cases about 2x slower but I had to up the iterations in the test to at least a million before it was even taking 1 ms so unless we are using it in a tight loop I'm sure other code will be a bottleneck before this.

derrickstolee: Well, thanks for checking the perf! That would have been a mistake to listen to me.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: `GetProjectedFolderEntryData` will not return entries that are not included and we we need to be able to return `Excluded` entries.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: This is for the create new folder scenario and we checked the sparse state and got back excluded in order to get to this code and excluded folders do not get added to the cache.  I will add some comments.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
kewillford: `TryGetOrAddFolderDataFromCache` will return false when it is not included
```
if (!this.projectionFolderCache.TryGetValue(folderPath, out folderData))
{
	if (!this.TryGetFolderDataFromTreeUsingPath(folderPath, out folderData) ||
		!folderData.IsIncluded)
	{
		folderData = null;
		return false;
	}

	this.projectionFolderCache.TryAdd(folderPath, folderData);
}

return true;
```

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1612,7 +1582,7 @@ private FSResult RemoveFolderPlaceholderIfEmpty(IPlaceholderData placeholder)
             string childName;
             string parentKey;
             this.GetChildNameAndParentKey(placeholder.Path, out childName, out parentKey);
-            PathSparseState sparseState = this.GetPathSparseState(placeholder.Path);
+            PathSparseState sparseState = this.GetFolderPathSparseState(parentKey);
```
wilbaker: Can this line be removed completely?  It looks like `sparseState` is only getting used for this check:

`sparseState != PathSparseState.Included`

And as best I can tell `this.TryGetSha` will return false for folders that are not included (it uses `GetProjectedFolderEntryData`).  If we can remove this call it will save us a lot of walking the tree in the more common scenario that placeholders on disk are still included.

kewillford: Looks like it.  I will give it a try and see.

PR #1348, opened 2019-07-23
Rename include to sparse throughout the code

kewillford:
This is only a renaming PR.  Decided that `sparse` would be a better verb and concept in the code since there is that concept in git with `sparse-checkout` and this feature is trying to match what the `sparse-checkout` in git does.


derrickstolee:
This is a good name change. I don't think there will be much confusion due to mental collisions with `sparse-checkout`.

PR #1351, opened 2019-07-24
No longer set PATH for git process and remove use of gvfs hooks root

kewillford:
The git process was launched using a PATH environment variable with the
gvfs hooks path to make sure that when a path in the repo was included in
the user's PATH environment variable that it would only search outside
the repo.  This was when the code was using git cat-file to hydrate files
and searching for gvfs hooks would cause it to hydrate files which would
cause it to search for hooks and it would be in a loop.

This is no longer the case, the code is using libgit2 and the hooks are not
called.  Since the gvfs hooks root was only used for setting the PATH for
the git process, all references to it can be removed.

Todo:
- [x] Test with the large repos with the `PATH` having paths in other VFS for git repos as well as the repo being mounted with various tools.

### DIFF
--- a/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
+++ b/GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
@@ -38,9 +38,9 @@ namespace GVFS.Platform.POSIX
         {
         }
 
-        public override bool TryGetGVFSHooksPathAndVersion(out string hooksPaths, out string hooksVersion, out string error)
+        public override bool TryGetGVFSHooksVersion(out string hooksVersion, out string error)
         {
-            hooksPaths = string.Empty;
+            string hooksPaths = string.Empty;
             string binPath = Path.Combine(this.Constants.GVFSBinDirectoryPath, GVFSPlatform.Instance.Constants.GVFSHooksExecutableName);
             if (File.Exists(binPath))
             {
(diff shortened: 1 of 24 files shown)


GVFS/GVFS.Platform.POSIX/POSIXPlatform.cs
```diff
@@ -38,9 +38,9 @@ public override void ConfigureVisualStudio(string gitBinPath, ITracer tracer)
         {
         }
 
-        public override bool TryGetGVFSHooksPathAndVersion(out string hooksPaths, out string hooksVersion, out string error)
+        public override bool TryGetGVFSHooksVersion(out string hooksVersion, out string error)
         {
-            hooksPaths = string.Empty;
+            string hooksPaths = string.Empty;
```
jrbriggs: Let's condense `hooksPath` and `binPath`, since I think they're effectively the same? They're also unused in this particular method, but I think the TODO tracks that work.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -1101,29 +1098,12 @@ private GVFSEnlistment CreateEnlistment(string enlistmentRootPath, GitAuthentica
                     this.ReportErrorAndExit("Error: " + GVFSConstants.GitIsNotInstalledError);
                 }
 
-                string hooksPath = null;
-                if (GVFSPlatform.Instance.UnderConstruction.RequiresDeprecatedGitHooksLoader)
```
wilbaker: It looks like this was the only place this `RequiresDeprecatedGitHooksLoader ` was being used.  Can you remove it?

PR #1363, opened 2019-07-25
Handle corrupt objects on Mac

### LINKED ISSUE #1218: Mac: Turn on Functional Test Categories.MacTODO.NeedsCorruptObjectFix
This may require both code and test changes.

jeschu1:
Remove Categories.MacTODO.NeedsCorruptObjectFix 

Handle corrupt objects on Mac by adding them to modifiedPaths.dat and printing an error.

resolves #1218

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
@@ -498,9 +498,7 @@ BOOL APIENTRY DllMain( HMODULE hModule,
             (badObject as FileInfo).ShouldNotBeNull().Length.ShouldEqual(objectFileInfo.Length);
         }
 
-        // TODO(#1218): Figure out why git for Mac is not requesting a redownload of the truncated object
         [TestCase, Order(17)]
-        [Category(Categories.MacTODO.NeedsCorruptObjectFix)]
         public void TruncatedObjectRedownloaded()
         {
             GitProcess.InvokeProcess(this.Enlistment.RepoRoot, "checkout " + this.Enlistment.Commitish);
@@ -521,6 +519,8 @@ BOOL APIENTRY DllMain( HMODULE hModule,
             GitProcess.InvokeProcess(this.Enlistment.RepoRoot, "rev-parse :Test_EPF_WorkingDirectoryTests/TruncatedObjectRedownloaded_copy.txt").Output.Trim().ShouldEqual(sha);
             string testFileContents = this.Enlistment.GetVirtualPathTo("Test_EPF_WorkingDirectoryTests", "TruncatedObjectRedownloaded_copy.txt").ShouldBeAFile(this.fileSystem).WithContents();
             objectPath.ShouldBeAFile(this.fileSystem);
+            string modifedFile = "Test_EPF_WorkingDirectoryTests/TruncatedObjectRedownloaded.txt";
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, modifedFile);
 
             // Truncate the contents of objectPath
             string tempTruncatedObjectPath = objectPath + "truncated";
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
```diff
@@ -541,11 +541,25 @@ public void TruncatedObjectRedownloaded()
             objectPath.ShouldBeAFile(this.fileSystem);
             new FileInfo(objectPath).Length.ShouldEqual(objectLength - 16);
 
-            // Read the original path and verify its contents are correct
-            this.Enlistment.GetVirtualPathTo("Test_EPF_WorkingDirectoryTests", "TruncatedObjectRedownloaded.txt").ShouldBeAFile(this.fileSystem).WithContents(testFileContents);
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                // Mac can't correct corrupt objects, but it should detect them and add to ModifiedPaths.dat
```
wilbaker: Should we also validate that the file is reported as modified by git status?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -380,6 +380,7 @@ private bool TryGetSymLinkTarget(string sha, out string symLinkTarget)
                             byte[] buffer = new byte[4096];
                             uint bufferIndex = 0;
                             int nextByte = stream.ReadByte();
+                            int bufferSize = 0;
```
wilbaker: Naming suggestion: `bufferSize` -> `bytesWritten`

The name `bufferSize` made me think this was somehow related to `byte[] buffer`

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -402,8 +403,19 @@ private bool TryGetSymLinkTarget(string sha, out string symLinkTarget)
...
                                 }
                             }
+                            bufferSize += Convert.ToInt32(bufferIndex);
+
+                            if (bufferSize != blobLength)
+                            {
+                                // If the read size does not match the expected size print an error and add the file to ModifiedPaths.dat
+                                // This allows the user to see that something went wrong with file hydration
+                                // Unfortunitely we must do this check *after* the file is hydrated since the header isn't corrupt for trunctated objects on mac
+                                this.OnLogError($"Read {relativePath} to {bufferSize}, not expected size of {blobLength}");
```
wilbaker: `this.OnLogError` is a callback from the ProjFS user-mode library.  

Rather than calling `this.OnLogError` ourselves we can use `this.Context.Tracer.RelatedError` (which keeps `OnLogError` specifically for logging errors from ProjFS)

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -402,8 +403,19 @@ private bool TryGetSymLinkTarget(string sha, out string symLinkTarget)
...
+                                this.FileSystemCallbacks.OnFileConvertedToFull(relativePath);
```
wilbaker: The end result is the same, but I think the intent of this call might be clearer if we call `OnFailedPlaceholderUpdate` instead.

Another option would be to add a new `OnFailedFileHydration` background task that does the same thing.

The only *potential* functionality impact would be if one day we decide to change what the BG thread does for these tasks.  If a user were to upgrade a new version of VFS4G with pending BG tasks, the persisted `OnFileConvertedToFull` or `OnFailedPlaceholderUpdate` tasks that we scheduled here could be processed differently.

cc: @kewillford for his thoughts, currently we only call `OnFailedPlaceholderUpdate` when we fail to update a placeholder as part of a projection change.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
jeschu1: Looks like the best way to do this is CheckGitCommandAgainstGVFSRepo (Since we're not in GitCommands).  Let me know if that is incorrect.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
jeschu1: I added `OnFailedFileHydration `, I think there is value in being explicit here.  As @wilbaker points out it also protects us if changes are made in the future to other tasks.

GVFS/GVFS.Virtualization/Background/FileSystemTask.cs
```diff
@@ -21,6 +21,7 @@ public enum OperationType
             OnFileOverwritten,
             OnFileSuperseded,
             OnFileConvertedToFull,
+            OnFailedFileHydration,
```
wilbaker: `OnFailedFileHydration` needs to be added to the end of the list.  

`OperationType` values are persisted, and if we add new values in the middle it could break things when VFS4G loads values from disk.

Could you add a comment that new values should be added to the end?  It isn't obvious looking at this code that we need to.

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
wilbaker: > I added OnFailedFileHydration 

This call to `OnFileConvertedToFull` still needs to be updated (to `OnFailedFileHydration`).

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
```diff
@@ -500,7 +500,6 @@ public void AllNullObjectRedownloaded()
 
         // TODO(Mac): Figure out why git for Mac is not requesting a redownload of the truncated object
```
wilbaker: @jeschu1 can you remove this TODO comment before merging?

PR #1368, opened 2019-07-25
More cleanup sparse

kewillford:


GVFS/GVFS.Common/Database/SparseTable.cs
```diff
@@ -68,22 +67,27 @@ public HashSet<string> GetAll()
...
             catch (Exception ex)
             {
-                throw new GVFSDatabaseException($"{nameof(SparseTable)}.{nameof(this.Remove)}({directory}) Exception: {ex.ToString()}", ex);
+                throw new GVFSDatabaseException($"{nameof(SparseTable)}.{nameof(this.Remove)}({directoryPath}) Exception: {ex.ToString()}", ex);
             }
         }
+
+        private static string NormalizePath(string path)
+        {
+            return path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar).Trim(Path.DirectorySeparatorChar);
```
wilbaker: - Do we also need to trim whitespace?  Or will that have been done by now?
- A unit test for this method might be a good idea.

GVFS/GVFS.FunctionalTests/Tools/GVFSProcess.cs
```diff
@@ -51,12 +55,28 @@ public bool TryMount(out string output)
...
-            return this.CallGVFS($"sparse {this.enlistmentRoot} -r {string.Join(";", folders)}");
+            return this.SparseCommand(addFolders: false, shouldSucceed: true, folders: folders);
+        }
+
+        public string RemoveSparseFolders(bool shouldSucceed, params string[] folders)
+        {
+            return this.SparseCommand(addFolders: false, shouldSucceed: shouldSucceed, folders: folders);
+        }
+
+        public string SparseCommand(bool addFolders, bool shouldSucceed, params string[] folders)
```
wilbaker: Are there any code paths where this will be called with `shouldSucceed` set to `false`?  Or will this feature be used in a subsequent commit?

kewillford: Yes see line 206, 255 in SparseTests.

GVFS/GVFS.Common/Database/SparseTable.cs
kewillford: Will add code to trim the whitespace and some functional tests.  The unit tests for this class are coming in another PR.

GVFS/GVFS.Common/Database/SparseTable.cs
```diff
@@ -68,22 +67,27 @@ public HashSet<string> GetAll()
...
             catch (Exception ex)
             {
-                throw new GVFSDatabaseException($"{nameof(SparseTable)}.{nameof(this.Remove)}({directory}) Exception: {ex.ToString()}", ex);
+                throw new GVFSDatabaseException($"{nameof(SparseTable)}.{nameof(this.Remove)}({directoryPath}) Exception: {ex.ToString()}", ex);
             }
         }
+
+        private static string NormalizePath(string path)
+        {
+            return path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar).Trim(Path.DirectorySeparatorChar, ' ');
```
wilbaker: Minor: If you call `Trim()` rather than passing in `' '` `Trim()` will trim all characters for which [`Char.IsWhiteSpace`](https://docs.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=netframework-4.8) would return `true`.

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -160,7 +161,9 @@ private string[] ParseFolderList(string folders)
             }
             else
             {
-                return folders.Split(new[] { FolderListSeparator }, StringSplitOptions.RemoveEmptyEntries);
+                return folders.Split(new[] { FolderListSeparator }, StringSplitOptions.RemoveEmptyEntries)
+                    .Select(x => x.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar).Trim(Path.DirectorySeparatorChar, ' '))
+                    .ToArray();
```
wilbaker: Could/should we call `NormalizePath` here (assuming it's made public) instead?

GVFS/GVFS.Common/Database/SparseTable.cs
kewillford: My question was should it be `.Trim().Trim(Path.DirectorySeparatorChar)` or `.Trim(Path.DirectorySeparatorChar).Trim()` or `.Trim().Trim(Path.DirectorySeparatorChar).Trim()` or ...

kewillford: Went with `.Trim().Trim(Path.DirectorySeparatorChar)` since if someone put a space before the directory separator then it will stay but that seems like a user error for the path to me.

wilbaker:
@kewillford the latest changes look good to me

PR #1375, opened 2019-07-25
Remove all TODO(POSIX) and TODO(Mac) comments.

### LINKED ISSUE #1357: Ensure all POSIX/Mac TODOs have work items or are removed
There are several TODOs in the code without work items tracking them

wilbaker:
Resolves #1357 

For each TODO(POSIX) or TODO(Mac) comment one of the following was done:

- Comment was updated with specific issue number
- Comment was removed if no longer relevant
- Change was made to address the TODO

### DIFF
--- a/GVFS/GVFS.Common/NativeMethods.cs
+++ b/GVFS/GVFS.Common/NativeMethods.cs
@@ -98,14 +98,6 @@ namespace GVFS.Common
             }
         }
 
-        public static void CreateHardLink(string newLinkFileName, string existingFileName)
-        {
-            if (!CreateHardLink(newLinkFileName, existingFileName, IntPtr.Zero))
-            {
-                ThrowLastWin32Exception($"Failed to create hard link from '{newLinkFileName}' to '{existingFileName}'");
-            }
-        }
-
         /// <summary>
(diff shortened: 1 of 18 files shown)


GVFS/GVFS.Common/NativeMethods.cs
```diff
@@ -98,14 +98,6 @@ public static void MoveFile(string existingFileName, string newFileName, MoveFil
             }
         }
 
-        public static void CreateHardLink(string newLinkFileName, string existingFileName)
```
jeschu1: I saw this was removed, did we not need it or is it unsupported?  I didn't see a HardLink issues, but I'm known to be blind!

wilbaker: @jeschu1 this is dead code.

I am guessing it was added when we were experimenting with using hardlinks to create GVFS_projection rather than performing a file copy.

cc: @kewillford as FYI that I'm removing this code.

kewillford: Yeah we can always add it back in if we want to try and use hard links again.

PR #1379, opened 2019-07-26
Add unit tests for SparseTable and pull out some common code to base …

kewillford:
…class


GVFS/GVFS.UnitTests/Common/Database/TableTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using GVFS.Tests.Should;
+using GVFS.UnitTests.Category;
+using Moq;
+using NUnit.Framework;
+using System;
+using System.Data;
+
+namespace GVFS.UnitTests.Common.Database
+{
+    public abstract class TableTests<T>
```
derrickstolee: This is a great pattern!

PR #1381, opened 2019-07-29
Upgrade: Fix test flakiness around upgrade reminders

### LINKED ISSUE #852: NoReminderForLeftOverDownloads test occasionally fails

derrickstolee:
Should fix #852.

The fix is to use an environment variable to always show the upgrade message, if present.

While looking at the post-command hook code, I discovered some simple cleanups that are isolated to different commits:

* Removed unreachable code.
* Changed a message to refer to `gvfs status` instead of `git status`.

### DIFF
--- a/GVFS/GVFS.Hooks/Program.cs
+++ b/GVFS/GVFS.Hooks/Program.cs
@@ -233,7 +233,7 @@ namespace GVFS.Hooks
                 {
                     if (response == null || response.ResponseData == null)
                     {
-                        Console.WriteLine("\nError communicating with GVFS: Run 'git status' to check the status of your repo");
+                        Console.WriteLine("\nError communicating with GVFS: Run 'gvfs status' to check the status of your repo");
                     }
                     else if (response.ResponseData.HasFailures)
                     {
--- a/GVFS/GVFS.Hooks/Program.cs
+++ b/GVFS/GVFS.Hooks/Program.cs
@@ -109,7 +109,7 @@ namespace GVFS.Hooks
             int reminderFrequency = 10;
             int randomValue = random.Next(0, 100);
 
-            if (randomValue <= reminderFrequency &&
+            if ((IsUpgradeMessageDeterministic() || randomValue <= reminderFrequency) &&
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Hooks/Program.cs
```diff
@@ -452,5 +422,17 @@ private static string GetGitCommandSessionId()
                 return string.Empty;
             }
         }
+
+        private static bool IsUpgradeMessageDeterministic()
+        {
+            try
+            {
+                return Environment.GetEnvironmentVariable("GVFS_UPGRADE_DETERMINISTIC", EnvironmentVariableTarget.Process) != null;
```
jeschu1: Just to check, getting an env variable is so cheap it wont affect performance?  Am I reading the code correctly that this will be run on every command?

derrickstolee: Environment variables are cheap. Reading Git config or a file on disk is problematic and slow.

wilbaker:
@derrickstolee with this change the functional tests are no longer testing/validating the upgrade reminder logic in the hooks because `GVFS_UPGRADE_DETERMINISTIC` is only used in the tests.

Rather than adding the check for `GVFS_UPGRADE_DETERMINISTIC` would we be better off disabling the test(s) until we have an approach for showing the upgrade message that's deterministic?

jamill:
> would we be better off disabling the test(s) until we have an approach for showing the upgrade message that's deterministic?

I think this change allows us to test *most* of the product code path, and provides value. I guess we could get even closer by tweaking the following line:

`if ((IsUpgradeMessageDeterministic() || randomValue <= reminderFrequency)`

to something like:

```
int reminderFrequency = IsUpgradeMessageDeterministic() ? 100 : 10;
...
`if ((randomValue <= reminderFrequency)`
...
```
So this environment variable would only change the frequency (to have it run every time).

wilbaker:
>  I guess we could get even closer by tweaking the following line:
> `int reminderFrequency = IsUpgradeMessageDeterministic() ? 100 : 10;`

@jamill thanks for posting that suggestion, it made me realize I misread the change to the `if` statement.  I incorrectly thought that `IsUpgradeMessageDeterministic` was skipping the rest of the checks (rather than just skipping the random number portion).

Thanks!

/cc: @derrickstolee

PR #1383, opened 2019-07-29
[TEST] Revert "Mac build scripts: consume git installer pkg directly"

derrickstolee:
Reverts microsoft/VFSForGit#1318

[The CI builds](https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10036740&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b) started failing with this merge. The issue is related to the mac GVFS.Service not loading correctly. PR builds are all failing right now with this issue.

This revert is to try and see if it fixes the problem, or if we need to get on the machine and do manual steps.


derrickstolee:
This was not the problem.

PR #1384, opened 2019-07-29
[Mac] Don't allow cloning or mounting on case sensitive volumes

### LINKED ISSUE #1222: Mac: Validate filesystem is configured to be case insensitive before cloning and mounting
Currently VFS4G only supports Mac when its file system is configured to be case insensitive.

jeschu1:
resolves #1222


GVFS/GVFS.Mount/InProcessMountVerb.cs
```diff
@@ -87,6 +87,11 @@ public void Execute()
                 // any platform specific preparation required to run as a background process.
                 GVFSPlatform.Instance.PrepareProcessToRunInBackground();
             }
+
+            if (GVFSPlatform.Instance.FileSystem.IsCaseSensitive(this.EnlistmentRootPathParameter))
```
derrickstolee: **This will block the Linux team.**

Instead: maybe make this an `IsFileSystemSupported()` method on the platform, and the Mac platform rejects case-sensitive filesystems?

jeschu1: @derrickstolee updated to IsFileSystemSupported

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -29,6 +30,24 @@ public override bool IsSocket(string fileName)
...
+        public override bool IsFileSystemSupported(string path, out string error)
+        {
+            string lowerCaseFile = Guid.NewGuid().ToString().ToLower();
+            string upperCaseFile = lowerCaseFile.ToUpper();
+
+            File.Create(Path.Combine(path, lowerCaseFile));
+            if (File.Exists(upperCaseFile))
+            {
+                File.Delete(lowerCaseFile);
+                error = "Case sensitive unsupported";
```
wilbaker: Minor: Can you make this error message a bit more descriptive?  Maybe something like "VFS for Git does not support case sensitive filesystems"

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -29,6 +30,24 @@ public override bool IsSocket(string fileName)
             return NativeStat.IsSock(statBuffer.Mode);
         }
 
+        public override bool IsFileSystemSupported(string path, out string error)
+        {
+            string lowerCaseFile = Guid.NewGuid().ToString().ToLower();
+            string upperCaseFile = lowerCaseFile.ToUpper();
+
+            File.Create(Path.Combine(path, lowerCaseFile));
```
wilbaker: `File.Create` can throw, catch/handle the exceptions here?

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -29,6 +30,24 @@ public override bool IsSocket(string fileName)
             return NativeStat.IsSock(statBuffer.Mode);
         }
 
+        public override bool IsFileSystemSupported(string path, out string error)
+        {
+            string lowerCaseFile = Guid.NewGuid().ToString().ToLower();
+            string upperCaseFile = lowerCaseFile.ToUpper();
+
+            File.Create(Path.Combine(path, lowerCaseFile));
+            if (File.Exists(upperCaseFile))
```
wilbaker: Should this check be reversed?  If we create the lower-case file and the upper-case file exists, doesn't that mean the file system is case insensitive?

GVFS/GVFS.Mount/InProcessMountVerb.cs
```diff
@@ -87,6 +87,11 @@ public void Execute()
                 // any platform specific preparation required to run as a background process.
                 GVFSPlatform.Instance.PrepareProcessToRunInBackground();
             }
+
+            if (!GVFSPlatform.Instance.FileSystem.IsFileSystemSupported(this.EnlistmentRootPathParameter, out string errorMessage))
```
wilbaker: A better spot for this check would be in `MountVerb` rather than `InProcessMountVerb`.  If we check in `MountVerb` the user will be able to see the error message in the console (`InProcessMountVerb` is running as a background process).

GVFS/GVFS.Common/GVFSEnlistment.cs
```diff
@@ -220,6 +220,11 @@ public bool TryCreateEnlistmentFolders()
             try
             {
                 Directory.CreateDirectory(this.EnlistmentRoot);
+                if (!GVFSPlatform.Instance.FileSystem.IsFileSystemSupported(this.EnlistmentRoot, out string error))
+                {
+                   throw new InvalidRepoException($"FileSystem unsupported: {error}");
```
wilbaker: It looks a bit strange to be throwing `InvalidRepoException` from inside a `Try` method that typically returns true/false.  

Could you either:

- Update the method to return an output error string and return false when IsFileSystemSupported returns false

-or-

- Move the call to `IsFileSystemSupported` up so that it's after the call to `TryCreateEnlistmentFolders` so that we can return a more specific error there?  In other words:

```
if (!enlistment.TryCreateEnlistmentFolders())
{
    string error = "Could not create enlistment directory";
    tracer.RelatedError(error);
    return new Result(error);
}

if (!enlistment.IsFileSystemSupported(this.EnlistmentRoot, out string error))
{
    string error = $"FileSystem unsupported: {error}";
    tracer.RelatedError(error);
    return new Result(error);
}

```

GVFS/GVFS.Mount/InProcessMountVerb.cs
jeschu1: I did think about that, but then if the service starts the process we won't see the check right?
It's such a rare case that the clone would success and the mount would not, I'm OK with the error being a bit harder to find. 

What do you think?

wilbaker: > I did think about that, but then if the service starts the process we won't see the check right?

The service calls `gvfs mount` and so the check will still be run:

https://github.com/microsoft/VFSForGit/blob/d0336e0690e58fd86c478f2c53e896dc3bed3b6e/GVFS/GVFS.Service/GVFSMountProcess.Mac.cs#L24

> I'm OK with the error being a bit harder to find.  
> What do you think?

Since `gvfs mount` is always called when mounting I think it makes sense to make the check there.  

Making the errors easier to surface to users was one of the motivations for updating GVFS.Service to call `gvfs mount` rather than `GVFS.Mount` directly.

jeschu1: @wilbaker I didn't realize we had moved the service to call gvfs mount'.  Thanks for the update there.  In that case we definitely want the check in MountVerb, updated.

Issue #1387, opened 2019-07-30
Make 'git gui' not block other git commands


jeschu1:
There are a few possible solutions here:

1. Windows creates 'git-gui.exe' which uses 'git update-index', instead of 'git gui' which uses 'git-update-index.exe'.  'git-gui.exe' or equivalent doesn't appear to be build for Mac.  @dscho, do you know why this is?
2. Make 'git-update-index.exe' use command hooks
3. Make 'git gui' use 'git update-index' instead of 'git-update-index.exe'

For now allowing 'git gui' to run and block other commands, is deemed better than it not giving it a way to run at all on mac.

dscho:
FWIW the `git-gui.exe` in Git for Windows exists mainly so that we have a neat target for the Start Menu item "Git GUI"...

PR #1388, opened 2019-07-30
[WIP] Proper error message for 'git gui' on Mac

### LINKED ISSUE #700: Mac: `git gui` doesn't work
```
~/BugBashRepos/Office.Prototype/src>git gui
To access the 'git gui' in a GVFS repo, please invoke 'git-gui.exe' instead.
fatal: pre-command hook aborted command
~/BugBashRepos/Office.Prototype/src>git-gui
-bash: git-gui: command not found
```

jeschu1:
resolves #700

#1387 has been created (not prioritized) to allow 'git-gui'

### DIFF
--- a/GVFS/GVFS.Hooks/Program.cs
+++ b/GVFS/GVFS.Hooks/Program.cs
@@ -132,7 +132,7 @@ namespace GVFS.Hooks
             switch (command)
             {
                 case "gui":
-                    ExitWithError("To access the 'git gui' in a GVFS repo, please invoke 'git-gui.exe' instead.");
+                    ExitWithError(GVFSHooksPlatform.GetGitGuiBlockedMessage());
                     break;
             }
         }
--- a/GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
+++ b/GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Mac.cs
@@ -23,5 +23,10 @@ namespace GVFS.Hooks.HooksPlatform
         {
             return MacPlatform.GetNamedPipeNameImplementation(enlistmentRoot);
         }
+
+        public static string GetGitGuiBlockedMessage()
+        {
+            return "git gui is not supported in VFS for Git repos on Mac";
+        }
     }
 }
--- a/GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Windows.cs
+++ b/GVFS/GVFS.Hooks/HooksPlatform/GVFSHooksPlatform.Windows.cs
@@ -46,5 +46,10 @@ namespace GVFS.Hooks.HooksPlatform
         {
             return WindowsPlatform.GetUpgradeHighestAvailableVersionDirectoryImplementation();
         }
+
+        public static string GetGitGuiBlockedMessage()
+        {
+            return "To access the 'git gui' in a GVFS repo, please invoke 'git-gui.exe' instead.";
+        }
     }
 }


GVFS/GVFS.Hooks/Program.cs
```diff
@@ -128,12 +129,15 @@ private static void ExitWithError(params string[] messages)
 
         private static void CheckForLegalCommands(string[] args)
         {
-            string command = GetGitCommand(args);
-            switch (command)
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
```
derrickstolee: I don't think this is the recommended pattern here. We try to rely on the `GVFSPlatform.Instance` and a property therein. 

Alternatively: why not let this be in all platforms? We would just delete this entire block and put up a notice in #1387 that we currently block when you use `git gui` but would not block if you run `git-gui` instead.

derrickstolee: s/block/lock

jeschu1: @wilbaker thoughts on allowing windows users to call 'git gui', knowing that it will be blocking?  They can still of course run 'git-gui' and it would simplify our code base.

kewillford: What is the point of letting users run `git gui` if nothing will work?  Is that the case or will some things work other not?  Are we going to document for users what works and what doesn't?  Seems like a bad user experience to say you can run `git gui` but we don't know what will work or what will be an issue.

jeschu1: @kewillford, to clarify since 'git gui' takes a GVFS lock everything run in that application works.  However if you run 'git status' while 'git gui' is open, you will get a message saying it needs to wait for the lock.

wilbaker: > I don't think this is the recommended pattern here. We try to rely on the GVFSPlatform.Instance and a property therein.

👍 

> @wilbaker thoughts on allowing windows users to call 'git gui', knowing that it will be blocking?

We used to have that behavior on Windows and received enough feedback regarding the poor experience that we decided to block it and force users to call `git-gui`.  I'm not sure that we should go back to allowing users to get into that situation on Windows.  

> Is that the case or will some things work other not? Are we going to document for users what works and what doesn't?

Another option here (if we think the blocking behavior of `git gui` is a bad enough experience that we should block users from getting into it) would be to keep the block but print a more accurate message on Mac (e.g. "`git gui` is not supported in VFS for Git repos").  

If we do allow `git gui` on Mac I like the idea of printing a notice to users to let them know that an open git gui window might block other git commands from running. 

jeschu1: spoke with @jrbriggs

For now we'll fix the error message to say "git gui is not supported in VFS for Git repos on Mac".  

If we get user feedback that "git gui" is needed, we should consider making "/usr/local/git/libexec/git-core" accessible via path on mac.

jeschu1: More fun!

Please see comments as to why I stuck with RuntimeInformation.  
Also this did require loading a tiny dll.  If we hate that idea, perhaps we can change the existing error message to just say 

"To access the 'git gui' in a GVFS repo, please invoke 'git-gui.exe' instead. Note you may have to access 'git-gui.exe' by it's full path on mac"

Thoughts @wilbaker?

wilbaker: > Please see comments as to why I stuck with RuntimeInformation.

I couldn't find any comments on this.  Rather than using `RuntimeInformation` could you add a new method to GVFSHooksPlatform? 

Something like this in GVFSHooksPlatform.Mac.cs:

```
        public static string GetGitGuiBlockedMessage()
        {
            return "git gui is not supported in VFS for Git repos on Mac";
        }
```

And the current message on Windows?

PR #1389, opened 2019-07-30
Update Git to v2.23.0

derrickstolee:
See microsoft/git#165 for the rebase onto v2.23.0.windows.1.

This PR has a few reactions to that update:

1. A newline was dropped in a message. Remove it from our expectation in a test.
2. Rebase now has unpredictable extra line of whitespace sometimes. However, `-q` doesn't work in tandem with `--merge`, but using `GIT_QUIET=true` makes the output quiet on Windows. Still doesn't work on OSX for some reason, so also ignore the error output and focus on the end-result. Use `git rev-parse HEAD^{tree}` to ensure we got the same resulting working directory.
3. `FileContentsShouldMatch()` used to fail if both of the files didn't exist. Something changed in `v2.23.0` such that a cherry-pick loses a file in both vanilla Git and VFS for Git, so the test started failing on that path. Relax the condition, since the files _do_ match.
4. `KnownGitCommands` needs to include `restore` and `switch`, resolving #1463.
5. Includes the performance fix for `git checkout -b` from microsoft/git#183.


kewillford:
Did we run the performance tests against this version?

Also there is a `IsCheckoutWithFilePaths` that we should update with the new `restore`

derrickstolee:
> Did we run the performance tests against this version?

Thank you for the reminder, @kewillford! I just ran the large repo perf tests and found that `git checkout -b <newbranch>` (i.e. creating a new branch at the current head) is as slow as moving the checkout to a new position.  That is, it took ~10s which was the same as moving ten commits away, but before it took ~0.3s.

It makes sense that our `-b` optimizations would break with the refactoring that was done related to `git switch`. Likely, Duy purposefully broke our fast code by now assuming a user would only do `git checkout -b` if they _really_ want to enforce the checkout and would use `git switch -c <newbranch>` if they don't want to change the working directory. The question is: do we go fix that `-b` behavior? (I think so, unless someone knows for sure why that was dropped.)

cc: @jrbriggs, @wilbaker, @jeffhostetler, @dscho

wilbaker:
> The question is: do we go fix that -b behavior? (I think so, unless someone knows for sure why that was dropped.)

I agree that it'd be best to fix that behavior.  Even if `switch -c` is fast users might have scripts that are using `checkout -b` and would expect it to be quick.

derrickstolee:
A follow-up question: do we block this PR until we track down the -b fix, or do we file an issue and fix it in another PR before cutting a new release?

derrickstolee:
Ok, I found the culprit, commit 65f099b3. It has this commit message:

```

    switch: no worktree status unless real branch switch happens

    When we switch from one branch to another, it makes sense to show a
    summary of local changes since there could be conflicts, or some files
    left modified.... When switch is used solely for creating a new
    branch (and "switch" to the same commit) or detaching, we don't really
    need to show anything.

    "git checkout" does it anyway for historical reasons. But we can start
    with a clean slate with switch and don't have to.

    This essentially reverts fa655d8411 (checkout: optimize "git checkout
    -b <new_branch>" - 2018-08-16) and make it default for switch,
    but also for -B and --detach. Users of big repos are encouraged to
    move to switch.
```

"Users of big repos are encouraged to move to switch."

I'll create a PR to revert this commit in our branch, then bring it up on the list to incorporate into core Git again.

derrickstolee:
[The one comment about this patch was in support of this change.](https://public-inbox.org/git/CAPig+cTR8px1MCq0qXijt1x2+BAgpsKCiRR1hw3hHiVc2NkHAA@mail.gmail.com/):

```
I like this last bit. The skip_merge_working_tree() function which
this removes was ugly, difficult to maintain, and difficult to get
just right (and easy to break -- even by changing parts of the system
which one might not expect to impact it).
```

Cc: @benpeart, @kewillford, @jrbriggs, @jeffhostetler: Shall we just live with this performance degradation and move people to `git switch -c`? or should we maintain a revert of this change?

jeffhostetler:
Is there a way to slip in a hint-when-slow into "checkout -b" directing them to "switch -c" -- much like we have the hint for no ahead/behind on status?

Then we could avoid having a tug-o-war on what "checkout -b" does.

The hint might even be welcomed upstream.

benpeart:
I wonder if we should treat the removal of this optimization more like a deprecation.  Go a release or two with 'checkout -b' recommending the new 'switch' command and as usage drops off, remove the optimization.

derrickstolee:
Thanks for the discussion, everyone. microsoft/git#183 has the revert and the deprecation warning.

derrickstolee:
Performance of `git checkout -b` is back to normal. Merging.

PR #1392, opened 2019-07-30
Mac: Add version information to kext and PrjFS logging daemon

### LINKED ISSUE #1181: Mac: Set proper version on kext and the kext logging daemon
The kext and PrjFSLogDaemon do not have the correct version number set on them.

Related issues have been opened for the other native binaries:

- Hooks: #1390 
- dylibs: #1391

wilbaker:
Resolves #1181 

The changes in this PR set the [CFBundleVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion?language=objc) and [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring?language=objc) of the kext and the kext logging daemon.

Apple uses a slightly different versioning scheme than we do, and I applied the following mapping:

- `CFBundleVersion` -> Our version numbers are `W.X.Y.Z`, and `CFBundleVersion` is set to `Y.Z`.  I went with `Y.Z` as it's the build number portion of the VFS4G version.

- `CFBundleShortVersionString` -> Officially only versions of the format `[Major].[Minor].[Patch]` are supported, but based on what I've found [on StackOverflow](https://stackoverflow.com/questions/21125159/which-ios-app-version-build-numbers-must-be-incremented-upon-app-store-release) as best I can tell this is only enforced when uploading to the app store.  For now I've left this as our usual version number, and it displays in the "Info" dialog of the kext and logger:

image


wilbaker:
@jrbriggs, @nickgra, @jeschu1, @alameenshah, @pmj 

I arrived at the changes in this PR using a lot of trial and error.  I don't have experience with the versioning schemes used by Apple, and if you have any recommendations for better choices for `CFBundleVersion` or `CFBundleShortVersionString` please let me know.

If in the future we need to move to a `[Major].[Minor].[Patch]` for `CFBundleShortVersionString` I was thinking we could apply the following transformation to our versions:

`W.X.Y.Z` ->  `W.X.YZ` (adding padding to `Z` so that it's always at least 3 digits).

Example:

`1.0.19175.3` -> `1.0.19175003`

ProjFS.Mac/Scripts/GeneratePrjFSXCConfig.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ -z $VERSION ]; then
+  VERSION="0.2.173.2"
+fi
+
+BUNDLE_VERSION="N/A"
+if [[ $VERSION =~ ([0-9])\.([0-9]*)\.([0-9]*)\.([0-9]*) ]]
+then
+  BUNDLE_VERSION="${BASH_REMATCH[3]}.${BASH_REMATCH[4]}"
+else
+  echo "Failed to determine MAJOR_VEBUILD_NUMBERRSION from $VERSION"
```
nickgra: Should we early exit here? Or is "N/A" a legal bundle version?

pmj:
Sorry, only just got around to taking a look at this. Those version numbers look like they'll conflict with some of [the limits Apple places](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364) on the various positions of the bundle version string:

> The build version number should be a string comprised of three non-negative, period-separated integers with the first integer being greater than zero—for example, 3.1.2. The string should only contain numeric (0-9) and period (.) characters. Leading zeros are truncated from each integer and will be ignored (that is, 1.02.3 is equivalent to 1.2.3). The meaning of each element is as follows:
> 
> * The first number represents the most recent major release and is limited to a maximum length of four digits.
> * The second number represents the most recent significant revision and is limited to a maximum length of two digits.
> * The third number represents the most recent minor bug fix and is limited to a maximum length of two digits.
>
> If the value of the third number is 0, you can omit it and the second period.
> 
> While developing a new version of your app, you can include a suffix after the number that is being updated; for example 3.1.3a1. The character in the suffix represents the stage of development for the new version. For example, you can represent development, alpha, beta, and final candidate, by d, a, b, and fc. The final number in the suffix is the build version, which cannot be 0 and cannot exceed 255. When you release the new version of your app, remove the suffix.

I suspect the OS isn't very fussy about enforcing these limits for apps and dylibs, but as far as I'm aware, the limits are enforced to some extent for kexts. I'm sure we can dig up the extent to which they're enforced in the kexttools source or similar, but it's probably best to stick with the published limitations. The only thing I wouldn't worry about is the first number being greater than 0 - the OS definitely accepts versions starting with 0.

Is the "Z" in our version number important to include in all public releases? If not, we could use it in only development builds by using a `W.X.YdZ` versioning scheme, and just `W.X.Y` for release builds.

That leaves the limit of 4 digits for `W` and `X` and 2 digits for `Z`. I don't know how our version numbers are generated, so I'm not sure to what extent that'll be a problem. If necessary, we'll need to subdivide the version number differently.

wilbaker:
@pmj thanks for the details on the version number limitations!  It explains why I'm seeing our official build pipeline fail with these changes (the version numbers are too long in that pipeline).

I think for now we can simplify the versioning scheme to:

`CFBundleVersion` -> [Major].[Minor]
`CFBundleShortVersionString` -> Our usual version number.

The main purpose of this PR is to get the version to appear correctly in Finder, and I'm seeing the behavior that you mentioned (it uses `CFBundleShortVersionString`).

Our versioning scheme frequently has 5 digits in the `Y` field of `W.X.Y.Z` (and the `Z` value is important) and so unless we find a technical reason for adjusting `CFBundleVersion` I don't think it's worth the complexity of mapping our version number to the format enforced by Apple.

The kext uses our regular version string for its interface version, and so that will protect us from kext <-> lib version mismatches.

wilbaker:
>I think for now we can simplify the versioning scheme to:
>
>CFBundleVersion -> [Major].[Minor]
>CFBundleShortVersionString -> Our usual version number.

cc: @jrbriggs as FYI

PR #1394, opened 2019-07-31
Increasing precision of timestamps being logged

turbonaitis:
Currently, we only log seconds, but for some operations, it would be great to be able to have better resolution of log timestamps, so I'm increasing the precision of timestamps in logs to 1/10^4 of seconds.

### DIFF
--- a/GVFS/GVFS.Common/Tracing/EventListener.cs
+++ b/GVFS/GVFS.Common/Tracing/EventListener.cs
@@ -41,7 +41,7 @@ namespace GVFS.Common.Tracing
         {
             // Make a smarter guess (than 16 characters) about initial size to reduce allocations
             StringBuilder message = new StringBuilder(1024);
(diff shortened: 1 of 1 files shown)


kewillford:
Do you have some examples of the operations that you need this extra precision for?  Maybe we need to add some explicit start activities to track the timing of them?

turbonaitis:
There isn't a specific set of operations, that I can recall right now. 

Another use case for more precision is correlating GVFS.Mount activity with external activity. (If we ran `build.sh` at 11:00:01.153 and it took 43.245s, having this added precision, we can filter all the events events, that happened during that time, presumably as a result of that build). 

Do you think there's a significant cost in increasing this precision?

kewillford:
>Do you think there's a significant cost in increasing this precision?

I haven't ever measured that so couldn't say but I would doubt it.  

Mainly I was wondering if there is code that we needed the `Start` and `Stop` in the logs which measures duration in milliseconds.  But your other use case also makes sense.

turbonaitis:
Thanks for your comments and approval!

Issue #1397, opened 2019-07-31
Mac: Do not download sizes when updating placeholders


wilbaker:
`PrjFS_UpdatePlaceholderFileIfNeeded` does not require file sizes.  The parameter should be removed and we should skip the file size download step when updating the projection on Mac.

This eliminates the need for #255

PR #1398, opened 2019-07-31
Mac: Don't download file sizes as part of OnEnumerateDirectory

### LINKED ISSUE #699: Mac: Navigating a fresh clone in Finder can be slow
Finder enumerates one level beyond what's shown (at least in column view).  This mean that expanding a folder (e.g. the src folder) the first time can take a long time.

Investigate if there are any options for speeding this up.  Some ideas include:

- Aggressively downloading sizes\loose objects during clone
- Expanding src (and one folder deep) during clone

wilbaker:
Resolves #699 

As part of #1130 PrjFSLib was updated to stop filling placeholders with empty data.  With that change, VFS4G on Mac no longer needs to download sizes when it writes placeholders.

With this optimization expanding the `src` folder of a fresh clone in Finder was reduced from ~2min30sec to ~5sec on my MacBook.

The same optimization can be made when VFS4G updates placeholders on Mac, and #1397 has been filed for that.

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
@@ -99,7 +99,6 @@ namespace GVFS.Platform.Mac
                     relativePath,
                     PlaceholderVersionId,
                     ToVersionIdByteArray(FileSystemVirtualizer.ConvertShaToContentId(sha)),
-                    (ulong)endOfFile,
                     fileMode);
 
                 return new FileSystemResult(ResultToFSResult(result), unchecked((int)result));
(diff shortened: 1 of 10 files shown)


GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -99,7 +99,6 @@ public override void Stop()
                     relativePath,
```
kewillford: Will you add a `summary` to this method indicating that the `endOfFile` parameter is not used?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -612,7 +599,9 @@ private Result CreatePlaceholders(string directoryRelativePath, IEnumerable<Proj
                 else
                 {
                     sha = fileInfo.Sha.ToString();
-                    fileSystemResult = this.WritePlaceholderFile(childRelativePath, fileInfo.Size, sha);
+
+                    // Writing placeholders on Mac does not require a file size
+                    fileSystemResult = this.WritePlaceholderFile(childRelativePath, endOfFile: 0, sha: sha);
```
kewillford: Add a `const` named something like `None` or `NotUsed` and maybe the value should be `-1`?

GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs
```diff
@@ -560,31 +559,19 @@ private void OnFileRenamed(string relativeDestinationPath, bool isDirectory)
...
-                    {
-                        projectedItems = this.FileSystemCallbacks.GitIndexProjection.GetProjectedItems(CancellationToken.None, blobSizesConnection, relativePath);
-                    }
+                IEnumerable<ProjectedFileInfo> projectedItems;
 
-                    result = this.CreatePlaceholders(relativePath, projectedItems, triggeringProcessName);
-                }
-                catch (SizesUnavailableException e)
+                // TODO: Pool these connections or schedule this work to run asynchronously using TryScheduleFileOrNetworkRequest
+                using (BlobSizes.BlobSizesConnection blobSizesConnection = this.FileSystemCallbacks.BlobSizes.CreateConnection())
```
kewillford: Since this isn't even used when `populateSizes` is `false`, should we pass `null` to `GetProjectionItems` and remove this code?

kewillford: And maybe that can be the indicator to get the sizes is if the `blobSizesConnection` is not null instead of adding the `populateSizes` parameter?

wilbaker: Both great suggestions, thanks!

PR #1399, opened 2019-07-31
Add SetupFixture for teardown to get the test results at the end

kewillford:
Currently the output for running functional tests put the service logs and the run times for each of the tests at the end after the actual test summary results.  This makes finding the actual results of failures harder to find since you have to scroll up an indeterminate amount to get to the results.
 
This PR will move the test run summary to the bottom of the output. So you can go straight to the bottom to see any failures.


GVFS/GVFS.FunctionalTests/Program.cs
```diff
@@ -108,7 +107,6 @@ public static void Main(string[] args)
 
             RunBeforeAnyTests();
```
wilbaker: @kewillford could the code in `RunBeforeAnyTests` be moved to `GlobalSetup.RunBeforeAnyTests`, or would that be called too early?

wilbaker: (I think the current changes are good, I was just wondering if this code could be moved too)

kewillford: I will test that and see.  I don't see anything in the code that should prevent it and from my understanding it will be ran at the beginning of the `RunTests` call below but I haven't looked into the code to see what it actually is doing.

PR #1401, opened 2019-08-01
Allow non-azure URLs for upgrade nuget feed

turbonaitis:
Currently, if the configured upgrade feed is not from azure devops, it doesn't work. Allowing the organisations to configure internal nuget feeds would allow them to better control the rollout of new versions.


derrickstolee:
Since we removed the nuget feed capability in #1746, this PR no longer applies.

PR #1402, opened 2019-08-01
Update Git to package including Linux installer

derrickstolee:
We now package a Linux installer with our Git builds. Update to include that build, even though Git itself did not change at all between those builds.


derrickstolee:
I'll get the Linux installer for free when I update to include microsoft/git#167.

PR #1403, opened 2019-08-01
Mac: Do not lookup or download file sizes when updating placeholders

### LINKED ISSUE #255: Mac: Batch file sizes up-front for new placeholders that are created during projection changes
On Mac `ReExpandFolder` might lay down new placeholders, and the sizes for these placeholders are not downloaded\batched up front (like they are for placeholders that are updated by VFSForGit).

### LINKED ISSUE #1397: Mac: Do not download sizes when updating placeholders

wilbaker:
Resolves #255 
Resolves #1397 

As part of #1130 PrjFSLib was updated to stop filling placeholders with empty data. With that change, VFS4G on Mac no longer needs to download sizes when it updates placeholders or re-expands folders in response to the index being updated.

A new `EmptyPlaceholdersRequireFileSize` property was added to `IKernelDriver` that VFS4G can use to check if the platform requires sizes for empty placeholders (allowing the current size lookup/download logic to continue working for Windows and Linux).

See related PR #1398 for a similar change to the enumeration callback path.

**Performance improvements**

*First Large Repo*

| Step | Master | This PR and #1398 | Difference
--- | --- | --- | ---
| Enumerating Folders | 147.08 | 138.86 | -8.22
| git checkout feature/gvfs/perftest/defaultBranch | 31.53 | 18.09 | -13.44

*Second Large Repo*

| Step | Master | This PR and #1398 | Difference
--- | --- | --- | ---
| Enumerating Folders | 193.69 | 172.87 | -20.82
| git checkout feature/gvfs/perftest/defaultBranch | 13.57 | 11.4 | -2.17

### DIFF
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -1210,14 +1210,9 @@ namespace GVFS.Virtualization.Projection
                 folderPlaceholdersToKeep.Add(string.Empty);
 
                 stopwatch.Restart();
-                this.ProcessListOnThreads(
-                    numThreads,
-                    placeholderFilesListCopy,
-                    (placeholderBatch, start, end, blobSizesConnection, availableSizes) =>
-                        this.BatchPopulateMissingSizesFromRemote(blobSizesConnection, placeholderBatch, start, end, availableSizes),
-                    (placeholder, blobSizesConnection, availableSizes) =>
-                        this.UpdateOrDeleteFilePlaceholder(blobSizesConnection, placeholder, folderPlaceholdersToKeep, availableSizes));
+                this.MultiThreadedPlaceholderUpdatesAndDeletes(numThreads, placeholderFilesListCopy, folderPlaceholdersToKeep);
                 stopwatch.Stop();
+
                 long millisecondsUpdatingFilePlaceholders = stopwatch.ElapsedMilliseconds;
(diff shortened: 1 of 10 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1210,14 +1210,9 @@ private void UpdatePlaceholders()
...
 
                 stopwatch.Restart();
-                this.ProcessListOnThreads(
-                    numThreads,
-                    placeholderFilesListCopy,
-                    (placeholderBatch, start, end, blobSizesConnection, availableSizes) =>
-                        this.BatchPopulateMissingSizesFromRemote(blobSizesConnection, placeholderBatch, start, end, availableSizes),
-                    (placeholder, blobSizesConnection, availableSizes) =>
-                        this.UpdateOrDeleteFilePlaceholder(blobSizesConnection, placeholder, folderPlaceholdersToKeep, availableSizes));
+                this.MultiThreadedPlaceholderUpdatesAndDeletes(numThreads, placeholderFilesListCopy, folderPlaceholdersToKeep);
```
kewillford: We could move the `numThreads` calculation into this method since that is the only place it is used.

PR #1406, opened 2019-08-02
auto download and install VFSForGit Git on Linux

chrisd8088:
These scripts are very similar to the Mac versions, but instead download the `gitforlinux.gvfs.installer` NuGet package and use the `.deb` package it contains to install the VFSForGet-enabled Git client on Linux.

We also add a `PrepFunctionalTests.sh` script to use with the functional test suite.

### DIFF
--- a/Scripts/Linux/BuildGVFSForLinux.sh
+++ b/Scripts/Linux/BuildGVFSForLinux.sh
@@ -41,17 +41,11 @@ if [ ! -d $BUILDDIR ]; then
   mkdir $BUILDDIR || exit 1
 fi
 
-# TODO(Linux): download VFS-enabled version of upstream Git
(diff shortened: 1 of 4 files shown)


chrisd8088:
@wilbaker: I only put @derrickstolee on the review list, but if you'd like to review this too, or instead, that's cool!

chrisd8088:
Thanks, @derrickstolee -- and you need to have [libprojfs](https://github.com/github/libprojfs) installed somewhere your compiler and linker can find it.  So either set `CPPFLAGS=-I/path/to/libprojfs/include` and `LD_LIBRARY_PATH=/path/to/libprojfs/lib` if you've got it somewhere unique; otherwise, if it's in `/usr` or `/usr/local`, we should find it automatically.

PR #1407, opened 2019-08-02
stop rename over .git/index on Linux without lock

chrisd8088:
This is a small fix to the work committed in 2e32e7694061668d8f3f7ee4e2c9b9220d654ece so that we also prevent renaming/overwriting the `.git/index` file unless the GVFS provider lock is held.  Previously we only prevented the `.git/index` file from being renamed, but failed to stop another file being renamed over it.

### DIFF
--- a/GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
@@ -512,9 +512,7 @@ namespace GVFS.Platform.Linux
         {
             try
             {
-                bool pathInsideDotGit = Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath);
-                if (pathInsideDotGit &&
-                    relativePath.Equals(GVFSConstants.DotGit.Index, StringComparison.Ordinal))
+                if (relativePath.Equals(GVFSConstants.DotGit.Index, GVFSPlatform.Instance.Constants.PathComparison))
                 {
                     string lockedGitCommand = this.Context.Repository.GVFSLock.GetLockedGitCommand();
                     if (string.IsNullOrEmpty(lockedGitCommand))
@@ -542,10 +540,8 @@ namespace GVFS.Platform.Linux
         {
             try
             {
-                bool pathInsideDotGit = Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath);
-                if (pathInsideDotGit &&
-                    (relativePath.Equals(GVFSConstants.DotGit.Index, StringComparison.Ordinal) ||
-                     relativeDestinationPath.Equals(GVFSConstants.DotGit.Index, StringComparison.Ordinal)))
+                if (relativePath.Equals(GVFSConstants.DotGit.Index, GVFSPlatform.Instance.Constants.PathComparison) ||
+                    relativeDestinationPath.Equals(GVFSConstants.DotGit.Index, GVFSPlatform.Instance.Constants.PathComparison))
                 {
                     string lockedGitCommand = this.Context.Repository.GVFSLock.GetLockedGitCommand();
                     if (string.IsNullOrEmpty(lockedGitCommand))


chrisd8088:
@wilbaker: I only put @derrickstolee on the review list, but if you'd like to review this too, or instead, that's cool!

GVFS/GVFS.Platform.Linux/LinuxFileSystemVirtualizer.cs
```diff
@@ -543,8 +543,10 @@ private Result OnPreRename(string relativePath, string relativeDestinationPath,
             try
             {
                 bool pathInsideDotGit = Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativePath);
-                if (pathInsideDotGit &&
-                    (relativePath.Equals(GVFSConstants.DotGit.Index, StringComparison.Ordinal) ||
+                bool destinationPathInsideDotGit = Virtualization.FileSystemCallbacks.IsPathInsideDotGit(relativeDestinationPath);
+                if ((pathInsideDotGit &&
+                     relativePath.Equals(GVFSConstants.DotGit.Index, StringComparison.Ordinal)) ||
+                    (destinationPathInsideDotGit &&
                      relativeDestinationPath.Equals(GVFSConstants.DotGit.Index, StringComparison.Ordinal)))
```
kewillford: Okay this code is a little confusing and is probably from when we were checking more things in the dot git folder.  But before it was if the `relativePath` started with `.git/` and either path equals the `.git/index`.  Now it is if `relativePath` starts with `.git/` and is the `.git/index` or the `relativeDestination` starts with `.git/` and is the `.git/index`

The `IsPathInsideDotGit` is doing this
```
        /// <summary>
        /// Returns true for paths that begin with ".git\" (regardless of case)
        /// </summary>
        public static bool IsPathInsideDotGit(string relativePath)
        {
            return relativePath.StartsWith(GVFSConstants.DotGit.Root + Path.DirectorySeparatorChar, StringComparison.OrdinalIgnoreCase);
        }
```
So it is using a `OrdinalIgnoreCase` when doing the starts with but then is using `Ordinal` when comparing the entire path against `.git/index`?  Can we remove the `IsPathInsideDotGit` check and simply make the `Equals` call?  It doesn't seem to be needed does it?

Do you have any tests that could demonstrate the issue and that this fixes it?

chrisd8088: Well, the check in `IsPathInsideDotGit()` is changed elsewhere, so there isn't a mismatch in case-sensitivity the longer run -- it just looks like it in this small PR.

The larger case-sensitivity change is in [#1412](https://github.com/microsoft/VFSForGit/pull/1412/files#diff-020796c269a4b274eb4d1ddd4a6730cc), and that changes it to use the new [`GVFSPlatform.Instance.Constants.PathComparison`](https://github.com/github/VFSForGit/blob/655673e559ec283f17861dc93f95085abd06f855/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs#L154).

It's a good point that we don't really need the `IsPathInsideDotGit()` changes, however, because we've refactored from the [Mac original](https://github.com/microsoft/VFSForGit/blob/a1053d2e516a29ecb665f69e94030b875029a47c/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L483) where it was used in a nested set of conditionals.  Let me see what I can do to simplify this and set up a simple test!

chrisd8088: @kewillford -- I [simplified](https://github.com/github/VFSForGit/commit/ec94d04d2a32594d2101f3c5d0a922b08eefcc33) this as you suggested -- thanks for the idea!

As for testing, I ended up putting together a separate PR #1428 because it turns out there are tests for most of the behaviour here (specifically, this small PR fixes the existing `LockPreventsRenameFromInsideWorkingTreeOnTopOfIndex()` test on Linux); however, we don't have a test that I'm aware of which tests the *other* rename check, i.e., renaming `.git/index` to something else without the lock on a platform which doesn't deliver a pre-delete event prior to renaming, such as our Linux implementation.

Anyway, I'm going to commit this small fix to `features/linuxprototype`, and put the new test up in a PR against master.  Thanks again for the advice!

PR #1408, opened 2019-08-02
Include panic logs for vfs in diagnose

### LINKED ISSUE #1395: Mac: `gvfs diagnose` should capture PrjFS kernel panics
`gvfs diagnose` should capture PrjFS kernel panics when it runs.

jeschu1:
resolves #1395

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacPlatform.cs
+++ b/GVFS/GVFS.Platform.Mac/MacPlatform.cs
@@ -15,6 +15,8 @@ namespace GVFS.Platform.Mac
     public partial class MacPlatform : POSIXPlatform
     {
         private const string UpgradeProtectedDataDirectory = "/usr/local/vfsforgit_upgrader";
+        private const string DiagnosticReportsDirectory = "/Library/Logs/DiagnosticReports";
+        private const string PanicFileNamePattern = "*panic";
 
         public MacPlatform() : base(
              underConstruction: new UnderConstructionFlags(
@@ -139,6 +141,48 @@ namespace GVFS.Platform.Mac
             MacDaemonController.DaemonInfo gvfsService = daemons.FirstOrDefault(sc => string.Equals(sc.Name, "org.vfsforgit.service"));
             installed = gvfsService != null;
             running = installed && gvfsService.IsRunning;
+        }
+
+        public override bool TryCopyPanicLogs(string copyToDir, out string error)
+        {
+            error = null;
+            try
+            {
+                if (!Directory.Exists(DiagnosticReportsDirectory))
+                {
+                    return true;
+                }
+
+                string copyToPanicDir = Path.Combine(copyToDir, ProjFSKext.DriverLogDirectory, "panic_logs");
+                Directory.CreateDirectory(copyToPanicDir);
+
+                foreach (string filePath in Directory.GetFiles(DiagnosticReportsDirectory, PanicFileNamePattern))
+                {
+                    try
+                    {
+                        // We only include panic logs caused by our kext
+                        // Panics caused by our kext will be in the form DriverName(Version)
+                        // We match the minimal requirement here
+                        if (File.ReadAllText(filePath).Contains(ProjFSKext.DriverName + "("))
+                        {
+                            File.Copy(filePath, Path.Combine(copyToPanicDir, Path.GetFileName(filePath)));
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        error = error == null ? string.Empty : error + "\n";
+                        error += $"{nameof(this.TryCopyPanicLogs)}: Failed to handle log {filePath}: {ex.ToString()}";
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                error = error == null ? string.Empty : error + "\n";
+                error += $"{nameof(this.TryCopyPanicLogs)}: Failed to copy panic logs: {ex.ToString()}";
+                return false;
+            }
+
+            return error == null;
         }
 
         public class MacPlatformConstants : POSIXPlatformConstants
--- a/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
+++ b/GVFS/GVFS/CommandLine/DiagnoseVerb.cs
@@ -156,6 +156,11 @@ namespace GVFS.CommandLine
                             this.CopyFile(GVFSPlatform.Instance.GetDataRootForGVFS(), archiveFolderPath, LocalGVFSConfig.FileName);
                         }
 
+                        if (!GVFSPlatform.Instance.TryCopyPanicLogs(archiveFolderPath, out string errorMessage))
+                        {
+                            this.WriteMessage(errorMessage);
+                        }
+
                         return true;
                     },
                     "Copying logs");
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -95,6 +95,8 @@ namespace GVFS.Common
 
         public abstract void ConfigureVisualStudio(string gitBinPath, ITracer tracer);
 
+        public abstract bool TryCopyPanicLogs(string copyToDir, out string error);
+
         public abstract bool TryGetGVFSHooksVersion(out string hooksVersion, out string error);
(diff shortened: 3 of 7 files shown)


GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -15,6 +15,8 @@ namespace GVFS.Platform.Mac
     public partial class MacPlatform : POSIXPlatform
     {
         private const string UpgradeProtectedDataDirectory = "/usr/local/vfsforgit_upgrader";
+        private const string DiagnosticReportsDirectory = "/Library/Logs/DiagnosticReports";
+        private const string PanicFileExtention = "*panic";
```
pmj: Nit: The extension is `"panic"`, this constant might be better described as `PanicFileNamePattern` or perhaps [`PanicFileGlob`](https://en.wikipedia.org/wiki/Glob_(programming)). (I think .NET normally calls them "patterns" so probably best to go with that.)

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -141,6 +143,36 @@ public override void IsServiceInstalledAndRunning(string name, out bool installe
...
+                {
+                    return true;
+                }
+
+                string copyToPanicDir = Path.Combine(copyToDir, ProjFSKext.DriverLogDirectory, "panic_logs");
+                Directory.CreateDirectory(copyToPanicDir);
+
+                foreach (string filePath in Directory.GetFiles(DiagnosticReportsDirectory, PanicFileExtention))
+                {
+                    if (File.ReadAllText(filePath).Contains(ProjFSKext.DriverName))
```
pmj: I think this will match any panics for which the ProjFS kext was *loaded*, not just those where it appeared in the stack trace. (There is a list of loaded kexts at the end of each panic log.) Is that the intended behaviour? Either way, it might be worth commenting on the *intention* of this check here.

wilbaker: Nice catch!

> Is that the intended behaviour?

I think the intended behavior is to only include the panics where ProjFS kext appeared in the stack trace.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -156,6 +156,12 @@ protected override void Execute(GVFSEnlistment enlistment)
                             this.CopyFile(GVFSPlatform.Instance.GetDataRootForGVFS(), archiveFolderPath, LocalGVFSConfig.FileName);
                         }
 
+                        // kernel panics
```
wilbaker: Nit: I don't think this comment is adding much, I think it can be removed.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -141,6 +143,36 @@ public override void IsServiceInstalledAndRunning(string name, out bool installe
...
+                    {
+                        File.Copy(filePath, Path.Combine(copyToPanicDir, Path.GetFileName(filePath)));
```
wilbaker: Should we have a try/catch specifically around this call to `File.Copy`?  That way if one copy fails we will can still try to copy any other panics on disk.

kewillford:
I'm wondering if we can remove the `IKernelDriver.LogsFolderPath` and `IKernelDriver.TryFlushLogs` and have one `IKernelDriver.TryCopyLogs` that will do the driver logs
```
// driver
if (this.FlushKernelDriverLogs())
{
	 string kernelLogsFolderPath = GVFSPlatform.Instance.KernelDriver.LogsFolderPath;

	// This copy sometimes fails because the OS has an exclusive lock on the etl files. The error is not actionable
	// for the user so we don't write the error message to stdout, just to our own log file.
	this.CopyAllFiles(Path.GetDirectoryName(kernelLogsFolderPath), archiveFolderPath, Path.GetFileName(kernelLogsFolderPath), copySubFolders: false, hideErrorsFromStdout: true);
}
``` 

plus the panic logs?

I'm not sure how much cleaner this would be or what you would need to pass to the method but I saw that we were already doing some `GVFSPlatform.Instance` specific things here and was wondering if this could fit in that same bucket or if it best to keep them separate.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
jeschu1: I can take it out.  We seem to do this pattern for most copies in the file which is why I did it here.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
jeschu1: I'm doing that, but I think that also begs the question should we fail everything if we can't read a log file (surrounding if).  In either scenario I imagine we'll need panics and will end up asking the user to manually copy.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
jeschu1: Looking at several panics it appears that if we're in the backtrace, we'll have the form `org.vfsforgit.PrjFSKext(`. 

Would you advise that as the simplest way to do this?

jeschu1:
> I'm wondering if we can remove the IKernelDriver.LogsFolderPath and IKernelDriver.TryFlushLogs and have one IKernelDriver.TryCopyLogs that will do the driver logs plus the panic logs?

@kewillford an exellent point and something I certainly though about.  TryFlushLogs has a different purpose and I think putting it all in one fuctions gives that function too much responsibility.  What we have here is pretty explicit (and hopefully easy to understand! :-) )

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
wilbaker: > We seem to do this pattern for most copies in the file which is why I did it here.

Apologies if that's the case and I missed it.  If we're doing it elsewhere it's good with me to do it here too.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -141,6 +143,45 @@ public override void IsServiceInstalledAndRunning(string name, out bool installe
...
+                {
+                    return true;
+                }
+
+                string copyToPanicDir = Path.Combine(copyToDir, ProjFSKext.DriverLogDirectory, "panic_logs");
+                Directory.CreateDirectory(copyToPanicDir);
+
+                foreach (string filePath in Directory.GetFiles(DiagnosticReportsDirectory, PanicFileNamePattern))
+                {
+                    if (File.ReadAllText(filePath).Contains(ProjFSKext.DriverName))
```
wilbaker: If you move the `File.ReadAllText` inside the inner `try` than a failure to read one panic won't stop us from checking the rest.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
jeschu1: @wilbaker, it honestly seemed redundant to me too.  I think it's fine to leave out.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
pmj: We discussed this out-of-band and this does seem like the simplest solution.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
pmj: I think if copying panic logs fails there's something pretty seriously wrong with the machine, so VFS4G diagnostics are going to be the last of the user's worries.

wilbaker: > I'm doing that, but I think that also begs the question should we fail everything if we can't read a log file (surrounding if). In either scenario I imagine we'll need panics and will end up asking the user to manually copy.

I think we should copy what we can but include in the diagnose log if we encountered errors while trying to copy.

jeschu1: @wilbaker  @pmj, logging added.  I agree let's make this as easy to diagnose form our side as possible!

PR #1411, opened 2019-08-05
[Linux] Convert tests and repair jobs to use backing root

chrisd8088:
Convert all relevant uses of `RepoRoot` and `WorkingDirectoryRoot` in the GVFS provider and functional tests to use the backing root, to ensure these tests will pass on systems where they are distinct (e.g., Linux).

For example. repair jobs run while the GVFS enlistment is not mounted, which on Linux requires us to use the backing root in order to access the `.git` contents; similarly, the `MountTests` are adjusted to ensure they don't expect file contents within an unmounted repository.

/cc @jrbriggs, @kivikakk.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
@@ -57,7 +57,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         {
             this.Enlistment.UnmountGVFS();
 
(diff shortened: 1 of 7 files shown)


chrisd8088:
Hi @derrickstolee -- do you want to look this one over as well, or are you comfortable with my merging it into master now that @wilbaker signed off?  Let me know either way, and many thanks!

wilbaker:
> Hi @derrickstolee -- do you want to look this one over as well, or are you comfortable with my merging it into master now that @wilbaker signed off? Let me know either way, and many thanks!

@chrisd8088 feel free to merge this PR in now, thanks!

cc: @derrickstolee

PR #1412, opened 2019-08-05
[Linux] Support case-sensitive filesystems

chrisd8088:
This PR looks to resolve a number of case-sensitivity issues with the current provider code and test suite, including:
1. Fixing case-differing references to files and path, e.g., `FastFetch.dll` not `fastfetch.dll`.  This includes references to Git object files, for which the SHA1 value needs to be lowercased.
2. Fixing hard-coded Windows path separators in messages and tests.
3. Ensuring `DiffHelper` functions correctly on case-sensitive filesystems; among other things, this allows `FastFetch` to succeed.
4. Adding case-sensitivity to the GVFS provider's databases and file path comparisons, so that, for example, the [`GitCommandsTests.CaseOnlyRenameFileAndChangeBranches()`](https://github.com/microsoft/VFSForGit/blob/3f322d3a44375d2108e1b70ee5329c6bfc565ec7/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs#L398) functional test succeeds.

The latter's failure on a case-sensitive filesystem is due to case-insensitivity in the current [modified paths database](https://github.com/microsoft/VFSForGit/blob/3f322d3a44375d2108e1b70ee5329c6bfc565ec7/GVFS/GVFS.Common/ModifiedPathsDatabase.cs#L21) and [virtualization code](https://github.com/microsoft/VFSForGit/blob/3f322d3a44375d2108e1b70ee5329c6bfc565ec7/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.SortedFolderEntries.cs#L180), among other things, which results in case-differing-only filenames being dropped from the data sent back from the GVFS provider, via the [`virtual-filesystem` hook](https://github.com/microsoft/git/blob/9b1adf3543bd5831c925fba1c4a6078a83118dfd/read-cache.c#L1887), to Git itself after it reads its `index` file.

---

Introduce case-sensitive file path matching and sorting of file and folder names within the repository as well as the modified paths and placeholder list databases, but only on Linux and other case-sensitive file systems, while retaining existing case-insensitive behaviour on Windows
and Mac platforms.

Use filesystem-specific case matching when comparing path throughout the functional test suite. 
 Also make sure to exclude the case-sensitive filesystem unit tests when running on Windows or Mac platforms.

Also fix the path matching in the `HydrateEntireRepo()` helper for `MultiEnlistmentTests.SharedCacheTests`, which was not excluding `{repoRoot}/.git/` paths on Mac (or Linux) due to hard-coded Windows path separators.

Divide the `HydratingFileUsesNameCaseFromRepo()` and `HydratingNestedFileUsesNameCaseFromRepo()` functional tests
into pairs, one for case-sensitive filesystems, and one for case-insensitive filesystems.

Includes some filename case corrections to prevent test case regressions on case-sensitive filesystems, use of lowercase paths generated from SHA1 values because that ensures consistency with Git's own naming scheme for files under `.git/objects` on case-sensitive filesystems, and fixes for some hard-coded uses of Windows path separators.

Resolves github#29.

/cc @jrbriggs, @kivikakk.

### DIFF
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -17,11 +17,12 @@ namespace GVFS.Common.Database
             this.connectionPool = connectionPool;
         }
 
-        public static void CreateTable(IDbConnection connection)
+        public static void CreateTable(IDbConnection connection, bool caseSensitiveFileSystem)
         {
             using (IDbCommand command = connection.CreateCommand())
             {
-                command.CommandText = "CREATE TABLE IF NOT EXISTS [Placeholder] (path TEXT PRIMARY KEY COLLATE NOCASE, pathType TINYINT NOT NULL, sha char(40) ) WITHOUT ROWID;";
+                string collateConstraint = caseSensitiveFileSystem ? string.Empty : " COLLATE NOCASE";
+                command.CommandText = $"CREATE TABLE IF NOT EXISTS [Placeholder] (path TEXT PRIMARY KEY{collateConstraint}, pathType TINYINT NOT NULL, sha char(40) ) WITHOUT ROWID;";
                 command.ExecuteNonQuery();
             }
         }
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -594,6 +594,8 @@ namespace GVFS.Common.Git
 
         private LooseObjectToWrite GetLooseObjectDestination(string sha)
         {
+            // Ensure SHA path is lowercase for case-sensitive filesystems
+            sha = sha.ToLower();
             string firstTwoDigits = sha.Substring(0, 2);
             string remainingDigits = sha.Substring(2);
             string twoLetterFolderName = Path.Combine(this.Enlistment.GitObjectsRoot, firstTwoDigits);
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -595,7 +595,11 @@ namespace GVFS.Common.Git
         private LooseObjectToWrite GetLooseObjectDestination(string sha)
         {
             // Ensure SHA path is lowercase for case-sensitive filesystems
-            sha = sha.ToLower();
+            if (GVFSPlatform.Instance.Constants.CaseSensitiveFileSystem)
+            {
+                sha = sha.ToLower();
+            }
+
             string firstTwoDigits = sha.Substring(0, 2);
             string remainingDigits = sha.Substring(2);
             string twoLetterFolderName = Path.Combine(this.Enlistment.GitObjectsRoot, firstTwoDigits);
--- a/GVFS/GVFS.Common/Git/GitRepo.cs
+++ b/GVFS/GVFS.Common/Git/GitRepo.cs
@@ -232,6 +232,8 @@ namespace GVFS.Common.Git
 
         private LooseBlobState GetLooseBlobState(string blobSha, Action<Stream, long> writeAction, out long size)
         {
+            // Ensure SHA path is lowercase for case-sensitive filesystems
+            blobSha = blobSha.ToLower();
             string blobPath = Path.Combine(
                 this.enlistment.GitObjectsRoot,
                 blobSha.Substring(0, 2),
--- a/GVFS/GVFS.Common/Git/GitRepo.cs
+++ b/GVFS/GVFS.Common/Git/GitRepo.cs
@@ -233,7 +233,11 @@ namespace GVFS.Common.Git
         private LooseBlobState GetLooseBlobState(string blobSha, Action<Stream, long> writeAction, out long size)
         {
             // Ensure SHA path is lowercase for case-sensitive filesystems
-            blobSha = blobSha.ToLower();
+            if (GVFSPlatform.Instance.Constants.CaseSensitiveFileSystem)
+            {
+                blobSha = blobSha.ToLower();
+            }
+
             string blobPath = Path.Combine(
                 this.enlistment.GitObjectsRoot,
                 blobSha.Substring(0, 2),
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
@@ -88,9 +88,9 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
                 {
                     "# A comment",
                     " ",
-                    "gvfs/",
-                    "gvfs/gvfs",
-                    "gvfs/"
+                    "GVFS/",
+                    "GVFS/GVFS",
+                    "GVFS/"
                 });
 
             this.ExpectBlobCount(this.Enlistment.Prefetch("--folders-list \"" + tempFilePath + "\""), 279);
@@ -166,7 +166,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
                 new[]
                 {
                     Path.Combine("GVFS", "GVFS", "packages.config"),
-                    Path.Combine("GVFS", "GVFS.FunctionalTests", "App.config")
+                    Path.Combine("GVFS", "GVFS.FunctionalTests", "app.config")
                 });
 
             this.ExpectBlobCount(this.Enlistment.Prefetch("--stdin-files-list", standardInput: input), 2);
@@ -181,9 +181,9 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
                 {
                     "# A comment",
                     " ",
-                    "gvfs/",
-                    "gvfs/gvfs",
-                    "gvfs/"
+                    "GVFS/",
+                    "GVFS/GVFS",
+                    "GVFS/"
                 });
 
             this.ExpectBlobCount(this.Enlistment.Prefetch("--stdin-folders-list", standardInput: input), 279);
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
@@ -127,7 +127,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.fileSystem
                 .EnumerateDirectory(this.Enlistment.GetPackRoot(this.fileSystem))
                 .Split()
-                .Where(file => string.Equals(Path.GetExtension(file), ".keep", StringComparison.OrdinalIgnoreCase))
+                .Where(file => string.Equals(Path.GetExtension(file), ".keep", FileSystemHelpers.PathComparison))
                 .Count()
                 .ShouldEqual(1, "Incorrect number of .keep files in pack directory");
 
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
@@ -386,30 +386,22 @@ BOOL APIENTRY DllMain( HMODULE hModule,
             this.PlaceholderHasVersionInfo(virtualFilePath, CurrentPlaceholderVersion, string.Empty).ShouldEqual(true);
         }
 
+        // Case-insensitive filesystems only
         [TestCase, Order(13)]
         [Category(Categories.GitCommands)]
+        [Category(Categories.CaseInsensitiveFileSystemOnly)]
         [Category(Categories.MacTODO.NeedsNewFolderCreateNotification)]
-        public void FolderContentsProjectedAfterFolderCreateAndCheckout()
+        public void FolderContentsProjectedAfterFolderCreateAndCheckoutOnCaseInsensitiveFileSystem()
         {
-            string folderName = "GVFlt_MultiThreadTest";
+            this.FolderContentsProjectedAfterFolderCreateAndCheckout(false);
+        }
 
-            // 54ea499de78eafb4dfd30b90e0bd4bcec26c4349 did not have the folder GVFlt_MultiThreadTest
-            GitProcess.InvokeProcess(this.Enlistment.RepoRoot, "checkout 54ea499de78eafb4dfd30b90e0bd4bcec26c4349");
-
-            // Confirm that no other test has created GVFlt_MultiThreadTest or put it in the modified files
-            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, folderName);
-
-            string virtualFolderPath = this.Enlistment.GetVirtualPathTo(folderName);
-            virtualFolderPath.ShouldNotExistOnDisk(this.fileSystem);
-            this.fileSystem.CreateDirectory(virtualFolderPath);
-
-            // b3ddcf43b997cba3fbf9d2341b297e22bf48601a was the commit prior to deleting GVFLT_MultiThreadTest
-            // 692765: Note that test also validates case insensitivity as GVFlt_MultiThreadTest is named GVFLT_MultiThreadTest
-            //         in this commit
-            GitProcess.InvokeProcess(this.Enlistment.RepoRoot, "checkout b3ddcf43b997cba3fbf9d2341b297e22bf48601a");
-
-            this.Enlistment.GetVirtualPathTo(folderName + "\\OpenForReadsSameTime\\test").ShouldBeAFile(this.fileSystem).WithContents("123 \r\n");
-            this.Enlistment.GetVirtualPathTo(folderName + "\\OpenForWritesSameTime\\test").ShouldBeAFile(this.fileSystem).WithContents("123 \r\n");
+        // Case-sensitive filesystems only
+        [TestCase, Order(13)]
+        [Category(Categories.CaseSensitiveFileSystemOnly)]
+        public void FolderContentsProjectedAfterFolderCreateAndCheckoutOnCaseSensitiveFileSystem()
+        {
+            this.FolderContentsProjectedAfterFolderCreateAndCheckout(true);
         }
 
         [TestCase, Order(14)]
@@ -687,6 +679,30 @@ BOOL APIENTRY DllMain( HMODULE hModule,
                 Assert.Fail("Unsupported platform");
                 return false;
             }
+        }
+
+        private void FolderContentsProjectedAfterFolderCreateAndCheckout(bool caseSensitiveFileSystem)
+        {
+            string folderName = "GVFlt_MultiThreadTest";
+
+            // 54ea499de78eafb4dfd30b90e0bd4bcec26c4349 did not have the folder GVFlt_MultiThreadTest
+            GitProcess.InvokeProcess(this.Enlistment.RepoRoot, "checkout 54ea499de78eafb4dfd30b90e0bd4bcec26c4349");
+
+            // Confirm that no other test has created GVFlt_MultiThreadTest or put it in the modified files
+            GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, folderName);
+
+            string virtualFolderPath = this.Enlistment.GetVirtualPathTo(folderName);
+            virtualFolderPath.ShouldNotExistOnDisk(this.fileSystem);
+            this.fileSystem.CreateDirectory(virtualFolderPath);
+
+            // b3ddcf43b997cba3fbf9d2341b297e22bf48601a was the commit prior to deleting GVFLT_MultiThreadTest
+            // 692765: Note that test also validates case insensitivity as GVFlt_MultiThreadTest is named GVFLT_MultiThreadTest
+            //         in this commit; on case-sensitive filesystems, case sensitivity is validated instead
+            GitProcess.InvokeProcess(this.Enlistment.RepoRoot, "checkout b3ddcf43b997cba3fbf9d2341b297e22bf48601a");
+
+            string testFolderName = caseSensitiveFileSystem ? "GVFLT_MultiThreadTest" : folderName;
+            this.Enlistment.GetVirtualPathTo(Path.Combine(testFolderName, "OpenForReadsSameTime", "test")).ShouldBeAFile(this.fileSystem).WithContents("123 \r\n");
+            this.Enlistment.GetVirtualPathTo(Path.Combine(testFolderName, "OpenForWritesSameTime", "test")).ShouldBeAFile(this.fileSystem).WithContents("123 \r\n");
         }
 
         private class NativeTests
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
@@ -303,45 +303,36 @@ BOOL APIENTRY DllMain( HMODULE hModule,
             folderVirtualPath.ShouldBeADirectory(this.fileSystem).WithItems();
         }
 
+        // Case-insensitive filesystems only
         [TestCase, Order(7)]
-        public void HydratingFileUsesNameCaseFromRepo()
+        [Category(Categories.CaseInsensitiveFileSystemOnly)]
+        public void HydratingFileUsesNameCaseFromRepoOnCaseInsensitiveFileSystem()
         {
-            string fileName = "Readme.md";
-            string parentFolderPath = this.Enlistment.GetVirtualPathTo(Path.GetDirectoryName(fileName));
-            parentFolderPath.ShouldBeADirectory(this.fileSystem).WithItems().ShouldContainSingle(info => info.Name.Equals(fileName, StringComparison.Ordinal));
-
-            // Hydrate file with a request using different file name case
-            string wrongCaseFilePath = this.Enlistment.GetVirtualPathTo(fileName.ToUpper());
-            string fileContents = wrongCaseFilePath.ShouldBeAFile(this.fileSystem).WithContents();
-
-            // File on disk should have original case projected from repo
-            parentFolderPath.ShouldBeADirectory(this.fileSystem).WithItems().ShouldContainSingle(info => info.Name.Equals(fileName, StringComparison.Ordinal));
+            this.HydratingFileUsesNameCaseFromRepo(false);
         }
 
-        [TestCase, Order(8)]
-        public void HydratingNestedFileUsesNameCaseFromRepo()
+        // Case-sensitive filesystems only
+        [TestCase, Order(7)]
+        [Category(Categories.CaseSensitiveFileSystemOnly)]
+        public void HydratingFileUsesNameCaseFromRepoOnCaseSensitiveFileSystem()
         {
-            string filePath = Path.Combine("GVFS", "FastFetch", "Properties", "AssemblyInfo.cs");
-            string filePathAllCaps = filePath.ToUpper();
-            string parentFolderVirtualPathAllCaps = this.Enlistment.GetVirtualPathTo(Path.GetDirectoryName(filePathAllCaps));
-            parentFolderVirtualPathAllCaps.ShouldBeADirectory(this.fileSystem).WithItems().ShouldContainSingle(info => info.Name.Equals(Path.GetFileName(filePath), StringComparison.Ordinal));
+            this.HydratingFileUsesNameCaseFromRepo(true);
+        }
 
-            // Hydrate file with a request using different file name case
-            string wrongCaseFilePath = this.Enlistment.GetVirtualPathTo(filePathAllCaps);
-            string fileContents = wrongCaseFilePath.ShouldBeAFile(this.fileSystem).WithContents();
+        // Case-insensitive filesystems only
+        [TestCase, Order(8)]
+        [Category(Categories.CaseInsensitiveFileSystemOnly)]
+        public void HydratingNestedFileUsesNameCaseFromRepoOnCaseInsensitiveFileSystem()
+        {
+            this.HydratingNestedFileUsesNameCaseFromRepo(false);
+        }
 
-            // File on disk should have original case projected from repo
-            string parentFolderVirtualPath = this.Enlistment.GetVirtualPathTo(Path.GetDirectoryName(filePath));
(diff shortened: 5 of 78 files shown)


GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -170,6 +170,8 @@ public override string GVFSBinDirectoryName
 
             // Documented here (in the addressing section): https://www.unix.com/man-page/mojave/4/unix/
             public override int MaxPipePathLength => 104;
+
+            public override bool CaseSensitiveFileSystem => false;
```
turbonaitis: This is not exactly true. One can have Mac with a case-sensitive file system too. (see https://apple.stackexchange.com/questions/71357/how-to-check-if-my-hd-is-case-sensitive-or-not) 
Perhaps at clone time, VFS4G could establish this? Also, even on windows and linux, there are cases, where case-sensitivity might not match the default, see cases listed here: https://github.com/dotnet/corefx/issues/3156#issuecomment-171455461

wilbaker: > One can have Mac with a case-sensitive file system too.

As best I could tell .NET Core has a mix of assumptions when it comes to Mac being case sensitive, which would make it difficult for us to support Mac file systems that are configured to be case sensitive.

https://github.com/dotnet/corefx/blob/e56e355d59380d7038040d57c01418f4950e47d0/src/Common/src/CoreLib/System/IO/Path.Unix.cs#L135
```
        /// <summary>Gets whether the system is case-sensitive.</summary>
        internal static bool IsCaseSensitive
        {
            get
            {
                #if PLATFORM_OSX
                    return false;
                #else
                    return true;
                #endif
            }
        }
```

https://github.com/dotnet/corefx/blob/e56e355d59380d7038040d57c01418f4950e47d0/src/Common/tests/Tests/System/IO/PathInternal.Unix.Tests.cs#L47
```
        [Fact]
        [PlatformSpecific(TestPlatforms.OSX)]
        public void IsCaseInsensitive_OSX()
        {
            // There have been reports of casing handling not being appropriate on MacOS
            // https://github.com/dotnet/corefx/issues/26797
            Assert.False(PathInternal.IsCaseSensitive);
        }
```

https://github.com/dotnet/corefx/blob/1bff7880bfa949e8c5e46039808ec412640bbb5e/src/Common/src/System/IO/PathInternal.CaseSensitivity.cs#L41

```
        /// <summary>
        /// Determines whether the file system is case sensitive.
        /// </summary>
        /// <remarks>
        /// Ideally we'd use something like pathconf with _PC_CASE_SENSITIVE, but that is non-portable, 
        /// not supported on Windows or Linux, etc. For now, this function creates a tmp file with capital letters 
        /// and then tests for its existence with lower-case letters.  This could return invalid results in corner 
        /// cases where, for example, different file systems are mounted with differing sensitivities.
        /// </remarks>
        private static bool GetIsCaseSensitive()
        {
            try
            {
                string pathWithUpperCase = Path.Combine(Path.GetTempPath(), "CASESENSITIVETEST" + Guid.NewGuid().ToString("N"));
                using (new FileStream(pathWithUpperCase, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.None, 0x1000, FileOptions.DeleteOnClose))
                {
                    string lowerCased = pathWithUpperCase.ToLowerInvariant();
                    return !File.Exists(lowerCased);
                }
            }
            catch (Exception exc)
            {
                // In case something goes terribly wrong, we don't want to fail just because
                // of a casing test, so we assume case-insensitive-but-preserving.
                Debug.Fail("Casing test failed: " + exc);
                return false;
            }
        }
```
To prevent someone from getting in a bad state #1384 updates the clone/mount verbs to fail if someone tries to use them on a Mac case sensitive filesystem.

chrisd8088: I was going to make the same observation that #1384 prevents mounting on case-sensitive Mac filesystems -- at least for now.

turbonaitis: Ok, if there's inconsistent support for case sensitive Mac file system in dotnet core itself, then it makes sense to not try to support them here.

chrisd8088:
Just to expand on our conversation today, I hope to add some further refinements in the non-common, platform-specific code (e.g., `GVFS.Platform.Mac`, etc.), which I think would help reduce the likelihood of re-introducing case-sensitivity issues in the common code.  The PR at the moment is the base set of changes needed, but there are a few more `OrdinalIgnoreCase` lurking around in some of the platform-specific classes.

chrisd8088:
Looks like the sparse mode changes have added a single new test failure ... will investigate today.

chrisd8088:
@wilbaker, @derrickstolee, @kewillford -- OK, I think I've resolved all outstanding "extras" (e.g., non-critical changes from `OrdinalIgnoreCase` in various places to ensure code consistency, along with a few other consistency fixups), as well as the issues from the sparse-checkout test.

That particular failure was due, I'm fairly sure, to a bit of a mismatch in the existing functional test suite.  The test in question is [GitCommands.DeleteFolderAndChangeBranchToFolderWithDifferentCase()](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs#L710), which deletes the directory `GVFlt_MultiThreadTest` and then checks out a commit which has the `GVFLT_MultiThreadTest` directory instead.  The `TestValidationAndCleanup()` cleanup then [checks for matching directory structure](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L169) against the control repo.  In all cases, the [`ignoreCase`](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L161) flag is set to `false`, i.e, be case-sensitive, even on Windows and Mac.  This flag is passed down to [`CompareDirectories()`](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L161).

The test succeeds right now on master, of course, but I believe it does so for a somewhat surprising reason.  The `CompareDirectories()` method [filters out](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs#L311) directories in the control repo which it doesn't care about testing using `IsMatchedPath()`, but that method is [always case-insensitive](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs#L280), regardless of the caller's `ignoreCase` flag.  So one might expect that it would incorrectly filter out some directories to compare, or include some it shouldn't.

In the specific case of the `DeleteFolderAndChangeBranchToFolderWithDifferentCase()` test, though, the test still happens to pass because the `GVFLT_MultiThreadTest` directory matches (case-insensitively) the `GVFlt_MultiThreadTest` path prefix, which appears in the [`SparseModeFolders`](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L29) list in `GitCommands/GitRepoTests.cs` list.

Sure enough, when I [adjusted the `IsMatchedPath()` filter method](https://github.com/github/VFSForGit/blob/b536003fa68806b5a743feecfca2e47522e83936/GVFS/GVFS.FunctionalTests/Should/FileSystemShouldExtensions.cs#L263) to inherit the `ignoreCase` flag from its `CompareDirectories()` caller -- which I think it reasonable behaviour, so you don't have case-insensitivity "hiding" inside a normally case-sensitive test -- that's when the test then began to fail.

Adding the [`GVFLT_MultiThreadTest` prefix](https://github.com/github/VFSForGit/blob/b536003fa68806b5a743feecfca2e47522e83936/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L47) to the `SparseModeFolders` list was what was required to get things passing again.  And on Linux, this was also required, but for a different reason; without that path in the list, `/GVFLT_MultiThreadTest` was excluded from the sparse set of paths to be projected, so the test [failed earlier](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs#L726) when the that directory was missing after the checkout.

GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -16,11 +16,12 @@ public PlaceholderTable(IGVFSConnectionPool connectionPool)
             this.connectionPool = connectionPool;
         }
 
-        public static void CreateTable(IDbConnection connection)
+        public static void CreateTable(IDbConnection connection, bool caseSensitiveFileSystem)
         {
             using (IDbCommand command = connection.CreateCommand())
             {
-                command.CommandText = "CREATE TABLE IF NOT EXISTS [Placeholder] (path TEXT PRIMARY KEY COLLATE NOCASE, pathType TINYINT NOT NULL, sha char(40) ) WITHOUT ROWID;";
+                string collateConstraint = caseSensitiveFileSystem ? string.Empty : " COLLATE NOCASE";
```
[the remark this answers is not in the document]
kewillford: SQLite default is case sensitive so this looks good to me.

GVFS/GVFS/RepairJobs/GitConfigRepairJob.cs
```diff
@@ -16,7 +16,7 @@ public GitConfigRepairJob(ITracer tracer, TextWriter output, GVFSEnlistment enli
 
         public override string Name
         {
-            get { return ".git\\config"; }
+            get { return $".git{Path.DirectorySeparatorChar}config"; }
```
kewillford: These changes don't seem like they are related to this PR.  I understand why you are making them but can they be in a separate PR.  Also they should be using the `GVFSConstants.DotGit` properties I would think.

chrisd8088: I've pulled these out and tidied them up further in PR #1448 -- thanks for the advice!

chrisd8088:
Thanks for the review, @kewillford!

I pulled out the directory separator fixes into a new PR #1448.

I've also updated the functional tests (and `CmdRunner`) as you suggested, which is much nicer, and added another case-sensitive test parallel to the [unit test](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs#L105) you mentioned.  Thank you!

I think the [`MockDirectory`](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.UnitTests/Mock/FileSystem/MockDirectory.cs#L123) instances you mention are OK as is, because as I understand it, the whole class is designed that way, and in particular has a [pair of hashes](https://github.com/microsoft/VFSForGit/blob/bf53ae9a1df4d71e31c71cc2dab1a223d1972c0a/GVFS/GVFS.UnitTests/Mock/FileSystem/MockDirectory.cs#L17) which store the mock directory and filenames, and those are also `InvariantCultureIgnoreCase`.  I don't see any issue with this, and it works fine as-is on Linux.

GVFS/GVFS.FunctionalTests/Tests/FastFetchTests.cs
```diff
@@ -321,8 +321,9 @@ public void SuccessfullyChecksOutCaseChanges()
 
             try
             {
+                // Ignore case differences on case-insensitive filesystems
                 this.fastFetchRepoRoot.ShouldBeADirectory(FileSystemRunner.DefaultRunner)
-                    .WithDeepStructure(FileSystemRunner.DefaultRunner, this.fastFetchControlRoot, ignoreCase: true);
+                    .WithDeepStructure(FileSystemRunner.DefaultRunner, this.fastFetchControlRoot, ignoreCase: !RuntimeInformation.IsOSPlatform(OSPlatform.Linux));
```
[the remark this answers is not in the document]
chrisd8088: Yes, done!

GVFS/GVFS.FunctionalTests/Tests/GitCommands/UpdateRefTests.cs
```diff
@@ -29,11 +30,19 @@ public void UpdateRefModifiesHeadThenResets()
 
         public override void TearDownForTest()
         {
-            // Need to ignore case changes in this test because the update-ref will have
-            // folder names that only changed the case and when checking the folder structure
-            // it will create partial folders with that case and will not get updated to the
-            // previous case when the reset --hard running in the tear down step
-            this.TestValidationAndCleanup(ignoreCase: true);
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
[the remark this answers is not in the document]
chrisd8088: I made `FileSystemHelpers.CaseSensitiveFileSystem` public so we could use it here -- thanks for the idea!

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -102,6 +103,7 @@ public void CanParseBackwardsDiff()
         // Delete a folder with two sub folders each with a single file
         // Readd it with a different casing and same contents
         [TestCase]
+        [Category(CategoryConstants.CaseInsensitiveFileSystemOnly)]
         public void ParsesCaseChangesAsAdds()
```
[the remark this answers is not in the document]
chrisd8088: I've written one and it passes on Linux, so I think we're good here -- thanks for the suggestion!

GVFS/GVFS.UnitTests/Virtualization/Projection/SortedFolderEntriesTests.cs
```diff
@@ -116,6 +129,21 @@ public void ValidateOrderOfDefaultEntries()
...
         }
 
+        [TestCase]
+        [Category(CategoryConstants.CaseSensitiveFileSystemOnly)]
+        public void ValidateCaseSensitiveOrderOfDefaultEntries()
+        {
+            List<string> allEntries = new List<string>(defaultFiles);
+            allEntries.AddRange(defaultFolders);
+            allEntries.Sort(CaseSensitiveStringCompare);
+            SortedFolderEntries sfe = SetupDefaultEntries();
```
kewillford: I'm wondering if we need to add some entries to the defaults that have different casing to make sure the sorting is happening correctly?  Like have a `Folder1` and `File1.txt`?  I think that could go in the default values because that would be good to have for the case insensitive tests as well.  Not sure if we would need the same entry that only differed in case?

chrisd8088: This turned out to be a little [more involved](https://github.com/microsoft/VFSForGit/pull/1412/commits/4435b1d03ed9a622fe80ca6ca4bfcdd53f0d61a3) than we might have hoped because [`GitIndexProjection.SortedFolderEntries.GetInsertionIndex()`](https://github.com/microsoft/VFSForGit/blob/9c733e5dd30b62d734eab6489d0410a17072b0c3/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.SortedFolderEntries.cs#L165) does [platform-specific indexing](https://github.com/github/VFSForGit/blob/4435b1d03ed9a622fe80ca6ca4bfcdd53f0d61a3/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.SortedFolderEntries.cs#L213) (as of this PR), so it throws an exception on Mac/Windows if their are case-differing-only entries in `defaultFiles` and `defaultFolders`.  Hence this more full-bodied approach.

chrisd8088:
OK -- @kewillford, I added the additional test coverage you suggested (modulo the fact that it requires a case-sensitive platform).  If that looks OK to you, and if @wilbaker is OK also, I'd love to merge ... let me know what you think!  And thanks so much for the feedback!

GVFS/GVFS.Common/Git/GitRepo.cs
```diff
@@ -232,6 +232,8 @@ private LooseBlobState GetLooseBlobStateAtPath(string blobPath, Action<Stream, l
 
         private LooseBlobState GetLooseBlobState(string blobSha, Action<Stream, long> writeAction, out long size)
         {
+            // Ensure SHA path is lowercase for case-sensitive filesystems
+            blobSha = blobSha.ToLower();
```
wilbaker: @chrisd8088 have you seen problems with `GetLooseBlobState` on Linux?

Rather than paying the cost of `ToLower` every time this method is called can you make sure that the source of the data (e.g. `Sha1Id.ToString()`) is always lowercase if it's not already?

If this is not feasible, can we check `CaseSensitiveFileSystem` before calling `ToLower()`?

GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -594,6 +594,8 @@ private Stream OpenTempLooseObjectStream(string path)
 
         private LooseObjectToWrite GetLooseObjectDestination(string sha)
         {
+            // Ensure SHA path is lowercase for case-sensitive filesystems
+            sha = sha.ToLower();
```
wilbaker: Same comment here as in `GitRepo. GetLooseBlobState`, can we avoid always calling `ToLower()` and make sure the caller is passing in the correct casing?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.LazyUTF8String.cs
```diff
@@ -138,6 +138,11 @@ public unsafe int CaseInsensitiveCompare(LazyUTF8String other)
                         byte thisC = *thisPtr;
                         byte otherC = *otherPtr;
 
+                        if (caseSensitive)
```
wilbaker: @chrisd8088 has the impact of this change on index parsing performance been measured?  From what I remember this path is hot enough that Saeed was able to get measurable perf improvements with every conditional check he was able to remove.

`GVFS.PerfProfiling` can be used to isolate/measure the performance of that code path, and we should verify this code path isn't impacted by the change.  @chrisd8088 let me know if you have a Windows box to test this, if not I can do a few runs for you with and without this change.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.LazyUTF8String.cs
```diff
@@ -183,11 +188,26 @@ public unsafe int CaseInsensitiveCompare(LazyUTF8String other)
...
             }
 
+            public unsafe int CaseInsensitiveCompare(LazyUTF8String other)
+            {
+                return this.Compare(other, false);
+            }
+
+            public unsafe int CaseSensitiveCompare(LazyUTF8String other)
+            {
+                return this.Compare(other, true);
```
wilbaker: Style nit:  Here and above can you use a named parameter for the bools?

*Example*

`return this.Compare(other, caseSensitive: true);`

chrisd8088: Definitely -- will do in next set of updates!

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.LazyUTF8String.cs
chrisd8088: Thanks for the attentive review, @wilbaker -- performance is a great question!  I think I can just hoist this check right out of the loop, now that I look more closely.  Let me do that, and see what I can do about the SHA casing as well, and then ping you for a Windows performance test run!  Thanks!

wilbaker: @chrisd8088 I didn't see a different in the time to rebuild the projection with the latest changes and `GVFS.PerfProfiling`.  Thanks!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
```diff
@@ -88,9 +88,9 @@ public void PrefetchFolderListFromFile()
                 {
                     "# A comment",
                     " ",
-                    "gvfs/",
-                    "gvfs/gvfs",
-                    "gvfs/"
+                    "GVFS/",
+                    "GVFS/GVFS",
+                    "GVFS/"
```
wilbaker: @derrickstolee @kewillford @chrisd8088 I'm thinking we might want to parameterize test data like this so that on case sensitive file system we use matching case (e.g. "GVFS") and on case insensitive file systems we use both cases ("gvfs" and "GVFS").

With the current set of changes we're losing validation that case-insensitive path matching works as expected on Mac and Windows.

Thoughts?

(This comment applies to all of the tests whose file paths were updated to match what's in the repo)

GVFS/GVFS.UnitTests/Common/Database/TableTests.cs
```diff
@@ -33,7 +33,7 @@ public void CreateTableTest()
             Mock<IDbConnection> mockConnection = new Mock<IDbConnection>(MockBehavior.Strict);
             mockConnection.Setup(x => x.CreateCommand()).Returns(mockCommand.Object);
 
-            this.CreateTable(mockConnection.Object);
+            this.CreateTable(mockConnection.Object, false);
```
wilbaker: Style nit here and on line 53: `false` -> `caseSensitiveFileSystem: false`

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -44,6 +44,7 @@ public abstract class GitRepoTests
             "foo.cpp", // Changes to a folder in one test
             "FilenameEncoding",
             "GitCommandsTests",
+            "GVFLT_MultiThreadTest",
```
wilbaker: Why was this folder added to the list?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
```diff
@@ -303,45 +303,36 @@ public void EnumerateAndReadDoesNotChangeEnumerationOrder()
...
-            string fileContents = wrongCaseFilePath.ShouldBeAFile(this.fileSystem).WithContents();
+            this.HydratingFileUsesNameCaseFromRepo(false);
+        }
 
-            // File on disk should have original case projected from repo
-            parentFolderPath.ShouldBeADirectory(this.fileSystem).WithItems().ShouldContainSingle(info => info.Name.Equals(fileName, StringComparison.Ordinal));
+        // Case-sensitive filesystems only
+        [TestCase, Order(7)]
+        [Category(Categories.CaseSensitiveFileSystemOnly)]
+        public void HydratingFileUsesNameCaseFromRepoOnCaseSensitiveFileSystem()
```
wilbaker: Rather than having two separate methods here tagged as `CaseInsensitiveFileSystemOnly` and `CaseSensitiveFileSystemOnly` could we go back to having a single method that would check if the current filesystem is case-sensitive?

In other words, there could be a single `HydratingFileUsesNameCaseFromRepo` test with a line like this

`string testFileName = FileSystemHelpers.CaseSensitiveFileSystem ? fileName : fileName.ToUpper();`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
```diff
@@ -303,45 +303,36 @@ public void EnumerateAndReadDoesNotChangeEnumerationOrder()
...
-            parentFolderVirtualPathAllCaps.ShouldBeADirectory(this.fileSystem).WithItems().ShouldContainSingle(info => info.Name.Equals(Path.GetFileName(filePath), StringComparison.Ordinal));
-
-            // Hydrate file with a request using different file name case
-            string wrongCaseFilePath = this.Enlistment.GetVirtualPathTo(filePathAllCaps);
-            string fileContents = wrongCaseFilePath.ShouldBeAFile(this.fileSystem).WithContents();
-
-            // File on disk should have original case projected from repo
-            string parentFolderVirtualPath = this.Enlistment.GetVirtualPathTo(Path.GetDirectoryName(filePath));
-            parentFolderVirtualPath.ShouldBeADirectory(this.fileSystem).WithItems().ShouldContainSingle(info => info.Name.Equals(Path.GetFileName(filePath), StringComparison.Ordinal));
+            this.HydratingNestedFileUsesNameCaseFromRepo(false);
```
wilbaker: Same comment here, single shared test that uses `FileSystemHelpers.CaseSensitiveFileSystem` rather than `[Category(Categories.CaseXXX)]` to drive behavior.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
```diff
@@ -451,19 +434,20 @@ public void FolderContentsCorrectAfterCreateNewFolderRenameAndCheckoutCommitWith
...
+        }
 
-            folderVirtualPath.ShouldBeADirectory(this.fileSystem).WithNoItems("test*");
-            folderVirtualPath.ShouldBeADirectory(this.fileSystem).WithNoItems("ريلٌأكتوب.TXT");
+        // Case-sensitive filesystems only
+        [TestCase, Order(15)]
+        [Category(Categories.CaseSensitiveFileSystemOnly)]
+        public void FilterNonUTF8FileNameOnCaseSensitiveFileSystem()
+        {
+            this.FilterNonUTF8FileName(true);
```
wilbaker: Same comment for these new test cases, could we use FileSystemHelpers.CaseSensitiveFileSystem instead of picking one test or another based on `Category`.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
chrisd8088: That allows the `TestValidationAndCleanup()` method to succeed on case-sensitive systems, because as memory serves, [`DeleteFolderAndChangeBranchToFolderWithDifferentCase()`](https://github.com/microsoft/VFSForGit/blob/e534dfc518249044b5edbb6de060a323310e8a0b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs#L742) leaves behind a top-level folder of that name, and the test cleanup only succeeds if it matches.

Specifically, [`TestValidationAndCleanup()`](https://github.com/microsoft/VFSForGit/blob/e534dfc518249044b5edbb6de060a323310e8a0b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs#L184) -- when running in sparse mode only -- passes the list of `SparseModeFolders` to `WithDeepStructure()`.  In this PR we also adjust

GVFS/GVFS.Common/Git/GitRepo.cs
chrisd8088: I'm just doing the latter (checking `CaseSensitiveFileSystem` before `ToLower()`) for the moment, although I agree it would be nice to track down all sources of uppercased SHA1s.  With luck, it's always from `Sha1Id`, but I'm not 100% sure of that yet.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
chrisd8088: Much better!  Thank you!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
chrisd8088: Done!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorkingDirectoryTests.cs
chrisd8088: Also done.  :-)

GVFS/GVFS.UnitTests/Common/Database/TableTests.cs
chrisd8088: Fixed, thanks!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
chrisd8088: I think I've updated this PR with all of your suggestions below, @wilbaker -- the exception is this one.  I could also just add some `FileSystemHelper.CaseSensitiveFileSystem ? ... : ...` code here as well, if you think that's better.  Let me know, or I'll try to tackle it early tomorrow, if time permits, and otherwise in a week or so.

Thanks again for all the review; much appreciated!

GVFS/GVFS.Common/Git/GitRepo.cs
wilbaker: Thanks! I'm marking this comment as resolved.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchVerbTests.cs
wilbaker: > I could also just add some FileSystemHelper.CaseSensitiveFileSystem ? ... : ... code here as well, if you think that's better.

I think that would work well.  

If you could leave some of these entries lower-case on case-insensitive file systems then I think this PR is ready to go.

chrisd8088: OK -- as we discussed just now, I'll put those changes into a followup PR next week, and maybe try to track down the source of the uppercase SHA1s, if I can, in another.  But I'm going to go ahead and merge this now -- thank you for all your support!

chrisd8088:
As we discussed, @wilbaker, I'll put the minor functional test changes into a small followup PR.  Thanks for all the help!!

chrisd8088:
As a postscript, #1503 should deal with some minor leftover issues, notably the functional test revisions @wilbaker wanted, and I also captured the question of whether it would be realistically eliminate all sources of uppercase SHA1s in github#35.

PR #1416, opened 2019-08-05
Mac: Fix and enable remaining dehydrate functional tests

### LINKED ISSUE #1216: Mac: Turn on Functional Test Categories.MacTODO.NeedsDehydrate
This may require both code and test changes.

wilbaker:
Resolves #1216 

VFS for Git was already doing the right thing.  The only changes needed to the test code:

- Switch from `CmdRunner.DeleteDirectoryWithUnlimitedRetries` to `RepositoryHelpers.DeleteTestDirectory`
- Updated `DirectoryShouldContain` to sort the entries before comparing them as the order is different on Mac and Windows

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
@@ -4,8 +4,10 @@ using GVFS.FunctionalTests.Tools;
 using GVFS.Tests.Should;
 using NUnit.Framework;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 
 namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
 {
@@ -31,7 +33,6 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         }
 
         [TestCase]
-        [Category(Categories.MacTODO.NeedsDehydrate)]  // Dehydrate is not successful
         public void DehydrateShouldSucceedInCommonCase()
         {
             this.DehydrateShouldSucceed("The repo was successfully dehydrated and remounted", confirm: true, noStatus: false);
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -157,12 +155,12 @@ public void DehydrateShouldFailOnWrongDiskLayoutVersion()
...
-            string[] folderItems = this.fileSystem.EnumerateDirectory(directory).Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
-            folderItems.Length.ShouldEqual(fileOrFolders.Length);
-            for (int i = 0; i < fileOrFolders.Length; i++)
-            {
-                Path.GetFileName(folderItems[i]).ShouldEqual(fileOrFolders[i]);
-            }
+            IEnumerable<string> onDiskItems =
+                this.fileSystem.EnumerateDirectory(directory)
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
+                .OrderByDescending(x => x);
```
wilbaker: Missing a call to `Path.GetFileName` here

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -157,12 +155,13 @@ public void DehydrateShouldFailOnWrongDiskLayoutVersion()
...
-            {
-                Path.GetFileName(folderItems[i]).ShouldEqual(fileOrFolders[i]);
-            }
+            IEnumerable<string> onDiskItems =
+                this.fileSystem.EnumerateDirectory(directory)
+                .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
+                .Select(path => Path.GetFileName(path))
+                .OrderByDescending(x => x);
+
+            onDiskItems.ShouldMatchInOrder(fileOrFolders.OrderByDescending(x => x));
```
jeschu1: Are we trying to verify that folderItems & fileOrFolders are equivalent?
If so this works it's just a shame you have to do the sorting.

Would this be a reasonable alternative?
`Assert.IsTrue(new HashSet<string>(onDiskItems).SetEquals(fileOrFolders));`

wilbaker: > Are we trying to verify that folderItems & fileOrFolders are equivalent?

Yep the issue here was that the order of the items in the folder is not the same on Windows and Mac.

> Would this be a reasonable alternative?

I had considered doing something similar to what you proposed, but what I liked about this option is that by using `ShouldMatchInOrder` we get nice output if the entries are not equal:

```
            if (groupList.Count != expectedValuesList.Count)
            {
                errorMessage.AppendLine(string.Format("{0} counts do not match. was: {1} expected: {2}", message, groupList.Count, expectedValuesList.Count));
            }

            foreach (T groupExtraItem in groupExtraItems)
            {
                errorMessage.AppendLine(string.Format("Extra: {0}", groupExtraItem));
            }

            foreach (T groupMissingItem in groupMissingItems)
            {
                errorMessage.AppendLine(string.Format("Missing: {0}", groupMissingItem));
            }

            if (errorMessage.Length > 0)
            {
                Assert.Fail("{0}\r\n{1}", message, errorMessage);
            }
```

However, just now I realized that `ShouldMatchInOrder` is not actually checking that they match in order so I'll push a commit to fix that.

jeschu1: > However, just now I realized that ShouldMatchInOrder is not actually checking that they match in order so I'll push a commit to fix that.

This may be fun and find other bugs in the code :-)

Since it gives us the nice output I'm find with using it.  Alternatively if you wanted to go a bit more crazy refactoring
`        public static IEnumerable<T> ShouldMatch<T>(this IEnumerable<T> group, params Action<T>[] itemCheckers, bool inOrder = false)
`
may be another option

Issue #1404, opened 2019-08-02
3-4 macOS functional test failures on master - "Unable to register repo because GVFS.Service is not responding."


pmj:
I was getting 3 identical, new functional test failures on different PRs, which didn't seem right so I kicked off a test run against master, and that now also fails.

The failing tests are:

* `GVFS.FunctionalTests.Tests.EnlistmentPerFixture.MountTests.MountCanProcessSavedBackgroundQueueTasks()`
* `GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.DehydrateShouldFailIfGitObjectsRootNotInMetadata()`
* `GVFS.FunctionalTests.Tests.EnlistmentPerFixture.CloneTests.CloneInsideUnmountedEnlistment()`

They all fail because the command output contains "`WARNING: Unable to register repo because GVFS.Service is not responding.`"
(The failing tests are different tests than #1204, hence tracking this as a new issue. Repro rate also seems much higher. Could be related of course, I can't tell.)

Retained logs:
 * https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10084307&view=results (PR #1208)
 * https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10083825&view=results (PR #1280)
 * https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10089649&view=results (master @ 34ed486) - on this run, 4 tests failed, the same 3 as above plus `GVFS.FunctionalTests.Tests.EnlistmentPerFixture.MountTests.MountChangesMountId()`, again for the same reason.

I don't really understand these test failures so I'd appreciate if someone more knowledgeable could take a look. Thanks!

pmj:
Note that repro rate is not 100%, I rebased PR #1280 on latest master earlier, and that run completed successfully: https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10089731&view=results

derrickstolee:
We had these happening early last week, and @jamill root-caused it to a long list of repos registered for auto-mount. I believe he is working on removing those repos before starting the functional test suite.

I see that all of your failures are on the GVFS-MACPRO-2 machine. Maybe that machine is overloaded already?

jamill:
I cleaned up the bad state on GVFS-MACPRO-2, we will keep an eye on it to make sure it is healthy again. I am looking at fixing the underlying issue.

pmj:
Thank you!

[...]

PR #1417, opened 2019-08-06
Update mac uninstall script to remove GVFS service directory

jamill:
The mac uninstall script did not remove the GVFS service directory, resulting
in GVFS artifacts being left around. This change updates the uninstall script
to also remove the GVFS service directory. This brings the behavior of the mac
uninstall script to match the Windows uninstall script in
src/Scripts/UninstallGVFS.bat.

One motivation for this change was that the Mac Functional test machines were
gathering cruft from repeated test runs, such as the list of registered
repositories in the repo-registry growing. This apparently caused the GVFS
service start up time to grow, resulting in test failures.

This address #1404

### DIFF
--- a/GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
+++ b/GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
@@ -6,6 +6,7 @@ PRJFSKEXTDIRECTORY="/Library/Extensions"
 LAUNCHDAEMONDIRECTORY="/Library/LaunchDaemons"
 LAUNCHAGENTDIRECTORY="/Library/LaunchAgents"
 LIBRARYAPPSUPPORTDIRECTORY="/Library/Application Support/VFS For Git"
+SERVICEAPPDIRECTORY="${HOME}/Library/Application Support/GVFS"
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Installer.Mac/uninstall_vfsforgit.sh
```diff
@@ -82,6 +83,12 @@ function UnInstallVFSForGit()
         echo "$rmCmd..."
         eval $rmCmd || { echo "Error: Could not delete ${LIBRARYAPPSUPPORTDIRECTORY}. Delete it manually."; exit 1; }
     fi
+
+    if [ -d "${SERVICEAPPDIRECTORY}" ]; then
+        rmCmd="/bin/rm -Rf \"${SERVICEAPPDIRECTORY}\""
```
kewillford: Most of the other `rm` commands use `sudo` and this one doesn't, is that because this one is in the user's home directory?  Will this never need `sudo` then?

jamill: > is that because this one is in the user's home directory?

Correct - these files are all user modifiable.

> Will this never need sudo then?

It shouldn't, unless something gets incorrect permissions... But maybe it is better to run this command as root in case something is set up unexpectedly, so we don't leave artifacts around for the next run.

Unless there are objections, I will update this to run as root.

PR #1419, opened 2019-08-06
Mac ProjFS kext: Only send NotifyFilePreDeleteFromRename on Mojave+

pmj:
isRename defaults to true on High Sierra where deletes cannot distinguish between rename and non-rename. In this case we've previously been sending a MessageType_KtoU_NotifyFilePreDeleteFromRename notification, which doesn't seem right. But I'm not 100% sure about the correct behaviour here.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -730,7 +730,7 @@ KEXT_STATIC int HandleVnodeOperation(
                 root,
                 isDirectory ?
                 MessageType_KtoU_NotifyDirectoryPreDelete :
-                isRename ? MessageType_KtoU_NotifyFilePreDeleteFromRename : MessageType_KtoU_NotifyFilePreDelete,
+                (isRename && s_osSupportsRenameDetection) ? MessageType_KtoU_NotifyFilePreDeleteFromRename : MessageType_KtoU_NotifyFilePreDelete,
                 currentVnode,
                 vnodeFsidInode,
                 nullptr, // path not needed, use fsid/inode


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -730,7 +730,7 @@ KEXT_STATIC int HandleVnodeOperation(
                 root,
                 isDirectory ?
                 MessageType_KtoU_NotifyDirectoryPreDelete :
-                isRename ? MessageType_KtoU_NotifyFilePreDeleteFromRename : MessageType_KtoU_NotifyFilePreDelete,
+                (isRename && s_osSupportsRenameDetection) ? MessageType_KtoU_NotifyFilePreDeleteFromRename : MessageType_KtoU_NotifyFilePreDelete,
```
wilbaker: I think we're better off leaving this as `MessageType_KtoU_NotifyFilePreDeleteFromRename` on High Sierra.

The main difference in the handling of these two events is here in PrjFSLib:

```
    // Remove placeholder xattrs for renames (PreDeleteFromRename) because:
    //  - Renames are treated as "Delete old path" + "create new path", and new files should not be marked as placeholders
    //  - The target of the rename might be outside of the root, and we should not allow placeholder files to
    //    live outside of the virtualization root (as the kext won't know what provider they belong to)
    if (result == PrjFS_Result_Success &&
        placeholderFile &&
        (PrjFS_NotificationType_PreConvertToFull == notificationType || PrjFS_NotificationType_PreDeleteFromRename == notificationType))
    {
        errno_t result = RemoveXAttrWithoutFollowingLinks(absolutePath, PrjFSFileXAttrName);
        if (0 != result && ENOATTR != result)
        {
            // It's expected that RemoveXAttrWithoutFollowingLinks return ENOATTR if
            // another thread has removed the attribute
            LogError("HandleFileNotification: RemoveXAttrWithoutFollowingLinks failed for '%s', error: %d", absolutePath, result);
        }
    }
```

I think it's better to default to removing xattrs even when we don't need to (i.e. deletes) than not remove them when we should (i.e. renames).

Hopefully in the next month or two we can mandate Mojave as the minimum version and won't need to worry about this at all anymore.

pmj: Ah, makes sense, thanks for clearing that up!

PR #1420, opened 2019-08-06
[M153 Hotfix] Update Git to revert slow commit-graph write

derrickstolee:
See microsoft/git#168 for full details.

v2.22.0 had a change that made writing commit-graph files much
slower by always parsing from pack-files instead of from the
commit-graph file. For some users of the OS repo, this caused
super-slow performance and high memory usage.

PR #1421, opened 2019-08-06
[master] Update Git, drop MIDX writes

derrickstolee:
This PR does two things:

1. Updates Git to include the slow commit-graph write fix from microsoft/git#168. Also see #1420 for the M153 hotfix.

2. Removes the multi-pack-index writes from the `PostFetchStep` and instead only writes the multi-pack-index during the `PackfileMaintenanceStep`. In order to get the most out of the step, we need to ensure we have a multi-pack-index before running the `expire` and `repack` steps. Also, we can only expire the packs from that day if they are contained in the multi-pack-index.

If we agree that we should send (2) as a hotfix to the M155 release (in addition to (1), which is necessary), then I'll create a hotfix PR to that branch.

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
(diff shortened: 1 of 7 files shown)


wilbaker:
Should the Milestone on this PR be M157 since it's targeting `master`?

PR #1422, opened 2019-08-06
[M155] Update Git, drop MIDX writes

derrickstolee:
This PR does two things:

1. Updates Git to include the slow commit-graph write fix from microsoft/git#168. Also see #1420 for the M153 hotfix.

2. Removes the multi-pack-index writes from the `PostFetchStep` and instead only writes the multi-pack-index during the `PackfileMaintenanceStep`. In order to get the most out of the step, we need to ensure we have a multi-pack-index before running the `expire` and `repack` steps. Also, we can only expire the packs from that day if they are contained in the multi-pack-index.

If we agree that we should send (2) as a hotfix to the M155 release (in addition to (1), which is necessary), then I'll create a hotfix PR to that branch.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
@@ -28,6 +28,7 @@ namespace GVFS.Common.Maintenance
     {
         public const string PackfileLastRunFileName = "pack-maintenance.time";
         public const string DefaultBatchSize = "2g";
+        private const string MultiPackIndexLock = "multi-pack-index.lock";
         private readonly bool forceRun;
         private readonly string batchSize;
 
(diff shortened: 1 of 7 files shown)


derrickstolee:
Related: #1421

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -105,15 +106,20 @@ protected override void PerformMaintenance()
...
 
                 if (!hasKeep)
                 {
                     activity.RelatedWarning(this.CreateEventMetadata(), "Skipping pack maintenance due to no .keep file.");
                     return;
-                }
+                }
+
+                string multiPackIndexLockPath = Path.Combine(this.Context.Enlistment.GitPackRoot, MultiPackIndexLock);
+                this.Context.FileSystem.TryDeleteFile(multiPackIndexLockPath);
```
wilbaker: I see we were doing this before too, but why do we always try to delete the lock file?

derrickstolee: Git will fail if the .lock file exists, as it thinks another process is writing to that file. It does not inspect handles. 

So, if a power-outage killed our Git process, then that user would never get an updated midx without this deletion of the .lock file.

We use the shared object cache lock to avoid multiple processes modifying this data.

PR #1429, opened 2019-08-07
Mac ProjFS kext: Deny I/O on offline roots, with exceptions, part 2

pmj:
This is a continuation of the work on issue #182. The remaining operations to block are:

- [x] A process attempts to read/execute an empty file in an offline root. The result of the read would be bad data, so failing with denied authorisation is preferable to letting the bad data propagate.
- [x] A process attempts to create files or directories in an offline root.

The commits implementing those 2 changes are fairly straightforward, however they did cause some knock-on effects:

 * During `gvfs clone`, the `git` subprocess itself wants to create files/directories in the offline root, so we need it to inherit the parent `gvfs` process's entitlement to access offline roots. The exception check therefore now walks up the process hierarchy.
 * One of the functional tests, `CheckoutBranchWithOpenHandleBlockingProjectionDeleteAndRepoMetdataUpdate` needs to mess with the repo while it's not mounted, so I've registered the test runner for offline I/O for the duration of that test.
 * To avoid weird edge cases, I've made the offline I/O registration in the native userlib thread-safe and reference counted, so registering multiple times is fine, and we only actually deregister once it's been matched with an equal number of unregister calls.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
@@ -16,6 +16,9 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
     [Category(Categories.GitCommands)]
     public class CheckoutTests : GitRepoTests
     {
+        private const string PrjFSLibPath = "libPrjFSLib.dylib";
+        private const int PrjFSResultSuccess = 1;
+
         public CheckoutTests(Settings.ValidateWorkingTreeMode validateWorkingTree)
             : base(enlistmentPerTest: true, validateWorkingTree: validateWorkingTree)
         {
@@ -498,6 +501,12 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         [TestCase]
         public void CheckoutBranchWithOpenHandleBlockingProjectionDeleteAndRepoMetdataUpdate()
         {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                uint result = RegisterForOfflineIO();
+                result.ShouldEqual<uint>(PrjFSResultSuccess, $"{nameof(RegisterForOfflineIO)} failed (result = {result})");
+            }
+
             this.ControlGitRepo.Fetch(GitRepoTests.ConflictSourceBranch);
             this.ControlGitRepo.Fetch(GitRepoTests.ConflictTargetBranch);
 
@@ -547,6 +556,13 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
                 // If the test fails, we should wait for the Task to complete so that it does not keep a handle open
                 task.Wait();
                 throw;
+            }
+            finally
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    UnregisterForOfflineIO();
+                }
             }
         }
 
@@ -956,5 +972,11 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
             [MarshalAs(UnmanagedType.U4)] FileMode creationDisposition,
             [MarshalAs(UnmanagedType.U4)] NativeFileAttributes flagsAndAttributes,
             [In] IntPtr templateFile);
+
+        [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_RegisterForOfflineIO")]
+        private static extern uint RegisterForOfflineIO();
+
+        [DllImport(PrjFSLibPath, EntryPoint = "PrjFS_UnregisterForOfflineIO")]
+        private static extern uint UnregisterForOfflineIO();
     }
 }
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -578,12 +578,21 @@ KEXT_STATIC int HandleVnodeOperation(
         {
             if (FileFlagsBitIsSet(currentVnodeFileFlags, FileFlags_IsEmpty))
             {
-                bool isWriteOperation = ActionBitIsSet(
-                    action,
-                    KAUTH_VNODE_WRITE_ATTRIBUTES |
-                    KAUTH_VNODE_WRITE_EXTATTRIBUTES |
-                    KAUTH_VNODE_WRITE_DATA |
-                    KAUTH_VNODE_APPEND_DATA);
+                // Prevent access to empty files in offline roots, except always allow the user to delete files.
+                bool shouldBlockIfOffline =
+                    (isRename && s_osSupportsRenameDetection) ||
+                    ActionBitIsSet(
+                        action,
+                        // Writes would get overwritten by subsequent hydration
+                        KAUTH_VNODE_WRITE_ATTRIBUTES |
+                        KAUTH_VNODE_WRITE_EXTATTRIBUTES |
+                        KAUTH_VNODE_WRITE_DATA |
+                        KAUTH_VNODE_APPEND_DATA |
+                        // Reads would yield bad (null) data
+                        KAUTH_VNODE_READ_DATA |
+                        KAUTH_VNODE_READ_ATTRIBUTES |
+                        KAUTH_VNODE_EXECUTE |
+                        KAUTH_VNODE_READ_EXTATTRIBUTES);
                 if (!TryGetVirtualizationRoot(
                         &perfTracer,
                         context,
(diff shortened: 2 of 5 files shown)


GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
```diff
@@ -16,6 +16,8 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
     [Category(Categories.GitCommands)]
     public class CheckoutTests : GitRepoTests
     {
+        private const string PrjFSLibPath = "libPrjFSLib.dylib";
```
jeschu1: nit: Do we need this variable?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
```diff
@@ -498,6 +500,12 @@ public void CheckoutBranchWithOpenHandleBlockingRepoMetdataUpdate()
         [TestCase]
         public void CheckoutBranchWithOpenHandleBlockingProjectionDeleteAndRepoMetdataUpdate()
         {
+            bool unregisterOfflineIO = false;
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                unregisterOfflineIO = 0 == RegisterForOfflineIO();
```
jeschu1: To be more in line with our C# syntax I would define 0 at the top of the file
```
private const int success = 0;
```

Then just fail if we can't register.  It will make it easier to debug the functional test that way 
```
If RegisterForOfflineIO() != success)
{
   Assert.Fail("Could not register for OfflineIO();
}
```

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -578,21 +596,29 @@ KEXT_STATIC int HandleVnodeOperation(
...
+                    (isRename && s_osSupportsRenameDetection) ||
+                    ActionBitIsSet(
+                        action,
+                        // Writes would get overwritten by subsequent hydration
+                        KAUTH_VNODE_WRITE_ATTRIBUTES |
+                        KAUTH_VNODE_WRITE_EXTATTRIBUTES |
+                        KAUTH_VNODE_WRITE_DATA |
+                        KAUTH_VNODE_APPEND_DATA |
+                        // Reads would yield bad (null) data
+                        KAUTH_VNODE_READ_DATA |
```
jeschu1: This change *I think* will have us call Hydrate on for some new attributes.  Does that sound correct?

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -664,14 +665,32 @@ bool VirtualizationRoots_ProcessMayAccessOfflineRoots(pid_t pid)
     
     RWLock_AcquireShared(s_virtualizationRootsLock);
     {
-        for (uint32_t i = 0; i < s_offlineIOPIDCount; ++i)
+        while (true)
```
jeschu1: Can we please add some comments here.  While true's in locks can be scary :-).

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -162,21 +165,36 @@ static mutex s_fileLocksMutex;
...
+    ++s_kernelServiceOfflineClientCount;
 
     return PrjFS_Result_Success;
 }
 
 PrjFS_Result PrjFS_UnregisterForOfflineIO()
 {
+    mutex_lock lock(s_kernelServiceOfflineClientMutex);
+    
+    assert(s_kernelServiceOfflineClientCount > 0);
```
jeschu1: If we have 2 people register for offline, and then 1 unregisters wouldn't we assert?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -162,21 +165,36 @@ static mutex s_fileLocksMutex;
...
     assert(s_kernelServiceOfflineWriterConnection != IO_OBJECT_NULL);
-
-    IOServiceClose(s_kernelServiceOfflineWriterConnection);
-    s_kernelServiceOfflineWriterConnection = IO_OBJECT_NULL;
+    --s_kernelServiceOfflineClientCount;
+    if (s_kernelServiceOfflineClientCount == 0)
+    {
+        IOServiceClose(s_kernelServiceOfflineWriterConnection);
+        s_kernelServiceOfflineWriterConnection = IO_OBJECT_NULL;
```
jeschu1: The next time register is called, will have a problem?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: Every successful call to `PrjFS_RegisterForOfflineIO()` increments the count via `++s_kernelServiceOfflineClientCount;`, so if 2 calls to register are made, the count will be 2. I'm not clear on why this would assert. The old code would have asserted on the second unregister.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: As the count has hit 0, the next register call will re-connect to the kext. So I think I've got this right - if you've worked out a sequence of events that might not be handled correctly, please elaborate!

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: That is certainly not the intention. The call to `ProviderMessaging_TrySendRequestAndWaitForResponse` below hasn't changed, the intended change here is to block reads from empty files when the provider is offline - of course precisely because it's not possible to hydrate in this case.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: I see so we assert if someone calls `PrjFS_UnregisterForOfflineIO` without a previous pairing  `PrjFS_RegisterForOfflineIO`.
That seems fair.

jeschu1: Let's please add a comment above the assert so it's 100% clear

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
```diff
@@ -664,14 +665,32 @@ bool VirtualizationRoots_ProcessMayAccessOfflineRoots(pid_t pid)
...
+            proc_t process = proc_find(pid);
+            if (process == nullptr)
+            {
+                break;
+            }
+            
+            pid = proc_ppid(process);
+            proc_rele(process);
+            
+            if (pid <= 1)
```
jeschu1: Is the idea we walk all the parents to check for the pid allowed to register.  Let's please add some comments here so it's super obvious whats going on.

jeschu1:
@pmj getting very very close, lets add some more comments and do the small change for the c# test

GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
pmj: Yes, that makes more sense. I've updated the code to something along those lines.

ProjFS.Mac/PrjFSKext/VirtualizationRoots.cpp
pmj: I've added some comments to this function - please let me know if there's any further detail I should cover in comments.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: I've added comments in both the registration and deregistration functions explaining how each of them deals with the refcounting.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
```diff
@@ -498,6 +501,12 @@ public void CheckoutBranchWithOpenHandleBlockingRepoMetdataUpdate()
         [TestCase]
         public void CheckoutBranchWithOpenHandleBlockingProjectionDeleteAndRepoMetdataUpdate()
         {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                uint result = RegisterForOfflineIO();
+                Assert.Equals(PrjFSResultSuccess, result);
```
wilbaker: Nit: We've been using the "Should" methods in the `.cs` tests, can you update this to use `ShouldEqual`:

`result.ShouldEqual(PrjFSResultSuccess, $"{nameof(RegisterForOfflineIO)} failed")`

PR #1431, opened 2019-08-07
Add lock around writes to SQLite database tables

kewillford:
Because SQLite only allows one writer at a time, we need to protect writes that might happen on multiple threads.  Most of the time this isn't an issue because in WAL mode it will only be locking, appending, unlocking which is very fast but there is still a small window where another thread tries to write and it is locked and will get an exception.

This change will force threads to be synchronized before calling `ExecuteNonQuery` to make sure only one thread is writing at a time.

In the performance tests there was not any change.  I also tried turning off the shared cache as a fix but some operations became 2X slower without the shared cache.

### DIFF
--- a/GVFS/GVFS.Common/Database/PlaceholderTable.cs
+++ b/GVFS/GVFS.Common/Database/PlaceholderTable.cs
@@ -10,6 +10,7 @@ namespace GVFS.Common.Database
     public class PlaceholderTable : IPlaceholderCollection
     {
         private IGVFSConnectionPool connectionPool;
+        private object writerLock = new object();
 
         public PlaceholderTable(IGVFSConnectionPool connectionPool)
         {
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Database/PlaceholderTable.cs
```diff
@@ -166,7 +169,16 @@ public void Remove(string path)
                 {
                     command.CommandText = "DELETE FROM Placeholder WHERE path = @path;";
                     command.AddParameter("@path", DbType.String, path);
-                    command.ExecuteNonQuery();
+
+                    try
+                    {
+                        this.writeToDatabaseResetEvent.WaitOne();
```
wilbaker: What was the reason for using `AutoResetEvent` instead of `lock`?

I look around a bit and saw a discussion [here](https://stackoverflow.com/questions/17273933/autoresetevent-as-a-lock-replacement-in-c) that `AutoResetEvent` does not provide the same fairness guarantees as `lock`:

> WaitHandles do not provide such a guarantee. Which is very detrimental if the lock is contended, the same thread can acquire it repeatedly and other threads can starve forever.
>
> Use an appropriate synchronization object for locks. Wait handles should only be used for signaling.

Is that something we need to be concerned about?

kewillford: I had thought that locks were more heavy weight but I ran the performance tests using a `lock` and there doesn't seem to be a difference in the performance at least in what our performance tests are doing.  So I changed to use a lock since that seems to be simpler.

PR #1413, opened 2019-08-05
[Linux] Revise functional tests for cross-device mounts

chrisd8088:
On POSIX systems, where a `rename(2)` returns `EXDEV` when trying to move a file or directory between mounted filesystems, the functional tests which do this currently fail if the
repository is mounted on a separate device (as is the case on Linux).

Therefore we create a `RepositoryMountsSameFileSystem` test category for use on VFSForGit implementations which allow renaming across repository boundaries (i.e., Windows PrjFlt
and Mac kext), and mark a number of tests with this category designation as they move files or folders in or out of the repository, or create hard links across the repository boundary.

In the `ModifiedPathsTests` suite, two specific tests which check the contents of the modified-paths database after remounting and after hard-linking, specifically, are subdivided into pairs of tests.  Each pair consists of one test which only checks changes that do not cross the repository boundary, suitable for all implementations, and one test which check changes spanning the repository boundary and which is marked with the `RepositoryMountsSameFileSystem`
category.

PR #1432, opened 2019-08-08
[Linux] enable most functional tests on Linux

chrisd8088:
This is an initial set of changes to enable most functional tests to run -- if not succeed -- on Linux.

We build the `GVFS.FunctionalTests` projects on Linux, change a number of `MacTODO` categories into `NonWindowsTODO` equivalents, and deal with some Linux-specific failures caused by the present lack of a `GVFS.Service` daemon and the temporary use of `flock(2)` in libprojfs (see https://github.com/github/libprojfs/issues/79 for details).

Enabling additional functional tests on Linux depends on the work in #1411, #1413, and especially #1412.

/cc @kivikakk, @wilbaker


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MultithreadedReadWriteTests.cs
```diff
@@ -133,32 +164,69 @@ public void CanReadWriteAFileInParallel(FileSystemRunner fileSystem)
...
+                    while (retry)
                     {
-                        while (keepRunning)
+                        Stream readStream = null;
+                        try
                         {
-                            Thread.Yield();
-                            fileContents[myIndex].Append(reader.ReadToEnd());
+                            readStream = File.Open(virtualPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+                            retry = false;
```
derrickstolee: Am I reading this right that this loop will repeat infinitely if `File.Open()` throws? What's the benefit there?

derrickstolee: (and that's the only way it will loop at all)

chrisd8088: Good catch and question -- the workaround here is to deal with `EAGAIN` from the `File.Open()`.  I think I'm missing some additional handling around the non-`EAGAIN` case, where we want to exit the loop.  Stay tuned and thanks!

Issue #1367, opened 2019-07-25
Mac: Fallback to using file path in HandleFileOpOperation if the vnode lookup fails


wilbaker:
`HandleFileOpOperation` will not deliver rename `KAUTH_FILEOP_RENAME` or `KAUTH_FILEOP_LINK` related notifications if `vnode_lookup` fails.

Rather than immediately giving up, we should see if we can find the root via the file's path.

pmj:
Are we seeing the `vnode_lookup()` call fail in the wild? (I'm not saying the fallback isn't worth implementing, just that if it's currently failing occasionally on user machines, I'd be curious to find a root cause.)

wilbaker:
> Are we seeing the vnode_lookup() call fail in the wild? (I'm not saying the fallback isn't worth implementing, just that if it's currently failing occasionally on user machines, I'd be curious to find a root cause.)

@pmj no we haven't had any reports of it, however, I do remember occasionally seeing functional test failures that could only be explained by a rename event not being delivered (it was too infrequent to root cause).

This issue got spun up when I checked the code for all `TODO(Mac)` comments and filed issues for each.

pmj:
> > Are we seeing the vnode_lookup() call fail in the wild? (I'm not saying the fallback isn't worth implementing, just that if it's currently failing occasionally on user machines, I'd be curious to find a root cause.)
> 
> @pmj no we haven't had any reports of it, however, I do remember occasionally seeing functional test failures that could only be explained by a rename event not being delivered (it was too infrequent to root cause).

Unfortunately the code behind `vnode_lookup()` is very complex, and of course calls into filesystem ops, etc., so it's difficult to say for sure if there could be spurious failures. The underlying code is the same as for the `open()` syscall, and the only difference for the latter is that the audit flag is set, which shouldn't affect success/failure. And `open()` isn't known to just arbitrarily fail to look up a path.

The `KAUTH_FILEOP_RENAME` notification should however be reliable, unlike `KAUTH_FILEOP_OPEN`, `_CLOSE`, and `_EXEC`, which rely on `vn_getpath`.

[...]

PR #1435, opened 2019-08-09
Mac ProjFS kext: Adds fallback for vnode_lookup failure in fileop handler

pmj:
In the case where looking up the vnode from the KAUTH_FILEOP_RENAME event path fails, fall back to finding the provider responsible via the path alone.

Various follow-on code assumed a valid vnode pointer, so some minor adjustments were necessary.

This implements #1367.


ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -786,11 +786,10 @@ KEXT_STATIC int HandleFileOpOperation(
         if (0 != toErr)
```
wilbaker: Minor: With this change the TODO comment can be removed.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -786,11 +786,10 @@ KEXT_STATIC int HandleFileOpOperation(
...
         {
             KextLog_Error("HandleFileOpOperation (KAUTH_FILEOP_RENAME): vnode_lookup failed, errno %d for path '%s'", toErr, newPath);
-            goto CleanupAndReturn;
```
wilbaker: Make these same change for KAUTH_FILEOP_LINK (see line 859 below)

pmj: I'm still iterating on this PR (hence draft status) - I'll ping you via the "request review" button when I'm done so you don't end up reviewing draft code.

wilbaker: Sure, sorry about missing that and jumping the gun!

pmj: No worries, just pointing it out so you don't end up wasting time! Github's "Draft" indicator is a bit too subtle I think, I'll try to remember to add a [WIP] or similar to the title again in future.

pmj:
The `vnode_lookup` failure fallback paths for both rename and hardlink events are now implemented. All tests pass, but as these failures are extremely rare, this is not surprising as the workaround code almost certainly won't run.

I'm therefore also pushing 2 commits (56e4c0b, 44eec3f) which randomly (10% probability) force taking the failure path as the new code paths are otherwise very unlikely to be taken on any given test run. Looking at the kext log for the test runs with fault injection, the failure paths are definitely taken frequently during functional testing:

 * https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10158430&view=results

It does raise the question: do we generally want to start running functional tests with fault injection enabled for certain API calls which may very rarely fail? (I'm thinking of `vnode_getparent`, `vn_getpath`, `vnode_lookup` in the fileop handler, etc.) We now have mitigation code for a lot of these failures, but we're unlikely to notice regressions.

We can of course make sure these code paths are covered with unit tests, but having effectively zero functional test coverage for them worries me a bit.

pmj:
So, the fault injection test found an issue: looks like *bad things happen* when the renamed vnode is a directory and we send a `MessageType_KtoU_NotifyFileRenamed` message as opposed to `MessageType_KtoU_NotifyDirectoryRenamed` (because we have no vnode, so we can't tell if it's a directory). I've pushed a possible workaround but it remains to be seen if that helps or makes things worse.

PR #1436, opened 2019-08-09
Allow privileged system service 'amfid' to hydrate

jeschu1:
A user attempted to run 'open project.app' and failed.  

This was because it called the system process 'amfid' used in signing.  We had blocked all system processes from hydrating.

This change creates an exception for 'amfid' and allows it to hydrate.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -70,7 +70,7 @@ static bool TryReadVNodeFileFlags(vnode_t vn, vfs_context_t _Nonnull context, ui
 KEXT_STATIC_INLINE bool FileFlagsBitIsSet(uint32_t fileFlags, uint32_t bit);
 KEXT_STATIC_INLINE bool TryGetFileIsFlaggedAsInRoot(vnode_t vnode, vfs_context_t _Nonnull context, bool* flaggedInRoot);
 KEXT_STATIC_INLINE bool ActionBitIsSet(kauth_action_t action, kauth_action_t mask);
-KEXT_STATIC bool CurrentProcessWasSpawnedByRegularUser();
+KEXT_STATIC bool CurrentProcessIsAllowedToHydrate();
 KEXT_STATIC bool IsFileSystemCrawler(const char* procname);
 
 static void WaitForListenerCompletion();
@@ -1265,10 +1265,10 @@ static bool TryGetVirtualizationRoot(
         return false;
     }
     
-    if (callbackPolicy == CallbackPolicy_UserInitiatedOnly && !CurrentProcessWasSpawnedByRegularUser())
+    if (callbackPolicy == CallbackPolicy_UserInitiatedOnly && !CurrentProcessIsAllowedToHydrate())
     {
         // Prevent hydration etc. by system services
-        KextLog_Info("TryGetVirtualizationRoot: process %d restricted due to owner UID.", pidMakingRequest);
+        KextLog_Info("TryGetVirtualizationRoot: process %d is not allowed to hydrate.", pidMakingRequest);
         perfTracer->IncrementCount(PrjFSPerfCounter_VnodeOp_GetVirtualizationRoot_UserRestriction);
         
         *kauthResult = KAUTH_RESULT_DENY;
@@ -1280,7 +1280,7 @@ static bool TryGetVirtualizationRoot(
     return true;
 }
 
-KEXT_STATIC bool CurrentProcessWasSpawnedByRegularUser()
+KEXT_STATIC bool CurrentProcessIsAllowedToHydrate()
 {
     bool nonServiceUser = false;
     
@@ -1309,7 +1309,7 @@ KEXT_STATIC bool CurrentProcessWasSpawnedByRegularUser()
         process = parentProcess;
         if (parentProcess == nullptr)
         {
-            KextLog_Error("CurrentProcessIsOwnedOrWasSpawnedByRegularUser: Failed to locate ancestor process %d for current process %d\n", parentPID, proc_selfpid());
+            KextLog_Error("CurrentProcessIsAllowedToHydrate: Failed to locate ancestor process %d for current process %d\n", parentPID, proc_selfpid());
             break;
         }
     }
@@ -1318,6 +1318,19 @@ KEXT_STATIC bool CurrentProcessWasSpawnedByRegularUser()
     if (process != nullptr)
     {
         proc_rele(process);
+    }
+    
+    if (!nonServiceUser)
+    {
+       // When amfid is invoked to check the code signature of a library which has not been hydrated,
+       // blocking hydration would cause the launch of an application which depends on the library to fail,
+       // so amfid should always be allowed to hydrate.
+       char buffer[MAXCOMLEN + 1] = "";
+       proc_selfname(buffer, sizeof(buffer));
+       if (0 == strcmp(buffer, "amfid"))
+       {
+          nonServiceUser = true;
+       }
     }
     
     return nonServiceUser;
--- a/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
+++ b/ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
@@ -29,7 +29,7 @@ using std::string;
     self->cacheWrapper.AllocateCache();
     context = vfs_context_create(nullptr);
     MockProcess_AddContext(context, 501 /*pid*/);
-    MockProcess_SetSelfPid(501);
+    MockProcess_SetSelfInfo(501, "Test");
     MockProcess_AddProcess(501 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "test" /*name*/);
 }
 
@@ -247,7 +247,7 @@ using std::string;
     // Test with file crawler trying to populate an empty file
     testVnode->attrValues.va_flags = FileFlags_IsEmpty | FileFlags_IsInVirtualizationRoot;
     MockProcess_Reset();
-    MockProcess_SetSelfPid(501);
+    MockProcess_SetSelfInfo(501, "Test");
     MockProcess_AddContext(context, 501 /*pid*/);
     MockProcess_AddProcess(501 /*pid*/, 1 /*credentialId*/, 1 /*ppid*/, "mds" /*name*/);
     XCTAssertFalse(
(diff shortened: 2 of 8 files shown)


jeschu1:
I should add, thanks again to @pmj for his help diagnosing here!

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -287,18 +287,26 @@ - (void)testCurrentProcessWasSpawnedByRegularUser {
...
     MockProcess_AddProcess(500 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
     XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
     MockProcess_Reset();
 
+    // Process is a 'amfid' service user and does not have a parent
+    MockProcess_AddContext(context, 500 /*pid*/);
+    MockProcess_SetSelfInfo(500, "amfid");
+    MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
+    MockProcess_AddProcess(500 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
+    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
```
jeschu1: `CurrentProcessWasSpawnedByRegularUser` is no longer the best name.

I could argue we only have 1 exclusion so keep the name, or `CurrentProcessWasSpawnedByAllowedUser `or similar?  @wilbaker @pmj

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -1320,6 +1320,18 @@ KEXT_STATIC bool CurrentProcessWasSpawnedByRegularUser()
         proc_rele(process);
     }
     
+    if (!nonServiceUser)
+    {
+       // We need to allow the system process amfid to run
+       // It is required for 'open project.app'
+       char buffer[MAXCOMLEN + 1] = "";
+       proc_selfname(buffer, sizeof(buffer));
+       if (0 == strcmp(buffer, "amfid"))
```
jamill: For my own curiosity, as someone not familiar with the code :) - why do we operate with c styled strings, instead of the string type? The proc name is already stored as a string, and we have to convert it to a c style string. Would it be possible to stay in the domain of the String type?

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
```diff
@@ -149,6 +151,11 @@ int proc_selfpid(void)
     return s_selfPid;
 }
 
+void proc_selfname(char* buf, int size)
+{
+    strlcpy(buf, s_selfName.c_str(), size);
```
jamill: Do we need to check the return value from `strlcpy` or ensure we were able to copy the full name into the provided buffer?

jeschu1: We haven't been doing that in the code base.  I don't think it's a necessity for this test code.  It would certainly double check the behavior but for strlcpy but I think the behavior should be well defined.  Open to others onions on this too!

jamill: cool - thanks for answering. I was just curious.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
wilbaker: > why do we operate with c styled strings, instead of the string type?  The proc name is already stored as a string, and we have to convert it to a c style string.

This is because STL is not supported for kernel extensions.  The spot you saw where the proc name is a string is in the unit tests (where we are allowed to use STL).

jamill: ah! thanks!

jeschu1: We don't have the 'string' type in KauthHandler.cpp.  This is on the 'hot' path for the kext so we want to stay as lean and fast as possible.  The test code does have 'string' for ease of use.

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
wilbaker: In this specific case we're actually matching the behavior of the OS:

```
void
proc_selfname(char * buf, int  size)
{
	proc_t p;

	if ((p = current_proc())!= (proc_t)0) {
		strlcpy(buf, &p->p_comm[0], size);
	}
}
```
https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/kern/kern_proc.c#L952

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
wilbaker: I like the idea of renaming the function but I am having trouble with name suggestions 😄 

What I'm not a big fan of with `CurrentProcessWasSpawnedByAllowedUser` is that is sounds like we're making the decision based on a user's identity alone.

Perhaps `CurrentProcessWasSpawnedByRegularUserOrCanAlwaysHydrate`? But I know it's a real mouthful.

I suppose another option is to put this new check in its own function `CurrentProcessCanAlwaysHydrate`, but then we have the trouble of always remembering to call both.

I'm also okay with leaving this if we can't think of a better name

jeschu1: @wilbaker I'm LOL here cause that was my exact thought process!  

I'm leaning it towards leaving it as is for now.  If we add a second exclusion in the future, I think we definitely need to rename and/or split it out.

wilbaker: One other option, if we name it based on what question it's trying to answer: `CurrentProcessIsAllowedToHydrate`, or if we flipped the return value `CurrentProcessIsRestrictedFromHydrating`

jeschu1: I went with `CurrentProcessIsAllowedToHydrate ` for clarity.  Thanks for the opinions here @wilbaker!

ProjFS.Mac/PrjFSKextTests/MockProc.cpp
pmj: The procname is always limited to `MAXCOMLEN` `char`s anyway, so any calling code will expect names to be truncated, as this is extremely common in practice.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -1265,7 +1265,7 @@ static bool TryGetVirtualizationRoot(
         return false;
     }
     
-    if (callbackPolicy == CallbackPolicy_UserInitiatedOnly && !CurrentProcessWasSpawnedByRegularUser())
+    if (callbackPolicy == CallbackPolicy_UserInitiatedOnly && !CurrentProcessIsAllowedToHydrate())
     {
         // Prevent hydration etc. by system services
         KextLog_Info("TryGetVirtualizationRoot: process %d restricted due to owner UID.", pidMakingRequest);
```
pmj: I guess this log message is no longer entirely accurate.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -1320,6 +1320,18 @@ KEXT_STATIC bool CurrentProcessWasSpawnedByRegularUser()
         proc_rele(process);
     }
     
+    if (!nonServiceUser)
+    {
+       // We need to allow the system process amfid to run
+       // It is required for 'open project.app'
```
pmj: `amfid` is responsible for code signature checks, including on dynamic libraries/frameworks. And this is where the problem was - the dynamic linker `dyld` detected a framework dependency of the app, and so queried `amfid` for the code signing status of the framework. The framework wasn't hydrated, and as `amfid` runs as root, it was blocked from accessing the framework's files instead of hydrating them.

So perhaps we can be more specific in the comment:

```
// When amfid is invoked to check the code signature of a library which has not been hydrated,
// blocking hydration would cause the launch of an application which depends on the library to fail,
// so amfid should always be allowed to hydrate.
```

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -282,47 +282,55 @@ - (void)testShouldHandleVnodeOpEvent {
...
 
-- (void)testCurrentProcessWasSpawnedByRegularUser {
+- (void)testCurrentProcessIsAllowedToHydrate {
     // Defaults should pass for all tests
-    XCTAssertTrue(CurrentProcessWasSpawnedByRegularUser());
+    XCTAssertTrue(CurrentProcessIsAllowedToHydrate());
     MockProcess_Reset();
 
-    // Process is a service user and does not have a parent
+    // Process is not a 'amfid' service user and does not have a parent
```
pmj: The comment here was previously wrong (all processes have parents), so if we're going to edit it, may as well get it right. Perhaps something like:

```
// Process and its parents are owned by a system user, and the executable is not a whitelisted service.
```

ProjFS.Mac/PrjFSKextTests/KauthHandlerTests.mm
```diff
@@ -282,47 +282,55 @@ - (void)testShouldHandleVnodeOpEvent {
...
     MockProcess_AddContext(context, 500 /*pid*/);
-    MockProcess_SetSelfPid(500);
+    MockProcess_SetSelfInfo(500, "Test");
     MockProcess_AddCredential(1 /*credentialId*/, 1 /*UID*/);
     MockProcess_AddProcess(500 /*pid*/, 1 /*credentialId*/, 501 /*ppid*/, "test" /*name*/);
-    XCTAssertFalse(CurrentProcessWasSpawnedByRegularUser());
+    XCTAssertFalse(CurrentProcessIsAllowedToHydrate());
+    MockProcess_Reset();
+
+    // Process is a 'amfid' service user and does not have a parent
```
pmj: Similar problem as above, e.g.
```
// The service "amfid" and its parents are owned by system users, but the process name is whitelisted for hydration.
```

jeschu1:
> The functionality is good, but I think some of the comments, naming, etc. could be improved.

Thanks for pointing this out @pmj, I've done the updates accordingly

PR #1438, opened 2019-08-09
[releases/shipped] Allow privileged system service 'amfid' to hydrate

jeschu1:
A user attempted to run 'open project.app' and failed.

This was because it called the system process 'amfid' used in signing. We had blocked all system processes from hydrating.

This change creates an exception for 'amfid' and allows it to hydrate.

cherry-pick from master


wilbaker:
It looks like the Sparse mode tests will always fail in this PR because this branch does not have that feature, and so I don't think we need to worry about it failing:

```
sc start GVFS.Service Errors: 
Installing Test.GVFS.Service
Using service executable: C:\Program Files\GVFS\GVFS.Service.exe
NUnitLite 3.10.1 (.NET Standard 2.0)
Copyright (c) 2018 Charlie Poole, Rob Prouse

Invalid argument: --sparse-mode
```

jeschu1:
Note the failing build is never worked on this branch and is unrelated to the changes.

PR #1440, opened 2019-08-12
Add LogInfo / LogWarning messages to PrjFSLib

### LINKED ISSUE #395: Mac: Add logging support to the ProjFS user-mode library

jeschu1:
I've verified LogInfo/LogWarning comes thru, but we should definitely double checks the spots I added logging and make sure no cases were missed.

resolves #395

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -140,6 +140,8 @@ static FileMutexMap::iterator CheckoutFileMutexIterator(const FsidInode& fsidIno
 static void ReturnFileMutexIterator(FileMutexMap::iterator lockIterator);
 
 static void LogError(const char* formatString, ...) __attribute__((__format__ (printf, 1, 2)));
+static void LogWarning(const char* formatString, ...) __attribute__((__format__ (printf, 1, 2)));
+static void LogInfo(const char* formatString, ...) __attribute__((__format__ (printf, 1, 2)));
 
 // State
 static io_connect_t s_kernelServiceConnection = IO_OBJECT_NULL;
@@ -193,6 +195,8 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
         << callbacks.GetFileStream << ", "
         << callbacks.NotifyOperation << ", "
         << callbacks.LogError << ","
+        << callbacks.LogWarning << ","
+        << callbacks.LogInfo << ","
         << poolThreadCount << ")" << endl;
 #endif
     
@@ -200,13 +204,16 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
         nullptr == callbacks.EnumerateDirectory ||
         nullptr == callbacks.GetFileStream ||
         nullptr == callbacks.NotifyOperation ||
-        nullptr == callbacks.LogError)
+        nullptr == callbacks.LogError ||
+        nullptr == callbacks.LogWarning ||
+        nullptr == callbacks.LogInfo)
     {
         return PrjFS_Result_EInvalidArgs;
     }
     
     if (!s_virtualizationRootFullPath.empty())
     {
+        LogError("PrjFS_StartVirtualizationInstance: s_virtualizationRootFullPath is not empty");
         return PrjFS_Result_EInvalidOperation;
     }
     
@@ -225,7 +232,7 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
     s_messageQueueDispatchQueue = dispatch_queue_create("PrjFS Kernel Message Handling", DISPATCH_QUEUE_SERIAL);
     if (!PrjFSService_DataQueueInit(&dataQueue, s_kernelServiceConnection, ProviderPortType_MessageQueue, ProviderMemoryType_MessageQueue, s_messageQueueDispatchQueue))
     {
-        cerr << "Failed to set up shared data queue.\n";
+        LogError("PrjFS_StartVirtualizationInstance: Failed to set up shared data queue.");
         return PrjFS_Result_EInvalidOperation;
     }
     
@@ -235,7 +242,7 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
     errno_t error = RegisterVirtualizationRootPath(virtualizationRootFullPath);
     if (error != 0)
     {
-        cerr << "Registering virtualization root failed: " << error << ", " << strerror(error) << endl;
+        LogError("PrjFS_StartVirtualizationInstance: Registering virtualization root failed: error=%d strerror=%s", error, strerror(error));
         return PrjFS_Result_EInvalidOperation;
     }
     
@@ -256,7 +263,7 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
             uint32_t messageSize = entry->size;
             if (messageSize < sizeof(Message))
             {
-                cerr << "Bad message size: got " << messageSize << " bytes, expected minimum of " << sizeof(Message) << ", skipping. Kernel/user version mismatch?\n";
+                LogError("PrjFS_StartVirtualizationInstance: Bad message size: got %d bytes, expected minimum of %lu, skipping. Kernel/user version mismatch?", messageSize, sizeof(Message));
                 DataQueue_Dequeue(dataQueue.queueMemory, nullptr, nullptr);
                 continue;
             }
@@ -266,7 +273,7 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
             IOReturn result = DataQueue_Dequeue(dataQueue.queueMemory, messageMemory, &dequeuedSize);
             if (kIOReturnSuccess != result || dequeuedSize != messageSize)
             {
-                cerr << "Unexpected result dequeueing message - result 0x" << hex << result << " dequeued " << dequeuedSize << "/" << messageSize << " bytes\n";
+                LogError("PrjFS_StartVirtualizationInstance: Unexpected result dequeueing message - result 0x%08x dequeued %d/%d bytes", result, dequeuedSize, messageSize);
                 abort();
             }
 
@@ -346,6 +353,7 @@ PrjFS_Result PrjFS_WritePlaceholderDirectory(
                 result = PrjFS_Result_EPathNotFound;
                 break;
             default:
+                LogWarning("PrjFS_WritePlaceholderDirectory: mkdir failed errno=%d stderror=%s", errno, strerror(errno));
                 result = PrjFS_Result_EIOError;
                 break;
         }
@@ -406,6 +414,7 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
                 break;
             case EEXIST: // The file already exists
             default:
+                LogWarning("PrjFS_WritePlaceholderFile: fopen failed filename=%s errorno=%d stderror=%s", fullPath, errno, strerror(errno));
                 result = PrjFS_Result_EIOError;
                 break;
         }
@@ -431,6 +440,7 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
     // TODO(#1370): Only call chmod if fileMode is different than the default file mode
     if (chmod(fullPath, fileMode))
     {
+        LogWarning("PrjFS_WritePlaceholderFile: failed to change permissions for %s errno=%d strerror=%s", fullPath, errno, strerror(errno));
         result = PrjFS_Result_EIOError;
         goto CleanupAndFail;
     }
@@ -470,6 +480,7 @@ PrjFS_Result PrjFS_WriteSymLink(
     
     if(symlink(symLinkTarget, fullPath))
     {
+        LogWarning("PrjFS_WriteSymLink: symLink call failed symLinkTarget=%s fullPath=%s errno=%d, strerror=%s", symLinkTarget, fullPath, errno, strerror(errno));
         goto CleanupAndFail;
     }
     
@@ -508,6 +519,7 @@ PrjFS_Result PrjFS_UpdatePlaceholderFileIfNeeded(
     PrjFS_Result result = PrjFS_DeleteFile(relativePath, updateFlags, failureCause);
     if (result != PrjFS_Result_Success)
     {
+       LogWarning("PrjFS_UpdatePlaceholderFileIfNeeded: PrjFS_DeleteFile call failed relativePath=%s updateFlags=%d failureCause=%d", relativePath, updateFlags, *failureCause);
        return result;
     }
 
@@ -532,6 +544,7 @@ PrjFS_Result PrjFS_ReplacePlaceholderFileWithSymLink(
     PrjFS_Result result = PrjFS_DeleteFile(relativePath, updateFlags, failureCause);
     if (result != PrjFS_Result_Success)
     {
+       LogWarning("PrjFS_ReplacePlaceholderFileWithSymLink: PrjFS_DeleteFile call failed relativePath=%s updateFlags=%d failureCause=%d", relativePath, updateFlags, *failureCause);
        return result;
     }
     
@@ -571,6 +584,7 @@ PrjFS_Result PrjFS_DeleteFile(
             case ENOTDIR: // A component of fullPath is not a directory
                 return PrjFS_Result_Success;
             default:
+                LogWarning("PrjFS_DeleteFile: stat call failed fullPath=%s errno=%d strerror=%s", fullPath, errno, strerror(errno));
                 return PrjFS_Result_EIOError;
         }
     }
@@ -579,6 +593,7 @@ PrjFS_Result PrjFS_DeleteFile(
     {
         // Only files and directories can be deleted with PrjFS_DeleteFile
         // Anything else should be treated as a full file
+        LogWarning("PrjFS_DeleteFile: path is not a regular file or directory %s, %d", fullPath, path_stat.st_mode);
         *failureCause = PrjFS_UpdateFailureCause_FullFile;
         return PrjFS_Result_EVirtualizationInvalidOperation;
     }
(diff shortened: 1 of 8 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1550,6 +1588,62 @@ static void LogError(const char* formatString, ...)
     vasprintf(&logString, formatString, dataArgs);
     va_end(dataArgs);
     
-    s_callbacks.LogError(logString);
+    if (s_callbacks.LogError != nullptr)
```
jeschu1: In testing I found that s_callbacks.LogError could be null depending on if the mount process is fully set up.  Instead of crashing and getting no error, I switched this to not crash and give an error in debug mode.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1550,6 +1588,62 @@ static void LogError(const char* formatString, ...)
...
+    va_end(dataArgs);
+    
+    if (s_callbacks.LogInfo != nullptr)
+    {
+       s_callbacks.LogInfo(logString);
+    }
+    else
+    {
+        #ifdef DEBUG
+        cout << "LogInfo: s_callbacks.Info is [NULL], Unable to log: %s " << logString << endl;
```
pmj: Nit: Should be `s_callbacks.LogInfo`

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1550,6 +1588,62 @@ static void LogError(const char* formatString, ...)
...
+    {
+        #ifdef DEBUG
+        cout << "LogWarning: s_callbacks.LogWarning is [NULL], Unable to log: %s " << logString << endl;
+        #endif
+    }
+    
+    free(logString);
+}
+
+static void LogInfo(const char* formatString, ...)
```
pmj: These 3 functions, `LogInfo`, `LogWarning`, and `LogError` are extremely similar. Surely we can put just about everything (apart from the `va_start` and `va_end` calls) into a helper function?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -565,6 +580,7 @@ PrjFS_Result PrjFS_DeleteFile(
     struct stat path_stat;
     if (0 != stat(fullPath, &path_stat))
     {
+        LogWarning("PrjFS_DeleteFile: stat call failed fullPath=%s errno=%d", fullPath, errno);
```
pmj: As a convenience, we might also want to log `strerror(errno)` where we log `errno` so we don't have to manually look up numeric error constants.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -554,6 +568,7 @@ PrjFS_Result PrjFS_DeleteFile(
     
     if (nullptr == relativePath)
     {
+        LogWarning("PrjFS_DeleteFile: relativePath is null");
```
pmj: Is there some context we could log here? The absolute path and expected prefix perhaps?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -383,6 +389,7 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
     
     if (nullptr == relativePath)
     {
+        LogWarning("PrjFS_WritePlaceholderFile: relative path is null %s ", relativePath);
```
pmj: I think you might have meant to log the absolute path here, not the relative path which we've just established is null?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -935,12 +958,14 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
...
         if (result != PrjFS_Result_Success)
         {
+            LogWarning("HandleRecursivelyEnumerateDirectoryRequest: HandleEnumerateDirectoryRequest failed on %s", path);
             goto CleanupAndReturn;
         }
         
         DIR* directory = opendir(path);
         if (nullptr == directory)
         {
+            LogWarning("HandleRecursivelyEnumerateDirectoryRequest: failed to open dir %s", path);
```
pmj: It might help to log the specific error (`errno`/`strerror(errno)`) here and in a few of the other new logging statements.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: On the GVFS side we did separate these out (RelatedWarning/RelatedInfo/RelatedError), so I kept that consistent.  

We could switch a general function Log(ErrorLevel level, const char* formatString, …)

I err'd on the side of consistency here.  I think the boilerplate is straightforward, and I also totally see why you called this out.

@wilbaker

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: At this point all we have is `relativePath`.  I've made the message a bit clearer though.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: This is an exellent point!  I've tried to update all places that could use this.  Please let me know if I missed one.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: I cleaned up the message here.  Unfortunately all we have here is the relative path.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I think it makes sense to have a common helper function (naming suggestion: `LogEvent`) that can be called by all three `Log` functions and takes an `EventLevel` parameter.  

We could also have the provider implement a single `LogEvent` callback rather than implementing 3.

pmj: > I err'd on the side of consistency here. I think the boilerplate is straightforward, and I also totally see why you called this out.

I don't know, 3 instances seems a good threshold to start applying DRY. And even if the backend of where this logging all ends up does repeat itself, we can plug into that with a simple `switch` block or whatever.

And in terms of consistency, the Kext's `KextLog_Error`, `KextLog_Info`, and `KextLog` are implemented as macros which expand to calls to `KextLog_Printf()` with `KEXTLOG_ERROR`, `KEXTLOG_INFO`, and `KEXTLOG_DEFAULT` as the first argument. So the user space native code would be consistent with that at least.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -383,6 +390,7 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
     
     if (nullptr == relativePath)
     {
+        LogWarning("PrjFS_WritePlaceholderFile: paramater relativePath should not be [NULL]");
```
pmj: I've now looked at the context of this function a bit more and I'm thinking that at least for this particular case of `relativePath` being `nullptr`, this seems like a programmer error (misuse of the API) which should just bring down the process (via `assert()`) rather than a true runtime error that should be logged.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -462,6 +473,7 @@ PrjFS_Result PrjFS_WriteSymLink(
     
     if (nullptr == relativePath || nullptr == symLinkTarget)
     {
+        LogWarning("PrjFS_WriteSymLink: relativePath=%s or symLinkTarget=%s is null", relativePath, symLinkTarget);
```
pmj: Similarly to `PrjFS_WritePlaceholderFile`, I'm thinking we should just `assert` that these are non-null?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: We generally try to avoid exiting unless we are in an unrecoverable bad state.

Up the [call stack](https://github.com/microsoft/VFSForGit/blob/aff5bb1f304afe326d218ccefc4b27c4e6795ae9/GVFS/GVFS.Platform.Mac/MacFileSystemVirtualizer.cs#L635)

We log this as an error which will go to telemetry.  We monitor that telemetry and can act accordingly.  Since at a higher level this is logged as an Error I think it's good to keep this as a Warning here to avoid double entries in telemetry.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: See previous comment.  We try to only exit the process if we are totally hosed.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -193,14 +195,18 @@ PrjFS_Result PrjFS_StartVirtualizationInstance(
...
         nullptr == callbacks.EnumerateDirectory ||
         nullptr == callbacks.GetFileStream ||
         nullptr == callbacks.NotifyOperation ||
-        nullptr == callbacks.LogError)
+        nullptr == callbacks.LogError ||
+        nullptr == callbacks.LogWarning ||
+        nullptr == callbacks.LogInfo)
     {
         return PrjFS_Result_EInvalidArgs;
     }
```
wilbaker: There are several error conditions in `PrjFS_StartVirtualizationInstance` with no logging where it would be helpful.

Here are two examples, could you add logging here and anywhere else that errors would cause `PrjFS_StartVirtualizationInstance` to fail?

```
    DataQueueResources dataQueue;
    s_messageQueueDispatchQueue = dispatch_queue_create("PrjFS Kernel Message Handling", DISPATCH_QUEUE_SERIAL);
    if (!PrjFSService_DataQueueInit(&dataQueue, s_kernelServiceConnection, ProviderPortType_MessageQueue, ProviderMemoryType_MessageQueue, s_messageQueueDispatchQueue))
    {
        cerr << "Failed to set up shared data queue.\n";
        return PrjFS_Result_EInvalidOperation;
    }
    
    s_virtualizationRootFullPath = virtualizationRootFullPath;
    s_callbacks = callbacks;
    
    errno_t error = RegisterVirtualizationRootPath(virtualizationRootFullPath);
    if (error != 0)
    {
        cerr << "Registering virtualization root failed: " << error << ", " << strerror(error) << endl;
        return PrjFS_Result_EInvalidOperation;
    }
```

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -424,13 +433,15 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
             &fileXattrData,
             PrjFSFileXAttrName))
     {
+        LogWarning("PrjFS_WritePlaceholderFile: failed to initialize placeholder for %s ", fullPath);
```
wilbaker: I noticed several other spots where we're checking `InitializeEmptyPlaceholder` and not logging anything if it fails.  Could you update those spots as well?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: Log these as errors?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -554,6 +569,7 @@ PrjFS_Result PrjFS_DeleteFile(
     
     if (nullptr == relativePath)
     {
+        LogWarning("PrjFS_DeleteFile: paramater relativePath should not be [NULL]");
```
wilbaker: Anything that indicates a programmer error like this should, at a minimum, be logged as an error.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -565,6 +581,7 @@ PrjFS_Result PrjFS_DeleteFile(
     struct stat path_stat;
     if (0 != stat(fullPath, &path_stat))
     {
+        LogWarning("PrjFS_DeleteFile: stat call failed fullPath=%s errno=%d strerror=%s", fullPath, errno, strerror(errno));
```
wilbaker: This is not always an error or unexpected, and so we should not always log at this point.  

In the case of `ENOENT` and `ENOTDIR` we return `PrjFS_Result_Success` (see below).

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -579,6 +596,7 @@ PrjFS_Result PrjFS_DeleteFile(
     {
         // Only files and directories can be deleted with PrjFS_DeleteFile
         // Anything else should be treated as a full file
+        LogWarning("PrjFS_DeleteFile: failing due to full file fullPath=%s", fullPath);
```
wilbaker: - Also log the values of `S_ISREG` and  `S_ISDIR`
- The message `"due to full file"` is a bit misleading, this check is unrelated to whether the file is full, it's checking if the path is a regular file or directory.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -589,13 +607,15 @@ PrjFS_Result PrjFS_DeleteFile(
...
         {
+            LogWarning("PrjFS_DeleteFile: failing because we were unable to get PrjFSFileXAttrName fullPath=%s", fullPath);
             *failureCause = PrjFS_UpdateFailureCause_FullFile;
             return PrjFS_Result_EVirtualizationInvalidOperation;
         }
     }
     
     if (0 != remove(fullPath))
     {
+        LogWarning("PrjFS_DeleteFile: remove failed fullPath=%s errno=%d strerror=%s", fullPath, errno, strerror(errno));
```
wilbaker: Similar to comment above, we should not log this message in the cases where the function would return `PrjFS_Result_Success`.

I think we could get by only logging this in the `PrjFS_Result_EIOError` case

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1423,6 +1460,7 @@ static PrjFS_Result RecursivelyMarkAllChildrenAsInRoot(const char* fullDirectory
                 CombinePaths(fullDirectoryPath, relativePath, fullPath);
                 if (!SetBitInFileFlags(fullPath, FileFlags_IsInVirtualizationRoot, true))
                 {
+                    LogWarning("RecursivelyMarkAllChildrenAsInRoot: FileFlags_IsInVirtualizationRoot not set for fullPath=%s", fullPath);
```
wilbaker: Minor: 

"FileFlags_IsInVirtualizationRoot not set for " could be interpreted as meaning that `FileFlags_IsInVirtualizationRoot` was not set when the code expected it to be set.

Maybe change this to something like : "SetBitInFileFlags failed to set FileFlags_IsInVirtualizationRoot"?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: > Log these as errors?
If we do that we'll get it thru telemetry twice since it's handled as an error farther up.  We can discuss more later today.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: As discussed we logged inside InitializeEmptyPlaceholder to capture all places.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -339,6 +346,7 @@ PrjFS_Result PrjFS_WritePlaceholderDirectory(
 
     if (mkdir(fullPath, 0777))
     {
+        LogWarning("PrjFS_WritePlaceholderDirectory: mkdir failed errno=%d stderror=%s", errno, strerror(errno));
```
wilbaker: I just did some more digging and found that the `ENOENT` case is actually expected sometimes, see:

https://github.com/microsoft/VFSForGit/blob/aff5bb1f304afe326d218ccefc4b27c4e6795ae9/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs#L1549

Could you move this `LogWarning` down to the `default` case?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -398,6 +406,7 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
     FILE* file = fopen(fullPath, "wx");
     if (nullptr == file)
     {
+        LogWarning("PrjFS_WritePlaceholderFile: fopen failed errorno=%d stderror=%s", errno, strerror(errno));
```
wilbaker: Same comment here.  It looks like the `ENOENT` case is expected, could you move the `LogWarning` down to the `default` case below?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
wilbaker: I discussed this offline with @jeschu1 and agreed that we can leave this as-is for now.  There's not currently a lot of code duplication that can be removed with a common helper, and we can always revisit this later if it grows any larger.

PR #1445, opened 2019-08-13
Check for new Panic Logs where PrjFSKextLogDaemon is started

### LINKED ISSUE #647: Mac: Detect\track\report ProjFS related kernel panics
Suggestion from @pmj:

> A good start would be to scan /Library/Logs/Kernel*.panic for any with PrjFS mentioned in “Kernel Extensions in backtrace”

This might be a good task for PrjFSLoggingDaemon or GVFS.Service.

In addition to detecting the panics alerts should fire if panics are found.

jeschu1:
- A time file is used to differentiate the last sweep
- The name of the panic file is sent to telemetry

resolves #647

### DIFF
--- a/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
+++ b/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
@@ -3,6 +3,7 @@
 #include "../PrjFSLib/Json/JsonWriter.hpp"
 #include "../PrjFSLib/PrjFSUser.hpp"
 #include <atomic>
+#include <dirent.h>
 #include <iostream>
 #include <mutex>
 #include <sstream>
@@ -10,7 +11,9 @@
 #include <IOKit/IOKitLib.h>
 #include <signal.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
 #include <sys/un.h>
+#include <unistd.h>
 
 using std::atomic_exchange;
 using std::atomic_uint32_t;
@@ -21,6 +24,9 @@ using std::ostringstream;
 using std::string;
 
 static const char PrjFSKextLogDaemon_OSLogSubsystem[] = "org.vfsforgit.prjfs.PrjFSKextLogDaemon";
+static const char PanicLogDirectory[] = "/Library/Logs/DiagnosticReports";
+static const char PanicLogTimestampDirectory[] = "/usr/local/vfsforgit/diagnostics";
+static const char PanicLogTimestampFile[] = "/usr/local/vfsforgit/diagnostics/LastRun.txt";
 static const int INVALID_SOCKET_FD = -1;
 
 static os_log_t s_daemonLogger, s_kextLogger;
@@ -40,6 +46,9 @@ static mutex s_messageListenerMutex;
 
 static atomic_uint32_t s_droppedMessageCount(0);
 
+static void LogPanics();
+static long GetFileModifiedTime(const char* path);
+static bool DoesPathExist(const char* path);
 static void StartLoggingKextMessages(io_connect_t connection, io_service_t service);
 static void HandleSigterm(int sig, siginfo_t* info, void* uc);
 static void SetupExitSignalHandler();
@@ -127,6 +136,7 @@ int main(int argc, const char* argv[])
 
     os_log(s_daemonLogger, "PrjFSKextLogDaemon running");
 
+    LogPanics();
     CFRunLoopRun();
 
     os_log(s_daemonLogger, "PrjFSKextLogDaemon shutting down");
@@ -153,6 +163,117 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
     default:
         return OS_LOG_TYPE_ERROR;
     }
+}
+
+static void LogPanics()
+{
+    DIR* dr = opendir(PanicLogDirectory);
+    if (dr == nullptr)
+    {
+        ostringstream errorMessage;
+        errorMessage << "Unable to open " << PanicLogDirectory << ", errno=" << errno << ", errorstr=" << strerror(errno);
+        LogDaemonError(errorMessage.str());
+        return;
+    }
+
+    // Look up the last run time
+    long lastRunTimestamp = GetFileModifiedTime(PanicLogTimestampFile);
+
+    struct dirent* dirEntry;
+    while ((dirEntry = readdir(dr)) != nullptr)
+    {
+        // Only check files
+        if(dirEntry->d_type == DT_REG)
+        {
+           // Check for .panic extension
+           const char* ext = strrchr(dirEntry->d_name,'.');
+           if(ext != nullptr && strcmp(ext, ".panic") == 0)
+           {
+               // Get the full name of the panic file
+               string fullFileName = PanicLogDirectory;
+               fullFileName += "/";
+               fullFileName += dirEntry->d_name;
+
+               // Verify it's after the last timestamp
+               long panicLogTime = GetFileModifiedTime(fullFileName.c_str());
+               if (panicLogTime == 0)
+               {
+                   // Record error if we can't determine timestamp
+                   os_log(s_daemonLogger, "Unable to run stat on panic at %s errno=%d strerror=%s\n", fullFileName.c_str(), errno, strerror(errno));
+                   continue;
+               }
+
+               if (panicLogTime >= lastRunTimestamp)
+               {
+                  // Check if 'org.vfsforgit.PrjFSKext(' occurs in the log
+                  // If it does, then PrjFSKext may have caused the panic and we should log an error
+                  FILE* fp = fopen(fullFileName.c_str(), "r");
+                  if (fp != nullptr)
+                  {
+                      char* line = nullptr;
+                      size_t len = 0;
+                      ssize_t read;
+                      while ((read = getline(&line, &len, fp)) != -1)
+                      {
+                          if (strstr(line, "org.vfsforgit.PrjFSKext(") != nullptr)
+                          {
+                              ostringstream errorMessage;
+                              errorMessage << "Found panic at " << fullFileName.c_str() << "\n";
+                              LogDaemonError(errorMessage.str());
+                              break;
+                          }
+                      }
+                      
+                      fclose(fp);
+                      free(line);
+                  }
+                  else
+                  {
+                      os_log(s_daemonLogger, "Unable open %s errno=%d strerror=%s\n", fullFileName.c_str(), errno, strerror(errno));
+                      continue;
+                  }
+              }
+           }
+        }
+    }
+    closedir(dr);
+
+    // Update timestamp file
+    if (!DoesPathExist(PanicLogTimestampDirectory))
+    {
+        mkdir(PanicLogTimestampDirectory, 0755);
+    }
+    
+    FILE* fp = fopen(PanicLogTimestampFile, "a");
+    if (fp != nullptr)
+    {
+       fprintf(fp, "Completed Panic Sweep for :%ld\n", lastRunTimestamp);
+       fclose(fp);
+    }
+    else
+    {
+        os_log(s_daemonLogger, "Unable to save panic timestamp file %s errno=%d strerror=%s\n", PanicLogTimestampFile, errno, strerror(errno));
+    }
+}
+
+static long GetFileModifiedTime(const char* path)
+{
+    struct stat attr;
+    if (stat(path, &attr) == -1)
+    {
+        return 0;
+    }
+    return attr.st_mtimespec.tv_sec;
+}
+
+static bool DoesPathExist(const char* path)
+{
+    struct stat attr;
+    if (stat(path, &attr) == -1)
+    {
+        return false;
+    }
+    return true;
 }
 
 static void StartLoggingKextMessages(io_connect_t connection, io_service_t prjfsService)


ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -21,6 +24,9 @@ using std::ostringstream;
 using std::string;
 
 static const char PrjFSKextLogDaemon_OSLogSubsystem[] = "org.vfsforgit.prjfs.PrjFSKextLogDaemon";
+static const char PanicLogDirectory[] = "/Library/Logs/DiagnosticReports";
+static const char PanicLogTimestampDirectory[] = "/Library/Logs/DiagnosticReports/VFS";
```
pmj: I have to admit I don't know whether `/Library/Logs/DiagnosticReports` is considered a place where 3rd party software is supposed to place state. Did you look into this at all? I could imagine Apple might have designated it as a location the OS is free to delete at will - admittedly this shouldn't have any adverse side effects for this particular purpose.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
     }
 }
 
+static void LogPanics()
+{
+    DIR *dr = opendir(PanicLogDirectory);
```
pmj: We've been using C++-style pointer asterisk placement so far:
```
    DIR* dr = opendir(PanicLogDirectory);
```

Any reason to deviate from this here?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    if (dr == NULL)
```
pmj: Similarly, the modern C++ way to write this is

```
    if (dr == nullptr)
```

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+
+    struct dirent *dirEntry;
+    while ((dirEntry = readdir(dr)) != NULL)
+    {
+        // Only check files
+        if(dirEntry->d_type == DT_REG)
+        {
+           // Check for .panic extension
+           const char *ext = strrchr(dirEntry->d_name,'.');
+           if(ext && strcmp(ext, ".panic") == 0)
```
pmj: Consider making the `nullptr` comparison for `ext` explicit.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+                      size_t len = 0;
+                      ssize_t read;
+                      while ((read = getline(&line, &len, fp)) != -1)
+                      {
+                          if (strstr(line, "org.vfsforgit.PrjFSKext(") != NULL)
+                          {
+                              ostringstream errorMessage;
+                              errorMessage << "Found panic at " << fullFileName.c_str() << "\n";
+                              LogDaemonError(errorMessage.str());
+                          }
```
pmj: I'm not sure if it's possible for the string `org.vfsforgit.PrjFSKext(` to appear multiple times in a file; if it does, this will generate multiple log entries. We could `break;` out of the loop as soon as we fine at least one occurrence.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+              }
+           }
+        }
+    }
+    closedir(dr);
+
+    // Update timestamp file
+    if (!DoesPathExist(PanicLogTimestampDirectory))
+    {
+        mkdir(PanicLogTimestampDirectory, 0777);
```
pmj: Any reason to make this world-writable?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    }
+    
+    FILE *fp = fopen(PanicLogTimestampFile, "aw+");
```
pmj: I'm not familiar with the mode `aw` and can't work out what it means from the manpage. Did you mean `a` for appending or `w` for overwriting? Also, the `+` additionally opens the file for reading, but there are no subsequent reads from the file.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: This is a very good point!  Thanks @pmj!  As well as thank you for the nits on style.  I'll definitely fix!

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: > I don't know whether /Library/Logs/DiagnosticReports is considered a place where 3rd party software is supposed to place state.

I agree, we should track state inside the directories that our applications own.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Same comment here on the location of '*' as well: `FILE *fp` -> `FILE* fp`

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    }
+
+    // Look up the last run time
+    long lastRunTimestamp = 0;
+    if (DoesPathExist(PanicLogTimestampFile))
+    {
+        lastRunTimestamp = GetFileModifiedTime(PanicLogTimestampFile);
+    }
+
+    struct dirent *dirEntry;
```
wilbaker: Nit: `dirent *dirEntry` -> `dirent* dirEntry`

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: There are several pointer declarations like this, could you update all of them?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: Similar comment elsewhere, all `NULL`s should be replaced with `nullptr`

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,123 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    if (dr == NULL)
+    {
+        ostringstream errorMessage;
+        errorMessage << "Unable to open " << PanicLogDirectory << ", errno=" << errno << ", errorstr=" << strerror(errno);
+        LogDaemonError(errorMessage.str());
+    }
+
+    // Look up the last run time
+    long lastRunTimestamp = 0;
+    if (DoesPathExist(PanicLogTimestampFile))
```
wilbaker: There's no need to call `DoesPathExist` here as `GetFileModifiedTime` will return 0 if the file does not exist.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: @wilbaker do you have a good suggestion for a location here?  Happy to move it.

wilbaker: How about a new "diagnostics" folder here:

`/usr/local/vfsforgit/diagnostics`

cc: @nickgra, @alameenshah for their thoughts as well

nickgra: I like that location, we should always have permissions to write to it.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: By default all of our C# directories are 0777 (Directory.CreateDirectory default perms).  It's also the same as the Placeholder dir.  Considering what the directory, I don't think we need to limit access.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: Just to check, this dir looks like we have to be 'sudo' to create. I think that's fine as long as PrjFSKextDaemon runs as root by default.

wilbaker: > as long as PrjFSKextDaemon runs as root by default

I think that should be the case, cc: @alameenshah to confirm.

jeschu1: @alameenshah and I looked together and it does run as root.  Yay :-)

@wilbaker this is ready for another look if you don't mind

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,120 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+
+               // Verify it's after the last timestamp
+               long panicLogTime = GetFileModifiedTime(fullFileName.c_str());
+               if (panicLogTime == 0)
+               {
+                   // Record error if we can't determine timestamp
+                   ostringstream errorMessage;
+                   errorMessage << "Unable to run stat on panic at " << fullFileName.c_str() << "\n";
+                   const char* errorMessageString = errorMessage.str().c_str();
+                   LogKextMessage(OS_LOG_TYPE_DEFAULT, 0, errorMessageString, static_cast<int>(strlen(errorMessageString)));
```
wilbaker: `LogKextMessage` is intended for messages from the kext.

Because this message is originating from the logging daemon's logic, we should use `s_daemonLogger` (search for `os_log(s_daemonLogger` for examples).

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,120 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    {
+       fprintf(fp, "Completed Panic Sweep for :%ld\n", lastRunTimestamp);
+       fclose(fp);
+    }
+    else
+    {
+        ostringstream errorMessage;
+        errorMessage << "Unable to save panic timestamp file " << PanicLogTimestampFile << "\n";
+        const char* errorMessageString = errorMessage.str().c_str();
+        LogKextMessage(OS_LOG_TYPE_DEFAULT, 0, errorMessageString, static_cast<int>(strlen(errorMessageString)));
```
wilbaker: Same comment here, `os_log(s_daemonLogger` instead of `LogKextMessage`.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,120 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+                   const char* errorMessageString = errorMessage.str().c_str();
+                   LogKextMessage(OS_LOG_TYPE_DEFAULT, 0, errorMessageString, static_cast<int>(strlen(errorMessageString)));
+                   continue;
+               }
+
+               if (panicLogTime >= lastRunTimestamp)
+               {
+                  // Verify 'org.vfsforgit.PrjFSKext(' occurs in the log
+                  FILE* fp = fopen(fullFileName.c_str(), "r");
+                  if (fp)
```
wilbaker: Log if the `fopen` fails?

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,120 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
 }
 
+static void LogPanics()
+{
+    DIR* dr = opendir(PanicLogDirectory);
+    if (dr == nullptr)
+    {
+        ostringstream errorMessage;
+        errorMessage << "Unable to open " << PanicLogDirectory << ", errno=" << errno << ", errorstr=" << strerror(errno);
+        LogDaemonError(errorMessage.str());
```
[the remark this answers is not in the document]
jeschu1: PanicLogDirectory should always exist.  This is where the panic log files are.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,120 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    }
+
+    // Look up the last run time
+    long lastRunTimestamp = GetFileModifiedTime(PanicLogTimestampFile);
```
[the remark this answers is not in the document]
jeschu1: Updated to return if dr is null, great point!

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
wilbaker: @jeschu1 what are you thoughts here, do you think we should log if `fopen` fails?

pmj: Probably. It's extremely unlikely that it will fail, but if it does, it suggests the system is in a bad state (overzealous security software, corrupted file system, hardware issues, etc.) so any VFSForGit errors are to be taken with a grain of salt.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,116 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    closedir(dr);
+
+    // Update timestamp file
+    if (!DoesPathExist(PanicLogTimestampDirectory))
+    {
+        mkdir(PanicLogTimestampDirectory, 0755);
+    }
+    
+    FILE* fp = fopen(PanicLogTimestampFile, "a");
+    if (fp)
```
pmj: Nit: `if (fp != nullptr)`

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,116 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+    {
+       fprintf(fp, "Completed Panic Sweep for :%ld\n", lastRunTimestamp);
+       fclose(fp);
+    }
+    else
+    {
+        os_log(s_daemonLogger, "Unable to save panic timestamp file %s errno=%d strerror=%s\n", PanicLogTimestampDirectory, errno, strerror(errno));
```
pmj: Typo? `PanicLogTimestampFile` rather than `PanicLogTimestampDirectory`.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
pmj: Also, nit: `if (fp != nullptr)`

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -155,6 +165,116 @@ static os_log_type_t KextLogLevelAsOSLogType(KextLog_Level level)
...
+                              ostringstream errorMessage;
+                              errorMessage << "Found panic at " << fullFileName.c_str() << "\n";
+                              LogDaemonError(errorMessage.str());
+                              break;
+                          }
+                      }
+                      
+                      fclose(fp);
+                   
+                      if (line)
```
pmj: Nit: Either `if (line != nullptr)` or even better, simply call `free(line);` without the `if` - `free(nullptr);` has well-defined behaviour.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: I definitely agree @wilbaker @pmj, I'm adding a guard against fopen failing.

ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
jeschu1: Exellent catch here!

PR #1450, opened 2019-08-14
Add information for when lock is release by process not owning the lock

kewillford:
There are cases when the lock is being released by a process that does not own the lock.  This is to add some additional log information so that we can better diagnose what is happening when we hit this error.

### DIFF
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -268,7 +268,33 @@ namespace GVFS.Virtualization.Projection
                 }
             }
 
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -268,7 +268,32 @@ public NamedPipeMessages.ReleaseLock.Response TryReleaseExternalLock(int pid)
...
+            }
+
+            if (externalHolder != null)
+            {
+                metadata.Add("ExternalHolder", externalHolder.ParsedCommand);
+                metadata.Add("ExternalHolderPID", externalHolder.PID);
+            }
+            else
+            {
+                metadata.Add("ExternalHolder", string.Empty);
```
wilbaker: Would it be useful to log if VFS4G itself holds the lock at this point?

PR #1452, opened 2019-08-15
Correct Upgrade Messages for Mac

### LINKED ISSUE #1442: Mac: `gvfs upgrade` prompt text should be updated for macOS
We currently display upgrade notifications like so:
```
>gvfs upgrade
New version 0.6.19224.2 is available.

When ready, run `gvfs upgrade --confirm` from an elevated command prompt.
```

We should update the last sentence to say:
```When ready, run `sudo gvfs upgrade --confirm` to upgrade.```

### PR DESCRIPTION
**Pull request by jeschu1**
I was hoping this was an easy change, but it ended up being a bit involved.
There were several places that required string changes, including the hooks.

resolves #1442

### DIFF
--- a/GVFS/GVFS.Common/GVFSPlatform.cs
+++ b/GVFS/GVFS.Common/GVFSPlatform.cs
@@ -149,6 +149,10 @@ namespace GVFS.Common
             /// the upgrade verb is running.
             /// </summary>
             public abstract bool SupportsUpgradeWhileRunning { get; }
+            public abstract string UpgradeInstallAdviceMessage { get; }
+            public abstract string RunUpdateMessage { get; }
+            public abstract string UpgradeConfirmCommandMessage { get; }
+            public abstract string StartServiceCommandMessage { get; }
             public abstract string WorkingDirectoryBackingRootPath { get; }
             public abstract string DotGVFSRoot { get; }
 
--- a/GVFS/GVFS.Common/InstallerPreRunChecker.cs
+++ b/GVFS/GVFS.Common/InstallerPreRunChecker.cs
@@ -177,14 +177,14 @@ namespace GVFS.Upgrader
 
         private bool IsGVFSUpgradeAllowed(out string consoleError)
         {
-            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm, StringComparison.OrdinalIgnoreCase);
+            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSPlatform.Instance.Constants.UpgradeConfirmCommandMessage, StringComparison.OrdinalIgnoreCase);
             string adviceText = null;
             if (!this.IsElevated())
             {
-                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = GVFSPlatform.Instance.Constants.RunUpdateMessage;
                 consoleError = string.Join(
                     Environment.NewLine,
-                    "The installer needs to be run from an elevated command prompt.",
+                    "The installer needs to be run with elevated permissions.",
                     adviceText);
                 this.tracer.RelatedWarning($"{nameof(this.IsGVFSUpgradeAllowed)}: Upgrade is not installable. {consoleError}");
                 return false;
@@ -202,7 +202,7 @@ namespace GVFS.Upgrader
 
             if (this.IsServiceInstalledAndNotRunning())
             {
-                adviceText = isConfirmed ? $"Run `sc start GVFS.Service` and run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run `sc start GVFS.Service` and run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = $"To install, run {GVFSPlatform.Instance.Constants.StartServiceCommandMessage} and run {GVFSPlatform.Instance.Constants.UpgradeConfirmCommandMessage}.";
                 consoleError = string.Join(
                     Environment.NewLine,
                     "GVFS Service is not running.",
--- a/GVFS/GVFS.Platform.Mac/MacPlatform.cs
+++ b/GVFS/GVFS.Platform.Mac/MacPlatform.cs
@@ -214,6 +214,26 @@ namespace GVFS.Platform.Mac
 
             // Documented here (in the addressing section): https://www.unix.com/man-page/mojave/4/unix/
             public override int MaxPipePathLength => 104;
+
+            public override string UpgradeInstallAdviceMessage
+            {
+                get { return $"When ready, run {this.UpgradeConfirmCommandMessage} to upgrade."; }
+            }
+
+            public override string UpgradeConfirmCommandMessage
+            {
+                get { return UpgradeConfirmMessage; }
+            }
+
+            public override string StartServiceCommandMessage
+            {
+                get { return "`launchctl load /Library/LaunchAgents/org.vfsforgit.service.plist`"; }
+            }
+
+            public override string RunUpdateMessage
+            {
+                get { return $"Run {UpgradeConfirmMessage}."; }
+            }
         }
     }
 }
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs
@@ -11,6 +11,7 @@ namespace GVFS.Platform.Windows
     public partial class WindowsPlatform
     {
         public const string DotGVFSRoot = ".gvfs";
+        public const string UpgradeConfirmMessage = "`gvfs upgrade --confirm`";
 
         private const int StillActive = 259; /* from Win32 STILL_ACTIVE */
 
@@ -119,6 +120,11 @@ namespace GVFS.Platform.Windows
         public static string GetUpgradeHighestAvailableVersionDirectoryImplementation()
         {
             return GetUpgradeProtectedDataDirectoryImplementation();
+        }
+
+        public static string GetUpgradeReminderNotificationImplementation()
+        {
+            return $"A new version of VFS for Git is available. Run {UpgradeConfirmMessage} from an elevated command prompt to upgrade.";
         }
 
         [DllImport("kernel32.dll")]
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -499,6 +499,26 @@ namespace GVFS.Platform.Windows
 
             // Tests show that 250 is the max supported pipe name length
             public override int MaxPipePathLength => 250;
+
+            public override string UpgradeInstallAdviceMessage
+            {
+                get { return $"When ready, run {this.UpgradeConfirmCommandMessage} from an elevated command prompt."; }
+            }
+
+            public override string UpgradeConfirmCommandMessage
+            {
+                get { return UpgradeConfirmMessage; }
+            }
+
+            public override string StartServiceCommandMessage
+            {
+                get { return $"`sc start GVFS.Service`";  }
+            }
+
+            public override string RunUpdateMessage
+            {
+                get { return $"Run {UpgradeConfirmMessage} from an elevated command prompt."; }
+            }
         }
     }
 }
--- a/GVFS/GVFS.Common/GVFSConstants.cs
+++ b/GVFS/GVFS.Common/GVFSConstants.cs
@@ -239,16 +239,13 @@ namespace GVFS.Common
         public static class UpgradeVerbMessages
         {
             public const string GVFSUpgrade = "`gvfs upgrade`";
-            public const string GVFSUpgradeConfirm = "`gvfs upgrade --confirm`";
             public const string GVFSUpgradeDryRun = "`gvfs upgrade --dry-run`";
             public const string NoUpgradeCheckPerformed = "No upgrade check was performed.";
             public const string NoneRingConsoleAlert = "Upgrade ring set to \"None\". " + NoUpgradeCheckPerformed;
             public const string NoRingConfigConsoleAlert = "Upgrade ring is not set. " + NoUpgradeCheckPerformed;
             public const string InvalidRingConsoleAlert = "Upgrade ring set to unknown value. " + NoUpgradeCheckPerformed;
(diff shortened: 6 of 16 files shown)

### CODE REVIEW

File: GVFS/GVFS.Platform.Mac/MacPlatform.cs [hunk 1]
```diff
@@ -214,6 +214,26 @@ public override string GVFSBinDirectoryName
...
+            }
+
+            public override string UpgradeConfirmCommand
+            {
+                get { return "`sudo gvfs upgrade --confirm`"; }
+            }
+
+            public override string StartServiceCommand
+            {
+                get { return "`launchctl start org.vfsforgit.service`"; }
```
jeschu1: @alameenshah / @jamill can you confirm this is the correct command.  I grabbed this from the start scripts.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -177,11 +177,11 @@ protected virtual bool TryRunGVFSWithArgs(string args, out string consoleError)
...
-            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm, StringComparison.OrdinalIgnoreCase);
+            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSPlatform.Instance.Constants.UpgradeConfirmCommand, StringComparison.OrdinalIgnoreCase);
             string adviceText = null;
             if (!this.IsElevated())
             {
-                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSPlatform.Instance.Constants.UpgradeConfirmCommand} from an elevated command prompt.";
                 consoleError = string.Join(
                     Environment.NewLine,
                     "The installer needs to be run from an elevated command prompt.",
```
jeschu1: IsElevated is only 'false' for windows so I didn't bother updating the string above.

GVFS/GVFS.Common/GVFSPlatform.cs
```diff
@@ -149,6 +149,10 @@ public abstract class GVFSPlatformConstants
             /// the upgrade verb is running.
             /// </summary>
             public abstract bool SupportsUpgradeWhileRunning { get; }
+            public abstract string UpgradeInstallAdvice { get; }
+            public abstract string ElevatedCommandPrompt { get; }
+            public abstract string UpgradeConfirmCommand { get; }
+            public abstract string StartServiceCommand { get; }
```
wilbaker: Naming suggestion:  Add "Message" to the end of each of these new property names.  

Previously they were in `UpgradeVerbMessages` (which made it clear these are user facing messages) but now glancing at the code someone might think `StartServiceCommand` is something VFS4G could call.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -499,6 +499,26 @@ public override HashSet<string> UpgradeBlockingProcesses
...
             public override int MaxPipePathLength => 250;
+
+            public override string UpgradeInstallAdvice
+            {
+                get { return "When ready, run `" + this.UpgradeConfirmCommand + "` from an elevated command prompt."; }
+            }
+
+            public override string UpgradeConfirmCommand
+            {
+                get { return "`gvfs upgrade --confirm`"; }
```
wilbaker: If you moved `gvfs upgrade --confirm` into a string constant in `WindowsPlatform.Shared.cs ` you could use it in `GetUpgradeReminderNotificationImplementation` as well:

        public static string GetUpgradeReminderNotificationImplementation()
        {
            return "A new version of GVFS is available. Run `gvfs upgrade --confirm` from an elevated command prompt to upgrade.";
        }

GVFS/GVFS.Platform.Mac/MacPlatform.cs
```diff
@@ -214,6 +214,26 @@ public override string GVFSBinDirectoryName
...
             public override int MaxPipePathLength => 104;
+
+            public override string UpgradeInstallAdvice
+            {
+                get { return "When ready, run " + this.UpgradeConfirmCommand + " to upgrade."; }
+            }
+
+            public override string UpgradeConfirmCommand
+            {
+                get { return "`sudo gvfs upgrade --confirm`"; }
```
wilbaker: Same comment here as in WindowsPlatform.  If this string were moved to a constant in the Shared.cs file, you could avoid duplicating it in `GetUpgradeReminderNotificationImplementation`

GVFS/GVFS.Common/InstallerPreRunChecker.cs
wilbaker: How can you tell it's only `false` for Windows?

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -177,11 +177,11 @@ protected virtual bool TryRunGVFSWithArgs(string args, out string consoleError)
 
         private bool IsGVFSUpgradeAllowed(out string consoleError)
         {
-            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm, StringComparison.OrdinalIgnoreCase);
+            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSPlatform.Instance.Constants.UpgradeConfirmCommand, StringComparison.OrdinalIgnoreCase);
             string adviceText = null;
             if (!this.IsElevated())
             {
-                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSPlatform.Instance.Constants.UpgradeConfirmCommand} from an elevated command prompt.";
```
wilbaker: Should ` from an elevated command prompt` be `{GVFSPlatform.Instance.Constants.ElevatedCommandPrompt}`?

GVFS/GVFS.Common/InstallerPreRunChecker.cs
jeschu1: I take that back, these need to get updated to, i'll do that

jeschu1:
@wilbaker thanks for the feedback, it should be addressed now.

jamill:
Did we consider if we could simplify / streamline any of the messages? To keep the _exact_ structure as we had previously, we now have to build the messages from multiple fragments, some of which start with a space, some don't, etc.. This makes it harder to determine the exact message the user is going to see and ensure it is correct.

GVFS/GVFS.Platform.Mac/MacPlatform.cs
jeschu1: Spoke with @alameenshah on this and instead we should have "launchctl load /Library/LaunchAgents/org.vfsforgit.service.plist' .  We also tested it out and it worked as expected.  Updating now.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -202,7 +202,9 @@ private bool IsGVFSUpgradeAllowed(out string consoleError)
 
             if (this.IsServiceInstalledAndNotRunning())
             {
-                adviceText = isConfirmed ? $"Run `sc start GVFS.Service` and run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run `sc start GVFS.Service` and run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = isConfirmed
```
jamill: Instead of building the sentence from multiple fragments, can we simplify this by just instructing the user to start the service and run the command again? We have already checked they are running from an elevated command prompt.

`To start the service, run {command} and run the command again`

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -177,14 +177,14 @@ protected virtual bool TryRunGVFSWithArgs(string args, out string consoleError)
 
         private bool IsGVFSUpgradeAllowed(out string consoleError)
         {
-            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm, StringComparison.OrdinalIgnoreCase);
+            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSPlatform.Instance.Constants.UpgradeConfirmCommandMessage, StringComparison.OrdinalIgnoreCase);
             string adviceText = null;
             if (!this.IsElevated())
             {
-                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
```
jamill: Can we simplify this to just instruct the user to re-run the command elevated?

`adviceText = GVFS.Platform.Instance.Constants.RunCommandElevated`, where we can say "Run the command from an elevated command prompt" or "Run the command as root (or with sudo)" on Mac...

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -499,6 +499,26 @@ public override HashSet<string> UpgradeBlockingProcesses
...
+            }
+
+            public override string StartServiceCommandMessage
+            {
+                get { return $"`sc start GVFS.Service`";  }
+            }
+
+            public override string ElevatedCommandPromptMessage
```
wilbaker: Do we still need `ElevatedCommandPromptMessage`?

If `UpgradeConfirmCommandMessage` were updated to return:

`"'gvfs upgrade --confirm'  from an elevated command prompt."`

Similar to how it returns the following on Mac:

`"'sudo gvfs upgrade --confirm'"`

Then we could probably remove both `ElevatedCommandPromptMessage` and `UpgradeInstallAdviceMessage` from the platform.

wilbaker: If we do still need `ElevatedCommandPromptMessage` can you update the name?  I think this property used to return something like "from an elevated command prompt" and now it's returning a different message.

jeschu1: `UpgradeConfirmCommandMessage` is also used in an error message for checking if the service is installed and not running, so we can't update that one.

The windows message does refer to an elevated prompt.  I'm updating it to be `RunUpdateMessage `which should be more clear.

Also as an FYI, potentially it's a bit of scope creep but @jamill and I went thru the messages and simplified what was there.  This should make life easier in the future.

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -177,14 +177,14 @@ protected virtual bool TryRunGVFSWithArgs(string args, out string consoleError)
...
+            bool isConfirmed = string.Equals(this.CommandToRerun, GVFSPlatform.Instance.Constants.UpgradeConfirmCommandMessage, StringComparison.OrdinalIgnoreCase);
             string adviceText = null;
             if (!this.IsElevated())
             {
-                adviceText = isConfirmed ? $"Run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = GVFSPlatform.Instance.Constants.RunUpdateMessage;
                 consoleError = string.Join(
                     Environment.NewLine,
-                    "The installer needs to be run from an elevated command prompt.",
+                    "The installer needs to be run elevated.",
```
derrickstolee: Nit: I had trouble parsing this. Perhaps "The installer needs to be run with elevated permissions"?

GVFS/GVFS.Platform.Windows/WindowsPlatform.Shared.cs
```diff
@@ -121,6 +122,11 @@ public static string GetUpgradeHighestAvailableVersionDirectoryImplementation()
             return GetUpgradeProtectedDataDirectoryImplementation();
         }
 
+        public static string GetUpgradeReminderNotificationImplementation()
+        {
+            return $"A new version of GVFS is available. Run {UpgradeConfirmMessage} from an elevated command prompt to upgrade.";
```
derrickstolee: GVFS -> VFS for Git?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -499,6 +499,26 @@ public override HashSet<string> UpgradeBlockingProcesses
 
             // Tests show that 250 is the max supported pipe name length
             public override int MaxPipePathLength => 250;
+
+            public override string UpgradeInstallAdviceMessage
+            {
+                get { return "When ready, run `" + this.UpgradeConfirmCommandMessage + "` from an elevated command prompt."; }
```
derrickstolee: Why not an interpolated string here? `$"When ready run `{this.UpgradeConfirmCommandMessage}`..."

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
wilbaker: @jeschu1 I unresolved this conversation.  It looks like this code can be updated to return `UpgradeConfirmMessage` rather than duplicating "'gvfs upgrade --confirm'"

GVFS/GVFS.Common/InstallerPreRunChecker.cs
```diff
@@ -202,7 +202,9 @@ private bool IsGVFSUpgradeAllowed(out string consoleError)
 
             if (this.IsServiceInstalledAndNotRunning())
             {
-                adviceText = isConfirmed ? $"Run `sc start GVFS.Service` and run {this.CommandToRerun} again from an elevated command prompt." : $"To install, run `sc start GVFS.Service` and run {GVFSConstants.UpgradeVerbMessages.GVFSUpgradeConfirm} from an elevated command prompt.";
+                adviceText = isConfirmed
+                    ? $"Run {GVFSPlatform.Instance.Constants.StartServiceCommandMessage} and run {this.CommandToRerun} again."
```
wilbaker: To we need two different messages here?

if `isConfirmed` is true, then `CommandToRerun` is the same as `UpgradeConfirmCommandMessage` and so the difference between messages is minimal:

```
 $"Run {GVFSPlatform.Instance.Constants.StartServiceCommandMessage} and run {UpgradeConfirmCommandMessage} again."
```
vs
```
$"To install, run {GVFSPlatform.Instance.Constants.StartServiceCommandMessage} and run {GVFSPlatform.Instance.Constants.UpgradeConfirmCommandMessage}."
```

PR #1456, opened 2019-08-19
Auto-unmount repos in pre-install

### LINKED ISSUE #1124: Installer should unmount first before installing
`preflight` script in the Mac installer should unmount repositories before proceeding with the install.

alameenshah:
Mac installer pkg would now automatically unmount all repositories registered for auto mount

Fixes #1124

### DIFF
--- a/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
+++ b/GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
@@ -34,7 +34,7 @@
         
         function GetBlockingProcesses()
         {
-            var watchList = [ "GVFS.Mount", "git", "gitk", "Wish" ];
+            var watchList = [ "git", "gitk", "Wish" ];
             var blockingProcesses = new Array();
             for (var process of watchList)
             {
--- a/GVFS/GVFS.Installer.Mac/scripts/preinstall
+++ b/GVFS/GVFS.Installer.Mac/scripts/preinstall
@@ -17,3 +17,10 @@ if [ "$isKextLoaded" -gt 0 ]; then
     echo $unloadCmd
     eval $unloadCmd || exit 1
 fi
+
+GVFSBINPATH="/usr/local/vfsforgit/gvfs"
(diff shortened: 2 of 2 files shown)


GVFS/GVFS.Installer.Mac/scripts/preinstall
```diff
@@ -17,3 +17,10 @@ if [ "$isKextLoaded" -gt 0 ]; then
     echo $unloadCmd
     eval $unloadCmd || exit 1
 fi
+
+GVFSBINPATH="/usr/local/vfsforgit/gvfs"
+if [ -f "${GVFSBINPATH}" ]; then
+    unmountCmd="${GVFSBINPATH} service --unmount-all"
+    echo $unmountCmd
+    eval $unmountCmd || exit 1
```
jeschu1: Just to double check will the eval print any errors that come up from the call?  
This follows the format of the other commands and lgtme.

alameenshah: yes it does.

GVFS/GVFS.Installer.Mac/scripts/Distribution.xml
```diff
@@ -34,7 +34,7 @@
         
         function GetBlockingProcesses()
```
wilbaker: Does this get called before or after `preinstall`?

alameenshah: Before `preinstall`. 

This is the order - `Distribution.xml::InstallationCheck`, `preinstall`, `postinstall`.

Ignas:
I think this will fail if you are running installer as root with a clean env (like using Munki or just sudo install) and should rather be:
```
for username in $(ps -Ac -o user,command | grep -iw "loginwindow" | awk '{print $1}'); do
    USER_HOME=$(eval echo "~$username")
    HOME="${USER_HOME}" gvfs service --unmount-all
done
```

Same as the service restart commands.

Issue #1437, opened 2019-08-09
Mac: Diagnostics for for Kext Events


jeschu1:
We found an issue when a user ran 'open project.app' where a system process was blocking hydration (and shouldn't have been).

To diagnose the issue I created a branch with logging for all the events called and ones blocked due to processId.  It took over an hour with @pmj/I working together to diagnose the underlying issue.

We want a better process for this.  Ideally a flag that can flipped that would log all events and rejections.  This would have allowed us to diagnose the problem much quicker and on the users machine.

I'd love to hear @pmj / @wilbaker  thoughts on ideas for this.  For now this is unprioritized but @jrbriggs I think pri1 or pri2 would be appropriate (I'm leaning torward pri1 depending on how complicated the solution is).

wilbaker:
@jeschu1 this sounds good to me.

@pmj I'm assigning this issue over to you.  How would you recommend that we communicate to the kext that we'd like this behavior?  I think for the purposes of this issue it would be okay to require reloading the kext to turn these diagnostics on/off (if that make the problem any easier).

cc: @jrbriggs

pmj:
> How would you recommend that we communicate to the kext that we'd like this behavior?

I'd probably do it either via a property on the PrjFS service object in the IOKit registry, or using a user client method in the logging user client. The latter is only really useful if it's the logging daemon that will enable or disable this tracing behaviour. The former would be convenient if we want a very small standalone CLI tool to toggle it.

pmj:
Just to keep everyone in the loop on this: I've now got a rough prototype of this that I've successfully used to trace what was causing the functional test failures in my previous draft of PR #1429. The filtering is still hardcoded, so I think I'm going to fix that aspect (otherwise the feature is of rather limited use) and then show you all what I've got in an RFC PR.

pmj:
I've thrown what I've got so far on https://github.com/pmj/VFSForGit/tree/mac-kext-event-tracing

Quick summary:
 * When enabled, events are logged to the KextLog mechanism. Currently vnode listener events only, FILEOP is still TODO.
 * New tool `prjfs-tracectl` which toggles this feature on or off with `--enable` or `--disable`. When enabling, other options control what is traced:
 * You can filter by path prefix using `--path-filter=/Users/test/devel/Adium4/` for example
 * you can also set a bitmask of VNODE actions using `--vnode-action-filter=<some-hex-value>` and only vnode events which bitwise AND with that mask to something nonzero will be logged
[...]

PR #1457, opened 2019-08-19
Mac ProjFS: In-kext diagnostic event tracing

pmj:
This is an early draft of my implementation of #1437 - event-level logging that can be turned on on demand.

The current approach is as follows:

 * When enabled, events are logged to the KextLog mechanism. Currently vnode listener events only, FILEOP is still TODO.
 * To avoid any runtime overhead when disabled, I've moved the code from `HandleVnodeOperation` into a templated function; the tracing calls *should* (I will verify this in due course) compile away to nothing when instantiated with the `NullTracer`. When tracing is enabled, the `KextLogTracer` instance of the template is used, in which the tracing calls actually do something and of course have a runtime overhead. Doing it this way avoids the need to have 2 versions of the kext or any such nonsense.
 * There is a new tool `prjfs-tracectl` which toggles this feature on or off with `--enable` or `--disable`. When enabling, other options control what is traced:
 * You must filter by path prefix using `--path-filter=/Users/test/devel/Adium4/` for example. If you want all files' events, set `/` as the filter. Note that globs are not supported, but `--path-filter=/Users/test/devel/Adium` (no trailing slash!) will match `/Users/test/devel/Adium1/`, `/Users/test/devel/Adium2/` and so on too.
 * You can also set a bitmask of VNODE actions using `--vnode-action-filter=<some-hex-value>` and only vnode events which bitwise AND with that mask to something nonzero will be logged.
 * `--vnode-message-events` turns on logging for any events that caused a message to be sent to the provider
 * `--vnode-events-denied turns on logging of anything which returned KAUTH_RESULT_DENY`
 * those 2 can be combined to log any event that either has a side effect via a message or caused an action to be denied.
 * Alternatively, `--vnode-all-events` will log any vnode event, even if it was a DEFER and didn't send a message

Log entries look something like this:
```
Event trace        11: Directory vnode '/Users/test/devel/Adium4/src/ASUnitTests' event by process 'bash' (PID = 616) action KAUTH_VNODE_READ_ATTRIBUTES KAUTH_VNODE_READ_SECURITY
Message to provider: 2 (RecursivelyEnumerateDirectory)
```
The above message is logged *before* the message is sent to the provider. Once it returns and the listener callback completes, you get:

```
Event trace        11: Directory vnode '/Users/test/devel/Adium4/src/ASUnitTests' event by process 'bash' (PID = 616) action KAUTH_VNODE_READ_ATTRIBUTES KAUTH_VNODE_READ_SECURITY
Message to provider: 2 (RecursivelyEnumerateDirectory) -> result: success
-> KAUTH_RESULT_DEFER
```

Note the index, so you can match them up and spot if the provider is taking a long time or not responding.

Things I would add if we like the broad approach:

 * Give the fileop handler the same treatment.
 * Try to add information for the *reason* why a vnode event was denied and the vnode or event metadata that led to the decision. Unfortunately the handler code is structured in a way that makes this difficult to do with zero runtime overhead when disabled (a lot of decisions are made in helper functions), so I'm probably going to have to compromise on level of detail or do some refactoring after all.
 * Require root privileges, or at minimum, membership of the "admin" group, for `prjfs-tracectl`. Only admins should be able to turn this on as it will log other users' file I/O metadata.
 * Add usage instructions in a `--help` mode for `prjfs-tracectl`

Note that the code still very much needs cleaning up (I need to move the tracers to a different file for example), so I'm definitely not interested in feedback on code style, organisation, etc. I **am** interested in feedback on:

 1. The UX of enabling/disabling the feature and the filtering options.
 2. The style of logging the events to the KextLog facility.
 3. Whether you like the general approach for inserting log statements using a template.

### DIFF
--- a/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
+++ b/ProjFS.Mac/PrjFSKext/KauthHandler.cpp
@@ -3,6 +3,7 @@
 #include <sys/proc.h>
 #include <kern/assert.h>
 #include <libkern/version.h>
+#include <libkern/libkern.h>
 #include <kern/thread.h>
 
 #include "KauthHandlerPrivate.hpp"
@@ -23,6 +24,7 @@
 #include "VnodeCache.hpp"
 #include "Memory.hpp"
 #include "ArrayUtilities.hpp"
+#include "KextLogEventTracer.hpp"
 
 #ifdef KEXT_UNIT_TESTING
 #include "KauthHandlerTestable.hpp"
@@ -45,6 +47,28 @@ struct PendingRenameOperation
 {
     vnode_t vnode;
     thread_t thread;
+};
+
+class NullTracer
+{
+    NullTracer() = delete;
+    NullTracer(const NullTracer&) = delete;
+    NullTracer& operator=(const NullTracer&) = delete;
+public:
+    explicit NullTracer(kauth_action_t vnodeAction, vnode_t vnode)
+    {}
+    explicit NullTracer(kauth_action_t fileopAction, uintptr_t arg0, uintptr_t arg1, uintptr_t arg2)
+    {}
+    void Flush()
+    {}
+    void Printf(const char* format, ...) __attribute__ ((format (printf, 2, 3)))
+    {}
+    void SendProviderMessage(MessageType providerMessage)
+    {}
+    void SendProviderMessageResult(bool success)
+    {}
+    void SetVnodeOpResult(int result)
+    {}
 };
 
 // Function prototypes
@@ -123,6 +147,16 @@ KEXT_STATIC bool InitPendingRenames();
 KEXT_STATIC void CleanupPendingRenames();
 KEXT_STATIC void ResizePendingRenames(uint32_t newMaxPendingRenames);
 
+template <typename EventTracer>
+int HandleVnodeOperationImpl(
+    kauth_cred_t    credential,
+    void*           idata,
+    kauth_action_t  action,
+    uintptr_t       arg0,
+    uintptr_t       arg1,
+    uintptr_t       arg2,
+    uintptr_t       arg3);
+
 // State
 static kauth_listener_t s_vnodeListener = nullptr;
 static kauth_listener_t s_fileopListener = nullptr;
@@ -161,8 +195,10 @@ kern_return_t KauthHandler_Init()
     {
         goto CleanupAndFail;
     }
+    
+    KextLogTracer::Initialize();
 
-    s_vnodeListener = kauth_listen_scope(KAUTH_SCOPE_VNODE, HandleVnodeOperation, nullptr);
+    s_vnodeListener = kauth_listen_scope(KAUTH_SCOPE_VNODE, HandleVnodeOperationImpl<NullTracer>, nullptr);
     if (nullptr == s_vnodeListener)
     {
         goto CleanupAndFail;
@@ -209,6 +245,8 @@ kern_return_t KauthHandler_Cleanup()
     ProviderMessaging_AbortAllOutstandingEvents();
     
     WaitForListenerCompletion();
+
+    KextLogTracer::Cleanup();
 
     CleanupPendingRenames();
     
@@ -408,8 +446,29 @@ KEXT_STATIC bool DeleteOpIsForRename(vnode_t vnode)
     return isRename;
 }
 
+template <typename EventTracer>
+    bool TraceAndTrySendRequestAndWaitForResponse(
+        EventTracer& eventTracer,
+        VirtualizationRootHandle root,
+        MessageType messageType,
+        const vnode_t vnode,
+        const FsidInode& vnodeFsidInode,
+        const char* vnodePath,
+        const char* fromPath,
+        int pid,
+        const char* procname,
+        int* kauthResult,
+        int* kauthError)
+{
+    eventTracer.SendProviderMessage(messageType);
+    bool messageSuccess = ProviderMessaging_TrySendRequestAndWaitForResponse(root, messageType, vnode, vnodeFsidInode, vnodePath, fromPath, pid, procname, kauthResult, kauthError);
+    eventTracer.SendProviderMessageResult(messageSuccess);
+    return messageSuccess;
+}
+
 // Private functions
-KEXT_STATIC int HandleVnodeOperation(
+template <typename EventTracer>
+int HandleVnodeOperationImpl(
     kauth_cred_t    credential,
     void*           idata,
     kauth_action_t  action,
@@ -432,6 +491,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
     UseMainForkIfNamedStream(currentVnode, putVnodeWhenDone);
 
+    EventTracer eventTracer(action, currentVnode);
+
     VirtualizationRootHandle root = RootHandle_None;
     uint32_t currentVnodeFileFlags;
     FsidInode vnodeFsidInode;
@@ -440,7 +501,7 @@ KEXT_STATIC int HandleVnodeOperation(
     bool isDeleteAction = false;
     bool isDirectory = false;
     bool isRename = false;
-
+    
     {
         PerfSample considerVnodeSample(&perfTracer, PrjFSPerfCounter_VnodeOp_BasicVnodeChecks);
         if (!VnodeIsEligibleForEventHandling(currentVnode))
@@ -509,7 +570,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
                 PerfSample recursivelyEnumerateSample(&perfTracer, PrjFSPerfCounter_VnodeOp_RecursivelyEnumerateDirectory);
         
-                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                if (!TraceAndTrySendRequestAndWaitForResponse(
+                        eventTracer,
                         root,
                         MessageType_KtoU_RecursivelyEnumerateDirectory,
                         currentVnode,
@@ -544,7 +606,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
                 PerfSample enumerateDirectorySample(&perfTracer, PrjFSPerfCounter_VnodeOp_EnumerateDirectory);
         
-                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                if (!TraceAndTrySendRequestAndWaitForResponse(
+                        eventTracer,
                         root,
                         MessageType_KtoU_EnumerateDirectory,
                         currentVnode,
@@ -628,7 +691,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
                 PerfSample enumerateDirectorySample(&perfTracer, PrjFSPerfCounter_VnodeOp_HydrateFile);
 
-                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                if (!TraceAndTrySendRequestAndWaitForResponse(
+                        eventTracer,
                         root,
                         MessageType_KtoU_HydrateFile,
                         currentVnode,
@@ -710,7 +774,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
                 PerfSample preConvertToFullSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreConvertToFull);
                 
-                if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+                if (!TraceAndTrySendRequestAndWaitForResponse(
+                        eventTracer,
                         root,
                         MessageType_KtoU_NotifyFilePreConvertToFull,
                         currentVnode,
@@ -751,7 +816,8 @@ KEXT_STATIC int HandleVnodeOperation(
         PerfSample preDeleteSample(&perfTracer, PrjFSPerfCounter_VnodeOp_PreDelete);
         
         // Predeletes must be sent after hydration since they may convert the file to full
-        if (!ProviderMessaging_TrySendRequestAndWaitForResponse(
+        if (!TraceAndTrySendRequestAndWaitForResponse(
+                eventTracer,
                 root,
                 isDirectory ?
                 MessageType_KtoU_NotifyDirectoryPreDelete :
@@ -771,6 +837,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
     
 CleanupAndReturn:
+    eventTracer.SetVnodeOpResult(kauthResult);
+
     if (putVnodeWhenDone)
     {
         vnode_put(currentVnode);
@@ -779,6 +847,20 @@ CleanupAndReturn:
     atomic_fetch_sub(&s_numActiveKauthEvents, 1);
     return kauthResult;
 }
+
+#ifdef KEXT_UNIT_TESTING
+KEXT_STATIC int HandleVnodeOperation(
+    kauth_cred_t    credential,
+    void*           idata,
+    kauth_action_t  action,
+    uintptr_t       arg0,
+    uintptr_t       arg1,
+    uintptr_t       arg2,
+    uintptr_t       arg3)
+{
+    return HandleVnodeOperationImpl<NullTracer>(credential, idata, action, arg0, arg1, arg2, arg3);
+}
+#endif
 
 // Note: a fileop listener MUST NOT return an error, or it will result in a kernel panic.
 // Fileop events are informational only.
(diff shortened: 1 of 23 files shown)


ProjFS.Mac/PrjFSKext/public/PrjFSCommon.h
```diff
@@ -45,6 +47,8 @@ enum PrjFSServiceUserClientType
 #define KEXT_STATIC_INLINE static inline
 #endif
 
+#ifdef __cplusplus
```
jeschu1: Curious why we added this

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+    void SetVnodeOpResult(int result)
+    {
+    }
+};
+
+static const char* MessageTypeString(MessageType messageType)
+{
+    switch(messageType)
+    {
+    case MessageType_KtoU_EnumerateDirectory:            return "EnumerateDirectory";
```
jeschu1: nit: identation

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+    NullTracer(const NullTracer&) = delete;
+    NullTracer& operator=(const NullTracer&) = delete;
+public:
+    explicit NullTracer(kauth_action_t vnodeAction, vnode_t vnode)
+    {}
+    explicit NullTracer(kauth_action_t fileopAction, uintptr_t arg0, uintptr_t arg1, uintptr_t arg2)
+    {}
+    
+    void Flush()
+    {
```
jeschu1: nit: {}

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+    case MessageType_KtoU_NotifyFileCreated:             return "NotifyFileCreated";
+    case MessageType_KtoU_NotifyFileRenamed:             return "NotifyFileRenamed";
+    case MessageType_KtoU_NotifyDirectoryRenamed:        return "NotifyDirectoryRenamed";
+    case MessageType_KtoU_NotifyFileHardLinkCreated:     return "NotifyFileHardLinkCreated";
+    case MessageType_KtoU_NotifyFilePreConvertToFull:    return "NotifyFilePreConvertToFull";
+    default:                                             return "Unknown";
+    };
+}
+
+class KextLogTracer
```
jeschu1: Would it be worth moving some of the logic to it's own class.  KextLogDiagnostics or similar?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+        }
+        else
+        {
+            return TraceBuffer { this->embeddedTraceBuffer, sizeof(this->embeddedTraceBuffer), this->traceBufferPosition };
+        }
+    }
+    
+    bool GrowBuffer(uint32_t minimumSize)
+    {
+        TraceBuffer currentBuffer = this->GetBuffer();
```
jeschu1: Do we have to worry about multiple threads here?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+        traceBufferPosition(0),
+        dynamicTraceBuffer(nullptr),
+        dynamicTraceBufferSize(0),
+        discarded(false),
+        willEmitTrace(false)
+    {
+        this->embeddedTraceBuffer[0] = '\0';
+        
+        kauth_action_t mask = atomic_load(&KextLogTracer::traceVnodeActionFilterMask);
+        if (0 == (mask & vnodeAction))
```
jeschu1: To be clear we drop actions that don't include the mask, but if a mask was never set we accept all?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+        }
+    }
+
+    void SetDeniedForCrawler()
+    {
+        if (!this->willEmitTrace && !this->discarded)
+        {
+            if (!atomic_load(&KextLogTracer::traceAllVnodeEvents) && atomic_load(&KextLogTracer::traceDeniedVnodeEvents))
+            {
+                // When tracing only denied events, throw out any denied crawlers as they tend to spam the trace
```
jeschu1: This seems ok, but is very good to document as you've done

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -772,6 +1172,8 @@ KEXT_STATIC int HandleVnodeOperation(
 
     
 CleanupAndReturn:
+    eventTracer.SetVnodeOpResult(kauthResult);
```
jeschu1: Would it be helpful to have the pathname here or is it easy enough to match up?

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+    
+    void SetDeniedForCrawler()
+    {}
+
+    void SetVnodeOpResult(int result)
+    {
+    }
+};
+
+static const char* MessageTypeString(MessageType messageType)
```
jeschu1: should this live in KextLog?

jeschu1:
@pmj I definitely like the idea of a separate program prjfs-tracel to turn the tracing on/off.

I think if we can move a lot of the new code to it's own class KextLogDiagnostics? that makes KauthHandler a lot easier to digest.

@wilbaker can you give a quick thumbs up/down on approach?  I started looking at cleanup and quickily noticed this is very much a WIP so don't bother with that :-).  @pmj is wisely getting an opinion on approach before getting too far.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -409,8 +777,29 @@ KEXT_STATIC bool DeleteOpIsForRename(vnode_t vnode)
...
+        const vnode_t vnode,
+        const FsidInode& vnodeFsidInode,
+        const char* vnodePath,
+        const char* fromPath,
+        int pid,
+        const char* procname,
+        int* kauthResult,
+        int* kauthError)
+{
+    eventTracer.SendProviderMessage(MessageType_KtoU_RecursivelyEnumerateDirectory);
```
jeschu1: Do you mean messageType above?

ProjFS.Mac/PrjFSKext/public/PrjFSCommon.h
pmj: C and Objective-C don't support namespaces, so if you `#include` this header from a .m or .c file, you'll get syntax errors on that part.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: What functionality would you split out? I guess the buffer could be encapsulated.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: No, each `KextLogTracer` instance has the lifetime of a vnode/fileop handler callback, and we stay on one thread for that.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: When tracing is first enabled in `PrjFSService::setProperties()`, the [mask defaults to `~0`](https://github.com/microsoft/VFSForGit/pull/1457/files/57466cbe5cb8001cfd1c9d27b701af40521429e2#diff-1beba556f03bf2aee251088144262bc0R150) so all actions allowed. Only if the user process sets it to something else do we deviate from that default.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yeah; we could add an extra option so that those denied events are also logged, but I suspect it's probably very rare that you need this.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: The buffer in the `eventTracer` will already contain the path at this point, so it'll be printed together with the return value.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes, or perhaps `Message_Kernel.cpp`. Plenty of code still needs rearranging in this PR; I didn't want to spend any time on aesthetics until I knew I was building the right thing.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
pmj: Yes of course, thanks for catching that!

wilbaker:
@pmj I just gave this a try and I can see prjfs-tracectl log the settings when I enable it:

image

But after that point I don't see anything logged when I have permission denied:

```
~/GVFSTest/src>git config --unset core.hookspath
error: could not lock config file .git/config: Permission denied
~/GVFSTest/src>cat License.md 
cat: License.md: Permission denied
~/GVFSTest/src>cat License.md 
cat: License.md: Permission denied
```

I am still on High Sierra, do you think that could be related?

pmj:
> @pmj I just gave this a try and I can see prjfs-tracectl log the settings when I enable it:
> 
> image

Where are you looking for the logs? This uses the existing KextLog system, so by default you will need to have `KextLogDaemon` running, and you need to be watching for the correct subsystem in Console.app or on the `log` CLI. As I don't see the output from [`KauthHandler_EnableTraceListeners`](https://github.com/microsoft/VFSForGit/pull/1457/files#diff-ab3be22ed1fbe1779471dbe568982eb0R2080) in your screenshot my guess is you're not filtering for the KextLog subsystem (`org.vfsforgit.prjfs.PrjFSKextLogDaemon`) but for something else.

> I am still on High Sierra, do you think that could be related?

I don't think the macOS version should make any difference here.

wilbaker:
> This uses the existing KextLog system, so by default you will need to have KextLogDaemon running, and you need to be watching for the correct subsystem in Console.app or on the log CLI.

Ahh that explains it I did not have a dev version of the logging daemon running.  Thanks!

wilbaker:
>@wilbaker can you give a quick thumbs up/down on approach?  I started looking at cleanup and quickily noticed this is very much a WIP so don't bother with that :-).  @pmj is wisely getting an opinion on approach before getting too far.
>
>The basic idea is @pmj adds an eventTracer that log when
>-  we deny due to a crawler
>- type before we sent message
>- result after we send message
>- outcome of the request (kauthResult)

Apologies for not getting to this sooner.  At a high level that approach sounds good to me, I've just started looking through the changes and will try out the feature first thing next week and reply to this PR with feedback.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -47,6 +48,359 @@ struct PendingRenameOperation
...
+        {
+            int pathLen = sizeof(pathBuffer);
+            errno_t error = vn_getpath(vnode, pathBuffer, &pathLen);
+            if (error == 0)
+            {
+                RWLock_AcquireShared(KextLogTracer::traceFilterLock);
+                {
+                    if (KextLogTracer::pathPrefixFilter != nullptr)
+                    {
+                        shouldTrace = strprefix(pathBuffer, KextLogTracer::pathPrefixFilter);
```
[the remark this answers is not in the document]
pmj: Given how high-frequency the check is, I think sticking with case sensitive is fine, as HFS & APFS are case-preserving.

ProjFS.Mac/prjfs-tracectl/prjfs-tracectl.m
```diff
@@ -0,0 +1,154 @@
...
+    io_service_t prjfsService = IOServiceGetMatchingService(kIOMasterPortDefault, matchDict); // matchDict consumed
+    
+    if (prjfsService == IO_OBJECT_NULL)
+    {
+        CFRelease(traceSettings);
+        fprintf(stderr, "PrjFS Service object not found.\n");
+        return 1;
+    }
+    
+    IORegistryEntrySetCFProperty(prjfsService, CFSTR(PrjFSEventTracingKey), traceSettings);
```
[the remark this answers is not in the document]
pmj: Sounds sensible! Detecting process death will require switching to a user client based approach, so there's a persistent connection between the kext and the trace process which gets severed when the process dies. I'll try to get that done in the next few days.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -409,8 +777,29 @@ KEXT_STATIC bool DeleteOpIsForRename(vnode_t vnode)
...
+        const char* vnodePath,
+        const char* fromPath,
+        int pid,
+        const char* procname,
+        int* kauthResult,
+        int* kauthError)
+{
+    eventTracer.SendProviderMessage(MessageType_KtoU_RecursivelyEnumerateDirectory);
+    bool messageSuccess = ProviderMessaging_TrySendRequestAndWaitForResponse(root, messageType, vnode, vnodeFsidInode, vnodePath, fromPath, pid, procname, kauthResult, kauthError);
+    eventTracer.SendProviderMessageResult(messageSuccess);
```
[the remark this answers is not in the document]
pmj: It's certainly an option, though I'm not sure it's going to be a clear win in the short term; none of the problems I've had to diagnose thus far have involved errors reported to the kext by the provider.

ProjFS.Mac/PrjFSKext/KextLog.hpp
```diff
@@ -107,47 +107,55 @@ template <typename... args>
...
+    ((action) & KAUTH_VNODE_EXECUTE)              ? itemPrefix "KAUTH_VNODE_EXECUTE" : "", \
+    ((action) & KAUTH_VNODE_DELETE)               ? itemPrefix "KAUTH_VNODE_DELETE" : "", \
+    ((action) & KAUTH_VNODE_APPEND_DATA)          ? itemPrefix "KAUTH_VNODE_APPEND_DATA" : "", \
+    ((action) & KAUTH_VNODE_READ_ATTRIBUTES)      ? itemPrefix "KAUTH_VNODE_READ_ATTRIBUTES" : "", \
+    ((action) & KAUTH_VNODE_WRITE_ATTRIBUTES)     ? itemPrefix "KAUTH_VNODE_WRITE_ATTRIBUTES" : "", \
+    ((action) & KAUTH_VNODE_READ_EXTATTRIBUTES)   ? itemPrefix "KAUTH_VNODE_READ_EXTATTRIBUTES" : "", \
+    ((action) & KAUTH_VNODE_WRITE_EXTATTRIBUTES)  ? itemPrefix "KAUTH_VNODE_WRITE_EXTATTRIBUTES" : "", \
+    ((action) & KAUTH_VNODE_READ_SECURITY)        ? itemPrefix "KAUTH_VNODE_READ_SECURITY" : "", \
+    ((action) & KAUTH_VNODE_WRITE_SECURITY)       ? itemPrefix "KAUTH_VNODE_WRITE_SECURITY" : "", \
+    ((action) & KAUTH_VNODE_TAKE_OWNERSHIP)       ? itemPrefix "KAUTH_VNODE_TAKE_OWNERSHIP" : ""
```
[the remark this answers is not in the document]
pmj: Yep, good idea.

ProjFS.Mac/PrjFSKext/KauthHandler.cpp
```diff
@@ -470,6 +861,10 @@ KEXT_STATIC int HandleVnodeOperation(
             &kauthResult,
             kauthError))
     {
+        if (kauthResult == KAUTH_RESULT_DENY && *kauthError != EBADF)
+        {
+            eventTracer.SetDeniedForCrawler();
```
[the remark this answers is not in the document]
pmj: Yes, tracing the reason for decisions is definitely something I'd like to do. As you've noticed, the way the code is currently structured didn't allow that without changes. I guess the suggested change makes this cluster of functions **yet** more complicated (see #900) which is something I really wanted to avoid.

I'm going to take another look at this and see if I can find a slightly cleaner approach.

pmj:
A few things that came up while using this for some debugging last night:

 [ ] It would be useful to filter crawler events even in `--vnode-all-events` mode. Perhaps we can have a `--vnode-all-events-even-crawlers` for those cases where we do care about them.
 [ ] I noticed there was no 2-part tracing for events causing messages in `--vnode-all-events`.
 [ ] Some trace indices were wrong (uninitialised?) in `--vnode-all-events` mode.

ProjFS.Mac/prjfs-tracectl/prjfs-tracectl.m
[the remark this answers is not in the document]
pmj: While implementing this it occurred to me that if we're going to go to the effort of an extra user client type, we could push the trace events directly to that client rather than going via the KextLog daemon. This has a few advantages:

 * The trace logs are *right there* on stdout from the `prjfs-tracectl` command. No need to hunt around in Console or use `log stream`, and no issues with which type of build is running.
 * We can make the trace buffer a *lot* larger than the KextLog buffer, so we're less likely to drop events. We can even make the buffer size configurable.

I'd probably use the same message framing protocol as with KextLog as that means we can share the encoding/decoding code, which could be extra handy when we add support for logging stack traces in KextLog.

What do you think of this approach @jeschu1? It's a little bit of extra work but nothing massive.

pmj:
I've pushed a series of significant updates to this feature:

 * The code is organised a bit better (the `KextLogTracer` code lives in a separate set of files, new helper functions are not just randomly lumped into `KauthHandler.cpp`)
 * `prjfs-tracectl` now uses a custom `IOUserClient` class in the kext to control tracing behaviour. Tracing is disabled automatically when the client disconnects, so running with `--disable` isn't really necessary. (I'll get rid of the flag completely in the next iteration.) This means you need to leave `prjfs-tracectl` running for the duration of the trace.
 * `prjfs-tracectl` now accepts a new flag `--vnode-crawler-events`. Without this, *all* crawler events are removed from the trace.
 * Various bug fixes.
 * `prjfs-tracectl` must now be run as root, as tracing will leak other users' data (mainly potentially-private file names) to the log.

pmj:
I've just tidied the commit series up somewhat, but beyond that I haven't really changed anything since my last comment. I have however been doing a bunch of manual testing with the feature, and while there's still plenty of room for improvement in terms of features, what's there works well, and I think the diff is already large enough for one PR, so I'd prefer to get it merged and then work on enhancements.

The functional test failure seems to be an issue with the machine it got assigned, I've kicked off a new build at https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=10447778&view=results which I'm hoping will succeed. I've also started a local functional test run over night just in case there's a real hang problem, in which case I can diagnose that tomorrow.

At any rate, please re-review and let me know what you think!

pmj:
This is now working reliably again, and as far as I'm concerned, ready to merge. When you have a chance, @wilbaker please review! Thanks.

wilbaker:

PR #1485, opened 2019-08-23
Mark panic strings as 'public' so they appear in Console

### LINKED ISSUE #1475: Mac: Panic related log messages in PrjFSKextLogDaemon should use {public} when formatting strings
The new log messages that were added in #1445 are not logging any secret/private information, and so `%{public}s` should be used instead of `%s` in formatting strings passed to `os_log`.  This will prevent the logs from being redacted.

jeschu1:
resolves #1475

### DIFF
--- a/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
+++ b/ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
@@ -199,7 +199,7 @@ static void LogPanics()
                if (panicLogTime == 0)
                {
                    // Record error if we can't determine timestamp
-                   os_log(s_daemonLogger, "Unable to run stat on panic at %s errno=%d strerror=%s\n", fullFileName.c_str(), errno, strerror(errno));
+                   os_log(s_daemonLogger, "Unable to run stat on panic at %{public}s errno=%d strerror=%{public}s\n", fullFileName.c_str(), errno, strerror(errno));
                    continue;
                }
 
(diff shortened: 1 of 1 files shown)


ProjFS.Mac/PrjFSKextLogDaemon/PrjFSKextLogDaemon.cpp
```diff
@@ -199,7 +199,7 @@ static void LogPanics()
                if (panicLogTime == 0)
                {
                    // Record error if we can't determine timestamp
-                   os_log(s_daemonLogger, "Unable to run stat on panic at %s errno=%d strerror=%s\n", fullFileName.c_str(), errno, strerror(errno));
```
pmj: Is the `strerror` similarly affected, do we need to mark that as public too? If that's not necessary, any idea why not?

wilbaker: I would guess that is affected.  In my experience every %s will be private by default.

pmj: This is what I thought too but there appears to be some kind of heuristic at play which doesn't redact everything. Presumably they run some regexes on the string to guess whether it might be a path, etc.

jeschu1: The error messages are not marked as private.  It appears there is a heuristic.  

[This article](https://www.lordcodes.com/posts/clear-and-searchable-logging-in-swift-with-oslog) suggests dynamic string content will be marked as private but that's not what I'm seeing here. 

I'll update the others to be %s{public} as well and verify the strings continue to come thru.  Better safe than sorry.

pmj: Actually, I wonder if it's the fact that the `strerror()` strings are most likely hardcoded and reside in the constant section of a binary, and the system is detecting this from the address and thus marking it as public automatically. Regardless, I'd explicitly mark them as public in case the heuristic changes in future.

Issue #1327, opened 2019-07-12
Mac: Make Functional Tests work on Catalina


jeschu1:

jeschu1:
With this change, https://github.com/microsoft/VFSForGit/pull/1337/files I can get all the tests to pass on Catalina Machine (update 4 19A512f)

wilbaker:
@jeschu1 @jrbriggs I removed the pri1 label from this issue

jrbriggs:
Should we just close it? @jeschu1 's comment above sounds like, as phrased, this issue is resolved.

wilbaker:
@jrbriggs I think we can either close this now or leave it open until Catalina ships and we confirm the functional tests pass there.  I don't have a strong feeling one way or the other as long as we continue to run the tests as new betas are released 😃

PR #1486, opened 2019-08-23
[WIP] Clean up Register/Unregister for OfflineIO

### LINKED ISSUE #1474: VFS4G verbs should fail if KernelDriver.RegisterForOfflineIO() fails
Currently VFS4G is not checking the return value of `RegisterForOfflineIO()`.  If `RegisterForOfflineIO()` fails the verb should not be allowed to continue as its attempts to perform offline IO will likely fail.

Steps to reproduce `RegisterForOfflineIO` failing:

1. Install an official VFS4G on Mac
2. Build a debug build of VFS4G locally
3. Attempt to clone a repo

*Output*

```
~/Repos/VFSForGit/src>./Scripts/Mac/GVFS_Clone.sh https://gvfs.visualstudio.com/ci/_git/ForTests
PrjFS kernel service interface version mismatch. Kernel: 0.6.19224.2, this library expects: 0.2.173.2
```

jeschu1:
- Add a call from MountSetsCoreHooksPath
- Ensure we check Register is successful when starting up
- Have verbs explicitly unregister

resolves #1474, #1327


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -83,6 +85,8 @@ public void MountSetsCoreHooksPath()
                 this.Enlistment.RepoRoot, "config core.hookspath")
                 .Trim('\n')
                 .ShouldEqual(expectedHooksPath);
+
+            GVFSHelpers.HandleUnregisterForOfflineIO();
```
wilbaker: Should `GVFSHelpers.HandleUnregisterForOfflineIO()` be put in the `finally` of a `try/finally` so that if there is an exception thrown in the test code we still call `HandleUnregisterForOfflineIO`?

GVFS/GVFS.FunctionalTests/Tools/GVFSHelpers.cs
```diff
@@ -179,6 +183,24 @@ public static string GetInternalParameter(string maintenanceJob = "null", string
...
+        public static void HandleRegisterForOfflineIO()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                uint result = RegisterForOfflineIO();
+                result.ShouldEqual<uint>(PrjFSResultSuccess, $"{nameof(HandleRegisterForOfflineIO)} failed (result = {result})");
+            }
+        }
+
+        public static void HandleUnregisterForOfflineIO()
```
wilbaker: Naming suggestion for `HandleRegisterForOfflineIO` and `HandleUnregisterForOfflineIO`.  

The name of these made me think they might be callback handlers.  I think you could leave them simply as `RegisterForOfflineIO` and `UnregisterForOfflineIO` and on non-OSX platforms the methods no-op.

GVFS/GVFS/Program.cs
```diff
@@ -13,7 +13,11 @@ public class Program
         public static void Main(string[] args)
         {
             GVFSPlatformLoader.Initialize();
-            GVFSPlatform.Instance.KernelDriver.RegisterForOfflineIO();
+            if (!GVFSPlatform.Instance.KernelDriver.RegisterForOfflineIO())
+            {
+                Console.WriteLine("Unable to register for OfflineIO");
```
wilbaker: What would you think of us providing a bit more detail here?  

Maybe something like:

> "Unable to register with the kernel for offline I/O.  Ensure that VFS for Git installed successfully and try again"

GVFS/GVFS/Program.cs
```diff
@@ -94,14 +98,24 @@ public static void Main(string[] args)
                             }
 
                             verb.Execute();
-                            Environment.Exit((int)ReturnCode.Success);
+                           Exit((int)ReturnCode.Success);
```
wilbaker: Nit: `verb` and `Exit` are not lining up vertically, could you fix the spacing of the code in this block?

PR #1498, opened 2019-08-29
Don't allow Executables to be Placeholders

### LINKED ISSUE #1495: Python Executables fail on first call
Repo Steps
1. Clone a repository with a Python Executable
2. Attempt to run (Without an explicit hydration first)

You will get an error about 'import' statements.  Second run is successful.

jeschu1:
resolves #1495

@pmj please have an early look here
If you want to advise any speedups in comments i'll apply them right away

I want to do one final look at the events we get for the executables to validate if any alternatives are possible.

Note PR #1497 shows the failure and the changes to test here prove they work

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
@@ -6,6 +6,7 @@ using GVFS.Tests.Should;
 using NUnit.Framework;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Threading;
@@ -917,6 +918,30 @@ namespace GVFS.FunctionalTests.Tests.LongRunningEnlistment
 
             FileSystemRunner.DefaultRunner.DeleteDirectory(parentDirectoryPath);
         }
+
+        [TestCase]
+        [Category(Categories.MacOnly)]
+        public void RunPythonExecutable()
+        {
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout FunctionalTests/PythonExecutable");
+
+            // Found an issue on Mac where running a python executable that is a placeholder, fails
+            // The fix was to always hydrate executables (no placeholders for this mode)
+            // To repro this issue in the C# framework the python executable must be run via a wrapper
+            string pythonDirectory = Path.Combine(this.Enlistment.RepoRoot, "Test_Executable");
+            string pythonExecutable = Path.Combine(pythonDirectory, "python_wrapper.sh");
+
+            ProcessStartInfo startInfo = new ProcessStartInfo(pythonExecutable);
+            startInfo.RedirectStandardOutput = true;
+            startInfo.RedirectStandardError = true;
+            startInfo.WorkingDirectory = pythonDirectory;
+
+            ProcessResult result = ProcessHelper.Run(startInfo);
+            result.ExitCode.ShouldEqual(0);
+            result.Output.ShouldContain("3.14");
+
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout " + this.Enlistment.Commitish);
+       }
 
         private class FileRunnersAndFolders
         {
--- a/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
@@ -2,6 +2,8 @@
 #include <cassert>
 #include <stddef.h>
 #include <sys/ioctl.h>
+#include <sys/mount.h>
+#include <sys/param.h>
 #include <sys/stat.h>
 #include <sys/sys_domain.h>
 #include <sys/xattr.h>
@@ -87,6 +89,8 @@ struct MutexAndUseCount
 
 typedef map<FsidInode, MutexAndUseCount, FsidInodeCompare> FileMutexMap;
 
+static fsid_t s_virtualizationRoot_fsid;
+
 // Function prototypes
 static bool SetBitInFileFlags(const char* fullPath, uint32_t bit, bool value);
 static bool IsBitSetInFileFlags(const char* fullPath, uint32_t bit);
@@ -112,6 +116,7 @@ static void HandleKernelRequest(void* messageMemory, uint32_t messageSize);
 static PrjFS_Result HandleEnumerateDirectoryRequest(const MessageHeader* request, const char* absolutePath, const char* relativePath);
 static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHeader* request, const char* absolutePath, const char* relativePath);
 static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const char* absolutePath, const char* relativePath);
+static PrjFS_Result HydrateFile(const char* absolutePath, const char* relativePath, FsidInode fsidInode, pid_t pid, const char* procname);
 static PrjFS_Result HandleNewFileInRootNotification(
     const MessageHeader* request,
     const char* relativePath,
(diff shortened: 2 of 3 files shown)


ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -87,6 +89,8 @@ struct MutexAndUseCount
 
 typedef map<FsidInode, MutexAndUseCount, FsidInodeCompare> FileMutexMap;
 
+static fsid_t virtualizationRoot_fsid;
```
pmj: I believe our coding guidelines require this to start with `s_` and be camelCase past that.

GVFS/GVFS.FunctionalTests/Settings.cs
```diff
@@ -35,7 +35,7 @@ public static void Initialize()
                 CurrentDirectory = Path.GetFullPath(Path.GetDirectoryName(Environment.GetCommandLineArgs()[0]));
 
                 RepoToClone = @"https://gvfs.visualstudio.com/ci/_git/ForTests";
-                Commitish = @"FunctionalTests/20180214";
+                Commitish = @"FunctionalTests/20190827";
```
pmj: For the benefit of the uninitiated (me), what does this do?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
```diff
@@ -918,6 +919,23 @@ public void CreateDirectoryInheritsParentDirectoryAttributes(string parentFolder
             FileSystemRunner.DefaultRunner.DeleteDirectory(parentDirectoryPath);
         }
 
+        [TestCase]
+        [Category(Categories.MacOnly)]
+        public void RunPythonExecutable()
+        {
+            string pythonDirectory = Path.Combine(this.Enlistment.RepoRoot, "Test_Executable");
+            string pythonExecutable = Path.Combine(pythonDirectory, "python_wrapper.sh");
```
pmj: Is there a reason the python script launch is wrapped in a shell script? (I’m just curious)

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -460,6 +473,28 @@ PrjFS_Result PrjFS_WritePlaceholderFile(
...
+           LogWarning("PrjFS_WritePlaceholderFile: lstat failed on %s errno=%d, strerror=%s", fullPath, errno, strerror(errno));
+           result = PrjFS_Result_EIOError;
+           goto CleanupAndFail;
+        }
+
+        FsidInode fsidInode;
+        fsidInode.inode = fileAttributes.st_ino;
+        fsidInode.fsid = virtualizationRoot_fsid;
+
+        if (HydrateFile(fullPath, relativePath, fsidInode, 1, "placeholder") != PrjFS_Result_Success)
```
pmj: Do we want to pass through the specific error here or explicitly EIOError as you’ve done?

GVFS/GVFS.FunctionalTests/Settings.cs
jeschu1: Great question ;-), I had to 2 new files to the 20190827 branch (wrapper.sh / test.py) to perform the test.  Generally if we add new files we create a new branch.

@kewillford to please validate this change.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
jeschu1: Exellent question.  While calling test.py on the command line repros the issue, calling it from C# does not.  C# does something to hydrate it before execution.  If I had to guess it's reading attributes.  By calling via a wrapper I can repro the issue.  

Side note, the wrapped added an extra level of fun to reproing in the unit test.

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: Under the covers it looks like it is all EIOError anyways.  

The EIOError is important because it signals to the provider that the affected file should be put in modified paths.  This means it will show up in 'git status' (which we want since it's likely different).

So in summary yes I think we want to explicity set this as PrjFS_Result_EIOError

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
```diff
@@ -918,6 +919,27 @@ public void CreateDirectoryInheritsParentDirectoryAttributes(string parentFolder
             FileSystemRunner.DefaultRunner.DeleteDirectory(parentDirectoryPath);
         }
 
+        [TestCase]
+        [Category(Categories.MacOnly)]
+        public void RunPythonExecutable()
+        {
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout FunctionalTests/20190827");
```
kewillford: I would rename this branch to something like `FunctionalTests/PythonExecutable` just because we have normally used the timestamp branch for the main test branch.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
```diff
@@ -918,6 +919,27 @@ public void CreateDirectoryInheritsParentDirectoryAttributes(string parentFolder
...
+        {
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout FunctionalTests/PythonExecutable");
+
+            string pythonDirectory = Path.Combine(this.Enlistment.RepoRoot, "Test_Executable");
+            string pythonExecutable = Path.Combine(pythonDirectory, "python_wrapper.sh");
+
+            ProcessStartInfo startInfo = new ProcessStartInfo(pythonExecutable);
+            startInfo.RedirectStandardOutput = true;
+            startInfo.RedirectStandardError = true;
+            startInfo.WorkingDirectory = pythonDirectory;
```
kewillford: Do we need to set any of the other properties like `CreateNoWindow` or `UseShellExecute`?  Not sure those make any difference in this test.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
pmj: OK, that's a good reason! Perhaps we'd like to document that in a comment in the test?

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
pmj: OK, sounds good!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
```diff
@@ -1016,18 +1051,8 @@ static PrjFS_Result HandleRecursivelyEnumerateDirectoryRequest(const MessageHead
     return result;
 }
 
-static PrjFS_Result HandleHydrateFileRequest(const MessageHeader* request, const char* absolutePath, const char* relativePath)
+static PrjFS_Result HydrateFile(const char* absolutePath, const char* relativePath, FsidInode fsidNode, pid_t pid, const char* procname)
```
pmj: Nit-picking, but I just noticed the naming on the parameter here - is the "node" in `fsidNode` a typo - shouldn't that be `fsidInode`? (Or `fsidAndInode` if you want maximum accuracy) The file ID in this struct is typically referred to as an ["inode"](https://en.wikipedia.org/wiki/Inode) which is what we've called it in the rest of the kext. HFS+ actually calls it the ["CNID"](https://en.wikipedia.org/wiki/Hierarchical_File_System#Design) internally but I don't think they kept that for APFS and it also isn't called that in any of the POSIX layer APIs of macOS. [The `fsgetpath` man page](https://www.unix.com/man-page/mojave/2/fsgetpath/) also calls it an "inode".

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
jeschu1: @pmj excellent point I'll add that documentation!

ProjFS.Mac/PrjFSLib/PrjFSLib.cpp
jeschu1: Nice catch, good variable names are important i'll update it!

PR #1499, opened 2019-08-30
Mac ProjFS: Don't perform user/kernel version validation for offline I/O

pmj:
Offline I/O registration uses a trivial user/kernel API which is not sensitive to version differences, and there have been instances where gvfs upgrade has failed and left VFS4G in an unusable state because version validation failed mid-upgrade.

This change disables validation for offline I/O registration only.

### DIFF
--- a/ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
+++ b/ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
@@ -57,7 +57,7 @@ CleanupAndFail:
     return false;
 }
 
-io_connect_t PrjFSService_ConnectToDriver(enum PrjFSServiceUserClientType clientType)
+io_connect_t PrjFSService_ConnectToDriver(enum PrjFSServiceUserClientType clientType, bool performVersionValidation)
 {
     CFDictionaryRef matchDict = IOServiceMatching(PrjFSServiceClass);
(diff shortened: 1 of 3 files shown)


ProjFS.Mac/PrjFSLib/PrjFSUser.cpp
```diff
@@ -70,7 +70,7 @@ io_connect_t PrjFSService_ConnectToDriver(enum PrjFSServiceUserClientType client
         return IO_OBJECT_NULL;
     }
     
-    if (!PrjFSService_ValidateVersion(prjfsService))
+    if (performVersionValidation && !PrjFSService_ValidateVersion(prjfsService))
```
jeschu1: @pmj thinking about this more it looks like the "ConnectToDriver" function is also used when we start the virtualization instance.  Will we fail to mount if there is a version mismatch?

pmj: Yes, if there's a version mismatch, 

    s_kernelServiceConnection = PrjFSService_ConnectToDriver(UserClientType_Provider);

will fail, as `performVersionValidation` defaults to `true` in this change. So no change of behaviour there.

jeschu1: Cool, I am still happy to take this change.  It's minimal and the right place to fail would be on setting the virtualization up.

PR #1500, opened 2019-08-30
Windows native toasts

alameenshah:
#### Supported notifications
| Notification | Screenshot | Action |
| --- | --- |---|
| Autmount start |Automount_Start|This notification is not actionable |
| Mount failure during auto-mount |Automount_Failure|Automount_Failure_Response|
| Upgrade available | Upgrade_Available| Upgrade_Available_Response|

#### Changes in this PR
- `VFSForGit` installer creates a `Start Menu` shortcut to `GVFS.Service.UI.exe` after successful install. (This shortcut is a requirement for enabling native toasters.)
- `VFSForGit` installer launches `GVFS.Service.UI.exe` in the currently logged-in user context after install
- `GVFS.Service` will launch `GVFS.Service.UI` when a new user logs in.
- `GVFS.Service.UI.exe` will always be running.
- `GVFS.Service.UI.exe` listens for supported notifications. Notifications that cannot be identified are ignored.
- Auto-mount and Upgrade available toasters are actionable. In response these actionable toasters, `GVFS.Service.UI` runs `gvfs mount` and `gvfs upgrade --confirm` (elevated) verbs in a new Command Prompt window.
- `GVFS.Service` waits for a minute before doing upgrade checks, just so that `VFSForGit` installer gets the time to launch `GVFS.Service.UI` that is needed to display the notifications.

#### TODO
- [x] Testing
- [x] Integrate with `GVFS.Service`
- [x] Unit testing
- [x] Update diagnose verb to cat `GVFS.Service.UI` logs

### DIFF
--- a/GVFS/GVFS.Installer.Windows/Setup.iss
+++ b/GVFS/GVFS.Installer.Windows/Setup.iss
@@ -27,6 +27,7 @@
 #define EnvironmentKey "SYSTEM\CurrentControlSet\Control\Session Manager\Environment"
 #define FileSystemKey "SYSTEM\CurrentControlSet\Control\FileSystem"
 #define GvFltAutologgerKey "SYSTEM\CurrentControlSet\Control\WMI\Autologger\Microsoft-Windows-Git-Filter-Log"
+#define ServiceUIName "VFS For Git"
 
 [Setup]
 AppId={{489CA581-F131-4C28-BE04-4FB178933E6D}
@@ -130,10 +131,11 @@ DestDir: "{app}"; Flags: ignoreversion; Source:"{#GVFSDir}\GVFS.Platform.Windows
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#GVFSDir}\GVFS.pdb"
 
 ; GVFS.Service.UI Files
-DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceUIDir}\GVFS.Service.UI.exe" 
+DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceUIDir}\GVFS.Service.UI.exe"
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceUIDir}\GVFS.Service.UI.exe.config" 
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceUIDir}\GVFS.Service.UI.pdb"
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceUIDir}\GitVirtualFileSystem.ico"
+DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceUIDir}\System.Runtime.dll"
 
 ; GVFS Files
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#GVFSDir}\CommandLine.dll"
@@ -176,6 +178,9 @@ DestDir: "{app}"; Flags: ignoreversion; Source:"{#GVFSDir}\System.IO.Compression
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.pdb"
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe.config"
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe"; AfterInstall: InstallGVFSService
+
+[Icons]
+Name: "{commonstartmenu}\{#ServiceUIName}"; Filename: "{app}\GVFS.Service.UI.exe"; AppUserModelID: "GVFS"
 
 [UninstallDelete]
 ; Deletes the entire installation directory, including files and subdirectories
@@ -350,6 +355,34 @@ begin
   if InstallSuccessful = False then
     begin
       RaiseException('Fatal: An error occured while installing GVFS.Service.');
+    end;
+end;
+
+procedure StartGVFSServiceUI();
+var
+  ResultCode: integer;
+begin
+  if ExecAsOriginalUser(ExpandConstant('{app}\GVFS.Service.UI.exe'), '', '', SW_HIDE, ewNoWait, ResultCode) then
+    begin
+      Log('StartGVFSServiceUI: Successfully launched GVFS.Service.UI');
+    end
+  else
+    begin
+      Log('StartGVFSServiceUI: Failed to launch GVFS.Service.UI');
+    end;
+end;
+
+procedure StopGVFSServiceUI();
+var
+  ResultCode: integer;
+begin
+  if Exec('powershell.exe', '-NoProfile "Stop-Process -Name GVFS.Service.UI"', '', SW_HIDE, ewNoWait, ResultCode) then
+    begin
+      Log('StopGVFSServiceUI: Successfully stopped GVFS.Service.UI');
+    end
+  else
+    begin
+      RaiseException('Fatal: Could not stop process: GVFS.Service.UI');
     end;
 end;
 
@@ -730,6 +763,7 @@ begin
       begin
         if ExpandConstant('{param:REMOUNTREPOS|true}') = 'true' then
           begin
+            StartGVFSServiceUI();
             MountRepos();
           end
       end;
@@ -746,6 +780,7 @@ begin
   case CurStep of
     usUninstall:
       begin
+        StopGVFSServiceUI();
         UninstallService('GVFS.Service', False);
         RemovePath(ExpandConstant('{app}'));
       end;
@@ -769,6 +804,7 @@ begin
       Abort();
     end;
   StopService('GVFS.Service');
+  StopGVFSServiceUI();
   UninstallGvFlt();
   UninstallProjFSIfNecessary();
 end;
--- a/GVFS/GVFS.Service.UI/GVFS.Service.UI.csproj
+++ b/GVFS/GVFS.Service.UI/GVFS.Service.UI.csproj
@@ -55,8 +55,12 @@
     <Compile Include="Data\BindingData.cs" />
     <Compile Include="Data\BindingItem.cs" />
     <Compile Include="Data\VisualData.cs" />
+    <Compile Include="GVFSServiceUI.cs" />
+    <Compile Include="GVFSToastRequestHandler.cs" />
+    <Compile Include="IToastNotifier.cs" />
     <Compile Include="Program.cs" />
     <Compile Include="Data\ToastData.cs" />
+    <Compile Include="WinToastNotifier.cs" />
     <Compile Include="XmlList.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
+++ b/GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
@@ -1,0 +1,190 @@
+﻿using GVFS.Common.NamedPipes;
+using GVFS.Common.Tracing;
+using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace GVFS.Service.UI
+{
+    public class GVFSToastRequestHandler
+    {
+        private const string VFSForGitAutomountStartTitle= "VFS For Git Automount";
+        private const string VFSForGitAutomountStartMessageFormat = "Attempting to mount {0} VFS For Git repos(s)";
+
+        private const string VFSForGitAutomountSuccessTitle = "VFS For Git Automount";
+        private const string VFSForGitAutomountSuccessMessageFormat = "The following VFS For Git repo is now mounted: {0}{1}";
+
+        private const string VFSForGitAutomountErrorTitle = "VFS For Git Automount";
+        private const string VFSForGitAutomountErrorMessageFormat = "The following VFS For Git repo failed to mount: {0}{1}";
+        private const string VFSForGitAutomountButtonTitle = "Retry";
+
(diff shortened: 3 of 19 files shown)


GVFS/GVFS.Installer.Windows/Setup.iss
```diff
@@ -27,6 +27,7 @@
 #define EnvironmentKey "SYSTEM\CurrentControlSet\Control\Session Manager\Environment"
 #define FileSystemKey "SYSTEM\CurrentControlSet\Control\FileSystem"
 #define GvFltAutologgerKey "SYSTEM\CurrentControlSet\Control\WMI\Autologger\Microsoft-Windows-Git-Filter-Log"
+#define ServiceUIName "GVFS.Service.UI"
```
mjcheetham: Do we want to give the UI a nicer name (assuming the notification sender is based on the shortcut name), like "VFS for Git"?

image


```suggestion
#define ServiceUIName "VFS for Git"
```

GVFS/GVFS.Service.UI/GVFS.Service.UI.csproj
```diff
@@ -86,5 +91,6 @@
     <Analyzer Include="..\..\..\packages\StyleCop.Analyzers.1.0.2\analyzers\dotnet\cs\StyleCop.Analyzers.CodeFixes.dll" />
     <Analyzer Include="..\..\..\packages\StyleCop.Analyzers.1.0.2\analyzers\dotnet\cs\StyleCop.Analyzers.dll" />
   </ItemGroup>
+  <ItemGroup />
```
mjcheetham: nit: Remove the empty `<ItemGroup/>` please 😄 

```suggestion
```

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+using GVFS.Common.NamedPipes;
+using GVFS.Common.Tracing;
+using System;
+using System.Diagnostics;
+
+namespace GVFS.Service.UI
+{
+    public class GVFSToastRequestHandler
+    {
+        private const string GVFSAutomountStartTitle= "GVFS Automount";
```
mjcheetham: Shouldn't these be named "VFS ..."?

```suggestion
        private const string GVFSAutomountStartTitle= "VFS Automount";
```

mjcheetham: Same applies to the rest of the strings here (and the property names?).

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+        private const string GVFSAutomountButtonTitle = "Retry";
+
+        private const string GVFSUpgradeTitleFormat = "New version {0} is available";
+        private const string GVFSUpgradeMessage = "Upgrade will unmount and remount gvfs repos, ensure you are at a stopping point. When ready, click Upgrade button to run upgrade.";
+        private const string GVFSUpgradeButtonTitle = "Upgrade";
+
+        private const string GvfsRemountActionPrefix = "gvfs mount ";
+        private const string GvfsUpgradeActionPrefix = "gvfs upgrade --confirm";
+
+        private ITracer tracer;
```
mjcheetham: Can make these fields `readonly`.

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+                        message = string.Format(GVFSAutomountSuccessMessageFormat, Environment.NewLine, request.Enlistment);
+                        break;
+                    }
+
+                case NamedPipeMessages.Notification.Request.Identifier.MountFailure:
+                    {
+                        title = GVFSAutomountErrorTitle;
+                        message = string.Format(GVFSAutomountErrorMessageFormat, Environment.NewLine, request.Enlistment);
+                        buttonTitle = GVFSAutomountButtonTitle;
+                        args = $"{GvfsRemountActionPrefix} {request.Enlistment}";
```
mjcheetham: Do we have to worry about escaping the enlistment here (spaces in the path) or is that done elsewhere?

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+        public void UserResponseCallback(string args)
+        {
+            if (!string.IsNullOrEmpty(args))
+            {
+                string gvfsArgs = null;
+                bool elevate = false;
+
+                if (GVFSPlatform.Instance.IsElevated())
+                {
+                    this.tracer.RelatedInfo($"GVFS.Serice.UI is currently running elevated.");
```
mjcheetham: Does this mean anyone can send a toast notification to `GVFS.Service.UI.exe` and have their code run elevated? I know we only accept some predefined 'actions' for the list of supported notification types, but if there are ways of breaking out, we should be careful here!

This is why I ask about escaping the 'enlistment path' args that we get back from a notification callback, if we're just shelling out to a possibly elevated cmd.exe.

GVFS/GVFS.Installer.Windows/Setup.iss
alameenshah: Updated it to `VFS For Git`

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
alameenshah: Updated the message text and properties in this file. (But there are so many properties, classes and files that use gvfs in their names today. )

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
alameenshah: Added path validation both while sending and while receiving callbacks from toaster.

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
alameenshah: >>Does this mean anyone can send a toast notification to GVFS.Service.UI.exe and have their code run elevated?

- `GVFS.Service.UI` is running un-elevated. So it cannot. (However it runs `gvfs upgrade --confirm` command elevated after user approval).
- Also added path validation.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -33,6 +34,8 @@ protected override string VerbName
 
         protected override void Execute(GVFSEnlistment enlistment)
         {
+            Debugger.Launch();
```
kewillford: Do we still need this?

GVFS/GVFS.Service/GVFSService.Windows.cs
```diff
@@ -365,5 +370,40 @@ private DirectorySecurity GetServiceDirectorySecurity(string serviceDataRootPath
...
         }
+
+        private void LaunchServiceUIIfNotRunning(int sessionId)
+        {
+            NamedPipeClient client;
+            using (client = new NamedPipeClient(GVFSConstants.Service.UIName))
+            {
+                if (!client.Connect())
+                {
+                    this.TerminateExistingProcess(GVFSConstants.Service.UIName);
```
kewillford: Help me understand why we are killing the process and restarting it if we can't connect?

GVFS/GVFS.Service/GVFSService.Windows.cs
```diff
@@ -320,23 +326,22 @@ private void CreateAndConfigureProgramDataDirectories()
             Directory.SetAccessControl(serviceDataRootPath, serviceDataRootSecurity);
 
             // Special rules for the upgrader logs, as non-elevated users need to be be able to write
-            this.CreateAndConfigureUpgradeLogDirectory();
+            this.CreateAndConfigureLogDirectory(ProductUpgraderInfo.GetLogDirectoryPath());
+            this.CreateAndConfigureLogDirectory(GVFSPlatform.Instance.GetDataRootForGVFSComponent(GVFSConstants.Service.UIName));
         }
 
-        private void CreateAndConfigureUpgradeLogDirectory()
+        private void CreateAndConfigureLogDirectory(string path)
```
kewillford: Do we need to pass something to tell us what path this is?  Either the upgrader or the service or will the name of the log be enough for us to know which one?

GVFS/GVFS.Service.UI/WinToastNotifier.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            ToastNotifier toastNotifier = ToastNotificationManager.CreateToastNotifier(this.appId);
+            toastNotifier.Show(toastNotification);
+        }
+
+        private void ToastActivated(ToastNotification sender, object e)
+        {
+            ToastActivatedEventArgs args = (ToastActivatedEventArgs)e;
+
+            this.UserResponseCallback?.Invoke(args.Arguments);
```
kewillford: Do we need a `try/catch` around this or in the callback method?

GVFS/GVFS.Service/GVFSService.Windows.cs
alameenshah: This is old code that is reused ([here is the original](https://github.com/microsoft/VFSForGit/blob/686b3b1f82021b515bf1d4dce728b9b90925a13a/GVFS/GVFS.Service/Handlers/NotificationHandler.Windows.cs#L19) in master.)

And probably is not needed. @wilbaker thoughts?

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
```diff
@@ -0,0 +1,170 @@
...
+            string message = null;
+            string buttonTitle = null;
+            string args = null;
+            string path = null;
+
+            switch (request.Id)
+            {
+                case NamedPipeMessages.Notification.Request.Identifier.AutomountStart:
+                    title = GVFSAutomountStartTitle;
+                    message = string.Format(GVFSAutomountStartMessageFormat, request.EnlistmentCount);
```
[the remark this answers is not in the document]
alameenshah: I'll then have to call it for each `switch` `case`.  This pattern (customize the local variables for each `switch case` and call `Notify()` just once in the end) is simpler.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
alameenshah: Nope. Removed.

GVFS/GVFS.Service/GVFSService.Windows.cs
alameenshah: `CreateAndConfigureLogDirectory` is a generic method. It doesn't need to know what process/function the log-directory is associated with.

GVFS/GVFS.Service.UI/WinToastNotifier.cs
alameenshah: Yes... I've added it around the `Process.Start` inside the callback. The other place that could throw exception is `TryValidatePath` which already has a `catch` block.

GVFS/GVFS.Service/GVFSService.Windows.cs
wilbaker: > And probably is not needed. @wilbaker thoughts?

I'm not familiar with the service UI code.  Have you tried going back in history (before the move to GitHub) to see if you can find any discussion on why the code is structured this way?

GVFS/GVFS.Service.UI/GVFSToastRequestHandler.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            const string cmdPath = "CMD.exe";
+            ProcessStartInfo processInfo = new ProcessStartInfo(cmdPath);
+            processInfo.UseShellExecute = true;
+            processInfo.RedirectStandardInput = false;
+            processInfo.RedirectStandardOutput = false;
+            processInfo.RedirectStandardError = false;
+            processInfo.WindowStyle = ProcessWindowStyle.Normal;
+            processInfo.CreateNoWindow = false;
+            processInfo.Arguments = "/K " + fullGvfsCmd;
```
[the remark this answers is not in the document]
alameenshah: Yes. This is especially useful during an error, user gets the time to read it on the `Command` window and then close it manually.

I will add a comment in the code as well.

alameenshah:
> Just some simple questions/comments. I was also unable to get the toasts to work. Do I need to do anything besides installing using the GVFS installer? I only had 1 repo that was automounted so maybe it happened too fast?

No, just installation is enought. We don't show notifications on Mount Failures (Not success) and when an upgrade is available. Can you simulate a mount failure, or an upgrade available situation and see.

GVFS/GVFS.Installer.Windows/Setup.iss
```diff
@@ -769,6 +803,7 @@ begin
       Abort();
     end;
   StopService('GVFS.Service');
+  StopGVFSServiceUI();
```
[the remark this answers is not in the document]
alameenshah: `GVFS.Service.UI` is not a service. It is launched as hidden process in the current users sessionId.

GVFS/GVFS.Service.UI/Program.cs
```diff
@@ -2,15 +2,9 @@
 using GVFS.Common.NamedPipes;
```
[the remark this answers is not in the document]
alameenshah: Moved `GVFS.Service.UI` to a separate file. `Program.cs` now only has the `Main` entry.

PR #1503, opened 2019-09-03
restore mixed-case prefetch folder list tests and tidy MirrorProviders

chrisd8088:
On case-insensitive filesystems, we can restore the use of non-case-matching mixed-case file and folder paths in several `PrefetchVerbTests`, while retaining case-sensitive exact matches on Linux.

These changes are a follow-up based on PR feedback from @wilbaker on PR #1412 and commit 5c78902.

We also use the `PathComparison` we set in our `MirrorProvider`s to actually perform path comparisons, and add a `PathComparer` for use on Windows in particular.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
@@ -44,7 +44,7 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
             "foo.cpp", // Changes to a folder in one test
             "FilenameEncoding",
             "GitCommandsTests",
-            "GVFLT_MultiThreadTest", // Changes case in one test
+            "GVFLT_MultiThreadTest", // Required by DeleteFolderAndChangeBranchToFolderWithDifferentCase test in sparse mode
             "GVFlt_BugRegressionTest",
             "GVFlt_DeleteFileTest",
             "GVFlt_DeleteFolderTest",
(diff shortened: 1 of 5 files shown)


chrisd8088:
@wilbaker -- This is the small extra PR I promised before our respective breaks; it should restore the functionality you wanted to retain in the functional tests (from #1412).  To get that I pulled out the hard-coded path lists and made a variable list based on the case-sensitivity flag.  I hope that's reasonably clear, but let me know if you see problems.  And thank you, again!

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitRepoTests.cs
```diff
@@ -44,7 +44,7 @@ public abstract class GitRepoTests
             "foo.cpp", // Changes to a folder in one test
             "FilenameEncoding",
             "GitCommandsTests",
-            "GVFLT_MultiThreadTest",
+            "GVFLT_MultiThreadTest", // Changes case in one test
```
wilbaker: > Changes case in one test

I'm not sure what this comment means, could you add a few more details to the comment?

chrisd8088: Sure, will do when I get back from an offsite meeting -- stay tuned, and thank you again!

chrisd8088: I hope I've clarified that one comment sufficiently!

wilbaker: Thanks looks good!

chrisd8088: OK -- will merge shortly!  Thank you!

PR #1507, opened 2019-09-04
Check the disposing flag in Dispose methods

kewillford:
When using `Dispose(bool disposing)`, the disposing flag needs to be checked so that objects that may have already been disposed of through garbage collection will not be used and cause a ObjectDisposedException.  When the disposing flag is `true` it is okay to call dispose on other member objects but when it is `false` it is being disposed through the finalizer by the garbage collector and the member objects may have already been disposed by garbage collection and should not be used.

In most cases in our code we are not following the exact pattern for dispose but it is okay because we are not holding onto unmanaged resources and are not implementing a finalizer to release the unmanaged resources.  

Please see [this article](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) for the suggested way to implement and use the dispose pattern.

### DIFF
--- a/GVFS/GVFS.Common/GVFSContext.cs
+++ b/GVFS/GVFS.Common/GVFSContext.cs
@@ -28,6 +28,7 @@ namespace GVFS.Common
         public void Dispose()
         {
             this.Dispose(true);
+            GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool disposing)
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Common/GVFSContext.cs
```diff
@@ -28,6 +28,7 @@ public GVFSContext(ITracer tracer, PhysicalFileSystem fileSystem, GitRepo reposi
         public void Dispose()
         {
             this.Dispose(true);
+            GC.SuppressFinalize(this);
```
jamill: Technically, I _think_ this is not necessary (and is a no-op), as this method does not have a finalizer, and hence no finalizer to suppress.

kewillford: Yeah but to keep it consistent with the suggested implementation and if by some tiny chance a finalizer is added we won't have to remember to add this.

PR #1510, opened 2019-09-05
Update upgrade messages for platforms that do not support NuGet package verification

jamill:
NuGet package verification is not supported by the NuGet client / net core on some platforms - in particular Linux / macOS. There has been some confusion around the messages we display. This PR is to improve the messaging to reduce confusion. I am open to feedback on the messaging. I listed some of the issues I could see with our current messaging at the bottom. I am also keeping in mind how much work it would be to improve the messages. The current PR is a more tactical change, but I could see that some of the issues might require more in-depth changes (and hence are not addressed in this PR) - but if we feel there is a driving need to update them, let's consider it.

There is a matrix of scenarios for us to consider based on the arguments passed to `gvfs upgrade` and whether it is run with elevated permissions:

| Confirm | NoVerify | Run elevated |
| -------- | -------- | ------------- |
 

With this change, here is the current output for several scenarios:

Scenario 1: `gvfs upgrade` (with or without `--no-verify` argument)

```
$ gvfs upgrade
New version 0.6.19224.2 is available.

When ready, run `sudo gvfs upgrade --confirm --no-verify` to upgrade.
```

Scenario 2: `gvfs upgrade` with `--confirm` argument, but not run elevated. With or without `--no-verify` argument

```
$ gvfs upgrade --confirm
Cannot upgrade GVFS on this machine.

ERROR: The installer needs to be run with elevated permissions.
Run `sudo gvfs upgrade --confirm --no-verify`.
```

Scenario 3: `gvfs upgrade` with confirm argument, run with elevated permissions:

```
$ sudo gvfs upgrade --confirm
New version 0.6.19224.2 is available.

NuGet package verification is not supported on this platform. In order to run upgrade, you must run upgrade without NuGet package verification.
To run upgrade without NuGet verification include the --no-verify option.
```

Issues:
1) In scenario 2, there are 2 items that prevent the user from upgrading - the command is not run elevated, and the `--no-verify` argument not supplied. We call out that the command needs to be run elevated, and the command we display includes the `--no-verify` option, but we don't otherwise call out the `--no-verify` argument.

2) The command that we recommend to run (`sudo gvfs upgrade --confirm --no-verify`) includes the `--no-verify` argument, but we do not call out what the user should consider before running this command as we do in the output for scenario 3.

3) Scenario 2 includes messaging that the user "Cannot upgrade GVFS on this machine", which is not entirely accurate. They just need to run the command elevated.

### DIFF
--- a/GVFS/GVFS/CommandLine/UpgradeVerb.cs
+++ b/GVFS/GVFS/CommandLine/UpgradeVerb.cs
@@ -68,7 +68,7 @@ namespace GVFS.CommandLine
             "no-verify",
             Default = false,
             Required = false,
-            HelpText = "This parameter is reserved for internal use.")]
+            HelpText = "Do not verify NuGet packages after downloading them. Some platforms do not support NuGet verification.")]
         public bool NoVerify { get; set; }
 
         protected override string VerbName
--- a/GVFS/GVFS/CommandLine/UpgradeVerb.cs
+++ b/GVFS/GVFS/CommandLine/UpgradeVerb.cs
@@ -198,6 +198,18 @@ namespace GVFS.CommandLine
                     return false;
                 }
 
+                // If we are on a platform that does not support nuget verification, and user has not
+                // specified the --no-verify flag, then print a helpful message here.
+                if (!GVFSPlatform.Instance.UnderConstruction.SupportsNuGetVerification && !this.NoVerify)
+                {
+                    string packageVerificationNotSupportedMessage = @"
+NuGet package verification is not supported on this platform. In order to run upgrade, you must run upgrade without NuGet package verification.
+To run upgrade without NuGet verification include the --no-verify option.
+";
+                    this.ReportInfoToConsole(packageVerificationNotSupportedMessage);
+                    return false;
+                }
+
                 if (!this.TryRunInstaller(out message))
                 {
                     this.tracer.RelatedError($"{nameof(this.TryRunProductUpgrade)}: Could not launch upgrade tool. {message}");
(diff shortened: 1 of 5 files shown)


GVFS/GVFS/CommandLine/UpgradeVerb.cs
```diff
@@ -198,6 +198,18 @@ private bool TryRunProductUpgrade()
                     return false;
                 }
 
+                // If we are on a platform that does not support nuget verification, and user has not
+                // specified the --no-verify flag, then print a helpful message here.
+                if (!GVFSPlatform.Instance.UnderConstruction.SupportsNuGetVerification && !this.NoVerify)
```
kewillford: If the platform supports nuget verification should we disallow the `--no-verify` or are there use cases for that?

jamill: The main use case is during development / authoring the upgrade package - it allows you test the upgrade package without running it through the full signing pipeline. I don't know of a use case for the mainline consumer.

I think it is useful enough for testing upgrade packages that it is worth leaving in (there are not that many switches) - but I don't feel strongly.

kewillford: I was mainly wondering about security and if leaving that option opens any attack vectors?

wilbaker: @kewillford @jrbriggs this is the only remaining comment on this PR, and outside of it I think the changes in this PR are ready to go.

@kewillford, I don't think the scenario you've brought up is something new (introduced by these changes), correct?  I'm wondering if we can/should leave these changes as-is for now and follow a file up issue to determine if the `--no-verify` flag should be supported.

Thoughts?

kewillford: @wilbaker yep sounds good to me.

wilbaker: Filed #1548 for this.

PR #1511, opened 2019-09-05
Add prune argument to the sparse verb

### LINKED ISSUE #1461: GVFS Sparse - Extra folders outside of sparse list showing up
After successfully running "gvfs sparse -a folder1" and "gvfs sparse -l" shows folder1, I have a bunch of other folders still in my repo. 

Response from @kewillford:

I’m guessing that you were using a previous repo and not a new clone and that there were modified files in the paths that continue to show up.  Right now there is not a way to clean up the modified files and the paths that have been affected until we have the dehydration of folders feature.  After we have that feature we can add an automatic cleanup after running a sparse command but this would also affect #2 because we would be removing all those files and would need to know they are not modified before that.

But my question is - git status was clean which means there were no uncommitted modified files - what exactly do you mean by "modified files?" - files in this repo that have ever been modified?

kewillford:
This change adds a `--prune` argument to the sparse verb.
This will take the current list of sparse folders and anything that is not in that set of folders will be removed via a call to the dehydrate verb using the `--folders` option.

Resolves #1461

### DIFF
--- a/GVFS/GVFS/CommandLine/DehydrateVerb.cs
+++ b/GVFS/GVFS/CommandLine/DehydrateVerb.cs
@@ -255,7 +255,7 @@ from a parent of the folders list.
                                 }
                                 else
                                 {
-                                    string fullPath = Path.Combine(enlistment.WorkingDirectoryRoot, folder);
+                                    string fullPath = Path.Combine(enlistment.WorkingDirectoryBackingRoot, folder);
                                     if (this.fileSystem.DirectoryExists(fullPath))
                                     {
                                         if (!this.TryIO(tracer, () => this.fileSystem.DeleteDirectory(fullPath), $"Deleting '{fullPath}'", out ioError))
@@ -448,35 +448,6 @@ from a parent of the folders list.
             }
 
             return false;
-        }
-
-        private void Unmount(ITracer tracer)
-        {
-            if (!this.ShowStatusWhileRunning(
-                () =>
-                {
-                    return
-                        this.ExecuteGVFSVerb<StatusVerb>(tracer) != ReturnCode.Success ||
-                        this.ExecuteGVFSVerb<UnmountVerb>(tracer) == ReturnCode.Success;
-                },
-                "Unmounting",
-                suppressGvfsLogMessage: true))
-            {
-                this.ReportErrorAndExit(tracer, "Unable to unmount.");
-            }
-        }
-
-        private void Mount(ITracer tracer)
-        {
-            if (!this.ShowStatusWhileRunning(
-                () =>
-                {
-                    return this.ExecuteGVFSVerb<MountVerb>(tracer) == ReturnCode.Success;
-                },
-                "Mounting"))
-            {
-                this.ReportErrorAndExit(tracer, "Failed to mount after dehydrating.");
-            }
         }
 
         private void PrepareSrcFolder(ITracer tracer, GVFSEnlistment enlistment)
@@ -679,43 +650,6 @@ from a parent of the folders list.
                 {
                     { TracingConstants.MessageKey.InfoMessage, message }
                 });
-        }
-
-        private ReturnCode ExecuteGVFSVerb<TVerb>(ITracer tracer)
-            where TVerb : GVFSVerb, new()
-        {
-            try
-            {
-                ReturnCode returnCode;
-                StringBuilder commandOutput = new StringBuilder();
-                using (StringWriter writer = new StringWriter(commandOutput))
-                {
-                    returnCode = this.Execute<TVerb>(this.EnlistmentRootPathParameter, verb => verb.Output = writer);
-                }
-
-                tracer.RelatedEvent(
-                    EventLevel.Informational,
-                    typeof(TVerb).Name,
-                    new EventMetadata
-                    {
-                        { "Output", commandOutput.ToString() },
-                        { "ReturnCode", returnCode }
-                    });
-
-                return returnCode;
-            }
-            catch (Exception e)
-            {
-                tracer.RelatedError(
-                    new EventMetadata
-                    {
-                        { "Verb", typeof(TVerb).Name },
-                        { "Exception", e.ToString() }
-                    },
-                    "ExecuteGVFSVerb: Caught exception");
-
-                return ReturnCode.GenericError;
-            }
         }
 
         private bool TryIO(ITracer tracer, Action action, string description, out string error)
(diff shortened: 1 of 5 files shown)


jamill:
Couple of quick initial thoughts / questions - I will take a deeper look later through the code.

1) I wasn't initially sure what "clean" meant in this context. After reading the docs, it is clear - so just throwing out some other names for your consideration:

- prune: as in "prune" entries that are not part of the sparse list

- apply / force-apply: as in "apply" the sparse list and remove other entries

2) Can / should we apply this logic as part of running the sparse command, without having to explicitly include `--clean`? Are there reasons why a user would run the sparse command but still want to have other directories outside of the sparse list arounds?

kewillford:
Thanks @jamill for your thoughts!

Thanks for the additional suggestions.  Another one I have thought of is "reset" which somewhat follows git's reset.  There are 2 concepts that are needed.
1. Getting the repo back to some state of sparseness.
2. Quickly removing the sparse feature and start projecting everything.

I've been trying to think if we can have one argument that does both but seems to me we will need 2 but if you have ideas there, I would love to hear them.  I suppose we could have:
```
1. gvfs sparse --reset "removes all items outside the current sparse list"
2. gvfs sparse --reset * "removes all sparse entries making the repo no longer sparse"
3. gvfs sparse --reset <folders> "removes all items outside the set of folders specified"
```
1 and 2 could be switched.  I'm not sure which one is more intuitive.  Let me know what you think.

>Can / should we apply this logic as part of running the sparse command, without having to explicitly include --clean?

We could but there has to be the option still because `gvfs sparse` outputs the list and users will need a way to "clean" without adding or removing any folders so running `gvfs sparse --clean` 

>Are there reasons why a user would run the sparse command but still want to have other directories outside of the sparse list arounds?

I don't know and I guess that is why flexibility is better at this point.  I don't want to force something on the user that they may want.

jamill:
Do we need a single verb to cover 2 / 3 different actions? I thought it might be confusing to have reset do both 1) remove all items outside the current sparse list (effectively making it more sparse?) and 2) remove all sparse entries making the repo no longer sparse.

As an possible alternative, a distinct set of subcommands / options:

`gvfs sparse --set <folderlist>` - would set the sparse list to the folder list

`gvfs sparse --prune` (or `--clean`) - remove all items outside of current sparse list

`gvfs sparse --make-unsparse` - make repo no longer sparse. Or is there an equivalent command in git we could use as inspiration?

might make the actions more clear?

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -153,6 +147,81 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+            foldersToEnumerate.Enqueue(rootPath);
+
+            List<string> foldersToRemove = new List<string>();
+            while (foldersToEnumerate.Count > 0)
+            {
+                string folderToEnumerate = foldersToEnumerate.Dequeue();
+                foreach (string directory in fileSystem.EnumerateDirectories(folderToEnumerate))
+                {
+                    string enlistmentRootRelativeFolderPath = GVFSDatabase.NormalizePath(directory.Substring(rootPath.Length));
+                    if (sparseFolders.Any(x => x.StartsWith(enlistmentRootRelativeFolderPath + Path.DirectorySeparatorChar)))
```
jamill: Do we need an explicit string comparer here?

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -153,6 +147,81 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+            }
+        }
+
+        private string[] GetDirectoriesOutsideSparse(string rootPath, HashSet<string> sparseFolders)
+        {
+            PhysicalFileSystem fileSystem = new PhysicalFileSystem();
+            Queue<string> foldersToEnumerate = new Queue<string>();
+            foldersToEnumerate.Enqueue(rootPath);
+
+            List<string> foldersToRemove = new List<string>();
```
jamill: This method does not actually remove folders - maybe something like `directoriesOutsideSparse` or `foldersOutsideSparse`?

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -153,6 +147,81 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+            string[] directoriesToDehydrate = this.GetDirectoriesOutsideSparse(enlistment.WorkingDirectoryBackingRoot, sparseFolders);
+            if (!this.ShowStatusWhileRunning(
+                () =>
+                {
+                    ReturnCode verbReturnCode = this.ExecuteGVFSVerb<DehydrateVerb>(
+                        tracer,
+                        verb =>
+                        {
+                            verb.Confirmed = true;
+                            verb.Folders = string.Join(";", directoriesToDehydrate);
```
jamill: I notice that we have hardcoded the string literal used to separate directory entries in several places (thinking back to the original PR) - should we use a named constant for this?

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -46,6 +46,14 @@ public class SparseVerb : GVFSVerb.ForExistingEnlistment
             HelpText = "List of folders in the sparse set for determining what to project.")]
         public bool List { get; set; }
 
+        [Option(
+            'c',
+            "clean",
+            Required = false,
+            Default = false,
+            HelpText = "Clean up any folders that are not in the list of sparse folders.")]
```
jamill: Is it clear what `Clean` means in this context? would `Remove folders that are not in the sparse list` make it more obvious that we will remove them?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -255,7 +255,7 @@ private void DehydrateFolders(JsonTracer tracer, GVFSEnlistment enlistment, stri
                                 }
                                 else
                                 {
-                                    string fullPath = Path.Combine(enlistment.WorkingDirectoryRoot, folder);
+                                    string fullPath = Path.Combine(enlistment.WorkingDirectoryBackingRoot, folder);
```
derrickstolee: Why is this change important?

kewillford: On Windows and Mac `WorkingDirectoryRoot == WorkingDirectoryBackingRoot` but on Linux that is not the case and when working with the unmounted repo it is in the `WorkingDirectoryBackingRoot`.  So this really is a fix for dehydrate folders to make sure it will work on Linux.

derrickstolee: Ah. I didn't realize we are unmounted at this point. Thanks!

PR #1512, opened 2019-09-05
[RFC] GVFS config: handle scenarios when gvfs config file is not present

jamill:
### Problem
Listing the GVFS configuration via the `gvfs config --list` command as a regular user will result in a program crash if the gvfs.config file has not already been created. This is not a common setup in current customer scenarios, as we usually provide a set of configurations with our deployments. However, basic installs of _just_ VFS for Git do not come with any configuration, and is susceptible to this issue.

A regular user should be able to run `gvfs config --list` without crashing.

### Cause
The GVFS config object uses `FileBasedDictionary` as its backing
store. FileBasedDictionary is backed by a file on disk, and it expects
the file to exist on disk, or to be able create one if it doesn't exist -
even if it has nothing to persist.

GVFS config has slightly different expectations - the backing file (gvfs.config) does not need to exist just to read the current configuration settings (we can assume there are no configuration settings), and it does not need to create the file just to read an empty configuration list. In order to modify this file, the `gvfs config` command has to be run with elevated permissions.

To support the GVFS config scenario of being able to report config settings when the gvfs.config file does not exist yet (and cannot be written), we teach `FileBasedDictionary` to optionally not attempt create the backing file if it does not already exist upon reading, and instead treat it as an empty collection.

I considered several other approaches, and am open to feedback on the preferred approach.

###  Alternative approach 1:
One alternative would be to not make this an optional behavior of
FileBasedDictionary, but rather move the GVFS config object to not use
FileBasedDictionary. This would require either re-implementing the
existing logic to read this file format, or to migrate to a simpler
format, but then we would need to migrate existing config files as
well.

### Alternative approach 2: 
Another approach could be to have this file written out by a
privileged process (either the installer or the service) so we can expect it to always be present.

## TODO:
- [ ] Add tests (after feedback on general approach)


kewillford:
This makes sense to me.  Do you know what code paths will try to save and get errors for the same reason?  Do we need to protect against that or is the logging we have sufficient?

derrickstolee:
This problem was fixed by #1625.

PR #1516, opened 2019-09-09
Sparse status loosen

### LINKED ISSUE #1460: Gvfs sparse -a fails with "dirty" git status even if "dirty" files are within sparse list
I’ve confirmed that gvfs sparse -a seems to fail whenever your repo is dirty at all, regardless of whether or not the added/edited files are in your sparse list at all. I realize the decision to fail if the repo is dirty at all opens up some weird edge cases, however I think it is way too brittle, especially since we are talking about having tooling (git hooks and scoper) call gvfs sparse unbeknownst to the user. Would a better approach be to only fail gvfs sparse if there are dirty folders present that are outside of the sparse list you are switching to?

kewillford:
Resolves #1460 

This change will allow the sparse set to change even when `git status` is not clean if the only "dirty" files are still within the sparse set after the change is applied.

This required a couple of changes to note:
1. Started using the `git status --porcelain -z -uall` so the status information can be parsed
2. Removed the force checkout of HEAD to update the projection but instead use the postindexchanged hook so that the changes in the index will be kept.

Todo:
- [ ] Add some more tests with new/modified/deleted files as well as changes added to the index


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -482,6 +482,12 @@ public Result Status(bool allowObjectDownloads, bool useStatusCache, bool showUn
             return this.InvokeGitInWorkingDirectoryRoot(command, useReadObjectHook: allowObjectDownloads);
         }
 
+        public Result StatusPorcelain()
```
derrickstolee: Should we have some kind of helper method and basic struct so callers don't need to roll their own parser?

kewillford: I thought about that and ended up not doing that since this is the only case that needs this functionality right now.  If there ends up being another use case then I would definitely do something like that.  If you feel like this would be used in other places, then I will make it more generic.  I know there are places that we are running status but only checking for `nothing to commit, working tree clean` which is really the wrong thing to do and we should be using the porcelain command.  Let me know what you think.

derrickstolee: Let's leave that as cleanup for later (i.e. when status changes output and we are forced to switch to porcelain)

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -167,11 +171,28 @@ private void ForceProjectionChange(ITracer tracer, GVFSEnlistment enlistment)
...
+                    {
+                        using (NamedPipeClient pipeClient = new NamedPipeClient(enlistment.NamedPipeName))
+                        {
+                            if (!pipeClient.Connect())
+                            {
+                                this.ReportErrorAndExit("Unable to connect to GVFS.  Try running 'gvfs mount'");
+                            }
+
+                            NamedPipeMessages.PostIndexChanged.Request request = new NamedPipeMessages.PostIndexChanged.Request(updatedWorkingDirectory: true, updatedSkipWorktreeBits: false);
+                            pipeClient.SendRequest(request.CreateMessage());
```
derrickstolee: Interesting that you run a post-index-changed message instead of `git checkout`, but that's a safe way to communicate with the Mount process without changing contents.

GVFS/GVFS.Common/NamedPipes/NamedPipeMessages.cs
```diff
@@ -158,9 +158,25 @@ public Request(Message message)
                     this.UpdatedSkipWorktreeBits = message.Body[1] == '1';
                 }
 
+                public Request(bool updatedWorkingDirectory, bool updatedSkipWorktreeBits)
```
derrickstolee: Just for others to avoid questions: the hook uses the `public Request(Message message)` overload above, so this new constructor does not require reaction work there.

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -88,26 +91,27 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                            needToChangeProjection = true;
+                            directories.Add(folder);
                         }
                     }
 
                     if (needToChangeProjection)
                     {
                         // Make sure there is a clean git status before allowing sparse set to change
-                        this.CheckGitStatus(tracer, enlistment);
+                        if (directories.Count > 0)
```
wilbaker: Why check `directories.Count > 0` before calling `CheckGitStatus`?  If `directories` is empty does that mean nothing is projected or everything is projected?

kewillford: That means that everything will be projected so there isn't any use checking status because the projection change will not affect any of the dirty files no matter where they are.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1554,6 +1554,14 @@ private string GetNewProjectedShaForPlaceholder(string path)
                             {
                                 this.placeholderDatabase.AddPartialFolder(childRelativePath);
                             }
+                            else if (result.Result == FSResult.IOError)
+                            {
+                                if (this.context.FileSystem.DirectoryExists(Path.Combine(this.context.Enlistment.WorkingDirectoryRoot, childRelativePath)))
+                                {
+                                    this.ReExpandFolder(childRelativePath, existingPlaceholders);
```
wilbaker: Offline you mentioned:

> since I changed to no longer call checkout -f and instead use the postindexchanged there were folders that were on disk but not in the placeholder list and would get an error when trying to write the placeholder directory

But I'm not quite following the scenario.  How/why are there already folders on disk, and how is it related to no longer calling `checkout -f`?  Are these folders full folders?

I was also wondering if there's any chance that the folders we're re-expanding here (or their children) are already in the placeholder list as expanded?

kewillford: The test that manifested this issue was `DeleteFileAndCommitThenChangingSparseFoldersShouldKeepFolderAndFile` which:
1. Deletes `GVFS\GVFS.Tests\packages.config` and commits that change so the path for that file is in the modified paths.  
2. It then adds `GVFS\GVFS` as a sparse folder so the file that was deleted is outside the sparse set and verifies that the file is still deleted.  
3. Then it checks out `HEAD~1` and verifies that the file comes back which it does because it is in the modified paths.  This checkout causes the `GVFS\GVFS.Tests` folder to be removed from the placeholders because it is not part of the projection since only `GVFS\GVFS` is being projected.
4. In the teardown all the sparse folders are removed which should return the repo to being fully projected, but when reexpanding the `GVFS` folder it tries to make a placeholder directory for `GVFS.Tests` but it already exists.

I think this would have been failing with the `checkout -f` as well because the `GVFS.Tests` directory would have still been on disk and the `WritePlaceholderDirectory` call would have failed.  So this was a bug in the original sparse code but not caught because the tests were only validating the top level folders and `DeleteFileAndCommitThenChangingSparseFoldersShouldKeepFolderAndFile` was the last test to run so there were not any subsequent tests that might have caught it.

PR #1521, opened 2019-09-13
Add disable option to quickly get out of sparse mode

### LINKED ISSUE #1502: GVFS Sparse -r *
Please enlighten me if I'm missing something but I'm not seeing a way to quickly clear the sparse list. This should be added to quickly turn on/off the feature

kewillford:
This PR builds on #1520 changes. [Here is the diff for just these changes](https://github.com/kewillford/VFSForGit/compare/sparse-set...kewillford:sparse-turn-off)

This adds a `--disable` option that isn't valid with other options that will take the list of directories in the sparse table and remove each one.  I thought about adding a `RemoveAll` to the `SparseTable` but decided that it might be helpful to have the paths that were removed in the log file so this uses the code that was already there simply adding each directory to the `foldersToRemove` list.

Changed the tests to use the `--disable` option in the `Teardown`. 

- [x] Add test that uses the option when sparse mode is already off.

Resolves #1502


wilbaker:
@kewillford these changes look good to me.  I will hold off on approving until the changes in #1520 are finalized.

PR #1524, opened 2019-09-17
GitMaintenanceStep: delete commit-graphs folder before rewriting

derrickstolee:
If "git commit-graph verify" fails, we call the method
LogErrorAndRewritecommitGraph(). This is supposed to get us
into a better state. It appears that it is not doing that
when we have an issue around our incremental commit-graph
format.

Fix that by deleting the commit-graphs folder in addition to
the commit-graph file. Report the status of the call, so we
know if we had such a folder.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
@@ -276,6 +276,14 @@ namespace GVFS.Common.Maintenance
             string commitGraphPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", "commit-graph");
             errorMetadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(commitGraphPath);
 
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Maintenance/GitMaintenanceStep.cs
```diff
@@ -276,6 +276,9 @@ protected void LogErrorAndRewriteCommitGraph(ITracer activity, List<string> pack
             string commitGraphPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", "commit-graph");
             errorMetadata["TryDeleteFileResult"] = this.Context.FileSystem.TryDeleteFile(commitGraphPath);
 
+            string commitGraphsPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, "info", "commit-graphs");
+            errorMetadata["TryDeleteDirectoryResult"] = this.Context.FileSystem.TryDeleteDirectory(commitGraphsPath, out Exception _);
```
wilbaker: Would it be useful to include the `Exception` (if present) in `errorMetadata` as well?

PR #1529, opened 2019-09-22
[POSIX] convert Mac-specific functional tests to POSIX where viable

chrisd8088:
Where appropriate, convert functional tests marked `MacOnly` into `POSIXOnly` tests which also execute on Linux.

The `rename(2)` function is identical on Linux and Mac, so update its `DllImport` wrapper name to `POSIXRename()` from `MacRename()` in `GVFSLockTests`.

Finally, use generic POSIX options to `ln(1`) in `BashRunner` because the `GitStatusReportsSymLinkChanges()` functional test will fail on Linux otherwise, as `ln -s -F` does not have the same meaning as on BSD/Darwin; the `-F` flag does not imply `-f` on
Linux but instead tries to hard link a directory.  Thus the deletion of the target `testFilePath` in the test does not occur and the test fails.  Instead, we use `ln -s -f` in the `CreateSymbolicLink()` method in `BashRunner` which suffices on both types of systems.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Categories.cs
+++ b/GVFS/GVFS.FunctionalTests/Categories.cs
@@ -8,6 +8,7 @@
 
         public const string WindowsOnly = "WindowsOnly";
         public const string MacOnly = "MacOnly";
+        public const string POSIXOnly = "POSIXOnly";
 
         public static class MacTODO
         {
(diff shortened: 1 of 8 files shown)


chrisd8088:
@wilbaker -- This PR now contains just the POSIX-specific bits from #1413 (which has been simplified as a result), as we [discussed](https://github.com/microsoft/VFSForGit/pull/1413#issuecomment-533658318).

And, as a bonus, I found some ways to simplify this set of changes too, so I'm glad I took the time to re-review these.  Let me know if you see anything of concern, though!

chrisd8088:
Thanks for the 👍 @wilbaker!!

PR #1531, opened 2019-09-29
[Linux] minor POSIX and path case comparison fixups

chrisd8088:
Convert a few recent inadvertent uses of `StringComparison.OrdinalIgnoreCase` in file path comparisons to the relevant OS-specific `PathComparison` helpers.

The array returned by `Directory.GetDirectories()` is not consistently ordered across different platforms, so we use `PathComparer` to sort the array prior to checking its contents in `CheckMainSparseFolder()` in the `SparseTests` functional tests.

We can also mark one recent new functional test, `RunPythonExecutable()`, as `POSIXOnly` since it applies to Linux as well as macOS.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
@@ -598,10 +598,9 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         private void CheckMainSparseFolder()
         {
             string[] directories = Directory.GetDirectories(this.Enlistment.RepoRoot);
-            Array.Sort(directories, FileSystemHelpers.PathComparer);
             directories.Length.ShouldEqual(2);
-            directories[0].ShouldEqual(Path.Combine(this.Enlistment.RepoRoot, ".git"));
-            directories[1].ShouldEqual(Path.Combine(this.Enlistment.RepoRoot, "GVFS"));
+            directories.ShouldContain(x => x == Path.Combine(this.Enlistment.RepoRoot, ".git"));
+            directories.ShouldContain(x => x == Path.Combine(this.Enlistment.RepoRoot, "GVFS"));
 
             string folder = this.Enlistment.GetVirtualPathTo(this.mainSparseFolder);
             folder.ShouldBeADirectory(this.fileSystem);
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
```diff
@@ -598,6 +598,7 @@ public void DisableWhenNotInSparseModeShouldBeNoop()
         private void CheckMainSparseFolder()
         {
             string[] directories = Directory.GetDirectories(this.Enlistment.RepoRoot);
+            Array.Sort(directories, FileSystemHelpers.PathComparer);
```
wilbaker: Why are we sorting using `FileSystemHelpers.PathComparer` here if we always are expecting the first two entries to be in a specific order (regardless of platform)?

chrisd8088: Good question!  And in fact I was poking around today and found [some comments](https://social.msdn.microsoft.com/Forums/vstudio/en-US/69762107-2736-4f51-89be-fd5d1cf78a54/which-attribute-would-be-used-to-sort-directoryinfogetdirectories-result-?forum=netfxbcl) which indicate that one shouldn't trust the ordering from `GetDirectories()`; the [`GetFiles()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.getfiles) docs specifically say not to expect consistent sorting, but the `GetDirectories()` docs are silent on that point.

Anyway, that made me think I should probably add sorting on [all the calls](https://github.com/microsoft/VFSForGit/blob/127ec0716f6cf853437ae6aa7bc23cc3f67fba76/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs#L29-L32) to `GetDirectories()` in `SparseTests`.  But I haven't gotten to that yet, and as you say, maybe using a fixed `Ordinal` sort would be better so it's deterministic on all platforms.

Thoughts?  Am I wrong about the output from `GetDirectories()` being unsorted, for one thing?

wilbaker: > Am I wrong about the output from GetDirectories() being unsorted, for one thing?

My interpretation is that the results will be sorted according to the rules of the underlying filesystem, but that they are not guaranteed to be consistent across platform or filesystem.

In most places in `SparseTests` it looks like we're comparing different `GetDirectories` calls against each other, and I would expect that to be OK because the filesystems we support should be consistent (against itself).

For this specific spot (in `CheckMainSparseFolder`), rather than sorting the results it might be simpler to use `ShouldContain` because as you've pointed out the ordering is not guaranteed across platforms.

chrisd8088: Good point -- OK, I can do that instead.  Thank you!

PR #1535, opened 2019-10-01
[Linux] Use backing root and check min version in dehydrate commands/tests

chrisd8088:
The `DehydrateVerb` `TryBackupFiles()` method currently fails on Linux due to the inadvertent use of the `WorkingDirectoryRoot` as the source location for the backup.  This is because on Linux the `src/` working directory is fully unavailable when the provider is unmounted.

Instead, by using the `WorkingDirectoryBackingRoot`, which is always available on all platforms (and is equivalent to the `WorkingDirectoryRoot` on Windows and Mac), we can allow most dehydrate tests to succeed on Linux.

Also, the `DehydrateShouldFailOnWrongDiskLayoutVersion()` functional test currently fails and leaves the test suite in a broken state if it attempts to set the disk layout major version number below a platform's minimum supported version.

By checking whether the existing version is higher than the minimum version first, we can proceed with this test only when safe to do so.  To accommodate this check, we move the platform-specific disk layout major version numbers and minimum version numbers into `GVFSHelpers` from `DiskLayoutVersionTests`.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/DiskLayoutVersionTests.cs
@@ -10,30 +10,7 @@ namespace GVFS.FunctionalTests.Tests
     [Category(Categories.ExtraCoverage)]
     public class DiskLayoutVersionTests : TestsWithEnlistmentPerTestCase
     {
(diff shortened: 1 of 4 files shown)


chrisd8088:
Thanks for these reviews, @wilbaker!!  Much appreciated!

PR #1537, opened 2019-10-01
When using --set add folders back to directories for checking git status

### LINKED ISSUE #1460: Gvfs sparse -a fails with "dirty" git status even if "dirty" files are within sparse list

kewillford:
The set option was removing folder from the directories hashset to know
what folders needed to be removed but the directories hashset was used
in checking status and for pruning so it needs to be the set of folders
that will be in the sparse set after determining the what to add and remove
so that git status and prune will have the correct set of sparse folders.

Fixes #1460

### DIFF
--- a/GVFS/GVFS/CommandLine/SparseVerb.cs
+++ b/GVFS/GVFS/CommandLine/SparseVerb.cs
@@ -166,6 +166,13 @@ Folders need to be relative to the repos root directory.")
                         {
                             needToChangeProjection = true;
                             foldersToRemove.AddRange(directories);
+                            directories.Clear();
+                        }
+
+                        // Need to add folders that will be in the projection back into directories for the status check
+                        foreach (string folder in folders)
+                        {
+                            directories.Add(folder);
                         }
                     }
                     else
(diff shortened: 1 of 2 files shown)


GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -166,6 +166,13 @@ protected override void Execute(GVFSEnlistment enlistment)
                         {
                             needToChangeProjection = true;
                             foldersToRemove.AddRange(directories);
+                            directories.Clear();
+                        }
+
+                        // Need to add folders that will be in the projection back into directories for the status check
+                        foreach (string folder in foldersToAdd)
+                        {
+                            directories.Add(folder);
```
wilbaker: Do we also need to add pack in directories that were in the set before the command was run?

Consider this scenario:

- InitialSet: { A, B, C }
- `set` is called with { B, C, D }

`foldersToRemove ` will be { A } and `foldersToAdd` will be { D } meaning that after this loop `directories` will only have { D }, when it should be { B, C, D }.

Rather than re-adding `foldersToAdd` should we be re-adding `folders`?

If there's not already it might be worth adding a test that would catch this.

wilbaker: To clarify, I think the problem would be if something was modified in { B, C } git status would flag it when it probably shouldn't.

kewillford: I will add a test case for this.

PR #1538, opened 2019-10-01
IsFileSystemSupported: cleanup test files on Mac

wilbaker:
IsFileSystemSupported was leaving around test files it uses to
test if the file system is case sensitive.  Fix up the paths it
uses for its tests.

### DIFF
--- a/GVFS/GVFS.Platform.Mac/MacFileSystem.cs
+++ b/GVFS/GVFS.Platform.Mac/MacFileSystem.cs
@@ -32,20 +32,21 @@ namespace GVFS.Platform.Mac
 
         public override bool IsFileSystemSupported(string path, out string error)
         {
-            string lowerCaseFile = Guid.NewGuid().ToString().ToLower();
-            string upperCaseFile = lowerCaseFile.ToUpper();
             error = null;
 
             try
             {
-                File.Create(Path.Combine(path, lowerCaseFile));
-                if (File.Exists(Path.Combine(path, upperCaseFile)))
+                string lowerCaseFilePath = Path.Combine(path, $"casetest{Guid.NewGuid().ToString()}");
+                string upperCaseFilePath = lowerCaseFilePath.ToUpper();
+
+                File.Create(lowerCaseFilePath);
+                if (File.Exists(upperCaseFilePath))
                 {
-                    File.Delete(lowerCaseFile);
+                    File.Delete(lowerCaseFilePath);
                     return true;
                 }
 
-                File.Delete(lowerCaseFile);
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -32,20 +32,21 @@ public override bool IsSocket(string fileName)
...
-            string lowerCaseFile = Guid.NewGuid().ToString().ToLower();
-            string upperCaseFile = lowerCaseFile.ToUpper();
             error = null;
 
             try
             {
-                File.Create(Path.Combine(path, lowerCaseFile));
-                if (File.Exists(Path.Combine(path, upperCaseFile)))
+                string lowerCaseFilePath = Path.Combine(path, Guid.NewGuid().ToString().ToLower());
+                string upperCaseFilePath = lowerCaseFilePath.ToUpper();
```
kewillford: This shouldn't make a difference but it is now upper casing the entire path instead of just the filename.  It should be a valid test for whether the file system is case sensitive or not still though so it should still work.

wilbaker: Yeah, I decided the code was simpler to upper-case everything in the path.

wilbaker:
> Any tests we can add?

I should be able to add tests that `clone` and `mount` only create what we expect in the root.

GVFS/GVFS.Platform.Mac/MacFileSystem.cs
```diff
@@ -32,20 +32,21 @@ public override bool IsSocket(string fileName)
...
         {
-            string lowerCaseFile = Guid.NewGuid().ToString().ToLower();
-            string upperCaseFile = lowerCaseFile.ToUpper();
             error = null;
 
             try
             {
-                File.Create(Path.Combine(path, lowerCaseFile));
-                if (File.Exists(Path.Combine(path, upperCaseFile)))
+                string lowerCaseFilePath = Path.Combine(path, Guid.NewGuid().ToString().ToLower());
```
derrickstolee: An aside: what are the chances that a GUID has no letters?

  10^32 / 16^32 = (5/8)^32 &approx;  1.7067^(-68)

OK. This is reasonable.

wilbaker: That's a good point, it would be more robust to name the file something like `$"casetest{GUID}"`

Any objections to make that change as part of this?

derrickstolee: Sounds like a simple change. Thanks!

PR #1545, opened 2019-10-02
Add message to sparse disable

### LINKED ISSUE #1458: GVFS sparse "No folders in sparse list" text should be expanded
As I understand it, the way gvfs sparse works is that if there are no folders in the sparse list, everything is projected. So, when I'm in that state, and I run gvfs sparse -l, I get:

No folders in sparse list.

I would like to suggest this get expanded to:

"No folders in sparse list. When the sparse list is empty, all folders are projected"

kewillford:
Give some extra information to the user about what will happen when running the `--disable` option

Resolves #1458

### DIFF
--- a/GVFS/GVFS/CommandLine/SparseVerb.cs
+++ b/GVFS/GVFS/CommandLine/SparseVerb.cs
@@ -106,6 +106,15 @@ Folders need to be relative to the repos root directory.")
                     EventLevel.Informational,
                     Keywords.Any);
 
+                EventMetadata metadata = new EventMetadata();
+                metadata.Add(nameof(this.Set), this.Set);
+                metadata.Add(nameof(this.File), this.File);
+                metadata.Add(nameof(this.Add), this.Add);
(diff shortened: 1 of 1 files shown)


GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -120,6 +120,7 @@ protected override void Execute(GVFSEnlistment enlistment)
                     {
                         if (directories.Count > 0)
                         {
+                            this.Output.WriteLine("Removing all folders from sparse list. When the sparse list is empty, all folders are projected.");
```
wilbaker: Use `this.WriteMessage` here?  It looks like we're writing the other messages in this method to the log.

A bit outside the scope of this PR, but are the values of the parameters being written to the log somewhere (I had trouble finding where they would be)?  If they're not, could we log them?  If we have support issues it would be helpful to see what parameters were passed to the verb.

PR #1550, opened 2019-10-09
Set git config autoRefreshIndex to true

alameenshah:
- setting the git config autoRefreshIndex to true. This is not needed anymore, because current version of gvfs does not use index.lock. (This change was introduced initially to prevent `git diff` from acquiring index.lock file.)

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
@@ -1020,6 +1020,27 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         }
 
         [TestCase]
+        public void ChangeTimestampAndDiff()
+        {
+            // User scenario -
+            // 1. Enlistment's "diff.autoRefreshIndex" config is set to false
+            // 2. A checked out file got into a state where it differs from the git copy
+            // only in its LastWriteTime metadata (no change in file contents.)
+            // Repro steps - This happens when user edits a file, saves it and later decides
+            // to undo the edit and save the file again.
+            // Once in this state, the unchanged file (only its timestamp has changed) shows
+            // up in `git difftool` creating noise. It also shows up in `git diff --raw` command,
+            // (but not in `git status` or `git diff`.)
+
+            // Change the timestamp - The lastwrite time can be close to the time this test method gets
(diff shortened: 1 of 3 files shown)


kewillford:
Can we write a test case for this issue?

wilbaker:
> Can we write a test case for this issue?

I agree, the change looks good but please add a functional test that would repro the bug without this fix.

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -1019,6 +1019,18 @@ public void EditFileNeedingUtf8Encoding()
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.FileSystem, relativeGitPath);
         }
 
+        [TestCase]
+        [Category(Categories.NewTest)]
```
wilbaker: Remove `[Category(Categories.NewTest)]`?  (I am guessing it's debug code that was missed during cleanup)

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
```diff
@@ -1019,6 +1019,17 @@ public void EditFileNeedingUtf8Encoding()
             GVFSHelpers.ModifiedPathsShouldContain(this.Enlistment, this.FileSystem, relativeGitPath);
         }
 
+        [TestCase]
+        public void ChangeTimestampAndDiff()
+        {
+            // Change the timestamp - The lastwrite time can be close to the time this test method gets
+            // run. Changing (Subtracting) it to the past so there will always be a difference.
```
wilbaker: Could you include a few details about scenarios where users would hit this?

In other words, what is the significance of adjusting the last right time?  How does that impact the behavior of `git diff --raw`?

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
wilbaker: Disregard, this was from an earlier push

GVFS/GVFS.FunctionalTests/Tests/GitCommands/GitCommandsTests.cs
alameenshah: Added a brief description.

PR #1553, opened 2019-10-11
Add sparse prune info

kewillford:
This changes adds more information to the user's console output to see what is happening when prune is running.

The one thing I don't really like is that there will be an entry in the sparse log file
```
DehydrateVerb {"Output":null,"ReturnCode":0}
```
since there isn't a `StringBuilder commandOutput` that is capturing the output to write to the logs.
There is a entire log file for the dehydrate that has all the log information though so maybe this is okay.  I could also remove the `"Output":null,` and just have the `ReturnCode` in this case or maybe put the information about the verb that was ran and to look in the logs for that verb.

Updated to have this in the sparse log:
```
DehydrateVerb {"Output":"Check dehydrate logs for output","ReturnCode":0}
```
Also thought about having all the information in the sparse log but that would require more extensive changes by having the base class create the tracer for the verb and all the derived class use that tracer.  Seems that change would be outside the scope of these changes though.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
@@ -16,6 +16,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
     [Category(Categories.ExtraCoverage)]
     public class DehydrateTests : TestsWithEnlistmentPerFixture
     {
+        private const string FolderDehydrateSuccessfulMessage = "folder dehydrate successful.";
         private const int GVFSGenericError = 3;
         private FileSystemRunner fileSystem;
 
@@ -183,7 +184,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             string subFolderToEnumerate = Path.Combine(pathToEnumerate, "GVFS");
             this.fileSystem.EnumerateDirectory(subFolderToEnumerate);
 
-            this.DehydrateShouldSucceed(new[] { "GVFS folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
+            this.DehydrateShouldSucceed(new[] { $"GVFS {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
             this.Enlistment.UnmountGVFS();
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -206,7 +207,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             {
             }
 
-            this.DehydrateShouldSucceed(new[] { "GVFS folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
+            this.DehydrateShouldSucceed(new[] { $"GVFS {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
             this.Enlistment.UnmountGVFS();
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -229,7 +230,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
 
             this.fileSystem.EnumerateDirectory(pathToReadFiles);
 
-            this.DehydrateShouldSucceed(new[] { "GVFS folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
+            this.DehydrateShouldSucceed(new[] { $"GVFS {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
             this.Enlistment.UnmountGVFS();
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -252,7 +253,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             GitProcess.Invoke(this.Enlistment.RepoRoot, "add .");
             GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -m Test");
 
-            this.DehydrateShouldSucceed(new[] { "GVFS folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
+            this.DehydrateShouldSucceed(new[] { $"GVFS {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
             this.Enlistment.UnmountGVFS();
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -273,7 +274,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.fileSystem.DeleteDirectory(pathToDelete);
             GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout -- Scripts");
 
-            this.DehydrateShouldSucceed(new[] { "Scripts folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: "Scripts");
+            this.DehydrateShouldSucceed(new[] { $"Scripts {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: "Scripts");
             this.Enlistment.UnmountGVFS();
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -302,7 +303,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.fileSystem.AppendAllText(fileNotDehydrated2, "Append content");
             GitProcess.Invoke(this.Enlistment.RepoRoot, $"reset --hard");
 
-            this.DehydrateShouldSucceed(new[] { $"{folderToDehydrate} folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: folderToDehydrate);
+            this.DehydrateShouldSucceed(new[] { $"{folderToDehydrate} {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: folderToDehydrate);
 
             this.PlaceholdersShouldNotContain(folderToDehydrate, Path.Combine(folderToDehydrate, "Program.cs"));
             GVFSHelpers.ModifiedPathsShouldNotContain(this.Enlistment, this.fileSystem, Path.Combine(folderToDehydrate, "App.config").Replace(Path.DirectorySeparatorChar, TestConstants.GitPathSeparator));
@@ -329,7 +330,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.fileSystem.ReadAllText(fileToRead2);
 
             this.DehydrateShouldSucceed(
-                new[] { $"{folderToDehydrate1} folder dehydrate successful.", $"{folderToDehydrate2} folder dehydrate successful." },
+                new[] { $"{folderToDehydrate1} {FolderDehydrateSuccessfulMessage}", $"{folderToDehydrate2} {FolderDehydrateSuccessfulMessage}" },
                 confirm: true,
                 noStatus: false,
                 foldersToDehydrate: string.Join(";", folderToDehydrate1, folderToDehydrate2));
@@ -351,7 +352,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.fileSystem.ReadAllText(fileToRead2);
 
             this.DehydrateShouldSucceed(
-                new[] { $"{folderToDehydrate1} folder dehydrate successful.", $"Cannot dehydrate folder '{folderToDehydrate2}': '{folderToDehydrate2}' does not exist." },
+                new[] { $"{folderToDehydrate1} {FolderDehydrateSuccessfulMessage}", $"Cannot dehydrate folder '{folderToDehydrate2}': '{folderToDehydrate2}' does not exist." },
                 confirm: true,
                 noStatus: false,
                 foldersToDehydrate: string.Join(";", folderToDehydrate1, folderToDehydrate2));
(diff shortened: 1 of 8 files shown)


GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -240,23 +240,32 @@ protected override void Execute(GVFSEnlistment enlistment)
...
-                            tracer,
-                            verb =>
-                            {
-                                verb.Confirmed = true;
-                                verb.Folders = string.Join(FolderListSeparator, directoriesToDehydrate);
-                            });
-
-                        return verbReturnCode == ReturnCode.Success;
+                        verb.Confirmed = true;
+                        verb.Folders = string.Join(FolderListSeparator, directoriesToDehydrate);
```
wilbaker: Alternative to consider for passing in `this.Output`.

You could just set it in the verb configuration action here:

```
verb.Output = this.Output
```

And then `ExecuteGVFSVerb` could be updated to only use the newly created writer if one wasn't set by the configure action:

```
    ReturnCode returnCode;
    returnCode = this.Execute<TVerb>(
    this.EnlistmentRootPathParameter,
    verb =>
    {
        configureVerb?.Invoke(verb);
        verb.Output = verb.Output ?? writer;
    });
```

(It would mean we'd occasionally create a `StringWriter` we don't need, but we've always been creating one so it's not making things any worse)

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -240,23 +240,32 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+                this.ReportErrorAndExit(tracer, "Failed to prune.");
+            }
+
+            this.WriteMessage(tracer, $"Found {directoriesToDehydrate.Length} folders to prune.");
+
             if (directoriesToDehydrate.Length > 0)
             {
-                if (!this.ShowStatusWhileRunning(
-                    () =>
+                ReturnCode verbReturnCode = this.ExecuteGVFSVerb<DehydrateVerb>(
```
wilbaker: Should the console messages in Dehydrate either be parameterized or made more general?  

It looks like this message is always printed:

https://github.com/microsoft/VFSForGit/blob/a537d002be8d48675f48e3e97fe1c8c6359133a9/GVFS/GVFS/CommandLine/DehydrateVerb.cs#L166

And there are some other error messages that use the work "dehydrate":

https://github.com/microsoft/VFSForGit/blob/a537d002be8d48675f48e3e97fe1c8c6359133a9/GVFS/GVFS/CommandLine/DehydrateVerb.cs#L209

Should they say prune (or some verb-neutral word)?

GVFS/GVFS/CommandLine/SparseVerb.cs
kewillford: The `GVFSVerb.Output` gets set in the constructor so I don't think it will every be null unless we make more changes and not set it in the constructor but I would need to see if that would have any bad side effects.
```
        public GVFSVerb(bool validateOrigin = true)
        {
            this.Output = Console.Out;
```

wilbaker: I totally missed that, thanks!

GVFS/GVFS/CommandLine/SparseVerb.cs
kewillford: Yeah that would make it easier for the user to understand and we would need possibly different terms for each because you would probably want
`WARNING: If you abort the dehydrate after this point, the repo may become corrupt`
to be
`WARNING: If you abort the sparse after this point, the repo may become corrupt`
but 
`Cannot dehydrate: must have a clean git status.`
to be
`Cannot prune: must have a clean git status.`

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -183,7 +183,7 @@ public void FolderDehydrateFolderThatWasEnumerated()
             string subFolderToEnumerate = Path.Combine(pathToEnumerate, "GVFS");
             this.fileSystem.EnumerateDirectory(subFolderToEnumerate);
 
-            this.DehydrateShouldSucceed(new[] { "GVFS folder successfully dehydrated." }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
+            this.DehydrateShouldSucceed(new[] { "GVFS folder dehydrate successful." }, confirm: true, noStatus: false, foldersToDehydrate: "GVFS");
```
derrickstolee: Should this string be a `static readonly` in the class?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -329,7 +329,7 @@ public void FolderDehydrateNestedFoldersChildBeforeParent()
             this.fileSystem.ReadAllText(fileToRead2);
 
             this.DehydrateShouldSucceed(
-                new[] { $"{folderToDehydrate1} folder successfully dehydrated.", $"{folderToDehydrate2} folder successfully dehydrated." },
+                new[] { $"{folderToDehydrate1} folder dehydrate successful.", $"{folderToDehydrate2} folder dehydrate successful." },
```
derrickstolee: Should these strings also be a constant? 

```
static readonly string FolderDehydrateSuccessfulFormat = "{0} folder dehydrate successful.";

...

this.DehydrateShouldSucceed(
    new []
    {
        String.Format(FolderDehydrateSuccessfulFormat, folderToDehydrate1), 
        String.Format(FolderDehydrateSuccessfulFormat, folderToDehydrate2),
    },
    confirm: true,
    noStatus: false,
    foldersToDehydrate: string.Join(";", folderToDehydrate1, foldertoDehydrate2));
```

kewillford:
azp run PR - Windows - Functional Tests (Sparse Mode)

kewillford:
azp run PR - Windows - Functional Tests (Sparse Mode)

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -334,12 +349,13 @@ private void SendDehydrateMessage(ITracer tracer, GVFSEnlistment enlistment, str
...
                 {
-                    this.WriteMessage(tracer, $"{folder} folder successfully dehydrated.");
+                    this.WriteMessage(tracer, $"{folder} folder {this.DisplayName} successful.");
                 }
 
                 foreach (string folder in response.FailedFolders)
                 {
-                    this.WriteMessage(tracer, $"{folder} folder failed to dehydrate. You may need to reset the working directory by deleting {folder}, running `git reset --hard`, and retry the dehydrate.");
+                    this.WriteMessage(tracer, $"{folder} folder failed to {this.DisplayName}. You may need to reset the working directory by deleting {folder}, running `git reset --hard`, and retry the {this.DisplayName}.");
+                    folderErrors.Add(folder);
```
wilbaker: Is `\0{ioError}` included in the entries in `response.FailedFolders`?

If not, should we an a `\0<error>` here?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -887,19 +889,23 @@ protected ReturnCode ExecuteGVFSVerb<TVerb>(ITracer tracer, Action<TVerb> config
...
-                        { "ReturnCode", returnCode }
-                        });
+                    EventMetadata metadata = new EventMetadata();
+                    if (outputWriter == null)
+                    {
+                        metadata.Add("Output", commandOutput.ToString());
+                    }
+                    else
+                    {
+                        metadata.Add("Output", $"Check {new TVerb().VerbName} logs for output");
```
wilbaker: I don't quite follow this check.  

If `outputWriter` is non-null (e.g. it has been set to `Console.Out`), how is that related to the verb log containing the output?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -92,6 +93,7 @@ public string ServicePipeName
         }
 
         public TextWriter Output { get; set; }
+        public TextWriter ErrorOutput { get; set; }
```
wilbaker: Should this only be used for machine readable output?  Or in the future should new error messages be written to `ErrorOutput`?

GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -260,9 +264,10 @@ private void DehydrateFolders(JsonTracer tracer, GVFSEnlistment enlistment, stri
                                     {
                                         if (!this.TryIO(tracer, () => this.fileSystem.DeleteDirectory(fullPath), $"Deleting '{fullPath}'", out ioError))
                                         {
-                                            this.WriteMessage(tracer, $"Cannot dehydrate folder '{folder}': removing '{folder}' failed.");
+                                            this.WriteMessage(tracer, $"Cannot {this.DisplayName} folder '{folder}': removing '{folder}' failed.");
                                             this.WriteMessage(tracer, "Ensure no applications are accessing the folder and retry.");
                                             this.WriteMessage(tracer, $"More details: {ioError}");
+                                            folderErrors.Add($"{folder}\0{ioError}");
```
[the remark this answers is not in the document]
kewillford: Looks like a space
```
Scripts The process cannot access the file 'C:\gvfs_repos\ft\src\Scripts' because it is being used by another process.
Failed to dehydrate 1 folder(s).
Failed to prune. Exit Code: DehydrateFolderFailures
```

GVFS/GVFS/CommandLine/DehydrateVerb.cs
kewillford: These folders come back from the mount process and I wasn't sure how a `\0` would be handled by our named pipe messages but let me take a look and try it out.

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: I really don't think we want to put that restriction on this.  We haven't been writing to stderr and it isn't displayed by default in the console.  If we want to really create a machine readable CLI then we should do something else but not sure how much work we want to put into it.

wilbaker: > it isn't displayed by default in the console

I did not realize this thanks!

Sounds like this change is good as it is now.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
wilbaker: Basic question, is the `\0<error>` just to provide more details to the caller (in case they care about the specific error reason)?

I wasn't sure if callers would expect every line to have a `\0<error>` even if it was just ` `\0{DisplayName} Failed` (or some other general failure message for this case)

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: The case I am trying to handle is the `sparse` command running the `DehyrateVerb` which the output is being redirected to `Console.Out` but it creates it's own log file for logging information so if you only look in the `SparseVerb` log you may not see everything that happened during the dehydrate.  That is why it is adding this to the sparse log so you know to look at the log file of the `new TVerb().VerbName` for details.

sparse log will have something like
```
sparse {"Message":"No folders to update in sparse set."}
sparse {"Message":"Found 1 folders to prune."}
DehydrateVerb {"Output":"Check dehydrate logs for output","ReturnCode":7}
Error {"ErrorMessage":"Failed to prune. Exit Code: DehydrateFolderFailures"}
```

Dehydrate log will have the full logs from the verb including
```
Error {"Description":"Deleting '<folder path>'","Error":"The process cannot access the file '<folder path>' because it is being used by another process.","ErrorMessage":"TryIO: Caught exception performing action"}
```

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: I will still add a comment about how we are using this right now in case something else wants to use it.

kewillford: Sorry it's not Console doesn't show stderr.  It does but I wasn't seeing the stderr stream from the DehydrateVerb in the sparse console output without using `2>&1`, so maybe there is something more I need to do here.

kewillford: Weird it is working the way you would expect.  Not sure what was going on when I didn't see the folder list in the console, maybe I didn't have everything connected yet.

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: Making the entire `TVerb` output logs go to the calling verb's log file is more work since the `ITracer` would need to be passed around and that seemed out of scope for these changes.

GVFS/GVFS/CommandLine/DehydrateVerb.cs
kewillford: Yeah we should always have at least the `\0` even if there isn't anything after so that it is an expected format where the `\0` is the field separator.

GVFS/GVFS/CommandLine/GVFSVerb.cs
wilbaker: > but it creates it's own log file for logging information

To confirm, in this example will `sparse`'s log file include the contents of whatever was written to `outputWriter`?  Or is this never recorded by a tracer?

This might be a dumb question but I can't figure out where this:

`{ "Output", commandOutput.ToString() },`

Was ever getting logged (e.g. I don't see it in the `clone` verb when it calls `mount`).

> you may not see everything

Maybe add a comment here, something like:

```
else
{
    // If a parent verb is redirecting the output of its child, include a reminder
    // that the child verb's activity was recorded in its own log file
    metadata.Add("Output", $"Check {new TVerb().VerbName} logs for output");
```

kewillford: >To confirm, in this example will sparse's log file include the contents of whatever was written to outputWriter? Or is this never recorded by a tracer?

No, writing to the `outputWriter` doesn't have anything to do with the `ITracer` that is being used for tracing.  That is what makes this kind of messy because we are redirecting where output is going (to the `Console.Out` of the sparse verb) but the `DehyrateVerb` is still creating its own tracer that is used and writes to the dehydrate log.

>Was ever getting logged (e.g. I don't see it in the clone verb when it calls mount).

Looks like in the case of the `CloneVerb` its tracer is being disposed before the call to `this.Execute<MountVerb>(`.
See https://github.com/microsoft/VFSForGit/blob/c5b4262e38870eb4c648a9bac48ca3c062aad5bc/GVFS/GVFS/CommandLine/CloneVerb.cs#L123-L204

wilbaker: > Looks like in the case of the CloneVerb its tracer is being disposed before the call to this.Execute<MountVerb>

Ahh that does sound familiar there was one point in time where we could not have two ETW tracer instances at the same time.  I don't think we still have that restriction but this code was never updated.

> That is what makes this kind of messy because we are redirecting where output is going (to the Console.Out of the sparse verb) but the DehyrateVerb is still creating its own tracer that is used and writes to the dehydrate log.

Thanks for all of the details here!  The `{ "Output", commandOutput.ToString() }` was what was really throwing me off, but it looks like it's probably never gone where we expected (after updating `clone` to dispose its tracer).

The latest changes look good to me, thanks!

PR #1557, opened 2019-10-16
Fix tests to check windows version for POSIX delete support

kewillford:
There are tests that specifically knew that windows did not support POSIX style deletes. 
Since support for POSIX style deletes has now been added to Windows these tests need to check the Windows version to see if it is supported or not for the specific checks.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
@@ -9,6 +9,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
 
 namespace GVFS.FunctionalTests.Tests.LongRunningEnlistment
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
```diff
@@ -941,7 +935,38 @@ public void RunPythonExecutable()
...
+        }
+
+        private bool SupportsPosixDelete()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // https://msdn.microsoft.com/en-us/library/windows/desktop/ms724429(v=vs.85).aspx
+                FileVersionInfo kernel32Info = FileVersionInfo.GetVersionInfo(Path.Combine(Environment.SystemDirectory, "kernel32.dll"));
+
+                // 18362 is first build with posix delete support in windows
```
wilbaker: Tiny nit: This is the first release where the DeleteFile API uses POSIX style deletes by default (earlier versions supported POSIX delete but it was not the default).

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/BasicFileSystemTests.cs
```diff
@@ -941,7 +935,38 @@ public void RunPythonExecutable()
             result.Output.ShouldContain("3.14");
 
             GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout " + this.Enlistment.Commitish);
-       }
+        }
+
+        private void CheckForDeletedFile(string filePath, FileSystemRunner fileSystem)
```
wilbaker: Naming suggestion: When reading the above tests it wasn't clear what `CheckForDeletedFile` would be checking.

Maybe rename this to something like `ConfirmDeletedWhileOpen`?

kewillford: How about `VerifyExistenceAfterDeleteWhileOpen`?

wilbaker: > How about `VerifyExistenceAfterDeleteWhileOpen`?

Even better! 😃

PR #1558, opened 2019-10-18
Do not try to uninstall prjflt unless flag is passed

kewillford:
Need to make sure that the build scripts don't change prjflt unless the flag is passed to do so.

### DIFF
--- a/Scripts/UninstallGVFS.bat
+++ b/Scripts/UninstallGVFS.bat
@@ -9,11 +9,13 @@ if not exist "c:\Program Files\GVFS" goto :end
 
 call %VFS_SCRIPTSDIR%\StopAllServices.bat
 
+if not "%2"=="--remove-prjflt" goto :uninstallGVFS
 REM The GVFS uninstaller will not remove prjflt, and so we must remove it ourselves first.  If we don't, the non-inbox ProjFS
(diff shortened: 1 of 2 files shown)


Scripts/UninstallGVFS.bat
```diff
@@ -9,11 +9,13 @@ if not exist "c:\Program Files\GVFS" goto :end
 
 call %VFS_SCRIPTSDIR%\StopAllServices.bat
 
+if not "%2"=="--remove-prjflt" goto :uninstallGVFS
```
jrbriggs: So we'll update our pipeline to call the uninstall script with this parameter on server2016, but leave the inbox version alone on 19h1?

kewillford: Yep and it will actually be the `ReinstallGVFS.bat` that will need to be passed the argument and it calls `UninstallGVFS.bat` passing the argument it got because it also needs to check for the `--remove-prjflt`

kewillford: https://github.com/microsoft/VFSForGit/blob/825202e19568d192ed406083e04be61664a221b8/Scripts/ReinstallGVFS.bat#L24-L29

jrbriggs: Cool, thanks!

PR #1569, opened 2019-11-05
Ensure git-credential cmds are run with correct wrkdir

### LINKED ISSUE #1568: Git credential approve/reject can fail to locate stored credentials
In the case that a repository has a local configuration entry for `credential.usehttppath = true`, calling `git credential reject` to remove a bad stored credential fails as it's unable to locate the same credential that `git credential fill` returns.

GVFS sets `credential.usehttppath = true` on the command line at invocation but this is only for "dev.azure.com" URLs, where as in the local config it's possible users that have cloned their repos _before_ #939 was released they will continue to have that unscoped `usehttppath` entry in the local config.

mjcheetham:
Ensure that all git-credential commands are run with the correct working
directory. Currently only 'fill' is invoked against the .git directory
and the 'reject' and 'approve' commands are invoked outside the
repository.

The result of this mismatch means that if the user has a local
configuration entry set, for say credential.usehttppath, then the
behaviour for locating credentials will be different between read and
modification operations.

Fixes #1568

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -173,7 +173,7 @@ namespace GVFS.Common.Git
 
             string stdinConfig = sb.ToString();
 
-            Result result = this.InvokeGitOutsideEnlistment(
+            Result result = this.InvokeGitAgainstDotGitFolder(
                 GenerateCredentialVerbCommand("reject"),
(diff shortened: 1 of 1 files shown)


mjcheetham:
@jrbriggs did we decide on skipping the Mac build steps here because they're failing due to Catalina?

jrbriggs:
@mjcheetham yes, merging since this is correct on Windows.

PR #1570, opened 2019-11-06
LooseObjectsStep: simply code for getting directory paths

wilbaker:
The code in CountLooseObjects can be simplified by avoiding
an unnecessary Path.Combine.

Additionally, fix a bug in the MockFileSystem where EnumerateDirectories
was not returning the full paths of directories and it should.

### DIFF
--- a/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
@@ -52,8 +52,7 @@ namespace GVFS.Common.Maintenance
 
                 if (GitObjects.IsLooseObjectsDirectory(directoryName))
                 {
-                    string dirPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath);
-                    List<DirectoryItemInfo> dirItems = this.Context.FileSystem.ItemsInDirectory(dirPath).ToList();
+                    List<DirectoryItemInfo> dirItems = this.Context.FileSystem.ItemsInDirectory(directoryPath).ToList();
                     count += dirItems.Count;
                     size += dirItems.Sum(item => item.Length);
                 }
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
```diff
@@ -52,7 +52,7 @@ public void CountLooseObjects(out int count, out long size)
 
                 if (GitObjects.IsLooseObjectsDirectory(directoryName))
                 {
-                    string dirPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryPath);
+                    string dirPath = Path.Combine(this.Context.Enlistment.GitObjectsRoot, directoryName);
```
kewillford: We could use `directoryPath` and delete this line right?

wilbaker: @kewillford, @derrickstolee another change I've been considering is replacing this line:

`string directoryName = directoryPath.TrimEnd(Path.DirectorySeparatorChar).Split(Path.DirectorySeparatorChar).Last();`

With

`Path.GetDirectoryName(directoryPath)`

Thoughts?  I was a bit on the fence about how much cleanup to do here.

derrickstolee: That looks much simpler. This is likely the only time we have to do that cleanup, so please do!

kewillford: Yeah I thought about commenting on that as well and since we were both thinking it then it is probably a good thing to change.

wilbaker:
It ends up that the code in `CountLooseObjects` actually has been working correctly because:

> If path2 includes a root, path2 is returned.

https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=netframework-4.8#System_IO_Path_Combine_System_String_System_String_

I will still make this change though as the behavior in the Mock is incorrect, and there is no need to perform the Path.Combine.

GVFS/GVFS.Common/Maintenance/LooseObjectsStep.cs
wilbaker: > Yeah I thought about commenting on that as well and since we were both thinking it then it is probably a good thing to change.

I am actually going to back the change to use `Path.GetDirectoryName(directoryPath)` out for now.  It ends up that `Path.GetDirectoryName(...)` doesn't play well with the `mock:` paths we use in the unit tests.

I've filed #1572 to clean this up.

PR #1571, opened 2019-11-06
Update Git to have proper version number

derrickstolee:
The previous installer was generated without the `v2.24.0.vfs.1.1` tag, which caused the Git version generation to fail. That is the only change between the `.1` and `.2` builds.

[Here is the build line with the version number.](https://dev.azure.com/gvfs/ci/_build/results?buildId=17360&view=logs&j=00af5505-930d-50dd-3844-b167e8b4bb13&t=6ec846bc-101e-588c-b676-9b550d603286&l=27)


derrickstolee:
Kevin found that even this build had a problem with the Mac version. The Windows version was fine.

PR #1586, opened 2019-11-12
Update sparse parent folder mtimes to trigger Explorer refresh

### LINKED ISSUE #1532: Gvfs sparse projection changes do not update open file explorer windows
Not sure if there is a possible fix for this but I can see this being a major pain point. Here is what I did:

1. Open file explorer to the root of our repo, see all folders
2. Call gvfs sparse --set word to scope it to only word
3. File explorer still shows all folders. Also, if I try to click on one of them I get this message:

image

For office, users will not be interacting with gvfs directly, scoper will change their projection for them. I can see it being confusing if a user opens a stale file explorer window and they click on a folder and get this error message and not know why.

wilbaker:
Resolves #1532 

If a File Explorer window is open to a directory in VFS4G, changes
to the files being projected are not enough to trigger a refresh.
This is especially noticeable when using the 'gvfs sparse' verb.

To force Explorer to refresh when 'gvfs sparse' triggers a projecion
changes, VFS4G will now update the last write time for every folder
placeholder that had children added or removed from the sparse set.

### DIFF
--- a/GVFS/GVFS.Common/NativeMethods.cs
+++ b/GVFS/GVFS.Common/NativeMethods.cs
@@ -102,7 +102,7 @@ namespace GVFS.Common
             }
         }
 
-        public static bool SetDirectoryLastWriteTimeIfOnDisk(string path, DateTime lastWriteTime)
+        public static void SetDirectoryLastWriteTime(string path, DateTime lastWriteTime, out bool directoryExists)
         {
             // We can't use Directory.SetLastWriteTime as it requests GENERIC_WRITE access
             // which will fail for directory placeholders.  The only access requried by SetFileTime
@@ -123,10 +123,11 @@ namespace GVFS.Common
                     int error = Marshal.GetLastWin32Error();
                     if (error == ERROR_FILE_NOT_FOUND || error == ERROR_PATH_NOT_FOUND)
                     {
-                        return false;
+                        directoryExists = false;
+                        return;
                     }
 
-                    throw new Win32Exception(error, $"{nameof(SetDirectoryLastWriteTimeIfOnDisk)}: Failed to open handle for '{path}'");
+                    throw new Win32Exception(error, $"{nameof(SetDirectoryLastWriteTime)}: Failed to open handle for '{path}'");
                 }
 
                 // SetFileTime will not update times with value 0
@@ -135,11 +136,12 @@ namespace GVFS.Common
                 long lastWriteFileTime = lastWriteTime.ToFileTime();
                 if (!SetFileTime(handle, ref creationFileTime, ref lastAccessFileTime, ref lastWriteFileTime))
                 {
-                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTimeIfOnDisk)}: Failed to update last write time for '{path}'");
+                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTime)}: Failed to update last write time for '{path}'");
                 }
             }
 
-            return true;
+            directoryExists = true;
+            return;
         }
 
         /// <summary>
--- a/GVFS/GVFS.Common/NativeMethods.cs
+++ b/GVFS/GVFS.Common/NativeMethods.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.Win32.SafeHandles;
 using System;
+using System.ComponentModel;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -15,6 +16,9 @@ namespace GVFS.Common
         private const uint FSCTL_GET_REPARSE_POINT = 0x000900a8;
 
         private const int ReparseDataPathBufferLength = 1000;
+
+        private const int ERROR_FILE_NOT_FOUND = 0x2;
+        private const int ERROR_PATH_NOT_FOUND = 0x3;
 
         [Flags]
         public enum MoveFileFlags : uint
@@ -98,7 +102,7 @@ namespace GVFS.Common
             }
         }
 
-        public static void SetDirectoryLastWriteTime(string path, DateTime lastWriteTime)
+        public static bool SetDirectoryLastWriteTimeIfOnDisk(string path, DateTime lastWriteTime)
         {
             // We can't use Directory.SetLastWriteTime as it requests GENERIC_WRITE access
             // which will fail for directory placeholders.  The only access requried by SetFileTime
@@ -116,7 +120,13 @@ namespace GVFS.Common
             {
                 if (handle.IsInvalid)
                 {
-                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTime)}: Failed to open handle for '{path}'");
+                    int error = Marshal.GetLastWin32Error();
+                    if (error == ERROR_FILE_NOT_FOUND || error == ERROR_PATH_NOT_FOUND)
+                    {
+                        return false;
+                    }
+
+                    throw new Win32Exception(error, $"{nameof(SetDirectoryLastWriteTimeIfOnDisk)}: Failed to open handle for '{path}'");
                 }
 
                 // SetFileTime will not update times with value 0
@@ -125,9 +135,11 @@ namespace GVFS.Common
                 long lastWriteFileTime = lastWriteTime.ToFileTime();
                 if (!SetFileTime(handle, ref creationFileTime, ref lastAccessFileTime, ref lastWriteFileTime))
                 {
-                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTime)}: Failed to update last write time for '{path}'");
+                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTimeIfOnDisk)}: Failed to update last write time for '{path}'");
                 }
             }
+
+            return true;
         }
 
         /// <summary>
--- a/GVFS/GVFS.Common/NativeMethods.cs
+++ b/GVFS/GVFS.Common/NativeMethods.cs
@@ -104,14 +104,15 @@ namespace GVFS.Common
             // which will fail for directory placeholders.  The only access requried by SetFileTime
             // is FILE_WRITE_ATTRIBUTES (which ProjFS does allow for placeholders)
 
-            using (SafeFileHandle handle = CreateFile(
-                path,
-                FileAccess.FILE_WRITE_ATTRIBUTES,
-                FileShare.ReadWrite | FileShare.Delete,
-                IntPtr.Zero,
-                FileMode.Open,
-                FileAttributes.FILE_FLAG_BACKUP_SEMANTICS,
-                IntPtr.Zero))
+            using (SafeFileHandle handle =
+                CreateFile(
+                    path,
+                    FileAccess.FILE_WRITE_ATTRIBUTES,
+                    FileShare.ReadWrite | FileShare.Delete,
+                    IntPtr.Zero,
+                    FileMode.Open,
+                    FileAttributes.FILE_FLAG_BACKUP_SEMANTICS,
+                    IntPtr.Zero))
             {
                 if (handle.IsInvalid)
                 {
--- a/GVFS/GVFS.Common/NativeMethods.cs
+++ b/GVFS/GVFS.Common/NativeMethods.cs
@@ -98,6 +98,37 @@ namespace GVFS.Common
             }
         }
 
+        public static void SetDirectoryLastWriteTime(string path, DateTime lastWriteTime)
+        {
+            // We can't use Directory.SetLastWriteTime as it requests GENERIC_WRITE access
+            // which will fail for directory placeholders.  The only access requried by SetFileTime
+            // is FILE_WRITE_ATTRIBUTES (which ProjFS does allow for placeholders)
+
+            using (SafeFileHandle handle = CreateFile(
+                path,
+                FileAccess.FILE_WRITE_ATTRIBUTES,
+                FileShare.ReadWrite | FileShare.Delete,
+                IntPtr.Zero,
+                FileMode.Open,
+                FileAttributes.FILE_FLAG_BACKUP_SEMANTICS,
+                IntPtr.Zero))
+            {
+                if (handle.IsInvalid)
+                {
+                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTime)}: Failed to open handle for '{path}'");
+                }
+
+                // SetFileTime will not update times with value 0
+                long creationFileTime = 0;
+                long lastAccessFileTime = 0;
+                long lastWriteFileTime = lastWriteTime.ToFileTime();
+                if (!SetFileTime(handle, ref creationFileTime, ref lastAccessFileTime, ref lastWriteFileTime))
+                {
+                    ThrowLastWin32Exception($"{nameof(SetDirectoryLastWriteTime)}: Failed to update last write time for '{path}'");
+                }
+            }
+        }
+
         /// <summary>
         /// Get the build number of the OS
         /// </summary>
@@ -200,6 +231,13 @@ namespace GVFS.Common
 
         [DllImport("kernel32.dll")]
         private static extern ulong GetTickCount64();
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern bool SetFileTime(
+            SafeFileHandle hFile,
+            [In] ref long creationTime,
+            [In] ref long lastAccessTime,
+            [In] ref long lastWriteTime);
 
         [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
         private struct REPARSE_DATA_BUFFER
--- a/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
+++ b/GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
@@ -14,6 +14,7 @@ using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -57,17 +58,8 @@ namespace GVFS.Virtualization.Projection
 
         private BlobSizes blobSizes;
         private IPlaceholderCollection placeholderDatabase;
-
         private ISparseCollection sparseCollection;
         private SparseFolderData rootSparseFolder;
-        private HashSet<string> sparsePaths;
-
-        // The set of sparse folders that had child folders added or removed
-        // during the last projection refresh.  If there are placeholders
-        // on disk for these folders we will update their last write time to
-        // force a refresh of the folders in File Explorer.
-        private HashSet<string> sparseFoldersNeedingRefresh;
-
         private GVFSGitObjects gitObjects;
         private BackgroundFileSystemTaskRunner backgroundFileSystemTaskRunner;
         private ReaderWriterLockSlim projectionReadWriteLock;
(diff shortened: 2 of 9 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,13 +798,33 @@ private void ClearProjectionCaches()
 
         private void RefreshSparseFolders()
         {
+            this.sparseFoldersNeedingRefresh = null;
             this.rootSparseFolder.Children.Clear();
             if (this.sparseCollection != null)
             {
+                this.sparseFoldersNeedingRefresh = new HashSet<string>(GVFSPlatform.Instance.Constants.PathComparer);
+
+                // Always refresh the root directory
```
wilbaker: There is probably no need to always refresh the root

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,13 +798,33 @@ private void ClearProjectionCaches()
...
-                foreach (string directoryPath in this.sparseCollection.GetAll())
+
+                HashSet<string> oldSparsePaths = this.sparsePaths;
+                this.sparsePaths = this.sparseCollection.GetAll();
+                foreach (string directoryPath in this.sparsePaths)
                 {
                     string[] folders = directoryPath.Split(new[] { Path.DirectorySeparatorChar }, StringSplitOptions.RemoveEmptyEntries);
+                    if (oldSparsePaths != null && !oldSparsePaths.Remove(directoryPath))
+                    {
+                        // directoryPath is a new entry, add its parent to updatedFolders
```
wilbaker: This might miss some required refresh scenarios.

*Example*

1. `gvfs sparse --set Scripts --prune`
2. `gvfs sparse --set Scripts;GVFS\GVFS --prune`

In this scenario we should refresh the root directory as well.
In this scenario

wilbaker:
@kewillford you can hold off on reviewing my changes to `RefreshSparseFolder` I realized the approach I'm using now won't work in all scenarios, and I'm working on an update.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: Actually leaving this in with a better comment in the next push

wilbaker:
> @kewillford you can hold off on reviewing my changes to RefreshSparseFolder I realized the approach I'm using now won't work in all scenarios, and I'm working on an update.

@kewillford okay this is ready for another look.  Please let me know if you can think of a better way of determining the list of directories to refresh.  With the most recent commit I took a simple approach:

- For any path that was added or removed from the set, refresh its ancestors (if they are on disk)

This could result in over-refreshing (i.e. there are scenarios where not *all* ancestors would need to be refreshed) but assuming there are not thousands of deeply nested folders being changed in each call to `gvfs sparse`  I'm not sure we need the extra complexity of a fancier algorithm.

cc: @derrickstolee for his thoughts as well

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,13 +798,44 @@ private void ClearProjectionCaches()
...
+                        // And now it contains:
+                        //    - A\B\C
+                        //    - A\B\D
+                        //
+                        // There is no need to refresh A, but with the code below we will
+                        // (assuming that A is on disk)
+                        int parentLevels = folders.Length - 1;
+                        for (int i = parentLevels; i > 0; --i)
+                        {
+                            this.sparseFoldersNeedingRefresh.Add(string.Join($"{Path.DirectorySeparatorChar}", folders, startIndex: 0, count: i));
```
wilbaker: Nit: Move `$"{Path.DirectorySeparatorChar}"` out of the loop

GVFS/GVFS.Common/NativeMethods.cs
```diff
@@ -98,6 +98,37 @@ public static void MoveFile(string existingFileName, string newFileName, MoveFil
...
+            // is FILE_WRITE_ATTRIBUTES (which ProjFS does allow for placeholders)
+
+            using (SafeFileHandle handle = CreateFile(
+                path,
+                FileAccess.FILE_WRITE_ATTRIBUTES,
+                FileShare.ReadWrite | FileShare.Delete,
+                IntPtr.Zero,
+                FileMode.Open,
+                FileAttributes.FILE_FLAG_BACKUP_SEMANTICS,
+                IntPtr.Zero))
```
derrickstolee: nit: left-spacing is odd.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -58,8 +57,17 @@ public partial class GitIndexProjection : IDisposable, IProfilerOnlyIndexProject
...
+
         private ISparseCollection sparseCollection;
         private SparseFolderData rootSparseFolder;
+        private HashSet<string> sparsePaths;
+
+        // The set of sparse folders that had child folders added or removed
+        // during the last projection refresh.  If there are placeholders
+        // on disk for these folders we will update their last write time to
+        // force a refresh of the folders in File Explorer.
+        private HashSet<string> sparseFoldersNeedingRefresh;
```
kewillford: What are your performance and memory numbers with this change?  Especially when there are a lot of folders?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -790,13 +798,46 @@ private void ClearProjectionCaches()
...
+                // To avoid a lot of no-op adds, always refresh the root
+                this.sparseFoldersNeedingRefresh.Add(string.Empty);
+
                 Dictionary<string, SparseFolderData> parentFolder = this.rootSparseFolder.Children;
-                foreach (string directoryPath in this.sparseCollection.GetAll())
+                HashSet<string> oldSparsePaths = this.sparsePaths;
+                this.sparsePaths = this.sparseCollection.GetAll();
+                foreach (string directoryPath in this.sparsePaths)
                 {
                     string[] folders = directoryPath.Split(new[] { Path.DirectorySeparatorChar }, StringSplitOptions.RemoveEmptyEntries);
```
kewillford: Why are we using `Split` here and then `LastIndexOf` below?  Are there differences in memory/performance?  Can we use a common method for both?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1322,7 +1391,42 @@ private void UpdatePlaceholders()
                 long millisecondsUpdatingFolderPlaceholders = stopwatch.ElapsedMilliseconds;
 
                 stopwatch.Restart();
+                DateTime updateTime = DateTime.Now;
+                int folderPlaceholdersRefreshAttempts = 0;
+                int folderPlaceholdersRefreshed = 0;
+                foreach (string folderPlaceholderPath in folderPlaceholdersToKeep)
```
kewillford: This will only refresh folders that are placeholders right?  Do we need to do anything for full folders that change or have changes in them?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: I will run some tests and post back with results, thanks!

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: Good point, these started out as different because in my first iteration I was only grabbing the immediate parent.  Now that we're processing all ancestors we should be able to make this code consistent.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
wilbaker: > Do we need to do anything for full folders that change or have changes in them?

I don't believe that we'll need to do anything.  

The behavior of Explorer with full folders will be the same as the behavior of a regular vanilla git repo in Explorer.  I think that *should* be OK because the problem noted in #1532 was with changes that don't result in anything actually changing on disk:

>The issue where Explorer doesn’t refresh when folders are removed from the projection sounds expected. Explorer keeps an open handle to the folder it is showing. I believe it uses directory change notifications to tell it to refresh the view. Altering the projection doesn’t fire a change notification because it doesn’t cause anything in the open folder to be changed, not for real anyway. 

I can test out this scenario a bit, but as long full folder behavior is consistent with native git I think that's sufficient.

wilbaker: @kewillford, I just tried testing this out, and found that I can't prune children of full folders.  Is that expected?

(In this test `GVFS\GVFS.FunctionalTests` is full)

```
M:\testSparse\ForTests>gvfs sparse --set GVFS\GVFS.FunctionalTests\Should --prune
Running git status...Succeeded
Updating sparse folder set...Succeeded
Forcing a projection change...Succeeded
Finding folders to prune...Succeeded
Found 5 folders to prune.
Running git status before prune to make sure you don't have any pending changes.

Running git status...Succeeded

WARNING: If you abort the sparse after this point, the repo may become corrupt

Unmounting...Succeeded
Cleaning up folders...-Cannot prune folder 'GVFS\GVFS.FunctionalTests\FileSystemRunners': Must prune parent folder 'GVFS/GVFS.FunctionalTests/'.
Cannot prune folder 'GVFS\GVFS.FunctionalTests\Properties': Must prune parent folder 'GVFS/GVFS.FunctionalTests/'.
Cannot prune folder 'GVFS\GVFS.FunctionalTests\TestData': Must prune parent folder 'GVFS/GVFS.FunctionalTests/'.
Cannot prune folder 'GVFS\GVFS.FunctionalTests\Tests': Must prune parent folder 'GVFS/GVFS.FunctionalTests/'.
Cannot prune folder 'GVFS\GVFS.FunctionalTests\Tools': Must prune parent folder 'GVFS/GVFS.FunctionalTests/'.
Cleaning up folders...Succeeded
Mounting...Succeeded
```

I *did* find a problematic scenario while doing this testing though:

1. `gvfs sparse --set GVFS\GVFS.FunctionalTests --prune`
2. `mkdir GVFS\FullFolder`
3. `gvfs sparse --set GVFS\GVFS;GVFS\FullFolder --prune`
4. `gvfs sparse --set GVFS\GVFS.FunctionalTests --prune`

After step 4 Explorer will refresh.

It looks like the issue is that the `GVFS` directory is not refreshed after the unmount\dehydrate\mount.

This is not an issue in the root directory because with the current changes it is always refreshed.

wilbaker: @kewillford and I just discussed this offline.  I am going to try moving the refresh logic out of `GitIndexProjection` and into the `sparse` and `dehydrate` verbs.

Advantages:

- `sparse` can perform the refresh after the final `mount`
- No risk of impacting the "update projection" hot path

wilbaker: >  I am going to try moving the refresh logic out of GitIndexProjection and into the sparse and dehydrate verbs.

Now that I've taken another look that won't be quite so simple, as we don't always need to run dehydrate.  I will tag this PR as WIP as I dig into some options here.

wilbaker: > Now that I've taken another look that won't be quite so simple, as we don't always need to run dehydrate. I will tag this PR as WIP as I dig into some options here.

I actually was able to do this and it's in the latest commit.  @kewillford I have not done any testing of these changes at all yet, and so please feel free to hold off on reviewing.

wilbaker: >  @kewillford I have not done any testing of these changes at all yet, and so please feel free to hold off on reviewing.

@kewillford I did some tests with the latest changes on my machine and found that adding/removing ~50 L3 folders (each with unique parents and grandparents) took between 20 and 40ms:

*Adding the directories to the projection (3 runs)*
```
[2019-11-15 15:31:54.6444 -08:00] UpdateParentFolderLastWriteTimes_Summary {"foldersToRefresh":103,"foldersRefreshed":103,"refreshErrors":0,"ElapsedMilliseconds":31,"Message":"Updated folder last write times"}
[2019-11-15 15:32:58.6538 -08:00] UpdateParentFolderLastWriteTimes_Summary {"foldersToRefresh":103,"foldersRefreshed":103,"refreshErrors":0,"ElapsedMilliseconds":19,"Message":"Updated folder last write times"}
[2019-11-15 15:33:44.5356 -08:00] UpdateParentFolderLastWriteTimes_Summary {"foldersToRefresh":103,"foldersRefreshed":103,"refreshErrors":0,"ElapsedMilliseconds":19,"Message":"Updated folder last write times"}
```

*Removing the directories to the projection (3 runs)*
```
[2019-11-15 15:32:37.7900 -08:00] UpdateParentFolderLastWriteTimes_Summary {"foldersToRefresh":103,"foldersRefreshed":1,"refreshErrors":0,"ElapsedMilliseconds":27,"Message":"Updated folder last write times"}
[2019-11-15 15:33:11.2884 -08:00] UpdateParentFolderLastWriteTimes_Summary {"foldersToRefresh":103,"foldersRefreshed":1,"refreshErrors":0,"ElapsedMilliseconds":37,"Message":"Updated folder last write times"}
[2019-11-15 15:37:36.6737 -08:00] UpdateParentFolderLastWriteTimes_Summary {"foldersToRefresh":103,"foldersRefreshed":1,"refreshErrors":0,"ElapsedMilliseconds":14,"Message":"Updated folder last write times"}
```

And so I am thinking that performance is sufficient for the use cases we're aware of today.

Please let me know what you think of the perf and the latest changes, thanks!

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -235,10 +237,105 @@ protected override void Execute(GVFSEnlistment enlistment)
                     {
                         this.PruneFoldersOutsideSparse(tracer, enlistment, sparseTable);
```
kewillford: I think we need to have this return the the folders that are outside the sparse set to add to the list to refresh parents because if I have sparse set `A\B` and create a new folder `A\C` with a file, add and commit, then run `gvfs sparse --prune` or `gvfs sparse --set A\B --prune`, `A\C` should get pruned but it will not be in `foldersToAdd` or `foldersToRemove`.

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -235,10 +237,105 @@ protected override void Execute(GVFSEnlistment enlistment)
...
                 }
             }
         }
 
+        private static void UpdateParentFolderLastWriteTimes(ITracer tracer, string rootPath, List<string> foldersToRemove, List<string> foldersToAdd)
+        {
+            Stopwatch updateTime = Stopwatch.StartNew();
+
+            HashSet<string> foldersToRefresh = new HashSet<string>(GVFSPlatform.Instance.Constants.PathComparer);
+            foreach (string path in foldersToRemove)
```
kewillford: I would either
```
foreach (string path in foldersToRemove.Concat(foldersToAdd).Concat(foldersPruned))
```
or make `AddNonRootParentPathsToSet` take the `List<string>` and have it do the `foreach` so that you would have
```
AddNonRootParentPathsToSet(foldersToRemove, foldersToRefresh);
AddNonRootParentPathsToSet(foldersToAdd, foldersToRefresh);
AddNonRootParentPathsToSet(foldersPruned, foldersToRefresh);
```

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -235,10 +237,105 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+            }
+            else if (refreshFailed)
+            {
+                ++refreshErrors;
+            }
+
+            string folderPathPrefix = $"{rootPath}{Path.DirectorySeparatorChar}";
+            foreach (string path in foldersToRefresh)
+            {
+                if (SetFolderLastWriteTimeIfOnDisk(tracer, folderPathPrefix + path, refreshTime, out refreshFailed))
```
kewillford: I had to go back up to the code to figure out why this is returning a `bool` and `refreshFailed` and what the difference between returning `false` and `refreshFailed` being `false` so maybe a comment would help.

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -235,10 +237,105 @@ protected override void Execute(GVFSEnlistment enlistment)
...
+            if (SetFolderLastWriteTimeIfOnDisk(tracer, rootPath, refreshTime, out refreshFailed))
+            {
+                ++foldersRefreshed;
+            }
+            else if (refreshFailed)
+            {
+                ++refreshErrors;
+            }
+
+            string folderPathPrefix = $"{rootPath}{Path.DirectorySeparatorChar}";
```
kewillford: Is this for perf so we don't have to use `Path.Combine`?

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: @kewillford great catch! I missed that we might prune folders that were never in the initial sparse set.

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: Yep it was for performance (although truth be told I did not measure the perf impact of using `Path.Combine` instead).

I figured that since `GVFSDatabase.NormalizePath` had already been run for all the paths we didn't need to add the extra overhead of `Path.Combine`.

If you think `Path.Combine` is preferable let me know and I can try running some performance tests using it instead.

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: @kewillford two other alternatives I considered for this method:

-  Make the method `void` and have an `out` parameter `directoryExists` (in addition to `refreshFailed`)
- Have the method return an result enum with possible values `Success`, `Failure`, `DirectoryDoesNotExist`

Do you think either of those would be preferable to the current approach?  Initially there was no `refreshFailed`, but I decided that it would be nice to log failures specifically (and not lump their count together with the directory not existing).

kewillford: Yeah we really have 3 different results here.
1. The directory was refreshed
2. The directory didn't exist
3. Tried to refresh the directory but it failed

I'm kind of leaning toward the result enum because it gives us the 3 states vs the 2 bools which give us 4 one of which is never possible.

Do we want to log the directories that don't exist?

GVFS/GVFS/CommandLine/SparseVerb.cs
kewillford: Nope this is good.  Thanks!

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: > Do we want to log the directories that don't exist?

It will be relatively common for some of the directories to not be on disk, and so I think we're ok just logging the counts.  It would be tough to differentiate between when it's expected and unexpected that a directory be on disk.

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: @kewillford I confirmed that this scenario works correctly with my latest commit (and that it was broken before that change), thanks!

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -370,6 +408,8 @@ private void PruneFoldersOutsideSparse(ITracer tracer, Enlistment enlistment, Sp
                     this.ReportErrorAndExit(tracer, verbReturnCode, $"Failed to {PruneOptionName}. Exit Code: {verbReturnCode}");
                 }
             }
+
+            return directoriesToDehydrate;
         }
 
         private string[] GetDirectoriesOutsideSparse(string rootPath, SparseTable sparseTable)
```
derrickstolee: I know this isn't new to this change, but why are we using `string[]` instead of `List<string>`? In `GetDirectoriesOutsideSparse()`, we even use a `List<string>` internally but then `.ToArray()` it (which is slower, as it allocs a new array and copies all values to the new array).

wilbaker: @derrickstolee that's a good question and I can't see why we'd need to use `string[]` either.  I will update these to `List<string>`.

PR #1593, opened 2019-11-15
Mac: EndpointSecurity MirrorProvider-like prototype

pmj:
This is an experimental implementation of a provider similar to MirrorProvider which uses the new EndpointSecurity API from macOS 10.15 Catalina. It is intended as an example for using the API as a possible future option for implementing a VFSForGit provider or file system watcher.

The prototype allows specifying a source directory and a target directory (which initially should not yet exist); the source directory is lazily replicated into the target by means of empty placeholder files whose contents are copied from their respective source equivalent on read or write access.

The EndpointSecurity API unfortunately does not offer the necessary events to fully replicate MirrorProvider; specifically, there is no blocking event for accesses to directories, so on-demand directory enumeration is not possible. As a result, all empty placeholder files must be created recursively during cloning.

The key EndpointSecurity event type for hydrating placeholders is `ES_EVENT_TYPE_AUTH_OPEN`, and this is the main functionality the prototype is built around.

Another critical limitation of the EndpointSecurity API is the hard time limit of 60 seconds. Particularly if file hydration hits the network, it can be impossible to complete the request within that time frame. The hard time limit therefore makes the API unsuitable for that type of application.

The remaining possible application for the EndpointSecurity API is to watch for file modifications and creation. For this, the various `ES_EVENT_TYPE_NOTIFY_*` events are indeed suitable.

### DIFF
--- a/ProjFS.Mac/EndpointSecurityPrototype/EndpointSecurityMirror.xcodeproj/project.pbxproj
+++ b/ProjFS.Mac/EndpointSecurityPrototype/EndpointSecurityMirror.xcodeproj/project.pbxproj
@@ -1,0 +1,300 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 50;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		4A6A654B22C227CA00E207E2 /* libEndpointSecurity.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = 4A6A654A22C227CA00E207E2 /* libEndpointSecurity.tbd */; };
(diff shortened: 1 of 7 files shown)


wilbaker:
@pmj thanks for posting this!

I saw the steps for running the provider in the README and I was wondering if you came across any special instructions/guidance on the proper way to install an EndpointSecurity client.

pmj:
> I saw the steps for running the provider in the README and I was wondering if you came across any special instructions/guidance on the proper way to install an EndpointSecurity client.

I've not come across any official documentation about it; the implication seems to be that you're supposed to use a "System Extension," similar to network extensions. I never managed to get this working, and the BuildXL port to EndpointSecurity just seems to use a LaunchDaemon, which seems reasonable.

PR #1597, opened 2019-11-18
Move config & upgrader files to ProgramFiles

### LINKED ISSUE #1585: Move config file and upgrader files into ProgramFiles
This will allow us to simplify the management of the GVFS.Service directory.

alameenshah:
- Changed GVFSDataRoot on Windows platform to
  "C:\Program Files\GVFS\ProgramData".
- Created new LogsDataRoot. LogsDataRoot is unchanged on all platforms.
- gvfs config file now gets created in the new GVFSDataRoot directory.
- Updated installer to copy gvfs.config file from old location to new.
- Updated diagnose verb to use new LogsDataRoot.

Fixes #1585

#TODO (Separate PR): Gather feedback on whether to keep (or remove) code that sets ACLs on DataRoot and LogsRoot directories.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -179,7 +179,13 @@ namespace GVFS.Common
             error = null;
             try
             {
-                this.fileSystem.CopyDirectoryRecursive(currentPath, upgradeApplicationDirectory);
+                // Copying C:\Program Files\GVFS to inside of C:\Program Files\GVFS\ProgramData\GVFS.Upgrade\Tools
+                // directory causes a cycle(at some point we start copying C:\Program Files\GVFS\ProgramData\GVFS.Upgrade
+                // and its contents into C:\Program Files\GVFS\ProgramData\GVFS.Upgrade\Tools). The exclusion below is
+                // added to avoid this loop.
+               HashSet<string> directoriesToExclude = new HashSet<string>();
+                directoriesToExclude.Add(GVFSPlatform.Instance.GetSecureDataRootForGVFS());
+                this.fileSystem.CopyDirectoryRecursive(currentPath, upgradeApplicationDirectory, directoriesToExclude);
             }
             catch (UnauthorizedAccessException e)
             {
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
@@ -29,8 +29,9 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         {
             this.fileSystem = new SystemIORunner();
             this.upgradeDownloadsDirectory = Path.Combine(
-                Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData, Environment.SpecialFolderOption.Create),
+                Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles, Environment.SpecialFolderOption.Create),
                 "GVFS",
+                "ProgramData",
                 "GVFS.Upgrade",
                 "Downloads");
         }
--- a/GVFS/GVFS.Installer.Windows/Setup.iss
+++ b/GVFS/GVFS.Installer.Windows/Setup.iss
@@ -27,6 +27,9 @@
 #define EnvironmentKey "SYSTEM\CurrentControlSet\Control\Session Manager\Environment"
 #define FileSystemKey "SYSTEM\CurrentControlSet\Control\FileSystem"
 #define GvFltAutologgerKey "SYSTEM\CurrentControlSet\Control\WMI\Autologger\Microsoft-Windows-Git-Filter-Log"
+#define GVFSConfigFileName "gvfs.config"
+#define GVFSStatuscacheTokenFileName "EnableGitStatusCacheToken.dat"
+#define ServiceName "GVFS.Service"
 #define ServiceUIName "VFS For Git"
 
 [Setup]
@@ -178,6 +181,9 @@ DestDir: "{app}"; Flags: ignoreversion; Source:"{#GVFSDir}\System.IO.Compression
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.pdb"
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe.config"
 DestDir: "{app}"; Flags: ignoreversion; Source:"{#ServiceDir}\GVFS.Service.exe"; AfterInstall: InstallGVFSService
+
+[Dirs]
+Name: "{app}\ProgramData\{#ServiceName}"; Permissions: users-readexec
 
 [Icons]
 Name: "{commonstartmenu}\{#ServiceUIName}"; Filename: "{app}\GVFS.Service.UI.exe"; AppUserModelID: "GVFS"
@@ -595,6 +601,37 @@ begin
     end;
 end;
 
+procedure MigrateFile(OldPath, NewPath : string);
+begin
+  Log('MigrateFile(' + OldPath + ', ' + NewPath + ')');
+  if (FileExists(OldPath)) then
+    begin
+      if (not FileExists(NewPath)) then
+        begin
+          if (not RenameFile(OldPath, NewPath)) then
+            Log('Could not move ' + OldPath + ' continuing anyway')
+          else
+            Log('Moved ' + OldPath + ' to ' + NewPath);
+        end
+      else
+        Log('Migration cancelled. Newer file exists at path ' + NewPath);
+    end
+  else
+    Log('Migration cancelled. ' + OldPath + ' does not exist');
+end;
+
+procedure MigrateConfigAndStatusCacheFiles();
+var
+  CommonAppDataDir: string;
+  SecureAppDataDir: string;
+begin
+  CommonAppDataDir := ExpandConstant('{commonappdata}\GVFS');
+  SecureAppDataDir := ExpandConstant('{app}\ProgramData');
+
+  MigrateFile(CommonAppDataDir + '\{#GVFSConfigFileName}', SecureAppDataDir + '\{#GVFSConfigFileName}');
+  MigrateFile(CommonAppDataDir + '\{#ServiceName}\{#GVFSStatuscacheTokenFileName}', SecureAppDataDir + '\{#ServiceName}\{#GVFSStatuscacheTokenFileName}');
+end;
+
 function ConfirmUnmountAll(): Boolean;
 var
   MsgBoxResult: integer;
@@ -761,9 +798,10 @@ begin
       end;
     ssPostInstall:
       begin
+        MigrateConfigAndStatusCacheFiles();
+        StartGVFSServiceUI();
         if ExpandConstant('{param:REMOUNTREPOS|true}') = 'true' then
           begin
-            StartGVFSServiceUI();
             MountRepos();
           end
       end;
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -40,7 +40,7 @@ namespace GVFS.Platform.Windows
         {
             get
             {
-                string servicePath = GVFSPlatform.Instance.GetDataRootForGVFSComponent(GVFSConstants.Service.ServiceName);
+                string servicePath = GVFSPlatform.Instance.GetSecureDataRootForGVFSComponent(GVFSConstants.Service.ServiceName);
                 string gvfsDirectory = Path.GetDirectoryName(servicePath);
 
                 return Path.Combine(gvfsDirectory, LocalGVFSConfig.FileName);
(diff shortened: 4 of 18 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
```diff
@@ -136,8 +137,7 @@ private bool ServiceLogContainsUpgradeMessaging()
             string serviceLogFolder = Path.Combine(
                 Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData),
                 "GVFS",
-                GVFSServiceProcess.TestServiceName,
-                "Logs");
+                GVFSServiceProcess.TestServiceName);
```
wilbaker: Why was "Logs" removed from the path?  If we keep a subfolder for log files it might be easier to adjust the ACLs for log files (assuming that's something that is still required).

GVFS/GVFS.Installer.Windows/Setup.iss
```diff
@@ -763,6 +776,7 @@ begin
       begin
         if ExpandConstant('{param:REMOUNTREPOS|true}') = 'true' then
           begin
+            MigrageVFSForGitConfigFile();
```
wilbaker: Should this be moved outside of the `'{param:REMOUNTREPOS|true}'` check?

On a related note, as best I can tell `REMOUNTREPOS` is false when upgrading from GitHub:

https://github.com/microsoft/VFSForGit/blob/4736473e8fc5c1703bb011ad314faf85a14a624b/GVFS/GVFS.Common/GitHubUpgrader.cs#L22

Do you remember why?

GVFS/GVFS.Installer.Windows/Setup.iss
```diff
@@ -595,6 +595,19 @@ begin
     end;
 end;
 
+procedure MigrageVFSForGitConfigFile();
+begin
+  Log('Migrating gvfs.config file from previous installation.');
+  if FileExists(ExpandConstant('{commonappdata}\GVFS\gvfs.config')) and not FileExists(ExpandConstant('{app}\ProgramData\gvfs.config')) then
+      begin
+        Log('Copying file ' + ExpandConstant('{commonappdata}\GVFS\gvfs.config') + ' to ' + ExpandConstant('{app}\ProgramData\gvfs.config'));
+        if (not FileCopy(ExpandConstant('{commonappdata}\GVFS\gvfs.config'),ExpandConstant('{app}\ProgramData\gvfs.config'),True)) then
```
wilbaker: Should we move/rename the file instead? (see [RenameFile](http://www.jrsoftware.org/ishelp/index.php?topic=isxfunc_renamefile)).

That would prevent us from having to worry about two copies existing, and should eliminate the risk of torn writes that comes with `FileCopy`.

GVFS/GVFS.Common/ProductUpgrader.cs
```diff
@@ -179,7 +179,9 @@ public virtual bool TrySetupUpgradeApplicationDirectory(out string upgradeApplic
             error = null;
             try
             {
-                this.fileSystem.CopyDirectoryRecursive(currentPath, upgradeApplicationDirectory);
+                HashSet<string> directoriesToExclude = new HashSet<string>();
+                directoriesToExclude.Add(GVFSPlatform.Instance.GetDataRootForGVFS());
+                this.fileSystem.CopyDirectoryRecursive(currentPath, upgradeApplicationDirectory, directoriesToExclude);
```
wilbaker: How did this work before?  I am wondering why we now need to specify directories to exclude.

wilbaker:
> Gather feedback on whether to keep (or remove) code that sets ACLs on DataRoot and LogsRoot directories.

How is the code that adjusts ACLs impacted by these changes?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/GVFSUpgradeReminderTests.cs
alameenshah: >>Why was "Logs" removed from the path?

After moving the secure files (`gvfs.config` and Upgrader files) only the log files remained in the old location (`C:\ProgramData\GVFS`).

But your thought about ACLs make sense. Also in future we can use this directory to keep non-log files as well.

Adding "Logs" subdirectory - [commit](https://github.com/microsoft/VFSForGit/pull/1597/commits/e052d104edc4a4fcca6bd822f2573d37049b5eb2)

GVFS/GVFS.Installer.Windows/Setup.iss
alameenshah: >>Should this be moved outside of the '{param:REMOUNTREPOS|true}' check?

Yes. Also moved the code that launches `Service UI` to outside this block.

>>Do you remember why?

Yes, b'coz Git upgrader programmatically un-mounts before install and re-mounts after install.

GVFS/GVFS.Common/ProductUpgrader.cs
alameenshah: It used to copy the contents of `C:\Program Files\GVFS` to `C:\ProgramData\GVFS.Upgrade\Tools`  directory and run the upgrade tool from `C:\ProgramData\GVFS.Upgrade\Tools`. 

But now we copy `C:\Program Files\GVFS` to inside of `C:\Program Files\GVFS\ProgramData\GVFS.Upgrade\Tools` directory. This causes a cycle (at some point we start copying `C:\Program Files\GVFS\ProgramData\GVFS.Upgrade` and its contents into `C:\Program Files\GVFS\ProgramData\GVFS.Upgrade\Tools`). The exclusion is added to avoid this loop.

GVFS/GVFS.Installer.Windows/Setup.iss
alameenshah: [Replaced FileCopy with Rename](https://github.com/microsoft/VFSForGit/pull/1597/commits/0e0ce78321b6b098aedbee889a1a0e04f3f5a73d)

GVFS/GVFS.Installer.Windows/Setup.iss
alameenshah: Yes - b'coz Github installer programmatically un-mounts before install (so that it can run Git installer before running `VFSForGit` installer) and re-mounts after install.

alameenshah:
> > Gather feedback on whether to keep (or remove) code that sets ACLs on DataRoot and LogsRoot directories.
> 
> How is the code that adjusts ACLs impacted by these changes?

This PR has No impact on the code that adjusts ACLs. There is a question whether we still need to adjust ACLs on `ProgramData` directory. I will discuss offline and create a separate PR for that.

GVFS/GVFS/CommandLine/DiagnoseVerb.cs
```diff
@@ -122,14 +122,14 @@ protected override void Execute(GVFSEnlistment enlistment)
...
-                            GVFSPlatform.Instance.GetDataRootForGVFS(),
+                            GVFSPlatform.Instance.GetCommonAppDataRootForGVFS(),
                             archiveFolderPath,
                             this.ServiceName,
                             copySubFolders: true);
 
                         // service ui
                         this.CopyAllFiles(
-                            GVFSPlatform.Instance.GetDataRootForGVFS(),
+                            GVFSPlatform.Instance.GetCommonAppDataRootForGVFS(),
```
wilbaker: Do we need additional calls to copy the service and service UI logs? (they are staying in their old location)

GVFS/GVFS.Common/ProductUpgrader.cs
wilbaker: Ok thanks!  Could you add a comment to the code here explaining why this directory is being excluded?

GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs
```diff
@@ -107,6 +107,16 @@ public override string GetDataRootForGVFSComponent(string componentName)
...
 
+        public override string GetCommonAppDataRootForGVFS()
+        {
+            return this.GetDataRootForGVFS();
+        }
+
+        public override string GetLogsDirectoryForGVFSComponent(string componentName)
+        {
+            return Path.Combine(this.GetCommonAppDataRootForGVFS(), componentName);
+        }
```
wilbaker: Should `GetLogsDirectoryForGVFSComponent` be updated to better match the behavior of VFS4G on Windows (the log directory not being a subdirectory of `GetCommonAppDataRootForGVFS`)?

wilbaker: If it's not too much work this might be a good time to also address the TODO in `GetDataRootForGVFS`:

```// TODO: Update this method to return non existant file path.```

File: GVFS/GVFS/CommandLine/DiagnoseVerb.cs (hunk 5 above)
Comment (alameenshah) [reply GVFS/GVFS/CommandLine/DiagnoseVerb.cs]: No need for additional calls - there is no change in how the Logs appear in diagnostics zip file. Here is the structure (it is same as the old).

Screen Shot 2019-12-06 at 7 21 48 AM

File: GVFS/GVFS.Common/ProductUpgrader.cs (hunk 4 above)
Comment (alameenshah) [reply GVFS/GVFS.Common/ProductUpgrader.cs]: [added](https://github.com/microsoft/VFSForGit/pull/1597/commits/bccac470eca503294beb8e58d9780991f9590af0)

File: GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs (hunk 6 above)
Comment (alameenshah) [reply GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs]: >>Should GetLogsDirectoryForGVFSComponent be updated to better match the behavior of VFS4G on Windows (the log directory not being a subdirectory of GetCommonAppDataRootForGVFS)?

Make sense. [Updated](https://github.com/microsoft/VFSForGit/pull/1597/commits/462ab783797119f076c8a9578f85deb284a14108).

Comment (alameenshah) [reply GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs]: >>If it's not too much work this might be a good time to also address the TODO in GetDataRootForGVFS:

Let me take a look.

File: GVFS/GVFS/CommandLine/DiagnoseVerb.cs (hunk 5 above)
Comment (wilbaker) [reply GVFS/GVFS/CommandLine/DiagnoseVerb.cs]: Ahh, I had these mixed up.  Previously we'd also capture the config file in `gvfs diagnose` and now we will not.  Can/should we still copy that over when `diagnose` is run?

I also have trouble keeping the difference between `GetDataRootForGVFS` and `GetCommonAppDataRootForGVFS` straight.  Are there any other method names we could use that would be more descriptive?

Comment (alameenshah) [reply GVFS/GVFS/CommandLine/DiagnoseVerb.cs]: >>Previously we'd also capture the config file in gvfs diagnose and now we will not. Can/should we still copy that over when diagnose is run?

We still capture `gvfs.config` file - the previous screenshot is showing only the directories (not files). Here is the one with root level files as well - 
Screen Shot 2019-12-06 at 12 42 43 PM

>>I also have trouble keeping the difference between GetDataRootForGVFS and GetCommonAppDataRootForGVFS straight. Are there any other method names we could use that would be more descriptive?

How about `GetSecureAppDataRootForGVFS` and `GetCommonAppDataRootForGVFS`

Comment (wilbaker) [reply GVFS/GVFS/CommandLine/DiagnoseVerb.cs]: > We still capture `gvfs.config` file - the previous screenshot is showing only the directories (not files). Here is the one with root level files as well

Where is this getting copied by `diagnose`?  Doesn't the config file live in `GetSecureAppDataRootForGVFS()`?

I'm wondering if you have an old gvfs.config file on your machine that was not migrated into ProgramFiles.

File: GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs (hunk 6 above)
Comment (alameenshah) [reply GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs]: This could be done. [Updated](https://github.com/microsoft/VFSForGit/pull/1597/commits/6776623ed0c0605871acd9911ee60a65da67cc8a)

File: GVFS/GVFS/CommandLine/DiagnoseVerb.cs (hunk 5 above)
Comment (alameenshah) [reply GVFS/GVFS/CommandLine/DiagnoseVerb.cs]: Take a look at [`DiagnoseVerb.cs` line no. 163.](https://github.com/microsoft/VFSForGit/pull/1597/commits/e3b2be03cc53222bff9c51cf8fb9ea0470f03566#diff-fd719e3f262436196484b5920d660d5d)

File: GVFS/GVFS.Platform.Windows/WindowsPlatform.cs [hunk 7]
```diff
@@ -352,7 +364,7 @@ public override bool IsConsoleOutputRedirectedToFile()
 
         public override bool IsGitStatusCacheSupported()
         {
-            return File.Exists(Path.Combine(GVFSPlatform.Instance.GetDataRootForGVFSComponent(GVFSConstants.Service.ServiceName), GVFSConstants.GitStatusCache.EnableGitStatusCacheTokenFile));
+            return File.Exists(Path.Combine(GVFSPlatform.Instance.GetSecureDataRootForGVFSComponent(GVFSConstants.Service.ServiceName), GVFSConstants.GitStatusCache.EnableGitStatusCacheTokenFile));
```
kewillford: Does this change where we are looking for the status cache token file?  Will this turn of status cache?

GVFS/GVFS.UnitTests/Mock/Common/MockPlatform.cs
```diff
@@ -94,17 +94,27 @@ public override string GetOSVersionInformation()
...
         }
 
-        public override string GetDataRootForGVFS()
+        public override string GetSecureDataRootForGVFS()
         {
-            // TODO: Update this method to return non existant file path.
-            return Path.Combine(
-                Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData),
-                "GVFS");
+            return "mock:\\mockRoot";
```
kewillford: `"mock:\\dataRoot"`?

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
alameenshah: As discussed [added ](https://github.com/microsoft/VFSForGit/pull/1597/commits/84d4016e3a2216d66579045eba8a72d9b7521699)a migration for status cache token file.

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
```diff
@@ -1,7 +1,7 @@
-﻿using System;
-using System.IO;
-using GVFS.Common;
+﻿using GVFS.Common;
 using GVFS.Platform.POSIX;
+using System;
+using System.IO;
```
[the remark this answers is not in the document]
alameenshah: I think it is the StyleCop - trying to arrange the headers alphabetically.

GVFS/GVFS.Platform.Mac/MacPlatform.Shared.cs
```diff
@@ -36,7 +36,7 @@ public static string GetUpgradeHighestAvailableVersionDirectoryImplementation()
 
         public static string GetUpgradeNonProtectedDirectoryImplementation()
```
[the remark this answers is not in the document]
alameenshah: On the Mac, unlike on Windows, there is no separate secure & common data roots. I reverted the method name changes in `MacPlatform.Shared.cs` and added comments in `MacPlatform.cs::GetSecureDataRoot*` method overrides.

PR #1600, opened 2019-11-19
WindowsGitHooksInstaller: allow configuring hooks in local git config

### LINKED ISSUE #1563: VFS4G does not merge post-command hooks that are configured per-repo
**Steps to Reproduce**

1. Configure a custom post-command hook in the repo specific git config
2. Remount the repo
3. Observe: The merged `post-command.hooks` file is missing the custom command.  If the setting is configured in the global or system config VFS4G will pick it up.

**Likely Cause**

https://github.com/microsoft/VFSForGit/blob/7390eba79729141c9042b93750a108cdf4dafc2a/GVFS/GVFS.Platform.Windows/WindowsGitHooksInstaller.cs#L46

`forceOutsideEnlistment: true` results in `git config` being called outside of the repo *and* without specifying a .git directory.

As a result, `MergeHooks` is only considering `gvfs.clone.default-post-command` settings that are configured at the global and system level

wilbaker:
Resolves #1563 

Prior to the changes in this commit gvfs.clone.default-pre-command
and gvfs.clone.default-pre-command would only be read from the system
and global git configs.  This meant that users were unable to config
custom pre/post command hooks at a per-repo level.

With this change InvokeGitAgainstDotGitFolder when reading the above
config values, and using InvokeGitAgainstDotGitFolder will give us
the standard git config behavior of local config "winning" over a
system or global config.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
@@ -99,26 +99,22 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         public void MountMergesLocalPrePostHooksConfig()
         {
             // Create some dummy pre/post command hooks
-            string dummyPreCommandHookBin = Path.Combine(this.Enlistment.EnlistmentRoot, "DummyPreCommandHook.bat");
-            this.fileSystem.WriteAllText(dummyPreCommandHookBin, "@echo off && exit 0");
-
-            string dummyPostCommandHookBin = Path.Combine(this.Enlistment.EnlistmentRoot, "DummyPostCommandHook.bat");
-            this.fileSystem.WriteAllText(dummyPreCommandHookBin, "@echo off && exit 0");
+            string dummyCommandHookBin = "cmd.exe /c exit 0";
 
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/MountTests.cs
```diff
@@ -94,6 +94,56 @@ public void MountSetsCoreHooksPath()
             }
         }
 
+        [TestCase]
+        [Category(Categories.WindowsOnly)] // Only Windows uses GitHooksLoader.exe and merges hooks
+        public void MountMergesLocalPrePostHooksConfig()
+        {
+            // Create some dummy pre/post command hooks
+            string dummyPreCommandHookBin = Path.Combine(this.Enlistment.EnlistmentRoot, "DummyPreCommandHook.bat");
+            this.fileSystem.WriteAllText(dummyPreCommandHookBin, "@echo off && exit 0");
```
wilbaker: Looks like the tests don't like this dummy hook:

> Could not execute 'C:\Repos\GVFSFunctionalTests\enlistment\6bba3385e1b54e35a2d4\DummyPostCommandHook.bat'. CreateProcess error (2).

This approach did work on my machine, I will do some digging and see if there's a better way to create a no-op hook in the test.

PR #1601, opened 2019-11-19
sparse: don't warn about dirty files inside the sparse set

### LINKED ISSUE #1584: GVFS Sparse --set "dirty git status" message displays folders that do not need dehydration
GVFS Sparse currently only fails because of a dirty git status if there are files that are dirty that are about to be dehydrated, which is great.

However, when gvfs sparse is called to dehydrate a folder that has a dirty file, it fails which is expected, but what is not expected is it prints to stderr all dirty files even if they are still going to be in the sparse set. 

For example:

D:\mono\src\otools>gvfs sparse --set engtools
Running git status...Failed

?? engtools/qtstore/dirs.txt ?? otools/blah.txt

git status reported that you have dirty files
Either commit your changes or reset and clean

Expected: Only report issue "?? otools/blah.txt" because it is blocking the dehydration of otools. Engtools will still be in the sparse set so we need not warn about it.

wilbaker:
Resolves #1584 

When 'gvfs sparse' found dirty files outside of the sparse set,
it would show all dirty paths to the user (even those inside the
sparse set).

Additionally, when 'gvfs dehydrate' was being called to prune, it
was unnecessarily running its own 'git status' that would fail
even when all of the dirty files were inside of the sparse set.

Two changes were made to address these issues:

 1. The sparse verb will now collect and report the dirty paths
    that were found outside of the sparse set (rather than all
    paths)

 2. A new StatusChecked property has been added to the dehydrate
    verb, and is set to true by the status verb.  When true,
    the dehydrate verb will not call 'git status' itself.

### DIFF
--- a/GVFS/GVFS/CommandLine/SparseVerb.cs
+++ b/GVFS/GVFS/CommandLine/SparseVerb.cs
@@ -399,6 +399,7 @@ Folders need to be relative to the repos root directory.")
                         verb.RunningVerbName = this.VerbName;
                         verb.ActionName = PruneOptionName;
                         verb.Confirmed = true;
+                        verb.StatusChecked = true;
                         verb.Folders = string.Join(FolderListSeparator, directoriesToDehydrate);
                     },
                     this.Output);
@@ -571,6 +572,7 @@ Folders need to be relative to the repos root directory.")
         private void CheckGitStatus(ITracer tracer, GVFSEnlistment enlistment, HashSet<string> sparseFolders)
         {
             GitProcess.Result statusResult = null;
+            HashSet<string> dirtyPathsNotInSparseSet = null;
             if (!this.ShowStatusWhileRunning(
                 () =>
                 {
@@ -581,12 +583,8 @@ Folders need to be relative to the repos root directory.")
                         return false;
                     }
 
-                    if (this.ContainsPathNotCoveredBySparseFolders(statusResult.Output, sparseFolders))
-                    {
-                        return false;
-                    }
-
-                    return true;
+                    dirtyPathsNotInSparseSet = this.GetPathsNotCoveredBySparseFolders(statusResult.Output, sparseFolders);
+                    return dirtyPathsNotInSparseSet.Count == 0;
                 },
                 "Running git status",
                 suppressGvfsLogMessage: true))
@@ -599,9 +597,17 @@ Folders need to be relative to the repos root directory.")
                 }
                 else
                 {
-                    this.WriteMessage(tracer, statusResult.Output);
-                    this.WriteMessage(tracer, "git status reported that you have dirty files");
-                    this.WriteMessage(tracer, "Either commit your changes or reset and clean");
+                    StringBuilder dirtyFilesMessage = new StringBuilder();
+                    dirtyFilesMessage.AppendLine("git status reported that you have dirty files:");
+                    dirtyFilesMessage.AppendLine();
+                    foreach (string path in dirtyPathsNotInSparseSet)
+                    {
+                        dirtyFilesMessage.AppendLine($"    {path}");
+                    }
+
+                    dirtyFilesMessage.AppendLine();
+                    dirtyFilesMessage.Append("Either commit your changes or reset and clean");
+                    this.WriteMessage(tracer, dirtyFilesMessage.ToString());
                 }
 
                 this.Output.WriteLine();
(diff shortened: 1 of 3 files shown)


wilbaker:
@kewillford, @derrickstolee if you have some time the product changes in this PR are ready for review.

I have left the PR as "Draft" for now as I'd like to see what tests I can add to validate these new scenarios (I've only done manual testing of the changes at this point).

wilbaker:
@judemars I made a few tweaks to how the dirty paths are displayed to users.  Rather than leaving them in git's porcelain format each path is put on its own line:

*Example (see output of `gvfs sparse --set ace --prune`)*

```
F:\testSparse\testRepo\src>git status
On branch master

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   ace/test.txt

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        newby/

no changes added to commit (use "git add" and/or "git commit -a")

F:\testSparse\testRepo\src>gvfs sparse --set ace --prune
Running git status...Failed

git status reported that you have dirty files:

    newby/new.txt
    newby/new2.txt

Either commit your changes or reset and clean

Sparse was aborted.
``` 

If that would cause problems for you let me know and I can switch this back to the old format (e.g. `?? newby/new.txt ?? newby/new2.txt`)

judemars:
This is great! Thanks!

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -599,45 +596,52 @@ private void CheckGitStatus(ITracer tracer, GVFSEnlistment enlistment, HashSet<s
...
-                    this.WriteMessage(tracer, statusResult.Output);
-                    this.WriteMessage(tracer, "git status reported that you have dirty files");
-                    this.WriteMessage(tracer, "Either commit your changes or reset and clean");
+                    this.WriteMessage(tracer, "git status reported that you have dirty files:\n");
+                    foreach (string path in dirtyUncoveredPaths)
+                    {
+                        this.WriteMessage(tracer, $"    {path}");
+                    }
+
+                    this.WriteMessage(tracer, "\nEither commit your changes or reset and clean");
```
derrickstolee: Should these `\n` newlines be special-cased depending on platform? Or should we write an empty message to create the blank lines?

wilbaker: > Should these \n newlines be special-cased depending on platform? 

Good point, iI can update these to `Environment.NewLine`.

> Or should we write an empty message to create the blank lines?

We could also write a blank line to the console, but I don't think we'd want to call `this.WriteMessage` as it writes to the log file as well (and it might look a bit weird to have the blank messages there).  We do have the option of just writing a new line to the console directly, but `Environment.NewLine` avoids mixing `WriteMessage` with calls to write directly to the console.

wilbaker: Next push will be a force-push to replace to fix up this first commit, I'll then add a second commit with functional test updates.

derrickstolee: Maybe construct the full multi-line string in a stringbuffer then write them at once in a single WriteMessage call?

wilbaker: That's a good idea, thanks!

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -599,45 +596,52 @@ private void CheckGitStatus(ITracer tracer, GVFSEnlistment enlistment, HashSet<s
...
                 }
 
                 this.Output.WriteLine();
                 this.ReportErrorAndExit(tracer, "Sparse was aborted.");
             }
         }
 
-        private bool ContainsPathNotCoveredBySparseFolders(string statusOutput, HashSet<string> sparseFolders)
+        private HashSet<string> GetPathsNotCoveredBySparseFolders(string statusOutput, HashSet<string> sparseFolders)
```
kewillford: I was thinking maybe we could gather the paths from git status output and then use Linq's `Except`?  I'm not sure how much if any it would simplify this code.

wilbaker: > I'm not sure how much if any it would simplify this code.

Yeah I think in this case it might make it a bit more complicated because of the rename logic (having to track what paths are for a rename) and so I'm thinking it's simpler to stick with the current approach.

PR #1605, opened 2019-11-21
HandleDehydrateFolders: update index using reset

### LINKED ISSUE #1603: 'gvfs sparse' overwrites users's changes if it needs to call dehydrate
Related to the changes in #1601 

If a user has dirty files inside the sparse set and then performs a --prune that needs to remove folder(s) the dirty files will lose their changes.

**Example:**

In this example `test.txt` loses its changes:

```
F:\testSparse\test\src>gvfs sparse --set ace --prune
Running git status...Succeeded
Updating sparse folder set...Succeeded
No folders to update in sparse set.
Finding folders to prune...Succeeded
Found 0 folders to prune.

F:\testSparse\test\src>mkdir new

F:\testSparse\test\src>echo foo >> ace\test.txt

F:\testSparse\test\src>git status
On branch master

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   ace/test.txt

no changes added to commit (use "git add" and/or "git commit -a")

F:\testSparse\test\src>gvfs sparse --set ace --prune
Running git status...Succeeded
Updating sparse folder set...Succeeded
No folders to update in sparse set.
Finding folders to prune...Succeeded
Found 1 folders to prune.

WARNING: If you abort the sparse after this point, the repo may become corrupt

Unmounting...Succeeded
Cleaning up folders...Succeeded
Mounting...Succeeded
new folder prune successful.

F:\testSparse\test\src>git status
On branch master

```

**Root Cause**

As part of dehydrating folders the `dehydrate` verb sends a message to the mount process.  In its handling of that message the mount process ends up calling `git checkout -f HEAD` and that causes the files to be overwritten:

https://github.com/microsoft/VFSForGit/blob/1fbf110dc0068070ac4282dbb03efb3309468423/GVFS/GVFS.Mount/InProcessMount.cs#L306

We should probably update this code to only perform the `checkout -f` against the folders that were dehydrated.

wilbaker:
Resolves #1603

Use `git reset <paths>` rather than `git checkout -f` (or `git checkout
-f <paths>`) because:

 - `git reset <paths>` will not overwrite modified files (like
   `checkout -f` would).
 - `reset` will update the index for paths that are outside of
   ModifiedPaths.dat (unlike `checkout -f <paths>`).

We need the index to be updated because when a folder is dehydrated
it is removed from ModifiedPaths.dat and we need to make sure its
entries in the index are updated.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -466,6 +466,11 @@ namespace GVFS.Common.Git
             return this.InvokeGitInWorkingDirectoryRoot("checkout -f " + target, useReadObjectHook: false);
         }
 
+        public Result Reset(string target, string paths)
+        {
+            return this.InvokeGitInWorkingDirectoryRoot($"reset {target} {paths}", useReadObjectHook: false);
+        }
+
         public Result Status(bool allowObjectDownloads, bool useStatusCache, bool showUntracked = false)
         {
             string command = "status";
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
@@ -407,7 +407,13 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             this.fileSystem.DeleteDirectory(secondFolderToDelete.VirtualPath);
             GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -a -m \"Delete directories\"");
 
+            folderToDelete.VirtualPath.ShouldNotExistOnDisk(this.fileSystem);
+            secondFolderToDelete.VirtualPath.ShouldNotExistOnDisk(this.fileSystem);
+
             GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout -f HEAD~1");
+
+            folderToDelete.VirtualPath.ShouldBeADirectory(this.fileSystem);
+            secondFolderToDelete.VirtualPath.ShouldBeADirectory(this.fileSystem);
 
             this.DehydrateShouldSucceed(
                 new[]
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
@@ -396,18 +396,31 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         }
 
         [TestCase]
-        public void FolderDehydratePreviouslyDeletedFolder()
+        public void FolderDehydratePreviouslyDeletedFolders()
         {
             string folderToDehydrate = "TrailingSlashTests";
             TestPath folderToDelete = new TestPath(this.Enlistment, folderToDehydrate);
+            string secondFolderToDehydrate = "FilenameEncoding";
+            TestPath secondFolderToDelete = new TestPath(this.Enlistment, secondFolderToDehydrate);
+
             this.fileSystem.DeleteDirectory(folderToDelete.VirtualPath);
-            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -a -m \"Delete a directory\"");
+            this.fileSystem.DeleteDirectory(secondFolderToDelete.VirtualPath);
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -a -m \"Delete directories\"");
 
             GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout -f HEAD~1");
 
-            this.DehydrateShouldSucceed(new[] { $"{folderToDehydrate} {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: folderToDehydrate);
+            this.DehydrateShouldSucceed(
+                new[]
+                {
+                    $"{folderToDehydrate} {FolderDehydrateSuccessfulMessage}",
+                    $"{secondFolderToDehydrate} {FolderDehydrateSuccessfulMessage}",
+                },
+                confirm: true,
+                noStatus: false,
+                foldersToDehydrate: new[] { folderToDehydrate, secondFolderToDehydrate });
 
             folderToDelete.VirtualPath.ShouldBeADirectory(this.fileSystem);
+            secondFolderToDelete.VirtualPath.ShouldBeADirectory(this.fileSystem);
         }
 
         [TestCase]
--- a/GVFS/GVFS.Mount/InProcessMount.cs
+++ b/GVFS/GVFS.Mount/InProcessMount.cs
@@ -44,6 +44,10 @@ namespace GVFS.Mount
         private HeartbeatThread heartbeat;
         private ManualResetEvent unmountEvent;
 
+        // True if InProcessMount is calling git reset as part of processing
+        // a folder dehydrate request
+        private volatile bool resetForDehydrateInProgress;
+
         public InProcessMount(ITracer tracer, GVFSEnlistment enlistment, CacheServerInfo cacheServer, RetryConfig retryConfig, GitStatusCacheConfig gitStatusCacheConfig, bool showDebugWindow)
         {
             this.tracer = tracer;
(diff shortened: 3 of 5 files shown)


wilbaker:
@derrickstolee @kewillford I've created this as a draft because I still need to add functional tests.  The product changes are ready for review though.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -461,9 +461,9 @@ public Result CreateBranchWithUpstream(string branchToCreate, string upstreamBra
             return this.InvokeGitAgainstDotGitFolder("branch " + branchToCreate + " --track " + upstreamBranch);
         }
 
-        public Result ForceCheckout(string target)
+        public Result ForceCheckout(string target, string paths = null)
         {
-            return this.InvokeGitInWorkingDirectoryRoot("checkout -f " + target, useReadObjectHook: false);
+            return this.InvokeGitInWorkingDirectoryRoot($"checkout -f {target} {paths}", useReadObjectHook: false);
```
derrickstolee: I had to check to make sure that `null` would write an empty string here, and it does. Carry on.

wilbaker:
This change has caused 2 functional test failures:

```
1) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.FolderDehydratePreviouslyDeletedFolder()
  Directory C:\Repos\GVFSFunctionalTests\enlistment\a6e2ef79cad04872a0c8\src\TrailingSlashTests does not exist
  Expected: True
  But was:  False
   at GVFS.Tests.Should.ValueShouldExtensions.ShouldEqual[T](T actualValue, T expectedValue, String message) in C:\agent\_work\18\s\GVFS\GVFS.Tests\Should\ValueShouldExtensions.cs:line 34
   at GVFS.FunctionalTests.Should.FileSystemShouldExtensions.DirectoryAdapter..ctor(String path, FileSystemRunner runner) in C:\agent\_work\18\s\GVFS\GVFS.FunctionalTests\Should\FileSystemShouldExtensions.cs:line 140
   at GVFS.FunctionalTests.Should.FileSystemShouldExtensions.ShouldBeADirectory(String path, FileSystemRunner runner) in C:\agent\_work\18\s\GVFS\GVFS.FunctionalTests\Should\FileSystemShouldExtensions.cs:line 46
   at GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.FolderDehydratePreviouslyDeletedFolder() in C:\agent\_work\18\s\GVFS\GVFS.FunctionalTests\Tests\EnlistmentPerFixture\DehydrateTests.cs:line 433

2) Failed : GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.FolderDehydrateTombstone()
  Directory C:\Repos\GVFSFunctionalTests\enlistment\a6e2ef79cad04872a0c8\src\TrailingSlashTests exists when it should not
  Expected: False
  But was:  True
   at GVFS.Tests.Should.ValueShouldExtensions.ShouldEqual[T](T actualValue, T expectedValue, String message) in C:\agent\_work\18\s\GVFS\GVFS.Tests\Should\ValueShouldExtensions.cs:line 34
   at GVFS.FunctionalTests.Should.FileSystemShouldExtensions.ShouldNotExistOnDisk(String path, FileSystemRunner runner) in C:\agent\_work\18\s\GVFS\GVFS.FunctionalTests\Should\FileSystemShouldExtensions.cs:line 57
   at GVFS.FunctionalTests.Tests.EnlistmentPerFixture.DehydrateTests.FolderDehydrateTombstone() in C:\agent\_work\18\s\GVFS\GVFS.FunctionalTests\Tests\EnlistmentPerFixture\DehydrateTests.cs:line 446
```

Looking into those next.

wilbaker:
Assuming tests pass, my next push will squash the current commits and update the PR description.

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -299,11 +304,25 @@ private void HandleDehydrateFolders(NamedPipeMessages.Message message, NamedPipe
...
+                EventMetadata resetIndexMetadata = new EventMetadata();
+                resetIndexMetadata.Add(nameof(resetPaths), resetPaths);
+                resetIndexMetadata.Add(nameof(refreshIndexResult.ExitCode), refreshIndexResult.ExitCode);
+                resetIndexMetadata.Add(nameof(refreshIndexResult.Output), refreshIndexResult.Output);
+                resetIndexMetadata.Add(nameof(refreshIndexResult.Errors), refreshIndexResult.Errors);
+                resetIndexMetadata.Add(TracingConstants.MessageKey.InfoMessage, $"{nameof(this.HandleDehydrateFolders)}: Reset git index");
+                this.tracer.RelatedEvent(EventLevel.Informational, $"{nameof(this.HandleDehydrateFolders)}_ResetIndex", resetIndexMetadata);
+
+                // Update the projection to ensure we start projecting files that were dehydrated
+                this.fileSystemCallbacks.ForceIndexProjectionUpdate(invalidateProjection: true, invalidateModifiedPaths: false);
```
wilbaker: Make sure we have tests that dehydrate and prune multiple folders

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -299,11 +304,25 @@ private void HandleDehydrateFolders(NamedPipeMessages.Message message, NamedPipe
...
+
+                    resetFolderPaths.Append($"\"{folder.Replace(Path.DirectorySeparatorChar, GVFSConstants.GitPathSeparator)}\" ");
                 }
 
-                // Since placeholders and modified paths could have changed with the dehydrate, the index needs to be rebuilt
+                // Since modified paths could have changed with the dehydrate, the paths that were dehydrated need to be reset in the index
+                string resetPaths = resetFolderPaths.ToString();
                 GitProcess gitProcess = new GitProcess(this.enlistment);
-                gitProcess.ForceCheckout(GVFSConstants.DotGit.HeadName);
+                GitProcess.Result refreshIndexResult = gitProcess.Reset(GVFSConstants.DotGit.HeadName, resetPaths);
```
kewillford: What do we do if this fails?  What state will things be in?  Will the user be able to recover and is there some message we can surface to the user?  I realize we weren't even checking the result before but we are now allowing this to run when the user doesn't have a clean status.

wilbaker: All great questions!

> What state will things be in?

Assuming that the reset completely fails (leaving things in the same state that they were in before) users will be unable to see some or all of the contents of the folder(s) that were just dehydrated.  The folder (or some of its contents) will be missing because:

- VFS4G only projects files whose skip-worktree bit is set
- The failure of `git reset` means that the entries in the index were not updated to have their skip worktree bit set.

Additionally, `git status` would not report these files as deleted because they are not in ModifiedPaths.dat.

> Will the user be able to recover and is there some message we can surface to the user?

It is possible to recover from this situation, but with the current code knowing how to recover (or even that recovering is required) would be difficult for most users.  We could report back that this portion of the `dehydrate` failed, but I think we'll also have to decide what step(s) to ask the user to run.

> What do we do if this fails?

Related to the question of what we surface to the user, I think we have a few options:

- If we're willing to take a perf hit, we could dehydrate/reset each folder one at a time.  If the dehydrate fails we could add the failed folder to ModifiedPaths.dat and ask the user to try again.

- If we want to avoid lots of calls to `git reset` we could be pessimistic and all of the folders that were dehydrated to ModifiedPaths.dat and ask the user to run the command again.

- We could tell the user that VFS4G failed to update the index, and ask them commit any changes they may have and run `git checkout HEAD`.  I am not sure if this advice will always be 100% good though, as if `dehydrate` is called from a post-command hook a user could find themselves in this situation while another git command (e.g. merge, rebase, etc.) is still in progress.

I am leaning toward option 2 (possibly as a follow up PR), but I'd like to hear if you can think of any alternatives.  

CC: @derrickstolee for his thoughts on how we can avoid straying too far from native git sparse-checkout behavior in similar situations.

wilbaker: Filed #1606 for this work.

kewillford: Yeah option 2 does seem like the best. Anything we can do to help the user know what to do but also not make to big an impact on performance would be best.

GVFS/GVFS.Mount/InProcessMount.cs
wilbaker: This call to `ForceIndexProjectionUpdate` is racing with the `OnIndexWriteRequiringModifiedPathsValidation` background task that's getting scheduled by the `git reset`.  See if we can piggy-pack off of the call to `HandlePostIndexChangedRequest` by `git reset` to rebuild the projection.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -395,19 +395,32 @@ public void FolderDehydrateCannotDehydrateDotGitFolder()
...
             TestPath folderToDelete = new TestPath(this.Enlistment, folderToDehydrate);
+            string secondFolderToDehydrate = "FilenameEncoding";
+            TestPath secondFolderToDelete = new TestPath(this.Enlistment, secondFolderToDehydrate);
+
             this.fileSystem.DeleteDirectory(folderToDelete.VirtualPath);
-            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -a -m \"Delete a directory\"");
+            this.fileSystem.DeleteDirectory(secondFolderToDelete.VirtualPath);
+            GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -a -m \"Delete directories\"");
 
             GitProcess.Invoke(this.Enlistment.RepoRoot, "checkout -f HEAD~1");
```
kewillford: I wonder if we should check for the folder existence before and after the `checkout`?

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -403,7 +437,20 @@ private void HandlePostIndexChangedRequest(NamedPipeMessages.Message message, Na
...
             {
-                this.fileSystemCallbacks.ForceIndexProjectionUpdate(request.UpdatedWorkingDirectory, request.UpdatedSkipWorktreeBits);
+                if (this.resetForDehydrateInProgress)
+                {
+                    // To avoid having to parse the index twice when dehydrating folders, repurpose the PostIndexChangedRequest
+                    // for git reset to rebuild the projection.  Additionally, if we were to call ForceIndexProjectionUpdate
+                    // directly in HandleDehydrateFolders we'd have a race condition where the OnIndexWriteRequiringModifiedPathsValidation
+                    // background task would be trying to parse the index at the same time as HandleDehydrateFolders
+
+                    this.fileSystemCallbacks.ForceIndexProjectionUpdate(invalidateProjection: true, invalidateModifiedPaths: false);
```
kewillford: Do we still need to pass `true` for the `invalidateModifiedPaths`?  `git reset` passes `true` for the `request.UpdatedSkipWorktreeBits` but in this case do we ever expect the reset that the dehydrate runs to change turn off skipworktree bits so those paths need to be added to the modified paths?  I don't think so since we are explictly passing in the paths that should be reset and they are the paths that were dehydrated and should get the skipworktree bits set.

wilbaker: > but in this case do we ever expect the reset that the dehydrate runs to change turn off skipworktree bits so those paths need to be added to the modified paths? I don't think so since we are explictly passing in the paths that should be reset and they are the paths that were dehydrated and should get the skipworktree bits set.

Yeah that was my thought as well.  

In my first round of changes I didn't explicitly set `invalidateModifiedPaths` to false and `AddMissingModifiedFiles` was always getting called.  As you've pointed out, the `git reset` that we're running should never turn skip-worktree bits off as we're calling it for folders we just dehydrated.  With that in mind I made the switch to pass in `false` for perf reasons.  It shouldn't hurt anything functionality wise to leave `invalidateModifiedPaths` as `request.UpdatedSkipWorktreeBits` though, and so let me know if you'd prefer that.

kewillford: Nope this looks good to me.

PR #1607, opened 2019-11-26
sparse: check for not recursive matches when finding dirty files

### LINKED ISSUE #1604: 'gvfs sparse' reports files as dirty (incorrectly) when their parent folders are not recursively in the sparse set
This looks like a case that was missed in #1601

Example:

```
F:\testSparse\test\src>gvfs sparse --set ace/bin --prune
Running git status...Failed

git status reported that you have dirty files:

    ace/test.txt

Either commit your changes or reset and clean

Sparse was aborted.
```

In this case `ace/test.txt` is still in the sparse set and so it should not be reported as dirty.

wilbaker:
Resolves #1604

The previous file path matching was only matching against recursive
sparse paths, and incorrectly flagging children of non-recursive folders
as dirty.  This matching has been updated to match both recursive and
non-recursive parents, and unit tests were added to validate the
matching behavior.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
@@ -485,7 +485,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             // Remove and prune folderToCreateFileIn
             string output = this.gvfsProcess.RemoveSparseFolders(shouldPrune: true, folders: folderToCreateFileIn);
             output.ShouldContain("Running git status...Succeeded");
-            this.ValidateFoldersInSparseList(this.mainSparseFolder);
+            this.ValidateFoldersInSparseList(this.mainSparseFolder, additionalSparseFolder);
 
             // Confirm the prune succeeded
             string folderPath = Path.Combine(this.Enlistment.RepoRoot, folderToCreateFileIn);
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
@@ -455,23 +455,31 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
         [TestCase, Order(22)]
         public void ModifiedFileInSparseSetShouldAllowPrune()
         {
-            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder);
-            this.ValidateFoldersInSparseList(this.mainSparseFolder);
+            string additionalSparseFolder = Path.Combine("GVFS", "GVFS.Tests", "Should");
+            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder, additionalSparseFolder);
+            this.ValidateFoldersInSparseList(this.mainSparseFolder, additionalSparseFolder);
 
             // Ensure that folderToCreateFileIn is on disk so that there's something to prune
             string folderToCreateFileIn = Path.Combine("GVFS", "GVFS.Common");
-            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder, folderToCreateFileIn);
-            this.ValidateFoldersInSparseList(this.mainSparseFolder, folderToCreateFileIn);
+            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder, additionalSparseFolder, folderToCreateFileIn);
+            this.ValidateFoldersInSparseList(this.mainSparseFolder, additionalSparseFolder, folderToCreateFileIn);
 
             string fileToCreate = Path.Combine(this.Enlistment.RepoRoot, folderToCreateFileIn, "newfile.txt");
             this.fileSystem.WriteAllText(fileToCreate, "New Contents");
             GitProcess.Invoke(this.Enlistment.RepoRoot, "add .");
             GitProcess.Invoke(this.Enlistment.RepoRoot, "commit -m Test");
 
-            // Modify a file that's in the sparse set
+            // Modify a file that's in the sparse set (recursively)
             string modifiedFileContents = "New Contents";
-            string modifiedPath = Path.Combine(this.Enlistment.RepoRoot, "GVFS", "GVFS", "Program.cs");
+            string modifiedPath = this.Enlistment.GetVirtualPathTo("GVFS", "GVFS", "Program.cs");
+            modifiedPath.ShouldBeAFile(this.fileSystem);
             this.fileSystem.WriteAllText(modifiedPath, modifiedFileContents);
+
+            // Modify a file that is in the sparse set (via a non-recursive parent)
+            string secondModifiedPath = this.Enlistment.GetVirtualPathTo("GVFS", "GVFS.Tests", "NUnitRunner.cs");
+            secondModifiedPath.ShouldBeAFile(this.fileSystem);
+            this.fileSystem.WriteAllText(secondModifiedPath, modifiedFileContents);
+
             string expecetedStatusOutput = GitProcess.Invoke(this.Enlistment.RepoRoot, "status --porcelain -uall");
 
             // Remove and prune folderToCreateFileIn
@@ -486,6 +494,7 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
 
             // Confirm the changes to the modified file are preserved and that status does not change
             modifiedPath.ShouldBeAFile(this.fileSystem).WithContents(modifiedFileContents);
+            secondModifiedPath.ShouldBeAFile(this.fileSystem).WithContents(modifiedFileContents);
             string statusOutput = GitProcess.Invoke(this.Enlistment.RepoRoot, "status --porcelain -uall");
             statusOutput.ShouldEqual(expecetedStatusOutput, "Status output should not change.");
         }
--- a/GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
+++ b/GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
@@ -17,50 +17,71 @@ namespace GVFS.UnitTests.Windows.Windows.CommandLine
         private static readonly HashSet<string> EmptySparseSet = new HashSet<string>();
 
         [TestCase]
+        public void GetNextGitPathGetsPaths()
+        {
+            string testStatusOutput = $"a.txt{StatusPathSeparatorToken}";
+            ConfirmGitPathsParsed(testStatusOutput, new List<string>() { "a.txt" });
+
+            testStatusOutput = $"a.txt{StatusPathSeparatorToken}b.txt{StatusPathSeparatorToken}c.txt{StatusPathSeparatorToken}";
+            ConfirmGitPathsParsed(testStatusOutput, new List<string>() { "a.txt", "b.txt", "c.txt" });
+
+            testStatusOutput = $"a.txt{StatusPathSeparatorToken}d/b.txt{StatusPathSeparatorToken}d/c.txt{StatusPathSeparatorToken}";
+            ConfirmGitPathsParsed(testStatusOutput, new List<string>() { "a.txt", "d/b.txt", "d/c.txt" });
+        }
+
+        [TestCase]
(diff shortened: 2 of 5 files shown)


GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -93,6 +93,44 @@ private enum SetDirectoryTimeResult
...
+            string parentPath = filePath.Substring(startIndex: 0, length: pathSeparatorIndex + 1);
+            if (sparseFolders.Any(x => x.StartsWith(parentPath, GVFSPlatform.Instance.Constants.PathComparison)))
+            {
+                // Path is a child of a non-recursive entry
+                //
+                // Example:
+                //  - Sparse set: A\B\C
+                //  - filePath: A\B\d.txt
+                //
+                // This file is in the sparse set because its parent ("A\B\") is an ancestory of a recursive
```
wilbaker: `ancestory` -> `ancestor`

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace GVFS.UnitTests.Windows.Windows.CommandLine
+{
+    [TestFixture]
+    public class SparseVerbTests
```
wilbaker: Update functional tests as well?

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -93,6 +93,44 @@ private enum SetDirectoryTimeResult
 
         protected override string VerbName => SparseVerbName;
 
+        internal static bool PathCoveredBySparseFolders(ref int index, string statusOutput, HashSet<string> sparseFolders, out string gitPath)
```
derrickstolee: `internal` :heart:

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        // Shorter name for readability
+        private static char dirSep = Path.DirectorySeparatorChar;
```
kewillford: Since this is only Windows why can't we just use a `\`?

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -93,6 +93,44 @@ private enum SetDirectoryTimeResult
...
         protected override string VerbName => SparseVerbName;
 
+        internal static bool PathCoveredBySparseFolders(ref int index, string statusOutput, HashSet<string> sparseFolders, out string gitPath)
+        {
+            int endOfPathIndex = statusOutput.IndexOf(StatusPathSeparatorToken, index);
+            gitPath = statusOutput.Substring(index, endOfPathIndex - index);
+            index = endOfPathIndex + 1;
+
+            string filePath = gitPath.Replace(GVFSConstants.GitPathSeparator, Path.DirectorySeparatorChar);
+            if (sparseFolders.Any(x => filePath.StartsWith(x + Path.DirectorySeparatorChar, GVFSPlatform.Instance.Constants.PathComparison)))
```
derrickstolee: Since `sparseFolders` is a `HashSet<string>` you are probably better off stripping parents from `filePath` and checking `sparseFolders.Contains(filePath)`. [Here is similar code in the `git sparse-checkout` implementation](https://github.com/derrickstolee/git/blob/7577ffc034be90bd60da17e8f3189d999e7f8149/dir.c#L731-L756).

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -93,6 +93,44 @@ private enum SetDirectoryTimeResult
...
+            int pathSeparatorIndex = filePath.LastIndexOf(Path.DirectorySeparatorChar);
+            if (pathSeparatorIndex < 0)
+            {
+                // Path is in the root, and root entries are always in the sparse set
+                return true;
+            }
+
+            // Get the parent path (including the path separator)
+            string parentPath = filePath.Substring(startIndex: 0, length: pathSeparatorIndex + 1);
+            if (sparseFolders.Any(x => x.StartsWith(parentPath, GVFSPlatform.Instance.Constants.PathComparison)))
```
derrickstolee: This has a less-obvious perf change, but creating a `HashSet<string>` full of the "parent paths" (i.e. those that are parents of the sparse folders, but not actually in the sparse folders) we can test `parentFolders.Contains(parentPath)`. [Here is the equivalent Git code](https://github.com/derrickstolee/git/blob/7577ffc034be90bd60da17e8f3189d999e7f8149/dir.c#L631-L717), but the sparse-checkout file already has the full list of parent paths as part of the list, so it may not be incredibly helpful.

The only reason I bring this up is because the current algorithm is quadratic: O(N*M) where N is the number of sparse folders and M is the number of dirty paths. Using the `HashSet<string>` method I describe will be O(N+M*d) (where d is the max depth of a dirty path) because O(N) time to construct the hash sets and O(M*d) time to check the values.

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        [TestCase]
+        public void PathCoveredBySparseFolders_RootPaths()
+        {
+            string rootPaths = $"a.txt{SparseVerb.StatusPathSeparatorToken}b.txt{SparseVerb.StatusPathSeparatorToken}c.txt{SparseVerb.StatusPathSeparatorToken}";
```
kewillford: I think that you are using `SparseVerb.StatusPathSeparatorToken` so that if it changes it won't break the tests but I would argue that the tests should __not__ use internal information and should break.  So my suggestion is to have
```
private const char StatusPathSeparatorToken = '\0';
```
in the test.  It will only be one place to change but tests will break when the internals of the SparseVerb change.

GVFS/GVFS/CommandLine/SparseVerb.cs
```diff
@@ -93,6 +93,44 @@ private enum SetDirectoryTimeResult
 
         protected override string VerbName => SparseVerbName;
 
+        internal static bool PathCoveredBySparseFolders(ref int index, string statusOutput, HashSet<string> sparseFolders, out string gitPath)
+        {
+            int endOfPathIndex = statusOutput.IndexOf(StatusPathSeparatorToken, index);
+            gitPath = statusOutput.Substring(index, endOfPathIndex - index);
+            index = endOfPathIndex + 1;
```
kewillford: Just thinking here and wondering if it would make the tests easier and more specific if we pulled this out into something like:
```
private static string GetNextGitPath(ref int index, string statusOutput)
{
    int endOfPathIndex = statusOutput.IndexOf(StatusPathSeparatorToken, index);
    gitPath = statusOutput.Substring(index, endOfPathIndex - index);
    index = endOfPathIndex + 1;
    return gitPath;
}
```
Then this method could just be 
```
internal static bool PathCoveredBySparseFolders(string gitPath, HashSet<string> sparseFolders)
```
`GetPathsNotCoveredBySparseFolders` would need to change to something like
```
string gitPath = GetNextGitPath(ref index, statusOutput);
if (!PathCoveredBySparseFolders(gitPath, sparseFolders))
{
...
```
Then the tests wouldn't need to be testing the parsing of the status output although it would be good to have some coverage there as well.  So I think after writing this I am more on the fence.

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            ConfirmAllPathsCovered(rootPaths, new HashSet<string>());
+            ConfirmAllPathsCovered(rootPaths, new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "A" });
+            ConfirmAllPathsCovered(rootPaths, new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "A", $"B{dirSep}C" });
```
kewillford: Maybe have some static `HashSet`s for these?

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
wilbaker: We could do that, and I will make the change.

I went back and forth on the best way to set these up.  The trouble is these tests aren't really *Windows* specific, but they depend on the `GVFS.Windows` project (which is Windows specific).

I will update these to use `\`, and if we decide to make the test cross-plat at some point we can move the code into a .NET Standard assembly or update `GVFS.Windows` to be .NET Core (assuming ProjFS is by that point).

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: Thanks for the great perf suggestions.  I think I will file a follow up for these so that those changes can be isolated as a performance only PR.

wilbaker: Filed #1608

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: Filed follow-up #1608 to address this.

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
wilbaker: I will give that a try and use them where it improves readability.

GVFS/GVFS/CommandLine/SparseVerb.cs
wilbaker: I like this suggestion, thanks!  I will update `SparseVerb` and add tests for `GetNextGitPath`.

GVFS/GVFS.UnitTests.Windows/Windows/CommandLine/SparseVerbTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            rootPaths.Append($"A/B/B.txt{StatusPathSeparatorToken}");
+            rootPaths.Append($"A/C.txt{StatusPathSeparatorToken}");
+            rootPaths.Append($"A/D/E/C.txt{StatusPathSeparatorToken}");
+
+            ConfirmAllPathsCovered(
+                rootPaths.ToString(),
+                new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+                {
+                    $"A\\B\\C\\D",
+                    $"A\\D\\E\\F\\G"
```
kewillford: We can get rid of `$` now and use `@`
```
@"A\B\C\D",
@"A\D\E\F\G"
```

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/SparseTests.cs
```diff
@@ -455,23 +455,31 @@ public void ModifiedFileOldSparseSetShouldNotAllowSparseFolderRemove()
...
+            string additionalSparseFolder = Path.Combine("GVFS", "GVFS.Tests", "Should");
+            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder, additionalSparseFolder);
+            this.ValidateFoldersInSparseList(this.mainSparseFolder, additionalSparseFolder);
 
             // Ensure that folderToCreateFileIn is on disk so that there's something to prune
             string folderToCreateFileIn = Path.Combine("GVFS", "GVFS.Common");
-            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder, folderToCreateFileIn);
-            this.ValidateFoldersInSparseList(this.mainSparseFolder, folderToCreateFileIn);
+            this.gvfsProcess.AddSparseFolders(this.mainSparseFolder, additionalSparseFolder, folderToCreateFileIn);
+            this.ValidateFoldersInSparseList(this.mainSparseFolder, additionalSparseFolder, folderToCreateFileIn);
```
wilbaker: This change is needed on line 488 as well

wilbaker:
Next commit will address @kewillford's comment here:

https://github.com/microsoft/VFSForGit/pull/1607#discussion_r350968841

PR #1610, opened 2019-12-05
dehydrate: skip mount validation steps

### LINKED ISSUE #1592: 'gvfs dehydrate' should configure 'MountVerb' to skip mounted and version checks
When `gvfs sparse` needs to dehydrate folders, the "Mounting..." step feels slow.

We should be able to speed it up be skipping the mounted check (`dehydrate` will have already unmounted the repo) and the version check (the repo needs to be mounted before `dehydrate` can run, and so the version must be correct).

wilbaker:
Resolves #1592 

When re-mouting a repo as part of dehydrate there is no need to
check if the repo is already mounted, as dehydrate itself unmounted it.

Additionally, there is no need to perform a version check (unless
--no-status was specified) as dehydrate would have required the repo
already be mounted to run 'git status' in it.

On my machine this reduced the time required by calls to `gvfs sparse --set <folder> --prune` that require dehydrate by 2 to 3 seconds (down from ~34 to ~31.5 seconds)

### DIFF
--- a/GVFS/GVFS/CommandLine/DehydrateVerb.cs
+++ b/GVFS/GVFS/CommandLine/DehydrateVerb.cs
@@ -298,7 +298,12 @@ from a parent of the folders list.
                 this.ReportErrorAndExit(tracer, $"{this.ActionName} for folders failed.");
             }
 
-            this.Mount(tracer);
+            // We can skip the version check because dehydrating folders requires that a git status
+            // be run first, and running git status requires that the repo already be mounted (meaning
+            // we don't need to perform another version check again)
+            this.Mount(
+                tracer,
+                skipVersionCheck: true);
 
             if (foldersToDehydrate.Count > 0)
             {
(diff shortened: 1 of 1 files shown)


GVFS/GVFS/CommandLine/DehydrateVerb.cs
```diff
@@ -298,7 +298,11 @@ private void DehydrateFolders(JsonTracer tracer, GVFSEnlistment enlistment, stri
                 this.ReportErrorAndExit(tracer, $"{this.ActionName} for folders failed.");
             }
 
-            this.Mount(tracer);
+            this.Mount(
+                tracer,
+                skipVersionCheck: true);  // Dehydrating folders requires that a git status be run first, and
+                                          // running git status requires that the repo already be mounted (meaning
+                                          // we don't need to check the versions again)
```
kewillford: Why did you format it this way?  Wouldn't it make just as much sense to do it like this?
```
// Dehydrating folders requires that a git status be run first, and
// running git status requires that the repo already be mounted (meaning
// we don't need to check the versions again)
this.Mount(tracer, skipVersionCheck: true);
```

Same for the other calls to `Mount`

wilbaker: Good point, I will move these comments, thanks!

PR #1611, opened 2019-12-05
Merge milestones/M162 to releases/shipped

wilbaker:
Release notes:

- Fixes an issue where File Explorer does not refresh after running `gvfs sparse`
- Allows hooks to be configured in local git config files
- Fixes an issue where `gvfs sparse` warns about dirty files inside the sparse set
- Fixes an issue where `gvfs sparse` warns about dirty files that are inside non-recursive entries in the sparse set
- Various other bug fixes and performance improvements


wilbaker:
@jrbriggs the results of "GitHub VFSForGit Large Repo Perf Tests" look roughly the same as the last run on `releases/shipped`.

PR #1614, opened 2019-12-09
Review git settings used in GVFSVerb

### LINKED ISSUE #1587: Old repo/local configuration entries are not removed
On clone/mount we **_add_** configuration entries to the `--local` Git configuration, things like `core.gvfs=1` and `credential.validate=false`.

The problem is that in the past we've set entries like `credential.usehttppath=true` which we no-longer set on new clones. The means the behaviour of older clones is different from newer ones.

We should ensure entries that we have set in previous version of VFS for Git / GVFS are removed or unset on mount to ensure consistent behaviour.

Aside: we no longer set `credential.usehttppath` but `credential.https://dev.azure.com.usehttppath` instead.

alameenshah:
#### Goal is to identify and remove/un-set git config settings that are not relevant today.

- Went through the commit history and documented git config settings that are set it GVFSVerb.
- Except for 3 config settings, am able to trace the commit history for the remaining. These settings seem to be still required.
- 3 settings for which I could not find history and it is not clear we still need them - `core.safecrlf`,  `core.repositoryformatversion` and `credential.validate`.

Fixes #1587

### DIFF
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -124,39 +124,112 @@ namespace GVFS.CommandLine
             //
             Dictionary<string, string> requiredSettings = new Dictionary<string, string>
             {
+                 // When running 'git am' it will remove the CRs from the patch file by default. This causes the patch to fail to apply because the
+                 // file that is getting the patch applied will still have the CRs. There is a --keep-cr option that you can pass the 'git am' command
+                 // but since we always want to keep CRs it is better to just set the config setting to always keep them so the user doesn't have to
+                 // remember to pass the flag. https://mseng.visualstudio.com/VSOnline/_git/GVFS/pullrequest/244891?_a=overview.
                 { "am.keepcr", "true" },
+
+                // Update git settings to enable optimizations in git 2.20
+                // Set 'checkout.optimizeNewBranch=true' to enable optimized 'checkout -b'
                 { "checkout.optimizenewbranch", "true" },
+
+                // Needed to prevent line ending conversion. Prevents git from auto-replacing line endings with Crlf.
                 { "core.autocrlf", "false" },
+
+                // commit graph. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/299081
                 { "core.commitGraph", "true" },
+
+                // Perf - Git for Windows uses this to bulk-read and cache lstat data of entire directories (instead of doing lstat file by file).
                 { "core.fscache", "true" },
+
+                // Turns on all special gvfs logic. https://github.com/microsoft/scalar/blob/935c50110b31c440aae8e1cdc4ffea72bc3be01f/Scalar/CommandLine/ScalarVerb.cs#L35
                 { "core.gvfs", "true" },
+
+                // Use 'multi-pack-index' builtin instead of 'midx' to match upstream implementation
                 { "core.multiPackIndex", "true" },
+
+                // Perf - Enable parallel index preload for operations like git diff
                 { "core.preloadIndex", "true" },
                 { "core.safecrlf", "false" },
+
+                // Possibly cause hydration while creating untrackedCache.
                 { "core.untrackedCache", "false" },
                 { "core.repositoryformatversion", "0" },
+
+                // Support file modes onr Mac & Linux.
                 { "core.filemode", GVFSPlatform.Instance.FileSystem.SupportsFileMode ? "true" : "false" },
+
+                // For consistency with git init. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/235679
                 { "core.bare", "false" },
+
+                // For consistency with git init. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/235679
                 { "core.logallrefupdates", "true" },
+
+                // Git to download objects on demand.
                 { GitConfigSetting.CoreVirtualizeObjectsName, "true" },
+
+                // Configure hook that git calls to ask GVFS abvout what fiels it cares about
                 { GitConfigSetting.CoreVirtualFileSystemName, Paths.ConvertPathToGitFormat(GVFSConstants.DotGit.Hooks.VirtualFileSystemPath) },
+
+                // Ensure hooks path is configured correctly.
                 { "core.hookspath", expectedHooksPath },
+
+                // Hostname is no longer sufficent for VSTS authentication. VSTS now requires dev.azure.com/account to determine the tenant.
+                // By setting useHttpPath, credential managers will get the path which contains the account as the first parameter. They can then use this information for auth appropriately.
                 { GitConfigSetting.CredentialUseHttpPath, "true" },
                 { "credential.validate", "false" },
+
+                // This setting is not needed anymore, because current version of gvfs does not use index.lock.
+                // (This change was introduced initially to prevent `git diff` from acquiring index.lock file.)
                 { "diff.autoRefreshIndex", "true" },
+
+                // Turn off of git garbage collection. Git garbage collection does not work with virtualized object. Check with Stolee - probably not needed.
                 { "gc.auto", "0" },
+
+                // Prevent git GUI from displaying a GC warning. https://mseng.visualstudio.com/AzureDevOps/_git/GVFS/pullrequest/184856?_a=overview
                 { "gui.gcwarning", "false" },
+
+                // Update git settings to enable optimizations in git 2.20
+                // Set 'index.threads=true' to enable multi-threaded index reads
                 { "index.threads", "true" },
+
+                // gvfs index parser would work only with index versions 4 and above.
                 { "index.version", "4" },
+
+                // Turned off for performance reasons.
                 { "merge.stat", "false" },
+
+                // Update git version to 2.18.0.gvfs.1.26.gf61ade4 - This updates the git version to 2.18.0 which includes changes to the rename detection to use
+                // config setting which was previously turned off with a patch in git.
                 { "merge.renames", "false" },
+
+                // Don't use bitmaps to determine pack file contents, b'coz we use MIDX for this.
                 { "pack.useBitmaps", "false" },
+
+                // Update Git to include sparse push algorithm
                 { "pack.useSparse", "true" },
+
+                // Stop automatic git GC
                 { "receive.autogc", "false" },
+
+                // Update git settings to enable optimizations in git 2.20
+                // Set 'reset.quiet=true' to speed up 'git reset <foo>"
                 { "reset.quiet", "true" },
+
+                // Configure git to use our serialize status file
                 { "status.deserializePath", gitStatusCachePath },
+
+                // In Git 2.24.0, some new config settings were created. Disable them locally in VFS for Git repos in case a user has set them globally.
+                // https://github.com/microsoft/VFSForGit/pull/1594
                 { "feature.manyFiles", "false" },
+
+                // In Git 2.24.0, some new config settings were created. Disable them locally in VFS for Git repos in case a user has set them globally.
+                // https://github.com/microsoft/VFSForGit/pull/1594
                 { "feature.experimental", "false" },
+
+                // In Git 2.24.0, some new config settings were created. Disable them locally in VFS for Git repos in case a user has set them globally.
+                // https://github.com/microsoft/VFSForGit/pull/1594
                 { "fetch.writeCommitGraph", "false" },
             };
 
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -127,23 +127,23 @@ namespace GVFS.CommandLine
                  // When running 'git am' it will remove the CRs from the patch file by default. This causes the patch to fail to apply because the
                  // file that is getting the patch applied will still have the CRs. There is a --keep-cr option that you can pass the 'git am' command
                  // but since we always want to keep CRs it is better to just set the config setting to always keep them so the user doesn't have to
-                 // remember to pass the flag. https://mseng.visualstudio.com/VSOnline/_git/GVFS/pullrequest/244891?_a=overview.
+                 // remember to pass the flag.
                 { "am.keepcr", "true" },
 
                 // Update git settings to enable optimizations in git 2.20
                 // Set 'checkout.optimizeNewBranch=true' to enable optimized 'checkout -b'
                 { "checkout.optimizenewbranch", "true" },
 
-                // Needed to prevent line ending conversion. Prevents git from auto-replacing line endings with Crlf.
+                // We don't support line ending conversions - automatic conversion of LF to Crlf by git would cause un-necessary hydration. Disabling it.
                 { "core.autocrlf", "false" },
 
-                // commit graph. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/299081
+                // Enable commit graph. https://devblogs.microsoft.com/devops/supercharging-the-git-commit-graph/
                 { "core.commitGraph", "true" },
 
                 // Perf - Git for Windows uses this to bulk-read and cache lstat data of entire directories (instead of doing lstat file by file).
                 { "core.fscache", "true" },
 
-                // Turns on all special gvfs logic. https://github.com/microsoft/scalar/blob/935c50110b31c440aae8e1cdc4ffea72bc3be01f/Scalar/CommandLine/ScalarVerb.cs#L35
+                // Turns on all special gvfs logic. https://github.com/microsoft/git/blob/be5e0bb969495c428e219091e6976b52fb33b301/gvfs.h
                 { "core.gvfs", "true" },
 
                 // Use 'multi-pack-index' builtin instead of 'midx' to match upstream implementation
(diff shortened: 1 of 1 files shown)


GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
             //
             Dictionary<string, string> requiredSettings = new Dictionary<string, string>
             {
+                 // When running 'git am' it will remove the CRs from the patch file by default. This causes the patch to fail to apply because the
+                 // file that is getting the patch applied will still have the CRs. There is a --keep-cr option that you can pass the 'git am' command
+                 // but since we always want to keep CRs it is better to just set the config setting to always keep them so the user doesn't have to
+                 // remember to pass the flag. https://mseng.visualstudio.com/VSOnline/_git/GVFS/pullrequest/244891?_a=overview.
```
kewillford: Please don't put links to non public code in the comments.  There are a bunch in this PR.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
             {
+                 // When running 'git am' it will remove the CRs from the patch file by default. This causes the patch to fail to apply because the
+                 // file that is getting the patch applied will still have the CRs. There is a --keep-cr option that you can pass the 'git am' command
+                 // but since we always want to keep CRs it is better to just set the config setting to always keep them so the user doesn't have to
+                 // remember to pass the flag. https://mseng.visualstudio.com/VSOnline/_git/GVFS/pullrequest/244891?_a=overview.
                 { "am.keepcr", "true" },
+
+                // Update git settings to enable optimizations in git 2.20
+                // Set 'checkout.optimizeNewBranch=true' to enable optimized 'checkout -b'
                 { "checkout.optimizenewbranch", "true" },
```
kewillford: I thought that something was upstreamed for this but not for all conditions so I'm not sure if we can remove this.  There were also changes when `git switch -c` was introduced which might make this obsolete if they have the optimizations there and for `checkout -b` but I would need to look into it.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+
+                // Needed to prevent line ending conversion. Prevents git from auto-replacing line endings with Crlf.
```
kewillford: Will you go a little into the "why" line ending conversions are off  here?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+                // Configure hook that git calls to ask GVFS abvout what fiels it cares about
                 { GitConfigSetting.CoreVirtualFileSystemName, Paths.ConvertPathToGitFormat(GVFSConstants.DotGit.Hooks.VirtualFileSystemPath) },
+
+                // Ensure hooks path is configured correctly.
                 { "core.hookspath", expectedHooksPath },
+
+                // Hostname is no longer sufficent for VSTS authentication. VSTS now requires dev.azure.com/account to determine the tenant.
+                // By setting useHttpPath, credential managers will get the path which contains the account as the first parameter. They can then use this information for auth appropriately.
                 { GitConfigSetting.CredentialUseHttpPath, "true" },
                 { "credential.validate", "false" },
```
wilbaker: >3 settings for which I could not find history and it is not clear we still need them - core.safecrlf, core.repositoryformatversion and credential.validate

For this one see https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/136038

From that PR's description:

> We already have logic to call git credential if we get back a 401, so there's no need to validate the PAT each time we ask for it.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+                // Use 'multi-pack-index' builtin instead of 'midx' to match upstream implementation
                 { "core.multiPackIndex", "true" },
+
+                // Perf - Enable parallel index preload for operations like git diff
                 { "core.preloadIndex", "true" },
                 { "core.safecrlf", "false" },
+
+                // Possibly cause hydration while creating untrackedCache.
                 { "core.untrackedCache", "false" },
                 { "core.repositoryformatversion", "0" },
```
wilbaker: This is the discussion from when `core.repositoryformatversion` was added (https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/235679):


Saeed Noursalehi
7/10/2017
Why are we calling 'git init'? Why not just create a new empty config file?


Saeed Noursalehi
7/10/2017
It looks like that's what you're doing down below. I don't understand what the 'git init' is for.


Forest Willard
7/10/2017
git init populates four entries that GVFS does not. I don't know much about them, but I figured this is how clone initializes the config, so it was best to keep it in line with that.

[core]
repositoryformatversion = 0
filemode = false
bare = false
logallrefupdates = true


Saeed Noursalehi
7/11/2017
It was just surprising to see "git init" called on an existing repo. What other side effects does it have besides recreating the config file?

If it's just a matter of creating those 4 config entries, I'd be fine with adding them to the list of entries in GVFSVerb, since they all seem like sensible values, and we're currently relying on them anyway so might as well make it explicit.


Forest Willard
7/11/2017
From the git man:
Running git init in an existing repository is safe. It will not overwrite things that are already there. The primary reason for rerunning git init is to pick up newly added templates (or to move the repository to another place if --separate-git-dir is given).

So really, the primary side effect would be "accidentally fixing" any missing folders in .git (refs/heads, objects, etc). but I'm not opposed to just adding those entries to simplify/clarify things here.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+                // Needed to prevent line ending conversion. Prevents git from auto-replacing line endings with Crlf.
                 { "core.autocrlf", "false" },
+
+                // commit graph. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/299081
                 { "core.commitGraph", "true" },
+
+                // Perf - Git for Windows uses this to bulk-read and cache lstat data of entire directories (instead of doing lstat file by file).
                 { "core.fscache", "true" },
+
+                // Turns on all special gvfs logic. https://github.com/microsoft/scalar/blob/935c50110b31c440aae8e1cdc4ffea72bc3be01f/Scalar/CommandLine/ScalarVerb.cs#L35
```
kewillford: I would use a link to [gvfs.h](https://github.com/microsoft/git/blob/be5e0bb969495c428e219091e6976b52fb33b301/gvfs.h) in the git code base not scalar.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+
+                // Perf - Enable parallel index preload for operations like git diff
                 { "core.preloadIndex", "true" },
                 { "core.safecrlf", "false" },
+
+                // Possibly cause hydration while creating untrackedCache.
                 { "core.untrackedCache", "false" },
                 { "core.repositoryformatversion", "0" },
+
+                // Support file modes onr Mac & Linux.
```
kewillford: typo: `on`

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+                // For consistency with git init. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/235679
                 { "core.bare", "false" },
+
+                // For consistency with git init. https://dev.azure.com/mseng/AzureDevOps/_git/GVFS/pullrequest/235679
                 { "core.logallrefupdates", "true" },
+
+                // Git to download objects on demand.
                 { GitConfigSetting.CoreVirtualizeObjectsName, "true" },
+
+                // Configure hook that git calls to ask GVFS abvout what fiels it cares about
```
kewillford: Suggested wording:
`Configure hook that git calls to get the paths git needs to consider for changes or untracked files`

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+
+                // Turns on all special gvfs logic. https://github.com/microsoft/scalar/blob/935c50110b31c440aae8e1cdc4ffea72bc3be01f/Scalar/CommandLine/ScalarVerb.cs#L35
                 { "core.gvfs", "true" },
+
+                // Use 'multi-pack-index' builtin instead of 'midx' to match upstream implementation
                 { "core.multiPackIndex", "true" },
+
+                // Perf - Enable parallel index preload for operations like git diff
                 { "core.preloadIndex", "true" },
                 { "core.safecrlf", "false" },
```
wilbaker: Details on this setting (https://git-scm.com/docs/git-config):

> If true, makes Git check if converting CRLF is reversible when end-of-line conversion is active. Git will verify if a command modifies a file in the work tree either directly or indirectly. For example, committing a file followed by checking out the same file should yield the original file in the work tree. If this is not the case for the current setting of core.autocrlf, Git will reject the file. The variable can be set to "warn", in which case Git will only warn about an irreversible conversion but continue the operation.

My guess is that because VFS4G never wants git to adjust line endings, we're explicitly setting `core.safecrlf` to false.

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: Add link to https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/master/Docs/Configuration.md#validate?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+
+                // Ensure hooks path is configured correctly.
                 { "core.hookspath", expectedHooksPath },
+
+                // Hostname is no longer sufficent for VSTS authentication. VSTS now requires dev.azure.com/account to determine the tenant.
+                // By setting useHttpPath, credential managers will get the path which contains the account as the first parameter. They can then use this information for auth appropriately.
                 { GitConfigSetting.CredentialUseHttpPath, "true" },
                 { "credential.validate", "false" },
+
+                // This setting is not needed anymore, because current version of gvfs does not use index.lock.
```
kewillford: I would explain that we need to keep this here because the repo could have been cloned when this was set to `false` so we need to explicitly set it to `true` even though it is the default value.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+                // (This change was introduced initially to prevent `git diff` from acquiring index.lock file.)
                 { "diff.autoRefreshIndex", "true" },
+
+                // Turn off of git garbage collection. Git garbage collection does not work with virtualized object. Check with Stolee - probably not needed.
                 { "gc.auto", "0" },
```
kewillford: We do run maintenance jobs now that do the packing of loose objects so in theory we shouldn't need this but I don't think it is hurting anything and it will prevent a `gc` from getting kicked off if for some reason the maintenance jobs have not been running and there are too many loose objects.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
+
+                // Prevent git GUI from displaying a GC warning. https://mseng.visualstudio.com/AzureDevOps/_git/GVFS/pullrequest/184856?_a=overview
                 { "gui.gcwarning", "false" },
+
+                // Update git settings to enable optimizations in git 2.20
+                // Set 'index.threads=true' to enable multi-threaded index reads
```
kewillford: Do we have any performance numbers how how much this helped reading the index?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
                 { "index.threads", "true" },
+
+                // gvfs index parser would work only with index versions 4 and above.
```
kewillford: Suggested wording:
`index parsing code in VFSForGit currently only supports version 4`.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
                 { "index.version", "4" },
+
+                // Turned off for performance reasons.
```
kewillford: Can you explain a little of the "why" here?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
                 { "merge.stat", "false" },
+
+                // Update git version to 2.18.0.gvfs.1.26.gf61ade4 - This updates the git version to 2.18.0 which includes changes to the rename detection to use
+                // config setting which was previously turned off with a patch in git.
                 { "merge.renames", "false" },
+
+                // Don't use bitmaps to determine pack file contents, b'coz we use MIDX for this.
```
kewillford: `because`?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
                 { "index.threads", "true" },
+
+                // gvfs index parser would work only with index versions 4 and above.
                 { "index.version", "4" },
+
+                // Turned off for performance reasons.
                 { "merge.stat", "false" },
+
+                // Update git version to 2.18.0.gvfs.1.26.gf61ade4 - This updates the git version to 2.18.0 which includes changes to the rename detection to use
+                // config setting which was previously turned off with a patch in git.
```
kewillford: Again why?  Performance? Will download objects when trying to find renamed files right?

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
                 { "pack.useSparse", "true" },
+
+                // Stop automatic git GC
                 { "receive.autogc", "false" },
+
+                // Update git settings to enable optimizations in git 2.20
+                // Set 'reset.quiet=true' to speed up 'git reset <foo>"
                 { "reset.quiet", "true" },
+
+                // Configure git to use our serialize status file
```
kewillford: for performance right?

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: This is to match what `git init` would do.

GVFS/GVFS/CommandLine/GVFSVerb.cs
kewillford: Since we don't support line ending conversions there is no reason to validate line endings either.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -124,39 +124,112 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
...
                 { "status.deserializePath", gitStatusCachePath },
+
+                // In Git 2.24.0, some new config settings were created. Disable them locally in VFS for Git repos in case a user has set them globally.
+                // https://github.com/microsoft/VFSForGit/pull/1594
```
kewillford: I would only have this comment once and have the 3 config settings it applies to follow it and have the comment be something like `The next 3 config settings were added in git 2.24.0. Disable ...`

GVFS/GVFS/CommandLine/GVFSVerb.cs
alameenshah: updated the comment to include this

GVFS/GVFS/CommandLine/GVFSVerb.cs
alameenshah: Added - `This is to match what git init does.`

GVFS/GVFS/CommandLine/GVFSVerb.cs
alameenshah: [updated](https://github.com/microsoft/VFSForGit/pull/1614/commits/a39ea1d2cf052db520e9440893b98ae3948ee1a6)

PR #1615, opened 2019-12-10
[master] PostFetchStep: do not immediately delete expired commit-graph files

derrickstolee:
We have noticed a lot of "cannot find all commit-graph files" errors,
more than is reasonable for simply hard-drive failures. Likely there is
an issue with how Git is comparing the paths of the commit-graph files
it is queueing for deletion and the ones it plans to keep, related to
how we use the --object-dir parameter.

The commit-graph code in Git will "touch" all the files that are
currently in the commit-graph-chain, so as long as we pass a non-zero
expire time, that will prevent us from deleting the files that are
still in the chain despite the path difference.

This is a faster fix than fixing Git first.

Cc: @jrbriggs, @wilbaker this may be worth taking into `milestones/162`.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -538,7 +538,9 @@ namespace GVFS.Common.Git
         /// </summary>
         public Result WriteCommitGraph(string objectDir, List<string> packs)
         {
-            string command = "commit-graph write --stdin-packs --split --size-multiple=4 --object-dir \"" + objectDir + "\"";
+            // Do not expire commit-graph files that have been modified in the last hour.
+            // This will prevent deleting any commit-graph files that are currently in the commit-graph-chain.
+            string command = $"commit-graph write --stdin-packs --split --size-multiple=4 --expire-time={1 * 60 * 60} --object-dir \"{objectDir}\"";
             return this.InvokeGitInWorkingDirectoryRoot(
                 command,
                 useReadObjectHook: true,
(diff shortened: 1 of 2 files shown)


derrickstolee:
And I just noticed the unit tests will break with my first commit. will update.

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -538,7 +538,9 @@ public Result PackObjects(string filenamePrefix, string gitObjectsDirectory, Act
         /// </summary>
         public Result WriteCommitGraph(string objectDir, List<string> packs)
         {
-            string command = "commit-graph write --stdin-packs --split --size-multiple=4 --object-dir \"" + objectDir + "\"";
+            // Do not expire commit-graph files that have been modified in the last hour.
+            // This will prevent deleting any commit-graph files that are currently in the commit-graph-chain.
+            string command = $"commit-graph write --stdin-packs --split --size-multiple=4 --expire-time={1 * 60 * 60} --object-dir \"{objectDir}\"";
```
kewillford: So if I understand correctly all the current commit-graph-chain files will be "touched" with this `commit-graph write` command? and setting the `--expire-time` will make it not consider those files for delete?  Is there any significance to the 1 hour?  Is that because the files could have been touched but the code to check the expire runs much later and a bunch of time could have passed?

derrickstolee: No significance to the one hour, except that it is definitely longer than the `git commit-graph write` command. The logic that expires first touches all the files in the chain then the next step is to compare the paths in the `objects/info/commit-graphs` dir to those in the `commit-graph-chain` file. Probably one second would be enough, but let's set the expire time to be relative to how often we expect the maintenance job to run (we prefetch about every 75 minutes, triggering a post-fetch job.)

PR #1621, opened 2019-12-13
[Linux] functional test backing root and DotGVFSRoot fixes

chrisd8088:
This PR includes two minor edits to enable the functional test suite on Linux: first, we replace one use of a `.gvfs` string constant in the functional tests which was added in 9fc6bc1, in order to align with the changes made in 8348777.  This allows the `CloneTests` to pass on Linux where `.vfsforgit` is used instead of `.gvfs`.

And second, the `MountSetsCoreHooksPath()` functional test performs an unmount step before resetting Git's configurations, and while this actually does succeed on Linux as it stands, the `src/` `RepoRoot` directory is entirely empty while its unmounted, so to be consistent, use the backing root instead.


chrisd8088:
Hi @derrickstolee!  I thought I'd ping you on this one, but let me know if it should go to someone else.  I hope this is sufficiently minor fixup that it's a fast review.

Hope all is well on your end!

chrisd8088:
Hi @jrbriggs also -- just adding you to this in case you're able to take a quick look; I hope this one is pretty trivial.

chrisd8088:
Thanks @kewillford!  I'll merge those in shortly!  Much appreciated, and happy holidays!

(Now I just need to figure out the one remaining inconsistent sparse test on Linux.  Hmm.  :-)

PR #1622, opened 2019-12-18
GVFSVerb: use flags for core.gvfs

derrickstolee:
The core.gvfs config option can take a boolen "true" to turn on
all options, but also an integer to show a bitmask of important
flags. For the first time, we have a new flag that we may not
want to enable right away: GVFS_PREFETCH_DURING_FETCH.

This flag will make the `git fetch` builtin run a `git gvfs-helper prefetch`
command. Until that hardens a bit, we do not want VFS for Git users
to use that code path.

Update Git to include this option.

### DIFF
--- a/GVFS/GVFS.Build/GVFS.props
+++ b/GVFS/GVFS.Build/GVFS.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup Label="Parameters">
     <GVFSVersion>0.2.173.2</GVFSVersion>
-    <GitPackageVersion>2.20191211.2</GitPackageVersion>
+    <GitPackageVersion>2.20191217.3</GitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="DefaultSettings">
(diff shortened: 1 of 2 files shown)


jeffhostetler:
@derrickstolee  @jrbriggs   I should have asked this question yesterday, but when I added gvfs-helper, I added explicit config settings `gvfs.cache-server` and `gvfs.sharedcache` and `core.usegvfshelper`.  Would it be better to set/unset this prefetch-during-fetch flag via a similar `core.*` or `gvfs.*` setting -- rather than setting it via another bit in `core.gvfs` ???

derrickstolee:
> @derrickstolee @jrbriggs I should have asked this question yesterday, but when I added gvfs-helper, I added explicit config settings `gvfs.cache-server` and `gvfs.sharedcache` and `core.usegvfshelper`. Would it be better to set/unset this prefetch-during-fetch flag via a similar `core.*` or `gvfs.*` setting -- rather than setting it via another bit in `core.gvfs` ???

I think there is value in making the prefetch-in-fetch work in VFS for Git eventually, as well as dropping the read-object hook in favor of your gvfs-helper.

The only thing I can thing is that `core.usegvfshelper` could become a flag in `core.gvfs`. Your `gvfs.cache-server` and `gvfs.sharedcache` don't make sense as they are string-valued.

PR #1635, opened 2020-02-19
sparse: allow dehydrate to continue if a directory delete fails

kewillford:
When the code gets to the point of deleting the directory it is empty
and it will not hurt anything leaving it on disk.  This also more
closely matches what git does when changing the sparse-checkout since
it ignores errors when trying to delete empty directories.

Todo:
- [x] Write a dehydrate functional test that will cause an error on the directory delete if possible.

### DIFF
--- a/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
+++ b/GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
@@ -13,7 +13,7 @@ namespace GVFS.Common.FileSystem
     {
         public const int DefaultStreamBufferSize = 8192;
 
-        public virtual void DeleteDirectory(string path, bool recursive = true)
+        public virtual void DeleteDirectory(string path, bool recursive = true, bool ignoreDirectoryDeleteExceptions = false)
         {
             if (!Directory.Exists(path))
             {
@@ -32,11 +32,21 @@ namespace GVFS.Common.FileSystem
 
                 foreach (DirectoryInfo subDirectory in directory.GetDirectories())
                 {
-                    this.DeleteDirectory(subDirectory.FullName);
+                    this.DeleteDirectory(subDirectory.FullName, recursive, ignoreDirectoryDeleteExceptions);
                 }
             }
 
-            directory.Delete();
+            try
+            {
+                directory.Delete();
+            }
+            catch (Exception)
+            {
+                if (!ignoreDirectoryDeleteExceptions)
+                {
+                    throw;
+                }
+            }
         }
 
         public virtual void CopyDirectoryRecursive(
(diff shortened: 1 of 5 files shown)


GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -36,7 +36,17 @@ public virtual void DeleteDirectory(string path, bool recursive = true)
                 }
             }
 
-            directory.Delete();
+            try
+            {
+                directory.Delete();
+            }
+            catch (Exception)
```
derrickstolee: Can we catch something more specific, like `IOException`?

kewillford: I thought about that but in the [git code](https://github.com/microsoft/git/blob/a925f8273bfa007717dbfe392a9b006fe89e51fd/symlinks.c#L281), they aren't checking for any specific type of error.  They simply leave the directory there.

The [other possible exceptions](https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.delete?view=netframework-4.8) `UnauthorizedAccessException` and `SecurityException` seem like we would want to keep going on these as well since this should be an empty folder and it's not going to hurt anything if it is left out there, but if we do fail dehydrate then there could be files that will show as deleted the next `git status` since we don't remove the directory from the modified paths and this is the issue we are trying to avoid.

derrickstolee: Those extra reception types make sense. Thanks!

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/DehydrateTests.cs
```diff
@@ -268,6 +270,24 @@ public void FolderDehydrateFolderThatWasDeleted()
...
+            TestPath folderToLock = new TestPath(this.Enlistment, Path.Combine(folderToDehydrate, "GVFS"));
+            using (SafeFileHandle handle = this.OpenFolderHandle(folderToLock.VirtualPath))
+            {
+                this.DehydrateShouldSucceed(new[] { $"{folderToDehydrate} {FolderDehydrateSuccessfulMessage}" }, confirm: true, noStatus: false, foldersToDehydrate: folderToDehydrate);
+            }
+
+            this.Enlistment.UnmountGVFS();
+
+            folderToLock.BackingPath.ShouldBeADirectory(this.fileSystem).WithNoItems();
+            folderPathDehydrated.BackingPath.ShouldBeADirectory(this.fileSystem).WithOneItem().Name.ShouldEqual("GVFS");
```
jrbriggs: nit:
```suggestion
            folderPathDehydrated.BackingPath.ShouldBeADirectory(this.fileSystem).WithOneItem().Name.ShouldEqual(folderToDehydrate);

```

PR #1642, opened 2020-03-09
Update Git to v2.26.0

derrickstolee:
See microsoft/git#251 for details.

This required an update to our test infrastructure. The rebase merge backend changed in a way that it would now open an editor during `git rebase --continue`, causing a test to wait for `vim` to close. Set the editor to be a no-op. This also changes the output to include the commit oid, so use `GIT_COMMITTER_TIME` to be a constant to keep the commits the same.


derrickstolee:
> Looks good. I don't see the change to "Ignore whitespace lines when checking output." Did we not need that anymore? Remove it from the PR description if that is the case.

You're right. That was a Scalar-only change. Thanks!

PR #1647, opened 2020-04-01
Fix commit-graph expiration

derrickstolee:
Wow, this was really not working as expected.

See microsoft/git#255 for how broken the `--expire-time` argument was.

Fix this by using the fixed argument and passing a datetime instead of an offset by seconds. This will provide a longer window for old commit-graph files, but apparently we've been leaving turd files around for a long time without anyone noticing.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -16,6 +16,7 @@ namespace GVFS.Common.Git
 
         private static readonly Encoding UTF8NoBOM = new UTF8Encoding(false);
         private static bool failedToSetEncoding = false;
+        private static string expireTimeDateString;
 
         /// <summary>
         /// Lock taken for duration of running executingProcess.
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -81,6 +82,19 @@ public GitProcess(string gitBinPath, string workingDirectoryRoot)
             }
         }
 
+        public static string ExpireTimeDateString
+        {
+            get
+            {
+                if (expireTimeDateString == null)
+                {
+                    expireTimeDateString = DateTime.Now.Subtract(TimeSpan.FromDays(1)).ToShortDateString();
```
kewillford, line 91: Before we chose to expire after 1 hour and now we are after 1 day.  I know that this was never working before and  I don't think that will be an issue because I don't think it should be creating that many commit graphs but wanted to get your thoughts.  Thanks!

derrickstolee: The important thing is that it doesn't expire things right away, and there are some concerns about timezones and things when it parses the expiry date. So either we get _really specific_ about the time (including timezone offset) or we can just go "at least 24 hours ago, give or take timezone shifts of +/-12 hours".

Since we've been running with the lack of expirations for this long, having a slightly longer expire window will not be a problem.

PR #1648, opened 2020-04-08
[WIP] Track folder changes to update USN journal

kewillford:
BuildXL was having issues detecting that a folder had changes when the projection changed because VFSForGit was simply updating its in memory projection and not doing anything with the file system that would trigger and update to the USN. It would hand back the new items in the directory when projfs would ask for the items for enumeration the next time an enumeration happened.

This change is to make the folder get a new USN when VFSForGit determines that a folder has changed.  This is done by keeping a SHA1 hash of all the SHAs of the files that are in the projection and when the projection changes calculate the new SHA for the folder and compare with what it had before.  If it changed then write and delete a file in the folder to trigger the USN update.

Something to note. We don't need any upgrade step because we are already storing the SHA for the folder in the placeholder database but it is currently `null` for all folders so the first time there is a projection change it will cause all the folders to get the calculated SHA for the folder and update the placeholder database. 

Todos:
- [x] Error handling
- [x] Functional tests
- [x] Check Memory and Performance
- [x] Give version to BuildXL to test with

### DIFF
--- a/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/WindowsFolderUsnUpdate.cs
+++ b/GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/WindowsFolderUsnUpdate.cs
@@ -1,0 +1,93 @@
+﻿using GVFS.FunctionalTests.FileSystemRunners;
+using GVFS.FunctionalTests.Tests.EnlistmentPerFixture;
+using GVFS.FunctionalTests.Tools;
+using GVFS.Tests.Should;
+using NUnit.Framework;
+using System;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace GVFS.FunctionalTests.Windows.Windows.Tests
+{
+    [TestFixture]
+    [Category(Categories.WindowsOnly)]
+    [Category(Categories.GitCommands)]
+    public class WindowsFolderUsnUpdate : TestsWithEnlistmentPerFixture
+    {
+        private const string StartingCommit = "ad87b3877c8fa6bebbe62330354f5c535875c4dd";
+        private const string CommitWithChanges = "e6d047cf65f4a384568b7a451530e18410bc8a12";
+        private FileSystemRunner fileSystem;
+
+        public WindowsFolderUsnUpdate()
+        {
+            this.fileSystem = new SystemIORunner();
+        }
+
+        [TestCase]
+        public void CheckoutUpdatesFolderUsnJournal()
+        {
+            this.GitCheckoutCommitId(StartingCommit);
+            this.GitStatusShouldBeClean(StartingCommit);
+
+            FolderPathUsn[] pathsToCheck = new FolderPathUsn[]
+            {
+                new FolderPathUsn(Path.Combine(this.Enlistment.RepoRoot, "Test_ConflictTests", "AddedFiles"), this.fileSystem),
+                new FolderPathUsn(Path.Combine(this.Enlistment.RepoRoot, "Test_ConflictTests", "DeletedFiles"), this.fileSystem),
+                new FolderPathUsn(Path.Combine(this.Enlistment.RepoRoot, "Test_ConflictTests", "ModifiedFiles"), this.fileSystem),
+            };
+
+            this.GitCheckoutCommitId(CommitWithChanges);
+            this.GitStatusShouldBeClean(CommitWithChanges);
+
+            foreach (FolderPathUsn folderPath in pathsToCheck)
+            {
+                folderPath.ValidateUsnChange();
+            }
+        }
+
+        private static string UsnFolderId(string path)
+        {
+            ProcessResult result = ProcessHelper.Run("fsutil", $"usn readdata \"{path}\"");
+            Match match = Regex.Match(result.Output, @"^Usn\s+:\s(\w+)", RegexOptions.Multiline);
+            if (match.Success)
+            {
+                return match.Value;
+            }
+
+            return string.Empty;
+        }
+
+        private void GitStatusShouldBeClean(string commitId)
+        {
+            GitHelpers.CheckGitCommandAgainstGVFSRepo(
+                this.Enlistment.RepoRoot,
+                "status",
+                "HEAD detached at " + commitId,
+                "nothing to commit, working tree clean");
+        }
+
+        private void GitCheckoutCommitId(string commitId)
+        {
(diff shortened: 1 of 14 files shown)


GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1304,6 +1310,21 @@ private void UpdatePlaceholders()
 
                     if (keepFolder)
                     {
+                        if (this.TryGetFolderDataFromTreeUsingPath(folderPlaceholder.Path, out FolderData folderData))
```
jrbriggs, line 1313: What do you think about guarding this section with a config setting? Of the two big internal monorepos, one needs this for correctness, the other does not. I worry that the cost here might be pretty large for no real gain if the dev doesn't need BuildXL. Thoughts?

kewillford: I am going to first see what performance on the windows repo looks like and how much this is going to add but this was definitely on my mind as I was writing this code.

kewillford:
I didn't find any change in the memory footprint with this version.

Differences in performance:
|Command|Folder Placeholders|Time before|Time After|
|---|---:|---:|---:|
|Repo 1 `checkout`|~100K|12s|13s|
|Repo 2 `checkout`|~150K|8.3s|9.5s|

So there is ~10% perf regression in `checkout` when there a greater than ~100K folder placeholders.  With fewer than 10K placeholders there was not a significant increase in the timings.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.FolderData.cs
```diff
@@ -49,6 +51,26 @@ public void Include()
                 }
             }
 
+            public string HashedFileShas()
```
kewillford, line 54: Rename this method since it will be file shas and folders that will be hashed.

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.FolderData.cs
```diff
@@ -49,6 +51,26 @@ public void Include()
...
+                using (HashAlgorithm hash = SHA1.Create())
+                {
+                    for (int i = 0; i < this.ChildEntries.Count; i++)
+                    {
+                        if (!this.ChildEntries[i].IsFolder)
+                        {
+                            FileData fileData = (FileData)this.ChildEntries[i];
+                            fileData.Sha.ToBuffer(sha);
+                            hash.TransformBlock(sha, 0, 20, null, 0);
+                        }
```
kewillford, line 66: This needs to be updated to also track any folders so it needs an else statement here.

Since the folder might not have a sha for it yet and we don't want to recurse through the folders to generate the sha, the logic should be to use the sha of the folder if there is one and if there isn't one use the name of the folder since that will be unique for the folder.

benwitman:
We don’t want to hash the shas, we only want to hash the names.  That should be changed for the file case too

kewillford:
> We don’t want to hash the shas, we only want to hash the names. That should be changed for the file case too

If there is a pure virtual file for the folder and the file's name doesn't change but the content does, the USN will not be updated for the folder.  Will that be okay because all the files of a folder will at least have a placeholder on disk?

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.cs
```diff
@@ -1304,6 +1311,25 @@ private void UpdatePlaceholders()
 
                     if (keepFolder)
                     {
+                        if (this.TryGetFolderDataFromTreeUsingPath(folderPlaceholder.Path, out FolderData folderData))
+                        {
+                            string newFolderSha = folderData.HashedFileShas();
```
narasamdya, line 1316: >HashedFileShas [](start = 61, length = 14)

Is this the hash of hashes of contents of files in that folder?

If yes, this means `.vsf_usn_folder` will be created if the content of a file in that folder changes. From BuildXL's point of view, this will force BuildXL to perform extra validation.

For example, suppose that a task only enumerates a folder, but does not read any file within that folder. BuildXL will track the fingerprint of that folder. The fingerprint is the hash of file names (along with some attributes -- i.e., if it's a file or a directory). If the content of any file in that folder changes, but there's no membership change, then BuildXL can safely assume that the fingerprint remains the same, and concludes that the task is still up-to-date. However, the introduction of `.vfs_usn_folder` triggers a membership change (due to creation and deletion). This forces BuildXL to re-validate if the new fingerprint is still the same.

Perhaps, instead of HashedFileShas, GVFS4Git should track the same form of fingerprint as BuildXL.

narasamdya:
I mentioned in the other comment that having the shas will not cause overbuild (I'll create a unit test on BuildXL's side). But it forces BuildXL to perform extra validation, that could've been avoided.

---
In reply to: [629356194](https://github.com/microsoft/VFSForGit/pull/1648#issuecomment-629356194) [](ancestors = 629356194)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.FolderData.cs
benwitman: We want to just hash the **names** of the files/folders in the directory and not the content (so it wouldn't need to recurse)

GVFS/GVFS.Virtualization/Projection/GitIndexProjection.FolderData.cs
benwitman: perhaps `HashedChildrenNamesSha()`

GVFS/GVFS.FunctionalTests.Windows/Windows/Tests/WindowsFolderUsnUpdate.cs
```diff
@@ -0,0 +1,94 @@
...
+using System;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace GVFS.FunctionalTests.Windows.Windows.Tests
+{
+    [TestFixture]
+    [Category(Categories.WindowsOnly)]
+    [Category(Categories.GitCommands)]
+    [Ignore("fsutil requires WSL be enabled.  Need to find a way to enable for builds or a different way to get the USN for the folder.")]
```
derrickstolee, line 15: making a note to self to investigate this.

derrickstolee:
Closing since this was replaced with #1664

PR #1653, opened 2020-04-14
Update Git to include v2.26.1

derrickstolee:
Re-pushed to include the correct version number after pushing the `v2.26.1.vfs.1.1` tag.


derrickstolee:
> How did the perf look for this version?

I haven't tested the perf because the only change should be the security fix.

PR #1658, opened 2020-04-17
ProjFSFilter: Be more robust to missing PowerShell

derrickstolee:
A user had issues and we started a support line. Their diagnose log
had _thousands_ of service logs. These were all failing with this call
stack pointing to the method changed here:

```
System.ComponentModel.Win32Exception (0x80004005): The system cannot find the file specified
 at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at GVFS.Common.ProcessHelper.StartProcess(Process executingProcess)
    at GVFS.Common.ProcessHelper.Run(ProcessStartInfo processInfo, String errorMsgDelimeter, Object executionLock)
     at GVFS.Platform.Windows.ProjFSFilter.IsGVFSUpgradeSupported() in E:\\A\\_work\\822\\s\\GVFS\\GVFS.Platform.Windows\\ProjFSFilter.cs:line 371 
       at GVFS.Upgrader.InstallerPreRunChecker.IsGVFSUpgradeAllowed(String& consoleError) 
        at GVFS.Upgrader.InstallerPreRunChecker.TryRunPreUpgradeChecks(String& consoleError) 
          at GVFS.Service.ProductUpgradeTimer.TimerCallback(Object unusedState) in E:\\A\\_work\\822\\s\\GVFS\\GVFS.Service\\ProductUpgradeTimer.cs:line 168
```

The only explanation I can think of is that PowerShell is not
available on their system!

### DIFF
--- a/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
+++ b/GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
@@ -679,9 +679,16 @@ namespace GVFS.Platform.Windows
 
         private static ProcessResult GetProjFSOptionalFeatureStatus()
         {
-            return CallPowershellCommand(
-                "$var=(Get-WindowsOptionalFeature -Online -FeatureName " + OptionalFeatureName + ");  if($var -eq $null){exit " +
-                (int)ProjFSInboxStatus.NotInbox + "}else{if($var.State -eq 'Enabled'){exit " + (int)ProjFSInboxStatus.Enabled + "}else{exit " + (int)ProjFSInboxStatus.Disabled + "}}");
+            try
+            {
+                return CallPowershellCommand(
+                    "$var=(Get-WindowsOptionalFeature -Online -FeatureName " + OptionalFeatureName + ");  if($var -eq $null){exit " +
+                    (int)ProjFSInboxStatus.NotInbox + "}else{if($var.State -eq 'Enabled'){exit " + (int)ProjFSInboxStatus.Enabled + "}else{exit " + (int)ProjFSInboxStatus.Disabled + "}}");
+            }
+            catch (Exception)
+            {
+                return new ProcessResult(string.Empty, "Failed. Possible that powershell does not exist on system", 1);
+            }
         }
 
         private static EventMetadata CreateEventMetadata(Exception e = null)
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Platform.Windows/ProjFSFilter.cs
```diff
@@ -679,9 +679,16 @@ private static bool TryEnableProjFSOptionalFeature(ITracer tracer, PhysicalFileS
...
+            try
+            {
+                return CallPowershellCommand(
+                    "$var=(Get-WindowsOptionalFeature -Online -FeatureName " + OptionalFeatureName + ");  if($var -eq $null){exit " +
+                    (int)ProjFSInboxStatus.NotInbox + "}else{if($var.State -eq 'Enabled'){exit " + (int)ProjFSInboxStatus.Enabled + "}else{exit " + (int)ProjFSInboxStatus.Disabled + "}}");
+            }
+            catch (Exception)
+            {
+                return new ProcessResult(string.Empty, "Failed. Possible that powershell does not exist on system", 1);
+            }
```
kewillford, line 691: Some questions to think about.
1. Should this be moved down a level into the `CallPowershellCommand` method so that anything else that calls it will catch it?
2. Should we be using `where` to try and locate powershell and use the full path or error out if not found?
3. Should we fallback on using `dism /online /get-featureinfo /featurename:Client-ProjFS` and trying to parse its output for `State : Enabled`?

derrickstolee: > Some questions to think about.
> 
> 1. Should this be moved down a level into the `CallPowershellCommand` method so that anything else that calls it will catch it? 

Yes. That would be better. Should have known better.

> 2. Should we be using `where` to try and locate powershell and use the full path or error out if not found?

Good idea. I haven't heard back from the user to see if they just don't have it at all, or if their PATH is messed up.

> 3. Should we fallback on using `dism /online /get-featureinfo /featurename:Client-ProjFS` and trying to parse its output for `State : Enabled`?

Good thing the service runs elevated or else `dism` wouldn't work. Worth a shot.

derrickstolee: Hm. The fallback is probably more complicated than we want, since the `ProcessResult` is taken and its exit code is converted into an `ProjFSInboxStatus` enum.

We should focus only on failing gracefully when not finding powershell. Using `where` to find it is a really valuable suggestion.

PR #1661, opened 2020-05-04
Update Git to include en/sparse-checkout

derrickstolee:
See microsoft/git#267.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/ResetMixedTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/ResetMixedTests.cs
@@ -79,22 +79,6 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         }
 
         [TestCase]
-        public void ResetMixedOnlyAddedThenCheckoutWithConflicts()
-        {
-            // The warning messages changed when using skip-worktree bits
-            // versus not using them. We cannot validate the output of these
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.FunctionalTests/Tests/GitCommands/ResetMixedTests.cs
```diff
@@ -78,19 +78,6 @@ public void ResetMixedThenCheckoutWithConflicts()
             this.FilesShouldMatchCheckoutOfTargetBranch();
         }
 
-        [TestCase]
-        public void ResetMixedOnlyAddedThenCheckoutWithConflicts()
```
jeffhostetler, line 82: I'm confused.  This test case was modified in the second commit to change `Validate...` to `Run...`, but then the whole test was deleted in the third commit.  Not sure if this was intended or what.

derrickstolee: Sorry for the commit confusion. I could not get this test to work the same because of how the (intended?) behavior changed for sparse paths.

derrickstolee: While rebasing onto v2.27.0-rc0, I think I found a reason why this test would fail. I'll close this PR and work on the rebased version instead.

PR #1672, opened 2020-06-02
ProductUpgrader: be extra careful about a copy loop

derrickstolee:
The `ProductUpgrader` makes a backup of the `C:\Program Files\GVFS` folder before running the install, as a mechanism for backing out a failed upgrade. However, it copies that data into `C:\Program Files\GVFS\ProgramData\GVFS.Upgrade\Tools`, which can create a copy loop! This affected an actual user.

I'm not sure why this affects that one user but not another. In addition to adding more defensive programming (adding the target of the copy to the list of exclusions), I also added some extra tracing so we can see how these strings differ on this machine.

### DIFF
--- a/GVFS/GVFS.Common/ProductUpgrader.cs
+++ b/GVFS/GVFS.Common/ProductUpgrader.cs
@@ -183,7 +183,7 @@ namespace GVFS.Common
                 // directory causes a cycle(at some point we start copying C:\Program Files\GVFS\ProgramData\GVFS.Upgrade
                 // and its contents into C:\Program Files\GVFS\ProgramData\GVFS.Upgrade\Tools). The exclusion below is
                 // added to avoid this loop.
-               HashSet<string> directoriesToExclude = new HashSet<string>();
+               HashSet<string> directoriesToExclude = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
                 string secureDataRoot = GVFSPlatform.Instance.GetSecureDataRootForGVFS();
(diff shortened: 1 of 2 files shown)


derrickstolee:
The log message said this on the user's machine:

```
Copying contents of 'C:\\Program Files\\GVFS' to 'C:\\Program Files\\GVFS\\ProgramData\\GVFS.Upgrade\\Tools',excluding 'C:\\Program Files\\GVFS\\ProgramData\\GVFS.Upgrade\\Tools' and 'C:\\Program Files\\GVFS\\ProgramData'
```

So, _not_ a case issue. I have no idea why this works for most users.

However, this situation did fix the issue for the user with the problem!

derrickstolee:
> The log message said this on the user's machine:
> 
> ```
> Copying contents of 'C:\\Program Files\\GVFS' to 'C:\\Program Files\\GVFS\\ProgramData\\GVFS.Upgrade\\Tools',excluding 'C:\\Program Files\\GVFS\\ProgramData\\GVFS.Upgrade\\Tools' and 'C:\\Program Files\\GVFS\\ProgramData'
> ```
> 
> So, _not_ a case issue. I have no idea why this works for most users.

Because **IT DOESN'T WORK**. This means `gvfs upgrade` is completely broken for 1.0.20112.1 and later. :(

derrickstolee:
I'm going to add documentation into the `troubleshooting.md` doc before I merge this into `master`, but I'll wait for a new version number before pushing those changes. The HOTFIX is merged so I am moving forward with the release process.

PR #1683, opened 2020-07-22
GVFS.SignFiles.csproj: include more DLLs for signing

derrickstolee:
It appears that during our build, we send some DLLs for signing, but the
versions we send have already been copied to other projects. Thus, we
ship versions of those DLLs that are unsigned.

Add extra versions of the DLLs that are currently being signed, except
include them in the GVFS.Windows project that is likely where they are
being copied as part of the build.


derrickstolee:
This PR used to target `releases/shipped` so I could easily test the signing build. Everything is working as intended:

image

PR #1689, opened 2020-08-10
Make delete docs more verbose.

kyle-rader:
## Potential issues
* Some folks reading the delete instructions might not know the best way to clean up after un-mounting. 
* They might expect un-mounting to also do this when it doesn't. 

Added an explicit step for removing and added warning not to try deleting before un-mounting.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -4,6 +4,12 @@ Troubleshooting
 Deleting a VFS for Git repo
 ---------------------------
 
+You must follow these steps to delete a VFS for Git repository.
+
+If you have attempted deletion before un-mounting jump to 
+[Recovering from an attempt to delete without un-mounting](#Recovering-from-an-attempt-to-delete-without-un-mounting).
+
+
 1. Un-mount the repo.
 
     Since a VFS for Git clone has a running `GVFS.Mount` process to track the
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -4,15 +4,32 @@ Troubleshooting
...
+
+    ```
+    C:\Users\you\big_repo\src\> cd ..\..
+    C:\Users\you\> gvfs unmount big_repo
+    ```
+
+    If you have deleted the enlistment or its `.gvfs` folder, then you will
+    likely see alerts saying "Failed to auto-mount at path `X`". To manually remove
+    this repo from the auto-mount feature, remove the appropriate line
+    from the `C:\ProgramData\GVFS\GVFS.Service\repo-registry` file.
```
derrickstolee, line 24: If you manually unmount the repo, then the repo should be removed from the repo-registry. Could this warning be placed outside of the bulleted list, to add that help only for users who tried deleting without following these instructions?

kyle-rader: Ya that makes sense, I thought this was sort of the existing "warning" that you shouldn't just delete it. We can make the an explicit warning at the top. If someone did try to delete it prematurely are the resolution steps to 
1. remove the auto-mount entry
2. finish actually deleting?

derrickstolee: Yes, that should work. It is possible that `GVFS.Mount.exe` is still running on their enlistment, preventing the `src` directory from being deleted. They need to go to Task Manager and find the `GVFS.Mount.exe` process with that enlistment in the "Command Line" column:

image

kyle-rader: Aha, I'll include this. Thanks!

kyle-rader:
@derrickstolee seems one of the CI was cancelled for a timeout maybe? Or the agent machine was restarting? I don't think I have perms to request it re-run but feel free to force merge this if you are comfortable, otherwise, somehow re-queue the CI?

PR #1702, opened 2020-10-21
[HOTFIX] Convert more projects to reduce build failures

derrickstolee:
While trying to get out the release, the builds were failing due to a different project order pulling different versions of NuGet packages, it seemed. We were getting some that were for .NET Framework 4.6+ instead of those compatible with .NET Standard 2.0. To get things working, I started converting projects away from dual-targeting .NET Core _and_ removing some test projects that were .NET Core and could be framework projects instead. This required renaming some test files.


derrickstolee:
Closing while I reevaluate the plan.

derrickstolee:
Re-opening to test an earlier commit.

PR #1703, opened 2020-10-22
Delete all macOS projects

derrickstolee:
Some macOS projects were causing failures in the release process, so
let's finally get around to deleting them.

Replaces #1700.


jeffhostetler (approved):

JunielKatarn:
Does this mean the macOS port is dead?

derrickstolee:
> Does this mean the macOS port is dead?

Yes, the macOS port was abandoned in August 2019 when Apple deprecated kernel extensions, which was how the macOS port was going to work. None of the options were satisfactory replacements, and we pivoted to creating and supporting [Scalar](https://github.com/microsoft/git/blob/vfs-2.37.2/contrib/scalar/docs/index.md) instead.

PR #1704, opened 2020-10-22
[HOTFIX] Convert projects away from .NET Core to .NET Standard

derrickstolee:
This required several updates to the build system:

* The files from `GVFS.UnitTests` are absorbed directly into `GVFS.UnitTests.Windows`.
* The `GVFS.PlatformLoader` project is removed (it was getting double-imported) and instead we register `WindowsPlatform` directly.
* A few more macOS-specific things are removed, including a `Marshal` method that doesn't exist in .NET Standard 2.0 (but we only needed it on Unix-based systems).


derrickstolee:
This may be a dead-end. Closing for now.

Issue #1708, opened 2020-12-10
Allow "dismount" to be used as a command line option (alias for "unmount")


GabeDeBacker:
I've been using power-shell to mount and dismount VHDs with the appropriately named cmdlet's "mount-vhd" and "dismount-vhd".

Every time I try to dismount a mounted GVFS repo, I type "gvfs dismount" because, well, I speak English.

Can we please PLEASE add gvfs dismount?

https://www.merriam-webster.com/dictionary/dismount

https://www.merriam-webster.com/dictionary/unmount

ranavale:
Please fix this ASAP

GabeDeBacker:
Bonus points if we fix change "unmount" to "dismount" in the output strings..

I am cloning the repo and taking a look...

dscho:
See https://github.com/microsoft/VFSForGit/pull/1709#issuecomment-744505629 for why this PR should not be accepted.

PR #1709, opened 2020-12-10
Add dismount command line option...

GabeDeBacker:
This addresses issue #1708 by simply adding a dismount verb that derives from "unmount" to add "dismount" to the command line options thereby making the command line options more friendly to use for those who prefer the English correct "dismount" to "unmount".

### DIFF
--- a/GVFS/GVFS/CommandLine/DismountVerb.cs
+++ b/GVFS/GVFS/CommandLine/DismountVerb.cs
@@ -1,0 +1,10 @@
+﻿using CommandLine;
+
+namespace GVFS.CommandLine
+{
+    [Verb(DismountVerb.DismountVerbName, HelpText = "Dismount a GVFS virtual repo")]
+    public class DismountVerb: UnmountVerb
+    {
+        private const string DismountVerbName = "dismount";
+    }
+}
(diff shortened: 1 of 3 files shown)


dscho:
Not a native English speaker here, so... take this with a grain of salt.

Having said that, I think you misunderstand the context. In VFS for Git, the verb ["to mount"](https://dictionary.cambridge.org/dictionary/learner-english/mount_1) is not used in the sense of "getting onto a horse or bicycle" but more in the sense of "fixing an object onto something". I.e. the verb ["to dismount"](https://dictionary.cambridge.org/dictionary/learner-english/dismount) would be incorrect here, contrary to your claim.

In any case, I oppose this on the following grounds:

- The "mount" action was developed in the context of a file system, for which it has been well established for about one eon and a half that the opposite of "mounting a file system" is to "unmount" it.
- VFS for Git is not actually developed actively, it is in maintenance mode, and all of the efforts go to establishing [Scalar](https://github.com/microsoft/scalar/) as its successor for as many of VFS of Git's use cases as possible. Implementing a new feature would send the wrong signal to users.

GabeDeBacker:
So I see that consistency for other commands such as:
https://docs.microsoft.com/en-us/powershell/module/hyper-v/dismount-vhd?view=win10-ps

Is not even a consideration?

dscho:
@GabeDeBacker I feel you did not even listen to anything I said.

derrickstolee:
Closing because stale.

PR #1719, opened 2021-02-02
Enforce only a minimum Git version

derrickstolee:
The interface between `microsoft/git` and VFS for Git has been stable for a while. We don't anticipate changing it any time soon. Thus, we can drop the requirement of an _exact_ Git version match.

This PR changes the current logic in a few ways:

1. Take the full GitVersion parsing logic from `microsoft/scalar`, as it is more robust.
2. Remove the exact version match with a minimum version match.
3. Use a fixed constant version string instead of a build-time string.
4. Require that the `Revision` of the version string is an _exact_ match. We will use this to signal that a `microsoft/git` version introduces a breaking change that requires a VFS for Git upgrade.

What is a `Revision` item? In a Git version string such as `v2.X.Y.vfs.Z.W`, the `Revision` is represented by `Z`. This allows us to modify the other version items arbitrarily, including taking version numbers `2.X.Y` from upstream and updating `.W` if we have any hotfixes to `microsoft/git`.


GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -34,60 +34,60 @@ public GVFSVerb(bool validateOrigin = true)
...
-            // currently tell the first access vs subsequent accesses so this
-            // flag just blocks them from occurring at all.
-            BlockFiltersAndEolConversions = 1 << 6,
-
-            // GVFS_PREFETCH_DURING_FETCH
-            // While performing a `git fetch` command, use the gvfs-helper to
-            // perform a "prefetch" of commits and trees.
-            PrefetchDuringFetch = 1 << 7,
+        }
+
```
jeffhostetler, line 38: Is this all whitespace changes or something?

derrickstolee: Yeah, one of the commits is just whitespace things. We still have a lot of files with mixed newlines due to the macOS effort. VS Code converts all to a single type on save.

derrickstolee: image

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -884,31 +884,23 @@ private void CheckGitVersion(ITracer tracer, GVFSEnlistment enlistment, out stri
...
-                    this.Output.WriteLine($"Warning: Installed git version {gitVersion} does not match supported version of {GVFSConstants.SupportedGitVersion}.");
-                }
+                this.ReportErrorAndExit(
+                    tracer,
+                    "Error: Installed git version {0} is less than the supported version of {1}.",
+                    gitVersion,
+                    GVFSConstants.SupportedGitVersion);
             }
-            else
+            else if (gitVersion.Revision != GVFSConstants.SupportedGitVersion.Revision)
```
jeffhostetler, line 895: I don't understand why we're having this `else if ... { ... }`.

I thought we were only enforcing a minimum installed Git version.  This reads like it is forcing an exact match.

derrickstolee: This is our escape hatch if we need to introduce a breaking change to `microsoft/git`: the version string `v2.XX.Y.vfs.Z.W` must match exactly only on the `.Z` part, but the rest is a minimum match.

jeffhostetler: oh ok.  thanks

jeffhostetler (approved):

PR #1727, opened 2021-03-19
Milestones/m185

jeffhostetler:
Major Changes
----------------

* Includes Git v2.31.0.
* VFS for Git now only performs a minimum version check with the installed version of Git. This means that VFS for Git is compatible with Git versions `v2.30.0.vfs.0.0` or greater, and users can upgrade Git independently from [the `microsoft/git` releases page](https://github.com/microsoft/git/releases) independently of VFS for Git.

Pull Requests
--------------

* #1716: Update Git to remove ODB string check
* #1719: Enforce only a minimum Git version
* #1726: Config: commitGraph.generationVersion=1


derrickstolee (approved):

dscho (approved):
Looks good to me!

PR #1730, opened 2021-03-26
Fix an issue with FastFetch when deleting files

### LINKED ISSUE #1729: FastFetch fails to delete files in some circumstances
If you have two branches, one with these files:

foo.xaml.cs
foo.xaml

When you switch from the branch with those files to the one without it, FastFetch will leave behind 'foo.xaml.cs' and correctly delete foo.xaml

SteveBenz:
This fixes #1729 and also takes care of a performance problem in the same method.


GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -166,37 +166,45 @@ public void ParseDiffFile(string filename)
...
                 {
-                    continue;
+                    string parentPath = Path.GetDirectoryName(filePath);
+                    if (!deletedDirectories.Contains(parentPath))
+                    {
+                        this.FileDeleteOperations.Enqueue(filePath);
+                    }
                 }
 
-                deletedPaths.Add(filePath);
```
SteveBenz, line 194: This line here is really the heart of the badness.  I believe if we just deleted this one line, it would not only fix the bug but also greatly reduce the performance problem.  We know for sure that 'filePath' is a file, and really 'deletedPaths' should always have been 'deletedFolders'.  It would also greatly reduce the performance problem, as it'd reduce it from O(n^2) to something more like O(n logn) (because the number of directories is more or less log(number-of-files).

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
```diff
@@ -90,10 +90,10 @@ public void CanParseBackwardsDiff()
             diffBackwards.RequiredBlobs.Count.ShouldEqual(10);
```
SteveBenz, line 90: This test is not all that it could be...

1) It just tests counts.  It'd be better if it ToString'ed the filenames that are in each set (where possible) so that when it fails you have some idea of what changed.
2) It calls 'ParseDiffFile' rather than 'PerformDiff'.  This misled me into thinking that 'PerformDiff' really didn't have any tests.  'ParseDiffFile', as it turns out, is just called by tests.  It'd be better (and not hard  to do) to make it call 'PerformDiff'.

Wouldn't take much to twist my arm to do that.

Scripts/BuildGVFSForWindows.bat
```diff
@@ -1,4 +1,4 @@
-@ECHO OFF
+@if "%_echo%"=="" (echo off) else (echo on)
 SETLOCAL
```
SteveBenz, line 2: I hope you don't mind a drive-by.  I had some struggles getting these to run.  We use this technique a bit around here to allow conditionally turning echo on to help diagnose the random "Error: wheel fell off" messages you get when running batch scripts in slightly incorrect environments"

derrickstolee: Sounds good to me. As long as the CI builds still pass with these messages, then I'm happy with the change.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
derrickstolee: I like that this has a performance improvement as well.

GVFS/GVFS.UnitTests/Prefetch/DiffHelperTests.cs
derrickstolee: I wouldn't worry too much about that. I imagine it would be difficult to modify the code in a way that is incorrect and still gets these numbers correct.

derrickstolee (approved):
Thank you for finding and fixing this.

Issue #1743, opened 2021-06-25
Race condition in BatchObjectDownloadStage


tyrielv:
I hit a situation when prefetching with "--files *" when almost all the blobs were already present. The logs showed that 3 blobs were missing, but that no download requests were made. The missing blobs were not actually downloaded (which wouldn't be a problem working with git through commandline or file system, but I was trying to use Libgit2 to build a tree using the blobs and it failed because they were missing). Running "gvfs prefetch --files *" again claimed that all blobs were prefetched so there was nothing to download, but they were still missing. Running "gvfs prefetch --files" and providing the actual files needed caused them to be prefetched.

I suspect that there is a race condition in BatchObjectDownloadStage between 
[L222 ](https://github.com/microsoft/VFSForGit/blob/a61e4dc82582d38efde0fd4c450efb5d1abe722a/GVFS/GVFS.Common/Prefetch/Pipeline/BatchObjectDownloadStage.cs#L222) and [L227](https://github.com/microsoft/VFSForGit/blob/a61e4dc82582d38efde0fd4c450efb5d1abe722a/GVFS/GVFS.Common/Prefetch/Pipeline/BatchObjectDownloadStage.cs#L227). It seems possible for the missingBlobs collection to finish adding in between those two lines, in which case any blobs that were added during that time would not be downloaded.

derrickstolee:
Maybe there is something about that `this.missingBlobs.IsAddingCompleted`, since otherwise it would loop 4000 times (waiting 100ms each time) before quitting if no blobs are added to the list.

If you use the `--hydrate` option on the `gvfs prefetch --files` command, does that populate the files?

tyrielv:
Now that you mention it, the 4000x loop seems like a more likely culprit than that tiny race condition, especially since I'm able to repro it on a fresh clone - it took nearly an hour (much longer than 400 seconds) for gvfs to finish determining that only 3 blobs needed to be downloaded.

PR #1744, opened 2021-06-28
Overhaul build and project systems and drop Mac/POSIX code

mjcheetham:
This PR simplifies and modernises the build system significantly. Move to the new MSBuild project file format for all .NET projects and avoid mixing ecosystems between `dotnet/msbuild`, `nuget/dotnet` and `packages.config/ProjectReference` (where possible).

All .NET projects target .NET Framework 4.6.1, and we drop all .NET Standard and .NET Core targets.

Native code uses the Windows SDK 10.0.16299 and the VS2019 toolset.

We also drop more/most of the macOS and POSIX code to aid in simplifying the build - VFS for Git on Mac and Linux were abandoned for various reasons.

Functional tests only run against the currently installed VFS for Git (on `%PATH%`) and the option to run against the just-built/dev version has been removed (it was not used in practice).

Add a GitHub Actions workflow for CI and PR builds. (Also added a YAML-based Azure DevOps pipeline for release builds, but this is not used.. yet).

**Notes:**

- The PR is best reviewed commit-by-commit, but note that each commit does _NOT_ build standalone, and the entire series must be taken as a whole to ensure a successful build.

- The majority of the new build design comes from Scalar and GCM Core, so there's some similarities in here to those systems. For example, the `GVFS.MSBuild` files come pretty much verbatim from the `Scalar.MSBuild` project in Scalar.

PR #1745, opened 2021-06-28
Fix for #1743 - do not exit early if no blobs found for a period when prefetching.

tyrielv:
Fix for #1743 - do not exit early if no blobs found for a period when prefetching.

### DIFF
--- a/GVFS/GVFS.Common/Prefetch/Pipeline/BatchObjectDownloadStage.cs
+++ b/GVFS/GVFS.Common/Prefetch/Pipeline/BatchObjectDownloadStage.cs
@@ -213,7 +213,7 @@ namespace GVFS.Common.Prefetch.Pipeline
             {
                 List<string> blobsInChunk = new List<string>();
 
-                for (int i = 0; i < this.chunkSize; ++i)
+                for (int i = 0; i < this.chunkSize;)
                 {
                     // Only wait a short while for new work to show up, otherwise go ahead and download what we have accumulated so far
                     const int TimeoutMs = 100;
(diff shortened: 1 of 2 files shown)


derrickstolee:
@tyrielv: did you build and install this locally to verify that it fixes your larger scenario? I'm not seeing a clear reason why this fixes your case.

tyrielv:
I'm still working on verifying it, but here's the reasoning:

If it takes more than ChunkSize * TimeoutMs (ie about 7 minutes) for the missingBlobs collection to produce its first value, then the loop currently exits and returns false. If it returns false, then the while loop on line 74 never runs and won't try again, even though the producer for missingBlobs is still running.

In my scenario it took about 1 hour for the missingBlobs producer to finish running and it only discovered 3 missing blobs out of  1.6 million, so it is very likely that there was at least one 7 minute gap before the 3 blobs were all discovered.

derrickstolee:
> I'm still working on verifying it, but here's the reasoning:
> 
> If it takes more than ChunkSize * TimeoutMs (ie about 7 minutes) for the missingBlobs collection to produce its first value, then the loop currently exits and returns false. If it returns false, then the while loop on line 74 never runs and won't try again, even though the producer for missingBlobs is still running.

I see. I now can connect the dots that the early stage is _really slow_ in some cases.

> In my scenario it took about 1 hour for the missingBlobs producer to finish running and it only discovered 3 missing blobs out of 1.6 million, so it is very likely that there was at least one 7 minute gap before the 3 blobs were all discovered.

And I didn't expect this to be so slow, but I suppose it would be when checking 1.6 million blobs via libgit2. I'm more convinced that this is a fix, but I appreciate you doing the extra testing.

We won't want to merge this until after #1744 is merged, anyway.

derrickstolee:
@tyrielv: #1744 is merged. Could you update your branch to kick off a new merge? That will allow us to run the PR validation builds. One easy way is to run `git commit --amend --no-edit && git push -f <remote> issue1743`

tyrielv:
@derrickstolee can you approve the workflow so this can run tests?

derrickstolee (approved):

PR #1746, opened 2021-06-30
Delete custom upgrader

derrickstolee:
Remove all code for the custom upgrader that ran the installer either from GitHub or from a NuGet feed.

This leaves `gvfs upgrade` non-functional, but a replacement will follow that calls `winget upgrade vfsforgit`.


mjcheetham (approved):
🎉

PR #1749, opened 2021-08-02
Update Microsoft.Windows.ProjFS version

derrickstolee:
This is mostly a test to check something with the build. While the fixed number allows taking a newer version, it focuses on taking the lowest available version that is at least this version. Let's try taking the latest version explicitly.


derrickstolee:
Of course this won't work because we use a custom NuGet feed to get our dependencies.

derrickstolee:
Trying again after removing our custom feed.

PR #1756, opened 2021-10-15
"winget install microsoft-git" finds entries in msstore

Eschryn:
To address this: using the id parameter we can make sure it will always install this package.

Original Content:
Adding the source parameter will make sure that the install command will actually install from winget

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -29,7 +29,7 @@ To install, use [`winget`](https://github.com/microsoft/winget) to install the
 using:
 
 ```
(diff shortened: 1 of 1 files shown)


derrickstolee:
Hi @eschryn. Could you explain more about what it means to "find entries in msstore"? This seems like a strange behavior for `winget` to not use the source that we have specified. The recommended change to add `-s winget` seems odd and it would be good to have some justification.

PR #1760, opened 2021-10-30
Update Git to v2.34.0.vfs.0.0

derrickstolee:
There were some build issues related to the install path moving from "GVFS" to "VFS for Git". The pack-file maintenance tests were flaky for reasons we've seen before on the Scalar side, so let's focus our efforts keeping the Scalar tests green.


dscho (approved):
All changes look good to me.

PR #1764, opened 2022-01-24
Update to Git 2.35.0

derrickstolee:
See microsoft/git#476

### DIFF
--- a/Version.props
+++ b/Version.props
@@ -18,7 +18,7 @@
        including taking version numbers 2.X.Y from upstream and updating .W if we have any
        hotfixes to microsoft/git.
     -->
-    <GitPackageVersion>2.20211115.1</GitPackageVersion>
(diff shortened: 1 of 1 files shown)


Version.props
```diff
@@ -18,7 +18,7 @@
        including taking version numbers 2.X.Y from upstream and updating .W if we have any
        hotfixes to microsoft/git.
     -->
-    <GitPackageVersion>2.20211115.1</GitPackageVersion>
+    <GitPackageVersion>2.20220123.2-pr</GitPackageVersion>
```
ldennington, line 21: Thanks for doing this, @derrickstolee! Would you mind sharing where to find the updated PackageVersion so I can add it to the docs?

derrickstolee: This is a manual process, separate from releasing microsoft/git. Our internal `MAINTAINING-MICROSOFT-GIT.md` document has some details.

PR #1787, opened 2022-09-02
file.Delete() if file.Exists

ankurash:
If some other process is deleting files at the same time, this fails the entire operation even though the intended action of files to be deleted has been successful.


GVFS/GVFS.Common/FileSystem/PhysicalFileSystem.cs
```diff
@@ -27,7 +27,10 @@ public virtual void DeleteDirectory(string path, bool recursive = true, bool ign
                 foreach (FileInfo file in directory.GetFiles())
                 {
                     file.Attributes = FileAttributes.Normal;
-                    file.Delete();
+                    if(file.Exists)
```
Porges, line 30: there is still a race here that the file exists and is then deleted before the delete is invoked; a better solution is to perform the delete unconditionally and then catch the exception if the file does not exist

Issue #1802, opened 2024-01-08
VFS for Git is incompatible with Azure DevOps OAuth


johnterickson:
See https://github.com/git-ecosystem/git-credential-manager/issues/1415 for details

There is a workaround to use a less secure option:
`git config --global credential.azreposCredentialType pat`

PR #1803, opened 2024-01-26
fix: #1802 confirm cached credential before rejecting

tyrielv:
**Issue**
When GCM is in OAuth mode, rejecting the credential is equivalent to a sign-out request, which means the next attempt to get a credential will always cause a pop-up. See #1802 

** Changes **
Before rejecting a credential try getting the credential from GCM again.
If GCM still returns the same credential, then tell GCM to reject it.
If GCM returns a different credential, don't attempt to reject the previous one - just return and let the caller try the new credential instead.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitAuthentication.cs
+++ b/GVFS/GVFS.Common/Git/GitAuthentication.cs
@@ -91,9 +91,27 @@ namespace GVFS.Common.Git
         {
             lock (this.gitAuthLock)
             {
+                var cachedCredentialAtStartOfReject = this.cachedCredentialString;
                 // Don't stomp a different credential
-                if (credentialString == this.cachedCredentialString && this.cachedCredentialString != null)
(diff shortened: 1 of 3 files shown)


tyrielv:
@microsoft-github-policy-service agree company="Microsoft"

derrickstolee (approved):
I left a few nits based on the repo style guide, but they aren't critical.

Also, I merged your change with mine in #1804 to make sure that everything built, passed unit tests, and ran against a full production repo just fine. The merge is in [`derrickstolee:oath-change-2024`](https://github.com/derrickstolee/VFSForGit/commits/oath-change-2024/).

I'm currently running the default functional test suite as an extra precaution, but I don't anticipate that this change will have any effect because the test suite operates against a public repo and thus does not test the auth stack in any way.

You may also want to cherry-pick abcb73f08f20111ff3519adde44f5714cc61acaf since that will set the default mode to oauth, giving us faster adoption of the mode in VFS for Git customers.

tyrielv:
@derrickstolee Feedback taken

derrickstolee (approved):

PR #1804, opened 2024-02-15
Building VFS for Git in 2024

derrickstolee:
In order to evaluate the merit of other changes, such as #1803, I needed these changes locally in order to get a clean functional test run. Each change is isolated to its own commit, so review commit-by-commit.

Generally, there are a few categories of problems:

1. README updates focusing on the actual requirements and dropping VS 2017 from mention.
2. A fix for a version parsing problem. Also update the version to the latest sprint (https://whatsprintis.it/).
3. Update a stale Git config key to avoid warnings.
4. Modify a couple failing functional test due to Git's updated handling of empty/truncated objects. They fail with v2.43.0.vfs.0.0 which is not currently used in the PR/CI builds.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tools/ControlGitRepo.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/ControlGitRepo.cs
@@ -55,7 +55,7 @@ namespace GVFS.FunctionalTests.Tools
             GitProcess.Invoke(this.RootPath, "config merge.renames false");
             GitProcess.Invoke(this.RootPath, "config advice.statusUoption false");
             GitProcess.Invoke(this.RootPath, "config core.abbrev 40");
-            GitProcess.Invoke(this.RootPath, "config core.useBuiltinFSMonitor false");
+            GitProcess.Invoke(this.RootPath, "config core.fsmonitor false");
             GitProcess.Invoke(this.RootPath, "config pack.useSparse true");
             GitProcess.Invoke(this.RootPath, "config reset.quiet true");
             GitProcess.Invoke(this.RootPath, "config status.aheadbehind false");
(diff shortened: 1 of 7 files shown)


GVFS/GVFS.FunctionalTests/Tools/ControlGitRepo.cs
```diff
@@ -55,7 +55,7 @@ public void Initialize()
             GitProcess.Invoke(this.RootPath, "config merge.renames false");
             GitProcess.Invoke(this.RootPath, "config advice.statusUoption false");
             GitProcess.Invoke(this.RootPath, "config core.abbrev 40");
-            GitProcess.Invoke(this.RootPath, "config core.useBuiltinFSMonitor false");
+            GitProcess.Invoke(this.RootPath, "config core.fsmonitor false");
```
derrickstolee, line 58: @jeffhostetler: Perhaps the old line should stay here, and we just _add_ the `core.fsmonitor` line, too. Keeping the old config won't hurt, and might help for users still on an older version of Git.

GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -311,7 +311,7 @@ public static bool TrySetRequiredGitConfigSettings(Enlistment enlistment)
                 { "commitGraph.generationVersion", "1" },
 
                 // Disable the builtin FS Monitor in case it was enabled globally.
-                { "core.useBuiltinFSMonitor", "false" },
+                { "core.fsmonitor", "false" },
```
jeffhostetler, line 314: same here.

Yedfree (approved):

PR #1808, opened 2024-03-13
Revert "GVFSVerb: Use OAuth credentials by default"

derrickstolee:
This reverts commit cbd0cefb9cc87744286fde23632860354f5c62ee. This commit was introduced in #1803 based on a comment I made recommending "faster adoption" but that was a mistake. When the release was shipped to `winget` yesterday, this led to breaks in some pipeline environments.

Using this authentication mode by default is breaking some automated scenarios on build agents that use "gvfs clone". Removing the new default is a fast measure to fix these users, while a longer-term consideration for increasing adoption of OAuth can be done at a later date with the proper parties engaged.

### DIFF
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -312,11 +312,6 @@ namespace GVFS.CommandLine
 
                 // Disable the builtin FS Monitor in case it was enabled globally.
                 { "core.useBuiltinFSMonitor", "false" },
-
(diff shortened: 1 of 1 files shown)


derrickstolee:
(Please ignore the confusion between commits cbd0cefb9cc87744286fde23632860354f5c62ee (actually merged in the release) and abcb73f08f20111ff3519adde44f5714cc61acaf (was in my PR, but cherry-picked into #1804).

vdye (approved):

PR #1822, opened 2024-10-24
Allow mounting on ReFS volumes

winstliu:
ReFS is the underlying filesystem that Dev Drives use and should be supported.

VFS for Git let ReFS drives through fine under https://github.com/microsoft/VFSForGit/releases/tag/v1.0.24074.1. I'm not sure why the volume check did not error out, but _something_ changed and after installing the latest prereleases, existing enlistments on Dev Drives are now blocked. They should not be.

(git bisect did not find the issue, nor did I expect it to given this code path did not change)


dscho (approved):
This change looks good to me, and I am just as puzzled why this change was not required before.

PR #1825, opened 2024-11-14
Attempt to mitigate locked machine.config

tyrielv:
Antivirus occasionally opens machine.config for writing. If a .NET Framework app happens to launch during that brief window, it will create a popup with an error message and not continue until popup is closed. Since GVFS periodically runs maintenance tasks, it is not uncommon for users to find the error popup waiting for them first thing in the morning.
In addition, any GVFS C# app that makes HTTP requests will initialize ServicePointManager. This call can also fail if machine.config is locked (ie if the lock started after the app started so it didn't do the pop-up, or if the pop-up has been closed/skipped and the machine.config is still locked), and once it does it is unrecoverable. 

This pull request attempts to mitigate the effects of machine.config locking in two ways:
1. GitHooksLoader is responsible for starting GVFS.Hooks.exe, by far the most likely place to hit this issue. By calling SetErrorMode(SEM_FAILCRITICALERRORS) before invoking GVFS.Hooks.exe, the popup is suppressed.
2. In HttpRequestor initialization, try briefly to lock machine.config ourselves for reading before initializing ServicePointManager.

### DIFF
--- a/GVFS/GVFS.Common/Http/HttpRequestor.cs
+++ b/GVFS/GVFS.Common/Http/HttpRequestor.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


derrickstolee (approved):
Thanks for the clear motivation, @tyrielv! The code changes look solid. Too bad that it's probably not worth trying to create a test for this scenario.

dscho (approved):

mjcheetham (approved):

PR #1815, opened 2024-08-21
Update dependencies

dscho:
In reaction to Component Detection pointing out two more issues where dependencies are missing legal information, this PR updates the dependencies (all of them, including the two pointed out by Component Detection) to their latest stable versions.

This includes updating the InnoSetup dependency to a newer version, the LibGit2Sharp dependency to a newer version, and all remaining dependencies to their latest stable versions.

PR #1827, opened 2024-12-03
Remove x64 check from installer

tyrielv:
PR #1815 updated the installer dependency version, which seems to have caused the installer to enforce the AllowedArchitectures=x64 check (which has been there for years). We've received complaints from users that the latest release refuses to install on their ARM machines, but previous versions did and worked fine with the default x86/64 emulation.

Although ARM is not explicitly supported, we should not prevent users from installing GVFS on ARM when it has been working for them so far.


derrickstolee (approved):

GVFS/GVFS.Installers/GVFS.Installers.csproj
```diff
@@ -28,9 +28,7 @@
     <Content Include="$(PkgGitForWindows_GVFS_Portable)\tools\*" />
     <Content Include="install.bat" />
     <Content Include="info.bat" />
-    <NuspecTemplate Include="GVFS.Installers.template.nuspec"
```
mjcheetham, line 31: Please can you revert the other changes to this file (not the `Tools.InnoSetup` version bump of course)?

mjcheetham (changes requested):
Looks good, just need to revert the `&quot;` that got added to the csproj file by Visual Studio. Thanks!

GVFS/GVFS.Installers/GVFS.Installers.csproj
tyrielv: Done

mjcheetham (approved):
Thank you @tyrielv!

PR #1830, opened 2025-02-13
Remove old unused upgrader methods and dependencies

mjcheetham:
Remove some of the unused methods left behind after the auto upgrader was shut down. Also remove the no longer needed dependencies.


dscho (approved):
🎉

PR #1833, opened 2025-04-02
Update and futureproofing of ConfigureVisualStudio

tyrielv:
Visual Studio contains a copy of mingit that it will use for Git operations in its UI (Team Explorer, etc.), which can be overridden by a registry key.

Previously, gvfs mount would set the registry key to point to the user's installed version of git (presumably the .vfs fork), but it only set the registry key for versions 15.0 and 16.0 of Visual Studio.

This commit instead looks for registry keys for versions of Visual Studio at least 15.0, in the same format as the known keys, and sets the key for them.

The current version of Visual Studio is 17.0, and this is fix is confirmed to work for it.

With luck this will also work for future versions of Visual Studio, but there's no guarantee of that.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
@@ -220,25 +220,28 @@ namespace GVFS.Platform.Windows
             try
             {
                 const string GitBinPathEnd = "\\cmd\\git.exe";
-                string[] gitVSRegistryKeyNames =
-                {
-                    "HKEY_CURRENT_USER\\Software\\Microsoft\\VSCommon\\15.0\\TeamFoundation\\GitSourceControl",
-                    "HKEY_CURRENT_USER\\Software\\Microsoft\\VSCommon\\16.0\\TeamFoundation\\GitSourceControl"
-                };
+                const string VSRegistryKeyRoot = @"Software\Microsoft\VSCommon";
+                const string GitVSRegistrySubKey = @"TeamFoundation\GitSourceControl";
                 const string GitVSRegistryValueName = "GitPath";
 
                 if (!gitBinPath.EndsWith(GitBinPathEnd))
                 {
-                    tracer.RelatedWarning(
-                        "Unable to configure Visual Studio’s GitSourceControl regkey because invalid git.exe path found: " + gitBinPath,
-                        Keywords.Telemetry);
-
                     return;
                 }
 
                 string regKeyValue = gitBinPath.Substring(0, gitBinPath.Length - GitBinPathEnd.Length);
-                foreach (string registryKeyName in gitVSRegistryKeyNames)
+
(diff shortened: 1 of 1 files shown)


dscho (approved):
Nice! In `microsoft/git`, we [extended this list, too](https://github.com/microsoft/git/blob/19154389a8f82b31971c9a8940eb15f061a024f6/.github/workflows/build-git-installers.yml#L259-L264) albeit by hard-coding the version numbers.

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
```diff
@@ -220,25 +220,28 @@ public override void ConfigureVisualStudio(string gitBinPath, ITracer tracer)
...
 
                 string regKeyValue = gitBinPath.Substring(0, gitBinPath.Length - GitBinPathEnd.Length);
-                foreach (string registryKeyName in gitVSRegistryKeyNames)
+
+                /* Get all versions of Visual Studio that exist in the registry at least 15.0.
+                 * This attempts to future proof (current version is 17.0), but may need to be
+                 * revisited in the future if VS changes how it stores this. */
+                var vsVersions = Registry.CurrentUser.OpenSubKey(VSRegistryKeyRoot)
+                    ?.GetSubKeyNames()
+                    .Where(name => float.TryParse(name, out var version) && version >= 15.0)
```
mjcheetham, line 239: The equality arm of `>= 15.0` unnerves me.. comparing equality of two floating-point numbers. But, if this has been tested for VS15, then I guess this is ok.

mjcheetham (approved):

GVFS/GVFS.Platform.Windows/WindowsPlatform.cs
mjcheetham: Could also use `System.Version.TryParse` and compare the `.Major` version parts.

mjcheetham (changes requested):
Upon second thought.. I think I'd prefer avoiding floats here and using `Version.Major` integer comparison instead.

tyrielv:
Created https://github.com/microsoft/git/issues/739 to track the winget installer issue.

dscho (approved):
Thank you!

mjcheetham (approved):
Thanks for the changes! Just one thing to note.. this change now means that you must install Visual Studio _before_ VFS for Git, otherwise these registry subkey versions will not be present and therefore the `GitSourceControl` key will not be written. I assume this is acceptable for your use case @tyrielv?

tyrielv:
> Thanks for the changes! Just one thing to note.. this change now means that you must install Visual Studio _before_ VFS for Git, otherwise these registry subkey versions will not be present and therefore the `GitSourceControl` key will not be written. I assume this is acceptable for your use case @tyrielv?

Yes, and also this runs on mount (not just on install) so a remount would fix it.

PR #1837, opened 2025-05-12
Fix std handle forwarding

tyrielv:
Currently, GitHooksLoader.exe does not forward any output from the hooks that it is configured to run. This is different from the expected git hooks behavior, and it also does not appear to be intentional - stdout and stderr are explicitly configured to be forwarded, but the CREATE_NO_WINDOW flag prevents that from happening. There are also outputs in GVFS.Hooks.exe (which in most cases is the only hook called by this) that are clearly expected to be visible to the user, but are not.

This PR removes CREATE_NO_WINDOW so that standard handles are forwarded. It also removes the explicit forwarding of stdout and stderr, which has the effect that stdin is also forwarded. I'm not aware of any reasons we wouldn't want to forward stdin - without it, if a configured hook expects user input then it will just hang.

### DIFF
--- a/GVFS/GitHooksLoader/GitHooksLoader.cpp
+++ b/GVFS/GitHooksLoader/GitHooksLoader.cpp
@@ -112,9 +112,6 @@ int ExecuteHook(const std::wstring &applicationName, wchar_t *hookName, int argc
     PROCESS_INFORMATION pi;
     ZeroMemory(&si, sizeof(si));
     si.cb = sizeof(si);
-    si.hStdOutput = GetStdHandle(STD_OUTPUT_HANDLE);
-    si.hStdError = GetStdHandle(STD_ERROR_HANDLE);
-    si.dwFlags = STARTF_USESTDHANDLES;
 
     ZeroMemory(&pi, sizeof(pi));
 
@@ -135,7 +132,7 @@ int ExecuteHook(const std::wstring &applicationName, wchar_t *hookName, int argc
         NULL,           // Process handle not inheritable
         NULL,           // Thread handle not inheritable
         TRUE,           // Set handle inheritance to TRUE
-        CREATE_NO_WINDOW, // Process creation flags
+        NULL,           // Process creation flags
         NULL,           // Use parent's environment block
         NULL,           // Use parent's starting directory 
         &si,            // Pointer to STARTUPINFO structure


GVFS/GitHooksLoader/GitHooksLoader.cpp
```diff
@@ -135,7 +132,7 @@ int ExecuteHook(const std::wstring &applicationName, wchar_t *hookName, int argc
         NULL,           // Process handle not inheritable
         NULL,           // Thread handle not inheritable
         TRUE,           // Set handle inheritance to TRUE
-        CREATE_NO_WINDOW, // Process creation flags
+        NULL,           // Process creation flags
```
dscho, line 135: I seriously wonder whether this flag is there on purpose, to avoid "flashing windows" when, say, running `git.exe` from Visual Studio.

GVFS/GitHooksLoader/GitHooksLoader.cpp
```diff
@@ -112,9 +112,6 @@ int ExecuteHook(const std::wstring &applicationName, wchar_t *hookName, int argc
     PROCESS_INFORMATION pi;
     ZeroMemory(&si, sizeof(si));
     si.cb = sizeof(si);
-    si.hStdOutput = GetStdHandle(STD_OUTPUT_HANDLE);
-    si.hStdError = GetStdHandle(STD_ERROR_HANDLE);
-    si.dwFlags = STARTF_USESTDHANDLES;
```
dscho, line 117: So what you're saying is that this explicit redirection of `stdout`/`stderr` currently does _not_ work? This is puzzling because the `GitHooksLoader.exe` process _should_ have the correct handles, and this way of creating the child process _should_ work...

dscho (commented):
I do not think that `stdin` is worth bothering about, given that Git [specifically prevents Git hooks from accepting interactive user input](https://github.com/git-for-windows/git/blob/v2.49.0.windows.1/hook.c#L66).

Also, I am quite puzzled that the redirection should not work.

The only thing I vaguely remember causing problems with the usage pattern of the code touched by this PR is that the handles need to be duplicated lest they are closed by the child process...

GVFS/GitHooksLoader/GitHooksLoader.cpp
tyrielv: Good point, I'll check that.

GVFS/GitHooksLoader/GitHooksLoader.cpp
tyrielv: Yes, it does not work. For example running git status on an unmounted gvfs repo displays only the standard "pre command hook failed" message, and not the specific message about needing to mount.

I wonder if it's somehow environment dependent; I have a vague memory of seeing that message years ago, but there has been no change to this file.

However the [documentation](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags) for CREATE_NO_WINDOW says `The process is a console application that is being run without a console window. Therefore, the console handle for the application is not set.`

GVFS/GitHooksLoader/GitHooksLoader.cpp
tyrielv: I observed the updated hook getting called in procmon from VS, but there was no flashing window that I could see. I would expect that VS is already launching its git command with CREATE_NO_WINDOW, so the hooks loader is inheriting that state and passing it on.

tyrielv: ~~I've updated the change to only modify the behavior if stdout or stderr is populated to reduce the risk of flashing windows, though I didn't observe any in testing.~~
See other comment thread.

GVFS/GitHooksLoader/GitHooksLoader.cpp
tyrielv: I have updated this to more closely match how [git-wrapper in microsoft/git ](https://github.com/git-for-windows/MINGW-packages/blob/994073c2964b48d9dfcf27124160cf0b237d0239/mingw-w64-git/git-wrapper.c#L882) works - checking explicitly for a console with CONOUT$, and if found then use default behavior (ie don't set CREATE_NO_WINDOW, don't explicitly forward handles as they are forwarded by default), falling back to the current behavior if CONOUT$ is not found (set CREATE_NO_WINDOW and explicitly forward handles)

dscho: Excellent, thank you! Did you test this a bit, including running `git status` on an unmounted GVFS repo?

tyrielv: Yes, it now displays `The repo does not appear to be mounted. Use 'gvfs status' to check.
fatal: pre-command hook aborted command`
instead of just the `fatal` line.

PR #1840, opened 2025-06-13
Always index locally for prefetch packs

tyrielv:
The GVFS protocol includes an index file along with pack file in the prefetch workflow (primarily used on a new clone to fetch all commits and trees).

Currently, GVFS accepts that index file without any verification.

This pull request changes GVFS prefetch to discard the index sent by the server and always create an index locally. This provides improved security and verification of the pack file, at the expense of performance for the first clone of a repository with a new local cache.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -706,64 +706,53 @@ namespace GVFS.Common.Git
 
                     bytesDownloaded += packLength;
 
-                    // We will try to build an index if the server does not send one
-                    if (pack.IndexStream == null)
+                    // We can't trust the index file from the server, so we will always build our own.
+                    // We still need to consume and handle any exceptions from the index stream though.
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -706,64 +706,53 @@ private LooseObjectToWrite GetLooseObjectDestination(string sha)
 
                     bytesDownloaded += packLength;
 
-                    // We will try to build an index if the server does not send one
-                    if (pack.IndexStream == null)
+                    // We can't trust the index file from the server, so we will always build our own.
+                    // We still need to consume and handle any exceptions from the index stream though.
```
mjcheetham, line 710: I wonder if we may want to add an option to use the pack index as given by the server (default to ignore it of course) for performance critical scenarios where we can also trust the server.

For now let's just be safe and never trust the pack index, and can add this in the future if needed.

PR #1843, opened 2025-06-16
Index prefetch packs in parallel

tyrielv:
#1840 required local indexing of pack files (an expensive task for large repositories) to `gvfs clone`. This pull request improves the performance of pack file indexing by:
1. Launching `git index-pack` instances asynchronously during prefetch, and
2. Specifying a number of threads equal to processor count to `git index-pack`, overriding the git default of half the processor count.

The prefetch API will usually return one large pack file comprising most of the commits and trees, followed by several smaller pack files representing the new commits and trees since the last maintenance job consolidated the main pack file.

`git index-pack` begins with a single-threaded task to index the objects in the pack file, followed by a multi-threaded task to resolve the deltas. With these changes, the smaller incremental pack files are typically indexed in full while the primary pack file is still in its single-threaded phase, and the primary pack file's multi-threaded phase is marginally faster.

In testing, these changes reduced clone time for a large repository by 15-30%.

PR #1846, opened 2025-06-23
Distrust received pack indexes (behind config flag, with perf fixes)

tyrielv:
This pull request contains 4 things:
1. #1840  to index packs locally, which was reverted due to unexpected performance issues
2. #1843 to increase parallelization when indexing packs, which was abandoned due to #1840 being reverted
3. Added a new configuration "gvfs.trust-pack-indexes", which defaults to true. The new behavior is only used when this is set to false.
This is intended to be a temporary configuration setting, to be removed (or defaulted to false) once more mitigations have been completed for the performance.
4. `git index-pack` is changed from running in the GVFS enlistment to running outside the enlistment. This was the reason for the unexpected performance issues. 
It was expected that the first prefetch on a new clone would take longer due to indexing the pack locally; however users who deleted their prefetch cache (but not the rest of the cache or local loose objects) in order to re-fetch it with local indexing enabled experienced many times longer delays than expected, because `git index-pack` reads all the existing pack indexes and loose objects and considers them when indexing a pack in order to support validating that all referenced objects exist - even when the command-line options to act on nonexistent references are not enabled. Since we aren't using --validate or its variants, we can run `git index-pack` outside the enlistment to avoid this issue.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitObjects.cs
+++ b/GVFS/GVFS.Common/Git/GitObjects.cs
@@ -131,11 +131,23 @@ namespace GVFS.Common.Git
             {
                 long bytesDownloaded = 0;
 
+                /* Distrusting the indexes from the server is a security feature to prevent a compromised server from sending a
+                 * pack file and an index file that do not match.
+                 * Eventually we will make this the default, but it has a high performance cost for the first prefetch after
+                 * cloning a large repository, so it must be explicitly enabled for now. */
+                bool trustPackIndexes = true;
+                if (gitProcess.TryGetFromConfig(GVFSConstants.GitConfig.TrustPackIndexes, forceOutsideEnlistment: false, out var valueString)
+                    && bool.TryParse(valueString, out var trustPackIndexesConfig))
+                {
+                    trustPackIndexes = trustPackIndexesConfig;
+                }
+                metadata.Add("trustPackIndexes", trustPackIndexes);
+
                 long requestId = HttpRequestor.GetNewRequestId();
                 List<string> innerPackIndexes = null;
                 RetryWrapper<GitObjectsHttpRequestor.GitObjectTaskResult>.InvocationResult result = this.GitObjectRequestor.TrySendProtocolRequest(
                     requestId: requestId,
-                    onSuccess: (tryCount, response) => this.DeserializePrefetchPacks(response, ref latestTimestamp, ref bytesDownloaded, ref innerPackIndexes, gitProcess),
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -580,7 +580,17 @@ public Result VerifyCommitGraph(string objectDir)
 
         public Result IndexPack(string packfilePath, string idxOutputPath)
         {
-            return this.InvokeGitAgainstDotGitFolder($"index-pack -o \"{idxOutputPath}\" \"{packfilePath}\"");
+            /* Git's default thread count is Environment.ProcessorCount / 2, with a maximum of 20.
+             * Testing shows that we can get a 5% decrease in gvfs clone time for large repositories by using more threads, but
+             * we won't go over ProcessorCount or 20. */
+            var threadCount = Math.Min(Environment.ProcessorCount, 20);
```
mjcheetham, line 586: Is there any negative impact on clone times for machines that have many fewer available cores compared to threads?

GVFS/GVFS.Common/Git/GitObjects.cs
```diff
@@ -394,6 +406,7 @@ public virtual GitProcess.Result IndexPackFile(string packfilePath, GitProcess g
                     gitProcess = new GitProcess(this.Enlistment);
                 }
 
+
```
mjcheetham: nit: extra new line not needed

mjcheetham (commented):
Just one question (and one nit pick 😄) about thread count.

GVFS/GVFS.Common/Git/GitObjects.cs
tyrielv: Fixed

GVFS/GVFS.Common/Git/GitProcess.cs
tyrielv: Do you mean, if there are cores with more hyperthreading available than the typical 2 threads per core?

In that case it's possible that there is even more room to raise the thread count, but in my experience Environment.ProcessorCount usually is the number of hardware threads rather than hardware cores (eg it will report 16 on an 8-core, 16-thread processor) so this will attempt to use all the hardware threads. The [comments](https://github.com/microsoft/git/blob/6e041f81f75a8db57d902e53a0cf0a313580dd5b/builtin/index-pack.c#L2061C4-L2068C6) around git code for the default number of threads suggest that number of cores is more important than number of threads though, which is why they default to half (ie they assume 2 threads per core) and is probably why I only observed 5% improvement when doubling that. (Most of the overall improvement comes from running the index tasks for smaller pack files in parallel while the big one is still in its single-threaded phase, not from using more threads in the multi-threaded phase)

PR #1849, opened 2025-06-27
Prefetch in background when cloning if pack indexes are not trusted

tyrielv:
Since prefetch can be quite long for large repos if pack indexes from the server are not trusted (ie need to be built locally after the pack files are downloaded), start prefetch as a separate background process in that case.

See also #1848 for making more history operations usable while the prefetch is still running.

### DIFF
--- a/GVFS/GVFS/CommandLine/CloneVerb.cs
+++ b/GVFS/GVFS/CommandLine/CloneVerb.cs
@@ -215,20 +215,40 @@ namespace GVFS.CommandLine
                 {
                     if (!this.NoPrefetch)
                     {
-                        ReturnCode result = this.Execute<PrefetchVerb>(
-                            enlistment,
-                            verb =>
-                            {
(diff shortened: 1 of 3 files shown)


GVFS/GVFS/CommandLine/CloneVerb.cs
```diff
@@ -215,20 +215,40 @@ public override void Execute()
                 {
                     if (!this.NoPrefetch)
                     {
-                        ReturnCode result = this.Execute<PrefetchVerb>(
-                            enlistment,
-                            verb =>
+                        bool trustPackIndexes = enlistment.GetTrustPackIndexesConfig();
```
derrickstolee, line 218: I see that this is already being loaded in the prefetch command, so we don't pass this as an option to the prefetch command as a CLI argument. That took me a tiny bit to figure out, so I thought it worth mentioning.

PR #1853, opened 2025-07-28
Download Microsoft Git from GitHub releases

mjcheetham:
We've not been running the functional tests against a new version of Microsoft Git in a while.. today we're still pointing at v2.35.3.vfs.0.0 which was released in 2022!

The [GVFS.Installers project](https://github.com/microsoft/VFSForGit/blob/2ad18e8ed9b070fed9f018eb1011e3ecfdddf8dc/GVFS/GVFS.Installers/GVFS.Installers.csproj#L16-L17) (that creates the NuGet packages that the Functional Tests consume) is set to download the Git installer and portable archives from the [ci/Dependencies](https://github.com/microsoft/VFSForGit/blob/2ad18e8ed9b070fed9f018eb1011e3ecfdddf8dc/nuget.config#L8) feed in the [GVFS Azure DevOps organisation](https://dev.azure.com/gvfs/ci/_artifacts/feed/Dependencies). The package version is set centrally for the build [here](https://github.com/microsoft/VFSForGit/blob/2ad18e8ed9b070fed9f018eb1011e3ecfdddf8dc/Version.props#L21), in the Version.props file.

The [build that publishes new versions of that package](https://dev.azure.com/gvfs/ci/_build?definitionId=11) are woefully out of date and no longer run. The last run was June 20, 2022. This was around the time official microsoft/git builds were [moved to GitHub workflows](https://github.com/microsoft/git/commit/5528db77358bbe0267ab183f6e2eb2581b76b277).

Here we instead source the Microsoft Git installers from GitHub releases in the build workflow. Updating the version of Git used in FTs can be updated by modifying the build.yaml workflow (or setting the variable at queue-time if doing a `workflow_dispatch` trigger).


dscho:
I believe that the tests [fail](https://github.com/microsoft/VFSForGit/actions/runs/16576741307/job/46883225301?pr=1853#step:8:38062) because of https://github.com/microsoft/git/commit/5efe36612ac493f5a38d46714ec92302d8cae656, as @tyrielv had pointed out to use in private communication earlier.

So we'll need to adjust _something_ in the tests to accommodate for the change of behavior.

I guess the easiest would be to patch https://github.com/microsoft/VFSForGit/blob/2ad18e8ed9b070fed9f018eb1011e3ecfdddf8dc/GVFS/GVFS.FunctionalTests/Tools/GitHelpers.cs#L112 to strip out the newly-added lines from `actualOutput`.

mjcheetham:
> but I'm debugging why it's not working at the moment.

Of course `status` prints to standard output, and the existing filtering logic works on standard _error_. Should work now - we're filtering both.

dscho (approved):
Awesome!

Issue #1855, opened 2025-08-11
`git restore` after `git cherry-pick -n` can leave GVFS in a bad state


tyrielv:
1. Use `git cherry-pick -n` to cherry-pick some changes to current branch, leaving them uncommitted
2. Use `git restore --staged` to unstage the changes

Expected:
All the changes from the cherry-pick move from staged to unstaged in git status.

Actual:
Sometimes some changes disappear from git status. 
**The files still have the changed contents**, but git no longer recognizes the files as being changed.

It's unclear so far whether cherry-pick -n is required or if manual changes with `git add` will also repro.

`git checkout` to another branch appears to properly reset the file contents. (I've heard others complain about situations where it did not, though I don't know if they were caused by this and I haven't repro'd that.)

tyrielv:
I think it has to do with handling of skip-worktree flag in the index, which was added in v4 of index format to support sparse-checkout.

In a GVFS repo most files are `skip-worktree=True` by default.

If you git add a file to stage it, then it changes to `skip-worktree=False` and `git restore --staged` will work as expected.

If you stage a file via another way, such as `git cherry-pick -n`, it does not change skip-worktree, and `git restore --staged` will [exit early](https://github.com/microsoft/git/blob/336bf91fc0b5811ba88bbce1bfec25b82a56770a/builtin/checkout-index.c#L93), which results in updating the index to unstage, leaving the file untouched (ie in modified state), but with `skip-worktree=true` so further git commands don't recognize that the file is in modified state.

I verified that it works as I described with another way to stage a file, `git reset --soft HEAD~1` to checkout the prior commit but leave the working directory files in their current state for the files changed in current commit and stage them in the index.

It looks like this is related to several prior issues around skip-worktree:
https://github.com/microsoft/git/issues/683
https://github.com/microsoft/git/issues/490
https://github.com/microsoft/git/issues/646

A fix might be for all ways to stage a file to set `skip-worktree=false` in a GVFS repo, or for whichever point is modifying the file on disk to do so. (Non-GVFS sparse-checkout repos (eg using Scalar) I'd expect to not modify the file on disk if it's outside the sparse-checkout, so inserting the change at point of modifying the file is probably simpler.)

tyrielv:
PostIndexChanged sends two flags: `UpdatedWorkingDirectory` and `UpdatedSkipWorktreeBits`. 

[...]

PR #1856, opened 2025-08-14
Functional test for cherry-pick -> restore corruption

tyrielv:
Adds a test to reproduce corruption of cherry-pick followed by restore. This should remove all staged changes in the index but leave the working directory unchanged.
It still leaves the working directory unchanged, but the index is in a corrupted state and git status no longer accurately reflects the working directory.

See #1855. It hasn't been determined yet whether microsoft/git or VFSForGit is the cause of the error - it could be both.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CorruptionReproTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CorruptionReproTests.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using GVFS.Common;
+using GVFS.FunctionalTests.Properties;
+using GVFS.FunctionalTests.Tests.EnlistmentPerTestCase;
+using NUnit.Framework;
+
+namespace GVFS.FunctionalTests.Tests.GitCommands
+{
+    /// <summary>
+    /// This class is used to reproduce corruption scenarios in the GVFS virtual projection.
+    /// </summary>
+    [Category(Categories.GitCommands)]
+    [TestFixtureSource(typeof(GitRepoTests), nameof(GitRepoTests.ValidateWorkingTree))]
+    public class CorruptionReproTests : GitRepoTests
+    {
+        public CorruptionReproTests(Settings.ValidateWorkingTreeMode validateWorkingTree)
+            : base(enlistmentPerTest: true, validateWorkingTree: validateWorkingTree)
+        {
+        }
+
+        [TestCase]
+        public void ReproCherryPickRestoreCorruption()
+        {
+            // Based on FunctionalTests/20170206_Conflict_Source
+            const string CherryPickCommit = "51d15f7584e81d59d44c1511ce17d7c493903390";
+            const string StartingCommit = "db95d631e379d366d26d899523f8136a77441914";            
+
+            this.ControlGitRepo.Fetch(StartingCommit);
+            this.ControlGitRepo.Fetch(CherryPickCommit);
+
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -365,6 +365,119 @@ public IEnumerable<string> GetAllModifiedPaths()
...
+            string[] pathspecs = null;
+            if (!string.IsNullOrEmpty(pathspec))
+            {
+                pathspecs = pathspec.Split('\0');
+            }
+
+            // Query all staged files in one call using --name-status -z.
+            // Output format: "A\0path1\0M\0path2\0D\0path3\0"
+            GitProcess.Result result = gitProcess.DiffCachedNameStatus(pathspecs);
+            if (result.ExitCodeIsSuccess && !string.IsNullOrEmpty(result.Output))
```
KeithIsSleeping, line 426: **HIGH: Fail-open — `AddStagedFilesToModifiedPaths` returns success when `git diff --cached` fails** (3/3 models agree)

`success` is initialized to `true` (line 382) and is only set to `false` inside the `ExitCodeIsSuccess` block. If `DiffCachedNameStatus` fails (non-zero exit — index locked, repo not ready, disk error), the method falls through and returns `true` with `addedCount = 0`.

The caller in `HandlePrepareForUnstageRequest` converts this to a success response, the hook allows the unstage to proceed, and no paths were added to ModifiedPaths. This re-opens the exact corruption this PR is meant to fix.

**Recommendation:** Treat `!result.ExitCodeIsSuccess` as hard failure:
```csharp
else if (!result.ExitCodeIsSuccess)
{
    success = false;
    // Log result.Errors for diagnostics
}
```

GVFS/GVFS.Hooks/Program.Unstage.cs
```diff
@@ -0,0 +1,131 @@
...
+            // Skip flags (--staged, -S, --source=, -s, etc.) and extract paths
+            List<string> paths = new List<string>();
+            bool pastDashDash = false;
+
+            for (int i = 2; i < args.Length; i++)
+            {
+                string arg = args[i];
+
+                if (arg.StartsWith("--git-pid="))
+                    continue;
```
KeithIsSleeping: **MEDIUM: Tree-ish arguments incorrectly forwarded as pathspecs** (2/4 models agree)

The parser adds every non-option token before `--` to the path list. For `git checkout HEAD -- foo.txt`, the extracted list becomes `["HEAD", "foo.txt"]`. If the repo has a file or directory matching the tree-ish name, the mount side will add/hydrate unrelated staged paths.

At OS repo scale (2.5M files), accidental scope widening is dangerous.

**Recommendation:** Parse `checkout` and `restore` separately — strip the checkout tree-ish (`HEAD`) and any `--source` value before collecting pathspecs.

GVFS/GVFS.Hooks/Program.Unstage.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Extracts pathspec arguments from a restore --staged command.
+        /// Returns null-separated pathspecs, or empty string for all staged files.
+        /// </summary>
+        private static string GetRestorePathspec(string command, string[] args)
+        {
```
KeithIsSleeping: **MEDIUM: `--pathspec-from-file` ignored, falls back to all-staged scan** (2/4 models agree)

`GetRestorePathspec()` skips all `-`-prefixed args without interpreting them. `git restore --staged --pathspec-from-file=list.txt` is treated as no pathspec, which downstream means "all staged files" (`DiffCachedNameStatus(null)`). On the OS repo, a narrowly-scoped unstage becomes a full-index walk.

**Recommendation:** Explicitly support `--pathspec-from-file`/`--pathspec-file-nul`, or fail closed with an error when those options are detected.

PR #1862, opened 2025-09-10
Enforce minimum Windows API set, and loosen SDK requirement checks

mjcheetham:
VFS for Git requires the Windows Projected File System ('ProjFS') to function, which was first introduced, in its non-final API design, with Windows 10 build 16299 (aka '1709', 'Redstone 3', 'Fall Creators Update').

Since 240f73ee5b987fa4e400251229582b64b3924d56, we now use the latest installed version of the Windows SDK to build the native projects. To prevent accidental use of APIs introduced after 16299, we should set the `TargetPlatformMinVersion` property to 10.0.16299.0.

Whilst we are here, we should also remove the now overzealous checks for the specific Windows SDK in the Build.bat script.

Note that the GitHub `windows-2025` runner does _not advertise_ that it has Windows SDK build 16299 installed, but it must do for the current builds to work.


dscho (approved):
Great!

PR #1866, opened 2025-09-12
ci: avoid unnecessary builds and test runs

dscho:
Sadly, VFSforGit's CI and PR builds are not exactly cheap. Not only in terms of combined runtime (where VFSforGit's 3 hours 45 minutes is not even close runner-up to [`git/git`'s 8.5 hours (!!!)](https://github.com/git/git/actions/runs/17628515153/usage). A Functional Test job typically runs for 35 minutes to test on x86_64, and an even worse 1 hour and 15 minutes on arm64. The wall-clock time for a CI or PR build of VFSforGit therefore weighs around 1 hour and 20 minutes.

This hurts a lot when trying to fast-track a change from a PR targeting `master` to a full VFSforGit release because not only does the entire build run _again_ on the (tree-same) `master` after merging, the next step (thanks to the original VFSforGit maintainer's decision to go for a Git Flow-like setup) requires the build to run _a third-time_ on the (also tree-same) PR merge commit.

Let's try to reuse build and test results from previous, successful runs on what is essentially the same code.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,6 +28,7 @@ jobs:
     steps:
     - name: Look for prior successful runs
       id: check
+      if: github.event.inputs.git_version == ''
       uses: actions/github-script@v7
       with:
         github-token: ${{secrets.GITHUB_TOKEN}}
@@ -72,6 +73,7 @@ jobs:
             for (const run of runs.workflow_runs.sort(demoteInProgressToEnd)) {
               if (head_sha !== run.head_sha && tree_id !== run.head_commit?.tree_id) continue
               if (context.runId === run.id) continue // do not wait for the current run to finish ;-)
+              if (run.event === 'workflow_dispatch') continue // skip runs that were started manually: they can override the Git version
 
               if (run.status === 'in_progress') {
                 // poll until the run is done


.github/workflows/build.yaml
```diff
@@ -22,11 +22,53 @@ jobs:
...
+
     steps:
+    - name: look for prior successful runs
+      id: check
+      uses: actions/github-script@v7
+      with:
+        github-token: ${{secrets.GITHUB_TOKEN}}
+        script: |
+          /*
+            * Look for any successful run for the same commit. No need to run it again, right?
```
mjcheetham: Except that we may want to be running this against the same commit, but using a different version of `microsoft/git` which can be specified as an input on `workflow_dispatch`:

```
  workflow_dispatch:
    inputs:
      git_version:
        description: 'Microsoft Git version tag to include in the build (leave empty for default)'
        required: false
        type: string

env:
  GIT_VERSION: ${{ github.event.inputs.git_version || 'v2.50.1.vfs.0.1' }}
```

Can we make the workflow _always_ run anew if this was `workflow_dispatch`?

.github/workflows/build.yaml
```diff
@@ -22,11 +22,53 @@ jobs:
   validate:
     runs-on: windows-2025
     name: Validation
+    outputs:
+      skip: ${{ steps.check.outputs.result }}
+
     steps:
+    - name: look for prior successful runs
```
mjcheetham:
```diff
- - name: look for prior successful runs
+ - name: Look for prior successful runs
```

.github/workflows/build.yaml
```diff
@@ -64,24 +94,50 @@ jobs:
...
+              else continue
+
+              const { data: jobs } = await github.rest.actions.listJobsForWorkflowRun({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                run_id: run.id
+              })
+              for (const job of jobs.jobs) {
+                if (job.conclusion !== 'success') continue
+                const jobLabel = job.name.startsWith('Build') ? 'build' : 'functional_test'
```
mjcheetham: Matching by job name feels fragile.. is there another, more robust way?

.github/workflows/build.yaml
```diff
@@ -64,24 +94,50 @@ jobs:
...
-            return ''
+            return result
           } catch (e) {
             core.error(e)
             core.warning(e)
           }
 
     - name: Checkout source
-      if: steps.check.outputs.result == ''
+      if: ${{ !(contains(fromJSON(steps.check.outputs.result).skip.build.*.label, 'Debug') && contains(fromJSON(steps.check.outputs.result).skip.build.*.label, 'Release')) }}
```
mjcheetham: I'm starting to feel like this is becoming overly complicated, hard to debug, and fragile to make modifications to. I'm having a hard time parsing the expression here.

I'd rather change the workflow and/or job structure to make this sort of 'skip on existing' logic work. Would splitting out the FT job to its own workflow help, or make it worse?

Can we take some of this (nearly-)repeated inline logic and expression and put it in an Action instead or in this workflow directly?

dscho: Unfortunately, the logic would have to remain complicated, no matter how much effort we put into it, due to the limitations of GitHub Actions' workflow YAML schema.

So I dropped the complex logic to reuse successful jobs even from failed runs.

dscho:
<details><summary>Range-diff</summary>


* -:  -------- >  1:  281c6b93 GVFS.Installers: update Inno Setup compiler
* 1:  906fc4fb =  2:  2568836a ci(build): use Unix line endings
* 2:  e34c0fbe =  3:  c5f1ca3c ci: explicitly restrict `GITHUB_TOKEN` permissions
* 3:  164e0b54 =  4:  5ca1b997 ci: avoid failing fast in the matrix jobs
* 4:  7617fd94 !  5:  a08ef4c4 ci: skip running if the same commit has already seen a successful run

   ``````diff
   @@ .github/workflows/build.yaml: jobs:
    +      skip: ${{ steps.check.outputs.result }}
    +
         steps:
   -+    - name: look for prior successful runs
   ++    - name: Look for prior successful runs
    +      id: check
    +      uses: actions/github-script@v7
    +      with:
   @@ .github/workflows/build.yaml: jobs:
    +              repo: context.repo.repo,
    +              per_page: 500,
    +              workflow_id,
   -+              status: 'success',,
   ++              status: 'success',
    +            });
    +            for (const run of runs.workflow_runs) {
    +              if (head_sha === run.head_sha) return true
   ``````

* 5:  bf4b95fb =  6:  2edc6c18 ci: skip also for successful runs for _tree-same_ commits
* 6:  eb9177be =  7:  4ee8c1c9 ci: when skipping runs, do not skip the required jobs
* 7:  4d4639b8 !  8:  50995009 ci: when skipping runs, report the previous successful one

   ``````diff
   @@ Commit message
    
     ## .github/workflows/build.yaml ##
    @@ .github/workflows/build.yaml: jobs:
   -               status: 'success',,
   +       uses: actions/github-script@v7
   +       with:
   +         github-token: ${{secrets.GITHUB_TOKEN}}
   ++        result-encoding: string
   +         script: |
   +           /*
   +             * It would be nice if GitHub Actions offered a convenient way to avoid running
   +@@ .github/workflows/build.yaml: jobs:
   +               status: 'success',
                 });
                 for (const run of runs.workflow_runs) {
    -              if (head_sha === run.head_sha) return true
   ``````

* 8:  effc5cb3 <  -:  -------- ci: do re-use successful jobs even from failing runs
* 9:  0f0f7543 !  9:  a9fc9a57 ci: when looking for successful workflow jobs, wait for in-progress ones

   ``````diff
   @@ Metadata
    Author: Johannes Schindelin <johannes.schindelin@gmx.de>
    
     ## Commit message ##
   -    ci: when looking for successful workflow jobs, wait for in-progress ones
   +    ci: when looking for successful workflow runs, wait for in-progress ones
    
   -    The runtime of the Functional Test jobs is long enough (35 minutes for
   -    x86_64, 1 hour 15 minutes for ARM64) that it is worth waiting for them
   -    to finish before kicking off a matrix build that tests the exact same
   -    code.
   +    The Functional Test jobs takes long enough (35 minutes for x86_64, 1
   +    hour 15 minutes for ARM64) that it is worth waiting for them to finish
   +    before kicking off a matrix build that tests the exact same code.
    
        So let's do that, wait for any in-progress run that might result in
        some saved time. Granted, this spends cycles of GitHub runners waiting
   @@ Commit message
    
     ## .github/workflows/build.yaml ##
    @@ .github/workflows/build.yaml: jobs:
   +               repo: context.repo.repo,
   +               per_page: 500,
                   workflow_id,
   +-              status: 'success',
                 });
                 for (const run of runs.workflow_runs) {
   +-              if (head_sha === run.head_sha) return run.html_url
   +-              else if (tree_id === run.head_commit?.tree_id) return run.html_url
   ++              if (head_sha !== run.head_sha && tree_id !== run.head_commit?.tree_id) continue
    +              if (context.runId === run.id) continue // do not wait for the current run to finish ;-)
    +
   -               let reason
   -               if (head_sha === run.head_sha) reason = 'the same commit'
   -               else if (tree_id === run.head_commit?.tree_id) reason = 'a tree-same commit'
   -               else continue
   - 
    +              if (run.status === 'in_progress') {
    +                // poll until the run is done
    +                const pollIntervalInSeconds = 30
   @@ .github/workflows/build.yaml: jobs:
    +                }
    +              }
    +
   -               const { data: jobs } = await github.rest.actions.listJobsForWorkflowRun({
   -                 owner: context.repo.owner,
   -                 repo: context.repo.repo,
   ++              if (run.status === 'success') return run.html_url
   +             }
   +             return ''
   +           } catch (e) {
   ``````

* 10:  9b6e6a7b ! 10:  65755726 ci: when looking for successful workflow jobs, use a smart order

   ``````diff
   @@ .github/workflows/build.yaml: jobs:
    +            const rank = (a) => (a.status === 'in_progress' ? 0 : (head_sha === a.head_sha ? 2 : 1))
    +            const demoteInProgressToEnd = (a, b) => (rank(b) - rank(a))
    +            for (const run of runs.workflow_runs.sort(demoteInProgressToEnd)) {
   +               if (head_sha !== run.head_sha && tree_id !== run.head_commit?.tree_id) continue
                   if (context.runId === run.id) continue // do not wait for the current run to finish ;-)
     
   -               let reason
   ``````

</details>

.github/workflows/build.yaml
dscho: Sadly, no. GitHub's REST API reports the job's labels as they are _displayed_, not as they are _defined_.

.github/workflows/build.yaml
dscho: Oops. I forgot about this...

dscho: Here you go: https://github.com/microsoft/VFSForGit/pull/1866/commits/45580086fb34570b48ef2f4b8d8c990aacf47450

PR #1867, opened 2025-09-12
Warn-and-continue if gvfs/config download fails

tyrielv:
The verbs 'clone', 'mount', 'prefetch', and 'cache-server' each perform a query to gvfs/config endpoint for two purposes: to get the list of allowed client versions for the server, and to get the list of preconfigured cache servers. Currently, if the query fails then the verb fails.

The most common reason for the query to fail appears to be transient network/authentication issues, making automount in particular more flaky than desired.

Mount and Prefetch verbs should always have a cache server configuration available already, and Clone and Cache-Server both have command-line options to allow specifying the cache server configuration.

This change modifies the verbs so that if a cache server url is already available (from local config or command-line option) then the verb will warn that the version could not be checked and continue instead of failing if the query to gvfs/config fails.

### DIFF
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -494,8 +494,10 @@ namespace GVFS.CommandLine
         }
 
         /// <summary>
-        /// Attempts to query the GVFS config endpoint. If it fails, but a valid cache server URL is available, returns a null object.
-        /// If neither is available, . If config endpoint fails, skips minimum version check and warns (unless skip-version-check is set).
+        /// Attempts to query the GVFS config endpoint. If successful, returns the config.
+        /// If the query fails but a valid fallback cache server URL is available, returns null and continues.
+        /// (A warning will be logged later.)
+        /// If the query fails and no valid fallback is available, reports an error and exits.
         /// </summary>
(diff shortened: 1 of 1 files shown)


GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -493,6 +493,48 @@ protected RetryConfig GetRetryConfig(ITracer tracer, GVFSEnlistment enlistment,
             return retryConfig;
         }
 
+        /// <summary>
+        /// Attempts to query the GVFS config endpoint. If it fails, but a valid cache server URL is available, returns a null object.
+        /// If neither is available, . If config endpoint fails, skips minimum version check and warns (unless skip-version-check is set).
```
huskydawg: A bit jumbled here

tyrielv: Fixed

derrickstolee (approved):
For `scalar clone`s, we had to do something similar by stopping the fallback to vanilla Git partial clones since the `gvfs/config` endpoint was flaky or caused issues if the sslbackend wasn't configured correctly. It's now a hard failure if the `gvfs/config` endpoint doesn't work at clone time. The only other time to use that endpoint is if the user asks to reconfigure the cache server info from the server.

So this seems like a valuable parallel change.

PR #1871, opened 2025-09-18
Fix `gvfs clone` version check when revision is included in the version string

dscho:
As reported by @tyrielv, `gvfs clone` would fail with https://github.com/microsoft/VFSForGit/releases/tag/v1.0.25260.1 when the repository to clone specifies a minimal VFSforGit version in its `gvfs/config` endpoint. The reason is that with the version bump from .NET SDK v3 to v9, the version string included in the assembly now contains the suffix `+<commit-SHA>`, which cannot be parsed by `System.Version`:

```
Cannot clone @ C:\path: System.FormatException: Input string was not in a correct format.
   at System.Version.VersionResult.SetFailure(ParseFailureKind failure, String argument)
   at System.Version.TryParseComponent(String component, String componentName, VersionResult& result, Int32& parsedComponent)
   at System.Version.TryParseVersion(String version, VersionResult& result)
   at System.Version.Parse(String input)
   at System.Version..ctor(String version)
   at GVFS.CommandLine.GVFSVerb.TryValidateGVFSVersion(GVFSEnlistment enlistment, ITracer tracer, ServerGVFSConfig config, String& errorMessage, Boolean& errorIsFatal)
   at GVFS.CommandLine.GVFSVerb.ValidateClientVersions(ITracer tracer, GVFSEnlistment enlistment, ServerGVFSConfig gvfsConfig, Boolean showWarnings)
   at GVFS.CommandLine.CloneVerb.Execute()
```

@mjcheetham provided a quick work-around in #1870 to opt out of the new behavior and report the version _without_ that suffix again; Here is a PR that prepares the code to handle such versions gracefully, to open up the door to opt back in to the new behavior (and let `gvfs version` report an even more informative output than right now).

### DIFF
--- a/GVFS/GVFS/CommandLine/GVFSVerb.cs
+++ b/GVFS/GVFS/CommandLine/GVFSVerb.cs
@@ -933,7 +933,11 @@ You can specify a URL, a name of a configured cache server, or the special names
                     return true;
                 }
 
-                Version currentVersion = new Version(ProcessHelper.GetCurrentProcessVersion());
+                string recordedVersion = ProcessHelper.GetCurrentProcessVersion();
+                // Work around the default behavior in .NET SDK 8 where the revision ID
+                // is appended after a '+' character, which cannot be parsed by `System.Version`.
+                int plus = recordedVersion.IndexOf('+');
+                Version currentVersion = new Version(plus < 0 ? recordedVersion : recordedVersion.Substring(0, plus));
                 IEnumerable<ServerGVFSConfig.VersionRange> allowedGvfsClientVersions =
                     config != null
                     ? config.AllowedGVFSClientVersions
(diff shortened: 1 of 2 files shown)


GVFS/GVFS/CommandLine/GVFSVerb.cs
```diff
@@ -933,7 +933,11 @@ private bool TryValidateGVFSVersion(GVFSEnlistment enlistment, ITracer tracer, S
                     return true;
                 }
 
-                Version currentVersion = new Version(ProcessHelper.GetCurrentProcessVersion());
+                string recordedVersion = ProcessHelper.GetCurrentProcessVersion();
```
mjcheetham, line 936: Are we sure this is the only place that `GetCurrentProcessVersion` is used? We should make sure that all callsites know how to handle the `+`.

dscho: I was coming from the other angle: The underlying problem is that `System.Version` cannot handle the new format. So I looked for all instances where `new Version()` is called, and I found two. One, I remove in this PR, the other one I [Postel](https://en.wikipedia.org/wiki/Postel's_Law)ize.

Would you agree that this is Good Enough™️?

PR #1874, opened 2025-09-24
ReadObjectHook: suppress false positive

dscho:
Currently, CodeQL mis-identifies a couple instances of the usually-legitimate "User-controlled data may not be null terminated" problem.

In these instances, CodeQL thinks that `packet_txt_read()` fails to NUL-terminate (not actually "null"... tsk, tsk) the string. But it totally does! Here is the current definition of that function:

	size_t packet_txt_read(char *buf, size_t count, FILE *stream)
	{
		size_t len;

		len = packet_bin_read(buf, count, stream);
		if (len && buf[len - 1] == '\n')
		{
			len--;
		}

		buf[len] = 0;
		return len;
	}

The `buf[len] = 0` statement guarantees that the string is NUL-terminated.

### DIFF
--- a/GVFS/GVFS.ReadObjectHook/main.cpp
+++ b/GVFS/GVFS.ReadObjectHook/main.cpp
@@ -84,13 +84,13 @@ int main(int, char *argv[])
     DisableCRLFTranslationOnStdPipes();
 
     packet_txt_read(packet_buffer, sizeof(packet_buffer));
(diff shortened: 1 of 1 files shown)


dscho:
> I wonder if CodeQL is being stupid and expects to see `buf[len] = NUL;` instead... 🤦‍♂️

I guess so, but I'd expect it to expect `'\0'`. Same problem, though, it's just inviting false positives.

PR #1875, opened 2025-09-24
Handle errors on background prefetch start

tyrielv:
In some environments "gvfs" may not be properly resolved by the shell when running the background prefetch process, so this change uses the current process path instead.

If the background prefetch process fails to start, the error is now displayed to the user with remediation steps and clone continues to the next step.


huskydawg (approved):
Looks good

PR #1878, opened 2025-10-06
Update heuristic for pre-prefetch commit loading

tyrielv:
# Description
This change is intended to improve the heuristic for loading commits when a prefetch has not completed, either because the clone was run with `--no-prefetch` or because `gvfs.trustPackIndexes` is set to `false`.

Previously the heuristic for loading commits before prefetch has completed is:
- When a commit is requested to be downloaded, record the tree it points to.
- At most once per 5 minutes
- When a tree is requested to be downloaded that was previously recorded and it has been at least 5 minutes since the last time a commit
- Then download the commit.

This works for the most basic case where a user clones a repo, then checks out a branch other than the default. It also limits over-downloading when history or commands are run that load many commits but need few objects from them.

However, it doesn't work well when these are combined (eg a history command is run first, then a checkout), or when multiple commands are run in far-apart sections of the commit graph in less than a 5-minute period.

The new heuristic is:
- When a commit is requested to be downloaded, record the tree it points to, associated with the commit.
- When a tree is downloaded that was previously recorded, record the subtrees that it references that have not been downloaded yet and associate them with the same commit.
- When a tree is requested to be downloaded, if the number of trees associated with a commit is greater than N, then download the commit.

N is currently set to 200, which is approximately the number of trees that are downloaded in the first 3 seconds of attempting to checkout a branch where many trees would be downloaded without batching.

Downloading the entire tree graph for a commit as a pack takes about 1 second, so this should limit the amount of time to download all trees for a commit to about 4 seconds, but also be unlikely to download the commit packs for history/blame operations that only need a few trees per commit.

# Changes
* Change the heuristic as described above in `InProcessMount.cs`
* Add support to Native layer for determining which subtrees of a tree are missing from the local cache.

### DIFF
--- a/GVFS/GVFS.Common/Git/LibGit2Repo.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2Repo.cs
@@ -1,5 +1,6 @@
 ﻿using GVFS.Common.Tracing;
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.Common/Git/LibGit2Repo.cs
```diff
@@ -147,6 +148,84 @@ public virtual bool TryCopyBlob(string sha, Action<Stream, long> writeAction)
...
+                || treeHandle == IntPtr.Zero)
+            {
+                return Array.Empty<string>();
+            }
+
+            try
+            {
+                if (Native.Object.GetType(treeHandle) != Native.ObjectTypes.Tree)
+                {
+                    return Array.Empty<string>();
```
ShiningMassXAcc, line 170: Should this be a more severe failure case rather than returning none (assuming we expect this not to happen).

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -27,6 +28,11 @@ public class InProcessMount
         private const int MaxPipeNameLength = 250;
         private const int MutexMaxWaitTimeMS = 500;
 
+        // This is value chosen based on tested scenarios to limit the required download time for
+        // all the trees in a commit to a few seconds, while not downloading pack files for commits
```
ShiningMassXAcc: nit: I'd just change 'few seconds' to exactly what you are observing (the 3 second / 1 second numbers etc)

GVFS/GVFS.Common/Git/LibGit2Repo.cs
tyrielv: No, I don't want to force callers to check the object type before calling this.

GVFS/GVFS.Mount/InProcessMount.cs
tyrielv: Updated comments

PR #1943, opened 2026-04-13
Improve resiliency of git worktree commands

tyrielv:
## Summary

`MountNewWorktree()` in the post-command hook silently discarded the exit code of `gvfs mount`, leaving worktrees unprojected when concurrent adds caused mount contention.

## Review Guide

**3 files changed** — read in this order:

### 1. `GVFS.Hooks/Program.cs` (smallest change)
- Added `GetHookExitCode()` — parses `--exit_code=N` from hook args. Generic infrastructure any post-hook handler can use. Placed alongside `GetHookType` and `GetGitCommand`.

### 2. `GVFS.Hooks/Program.Worktree.cs` (product fix)
- **`RunWorktreePostCommand`** — now checks exit code. Skips `add`/`move` post-processing if git itself failed; always runs `remove` cleanup.
- **`TryMountWithRetry`** (new) — retries `gvfs mount` 2× with 100ms/250ms delays. First attempt shows console output; retries are quiet. Returns bool.
- **`MountNewWorktree`** — captures `git checkout` exit code (was discarded). On checkout failure, emits warning with recovery command and skips mount. On mount failure, emits warning.
- **`MountMovedWorktree`** — same `TryMountWithRetry` treatment.

### 3. `WorktreeTests.cs` (test improvements)
- **Adaptive scaling** — starts at `Environment.ProcessorCount` worktrees (min 4). If concurrent adds overwhelm the primary mount ("does not appear to be mounted"), halves count and retries.
- **Pipe-based mount assertion** — probes worktree-specific named pipe directly instead of `File.Exists`. Definitive signal that ProjFS is active.
- **Barrier synchronization** — `CountdownEvent` ensures all threads launch simultaneously.
- **Diagnostics on failure** — captures directory listing, .git contents on assertion failure.

## Verification

Tested locally with 16 concurrent worktrees (`Environment.ProcessorCount` on dev machine):

| Scenario | Runs | Failures | Rate |
|----------|------|----------|------|
| Without product fix | 5 | 5 | **100%** |
| With product fix | 5 | 0 | **0%** |

**Note:** On CI runners with fewer cores (4), the lower concurrency may not trigger the race. The test still validates correctness of concurrent worktree operations and the adaptive scaling logic — the product fix is independently justified by code inspection (silent exit code discard).

CI: all 45 checks passed.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
@@ -1,9 +1,15 @@
-﻿using GVFS.FunctionalTests.Tools;
+﻿using GVFS.Common;
+using GVFS.Common.NamedPipes;
+using GVFS.FunctionalTests.Tools;
 using GVFS.Tests.Should;
 using NUnit.Framework;
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using ProcessResult = GVFS.FunctionalTests.Tools.ProcessResult;
 
 namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
 {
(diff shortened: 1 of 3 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
```diff
@@ -1,112 +1,303 @@
...
+
+        private void InitWorktreeArrays(int count, out string[] paths, out string[] branches)
+        {
+            paths = new string[count];
+            branches = new string[count];
+            for (int i = 0; i < count; i++)
+            {
+                string suffix = Guid.NewGuid().ToString("N").Substring(0, 8);
+                paths[i] = Path.Combine(this.Enlistment.EnlistmentRoot, $"test-wt-{(char)('a' + i)}-{suffix}");
+                branches[i] = $"worktree-test-branch-{(char)('a' + i)}-{suffix}";
```
KeithIsSleeping: There's a bug here, if there's more than 27 logical processors, which is common, char starts evaluating to `|` at 27 and `~` at 29

GVFS/GVFS.Hooks/Program.Worktree.cs
```diff
@@ -51,22 +51,61 @@ private static void RunWorktreePreCommand(string[] args)
...
                     break;
             }
         }
 
+        /// <summary>
+        /// Attempts to mount GVFS for a worktree, retrying on transient failures.
+        /// The first attempt shows output to the console; retries are quiet.
+        /// Returns true if mount succeeded.
+        /// </summary>
+        private static bool TryMountWithRetry(string fullPath, int maxRetries = 2)
```
KeithIsSleeping: Bug here: maxRetries can be overridden with a parameter, but a value great than 2 would crash the hook process (IndexOutOfRangeException) in the post-command hook.

GVFS/GVFS.Hooks/Program.cs
```diff
@@ -508,6 +508,26 @@ private static bool IsAlias(string command)
...
+                {
+                    if (int.TryParse(args[i].Substring("--exit_code=".Length), out int code))
+                    {
+                        return code;
+                    }
+                }
+            }
+
+            return 0;
+        }
```
KeithIsSleeping, line 531: Minor: The default here is 0 (success) do we want this to fail open?

GVFS/GVFS.Hooks/Program.Worktree.cs
```diff
@@ -51,22 +51,61 @@ private static void RunWorktreePreCommand(string[] args)
...
                 case "remove":
+                    // Always run cleanup regardless of git exit code — need to
+                    // remount if remove failed, and clean markers either way.
                     RemountWorktreeIfRemoveFailed(args);
                     CleanupSkipCleanCheckMarker(args);
                     break;
                 case "move":
-                    // Mount at the new location after git moved the directory
-                    MountMovedWorktree(args);
+                    if (gitExitCode == 0)
```
KeithIsSleeping: When 'git worktree move' fails, the pre-hook has already unmounted the old worktree, but the post-hook now does nothing, leaving the old worktree unmounted and unusable. 

The 'remove' case has RemountWorktreeIfRemoveFailed for this pattern, but 'move' does not.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
```diff
@@ -1,112 +1,303 @@
...
+            // Adaptively scale down if concurrent adds overwhelm the primary
+            // GVFS mount. CI runners with fewer resources may not handle as
+            // many concurrent git operations as a developer workstation.
+            while (true)
+            {
+                this.InitWorktreeArrays(count, out worktreePaths, out branchNames);
+                ProcessResult[] addResults = this.ConcurrentWorktreeAdd(worktreePaths, branchNames, count);
+
+                bool overloaded = addResults.Any(r =>
+                    r.ExitCode != 0 &&
```
KeithIsSleeping, line 38: The adaptive overload-detection loop discards ALL results when any single result contains 'does not appear to be mounted'. 

Mixed failure sets (one overload + one real regression) get masked.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
tyrielv: Fixed

GVFS/GVFS.Hooks/Program.Worktree.cs
tyrielv: Fixed

GVFS/GVFS.Hooks/Program.cs
tyrielv: Fixed to return null if not present, but at least for how we're using it here I do want it to fail open.

GVFS/GVFS.Hooks/Program.Worktree.cs
tyrielv: Fixed

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
tyrielv: Fixed

KeithIsSleeping (approved):

PR #1953, opened 2026-04-29
.NET 10 NativeAOT migration

tyrielv:
# .NET 10 NativeAOT Migration

Retarget VFSForGit from .NET Framework 4.7.1 to .NET 10 with NativeAOT self-contained deployment. Native binaries eliminate the .NET runtime dependency (~20 MB GVFS.exe, 36.7 MB installer vs 107 MB self-contained).

Based on [this branch](https://github.com/microsoft/VFSForGit/compare/master...miniksa:user/miniksa/net10-nativeaot?expand=1&w=1) by @miniksa

## Commit walkthrough

| # | Commit | Summary | Files | What to look for |
|---|--------|---------|-------|------------------|
| 1 | [`66334474`](https://github.com/microsoft/VFSForGit/pull/1953/changes/663344744c1fc30ddd85a2d196b94f2e9b6276d0) | Retarget to .NET 10 + NativeAOT | 33 | TFM change, package updates, Directory.Build.props/targets for AOT, Build.bat 3-step build, CI yaml, installer, script output paths |
| 2 | [`f2df4049`](https://github.com/microsoft/VFSForGit/pull/1953/changes/f2df4049) | Replace System.Reflection.Assembly | 7 | Assembly.Location ΓåÆ Environment.ProcessPath with null guard, AppDomain.FriendlyName for process name |
| 3 | [`c5b7ca75`](https://github.com/microsoft/VFSForGit/pull/1953/changes/c5b7ca75) | Replace changed .NET 10 APIs | 10 | NamedPipeServerStreamAcl, DirectorySecurity instance methods, Uri.EscapeDataString, UseShellExecute, CountingStream for DeflateStream truncation |
| 4 | [`bb787608`](https://github.com/microsoft/VFSForGit/pull/1953/changes/bb787608) | Replace WMI with kernel32 P/Invoke | 1 | WindowsPhysicalDiskInfo: DeviceIoControl replaces System.Management (COM interop incompatible with AOT) |
| 5 | [`c36bc237`](https://github.com/microsoft/VFSForGit/pull/1953/changes/c36bc237) | ProjFS null guard | 1 | FileSystemCallbacks: null-coalesce triggeringProcessImageFileName (ProjFS v2.1.0 returns null for kernel ops) |
[...]

PR #1954, opened 2026-05-01
HealthTests: diagnostic output for flaky test investigation

tyrielv:
## Problem
HealthTests.PlaceholdersChangeHealthScores and ModifiedPathsChangeHealthScores fail intermittently in CI with ArgumentOutOfRangeException — the test parses gvfs health output by line index without validating the output first.

## Changes
- Capture raw `gvfs health` output and validate line count before index-based parsing
- On failure, include the full output in the assertion message for root-cause analysis
- CI matrix restricted to slice 7 only (where HealthTests lives) for faster iteration

## Context
Observed in [run 25186133967](https://github.com/microsoft/VFSForGit/actions/runs/25186133967) on branch tyrielv/net10-pr (3 of 4 failed jobs).

ADO: #62098957


tyrielv:
Root cause found: stdout truncation in WaitForExit(timeout). Fix applied to net10-pr (#1953) instead.

PR #1955, opened 2026-05-01
functional tests: diagnostics for flaky CheckoutCleansUpTombstones

tyrielv:
## Problem

\WindowsTombstoneTests.CheckoutCleansUpTombstones\ fails intermittently on arm64 CI (slice 9, Debug config). The failure manifests as a mount failure (\GVFS did not mount\) at line 46 after the test modifies \ModifiedPaths.dat\ while GVFS is unmounted.

**Failed run:** https://github.com/microsoft/VFSForGit/actions/runs/25186133967
**ADO work item:** AB#62098958
**Parent deliverable:** AB#61580834

## Hypothesis

The test calls \UnmountGVFS()\ then immediately reads/writes \ModifiedPaths.dat\. The unmount command may return before the GVFS process fully releases file locks, causing the \ReadAllText\/\WriteAllText\ to either fail with \IOException\ or silently corrupt the file, which then prevents the subsequent mount from succeeding.

### Diagnostic logging (\WindowsTombstoneTests.cs\)
- Timing for unmount/mount operations  
- File metadata (existence, size, timestamps) before file access
- Content verification after write (read-back check)
- Retry with 500ms backoff for file I/O (up to 10 attempts) to handle transient locks
- On mount failure: dump ModifiedPaths content, list GVFS log files, tail latest mount log
- Uses \TryMountGVFS\ with explicit output capture instead of hard-assert \MountGVFS\

### CI restriction (\unctional-tests.yaml\)
- Matrix restricted to \rm64 / Debug / slice 9\ to focus runner time on reproducing the failure
- **Must revert before merge**

## Next steps

Once diagnostics confirm the root cause:
1. Add proper file-lock wait after unmount (or retry in \MountGVFS\)
2. Restore full CI matrix
3. Merge fix


tyrielv:
Root cause found: stdout truncation in WaitForExit(timeout). Fix applied to net10-pr (#1953) instead.

PR #1958, opened 2026-05-04
Non-disruptive upgrade: staged install with in-process file copy

tyrielv:
## Motivation

To roll out upgrades automatically (e.g. via fleet management or scheduled tasks), the installer needs to work without disrupting active GVFS mounts. Today, upgrading requires unmounting all repos — interrupting builds, IDE indexing, and any in-progress work. This PR adds a staged upgrade mode so users can be moved to newer versions transparently.

### Installer (Setup.iss)

New `/STAGEIFMOUNTED=true` flag for silent installs (default: false — existing behavior preserved). Interactive installs show a dialog when mounts are detected. When staging:

1. Files are written to `{app}\PendingUpgrade\` instead of replacing in-place
2. `GVFS.Service.exe` is replaced directly (brief stop/start — mounts unaffected)
3. A `.ready` marker is written after all files are staged (guards against partial writes from interrupted installs)

`CloseApplications=no` prevents Restart Manager from killing GVFS processes. The clean upgrade path now force-kills processes if `unmount-all` fails to clean up.

### Service (PendingUpgradeHandler.cs, RequestHandler.cs)

Completes the staged upgrade when no mounts are running:

- **On service start:** checks before automount
- **After unmount:** timer-based debounce (5s) fires once after the last repo unmounts

The upgrade uses atomic `File.Move` in two phases:
1. Back up old files to `PreviousVersion\` (with `.phase1-complete` marker for crash recovery)
2. Move staged files to install dir

Safety: defers if GVFS.Mount processes are still running. Rejects PendingUpgrade without `.ready` marker. Crash mid-Phase-1 restores backed-up files and retries.

### CI (upgrade-tests.yaml)

Five test scenarios: staging upgrade, clean upgrade, double staging, staging-then-clean, and mount safety deferral.

### DIFF
--- a/.github/workflows/upgrade-tests.yaml
+++ b/.github/workflows/upgrade-tests.yaml
@@ -1,0 +1,290 @@
+name: Upgrade Tests
+
+on:
+  workflow_call:
+    inputs:
+      skip:
+        description: 'URL of a previous successful run; if non-empty, all steps are skipped'
+        required: false
+        type: string
+        default: ''
+      lkg_release_tag:
+        description: 'Tag of the last known good release to upgrade from (default: latest release)'
+        required: false
+        type: string
+        default: ''
+
+permissions:
+  contents: read
+  actions: read
+
+jobs:
+  upgrade_test:
+    runs-on: windows-2025
+    name: Upgrade
+    timeout-minutes: 30
+
+    strategy:
+      matrix:
+        configuration: [ Debug ]
+        scenario:
+          - staging-upgrade
+          - clean-upgrade
+          - double-staging
+          - staging-then-clean
+          - mount-safety-deferral
+      fail-fast: false
+
+    steps:
+    - name: Skip this job if there is a previous successful run
+      if: inputs.skip != ''
+      id: skip
+      uses: actions/github-script@v9
+      with:
+        script: |
+          core.info(`Skipping: There already is a successful run: ${{ inputs.skip }}`)
+          return true
+
+    # -- Artifacts --
+
+    - name: Download LKG release installer
+      if: steps.skip.outputs.result != 'true'
+      shell: pwsh
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: |
+        $tag = "${{ inputs.lkg_release_tag }}"
+        if (-not $tag) {
+          $tag = gh api repos/microsoft/VFSForGit/releases/latest --jq '.tag_name'
+          Write-Host "Auto-detected latest release: $tag"
+        }
+        New-Item -ItemType Directory -Path gvfs-lkg -Force | Out-Null
+        gh release download $tag --repo microsoft/VFSForGit --pattern "SetupGVFS*.exe" --dir gvfs-lkg
+
+    - name: Download Git installer
+      if: steps.skip.outputs.result != 'true'
+      uses: actions/download-artifact@v8
+      with:
+        name: MicrosoftGit
+        path: git
+
+    - name: Download current GVFS installer
+      if: steps.skip.outputs.result != 'true'
+      uses: actions/download-artifact@v8
+      with:
+        name: GVFS_${{ matrix.configuration }}
+        path: gvfs-new
+
+    # -- Setup --
+
+    - name: Install Git
+      if: steps.skip.outputs.result != 'true'
+      shell: cmd
+      run: git\install.bat
+
+    - name: Enable ProjFS
+      if: steps.skip.outputs.result != 'true'
+      shell: pwsh
+      run: |
+        $feature = Get-WindowsOptionalFeature -Online -FeatureName Client-ProjFS
+        if ($feature.State -ne 'Enabled') {
+          Enable-WindowsOptionalFeature -Online -FeatureName Client-ProjFS -NoRestart
+        }
+
+    # -- Test Execution --
+
+    - name: Run upgrade test - ${{ matrix.scenario }}
+      if: steps.skip.outputs.result != 'true'
+      shell: pwsh
+      run: |
+        $ErrorActionPreference = 'Stop'
+
+        $lkgInstaller = (Get-ChildItem gvfs-lkg\SetupGVFS*.exe).FullName
+        $newInstaller = (Get-ChildItem gvfs-new\SetupGVFS*.exe).FullName
+        $installDir = "C:\Program Files\VFS for Git"
+        $testRepo = "https://dev.azure.com/gvfs/ci/_git/ForTests"
+        $enlistment = "C:\gvfs-upgrade-test"
+
+        function Install-GVFS($installer, [string[]]$extraArgs = @()) {
+          $logDir = "C:\temp\gvfs-install-logs"
+          New-Item -ItemType Directory -Path $logDir -Force | Out-Null
+          $logFile = Join-Path $logDir "gvfs-$(Get-Date -Format 'yyyyMMdd-HHmmss').log"
+          $allArgs = @("/VERYSILENT", "/SUPPRESSMSGBOXES", "/NORESTART", "/LOG=$logFile") + $extraArgs
+          Write-Host "Installing: $installer $($allArgs -join ' ')"
+          # Start without -Wait: Inno Setup launches child processes
+          # (e.g. GVFS.Service.UI) that stay running, causing -Wait to
+          # hang. Instead, wait only for the installer process itself.
+          $proc = Start-Process -FilePath $installer -ArgumentList $allArgs -PassThru
+          $proc.WaitForExit()
+          if ($proc.ExitCode -ne 0) {
+            Get-Content $logFile -Tail 30 -ErrorAction SilentlyContinue
+            throw "Installer failed with exit code $($proc.ExitCode)"
+          }
+          Write-Host "Installed successfully"
+        }
+
+        function Assert-ServiceRunning {
+          $svc = sc.exe query GVFS.Service 2>&1 | Select-String "STATE"
+          if ($svc -notmatch "RUNNING") { throw "GVFS.Service is not running: $svc" }
+        }
+
+        function Mount-TestRepo {
+          if (Test-Path $enlistment) {
+            & "$installDir\gvfs.exe" mount $enlistment 2>&1 | Write-Host
+          } else {
+            & "$installDir\gvfs.exe" clone $testRepo $enlistment 2>&1 | Write-Host
+          }
+          if ($LASTEXITCODE -ne 0) { throw "Mount/clone failed" }
+          $mountProc = Get-Process -Name "GVFS.Mount" -ErrorAction SilentlyContinue
+          if (-not $mountProc) { throw "No GVFS.Mount process after mount" }
+          return $mountProc.Id
+        }
+
+        function Assert-MountAlive($expectedPid) {
+          $proc = Get-Process -Id $expectedPid -ErrorAction SilentlyContinue
+          if (-not $proc -or $proc.ProcessName -ne "GVFS.Mount") {
+            throw "Mount process $expectedPid is no longer running"
+          }
+          # Verify the mount is functional by accessing a file
+          $readmePath = Join-Path $enlistment "src\Readme.md"
+          if (-not (Test-Path $readmePath)) {
+            throw "Mount is running but cannot access $readmePath"
+          }
+        }
+
+        function Unmount-TestRepo {
+          & "$installDir\gvfs.exe" unmount $enlistment 2>&1
+          Start-Sleep -Seconds 3
+        }
+
+        function Restart-Service {
+          sc.exe stop GVFS.Service | Out-Null
+          Start-Sleep -Seconds 10
+          sc.exe start GVFS.Service | Out-Null
+          Start-Sleep -Seconds 10
+          Assert-ServiceRunning
+        }
+
+        function Assert-PendingUpgrade($expected) {
+          $exists = Test-Path "$installDir\PendingUpgrade"
+          if ($exists -ne $expected) {
+            throw "PendingUpgrade directory: expected=$expected, actual=$exists"
+          }
+        }
+
+        # =============================================
+        # Test scenarios
+        # =============================================
+
+        switch ("${{ matrix.scenario }}") {
+
+          "staging-upgrade" {
+            Write-Host "=== Scenario: Staging upgrade e2e ==="
+            # Install LKG, mount, staging upgrade, unmount, verify completion
+            Install-GVFS $lkgInstaller
+            Assert-ServiceRunning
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Installers/Setup.iss
```diff
@@ -156,6 +173,50 @@ begin
...
+          Log('WaitForServiceProcessToExit: sc query failed, assuming service is gone');
+          break;
+        end;
+      Attempts := Attempts + 1;
+      Log('WaitForServiceProcessToExit: Waiting for service to stop (attempt ' + IntToStr(Attempts) + ')');
+      Sleep(1000);
+    end;
+  DeleteFile(TempFile);
+  if Attempts >= 30 then
+    Log('WaitForServiceProcessToExit: Timed out waiting for service to stop');
```
dscho: Might make sense to log the contents of the `~scquery` file (_before_ deleting it 😉) in that case? And: did you want to error out if the service could not be stopped?

GVFS/GVFS.Installers/Setup.iss
```diff
@@ -677,22 +743,124 @@ begin
...
   SetNuGetFeedIfNecessary();
-  if ConfirmUnmountAll() then
+
+  // Check for mounted repos by querying the service, and also check for
+  // running GVFS processes (a mount can be running without being registered
+  // in the service's repo-registry, e.g., after a reinstall).
+  HasMounts := False;
+  if ExecWithResult('gvfs.exe', 'service --list-mounted', '', SW_HIDE, ewWaitUntilTerminated, ResultCode, Repos) then
     begin
-      if ExpandConstant('{param:REMOUNTREPOS|true}') = 'true' then
```
dscho, line 686: For smooth upgrade paths, it might be nice to keep that `/REMOUNTREPOS` parameter kind of working by silently following the `/KEEPMOUNTED` code path?

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,280 @@
...
+    # -- Artifacts --
+
+    - name: Download LKG release installer
+      if: steps.skip.outputs.result != 'true'
+      shell: pwsh
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: |
+        New-Item -ItemType Directory -Path gvfs-lkg -Force | Out-Null
+        gh release download "${{ inputs.lkg_release_tag }}" --repo ${{ github.repository }} --pattern "SetupGVFS*.exe" --dir gvfs-lkg
```
dscho: Hmm. This workflow is triggered on push or on workflow_dispatch, and that can be done on forks, too. They don't have that GitHub Release, though. So we should probably hard-code the `--repo microsoft/VFSforGit` parameter here instead of using `${{ github.repository }}`.

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,280 @@
...
+          if (-not $mountProc) { throw "No GVFS.Mount process after mount" }
+          return $mountProc.Id
+        }
+
+        function Assert-MountAlive($expectedPid) {
+          $proc = Get-Process -Id $expectedPid -ErrorAction SilentlyContinue
+          if (-not $proc -or $proc.ProcessName -ne "GVFS.Mount") {
+            throw "Mount process $expectedPid is no longer running"
+          }
+        }
```
dscho, line 153: Maybe also access a file that needs to exist?

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,280 @@
...
+            Restart-Service
+            Assert-PendingUpgrade $false
+            Write-Host "PASS: Staging upgrade completed"
+          }
+
+          "clean-upgrade" {
+            Write-Host "=== Scenario: Clean upgrade (traditional) ==="
+            Install-GVFS $lkgInstaller
+            Assert-ServiceRunning
+            Mount-TestRepo | Out-Null
```
dscho: Why `| Out-Null`? Do we expect this to be failing, and want to suppress the error message? I'd recommend to at least show error messages, for better debuggability.

dscho (approved):

.github/workflows/upgrade-tests.yaml
tyrielv: Feedback taken

GVFS/GVFS.Installers/Setup.iss
tyrielv: Feedback taken

.github/workflows/upgrade-tests.yaml
tyrielv: Feedback taken

.github/workflows/upgrade-tests.yaml
tyrielv: Feedback taken

GVFS/GVFS.Installers/Setup.iss
tyrielv: Thanks for pointing this out. `/REMOUNTREPOS` is true by default, so anything existing that specified a value for it would have specified false - ie kill the existing mounts without giving them a graceful unmount, and expecting to manually remount afterwards if needed. 

So I don't think we want /REMOUNTREPOS=true to silently follow the /KEEPMOUNTED code path - that could break the assumption of existing scripts that their upgrade is complete if the installer succeeds. I also need to make KEEPMOUNTED default to false for the same reason (ie make it opt-in instead of opt-out when run with /VERYSILENT from a script).

I've also realized KEEPMOUNTED and REMOUNTREPOS are too similarly named - I'm going to rename KEEPMOUNTED to STAGEIFMOUNTED. If STAGEIFMOUNTED=true and REMOUNTREPOS=false (both non-default), then STAGEIFMOUNTED will activate and REMOUNTREPOS will have nothing to do (nothing was unmounted, nothing to remount).

GVFS/GVFS.Installers/Setup.iss
```diff
@@ -59,8 +59,14 @@ Name: "full"; Description: "Full installation"; Flags: iscustom;
...
 [Files]
-DestDir: "{app}"; Flags: ignoreversion recursesubdirs; Source:"{#LayoutDir}\*"
-DestDir: "{app}"; Flags: ignoreversion; Source:"{#LayoutDir}\GVFS.Service.exe"; AfterInstall: InstallGVFSService
+; Normal install: all files go to {app}, service gets AfterInstall callback
+DestDir: "{app}"; Flags: ignoreversion recursesubdirs; Source:"{#LayoutDir}\*"; Check: IsNormalInstall
+DestDir: "{app}"; Flags: ignoreversion; Source:"{#LayoutDir}\GVFS.Service.exe"; AfterInstall: InstallGVFSService; Check: IsNormalInstall
+; Staging install: most files go to {app}\PendingUpgrade, but GVFS.Service.exe
+; goes directly to {app} so the restarted service has PendingUpgradeHandler code.
+; The service is briefly stopped/restarted (mounts are independent processes).
+DestDir: "{app}\PendingUpgrade"; Flags: ignoreversion recursesubdirs; Source:"{#LayoutDir}\*"; Check: IsStagingInstall
```
mjcheetham, line 68: So with `Source:"{#LayoutDir}\*"` we also put a copy of the new GVFS.Service.exe in the PendingUpgrade directory, as well as another copy in the root? Just clarifying my understanding.

GVFS/GVFS.Service/PendingUpgradeHandler.cs
```diff
@@ -0,0 +1,309 @@
...
+        private const string ReadyMarkerFileName = ".ready";
+        private const string Phase1CompleteMarkerFileName = ".phase1-complete";
+        private const string ServiceExeName = "GVFS.Service.exe";
+        private const string MountProcessName = "GVFS.Mount";
+
+        /// <summary>
+        /// Checks for and applies a pending staged upgrade. Returns true if
+        /// an upgrade was applied (caller should proceed with normal startup).
+        /// </summary>
+        public static bool TryApplyPendingUpgrade(ITracer tracer)
```
mjcheetham: We return `true` or `false` from this method, but the call sites don't actually do anything different. The comment says:

> Returns true if an upgrade was applied (caller should proceed with normal startup).

GVFS/GVFS.Service/PendingUpgradeHandler.cs
```diff
@@ -0,0 +1,309 @@
...
+
+            tracer.RelatedInfo($"{nameof(PendingUpgradeHandler)}: Pending upgrade detected at {pendingUpgradeDir}");
+
+            // Don't apply if GVFS.Mount processes are still running — their
+            // executables are locked and moves would fail. Upgrade will be
+            // retried on next service start when no mounts are active.
+            Process[] mountProcesses = Array.Empty<Process>();
+            try
+            {
+                mountProcesses = Process.GetProcessesByName(MountProcessName);
```
mjcheetham, line 93: What do we do to ensure a new mount process isn't started mid-upgrade?

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,285 @@
...
+      skip:
+        description: 'URL of a previous successful run; if non-empty, all steps are skipped'
+        required: false
+        type: string
+        default: ''
+      lkg_release_tag:
+        description: 'Tag of the last known good release to upgrade from'
+        required: false
+        type: string
+        default: 'v1.0.26098.1'
```
mjcheetham: I wonder if this can be determined from the GitHub API? We could define the LKG as the latest non-prerelease?

GVFS/GVFS.Installers/Setup.iss
tyrielv: Yes, there's a duplicate copy in PendingUpgrade that the service has to ignore/delete when doing the moves later.

.github/workflows/upgrade-tests.yaml
tyrielv: Feedback taken

GVFS/GVFS.Service/PendingUpgradeHandler.cs
tyrielv: Good point, this was left over from an earlier iteration. Removed the return value.

GVFS/GVFS.Service/PendingUpgradeHandler.cs
tyrielv: I couldn't think of a way to definitely block one from starting, so here's what I did in response to this feedback:
1. Move the 3 exes most likely to get called out first, and in last. (gvfs.hooks.exe, gvfs.exe, gvfs.mount.exe)
2. Add a 'kill process + retry' attempt if move fails

GVFS/GVFS.Installers/Setup.iss
```diff
@@ -245,6 +313,33 @@ begin
...
+  // could be replaced. Now start it with the new binary. The new service has
+  // PendingUpgradeHandler which will complete the upgrade on next restart
+  // when no mounts are running.
+  StatusText := WizardForm.StatusLabel.Caption;
+  WizardForm.StatusLabel.Caption := 'Starting GVFS.Service.';
+  WizardForm.ProgressGauge.Style := npbstMarquee;
+
+  try
+    Log('StagingUpdateService: Starting service with new binary');
+    if not Exec(ExpandConstant('{sys}\SC.EXE'), 'start GVFS.Service', '', SW_HIDE, ewWaitUntilTerminated, ResultCode) then
```
KeithIsSleeping, line 331: Exec returns false only when sc.exe can't be launched. This misses the case where sc.exe runs and returns a non zero, not logging or surfacing anything.

Also in the `not Exec()` ResultCode is undefined so SysErrorMessage(ResultCode) would print "operation completed successfully"

GVFS/GVFS.Service/Handlers/RequestHandler.cs
```diff
@@ -121,5 +133,35 @@ private void TrySendResponse(
...
+            string pendingUpgradeDir = System.IO.Path.Combine(installDir, PendingUpgradeHandler.PendingUpgradeDirectoryName);
+            if (!System.IO.Directory.Exists(pendingUpgradeDir))
+            {
+                return;
+            }
+
+            // Debounce: reset the timer on each unmount so the check fires
+            // once after the last unmount settles. If multiple repos unmount
+            // in quick succession, only one upgrade attempt runs.
+            if (this.pendingUpgradeTimer == null)
```
KeithIsSleeping, line 149: Copilot flagged this:

`WindowsRequestHandler` is a singleton. `NamedPipeServer.OnNewConnection` spawns a new thread per connection ("we must not do any blocking IO on this thread"). Two concurrent unmounts can both observe `null` and both construct a `Timer` — the loser is leaked but its callback still fires, so TryApplyPendingUpgrade runs twice in parallel and races on File.Move.


Wrap the read+write under a `lock`, dispose the prior timer when replacing it. Optional belt-and-suspenders: gate `TryApplyPendingUpgrade` itself with `Interlocked.CompareExchange`

GVFS/GVFS.Installers/Setup.iss
```diff
@@ -647,12 +709,21 @@ begin
   case CurStep of
     ssInstall:
       begin
-        UninstallService('GVFS.Service', True);
+        if not KeepMountsRunning then
+          UninstallService('GVFS.Service', True);
       end;
     ssPostInstall:
       begin
+        if KeepMountsRunning then
```
KeithIsSleeping, line 717: For `ssPostInstall` (line 69) and also `AfterInstallHook` (here):

If the last mount happens to exit during this install window, the 5s debounce timer fires before `.ready` is written, TryApplyPendingUpgrade skips, and nothing pokes the service afterward. The staged files then sit there until a future unmount or service restart.

Could we either start the service only after `.ready` is written, or have `ssPostInstall` send a "check pending now" pipe message after creating the marker?

KeithIsSleeping (approved):

GVFS/GVFS.Service/PendingUpgradeHandler.cs
mjcheetham: Only other thing we could have done is add a cross-process semaphore that these programs check for, first thing, and exit fast with a 'simple upgrade in progress' message and unique exit code 42 (for filtering out of any logs/events).
The GVFS.Service could create this blocking semaphore and release upon completion.

If not a semaphore, then some other sentinel or marker file `%ProgramFiles%\VFS for Git\.upgrade-in-progress`.

tyrielv: That would help a little bit for future rollouts, not for the first one (since the processes starting wouldn't have code to check for the semaphore). I think it will be unnecessary with the kill/retry attempt, but will keep it in mind if telemetry shows any issues.

PR #1959, opened 2026-05-06
Fix flaky WriteWithoutClose/CreateFileWithoutClose functional tests

tyrielv:
## Summary

Fix GC-dependent race condition causing flaky failures in `StatusTests.WriteWithoutClose()` and `StatusTests.CreateFileWithoutClose()`.

## Problem

These tests intentionally open file handles without closing them to validate that `git status` works with locked files. The handles were created as local variables inside `OpenFileAndWriteWithoutClose`/`CreateFileWithoutClose` with no reference kept — making them eligible for GC immediately. If the GC finalized the handles before TearDown ran `git reset --hard`, the reset would succeed (no lock errors) while the control repo might still have errors, causing a flaky error mismatch in TearDown.

## Fix

- Change `FileSystemRunner.CreateFileWithoutClose` and `OpenFileAndWriteWithoutClose` to return `IDisposable` instead of `void`
- `SystemIORunner` returns the actual `FileStream`/`StreamWriter` handle, with an explicit `Flush()` to ensure data is on disk
- `GitRepoTests` helpers return a `CompositeDisposable` wrapping handles for both GVFS and control repos
- `StatusTests` uses `using` blocks — handles stay alive during the assertion scope and are deterministically disposed before TearDown runs

This eliminates the GC race: both repos have unlocked files when `reset --hard` runs in TearDown, so errors match (both 0).

## Testing

- Built GVFS.FunctionalTests project successfully (C# compilation clean)
- The fix is structural — deterministic disposal replaces non-deterministic GC finalization

### DIFF
--- a/GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
+++ b/GVFS/GVFS.FunctionalTests/FileSystemRunners/BashRunner.cs
@@ -304,12 +304,12 @@ namespace GVFS.FunctionalTests.FileSystemRunners
             return long.Parse(this.RunProcess(statCommand));
         }
 
-        public override void CreateFileWithoutClose(string path)
+        public override IDisposable CreateFileWithoutClose(string path)
         {
             throw new NotImplementedException();
         }
 
-        public override void OpenFileAndWriteWithoutClose(string path, string data)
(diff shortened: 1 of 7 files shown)


dscho (approved):
Nice!

PR #1960, opened 2026-05-07
release.yml: update the AzDO release pipeline YAML

mjcheetham:
Previously the .azure-pipelines/release.yml file was not being used for release builds, but instead a private repository containing the real YAML (in Azure Repos) was being used.

Let's bring the actual pipeline YAML into the main repo on GitHub. This is what we do in other projects like microsoft/git and git-ecosystem/git-credential-manager, and allows us to keep build changes in sync between GH and AzDO.


dscho (approved):
❤️

PR #1962, opened 2026-05-11
Update GVFS version to 2.0 in release pipeline

tyrielv:
Upgrade to .NET 10 is a major change, and excluding bundled ProjFS driver for older systems is potentially breaking.


dscho (approved):
Yeah!

mjcheetham (approved):

PR #1964, opened 2026-05-11
Handle case-only file and directory renames in FastFetch on Windows

erickulcyk:
Previously, FastFetch detected case-only renames (e.g. TestFolder -> testfolder) but suppressed them with a warning, leaving the working directory with stale casing. This change makes FastFetch propagate case renames to disk, matching the behavior of regular git checkout.

Changes:
- DiffTreeResult: Add SourcePath property to carry old-cased path for directory case renames
- DiffHelper: Detect case-only renames and flow them through the pipeline instead of suppressing. Track case-renamed directories to properly filter child operations in FlushStagedQueues
- CheckoutStage: Perform two-step directory rename (old -> temp -> new) to work around Directory.Move() failing on case-only renames on Windows
- Update unit tests to verify SourcePath, delete counts, and directory operation filtering for case renames
- Update functional test to verify casing with ignoreCase: false


derrickstolee:
@copilot Could you help me understand how directory structures interact with parallel file movement in FastFetch? Is there concern that a parent directory will be renamed in one thread while a child file will want to be renamed in another? What protections in this codebase prevent that from happening?

tyrielv (approved):

PR #1966, opened 2026-05-12
azure-pipelines: move to new MicroBuild pool

mjcheetham:
Move from the `VSEngSS-MicroBuild2022-1ES` pool to the new `VSEng-MicroBuildVSStable` pool, which has Windows Server 2025 + Visual Studio 2026 images instead.

This is targeting the `releases/shipped` branch. Once merged will port back to `master.`


dscho (approved):
Hopefully we can stay with this pool for a while.

PR #1973, opened 2026-05-14
build: collapse chains of skipped workflow runs

mjcheetham:
The validate job's 'Look for prior successful runs' step finds a prior successful run that is commit- or tree-same and records its URL via `core.notice` and as the `validate.outputs.skip` output, which downstream jobs surface to users.

When the prior run was itself a skip-run (i.e. its only successful work was recording a notice pointing at a yet earlier run), we end up with a chain of skip-runs each pointing at the previous skip-run, rather than at the run that actually built and tested the tree. Following such a chain by hand to find the real build is tedious.

Bottom out that chain in the script: after identifying a candidate run, inspect its Validation job's annotations for the same 'Skipping: There already is a successful run: <url>' notice. If present, fetch the referenced run, verify it is still completed and successful, and repeat until we reach a run that is not itself a skip-run. Use that URL for both the notice and the job output, so consumers always see the real underlying run.

Also add console logs along the way (initial match, each chain hop, stop conditions) to make future debugging straightforward, and grant the workflow checks: read so the script can call `checks.listAnnotations`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,6 +21,7 @@ on:
 permissions:
   contents: read
   actions: read
+  checks: read
 
 env:
   GIT_VERSION: ${{ github.event.inputs.git_version || 'v2.53.0.vfs.0.7' }}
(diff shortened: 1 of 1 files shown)


mjcheetham:
This issue was noticed with the failure of this check run in `microsoft/git`: https://github.com/microsoft/git/actions/runs/25853366997/job/75966092385?pr=914

That workflow followed the skip-run to _another skip-run_, which didn't have the required artifacts (to run the functional tests). Whilst we could also update `microsoft/git`'s workflow here to do the skip-chain collapsing, it makes sense to do this _here_ instead.

dscho (approved):

derrickstolee (approved):

PR #1975, opened 2026-05-15
Fix: --unmount-all now triggers pending staged upgrade

tyrielv:
## Problem

When an installer runs while mounts are active, it stages new files to
`PendingUpgrade\` instead of replacing files in-place. The service is
supposed to apply the upgrade when all mounts are unmounted.

This works correctly for `gvfs unmount` because each unmount sends an
`UnregisterRepoRequest` to the service, which triggers
`TryDeferredPendingUpgradeCheck` in `RequestHandler`.

However, `gvfs service --unmount-all` sets `SkipUnregister = true` on
each unmount (to preserve automount registration), so **no message reaches
the service** and the staged upgrade is never applied — it sits in
`PendingUpgrade\` indefinitely until the next service restart.

## Fix

**New `PendingUpgradeCheckRequest` named pipe message** (3 files, ~65 lines):

1. **`NamedPipeMessages.cs`** — Added `PendingUpgradeCheckRequest` with
   header, `FromMessage`/`ToMessage`, and `Response` (follows existing
   `EnableAndAttachProjFSRequest` pattern).

2. **`RequestHandler.cs`** — Handle the new message by calling the existing
   `TryDeferredPendingUpgradeCheck` (same debounced timer that individual
   unmounts use) and returning a success response.

3. **`ServiceVerb.cs`** — After the `--unmount-all` loop completes, send
   `PendingUpgradeCheckRequest` to the service pipe. Placed **before** the
   failure exit path so partial unmount failures still trigger the check.
   Notification failure is treated as a warning (unmounts themselves succeeded).

The service-side `PendingUpgradeHandler.TryApplyPendingUpgrade` already
checks for running `GVFS.Mount` processes and defers if any are present,
so sending the notification even after partial unmount failures is safe.

## Test

Added `unmount-all-triggers-upgrade` scenario to the CI upgrade test matrix
(`upgrade-tests.yaml`). It follows the same pattern as `staging-upgrade`
but uses `gvfs service --unmount-all` instead of per-repo unmount, then
polls for `PendingUpgrade\` directory removal to confirm the upgrade
applied without needing a service restart.

### DIFF
--- a/.github/workflows/upgrade-tests.yaml
+++ b/.github/workflows/upgrade-tests.yaml
@@ -33,6 +33,7 @@ jobs:
           - double-staging
           - staging-then-clean
           - mount-safety-deferral
+          - unmount-all-triggers-upgrade
       fail-fast: false
 
     steps:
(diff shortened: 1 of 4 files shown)


tyrielv:
Closing for now - this approach doesn't work when upgrading from older version to the first version that has this.

PR #1976, opened 2026-05-15
Apply staged upgrade when mount processes exit

tyrielv:
## Problem

When the installer runs with mounts active, it stages new files to
`PendingUpgrade\`. The service applies this upgrade when all mounts are
gone — but the trigger relied on `UnregisterRepoRequest`, which only
`gvfs unmount` sends.

`gvfs service --unmount-all` sets `SkipUnregister=true` (to preserve
automount registration), so **no message reached the service** and staged
upgrades sat in `PendingUpgrade\` indefinitely until the next service
restart.

## Solution: Process Exit Monitor

The installer always replaces `GVFS.Service.exe` in-place (even during
staging), so the running service is always the new version. Instead of
requiring the client `gvfs.exe` (which may be an old version) to send a
new message, the service proactively monitors mount process exits:

- **`PendingUpgradeMonitor`** — registers `Process.Exited` callbacks on
  all `GVFS.Mount` processes. When one exits, a debounced timer (5s)
  fires and calls `TryApplyPendingUpgrade`. If new mounts started, it
  re-registers. When the upgrade completes or no pending upgrade remains,
  the monitor stops itself. Zero polling cost — purely event-driven.

- **`PendingUpgradeHandler`** — now returns an `UpgradeResult` enum
  (`NoPending`, `Applied`, `DeferredMountsRunning`, `NotReady`,
  `Failed`) so callers can react. Adds a static lock to prevent
  concurrent apply attempts from startup, RequestHandler, and monitor.

- **`GVFSService`** — starts the monitor when the startup upgrade check
  is deferred due to running mounts. Disposes on shutdown.

The existing `RequestHandler` unmount trigger is preserved as
belt-and-suspenders for the normal `gvfs unmount` path.

## Test

New `unmount-all-triggers-upgrade` scenario in the CI upgrade test
matrix. Installs LKG, mounts, stages new version, uses `--unmount-all`
(which runs the old LKG `gvfs.exe`), and verifies `PendingUpgrade\`
is consumed by the monitor without a service restart.

### DIFF
--- a/GVFS/GVFS.Service/PendingUpgradeHandler.cs
+++ b/GVFS/GVFS.Service/PendingUpgradeHandler.cs
@@ -14,8 +14,9 @@ namespace GVFS.Service
     /// When the installer runs with mounts active, it stages new files to
     /// {installDir}\PendingUpgrade\ instead of replacing files in-place.
     /// This class applies the upgrade when no GVFS.Mount processes are
-    /// running — either on service start (before automount) or after a
-    /// repo unmount (via deferred check from RequestHandler).
+    /// running — either on service start (before automount), after a
+    /// repo unmount (via deferred check from RequestHandler), or when
+    /// PendingUpgradeMonitor detects all mount processes have exited.
     ///
     ///   1. Move old files from install dir → PreviousVersion\
     ///   2. Move new files from PendingUpgrade\ → install dir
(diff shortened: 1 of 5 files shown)


derrickstolee (approved):

GVFS/GVFS.Service/PendingUpgradeHandler.cs
```diff
@@ -36,6 +37,9 @@ public static class PendingUpgradeHandler
         private const string Phase1CompleteMarkerFileName = ".phase1-complete";
         private const string ServiceExeName = "GVFS.Service.exe";
         private const string MountProcessName = "GVFS.Mount";
+        private const string MountExeName = "GVFS.Mount.exe";
+
+        private static readonly object ApplyLock = new object();
```
winstliu, line 42: Given this is now on .NET 10, use [Lock](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-13#new-lock-object) instead? Same in PendingUpgradeMonitor.

tyrielv: Turns out there's like 20 other cases like this in the existing code, so I'm going to check this one in as-is then open another PR to switch them all at once.

PR #1978, opened 2026-05-19
Fix handle leak in GetFinalPathName

tyrielv:
The fileHandle obtained from CreateFileW in GetFinalPathName (common.windows.cpp) was never closed after calling GetFinalPathNameByHandleW, leaking a kernel handle on every invocation.

This adds CloseHandle(fileHandle) immediately after GetFinalPathNameByHandleW returns, before the error check — the handle is no longer needed once the API call completes.


huskydawg (approved):
looks like an old issue!

PR #1979, opened 2026-05-19
hooks: resolve enlistment root for worktrees outside enlistment tree

tyrielv:
## Summary

All native hooks (virtual-filesystem, read-object, post-index-changed) and managed hooks (GVFS.Hooks) resolved the primary enlistment root by walking up from CWD looking for a `.gvfs/` directory. Worktrees placed outside the enlistment directory tree (e.g. in a temp directory or as a sibling) could not use git commands — hooks failed with *'must be run from inside a GVFS enlistment'*.

## Fix

Added a worktree fallback path: when `.gvfs` walk-up fails, walk up looking for a `.git` file (indicating a linked worktree), then resolve the primary enlistment root through the gitdir chain — first from the `gvfs-enlistment-root` marker file, then by deriving from `commondir`. The resolved root is validated by checking that `.gvfs/` exists.

### Native hooks (`common.windows.cpp`)
- Extract `ReadFirstLine`, `Utf8ToWide`, `TryParseGitFile` helpers
- Refactor `GetWorktreePipeSuffix` to use shared helpers
- Add `TryResolveFromWorktree` returning both enlistment root and pipe suffix
- `GetGVFSPipeName` tries `.gvfs` walk-up first, then worktree fallback

### Managed hooks (`Program.cs`)
- After `TryGetGVFSEnlistmentRoot` fails, try `TryGetWorktreeInfo` → `GetEnlistmentRoot()` before exiting
- Validate `.gvfs` exists at the resolved root

## Testing
- 3 unit tests for `GetEnlistmentRoot` (marker file, SharedGitDir fallback, marker-preferred)
- Functional test creating worktree in temp directory, verifying git status from root and subdirectory, file projection, and commits
- All 806 unit tests pass

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
@@ -191,6 +191,83 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             }
         }
 
+        [TestCase]
+        public void WorktreeOutsideEnlistmentTree()
+        {
+            string suffix = Guid.NewGuid().ToString("N").Substring(0, 8);
+            string tempDir = Path.Combine(Path.GetTempPath(), $"gvfs-remote-wt-{suffix}");
+            string worktreePath = Path.Combine(tempDir, "wt");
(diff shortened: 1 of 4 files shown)


dscho (approved):
It's a lot of added code, but I agree that it is necessary (the alternative would be to shell out to `git.exe` to determine the top-level directory, but that's prohibitively slow).

PR #1981, opened 2026-05-21
Eliminate conhost.exe from git.exe and hooks process launches

ShiningMassXAcc:
## Summary

GVFS launches git.exe processes with `CreateNoWindow=true` (or `CREATE_NO_WINDOW` in native code). Despite the name, this flag tells Windows to create a **new hidden console** for each child process, which allocates a `conhost.exe` instance. For frequent, small git operations (e.g., during prefetch), the per-process conhost creation/teardown overhead is disproportionately large.

## Changes

| File | Change | Why |
|------|--------|-----|
| `GitProcess.cs:827` | `CreateNoWindow = true` to `false` | Central git.exe wrapper (all prefetch calls route here) |
| `ProcessHelper.cs:21` | `CreateNoWindow = redirectOutput` to `false` | Generic process helper |
| `GitHooksLoader.cpp:132` | `CREATE_NO_WINDOW` to `DETACHED_PROCESS` | Native hooks loader — truly detaches from console |

## How it works

- **`CreateNoWindow = false`** with `UseShellExecute = false`: child inherits parent console state. Since GVFS runs as a service with no console, child gets no console and no conhost.
- **`DETACHED_PROCESS`** (native): explicitly detaches from any console without allocating a new one.

### Conhost elimination confirmed

| Setting | conhost.exe per git.exe |
|---------|------------------------|
| `CreateNoWindow=true` (old) | **1** conhost.exe spawned |
| `CreateNoWindow=false` (new) | **0** — eliminated |

### Benchmark (3 runs x 50 git.exe invocations)

| Metric | Old | New | Improvement |
|--------|-----|-----|-------------|
| Total (50 processes) | 3905ms | 2949ms | **956ms saved (24.5%)** |
| Per-process | 78.1ms | 59.0ms | **19.1ms saved** |

For a prefetch spawning 20-50 git.exe processes, this eliminates ~380-960ms of pure conhost overhead per cycle.

## Reference

- Same pattern as ToolLocationTransformer PR 15071136 (merged)
- ADO 62385282

### DIFF
--- a/GVFS/GVFS.Common/ProcessHelper.cs
+++ b/GVFS/GVFS.Common/ProcessHelper.cs
@@ -18,7 +18,16 @@ namespace GVFS.Common
             processInfo.RedirectStandardOutput = redirectOutput;
             processInfo.RedirectStandardError = redirectOutput;
             processInfo.WindowStyle = ProcessWindowStyle.Hidden;
-            processInfo.CreateNoWindow = redirectOutput;
+
+            // CreateNoWindow=false avoids allocating a hidden conhost.exe per child
+            // process. When redirectOutput is true, I/O goes through pipes so no
+            // console is needed. When redirectOutput is false, the child inherits the
+            // parent's console handles — this works when the parent has a console
+            // (e.g., GVFS.Hooks invoked from a terminal), but output is silently lost
+            // when the parent has no console (e.g., service context). This is
+            // acceptable because CreateNoWindow=true would only send that output to
+            // an invisible hidden console instead.
+            processInfo.CreateNoWindow = false;
             processInfo.Arguments = args;
 
             return Run(processInfo);
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -815,16 +815,24 @@ namespace GVFS.Common.Git
             return this.InvokeGitAgainstDotGitFolder($"-c pack.threads=1 -c repack.packKeptObjects=true multi-pack-index repack --object-dir=\"{gitObjectDirectory}\" --batch-size={batchSize} --no-progress");
         }
 
-        public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, bool redirectStandardError, string gitObjectsDirectory, bool usePreCommandHook)
+        public Process GetGitProcess(string command, string workingDirectory, string dotGitDirectory, bool useReadObjectHook, string gitObjectsDirectory, bool usePreCommandHook)
         {
             ProcessStartInfo processInfo = new ProcessStartInfo(this.gitBinPath);
             processInfo.WorkingDirectory = workingDirectory;
             processInfo.UseShellExecute = false;
             processInfo.RedirectStandardInput = true;
             processInfo.RedirectStandardOutput = true;
-            processInfo.RedirectStandardError = redirectStandardError;
+            processInfo.RedirectStandardError = true;
             processInfo.WindowStyle = ProcessWindowStyle.Hidden;
-            processInfo.CreateNoWindow = true;
+
+            // CreateNoWindow=false avoids allocating a hidden conhost.exe per child
+            // process. This is safe because both stdout and stderr are redirected via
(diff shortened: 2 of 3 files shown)


GVFS/GVFS.Common/ProcessHelper.cs
```diff
@@ -18,7 +18,7 @@ public static ProcessResult Run(string programName, string args, bool redirectOu
             processInfo.RedirectStandardOutput = redirectOutput;
             processInfo.RedirectStandardError = redirectOutput;
             processInfo.WindowStyle = ProcessWindowStyle.Hidden;
-            processInfo.CreateNoWindow = redirectOutput;
+            processInfo.CreateNoWindow = false;
```
wilbaker, line 30: Will always setting `CreateNoWindow` to `false` break callers who want output redirected?

I saw some discussion here: https://learn.microsoft.com/en-us/answers/questions/671607/writing-to-stdout-with-detached-process that child processes require a console to write to stdout.

ShiningMassXAcc: Good question. When `redirectOutput = true`, .NET creates anonymous pipes for stdout/stderr via `CreatePipe()` before calling `CreateProcess`. The child writes to pipe handles, not a console — so no console is needed for redirected I/O.

The linked discussion is about `DETACHED_PROCESS` with *non-redirected* stdout, which is a different scenario. Here with `UseShellExecute = false` and `RedirectStandardOutput = true`, the handles are pipes regardless of `CreateNoWindow`.

That said, this change in ProcessHelper is broader than strictly needed for the prefetch fix. The essential fix is in `GitProcess.cs` (the central git.exe wrapper). Happy to revert this file and scope down to just `GitProcess.cs` if you'd prefer to keep ProcessHelper conservative.

tyrielv: +1, I'm concerned this will break output forwarding for hooks. It makes sense for git.exe calls launched by gvfs.mount.exe, but not for ones called by gvfs.hooks.exe or gvfs.exe.

ShiningMassXAcc: Good point — reverted ProcessHelper.cs. The fix is now scoped to just GitProcess.cs (used by gvfs.mount.exe / service path) and GitHooksLoader.cpp (native, uses DETACHED_PROCESS with explicit STARTF_USESTDHANDLES). ProcessHelper stays as-is for hooks and CLI contexts.

tyrielv: GitHooksLoader.cpp is the caller for Gvfs.Hooks.exe, so it should be reverted too.

tyrielv: Call chain is user->git.exe->GitHooksLoader.exe->Gvfs.Hooks.exe->possibly more calls to git or gvfs

ShiningMassXAcc: Thanks for the call chain context — that makes sense. Reverted GitHooksLoader.cpp too. The fix is now a single-line change in GitProcess.cs only, targeting the gvfs.mount.exe service path where there's no console to inherit.

tyrielv: I did a bunch of testing and all the original changes were fine for the scenarios I was worried about. I've restored them.

tyrielv (approved):

wilbaker (approved):

ShiningMassXAcc:
All checks green and both approvals in — could one of you merge when ready? Thanks!

tyrielv:
> All checks green and both approvals in — could one of you merge when ready? Thanks!

@ShiningMassXAcc 
It's still blocked - because I'm the one that pushed most recently, and Will doesn't have approval permissions. We need another approver with permissions, or for you to re-push the changes I contributed.

dscho (approved):

dscho:
The only time when I _think_ this might come back to bite us is the scenario where users want to type in their (e.g. HTTPS) credentials via the Console. But since Azure Repos moved away from passwords/PATs, and VFSforGit really only works with Azure Repos, I think we're on the safe side.

PR #1984, opened 2026-05-22
ci: add retry for artifact download steps

tyrielv:
Transient failures in `actions/download-artifact` are a frequent source of flaky CI runs — I see these several times a week, always succeeding on manual re-run.

This adds a single automatic retry for each `download-artifact` step using the `continue-on-error` + `outcome == 'failure'` pattern:

- **functional-tests.yaml**: retry for Git installer, GVFS installer, and functional tests drop (3 download steps)
- **upgrade-tests.yaml**: retry for Git installer and GVFS installer (2 download steps)

**How it works:** Each download step gets `id` + `continue-on-error: true`. A duplicate step follows that only runs when the first attempt failed (`steps.<id>.outcome == 'failure'`). No new dependencies, no third-party actions.

Example of a recent failure this would fix: https://github.com/microsoft/VFSForGit/actions/runs/26310137269/job/77463758324?pr=1981

### DIFF
--- a/.github/workflows/functional-tests.yaml
+++ b/.github/workflows/functional-tests.yaml
@@ -77,7 +77,19 @@ jobs:
           return true
 
     - name: Download Git installer
+      id: download-git
       if: steps.skip.outputs.result != 'true'
+      continue-on-error: true
+      uses: actions/download-artifact@v8
+      with:
+        name: ${{ inputs.git_artifact_name }}
+        path: git
+        repository: ${{ inputs.git_repository || github.repository }}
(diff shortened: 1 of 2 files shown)


derrickstolee (approved):
It would be nice if the task itself could have retries, but this "maybe just run it again" will help.

tyrielv:
> It would be nice if the task itself could have retries, but this "maybe just run it again" will help.

Agreed - that feature isn't built in to github actions though, it would require referencing a third party action.

PR #1986, opened 2026-05-22
GVFS.Service: defer and retry telemetry pipe attachment

tyrielv:
## Summary

GVFS.Service runs as SYSTEM and cannot read the user's global git config at startup, so `TelemetryDaemonEventListener` was never created and all service telemetry events were silently lost.

This PR adds deferred telemetry pipe attachment without modifying `JsonTracer` — the logic lives in two new focused classes that are reusable by other processes (e.g. GVFS.Mount).

### `BufferingTelemetryListener` (GVFS.Common)
An `EventListener` subclass that buffers `TraceEventMessage` objects in a bounded `ConcurrentQueue`. When a real listener becomes available, `ReplayAndStop()` replays buffered messages and switches to no-op mode.

### `DeferredTelemetryAttacher` (GVFS.Common)
Manages the deferred attachment lifecycle:
1. Adds a `BufferingTelemetryListener` to the tracer at construction
2. Runs a retry timer with exponential backoff (10s, 30s, 1m, 5m steady state)
3. On each retry, calls `TelemetryDaemonEventListener.CreateIfEnabled`
4. On success: adds the real listener, replays buffered messages, stops the timer

Callers can also trigger an explicit attach via `TryAttach(gitBinRoot)` when conditions change (e.g. user session becomes available).

### Service integration (GVFSService.Windows.cs)
- Creates a `DeferredTelemetryAttacher` at startup
- Starts the retry timer for cases where the telemetry collector is installed later
- On `SessionLogon`: temporarily overrides HOME to the user's profile (serialized with a lock) so git can read `.gitconfig`, then calls `TryAttach`
- Also checks for an active session at startup (handles service restart during active session)

## Files changed
- `GVFS.Common/NativeMethods.cs` — `WTSGetActiveConsoleSessionId` P/Invoke
- `GVFS.Common/Tracing/BufferingTelemetryListener.cs` — new
- `GVFS.Common/Tracing/DeferredTelemetryAttacher.cs` — new
- `GVFS.Service/GVFSService.Windows.cs` — deferred attach integration
- `GVFS.UnitTests/Common/JsonTracerDeferredTests.cs` — tests for buffering, replay, cap, and retry intervals

## Testing
- 8 new unit tests covering buffer/replay, bounded cap, stop-after-replay, retry backoff values, and null handling
- All 814 existing unit tests pass

### DIFF
--- a/GVFS/GVFS.Common/Tracing/DeferredTelemetryAttacher.cs
+++ b/GVFS/GVFS.Common/Tracing/DeferredTelemetryAttacher.cs
@@ -1,0 +1,208 @@
+﻿using System;
+using System.Threading;
+
+namespace GVFS.Common.Tracing
+{
+    /// <summary>
+    /// Manages deferred telemetry pipe attachment for processes that cannot
+    /// read the pipe config at startup (e.g. GVFS.Service running as SYSTEM,
+    /// or any process started before the telemetry collector is installed).
+    ///
+    /// Adds a <see cref="BufferingTelemetryListener"/> to the tracer at
+    /// construction time, then periodically retries creating a real
+    /// <see cref="TelemetryDaemonEventListener"/>.  On success, buffered
+    /// messages are replayed and the retry timer stops.
+    ///
+    /// Callers can also trigger an explicit attach attempt via
+    /// <see cref="TryAttach(string)"/> — e.g. on session logon when the
+    /// user's HOME is available.
+    ///
+    /// Designed for reuse by both GVFS.Service and GVFS.Mount.
+    /// </summary>
+    public class DeferredTelemetryAttacher : IDisposable
+    {
+        private readonly JsonTracer tracer;
+        private readonly BufferingTelemetryListener buffer;
+        private readonly string providerName;
+        private readonly string enlistmentId;
+        private readonly string mountId;
+        private readonly Lock attachLock = new Lock();
+
+        private Timer retryTimer;
+        private string retryGitBinRoot;
+        private int retryCount;
+        private bool attached;
+        private bool disposed;
+
+        public DeferredTelemetryAttacher(
+            JsonTracer tracer,
+            string providerName,
+            string enlistmentId,
+            string mountId)
+        {
+            this.tracer = tracer;
+            this.providerName = providerName;
+            this.enlistmentId = enlistmentId;
+            this.mountId = mountId;
+            this.buffer = new BufferingTelemetryListener();
+            tracer.AddEventListener(this.buffer);
+        }
+
+        public bool IsAttached
+        {
+            get
+            {
+                lock (this.attachLock)
+                {
+                    return this.attached;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Starts a background retry timer that periodically calls
+        /// <see cref="TryAttach"/> with the given gitBinRoot.  Uses
+        /// exponential backoff: 10s, 30s, 1m, then 5m steady state.
+        /// </summary>
+        public void StartRetryTimer(string gitBinRoot)
+        {
+            lock (this.attachLock)
+            {
+                if (this.attached || this.disposed || this.retryTimer != null)
+                {
+                    return;
+                }
+
+                this.retryGitBinRoot = gitBinRoot;
+                this.retryCount = 0;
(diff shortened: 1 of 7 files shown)


GVFS/GVFS.Common/Tracing/DeferredTelemetryAttacher.cs
```diff
@@ -0,0 +1,193 @@
...
+                {
+                    return false;
+                }
+
+                if (daemonListener == null)
+                {
+                    return false;
+                }
+
+                this.tracer.AddEventListener(daemonListener);
```
KeithIsSleeping, line 134: Could you double-check the interaction between this `AddEventListener` call and the `CreateIfEnabled` call that `JsonTracer`'s own constructor already makes (JsonTracer.cs ~line 57)? 

The doc comment says this class is designed for reuse by GVFS.Mount, where the constructor-time attach normally succeeds - that path seems even more likely to hit this.

GVFS/GVFS.Common/Tracing/DeferredTelemetryAttacher.cs
```diff
@@ -0,0 +1,193 @@
...
+            if (this.retryTimer != null)
+            {
+                this.retryTimer.Dispose();
+                this.retryTimer = null;
+            }
+        }
+
+        private void OnRetryTimer(object state)
+        {
+            bool success = this.TryAttach(this.retryGitBinRoot);
```
KeithIsSleeping: OnRetryTimer doesn't wrap TryAttach in try/catch, is this expected?

GVFS/GVFS.Service/GVFSService.Windows.cs
```diff
@@ -391,6 +428,79 @@ private void CreateAndConfigureLogDirectory(string path)
...
+                    }
+
+                    // HOME is process-wide; serialize to prevent concurrent
+                    // session events from clobbering each other's override.
+                    lock (this.telemetryHomeLock)
+                    {
+                        string oldHome = Environment.GetEnvironmentVariable("HOME");
+                        try
+                        {
+                            Environment.SetEnvironmentVariable("HOME", userProfile);
```
KeithIsSleeping: Worth a second look at the process-wide `HOME` override here? 

`Environment.SetEnvironmentVariable` mutates the **process** environment, and any other code running concurrently in the service that shells out to git (RepoRegistry mounts, maintenance steps, request handlers, the upgrade monitor) will inherit `HOME=<logged-on-user-profile>` for the duration of the lock - potentially reading the interactive user's `.gitconfig` while running as SYSTEM.

GVFS/GVFS.Common/Tracing/DeferredTelemetryAttacher.cs
tyrielv: Fixed — `OnRetryTimer` now wraps `TryAttach` in try/catch. On exception the timer stops (doesn't reschedule), but the explicit `TryAttach` path (e.g. on SessionLogon) can still succeed.

GVFS/GVFS.Common/Tracing/DeferredTelemetryAttacher.cs
tyrielv: Fixed — `DeferredTelemetryAttacher.TryAttach` now checks `tracer.HasTelemetryDaemonListener` before adding a new one. If the JsonTracer constructor already attached a daemon listener (the normal user-context path for GVFS.Mount), `TryAttach` returns true immediately without creating a duplicate. New property `HasTelemetryDaemonListener` on JsonTracer follows the same pattern as the existing `HasLogFileEventListener`.

GVFS/GVFS.Service/GVFSService.Windows.cs
tyrielv: Agreed — rewrote to avoid mutating HOME entirely. `TelemetryDaemonEventListener.CreateIfEnabled` now accepts an optional `globalConfigPath`; when provided it uses `git config --file` instead of `--global`. The service resolves the user's profile via `RunImpersonated` and passes `Path.Combine(userProfile, ".gitconfig")` directly. No process-wide environment mutation.

KeithIsSleeping (approved):

PR #1987, opened 2026-05-26
Fix flaky CreatePlaceholderTests stderr comparison

tyrielv:
The `AllowsPlaceholderCreationWhileGitCommandIsRunning` test fails intermittently because `git reset --hard` produces non-deterministic `Updating files` progress output on stderr. When one repo emits progress and the other does not, `ErrorsShouldMatch` fails.

Set `GIT_PROGRESS_DELAY=3600` on all test git invocations (in `GitProcess.InvokeProcess`) so progress output is never emitted for commands completing under an hour. This eliminates the flake at the source without skipping error validation.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tools/GitProcess.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GitProcess.cs
@@ -35,6 +35,9 @@ namespace GVFS.FunctionalTests.Tools
             }
 
             processInfo.EnvironmentVariables["GIT_TERMINAL_PROMPT"] = "0";
+            // Suppress progress output (e.g. "Updating files: 100%") which is timing-dependent
+            // and causes flaky stderr comparisons between control and GVFS repos.
+            processInfo.EnvironmentVariables["GIT_PROGRESS_DELAY"] = "3600";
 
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.FunctionalTests/Tests/GitCommands/CreatePlaceholderTests.cs
```diff
@@ -69,7 +69,7 @@ private void CheckPlaceholderCreation(string command, bool shouldAllow)
 
             this.ValidateGitCommand("--no-optional-locks status");
             resetEvent.Wait();
-            this.RunGitCommand("reset --hard");
+            this.RunGitCommand("reset --hard", ignoreErrors: true);
```
derrickstolee: This is probably fine, but I wonder if `RunGitCommand()` should set `GIT_PROGRESS_DELAY=3600` to avoid any progress output for commands that last less than an hour. That may be a more robust solution.

tyrielv: Good idea! Will update shortly.

dscho (approved):
Nice!

PR #1992, opened 2026-06-01
Fix deferred telemetry pipe attach for RDP sessions

tyrielv:
At startup, the deferred telemetry attacher only checked the physical console session via `WTSGetActiveConsoleSessionId`. On Cloud PCs and RDP-only machines, the console session has no logged-in user — the real user is in an RDP session. This meant the startup attach always failed, the retry timer could never succeed (SYSTEM has no git global config), and no `SessionLogon` event fires for already-connected users.

## Fix

Enumerate all interactive sessions (Active/Connected, session > 0) via `WTSEnumerateSessions` and try each until the pipe attaches.

## Changes

- **`CurrentUser.cs`**: Add public `GetInteractiveSessionIds` that filters `WTSEnumerateSessions` to Active/Connected sessions with ID > 0
- **`GVFSService.Windows.cs`**: Replace single console session check with `TryAttachTelemetryPipeForAnySessions` that iterates all interactive sessions; move telemetry attacher setup before `PendingUpgradeHandler` so deferred-upgrade events are buffered
- **Log levels**: Downgrade token-query failure from Error to Warning (expected on Cloud PCs where the console session has no user)
- **AOT warnings**: Use generic `Marshal.SizeOf<T>()` and `Marshal.PtrToStructure<T>()` instead of `typeof()` overloads

## Testing

- All 818 unit tests pass
- Verified locally on Cloud PC (RDP-only): telemetry pipe attaches via RDP session after console session fails
- `VFS.Service.PendingUpgradeHandler_Deferred` and `VFS.Service.PendingUpgradeHandler_Complete` confirmed in App Insights

### DIFF
--- a/GVFS/GVFS.Platform.Windows/CurrentUser.cs
+++ b/GVFS/GVFS.Platform.Windows/CurrentUser.cs
@@ -134,7 +134,7 @@ namespace GVFS.Platform.Windows
                     if (CreateEnvironmentBlock(ref environment, duplicate, false))
                     {
                         STARTUP_INFO info = new STARTUP_INFO();
-                        info.Length = Marshal.SizeOf(typeof(STARTUP_INFO));
+                        info.Length = Marshal.SizeOf<STARTUP_INFO>();
 
                         PROCESS_INFORMATION procInfo = new PROCESS_INFORMATION();
                         if (CreateProcessAsUser(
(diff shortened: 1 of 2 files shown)


GVFS/GVFS.Platform.Windows/CurrentUser.cs
```diff
@@ -193,6 +193,28 @@ public bool RunAs(string processName, string args)
...
+        /// </summary>
+        public static List<int> GetInteractiveSessionIds(ITracer tracer)
+        {
+            List<int> sessionIds = new List<int>();
+            List<WTS_SESSION_INFO> sessions = ListSessions(tracer);
+            foreach (WTS_SESSION_INFO session in sessions)
+            {
+                if (session.SessionID > 0 &&
+                    (session.State == ConnectionState.Active ||
+                     session.State == ConnectionState.Connected))
```
KeithIsSleeping: The session-state filter looks like it may be inverted: Connected sessions typically have no user token (so attach will always fail there), while Disconnected sessions usually do have a logged-in user available for impersonation, but they're currently excluded.

tyrielv: Fixed to Active || Disconnected, and added comments clarifying intent.

KeithIsSleeping (approved):

PR #1996, opened 2026-06-01
Add diagnostic logging to GetFileStreamException catch block

tyrielv:
The `GetFileStreamException` catch in `GetFileStreamHandlerAsyncHandler` silently completed the ProjFS command without logging `requestMetadata` (SHA, virtual path). The other catch blocks in the same handler (`OperationCanceledException`, generic `Exception`) all log this metadata.

This adds a `RelatedWarning` log call with the HResult value before `TryCompleteCommand`, providing diagnostic visibility while keeping the severity appropriate for expected/benign `HResult.Handle` results.


derrickstolee (approved):
Good find. Obvious solution.

PR #1998, opened 2026-06-02
fix worktree mounts using primary enlistment's .gvfs paths

tyrielv:
## Problem

Worktree mounts shared the primary enlistment's `.gvfs` metadata (placeholder database, RepoMetadata, diagnostics, corrupt objects) instead of using their own per-worktree `.gvfs` directory.

**Symptoms:** after `git reset --hard` in a worktree, fully-materialized files retained stale content because placeholder tracking was shared with the primary mount. The skip-worktree bit hid the stale content from `git status`/`git diff`.

**Root cause:** code throughout the codebase derived `.gvfs` paths from `EnlistmentRoot`, which always resolves to the primary enlistment root -- even for worktree mounts. The correct property is `DotGVFSRoot`, which resolves to the per-worktree `.gvfs` for worktree mounts.

### Product fixes
- **GVFSDatabase**: accept `dotGVFSRoot` directly instead of deriving from `EnlistmentRoot`; update all call sites
- **GitRepo**: resolve `DotGVFSRoot` at construction for corrupt object paths
- **GVFSVerb.InitializeLocalCacheAndObjectsPaths**: use `DotGVFSRoot` for RepoMetadata (affects dehydrate, prefetch)
- **DiagnoseVerb**: use `DotGVFSRoot` for .gvfs file collection and RepoMetadata
- **CacheVerb**: use `DotGVFSRoot` for RepoMetadata
- **RequiredGitConfig**: narrow parameter to `GVFSEnlistment`, use `DotGVFSRoot`
- **GVFSLogsRoot**: derive from `DotGVFSRoot` instead of re-deriving
- **GitMaintenanceStep**: check `WorkingDirectoryRoot` for existence
- **HooksInstaller**: use `WorkingDirectoryRoot` in error messages

### Rename to prevent recurrence
Rename `Enlistment.EnlistmentRoot` to `PrimaryEnlistmentRoot`, forcing every caller to explicitly choose the right path property. Audit of all ~280 references across ~75 files confirmed each usage is correct.

## Test

Added `WorktreeUsesPerWorktreePlaceholderDatabase` functional test: creates a worktree outside the enlistment tree, materializes a file, and verifies the per-worktree `VFSForGit.sqlite` contains placeholder entries.

## Validation

- Unit tests: 807 passed, 0 failed
- All projects compile clean (GVFS, GVFS.Common, GVFS.Mount, GVFS.Service, GVFS.Hooks, FastFetch, UnitTests)

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
@@ -268,6 +268,101 @@ namespace GVFS.FunctionalTests.Tests.EnlistmentPerFixture
             }
         }
 
+        [TestCase]
+        public void WorktreeUsesPerWorktreePlaceholderDatabase()
+        {
+            // Two commits where test4.txt has different content
+            const string CommitA = "5d7a7d4db1734fb468a4094469ec58d26301b59d";
+            const string CommitB = "fec239ea12de1eda6ae5329d4f345784d5b61ff9";
+            string testFileRelativePath = Path.Combine(
+                "Test_EPF_UpdatePlaceholderTests",
+                "LockToPreventUpdateAndDelete",
+                "test4.txt");
+            const string ContentAtCommitA = "TestFileLockToPreventUpdateAndDelete4 \r\n";
+            const string ContentAtCommitB = "Commit2LockToPreventUpdateAndDelete4 \r\n";
+
+            string suffix = Guid.NewGuid().ToString("N").Substring(0, 8);
+            string tempDir = Path.Combine(Path.GetTempPath(), $"gvfs-db-test-{suffix}");
+            string worktreePath = Path.Combine(tempDir, "wt");
+            string branchName = $"db-test-branch-{suffix}";
+
+            try
+            {
+                Directory.CreateDirectory(tempDir);
+
+                // 1. Checkout commitB in the primary and read the file to
+                //    create a placeholder with version B content.
+                GitHelpers.InvokeGitAgainstGVFSRepo(
+                    this.Enlistment.RepoRoot, $"checkout {CommitB}")
+                    .ExitCode.ShouldEqual(0, "checkout commitB in primary failed");
+
+                string primaryFilePath = Path.Combine(this.Enlistment.RepoRoot, testFileRelativePath);
+                File.ReadAllText(primaryFilePath).ShouldEqual(ContentAtCommitB,
+                    "Primary should have version B content");
+
+                // 2. Create a worktree at commitA and read the same file.
+                //    Without the fix, the worktree mount writes the placeholder
+                //    entry (SHA_A) to the primary's DB, overwriting SHA_B.
+                ProcessResult addResult = GitHelpers.InvokeGitAgainstGVFSRepo(
+                    this.Enlistment.RepoRoot,
+                    $"worktree add -b {branchName} \"{worktreePath}\" {CommitA}");
+                addResult.ExitCode.ShouldEqual(0,
+                    $"worktree add failed: {addResult.Errors}");
+
+                this.AssertWorktreeMounted(worktreePath, "db-test worktree");
+
+                string worktreeFilePath = Path.Combine(worktreePath, testFileRelativePath);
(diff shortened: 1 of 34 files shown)


GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
```diff
@@ -268,6 +268,82 @@ public void WorktreeOutsideEnlistmentTree()
...
+                    $"Per-worktree placeholder DB should contain Readme.md entry.\n" +
+                    $"DB path: {worktreeDbPath}\n" +
+                    $"DB contents:\n{worktreePlaceholders}");
+
+                // 7. Verify the primary's placeholder DB does NOT contain a
+                //    duplicate entry from worktree materialization.  The primary
+                //    DB tracks the primary enlistment's placeholders only.
+                string primaryDbPath = Path.Combine(
+                    this.Enlistment.DotGVFSRoot, "databases", "VFSForGit.sqlite");
+                string primaryPlaceholders = GVFSHelpers.GetAllSQLitePlaceholdersAsString(primaryDbPath);
```
KeithIsSleeping: The new test computes primaryPlaceholders but never asserts on it; the comment two lines above promises an isolation check that the test does not perform.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
```diff
@@ -268,6 +268,82 @@ public void WorktreeOutsideEnlistmentTree()
             }
         }
 
+        [TestCase]
+        public void WorktreeUsesPerWorktreePlaceholderDatabase()
```
KeithIsSleeping, line 272: The new test verifies a placeholder row exists, but doesn't capture stale content after 'git reset --hard' in a worktree.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
```diff
@@ -268,6 +268,82 @@ public void WorktreeOutsideEnlistmentTree()
...
+
+                // 4. Materialize a file in the worktree by reading it
+                string testFilePath = Path.Combine(worktreePath, "Readme.md");
+                File.Exists(testFilePath).ShouldBeTrue("Readme.md should be projected");
+                File.ReadAllText(testFilePath);
+
+                // 5. Wait for background operations to flush placeholder entries
+                //    The primary mount handles background tasks; give it time to
+                //    process the worktree's placeholder creation.
+                System.Threading.Thread.Sleep(2000);
```
KeithIsSleeping: A thread.sleep(2000) could create a flaky test/output, is there a more deterministic approach we could use here?

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
tyrielv: Fixed, removed dead code.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
tyrielv: Determinism isn't possible. Replaced with polling every 500ms up to 10 seconds so most of the time it will be faster, but also more resilient to slow cases.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/WorktreeTests.cs
tyrielv: 'git reset' isn't exactly necessary. The flow is:
1. One of the worktrees reads file version A, which creates its placeholder and updates placeholder DB
2. The other worktree reads file version B, which creates its placeholder and overwrites the placeholder DB (because the bug is that worktrees are sharing primary DB instead of using their own)
3. First worktree checks out (or resets to) a commit that has version B. GVFS compares the new version (B) to what's in its DB (also B), and determines it doesn't need to update the placeholder (which is incorrect, the placeholder still has version A from step 1).

Test updated to specifically repro this scenario.

KeithIsSleeping (approved):

PR #2002, opened 2026-06-04
Offload prefetch to mount process for warm auth

tyrielv:
## Summary

When a GVFS mount is running, `gvfs prefetch --commits` now sends a `PrefetchCommits` request to the mount process via named pipe IPC. The mount process executes the prefetch using its already-warm authentication, avoiding the slow cold-auth path (anonymous HTTP probe + `git credential` helper invocation).

If the mount is not running, not ready, or is an older version that does not recognize the `PrefetchCommits` message, the verb falls back to the existing direct-auth path transparently.

## Motivation

Authentication for prefetch is always slow — the anonymous probe + credential helper invocation adds multiple seconds. The mount process already has warm auth from startup. By offloading to the mount, we skip this cold-auth cost entirely for the common case where the repo is mounted.

## Changes

- **NamedPipeMessages.cs**: Add `PrefetchCommits` request/response message types
- **PrefetchStep.cs**: Add injectable post-fetch callback parameter to avoid re-entrant named pipe IPC when running inside the mount process  
- **InProcessMount.cs**: Add `HandlePrefetchCommitsRequest` handler that runs `PrefetchStep` synchronously on the connection thread using the mount's warm `GVFSGitObjects`
- **PrefetchVerb.cs**: Try mount offload first; fall back to direct auth only when mount is unavailable (connect failure, `MountNotReady`, or `UnknownRequest` from old mount). Mount-side failures are surfaced directly — no fallback on real errors.
- **PrefetchCommitsOffloadTests.cs**: Functional tests covering mounted prefetch, idempotent re-prefetch, unmounted fallback to direct auth, and remount scenarios

## Design Decisions

1. **Single request/response** — no streaming progress in this PR. The spinner covers the wait. Auth speedup is the primary win.
2. **Synchronous on connection thread** — connection threads are per-client, won't block other mount IPC. The file-based lock serializes with the mount's own scheduled prefetch.
3. **Post-fetch callback injection** — avoids `PrefetchStep.SchedulePostFetchJob()` opening a named pipe back to the same mount (re-entrant IPC). The mount handler directly enqueues `PostFetchStep` on its maintenance scheduler.
4. **Fallback only on unavailable** — mount-side failure means a real network/auth/pack error; falling back would duplicate work and mask the error.

## Testing

- All 818 unit tests pass
- Functional tests cover mounted, unmounted, idempotent, and remount scenarios

### DIFF
--- a/GVFS/GVFS/CommandLine/PrefetchVerb.cs
+++ b/GVFS/GVFS/CommandLine/PrefetchVerb.cs
@@ -3,11 +3,15 @@ using GVFS.Common.FileSystem;
 using GVFS.Common.Git;
 using GVFS.Common.Http;
 using GVFS.Common.Maintenance;
+using GVFS.Common.NamedPipes;
 using GVFS.Common.Prefetch;
 using GVFS.Common.Tracing;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace GVFS.CommandLine
 {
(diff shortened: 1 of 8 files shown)


GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -296,6 +339,142 @@ private void InitializeServerConnection(
...
+                    tracer.RelatedInfo("TryPrefetchBlobsViaMountProcess: Mount not running, falling back to direct prefetch");
+                    return false;
+                }
+
+                NamedPipeMessages.PrefetchBlobs.Request request = new NamedPipeMessages.PrefetchBlobs.Request
+                {
+                    Files = filesList,
+                    Folders = foldersList,
+                    HeadCommitId = headCommitId,
+                    HydrateFiles = this.HydrateFiles,
```
KeithIsSleeping: In the --hydrate flow the offload request tells the mount to hydrate (HydrateFiles=true), so files get hydrated on the mount and then again locally, and the mount-side hydration is the same ProjFS self-callback the code comment says it is avoiding.

tyrielv: Fixed

KeithIsSleeping (approved):

PR #2006, opened 2026-06-04
Use git_odb_exists instead of git_revparse_single for ObjectExists

tyrielv:
## Summary

Replaces the heavyweight `git_revparse_single` call in `LibGit2Repo.ObjectExists` with `git_odb_exists`, a purpose-built existence check that skips revparse expression parsing and `git_object` handle allocation.

### Benchmark (os.2020 enlistment, 59.7M objects, 14 packs)

| Scenario | Before (revparse) | After (odb_exists) | Speedup |
|----------|-------------------|-------------------|---------|
| Existing objects | ~800 ns/op | ~800 ns/op | ~1x |
| Missing objects | 2,800,000 ns/op | 1,320,000 ns/op | **2.1x** |

### Implementation

- Add P/Invoke bindings for `git_repository_odb`, `git_odb_exists`, `git_odb_free`, `git_oid_fromstr`
- ODB handle lazily acquired on first `ObjectExists` call, freed in `Dispose`
- Thread-safe lazy init via `Interlocked.CompareExchange` -- if two threads race on the first call, the loser frees its duplicate handle
- Falls back to `git_revparse_single` if ODB handle acquisition fails

### Full-SHA requirement and caller audit

`git_odb_exists` + `git_oid_fromstr` requires a full 40-character hex SHA (unlike `git_revparse_single` which also resolves refs and abbreviated SHAs). All callers were audited:

| Caller | Input source | Full SHA? | Action |
|--------|-------------|-----------|--------|
| `FindBlobsStage` | Blob SHAs from diff output | Yes | Uses `ObjectExists` |
| `EnumerateMissingTreeEntries` | `GitOid.ToString()` | Yes | Uses `ObjectExists` |
| `CommitAndRootTreeExists` | Tree SHA from `GitOid.ToString()` (commitish resolved via `GetTreeSha` which still uses revparse) | Yes | Uses `ObjectExists` |
| `GVFSVerb` (gitattributes) | `DiffTreeResult.TargetSha` from ls-tree | Yes | Uses `ObjectExists` |
| `GVFSVerb.TryDownloadCommit` | `refs.GetTipCommitId()` (CloneVerb), `git rev-parse HEAD` output (DehydrateVerb) | Yes | Uses `CommitAndRootTreeExists` |
| `InProcessMount` | Object SHA from download pipe | Yes | Uses `CommitAndRootTreeExists` |
| `LibGit2RepoInvoker.InitializeSharedRepo` | Hardcoded 40-char SHA | Yes | Uses `ObjectExists` |
| **`BlobPrefetcher.DownloadMissingCommit`** | **Raw `--commit` CLI arg from FastFetch (may be ref or abbreviated SHA)** | **No** | **Switched to `ObjectCanBeParsed`** |
| `LooseObjectsStep.ClearCorruptLooseObjects` | Loose object file names | Yes (but needs parse validation) | Switched to `ObjectCanBeParsed` |

### Semantic split: ObjectExists vs ObjectCanBeParsed

`git_odb_exists` only checks whether an OID is present in the object store -- it does not decompress or validate the object, and it requires a full 40-char hex SHA. Two callers need the old `git_revparse_single` behavior:

1. **`LooseObjectsStep.ClearCorruptLooseObjects`** -- needs corruption detection (revparse fails on corrupt objects, `odb_exists` does not)
2. **`BlobPrefetcher.DownloadMissingCommit`** -- receives raw CLI input that may be a ref name or abbreviated SHA

For these callers, `ObjectCanBeParsed(sha)` retains the old revparse path.

| Method | Implementation | Use case |
|--------|---------------|----------|
| `ObjectExists` | `git_odb_exists` (fast) | Existence checks with full 40-char SHAs |
| `ObjectCanBeParsed` | `git_revparse_single` (slow but resolves refs and validates) | Corruption detection, non-SHA input |

### Files changed

- `LibGit2Repo.cs` -- new P/Invokes, `ObjectExists` uses `git_odb_exists`, new `ObjectCanBeParsed`, thread-safe lazy ODB init
- `GitRepo.cs` -- expose `ObjectCanBeParsed` through the invoker
- `LooseObjectsStep.cs` -- call `ObjectCanBeParsed` instead of `ObjectExists`
- `BlobPrefetcher.cs` -- `DownloadMissingCommit` calls `ObjectCanBeParsed` instead of `ObjectExists`

### DIFF
--- a/GVFS/GVFS.Common/Git/LibGit2Repo.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2Repo.cs
@@ -3,12 +3,14 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Threading;
 
 namespace GVFS.Common.Git
 {
     public class LibGit2Repo : IDisposable
     {
         private bool disposedValue = false;
+        private IntPtr odbHandle = IntPtr.Zero;
 
         public delegate void MultiVarConfigCallback(string value);
 
@@ -104,6 +106,55 @@ namespace GVFS.Common.Git
         }
 
         public virtual bool ObjectExists(string sha)
+        {
+            IntPtr odb = this.odbHandle;
+            if (odb == IntPtr.Zero)
+            {
+                if (Native.Odb.GetOdb(out IntPtr newOdb, this.RepoHandle) != Native.ResultCode.Success)
+                {
+                    return this.ObjectExistsFallback(sha);
+                }
+
+                IntPtr existing = Interlocked.CompareExchange(ref this.odbHandle, newOdb, IntPtr.Zero);
+                if (existing != IntPtr.Zero)
+                {
+                    // Another thread won the race — free our duplicate and use theirs
+                    Native.Odb.Free(newOdb);
+                    odb = existing;
+                }
+                else
+                {
+                    odb = newOdb;
+                }
+            }
+
+            GitOid oid;
+            if (Native.Odb.OidFromStr(out oid, sha) != Native.ResultCode.Success)
+            {
+                return false;
+            }
+
+            return Native.Odb.Exists(odb, ref oid) == 1;
+        }
+
+        private bool ObjectExistsFallback(string sha)
+        {
+            IntPtr objHandle;
+            if (Native.RevParseSingle(out objHandle, this.RepoHandle, sha) != Native.ResultCode.Success)
+            {
+                return false;
+            }
+
+            Native.Object.Free(objHandle);
+            return true;
+        }
+
+        /// <summary>
+        /// Checks whether the object can be fully parsed by libgit2 (not just that it exists).
+        /// Use this when you need to detect corrupt objects. For simple existence checks,
+        /// prefer <see cref="ObjectExists"/> which is faster.
+        /// </summary>
+        public virtual bool ObjectCanBeParsed(string sha)
         {
             IntPtr objHandle;
             if (Native.RevParseSingle(out objHandle, this.RepoHandle, sha) != Native.ResultCode.Success)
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Common/Git/LibGit2Repo.cs
```diff
@@ -104,6 +105,42 @@ public virtual bool CommitAndRootTreeExists(string commitish, out string treeSha
...
+            if (this.odbHandle == IntPtr.Zero)
+            {
+                if (Native.Odb.GetOdb(out this.odbHandle, this.RepoHandle) != Native.ResultCode.Success)
+                {
+                    return this.ObjectExistsFallback(sha);
+                }
+            }
+
+            GitOid oid;
+            if (Native.Odb.OidFromStr(out oid, sha) != Native.ResultCode.Success)
```
KeithIsSleeping, line 132: ObjectExists now only works for full 40-char SHAs; refs or abbreviated SHAs (like FastFetch's raw --commit value) that revparse used to resolve will now be reported as missing.

GVFS/GVFS.Common/Git/LibGit2Repo.cs
```diff
@@ -104,6 +105,42 @@ public virtual bool CommitAndRootTreeExists(string commitish, out string treeSha
         }
 
         public virtual bool ObjectExists(string sha)
+        {
+            if (this.odbHandle == IntPtr.Zero)
```
KeithIsSleeping: The shared LibGit2Repo can run ObjectExists on multiple threads at once, and the lazy ODB-handle setup has no lock, so two first-time calls could each acquire the odb and leak one reference.

GVFS/GVFS.Common/Git/LibGit2Repo.cs
tyrielv: Fixed for FastFetch path to use rev-parse, though to be fair if you're calling FastFetch --commit you should expect to need the full commit ID.

GVFS/GVFS.Common/Git/LibGit2Repo.cs
tyrielv: Fixed

KeithIsSleeping (approved):

PR #1997, opened 2026-06-02
Remove ExtraCoverage category — run all functional tests in CI

tyrielv:
## Summary

The `ExtraCoverage` category excluded ~110 functional test methods (19 test classes) from the default test run and CI. These tests cover critical functionality - mount edge cases, dehydrate, repair, shared cache, disk layout upgrades, junctions, and more - but were never validated in CI, only via the `--extra-only` flag.

This PR removes ExtraCoverage filtering, fixes atrophied tests, and introduces infrastructure to track tests that still need work.

## Infrastructure changes

- **Remove ExtraCoverage filtering** - delete `ExtraCoverage` constant, `--extra-only` flag handling, and all `[Category(Categories.ExtraCoverage)]` annotations (19 usages)
- **Download FastFetch artifact** in `functional-tests.yaml` so FastFetch tests find the exe
- **Fix NUnitRunner slice grouping** - extend regex to include `MultiEnlistmentTests` alongside `EnlistmentPerFixture`, preventing Order-dependent tests from splitting across slices
- **Introduce `SkipInCIAttribute`** with a required reason string, replacing the bare `NeedsReactionInCI` category. Makes it clear why each test is skipped
- **Increase test slices** from 10 to 12 to distribute the additional test load
- **Resilient teardown** - `UnmountAndDeleteAll` catches `TimeoutException` from stuck unmounts and kills the GVFS.Mount process as a fallback, preventing a single hung unmount from eating 5 minutes of CI time
- **Update `AuthoringTests.md`** documentation

## Fixed tests (now running in CI)

| Test | Fix |
|------|-----|
| FastFetchTests (15 methods) | FastFetch artifact now downloaded in CI |
| ConfigVerbTests (8 methods) | Slice grouping keeps Order-dependent tests together |
| RepairTests (3 methods) | Removed stale mount-fail assertions - GVFS now tolerates corrupt index |
| FastFetchTests.CanFetchIntoEmptyGitRepoAndCheckoutWithGit | Fixed git output assertion (casing changed between git versions) |
| MountTests.MountFailsWhenNoOnDiskVersion | Wrapped in try/finally for reliable metadata restoration |
| MountTests.MountFailsWhenNoLocalCacheRootInRepoMetadata | Check exit code only (error messages now go to GVFS log, not stdout) |
| MountTests.MountFailsWhenNoGitObjectsRootInRepoMetadata | Same as above |
| MountTests.MountShouldFail helper | Now captures stderr in addition to stdout |
[...]

PR #2008, opened 2026-06-04
Fix repair of corrupt BlobSizes.sql and mount tolerance

tyrielv:
## Summary

Fix `gvfs repair --confirm` failing to delete corrupt BlobSizes.sql on Windows, and add defense-in-depth so mount tolerates corruption.

**Stacked on #1997**

## Root Cause

`SqliteDatabase.HasIssue()` opened a pooled SQLite connection for the integrity check. On Windows, disposing the connection returned the handle to the pool but kept the file locked. When repair tried to delete the corrupt folder, `TryDeleteFolder` failed silently (repair still exited 0). The corrupt file remained, causing mount to crash with a broken pipe.

## Changes

1. **`SqliteDatabase.HasIssue`** — Use `Pooling=False` for integrity-check connections so handles are released on dispose.
2. **`BlobSizes.Initialize`** — Catch `SQLITE_CORRUPT`/`SQLITE_NOTADB`, call `ClearAllPools()`, delete corrupt DB + WAL/SHM sidecars, recreate fresh. BlobSizes is a cache, so recreation is safe.
3. **`SqliteErrorCodes.cs`** — New constants for SQLite result codes.
4. **Test** — Remove `[SkipInCI]` from `RepairFixesCorruptBlobSizesDatabase`, add assertion that repair deletes the blob sizes folder.

### DIFF
--- a/GVFS/GVFS.Common/Database/SqliteDatabase.cs
+++ b/GVFS/GVFS.Common/Database/SqliteDatabase.cs
@@ -21,7 +21,7 @@ namespace GVFS.Common.Database
 
                 try
                 {
-                    string sqliteConnectionString = CreateConnectionString(databasePath);
+                    string sqliteConnectionString = $"data source={databasePath};Pooling=False";
                     using (SqliteConnection integrityConnection = new SqliteConnection(sqliteConnectionString))
                     {
                         integrityConnection.Open();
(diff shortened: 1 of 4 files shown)


tyrielv:
Already merged as part of #1997.

PR #2009, opened 2026-06-04
Fix CA2022 warnings: avoid inexact Stream.Read calls

tyrielv:
Replace `Stream.Read` with `Stream.ReadExactly` where the caller expects all requested bytes (`GitRepo`, `EnlistmentHydrationSummary`, `ReusableMemoryStream`). Suppress CA2022 in `GitIndexParser.ReadNextPage` where partial last-page reads are intentional.

**Changes (4 files):**
- `GitRepo.cs` — header read must be exact 5 bytes
- `EnlistmentHydrationSummary.cs` — index entry count is exact 4 bytes
- `ReusableMemoryStream.cs` — mock helper reads exact length
- `GitIndexProjection.GitIndexParser.cs` — pragma suppression; last page of paged index read is intentionally partial

**Validation:** 0 warnings, 0 errors. 807/807 unit tests pass.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitRepo.cs
+++ b/GVFS/GVFS.Common/Git/GitRepo.cs
@@ -153,8 +153,13 @@ namespace GVFS.Common.Git
             size = 0;
 
             byte[] buffer = new byte[5];
-            input.Read(buffer, 0, buffer.Length);
-            if (!Enumerable.SequenceEqual(buffer, LooseBlobHeader))
+
+            // Verify bytesRead instead of using ReadExactly: a truncated header must
+            // return false (Corrupt) so the caller quarantines the file, rather than
+            // throwing EndOfStreamException which would be caught as IOException
+            // (Unknown) and skip quarantine.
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Common/Git/GitRepo.cs
```diff
@@ -153,7 +153,7 @@ private static bool ReadLooseObjectHeader(Stream input, out long size)
             size = 0;
 
             byte[] buffer = new byte[5];
-            input.Read(buffer, 0, buffer.Length);
+            input.ReadExactly(buffer, 0, buffer.Length);
```
KeithIsSleeping: Switching to ReadExactly means a truncated loose-object header now throws instead of failing the header check, which routes it to a different (non-quarantining) error path.

tyrielv: Fixed

KeithIsSleeping (approved):

PR #2013, opened 2026-06-09
Use git ls-files -s instead of ls-tree for full-tree enumeration

tyrielv:
## Summary

When no previous commit exists to diff against (`sourceTreeSha == null`), `DiffHelper.PerformDiff` runs `git ls-tree -r -t HEAD` to enumerate all blobs and trees. This walks every tree object — very slow on large repos.

Replace with `git ls-files -s`, which reads the git index instead of walking tree objects. The index is already materialized in GVFS-mounted repos, making this significantly faster.

The optimization only applies when the target tree matches HEAD (i.e., the index reflects the tree we need). This is always the case for `gvfs prefetch`, which resolves HEAD as its target (`PrefetchVerb.LoadBlobPrefetchArgs` → `RevParse(HEAD)`). For other callers like FastFetch force-checkout (which can target a non-HEAD commit), the code falls back to `ls-tree` to preserve correctness.

## Benchmark (repo with ~2.5M files)

| Approach | Time | Speedup |
|----------|------|---------|
| `git ls-tree -r -t HEAD` (before) | ~24s | baseline |
| **`git ls-files -s` (after)** | **~6.5s** | **3.7×** |
| libgit2 in-process tree walk | ~8.2s | 2.9× |
| libgit2 in-process index read | ~12.9s | 1.9× |

Also benchmarked libgit2 alternatives: in-process recursive tree walk (2.9× faster than ls-tree) and in-process index read (1.9× — marshaling overhead). `git ls-files -s` was the fastest and simplest option.

## Changes

- **GitProcess.cs** — New `LsFilesStaging()` method that runs `git ls-files -s`
- **DiffTreeResult.cs** — New `ParseFromLsFilesStagingLine()` parser for the `<mode> <sha> <stage>\t<path>` format
- **DiffHelper.cs** — `PerformDiff` now uses `ls-files -s` when `sourceTreeSha == null` **and** `targetTreeSha` matches HEAD's tree (verified via libgit2). Falls back to `ls-tree` otherwise.
- **DiffTreeResultTests.cs** — 7 new unit tests for the parser

## Safety

- `TargetMatchesHeadTree()` resolves HEAD's tree SHA via libgit2 and compares to the requested `targetTreeSha`. Only uses the index-based path when they match.
- Falls back to `ls-tree` if the index is unavailable, HEAD can't be resolved, or the target differs from HEAD.
- `ls-files -s` only returns file entries (not tree entries). Tree entries from `ls-tree` were only used for directory creation, which `FlushStagedQueues` handles from file paths anyway.

### DIFF
--- a/GVFS/GVFS.Common/Git/DiffTreeResult.cs
+++ b/GVFS/GVFS.Common/Git/DiffTreeResult.cs
@@ -195,6 +195,47 @@ namespace GVFS.Common.Git
             return line.IndexOf(typeMarker, TypeMarkerStartIndex, typeMarker.Length, StringComparison.OrdinalIgnoreCase) == TypeMarkerStartIndex;
         }
 
+        /// <summary>
+        /// Parse the output of calling git ls-files -s (staging info).
+        /// This reads from the index, which is much faster than ls-tree on large repos.
+        /// ls-files only returns file entries (no tree entries).
+        /// </summary>
+        public static DiffTreeResult ParseFromLsFilesStagingLine(string line)
+        {
+            if (string.IsNullOrEmpty(line))
+            {
+                throw new ArgumentException("Line to parse cannot be null or empty", nameof(line));
+            }
+
+            /*
+             * Example output lines from ls-files -s
+             *
+             * 100644 44c5f5cba4b29d31c2ad06eed51ea02af76c27c0 0\tReadme.md
+             * 100755 196142fbb753c0a3c7c6690323db7aa0a11f41ec 0\tScripts/BuildGVFSForMac.sh
+             * ^-mode ^-sha                                    ^stage
+             *                                                   ^-tab
+             *                                                     ^-path
+             *
+             * Format: <mode> <sha> <stage>\t<path>
+             * Mode is 6 chars, space, SHA is 40 chars, space, stage digit(s), tab, path
+             */
+
+            int tabIndex = line.IndexOf('\t');
+            if (tabIndex < 0 || line.Length < 50)
+            {
+                return null;
+            }
+
+            DiffTreeResult blobAdd = new DiffTreeResult();
+            blobAdd.TargetMode = Convert.ToUInt16(line.Substring(0, 6), 8);
+            blobAdd.TargetIsSymLink = blobAdd.TargetMode == SymLinkFileIndexEntry;
+            blobAdd.TargetSha = line.Substring(7, GVFSConstants.ShaStringLength);
+            blobAdd.TargetPath = ConvertPathToUtf8Path(line.Substring(tabIndex + 1));
+            blobAdd.Operation = Operations.Add;
+
+            return blobAdd;
+        }
+
         private static string AppendPathSeparatorIfNeeded(string path)
         {
             return path.Last() == Path.DirectorySeparatorChar ? path : path + Path.DirectorySeparatorChar;
--- a/GVFS/GVFS.Common/Git/DiffTreeResult.cs
+++ b/GVFS/GVFS.Common/Git/DiffTreeResult.cs
@@ -199,6 +199,7 @@ namespace GVFS.Common.Git
         /// Parse the output of calling git ls-files -s (staging info).
         /// This reads from the index, which is much faster than ls-tree on large repos.
         /// ls-files only returns file entries (no tree entries).
+        /// Entries with stage != 0 (unmerged) are skipped to avoid duplicate/conflicting adds.
         /// </summary>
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -235,6 +260,37 @@ private void FlushStagedQueues()
...
+        /// git ls-tree (which walks a specific tree object).
+        /// </summary>
+        private bool TargetMatchesHeadTree(string targetTreeSha)
+        {
+            try
+            {
+                using (LibGit2Repo repo = new LibGit2Repo(this.tracer, this.enlistment.WorkingDirectoryBackingRoot))
+                {
+                    string headTreeSha = repo.GetTreeSha("HEAD");
+                    if (headTreeSha != null && string.Equals(headTreeSha, targetTreeSha, StringComparison.OrdinalIgnoreCase))
```
KeithIsSleeping: The guard that decides whether to use the fast ls-files path compares HEAD's tree SHA against a value the real callers pass as a commit SHA, so the comparison may never be true and the optimization may never actually run.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
```diff
@@ -268,6 +324,24 @@ private void EnqueueOperationsFromLsTreeLine(ITracer activity, string line)
...
+                return;
+            }
+
+            if (!this.ShouldIncludeResult(result))
+            {
+                return;
+            }
+
+            // ls-files -s only returns file entries, never trees
+            this.EnqueueFileAddOperation(activity, result);
```
KeithIsSleeping, line 358: The new ls-files path only enqueues file adds and never directory entries. FastFetch's checkout relies on directory operations to create folders before writing files, and the file writer does not create missing parent directories.

GVFS/GVFS.Common/Git/DiffTreeResult.cs
```diff
@@ -195,6 +195,47 @@ public static bool IsLsTreeLineOfType(string line, string typeMarker)
...
+            int tabIndex = line.IndexOf('\t');
+            if (tabIndex < 0 || line.Length < 50)
+            {
+                return null;
+            }
+
+            DiffTreeResult blobAdd = new DiffTreeResult();
+            blobAdd.TargetMode = Convert.ToUInt16(line.Substring(0, 6), 8);
+            blobAdd.TargetIsSymLink = blobAdd.TargetMode == SymLinkFileIndexEntry;
+            blobAdd.TargetSha = line.Substring(7, GVFSConstants.ShaStringLength);
```
KeithIsSleeping, line 247: ls-files -s reads the index, which can differ from HEAD's tree (staged or unmerged entries). The parser ignores the stage column, so unmerged paths could produce duplicate adds with the wrong blob SHA.

KeithIsSleeping (approved):

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
tyrielv: Good catch — this was a real bug. Callers pass commit SHAs but \GetTreeSha(\HEAD\)\ returns a tree SHA, so the comparison never matched and the optimization silently fell back to ls-tree every time. Fixed in 62ff7fa by resolving \	argetTreeSha\ to its tree SHA via \GetTreeSha()\ before comparing.

GVFS/GVFS.Common/Prefetch/Git/DiffHelper.cs
tyrielv: This is intentional — the ls-files path only fires when \	argetTreeSha\ matches HEAD, which only happens for \gvfs prefetch\ on a GVFS-mounted repo where directories are virtualized by PrjFlt and don't need explicit creation. FastFetch force-checkout targets a non-HEAD commit, so it falls back to ls-tree which includes directory entries. Added comments in 62ff7fa clarifying this.

GVFS/GVFS.Common/Git/DiffTreeResult.cs
tyrielv: Valid concern. While GVFS-mounted repos shouldn't have merge conflicts in the index, added defensive filtering in 62ff7fa — the parser now skips entries with stage != 0, returning null so they're ignored like invalid lines.

KeithIsSleeping (approved):

PR #2016, opened 2026-06-10
Fix reset --mixed skipping hydrated files due to skip-worktree

tyrielv:
## Problem

After `git reset --mixed HEAD~N`, hydrated files (read via blame/cat-file but not modified) retain skip-worktree because they are not in ModifiedPaths. Git skips the working-tree vs index comparison for these files, so `status` reports clean when the on-disk content doesn't match the new index entry.

**Manually confirmed:** control repo outputs 224 modified files including `M Readme.md`; GVFS repo outputs 223 — Readme.md missing from reset output and status.

## Fix

Adds a `PrepareForReset` pre-command hook (mirroring the existing `PrepareForUnstage` pattern for `
estore --staged`):

1. **ResetCommandParser** — detects mixed resets, extracts target commit
2. **Program.Reset** → IPC → **InProcessMount** → **FileSystemCallbacks.AddResetDiffToModifiedPaths** — runs `git diff-tree HEAD <target>`, adds changed paths to ModifiedPaths
3. Skip-worktree cleared for affected files → git correctly reports them

## Testing

- **Repro test** (`ReproResetMixedSkipWorktree`): hydrates Readme.md via blame, then resets to HEAD~1 where it changed. Expected to fail without the fix (pushed as first commit to verify CI catches it).
- **Unit tests**: 12 tests for `ResetCommandParser` arg parsing
- All 830 existing unit tests pass

## Commits

1. ⬅️ **Test only** (current push) — should FAIL in CI, proving the repro test catches the bug
2. **Fix** — will be pushed after CI confirms failure

### DIFF
--- a/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
+++ b/GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
@@ -493,6 +493,69 @@ namespace GVFS.Virtualization
         }
 
         /// <summary>
+        /// Finds files that differ between HEAD and the given target commit and adds
+        /// them to ModifiedPaths. This prepares for a mixed reset operation by ensuring
+        /// git will clear skip-worktree for these entries, allowing it to detect
+        /// working-tree mismatches for files that were hydrated (read from ProjFS but
+        /// not modified, so not yet in ModifiedPaths).
+        /// </summary>
+        /// <param name="targetCommit">
+        /// The commit to diff against HEAD. If null or empty, defaults to HEAD (no-op).
+        /// </param>
+        /// <param name="addedCount">Number of paths added to ModifiedPaths.</param>
+        /// <returns>True if the operation succeeded, false on failure.</returns>
+        public bool AddResetDiffToModifiedPaths(string targetCommit, out int addedCount)
+        {
+            addedCount = 0;
+
+            if (string.IsNullOrEmpty(targetCommit))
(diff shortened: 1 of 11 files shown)


GVFS/GVFS.Virtualization/FileSystemCallbacks.cs
```diff
@@ -492,6 +492,69 @@ public bool AddStagedFilesToModifiedPaths(string messageBody, out int addedCount
...
+            }
+
+            GitProcess gitProcess = new GitProcess(this.context.Enlistment);
+
+            GitProcess.Result result = gitProcess.DiffTreeNameOnly("HEAD", targetCommit);
+            if (result.ExitCodeIsSuccess && !string.IsNullOrEmpty(result.Output))
+            {
+                // Output is null-separated paths. The first entry from diff-tree is
+                // the target commit SHA (when given a commit, not a tree), skip it.
+                string[] parts = result.Output.Split(new[] { '\0' }, StringSplitOptions.RemoveEmptyEntries);
```
KeithIsSleeping, line 524: Splitting the diff-tree -z output  can leave a stray newline element that gets added to ModifiedPaths.

tyrielv:
More investigation is suggesting this needs to be fixed in git.exe, which makes a wrong assumption that placeholder files don't exist on disk (but they do if they've had contents read)

PR #2017, opened 2026-06-10
Fix flaky ModifiedPathsTests by replaying the on-disk log

tyrielv:
## Summary

Fixes the flaky `ModifiedPathsTests.DeletedTempFileIsRemovedFromModifiedFiles` test (and any other modified-paths assertion that hits the same race).

```
System.InvalidOperationException : Sequence contains more than one matching element
   at GVFS.Tests.Should.EnumerableShouldExtensions.ShouldNotContain[T](IEnumerable`1, Func`2)
   at GVFS.FunctionalTests.Tools.GVFSHelpers.ModifiedPathsShouldNotContain(...)
```

## Root cause

`ModifiedPathsDatabase` is an append-only A/D log that gets compacted at the end of each background-op batch via `WriteAllEntriesAndFlush` in `PostBackgroundOperation`. `BackgroundOperationCount` drops to 0 inside `DequeueAndFlush` for the last task — **before** the post-callback (and therefore the compaction) runs. `WaitForBackgroundOperations` polls until count==0 and can return between those two steps, observing a transient state like:

```
A temp.txt
D temp.txt
```

The old `ModifiedPathsShouldNotContain` matched both lines, fed two results into `EnumerableShouldExtensions.ShouldNotContain`, and `SingleOrDefault` threw `InvalidOperationException` — masking the real race with a confusing exception.

## Product code is correct

`FileBasedCollection.TryLoadFromDisk` replays the A/D log on mount and only the final state reaches the in-memory set, so the on-disk file is always recoverable. The append-then-amortized-rewrite pattern is intentional (one O(1) write per FS event vs. O(N) on every op). No product change needed. The new `AddFollowedByDeleteIsRecoveredOnLoad` unit test pins down this recovery contract explicitly.

## Changes (all test-only)

- **`GVFSHelpers.ModifiedPathsShouldContain` / `ModifiedPathsShouldNotContain`** — Replay the A/D log into a `HashSet<string>` (same algorithm as `TryLoadFromDisk`) and check semantic membership. Matches the intent of all ~110 existing callers and is robust to the flush-race window.
- **`GVFSHelpers.ModifiedPathsContentsShouldEqual` → `ModifiedPathsRawFileContentsShouldEqual`** — Renamed and documented to make it obvious this asserts on the *raw compacted file layout*, not the semantic set. Two callers in `CheckoutTests.cs` updated.
- **`EnumerableShouldExtensions.ShouldNotContain<T>(predicate)`** — Uses `Where(predicate)` instead of `SingleOrDefault(predicate)` so multi-match cases yield a useful `Assert.Fail` message instead of `InvalidOperationException`.
- **`ModifiedPathsDatabaseTests.AddFollowedByDeleteIsRecoveredOnLoad`** — New unit test asserting that loading `"A temp.txt\r\nD temp.txt\r\n"` produces an empty set (only the default `.gitattributes`).

## Verification

- ✅ Full `GVFS.UnitTests` suite passes (819/819) on Windows.
- ✅ New `AddFollowedByDeleteIsRecoveredOnLoad` test passes.
- ✅ `GVFS.FunctionalTests.csproj` builds clean.

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
@@ -248,7 +248,7 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
             this.FilesShouldMatchCheckoutOfSourceBranch();
 
             // Verify modified paths contents
-            GVFSHelpers.ModifiedPathsRawFileContentsShouldEqual(this.Enlistment, this.FileSystem, "A .gitattributes" + GVFSHelpers.ModifiedPathsNewLine);
+            GVFSHelpers.ModifiedPathsShouldOnlyContain(this.Enlistment, this.FileSystem, ".gitattributes");
         }
 
         [TestCase]
@@ -266,7 +266,7 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
                 .WithDeepStructure(this.FileSystem, this.ControlGitRepo.RootPath, compareContent: true, withinPrefixes: this.pathPrefixes);
 
             // Verify modified paths contents
-            GVFSHelpers.ModifiedPathsRawFileContentsShouldEqual(this.Enlistment, this.FileSystem, "A .gitattributes" + GVFSHelpers.ModifiedPathsNewLine);
+            GVFSHelpers.ModifiedPathsShouldOnlyContain(this.Enlistment, this.FileSystem, ".gitattributes");
         }
 
         [TestCase]
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
@@ -248,7 +248,7 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
             this.FilesShouldMatchCheckoutOfSourceBranch();
 
             // Verify modified paths contents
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.FunctionalTests/Tests/GitCommands/CheckoutTests.cs
```diff
@@ -248,7 +248,7 @@ public void CheckoutBranchAfterReadingFileAndVerifyContentsCorrect()
             this.FilesShouldMatchCheckoutOfSourceBranch();
 
             // Verify modified paths contents
-            GVFSHelpers.ModifiedPathsContentsShouldEqual(this.Enlistment, this.FileSystem, "A .gitattributes" + GVFSHelpers.ModifiedPathsNewLine);
+            GVFSHelpers.ModifiedPathsRawFileContentsShouldEqual(this.Enlistment, this.FileSystem, "A .gitattributes" + GVFSHelpers.ModifiedPathsNewLine);
```
KeithIsSleeping: The two renamed raw-content assertions in CheckoutTests still read the uncompacted file right after WaitForBackgroundOperations, so they may remain exposed to the same compaction race this PR fixes for the semantic helpers.

tyrielv: Good point - after reviewing the tests using it again, I've replaced ModifiedPathsRawFileContentsShouldEqual with a new method ModifiedPathsShouldOnlyContain which uses the same semantics for loading the contents as the other fixes in this PR.

KeithIsSleeping (approved):

PR #2018, opened 2026-06-11
Add functional test for reset --mixed skip-worktree bug (depends on microsoft/git#935)

tyrielv:
## Functional test for reset --mixed skip-worktree bug

Adds `ReproResetMixedSkipWorktree` to `CorruptionReproTests`. This test validates the fix in microsoft/git#935.

### Bug

After `git reset --mixed HEAD~N`, hydrated files (read via `blame`/`cat-file` but not modified) retained `skip-worktree` because git's `update_index_from_diff()` only cleared it for files not on disk. Hydrated files were invisible to `git status` after the reset.

### Test

On the `FunctionalTests/20201014` branch, `Readme.md` is the only file that differs between HEAD and HEAD~1:

1. `blame Readme.md` — hydrates the file (ProjFS materializes on disk, not in ModifiedPaths)
2. `reset HEAD~1` — should report `M Readme.md` in output

Without the git fix: GVFS repo reports 223 files, control reports 224 (Readme.md missing).
With the git fix: both report 224.

### Dependencies

- **microsoft/git#935** — the actual fix (in `builtin/reset.c`)
- This test will **fail** until the git fix is included in the GVFS build's git dependency

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CorruptionReproTests.cs
+++ b/GVFS/GVFS.FunctionalTests/Tests/GitCommands/CorruptionReproTests.cs
@@ -77,6 +77,47 @@ namespace GVFS.FunctionalTests.Tests.GitCommands
         }
 
         /// <summary>
+        /// Reproduces a bug where "git reset --mixed" fails to report hydrated files
+        /// as modified when skip-worktree hides the working-tree vs index mismatch.
+        ///
+        /// After a mixed reset, the index is updated to the target commit's tree, but
+        /// the working tree is left untouched. For non-hydrated placeholders this is
+        /// invisible (ProjFS serves the new content transparently). But for hydrated
+        /// files — files that have been read and materialized on disk — the on-disk
+        /// content still matches the OLD HEAD. Because the file was never modified
+        /// (just read), it is not in ModifiedPaths, so skip-worktree remains set.
+        /// This causes "git status" to skip the file entirely, hiding the mismatch
+        /// between the stale on-disk content and the new index entry.
+        ///
+        /// On the FunctionalTests/20201014 branch, Readme.md is the only file that
+        /// differs between HEAD and HEAD~1, making it a clean single-file repro.
+        ///
+        /// Expected: reset output includes "M Readme.md", status shows it as modified.
+        /// Actual (bug): reset output is empty, status reports clean.
+        /// </summary>
+        [TestCase]
+        public void ReproResetMixedSkipWorktree()
+        {
+            // Hydrate Readme.md by reading it via blame. In GVFS, this triggers a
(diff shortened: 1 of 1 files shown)


ShiningMassXAcc (changes requested):
Adding a repro test for a known projection bug is exactly the right instinct, and the `blame`-to-hydrate trick plus the note about `Readme.md` being the only file differing between HEAD and HEAD~1 on `FunctionalTests/20201014` is genuinely useful context. One blocking question and a few follow-ups.

Four findings, ranked.

---

### 1. High — the test is green, so what is it proving?

`ValidateGitCommand` fails when GVFS output diverges from the control repo. If the bug described in the doc comment were live, this test would be red. CI is fully green (35/35).

So which is it?

- microsoft/git#935 is already in the pinned git, and this is a genuine regression guard — great, and worth saying so in the PR body; or
- the test does not actually exercise the divergence, in which case it is a no-op that will never catch the regression it is named for.

The title says "depends on microsoft/git#935", which suggests the fix may not be in yet — and if so, a passing test is the wrong signal. Could you state in the PR which git version first makes this go from failing to passing? That is the one fact that tells a future reader whether this test has teeth.

### 2. Medium — the doc comment promises `git status`, but the test never runs it

The comment mentions status twice ("status shows it as modified", "status incorrectly reports clean"), yet the test only runs `blame`, `checkout -b` and `reset HEAD~1`.

And teardown will not cover for it: `GitRepoTests.TearDownForTest` -> `TestValidationAndCleanup` does `CheckHeadCommitTree()` and a deep **working-tree structure** comparison against the control repo. After a mixed reset the working trees are byte-identical in both repos — mixed reset does not touch the working tree — so the skip-worktree/status divergence cannot be caught there.

Everything therefore rests on `reset`'s stdout differing. Adding `this.ValidateGitCommand("status")` is the direct assertion for the documented symptom and costs nothing.

### 3. Medium — "Expected / Actual (bug)" will ship as wrong documentation

Given CI is green, the block asserting "Actual (bug): reset output is empty, status reports clean" describes a state that is probably no longer true, so it lands in the repo already stale.

Suggest retargeting the comment at what the test *guards* — asserts that reset reports `M Readme.md`, which regresses when skip-worktree hides the working-tree/index mismatch — and moving the investigation narrative to the PR description. The branch/commit note about `Readme.md` is worth keeping verbatim.

### 4. Low — no closing state assertion

The sibling `ReproCherryPickRestoreCorruption` ends with `this.FilesShouldMatchCheckoutOfSourceBranch();`. This one ends on a bare command. Worth a closing assertion for symmetry, especially alongside finding 2.

(No branch cleanup needed — the fixture is `enlistmentPerTest: true`, so the `tests/functional/ReproResetMixedSkipWorktree` branch goes away with the enlistment.)

PR #2023, opened 2026-06-12
Fix build break: bad merge in prefetch cache parameters

tyrielv:
Fixes build break introduced by the merge of #2002 on top of #2004.

Two call sites referenced the removed `lastPrefetchArgs` variable instead of the `prefetchCache` + `prefetchCacheSize` parameters added by #2004:

- **PrefetchVerb.cs**: hydration-fallback path now passes `prefetchCache, prefetchCacheSize` (matching the working call site on line 252)
- **InProcessMount.cs**: mount-side prefetch handler passes `null, 0` (one-shot download, no noop cache)


GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -231,11 +231,12 @@ protected override void Execute(GVFSEnlistment enlistment)
...
                                     out objectRequestor,
                                     out resolvedCacheServer);
-                                this.PrefetchBlobs(tracer, enlistment, headCommitId, filesList, foldersList, lastPrefetchArgs, objectRequestor, resolvedCacheServer);
+                                this.PrefetchBlobs(tracer, enlistment, headCommitId, filesList, foldersList, prefetchCache, prefetchCacheSize, objectRequestor, resolvedCacheServer);
                             }
                             else
                             {
-                                // Mount handled download — now hydrate locally
+                                // Mount handled download — update noop cache, then hydrate locally
+                                BlobPrefetcher.UpdateNoopCache(prefetchCache, prefetchCacheSize, headCommitId, filesList, foldersList, this.HydrateFiles);
```
KeithIsSleeping: The noop cache is marked done before local hydration runs, so if hydration fails the next run may wrongly skip it.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
```diff
@@ -251,6 +252,11 @@ protected override void Execute(GVFSEnlistment enlistment)
                                 out resolvedCacheServer);
                             this.PrefetchBlobs(tracer, enlistment, headCommitId, filesList, foldersList, prefetchCache, prefetchCacheSize, objectRequestor, resolvedCacheServer);
                         }
+                        else
+                        {
+                            // Mount handled download — update noop cache so repeat runs are skipped
+                            BlobPrefetcher.UpdateNoopCache(prefetchCache, prefetchCacheSize, headCommitId, filesList, foldersList, hydrate: false);
```
KeithIsSleeping, line 260: Opus 4.8 detected this, can you double check:
Both UpdateNoopCache calls (lines 239 and 258) run when TryPrefetchBlobsViaMountProcess returns true. If that method can return true while the in-mount download had failures, a failed prefetch would be cached as a noop and silently skipped on later runs.

GVFS/GVFS.FunctionalTests/Tests/EnlistmentPerFixture/PrefetchBlobsOffloadTests.cs
```diff
@@ -69,12 +69,14 @@ public void PrefetchBlobsMountedWithFolders()
...
-            // the mount process again.
+            // the mount process again. Since this file was already
+            // prefetched in Order(1), the noop cache correctly detects
+            // there's nothing new to download.
             this.Enlistment.UnmountGVFS();
             this.Enlistment.MountGVFS();
 
             string output = this.Enlistment.Prefetch($"--files {Path.Combine("GVFS", "GVFS", "Program.cs")}");
-            output.ShouldContain("Matched blobs:");
+            output.ShouldContain("Nothing new to prefetch.");
```
KeithIsSleeping, line 80: The updated assertion checks the noop short-circuit and now depends on an earlier ordered test having seeded the cache.

tyrielv: By design - see the [Order] attributes.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
tyrielv: The failure path calls ReportErrorAndExit, so the path you're worried about should be unreachable.

GVFS/GVFS/CommandLine/PrefetchVerb.cs
tyrielv: Fixed

KeithIsSleeping (approved):

PR #2025, opened 2026-06-15
ci: drop Debug configuration from build and test matrices

tyrielv:
## Motivation

The CI test matrix is about to get more complex: user-level install testing and arm64-native builds will each add new matrix dimensions. Before that expansion, simplify by removing the Debug configuration — which currently doubles every build and test job for minimal benefit.

## What this does

**Removes Debug from CI matrices** in all three workflows:
- `build.yaml` — build + unit test matrix: `[Debug, Release]` → `[Release]`
- `functional-tests.yaml` — functional test matrix: `[Debug, Release]` → `[Release]`
- `upgrade-tests.yaml` — upgrade test matrix: `[Debug]` → `[Release]`

**Promotes Debug.Assert to runtime checks** so coverage is strictly *better* than before:
- `GitStatusCache.cs` — two lock-discipline assertions → `InvalidOperationException`
- `QueuedPipeStringWriter.cs` — queue-drained assertion → `InvalidOperationException`
- `MockPhysicalGitObjects.cs` — null-stream assertion → `ArgumentNullException.ThrowIfNull`

## Why Debug CI had little value here

- **Zero `#if DEBUG` blocks** in the codebase — no Debug-only code paths exist
- **Only 3 `Debug.Assert` calls** total in production code (now promoted)
- **No custom `DefineConstants` or `CheckForOverflowUnderflow`** in build props
- The practical difference was: JIT/AOT optimizations on/off, and 3 assertions that are now always-on

## Impact

- ~50% reduction in CI runner usage per PR (1 fewer build, 20 fewer FT shards, 6 fewer upgrade-test jobs)
- Debug configuration remains fully buildable locally for dev inner-loop

### DIFF
--- a/GVFS/GVFS.Common/GitStatusCache.cs
+++ b/GVFS/GVFS.Common/GitStatusCache.cs
@@ -3,7 +3,7 @@ using GVFS.Common.NamedPipes;
 using GVFS.Common.Tracing;
 using System;
 using System.ComponentModel;
-using System.Diagnostics;
+
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
@@ -597,7 +597,10 @@ namespace GVFS.Common
 
         private bool TryDeleteStatusCacheFile()
         {
-            Debug.Assert(this.cacheFileLock.IsHeldByCurrentThread, "Attempting to delete the git status cache file without the cacheFileLock");
+            if (!this.cacheFileLock.IsHeldByCurrentThread)
+            {
+                throw new InvalidOperationException("Attempting to delete the git status cache file without the cacheFileLock");
+            }
 
             try
             {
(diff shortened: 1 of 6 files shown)


KeithIsSleeping (approved):

GVFS/GVFS.Common/GitStatusCache.cs
```diff
@@ -3,7 +3,7 @@
 using GVFS.Common.Tracing;
 using System;
 using System.ComponentModel;
-using System.Diagnostics;
+
```
mjcheetham, line 6: Odd to leave this as a blank line instead of just dropping. But not important.

GVFS/GVFS.UnitTests/Mock/Common/MockPhysicalGitObjects.cs
```diff
@@ -3,7 +3,8 @@
 using GVFS.Common.Git;
 using GVFS.Common.Http;
 using GVFS.Common.Tracing;
-using System.Diagnostics;
+
```
mjcheetham, line 6: Same nit.

mjcheetham (approved):
This makes sense to me! The build+test against release and debug is vestigial at this point.

PR #2026, opened 2026-06-15
Add ARM64 native build support

tyrielv:
# Add ARM64 native build support

Adds native ARM64 build support alongside the existing x64 build, plus the
CI matrix to PR-validate both arches and the telemetry plumbing to identify
which arch is running.

**Status:** draft — do **not** mark ready for review without going through
the open items at the bottom.

## Why

VFS for Git 2.0 moved from .NET Framework 4.7 to .NET 10 NativeAOT. The
single self-contained x64 binary we ship today runs under Prism on ARM64
Windows. A native ARM64 build eliminates that overhead.

## Perf results

Measured on a Snapdragon Windows host, Release / NativeAOT / SelfContained,
both built from this branch. Methodology and raw JSONL data in the perf
writeup (kept out of the repo, in session state).

| Workload | ARM64 native vs x64-under-Prism |
|---|---|
| `gvfs version` cold | **1.51× faster** (46.5 ms vs 70.1 ms median) |
| `gvfs version` warm | **1.47× faster** |
| `gvfs --help` cold | **1.49× faster** |
| `GVFS.PostIndexChangedHook` cold (per `git checkout` etc.) | **1.53× faster** |
| GVFS.UnitTests full suite | **1.21× faster** (9.7s vs 11.2s median) |
| GVFS.FunctionalTests full suite | **1.37× faster** (27.8 min vs 38.1 min) |
| large repo workload sample (clone + checkout + hydrate + blame) | **~tied** (within 10% once blob cache warmth is controlled) |

Functional test results matched arch-for-arch: 552 pass / 1 fail / 16 skip
on both arches, same single known-flaky test (`LooseObjectStepTests.CorruptLooseObjectIsDeleted`).

## What's in the PR

Two commits, intentionally separable so the telemetry change can be
reviewed (and if needed, reverted) independently of the build work.

### `fe5bfdb0` Add architecture field to telemetry events
6 files, +38/-1.

- New `ProcessHelper.GetCurrentProcessArchitecture()` — RID-style lowercase
  string ("x64" / "arm64"), cached, parallel to the existing
  `GetCurrentProcessVersion()`.
- Pair `Architecture` alongside `Version` in the three in-process telemetry
  emit sites: `JsonTracer.WriteStartEvent`, `HeartbeatThread.EmitHeartbeat`,
  `GVFSService.Windows.Run`.
- Add top-level `architecture` property to the `PipeMessage` schema in
  `TelemetryDaemonEventListener` — peer to `version`. Cached in constructor,
  attached to every outgoing pipe message. Updated the test for the new
  field (top-level property count 6 → 7).

The collector side (devprod.git.telemetry repo) needs a matching change
to pick up the new field and pass it through to ETW / AppInsights. That
PR is also in draft; ordering between the two doesn't matter — old
collector silently drops the new field, new collector tolerates clients
that don't send it.

### `9199ea64` Add ARM64 native build support
23 files, +451/-153.

Architecture parameterization:
- `Directory.Build.props` introduces `$(VfsArch)` (lowercase RID/vcpkg form)
  and `$(VfsNativePlatform)` (mixed-case vcxproj form). All hardcoded
  `win-x64`, `x64-windows-*`, `<Platform>x64</Platform>` references go
  through these properties.
- `Build.bat` gains a 4th `ARCH` argument defaulting to `x64` (existing
  invocations keep their exact behaviour). Threads through vcpkg triplet
  selection, MSBuild Platform, dotnet RID. Two cross-cutting fixes
  accompany this:
  - Prepend the VS Installer dir to PATH so `ilc` can find `vswhere` →
    `link.exe` when invoked from a non-developer cmd shell. Pre-existing
    trip hazard.
  - Call `vcvarsall.bat <arch>` before the native C++ loop. Without this,
    MSBuild finds the v145 ARM64 toolset's `.targets` but cannot locate
    the matching `cl.exe` / `link.exe`.
  - Clear `Platform` env var (vcvarsall sets it) before the managed
    publish loop. Otherwise csproj eval inserts a spurious `\<arch>\`
    segment into managed output paths.
- All 5 `.vcxproj` files gain `Debug|ARM64` / `Release|ARM64`
  `ProjectConfiguration` entries. Configuration|Platform conditions
  simplified to Configuration-only (their bodies were never
  platform-specific). UCRT lib arch path parameterized to
  `\$(Platform.ToLower())`. `<PlatformToolset>` parameterized via new
  `$(VfsPlatformToolset)` — `v143` for x64 (preserves baseline
  byte-for-byte), `v145` for arm64 (only toolset with ARM64
  cross-compile binaries here).
- `GVFS.sln` gains `Debug|ARM64` / `Release|ARM64` solution configs and
  per-project mappings.
- `GVFS.NativeTests.vcxproj` swaps the x64-only `GVFS.ProjFS` NuGet's
  `ProjectedFSLib.lib` path for Windows SDK 10.0.26100's
  `Lib\…\um\$(Platform.ToLower())\`, which ships ARM64 natively.
  Unused `packages.config` + the `GVFS.ProjFS 2019.411.1` reference
  removed.
- `layout.bat` parameterized; for ARM64 the VC runtime DLLs come from
  `%VCToolsRedistDir%\arm64\Microsoft.VC*.CRT\` because the
  `GVFS.VCRuntime` NuGet is x64-only.
- `GVFS.Installers.csproj` `LayoutPath` uses `win-$(VfsArch)`. New
  `InstallerArchSuffix` property (empty for x64, `-arm64` for arm64) is
  passed to Inno Setup via `/DArchSuffix`. `Setup.iss` `OutputBaseFilename`
  uses `{#ArchSuffix}` so the **x64 installer keeps its historical name**
  (`SetupGVFS.<v>.exe`) and the **arm64 installer gets a `-arm64`
  suffix** (`SetupGVFS.<v>-arm64.exe`). Both files can coexist on a
  single GitHub release.
- New vcpkg triplets: `triplets/arm64-windows-static-aot.cmake` and
  `arm64-windows-dynamic.cmake`.

Functional-test infrastructure:
- `GVFS.FunctionalTests/Settings.cs`: dev-mode payload discovery uses
  `RuntimeInformation.ProcessArchitecture` instead of hardcoded
  `win-x64`. So an ARM64 functional-test driver targets the ARM64
  payload it was built alongside, instead of silently falling through
  to the installed system gvfs.
- `scripts/RunFunctionalTests-Dev.ps1` gains `-Arch` parameter. The
  `$payloadDir` computation is fixed to match the Payload csproj's
  actual output layout (`bin\<cfg>\win-<arch>\` — the Payload sets
  `AppendTargetFrameworkToOutputPath=false`), with a missing-file
  guard so the bug above can't silently recur.

CI matrix (per-arch native builds, no arch-cross):
- `.github/workflows/build.yaml`: matrix gains an `architecture`
  dimension. `runs-on` routes to `windows-11-arm` for arm64 / `windows-2025`
  for x64, same pattern the functional-tests workflow already uses.
  Artifact names get an `_<arch>` suffix.
- `.github/workflows/functional-tests.yaml`: downloads arch-keyed
  `GVFS_<cfg>_<arch>` and `FunctionalTests_<cfg>_<arch>` artifacts.
  Matrix value `x86_64` renamed to `x64` for consistency with the
  artifact-name suffix.
- `.github/workflows/upgrade-tests.yaml` is x64-only; its `GVFS_<cfg>`
  download is updated to `GVFS_<cfg>_x64` to match the new artifact name.

## Verification done locally

- x64 Release build end-to-end: clean. Installer at expected path.
- ARM64 Release build end-to-end: clean. Installer at expected path with
  `-arm64` suffix.
- Both payloads audited: 13/13 binaries are the expected arch (verified
  via dumpbin /HEADERS — AA64/8664).
- `GVFS.UnitTests.exe` Release passes on both arches (14.0s arm64, 17.2s x64).
- `TelemetryDaemonEventListenerTests.TraceMessageDataIsCorrectFormat`
  still passes (new field assertions in place).

## Not yet verified (require a real release pipeline run)

- **ESRP signing of ARM64 PE files:** the release pipeline YAML is
  correct (uses the same CP-230012 key, which is arch-agnostic), but
  the first actual ARM64 release build will be the proof that ESRP
  accepts the ARM64 binaries without issue.
- **First install of the ARM64 installer on a clean ARM64 box:** CI
  functional tests (which install via Inno Setup + run the full test
  suite on `windows-11-arm` runners) serve as the functional
  equivalent, but a hand-verified fresh-box install hasn't been done.
## Open items — all resolved

All items verified by successful CI runs (build + functional tests
green on both arches). See previous PR description revision for the
original list.
---

Ready for review.

Assisted-by: Claude Opus 4.7


tyrielv:
Closing in favor of a new PR using the branch pushed directly to upstream (avoids needing to sync two remotes).

PR #2031, opened 2026-06-18
PlaceholderTable/SparseTable: add transient SQLite error resilience via GVFSTable base class

tyrielv:
## Problem

Two classes of transient SQLite errors hit PlaceholderTable in production:

1. **SQLITE_IOERR (10)** - telemetry shows repeated disk I/O errors in `GetFilePlaceholdersCount()` during heartbeat, caused by ReFS snapshots, antivirus, or momentary disk busyness.

2. **SQLITE_LOCKED (6)** - table lock contention on the Placeholder table during concurrent operations.

Both PlaceholderTable and SparseTable shared an identical pattern (connection pool, writer lock, try/catch wrapping) with no retry or transient error handling.

## Fix

Extract shared logic into `GVFSTable` base class with four execution primitives:

| Method | Behavior |
|--------|----------|
| `ExecuteWrite` | Under writer lock, retry up to 5x on transient errors |
| `ExecuteRead` | Retry up to 5x on transient errors |
| `ExecuteNonCriticalRead` | Return fallback value on transient error (no throw) |
| `ExecuteReadThenWrite` | Mixed read+write on same connection, retry on transient errors |

**Transient errors retried** (linear backoff: 50ms, 100ms, ... 250ms):
- `SQLITE_BUSY (5)` - connection-level lock contention
- `SQLITE_LOCKED (6)` - table-level lock contention
- `SQLITE_IOERR (10)` - disk I/O errors

**Non-critical count methods** (`GetCount`, `GetFilePlaceholdersCount`, `GetFolderPlaceholdersCount`) return -1 on transient failure - only consumed by heartbeat telemetry.

Also fixes pre-existing copy-paste bug where `GetFilePlaceholdersCount`/`GetFolderPlaceholdersCount` exception messages incorrectly said `GetCount`.

## Files Changed

- **New:** `GVFSTable.cs` - base class with retry/lock/error infrastructure
- **Refactored:** `PlaceholderTable.cs` - inherits GVFSTable, all operations use base methods
- **Refactored:** `SparseTable.cs` - same treatment
- **Extended:** `SqliteErrorCodes.cs` - added BUSY, LOCKED, IOERR constants + `IsTransientError()`
- **Updated:** Unit tests - adjusted exception message assertions to match simplified format

## Validation

- All 872 unit tests pass (0 failures, 11 pre-existing skips)
- Self-reviewed: caught and fixed a retry-duplicate bug in `GetAllEntries`

### DIFF
--- a/GVFS/GVFS.Common/Database/GVFSTable.cs
+++ b/GVFS/GVFS.Common/Database/GVFSTable.cs
@@ -1,0 +1,154 @@
+﻿using Microsoft.Data.Sqlite;
+using System;
+using System.Data;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace GVFS.Common.Database
+{
+    /// <summary>
+    /// Base class for GVFS SQLite tables. Provides connection pooling,
+    /// writer serialization, and transient error retry logic.
+    /// </summary>
+    public abstract class GVFSTable
+    {
+        private const int MaxRetries = 5;
+        private const int BaseRetryDelayMs = 50;
+
+        private readonly IGVFSConnectionPool connectionPool;
+        private readonly Lock writerLock = new Lock();
+
+        protected GVFSTable(IGVFSConnectionPool connectionPool)
+        {
+            this.connectionPool = connectionPool;
+        }
+
+        /// <summary>
+        /// Name of the concrete table class, used in exception messages.
+        /// </summary>
+        protected abstract string TableName { get; }
+
+        /// <summary>
+        /// Executes a read operation with retry on transient SQLite errors.
+        /// Throws GVFSDatabaseException on non-transient or exhausted retries.
+        /// </summary>
+        protected T ExecuteRead<T>(Func<IDbCommand, T> operation, [CallerMemberName] string caller = null)
+        {
+            int attempt = 0;
+            while (true)
+            {
+                try
+                {
+                    using (IDbConnection connection = this.connectionPool.GetConnection())
+                    using (IDbCommand command = connection.CreateCommand())
+                    {
+                        return operation(command);
+                    }
+                }
+                catch (SqliteException ex) when (SqliteErrorCodes.IsTransientError(ex.SqliteErrorCode) && attempt < MaxRetries)
+                {
+                    attempt++;
+                    Thread.Sleep(BaseRetryDelayMs * attempt);
+                }
+                catch (Exception ex)
+                {
+                    throw new GVFSDatabaseException($"{this.TableName}.{caller} Exception", ex);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Executes a read operation that tolerates transient errors by returning
+        /// a fallback value. Used for non-critical paths like heartbeat telemetry.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


GVFS/GVFS.Common/Database/GVFSTable.cs
```diff
@@ -0,0 +1,154 @@
...
+                    using (IDbConnection connection = this.connectionPool.GetConnection())
+                    using (IDbCommand command = connection.CreateCommand())
+                    {
+                        return operation(command);
+                    }
+                }
+                catch (SqliteException ex) when (SqliteErrorCodes.IsTransientError(ex.SqliteErrorCode) && attempt < MaxRetries)
+                {
+                    attempt++;
+                    Thread.Sleep(BaseRetryDelayMs * attempt);
```
KeithIsSleeping: Microsoft.Data.Sqlite already retries SQLITE_BUSY/LOCKED internally until the command timeout (default 30s) before throwing, so these outer retries can stack on top of it. 

The lock contention could block far longer than the intended ~750ms.

GVFS/GVFS.Common/Database/GVFSTable.cs
```diff
@@ -0,0 +1,154 @@
...
+                }
+                catch (Exception ex)
+                {
+                    throw new GVFSDatabaseException($"{this.TableName}.{caller} Exception", ex);
```
KeithIsSleeping: The shared exception wrapper drops the path/sha that the old messages carried, and all four Add* methods now surface as 'InsertPlaceholder' via CallerMemberName.

GVFS/GVFS.Common/Database/GVFSTable.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                using (IDbConnection connection = this.connectionPool.GetConnection())
+                using (IDbCommand command = connection.CreateCommand())
+                {
+                    return operation(command);
+                }
+            }
+            catch (SqliteException ex) when (SqliteErrorCodes.IsTransientError(ex.SqliteErrorCode))
+            {
+                return fallbackValue;
```
KeithIsSleeping, line 68: ExecuteNonCriticalRead returns -1 on the first transient error with no trace/metric and no retry, hiding the very IOERR/BUSY/LOCKED failures this PR is meant to address.

GVFS/GVFS.Common/Database/GVFSTable.cs
tyrielv: Good catch! I've addressed this by setting the default command timeout to 2s (down from 30s) so with the outer retries the total wait time ~11s (5x2s+750ms).

GVFS/GVFS.Common/Database/GVFSTable.cs
tyrielv: This is by design - NonCriticalRead is exactly that, non-critical. It's only used for an additional telemetry field on heartbeat events. This PR is primarily designed to add resilience to the critical operations, but includes changes to NonCriticalRead to emphasize the difference.

GVFS/GVFS.Common/Database/GVFSTable.cs
tyrielv: Fixed - PlaceholderTable includes the path/sha, CallerMemberName usage fixed.

KeithIsSleeping (approved):

PR #2033, opened 2026-06-19
Add telemetry events to mount initialization phases

tyrielv:
## Problem

When the GVFS mount process hangs during initialization, there is NO telemetry in Application Insights between the `VFS.EnlistmentInfo` event and `VFS.Mount 'Virtual repo is ready'`. This makes it impossible to diagnose which phase is stuck from server-side telemetry.

A user experienced a mount that took 136 seconds to initialize (normally <1s), and we could only see a 60-second gap in telemetry with no events. The mount was stuck in the auth/network validation phase.

## Solution

Add `Keywords.Telemetry` events at each phase transition during `MountWithLockAcquired`. Each event uses a consistent `MountPhase` event name with a `Phase` field and an `ElapsedMs` field measuring time since mount start.

### New MountPhase events

| Phase | When |
|-------|------|
| `ParallelMountStarted` | Parallel auth + local validation begins |
| `NetworkValidationComplete` | Auth + /gvfs/config query done |
| `LocalValidationComplete` | Git/hooks/fs validation + config done |
| `ParallelMountComplete` | Both parallel tasks finished |
| `CacheServerResolved` | Cache server URL resolved |
| `LocalCacheHealthy` | Local object cache validated |
| `ContextCreated` | GVFSContext initialized |
| `HooksUpdated` | Hook binaries installed (or skipped for worktrees) |
| `VirtualizationStarting` | About to start ProjFS callbacks |

The existing `VFS.Mount 'Virtual repo is ready'` event serves as the final phase.

## Analysis: Network timeout budget

`RetryConfig` defaults: 6 retries x 30s timeout = **210s worst case** for `TryInitializeAndQueryGVFSConfig`. When auth is expired or network is unreachable, the mount can block for up to 210s in the network task with no telemetry — these new events make that gap visible and pinpoint exactly which phase is stuck.

## Testing

- Builds successfully (`dotnet build GVFS.Mount.csproj`)
- Events are informational-level with `Keywords.Telemetry` so they flow to Application Insights
- No behavioral changes — purely additive observability

### DIFF
--- a/GVFS/GVFS.Mount/InProcessMount.cs
+++ b/GVFS/GVFS.Mount/InProcessMount.cs
@@ -128,8 +128,19 @@ namespace GVFS.Mount
             // TryInitializeAndQueryGVFSConfig combines the anonymous probe, credential fetch,
             // and config query into at most 2 HTTP requests (1 for anonymous repos), reusing
             // the same HttpClient/TCP connection.
+            Stopwatch mountPhaseTimer = Stopwatch.StartNew();
             Stopwatch parallelTimer = Stopwatch.StartNew();
             bool hasCacheServer = this.cacheServer != null && !string.IsNullOrWhiteSpace(this.cacheServer.Url);
+
+            this.tracer.RelatedEvent(
+                EventLevel.Informational,
+                "MountPhase",
+                new EventMetadata
+                {
+                    { "Phase", "ParallelMountStarted" },
+                    { "ElapsedMs", mountPhaseTimer.ElapsedMilliseconds },
+                },
+                Keywords.Telemetry);
 
             // When a cache server is configured locally, auth/config is best-effort:
             // mount can proceed without it. We still attempt it so GCM can pop up a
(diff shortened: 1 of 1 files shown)


GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -168,6 +179,19 @@ private void MountWithLockAcquired(EventLevel verbosity, Keywords keywords)
                 }
 
                 this.ValidateGVFSVersion(config, failOnError: !hasCacheServer);
+
+                this.tracer.RelatedEvent(
+                    EventLevel.Informational,
+                    "MountPhase",
+                    new EventMetadata
+                    {
+                        { "Phase", "NetworkValidationComplete" },
```
KeithIsSleeping, line 188: When a cache server is configured, the network task is not awaited, so its NetworkValidationComplete event can be emitted after later phases (even after 'Virtual repo is ready') or be missing entirely.

tyrielv: This is by design - the network task is not essential if cache server is configured, it just helps to get the auth stack warm.

KeithIsSleeping (approved):

PR #2034, opened 2026-06-19
Mount: don't block startup on auth when cache server is configured

tyrielv:
## Problem

A user reported *'Unable to mount because the GVFS.Mount process is not responding'* when GCM was blocked waiting for interactive auth on a headless server. The mount process was alive but stuck in `git credential fill`, and MountVerb's 60-second pipe connect expired with a misleading error.

More broadly: when a cache server URL is already in local git config, there's no reason for mount to block on authentication. The enlistment can serve objects from the local cache immediately, and knows the cache server to use when not available locally. If a cache server has not been configured already though (or explicitly set to not use cache server) then the mount should try to configure one upon startup.

## Fix

**Background auth with cache server** — When a cache server URL is configured locally, the auth + config query runs as a fire-and-forget background task. Mount proceeds immediately after local validations complete. GCM still gets a chance to pop up a renewal prompt so the token is fresh for subsequent object downloads, but it cannot delay mount.

**Without a cache server** — auth remains synchronous and blocking (as before), but with better diagnostics:

| Exit code | Name | Meaning |
|-----------|------|---------|
| 9 | `AuthenticationError` | Credentials rejected by server |
| 10 | `CredentialTimeout` | `git credential fill` hung for 30s (GCM waiting for interactive auth nobody is answering) |
| 11 | `RemoteGvfsConfigError` | `/gvfs/config` query failed for non-auth reasons (DNS, timeout, 5xx) |

**Process tracking in MountVerb** — `WaitUntilMounted` now uses short-interval (500ms) connect retries with process liveness checks instead of a single blocking 60s connect. If GVFS.Mount exits early, MountVerb detects it within 500ms.

## Implementation details

- **Credential gate** — `SemaphoreSlim(1,1)` in `GitAuthentication` serializes all `git-credential-fill` calls. Prevents duplicate GCM prompts when the background auth task and foreground object downloads race to fetch credentials.
- **30s credential timeout (no-cache-server only)** — `git credential fill` is killed after 30 seconds so mount fails fast with exit code 10 instead of hanging indefinitely. The timeout is added to `ICredentialStore.TryGetCredential` (default infinite, preserving existing behavior for all other callers).
- **Version check is log-only in background** — `ValidateGVFSVersion` logs but doesn't `FailMountAndExit` when running as the background auth task, avoiding a race where the process could terminate after mount already reported success.
- **isInitialized on credential failure** — `GitAuthentication` marks itself initialized even when credential fetch fails, so `TryGetCredentials` can retry during later object downloads rather than throwing `InvalidOperationException`.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitAuthentication.cs
+++ b/GVFS/GVFS.Common/Git/GitAuthentication.cs
@@ -13,8 +13,11 @@ namespace GVFS.Common.Git
     public class GitAuthentication
     {
         private const double MaxBackoffSeconds = 30;
+        public const int DefaultCredentialTimeoutMs = 30_000;
+        public const int BackgroundCredentialTimeoutMs = 120_000;
 
         private readonly Lock gitAuthLock = new Lock();
+        private readonly SemaphoreSlim credentialGate = new SemaphoreSlim(1, 1);
         private readonly ICredentialStore credentialStore;
         private readonly string repoUrl;
 
@@ -219,7 +222,8 @@ namespace GVFS.Common.Git
             RetryConfig retryConfig,
             out ServerGVFSConfig serverGVFSConfig,
             out string errorMessage,
-            out bool isAuthFailure)
+            out bool isAuthFailure,
+            int credentialTimeoutMs = DefaultCredentialTimeoutMs)
         {
             if (this.isInitialized)
             {
@@ -246,6 +250,7 @@ namespace GVFS.Common.Git
 
                 if (httpStatus != HttpStatusCode.Unauthorized)
                 {
+                    this.isInitialized = true;
                     errorMessage = "Unable to query /gvfs/config";
                     tracer.RelatedWarning("{0}: Config query failed with status {1}", nameof(this.TryInitializeAndQueryGVFSConfig), httpStatus?.ToString() ?? "None");
                     return false;
@@ -254,9 +259,13 @@ namespace GVFS.Common.Git
                 // Server requires authentication — fetch credentials
                 this.IsAnonymous = false;
 
-                if (!this.TryCallGitCredential(tracer, out errorMessage))
+                if (!this.TryCallGitCredential(tracer, out errorMessage, credentialTimeoutMs))
                 {
                     isAuthFailure = true;
+                    // Mark initialized even on failure so TryGetCredentials can
+                    // retry later (e.g., when mount proceeds with a cache server
+                    // and object downloads need auth).
+                    this.isInitialized = true;
                     tracer.RelatedWarning("{0}: Credential fetch failed: {1}", nameof(this.TryInitializeAndQueryGVFSConfig), errorMessage);
                     return false;
                 }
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -252,7 +274,7 @@ private void MountWithLockAcquired(EventLevel verbosity, Keywords keywords)
                 parallelTimer.Stop();
                 this.tracer.RelatedInfo("ParallelMount: All parallel tasks completed in {0}ms", parallelTimer.ElapsedMilliseconds);
 
-                ServerGVFSConfig serverGVFSConfig = networkTask.Result;
+                ServerGVFSConfig serverGVFSConfig = hasCacheServer ? null : networkTask.Result;
```
KeithIsSleeping, line 283: In the cache-server path the background networkTask is never awaited or read, so exceptions inside it are silently dropped.

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -376,28 +383,39 @@ private void UpdateBackoff()
...
-        private bool TryCallGitCredential(ITracer tracer, out string errorMessage)
+        private bool TryCallGitCredential(ITracer tracer, out string errorMessage, int timeoutMs = -1)
         {
-            string gitUsername;
-            string gitPassword;
-            if (!this.credentialStore.TryGetCredential(tracer, this.repoUrl, out gitUsername, out gitPassword, out errorMessage))
+            // Serialize credential fetches so only one git-credential-fill
+            // process runs at a time. Without this, a background auth task
+            // and a foreground object download could both spawn GCM prompts.
+            this.credentialGate.Wait();
```
KeithIsSleeping: The credential gate has no wait timeout while other callers still use an infinite credential-fill timeout, so a hung fill can stall all auth.

GVFS/GVFS.Mount/InProcessMount.cs
```diff
@@ -141,20 +145,28 @@ private void MountWithLockAcquired(EventLevel verbosity, Keywords keywords)
                     this.tracer, this.enlistment, this.retryConfig,
```
KeithIsSleeping, line 145: The cache-server background auth also kills git credential fill at 30s, which can cut off an interactive GCM refresh.

GVFS/GVFS.Common/Git/GitAuthentication.cs
```diff
@@ -254,9 +257,13 @@ public bool TryInitializeAndQueryGVFSConfig(
                 // Server requires authentication — fetch credentials
                 this.IsAnonymous = false;
```
KeithIsSleeping, line 260: With a cache server, mount can proceed while the auth object is still not initialized, so a later credential request during hydration can throw InvalidOperationException.

tyrielv: Fixed by marking initialized on all paths.

GVFS/GVFS.Mount/InProcessMount.cs
tyrielv: Fixed by using longer timeout for background auth.

GVFS/GVFS.Mount/InProcessMount.cs
tyrielv: Fixed with fault-only continuation to log.

GVFS/GVFS.Common/Git/GitAuthentication.cs
tyrielv: Fixed with 60s timeout on credentialGate

KeithIsSleeping (approved):

PR #2042, opened 2026-07-06
Harden GVFS.Mount against native ProjFS failures under memory pressure (0xC000CE01)

tyrielv:
## Problem

`GVFS.Mount` crashes are being reported that surface downstream as other tools exiting with `-1073689087` = `0xC000CE01` = `STATUS_FILE_SYSTEM_VIRTUALIZATION_UNAVAILABLE`. The chain is: **GVFS.Mount crashes → the ProjFS virtualization root is orphaned → every subsequent placeholder access on that enlistment returns `0xC000CE01`** until remount.

Watson attributes the crashes to `radar_high_memory projectedfslib.dll!prjcompletecommand`, alongside `System.OutOfMemoryException` buckets and GC access violations. The dominant managed signatures are `NullReferenceException` surfaced from the **native** ProjFS command-completion (`PrjCompleteCommand`, via `StartDirectoryEnumerationAsyncHandler`) and delete (`PrjDeleteFile`) paths — the native code faults under memory exhaustion and the CLR raises a (catchable) `NullReferenceException` on the worker thread, which today reaches `ExecuteFileOrNetworkRequest` and calls `Environment.Exit`.

VFS for Git is already on the latest `Microsoft.Windows.ProjFS` (2.1.0), so the native `projectedfslib.dll` fault cannot be fixed from here. The only managed-side lever this PR pulls is: **stop converting a single failed native ProjFS call into a whole-mount crash.** It does **not** claim to eliminate the native OOM fault itself.

> Not a v2.0 regression: the crash-prone call sites date to 2018 (`785ccb4a`, .NET Framework era). The v2.0 pure-C# ProjFS rewrite (`66334474`) made the same latent native fault *legible* as a managed NRE in GVFS telemetry — so the recent spike in reports is largely improved visibility, which this PR's telemetry is designed to quantify.

## Scope of this PR — the stabilization-safe core

This is the **narrow, hot-path-neutral** half of the investigation, intended for the 2.0 stabilization prerelease. A second, *speculative* half — bounding the leaked `activeEnumerations` collection — has been **split out into a separate draft PR, #2053**, because it alters the directory-enumeration hot path and its premise is not yet proven (see Repro below). #2053 keeps that change behind an off-by-default config flag and adds always-on Heartbeat telemetry to measure the leak first. This PR does not touch the enumeration hot path.

## Don't crash the mount on a native ProjFS failure

A single boundary helper, `TryInvokeProjFS(Func<HResult>, operationName, …)`, wraps each **GVFS-initiated** native ProjFS call: it catches the exception, emits high-signal telemetry (`*_NativeFailure`, with `activeEnumerations`/`activeCommands` counts and `GC.GetTotalMemory`), and returns `HResult.InternalError` so the caller maps it to a failure result and keeps serving.

Routed through the swallow-and-continue helper: `CompleteCommand` (via `TryCompleteCommand` — covers the completion leg of **all** async callbacks: enumeration, placeholder-info, file-data), `DeleteFile`, `WritePlaceholderFile`, `WritePlaceholderDirectory`, `UpdatePlaceholderIfNeeded`, `ClearNegativePathCache`.

**`MarkDirectoryAsPlaceholder`** mutates projection state, so swallowing is unsafe. It uses a sibling helper `InvokeProjFSOrThrow` that emits the **same** `*_NativeFailure` telemetry (so the memory diagnostics are captured) and then **rethrows** — the exception still reaches `LogUnhandledExceptionAndExit`, preserving fail-fast. Both helpers share one `LogProjFSNativeFailure`.

**Left as-is (already non-fatal):** `GetFileStreamHandlerAsyncHandler` already fails a single hydration on any exception (completes the command with `FileNotAvailable`, no exit), so `CreateWriteBuffer`/`WriteFileData` are already non-fatal.

> The exception is demonstrably catchable — it already reaches GVFS's `LogUnhandledExceptionAndExit` today (the source of the existing `VFS.Error "exiting process"` events). This change intercepts it one frame earlier and either declines to exit or (for state-mutating ops) enriches the crash with memory telemetry first.

## Local repro attempt (empirical) — why the enumeration-leak fix was split out

I built a variant of this change and, on top of it, tried to reproduce an enumeration leak → OOM → native crash: installed it, cloned a large enlistment, and hammered it with cancelled directory enumerations (`FindFirstFileExW` + `CancelSynchronousIo` mid-flight).

Controlled A/B, each touching 15k **cold** directories exactly once:

| Run | Aborted enums | Mount ΔWorkingSet |
|---|---|---|
| Control (no cancel) | 0 | **+10 MB** |
| Test (cancel) | 10,034 | **+7 MB** |

The cancel run grew *less* than the no-cancel control. A sustained 180 s / 869k-enumeration / 110k-abort run added only ~37 MB of slow creep dominated by legitimate cold→warm projection cache, stayed flat on idle, and produced **zero** `*_NativeFailure` / OOM / crash. Telemetry showed **883** async-path cancellations, **all** with `activeEnumerationsUpdated:true` — i.e. the existing async-cancel cleanup (`StartDirectoryEnumerationAsyncHandler` → `TryCompleteCommand` returns false → `TryRemove`) already removes every cancelled enumeration.

**Takeaways:**
- The simplest, most-suspected leak vector — enumeration cancellation — **does not** produce runaway growth in the current code; it self-heals. That is exactly why the `activeEnumerations` eviction was pulled out of this PR into **#2053** and gated off-by-default: it is unproven insurance, not the demonstrated driver, and it should not perturb the enumeration hot path in a stabilization build.
- The real-world OOM therefore comes from a different source (sync-path `Start`-without-`End`, abnormal process termination, or genuine kernel-memory exhaustion). That is exactly what **this PR's `*_NativeFailure` + `activeEnumerations`/memory telemetry is there to identify** once rolled out.

## Tests

- `DeleteFileReturnsIOErrorWhenVirtualizationInstanceThrows`, `OutboundProjFSOperationsReturnFailureWhenVirtualizationInstanceThrows` (ClearNegativePathCache, UpdatePlaceholderIfNeeded), `WritePlaceholderOperationsReturnFailureWhenVirtualizationInstanceThrows` (tester-based) — a throwing native call yields `IOError`, not a propagated exception.
- (`InvokeProjFSOrThrow`'s rethrow path ends in `Environment.Exit`, so it isn't safely unit-testable in-process; it's symmetric with the tested `TryInvokeProjFS`.)
- Full unit suite: **879 passed, 0 failed** (11 skipped); clean build.

## Open question for review

One design question remains genuinely open; I'd appreciate reviewer input.

**Swallow-and-continue vs. controlled remount.** When a native ProjFS call faults, this change logs `*_NativeFailure` and fails just that one operation, keeping the mount alive. That is strictly better than today's behavior (crash → orphaned root → `0xC000CE01` for everything). But if the native virtualization context is *genuinely* dead — not merely a transient allocation failure — then every subsequent operation will also fault, and the mount is effectively limping while still reporting itself as up. In that case a **deliberate, clean unmount + remount** on repeated `*_NativeFailure` (e.g. N failures within a window) might be better recovery than indefinitely serving errors. Open points:
- Is limping-but-alive the right default, or should repeated native failures escalate to a controlled remount / self-restart of the mount process?
- If we do escalate, what signal distinguishes "transient" from "context is dead"? A simple failure-rate threshold, or something more specific?
- I could **not settle this locally**: the native OOM fault was not reproducible via cancelled enumerations (see Repro — cancellation is self-healed), so I have no local scenario that reaches the dead-context state to test remount-vs-limp against. This needs either a heavier memory-pressure harness or production data.

## Summary / conclusion

The core value of this PR is **turning an opaque, fatal crash into a survivable, observable event.** Today the native ProjFS fault kills GVFS.Mount and all we get is a `radar_high_memory` Watson bucket with no GVFS-side context. After this change the mount stays alive, and every native failure emits high-signal telemetry (`activeEnumerations` / `activeCommands` counts and `GC.GetTotalMemory` at the exact failure point).

That telemetry is precisely what will let us **answer the open question above with real data instead of guesses**, and — together with the Heartbeat `ActiveEnumerationCount` added in **#2053** — tell us whether the enumeration-leak eviction is ever even needed in the field. The local repro already narrowed the search: enumeration cancellation is self-healed and is *not* the OOM driver, so the real memory source lies elsewhere. Land this to stop the bleeding and start collecting the data needed to fix the root cause.

---

*Assisted-by: Claude Opus 4.8*

### DIFF
--- a/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
+++ b/GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
@@ -119,14 +119,21 @@ namespace GVFS.Platform.Windows
 
         public override FileSystemResult ClearNegativePathCache(out uint totalEntryCount)
         {
-            HResult result = this.virtualizationInstance.ClearNegativePathCache(out totalEntryCount);
+            uint entryCount = 0;
+            HResult result = this.TryInvokeProjFS(
+                () => this.virtualizationInstance.ClearNegativePathCache(out entryCount),
+                nameof(this.ClearNegativePathCache));
+            totalEntryCount = entryCount;
             return new FileSystemResult(HResultToFSResult(result), unchecked((int)result));
         }
 
         public override FileSystemResult DeleteFile(string relativePath, UpdatePlaceholderType updateFlags, out UpdateFailureReason failureReason)
         {
             UpdateFailureCause failureCause = UpdateFailureCause.NoFailure;
-            HResult result = this.virtualizationInstance.DeleteFile(relativePath, (UpdateType)updateFlags, out failureCause);
+            HResult result = this.TryInvokeProjFS(
+                () => this.virtualizationInstance.DeleteFile(relativePath, (UpdateType)updateFlags, out failureCause),
+                nameof(this.DeleteFile),
+                relativePath);
             failureReason = (UpdateFailureReason)failureCause;
             return new FileSystemResult(HResultToFSResult(result), unchecked((int)result));
         }
@@ -137,17 +144,20 @@ namespace GVFS.Platform.Windows
             string sha)
         {
             FileProperties properties = this.FileSystemCallbacks.GetLogsHeadFileProperties();
-            HResult result = this.virtualizationInstance.WritePlaceholderInfo(
-                relativePath,
-                properties.CreationTimeUTC,
-                properties.LastAccessTimeUTC,
-                properties.LastWriteTimeUTC,
-                changeTime: properties.LastWriteTimeUTC,
-                fileAttributes: FileAttributes.Archive,
-                endOfFile: endOfFile,
-                isDirectory: false,
-                contentId: FileSystemVirtualizer.ConvertShaToContentId(sha),
-                providerId: PlaceholderVersionId);
+            HResult result = this.TryInvokeProjFS(
+                () => this.virtualizationInstance.WritePlaceholderInfo(
+                    relativePath,
+                    properties.CreationTimeUTC,
+                    properties.LastAccessTimeUTC,
+                    properties.LastWriteTimeUTC,
+                    changeTime: properties.LastWriteTimeUTC,
+                    fileAttributes: FileAttributes.Archive,
+                    endOfFile: endOfFile,
+                    isDirectory: false,
+                    contentId: FileSystemVirtualizer.ConvertShaToContentId(sha),
+                    providerId: PlaceholderVersionId),
+                nameof(this.WritePlaceholderFile),
+                relativePath);
 
             return new FileSystemResult(HResultToFSResult(result), unchecked((int)result));
         }
@@ -155,17 +165,20 @@ namespace GVFS.Platform.Windows
         public override FileSystemResult WritePlaceholderDirectory(string relativePath)
         {
             FileProperties properties = this.FileSystemCallbacks.GetLogsHeadFileProperties();
-            HResult result = this.virtualizationInstance.WritePlaceholderInfo(
-                relativePath,
-                properties.CreationTimeUTC,
-                properties.LastAccessTimeUTC,
-                properties.LastWriteTimeUTC,
-                changeTime: properties.LastWriteTimeUTC,
-                fileAttributes: FileAttributes.Directory,
-                endOfFile: 0,
-                isDirectory: true,
-                contentId: FolderContentId,
-                providerId: PlaceholderVersionId);
+            HResult result = this.TryInvokeProjFS(
+                () => this.virtualizationInstance.WritePlaceholderInfo(
+                    relativePath,
+                    properties.CreationTimeUTC,
+                    properties.LastAccessTimeUTC,
+                    properties.LastWriteTimeUTC,
+                    changeTime: properties.LastWriteTimeUTC,
+                    fileAttributes: FileAttributes.Directory,
+                    endOfFile: 0,
+                    isDirectory: true,
+                    contentId: FolderContentId,
+                    providerId: PlaceholderVersionId),
+                nameof(this.WritePlaceholderDirectory),
+                relativePath);
 
             return new FileSystemResult(HResultToFSResult(result), unchecked((int)result));
         }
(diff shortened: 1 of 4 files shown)


ShiningMassXAcc (approved):

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -652,13 +668,89 @@ private bool TryCompleteCommand(int commandId, HResult result)
...
+                this.TryInvokeProjFS(
+                    () => this.virtualizationInstance.CompleteCommand(commandId, result),
+                    nameof(this.TryCompleteCommand),
+                    addDetails: metadata =>
+                    {
+                        metadata.Add("commandId", commandId);
+                        metadata.Add("completionResult", result.ToString("X") + "(" + result.ToString("G") + ")");
+                    });
+
                 return true;
```
KeithIsSleeping: If CompleteCommand throws after an enumeration was registered, this returns true and skips the existing cleanup path.

tyrielv: Good catch — this was a real leak. Previously `TryCompleteCommand` always returned `true` whenever the command was still in `activeCommands`, even though `TryInvokeProjFS` swallows a native `CompleteCommand` fault and returns `InternalError`. When that fault happened after the enumeration was registered, ProjFS never accepted the completion (so no `EndDirectoryEnumeration` callback would arrive), but the caller saw `true` and skipped the cleanup path — leaking the `ActiveEnumeration`.

Fixed in the latest push: `TryCompleteCommand` now returns `true` only if the native call actually ran without throwing (`completionHandedOffToProjFS`). On a swallowed fault it returns `false`, so `StartDirectoryEnumerationAsyncHandler` runs its existing cleanup and removes the enumeration. Also generalized the trace/comment there (`_CommandAlreadyCanceled` → `_CommandNotCompleted`) since that branch now covers both cancellation and native-completion failure.

Added a regression test (`OnStartDirectoryEnumerationCleansUpEnumerationWhenNativeCompletionFails`) that forces a native completion fault and verifies the enumeration is removed rather than leaked.

KeithIsSleeping (approved):

PR #2052, opened 2026-07-09
Auto-recover corrupt packfiles in packfile maintenance

tyrielv:
> Independent change, based on `master`. Related to #2048 (both stabilize the same corrupt-packfile scenario) but has no code dependency on it. For maximum robustness, #2048 (which bounds git output) ideally lands first so recovery can't OOM on a very noisy stderr mid-recovery, but the two can merge in either order.

## Summary

When a packfile in the shared object cache is corrupt/truncated (e.g. from a past disk-full event), `git multi-pack-index write/verify` fails with `could not load pack N`. The existing self-heal only deletes+rewrites the multi-pack-index (MIDX), which **does not fix the underlying pack** — the rewrite re-scans the same bad pack and keeps failing, so the corruption recurs indefinitely.

## Change

`PackfileMaintenanceStep` routes write/verify failures through a recovery path. When git reports a pack-load failure:

- **Detection (always runs, even with recovery disabled):** verifies each pack with `git verify-pack` and reports every unreadable pack via telemetry (`Operation=FoundCorruptPack`). The `N` in `could not load pack N` is an internal MIDX position, not a filename, so per-pack verification is how we find the actual bad file.
- **Removal (gated — see kill switch):** deletes each corrupt pack's files (`.pack/.idx/.keep/.rev`; `Operation=DeletedCorruptPack`), then deletes+regenerates the MIDX from the packs that remain (fast path, **no full repack**). Missing objects are re-fetched on demand.
- **Corrupt prefetch pack (special case):** prefetch packs are incremental/ordered by timestamp, so a corrupt one invalidates every **later** prefetch pack too. Recovery removes the corrupt prefetch pack **and every later prefetch pack** (`Operation=DeletedHealthyPrefetchPack` for the healthy ones removed purely due to ordering), then **requests a prefetch** (via a callback `GitMaintenanceScheduler` wires to a `PrefetchStep`, only when using a cache server) to re-download them and rebuild the commit-graph.

## Kill switch

The destructive pack removal is gated by a new git config **`gvfs.enable-packfile-recovery` (default `true`)**. When `false`, GVFS still **detects and reports** corrupt packs (`FoundCorruptPack` → `CorruptPackRecoverySkipped`) but deletes nothing and requests no prefetch; the non-destructive MIDX rewrite still runs, so behavior degrades to today's. This is a field kill switch — no redeploy needed if the destructive path ever misbehaves.

## Telemetry

Distinct, queryable `Operation` values so rollout can be watched: `FoundCorruptPack` (even when disabled), `CorruptPackRecoverySkipped`, `DeletedCorruptPack`, `DeletedHealthyPrefetchPack`, `PrefetchRestoreUnavailable`.

## Tests

- A verify failure reporting a pack-load error removes the corrupt pack and rewrites the MIDX from the remaining good packs (recovery enabled).
- With recovery **disabled**, the same failure still verifies each pack and reports the corrupt one but **deletes nothing**.
- A corrupt prefetch pack removes it **and** every later prefetch pack, keeps the earlier healthy one, and requests a prefetch.

Full unit suite: **885 passed, 0 failed** (11 pre-existing native-hook skips). The pre-existing `PackfileMaintenanceRewriteOnBadVerify` (empty-stderr verify failure → MIDX rewrite only, no pack detection) is unchanged.

## Follow-up

A functional test that corrupts a real pack and asserts the mount self-heals (rather than looping) is tracked for the streaming follow-up PR.

### DIFF
--- a/GVFS/GVFS.Common/Git/GitProcess.cs
+++ b/GVFS/GVFS.Common/Git/GitProcess.cs
@@ -784,6 +784,17 @@ namespace GVFS.Common.Git
             return this.InvokeGitAgainstDotGitFolder("-c core.multiPackIndex=true multi-pack-index verify --object-dir=\"" + objectDir + "\" --no-progress");
         }
 
+        /// <summary>
+        /// Verifies the integrity of a single packfile via its .idx. Returns a failure exit code if the
+        /// pack is truncated or otherwise unreadable. Used by pack maintenance recovery to determine
+        /// which pack is corrupt - the "could not load pack N" ordinal reported by the multi-pack-index
+        /// is an internal position, not a filename, so it cannot be mapped to a file directly.
+        /// </summary>
+        public Result VerifyPack(string packIndexPath)
+        {
+            return this.InvokeGitAgainstDotGitFolder("verify-pack \"" + packIndexPath + "\"");
+        }
+
         public Result RemoteAdd(string remoteName, string url)
         {
             return this.InvokeGitAgainstDotGitFolder("remote add " + remoteName + " " + url);
--- a/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
+++ b/GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace GVFS.Common.Maintenance
 {
@@ -31,17 +32,26 @@ namespace GVFS.Common.Maintenance
         private const string MultiPackIndexLock = "multi-pack-index.lock";
         private readonly bool forceRun;
         private readonly string batchSize;
+        private readonly Action requestPrefetch;
+
+        // Set once corrupt packs have been detected and reported with recovery disabled. Recovery leaves
+        // the corrupt packs in place, so 'git multi-pack-index write/verify' keeps failing on them for
+        // the rest of this maintenance run - once reported, skip re-verifying every pack on each
+        // subsequent failure in the same run rather than repeating an identical, already-known result.
+        private bool reportedCorruptPacksWithRecoveryDisabled;
 
         public PackfileMaintenanceStep(
             GVFSContext context,
             bool requireObjectCacheLock = true,
             bool forceRun = false,
             string batchSize = DefaultBatchSize,
-            GitProcessChecker gitProcessChecker = null)
+            GitProcessChecker gitProcessChecker = null,
+            Action requestPrefetch = null)
             : base(context, requireObjectCacheLock, gitProcessChecker)
         {
             this.forceRun = forceRun;
             this.batchSize = batchSize;
+            this.requestPrefetch = requestPrefetch;
         }
 
         public override string Area => nameof(PackfileMaintenanceStep);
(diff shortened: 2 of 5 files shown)


GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -171,5 +187,276 @@ protected override void PerformMaintenance()
...
+                return;
+            }
+
+            // Phase 2 - deletion (gated). Build the set of prefetch packs to remove: the corrupt one and
+            // every later (>= timestamp) prefetch pack, whether or not those later packs are themselves
+            // corrupt, because prefetch packs are incremental.
+            List<string> laterPrefetchIdxPaths = new List<string>();
+            if (minCorruptPrefetchTimestamp.HasValue)
+            {
+                prefetchRestoreNeeded = true;
```
KeithIsSleeping: Could we set this only after at least one prefetch pack set is actually removed? If deletion is blocked, the restore prefetch may run while the corrupt pack is still present.

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
```diff
@@ -171,5 +187,276 @@ protected override void PerformMaintenance()
...
+        /// pack, a packfile itself is corrupt (e.g. truncated by a past disk-full event) and regenerating
+        /// the multi-pack-index (MIDX) alone keeps failing because the rewrite re-scans the same bad pack.
+        /// Detect the corrupt pack(s) - and, when recovery is enabled, remove them - then delete and
+        /// regenerate the MIDX from the packs that remain (fast path, no full repack).
+        /// </summary>
+        private void RepairMultiPackIndex(ITracer activity, GitProcess.Result failure, bool recoveryEnabled)
+        {
+            bool prefetchRestoreNeeded = false;
+
+            if (!this.Stopping && ResultIndicatesCorruptPack(failure))
```
KeithIsSleeping, line 260: With recovery disabled, the corrupt pack remains, so this path can verify every pack again on later MIDX failures. Could we avoid repeating the scan after the first report in a maintenance run?

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
tyrielv: Fixed

GVFS/GVFS.Common/Maintenance/PackfileMaintenanceStep.cs
tyrielv: Fixed - saved on an instance field (each instance of the object is only used for one maintenance run, but the instance field prevents rechecking on retries)

GVFS/GVFS.Common/Git/GitProcess.cs
```diff
@@ -806,6 +784,17 @@ public Result VerifyMultiPackIndex(string objectDir)
...
 
+        /// <summary>
+        /// Verifies the integrity of a single packfile via its .idx. Returns a failure exit code if the
+        /// pack is truncated or otherwise unreadable. Used by pack maintenance recovery to determine
+        /// which pack is corrupt - the "could not load pack N" ordinal reported by the multi-pack-index
+        /// is an internal position, not a filename, so it cannot be mapped to a file directly.
+        /// </summary>
+        public Result VerifyPack(string packIndexPath)
+        {
+            return this.InvokeGitAgainstDotGitFolder("verify-pack \"" + packIndexPath + "\"");
```
derrickstolee, line 795: I guess this is the most robust to interesting issues, but could we do a faster check that the trailing hash matches its contents up to that point? Perhaps `git verify-pack` fails fast in that case?

derrickstolee (approved):

GVFS/GVFS.Common/Git/GitProcess.cs
tyrielv: Good instinct — went and checked this properly. The write-path's ordinal ("could not load pack N") genuinely can't be resolved to a filename (confirmed in git's midx-write.c: the die() message never includes the name, and reconstructing it would mean depending on git's internal pack-enumeration order, which isn't a stable public contract).

But the actual detection path we hit here is multi-pack-index verify, not write - and verify's error text does already name the specific pack directly (e.g. "packfile pack-1234.pack does not match index", "wrong index v2 file size in pack-1234.idx"). We were throwing that away and re-verifying every pack in the cache unconditionally. Pushed a fix that parses the pack path straight out of the failure text and verifies only that candidate, falling back to the full scan if nothing can be parsed or the named pack turns out healthy. Thanks for pushing on this!

KeithIsSleeping (approved):

PR #2058, opened 2026-07-10
Split ServiceVerb repo-state check into intent-specific predicates

tyrielv:
## Problem

`gvfs service --unmount-all` intermittently fails with exit code 3 and `Already unmounting, please wait`, surfacing as the flaky functional test `GVFS.FunctionalTests.Tests.MultiEnlistmentTests.ServiceVerbTests.ServiceCommandsWithNoRepos`.

## Root cause

The service verbs (`--list-mounted`, `--mount-all`, `--unmount-all`) iterate the service's global repo registry and used a single `IsRepoMounted(repoRoot)` helper. That helper returned `true` whenever the mount named pipe merely **answered** a `GetStatus` request — regardless of the actual `MountState`. `StatusVerb` reports success as long as it can connect, so a `GVFS.Mount` process in the transient `Unmounting` state (a concurrent or prior unmount still shutting it down) counts as "mounted."

`--unmount-all` therefore attempts to unmount a repo that is already on its way out, gets the transient `Already unmounting` response from the mount process, and treats it as a hard failure — even though the desired end state (repo unmounted) is already being reached.

## Fix

Replace `IsRepoMounted` with three intent-specific predicates, all built on a `TryGetRepoMountStatus` primitive that reads the actual `MountState`:

| Predicate | Meaning | Caller |
|---|---|---|
| `IsRepoReady` | `MountState == Ready` | `--list-mounted` |
| `IsRepoAvailableToMount` | no live mount process answering | `--mount-all` |
| `IsRepoAvailableToUnmount` | `MountState ∈ {Ready, MountFailed}` | `--unmount-all` |

`--unmount-all` now skips repos that are `Unmounting` (already reaching the desired state) or `Mounting` (an unmount request would be rejected anyway), eliminating the spurious failure.

`StatusVerb` exposes the `"Mount status: "` output prefix as a shared constant so `ServiceVerb` recovers the `MountState` without re-implementing the pipe protocol (and without changing `StatusVerb`'s worktree/enlistment resolution, so there's no behavior regression for worktree mounts).

## Behavior changes

- **`--unmount-all`**: skips repos in `Unmounting`/`Mounting` instead of failing on them. (The fix.)
- **`--list-mounted`**: now reports only fully-`Ready` repos (previously also listed `Mounting`/`Unmounting`/`MountFailed` processes that happened to answer the pipe).
- **`--mount-all`**: unchanged — still mounts only when no live mount process is answering.

## Testing

- Builds clean (`dotnet build GVFS\GVFS\GVFS.csproj -c Debug`, 0 warnings).
- Covered by the existing `ServiceVerbTests` functional suite (`ServiceCommandsWithNoRepos`, `ServiceCommandsWithMultipleRepos`, `ServiceCommandsWithMountAndUnmount`), whose expectations remain consistent with the new state-aware behavior. There is no unit-test seam for the service verbs (they require a live service pipe).

## Related

This is the first of a set of fixes for the same flaky test. Two follow-ups are tracked separately: closing the mount-process lifecycle gap (a repo committed to unmounting still presents as a normal mounted repo to concurrent callers), and scoping the functional-test service registry so the global `--mount-all`/`--unmount-all` verbs can't reach across parallel fixtures.

### DIFF
--- a/GVFS/GVFS/CommandLine/ServiceVerb.cs
+++ b/GVFS/GVFS/CommandLine/ServiceVerb.cs
@@ -73,7 +73,7 @@ namespace GVFS.CommandLine
             {
                 foreach (string repoRoot in repoList)
                 {
-                    if (this.IsRepoMounted(repoRoot))
+                    if (this.IsRepoReady(repoRoot))
                     {
                         this.Output.WriteLine(repoRoot);
                     }
@@ -93,7 +93,7 @@ namespace GVFS.CommandLine
 
                 foreach (string repoRoot in repoList)
                 {
-                    if (!this.IsRepoMounted(repoRoot))
+                    if (this.IsRepoAvailableToMount(repoRoot))
                     {
                         this.Output.WriteLine("\r\nMounting repo at " + repoRoot);
                         ReturnCode result = this.Execute<MountVerb>(repoRoot);
@@ -118,7 +118,7 @@ namespace GVFS.CommandLine
 
                 foreach (string repoRoot in repoList)
                 {
-                    if (this.IsRepoMounted(repoRoot))
+                    if (this.IsRepoAvailableToUnmount(repoRoot))
                     {
                         this.Output.WriteLine("\r\nUnmounting repo at " + repoRoot);
                         ReturnCode result = this.Execute<UnmountVerb>(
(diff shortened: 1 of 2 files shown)


GVFS/GVFS/CommandLine/ServiceVerb.cs
```diff
@@ -73,7 +73,7 @@ public override void Execute()
             {
                 foreach (string repoRoot in repoList)
                 {
-                    if (this.IsRepoMounted(repoRoot))
+                    if (this.IsRepoReady(repoRoot))
```
KeithIsSleeping, line 76: The installer's PrepareToInstall uses --list-mounted output to decide HasMounts; with the Ready-only narrowing, a MountFailed or Mounting repo drops off the list.

tyrielv: That's probably fine. MountFailed state is very short and leads to mount process exit, and odds of installer trying to run while a repo is mounting are low enough that I think we can accept the UX (for interactive) - automated upgrades typically should use /StageIfMounted so they don't unmount existing processes anyway.

KeithIsSleeping (approved):

PR #2059, opened 2026-07-10
Scope ServiceVerbTests to a fixture-local GVFS service

tyrielv:
## Purpose
Tests for `gvfs service --mount-all` and `gvfs service --unmount-all` currently can be flaky or cause flakiness in other tests because the service is system-scope. This PR changes the service scope to be a single test fixture while running functional tests.

## Summary
- Run ServiceVerbTests against a dedicated service name per fixture instance instead of the shared test service
- Thread optional service-name overrides through GVFSHelpers, GVFSProcess, and GVFSFunctionalTestEnlistment so clone/mount/unmount/register/service verbs all hit the same fixture-local registry
- Add overloads in GVFSServiceProcess to install/start/stop/uninstall a named test service while keeping existing defaults intact

## Validation
- `dotnet build GVFS/GVFS.FunctionalTests/GVFS.FunctionalTests.csproj -c Debug`
- attempted targeted functional test run of `ServiceVerbTests.ServiceCommandsWithNoRepos` (blocked locally by existing ProgramData access denial while cleaning `Test.GVFS.Service` logs before tests start)

### DIFF
--- a/GVFS/GVFS.FunctionalTests/Tools/GVFSServiceProcess.cs
+++ b/GVFS/GVFS.FunctionalTests/Tools/GVFSServiceProcess.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.ServiceProcess;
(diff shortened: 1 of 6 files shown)


GVFS/GVFS.FunctionalTests/Tools/GVFSServiceProcess.cs
```diff
@@ -178,48 +204,48 @@ private static void InstallWindowsService()
 
             string createServiceArguments = string.Format(
```
KeithIsSleeping, line 205: The fixture service name is passed only when starting the Windows service, not in the installed command line. If SCM needs the service identity before OnStart receives start arguments, the named service path may fail outside dev mode.

tyrielv: Not a problem - this passed CI functional tests, and only other use case is dev mode.

KeithIsSleeping (approved):

PR #2065, opened 2026-07-15
Fix NullReferenceException when cloning into a non-empty directory

tyrielv:
## Summary

`gvfs clone <url> <target>` into a non-empty `<target>` directory throws an unhandled `NullReferenceException` instead of reporting the expected error:

```
Cannot clone @ <target>: System.NullReferenceException: Object reference not set to an instance of an object.
   at GVFS.CommandLine.CloneVerb.Execute() + 0x9fc
```

v1.0 behavior (expected):

```
Cannot clone @ <target>
Error: Clone directory '<target>' exists and is not empty
```

## Root cause

`CloneVerb.Execute()` unconditionally read `enlistment.WorkingDirectoryBackingRoot` to determine `trustPackIndexes`, even when `TryCreateEnlistment` failed (e.g. because the target directory already exists and is not empty). In that failure path `enlistment` is `null`, so the dereference throws.

## Fix

Extracted the lookup into `internal bool GetTrustPackIndexes(ITracer, Result, GVFSEnlistment)`, which only touches `enlistment` when `cloneResult.Success` is `true`; otherwise it returns the default without dereferencing `enlistment`. Delegates to the shared `LibGit2Repo.GetConfigBoolOrDefault` helper added in #2066.

## Testing

`GVFS.UnitTests.CommandLine.CloneVerbTests`:
- `TryCreateEnlistmentFailsWithoutEnlistmentWhenTargetDirectoryIsNotEmpty` — confirms the failure precondition (null enlistment on non-empty target dir).
- `TryCreateEnlistmentDoesNotFailForEmptyTargetDirectory` — boundary case: an empty target directory does not trigger the "exists and is not empty" error.
- `TryCreateEnlistmentReportsNormalizedPathWhenItDiffersFromFullPath` — covers the divergent full-vs-normalized-path error message branch.
- `GetTrustPackIndexesDoesNotThrowWhenCloneFailedAndEnlistmentIsNull` — the actual regression test: drives the exact failed-clone/null-enlistment composition that used to throw. Verified by temporarily removing the `cloneResult.Success` gate — the test failed with the original `NullReferenceException`, then passed again once the gate was restored.

## Review

Reviewed with an internal 6-lens review-swarm pass. The main finding: an earlier draft of the regression test only proved `TryCreateEnlistment`'s own contract (already true pre-fix) without exercising the actual `Execute()` null-dereference — addressed by the extraction and new test above.

## Validation

Full unit test suite: 895 passed, 0 failed (11 pre-existing skips, unrelated to this change).


GVFS/GVFS.UnitTests/Common/LibGit2RepoConfigLookupTests.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // The LibGit2Repo constructor logs a RelatedWarning with the native open-failure
+            // reason before throwing InvalidDataException; the static helper's catch does not
+            // log a second time for that case (see LibGit2Repo.GetConfigBoolOrDefault), so
+            // exactly one warning is expected here.
+            tracer.RelatedWarningEvents.Count.ShouldEqual(1);
+            tracer.RelatedWarningEvents[0].ShouldContain("Couldn't open repo at");
+        }
+
+        private class MockConfigRepo : LibGit2Repo
```
KeithIsSleeping, line 99: MockConfigRepo does not override Dispose(bool), so disposing it invokes native cleanup on a repo that never initialized libgit2.

tyrielv: Good catch --- fixed in the latest push. `MockConfigRepo` now overrides `Dispose(bool)` as a no-op, matching the pattern used by `LibGit2RepoInvokerTests.MockLibGit2Repo` and `LibGit2RepoSafeDirectoryTests.MockSafeDirectoryRepo`, since this mock never calls the native-initializing constructor and has no real handle to free.

ShiningMassXAcc:
Now that #2066 has merged to `master`, could you rebase this branch onto `master`? The diff still carries the #2066 `LibGit2Repo` config-lookup helper (the "Share transient libgit2 config lookup helper" commit), so a rebase should reduce this to the single fix commit described in the note. I'd like that clean single-commit diff before I give final approval.

ShiningMassXAcc (changes requested):
# Review — PR #2065: Fix NullReferenceException when cloning into a non-empty directory

**Recommendation: request changes** (not a blocker on correctness — the change is safe — but the PR
no longer does what it says, and one of the four new tests doesn't test anything).

## 1. Blocking: after the rebase, the NRE is already fixed on `master`

You rebased onto `master` as I asked, and that rebase removed the actual fix. This branch's
merge base is now `92020a4f`, and at that commit `CloneVerb.Execute()` already reads:

```csharp
if (!cloneResult.Success)
{
    tracer.RelatedError(cloneResult.ErrorMessage);
}
else
{
    trustPackIndexes = LibGit2Repo.GetConfigBoolOrDefault(
        tracer,
        enlistment.WorkingDirectoryBackingRoot,
        ...);
}
```

`enlistment` is only dereferenced when `cloneResult.Success` is `true`, so the crash described in
the PR body cannot happen on the base branch.

The NRE was real — in the pre-#2066 shape the config read sat *outside* the `if`:

```csharp
if (!cloneResult.Success)
{
    tracer.RelatedError(cloneResult.ErrorMessage);
}

using (var repo = new LibGit2RepoInvoker(tracer, enlistment.WorkingDirectoryBackingRoot))  // <- NRE
```

but `6ba0bb2a` ("Share transient libgit2 config lookup helper", #2066, merged 2026-08-27)
introduced both the `else` gate and the `trustPackIndexes = TrustPackIndexesDefault` initializer,
which is exactly the fix. So #2066 already shipped it.

What's left in this PR is a behaviour-preserving refactor plus tests. That's fine to take, but
please retitle/rewrite the description accordingly — and drop the "This gating is what fixes the
NullReferenceException regression" sentence from the `GetTrustPackIndexes` XML doc, which will
mislead the next reader into thinking the gate originated here.

## 2. Blocking: `TryCreateEnlistmentDoesNotFailForEmptyTargetDirectory` is vacuous

I instrumented the test and ran it. For the empty-directory case it observes:

```
Success=False, ErrorMessage=<Could not find git.exe.  Ensure that Git is installed.>, enlistment=null
```

`MockGitInstallation.GetInstalledGitBinPath()` returns `null`, so `TryCreateEnlistment` bails at the
`GitIsNotInstalledError` branch and never reaches enlistment creation. The test is named
"DoesNotFail" but the call *does* fail; the only assertion
(`DoesNotContain("exists and is not empty", ...)`) is trivially satisfied by any unrelated error, so
it would keep passing even if the empty-directory boundary broke completely.

It's also fragile in a non-obvious way: the only reason it doesn't crash is the early `null`
return. If `MockGitInstallation` ever returns a path, the next statement is `CheckGVFSHooksVersion`,
and `MockPlatform.TryGetGVFSHooksVersion` throws `NotSupportedException` (and past that,
`ReportErrorAndExit` would tear down the process).

Either assert the actual contract (`result.ErrorMessage.ShouldEqual(GVFSConstants.GitIsNotInstalledError)`,
i.e. "the empty directory was accepted and we moved past the emptiness check"), or drop the test.

## 3. The "regression test" doesn't guard the regression

`GetTrustPackIndexesDoesNotThrowWhenCloneFailedAndEnlistmentIsNull` exercises the extracted helper,
not the call site that used to crash. I verified this by mutation: I restored `Execute()` to the
exact pre-#2066 buggy form (unconditional `enlistment.WorkingDirectoryBackingRoot` deref) while
leaving `GetTrustPackIndexes` in place, rebuilt, and re-ran the fixture:

```
Overall result: Passed
Test Count: 4, Passed: 4, Failed: 0
```

So the test cannot fail for the bug it names. The PR body is candid that `Execute()` isn't unit
testable, which is true — but then the honest framing is "coverage for the new helper", not a
regression guard. Worth saying so in the comment above the test.

## 4. Non-blocking

- `GetTrustPackIndexes` only reads `cloneResult.Success`. Taking a `bool` instead of a `Result`
  would avoid widening the nested `Result` class from `private` to `internal` purely for test
  reachability.
- `GetTrustPackIndexes` doesn't touch instance state and could be `internal static`.
- `Setup`/`TearDown` create and delete a real temp directory per test. That's fine here, but
  `GVFS.UnitTests` is otherwise mock-based; a short comment on why real I/O is needed would help.

## Verification performed

- Built `GVFS.UnitTests` at `d487f91b` (x64 Debug, net10.0-windows, real vcpkg libgit2): 0 warnings,
  0 errors.
- Full unit suite: **986 tests, 975 passed, 0 failed, 11 skipped**.
- `CloneVerbTests` alone: 4/4 passed.
- Mutation A (restore buggy `Execute()`): 4/4 still passed — see §3.
- Probe of the empty-directory case: see §2.
- Confirmed `git log -S` that `6ba0bb2a` is the sole origin of the `else` gate and the
  `TrustPackIndexesDefault` initializer.

tyrielv:
Closing without merging. @ShiningMassXAcc's review caught that the actual fix for the NullReferenceException landed in #2066 (the `else` gate added around the `trustPackIndexes` config read in `6ba0bb2a`), which already merged to `master`. What remained here after rebasing onto `master` was a behavior-preserving refactor (extracting `GetTrustPackIndexes`) plus test coverage, two of which don't hold up under mutation testing as regression guards.

Thanks for the thorough review — no further action needed on this one.

PR #2066, opened 2026-07-15
Share transient libgit2 config lookup helper

tyrielv:
> [!NOTE]
> This PR is now a single commit (5471e467), squashed for a clean review/merge history.

## Summary
- add `LibGit2Repo.GetConfigBoolOrDefault(...)` (instance + static overloads) for one-off boolean config reads, replacing scattered short-lived `LibGit2Repo`/`LibGit2RepoInvoker` usage at 4 call sites
- avoid `LibGit2RepoInvoker`'s object-store preload for reads that only need a single config value and immediately dispose the repo
- cover happy-path, unset-key, and exception-fallback behavior with unit tests

## Motivation
`LibGit2RepoInvoker.InitializeSharedRepo()` intentionally forces an object-store probe so long-lived/shared callers can amortize object-store load costs. That is wasted work for one-off config reads that immediately dispose the repo. This change centralizes the short-lived config-read pattern behind two methods.

## Design
The helper methods live directly on `LibGit2Repo` rather than a separate extension class — this matches the `repo.GetConfigBoolOrDefault(name, default)` convention already documented in `AGENTS.md`, and avoids unnecessary indirection for a class the team owns in the same assembly:
- `public bool GetConfigBoolOrDefault(string key, bool defaultValue)` — instance method, reads from an already-open repo using `this.Tracer`.
- `public static bool GetConfigBoolOrDefault(ITracer tracer, string repoPath, string key, bool defaultValue)` — opens a transient `LibGit2Repo`, delegates to the instance method, and disposes it.

Both fall back to `defaultValue` and log a `RelatedWarning` on any failure (matching the "default on any failure" contract each call site previously implemented independently).

Also added a `protected LibGit2Repo(ITracer tracer)` constructor purely to support test doubles that need to inject a mock tracer without opening a real repo.

## Refactored call sites
- `GVFS/GVFS/CommandLine/CloneVerb.cs` — `gvfs.trust-pack-indexes` (only read when clone succeeded)
- `GVFS/GVFS.Hooks/Program.cs` — `gvfs.show-hydration-status` (runs on every git hook invocation)
- `GVFS/GVFS.Mount/InProcessMount.cs` — `gvfs.background-cache-auth`
- `GVFS/GVFS/CommandLine/PrefetchVerb.cs` — `gvfs.prefetch-offload`
- `GVFS/GVFS.Hooks/GVFS.Hooks.csproj` — updated to drop the now-removed extension-class source file

## Survey notes / intentionally left alone
Surveyed `master` for short-lived config-only `LibGit2Repo`/`LibGit2RepoInvoker` usage. These were left alone because they are shared/long-lived repo access, not the transient anti-pattern this PR addresses:
- `GVFS/GVFS.Common/Maintenance/PrefetchStep.cs`
- `GVFS/GVFS.Common/GitStatusCache.cs`
- `GVFS/GVFS.Common/Git/GitRepo.cs`

## Stacked PR
#2065 ("Fix NullReferenceException when cloning into a non-empty directory") is rebased on top of this branch and depends on the `LibGit2Repo.GetConfigBoolOrDefault` helper added here. **This PR should merge first.**

## Review
Reviewed with an internal 6-lens review-swarm pass (correctness, security, design, tests, async-parallelism, risk-rollout). All actionable findings were addressed:
- Widened the shared helper's exception handling back to a plain `catch (Exception)`, restoring the "default on any failure" guarantee `InProcessMount`/`PrefetchVerb` relied on before this refactor.
- Fixed a double-`RelatedWarning` log on the repo-open-failure path.
- Replaced a hardcoded, non-portable `Z:\...` path in a unit test with a GUID-suffixed temp path.
- Added missing test coverage for the unset-key (`null`-coalescing) branch and the `InvalidDataException` catch arm.
- Simplified the parameterless constructor to delegate to the tracer-accepting one.

## Validation
- `dotnet build GVFS\GVFS.UnitTests\GVFS.UnitTests.csproj -c Debug`
- `out\GVFS.UnitTests\bin\Debug\net10.0-windows10.0.17763.0\win-x64\GVFS.UnitTests.exe`
  - Result: 891 passed, 0 failed, 11 skipped (pre-existing, unrelated)


GVFS/GVFS.UnitTests/Common/LibGit2RepoConfigLookupTests.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            // The LibGit2Repo constructor logs a RelatedWarning with the native open-failure
+            // reason before throwing InvalidDataException; the static helper's catch does not
+            // log a second time for that case (see LibGit2Repo.GetConfigBoolOrDefault), so
+            // exactly one warning is expected here.
+            tracer.RelatedWarningEvents.Count.ShouldEqual(1);
+            tracer.RelatedWarningEvents[0].ShouldContain("Couldn't open repo at");
+        }
+
+        private class MockConfigRepo : LibGit2Repo
```
KeithIsSleeping, line 99: The new MockConfigRepo does not override Dispose(bool), so disposing it runs the base finalizable path that calls native libgit2 Native.Repo.Free/Native.Shutdown without a matching Init, unlike the other LibGit2Repo mocks which override Dispose to avoid the base native calls.

tyrielv: Good catch --- fixed in the latest push. `MockConfigRepo` now overrides `Dispose(bool)` as a no-op, matching the pattern used by `LibGit2RepoInvokerTests.MockLibGit2Repo` and `LibGit2RepoSafeDirectoryTests.MockSafeDirectoryRepo`, since this mock never calls the native-initializing constructor and has no real handle to free.

derrickstolee (approved):
Looks reasonable to me. Config lookups should be simple.

PR #2076, opened 2026-08-05
Split enumeration-miss cause + extract EnumerationFailureTracker + de-duplicate error telemetry

tyrielv:
## What

Extends the directory-enumeration cause tagging from #2071 with a finer split, de-duplicates the error
telemetry, and moves the whole classification/bookkeeping into a new `EnumerationFailureTracker` class.
**No behavior change** - metadata and log volume only; the returned `HResult` is unchanged.

## Why

"Directory enumeration failure" ("Failed to find active enumeration ID") is the largest named
regression signature in the field telemetry. #2071 tags the miss as `Evicted` vs `Unknown`. Analysis
of that `Unknown` bucket shows two distinct causes that need to be told apart, plus a per-machine event
storm that inflates the signal:

- The kernel (ProjFS) can deliver a `GetDirectoryEnumeration` that races the `EndDirectoryEnumeration`
  for the same handle - a query in flight while the directory handle is closing, or the querying
  process dying mid-enumeration. This is a benign close/query race, not an ID GVFS never held.
- A caller re-enumerating a handle whose Start GVFS lost (for example across a provider restart) can
  re-issue the same failing Get in a tight loop, producing a very large number of error events on one
  machine and drowning the machine-based signal.

## Changes

- **New `EnumerationFailureTracker` class.** Extracts the classification, once-per-ID de-duplication,
  and the bounded tracking maps into a cohesive, independently unit-testable class (mirroring the
  `MissingTreeTracker` pattern). It owns all three "why is this ID absent" maps - recently evicted,
  recently ended, and recently reported - and exposes `RecordEvicted`, `RecordEnded`, `ClassifyMiss`,
  and `TryReserveReport`. It keeps lock-free `ConcurrentDictionary` state; a coarse lock would
  serialize the hot enumeration path. The virtualizer keeps only the eviction *policy* (deciding what
  is stale) and delegates the bookkeeping.

- **Split the miss reason.** `EnumerationFailureReason` renames `Unknown` -> `NeverSeen` and adds
  `EndedRecently`:
  - `Evicted` - GVFS's own stale-enumeration eviction removed a live enumeration (self-inflicted).
  - `EndedRecently` - a Get raced or followed the End for the same enumeration (benign kernel
    close/query race; outside gvfs.exe's control).
  - `NeverSeen` - GVFS never held the ID: never started, or from before a provider restart (outside
    gvfs.exe's control).
  The virtualizer records an end (and an eviction) via the tracker **before** removing the ID from
  `activeEnumerations`, so a racing Get always finds it in one collection or the other. `ClassifyMiss`
  attributes `Evicted` (most actionable) first, then `EndedRecently`, then `NeverSeen`.

- **De-duplicate the error telemetry.** The tracker emits the full error once per enumeration ID within
  a window; repeats from a caller's retry loop are suppressed. The first occurrence still logs at Error,
  so the machine-based signal is preserved.

- **Bound the tracking maps.** The tracker prunes on a throttle from every record point, so no map can
  grow unbounded when one callback (e.g. End) stops arriving - the never-ended scenario this
  instrumentation targets.

## Testing

- `EnumerationFailureTrackerTests` + `WindowsFileSystemVirtualizerTests`: **44/44 passed**.
- Tracker unit tests cover classification, the `Evicted`-over-`EndedRecently` precedence, the eviction
  undo (lost-race), the dedup, and the prune/retention of each map. The virtualizer tests cover the
  end-to-end wiring, the record-before-remove ordering, and the precedence through the real callbacks.

### DIFF
--- a/GVFS/GVFS.Platform.Windows/EnumerationFailureTracker.cs
+++ b/GVFS/GVFS.Platform.Windows/EnumerationFailureTracker.cs
@@ -1,0 +1,230 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace GVFS.Platform.Windows
+{
+    /// <summary>
+    /// Why a directory-enumeration Get failed to find its enumeration ID. Recorded on the failure
+    /// telemetry so self-inflicted causes can be told apart from ProjFS races outside gvfs.exe's
+    /// control. These values are a case-sensitive contract consumed by the release-readiness
+    /// telemetry dashboard; keep them in sync with its cause bucketing.
+    /// </summary>
+    public enum EnumerationFailureReason
+    {
+        NeverSeen = 0,   // ProjFS delivered an ID GVFS never held: never started, or from before a provider restart (outside gvfs.exe's control).
+        Evicted,         // GVFS's own stale-enumeration eviction removed a live enumeration (self-inflicted).
+        EndedRecently,   // ProjFS delivered a Get racing or following the End for the same enumeration - a benign close/query race (outside gvfs.exe's control).
+    }
+
+    /// <summary>
+    /// Tracks the state needed to classify and rate-limit "Failed to find active enumeration ID"
+    /// failures, keyed by ProjFS enumeration GUID:
+    ///
+    /// - Recently ended IDs, so a Get that races or follows the End for the same enumeration is
+    ///   attributed to a benign close/query race (<see cref="EnumerationFailureReason.EndedRecently"/>)
+    ///   rather than an ID GVFS never held.
+    /// - Recently reported IDs, so the error is emitted once per ID within a window instead of once
+    ///   per retry when a caller re-enumerates a lost handle in a loop.
+    ///
+    /// Eviction (a separate concern owned by the virtualizer) is passed in to
+    /// <see cref="ClassifyMiss"/> as a flag rather than tracked here.
+    ///
+    /// Thread-safety: the enumeration callbacks run concurrently on many ProjFS worker threads, so
+    /// this deliberately uses lock-free <see cref="ConcurrentDictionary{Guid, Int64}"/> state rather
+    /// than a coarse lock, which would serialize the hot enumeration path. GUIDs are never reused, so
+    /// entries are bounded purely by age.
+    /// </summary>
+    public class EnumerationFailureTracker
+    {
+        // ProjFS can deliver a Get that races the End for the same handle (a query in flight while the
+        // directory handle is closing, or the querying process dying mid-enumeration). Ended IDs are
+        // retained this long so such a Get is attributed to a recently-ended enumeration.
+        private static readonly TimeSpan DefaultRecentlyEndedRetention = TimeSpan.FromSeconds(30);
+
+        // A Get miss for the same ID can repeat in a tight loop (a caller re-enumerating a handle
+        // whose Start GVFS lost, e.g. across a provider restart). The error is emitted once per ID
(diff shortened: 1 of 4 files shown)


GVFS/GVFS.Platform.Windows/EnumerationFailureTracker.cs
```diff
@@ -0,0 +1,212 @@
...
+{
+    /// <summary>
+    /// Why a directory-enumeration Get failed to find its enumeration ID. Recorded on the failure
+    /// telemetry so self-inflicted causes can be told apart from ProjFS races outside gvfs.exe's
+    /// control. These values are a case-sensitive contract consumed by the release-readiness
+    /// telemetry dashboard; keep them in sync with its cause bucketing.
+    /// </summary>
+    public enum EnumerationFailureReason
+    {
+        NeverSeen = 0,   // ProjFS delivered an ID GVFS never held: never started, or from before a provider restart (outside gvfs.exe's control).
```
KeithIsSleeping, line 16: Since the comment notes these strings are a case-sensitive contract for the telemetry dashboard, it might be worth confirming the KQL bucketing is updated in lockstep so the 'Unknown' rename does not silently drop the signal. Just flagging in case that dashboard change lives elsewhere.

GVFS/GVFS.Platform.Windows/WindowsFileSystemVirtualizer.cs
```diff
@@ -597,6 +588,15 @@ public HResult EndDirectoryEnumerationCallback(Guid enumerationId)
         {
             try
             {
+                // Record the end BEFORE removing from activeEnumerations so a GetDirectoryEnumeration
+                // that races this end - ProjFS can deliver an in-flight Get concurrently with the
+                // handle-close End for the same enumeration - is attributed to a recently-ended
+                // enumeration rather than an ID GVFS never held. RecordEnded also prunes the tracking
+                // maps on a throttle, so they stay bounded from this path.
+                this.enumerationFailureTracker.RecordEnded(enumerationId);
```
KeithIsSleeping, line 596: Could we consider undoing the ended marker when TryRemove fails and the ID is not known-evicted? That would keep the race-safe ordering while avoiding telemetry skew toward EndedRecently for unknown IDs.

tyrielv: Done — good call. In `EndDirectoryEnumerationCallback`, when `TryRemove` finds nothing and the ID wasn't evicted, I now undo the ended marker (`UndoEnded`), so a later Get for a never-held ID classifies `NeverSeen` instead of skewing to `EndedRecently`. The record-before-remove ordering is preserved for the normal successful-remove path, so a Get racing a real End is still race-safe; the evicted case is left alone since `Evicted` wins classification anyway.

Added `IsRecentlyEvicted`/`UndoEnded` to the tracker plus tests: tracker-level `UndoEnded` reclassification, and a virtualizer test asserting an End for a never-held ID doesn't skew a later Get. Pushed in the latest revision.

GVFS/GVFS.Platform.Windows/EnumerationFailureTracker.cs
tyrielv: Good flag, I'll make sure our dashboards are ready for this before deploying

ShiningMassXAcc (approved):
Approved. Clean no-behavior-change refactor: split enum-miss cause, extract EnumerationFailureTracker, de-dup error telemetry. Lock-free ConcurrentDictionary, tests green.

PR #2084, opened 2026-08-11
Update default Microsoft Git version to v2.55.0.vfs.0.8

dscho:
Update the default Microsoft Git version used by VFS for Git
to the newly promoted [`v2.55.0.vfs.0.8`](https://github.com/microsoft/git/releases/tag/v2.55.0.vfs.0.8) release.

PR #2094, opened 2026-08-18
Merge master into vnext (20260818)

tyrielv:
Routine forward-merge of `master` into `vnext` to keep the integration line
current after the 2.0.26229.1 (M279.3) prerelease. Merge commit only — no squash.

Clean merge (no conflicts). Validated locally: `GVFS.Common` builds clean and the
full unit suite passes (917 passed, 0 failed, 11 skipped).

**master commits brought into vnext:**

- Update default Microsoft Git version to v2.55.0.vfs.0.8 (#2084)
- Reject malformed object SHAs before the cache-server download and fail blob hydration cleanly on a malformed (NUL-byte) placeholder SHA (#2074)
- Record HTTP status code on blob-hydration failure telemetry (#2085)
- Read git config strings via snapshot and owned marshalling in LibGit2Repo, with a functional test for libgit2 config reads (#2086)
- Return DownloadAttemptResult from the malformed-SHA guard (#2091)

### DIFF
--- a/GVFS/GVFS.Common/Git/LibGit2Repo.cs
+++ b/GVFS/GVFS.Common/Git/LibGit2Repo.cs
@@ -259,18 +259,39 @@ namespace GVFS.Common.Git
             }
             try
             {
-                string value;
-                Native.ResultCode resultCode = Native.Config.GetString(out value, configHandle, name);
-                if (resultCode == Native.ResultCode.NotFound)
+                // git_config_get_string returns a borrowed pointer whose lifetime is tied to the
+                // config, so libgit2 only allows it on a snapshot (read-only) config. Calling it on
+                // the live config returned by git_repository_config fails with "get_string called on
+                // a live config object". Snapshot the config first, then read the string from it.
+                IntPtr snapshotHandle;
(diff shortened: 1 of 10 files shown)


GVFS/GVFS.Common/Git/LibGit2Repo.cs
```diff
@@ -259,18 +259,39 @@ public virtual string GetConfigString(string name)
...
+                    {
+                        return null;
+                    }
+                    else if (resultCode != Native.ResultCode.Success)
+                    {
+                        throw new LibGit2Exception($"Failed to get config value for '{name}': {Native.GetLastError()}");
+                    }
 
-                return value;
+                    return valuePtr == IntPtr.Zero ? null : Marshal.PtrToStringUTF8(valuePtr);
```
KeithIsSleeping, line 289: `Marshal.PtrToStringUTF8` is not available on .NET Framework 4.x, and nearby comments still mention 4.7.1, so this may not compile there.

tyrielv: This is an 'automation' PR - taking everything as-is from master to vnext. No changes will be accepted.

KeithIsSleeping (approved):

